Issue #144, opened 2020-08-17
Migrate controllers and factories to TypeScript and extract services


liangyuanruo:
Blocked by #22 

TODOs
- Migrate controllers/factories to TypeScript
- For each controller, extract relevant business logic into services
- Refactor implementations to use up-to-date patterns where possible e.g. from callbacks to async/await
- Add new Jest tests, or migrate existing tests from Jasmine to Jest

Controllers
- [x] admin-console.server.controller.js (@karrui)
- [ ] admin-forms.server.controller.js (@karrui)
- [x] authentication.server.controller.js (@liangyuanruo)
- [x] core.server.controller.js (@karrui)
- [x] email-submissions.server.controller.js (@mantariksh)
- [ ] encrypt-submissions.server.controller.js
- [x] errors.server.controller.js (@tshuli)
- [x] myinfo.server.controller.js (@mantariksh )
- [x] public-forms.server.controller.js (@karrui, 2/6 done. Blocked by spcp server controller refactor)
- [x] spcp.server.controller.js (@mantariksh)
- [x] submissions.server.controller.js (@mantariksh)
- [x] verification.server.controller.js (@mantariksh)
- [x] webhooks.server.controller.js (@mantariksh)
- [x] user.server.controller

Factories
- [x] aggregate-stats.factory.js (@karrui)
- [x] captcha.factory.js (@karrui)
- [ ] google-analytics.factory.js (@tshuli)
- [x] sms.factory.js (@karrui)
- [x] verified-fields.factory.js (@mantariksh)
- [ ] webhook-verified-content.factory.js

liangyuanruo:
Superseded by #1434, #1435, #1437, #1438.

Issue #948, opened 2020-12-28
Use `never` instead of `ParamsDictionary` in middleware


mantariksh:
`ParamsDictionary` from `express-serve-static-core` is used in nearly every middleware function as the first generic type argument to `RequestHandler`, but this exposes developers to the risk of assuming that a parameter is defined when it actually isn't, since `ParamsDictionary` is typed as `{ [key: string]: string }`. `never` should be used instead if the parameter is unused.

PR #1405, opened 2021-03-18
refactor: migrate frontend routes and google analytics factory to ts

tshuli:
## Problem
- Part of #144 
- Migrates frontend routes and google analytics factory to typescript

## Tests
- [ ] Create a form. Open. Submit. Check that event shows up correctly on google analytics
- [ ] Check that `/frontend/environment` endpoint displays the envvars
- [ ] Check that `/frontend/features` endpoint displays the list of features and their activation state


mantariksh:
actually since you're touching the `/frontend` code, maybe you can convert all the frontend routes and controller to TS? it's a tiny module anyway

src/app/controllers/frontend.server.controller.ts
```diff
@@ -1,16 +1,24 @@
...
- * Google Tag Manager initialisation Javascript code templated
- * with environment variables.
- * @returns {String} Templated Javascript code for the frontend
+ * Handler for GET /frontend/datalayer endpoint.
+ * @param req - Express request object
+ * @param res - Express response object
+ * @returns {String} Templated Javascript code for the frontend to initialise Google Tag Manager
  */
-module.exports.datalayer = function (req, res) {
+export const datalayer: RequestHandler<
```
mantariksh, line 18: might it be better to rename all the controllers in this file to be more informative, e.g. `addGoogleAnalyticsData`?

src/app/controllers/frontend.server.controller.ts
```diff
@@ -1,16 +1,24 @@
-'use strict'
+import ejs from 'ejs'
```
mantariksh, line 1: this file should be moved into a `frontend` module

src/app/routes/frontend.server.routes.ts
```diff
@@ -0,0 +1,52 @@
...
+export const FrontendRouter = Router()
+
+/**
+ * Generate the templated Javascript code for the frontend to initialise Google Tag Manager
+ * Code depends on whether googleAnalyticsFeature.isEnabled
+ * @route GET /frontend/datalayer
+ * @return 200 when code generation is successful
+ * @return 400 when code generation fails
+ */
+FrontendRouter.get('/frontend/datalayer', GoogleAnalyticsFactory.datalayer)
```
mantariksh, line 16: all these routes can have their `/frontend` prefix removed, and you can add the `/frontend` prefix in `src/loaders/express/index.ts`. this is so that we can establish a clear hierarchy of APIs

src/loaders/express/index.ts
```diff
@@ -147,6 +148,7 @@ const loadExpressApp = async (connection: Connection) => {
     routeFunction(app)
   })
 
+  app.use(FrontendRouter)
```
mantariksh, line 151: can remove the /frontend prefix in `frontend.routes` and add it here instead

src/app/routes/frontend.server.routes.ts
```diff
@@ -0,0 +1,52 @@
+import { celebrate, Joi, Segments } from 'celebrate'
```
mantariksh, line 1: this file should be moved into `src/app/modules/frontend/frontend.routes.ts`

mantariksh (changes requested):

src/app/controllers/frontend.server.controller.ts
tshuli: renamed

src/app/controllers/frontend.server.controller.ts
tshuli: shifted

src/app/routes/frontend.server.routes.ts
tshuli: shifted

src/app/routes/frontend.server.routes.ts
tshuli: shifted

tshuli:
@mantariksh thanks for reviewing, for re-review

src/app/factories/google-analytics.factory.ts
```diff
@@ -0,0 +1,42 @@
+import { RequestHandler } from 'express'
```
mantariksh, line 1: instead of artificially separating this from the frontend module, should we just move it into that module instead?

src/app/factories/__tests__/google-analytics.factory.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { StatusCodes } from 'http-status-codes'
```
mantariksh, line 1: same comment as on the factory - should we move these into the frontend module tests?

mantariksh (approved):
lgtm, with a final question on file placement

src/app/factories/__tests__/google-analytics.factory.spec.ts
tshuli: thanks, shifted accordingly

src/app/modules/frontend/frontend.server.controller.ts
```diff
@@ -97,3 +114,16 @@ module.exports.redirectLayer = function (req, res) {
...
 }
+
+/**
+ * Handler for GET /frontend/features endpoint.
+ * @param req - Express request object
+ * @param res - Express response object
+ * @returns {String} Current featureManager states
+ */
+export const showFeaturesStates: RequestHandler<
+  ParamsDictionary,
```
mantariksh, line 125: sorry one more thing, I realised that all these controllers still use the old method of supplying `ParamsDictionary` as the first generic type argument, which we don't want as per #948. could you help change all of these to `unknown` pls?

src/app/modules/frontend/google-analytics.factory.ts
```diff
@@ -0,0 +1,43 @@
...
+import FeatureManager, {
+  FeatureNames,
+  RegisteredFeature,
+} from '../../../config/feature-manager'
+
+import * as FrontendServerController from './frontend.server.controller'
+
+interface IGoogleAnalyticsFactory {
+  addGoogleAnalyticsData: RequestHandler<
+    ParamsDictionary,
```
mantariksh, line 14: `unknown` here as well

src/app/modules/frontend/frontend.server.controller.ts
tshuli: Ok edited; apart from `showFeaturesStates`, the rest require `ParamsDictionary` as the first generic type argument because they call `createReqMeta(req)` and req is of interface `Request` which requires `ParamsDictionary` as the first generic type argument

src/app/modules/frontend/google-analytics.factory.ts
tshuli: as above, ParamsDictionary is needed

PR #1424, opened 2021-03-22
build: automate dependency upgrades using mergify

mantariksh:
## Problem

This project has over 200 dependencies, resulting in around 5-10 dependency upgrades a day. All these upgrades are currently manually approved and merged by engineers, which requires time and effort.

## Solution

Use [mergify](https://mergify.io/) to merge dependabot PRs which pass tests, except major version changes.

This screenshot from the Mergify [config editor](https://dashboard.mergify.io/github/opengovsg/config-editor) demonstrates that PR #1417 (patch version change, CI passing) would be automatically merged using this config:
Screenshot 2021-03-22 at 5 03 46 PM

Whereas this screenshot shows that PR #1422 (patch version change, CI failing) would not be merged:
image

And this screenshot shows that PR #1423 (major version change, CI passing) would not be merged:
image

### DIFF
--- a/.github/.mergify.yml
+++ b/.github/.mergify.yml
@@ -3,6 +3,7 @@ pull_request_rules:
     conditions:
       - 'author=dependabot[bot]'
       - '#check-failure=0'
+      - 'title~=bump [^\s]+ from ([\d]+)\..+ to \1\.'
     actions:
       merge:
         method: squash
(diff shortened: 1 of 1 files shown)


liangyuanruo (commented):
Good initiative. Is there a way to automatically merge only when it is not a major version change? I sometimes find that it's prudent to take a look at the changelog when it could _potentially_ be breaking.

mantariksh:
>Is there a way to automatically merge only when it is not a major version change?

Great idea, added and updated the screenshots from the config editor

liangyuanruo (approved):

Issue #1355, opened 2021-03-11
Settings API refactor


syan-syan:
Related to #815, shard out for priority

PR #1425, opened 2021-03-22
feat: rename /admin/form endpoint to /admin/forms, correct dir structure of /api/v3 routes

karrui:
## Problem

API endpoint for recently merged #1355 should be pluralized to follow REST resource conventions when the resource is in a collection. Thus, the plural form of `form` should be used for the new `/api/v3/admin/forms` endpoint introduced.

This PR also cleans up the directory structure of /api/v3 routes, giving the correct nested express routers

> This should be merged before the next release to ensure backwards compatibility. If it is not merged before that... I guess we will need to live with the old-new endpoint....


liangyuanruo:
Btw can I also suggest separating the `ApiRouter` into two, for `/api` and `/v3`?

liangyuanruo (approved):

karrui:
> Btw can I also suggest separating the `ApiRouter` into two, for `/api` and `/v3`?

Noted. Will do it in this PR

src/app/routes/admin/form/__tests__/admin-form.routes.spec.ts
```diff
@@ -15,7 +15,7 @@ import { AdminFormRouter } from '../admin-form.routes'
 
 const UserModel = getUserModel(mongoose)
 
-const app = setupApp('/admin/form', AdminFormRouter, {
+const app = setupApp('/admin/forms', AdminFormRouter, {
```
liangyuanruo, line 18: change the folder name as well?

karrui:
Updated directory structure and split `ApiRouter` to `ApiRouter` and `V3Router`

liangyuanruo (approved):

PR #1440, opened 2021-03-23
refactor: use neverthrow in verification module

mantariksh:
**Not to be merged until #1443 goes into production.**

## Problem

1. Most of the backend is written using `neverthrow` to chain asynchronous operations, but the verification module is not.
2. The verification module handles errors by sending custom strings (e.g. `'SEND_OTP_FAILED'`, '`INVALID_OTP`') to the frontend, which then uses a switch statement to determine the error message to render. This is an inflexible approach as any change in the error message cannot be propagated to existing clients.

## Solution

1. Rewrite the verification module using `neverthrow`. All existing functionality should be preserved.
2. Send error messages from the backend and render them in the frontend (see `verifiable-field.client.component.js` for frontend changes).

## Other improvements
Test suite has been enhanced with factory and integration tests.

## Tests
**Email OTP verification**
- [ ] Request for OTP and enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 3 times, then enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 4 times, then enter correct OTP. Field should NOT be able to be verified as max retry count has been exceeded.
- [ ] Open network tab. Request for OTP, then within 30 seconds, replay the network request. You should get a 400 as the waiting time between OTP requests has not elapsed.
- [ ] Request for OTP then wait for 10 minutes, then enter the OTP. OTP should not be verified as it has expired.

**Mobile OTP verification**
- [ ] Request for OTP and enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 3 times, then enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 4 times, then enter correct OTP. Field should NOT be able to be verified as max retry count has been exceeded.
- [ ] Open network tab. Request for OTP, then within 30 seconds, replay the network request. You should get a 400 as the waiting time between OTP requests has not elapsed.
- [ ] Request for OTP then wait for 10 minutes, then enter the OTP. OTP should not be verified as it has expired.

PR #1443, opened 2021-03-24
refactor: check for .message in verification error

mantariksh:
## Problem

Across nearly all APIs in this repo, the server returns a JSON with a `message` key when an error occurs. However, the Verification module is an exception, as it returns one of a set of custom strings (e.g. `'SEND_OTP_FAILED'`, '`INVALID_OTP`'), and the frontend contains a switch statement to determine what error message to display.

We want APIs to behave consistently, so it would be good for the Verification module to return error objects in the form `{ message: 'some error message' }`. However, this would be a backwards-incompatible change as the current frontend would render the entire JSON object instead of just the message string. 

## Solution

This PR ensures that the API change will be backwards-compatible by checking for a `message` property on the error. #1440 will address the actual API change.

### DIFF
--- a/src/public/modules/forms/base/components/verifiable-field.client.component.js
+++ b/src/public/modules/forms/base/components/verifiable-field.client.component.js
@@ -244,6 +244,10 @@ function verifiableFieldController($q, $timeout, $interval) {
   }
 
   const getErrorMessage = (err) => {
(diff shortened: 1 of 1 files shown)


tshuli (approved):
lgtm!

karrui (approved):

PR #1450, opened 2021-03-24
refactor: move Verification files together (part 1)

mantariksh:
**Not to be merged until #1443 goes into production.**

## Problem

1. Most of the backend is written using `neverthrow` to chain asynchronous operations, but the verification module is not.
2. The verification module handles errors by sending custom strings (e.g. `'SEND_OTP_FAILED'`, '`INVALID_OTP`') to the frontend, which then uses a switch statement to determine the error message to render. This is an inflexible approach as any change in the error message cannot be propagated to existing clients.

## Solution

This is the first part of the `neverthrow` rewrite. It includes the following preliminary changes:
1. Moves the Verification model into the verification module
2. Moves the verification module tests into the module
3. Refactors the existing `verification.factory` to make it clear that it is exporting middleware
4. Adds integration tests
5. Adds Express route validation to validate the shape of transaction IDs

## Tests
**Email OTP verification**
- [ ] Request for OTP and enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 3 times, then enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 4 times, then enter correct OTP. Field should NOT be able to be verified as max retry count has been exceeded.
- [ ] Open network tab. Request for OTP, then within 30 seconds, replay the network request. You should get a 400 as the waiting time between OTP requests has not elapsed.
- [ ] Request for OTP then wait for 10 minutes, then enter the OTP. OTP should not be verified as it has expired.

**Mobile OTP verification**
- [ ] Request for OTP and enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 3 times, then enter correct OTP. Field should be shown as verified.
- [ ] Request for OTP and enter incorrect OTP 4 times, then enter correct OTP. Field should NOT be able to be verified as max retry count has been exceeded.
- [ ] Open network tab. Request for OTP, then within 30 seconds, replay the network request. You should get a 400 as the waiting time between OTP requests has not elapsed.
- [ ] Request for OTP then wait for 10 minutes, then enter the OTP. OTP should not be verified as it has expired.

### DIFF
--- a/src/app/modules/verification/__tests__/verification.routes.spec.ts
+++ b/src/app/modules/verification/__tests__/verification.routes.spec.ts
@@ -1,0 +1,359 @@
+import bcrypt from 'bcrypt'
+import { StatusCodes } from 'http-status-codes'
+import mongoose from 'mongoose'
+import session, { Session } from 'supertest-session'
+
+import { BasicField, IVerificationSchema } from 'src/types'
+
+import { setupApp } from 'tests/integration/helpers/express-setup'
+import { buildCelebrateError } from 'tests/unit/backend/helpers/celebrate'
+import { generateDefaultField } from 'tests/unit/backend/helpers/generate-form-data'
+import dbHandler from 'tests/unit/backend/helpers/jest-db'
+
+import getVerificationModel from '../verification.model'
+import { VfnRouter } from '../verification.routes'
+
+import {
+  generateFieldParams,
+  MOCK_HASHED_OTP,
+  MOCK_OTP,
+  MOCK_RECIPIENT,
+  MOCK_SIGNED_DATA,
+} from './verification.test.helpers'
+
+const verificationApp = setupApp('/transaction', VfnRouter)
+const VerificationModel = getVerificationModel(mongoose)
+
+jest.mock('nodemailer', () => ({
+  createTransport: jest.fn().mockReturnValue({
+    sendMail: jest.fn().mockResolvedValue(true),
+  }),
+}))
+// Default Twilio export is a function
+jest.mock('twilio', () => () => ({
+  messages: {
+    create: jest.fn().mockResolvedValue({
+      sid: 'mockSid',
+    }),
+  },
+}))
+
+describe('verification.routes', () => {
+  let mockTransaction: IVerificationSchema
+  let mockTransactionId: string
+  let mockEmptyFormId: string
+  let mockVerifiableFormId: string
+  let mockEmailFieldId: string
+  let mockMobileFieldId: string
+  let request: Session
+
+  beforeAll(async () => await dbHandler.connect())
+
+  beforeEach(async () => {
+    request = session(verificationApp)
+    jest.clearAllMocks()
+    await dbHandler.clearDatabase()
+
+    // Form without verifiable fields
+    const { form: emptyForm } = await dbHandler.insertEmailForm()
+    mockEmptyFormId = String(emptyForm._id)
+
+    // Form with verifiable fields
+    const emailField = generateDefaultField(BasicField.Email, {
+      isVerifiable: true,
+    })
+    const mobileField = generateDefaultField(BasicField.Mobile, {
+      isVerifiable: true,
+    })
+    mockEmailFieldId = String(emailField._id)
+    mockMobileFieldId = String(mobileField._id)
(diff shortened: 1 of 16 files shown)


src/app/modules/verification/__tests__/verification.routes.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+      const response = await request.post(ROUTE)
+
+      expect(response.status).toBe(StatusCodes.BAD_REQUEST)
+      expect(response.body).toEqual(
+        buildCelebrateError({ body: { key: 'formId' } }),
+      )
+    })
+
+    it('should return 400 when formId is malformed', async () => {
+      const response = await request.post(ROUTE).send({ formId: 'malformed' })
```
karrui, line 123: ...

src/app/modules/verification/__tests__/verification.routes.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+        }),
+      )
+    })
+
+    it('should return 200 when transactionId and fieldId for email field are valid', async () => {
+      const response = await request
+        .post(`/transaction/${mockTransactionId}/reset`)
+        .send({ fieldId: mockEmailFieldId })
+
+      expect(response.status).toBe(StatusCodes.OK)
```
karrui, line 206: do you think we should verify the shape of the body also? can use `http-status-code`'s `getReasonPhrase`

karrui (approved):
lgtm

src/app/modules/verification/__tests__/verification.routes.spec.ts
mantariksh: oh this wasn't meant to be a joke but I guess it's still funny

src/app/modules/verification/__tests__/verification.routes.spec.ts
mantariksh: sure

src/app/modules/verification/__tests__/verification.routes.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    // Mock transaction with both email and mobile fields
+    mockTransaction = await VerificationModel.create({
+      formId: mockVerifiableFormId,
+      fields: [
+        generateFieldParams({
+          _id: mockEmailFieldId,
+          // Hash created 1 minute ago, so we can test requesting for new OTP
+          // without being rejected due to minimum waiting time
+          hashCreatedAt: new Date(Date.now() - 60 * 1000),
```
seaerchin, line 87: should we consider adding `luxon` so that this is clearer? because as it is, 60 * 1000 is abit cryptic and requires prior knowledge that date arithmetic is in milliseconds

src/app/modules/verification/__tests__/verification.routes.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+  messages: {
+    create: jest.fn().mockResolvedValue({
+      sid: 'mockSid',
+    }),
+  },
+}))
+
+describe('verification.routes', () => {
+  let mockTransaction: IVerificationSchema
+  let mockTransactionId: string
```
seaerchin, line 43: why use `let` over declaring each one as `const` and initializing in respective blocks? `let` implies that we are mutating the variables, which should not happen if this is a test? (not sure about this) this is additionally complicated by the fact that within each test block, if we modify the variables, the change will be visible throughout the `describe` block, which might complicate testing later on

src/app/modules/verification/__tests__/verification.routes.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+    mockEmptyFormId = String(emptyForm._id)
+
+    // Form with verifiable fields
+    const emailField = generateDefaultField(BasicField.Email, {
+      isVerifiable: true,
+    })
+    const mobileField = generateDefaultField(BasicField.Mobile, {
+      isVerifiable: true,
+    })
+    mockEmailFieldId = String(emailField._id)
```
seaerchin, line 68: from above comment, can't we shift this into each individual test block and let them instantiate

PR #1451, opened 2021-03-24
refactor: use neverthrow in /transaction/:transactionId (part 2)

mantariksh:
Builds on #1450

## Problem

See #1450

## Solution

This is the 2nd part of the solution, which focuses on the `GET /transaction/:transactionId` endpoint. It has the following additional changes:
- Renames the old `verification.factory` to `verification.middleware`, and adds an actual `verification.factory` to export service methods

## Tests
See #1450

### DIFF
--- a/src/app/modules/verification/__tests__/verification.controller.new.spec.ts
+++ b/src/app/modules/verification/__tests__/verification.controller.new.spec.ts
@@ -1,0 +1,121 @@
+import { ObjectId } from 'bson'
+import { Response } from 'express'
+import { StatusCodes } from 'http-status-codes'
+import mongoose from 'mongoose'
+import { errAsync, okAsync } from 'neverthrow'
+import { mocked } from 'ts-jest/utils'
+
+import { IVerificationSchema } from 'src/types'
+
+import dbHandler from 'tests/unit/backend/helpers/jest-db'
+
+import expressHandler from '../../../../../tests/unit/backend/helpers/jest-express'
+import { DatabaseError } from '../../core/core.errors'
+import * as VerificationController from '../verification.controller'
+import { TransactionNotFoundError } from '../verification.errors'
+import { VerificationFactory } from '../verification.factory'
+import getVerificationModel from '../verification.model'
+
+const VerificationModel = getVerificationModel(mongoose)
+
+jest.mock('../verification.factory')
+const MockVerificationFactory = mocked(VerificationFactory, true)
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {}
+const MOCK_FORM_ID = new ObjectId().toHexString()
+const MOCK_TRANSACTION_ID = new ObjectId().toHexString()
+
+describe('Verification controller', () => {
+  let mockTransaction: IVerificationSchema
+  let mockRes: Response
+
+  beforeAll(async () => {
+    await dbHandler.connect()
+    mockTransaction = await VerificationModel.create({
+      _id: MOCK_TRANSACTION_ID,
+      formId: MOCK_FORM_ID,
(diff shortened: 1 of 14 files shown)


src/app/modules/verification/__tests__/verification.controller.new.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  beforeEach(() => {
+    mockRes = expressHandler.mockResponse()
+  })
+
+  afterEach(() => jest.resetAllMocks())
+
+  afterAll(async () => {
+    // mockTransaction is reused throughout the tests
+    await dbHandler.clearDatabase()
```
seaerchin, line 51: in the previous PR, `clearDatabase` is done in the `afterEach` block but here, it is done `afterAll` - would this create bugs where previous transactions interfere with tests and create flaky tests?

src/app/modules/verification/__tests__/verification.controller.new.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        mockRes,
+        noop,
+      )
+
+      expect(
+        MockVerificationFactory.getTransactionMetadata,
+      ).toHaveBeenCalledWith(MOCK_TRANSACTION_ID)
+      expect(mockRes.status).toHaveBeenCalledWith(StatusCodes.NOT_FOUND)
+      expect(mockRes.json).toHaveBeenCalledWith({
+        message: expect.any(String),
```
seaerchin, line 95: @karrui should this enumerate over all possible 404s (from `mapRouteError`)?

src/app/modules/verification/verification.util.ts
```diff
@@ -0,0 +1,107 @@
...
+        errorMessage: coreErrorMsg,
+        statusCode: StatusCodes.BAD_REQUEST,
+      }
+    case FieldNotFoundInTransactionError:
+    case TransactionNotFoundError:
+      return {
+        errorMessage: coreErrorMsg,
+        statusCode: StatusCodes.NOT_FOUND,
+      }
+    case FormNotFoundError:
```
seaerchin, line 82: nit: combine with above

src/app/modules/verification/__tests__/verification.controller.new.spec.ts
mantariksh: as discussed, this particular controller doesn't return FormNotFoundError

src/app/modules/verification/__tests__/verification.controller.new.spec.ts
mantariksh: as commented

src/app/modules/verification/verification.middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+      createTransaction: verification.createTransaction,
+      resetFieldInTransaction: verification.resetFieldInTransaction,
+      getNewOtp: verification.getNewOtp,
+      verifyOtp: verification.verifyOtp,
+    }
+  } else {
+    const errMsg = 'Verified fields feature is not enabled'
+    return {
+      createTransaction: (req, res) =>
+        res.status(StatusCodes.INTERNAL_SERVER_ERROR).json({ message: errMsg }),
```
karrui, line 31: do we not use `MissingFeatureError` anymore?

karrui (approved):
el gee tee em

src/app/modules/verification/verification.middleware.ts
mantariksh: we do! it's just that these are the middleware functions which haven't been migrated yet. every subsequent PR in this series will remove one of these functions and replace it with one in `verification.controller.ts`

PR #1452, opened 2021-03-24
refactor: use neverthrow in /transaction (part 3)

mantariksh:
Builds on #1451

## Problem

See #1450

## Solution

This is the 3rd part of the solution, which focuses on the `POST /transaction` endpoint.

## Tests
See #1450


karrui (approved):
lgtm

PR #1453, opened 2021-03-24
refactor: use neverthrow in /:transactionId/reset (part 4)

mantariksh:
Builds on #1452

## Problem

See #1450

## Solution

This is the 4th part of the solution, which focuses on the `POST /transaction/:transactionId/reset` endpoint.

## Tests
See #1450

### DIFF
--- a/src/app/modules/verification/verification.controller.ts
+++ b/src/app/modules/verification/verification.controller.ts
@@ -87,28 +87,31 @@ export const handleGetTransactionMetadata: RequestHandler<
  * @param req
  * @param res
  */
-export const resetFieldInTransaction: RequestHandler<
+export const handleResetField: RequestHandler<
   { transactionId: string },
-  string,
+  { message: string },
   { fieldId: string }
 > = async (req, res) => {
-  try {
-    const { transactionId } = req.params
-    const { fieldId } = req.body
-    const transaction = await VerificationService.getTransaction(transactionId)
-    await VerificationService.resetFieldInTransaction(transaction, fieldId)
-    return res.sendStatus(StatusCodes.OK)
-  } catch (error) {
-    logger.error({
-      message: 'Error resetting field in transaction',
-      meta: {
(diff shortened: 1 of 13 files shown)


src/app/modules/verification/verification.controller.ts
```diff
@@ -87,28 +87,31 @@ export const handleGetTransactionMetadata: RequestHandler<
  * @param req
  * @param res
  */
-export const resetFieldInTransaction: RequestHandler<
+export const handleResetField: RequestHandler<
   { transactionId: string },
-  string,
+  { message: string },
```
karrui, line 92: nit: I added a DTO type `ErrorDto` somewhere in the codebase to make it obvious what we are returning if you want to use it

src/app/modules/verification/verification.service.ts
```diff
@@ -159,10 +267,10 @@ export const getNewOtp = async (
   answer: string,
 ): Promise<void> => {
   // TODO (#317): remove usage of non-null assertion
-  if (isTransactionExpired(transaction.expireAt!)) {
+  if (isDateExpired(transaction.expireAt!)) {
```
karrui, line 270: can we not use `isTransactionExpired` function exported from `verification.util`? Seems like the same thign

karrui (approved):
lgtm with nits

src/app/modules/verification/verification.service.ts
mantariksh: yes it is! this is just a rename because the name conflicts with the new `isTransactionExpired`. I'm trying to isolate the changes to only `/:transactionId/reset` in this PR. `isDateExpired` gets replaced with `isTransactionExpired` in a subsequent PR.

src/app/modules/verification/verification.controller.ts
mantariksh: looks good, will make the change in #1455 to avoid merge conflicts

PR #1454, opened 2021-03-24
refactor: use neverthrow in /:transactionId/otp (part 5)

mantariksh:
Builds on #1453

## Problem

See #1450

## Solution

This is the 5th part of the solution, which focuses on the `POST /transaction/:transactionId/otp` endpoint.

## Tests
See #1450

PR #1455, opened 2021-03-24
refactor: use neverthrow in /:transactionId/otp/verify (part 6)

mantariksh:
Builds on #1454

## Problem

See #1450

## Solution

This is the 6th part of the solution, which focuses on the `POST /transaction/:transactionId/otp/verify` endpoint. It includes the following additional improvements:
- Deletes the old `verification.(controller|service).spec.ts` and replaces them with `verification.(controller|service).new.spec.ts`
- Adds factory tests
- Logs form ID in all verification service functions

## For reviewers
Note that in `verification.service.spec.ts` and `verification.controller.spec.ts`, the only new tests added were for `verifyOtp` and `handleVerifyOtp` respectively. The rest of the changes are due to the file renaming.

## Tests
See #1450


mantariksh:
TODO from previous PRs
- [ ] Use ErrorDto type
- [ ] Use `transformMongoError`
- [ ] Set default mock value in beforeAll instead of using mockReset
- [ ] Use date-fns
- [ ] Log form ID as well as transaction/field ID

karrui (approved):
lgtm

PR #1456, opened 2021-03-24
build: fix mergify config

mantariksh:
Mergify likely not running because the config file has an extra `.` in front, see [docs](https://docs.mergify.io/getting-started.html#configuration) and [this forum post](https://github.com/Mergifyio/mergify-engine/discussions/2189) for confirmation. Also adds an approval step.


karrui (approved):
lmao

Issue #149, opened 2020-08-17
Combine submission middlewares into controllers


liangyuanruo:
Blocked by #144

Many of our middleware functions are being chained together and used as services, which is an [anti-pattern](https://insertafter.com/en/blog/no_more_middlewares.html). This causes problems such as having long-range object and type dependencies, which must be overridden by the developer with type assertions.

A better solution where possible, is to merge middlewares into a single controller function as far as possible so that dependencies are clearly defined.

arshadali172:
- [ ] Webhook controller should be removed and webhook service referenced in the submission controller

liangyuanruo:
@karrui @tshuli i had a discussion with @mantariksh today, and we thought that instead of chaining middlewares, we can group relevant parts of the `celebrate` middleware together.

For example, instead of doing this:

```typescript
app.route(...).post(
  celebrate({
    body: Joi.object({
        valueA: Joi.string(),
        valueB: Joi.boolean()
    })
  })
  middlewareUsingValueA, // naked middleware function with implicit
  middlewareUsingValueB, // dependency on upstream middleware for protection
)
```

We can do this instead:

```typescript
// Exported together so that it's not possible to
// use middleware without protection
export const middlewareUsingValueA = [
  celebrate({
    body: Joi.object({
        valueA: Joi.string(),
    })
  }),
  (req, res, next) => {...} // or implemented as a separate function
]

export const middlewareUsingValueB = [
  celebrate({
    body: Joi.object({
        valueB: Joi.string(),
    })
  }),
  (req, res, next) => {...}
]

app.route(...).post(
  middlewareUsingValueA,
  middlewareUsingValueB
)
```

liangyuanruo:
Note: to also refactor associated routes for

- [ ] VerifiedContentMiddleware.encryptVerifiedSpcpFields
- [ ] EncryptSubmissionMiddleware.prepareEncryptSubmission
- [ ] encryptSubmissions.saveResponseToDb
- [ ] webhookVerifiedContentFactory.post
- [ ] SubmissionsMiddleware.sendEmailConfirmations

PR #1474, opened 2021-03-29
refactor: Migrate the rest of the filters to handleEncryptedSubmission

frankchn:
## Problem
Part of the set of pull requests that addresses #149

## Solution
Migrate `VerifiedContentMiddleware.encryptVerifiedSpcpFields`, `EncryptSubmissionMiddleware.prepareEncryptSubmission`, `encryptSubmissions.saveResponseToDb`, `webhookVerifiedContentFactory.post`, `SubmissionsMiddleware.sendEmailConfirmations`.

I believe all these filters are used elsewhere in the code, so not deleting them yet.


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -231,7 +266,139 @@ export const handleEncryptedSubmission: RequestHandler = async (
...
+    attachmentUploadPromises.push(
+      AwsConfig.s3
+        .upload({
+          Bucket: AwsConfig.attachmentS3Bucket,
+          Key: uploadKey,
+          Body: Buffer.from(individualAttachment),
+        })
+        .promise(),
+    )
+  }
```
liangyuanruo, line 335: can help to see if it's possible to extract some of this detail into a service? we're trying to keep things simple in the controller so that no real work is performed here; it should only be determining HTTP response statuses

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -231,7 +266,139 @@ export const handleEncryptedSubmission: RequestHandler = async (
...
+    }
+  }
+
+  // Verify Structure of Encrypted Response
+  // Step 1: Add req.body.encryptedContent to req.formData
+  // eslint-disable-next-line @typescript-eslint/no-extra-semi
+  ;(req as WithFormData<typeof req>).formData = req.body.encryptedContent
+  // Step 2: Add req.body.attachments to req.attachmentData
+  ;(req as WithAttachmentsData<typeof req>).attachmentData =
+    req.body.attachments || {}
```
liangyuanruo, line 305: previously we had to tack on `attachmentData` to `req` because of long-range dependencies across multiple middleware, but i think we can dispense with this pattern now :)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -231,7 +266,139 @@ export const handleEncryptedSubmission: RequestHandler = async (
...
+  })
+  req.submission = savedSubmission
+
+  // Fire webhooks if available
+  const webhookUrl = form.webhook?.url
+  const submissionWebhookView = submission.getWebhookView()
+  if (webhookUrl) {
+    // Note that we push data to webhook endpoints on a best effort basis
+    // As such, we should not await on these post requests
+    void pushData(webhookUrl, submissionWebhookView)
```
liangyuanruo, line 380: can help to rename `pushData` to something more informative, like `webhook.send()`?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
frankchn: Maybe on the next PR? I don't want to do a bunch of things at the same time :p Adding a TODO

liangyuanruo (approved):
> I believe all these filters are used elsewhere in the code, so not deleting them yet.

thanks for pointing this out - could we get your help to update those to use the same controller pattern too? perhaps in a separate PR...

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
frankchn: Yeah, realized this after I finished haha. Removing...

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
frankchn: I am renaming this as `webhookPushData` in this, maybe the rename can come later?

frankchn:
@liangyuanruo fixed tests :)

liangyuanruo (approved):
LGTM barring minor merge conflicts

Issue #1434, opened 2021-03-23
Migrate admin-forms.server.controller.js to TS


liangyuanruo:
- [ ] Collapse middlewares in routes

PR #1526, opened 2021-03-31
ref: collapse middlewares of GET /adminform/submissions/metadata route

karrui:
## Problem

This PR cleans up the janky `authEncryptedResponseAccess` middleware used in the `/adminform/submissions/metadata` endpoint and collapses the used middlewares in the `handleGetMetadata` controller handler.

This PR is just one of the few PRs needed to remove total usage of the `authEncryptedResponseAccess` middleware.

Related to #1434

## Solution

**Improvements**:

- feat: inline auth middlewares for `handleGetMetadata`
- move `withUserAuthentication` middleware of submissions count API route to execute first

## Tests

- unit tests have been added for the refactored controller handler function

### Manual tests
- [ ] Go to view responses for a storage mode form. The list of responses should be shown properly.

### DIFF
--- a/src/app/modules/submission/encrypt-submission/__tests__/encrypt-submission.controller.spec.ts
+++ b/src/app/modules/submission/encrypt-submission/__tests__/encrypt-submission.controller.spec.ts
@@ -158,6 +158,16 @@ describe('encrypt-submission.controller', () => {
       title: 'mock title',
     } as IPopulatedForm
 
+    beforeEach(() => {
+      MockUserService.getPopulatedUserById.mockReturnValue(okAsync(MOCK_USER))
+      MockAuthService.getFormAfterPermissionChecks.mockReturnValue(
+        okAsync(MOCK_FORM),
+      )
+      MockEncryptSubService.checkFormIsEncryptMode.mockReturnValue(
+        ok(MOCK_FORM as IPopulatedEncryptedForm),
+      )
+    })
+
     it('should return 200 with single submission metadata when query.submissionId is provided and can be found', async () => {
       // Arrange
       const mockSubmissionId = new ObjectId().toHexString()
@@ -180,15 +190,6 @@ describe('encrypt-submission.controller', () => {
         submissionTime: 'some submission time',
       }
 
-      MockUserService.getPopulatedUserById.mockReturnValueOnce(
-        okAsync(MOCK_USER),
-      )
-      MockAuthService.getFormAfterPermissionChecks.mockReturnValueOnce(
-        okAsync(MOCK_FORM),
-      )
-      MockEncryptSubService.checkFormIsEncryptMode.mockReturnValueOnce(
-        ok(MOCK_FORM as IPopulatedEncryptedForm),
-      )
       MockEncryptSubService.getSubmissionMetadata.mockReturnValueOnce(
         okAsync(expectedMetadata),
       )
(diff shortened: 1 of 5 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -619,14 +628,33 @@ export const handleGetMetadata: RequestHandler<
...
+        getFormAfterPermissionChecks({
+          user,
+          formId,
+          level: PermissionLevel.Read,
+        }),
+      )
+      // Step 3: Check whether form is encrypt mode.
+      .andThen(checkFormIsEncryptMode)
+      // Step 4: Retrieve submission metadata.
+      .andThen<SubmissionMetadataList, DatabaseError>(() => {
```
seaerchin, line 645: oh wow i didn't know you could do this. this is just for cleanliness so that on the `map`, we can hover and see that it's `SubmissionMetadataList` instead of a long type right

seaerchin (commented):
lgtm

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
karrui: no la, sometimes neverthrow not smart enough to deduce what it is from the inner block, so this is more akin to type casting the returnable types

mantariksh: actually you don't need this casting right? I tried removing it and it works fine

karrui: I could have sworn the types were `unknown` before... Nonetheless fixed in 0d80908

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -619,14 +628,34 @@ export const handleGetMetadata: RequestHandler<
...
+      .andThen(checkFormIsEncryptMode)
+      // Step 4: Retrieve submission metadata.
+      .andThen(() => {
+        // Step 4a: Retrieve specific submission id.
+        if (submissionId) {
+          return getSubmissionMetadata(formId, submissionId).map((metadata) => {
+            const metadataList: SubmissionMetadataList = metadata
+              ? { metadata: [metadata], count: 1 }
+              : { metadata: [], count: 0 }
+            return metadataList
```
mantariksh, line 652:
```suggestion
            return metadata
              ? { metadata: [metadata], count: 1 }
              : { metadata: [], count: 0 }
```

src/app/modules/submission/encrypt-submission/__tests__/encrypt-submission.controller.spec.ts
```diff
@@ -252,6 +306,205 @@ describe('encrypt-submission.controller', () => {
...
+      })
+      const mockRes = expressHandler.mockResponse()
+
+      MockUserService.getPopulatedUserById.mockReturnValueOnce(
+        okAsync(MOCK_USER),
+      )
+      const expectedError = new ForbiddenFormError('no access')
+      MockAuthService.getFormAfterPermissionChecks.mockReturnValueOnce(
+        errAsync(expectedError),
+      )
```
mantariksh, line 376: might it be better to stick all these mocks in a `beforeEach` and vary them only when necessary? rn the mocks are v long and repetitive, so it's hard to tell at a glance which part of the mock is varying, and what the test is trying to do

mantariksh (approved):
gmtl

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
karrui: i wanted to add typing to it haha

mantariksh: ah makes sense

src/app/modules/submission/encrypt-submission/__tests__/encrypt-submission.controller.spec.ts
karrui: Very good point, resolved in d927feb

PR #1531, opened 2021-03-31
refactor: migrate admin-console client service to ts

### LINKED ISSUE #1521: Convert  admin-console.client.service.js to TypeScript
In addition to conversion to TS, suggest splitting the existing service in two (just like the backend modules), into an Examples client service and Billing client service.

seaerchin:
## Problem
Closes #1521 - refactors `admin-console.client.service.js` into two separate frontend services and writes adds tests for them

## Solution
Rewrote the `admin-console` service as two separate services in typescript.
Followed backend structure and structured it as `exampleService` and `billingService`

## Tests
1.tests for `billingService`/`exampleService` to ensure that api requests are being carried out and to ensure that errors are being thrown if promises are rejected

## Manual tests 

- [ ] Click on `examples` after logging in - there should be multiple examples loaded
- [ ] Login and create a form. Navigate to share and then click copy on Share Template. Open the copied url in another tab - the form should open as a template.
- [ ] Login and navigate to billing. 
    1. entering an invalid eserviceId should have 0 logins
    2. entering a valid eserviceId should have the logins for the specified time period.


liangyuanruo (changes requested):
> the JSDoc for getExampleForms is incomplete because i'm unsure how to describe them. the corresponding backend controller examples.controller doesn't have the documentation defined on the parameter/type. will ask around and write it up

Thanks, noted. Appreciate it if you could derive and fill in the backend JSDocs, if they're missing too?

Meanwhile, you should actually switch over to using the TS implementation and delete the old JS files. Don't forget to test on staging if you have to.

seaerchin:
@liangyuanruo yep, i'm going to check with either @karrui or @mantariksh and fill in both fe/be documentation

src/public/services/BillingService.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param esrvcId The id of the form
+ * @returns Promise<BillingResult> The billing statistics of the given month
+ */
+export const getBillingInfo = (
+  yr: string,
+  mth: string,
+  esrvcId: string,
+): Promise<BillingResult> =>
+  axios
+    .get(BILLING_ENDPOINT, {
```
liangyuanruo, line 22: You should fill in the type parameter which represents the shape of data that the backend is returning. Something like this:

```typescript
  axios
    .get<BillingResult>(BILLING_ENDPOINT, {
```

src/public/services/BillingService.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param mth The month to get the billing information for
+ * @param esrvcId The id of the form
+ * @returns Promise<BillingResult> The billing statistics of the given month
+ */
+export const getBillingInfo = (
+  yr: string,
+  mth: string,
+  esrvcId: string,
+): Promise<BillingResult> =>
+  axios
```
karrui, line 21: nit: prefer explicit return when the return cannot fit into a single line for readability

```ts
): Promise<BillingResult> => {
  return axios.get(...)
}
```

src/public/services/ExampleService.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @returns The list of retrieved examples if `shouldGetTotalNumResults` is false
+ * @returns The list of retrieved examples with the total results if `shouldGetTotalNumResults` is true
+ */
+export const getExampleForms = (
+  pageNo: number,
+  searchTerm?: string,
+  agency?: string,
+  shouldGetTotalNumResults?: boolean,
+): Promise<ExampleFormsResult> =>
+  axios
```
karrui, line 30: nit: same for this

src/public/services/BillingService.ts
```diff
@@ -0,0 +1,29 @@
+import axios from 'axios'
+
+import { LoginStatistic } from 'src/types/login'
+
+// Exported for testing
+export const BILLING_ENDPOINT = '/billing'
+type BillingResult = LoginStatistic
```
liangyuanruo, line 7: This looks dodgy - BillingResult should be typed as a _data transfer object_ or DTO, which describes the shape of data returned by the backend. You can refer to _AnalyticStatsDto_ type to see how it is being used in both frontend and backend - this is to ensure both client and server have a common contractual understanding of the shape of data.

seaerchin: fixed - adding dto for `exampleService` and `billingService` fe/be. i think we do need a utility type that is just `PossibleError<T>` that wraps `T` to be either `T | ErrorDto` just for convenience. 

also, i wrote the explicit parameter documentation on the dto rather than the function itself. function just has a high level view of the actual parameters.

src/public/services/__tests__/BillingService.test.ts
```diff
@@ -0,0 +1,52 @@
...
+          params: MOCK_PARAMS,
+        },
+      )
+      expect(actual).toEqual(MOCK_DATA)
+    })
+
+    it('should reject with the provided error message when the GET request fails', async () => {
+      // Arrange
+      const expected = new Error('Mock Error')
+      const shouldReject = () => BillingService.getBillingInfo(MOCK_PARAMS)
```
karrui, line 37: should this line be in the act block instead?

src/public/services/__tests__/BillingService.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      MockAxios.get.mockRejectedValueOnce(expected)
+
+      // Act
+      await expect(shouldReject).rejects.toEqual(expected)
```
karrui, line 41: should this line be in the assert block?  fix for all other tests?

src/types/api/billing.ts
```diff
@@ -0,0 +1,22 @@
...
+// mth: The month to get the billing information for
+// esrvcId: The id of the form
+export type BillingQueryDto = {
+  esrvcId: string
+  yr: string
+  mth: string
+}
+
+// NOTE: LoginStatistic is defined here and in src/types/login.
+// This is to allow the communication layer (the dto) to be agnostic of intermediate representations.
```
karrui, line 13: It's ok, just import instead so we have a single source of truth

src/app/modules/billing/billing.controller.ts
```diff
@@ -54,7 +51,7 @@ export const handleGetBillInfo: RequestHandler<
     })
     return res
       .status(StatusCodes.INTERNAL_SERVER_ERROR)
-      .json('Error in retrieving billing records')
```
karrui, line 57: double check to see if frontend client acts on the error message in any way for backwards compatibility.

karrui (commented):
some queries

src/public/services/__tests__/BillingService.test.ts
seaerchin: i don't have a strong opinion on this; i thought it belongs here because it has a act component where it rejects 

i think for clarity, i'll just split this into two portions accordingly

src/app/modules/billing/billing.controller.ts
seaerchin: the frontend client only logs the `error` object at most and does not do anything meaningful beyond that

karrui (approved):
lgtm, please add manual test steps for next release

karrui:
This PR may have conflicts with #1573, please coordinate :P
@orbitalsqwib FYI

seaerchin:
> This PR may have conflicts with #1573, please coordinate :P
> @orbitalsqwib FYI

talked to @orbitalsqwib on this. we finalized that i'll merge my PR in first and he'll rebase/merge off this accordingly.  @karrui fyi

src/types/api/billing.ts
```diff
@@ -0,0 +1,12 @@
+import { LoginStatistic } from '../../types'
+
+// yr: The year to get the billing information for
+// mth: The month to get the billing information for
+// esrvcId: The id of the form
+export type BillingQueryDto = {
+  esrvcId: string
+  yr: string
+  mth: string
+}
```
mantariksh, line 10: is this really a DTO? it's only used internally by the frontend right?

src/types/api/examples.ts
```diff
@@ -0,0 +1,20 @@
...
+// pageNo: The page to render
+// searchTerm: The term to search on
+// agency: The agency to search on - this can be all agencies or the user's agency
+// shouldGetTotalNumResults: Whether to return all the results or not
+export type ExampleFormsQueryDto = {
+  pageNo: number
+  searchTerm?: string
+  agency?: string
+  shouldGetTotalNumResults?: boolean
+}
```
mantariksh, line 16: is this really a DTO if it's only used internally by the frontend?

src/public/services/__tests__/BillingService.test.ts
```diff
@@ -0,0 +1,52 @@
+import MockAxios from 'jest-mock-axios'
+
+import * as BillingService from '../BillingService'
+
+jest.mock('axios', () => MockAxios)
+
+describe('BillingService', () => {
+  describe('getBillingInfo', () => {
+    const MOCK_DATA = {
+      adminEmail: 'Big Mock',
```
mantariksh, line 10: hahahahaah

src/types/api/billing.ts
seaerchin: it's used by fe and be. the be uses it as part of `req.body`'s shape in the type of `RequestHandler`

seaerchin: it's used in `src/app/modules/billing/billing.controller.ts`

src/types/api/examples.ts
seaerchin: used in `src/app/modules/examples/examples.controller.ts` as part of be to verify request's body's shape

src/public/services/__tests__/BillingService.test.ts
seaerchin: i aim to make reading tests joyful HAHAHAH

liangyuanruo (approved):

src/types/api/billing.ts
mantariksh: ahh I see that they are query params, got it

mantariksh (approved):

liangyuanruo (commented):
Code LGTM, do address/resolve Antariksh's comments before merging.

PR #1539, opened 2021-04-01
refactor: convert submission limit service function to ResultAsync

mantariksh:
## Problem

The `checkFormSubmissionLimitAndDeactivateForm` function returns a `Promise<Result>` instead of `ResultAsync`, making it impossible to chain with other async operations using `andThen`.

## Solution

Refactor the function to return a `ResultAsync` instead. In the process, the return type was expanded to include `PossibleDatabaseError`. This made it necessary to include all the `PossibleDatabaseError` types in the respective `mapRouteError` functions used by controllers which depend on `checkFormSubmissionLimitAndDeactivateForm`.

### DIFF
--- a/src/app/modules/form/__tests__/form.service.spec.ts
+++ b/src/app/modules/form/__tests__/form.service.spec.ts
@@ -302,7 +302,12 @@ describe('FormService', () => {
       )
 
       // Assert
-      expect(actual._unsafeUnwrapErr()).toEqual(new PrivateFormError())
+      expect(actual._unsafeUnwrapErr()).toEqual(
+        new PrivateFormError(
+          'Submission made after form submission limit was reached',
+          form.title,
+        ),
+      )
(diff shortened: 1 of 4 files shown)


liangyuanruo (approved):
LGTM.

This is a nitpick, but I notice that `PossibleDatabaseError` was introduced a couple of days ago as a type alias of all the possible database error types. For clarity it could be a good idea to rename it to `DatabaseError`, and rename the existing class to `GenericDatabaseError` since that is what we seem to be using it as?

mantariksh:
> For clarity it could be a good idea to rename it to DatabaseError, and rename the existing class to GenericDatabaseError since that is what we seem to be using it as?

Yup I agree with this, but have been putting it off because it would cause merge conflicts absolutely everywhere. I was thinking it might be better to do it once everyone is mostly working on the frontend.

PR #1541, opened 2021-04-01
build: release 5.4.1 - server to respond with status code if storage mode form is archived

### LINKED ISSUE #1540: Submitting an archived storage mode form returns 504
To replicate:
1. Create storage mode form. Activate and open it.
2. In a separate tab, go to the form dashboard and archive the form.
3. Submit the form.

This is because the code calls res.status(410) without calling .json or .send.

liangyuanruo:
## Problem

Closes #1540

## Solution

Affected LOC to use `res.sendStatus()` instead.


mantariksh (approved):
should we bump the version number and update the CHANGELOG as well?

mantariksh (approved):

PR #1547, opened 2021-04-04
refactor: Refactor attachment upload into a service

frankchn:
## Problem
Part of the set of pull requests that addresses #149

## Solution
Refactor upload attachments to its own attachments service.


src/app/modules/attachment/attachment.service.ts
```diff
@@ -0,0 +1,51 @@
+import crypto from 'crypto'
```
mantariksh, line 1: I'm wondering whether this should be part of `encrypt-submission.service.ts`, since it's very specific to the encrypt submissions pipeline. what do you think?

src/app/modules/attachment/attachment.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import { aws as AwsConfig } from '../../../config/config'
+import { createLoggerWithLabel } from '../../../config/logger'
+import { AttachmentUploadError } from '../core/core.errors'
+
+const logger = createLoggerWithLabel(module)
+
+export const uploadAttachments = (
+  formId: string,
+  attachmentData: Record<string, unknown>,
+): ResultAsync<Map<string, string>, AttachmentUploadError> => {
```
mantariksh, line 13: perhaps it would be clearer to define a type alias `type AttachmentMetadata = Map<string, string>` in `encrypt-submission.types, and also to reduce repetition

src/app/modules/attachment/attachment.service.ts
```diff
@@ -0,0 +1,51 @@
...
+  const attachmentMetadata = new Map()
```
mantariksh, line 14: I think we should specify that this is a `new Map<string, string>()`, as it gets typed as `Map<any, any>` otherwise. or you could go with my other suggestion to define a type alias

src/app/modules/attachment/attachment.service.ts
```diff
@@ -0,0 +1,51 @@
...
+  const attachmentUploadPromises = []
+
+  for (const fieldId in attachmentData) {
```
mantariksh, line 17: this is a matter of personal preference, but you could consider using `Object.keys(attachmentData).reduce(...)` rather than a `for` loop, to make this code more maintainable. the accumulated data might look something like this:
```
import { ManagedUpload } from 'aws-sdk/clients/s3'

type AttachmentReducerData = {
  attachmentMetadata: Map<string, string>
  attachmentUploadPromises: Promise<ManagedUpload.SendData>[]
}
```

but again this is subjective, since arguably a `reduce` might be less readable.

src/app/modules/attachment/attachment.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    Promise.all(attachmentUploadPromises).then(() => attachmentMetadata),
+    (error) => {
+      logger.error({
+        message: 'S3 attachment upload error',
+        meta: { action: 'uploadAttachments' },
+        error,
+      })
+      return new AttachmentUploadError()
+    },
+  )
```
mantariksh, line 50: just as a matter of convention, could you use `.map(() => attachmentMetadata)` here instead of the `Promise.then` above? this makes it slightly easier for other developers to find what is being returned, since we use `.map` for this use case in most of the rest of the codebase

src/app/modules/core/core.errors.ts
```diff
@@ -73,3 +73,12 @@ export class MissingFeatureError extends ApplicationError {
     )
   }
 }
+
+/**
+ * Error thrown when attachment upload fails
+ */
+export class AttachmentUploadError extends ApplicationError {
+  constructor(message?: string) {
```
mantariksh, line 81: could you provide a simple default error message here (e.g. `constructor(message = 'Error while uploading encrypted attachments to S3')`? this is to make our logging clearer

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -279,35 +271,25 @@ export const handleEncryptedSubmission: RequestHandler = async (req, res) => {
...
-    attachmentMetadata.set(fieldId, uploadKey)
-    attachmentUploadPromises.push(
-      AwsConfig.s3
-        .upload({
-          Bucket: AwsConfig.attachmentS3Bucket,
-          Key: uploadKey,
-          Body: Buffer.from(individualAttachment),
-        })
-        .promise(),
+  let attachmentMetadata = new Map()
```
mantariksh, line 276: again, you can use `Map<string, string>` or a type alias to ensure that TypeScript doesn't revert to the default of `Map<any, any>`

mantariksh (changes requested):
thank you for this, the controller is much cleaner now! in addition to the suggestions in the comments, could you also help with the following:
1. add generic type arguments to the `RequestHandler` type for `handleEncryptedSubmission`. this is because when we don't supply generic type arguments, `req.body` is typed as `any`, which means we lack type safety in the controller. you can look at `handleEmailSubmission` for an example, and you can make use of the type `EncryptSubmissionBody` which is already defined in `encrypt-submission.types.ts`. I think the `attachments` key in `EncryptSubmissionBody` might be typed wrongly because it's missing the field ID, so you may have to fix that.
2. avoid using `res.locals` within the controller, as everything in `res.locals` is typed as `any` by default as well. we can just use local variables within the controller.

mantariksh (changes requested):
also, at the v2/submissions/encrypt endpoint (where `handleEncryptedSubmission` is called), we are calling `forms.formById` an extra time - could you remove that pls?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -279,35 +271,25 @@ export const handleEncryptedSubmission: RequestHandler = async (req, res) => {
...
+      form._id,
+      req.body.attachments,
     )
+
+    if (attachmentUploadResult.isErr()) {
+      return res.status(StatusCodes.BAD_REQUEST).json({
+        message:
+          'Could not upload attachments for submission. For assistance, please contact the person who asked you to fill in this form.',
+        spcpSubmissionFailure: false,
+      })
```
mantariksh, line 289: rather than hardcoding the message here, it might be better to add the new `AttachmentUploadError` to the `mapRouteError` function so we can view all the status codes and messages in one place

src/app/modules/attachment/attachment.service.ts
frankchn: Makes sense, moving it over.

src/app/modules/attachment/attachment.service.ts
frankchn: Changed to `AttachmentMetadata`

src/app/modules/attachment/attachment.service.ts
frankchn: Ah my preference is usually towards imperative style programming, but sure that works. Changed.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
frankchn: Changed to `AttachmentMetadata` type

frankchn: ah never mind, typescript doesn't support a straight `new AttachmentMetadata()` thing. changing to `Map<string, string>`

src/app/modules/attachment/attachment.service.ts
frankchn: ah never mind, typescript doesn't support a straight new AttachmentMetadata() thing. changing to Map<string, string>

frankchn:
@mantariksh made changes based on your comments, PTAL again, thanks!

mantariksh:
@frankchn thanks for the latest changes! could you also help with the following please:
1. add generic type arguments to the RequestHandler type for handleEncryptedSubmission. this is because when we don't supply generic type arguments, req.body is typed as any, which means we lack type safety in the controller. you can look at handleEmailSubmission for an example, and you can make use of the type EncryptSubmissionBody which is already defined in encrypt-submission.types.ts. I think the attachments key in EncryptSubmissionBody might be typed wrongly because it's missing the field ID, so you may have to fix that.
2. avoid using res.locals within the controller, as everything in res.locals is typed as any by default as well. we can just use local variables within the controller.
3. at the v2/submissions/encrypt endpoint (where handleEncryptedSubmission is called), we are calling forms.formById an extra time - could you remove that pls?

frankchn:
@mantariksh Should I do this in another PR? The other refactorings are not really part of this PR and I don't really think should be rolled into this?

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -17,17 +19,87 @@ import {
...
+      attachmentUploadPromises: [],
+    },
+  )
+
+  return ResultAsync.fromPromise(
+    Promise.all(attachmentUploadPromises),
+    (error) => {
+      logger.error({
+        message: 'S3 attachment upload error',
+        meta: { action: 'uploadAttachments' },
```
mantariksh, line 95: could you include the `formId` in the `meta` for logging purposes?

mantariksh (commented):
@frankchn sure, we can incorporate those changes in another PR. the code looks good other than a minor point on logging as well as merge conflicts (which should be easy to resolve).

mantariksh (approved):
lgtm other than merge conflicts!

PR #1567, opened 2021-04-06
ref: migrate (most) /adminform endpoints to TypeScript, add integration tests

karrui:
## Problem

This PR migrates most /adminform endpoints in `admin-forms.server.routes.js` to their Typescript counterpart.

Related to #1434. Unable to close as there are two routes left (preview submission for email/encrypt)

## Solution

**Features**:
- migrate most admin form endpoints to Typescript
- add `jsonParseStringify` test utility 

**Improvements**
- validate that there is at least 1 email provided when creating/duping email form
  - There is already validation on the model layer when creating the document. This change just brings the validation forward to the Joi validation stage instead of document creation stage.
- use joi date extension to validate dates in YYYY-MM-DD format instead of regex

## Tests

- add integration tests for GET /adminform
- add integration tests for POST /adminform
- add integration tests for GET /:formId/adminform
- add integration tests for PUT /:formId/adminform
- add integration tests for DELETE /:formId/adminform
- add integration tests for POST /:formId/adminform
- add integration tests for POST /:formId/adminform/transfer-owner
- add integration tests for GET /:formId/adminform/template
- add integration tests for POST /:formId/adminform/copy
- add integration tests for GET /:formId/adminform/preview
- add integration tests for GET /:formId/adminform/feedback
- add integration tests for GET /:formId/adminform/feedback/count
- add integration tests for GET /:formId/adminform/feedback/download
- add integration tests for GET /:formId/adminform/submissions
- add integration tests for GET /:formId/adminform/submissions/count
- add integration tests for GET /:formId/adminform/submissions/metadata
- add integration tests for GET /:formId/adminform/submissions/download
- add integration tests for POST /:formId/adminform/images
- add integration tests for POST /:formId/adminform/logos


src/app/modules/form/admin-form/admin-form.routes.ts
```diff
@@ -0,0 +1,497 @@
...
+    fileMd5Hash: Joi.string().base64().required(),
+    fileType: Joi.string()
+      .valid(...VALID_UPLOAD_FILE_TYPES)
+      .required(),
+  },
+})
+
+AdminFormsRouter.route('/adminform')
+  // All HTTP methods of route protected with authentication.
+  .all(withUserAuthentication)
```
mantariksh, line 92: is it possible to just use `AdminFormsRouter.use(withUserAuthentication)` at the top?

liangyuanruo (commented):
Appreciate the heroic effort. Is it possible to split the tests into separate files?

src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
```diff
@@ -0,0 +1,4621 @@
...
+        title: 'cool new title',
+      }
+
+      // Act
+      const response = await request
+        .post(`/${formToDupe._id}/adminform`)
+        .send(invalidEmailDupeParams)
+
+      // Assert
+      expect(response.status).toEqual(500)
```
mantariksh, line 1400: should this be a 500? looks like a validation error, so should be 400 right?

mantariksh (approved):

src/app/modules/form/admin-form/admin-form.routes.ts
karrui: Not possible, due to our non-RESTful API hierarchy :(.

See d741b76;

> apparently the way express handles pathing is that they flatten all paths; and since AdminFormRouter (due to lack of RESTfulness) is on the root route '/', the withUserAuthentication middleware is now applied across the application, which is not ideal.

src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
karrui: Yea, but I did not want to pollute this PR, and that was the old behaviour. Do you think I should just fix it here?

src/app/modules/form/admin-form/admin-form.routes.ts
mantariksh: ahh I see

src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
mantariksh: I see, nah it seems out of scope

karrui:
> Appreciate the heroic effort. Is it possible to split the tests into separate files?

The tests will be split when the new routes gets created due to API refactor (and copy-pasting the old integration tests into the new tests), so I think we can just keep them here for the time being. 

These tests will eventually be deleted after they get migrated. I just had to write them first for some semblance of safety before we migrate /extract APIs

Issue #1513, opened 2021-03-30
[API] Duplicate public form related endpoints to their new /api counterparts


karrui:
This issue tracks the straight duplication of the various public form endpoints to their `/api` rooted endpoints to follow a RESTful hierarchy. 

The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

The current to new mapping is shown below:

\# | Existing endpoints | New endpoints<br>(note that all routes are prefixed with /api/v3) | Description
-- | -- | --|--
1 | GET /:formId/publicform | GET /forms/:formId | retrieve public form data 
2 | POST /:formId/feedback | POST /forms/:formId/feedback | submit form feedback
3 | POST /v2/submissions/encrypt/:formId | POST /forms/:formId/submissions/encrypt | submit storage mode form
4 | POST /v2/submissions/email/:formId | POST /forms/:formId/submissions/email | submit email mode form

karrui:
Blocked by #1385

karrui:
row 3 is also still blocked pending collapsing of encrypt submission endpoint

karrui:
row 3 is unblocked

karrui:
Found an inconsistency in the endpoint for POSTing public form feedback. the new endpoint should be 

```diff
- POST /forms/:formId/feedback
+ POST /forms/:formId/feedbacks
```

PR #1572, opened 2021-04-07
Test: publicform integration tests

seaerchin:
## Problem

current public form end point does not have any integration tests and only has unit tests on the controller. this PR adds integration tests for HTTP errors.

Part of #1513 

## Solution

Added integration tests for `GET /:formId/publicform` endpoint.

### DIFF
--- a/src/app/modules/form/public-form/__tests__/public-form.routes.spec.ts
+++ b/src/app/modules/form/public-form/__tests__/public-form.routes.spec.ts
@@ -1,5 +1,5 @@
 import SPCPAuthClient from '@opengovsg/spcp-auth-client'
-import _ from 'lodash'
+import { ObjectId } from 'bson-ext'
 import { errAsync } from 'neverthrow'
 import supertest, { Session } from 'supertest-session'
 import { mocked } from 'ts-jest/utils'
@@ -196,7 +196,7 @@ describe('public-form.routes', () => {
         usedCount: 0,
         state: MyInfoCookieState.Success,
       })
-      const MOCK_FORM_ID = _.pad('', 24, '1')
+      const MOCK_FORM_ID = new ObjectId().toHexString()
       const expectedResponseBody = JSON.parse(
         JSON.stringify({
           message: 'Form not found',
@@ -223,8 +223,9 @@ describe('public-form.routes', () => {
       })
       const expectedResponseBody = JSON.parse(
         JSON.stringify({
-          message:
-            'If you think this is a mistake, please contact the agency that gave you the form link.',
+          message: form.inactiveMessage,
+          formTitle: form.title,
+          isPageFound: true,
         }),
       )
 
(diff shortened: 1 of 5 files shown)


src/app/modules/form/public-form/__tests__/public-form.routes.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+    })
+
+    it('should return 404 if the form does not exist', async () => {
+      // Arrange
+      const cookie = JSON.stringify({
+        accessToken: 'mockAccessToken',
+        usedCount: 0,
+        state: MyInfoCookieState.Success,
+      })
+      const MOCK_FORM_ID = _.pad('', 24, '1')
```
karrui, line 199: nit: Can just use `new ObjectId().toHextString()`

src/app/modules/form/public-form/__tests__/public-form.routes.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+
+    it('should return 404 if the form is private', async () => {
+      // Arrange
+      const { form } = await dbHandler.insertEmailForm({
+        formOptions: { status: Status.Private },
+      })
+      const expectedResponseBody = JSON.parse(
+        JSON.stringify({
+          message:
+            'If you think this is a mistake, please contact the agency that gave you the form link.',
```
karrui, line 227: The body response should have 2 more key-values:

```ts
formTitle: string
isPageFound: boolean
```

check if this is a regression

tests/unit/backend/helpers/jest-db.ts
```diff
@@ -214,6 +215,21 @@ const insertEncryptForm = async ({
...
 }
 
+const getFullFormById = async (
+  formId: string,
+): Promise<IPopulatedForm | null> =>
+  await getEmailFormModel(mongoose).getFullFormById(formId)
+
+// 1. Get full form by id
+// 2. Patch in new settings
+const updateForm = async (
```
karrui, line 225: I don't see this being used

karrui (approved):
tests are fine, but check if there is a regression on private form response

src/app/modules/form/public-form/__tests__/public-form.routes.spec.ts
seaerchin: good catch. this is a regression. will update tests and the controller

src/types/api/core.ts
```diff
@@ -1,3 +1,8 @@
-export type ErrorDto = {
+export interface ErrorDto {
   message: string
 }
+
+export interface PrivateFormErrorDto extends ErrorDto {
```
mantariksh, line 5: it doesn't make a difference but for the record, it wasn't necessary to change this to an interface, since you can use `&` to achieve the same thing

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -214,6 +214,19 @@ export const handleGetPublicForm: RequestHandler<
       })
     }
     const { errorMessage, statusCode } = mapRouteError(error)
+
+    // Specialized error response for PrivateFormError.
+    // This is to maintain backwards compatibility with the middleware implementation
+    if (error instanceof PrivateFormError) {
+      return res.status(statusCode).json({
```
mantariksh, line 221: noticed you fixed a regression here. in future it would be good to separate this kind of change into a different PR which can be reviewed with greater urgency. if you name your PR as you did here, other devs may (rightly) assume (as I did) that this PR only contains tests.

mantariksh (approved):

PR #1586, opened 2021-04-08
refactor: Extract delete logic endpoint

### LINKED ISSUE #1490: [API] Extract delete logic endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **deleting form logic**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
DELETE /api/v3/admin/forms/:formId/logic/:logicId
```

tshuli:
## Problem
Closes #1490

## Tests
- Create a form with single logic. 
  - [ ] Check that delete logic works on admin view Logic Tab. The network call should return 200.
  - [ ] Repeat the above with multiple logic. Check that only one logic is deleted.
- Test using manual API calls
  - [ ] Check that new delete endpoint `DELETE /forms/:formId/logic/:logicId` returns 200 when formId and logicId are correct and that the specific logicId is deleted from db
  - [ ] Check that endpoint returns 404 when formId or logicId are incorrect


src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -554,3 +558,59 @@ export const updateFormSettings = (
...
+      },
+    })
+    return err(new LogicNotFoundError('logicId does not exist on form'))
+  }
+
+  // Remove specified logic and then update form logic
+  const updated_form_logic = form.form_logics.filter(
+    (logic) => logic.id !== logicId,
+  )
+  const ModelToUse = getFormModelByResponseMode(form.responseMode)
```
karrui, line 591: I think general form model would work, don't need to retrieve form model by response mode right?

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -554,3 +558,59 @@ export const updateFormSettings = (
...
+
+  return ResultAsync.fromPromise(
+    ModelToUse.findByIdAndUpdate(
+      form._id,
+      { form_logics: updated_form_logic },
```
karrui, line 596: should we use mongo's [`$pull`](https://mongoosejs.com/docs/api/array.html#mongoosearray_MongooseArray-pull) operator instead? 

```ts
{ 
  form_logics: {
    $pull: { _id: logicId }
  }
}
```

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -554,3 +558,59 @@ export const updateFormSettings = (
...
+      logger.error({
+        message: 'Error occurred when deleting form logic',
+        meta: {
+          action: 'deleteFormLogic',
+          formId: form._id,
+          logicId,
+        },
+        error,
+      })
+      return new DatabaseError(getMongoErrorMessage(error))
```
karrui, line 612: should we return with `transformMongoError` instead?

```ts
return transformMongoError(error)
```

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.routes.spec.ts
```diff
@@ -265,4 +265,179 @@ describe('admin-form.routes', () => {
       })
     })
   })
+
+  describe('DELETE /forms/:formId/logic/:logicId', () => {
+    it('should return 200 with success message on successful form logic delete for email mode form', async () => {
+      // Arrange
+      const formLogicId = '606d408f37af650047dd1a6a'
```
karrui, line 272: should we use `new ObjectId()` instead of hardcoding a string?

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -554,3 +558,59 @@ export const updateFormSettings = (
...
+
+/**
+ * Deletes form logic.
+ * @param form The original form to delete logic in
+ * @param logicId the logicId to delete
+ * @returns ok(true) on success
+ * @returns err(database errors) if db error is thrown during logic delete
+ * @returns err(LogicNotFoundError) if logicId does not exist on form
+ */
+export const deleteFormLogic = (
```
karrui, line 570: can we add unit tests for this function too?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1010,3 +1011,49 @@ export const handleUpdateSettings: RequestHandler<
...
+ * Handler for DELETE /forms/:formId/logic/:logicId
+ * @security session
+ *
+ * @returns 200 with success message when successfully deleted
+ * @returns 403 when user does not have permissions to delete logic
+ * @returns 404 when form cannot be found
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database error occurs
+ */
+export const handleDeleteLogic: RequestHandler = (req, res) => {
```
karrui, line 1025: can we add unit tests for this controller function?

src/app/modules/form/form.errors.ts
```diff
@@ -51,3 +51,12 @@ export class TransferOwnershipError extends ApplicationError {
     super(message)
   }
 }
+
+/**
+ * Error to be returned when form logic cannot be found
+ */
+export class LogicNotFoundError extends ApplicationError {
+  constructor(message: string) {
```
karrui, line 59: nit: should we have a default message for this error? Then instantiating this error won't require an argument anymore

karrui (changes requested):
some changes~

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: added

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.routes.spec.ts
tshuli: edited

src/app/modules/form/form.errors.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: added

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: edited - but i think form_logics should be nested inside pull :)

tshuli:
@karrui thanks for reviewing, for re-review please

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -538,3 +542,58 @@ export const updateFormSettings = (
...
+        action: 'deleteFormLogic',
+        formId: form._id,
+        logicId,
+      },
+    })
+    return err(new LogicNotFoundError())
+  }
+
+  // Remove specified logic and then update form logic
+  const ModelToUse = getFormModel(mongoose)
```
karrui, line 572: nit: call this `FormModel`?

nonnit: should we instantiate this outside of this function so this retrieval of model only gets called a single time?

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.settings.routes.spec.ts
```diff
@@ -265,4 +265,179 @@ describe('admin-form.settings.routes', () => {
       })
     })
   })
+
+  describe('DELETE /forms/:formId/logic/:logicId', () => {
```
karrui, line 269: should this test block be in a separate `admin-form.logic.routes.spec.ts` file?

karrui (changes requested):

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.settings.routes.spec.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: edited

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -1191,4 +1193,83 @@ describe('admin-form.service', () => {
...
+    const logicId = new ObjectId().toHexString()
+    const mockFormLogic = {
+      form_logics: [
+        {
+          _id: logicId,
+          id: logicId,
+        } as ILogicSchema,
+      ],
+    }
+    const MOCK_EMAIL_FORM = ({
```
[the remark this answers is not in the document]
tshuli: edited

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -6564,4 +6565,174 @@ describe('admin-form.controller', () => {
...
+      expect(MockAdminFormService.deleteFormLogic).not.toHaveBeenCalled()
+
+      expect(mockRes.status).toHaveBeenCalledWith(403)
+
+      expect(mockRes.json).toHaveBeenCalledWith({
+        message: 'not authorized to perform write operation',
+      })
+    })
+
+    it('should return 404 when form cannot be found', async () => {
```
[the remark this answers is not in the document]
tshuli: added

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -6564,4 +6565,174 @@ describe('admin-form.controller', () => {
...
+          formId: MOCK_FORM_ID,
+          level: PermissionLevel.Write,
+        },
+      )
+      expect(MockAdminFormService.deleteFormLogic).not.toHaveBeenCalled()
+
+      expect(mockRes.status).toHaveBeenCalledWith(403)
+
+      expect(mockRes.json).toHaveBeenCalledWith({
+        message: 'not authorized to perform write operation',
```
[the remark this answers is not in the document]
tshuli: It works, because we mocked it in line 6641

```
      MockAuthService.getFormAfterPermissionChecks.mockReturnValue(
        errAsync(
          new ForbiddenFormError('not authorized to perform write operation'),
        ),
      )
```

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -538,3 +542,58 @@ export const updateFormSettings = (
...
+ * @param form The original form to delete logic in
+ * @param logicId the logicId to delete
+ * @returns ok(true) on success
+ * @returns err(database errors) if db error is thrown during logic delete
+ * @returns err(LogicNotFoundError) if logicId does not exist on form
+ */
+export const deleteFormLogic = (
+  form: IPopulatedForm,
+  logicId: string,
+): Result<never, LogicNotFoundError> | ResultAsync<true, DatabaseError> => {
```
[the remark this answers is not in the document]
tshuli: `Result` is returned only for `err(new LogicNotFoundError())`. There is no case where we return `ok(true)`. But we do have `ResultAsync<true`

tshuli:
@karrui thanks for reviewing, for re-review pls

karrui (changes requested):
some more changes;

will the changes to make client use this endpoint will be in a separate PR or was it just missed?

tshuli:
@karrui simplified the return; frontend changes will come in separate PR

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -538,3 +542,56 @@ export const updateFormSettings = (
...
+    FormModel.findByIdAndUpdate(
+      form._id,
+      {
+        $pull: { form_logics: { _id: logicId } },
+      },
+      {
+        new: true,
+        runValidators: true,
+      },
+    ).exec(),
```
mantariksh, line 582: should this be a model static method? the idea of the repository layer is that it encapsulates the data and also all the mutations that can be performed on the data

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -538,3 +542,56 @@ export const updateFormSettings = (
...
+          action: 'deleteFormLogic',
+          formId: form._id,
+          logicId,
+        },
+        error,
+      })
+      return transformMongoError(error)
+    },
+    // On success, return true
+  ).map(() => true)
```
mantariksh, line 596: instead of throwing away information by returning `true`, might it be better to return the saved form so the caller can decide what to do with it?

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -6564,4 +6566,221 @@ describe('admin-form.controller', () => {
...
+        params: {
+          formId: MOCK_FORM_ID,
+          logicId: wrongLogicId,
+        },
+        session: {
+          user: {
+            _id: MOCK_USER_ID,
+          },
+        },
+      })
```
mantariksh, line 6679: this isn't necessary right, since you've mocked `deleteFormLogic`?

tshuli: yes, it isn't necessary in order to test `deleteFormLogic`, but since we're testing `handleDeleteLogic`, this allows us to test that `handleDeleteLogic` passes the correct param value for `logicId` into `deleteFormLogic`

mantariksh: but you've already defined `mockReq` right? so you just need to test that `deleteFormLogic` is called with `logicId` instead of `wrongLogicId`. just trying to streamline the testing code here.

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -6564,4 +6566,221 @@ describe('admin-form.controller', () => {
...
+      admin: MOCK_USER,
+      _id: MOCK_FORM_ID,
+      title: 'mock title',
+      ...mockFormLogic,
+    } as IPopulatedForm
+
+    const mockReq = expressHandler.mockRequest({
+      params: {
+        formId: MOCK_FORM_ID,
+        logicId: logicId,
```
mantariksh, line 6598:
```suggestion
        logicId,
```

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
tshuli: ok, no strong views here. edited.

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: since we are deleting logic here, i thought it would suffice to return true since there is no new information created from the API Call (similar to archive form)

tshuli:
@karrui actually I added on the frontend changes, for review pls

src/public/modules/forms/admin/components/edit-logic.client.component.js
```diff
@@ -74,8 +75,9 @@ function editLogicComponentController($uibModal, FormFields) {
   }
 
   vm.deleteLogic = function (logicIndex) {
+    const logicIdToDelete = vm.myform.form_logics[logicIndex]._id
+    AdminFormService.deleteFormLogic(vm.myform._id, logicIdToDelete)
```
karrui, line 79: Since this is a promise, should we use `$q.when` and only splice the logic on success?

karrui (changes requested):
some changes on the frontend handling requested, lgtm otherwise

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: thanks for discussion; edited

src/public/modules/forms/admin/components/edit-logic.client.component.js
tshuli: thanks, edited

src/app/models/form.server.model.ts
```diff
@@ -577,6 +577,24 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     )
   }
 
+  // Deletes specified form logic.
+  FormSchema.statics.deleteFormLogic = async function (
+    this: IFormModel,
+    form: IPopulatedForm,
```
mantariksh, line 583: do you need to pass the entire form? would the form ID be sufficient? this would also make this function more extensible because it can be called without having to retrieve the entire populated form

src/app/modules/form/admin-form/admin-form.service.ts
mantariksh: I guess, but then if the caller wanted to use the new form document, they would have to re-derive the new version by ownself deleting the relevant logic ID right? wouldn't it be easier to just pass back the updated document?

I know returning `true` is a pattern we use quite often but I don't think it's a good practice, except if there genuinely isn't any information to return

src/types/form.ts
```diff
@@ -248,6 +248,10 @@ export type IPopulatedEmailForm = IPopulatedForm & IEmailForm
 
 export interface IFormModel extends Model<IFormSchema> {
   getOtpData(formId: string): Promise<FormOtpData | null>
+  deleteFormLogic(
+    formId: IPopulatedForm,
```
mantariksh, line 252: we should use consistent naming - if we're passing the full form then call the parameter `form`, but if we can afford to just pass the form ID then we should stick to `formId`

src/app/models/form.server.model.ts
```diff
@@ -577,6 +577,24 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     )
   }
 
+  // Deletes specified form logic.
+  FormSchema.statics.deleteFormLogic = async function (
```
mantariksh, line 581: can we add unit tests for this also pls?

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -1191,4 +1193,89 @@ describe('admin-form.service', () => {
...
+      form_logics: [
+        {
+          _id: logicId,
+          id: logicId,
+        } as ILogicSchema,
+      ],
+    }
+
+    const UPDATE_SPY = jest
+      .spyOn(FormModel, 'findByIdAndUpdate')
```
mantariksh, line 1209: once you write unit tests for `deleteFormLogic`, you can change this to spy on `deleteFormLogic` instead

mantariksh (changes requested):

src/types/form.ts
tshuli: edited

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
tshuli: added spy for deleteFormLogic

src/app/models/form.server.model.ts
tshuli: added

src/app/models/form.server.model.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: edited

tshuli:

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1048,6 +1049,58 @@ describe('Form Model', () => {
...
+      const mockFormLogic = {
+        form_logics: [
+          {
+            _id: logicId,
+            id: logicId,
+          } as ILogicSchema,
+        ],
+      }
+
+      it('should return form upon for successful delete', async () => {
```
mantariksh, line 1064:
```suggestion
      it('should return form upon successful delete', async () => {
```

mantariksh, line 1064: can we test deleting from an array containing multiple logic fields as well pls?

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -538,3 +542,52 @@ export const updateFormSettings = (
...
+ * @returns err(LogicNotFoundError) if logicId does not exist on form
+ */
+export const deleteFormLogic = (
+  form: IPopulatedForm,
+  logicId: string,
+): ResultAsync<IFormSchema, DatabaseError | LogicNotFoundError> => {
+  // First check if specified logic exists
+  if (!form.form_logics.some((logic) => logic.id === logicId)) {
+    logger.error({
+      message: 'Error occurred when deleting form logic',
```
mantariksh, line 561: can we make this error message more specific, since we know exactly what the problem is? maybe "Logic ID to be deleted does not exist"

tshuli: edited

src/app/models/__tests__/form.server.model.spec.ts
tshuli: added

tshuli:

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1053,6 +1054,103 @@ describe('Form Model', () => {
...
+        expect(modifiedForm?.responseMode).not.toBeNull()
+        expect(modifiedForm?.responseMode).toEqual(ResponseMode.Email)
+        expect(modifiedForm?.status).not.toBeNull()
+        expect(modifiedForm?.status).toEqual(Status.Public)
+
+        // Check that correct form logic has been deleted
+        expect(modifiedForm?.form_logics).toBeDefined()
+        expect(modifiedForm?.form_logics).toHaveLength(1)
+        const logic = modifiedForm?.form_logics || ['some logic']
+        expect((logic[0] as any)['_id'].toString()).toEqual(logicId2)
```
mantariksh, line 1138: in future, you can use a non-null assertion (`!`) to deal with this situation. we've disallowed it in the main code using `eslint`, but it's still allowed in tests.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1494,3 +1494,50 @@ export const handleUpdateFormField = [
...
+          user,
+          formId,
+          level: PermissionLevel.Write,
+        }),
+      )
+      // Step 3: Delete form logic
+      .andThen((retrievedForm) =>
+        AdminFormService.deleteFormLogic(retrievedForm, logicId),
+      )
+      .map(() => res.json({ message: 'Logic deleted successfully' }))
```
mantariksh, line 1526: rather than sending a superfluous message which we don't intend for the client to use, might it be better to just do `res.sendStatus(200)`?

src/public/services/AdminFormService.ts
```diff
@@ -29,3 +29,10 @@ export const updateSingleFormField = async (
     )
     .then(({ data }) => data)
 }
+
+export const deleteFormLogic = async (
+  formId: string,
+  logicId: string,
+): Promise<true> => {
```
mantariksh, line 36: is this function typed correctly? it seems to be returning an `AxiosResponse`, not `true`. looks like we need a `.then(() => true)`.

src/public/modules/forms/admin/components/edit-logic.client.component.js
```diff
@@ -74,8 +81,19 @@ function editLogicComponentController($uibModal, FormFields) {
   }
 
   vm.deleteLogic = function (logicIndex) {
-    vm.myform.form_logics.splice(logicIndex, 1)
-    updateLogic({ form_logics: vm.myform.form_logics })
+    const logicIdToDelete = vm.myform.form_logics[logicIndex]._id
+    $q.when(AdminFormService.deleteFormLogic(vm.myform._id, logicIdToDelete))
+      .then((result) => {
+        if (result.request.status == 200) {
```
mantariksh, line 87: I don't think this is necessary. `axios` will reject the Promise if a non-2xx status is returned anyway, so this chunk should only have to deal with the success case (i.e. you can go straight to `vm.myform.form_logics.splice(logicIndex, 1)`)

tshuli: thanks! edited

src/public/services/AdminFormService.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.controller.ts

src/app/models/__tests__/form.server.model.spec.ts
tshuli: got it

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -582,3 +586,52 @@ export const updateFormSettings = (
...
+        formId: form._id,
+        logicId,
+      },
+    })
+    return errAsync(new LogicNotFoundError())
+  }
+
+  // Remove specified logic and then update form logic
+  return ResultAsync.fromPromise(
+    FormModel.deleteFormLogic(form._id, logicId),
```
mantariksh, line 617: sorry final nit - should we do `String(form._id)` instead? `._id` has a nasty habit of giving us an unexpected type (e.g. ObjectId) at runtime.

mantariksh (approved):

src/app/models/form.server.model.ts
```diff
@@ -596,6 +596,24 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+    this: IFormModel,
+    formId: string,
+    logicId: string,
+  ): Promise<IFormSchema | null> {
+    return this.findByIdAndUpdate(
+      mongoose.Types.ObjectId(formId),
```
karrui, line 606: do not need to cast, mongoose will automatically cast for you; `string` type is fine

src/app/modules/form/admin-form/admin-form.service.ts
karrui: in this case it does not matter since mongoose will cast it properly, but yea why not

karrui (approved):
lgtm with some nits

src/app/modules/form/admin-form/admin-form.service.ts
mantariksh: it'll cast properly when we call findByIdAndUpdate, but the typing is still wrong

karrui: yea, better be safe than sorry!

tshuli: updated

PR #1588, opened 2021-04-08
refactor(PublicFormRoutes): duplicate /publicform to /api/v3 endpoints

### LINKED ISSUE #1513: [API] Duplicate public form related endpoints to their new /api counterparts

seaerchin:
## Problem
duplicates js endpoints in `public-forms.server.routes` to their ts counterparts 
Closes #1513.

## Solution

Duplicated over existing endpoints for `/forms/:formId/feedback` and `/forms/:formId/submissions/email` to the new routes file.

## Tests

1. ported over integration tests for `:/formId/submissions/email`
2. ported over integration tests for `:/formId/submissions/encrypt`
3. ported over integration tests for `:/formId`
4. wrote integration tests for `/:formId/feedback`

## Manual Tests 
**NOTE: Attempt the below tests in all auth types**

- [ ] Create a normal (NIL Auth) email mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel
- [ ] Create a SP email mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel
- [ ] Create a CP email mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel
- [ ] Create a MyInfo email mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel
- [ ] Create a normal (NIL Auth) storage mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel
- [ ] Create a SP storage mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel
- [ ] Create a CP storage mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel
- [ ] Create a MyInfo storage mode form with a few fields and publish it. Open the form and fill in the form together with feedback. The responses should show up in email while the feedback should show up in the admin panel

**MyInfo-specific**

- [ ] 1. Create an email mode form and publish it. When signing in using MockPass, deny authentication. The form should not have been signed in and there should be an error message with MyInfo error displayed


src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -2,6 +2,8 @@ import { RequestHandler } from 'express'
 import { StatusCodes } from 'http-status-codes'
 import querystring from 'querystring'
 
+import { ErrorDto } from 'src/types/api'
```
karrui, line 5: use relative imports instead?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -25,7 +27,7 @@ const logger = createLoggerWithLabel(module)
  */
 export const handleSubmitFeedback: RequestHandler<
   { formId: string },
-  unknown,
+  ErrorDto | { message: string; isPageFound: boolean; formTitle: string },
```
karrui, line 30: nit: prefer if the return type is typed to something like `PrivateFormErrorDto`

src/app/routes/api/v3/public/__tests__/public-forms.routes.spec.ts
```diff
@@ -0,0 +1,1140 @@
...
+MockMyInfoGovClient.mockImplementation(
+  () =>
+    (({
+      extractUinFin: mockExtractUinFin,
+    } as unknown) as MyInfoGovClient),
+)
+
+// Import last so mocks are imported correctly
+// eslint-disable-next-line import/first
+import { PublicFormsRouter } from '../public-forms.routes'
```
karrui, line 71: this probably does not matter anymore, can move the import to its correct position and remove the comment and lint ignore line

karrui (changes requested):
you might have forgotten to change all client api calls to the old endpoint to call the new endpoint...

src/app/modules/form/public-form/public-form.controller.ts
seaerchin: i agree - waiting on anty's review on [this pr](https://github.com/opengovsg/FormSG/pull/1572/files), which has the dto and will fix once the other pr is merged.

seaerchin:
> you might have forgotten to change all client api calls to the old endpoint to call the new endpoint...

just to clarify, in the **notes** section, i wrote that i didn't change due to what was specified on #1513; do we want to do so immediately or am i mistaken on what #1513 states? (because we should not delete the old endpoints by changing client to call the new ones)

src/app/routes/api/v3/public/__tests__/public-forms.routes.spec.ts
seaerchin: this does matter and removing the import causes the successful `MyInfo` test to fail. This is because the `mockExtractUinFin`, which needs to be setup in order for the test to pass successfully.

karrui:
> (because we should not delete the old endpoints by changing client to call the new ones)

Clarified IRL; call new endpoints with the client (without removing old endpoints); if not we will never be able to delete the old endpoints 😆

src/public/modules/forms/services/form-api.client.factory.js
```diff
@@ -78,6 +78,7 @@ function FormApi($resource, FormErrorService, FormFields) {
 
   // accessMode is either adminForm or publicForm
   let resourceUrl = '/:formId/:accessMode'
+  const V3_PUBLICFORM_URL = ':/formId'
```
seaerchin, line 81: @karrui did this for visibility rather than just changing '/:accessMode' to be `''` since we are going to port over all these to `react-query` eventually.

src/app/routes/api/v3/public/public-forms.routes.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Please strictly ensure backwards compatibility.
+ * @route GET /:formId
+ *
+ * @returns 200 with form when form exists and is public
+ * @returns 404 when form is private or form with given ID does not exist
+ * @returns 410 when form is archived
+ * @returns 500 when database error occurs
+ */
+PublicFormsRouter.get(
+  '/:formId([a-fA-F0-9]{24})/',
```
mantariksh, line 101: for consistency
```suggestion
  '/:formId([a-fA-F0-9]{24})',
```

src/app/routes/api/v3/public/public-forms.routes.ts
```diff
@@ -0,0 +1,103 @@
...
+  celebrate({
+    [Segments.BODY]: Joi.object()
+      .keys({
+        rating: Joi.number().min(1).max(5).cast('string').required(),
+        comment: Joi.string().allow('').required(),
+      })
+      // Allow other keys for backwards compability as frontend might put
+      // extra keys in the body.
+      .unknown(true),
+  }),
```
mantariksh, line 39: what do you think about moving this validation middleware into PublicFormController, and exporting both the validation middleware and `handleSubmitFeedback` as an array of middleware instead of separately?

we actually decided to practice this a while ago, but we haven't been very good about enforcing it. it makes sense because `handleSubmitFeedback` assumes that the request has a certain shape, and can only do so safely when the validation middleware comes before it. so the validation and the handler must always go together, so it makes sense to export them as a single unit.

src/public/modules/forms/services/form-api.client.factory.js
```diff
@@ -98,6 +99,7 @@ function FormApi($resource, FormErrorService, FormFields) {
         // disable IE ajax request caching (so new fields show on build panel)
       },
       getPublic: {
+        url: V3_PUBLICFORM_URL,
         method: 'GET',
         // disable IE ajax request caching (so new fields show on build panel)
         headers: { 'If-Modified-Since': '0' },
```
mantariksh, line 105: you can delete the line immediately after this one, since we don't need the `accessMode` param anymore for this

src/app/routes/api/v3/public/public-forms.routes.ts
seaerchin: do you mean something along the lines of 

```ts
const handleSubmitFeedback = ... (same as current) 
export const handleSubmitFeedbackWithValidation = [celebrate(...), handleSubmitFeedback]
```

seaerchin: @mantariksh following up on this - i will collapse 2 other separate endpoints as well 

1. handleEncryptedSubmission
2. handleEmailSubmission 

the only visible middleware should be the rate limiter. This is because we want to always validate captcha and submission parameters before deferring to the respective controllers but limit rate is tweakable and should not be viewed as being crucial to handling the respective submissions.

feel free to correct me if i'm wrong on this!

mantariksh: what about `receiveEmailSubmission`?

seaerchin: hidden from routes and is exported as part of `handleEmailSubmission`

mantariksh: excellent!

mantariksh (commented):
lgtm, but could you test on staging before approval?

seaerchin:
@mantariksh will add manual tests in also but it'll be happy path tests because this is a duplication

mantariksh (approved):

karrui (approved):

PR #1591, opened 2021-04-08
refactor: convert Bounce module to use neverthrow

### LINKED ISSUE #193: Convert Bounce, Webhook services to neverthrow
Part of typescript server migration.

`ApplicationError` is a new base error class [introduced](https://github.com/opengovsg/FormSG/blob/82e853e17cbd282db1f769a52db8c4fa0e7b7250/src/app/modules/core/core.errors.ts) in the emergency contact feature PR #151.

Tracking for https://github.com/opengovsg/FormSG/pull/176#discussion_r474792883 

EDIT: Most of the controllers/services refactored to Typescript has already been returning errors that are extended from the `ApplicationError` class. 

The remaining spots of Typescript that are still throwing errors and not using `neverthrow` (to return `ApplicationError` classes instead of throwing) are in:
- [ ] `bounce.service.ts`
- [x] `verification.service.ts` (being completed by @mantariksh)
- [x] `webhook.service.ts` (@orbitalsqwib)

Should fix those before closing this issue.

mantariksh:
Updates the Bounce module to use `neverthrow`.

A note on PR scope: there is a lot of refactoring that can be done in the Bounce module, but I don't think it should be a priority right now. The module is complex and was based on a long process of designing and refining, and refactoring it will not only take a long time but risks disrupting a service which has been automated for nearly a year. Hence I think a deeper refactor is out of scope of this PR, unless any part of it is genuinely un-readable.

Closes #193

## Manual tests
- [ ] Create an email mode form. Add a valid collaborator with a registered contact number, and add a non-existent email as one of the email recipients. Submit the form. The form should NOT be deactivated, and the valid email recipient should not receive any emails/SMSes. Check the document in the Bounces collection corresponding to the form ID. The `bounces` array should contain the correct `hasBounced` key for each email addres.
- [ ] Remove all valid email recipients and leave only non-existent emails. Submit the form. The form should be deactivated, and the form admin + collaborators should receive ONE email and TWO smses each, informing them of the bounce and form deactivation.

### DIFF
--- a/src/app/modules/bounce/__tests__/bounce.controller.spec.ts
+++ b/src/app/modules/bounce/__tests__/bounce.controller.spec.ts
@@ -1,9 +1,11 @@
 import { ObjectId } from 'bson'
 import mongoose from 'mongoose'
-import { errAsync, okAsync } from 'neverthrow'
+import { errAsync, ok, okAsync } from 'neverthrow'
 import { mocked } from 'ts-jest/utils'
 
 import getFormModel from 'src/app/models/form.server.model'
+// eslint-disable-next-line import/first
+import { handleSns } from 'src/app/modules/bounce/bounce.controller'
 import getBounceModel from 'src/app/modules/bounce/bounce.model'
 import * as BounceService from 'src/app/modules/bounce/bounce.service'
 import * as FormService from 'src/app/modules/form/form.service'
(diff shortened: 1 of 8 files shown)


src/app/modules/bounce/__tests__/bounce.controller.spec.ts
```diff
@@ -1,9 +1,11 @@
 import { ObjectId } from 'bson'
 import mongoose from 'mongoose'
-import { errAsync, okAsync } from 'neverthrow'
+import { errAsync, ok, okAsync } from 'neverthrow'
 import { mocked } from 'ts-jest/utils'
 
 import getFormModel from 'src/app/models/form.server.model'
+// eslint-disable-next-line import/first
```
karrui, line 7: is this line needed?

karrui (approved):
lgtm, thanks for the effort

src/app/modules/bounce/__tests__/bounce.controller.spec.ts
mantariksh: nope, thanks for the catch!

Issue #1571, opened 2021-04-07
Add unit tests for handleEmailSubmission


mantariksh:
The email submissions pipeline has unit tests for all services as well as integration tests, but there should also be unit tests to confirm that the controller is calling the services correctly.

LoneRifle:
Blocked by #1594

PR #1594, opened 2021-04-08
refactor(email-submission/controller): refactored email submission 

seaerchin:
## Problem
`handleEmailSubmission` uses alot of mutated variables and `if/else`, which could have been better written in order to avoid mental overhead on the maintainer and future developers

## Solution

This PR refactors the `handleEmailSubmission` so that it reads in a sequential fashion and there is only one `let` declared variable. 

## Design Considerations 
1. `let` scoped variable - `spcpSubmissionFailure`. This is because the flag is set only in 1 place and the frontend **does** do work if it is `true`. Although there could be a better alternative to avoid the `let`, it is typed as `undefined | true` for explicitness and i don't want to significantly refactor the controller just for this. 
2. the generation of `hashedFields/parsedResponses` is still pretty unwieldy and done as a `switch` case - this is because `MyInfo` treats this differently 

### Manual Tests
- [x] 1. Create a single field form without authentication and set its mode to email. Submitting the form should have the field data 
- [x] 2. Create a single field SP form and set its mode to email. Submitting the form should have the field data
- [x] 3. Create a single field CP form and set its mode to email. Submitting the form should have the field data
- [x] 4. Create a single field MyInfo form and set its mode to email. Submitting the form should have the field data
- [x] 5. Submit storage mode form to email mode endpoint
- [x] 6. Submit form which has been made private
- [x] 7. Submit form which has been archived
- [x] 8. Submit captcha-enabled form without captcha
- [x] 9. Submit captcha-disabled form
- [x] 10. Submit form which has exceeded submission limit
- [x] 11. Submit form with invalid responses
- [x] 12. Create multiple email fields with email confirmation, check that the correct confirmation email is sent for each one
- [x] 13. Submit form with invalid attachments


karrui:
@LoneRifle any conflicts here and what you're doing?

LoneRifle:
Blocks #1571

seaerchin:
> Blocks #1571

@LoneRifle should be fine now; seeing as you will write unit tests for this file, lmk if it reads better/what you'd liek to see improved!

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +33,244 @@ const logger = createLoggerWithLabel(module)
...
+      .andThen((form) =>
+        // Validate responses
+        EmailSubmissionService.validateAttachments(req.body.responses)
+          .andThen(() =>
+            SubmissionService.getProcessedResponses(form, req.body.responses),
+          )
+          .map((parsedResponses) => ({ parsedResponses, form }))
+          .mapErr(logErrorWithReqMeta('Error processing responses')),
+      )
+      .andThen(({ parsedResponses, form }) => {
```
seaerchin, line 115: point 3 in design considerations

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +33,244 @@ const logger = createLoggerWithLabel(module)
...
+        const { authType } = form
+        switch (authType) {
```
seaerchin, line 117: point 3 in design considerations

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +33,247 @@ const logger = createLoggerWithLabel(module)
...
 
-  // Check that form is public
-  const formPublicResult = FormService.isFormPublic(form)
-  if (formPublicResult.isErr()) {
-    logger.warn({
-      message: 'Attempt to submit non-public form',
-      meta: logMeta,
-      error: formPublicResult.error,
+  // Stores a predefined message and returns a function that can be used for mapErr
+  const logErrorWithMeta = (
```
karrui, line 52: i think this is an overoptimization; we should just verbosely write out the logging. It makes it easier to reason about and aids when developers use the logger to check logs.

mantariksh: ^ 100% agree

seaerchin:
@karrui @mantariksh are there any issues apart from the unclear logging?

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +33,247 @@ const logger = createLoggerWithLabel(module)
...
+        res.json({
+          message: 'Form submission successful.',
+          submissionId: submission.id,
         })
-        const { errorMessage, statusCode } = mapRouteError(
-          verifyMyInfoResult.error,
-        )
+
+        // Send email confirmations
+        return SubmissionService.sendEmailConfirmations({
```
mantariksh, line 264: the intent of the code might be clearer if we prefix this with `void` and shift it above the call to `res.json`, to make it more obvious that we're not awaiting on the email confirmations

seaerchin: this would result in a behavioral change from the previous implementation - if an error exists on sending email confirmations, this would not be reflected to the client.

if this is what's desired (which i think it is, because the submission should not affect the webpage view), then i think a better change would be to just segregate this into 2 blocks - `andThen` to send the confirmations and a `map` to send the json response back to the client

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +33,247 @@ const logger = createLoggerWithLabel(module)
...
+                  error,
+                )
+              })
+          default:
+            return ok({
+              form,
+              parsedResponses,
+              hashedFields: new Set<string>(),
+            }) as Result<
+              {
```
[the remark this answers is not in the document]
seaerchin: i agree with the first point. extracting it has value, because it helps the reader to understand this whole chunk without getting lost in the details.

for the second point, this is to aid in typehinting for typescript, and even if we were to add it as a generic argument to this particular `andThen`, we would still need the annotation on the errors just so that type level information is preserved and not lost (because of the `ok`).

i think this would impact readability just by virtue of how verbose it is (due to the errors) and wouldn't significantly impact clarity because the destructuring of the next `andThen` makes the shape and types clear

src/app/modules/submission/email-submission/email-submission.controller.ts
seaerchin: i combined it into a single `map` block and prefixed it with `void`; main reason is because we need to call `orElse` on `sendEmailConfirmations` which requires typehints. The typehinting here is rather verbose due to the parameter forwarding

mantariksh (commented):
could you also add manual tests for the following (and anything else you can think of)? can't take any chances with email submissions endpoint.
- submit storage mode form to email mode endpoint
- submit form which has been made private
- submit form which has been archived
- submit captcha-enabled form without captcha
- submit captcha-disabled form
- submit form which has exceeded submission limit
- submit form with invalid responses
- create multiple email fields with email confirmation, check that the correct confirmation email is sent for each one

seaerchin:
@mantariksh based on the old code, there are the following decision points

1. check if form is email mode
2. check if form is public
3. check if there is a captcha
4. check if form has hit submission limits
5. check if attachments are valid 
6. sending of admin email
7. sending of email confirmation (to person who submits form) 

i added 1 test based on point 5; unsure how to test point 7. i think the tests that you've described covers the rest of the decision points and i've verified them on local (with the exception of captcha, because my env var doesn't have it)

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +32,274 @@ const logger = createLoggerWithLabel(module)
...
+        })
+        // MyInfo access token is single-use, so clear it
+        return res.clearCookie(MYINFO_COOKIE_NAME, MYINFO_COOKIE_OPTIONS).json({
+          // Return the reply early to the submitter
+          message: 'Form submission successful.',
+          submissionId: submission.id,
+        })
+      })
+      .mapErr((error) => {
+        const { errorMessage, statusCode } = mapRouteError(error)
```
karrui, line 299: This is not in the original code; but since myinfo token is cleared in the success case, we should also clear it in the error case right? 

double check my assertion @mantariksh

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +32,274 @@ const logger = createLoggerWithLabel(module)
...
+          case AuthType.SP:
+            return SpcpFactory.extractJwt(req.cookies, authType)
+              .asyncAndThen((jwt) => SpcpFactory.extractSingpassJwtPayload(jwt))
+              .map((jwt) => ({
+                form,
+                parsedResponses: [
+                  ...parsedResponses,
+                  ...createSingpassParsedResponses(jwt.userName),
+                ],
+                hashedFields: new Set<string>(),
```
karrui, line 157: since this is a refactor, should we just not pass in hashedFields if we do not need to, and let `SubmissionEmailObj` assign a default?

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +32,274 @@ const logger = createLoggerWithLabel(module)
...
+                  meta: logMeta,
+                  error,
+                })
+                return error
+              })
+          default:
+            return ok({
+              form,
+              parsedResponses,
+              hashedFields: new Set<string>(),
```
karrui, line 201: would save on redundant declarations like this

src/app/modules/submission/email-submission/email-submission.controller.ts
seaerchin: in the original code, we clear the cookie regardless once the email submission is successfully sent isn't it?

karrui: Yea, i meant we should also clear the token in the error case; and this clearing of the token in the error case is not in the original code.

mantariksh: no, we should not clear it in the error case. if the submission fails through no fault of the user (eg they submitted before captcha could connect, or the admin email failed to send), they should not be randomly signed out.

karrui: I see, thanks for the explanation! Should we add it in the comments too?

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +32,274 @@ const logger = createLoggerWithLabel(module)
...
+          attachments,
+          dataCollationData: emailData.dataCollationData,
+          formData: emailData.formData,
+        })
+          .map(() => ({
+            form,
+            parsedResponses,
+            submission,
+            emailData,
+            logMeta,
```
mantariksh, line 265: why do you need to pass `logMeta` here? isn't it in scope throughout the function?

seaerchin: each `andThen` is a function. hence, variables created in `andThen` is only in scope for that single `andThen` :(

src/app/modules/submission/email-submission/email-submission.controller.ts
seaerchin: thanks for explaining; will add this in comments for future devs!

src/app/modules/submission/email-submission/email-submission.controller.ts
seaerchin: agreed, this is a cleaner implementation; i did that originally for clarity to see how `hashedFields` was being used but i think it's clear regardless

src/app/modules/submission/email-submission/email-submission.controller.ts
seaerchin: on i see the confusion; this should be `logMetaWithSubmission` - this is bad namign from me @_@ thanks for pointing it out

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -30,267 +32,286 @@ const logger = createLoggerWithLabel(module)
...
+        }) => {
+          // Send email confirmations
+          void SubmissionService.sendEmailConfirmations({
+            form,
+            parsedResponses,
+            submission,
+            attachments,
+            autoReplyData: emailData.autoReplyData,
+          }).mapErr((error) => {
+            // NOTE: MyInfo accses token is not cleared here.
```
karrui, line 291: nit: typo

karrui (approved):
lgtm; update your PR description? Remove the `logErrorWithMeta` block hehe since it is already removed

PR #1604, opened 2021-04-09
test: fix flakiness (again) by hardcoding created dates

karrui:
Tests seems to still be failing for a particular test.
This PR (attempts) to fix the flakiness by using hardcoded dates instead of using relative dates for the affected test, which may behave differently in different machines (like CI/CD)

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
@@ -4719,7 +4719,6 @@ describe('admin-form.routes', () => {
 
     it('should return 200 with stream of encrypted responses between given query.startDate and query.endDate', async () => {
       // Arrange
-      const now = new Date()
       const submissions = await Promise.all(
         times(5, (count) =>
           createSubmission({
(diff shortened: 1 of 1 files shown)


src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
```diff
@@ -4748,8 +4746,8 @@ describe('admin-form.routes', () => {
       const response = await request
         .get(`/${defaultForm._id}/adminform/submissions/download`)
         .query({
-          startDate: format(subDays(now, 4), 'yyyy-MM-dd'),
-          endDate: format(subDays(now, 3), 'yyyy-MM-dd'),
+          startDate: '2020-02-03',
+          endDate: '2020-02-04',
```
mantariksh, line 4750: should we save these as a string to enforce consistency between `submission.created` and the query?

mantariksh (approved):

src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
karrui: good point, will do

PR #1605, opened 2021-04-09
chore: uninstall sinon

mantariksh:
## Problem

`sinon` no longer required since `jasmine` has been removed.

## Solution

I think army is not the career for me, it's time to `uninstall-sinon`


mantariksh:
Closing because apparently uninstalling breaks `ts-mock-imports`

PR #1606, opened 2021-04-09
chore: rearrange directory structure

mantariksh:
## Problem

Repo structure is inconsistent in a couple of ways:
- Some backend tests are within `__tests__` folders, while some are in `tests/unit/backend`
- `src/loaders` and `src/config` are outside `src/app`, even though they are used purely on the server

## Solution

- Move all tests into `__tests__` folder so they are as close as possible to the modules being tested.
- Move `src/loaders` and `src/config` into `src/app` and fix all relevant imports and mocks.
- Update the backend structure documented in `ARCHITECTURE.md`.


karrui (commented):
thanks for the cleanup. 
couldn't comment on the PR since they were just renamed, but should `logic.spec.ts` and `file-validation.spec.ts` be in the shared folder tests instead, since the files they are testing are not in `/app/utils`?

lgtm otherwise.

mantariksh:
@karrui good catch!

karrui (approved):
lgtm

PR #1620, opened 2021-04-12
test: fix flaky form feedback test

mantariksh:
## Problem

Form feedback unit tests have been flaking for months because of a non-deterministic `expect` statement. The issue is that form feedback is sorted by creation order when it is retrieved, but the feedback documents in the test are created asynchronously, so sometimes the array of documents is not in the expected order.

## Solution

Sort the array of feedback documents before comparing.

### DIFF
--- a/src/app/modules/feedback/__tests__/feedback.service.spec.ts
+++ b/src/app/modules/feedback/__tests__/feedback.service.spec.ts
@@ -1,4 +1,5 @@
 import { ObjectId } from 'bson-ext'
+import compareAsc from 'date-fns/compareAsc'
 import { times } from 'lodash'
 import moment from 'moment-timezone'
 import mongoose from 'mongoose'
@@ -126,7 +127,8 @@ describe('feedback.service', () => {
       })
       const expectedCreatedFbs = await Promise.all(expectedFbPromises)
       const expectedFeedbackList = expectedCreatedFbs
-        .sort((a, b) => a.created!.getTime() - b.created!.getTime())
+        // Feedback is returned in date order
+        .sort((a, b) => compareAsc(a.created!, b.created!))
         .map((fb, idx) => ({
           index: idx + 1,
           timestamp: moment(fb.created).valueOf(),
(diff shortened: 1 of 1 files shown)


karrui (approved):
><

src/app/modules/feedback/__tests__/feedback.service.spec.ts
```diff
@@ -144,11 +144,11 @@ describe('feedback.service', () => {
...
       ).toFixed(2)
       expect(actualResult.isOk()).toEqual(true)
-      expect(actualResult._unsafeUnwrap()).toEqual({
-        average: expectedAverage,
-        count: expectedCount,
-        feedback: expectedFeedbackList,
-      })
+      expect(actualResult._unsafeUnwrap().average).toEqual(expectedAverage)
+      expect(actualResult._unsafeUnwrap().count).toEqual(expectedCount)
+      expect(actualResult._unsafeUnwrap().feedback.sort()).toEqual(
```
karrui, line 149: actually, do we need to sort the feedback by date? Instead of running sort on the entire array

mantariksh: sure, that would be more efficient

src/app/modules/feedback/__tests__/feedback.service.spec.ts
```diff
@@ -125,15 +125,16 @@ describe('feedback.service', () => {
...
-        index: idx + 1,
-        timestamp: moment(fb.created).valueOf(),
-        rating: fb.rating,
-        comment: fb.comment,
-        date: moment(fb.created).tz('Asia/Singapore').format('D MMM YYYY'),
-        dateShort: moment(fb.created).tz('Asia/Singapore').format('D MMM'),
-      }))
-
+      const expectedFeedbackList = expectedCreatedFbs
+        .sort((a, b) => a.created!.getTime() - b.created!.getTime())
```
karrui, line 129: nit: [date-fns#compareAsc](https://date-fns.org/v2.19.0/docs/compareAsc)?

karrui (approved):
lgtm

PR #1621, opened 2021-04-12
refactor: Minor cleanup to avoid using res.locals

frankchn:
## Problem
Minor cleanup to avoid using `res.locals` in `handleEncryptedSubmission`

## Solution
Removed uses of `res.locals`.


mantariksh (approved):
lgtm, thanks!

PR #1622, opened 2021-04-12
Refactor(admin-form.controller): encrypt preview submission

seaerchin:
## Problem
The current `admin-form/controller` uses `if/else` within the file, which makes it difficult to understand and maintain.

## Solution

Rewrote the `handleEncryptPreviewSubmission` function to use a neverthrow chain, which makes it simpler to understand and parse. 

**Design Considerations** 
1. rewriting to neverthrow chains means that there are no explicit unwrapping and no more "wrapper-variables", whose sole function is to denote a `result` type and then unwrap it. 
2. simpler error handling - `mapErr` at the end makes it clear what happens rather than having control flow cut short halfway within the controller

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -7,6 +7,7 @@ import { ResultAsync } from 'neverthrow'
 import {
   AuthType,
   FieldResponse,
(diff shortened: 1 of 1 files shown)


mantariksh (approved):
nice, probably the only PR ever where no tests had to be changed!

PR #1383, opened 2021-03-16
feat: create new PATCH endpoint for updating form settings

karrui:
Ready for review

## Problem

This PR extracts out updating of form settings into its own endpoint `PATCH /form/:formId/settings`.
In addition, this PR lays the groundwork for our API refactor with the `/api/v3` API base url.

Closes #1355, related to #1210

## Solution

**Features**:

- create new ApiRouter accessed via '/api/v3'
- add backend AdminFormService#updateFormSettings fn
- add settings route handler with Joi validation
- add client AdminFormService for form settings update
- add new FormModel instance method `getSettings`

**Improvements**:

- use new updateFormSettings fn to update settings in client

## Tests

Both unit and integration tests have been added for the new PATCH endpoint.

**Manual tests**
- [ ] Update form settings for an **email** mode form. All possible updates should work:
  - [ ] Activate/deactivate form
  - [ ] Toggle submission limit, change limit number
  - [ ] Set inactive form message
  - [ ] Authentication types,  esrvcId, etc.
  - [ ] Toggle on/off captcha
  - [ ] Update form name
  - [ ] Update emails
- [ ] Update form settings for a **storage** mode form. All possible updates should work:
  - [ ] Activate/deactivate form
  - [ ] Toggle submission limit, change limit number
  - [ ] Set inactive form message
  - [ ] Authentication types,  esrvcId, etc.
  - [ ] Toggle on/off captcha
  - [ ] Update form name
  - [ ] Update webhook URL
- [ ] For any mode form, try with random keys to update settings. Should return 400 error

PR #1633, opened 2021-04-13
feat(admin-forms): implement retrieval of form settings

### LINKED ISSUE #1496: [API] Extract get form settings view endpoint from GET /:formId/adminform

This issue tracks the creation of a new endpoint just to **retrieving the subset relevant for form settings**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
GET /api/v3/admin/forms/:formId/settings
```

> The return shape of this view subset is up to the implementer, as long as the relevant details can still be displayed when rendering form settings.

LoneRifle:
## Problem
the retrieval counterpart to #1383 is missing

Fixes #1496 

## Solution
[form]
- augment `FormSchema.getFullFormById()`,
  which will partially populate an IPopulatedForm with specified keys,
  if given
- feed the `fields` argument supplied in the abovementioned fn as
  the projection arg of `Schema.findById()`
- implement `FormSchema.pick()`, allowing one to pick only the fields
  we want out of a given IFormSchema

[auth-service]
- implement `getFormFieldsAfterPermissionChecks()`, exploiting
  type-generic parameters to allow callers to specify the shape
  of the return value
- take care to only return fields specified by the caller

[admin-form]
- declare and implement `handleGetSettings()` in controller and route,
  taking inspiration from `handleUpdateSettings()`
  - import `FORM_SETTING_FIELDS` from form.server.model to specify the
    form setting fields we need

TODO: Unit-testing

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
@@ -1,7 +1,7 @@
 import { PresignedPost } from 'aws-sdk/clients/s3'
 import { ObjectId } from 'bson-ext'
 import { assignIn, cloneDeep, merge } from 'lodash'
-import { err, errAsync, ok, okAsync } from 'neverthrow'
+import { err, errAsync, ok, okAsync, Result } from 'neverthrow'
 import { PassThrough } from 'stream'
 import { MockedObject } from 'ts-jest/dist/utils/testing'
 import { mocked } from 'ts-jest/utils'
@@ -74,6 +74,7 @@ import {
   PrivateFormError,
   TransferOwnershipError,
 } from '../../form.errors'
+import * as FormService from '../../form.service'
 import * as AdminFormController from '../admin-form.controller'
 import {
   CreatePresignedUrlError,
@@ -105,6 +106,8 @@ jest.mock('src/app/utils/encryption')
 const MockEncryptionUtils = mocked(EncryptionUtils)
 jest.mock('../admin-form.service')
 const MockAdminFormService = mocked(AdminFormService)
+jest.mock('../../form.service')
+const MockFormService = mocked(FormService)
 jest.mock('../../../user/user.service')
 const MockUserService = mocked(UserService)
 jest.mock('src/app/services/mail/mail.service')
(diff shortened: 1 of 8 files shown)


liangyuanruo (dismissed):
As discussed, to separate authorisation concerns from business logic (such as form settings).

I'm also not sure about the whole `pick` thing - I think the _principle of least power_ would discourage such code? unless there is a clear case for its use in future.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1020,6 +1021,52 @@ export const handleUpdateSettings: RequestHandler<
...
+    .andThen((user) =>
+      // Retrieve form for settings as well as for permissions checking
+      FormService.retrieveFullFormById(formId).map((form) => ({
+        form,
+        user,
+      })),
+    )
+    .andThen(AuthService.checkFormForPermissions(PermissionLevel.Read))
+    .map((form) =>
+      res.status(StatusCodes.OK).json(form.getSettings() as IPopulatedForm),
```
mantariksh, line 1052: why is this type cast necessary? I tried removing it and it seems to work fine, plus we've specified in the `RequestHandler` generic args that we are returning a `FormSettings`

src/app/modules/form/form.service.ts
```diff
@@ -112,6 +112,39 @@ export const retrieveFullFormById = (
...
 
+/**
+ * Retrieves the specified form fields of the given formId
+ * @param formId the id of the form
+ * @param fields an array of field names to retrieve
+ * @returns ok(form) if form exists
+ * @returns err(FormNotFoundError) if the form or form admin does not exist
+ * @returns err(DatabaseError) if error occurs whilst querying the database
+ */
+export const retrieveFormFieldsById = (
```
mantariksh, line 123: what's the purpose of this function? doesn't seem to be called anywhere

mantariksh: if we remove it then it seems we can remove a bunch of other changes also

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -4827,6 +4830,255 @@ describe('admin-form.controller', () => {
     })
   })
 
+  describe('handleGetSettings', () => {
+    const MOCK_FORM_SETTINGS: FormSettings = {
+      authType: AuthType.NIL,
+      hasCaptcha: false,
+      inactiveMessage: 'some inactive message',
+      status: Status.Private,
+      submissionLimit: 42069,
```
mantariksh, line 4839: lmao

src/app/modules/form/form.service.ts
LoneRifle: 'twas implementation that I did to facilitate partial retrieval of fields within a form, to conserve bandwidth between app and data store. While not actually used in this PR, I feel that you would find this helpful for subsequent PRs for the API refactor.

mantariksh: I see, in that case might it be better to call it `retrieveFormKeysById`, rather than `FormFields`? I was confused because when I see "form fields" I think form fields as in fields to fill in. obviously this is subjective so I'm ok either way

mantariksh (approved):
lgtm, but we have to remember that `retrieveFormKeysById` exists and use it when the time comes!

src/app/routes/api/v3/admin/forms/admin-forms.settings.routes.ts
```diff
@@ -52,3 +55,19 @@ AdminFormsSettingsRouter.route('/:formId([a-fA-F0-9]{24})/settings').patch(
...
+ * @returns 200 with latest form settings on successful update
+ * @returns 401 when current user is not logged in
+ * @returns 403 when current user does not have permissions to obtain form settings
+ * @returns 404 when form to retrieve settings for cannot be found
+ * @returns 409 when saving form settings incurs a conflict in the database
+ * @returns 500 when database error occurs
+ */
+AdminFormsSettingsRouter.route('/:formId([a-fA-F0-9]{24})/settings').get(
+  handleGetSettings,
+)
```
mantariksh, line 73: sorry last comment - since this and the above endpoint are the same, can we just chain the `patch` and `get` so it's clear that they're different methods on the same endpoint?

LoneRifle: ... How would the jsDoc work though

mantariksh: can search for `AdminFormsRouter.route('/:formId([a-fA-F0-9]{24})/adminform')` to see an example

mantariksh: idk if there's a best practice for this though - I find the chaining to be more clear, but if the folklore says otherwise then it's fine

PR #1640, opened 2021-04-14
feat(api-refactor): add specific API for updating of single form field

### LINKED ISSUE #1481: [API] Extract form field update endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to update form fields. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
`PUT /api/v3/admin/forms/:formId/fields/:fieldId`

karrui:

## Problem

This PR adds a specific endpoint to update a single form field, and updates the frontend client to use that endpoint when updating fields.

Closes #1481

## Solution

**Features**:

- feat(AdminFormService): add updateFormField to update single field
- feat(AdminFormCtl): add handleUpdateFormField fn
- feat: update types of mongo subdocs to allow for DocumentArray
  - this allows for usage of special mongoose array methods such as `id()`, `pull()`, etc
- feat(formUtils): add getFormFieldById helper method
- feat: add Form instance method updateFormFieldById
  - NOTE: This is a change from using the globalId to do field comparisons, and we are doing comparisons with `_id` instead. 
  - This is because there are 0 fields that do not have an _id as queried.
```
db.getCollection('forms')
    .find({ 'form_fields._id': { $exists: false }, 'form_fields': { $ne: [] } })
    .limit(10)
```

**Improvements**:

- return http 422 status code on EditFieldError instead of 400

## Tests

Integration tests will be added in a separate PR

- test(AdminFormSvc): add unit tests for updateFormField fn
- test(AdminFormCtl): add unit tests for handleUpdateFormField
- test(FormModel): add unit tests for updateFormFieldById instance fn

## Manual tests
- [ ] Create a new field and then edit and save it. The field should successfully be updated (go into the field and check again)
  - [ ] Update image fields (change image). Should update correctly.
  - [ ] Update attachment fields (size, description, etc). Should update correctly.
  - [ ] Try a sample various variants of other fields. All should be successfully saved.
- [ ] Create a new myinfo field and then edit and save it. The field should be "updated" without changes


src/app/modules/form/form.utils.ts
```diff
@@ -76,3 +79,37 @@ export const isEmailModeForm = (
...
+ * @param formFields the form fields to search from
+ * @param fieldId the id of the field to retrieve
+ * @returns the form field if found, `null` otherwise
+ */
+export const getFormFieldById = (
+  formFields: IFormSchema['form_fields'],
+  fieldId: IFieldSchema['_id'],
+): IFieldSchema | null => {
+  if (!formFields) {
+    return null
```
seaerchin, line 107: should we consider changing this to return a custom error (eg: `FieldNotFoundError`)? There's only 1 callsite in `FormDocumentSchema.methods.updateFormFieldById` and if this change is made, i think we could rewrite that as a chained method

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1266,3 +1322,42 @@ export const handleEmailPreviewSubmission: RequestHandler<
...
+        disabled: Joi.boolean().required(),
+        // Allow other field related key-values to be provided and let the model
+        // layer handle the validation.
+      }).unknown(true),
+    },
+    undefined,
+    // Required so req.body can be validated against values in req.params.
+    // See https://github.com/arb/celebrate#celebrateschema-joioptions-opts.
+    { reqContext: true },
+  ),
```
seaerchin, line 1361: should we extract this into a variable so that the name can provide context for this validator?

karrui: The validator being right before the controller function arguably has better context than moving the validator to a constant

src/app/modules/form/admin-form/admin-form.errors.ts
```diff
@@ -17,3 +17,9 @@ export class EditFieldError extends ApplicationError {
     super(message)
   }
 }
+
+export class FieldNotFoundError extends ApplicationError {
```
seaerchin, line 21: i think the callsites for `FieldNotFoundError` have a variable called `fieldId` in scope and if we are to log this, should we also attach the variable `fieldId` to the constructor to aid in troubleshooting?

src/types/field/index.ts
```diff
@@ -19,3 +39,50 @@ export * from './shortTextField'
 export * from './statementField'
 export * from './tableField'
 export * from './yesNoField'
+
+export type FormFieldSchema =
```
seaerchin, line 43: thx 4 ur hard work...

src/app/modules/form/admin-form/admin-form.controller.ts
seaerchin: fair point!

src/app/modules/form/form.utils.ts
karrui: this util is actually used in multiple places; the Form instance method and in admin form service.

src/app/modules/form/admin-form/admin-form.errors.ts
karrui: no need; knowing the fieldId won't help since the field is not found in the database :P

src/app/modules/form/form.utils.ts
```diff
@@ -76,3 +79,37 @@ export const isEmailModeForm = (
...
+  }
+
+  if (isMongooseDocumentArray(formFields)) {
+    return formFields.id(fieldId)
+  } else {
+    return formFields.find((f) => fieldId === String(f._id)) ?? null
+  }
```
mantariksh, line 114: since when did you use `else`??? lol

src/public/modules/forms/admin/constants/update-form-types.js
```diff
@@ -0,0 +1,5 @@
+const UPDATE_FORM_TYPES = {
```
mantariksh, line 1: this file can be in TS right? it'll save us some trouble later on

src/public/modules/forms/admin/controllers/admin-form.client.controller.js
```diff
@@ -179,13 +181,39 @@ function AdminFormController(
...
+              fieldId,
+              body,
+            ),
+          )
+          .then((updatedFormField) => {
+            // merge back into the form fields
+            const updateIndex = $scope.myform.form_fields.findIndex(
+              (f) => f._id === fieldId,
+            )
+            if (updateIndex !== -1) {
```
mantariksh, line 202: should we throw an error here if `updateIndex === -1`?

src/public/modules/forms/admin/constants/update-form-types.js
karrui: True, but probably won't use this file in our react rewrite. Still will change to TS anyways

src/app/modules/form/form.utils.ts
```diff
@@ -76,3 +79,37 @@ export const isEmailModeForm = (
...
+ * @param array the array to check
+ */
+export const isMongooseDocumentArray = <T extends Document>(
+  array: T[] & { isMongooseDocumentArray?: boolean },
+): array is Types.DocumentArray<T> => {
+  /**
+   * @see {mongoose.Types.DocumentArray.isMongooseDocumentArray}
+   */
+  return !!array.isMongooseDocumentArray
+}
```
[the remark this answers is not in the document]
karrui: Fixed in 8344f3a

src/app/modules/form/form.utils.ts
karrui: i am disgusted with myself; fixed in ea629fd

src/public/modules/forms/admin/constants/update-form-types.js
karrui: 8e20006

src/public/modules/forms/admin/controllers/admin-form.client.controller.js
karrui: a58c9d2

mantariksh (approved):
remember to write extensive manual tests!

karrui:
> remember to write extensive manual tests!

Idk what extensive means haha, I can only think of checking whether updates go through lol

mantariksh:
@karrui I was thinking we should go through every single field and check every single setting

karrui:
> @karrui I was thinking we should go through every single field and check every single setting

Noted. Will write it down

Issue #1520, opened 2021-03-30
[API] Migrate GET submission metadata related endpoints to /api counterpart


karrui:
This issue tracks the migration of `GET /:formId/adminform/submissions/` and `GET /:formId/adminform/submissions/metadata` to their new `/api` counterparts.

This issue is distinct from #1511 due to the request query parameter shifting into the endpoint itself, which requires a new controller to be written instead of just duplicating the route. 
This change is proposed as the current endpoints are not RESTful, as queries should be for properties of a resource, not the identifier for a resource.

\# | Existing endpoints | New endpoints<br>(note that all routes are prefixed with /api/v3) | Description
-- | -- | --|--|
1 | GET /:formId/adminform/submissions/<br>{ query: { submissionId } } | GET /admin/forms/:formId/submissions/:id | Retrieve submission by id
2 | GET /:formId/adminform/submissions/metadata<br>{ query: { submissionId, page } } | GET /admin/forms/:formId/submissions/metadata<br>{ query: { searchId?, page = 1 } } | Search by submission

> For the first endpoint, note the inlining of `query.submissionId` to the resource identifier `:id` in the endpoint itself.

The second endpoint just has a change in query params from `submissionId` to `searchId` for better clarity as to what the query is used for

seaerchin:
@karrui i will name the new controller `handleGetEncryptedResponseWithUrlParams` just to be clear. this is because we can't delete the old one so the distinction should be made explicit ;--;

seaerchin:
as discussed, for the second point, will only migrate the `routes` but not the actual `controller` for 2 reasons: 

1. FE is now doing a search on an exact match rather than a substring search so the renaming might not be very applicable 
2. Code duplication on controller might not be worth the trouble

seaerchin:
@karrui could i get some clarification on the intent behind the page number defaulting to 1? Currently, the validator ensures that page number is set only when `submissionId` is not present. however, this leads to a case where users might have a less than pleasant experience because they will hit a 400 error then they do not specify both submissionId and page number. 

Hence, if the intention behind the parameter is to be more permissive to the users, should we delete the validation altogether and set a default parameter on the pageNumber instead?

seaerchin:
clarified the above issue - validation is still needed to ensure the existence of at least one of the two parameters and to prevent additional parameters from being queried by the client

PR #1645, opened 2021-04-14
refactor: migrate GET /submission to new endpt

seaerchin:
## Problem
This is part 1 of #1520. This PR addresses the fact that URL query parameters should not be used to identify a resource. rather, they should be used to identify _properties_ of a resource. Hence, there is a new controller + endpoint set up so that the api endpoint is restful.

## Solution
1. Remove request parameter from `RequestHandler` and shift it to the url isntead. 
2. Ports over existing controller unit tests and changes it so that the parameter is in the url 
3. Ports over existing integration tests and changes it so that the parameter is in the url

## Manual Tests
1. Create a new storage mode form and make it public. Go to the form and submit a few responses. Go to the admin panel and click on data and click on any response. The panel should be populated with the information of the responses

## Notes 
1. Shifted from `/admin/forms` to`${ADMIN_FORMS_PREFIX}` in tests. Maybe this should have been done in a separate PR? Felt that it was small (and related to the current PR) so i did it here but comments welcome

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -604,6 +604,82 @@ export const handleGetEncryptedResponseUsingQueryParams = [
 ] as RequestHandler[]
 
 /**
+ * Handler for GET /:formId/submissions/submissionId
+ * NOTE: This is exported solely for testing
+ * @security session
+ *
+ * @returns 200 with encrypted submission data response
+ * @returns 400 when form is not an encrypt mode form
+ * @returns 403 when user does not have read permissions for form
+ * @returns 404 when submissionId cannot be found in the database
+ * @returns 404 when form cannot be found
+ * @returns 410 when form is archived
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when any errors occurs in database query or generating signed URL
+ */
+export const getEncryptedResponse: RequestHandler<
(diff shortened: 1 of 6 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -586,6 +595,81 @@ export const handleGetEncryptedResponse: RequestHandler<
   )
 }
 
+// Exported as an array to ensure that the handler always a valid submissionId
+export const handleGetEncryptedResponseUsingQueryParams = [
```
karrui, line 599: can we mark this as `@deprecated` and point it to our new api end point? Now that I think about it, we should have marked all our migrated endpoints as deprecated

karrui (approved):
lgtm, pls add deprecate jsdoc

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
seaerchin: good point and i'll do so! for the pointing, i'll put the absolute path there just for maximum clarity. will do so for currently open PRs by me

seaerchin: @karrui actually i think i'll do this on the routes. makes more sense rather than placing it on the controller; if there are new controllers (like what we have here), i'll add the @deprecate but point it to the new controller

PR #1651, opened 2021-04-15
refactor: migrate submissions metadata

### LINKED ISSUE #1520: [API] Migrate GET submission metadata related endpoints to /api counterpart

seaerchin:
This is part 2 of #1520. 

## Problem
This PR migrates the old API to the new API rooted at `/api/v3/admin/forms`. The current implementation is ported over as is and something to note is that currently, we are not validating that there is either submission or formId, and that formId should be at least 1. 

This will be rectified in a future PR to avoid scope creep here and this PR should be reviewed as an almost direct port over. 

Closes #1520 

## Solution
1. Adds new endpoint to `AdminFormSubmissions` router for metadata and changed client callsites
2. Ported over integration tests

## Manual Tests

- [ ] Create a storage mode form and activate it. Submit **>10** responses to the form. Go to the admin panel and click on data
    - Verify that clicking on any page returns a list of responses  
    - Verify that searching on any single response id returns exactly that response   
    - Verify that searching on any substring of a response id returns nothing

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -5,7 +5,7 @@ import { Query } from 'express-serve-static-core'
 import { StatusCodes } from 'http-status-codes'
 import JSONStream from 'JSONStream'
 import mongoose from 'mongoose'
-import { SetOptional } from 'type-fest'
+import { RequireAtLeastOne, SetOptional } from 'type-fest'
 
 import {
   AuthType,
@@ -681,8 +681,21 @@ export const handleGetEncryptedResponse: RequestHandler<
   )
 }
 
+// NOTE: If submissionId is set, then page is optional.
+// Otherwise, if there is no submissionId, then page >= 1
+const validateSubmissionIdOrPageNum = celebrate({
+  [Segments.QUERY]: {
+    submissionId: Joi.string().optional(),
+    page: Joi.number().min(1).when('submissionId', {
+      not: Joi.exist(),
+      then: Joi.required(),
+    }),
+  },
+})
+
 /**
- * Handler for GET /:formId([a-fA-F0-9]{24})/adminform/submissions/metadata
+ * Handler for GET /:formId/submissions/metadata
+ * This is exported solely for testing purposes
  *
  * @returns 200 with single submission metadata if query.submissionId is provided
  * @returns 200 with list of submission metadata with total count (and optional offset if query.page is provided) if query.submissionId is not provided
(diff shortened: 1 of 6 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -693,15 +706,19 @@ export const handleGetEncryptedResponse: RequestHandler<
...
+  Query &
+    RequireAtLeastOne<
+      { page?: number; submissionId?: string },
+      'page' | 'submissionId'
+    >
 > = async (req, res) => {
   const sessionUserId = (req.session as Express.AuthedSession).user._id
   const { formId } = req.params
-  const { page, submissionId } = req.query
+  const { page = 1, submissionId } = req.query
```
mantariksh, line 721: can we restrict the scope of this PR to the API refactor only? let's not repeat the mistakes of the previous production bug by mixing refactors and logic changes

mantariksh (commented):
tested on staging?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
seaerchin: sure, i'll revert this change to adhere to before

seaerchin:
@mantariksh tested on local, haven't tested on staging. will borrow staging tomorrow and test

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -681,8 +681,21 @@ export const handleGetEncryptedResponse: RequestHandler<
   )
 }
 
+// NOTE: If submissionId is set, then page is optional.
+// Otherwise, if there is no submissionId, then page >= 1
+const validateSubmissionIdOrPageNum = celebrate({
```
karrui, line 686: nit: if this is only used in one place, can we inline it instead?

karrui (approved):
lgtm, can squash after fixing PR according to anti's comments

mantariksh:
pls test on staging before merging

seaerchin:
I got 429 `too many requests` while testing. this occurred at the start of testing, when clicking on teh second page of data. Didn't occur later on after refreshing the page and retrying. 

I don't think the error is due to this PR and more likely due to middleware set up for rate limiting.

seaerchin:
tested on staging again; no issues found. will merge into develop

PR #1663, opened 2021-04-16
fix: loosen validation for sg home numbers

tshuli:
## Problem

Fixes the problem of sg home numbers starting with '666' being incorrectly rejected.

## Test

- [ ] Create form with home number field. Check that the number `66665555` can be successfully submitted.

### DIFF
--- a/src/shared/util/phone-num-validation.ts
+++ b/src/shared/util/phone-num-validation.ts
@@ -58,6 +58,10 @@ export const isHomePhoneNumber = (phoneNum: string): boolean => {
   if (!parsedNumber) return false
 
   // For SG numbers only check if it starts with 3 or 6 and has 8 digits
+  // Leading number 3 is for IP Telephony (IPT) service and User-Centric Data-Only (UCDO) service
+  // Leading number 6 is for  PSTN service and IP Telephony (IPT) service.
+  // In both cases the length is 8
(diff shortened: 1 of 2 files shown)


src/shared/util/phone-num-validation.ts
```diff
@@ -54,11 +54,21 @@ export const isMobilePhoneNumber = (mobileNumber: string): boolean => {
  */
 export const isHomePhoneNumber = (phoneNum: string): boolean => {
   const parsedNumber = parsePhoneNumberFromString(phoneNum)
+
   if (!parsedNumber) return false
 
+  // For SG numbers only check if it starts with 3 or 6 and has 8 digits
+  if (parsedNumber.countryCallingCode === '65') {
```
mantariksh, line 61: is it possible to reuse `startsWithSgPrefix` here?

src/shared/util/phone-num-validation.ts
```diff
@@ -54,11 +54,21 @@ export const isMobilePhoneNumber = (mobileNumber: string): boolean => {
...
+    return (
+      isPhoneNumber(phoneNum) &&
+      // Regex checks if the national number starts with 3 or 6, and is of
+      // length 8.
```
mantariksh, line 65: where do these rules come from? eg how do we know that 3 is allowed? and should we document the source?

src/shared/util/phone-num-validation.ts
tshuli: edited

src/shared/util/phone-num-validation.ts
tshuli: added

tshuli:

mantariksh (approved):

PR #1664, opened 2021-04-17
feat: Add a limit for the number of MyInfo fields that can be added

### LINKED ISSUE #1305: Limit to 30 MyInfo fields
MyInfo fields must be limited to 30 per form on Email mode to comply with PII guidelines. In exchange, we get access to all fields.

frankchn:
## Problem
MyInfo fields must be limited to 30 per form on Email mode to comply with PII guidelines. In exchange, we get access to all fields.

Closes #1305

## Solution
Adds limits and validators to the appropriate places to limit MyInfo fields to 30.

## Screenshots
image

image

## Tests
- [ ] Add 30 MyInfo fields to a form.
- [ ] Check that under the build tab, MyInfo fields can no longer be added
- [ ] Check that MyInfo fields cannot be duplicated
- [ ] Delete a Myinfo field (such that the form now has <30) and check that you can now add and duplicate Myinfo fields
- [ ] Manually add an additional MyInfo field to the form by editing the database. Check that when you try to edit the form subsequently, you get an error message saying to check that the form has max 30 MyInfo fields.
- [ ] Run the following query to ensure that there are no existing forms with >30 MyInfo fields:
```
db.getMongo().setReadPref('secondary')
db.getCollection('forms').aggregate([
    { $match: { 'form_fields.myInfo': { $exists: true } } },
    { $unwind: '$form_fields' },
    { $match: { 'form_fields.myInfo': { $exists: true } } },
    { $group: { _id: '$_id', count: { $sum: 1 } } },
    { $match: { count: { $gt: 30 } } }
])
```
- [ ] **After release has been deployed**, run the above query again to check that there are still no forms with >30 MyInfo fields. If there are, the form admin must be contacted.

### DIFF
--- a/src/app/models/field/baseField.ts
+++ b/src/app/models/field/baseField.ts
@@ -22,14 +22,24 @@ export const MyInfoSchema = new Schema<IMyInfoSchema>(
       enum: Object.values(MyInfoAttribute),
       validate: {
         validator: function (this: IMyInfoSchema) {
-          const { authType, responseMode } = this.ownerDocument()
+          const form = this.ownerDocument()
+          const { authType, responseMode } = form
+          let myInfoCount = 0
+          if (form.form_fields !== undefined) {
+            form.form_fields.forEach((field) => {
+              if (field.myInfo !== undefined) {
+                myInfoCount++
+              }
+            })
+          }
           return (
             authType === AuthType.MyInfo &&
-            responseMode !== ResponseMode.Encrypt
+            responseMode !== ResponseMode.Encrypt &&
+            myInfoCount <= 30
           )
         },
         message:
-          'MyInfo field is invalid. Check that your form has MyInfo enabled, or is not an encrypted mode form.',
+          'MyInfo field is invalid. Check that your form has MyInfo enabled, or is not an encrypted mode form, and has 30 or fewer MyInfo fields.',
       },
     },
   },
--- a/src/app/models/field/baseField.ts
+++ b/src/app/models/field/baseField.ts
@@ -2,7 +2,6 @@ import { Schema } from 'mongoose'
 import UIDGenerator from 'uid-generator'
 
 import {
-  AuthType,
   BasicField,
   IFieldSchema,
   IMyInfoSchema,
@@ -20,27 +19,6 @@ export const MyInfoSchema = new Schema<IMyInfoSchema>(
     attr: {
       type: String,
       enum: Object.values(MyInfoAttribute),
-      validate: {
-        validator: function (this: IMyInfoSchema) {
-          const form = this.ownerDocument()
-          const { authType, responseMode } = form
-          let myInfoCount = 0
-          if (form.form_fields !== undefined) {
-            form.form_fields.forEach((field) => {
-              if (field.myInfo !== undefined) {
-                myInfoCount++
-              }
-            })
-          }
-          return (
-            authType === AuthType.MyInfo &&
-            responseMode !== ResponseMode.Encrypt &&
-            myInfoCount <= 30
-          )
-        },
-        message:
-          'MyInfo field is invalid. Check that your form has MyInfo enabled, or is not an encrypted mode form, and has 30 or fewer MyInfo fields.',
-      },
     },
   },
   {
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -586,6 +586,37 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       return next(err)
     }
 
+    // Check if MyInfo fields exist and if they do, the total number of fields <= 30, and that
+    // the response mode and authType fields are set appropriately.
+    let myInfoFieldCount = 0
+    if (this.form_fields !== undefined) {
+      this.form_fields.forEach((field) => {
+        if (field.myInfo !== undefined) {
+          myInfoFieldCount++
+        }
+      })
+    }
+
+    if (
+      myInfoFieldCount > 0 &&
+      (this.authType !== AuthType.MyInfo ||
+        this.responseMode !== ResponseMode.Encrypt)
+    ) {
+      const validationError = this.invalidate(
+        'form_fields',
+        'This form contains MyInfo fields but does not have MyInfo enabled or is a storage mode form.',
+      ) as mongoose.Error.ValidationError
+      return next(validationError)
+    }
+
+    if (myInfoFieldCount > 30) {
+      const validationError = this.invalidate(
+        'form_fields',
+        'This form contains too many MyInfo fields. A maximum of 30 MyInfo fields are allowed.',
+      ) as mongoose.Error.ValidationError
+      return next(validationError)
+    }
+
     // Validate that admin exists before form is created.
     return User.findById(this.admin).then((admin) => {
       if (!admin) {
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -600,7 +600,7 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     if (
       myInfoFieldCount > 0 &&
       (this.authType !== AuthType.MyInfo ||
-        this.responseMode !== ResponseMode.Encrypt)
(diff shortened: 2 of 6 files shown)


src/public/modules/forms/admin/directives/edit-form.client.directive.js
```diff
@@ -442,6 +468,11 @@ function editFormController(
   }
 
   const duplicateField = (fieldToDuplicate) => {
+    if (
+      fieldToDuplicate.myInfo !== undefined &&
+      $scope.numMyInfoFields >= $scope.maxMyInfoFields
+    )
+      return
```
mantariksh, line 475: what's the UX here if they attempt to duplicate a MyInfo field when they already have 30? should we show some sort of error message (maybe a Toastr)?

src/public/modules/forms/admin/directiveViews/edit-form.client.view.html
```diff
@@ -304,6 +329,7 @@
                   <button
                     class="current-field-opt-icon"
                     ng-click="tryDuplicateField($index, isCondition(field) || isHidden(field)); $event.stopPropagation()"
+                    ng-show="numMyInfoFields < maxMyInfoFields"
```
frankchn, line 332: Make sure only disable the duplication on the MyInfo field.

src/public/modules/forms/admin/directives/edit-form.client.directive.js
frankchn: The button disappears when they are on 30, so this code shouldn't actually trigger.

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -287,13 +294,27 @@ export const createRelayState = (formId: string): string =>
...
+  if (form.form_fields !== undefined) {
+    form.form_fields.forEach((field) => {
+      if (field.myInfo !== undefined) {
+        count++
+      }
+    })
+  }
+  if (count > 30) {
+    return err(new MyInfoTooManyFieldsError())
+  }
```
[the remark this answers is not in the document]
frankchn: moving to MyInfo field schema validation.

mantariksh (commented):
thanks @frankchn! this looks good, but I'll need to write and run a database query first to check that there aren't already forms with >30 MyInfo fields. ~if the current database state checks out then I'll approve and merge this.~ pending last comment below

src/app/models/field/baseField.ts
```diff
@@ -22,14 +22,24 @@ export const MyInfoSchema = new Schema<IMyInfoSchema>(
...
+                myInfoCount++
+              }
+            })
+          }
           return (
             authType === AuthType.MyInfo &&
-            responseMode !== ResponseMode.Encrypt
+            responseMode !== ResponseMode.Encrypt &&
+            myInfoCount <= 30
           )
```
mantariksh, line 39: I'm wondering whether we can improve this chunk of code by moving it into the validation for the Form schema. I think that's where this code really belongs, as this is really form-level validation ("every form should have no more than 30 MyInfo fields"), not field-level validation. This would have the added advantages of 1) running the validator only once per form, rather than once per field and 2) removing the need for breaking encapsulation by calling `this.ownerDocument()`. what do you think?

liangyuanruo:
> thanks @frankchn! this looks good, but I'll need to write and run a database query first to check that there aren't already forms with >30 MyInfo fields. if the current database state checks out then I'll approve and merge this.

Update on behalf on @mantariksh - there aren't any such forms, so the PR is safe.

src/app/models/field/baseField.ts
frankchn: Makes sense, moving this to the main form schema to validate before saving.

src/app/models/form.server.model.ts
```diff
@@ -586,6 +586,37 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+        if (field.myInfo !== undefined) {
+          myInfoFieldCount++
+        }
+      })
+    }
+
+    if (
+      myInfoFieldCount > 0 &&
+      (this.authType !== AuthType.MyInfo ||
+        this.responseMode !== ResponseMode.Encrypt)
```
mantariksh, line 603: I think this needs to be flipped
```suggestion
        this.responseMode === ResponseMode.Encrypt)
```

src/app/models/form.server.model.ts
```diff
@@ -586,6 +586,37 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+      ) as mongoose.Error.ValidationError
+      return next(validationError)
+    }
+
+    if (myInfoFieldCount > 30) {
+      const validationError = this.invalidate(
+        'form_fields',
+        'This form contains too many MyInfo fields. A maximum of 30 MyInfo fields are allowed.',
+      ) as mongoose.Error.ValidationError
+      return next(validationError)
```
mantariksh, line 617: actually should we change this to be a key-specific validator for `form_fields` instead? I think this pre-validate hook looks at risk of becoming a mash of unrelated functionality. you can look at this chunk of code in `src/app/models/field/tableField.ts` for an example of how to access `this` in a key-specific validator:
```
maximumRows: {
      type: Number,
      min: 2,
      validate: {
        validator: function (this: ITableFieldSchema, v?: number) {
          return !v || v > this.minimumRows
        },
        message: 'Maximum number of rows must be greater than minimum.',
      },
    },
```

frankchn: Currently `form_fields` is defined as `form_fields: [BaseFieldSchema]` and I am not sure you can add a validator there easily?

So the idea is to move this code into `BaseFieldSchema`, but instead of sticking it into https://github.com/opengovsg/FormSG/blob/32a18fc781a125046d343ede2fed03205533be42/src/app/models/field/baseField.ts#L73-L88, you want to create another property of BaseFieldSchema and add the validator for the count there? 

Isn't that more complicated than either adding the validator to `MyInfoSchema` (what was originally written) or leaving it here?

mantariksh: the current code is this:
```
form_fields: [BaseFieldSchema],
```

since we're performing validation on form fields, I'm thinking something we can change it to something like this:
```
form_fields: {
        type: [BaseFieldSchema],
        validate: {
          validator: function (this: IFormSchema) {
            const myInfoFieldCount = (this.form_fields ?? []).reduce(
              (acc, field) => acc + (field.myInfo ? 1 : 0),
              0,
            )
            if (myInfoFieldCount === 0) return true
            return (
              this.authType === AuthType.MyInfo &&
              this.responseMode === ResponseMode.Email &&
              myInfoFieldCount <= 30
            )
          },
          message:
            'Check that your form is Myinfo-authenticated, is an email mode form and has 30 or fewer MyInfo fields.',
        },
      },
```

what do you think? the idea is that since the validation is specific to `form_fields`, we should add it to the validator for that particular key.

frankchn: Yeah I guess this is fair. Let me make the change.

src/public/modules/forms/admin/directiveViews/edit-form.client.view.html
```diff
@@ -304,6 +329,7 @@
                   <button
                     class="current-field-opt-icon"
                     ng-click="tryDuplicateField($index, isCondition(field) || isHidden(field)); $event.stopPropagation()"
+                    ng-show="field.myInfo && numMyInfoFields < maxMyInfoFields"
```
mantariksh, line 332: this disallows duplication of non-MyInfo fields. I think this might need to change to the following:
```suggestion
                    ng-show="!field.myInfo || numMyInfoFields < maxMyInfoFields"
```

mantariksh (commented):
one remaining bug, otherwise looks good!

src/public/modules/forms/admin/directiveViews/edit-form.client.view.html
frankchn: Good catch!

mantariksh (approved):

PR #1669, opened 2021-04-19
refactor: shard public forms router

seaerchin:
## Problem
`PublicFormsRouter` is getting too big and we should shard each sub component out into their own folder for greater maintainability. This follows the `AdminFormsRouter` hierarchy. 

## Solution
Shard them out into their own folder and shards tests also.


mantariksh (approved):

src/app/routes/api/v3/public/public-forms.feedback.routes.ts
```diff
@@ -0,0 +1,23 @@
+import { Router } from 'express'
```
karrui, line 1: i didn't notice the file naming in the previous PR; the directory structure should mimic the real endpoint path. Can we change the `/v3/public/` directory to `/v3/forms/` directory instead?

karrui (changes requested):
we should change the directory name

karrui (approved):
lgtm

PR #1671, opened 2021-04-19
feat(api-refactor): implement specific create field api

### LINKED ISSUE #1484: [API] Extract form field creation endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **create new form fields**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
POST /api/v3/admin/forms/:formId/fields/
```

> Note that the proposed endpoint is `POST` and not `PUT`, since field creation is not idempotent

karrui:

## Problem

This PR adds a specific endpoint to create a single form field, and updates the frontend client to use that endpoint when creating fields.

Closes #1484

## Solution

**Features**:
- feat(AdminFormCtl): add handleCreateFormField controller and router fn
- feat(FormModel): add insertFormField model instance method
  - Unable to use model static function due to usage of `this.getParent()` method in some field validators which is not available when using MongoDB's `update` methods.
  - NOTE: This is a change from using the globalId to do field comparisons, and we are doing comparisons with _id instead.
- feat(client): impl createSingleFormField service fn for field creation

## Tests
- test(AdminFormCtl): add unit tests for _handleCreateFormField fn
- test(AdminFormSvc): add unit tests for createFormField fn
- test(FormModel): add unit tests for insertFormField instance method

## Manual tests
- [ ] Create a new field. Should be created successfully. Network tab should show a call to the new endpoint.
- [ ] Create a new MyInfo field. Should be created successfully. Network tab should show a call to the new endpoint.

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -1566,10 +1566,10 @@ export const handleCreateFormField = [
       fieldType: Joi.string()
         .valid(...Object.values(BasicField))
         .required(),
-      description: Joi.string().allow('').required(),
-      required: Joi.boolean().required(),
       title: Joi.string().required(),
-      disabled: Joi.boolean().required(),
+      description: Joi.string().allow(''),
+      required: Joi.boolean(),
+      disabled: Joi.boolean(),
       // Allow other field related key-values to be provided and let the model
       // layer handle the validation.
     }).unknown(true),
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -1562,6 +1562,7 @@ export const handleCreateFormField = [
     [Segments.BODY]: Joi.object({
       // Ensures id is not provided.
       _id: Joi.any().forbidden(),
+      globalId: Joi.any().forbidden(),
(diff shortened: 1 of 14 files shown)


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1449,3 +1450,83 @@ export const handleUpdateFormField = [
...
+      // Ensures id is not provided.
+      _id: Joi.any().forbidden(),
+      globalId: Joi.any().forbidden(),
+      fieldType: Joi.string()
+        .valid(...Object.values(BasicField))
+        .required(),
+      title: Joi.string().required(),
+      description: Joi.string().allow(''),
+      required: Joi.boolean(),
+      disabled: Joi.boolean(),
```
mantariksh, line 1526: not sure but should these be required?

mantariksh (approved):

src/app/modules/form/admin-form/admin-form.controller.ts
karrui: These params have a model default already (yea, probably can set the update field one to not be required too)

karrui:
I see that tests are in another PR; I think unit tests should be contained in the same PR tbh

Issue #1517, opened 2021-03-30
[API] Collapse SP/CP/MyInfo login endpoints to new /api counterpart


karrui:
This issue tracks the collapsing of `/spcp`, `/myinfo` related endpoints to their new `/api/forms/:formId/auth` counterparts.

The current `/spcp` routes require the client  to pass in information such as the `target` endpoint, the current `authType`, and the `esrvcId` of the current form. However, all this information should be retrieved from the backend itself, and the new endpoint should only accept `formId` as a parameter, like the current `/myinfo` routes for consistency.

This also allows for the form's `authType` to be obtained from the form referenced by its id, which is the SSOT.

Collapsing all the various endpoints to just two endpoints (`/redirect` and `/validate`) also allows for less code to maintain and increases the consistency in the APIs.

The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

The current to new mapping is shown below:

\# | Existing endpoints | New endpoints<br>(note that all routes are prefixed with /api/v3) | Description
-- | -- | --|--|
20 | GET /spcp/redirect<br>query: { target, authType, esrvcId } | GET /forms/:formId/auth/redirect| redirect to SPCP endpoint to login
21 | GET /spcp/validate <br>query: { target, authType, esrvcId } | GET  /forms/:formId/auth/validate| gets the spcp redirect URL and parses the returned page to check for error codes
22 | GET /myinfo/redirect<br>query: {formId} | GET /forms/:formId/auth/redirect
23 | GET /myinfo/validate<br>query: {formId} | GET /forms/:formId/auth/validate

seaerchin:
@karrui clarifications on this issue: this is actually a few things right?

1. collapse spcp
2. collapse myinfo
3. migrate to new router

seaerchin:
@karrui should we omit the `formId` query parameter since it is already part of the url path?

karrui:
> @karrui should we omit the `formId` query parameter since it is already part of the url path?

good point, yea no need thanks for the catch; but add `persistantLogin` query param, to be used my spcp auth

PR #1672, opened 2021-04-19
feat(api): collapse spcp/myinfo redirect endpoint to new endpoint

seaerchin:
Part 1 of #1517 

## Problem
Previously, our `/auth/redirect` endpoints were split between the `MyInfo` and `spcp` sections, which made code harder to maintain and our frontend had separated logic. 

Separately, this is also part of the `/api/v3` migration to shift endpoints to their new homes

## Solution
This PR merges the 2 endpoints into 1 and shifts it into a new route rooted at `/api/v3`

**Features**:

1. Created a new controller method under `auth.controller` to handle the redirects for backend
2. Created a new router to service the new `/api/v3` routes
3. New `AuthService` on frontend side as part of react migration 

**Notes**
1. there is no decision making on the frontend now. instead, decisions are deferred to the backend and the frontend just transparently sends across its data
2. `MyInfoService` will be fully deleted in part 2 of this PR. it is left as is first. 

## Tests
Tests are [here](https://github.com/opengovsg/FormSG/pull/1685) to avoid review fatigue

## Manual Tests
1. Create a SP form. Click on activate and go to the form. Click on keep me logged in. Clicking on sign in with singpass should redirect you to the Singpass page. Ensure that after signing in, refreshing the page still keeps you signed in.
2. Create a SP form. Click on activate and go to the form. Clicking on sign in with singpass should redirect you to the Singpass page.
3. Create a email mode form with authentication set to myInfo. Click on activate and go to the form. Clicking on sign in with singpass should redirect to the singpass page.


src/app/modules/auth/auth.controller.ts
```diff
@@ -217,3 +227,92 @@ export const handleSignout: RequestHandler = async (req, res) => {
...
+ * @param isPersistentLogin whether the client wants to have their login information stored
+ * @returns 200 with the redirect url when the user authenticates successfully
+ * @returns 400 when there is an error on the authType of the form
+ * @returns 400 when the eServiceId of the form does not exist
+ * @returns 404 when form with given ID does not exist
+ * @returns 500 when database error occurs
+ * @returns 500 when the redirect url could not be created
+ * @returns 500 when the redirect feature is not enabled
+ */
+export const getRedirectLink: RequestHandler<
```
karrui, line 243: I think this function should be in the public-form controller instead? The auth controller is for admin login and logout

src/app/modules/auth/auth.controller.ts
```diff
@@ -217,3 +227,92 @@ export const handleSignout: RequestHandler = async (req, res) => {
...
+/**
+ * Handler for /forms/:formId/auth/redirect
+ */
+export const handleRedirect = [
+  celebrate({
+    [Segments.QUERY]: Joi.object({
+      isPersistentLogin: Joi.boolean().optional(),
+    }),
+  }),
+  getRedirectLink,
```
karrui, line 315: I think we should rename this to _handleFormAuthRedirect too, with the underscore making it explicit that it should not be used and is "private".

src/app/modules/auth/auth.utils.ts
```diff
@@ -63,3 +74,60 @@ export const getUserIdFromSession = (
 ): string | undefined => {
   return session?.user?._id as string | undefined
 }
+
+/**
+ * Maps errors from creating a redirect URL to status codes and error messages to return to frontend.
+ * @param error
+ */
+export const mapRedirectUrlError: MapRouteError = (
```
karrui, line 82: Same here, this whole function should be in public form utils; and should we also rename it to `mapFormAuthRedirectRouteError`.

Actually, is there a reason why we do not use the usual `mapRouteError`?

src/app/modules/spcp/spcp.types.ts
```diff
@@ -1,3 +1,5 @@
+import { AuthType, IFormSchema } from 'src/types'
```
karrui, line 1: nit: use relative path

src/app/modules/spcp/spcp.types.ts
```diff
@@ -43,3 +45,8 @@ export interface ParsedSpcpParams {
   rememberMe: boolean
   cookieDuration: number
 }
+
+export interface ISPCPForm extends IFormSchema {
```
karrui, line 49: nit: rename to `ISpcpForm`, we tend to not use UPPERCASE for abbreviations/initialisms

src/public/services/AuthService.ts
```diff
@@ -0,0 +1,22 @@
+import axios from 'axios'
```
karrui, line 1: nit: pls call this `PublicFormAuthService`, `AuthService` will be extremely confusing

src/public/services/AuthService.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { RedirectUrlDto } from 'src/types/api/auth'
```
karrui, line 3: nit: use relative paths

src/public/services/AuthService.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// Exported for testing
+export const PUBLIC_FORMS_ENDPOINT = '/api/v3/forms'
```
karrui, line 6: I think it's fine to inline them since they are only used once

karrui (changes requested):
some changes requested; pls move the new functions to public-forms module instead of auth
can we also have unit tests at the very least for all the new backend functions?

src/app/modules/auth/auth.controller.ts
```diff
@@ -311,7 +309,7 @@ export const getRedirectLink: RequestHandler<
 export const handleRedirect = [
   celebrate({
     [Segments.QUERY]: Joi.object({
-      isPersistentLogin: Joi.boolean().optional(),
+      isPersistentLogin: Joi.boolean(),
```
karrui, line 312: this is a backwards incompatible change, I'd suggest reverting this.

seaerchin:
@karrui the unit tests are in the linked PR together with integration tests (see tests section) for both FE/BE. This PR is only the logical changes to avoid review fatigue.

src/app/modules/auth/auth.controller.ts
```diff
@@ -217,3 +227,92 @@ export const handleSignout: RequestHandler = async (req, res) => {
...
+export const getRedirectLink: RequestHandler<
+  { formId: string },
+  RedirectUrlDto | ErrorDto,
+  unknown,
+  Query & { isPersistentLogin: boolean }
+> = (req, res) => {
+  const { formId } = req.params
+  const { isPersistentLogin } = req.query
+  const logMeta = {
+    action: 'handleRedirect',
```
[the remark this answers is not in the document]
seaerchin: ohyea, good catch

karrui:
I see that tests are in another PR; fwiw I think unit tests should be contained in the same PR tbh, makes it easier for me

src/app/modules/auth/auth.controller.ts
seaerchin: ah ok. good point! i'll shift this over to the appropriate folder!

src/app/modules/auth/auth.controller.ts
seaerchin: sure. i'll still leave the frontend with default parameters but will revert this change to make it backwards compatible

karrui: Will need to revert the `!!` change since now there is a possibility for undefinedness

src/app/modules/auth/auth.utils.ts
seaerchin: i think the main motivating factor for not using the `mapRouteError` was the fear that there might be identical errors that are handled differently. i think this is premature  on my end and i'll combine it with the appropriate `mapRouteError` in public-forms

src/app/modules/auth/auth.controller.ts
seaerchin: i think this is good convention and will do this going forwards

src/app/modules/auth/auth.utils.ts
mantariksh: I separated the two mapRouteError functions in the same module because it was easier to check that the correct errors were being covered for each one. not sure if there's any benefit to smooshing all the errors together into a gigantic switch statement even if they're for different endpoints.

src/public/services/AuthService.ts
seaerchin: i would prefer not to for `PUBLIC_FORMS_ENDPOINT` because for `validate`, it's going to be reused but i'll fix it for the rest

src/app/modules/auth/auth.utils.ts
seaerchin: i do think that if we want `mapRouteError` to just be a black box and let consumers call and get the appropriate error code/message, it should just be 1 huge switch case HAHA

after checking, i don't think there are any errors that we treat differently for redirect, so i combined it into public-form's `mapRouteError`

mantariksh: >if we want mapRouteError to just be a black box and let consumers call and get the appropriate error code/message

I don't understand your point. we do want this - that's what a helper function is for - but why combine the switch cases for unrelated endpoints?

seaerchin: wouldn't the value of `mapRouteError` be for the module where it's rooted in - i.e., that for all routes rooted at `/forms`, we treat all errors similarly and defer the appropriate response to `mapRouteError`?

mantariksh: it's a minor thing so I'm not going to belabour this discussion anymore, but I don't think it's the right approach to anyhow throw unrelated errors together in the same switch case. it makes it so much harder to check that the errors are being covered exhaustively.

src/app/modules/spcp/spcp.types.ts
```diff
@@ -43,3 +45,8 @@ export interface ParsedSpcpParams {
   rememberMe: boolean
   cookieDuration: number
 }
+
+export interface ISpcpForm extends IFormSchema {
+  authType: AuthType.SP | AuthType.CP
+  esrvcId: string
+}
```
karrui, line 52: Suggest a composite type so the original type is preserved:

```ts
export type SpcpForm<T extends IFormSchema> = T & {
  authType: AuthType.SP | AuthType.CP
  esrvcId: string
}
```

src/app/modules/spcp/spcp.util.ts
```diff
@@ -210,6 +220,22 @@ export const createCorppassParsedResponses = (
...
+ * Validates that a form is a SPCP form with an e-service ID
+ * @param form Form to validate
+ */
+export const validateSpcpForm = (
+  form: IFormSchema | IPopulatedForm,
+): Result<ISpcpForm, SpcpNoESrvcIdError | SpcpAuthTypeError> => {
+  if (!form.esrvcId) {
+    return err(new SpcpNoESrvcIdError())
+  }
+  if (form.authType !== AuthType.SP && form.authType !== AuthType.CP) {
```
karrui, line 233: With the composite type above, if we make this into a type guard instead we can get better typings and remove the need for type casting

```ts
const isSpcpForm = <F extends IFormSchema>(form: F): form is SpcpForm<F> => {
  return [AuthType.SP, AuthType.CP].includes(form.authType)
}
```

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -90,44 +88,29 @@ function submitFormDirective(
...
-              $window.location.href = response.redirectURL
-            })
-            .catch((error) => {
-              console.error(error)
-              Toastr.error(
-                'Sorry, there was a problem with your login. Please try again.',
-              )
-            })
-        }
+      scope.formLogin = function (authType, isPersistentLogin) {
```
karrui, line 91: you don't even need the `authType` parameter here right?

karrui (changes requested):
a couple more changes; pls test on staging/local

src/app/modules/auth/auth.utils.ts
karrui: @seaerchin i defer to anti, you decide if you want to extract out the mapXError function to a specific form auth one. Fwiw I also agree with him

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -350,3 +364,92 @@ export const handleGetPublicForm: RequestHandler<
...
+        case AuthType.CP: {
+          // NOTE: Persistent login is only set (and relevant) when the authType is SP.
+          // If authType is not SP, assume that it was set erroneously and default it to false
+          return validateSpcpForm(form).andThen((form) => {
+            const target = `/${formId},${
+              // Need to cast to boolean because undefined is allowed as a valid value
+              // We are not following corppass's official spec for
+              // the target parameter
+              form.authType === AuthType.SP ? !!isPersistentLogin : false
+            }`
```
mantariksh, line 415: I think this business logic (only SP forms can have persistent login) should be in a service

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -350,3 +364,92 @@ export const handleGetPublicForm: RequestHandler<
...
+              form.authType,
+              target,
+              form.esrvcId,
+            )
+          })
+        }
+        // NOTE: Only MyInfo and SPCP should have redirects as the point of a redirect is
+        // to provide auth for users from a third party
+        default:
+          return err<never, AuthTypeMismatchError>(
```
mantariksh, line 426: am I right to assume that providing the generic types is necessary so that `mapErr` and `map` have the correct types?

src/app/modules/submission/email-submission/__tests__/email-submission.routes.spec.ts
```diff
@@ -44,21 +44,20 @@ jest.mock('nodemailer', () => ({
...
-const MockMyInfoGovClient = mocked(MyInfoGovClient, true)
-const mockExtractUinFin = jest.fn()
-MockMyInfoGovClient.mockImplementation(
-  () =>
-    (({
-      extractUinFin: mockExtractUinFin,
-    } as unknown) as MyInfoGovClient),
+
+const MockMyInfoGovClient = mocked(
+  new MyInfoClient.MyInfoGovClient({} as IMyInfoConfig),
```
mantariksh, line 59: wow how does this not throw an error? myinfogovclient definitely throws if params are missing

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -90,44 +88,29 @@ function submitFormDirective(
...
         // Fire GA tracking event
-        if (rememberMe) {
-          GTag.persistentLoginUse(scope.form)
-        }
+        if (isPersistentLogin) GTag.persistentLoginUse(scope.form)
 
-        // redirect to SPCP log in page
-        let destination =
-          '/' + scope.form._id + (scope.form.isPreview ? '/preview' : '')
```
mantariksh, line 115: I don't think this is a regression because we don't allow logins in preview mode anyway, but could you help double check that we're not losing any functionality here?

src/app/modules/form/public-form/public-form.controller.ts
seaerchin: yeap! in general, we actually need to hint when we construct `ok` or `err` directly. this is because they are type constructors and take 2 arguments, 1 of which can never occur.

src/app/modules/submission/email-submission/__tests__/email-submission.routes.spec.ts
seaerchin: i think this is allowed because we call `jest.mock` on the containing module. this transforms MyInfoGovClient into a mock function, whose return value is then set later on!

mantariksh: ah so you're calling `new` on the mocked module. got it

src/app/modules/spcp/spcp.errors.ts
```diff
@@ -96,3 +96,24 @@ export class MissingJwtError extends ApplicationError {
     super(message)
   }
 }
+
+/**
+ * SPCP form missing e-service ID.
+ */
+export class SpcpNoESrvcIdError extends ApplicationError {
```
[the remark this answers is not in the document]
seaerchin: i didn't do this because `MyInfoNoESrvcIdError` has another callsite hwich is not related to this (and is `myInfo` specific). however, if this is desired, we can still keep this error and create a new sum type that is either `{Spcp | MyInfo}...Error`. 

i think you're correct and this serves to make code more readable so i'll go with teh above change!

src/app/modules/auth/auth.utils.ts
seaerchin: ok sure! will do

src/public/modules/forms/base/directives/submit-form.directive.js
seaerchin: good catch! thanks for reminding me and i'll remove it

src/app/modules/spcp/spcp.errors.ts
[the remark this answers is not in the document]
mantariksh: you can just use `FormAuthNoEsrvcIdError`, I don't think it's necessary to keep the MyInfo one separate

src/app/modules/form/public-form/public-form.controller.ts
seaerchin: good point - i think this should be part of `SpcpFactory.createRedirectUrl` and i'll shift it there!

src/app/modules/spcp/spcp.util.ts
seaerchin: good suggestion! will also do teh same for `MyInfo` and look to preserve type information through typeguards in the future

src/app/modules/spcp/spcp.errors.ts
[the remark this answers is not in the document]
seaerchin: sure, this sounds good and i'll make the changes accordingly! on that note, there's a `MyInfoAuthTypeError` along with `AuthTypeMismatchError` (in spcp) - i think i'll combine these also into `AuthTypeMismatchError` and shift it out into public-forms!

src/app/modules/spcp/spcp.errors.ts
```diff
@@ -96,3 +96,24 @@ export class MissingJwtError extends ApplicationError {
...
+  constructor(message = 'Form does not have e-service ID') {
+    super(message)
+  }
+}
+
+/**
+ * Attempt to perform a Spcp-related operation on a form without Spcp
+ * authentication enabled.
+ */
+export class SpcpAuthTypeError extends ApplicationError {
```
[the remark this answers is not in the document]
seaerchin: good catch! i'll combine this with the existing `AuthTypeMismatchError`!

src/public/modules/forms/base/directives/submit-form.directive.js
seaerchin: yeap, checked staging behaviour with prod - we present users with a view of the logged in form so we aren't losing functionality. users also can't logout (all buttons have no effect besides edit form)

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -285,16 +286,23 @@ export const createRelayState = (formId: string): string =>
...
   }
-  return ok(form as IMyInfoForm)
+  return err(new AuthTypeMismatchError(AuthType.MyInfo, form.authType))
+}
+
+// Typeguard to ensure that form has eserviceId and MyInfo authType
+const isMyInfoForm = <F extends IFormSchema>(
+  form: F,
+): form is MyInfoForm<F> => {
+  return form.authType === AuthType.MyInfo && !!form.esrvcId
```
karrui, line 305: Doesn't seem like it, but ensure that the additional `!!form.esrvcId` check is not a regression!

src/app/modules/spcp/spcp.util.ts
```diff
@@ -210,6 +221,32 @@ export const createCorppassParsedResponses = (
...
+  }
+  return err(new AuthTypeMismatchError(AuthType.CP, form.authType))
+}
+
+// Typeguard to ensure that form has eserviceId and correct authType
+const isSpcpForm = <F extends IFormSchema>(form: F): form is SpcpForm<F> => {
+  return (
+    !!form.authType &&
+    [AuthType.SP, AuthType.CP].includes(form.authType) &&
+    !!form.esrvcId
```
karrui, line 246: same here

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -90,44 +88,29 @@ function submitFormDirective(
         $('head').append(meta)
       }
 
-      scope.formLogin = function (authType, rememberMe) {
-        if (authType === 'MyInfo') {
-          return $q
-            .when(MyInfoService.createRedirectURL(scope.form._id))
```
karrui, line 96: `MyInfoService` seems to now be unused, should delete in this PR too

src/types/api/auth.ts
```diff
@@ -0,0 +1 @@
+export type RedirectUrlDto = { redirectURL: string }
```
karrui, line 1: nit: change the name to `PublicFormAuthRedirectDto`?

karrui (approved):
lgtm with comments

src/app/modules/myinfo/myinfo.util.ts
seaerchin: good catch! will do so (also for the below one)

src/public/modules/forms/base/directives/submit-form.directive.js
seaerchin: `MyInfoService` has `validateESrvcId`, which is why i left it here cos i wanted to delete the whole file at once HAHA more satisfying but i can delete `createRedirectURL` if you want!

src/types/api/auth.ts
seaerchin: np will do

src/public/modules/forms/base/directives/submit-form.directive.js
karrui: delete the unused code

src/app/modules/form/form.errors.ts
```diff
@@ -60,3 +62,26 @@ export class LogicNotFoundError extends ApplicationError {
...
+export class AuthTypeMismatchError extends ApplicationError {
+  constructor(attemptedAuthType: AuthType, formAuthType?: AuthType) {
+    super(
+      `Attempted authentication type ${attemptedAuthType} did not match form auth type ${formAuthType}`,
+    )
+  }
+}
+
+/**
+ * Error to be returned when the form is lacking an eServiceId
```
mantariksh, line 78: nit: to be more precise, you can add that the form has SP, CP or MyInfo enabled but is missing an e-service ID

src/app/modules/myinfo/myinfo.util.ts
seaerchin: on prod, we don't allow the eserviceId field to be empty. this would get caught by the first layer of checks (and the correct error returned). hence, this isn't a regression!

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -350,3 +360,91 @@ export const handleGetPublicForm: RequestHandler<
...
+}
+
+/**
+ * Handler for /forms/:formId/auth/redirect
+ */
+export const handleFormAuthRedirect = [
+  celebrate({
+    [Segments.QUERY]: Joi.object({
+      isPersistentLogin: Joi.boolean().optional(),
+    }),
```
mantariksh, line 447: should we validate for the form ID in `Segments.PARAMs` as well? I know we already validate using Express route  matching, but it's nice to have all the validation in one place

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -285,16 +286,23 @@ export const createRelayState = (formId: string): string =>
...
-    return err(new MyInfoAuthTypeError())
+  if (isMyInfoForm(form)) {
+    return ok(form)
   }
-  return ok(form as IMyInfoForm)
+  return err(new AuthTypeMismatchError(AuthType.MyInfo, form.authType))
+}
+
+// Typeguard to ensure that form has eserviceId and MyInfo authType
+const isMyInfoForm = <F extends IFormSchema>(
```
mantariksh, line 302: technically a MyInfo form can exist without an e-service ID as long as it's not public. can we name this more specifically to reflect that? eg isMyInfoFormWithEsrvcId

src/public/modules/forms/base/directiveViews/submit-form.directive.view.html
```diff
@@ -11,7 +11,7 @@
       auth-type="form.authType"
       is-admin-preview="false"
       is-template="form.isTemplate"
-      form-login="formLogin(authType, rememberMe)"
+      form-login="formLogin(isPersistentLogin)"
```
mantariksh, line 14: is this set on the `scope`? I don't think it is, which means it is always `undefined`

mantariksh (changes requested):
possible bug with persistent logins. have you tested that persistent logins are working?

src/public/modules/forms/base/directiveViews/submit-form.directive.view.html
mantariksh: ah nvm, I see that it's using `&` binding

mantariksh (approved):
approving because I don't think the `isPersistentLogin` thing is a bug, but do test that extended logins work before merging

src/public/modules/forms/base/directiveViews/submit-form.directive.view.html
karrui: angularjs strikes again

seaerchin:
@mantariksh could you clarify abit on the extended login? to be safe, i'll push this to staging and test it there!

src/app/modules/form/public-form/public-form.controller.ts
seaerchin: oh sure! just to make sure we're on the same wavelength, the express regex should still remain also right?

mantariksh: yep

src/app/modules/myinfo/myinfo.util.ts
seaerchin: oh yea, definitely. good catch and i'll do so!

mantariksh:
@seaerchin extended login means the Singpass cookie should expire after 30 days instead of the usual 1 hour

src/public/modules/forms/base/directiveViews/submit-form.directive.view.html
seaerchin: the binding thing is still very trippy tbh. just leaving [this](https://www.codelord.net/2016/05/13/understanding-angulars-and-binding/) here for the weird & binding thing in case anyone needs it (i.e, future me)

seaerchin:
@mantariksh gotcha, will do!

PR #1695, opened 2021-04-21
refactor: extract update logic endpoint

### LINKED ISSUE #1488: [API] Extract update logic endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **updating form logic**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
PUT /api/v3/admin/forms/:formId/logic/:logicId
```

> Note the `PUT` HTTP method, endpoint should be idempotent

tshuli:
## Problem

Closes #1488

## Tests

- [ ] Create field with one logic field. Edit logic - both from show field to prevent submit and vice versa. Check that logic is successfully edited.
- [ ] Check that build tab is updated correctly (hidden fields should be greyed out based on logic)
- [ ] Repeat the above with more than one logic field.

Test using manual API calls
- [ ] Check that endpoint returns 404 when formId or logicId are incorrect

### DIFF
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -1288,12 +1288,10 @@ describe('Form Model', () => {
         form_logics: [
           {
             _id: logicId1,
-            id: logicId1,
             logicType: LogicType.ShowFields,
           } as ILogicSchema,
           {
             _id: logicId2,
-            id: logicId2,
             logicType: LogicType.ShowFields,
           } as ILogicSchema,
         ],
@@ -1301,7 +1299,6 @@ describe('Form Model', () => {
 
       const mockUpdatedFormLogic = {
         _id: logicId1,
-        id: logicId1,
         logicType: LogicType.PreventSubmit,
       } as ILogicSchema
 
@@ -1311,7 +1308,6 @@ describe('Form Model', () => {
           form_logics: [
             {
               _id: logicId1,
-              id: logicId1,
               logicType: LogicType.ShowFields,
             } as ILogicSchema,
           ],
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -17,6 +17,7 @@ import {
   IFormSchema,
   ILogicSchema,
   IPopulatedUser,
+  LogicType,
   Permission,
   ResponseMode,
   Status,
@@ -1276,6 +1277,131 @@ describe('Form Model', () => {
         // Assert
         expect(actual).toEqual(null)
         await expect(Form.countDocuments()).resolves.toEqual(0)
+      })
+    })
+  })
+    describe('updateFormLogic', () => {
+      const logicId1 = new ObjectId().toHexString()
+      const logicId2 = new ObjectId().toHexString()
+
+      const mockExistingFormLogic = {
+        form_logics: [
+          {
+            _id: logicId1,
+            id: logicId1,
+            logicType: LogicType.ShowFields,
+          } as ILogicSchema,
+          {
+            _id: logicId2,
+            id: logicId2,
+            logicType: LogicType.ShowFields,
+          } as ILogicSchema,
+        ],
+      }
+
+      const mockUpdatedFormLogic = {
+        _id: logicId1,
+        id: logicId1,
+        logicType: LogicType.PreventSubmit,
+      } as ILogicSchema
+
+      it('should return form upon successful update of logic when there is one logic', async () => {
+        // arrange
+        const mockExistingFormLogicSingle = {
+          form_logics: [
+            {
+              _id: logicId1,
+              id: logicId1,
+              logicType: LogicType.ShowFields,
+            } as ILogicSchema,
+          ],
+        }
+
+        const formParams = merge({}, MOCK_EMAIL_FORM_PARAMS, {
+          admin: populatedAdmin,
+          status: Status.Public,
+          responseMode: ResponseMode.Email,
+          ...mockExistingFormLogicSingle,
+        })
+        const form = await Form.create(formParams)
+
+        // act
+        const modifiedForm = await Form.updateFormLogic(
+          form._id,
+          logicId1,
+          mockUpdatedFormLogic,
+        )
+
+        // assert
+        // Form should be returned
+        expect(modifiedForm).not.toBeNull()
+
+        // Form should have correct status, responsemode
+        expect(modifiedForm?.responseMode).not.toBeNull()
+        expect(modifiedForm?.responseMode).toEqual(ResponseMode.Email)
+        expect(modifiedForm?.status).not.toBeNull()
+        expect(modifiedForm?.status).toEqual(Status.Public)
+
+        // Check that form logic has been updated
+        expect(modifiedForm?.form_logics).toBeDefined()
+        expect(modifiedForm?.form_logics).toHaveLength(1)
+        expect(modifiedForm!.form_logics![0].logicType).toEqual(
+          LogicType.PreventSubmit,
+        )
+      })
+
+      it('should return form upon successful update of logic when there are more than one logics', async () => {
+        // arrange
+        const formParams = merge({}, MOCK_EMAIL_FORM_PARAMS, {
+          admin: populatedAdmin,
+          status: Status.Public,
+          responseMode: ResponseMode.Email,
+          ...mockExistingFormLogic,
+        })
+        const form = await Form.create(formParams)
+
+        // act
+        const modifiedForm = await Form.updateFormLogic(
+          form._id,
+          logicId1,
+          mockUpdatedFormLogic,
+        )
+
+        // assert
+        // Form should be returned
+        expect(modifiedForm).not.toBeNull()
+
+        // Form should have correct status, responsemode
+        expect(modifiedForm?.responseMode).not.toBeNull()
+        expect(modifiedForm?.responseMode).toEqual(ResponseMode.Email)
+        expect(modifiedForm?.status).not.toBeNull()
+        expect(modifiedForm?.status).toEqual(Status.Public)
+
+        // Check that first form logic has been updated but second is unchanges
+        expect(modifiedForm?.form_logics).toBeDefined()
+        expect(modifiedForm?.form_logics).toHaveLength(2)
+        expect(modifiedForm!.form_logics![0].logicType).toEqual(
+          LogicType.PreventSubmit,
+        )
+        expect(modifiedForm!.form_logics![1].logicType).toEqual(
+          LogicType.ShowFields,
+        )
+      })
+
+      it('should return null if formId is invalid', async () => {
+        // arrange
+        const invalidFormId = new ObjectId().toHexString()
+
+        // act
+        const modifiedForm = await Form.updateFormLogic(
+          invalidFormId,
+          logicId1,
+          mockUpdatedFormLogic,
+        )
+
+        // assert
+        // should return null
+        expect(modifiedForm).toBeNull()
       })
     })
   })
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -1279,7 +1279,7 @@ describe('Form Model', () => {
         await expect(Form.countDocuments()).resolves.toEqual(0)
       })
     })
-  })
+
     describe('updateFormLogic', () => {
       const logicId1 = new ObjectId().toHexString()
       const logicId2 = new ObjectId().toHexString()
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -1399,6 +1399,51 @@ describe('Form Model', () => {
         // should return null
         expect(modifiedForm).toBeNull()
       })
+
+      it('should return unmodified form if logicId is invalid', async () => {
+        // arrange
+        const invalidLogicId = new ObjectId().toHexString()
+        const mockExistingFormLogicSingle = {
+          form_logics: [
+            {
+              _id: invalidLogicId,
+              logicType: LogicType.ShowFields,
+            } as ILogicSchema,
+          ],
+        }
+
+        const formParams = merge({}, MOCK_EMAIL_FORM_PARAMS, {
+          admin: populatedAdmin,
+          status: Status.Public,
+          responseMode: ResponseMode.Email,
+          ...mockExistingFormLogicSingle,
+        })
+        const form = await Form.create(formParams)
+
+        // act
+        const modifiedForm = await Form.updateFormLogic(
+          form._id,
+          logicId1,
+          mockUpdatedFormLogic,
+        )
+
+        // assert
+        // Form should be returned
+        expect(modifiedForm).not.toBeNull()
+
+        // Form should have correct status, responsemode
+        expect(modifiedForm?.responseMode).not.toBeNull()
+        expect(modifiedForm?.responseMode).toEqual(ResponseMode.Email)
+        expect(modifiedForm?.status).not.toBeNull()
+        expect(modifiedForm?.status).toEqual(Status.Public)
+
+        // Check that form logic has not been updated and there are no new form logics introduced
+        expect(modifiedForm?.form_logics).toBeDefined()
+        expect(modifiedForm?.form_logics).toHaveLength(1)
+        expect(modifiedForm!.form_logics![0].logicType).toEqual(
+          LogicType.ShowFields,
+        )
+      })
     })
   })
 
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -29,7 +29,7 @@ import {
   IFormDocument,
   IFormModel,
   IFormSchema,
-  ILogicSchema,
+  ILogicWithId,
   IPopulatedForm,
   LogicType,
   Permission,
@@ -690,7 +690,7 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     this: IFormModel,
     formId: string,
     logicId: string,
-    updatedLogic: ILogicSchema,
+    updatedLogic: ILogicWithId,
   ): Promise<IFormSchema | null> {
     return this.findByIdAndUpdate(
       formId,
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -29,6 +29,7 @@ import {
   IFormDocument,
   IFormModel,
   IFormSchema,
+  ILogicSchema,
   IPopulatedForm,
   LogicType,
   Permission,
@@ -682,6 +683,25 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       formId,
       { $pull: { form_fields: { _id: fieldId } } },
       { new: true, runValidators: true },
+    ).exec()
+  }
+  // Updates specified form logic.
+  FormSchema.statics.updateFormLogic = async function (
+    this: IFormModel,
+    formId: string,
+    logicId: string,
+    updatedLogic: ILogicSchema,
+  ): Promise<IFormSchema | null> {
+    return this.findByIdAndUpdate(
+      mongoose.Types.ObjectId(formId),
+      {
+        $set: { 'form_logics.$[object]': updatedLogic },
+      },
+      {
+        arrayFilters: [{ 'object._id': logicId }],
+        new: true,
+        runValidators: true,
+      },
     ).exec()
   }
 
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -693,7 +693,7 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     updatedLogic: ILogicSchema,
   ): Promise<IFormSchema | null> {
     return this.findByIdAndUpdate(
-      mongoose.Types.ObjectId(formId),
+      formId,
       {
         $set: { 'form_logics.$[object]': updatedLogic },
       },
--- a/src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
@@ -1751,7 +1751,7 @@ describe('admin-form.service', () => {
 
     it('should return ok(updated logic) on successful form logic update for email mode form', async () => {
       // Arrange
-      UPDATE_SPY.mockResolvedValue(mockEmailFormUpdated)
+      UPDATE_SPY.mockResolvedValue(mockEmailFormUpdated as IFormSchema)
 
       // Act
       const actualResult = await updateFormLogic(
@@ -1773,7 +1773,7 @@ describe('admin-form.service', () => {
 
     it('should return ok(updated logic) on successful form logic update for encrypt mode form', async () => {
       // Arrange
-      UPDATE_SPY.mockResolvedValue(mockEncryptFormUpdated)
+      UPDATE_SPY.mockResolvedValue(mockEncryptFormUpdated as IFormSchema)
 
       // Act
       const actualResult = await updateFormLogic(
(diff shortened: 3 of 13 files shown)


src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -676,3 +677,55 @@ export const deleteFormLogic = (
...
+          logicId,
+          updatedLogic,
+        },
+        error,
+      })
+      return transformMongoError(error)
+    },
+    // On success, return form
+  ).andThen((updatedForm) => {
+    if (!updatedForm) {
```
seaerchin, line 726: can i ask about when will this conditional be hit? in `handleUpdateLogic`, this is called after already retrieving a form from the database, so wouldn't the form always be updated?

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
```diff
@@ -263,19 +268,40 @@ function EditLogicModalController(
...
+        }
+      })
       // Not new, and logic index is provided
     } else if (logicIndex !== -1) {
-      vm.formLogics[logicIndex] = vm.logic
+      vm.updateExistingLogic(logicIndex, vm.logic).then((error) => {
+        if (!error) {
+          $uibModalInstance.close()
+        }
+      })
```
seaerchin, line 282: should we change this to be within `updateExistingLogic` as i'm unsure if an error would actually be thrown here (as it is caught within the `updateExistingLogic` portion)

seaerchin (commented):
i've a few questions below, mainly stemming from my knowledge gaps in angular.

src/app/models/form.server.model.ts
```diff
@@ -631,6 +632,26 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
 
+  // Updates specified form logic.
+  FormSchema.statics.updateFormLogic = async function (
+    this: IFormModel,
+    formId: string,
+    logicId: string,
+    updatedLogic: ILogicSchema,
+  ): Promise<IFormSchema | null> {
+    return this.findByIdAndUpdate(
+      mongoose.Types.ObjectId(formId),
```
mantariksh, line 643: no need for this, you can leave `formId` as a string

src/app/models/form.server.model.ts
```diff
@@ -631,6 +632,26 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+      {
+        $set: { 'form_logics.$[object]': updatedLogic },
+      },
+      {
+        arrayFilters: [{ 'object._id': logicId }],
```
mantariksh, line 648: very noice

src/app/modules/form/admin-form/admin-form.service.ts
mantariksh: you're right, this is just to be defensive

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
```diff
@@ -263,19 +268,40 @@ function EditLogicModalController(
...
+    const logicIdToUpdate = vm.formLogics[logicIndex]._id
+    $q.when(
+      AdminFormService.updateFormLogic(
+        vm.myform._id,
+        logicIdToUpdate,
+        updatedLogic,
+      ),
+    )
+      .then(() => {
+        vm.formLogics[logicIndex] = updatedLogic
```
mantariksh, line 300: do we need to update the global `myform` object here? also do we need to do `$uibModalInstance.close()`?

mantariksh (commented):
preliminary comments!

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
tshuli: thanks for suggestion, shifted

src/app/models/form.server.model.ts
tshuli: edited

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
tshuli: thanks, edited

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1717,6 +1718,60 @@ export const _handleReorderFormField: RequestHandler<
...
+          formId,
+          level: PermissionLevel.Write,
+        }),
+      )
+
+      // Step 3: Update form logic
+      .andThen((retrievedForm) =>
+        AdminFormService.updateFormLogic(retrievedForm, logicId, updatedLogic),
+      )
+      .map(() => res.sendStatus(StatusCodes.OK))
```
mantariksh, line 1756: we should be returning the saved updated logic field so the client always gets a copy which is consistent with the current state of the db

src/public/services/AdminFormService.ts
```diff
@@ -76,3 +76,13 @@ export const deleteFormLogic = async (
...
 }
+
+export const updateFormLogic = async (
+  formId: string,
+  logicId: string,
+  updatedLogic: ILogicSchema,
+): Promise<true> => {
+  return axios
+    .put(`${ADMIN_FORM_ENDPOINT}/${formId}/logic/${logicId}`, { updatedLogic })
+    .then(() => true)
```
mantariksh, line 87: we should be returning the updated logic field here

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
```diff
@@ -263,19 +268,38 @@ function EditLogicModalController(
...
```
mantariksh, line 296: `updatedLogic` should be the logic element returned by the backend

mantariksh (changes requested):

src/app/models/form.server.model.ts
```diff
@@ -671,6 +672,26 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     ).exec()
   }
 
+  // Updates specified form logic.
+  FormSchema.statics.updateFormLogic = async function (
+    this: IFormModel,
+    formId: string,
+    logicId: string,
+    updatedLogic: ILogicSchema,
```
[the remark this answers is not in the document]
tshuli: thanks, edited

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1717,6 +1718,60 @@ export const _handleReorderFormField: RequestHandler<
...
+ * @security session
+ *
+ * @returns 200 with success message when successfully updated
+ * @returns 403 when user does not have permissions to update logic
+ * @returns 404 when form cannot be found
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database error occurs
+ */
+export const handleUpdateLogic: RequestHandler<
+  ParamsDictionary,
```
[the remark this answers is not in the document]

src/app/routes/api/v3/admin/forms/admin-forms.logic.routes.ts
```diff
@@ -19,3 +19,19 @@ export const AdminFormsLogicRouter = Router()
...
+ * @consumes application/json
+ * @returns 200 with success message when successfully updated
+ * @returns 403 when user does not have permissions to update logic
+ * @returns 404 when form cannot be found
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database error occurs
+ */
+AdminFormsLogicRouter.route(
+  '/:formId([a-fA-F0-9]{24})/logic/:logicId([a-fA-F0-9]{24})',
+).put(AdminFormController.handleUpdateLogic)
```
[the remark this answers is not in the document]
tshuli: edited

src/app/modules/form/admin-form/admin-form.controller.ts

src/public/services/AdminFormService.ts
tshuli: edited

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
tshuli: edited

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
```diff
@@ -263,19 +268,38 @@ function EditLogicModalController(
...
+  vm.updateExistingLogic = function (logicIndex, updatedLogic) {
+    const logicIdToUpdate = vm.formLogics[logicIndex]._id
+    $q.when(
+      AdminFormService.updateFormLogic(
+        vm.myform._id,
+        logicIdToUpdate,
+        updatedLogic,
+      ),
+    )
+      .then(({ form_logics }) => {
```
mantariksh, line 295: the API should just return the individual logic element, so you can do `vm.formLogics[logicIndex] = updatedLogic`

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1730,6 +1732,98 @@ export const handleReorderFormField = [
...
+ * @returns 200 with success message when successfully updated
+ * @returns 403 when user does not have permissions to update logic
+ * @returns 404 when form cannot be found
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database error occurs
+ */
+export const _handleUpdateLogic: RequestHandler<
+  { formId: string; logicId: string },
+  unknown,
+  { updatedLogic: ILogicSchema }
```
mantariksh, line 1750: in addition, the request and response bodies should be typed as DTOs. you can look at other endpoints (e.g. update form field) as a reference

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -733,6 +734,58 @@ export const deleteFormLogic = (
...
+ * @returns err(database errors) if db error is thrown during logic update
+ * @returns err(LogicNotFoundError) if logicId does not exist on form
+ */
+export const updateFormLogic = (
+  form: IPopulatedForm,
+  logicId: string,
+  updatedLogic: ILogicSchema,
+): ResultAsync<IFormSchema, DatabaseError | LogicNotFoundError> => {
+  // First check if specified logic exists
+  if (!form.form_logics.some((logic) => logic.id === logicId)) {
```
mantariksh, line 752: would it be better to standardise to using `_id`?

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1202,6 +1203,131 @@ describe('Form Model', () => {
...
+
+    describe('updateFormLogic', () => {
+      const logicId1 = new ObjectId().toHexString()
+      const logicId2 = new ObjectId().toHexString()
+
+      const mockExistingFormLogic = {
+        form_logics: [
+          {
+            _id: logicId1,
+            id: logicId1,
```
mantariksh, line 1215: I'm in favour of standardising to `_id` because it would remove the need for stuff like this

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1202,6 +1203,131 @@ describe('Form Model', () => {
         })
       })
     })
+
+    describe('updateFormLogic', () => {
```
mantariksh, line 1207: should we have a test for when the logic ID doesn't exist? just to be defensive, I know that it's already validated in the service, but there may be very rare race conditions where the logic gets deleted after the service check but before the db update

mantariksh (changes requested):

src/app/models/__tests__/form.server.model.spec.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: edited

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: edited

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
tshuli: edited

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -733,6 +734,58 @@ export const deleteFormLogic = (
...
+        error,
+      })
+      return transformMongoError(error)
+    },
+    // On success, return form
+  ).andThen((updatedForm) => {
+    if (!updatedForm) {
+      return errAsync(new FormNotFoundError())
+    }
+    return okAsync(updatedForm)
```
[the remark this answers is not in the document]
tshuli: edited

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1730,6 +1732,98 @@ export const handleReorderFormField = [
...
+  celebrate(
+    {
+      [Segments.BODY]: Joi.object({
+        // Ensures given logic is same as accessed logic
+        updatedLogic: {
+          _id: Joi.string().valid(Joi.ref('$params.logicId')).required(),
+          logicType: Joi.string()
+            .valid(...Object.values(LogicType))
+            .required(),
+          conditions: Joi.array().required(),
```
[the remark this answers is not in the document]
tshuli: added in further validation for the conditions array

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1730,6 +1732,98 @@ export const handleReorderFormField = [
...
+          formId,
+          level: PermissionLevel.Write,
+        }),
+      )
+
+      // Step 3: Update form logic
+      .andThen((retrievedForm) =>
+        AdminFormService.updateFormLogic(retrievedForm, logicId, updatedLogic),
+      )
+      .map((updatedForm) => res.status(StatusCodes.OK).json(updatedForm))
```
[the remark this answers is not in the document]
tshuli: edited

src/app/modules/form/admin-form/admin-form.controller.ts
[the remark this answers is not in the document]
tshuli: edited

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1730,6 +1732,98 @@ export const handleReorderFormField = [
...
+ *
+ * @returns 200 with success message when successfully updated
+ * @returns 403 when user does not have permissions to update logic
+ * @returns 404 when form cannot be found
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database error occurs
+ */
+export const _handleUpdateLogic: RequestHandler<
+  { formId: string; logicId: string },
+  unknown,
```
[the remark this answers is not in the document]
tshuli: edited

src/app/models/__tests__/form.server.model.spec.ts
tshuli: thanks for pointing out - in this case, the DB would return an unmodified form, and at the service level, we check if the logic exists in the returned form and if not, then return logicNotFoundError i.e. the logic was deleted in between the service check and db update

tshuli:
@mantariksh thanks for reviewing, for your re-review pls

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -1625,4 +1626,167 @@ describe('admin-form.service', () => {
...
+      mockEncryptFormUpdated = ({
+        ...mockEncryptForm,
+        ...mockFormLogicUpdated,
+      } as unknown) as IPopulatedForm
+    })
+
+    it('should return ok(updated logic) on successful form logic update for email mode form', async () => {
+      // Arrange
+      const STATIC_UPDATE_SPY = jest
+        .spyOn(FormModel, 'findByIdAndUpdate')
```
mantariksh, line 1700: is it easier to write the tests if you spy on `updateFormLogic` instead?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1730,6 +1733,106 @@ export const handleReorderFormField = [
...
+
+/**
+ * Handler for PUT /forms/:formId/logic/:logicId
+ */
+export const handleUpdateLogic = [
+  celebrate(
+    {
+      [Segments.BODY]: Joi.object({
+        // Ensures given logic is same as accessed logic
+        updatedLogic: {
```
mantariksh, line 1800: is it necessary to have `updatedLogic` as an intermediate key? if not, would it be better to remove `LogicUpdateDto` and just have `LogicDto` for both the request and response?

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -733,6 +735,63 @@ export const deleteFormLogic = (
...
+    },
+    // On success, return updated form logic
+  ).andThen((updatedForm) => {
+    if (!updatedForm) {
+      return errAsync(new FormNotFoundError())
+    }
+    const updatedLogic = updatedForm.form_logics?.filter(
+      (logic) => logic._id.toHexString() === logicId,
+    )[0]
+    if (!updatedLogic) return errAsync(new LogicNotFoundError()) // Possible race condition if logic gets deleted after the initial logicId check but before the db update
```
mantariksh, line 789: possible to implement a `getLogicById`, similar to `getFormFieldById`?

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
```diff
@@ -263,19 +268,38 @@ function EditLogicModalController(
...
+    $q.when(
+      AdminFormService.updateFormLogic(
+        vm.myform._id,
+        logicIdToUpdate,
+        updatedLogic,
+      ),
+    )
+      .then((updatedLogic) => {
+        vm.formLogics[logicIndex] = updatedLogic
+        externalScope.myform.form_logics[logicIndex] = updatedLogic // update global myform
```
mantariksh, line 297: could you double check that all logic dependencies in the admin panel (eg in the Build tab, the warning for invalid logic and the opacity of hidden fields) are correctly updated when you do this? I'm not sure this triggers the digest cycle correctly for the various `$scope.$watch` that we have.

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: thanks, edited

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: edited

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
tshuli: thanks, its already spied on in L1667 and checked in L1730, etc

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
tshuli: see if edits work

tshuli:
@mantariksh thanks for reviewing, for your re-review pls

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -1625,4 +1626,167 @@ describe('admin-form.service', () => {
...
+        mockEmailFormId.toHexString(),
+        {
+          $set: { 'form_logics.$[object]': updatedLogic },
+        },
+        {
+          arrayFilters: [{ 'object._id': logicId1.toHexString() }],
+          new: true,
+          runValidators: true,
+        },
+      )
```
mantariksh, line 1728: I'm not sure this test is behaving as intended. if you've already spied on `FormModel.updateFormLogic`, then `FormModel.findByIdAndUpdate` shouldn't be called, because `updateFormLogic`. could you help me understand how this test works?

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.logic.routes.spec.ts
```diff
@@ -196,4 +196,221 @@ describe('admin-form.logic.routes', () => {
...
+        conditions: [],
+        preventSubmitMessage: 'Some message',
+      } as unknown) as ILogicSchema
+
+      const session = await createAuthedSession(user.email, request)
+
+      // Act
+      const response = await session
+        .put(`/admin/forms/${formToUpdate._id}/logic/${formLogicId}`)
+        .send({ ...updatedLogic })
```
mantariksh, line 227:
```suggestion
        .send(updatedLogic)
```

mantariksh: same for the rest of the tests

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
```diff
@@ -263,19 +268,40 @@ function EditLogicModalController(
...
+        logicIdToUpdate,
+        updatedLogic,
+      ),
+    )
+      .then((updatedLogic) => {
+        const updatedFormLogics = [...vm.formLogics]
+        updatedFormLogics[logicIndex] = updatedLogic
+        vm.formLogics = updatedFormLogics
+        externalScope.myform.form_logics = updatedFormLogics // update global myform
+        $uibModalInstance.close()
```
mantariksh, line 300: can I check if you verified that this executes the digest cycle correctly, and hence updates the Build tab correctly?

tshuli: Yes! it does

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
tshuli: In this case, I didn't mock the return value of `updateFormLogic`, only checked that it was called with the correct parameters by spying on it. Hence `findByIdAndUpdate` is still called (see below from the static method declaration)

```
// Updates specified form logic.
  FormSchema.statics.updateFormLogic = async function (
    this: IFormModel,
    formId: string,
    logicId: string,
    updatedLogic: ILogicWithId,
  ): Promise<IFormSchema | null> {
    return this.findByIdAndUpdate(
      formId,
      {
        $set: { 'form_logics.$[object]': updatedLogic },
      },
      {
        arrayFilters: [{ 'object._id': logicId }],
        new: true,
        runValidators: true,
      },
    ).exec()
  }
```

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
tshuli: ^ I had also updated the manual test to cover this

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.logic.routes.spec.ts
tshuli: Thanks! edited

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
mantariksh: I see. is it necessary to check the args for `findByIdAndUpdate` as well? seems a bit overkill since we already have tests for `updateFormLogic`. imo it's cleaner to just mock the layer below without digging into the implementation details.

tshuli:
@mantariksh thanks for reviewing, for your re-review pls

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
tshuli: edited; no strong views on this

mantariksh (commented):
lgtm, will approve upon fixing merge conflicts.

also...sigh...if you want to try `$watchCollection` pls go ahead! although I am still not a fan of trying new AngularJS things in production when we're so close to migrating the frontend, canonical solution or not

tshuli:
@mantariksh thanks! fixed the merge conflicts
will look at $watchCollection in the next PR

mantariksh (approved):

PR #1696, opened 2021-04-21
feat(webhooks): streamline webhook response data

mantariksh:
## Problem

The webhook response data stored in the database has a couple of superfluous fields:
1. `errorMessage` is useless because it is always a standard `axios` error message, e.g. "Request failed with status code 400".
2. `statusText` is useless because it is just the standard HTTP status text, e.g. `'OK'` for 200

Furthermore, fields that should always be present (webhook URL, signature, status code) are not marked as required in the database schema.

## Solution

Remove the superfluous fields and mark the rest as `required` where appropriate. `headers` and `data` were not made required as they may be empty strings.


karrui (approved):
lgtm

Issue #1312, opened 2021-03-08
Constrain email field domain model to prevent inconsistencies


liangyuanruo:
The `hasAllowedEmailDomains` boolean property on the email field schema is used to support the toggle in the UI. Explicitly saving this boolean into the database can be problematic because it can result in inconsistent database state, which occurs when `hasAllowedEmailDomains` is `true` and yet `allowedEmailDomains` is empty.

We can constrain the model by using an empty `allowedEmailDomains` array to replace the false value of `hasAllowedEmailDomains`. The toggle behavior can be recovered by simply having the frontend controller inspect the length of the array.

karrui:
Blocked by https://github.com/opengovsg/FormSG/issues/1481

PR #1697, opened 2021-04-21
fix: sync email field state between hasAllowedEmailDomains and allowedEmailDomains

karrui:
> This is the first PR of our journey to remove `emailField.hasAllowedEmailDomains` from the schema. 

## Problem

Currently, it is somehow possible (and we have some states in the database where this is so) to have the following email field states:

|`hasAllowedEmailDomains`|`allowedEmailDomains`| Remarks|
|---|---|---|
|`true`| `[]`| Not ideal, since the source of truth is muddled, but no real harm since there are sanity checks in the code where isVerifiable && hasAllowedEmailDomains && allowedEmailDomains.length must all be true before restricting domains|
| `false` | `["@someDomain.com", "@anotherDomain.com"]` | Was supposed to be a feature such that admins will be able to toggle restriction on/off and have their previously entered domains. Scrapping this state for cleanliness|
| `true`| `["@someDomain.com", "@anotherDomain.com"]`| consistent and expected shape |
| `false`| `[]`|consistent and expected shape |

This PR removes the first two possible states and syncs all newly created or updated form into the two consistent shapes with both client and server-side validation.

The end goal is to remove `emailField.hasAllowedEmailDOmains` from the email field schema such that the decision of whether or not to restrict email verification to some whitelisted email domains is **determined solely** by the `emailField.allowedEmailDomains` array values.

The PR to fully remove the `hasAllowedEmailDomains` key will come in a later PR when users have stopped using the old client without the frontend sync checks, as it is not safe to migrate now due to possible data inconsistencies whilst running the migration scripts.

As such, a logger is also implemented to keep track of whether any old clients are still calling the deprecated updateField endpoints.

Related to #1312

## Solution

**Features**:
- feat: add logger to track deprecated updateForm endpoint calls

**Bug Fixes**:

- fix: sync client hasAllowedEmailDomains and allowedEmailDomains states
- fix: watch field.isVerifiable to show relevant tooltip
- fix: sync inconsistent email field states in the backend

## Manual Tests
- [ ] Create an email field with OTP verification toggled on, then toggle on Restrict email domains, then fill in the allowed domains. Save the field. The returned field from the network request should contain 
  - `isVerifiable=true`, 
  - `hasAllowedEmailDomains=true`, 
  - `allowedEmailDomains=[<whatever you filled in>]`
- [ ] For the same field, toggle off OTP verification. Restrict email domains should automatically be toggled off and a toast for restrict email domains should show up. Save the field. The returned field from the network request should contain
  - `isVerifiable=false`,
  - `hasAllowedEmailDomains=false`,
  - `allowedEmailDomains=[]` // empty array
- [ ] For the same field, toggle OTP verification back on and save the field. The returned field should contain
  - `isVerifiable=true`,
  - `hasAllowedEmailDomains=false`,
  - `allowedEmailDomains=[]`

## Manual Tests (Programmatic)
Programatically submit an field creation request to the OLD endpoint (or copy and paste the below body snippet changing the `formId` to a form that you control): 

`POST <baseUrl>/<formId>/adminform`

```
{
  "form": {
    "editFormField": {
      "action": {
        "name": "CREATE"
      },
      "field": {
        "required": true,
        "disabled": false,
        "autoReplyOptions": {
          "hasAutoReply": false,
          "autoReplyMessage": "",
          "includeFormSummary": false
        },
        "isVerifiable": false,
        "hasAllowedEmailDomains": true,
        "allowedEmailDomains": [],
        "fieldName": "Email",
        "title": "Email",
        "fieldType": "email",
      }
    }
  }
}
```

Note the body is mimicking state 1 in the table above, where hasAllowedEmailDomains is true, but the array is empty. 
- [ ] The returned form field should manipulate the created field into state 4, a valid state
- [ ] Change the above body snippet to `hasAllowedEmailDomains=false` with a non-empty `allowedEmailDomains` array value. The returned form field should manipulate the created field also into state 4, since hasAllowedEmailDomains is false.

### DIFF
--- a/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
@@ -90,7 +90,7 @@ function EditFieldsModalController(
         '\n',
       )
     }
-    $scope.$watch('vm.field.allowedEmailDomainsFromText', (newValue) => {
+    $scope.$watch('vm.field.isVerifiable', (newValue) => {
       if (newValue) {
         vm.tooltipHtml = 'e.g. @mom.gov.sg, @moe.gov.sg'
       } else {
--- a/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
@@ -74,8 +74,18 @@ function EditFieldsModalController(
   vm.user = Auth.getUser() || $state.go('signin')
   if (vm.field.fieldType === 'email') {
     const userEmailDomain = '@' + vm.user.email.split('@').pop()
+
+    // Backwards compatibility and inconsistency fix.
+    // Set allowedEmailDomains array to empty if allow domains toggle is off.
+    if (vm.field.hasAllowedEmailDomains === false) {
+      vm.field.allowedEmailDomains = []
+    } else {
+      // hasAllowedEmailDomains is true, set "true" state based on length of allowedEmailDomains.
+      vm.field.hasAllowedEmailDomains = vm.field.allowedEmailDomains.length > 0
+    }
+
     vm.field.allowedEmailDomainsPlaceholder = `${userEmailDomain}\n@agency.gov.sg`
-    if (vm.field.allowedEmailDomains.length > 0) {
+    if (vm.field.hasAllowedEmailDomains) {
       vm.field.allowedEmailDomainsFromText = vm.field.allowedEmailDomains.join(
         '\n',
       )
@@ -142,6 +152,14 @@ function EditFieldsModalController(
       field.fieldOptions = field.fieldOptionsFromText.split('\n')
     } else {
       field.fieldOptions = []
+    }
+  }
+
+  vm.handleRestrictEmailDomainsToggle = function () {
+    const field = vm.field
+    if (field.hasAllowedEmailDomains === false) {
+      // Reset email domains.
+      field.allowedEmailDomainsFromText = ''
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
```diff
@@ -74,13 +74,23 @@ function EditFieldsModalController(
...
+    if (vm.field.hasAllowedEmailDomains === false) {
+      vm.field.allowedEmailDomains = []
+    } else {
+      // hasAllowedEmailDomains is true, set "true" state based on length of allowedEmailDomains.
+      vm.field.hasAllowedEmailDomains = vm.field.allowedEmailDomains.length > 0
+    }
+
     vm.field.allowedEmailDomainsPlaceholder = `${userEmailDomain}\n@agency.gov.sg`
-    if (vm.field.allowedEmailDomains.length > 0) {
+    if (vm.field.hasAllowedEmailDomains) {
```
mantariksh, line 88: if the end goal is to remove `hasAllowedEmailDomains`, why not just stick to using `allowedEmailDomains.length`?

karrui: Case 2 in the table above would incorrectly be enabled

karrui: Oh but you are right, we have already reset the array before his

mantariksh (approved):

src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
karrui: Clarified irl: hasAllowedEmailDomains is still needed in the client to show the state of the "Restrict email domains" toggle.

karrui:
Added manual tests for completeness, will still require DB script to clean up current state 1s and 2s in the DB. This is tracked in #1698

PR #1698, opened 2021-04-21
feat: remove emailField.hasAllowedEmailDomains key from schema + db script

### LINKED ISSUE #1312: Constrain email field domain model to prevent inconsistencies

karrui:
> This is the final PR of our journey to remove emailField.hasAllowedEmailDomains from the schema

**!!! This PR should only be merged after checking all old clients have stopped calling the old endpoints AND after the DB script has been run**

See #1697

This PR cannot be rolled back due to a db migration.

## Problem

This PR removes the `emailField.hasAllowedEmailDomains` key from email form field schema.

This PR also contains a DB script to remove the `emailField.hasAllowedEmailDomains` key and corrects all inconsistent values in the field documents.

Closes #1312

## Solution

**Features**:

- feat: remove hasAllowedEmailDomains key from email field
- feat: add cleanup script to sync email field states and remove bool

## Deploy Notes

Will need to run the migration script before and after the release.


karrui:
6 May 2021 update; still 16 calls in the last 3 days hitting the old endpoint, but none in the past day. Will monitor move.

karrui:
25 May 2021 update; 1 call on 20 May hitting the endpoint, will fix merge conflicts and schedule migration next week

karrui:
Closing until just before React

karrui:
17 June update, one last form still calling the endpoint, but the endpoint is not updating email fields. Should be gtg

mantariksh (commented):
don't we need to remove `hasAllowedEmailDomains` from the model as well?

scripts/20210421_has-allowed-email-domains/cleanup.js
```diff
@@ -0,0 +1,167 @@
+/* eslint-disable */
+
+// Optional: Retrieve forms with fields that has a populated allowedEmailDomains but has no allowed email domains.
+db.getCollection('forms').aggregate([
```
mantariksh, line 4: what does this query tell us?

scripts/20210421_has-allowed-email-domains/cleanup.js
```diff
@@ -0,0 +1,167 @@
...
+// Count forms with email fields isVerifiable=true, hasAllowedEmailDomains=true, non-empty allowedEmailDomains
+// Basically forms with restricted domain email fields
+
+// COUNT_A =
+
+db.getCollection('forms').count({
+  'form_fields.fieldType': 'email',
+  'form_fields.isVerifiable': true,
+  'form_fields.hasAllowedEmailDomains': true,
+  'form_fields.allowedEmailDomains': { $ne: [] },
```
mantariksh, line 31: we need an `$elemMatch` here right? otherwise this will pick up e.g. two separate email fields, one with isVerifiable true and one with hasAllowedEmailDomains true, but neither of which satisfies all the criteria

scripts/20210421_has-allowed-email-domains/cleanup.js
```diff
@@ -0,0 +1,167 @@
...
+})
+
+// Count forms with isVerifiable=true, but hasAllowedEmailDomains=false
+// Basically forms with verification but not restrictions
+
+// COUNT_B = 
+
+db.getCollection('forms').count({
+  'form_fields.fieldType': 'email',
```
mantariksh, line 40: `$elemMatch` needed here?

scripts/20210421_has-allowed-email-domains/cleanup.js
```diff
@@ -0,0 +1,167 @@
...
+  'form_fields.hasAllowedEmailDomains': false,
+})
+
+// Count forms with non-empty allowed email domains even if hasAllowedEmailDomains is false, 
+// so we can reset array before removing hasAllowedEmailDomains key.
+
+// COUNT_C = 
+
+db.getCollection('forms').count({
+  'form_fields.fieldType': 'email',
```
mantariksh, line 51: `$elemMatch` needed here?

scripts/20210421_has-allowed-email-domains/cleanup.js
```diff
@@ -0,0 +1,167 @@
...
+})
+
+// Count forms with inconsistent email states
+// hasAllowedEmailDomains = true but allowedEmailDomains is empty
+// Should not matter since allowedEmailDomains is the source of truth
+
+// COUNT_D = 
+
+db.getCollection('forms').count({
+  'form_fields.fieldType': 'email',
```
mantariksh, line 63: same here

scripts/20210421_has-allowed-email-domains/cleanup.js
```diff
@@ -0,0 +1,167 @@
...
+    $unset: {
+      'form_fields.$[elem].hasAllowedEmailDomains': 1,
+    }
+  },
+  {
+    arrayFilters: [
+      {
+        'elem.fieldType': 'email',
+        'elem.hasAllowedEmailDomains': false,
+        'elem.allowedEmailDomains': { $ne: [] },
```
mantariksh, line 93: this removes the need for the `$elemMatch` in the query portion on L76-78, but I think it might still be clearer to add it so that we only match fields which meet all the criteria

scripts/20210421_has-allowed-email-domains/cleanup.js
karrui: finding the inconsistent email fields itself so we have some forms to verify that the change has occurred

karrui:
we will need to perform the db migration script before releasing this, due to the client check that converts any non-empty emailAllowedDomain array into the toggle being enabled, and we may have forms with inconsistent checks

karrui:
> don't we need to remove `hasAllowedEmailDomains` from the model as well?

Thanks for the catch! What a thing to miss....

mantariksh (commented):
as discussed, we need to ensure that the public form side does not break when hasAllowedEmailDomains is purged from the database. preferably we should remove the dependence of `validate-email-domain` on hasAllowedEmailDomains

scripts/20210421_has-allowed-email-domains/cleanup.js
```diff
@@ -0,0 +1,209 @@
...
+
+// Optional: Retrieve forms with fields that has a populated allowedEmailDomains but has no allowed email domains.
+// For sampling to check if form has been fixed properly after migration.
+db.getCollection('forms').aggregate([
+  { $unwind: '$form_fields' },
+  {
+    $match: {
+      'form_fields.fieldType': 'email',
+      'form_fields.hasAllowedEmailDomains': false,
+      'form_fields.allowedEmailDomains': { $ne: [] },
```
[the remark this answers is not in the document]
mantariksh: edit: this won't work because of `$unwind`

mantariksh:
>we will need to perform the db migration script before releasing this

can we pls write down in the script (maybe even put in separate files) what exactly needs to be run before, and what exactly needs to be run after the code is released? it seems to me that if we do the whole db migration before releasing this, email domain validation will break because `hasAllowedEmailDomains` will always be `false` in the backend email validator

karrui:
> > we will need to perform the db migration script before releasing this
> 
> can we pls write down in the script (maybe even put in separate files) what exactly needs to be run before, and what exactly needs to be run after the code is released? it seems to me that if we do the whole db migration before releasing this, email domain validation will break because `hasAllowedEmailDomains` will always be `false` in the backend email validator

You're right... we STILL can't remove hasAllowedEmailDomains key, since old clients are still using that check...

i'll split it up into three PRs then.

First PR will handle
1. Correcting state inconsistency in DB that will arise when we switch from using `hasAllowedEmailDomains` as the source of truth to `hasAllowedEmailDomains`.
2. After DB migration is complete, possible inconsistencies should be fixed.
  - all `hasAllowedEmailDomains = false` will have `allowedEmailDomains = []`

Second PR will handle:
1. Updating the client to refer to allowedEmailDomains as the source of truth and stop referring to field's `hasAllowedEmailDomains` key and instead compute it from the length of `allowedEmailDomains`. Frontend will stop sending `hasAllowedEmailDomains` key.
2. Update schema to remove `hasAllowedEmailDomains` and instead set it as a virtual that computes from `allowedEmailDomains`....
3.  Add logger to see if frontend is still sending back `hasAllowedEmailDomains`

Third PR will:
1. Remove hasAllowedEmailDomains completely from the schema once the logging confirms that frontend has stopped sending that key.

karrui:
Closing this in favour of creating 3 separate piecemeal PRs... Hopefully for the last time.

PR #1701, opened 2021-04-22
fix: remove verified prefix on blank verified fields

### LINKED ISSUE #844: [verified] prefix should not be present on blank verified fields
Optional verified fields have the [verified] prefix if left blank

Screenshot 2020-12-10 at 3 49 32 PM

yong-jie:
## Problem
In the submission emails sent to form admins, verified fields have the [verified] prefix. However, this prefix is also present for optional verified fields that were left blank.

Closes #844

## Solution
Include an extra check to ensure that the field is not blank before adding the [verified] prefix.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


mantariksh (approved):
LG<sup>TM</sup>

PR #1702, opened 2021-04-22
feat: prefill mainstream launch

### LINKED ISSUE #1593: Prefill mainstream launch and scamproofing
Mainstream launch prefill for all short text fields
- Enable prefill toggle on short text fields
- Highlight prefilled fields on public forms

Design https://www.figma.com/file/ta5tjjBqphML8LPRkmxD53/DesignMaster_v2

tshuli:
## Problem

Closes #1593

Forms guide PR [here](https://github.com/opendocsg/opendoc-formsg-faq/pull/78) 

## Solution
- Add prefill toggle for short text field and highlight prefills for public user
- Note that fieldId is generated only upon field create

## Screenshots

**Admin page - Create Field**
Screenshot 2021-04-27 at 9 21 30 AM
Screenshot 2021-04-27 at 9 21 24 AM

**Admin Page - Edit Field**
Screenshot 2021-04-27 at 9 21 44 AM
Screenshot 2021-04-27 at 9 21 38 AM

**Public Form View**
Screenshot 2021-04-22 at 12 21 02 PM
Screenshot 2021-04-22 at 12 21 09 PM

## Tests
- [ ] Create two short text fields and set prefill toggle to true. Open form in public view. Check that the fields are prefillable by adding the query params to URL
- [ ] Check that prefill fields are highlighted in yellow
- [ ] Check that warning label appears if there are prefillable fields on form, and does not appear if there are no prefillable fields on form
- [ ] Hide prefill field with logic. Check that warning label only appears when field is visible.

### DIFF
--- a/src/public/modules/forms/admin/views/edit-fields.client.modal.html
+++ b/src/public/modules/forms/admin/views/edit-fields.client.modal.html
@@ -1110,7 +1110,6 @@
               <label
                 class="toggle-selector pull-right"
                 ng-class="vm.field.allowPrefill ? 'toggle-selector-on' : ''"
-                onclick=""
               >
                 <input type="checkbox" ng-model="vm.field.allowPrefill" />
                 <div class="toggle-selector-switch">
--- a/src/public/modules/forms/admin/views/edit-fields.client.modal.html
+++ b/src/public/modules/forms/admin/views/edit-fields.client.modal.html
@@ -1134,16 +1134,20 @@
           <div class="col-md-12">
             <input
               type="text"
-              ng-disabled="true"
+              ng-readonly="true"
               class="input-custom input-medium"
               value="{{vm.field._id || 'Field ID will be generated after field is saved'}}"
+              id="prefill-id"
             />
             <i
               class="bx bx-copy bx-md copy-icon-overlay prefill-copy-icon"
               ng-if="vm.field._id"
               ngclipboard
-              ng-click="vm.fieldIdCopied()"
-              data-clipboard-text="{{vm.field._id}}"
+              ng-click=""
+              data-clipboard-action="copy"
+              data-clipboard-target="#prefill-id"
+              ngclipboard-success="vm.toastSuccessfulFieldIdCopy()"
+              ngclipboard-error="vm.toastFailedFieldIdCopy()"
             ></i>
           </div>
           <div class="col-md-12">
--- a/src/public/modules/forms/admin/views/edit-fields.client.modal.html
+++ b/src/public/modules/forms/admin/views/edit-fields.client.modal.html
@@ -1087,6 +1087,44 @@
         </div>
         <!-- End custom validation field -->
 
+        <!-- Short Text -->
+
+        <div
+          ng-show="vm.field.fieldType === 'textfield'"
+          class="edit-fields-line"
+        ></div>
+
+        <div ng-if="vm.field.fieldType === 'textfield'" class="row">
+          <div class="row"><br /></div>
+          <div class="toggle-option">
+            <div class="col-xs-8 label-custom label-medium">
+              Enable pre-fill
+              <i
+                class="glyphicon glyphicon-question-sign"
+                uib-tooltip="Allow this field to be pre-populated via URL"
+                tooltip-trigger="'click mouseenter'"
+              >
+              </i>
+            </div>
+            <div class="col-xs-4 field-input">
+              <label
+                class="toggle-selector pull-right"
+                ng-class="vm.field.allowPrefill ? 'toggle-selector-on' : ''"
+                onclick=""
+              >
+                <input type="checkbox" ng-model="vm.field.allowPrefill" />
+                <div class="toggle-selector-switch">
+                  <i
+                    ng-class="vm.field.allowPrefill ? 'bx bx-check' : 'bx bx-x' "
+                  ></i>
+                </div>
+              </label>
+            </div>
+          </div>
+        </div>
+
+        <!-- End Short Text -->
+
         <!-- Decimal -->
         <div ng-if="vm.field.fieldType === 'decimal'">
           <div class="row">
--- a/src/public/modules/forms/admin/views/edit-fields.client.modal.html
+++ b/src/public/modules/forms/admin/views/edit-fields.client.modal.html
@@ -1123,6 +1123,37 @@
           </div>
         </div>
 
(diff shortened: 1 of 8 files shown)


src/public/modules/forms/base/componentViews/field-textfield.client.view.html
```diff
@@ -29,6 +29,7 @@
       ng-required="vm.field.required"
       ng-disabled="vm.field.disabled"
       ng-pattern="vm.pattern"
+      ng-style="vm.field.allowPrefill && {'background-color': '#FEF8E3'}"
```
liangyuanruo, line 32: would it be possible to avoid the use of `ng-style` and rely on `ng-class` instead, to allow the `background-color` CSS property be specified in a CSS file?

src/public/modules/forms/admin/views/edit-fields.client.modal.html
```diff
@@ -1087,6 +1087,85 @@ <h2 class="modal-title">
         </div>
         <!-- End custom validation field -->
 
+        <!-- Short Text -->
```
liangyuanruo, line 1090: is there an existing block of HTML for short text? can these HTML elements be relocated there? i know this block of code is in need of a big refactor, but we won't be needing it in the React refactor.

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -429,6 +429,10 @@ function submitFormDirective(
           }
         })
       }
+      // Variable to indicate if there are prefill fields
+      scope.hasPrefill = scope.form.form_fields.some((field) => {
+        return field.allowPrefill
+      })
```
liangyuanruo, line 435: can I clarify on what's the expected behavior if prefill is used in conjunction with logic? would that necessitate the use of `scope.$watch`?

src/public/modules/forms/admin/views/edit-fields.client.modal.html
```diff
@@ -1087,6 +1087,85 @@ <h2 class="modal-title">
...
+                uib-tooltip="Allow this field to be pre-populated via URL"
+                tooltip-trigger="'click mouseenter'"
+              >
+              </i>
+            </div>
+            <div class="col-xs-4 field-input">
+              <label
+                class="toggle-selector pull-right"
+                ng-class="vm.field.allowPrefill ? 'toggle-selector-on' : ''"
+                onclick=""
```
liangyuanruo, line 1113: is the `onclick` property suppose to be empty?

src/public/modules/forms/admin/views/edit-fields.client.modal.html
```diff
@@ -1087,6 +1087,85 @@ <h2 class="modal-title">
...
+              class="input-custom input-medium"
+              value="{{vm.field._id || 'Field ID will be generated after field is saved'}}"
+            />
+            <i
+              class="bx bx-copy bx-md copy-icon-overlay prefill-copy-icon"
+              ng-if="vm.field._id"
+              ngclipboard
+              ng-click="vm.fieldIdCopied()"
+              data-clipboard-text="{{vm.field._id}}"
+            ></i>
```
liangyuanruo, line 1147: i'm looking at the code example from the [docs](http://sachinchoolur.github.io/ngclipboard/) - should we be using `data-clipboard-action="copy"` and `data-clipboard-target`, as well as using `ngclipboard-success` (and `ngclipboard-error` instead of `ng-click`?

```html
<!-- Target -->
<textarea id="bar">Mussum ipsum cacilds...</textarea>

<!-- Trigger -->
<button class="btn" ngclipboard data-clipboard-action="cut" data-clipboard-target="#bar">
    Cut to clipboard
</button>
```

liangyuanruo (changes requested):
Proposing some minor refactors as well as clarification on behavior - thanks!

src/public/modules/forms/admin/views/edit-fields.client.modal.html
tshuli: not as far as I can see... Short Text shares the same elements as Long Text and Number. This would be the first unique element for Short Text.

src/public/modules/forms/base/directives/submit-form.directive.js
tshuli: thanks, the warning message should show only when the field is visible. adjusted accordingly.

src/public/modules/forms/admin/views/edit-fields.client.modal.html
tshuli: yes, it is to provide the mouseover hand, the actual logic comes from the checkbox being ticked

src/public/modules/forms/admin/views/edit-fields.client.modal.html
tshuli: edited, although this means that the fieldId field has to be set to readonly instead of disabled, slight change from original design (see [link](https://github.com/zenorocha/clipboard.js/issues/41), the issue seems to affect chrome now)

src/public/modules/forms/base/componentViews/field-textfield.client.view.html
tshuli: edited

tshuli:
@liangyuanruo thanks for reviewing, for re-review pls

src/public/modules/forms/admin/views/edit-fields.client.modal.html
liangyuanruo: you should use the CSS `cursor` property

src/public/modules/forms/admin/views/edit-fields.client.modal.html
liangyuanruo: in that case, we should invest in the CSS to have it work with `readonly`. generally, we shouldn't be overriding designers' design without consulting with them first.

liangyuanruo (changes requested):
need to align implementation according to the designers' intent - I was wondering why the input boxes were not greyed out.

src/public/modules/forms/admin/views/edit-fields.client.modal.html
tshuli: edited

src/public/modules/forms/admin/views/edit-fields.client.modal.html
tshuli: edited

liangyuanruo (approved):

PR #1713, opened 2021-04-24
feat: Add webhook support for storage mode attachments

frankchn:
## Problem
This adds webhook support for encrypted attachments on the server side. This does not include any SDK changes that needs to go in for full support

The design sketch for webhook support can be found [here](https://docs.google.com/document/d/e/2PACX-1vRhiNPSNeitXfQXoFYLjOReunLs7nJ8sS08dFiRch38ZUL-bKct1OyZNalW6j4wcRrUkcgujklmIqOh/pub).


src/app/models/submission.server.model.ts
```diff
@@ -182,13 +182,21 @@ const EncryptSubmissionSchema = new Schema<
 EncryptSubmissionSchema.methods.getWebhookView = function (
   this: IEncryptedSubmissionSchema,
 ): WebhookView {
+  const attachmentRecords: Record<string, string> = {}
+  if (this.attachmentMetadata) {
+    for (const [k, v] of this.attachmentMetadata) {
+      attachmentRecords[k] = v
+    }
+  }
```
karrui, line 190: nit: can be converted from Map to Object with

```ts
const attachmentRecords = Object.fromEntries(this.attachmentMetadata)
```

src/app/modules/webhook/webhook.service.ts
```diff
@@ -97,81 +111,91 @@ export const sendWebhook = (
...
-      logger.info({
-        message: 'Webhook POST succeeded',
-        meta: {
-          ...logMeta,
-          status: get(response, 'status'),
-        },
+    return new WebhookFailedWithUnknownError(error)
+  }).andThen((signedUrls) => {
+    submissionWebhookView.data.attachmentDownloadUrls = signedUrls
+    return ResultAsync.fromPromise(validateWebhookUrl(webhookUrl), (error) => {
```
karrui, line 123: should we validate the webhook as the first step in the chain? Might be generating attachment URLs for naught if the webhook is invalid

src/public/modules/forms/admin/directiveViews/settings-form.client.view.html
```diff
@@ -410,7 +401,7 @@
               ng-model="tempForm.webhook.url"
               placeholder="https://your-webhook.com/url"
               ng-required="false"
-              ng-disabled="isFormPublic() || doesFormContainAttachments()"
```
karrui, line 413: may still want to block it on the client first, since there is no way to actually consume the attachment URLs if its sent to the webhook

src/app/modules/webhook/webhook.service.ts
```diff
@@ -88,6 +90,18 @@ export const sendWebhook = (
     epoch: now,
   })
 
+  // Generate S3 signed urls
+  const signedUrlPromises: Record<string, Promise<string>> = {}
```
karrui, line 94: should unit tests be added in `webhook.service.spec.ts` for checking if the webhook is being sent with the expected attachment URLs?

aws can be mocked in the unit test with 

```ts
// webhook.service.spec.ts
import { aws as AwsConfig } from 'src/app/config/config'

...
jest
  .spyOn(AwsConfig.s3, 'getSignedUrlPromise')
  .mockImplementation((_operation, params) => {
    return Promise.resolve(`https://some-fake-url/${params.Key}/${params.Expires}`)
  })
```

src/app/models/submission.server.model.ts
mantariksh: ^ just beware that this will cause a TypeError if `this.attachmentMetadata` is `undefined`, which it may be for submissions created before we had storage mode attachments. so we need to either retain the `if (this.attachmentMetadata.)` or do `Object.fromEntries(this.attachmentMetadata ?? new Map())`.

src/public/modules/forms/admin/directiveViews/settings-form.client.view.html
mantariksh: ^ sorry what do you mean by this?

karrui: undo this client change? since the formsg sdk still does not have an utility function to decrypt the attachment object? May result in some form admins adding a webhook for their form with attachments that they do not know what to deal with. 

but then again, not like responses are lost; the attachments can still be retrieved from the responses tab. May be a moot point!

mantariksh: oh yeah sure, I agree with that. unless we can make the SDK changes before this code is merged. the SDK changes seem pretty straightforward, since we already have all the required utilities for decryption.

src/app/modules/webhook/webhook.service.ts
```diff
@@ -97,81 +111,91 @@ export const sendWebhook = (
...
-    )
-    .map((response) => {
-      // Capture response for logging purposes
-      logger.info({
-        message: 'Webhook POST succeeded',
-        meta: {
-          ...logMeta,
-          status: get(response, 'status'),
-        },
+    return new WebhookFailedWithUnknownError(error)
```
mantariksh, line 120: I realise this is not apparent from the error name (sorry about that) but `WebhookFailedWithUnknownError` was intended for when the webhook POST request fails with an unknown error. we should create a new error class for when the presigned URLs fail to be generated.

src/public/modules/forms/admin/directiveViews/settings-form.client.view.html
frankchn: Yeah we can leave this PR out of the develop branch until the corresponding SDK changes goes in too.

src/app/modules/webhook/webhook.service.ts
frankchn: No worries, adding `WebhookFailedWithPresignedUrlGenerationError`

src/app/modules/webhook/webhook.service.ts
```diff
@@ -97,81 +111,91 @@ export const sendWebhook = (
...
+  }).andThen((signedUrls) => {
+    submissionWebhookView.data.attachmentDownloadUrls = signedUrls
```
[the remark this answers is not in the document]
frankchn: So that was what I was thinking originally when I suggesting putting the generation in `getWebhook` so it returns the full S3 presigned URL. Should I put it there or create a new function to convert what is returned by `getWebhook`?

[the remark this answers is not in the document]
mantariksh: I think the latter! ideally `getWebhookView` should only return some view of the data, and we should keep application logic in the service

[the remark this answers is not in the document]
frankchn: Cool, split it into a separate function.

mantariksh (commented):
lgtm, let's wait until the corresponding SDK changes are ready and tested before we merge this to develop

karrui (commented):
lgtm

karrui:
Adding a blocked label until the sdk merging is resolved

mantariksh (changes requested):
a couple of issues I found while testing on staging:
1. attachment fields are still disabled in the Build tab when the form has a webhook. we should remove that restriction for both the desktop and mobile views.
2. there is model-level validation to ensure that attachment fields do not get saved when the form has a webhook. we should remove that as well.

mantariksh (dismissed):

mantariksh:
lgtm, will wait for SDK release before merging

liangyuanruo:
> Waiting for a few weeks for webhook developers to make the necessary changes before releasing this.

I think we could also ask form admins to write back to us if it's a breaking change for them - maybe that could bring forward the timeline?

src/public/modules/forms/admin/directives/edit-form.client.directive.js
```diff
@@ -341,9 +341,8 @@ function editFormController(
 
   $scope.isStorageForm = $scope.myform.responseMode === responseModeEnum.ENCRYPT
 
-  // Disable attachment fields when we have webhooks
-  $scope.isDisabledField = function (fieldType) {
-    return fieldType.name === 'attachment' && $scope.myform.webhook.url !== ''
+  $scope.isDisabledField = function () {
+    return false
```
liangyuanruo, line 345: can we simply remove this function entirely?

frankchn: Sure, but I'll do it in a separate PR?

PR #1722, opened 2021-04-26
fix: show correct error message for e-service ID validation

mantariksh:
## Problem

Singpass e-service ID validation relies on parsing the HTML of the login page. The login page recently added a semicolon in the line showing the error code, resulting in the FormSG server returning `null` for the error code. This causes the frontend to show an error message saying "Could not connect SingPass to validate the e-service ID", instead of showing that the e-service ID is invalid.

## Solution

Add the semicolon to the parsing code, and also make the comparison case-insensitive.

### Before
Screenshot 2021-04-26 at 10 47 52 AM

### After
image

### DIFF
--- a/src/app/modules/spcp/spcp.service.ts
+++ b/src/app/modules/spcp/spcp.service.ts
@@ -178,8 +178,8 @@ export class SpcpService {
     } else {
       // The error page should have text like 'System Code:&nbsp<b>138</b>'
       const errorCode = getSubstringBetween(
-        loginHtml,
-        'System Code:&nbsp<b>',
+        loginHtml.toLowerCase(),
+        'System code:&nbsp;<b>'.toLowerCase(),
         '</b>',
       )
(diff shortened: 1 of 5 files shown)


src/app/modules/spcp/spcp.service.ts
```diff
@@ -178,8 +178,8 @@ export class SpcpService {
     } else {
       // The error page should have text like 'System Code:&nbsp<b>138</b>'
       const errorCode = getSubstringBetween(
-        loginHtml,
-        'System Code:&nbsp<b>',
+        loginHtml.toLowerCase(),
+        'System code:&nbsp;<b>'.toLowerCase(),
```
karrui, line 182: nit: can't we just hardcode the lowercased version of this string haha
```suggestion
        'system code:&nbsp;<b>',
```

karrui (approved):
lgtm

PR #1723, opened 2021-04-26
build(deps): bump mockpass and spcp-auth-client

LoneRifle:
## Problem and Solution
Update mockpass and spcp-auth-client to pick up the latest transitive dependencies, 
notably de-duping xpath versions thanks to updates to xml-encryption to bring it in
line with xml-crypto


mantariksh (approved):
lgtm although the additional tab width is offensive to @karrui's faith

LoneRifle:
How did that even happen... let me rework this.

PR #749, opened 2020-11-30
refactor: migrate date validator to ts

tshuli:
## Problem
Part of #7

## Solution

**Improvements**

- Migrates Date validation to Typescript

## Tests
- [ ] Create two date fields. Hide one of them behind logic.
- [ ] Check that validation on both visible and hidden date fields work when submitting programatically
 - Date input is rejected if it is not in the format 'DD MMM YYYY'
 - Date input is rejected if it is not within the validation range (isFutureOnly, isPastOnly, Custom Min / Max)

PR #1724, opened 2021-04-26
refactor: use shared DateSelectedValidation enum instead of DATE_VALIDATION_OPTIONS

### LINKED ISSUE #777: Frontend to use shared DateValidationOptions enum instead of DATE_VALIDATION_OPTIONS
Related to #749

yong-jie:
## Problem
An enum called `DateSelectedValidation` was introduced during a typescript migration task #749, but the frontend codebase has not been updated to make use of this enum.

Closes #777 

## Solution
Remove the naming conflict between `DateValidationOptions` and `DATE_VALIDATION_OPTIONS`. The former is actually a list of options and the latter is the enum. The former is now called DateValidationOptionList.

Edit usages of the enum from lower-case to title-case to match the new enum. i.e. `DateValidationOptions.custom` to `DateValidationOptions.Custom`

## Reviewers

@tshuli appreciate if you could take a quick look to see if these changes are indeed what you had intended.

### DIFF
--- a/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
@@ -266,7 +266,7 @@ function EditFieldsModalController(
 
   vm.dateValidationOptionList = values(DateValidationOptions)
 
-  // Make DATE_VALIDATION_OPTIONS accessible to view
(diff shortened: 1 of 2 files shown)


karrui (commented):
lgtm, but wait for @tshuli for final say hehe

tshuli (approved):
thanks! lgtm :)

PR #1725, opened 2021-04-26
refactor(api): collapse /validate into single endpoint

### LINKED ISSUE #1517: [API] Collapse SP/CP/MyInfo login endpoints to new /api counterpart

### LINKED ISSUE #1518: Convert SPCP factories to TypeScript
This issue encompasses both _spcp-redirect.client.factory.js_ and _spcp-validate-esrvcid.client.factory.js_, which should be merged into a single Singpass / Corppass service.

seaerchin:
## Problem
This is part 2 of #1517. See write up in part 1 (#1672 for reasoning behind why this is needed) 

Closes #1517 
Closes #1518 

## Solution
This collapses the `myInfo` and `spcp` validate endpoints into one. similarly, the frontend side calls `PublicFormAuthService` exclusively now. 

## Notes
Unit tests are contained within this PR but integration tests for `/validate` will be in a follow up PR

## Manual Tests
**Note**: the following have already been tested on staging. lmk if there's anything i missed out!

- [ ] Create a new SP form. Go into the admin panel and key in an **invalid** eServiceId. Attempt to activate the form. The form should **not** be activated. 
- [ ] Create a new SP form. Go into the admin panel and key in a **valid** eServiceId. Attempt to activate the form. The form **should** be activated. 
- [ ] Create a new MyInfo form. Go into the admin panel and key in an **invalid** eServiceId. Attempt to activate the form. The form should **not** be activated. 
- [ ] Create a new MyInfo form. Go into the admin panel and key in a **valid** eServiceId. Attempt to activate the form. The form **should** be activated.


src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -451,3 +456,52 @@ export const handleFormAuthRedirect = [
...
+      // And because MyInfo login is beyond our control, we coerce MyInfo to SP.
+      // This is valid because a valid MyInfo eserviceId is also a valid SP eserviceId
+      if (form.authType === AuthType.MyInfo) {
+        return validateMyInfoForm(form).andThen((form) =>
+          SpcpFactory.createRedirectUrl(AuthType.SP, formId, form.esrvcId),
+        )
+      }
+      return validateSpcpForm(form).andThen((form) =>
+        SpcpFactory.createRedirectUrl(form.authType, formId, form.esrvcId),
+      )
```
karrui, line 489: (i know this is wip, just happened to see it)

should we split up the steps to 
1. a validate step and
2. the creation of redirect url step so we don't need to repeat the same steps twice?

```ts
// redirect step
const redirectAuthType = form.authType === AuthType.MyInfo ? AuthType.SP : form.authType
return SpcpFactory.createRedirectUrl(redirectAuthType, formId, form.esrvcId),
```

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -451,3 +456,52 @@ export const handleFormAuthRedirect = [
...
+  { formId: string },
+  PublicFormAuthValidateEsrvcIdDto | ErrorDto
+> = (req, res) => {
+  const { formId } = req.params
+  return FormService.retrieveFormById(formId)
+    .andThen((form) => {
+      // NOTE: Because the check is based on parsing the html of the returned webpage,
+      // And because MyInfo login is beyond our control, we coerce MyInfo to SP.
+      // This is valid because a valid MyInfo eserviceId is also a valid SP eserviceId
+      if (form.authType === AuthType.MyInfo) {
```
karrui, line 482: Suggest switch-case, what if the auth type is NIL?
is there also a regression? this used to only be called by SP/MyInfo forms, and not for CP forms since we cannot validate CP forms. 

Should we also check if that has changed, since CP is now logged in via SP?

karrui (commented):
i know wip, just preliminary review since I was waiting for my local dev to compile lol

src/app/modules/form/public-form/public-form.controller.ts
seaerchin: didn't do this because i wanted the coercion from `MyInfo` to `SP` to be clear. although it's slightly longer, i think the explicitness is worth the length increase!

src/app/modules/form/public-form/public-form.controller.ts
seaerchin: good catch on the first point! i'll maintain backwards compatibility for the `CP` one first and then raise an issue to investigate if it's possible!

src/public/modules/forms/admin/controllers/activate-form-modal.client.controller.js
```diff
@@ -98,8 +96,9 @@ function ActivateFormController(
       vm.esrvcIdStatus = 1
     })
 
-    if (authType === 'SP' && esrvcId !== '') {
-      return SpcpValidateEsrvcId(target, authType, esrvcId)
+    if ((authType === 'SP' && esrvcId !== '') || authType === 'MyInfo') {
```
karrui, line 99: is the difference in handling for SP vs MyInfo forms required?

src/public/modules/forms/admin/controllers/activate-form-modal.client.controller.js
```diff
@@ -119,24 +118,6 @@ function ActivateFormController(
       // CorpPass doesn't return any error page even with the wrong e-service id
```
karrui, line 118: if the endpoint is not called for CP forms, why is there a switch case for CP forms in server controller handleValidateFormEsrvcId function?

src/public/services/__tests__/PublicFormAuthService.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+jest.mock('axios')
+const MockAxios = mocked(axios, true)
+
+const MOCK_REDIRECT_URL = 'redirectURL'
+const MOCK_FORM_ID = new ObjectId().toHexString()
+
+describe('PublicFormAuthService', () => {
+  afterEach(() => jest.resetAllMocks())
+  describe('createRedirectURL', () => {
```
karrui, line 15: I think this test suite is left over from rebasing

src/public/services/__tests__/PublicFormAuthService.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      expect(MockAxios.get).toHaveBeenCalledWith(
+        `${PublicFormAuthService.PUBLIC_FORMS_ENDPOINT}/${MOCK_FORM_ID}/auth/redirect`,
+        {
+          params: { isPersistentLogin: false },
+        },
+      )
+    })
+  })
+
+  describe('validateESrvcId', () => {
```
karrui, line 78: nit: typo in casing

src/public/services/__tests__/PublicFormAuthService.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      const result = await PublicFormAuthService.validateEsrvcId(MOCK_FORM_ID)
+
+      expect(MockAxios.get).toHaveBeenCalledWith(
+        `${PublicFormAuthService.PUBLIC_FORMS_ENDPOINT}/${MOCK_FORM_ID}/auth/validate`,
+      )
+      expect(result).toEqual(mockData)
+    })
+
+    it('should return isValid as false and an errorcode when the e-service ID is invalid', async () => {
```
karrui, line 91: is this second test actually needed? this test is just mocking the result and asserting the result, and seems no different from the first test.

should we just have a single test with the description 

```ts
it('should call validate endpoint successfully', async () => { ... })
```

src/app/modules/form/public-form/public-form.utils.ts
```diff
@@ -110,3 +114,59 @@ export const mapFormAuthRedirectError: MapRouteError = (
       }
   }
 }
+
+/**
+ * Handler to map ApplicationErrors from validating a form's eServiceId to HTTP errors
+ * @param error The error to retrieve the status codes and error messages
+ * @param coreErrorMessage Any error message to return instead of the default core error message, if any
+ */
+export const mapValidateEsrvcIdError: MapRouteError = (
```
karrui, line 123: is a totally separate error mapping for the validate endpoint needed? Seems to be the same as the auth redirect mapping. Should the two be combined, e.g. `mapFormAuthError`?

karrui (changes requested):
preliminary comments and changes requested

src/public/modules/forms/admin/controllers/activate-form-modal.client.controller.js
seaerchin: good catch. i'll edit this out!

src/public/modules/forms/admin/controllers/activate-form-modal.client.controller.js
seaerchin: the switch case isn't called from the frontend but for the old backend endpoint, it accepts both `SP | CP` and treats them equally. hence, this change is made here to ensure that the behavior is identical (but an issue has been raised so that the frontend will call when `authType === CP`)!

src/public/services/__tests__/PublicFormAuthService.test.ts
seaerchin: yeap! this is because the tests PR isn't in yet due to an error from my end. idt this will show up when it is merged into `develop`!

src/public/services/__tests__/PublicFormAuthService.test.ts
seaerchin: you've made a really good point here. i think my approach to this was just to cover grounds and made sure that all the cases were hit but should've thought throw the reasoning behind each unit test. 

will edit and catch this in the future!

src/app/modules/form/public-form/public-form.utils.ts
seaerchin: i'm not too sure about this. my primary reasoning was because as per the previous one, maybe they are different enough to be separated (and exhaustive checking for cases). 

i'll merge them in if you feel it's too much!

karrui (approved):
lgtm

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -451,3 +452,59 @@ export const handleFormAuthRedirect = [
...
+            SpcpFactory.createRedirectUrl(form.authType, formId, form.esrvcId),
+          )
+        default:
+          return err<never, AuthTypeMismatchError>(
+            new AuthTypeMismatchError(AuthType.SP, form.authType),
+          )
+      }
+    })
+    .andThen(SpcpFactory.fetchLoginPage)
+    .andThen(SpcpFactory.validateLoginPage)
```
mantariksh, line 495: very nice

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -451,3 +452,59 @@ export const handleFormAuthRedirect = [
...
+      switch (form.authType) {
+        case AuthType.MyInfo:
+          return validateMyInfoForm(form).andThen((form) =>
+            SpcpFactory.createRedirectUrl(AuthType.SP, formId, form.esrvcId),
+          )
+        case AuthType.SP:
+        case AuthType.CP:
+          return validateSpcpForm(form).andThen((form) =>
+            SpcpFactory.createRedirectUrl(form.authType, formId, form.esrvcId),
+          )
```
mantariksh, line 487: I don't think we should be supporting Corppass here, as the error page is different from the Singpass error page
Screenshot 2021-04-28 at 10 35 48 AM

src/public/modules/forms/admin/controllers/activate-form-modal.client.controller.js
mantariksh: as commented elsewhere, I don't think we should support corppass for this

mantariksh (changes requested):
to clarify whether we are actually able to support a common validation flow for singpass and corppass

src/app/modules/form/public-form/public-form.controller.ts
seaerchin: Good point! I can take a (brief) look at this and the html structure of the page just to check if it's possible to validate using html and string matching. 

If it's not, I'll probably just lump this together with the default case!

mantariksh: err don't waste time on that imo, we have no intention of supporting corppass e-service ID validation anytime soon. we would much rather push for a proper validation endpoint.

src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
```diff
@@ -1488,4 +1490,335 @@ describe('public-form.controller', () => {
...
+        mockRes,
+        jest.fn(),
+      )
+
+      // Assert
+      expect(mockRes.status).toBeCalledWith(404)
+      expect(mockRes.json).toBeCalledWith(expectedError)
+    })
+
+    it('should return 500 when the returned html has no title', async () => {
```
mantariksh, line 1714: actually I think we shouldn't be returning a 5xx here, but let's preserve existing behaviour for now
```suggestion
    it('should return 502 when the returned html has no title', async () => {
```

src/app/modules/form/public-form/public-form.utils.ts
```diff
@@ -89,6 +93,16 @@ export const mapFormAuthRedirectError: MapRouteError = (
...
+        errorMessage: 'Failed to contact SingPass. Please try again.',
+      }
+    case LoginPageValidationError:
+      return {
+        statusCode: StatusCodes.BAD_GATEWAY,
+        errorMessage: 'Error while contacting SingPass. Please try again.',
+      }
     case DatabaseError:
     case CreateRedirectUrlError:
     case MissingFeatureError:
```
mantariksh, line 108: can change the action in the log message below to 'mapFormAuthError'

mantariksh (approved):

PR #1726, opened 2021-04-26
feat(api-refactor): implement specific reorder field api

### LINKED ISSUE #1482: [API] Extract form field reorder endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **reorder form fields**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## Proposed new endpoint
With that, I propose the new endpoint to be

```
POST /:formId/fields/:fieldId/reorder?to=<newIndex>
```

karrui:

## Problem

This PR adds a specific endpoint to reorder form fields in the form, and updates the frontend client to use that endpoint when reordering fields. There is also a feature to undo the field reordering if the reorder network request fails for one reason or another.

Closes #1482

## Solution

**Backend Features**:
- feat: add ExtractTypeFromArray type
- feat(FormModel): add reorderFormFieldById instance method
- feat(AdminFormSvc): add reorderFormField service function
- feat(AdminFormCtl): add handleReorderFormField controller handler fn
- feat(AdminFormRoutes): add route for reordering form field

**Client Features**
- feat(client): add reorderSingleFormField client service fn
- feat(client): call new reorder field endpoint

**Improvements**:

- ref: move immutable-array-fns to /shared directory 

**Bug Fixes**:
- feat: rollback form fields if reordering encounters error

## Tests

- test(FormModel): add tests for reorderFormFieldById
- test(AdminFormSvc): add unit tests for reorderFormField
- test(AdminFormCtl): add unit tests for _handleReorderFormField

## Manual tests
- [ ] Turn off network (from chrome devtools) and attempt to reorder a field. An error toast should occur and the form field ordering should revert to initial ordering
- [ ] Reorder a field. The returned form_fields shape should be the new ordering.
- [ ] Open two builder tabs in your browser for the same form (with at least 2 initial fields). Add a field on the second tab (should now have +1 fields). Go back to the first tab (with the initial field count) and reorder the fields. 3 fields should show up but correctly ordered.


src/app/models/form.server.model.ts
```diff
@@ -530,6 +531,27 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     return this.save()
   }
 
+  FormDocumentSchema.methods.reorderFormFieldById = function (
+    this: IFormDocument,
+    fieldId: string,
+    newPosition: number,
+  ) {
```
mantariksh, line 538: nit: can specify return type

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1545,5 +1545,61 @@ describe('Form Model', () => {
...
+        ).toEqual([
+          // Should be rearranged to the 0th index position, and the previously
+          // 0th index field should be pushed to 1st index.
+          originalFields[1],
+          originalFields[0],
+          originalFields[2],
+        ])
+      })
+
+      it('should return null if given fieldId is invalid', async () => {
```
mantariksh, line 1594: do we need a test for when the new position is greater than the length of the array?

karrui: nice catch, i thought there were already tests for `immutable-array-fns`. Will add them. 

for now, the newPosition is determined by slice and can be negative or positive etc, slice will either slice backwards or slice to the end. do you think this should be the behaviour or do we need guards for that?

mantariksh: I remember we had this discussion the first time you implemented `immutable-array-fns`, and I think we concluded that the functions should just be dumb and it should be the responsibility fo the caller to validate the `newIndex`. in this case we already have Joi validation to ensure that the new index is >0, which I think is sufficient to cover the negative case, which is why I figured we should just add tests for the >length case.

idt we need additional tests for the negative case. it's like any other Joi validation, eg we don't have tests for when a service is called with an invalid fieldType because we assume the fieldType has been Joi-validated

karrui: noted, will add tests for the newPosition > fields.length. (but yea, since immutable-array-fns is using slice under the hood, will be safe)

src/app/models/form.server.model.ts
karrui: resolved in 43381c6

src/app/models/__tests__/form.server.model.spec.ts
karrui: Resolved in 7f46d72

PR #1727, opened 2021-04-26
feat(spcp): raise sp cookie max age to 3 hrs

### LINKED ISSUE #1674: raise minimum timeout for sp, cp, myinfo to 3h
past feedback that 1h was too short for some long forms

yong-jie:
## Problem
Past feedback that 1h was too short for some long forms

Closes #1674 

## Solution
Modify the config's default for SP session age to 3 hours, up from the original 1 hour. Corp pass already has their session age to be 6 hours, so no change required on that front.


karrui (approved):
el gee tee em

PR #1685, opened 2021-04-20
refactor(test): integration and unit tests for redirect

seaerchin:
This PR is a follow on from #1672. This PR adds the following tests 

1. Unit tests for PublicFormController's new method (`_handleFormAuthRedirect`)
2. Integration tests for the new api end point (`/auth/redirect`) 
3. Frontend unit test for `AuthService`

PR #1728, opened 2021-04-26
refactor(test): integration and unit tests for redirect 

seaerchin:
This is reopened because #1685 got merged into its dependent branch instead of develop @__@ sorry for the time wastage


src/app/routes/api/v3/forms/__tests__/public-forms.auth.routes.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+
+      // Act
+      const response = await request
+        .get(`/forms/${form._id}/auth/redirect`)
+        .query({ isPersistentLogin: false })
+
+      // Assert
+      expect(response.status).toEqual(StatusCodes.OK)
+      expect(response.body).toMatchObject({
+        redirectURL: expect.any(String),
```
karrui, line 56: do we know the shape of the redirectURL and how it differs when isPersistentLogin is true vs false? We should test that the return string fits a substring of some valid string etc.

karrui (approved):
lgtm with one question

src/app/routes/api/v3/forms/__tests__/public-forms.auth.routes.spec.ts
seaerchin: i could check with anty/alwyn on this and make the check stronger. cos atm this is actually handled by an external library that's written by them HAHA

seaerchin: the redirectURL is created using an external library and for `authType = SP | CP`, the target has the form `(formId, {isPersistentLogin})` but not too sure for `MyInfo`. didn't wanna do that cos i thought external library supposed to be a black box @_@

i can clarify with anty/alwyn (cos they wrote the library) and update this!

karrui: since this is an integration test, we should assert the string returned by the third party does not change (in the syntax/style) at least

seaerchin: ah ok, good point! will do so (and will clarify for `MyInfo`)

seaerchin: read through `MyInfo` implementation, the implementation is [here](https://github.com/opengovsg/myinfo-gov-client/blob/c6d7d24b5a7813631ea12005b3446641d8261bec/src/MyInfoGovClient.class.ts#L172) and i'll update it accordingly!

src/app/routes/api/v3/forms/__tests__/public-forms.auth.routes.spec.ts
```diff
@@ -53,7 +54,12 @@ describe('public-form.auth.routes', () => {
       // Assert
       expect(response.status).toEqual(StatusCodes.OK)
       expect(response.body).toMatchObject({
-        redirectURL: expect.any(String),
+        redirectURL: expect.stringContaining(
+          encodeURI(getRedirectTarget(form._id, AuthType.SP, false)),
+        ),
+      })
+      expect(response.body).toMatchObject({
+        redirectURL: expect.stringContaining(form.esrvcId!),
```
karrui, line 62: nit:  can be refactored to check everything in 1 block:

```ts
expect(response.body).toMatchObject({
  redirectUrl: expect.toIncludeMultiple([
    encodeURI(getRedirectTarget(form._id, AuthType.SP, false)), 
    form.esrvcId!
  ])
})
```

src/app/routes/api/v3/forms/__tests__/public-forms.auth.routes.spec.ts
```diff
@@ -75,7 +81,12 @@ describe('public-form.auth.routes', () => {
       // Assert
       expect(response.status).toEqual(StatusCodes.OK)
       expect(response.body).toMatchObject({
-        redirectURL: expect.any(String),
+        redirectURL: expect.stringContaining(
```
karrui, line 84: samesies

src/app/routes/api/v3/forms/__tests__/public-forms.auth.routes.spec.ts
```diff
@@ -97,7 +108,10 @@ describe('public-form.auth.routes', () => {
       // Assert
       expect(response.status).toEqual(StatusCodes.OK)
       expect(response.body).toMatchObject({
-        redirectURL: expect.any(String),
+        redirectURL: expect.stringContaining(form.esrvcId!),
```
karrui, line 111: samesies

karrui (approved):
lgtm with nit

src/app/routes/api/v3/forms/__tests__/public-forms.auth.routes.spec.ts
seaerchin: ah good one. wasn't aware of this because i was looking through jest docs and they didn't have this oddly enough

PR #1738, opened 2021-04-27
fix: update MyInfo field count correctly, show correct error

mantariksh:
## Problem

1. When user attempts to save more MyInfo fields than is allowed, a generic error message shows in the Toastr, rather than the specific one returned by the backend.
2. On the build tab, MyInfo field count does not update upon creation of MyInfo field.

### Fixing error message
The initial error message was:
image

This was fixed by accessing `error.response.status` rather than `error.status`, since `error` is now an `AxiosError`.

The error message then changed to this:
Screenshot 2021-04-27 at 10 54 16 AM

This was because we return a 422 when an error occurs while editing fields, and the `switch` case handles 422s by attempting to access `error.data.message`. This was fixed by accessing `error.response.data.message` instead.

This finally fixed the error message:
Screenshot 2021-04-27 at 10 54 55 AM

### Updating MyInfo field count
This was fixed by assigning to `$scope.myform.form_fields` instead of pushing to the array, which triggers the AngularJS digest cycle correctly.

### DIFF
--- a/src/public/modules/forms/admin/controllers/admin-form.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/admin-form.client.controller.js
@@ -141,7 +141,8 @@ function AdminFormController(
       return
     }
     let errorMessage
-    switch (error.status) {
+    const status = get(error, 'response.status') || get(error, 'status')
+    switch (status) {
       case StatusCodes.CONFLICT:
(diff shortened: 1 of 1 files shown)


karrui (approved):
lgtm

PR #1744, opened 2021-04-27
refactor(adminform): update form collab

### LINKED ISSUE #1491: [API] Extract update form collaborators endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **updating form collaborators**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
PATCH /api/v3/admin/forms/:formId/collaborators
```

> No need for PUT/POST/DELETE HTTP methods since it is just manipulating `form.permissionList` array values and collaborators have no identifier to act on

seaerchin:
Integration tests will be in another PR!

## Problem
This PR addresses the sharding of the `/adminform` endpoint so that it is more maintainable in the future. This PR shifts out the adding/deleting/updating of collaborators into its own endpoint and replaces the frontend call site so that services are more modular. 

Closes #1491 

## Solution
This implements a different endpoint for handling CRUD operations related to updating collaborators, along with the associated typing/logical updates on the backend and db model

## Design Choices (and misc questions)
1. Why are some `.save()` methods typed as `T | null`? In the mongoose [docs](https://mongoosejs.com/docs/documents.html#updating-using-queries), `.save()` will only return `T | Error`
2. PUT chosen over PATCH because of ease of implementation. making this a PATCH implies that we have to do stateful updates on each click in the modal - i.e, whenever someone clicks add/delete/update permissions, we will call the new endpoint, which requires additional parameters to indicate the type of request or a http verb for each specific operation
3. `permissionList` which is defined on the mongo schema is typed as being `NonNullable` in our types but no default is set on the schema - can this actually default to `undefined`?

## Manual Test 
- [ ] Add a collaborator with a valid `.gov.sg` email as `editor` to the form. After doing so, attempt to change their role to `read-only`. Thereafter, delete the collaborator. This should all succeed. 
- [ ] Add a collaborator with a valid `.gov.sg` email as `read-only` to the form. After doing so, attempt to change their role to `editor`. Thereafter, delete the collaborator. This should all succeed. 
- [ ] Attempt to add a collaborator with an invalid email to the form. This should fail with an error message.
    - [ ] try using `.gov.sga` 
    - [ ] try using invalid email domains `@imamonster.com` 
    - [ ] try using personal emails 
    - [ ] please feel free to add more
   
**Note**: This PR cannot be reverted due to changing FE calls to the BE


src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -88,20 +91,25 @@ function CollaboratorModalController(
...
+      .then((updatedCollaborators) => {
+        $scope.myform = {
+          ...$scope.myform,
+          ...updatedCollaborators,
+        }
         externalScope.refreshFormDataFromCollab($scope.myform)
       })
       .catch((err) => {
-        Toastr.error(err.data.message)
+        Toastr.error(err.message)
```
mantariksh, line 112: have you tested this? may need to be `err.response.data.message`, or even better, `_.get(err, 'response.data.message, 'Some default message')` since we don't trust types on the frontend

mantariksh (changes requested):

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1729,3 +1730,79 @@ export const handleReorderFormField = [
...
+ * Handler for PUT /api/v3/admin/forms/:formId/collaborators
+ */
+export const handleUpdateCollaborators = [
+  celebrate({
+    [Segments.BODY]: {
+      permissionList: Joi.array().items(
+        Joi.object({
+          email: Joi.string().email().required(),
+          write: Joi.bool().optional(),
+          // There might be other keys sent over with the request but this is not an error.
```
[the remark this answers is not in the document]
seaerchin: yea, this seems like a better solution! will do and test

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -636,6 +637,35 @@ export const updateForm = (
...
+ * @param form the form to update collaborators fo
+ * @param updatedCollaborators the new list of collaborators
+ *
+ * @returns ok(collaborators) if form updates successfully
+ * @returns err(PossibleDatabaseError) if any database errors occurs
+ */
+export const updateFormCollaborators = (
+  form: IPopulatedForm,
+  updatedCollaborators: Permission[],
+): ResultAsync<Permission[], PossibleDatabaseError> => {
```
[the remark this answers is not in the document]
seaerchin: this is a good point - i'll add `_id?` to the type of `Permission` and fix the tests accordingly!

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1729,3 +1730,79 @@ export const handleReorderFormField = [
...
+        }),
+      )
+      // Step 2: Update the form collaborators
+      .andThen((form) =>
+        AdminFormService.updateFormCollaborators(form, req.body.permissionList),
+      )
+      .map((updatedCollaborators) =>
+        res
+          .status(StatusCodes.OK)
+          .json({ permissionList: updatedCollaborators }),
```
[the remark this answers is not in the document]
seaerchin: @karrui not in favour of fixing because idt it's a big issue but @mantariksh raised the concern that the shapes might be different - do you think it's important enough to fix?

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
seaerchin: great catch! will read up on axios and fix + document this!

src/app/modules/form/admin-form/admin-form.controller.ts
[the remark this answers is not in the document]
karrui: so this was the discussion yesterday; for my 2 cents I think anti makes sense that updating form collaborators return the new form collaborator array

should we change it to return an array of new collaborators?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1729,3 +1730,79 @@ export const handleReorderFormField = [
...
+  )
+}
+
+/**
+ * Handler for PUT /api/v3/admin/forms/:formId/collaborators
+ */
+export const handleUpdateCollaborators = [
+  celebrate({
+    [Segments.BODY]: {
+      permissionList: Joi.array().items(
```
karrui, line 1798: the shape of this should be an array of new collaborators to update instead of an object, since it is updating the `permissionList` resource directly. So on that front, I think the return response of this handler should also be the updated array of collaborators.

src/app/modules/form/admin-form/admin-form.controller.ts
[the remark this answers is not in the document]
seaerchin: Thanks for the input! I'll update it accordingly

src/app/modules/form/admin-form/admin-form.service.ts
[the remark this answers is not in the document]
karrui: I think `_id` was omitted in the types because we don't actually use them. I guess there is no harm adding it in as a type anyways and probably adds to the correctness haha

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -88,20 +91,25 @@ function CollaboratorModalController(
...
-      .$promise.then((savedForm) => {
-        $scope.myform = savedForm
+    return $q
+      .when(
+        AdminFormService.updateCollaborators($scope.myform._id, {
+          permissionList,
+        }),
+      )
+      .then((updatedCollaborators) => {
+        $scope.myform = {
```
karrui, line 105: nit: does just plain assigning the new collaborator list not work?

```js
$scope.myform.permissionList = ...updatedCollaborators
```

but since we are changing the return type to an array we can just do a straight assignment after that change

seaerchin: haven't tried; primary reason was cos i heard something about assignment bugging on fe ui and should've cloned so that caused me to clone here HAHAHA

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1730,6 +1731,62 @@ export const handleReorderFormField = [
...
+      )
+      .mapErr((error) => {
+        logger.error({
+          message: 'Error occurred when updating collaborators',
+          meta: {
+            action: '_handleUpdateCollaborators',
+            ...createReqMeta(req),
+            userId: sessionUserId,
+            formId,
+            formCollaborators: req.query,
```
mantariksh, line 1780: are there query params for this endpoint?

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1676,5 +1676,45 @@ describe('Form Model', () => {
...
+          },
+        ]
+
+        // Act
+        const actual = await validForm
+          .updateFormCollaborators(newCollaborators)
+          .catch((error) => error)
+
+        // Assert
+        expect(actual).toBeInstanceOf(mongoose.Error.ValidationError)
```
mantariksh, line 1716: a more canonical way to write this would be to use jest's `.rejects`

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1730,6 +1731,62 @@ export const handleReorderFormField = [
...
+          formId,
+          level: PermissionLevel.Write,
+        }),
+      )
+      // Step 2: Update the form collaborators
+      .andThen((form) =>
+        AdminFormService.updateFormCollaborators(form, req.body),
+      )
+      .map((updatedCollaborators) =>
+        res.status(StatusCodes.OK).send(updatedCollaborators),
```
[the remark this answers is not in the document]
seaerchin: putting this here for visibility: `res.json` encodes it as a stringified json ([docs](http://expressjs.com/en/api.html#res.json)). `res.send` does coerce it to json but it's unclear if it sets the content type to `application/json`

src/app/modules/form/admin-form/admin-form.controller.ts
seaerchin: good catch, will fix!

src/app/models/__tests__/form.server.model.spec.ts
seaerchin: good point, will do!

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1676,5 +1676,45 @@ describe('Form Model', () => {
...
+            email: `fakeuser@${MOCK_ADMIN_DOMAIN}`,
+            write: false,
+          },
+        ]
+
+        // Act
+        const actual = await validForm.updateFormCollaborators(newCollaborators)
+
+        // Assert
+        const actualPermissionsWithoutId = map(
```
mantariksh, line 1694: hahahahahaha @ using `_.map` instead of native map

mantariksh (approved):
lgtm, do test on staging

src/app/models/__tests__/form.server.model.spec.ts
seaerchin: y u laff

PR #1748, opened 2021-04-28
chore(mergify): remove 0 check failure condition

mantariksh:
## Problem

The [Mergify documentation](https://docs.mergify.io/conditions/#validating-all-status-checks) states that `#check-failure=0` should not be used as a condition because of a possible race condition where the PR might be merged before any service reports its status check.

Credit to @lamkeewei for pointing this out.

## Solution

Remove the `#check-failure=0` condition. We can rely on branch protections to ensure that all status checks pass before PRs can be merged (which is what we were implicitly doing in the first place).


karrui (approved):
imagetm

PR #1753, opened 2021-04-28
test(admin-forms): integration tests for updating of collaborators

seaerchin:
## Problem
This is a follow up to #1744 and adds integration tests to the PR.


mantariksh (commented):
lgtm but I noticed the request/response shapes haven't been updated to the array format. will approve when that has been updated.

mantariksh (approved):

PR #1760, opened 2021-04-29
feat(api-refactor): add specific update end page endpoint in server

### LINKED ISSUE #1487: [API] Extract update form end page endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **update form's end page**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
PATCH /api/v3/admin/forms/:formId/end-page
```

karrui:
## Problem

This PR adds a specific endpoint `PUT /api/v3/admin/forms/:formId/end-page` for updating of the form's end page subslice.
The client changes will continue in a separate PR for ease of review.

Note that this new endpoint has stricter Joi validation than before: 
```ts
celebrate({
    [Segments.BODY]: {
      title: Joi.string(),
      paragraph: Joi.string().allow(''),
      buttonLink: Joi.string().uri().allow(''), // model validation did not test for `uri`-ness
      buttonText: Joi.string().allow(''),
    },
})
```

Previously, there was no route level validation, and instead relying on model level validation, which did not test whether a buttonLink was an URI.

Closes #1487

## Solution

**Features**:
- feat(FormModel): add updateEndPageById form static method
- feat(AdminFormSvc): add updateEndPage fn
- feat(AdminFormCtl): add handler for updating form end page

## Tests

- Unit tests have been added to the model, service and controller layers for the various updateEndPage functions.

### DIFF
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -1217,6 +1217,7 @@ describe('Form Model', () => {
         const updatedEndPage: EndPage = {
           title: 'some new title',
           paragraph: 'some description paragraph',
+          buttonText: 'custom button text',
         }
 
         // Act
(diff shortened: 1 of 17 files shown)


seaerchin:
lgtm

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1780,3 +1782,71 @@ export const handleDeleteFormField: RequestHandler<
...
+  )
+}
+
+/**
+ * Handler for PUT /forms/:formId/end-page
+ */
+export const handleUpdateEndPage = [
+  celebrate({
+    [Segments.BODY]: {
+      title: Joi.string().required(),
```
mantariksh, line 1845: as per discussion, this doesn't need to be required because it has a default in the model

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1202,6 +1203,53 @@ describe('Form Model', () => {
         })
       })
     })
+
+    describe('updateEndPageById', () => {
+      it('should update end page and return updated form when successful', async () => {
```
mantariksh, line 1208: if you think it's important, we can also test that the correct defaults are inserted when we don't provide `title` and `buttonText`

karrui: buttonText is already tested in this test case, but i'll add one for title too just in case

karrui:
Ready for re-review

mantariksh (approved):

PR #1762, opened 2021-04-29
refactor(adminform): extract get collaborators

### LINKED ISSUE #1497: [API] Extract get form collaborators view endpoint from GET /:formId/adminform

This issue tracks the creation of a new endpoint just to **retrieving the subset view relevant to show form collaborators**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
GET /api/v3/admin/forms/:formId/collaborators
```

> The return shape of this view subset is up to the implementer, as long as the relevant details can still be displayed when rendering form collaborators

seaerchin:
## Problem
Currently, our `/adminform` endpoint is returning the full adminform object. This endpoint has alot of responsibilities and this pr shards out the responsibility of returning the collaborators to the individual endpoints.

Closes #1497

## Solution
This PR adds a new endpoint for `GET /admin/form/collaborators`, which is solely responsible for returning the slice of the form state that holds the collaborators


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -234,6 +234,54 @@ export const handleGetAdminForm: RequestHandler<{ formId: string }> = (
...
+      .andThen((user) =>
+        // Step 2: Check whether user has read permissions to form
+        AuthService.getFormAfterPermissionChecks({
+          user,
+          formId,
+          level: PermissionLevel.Read,
+        }),
+      )
+      .map(({ permissionList }) =>
+        res.status(StatusCodes.OK).send(permissionList),
```
mantariksh, line 267:
```suggestion
        res.status(StatusCodes.OK).json(permissionList),
```

src/app/routes/api/v3/admin/forms/admin-forms.settings.routes.ts
```diff
@@ -78,6 +78,6 @@ AdminFormsSettingsRouter.route('/:formId([a-fA-F0-9]{24})/settings')
  * @returns 422 when user in session cannot be retrieved from the database
  * @returns 500 when database error occurs
  */
-AdminFormsSettingsRouter.route('/:formId([a-fA-F0-9]{24})/collaborators').put(
-  AdminFormController.handleUpdateCollaborators,
-)
+AdminFormsSettingsRouter.route('/:formId([a-fA-F0-9]{24})/collaborators')
+  .put(AdminFormController.handleUpdateCollaborators)
```
mantariksh, line 82: can we put the documentation above each method, as per the block above?

mantariksh (approved):

PR #1763, opened 2021-04-29
build: allow local FormSG container to depend on Localstack

mantariksh:
TODO: suppress unnecessary output, ensure that adding SQS works


mantariksh:
Closing in favour of adding a bash script which checks that the queue is running before starting the FormSG server

PR #1782, opened 2021-05-04
fix: allow commas in email confirmation sender

mantariksh:
## Problem

We allow admins to specify the email sender for email confirmations. However, when the sender contains a comma, the text before the comma gets cut off.

### Example from production

**Specifying an email sender with a comma**
image

**Actual email sender**
Screenshot 2021-05-04 at 11 21 44 AM

## Solution

According to [this issue](https://github.com/nodemailer/nodemailer/issues/377), sender names containing commas must be wrapped in quotes. This fixed the problem.

### Example from staging

**Specifying an email sender with a comma**
image

**Actual email sender**
Screenshot 2021-05-04 at 11 24 43 AM

### Backwards compatibility

Note that for senders without special characters, the email headers actually remain unchanged, so this change is backwards compatible even for receivers which are manually parsing email headers. For senders with special characters, the email headers get fixed by the inclusion of quotes.

**Without comma, before (production)**
Screenshot 2021-05-04 at 11 28 28 AM

**Without comma, after (staging)**
Screenshot 2021-05-04 at 11 27 19 AM

**With comma, before (production)**
Screenshot 2021-05-04 at 11 28 50 AM

**With comma, after (staging)**
Screenshot 2021-05-04 at 11 29 19 AM

### DIFF
--- a/src/app/services/mail/__tests__/mail.service.spec.ts
+++ b/src/app/services/mail/__tests__/mail.service.spec.ts
@@ -738,7 +738,7 @@ describe('mail.service', () => {
 
       defaultExpectedArg = {
         to: MOCK_AUTOREPLY_PARAMS.autoReplyMailDatas[0].email,
-        from: `${MOCK_AUTOREPLY_PARAMS.form.admin.agency.fullName} <${MOCK_SENDER_EMAIL}>`,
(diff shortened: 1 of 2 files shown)


karrui (approved):
lgtm

tshuli (approved):

PR #1789, opened 2021-05-04
feat: add cors middleware to express server

karrui:
## Problem

This PR adds the `cors` express middleware to the server, so that the React client will be able to call the server APIs successfully.

This is one of the PRs to prepare for the React client.

## Solution

**Features**:
- feat: add corsWhitelist variable to convict schema
- feat: use cors middleware in express

### Manual tests
- [ ] All requests in the network tab should still pass (barring adblock)
- [ ] Create an email form with an invalid admin email. Submit a form and check whether the `/emailnotifications` endpoint has been hit successfully. Check the bounce collection in the database and assert that the document is correctly created.


src/app/loaders/express/cors.ts
```diff
@@ -0,0 +1,19 @@
+import cors from 'cors'
+
+import { app, corsWhiteList } from '../../config/config'
+
+export const corsMiddleware = () => {
+  return cors({
+    origin: function (origin, callback) {
+      // If no origin or from self, disable CORS
+      if (!origin || app.appUrl === origin) {
+        callback(null, false)
```
mantariksh, line 10: sorry I may be misunderstanding this but shouldn't this be `true`? if it's a server to server request or the request originates from our own domain, we should allow it and set the ACAO header to reflect the request origin?

mantariksh, line 10: do you think this would be clearer if we just create one boolean which explains all the cases? e.g.
```
const isOriginAllowed = 
  // allow server-to-server requests
  !origin ||
  // allow requests from own origin
  app.appUrl === origin ||
  // allow requests from whitelisted domains
  corsWhiteList.indexOf(origin) !== -1
callback(null, isOriginAllowed)
```

mantariksh: might need to flip some of the booleans depending on my other comment

src/app/loaders/express/cors.ts
```diff
@@ -0,0 +1,19 @@
...
+      } else {
+        // Enable CORS only if whitelisted
+        callback(null, corsWhiteList.indexOf(origin) !== -1)
```
mantariksh, line 13: is it better to just use `corsWhiteList.includes(origin)`?

src/app/config/config.ts
```diff
@@ -214,6 +214,7 @@ const config: Config = {
   bounceLifeSpan: basicVars.mail.bounceLifeSpan,
   chromiumBin: basicVars.mail.chromiumBin,
   port: basicVars.core.port,
+  corsWhiteList: basicVars.corsWhitelist,
```
mantariksh, line 217: can we standardise this to capital L or small l pls?

src/app/loaders/express/cors.ts
mantariksh: also just double checking, does this allow incoming requests from AWS for email notifications?

karrui: disabling cors will just result in status quo, no?

good point about the AWS email notifications, can we check that easily?

src/app/config/config.ts
karrui: good catch, haha will fix

src/app/loaders/express/cors.ts
mantariksh: >disabling cors will just result in status quo, no?

yes for same-origin requests, but I'm not sure about server-to-server requests (eg email notifications). can check on staging, just submit an email mode form and see if the email notification gets logged

karrui: i assume aws will send without origin since it's not a browser and should not be affected

karrui: i'll test on staging

karrui: will defer and use this, probably fine to set origin even when from our own domain

karrui: works on staging without issues

src/app/config/schema.ts
```diff
@@ -286,6 +286,12 @@ export const optionalVarsSchema: Schema<IOptionalVarsSchema> = {
       env: 'SEND_AUTH_OTP_RATE_LIMIT',
     },
   },
+  corsWhitelist: {
+    doc: 'Comma separated domain strings to enable CORS for',
```
mantariksh, line 290: does `convict` parse comma-separated strings correctly as arrays? can't find it in the docs

mantariksh (approved):

src/app/config/schema.ts
karrui: yea, it's a convict type we created in `src/app/config/schema.ts#19`, coerces the input (which is always string due to being an env var) into an array

```ts
convict.addFormat({
  name: 'string[]',
  validate: (val: string[]) => {
    if (!Array.isArray(val)) {
      throw new Error('must be of type Array')
    }
    if (val.some((i) => typeof i !== 'string')) {
      throw new Error('Elements must be of type string')
    }
  },
  coerce: (val: string): string[] => {
    return val.split(',')
  },
})
```

liangyuanruo (changes requested):
As discussed, it is currently premature to allow cross-origin requests.

PR #1796, opened 2021-05-05
refactor: extract create logic endpoint / refactor $watchCollection

### LINKED ISSUE #1489: [API] Extract create logic endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **creating new form logic**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
POST /api/v3/admin/forms/:formId/logic
```

tshuli:
## Problem

- Closes #1489
- Also refactors edit logic client controller to use $watchCollection to track changes in form_logic

## Tests

**Create Logic Endpoint**
- [ ] Create a showFields logic. Check that logic is successfully added and build tab is successfully updated (hidden fields greyed out).
- [ ] Repeat the above with a second logic

Test using manual API calls
- [ ] Check that endpoint returns 404 when formId is incorrect

**$watchCollection Refactor**
- [ ] Create a form with multiple showFields logic. Delete one of them. Check that the Build tab is updated correctly (field should no longer be greyed out)
- [ ] Now change one of the logic to preventSubmit. Check that the Build tab is updated correctly (field should no longer be greyed out). Change it back to showFields. Check that the field is now greyed out on build tab

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
@@ -1693,10 +1693,7 @@ describe('admin-form.service', () => {
         expect.objectContaining(createdLogic),
       )
 
-      expect(CREATE_SPY).toHaveBeenCalledWith(
-        mockEmailForm._id.toHexString(),
-        createdLogic,
-      )
+      expect(CREATE_SPY).toHaveBeenCalledWith(mockEmailForm._id, createdLogic)
     })
 
     it('should return ok(created logic) on successful form logic create for encrypt mode form', async () => {
@@ -1712,10 +1709,7 @@ describe('admin-form.service', () => {
         expect.objectContaining(createdLogic),
       )
 
-      expect(CREATE_SPY).toHaveBeenCalledWith(
-        mockEncryptFormId.toHexString(),
-        createdLogic,
-      )
+      expect(CREATE_SPY).toHaveBeenCalledWith(mockEncryptFormId, createdLogic)
     })
   })
 
--- a/src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
@@ -37,6 +37,7 @@ import {
   IPopulatedForm,
   IPopulatedUser,
   IUserSchema,
+  LogicType,
   PickDuplicateForm,
   ResponseMode,
   Status,
@@ -64,6 +65,7 @@ import {
   archiveForm,
   createForm,
   createFormField,
+  createFormLogic,
   createPresignedPostUrlForImages,
   createPresignedPostUrlForLogos,
   deleteFormField,
@@ -1606,6 +1608,117 @@ describe('admin-form.service', () => {
       expect(actual._unsafeUnwrapErr()).toBeInstanceOf(ApplicationError)
     })
   })
+
+  describe('createFormLogic', () => {
+    const logicId1 = new ObjectId()
+    const logicId2 = new ObjectId()
+    const logicId3 = new ObjectId()
+    const mockEmailFormId = new ObjectId()
+    const mockEncryptFormId = new ObjectId()
+
+    const mockFormLogicOld = {
+      form_logics: [
+        {
+          _id: logicId1,
+          logicType: LogicType.ShowFields,
+        } as ILogicSchema,
+        {
+          _id: logicId2,
+          logicType: LogicType.ShowFields,
+        } as ILogicSchema,
+      ],
+    }
+
+    const createdLogic = {
+      logicType: LogicType.PreventSubmit,
+    } as ILogicSchema
+
+    const mockFormLogicUpdated = {
+      form_logics: [
+        {
+          _id: logicId1,
+          logicType: LogicType.ShowFields,
+        } as ILogicSchema,
+        {
+          _id: logicId2,
+          logicType: LogicType.ShowFields,
+        } as ILogicSchema,
+        {
+          _id: logicId3,
+          logicType: LogicType.PreventSubmit,
+        } as ILogicSchema,
+      ],
+    }
+
+    const CREATE_SPY = jest.spyOn(FormModel, 'createFormLogic')
+
+    let mockEmailForm: IPopulatedForm,
+      mockEncryptForm: IPopulatedForm,
+      mockEmailFormUpdated: IPopulatedForm,
+      mockEncryptFormUpdated: IPopulatedForm
+
+    beforeEach(() => {
+      mockEmailForm = ({
+        _id: mockEmailFormId,
+        status: Status.Public,
+        responseMode: ResponseMode.Email,
+        ...mockFormLogicOld,
+      } as unknown) as IPopulatedForm
+      mockEncryptForm = ({
+        _id: mockEncryptFormId,
+        status: Status.Public,
+        responseMode: ResponseMode.Encrypt,
+        ...mockFormLogicOld,
+      } as unknown) as IPopulatedForm
+      mockEmailFormUpdated = ({
+        ...mockEmailForm,
+        ...mockFormLogicUpdated,
+      } as unknown) as IPopulatedForm
+      mockEncryptFormUpdated = ({
+        ...mockEncryptForm,
+        ...mockFormLogicUpdated,
+      } as unknown) as IPopulatedForm
+    })
+
+    it('should return ok(created logic) on successful form logic create for email mode form', async () => {
+      // Arrange
+      CREATE_SPY.mockResolvedValue(mockEmailFormUpdated as IFormSchema)
+
+      // Act
+      const actualResult = await createFormLogic(mockEmailForm, createdLogic)
+
+      // Assert
+      expect(actualResult.isOk()).toEqual(true)
+      expect(actualResult._unsafeUnwrap()).toEqual(
+        expect.objectContaining(createdLogic),
+      )
+
+      expect(CREATE_SPY).toHaveBeenCalledWith(
+        mockEmailForm._id.toHexString(),
+        createdLogic,
+      )
+    })
+
+    it('should return ok(created logic) on successful form logic create for encrypt mode form', async () => {
+      // Arrange
+      CREATE_SPY.mockResolvedValue(mockEncryptFormUpdated as IFormSchema)
+
+      // Act
+      const actualResult = await createFormLogic(mockEncryptForm, createdLogic)
+
+      // Assert
+      expect(actualResult.isOk()).toEqual(true)
+      expect(actualResult._unsafeUnwrap()).toEqual(
+        expect.objectContaining(createdLogic),
+      )
+
+      expect(CREATE_SPY).toHaveBeenCalledWith(
+        mockEncryptFormId.toHexString(),
+        createdLogic,
+      )
+    })
+  })
+
   describe('deleteFormField', () => {
     let deleteSpy: jest.SpyInstance
 
@@ -1746,29 +1859,29 @@ describe('admin-form.service', () => {
       form_logics: [
         {
           _id: logicId1,
-          logicType: 'showFields',
+          logicType: LogicType.ShowFields,
         } as ILogicSchema,
         {
           _id: logicId2,
-          logicType: 'showFields',
+          logicType: LogicType.ShowFields,
         } as ILogicSchema,
       ],
     }
 
     const updatedLogic = {
       _id: logicId1,
-      logicType: 'preventSubmit',
+      logicType: LogicType.PreventSubmit,
     } as ILogicSchema
 
     const mockFormLogicUpdated = {
       form_logics: [
         {
           _id: logicId1,
-          logicType: 'preventSubmit',
+          logicType: LogicType.PreventSubmit,
         } as ILogicSchema,
         {
           _id: logicId2,
-          logicType: 'showFields',
+          logicType: LogicType.ShowFields,
         } as ILogicSchema,
       ],
     }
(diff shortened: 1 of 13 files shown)


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1640,6 +1642,97 @@ export const _handleCreateFormField: RequestHandler<
...
+        .valid(...Object.values(LogicType))
+        .required(),
+      conditions: Joi.array()
+        .items(
+          Joi.object({
+            field: Joi.string().required(),
+            state: Joi.string()
+              .valid(...Object.values(LogicConditionState))
+              .required(),
+            value: Joi.any().required(),
```
seaerchin, line 1716: would it be better to constrain this to the type of `ICondition.value` using [`Joi.alternative()`](https://joi.dev/api/?v=17.4.0#alternatives)?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1640,6 +1642,97 @@ export const _handleCreateFormField: RequestHandler<
...
+            ifValueType: Joi.valid(...Object.values(LogicIfValue)),
+          }).unknown(true),
```
seaerchin, line 1718: maybe this is because of my lack of context but `ICondition` only seems to hold these 4 properties (`field/state/value/ifValueType`) so i'm unsure if the unknown is here only because of mongoose's `_id` property but if that's the case, maybe it might be better to add it as `_id: Joi.string().optional()`

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1640,6 +1642,97 @@ export const _handleCreateFormField: RequestHandler<
...
+        is: LogicType.ShowFields,
+        then: Joi.array().items(Joi.string()).required(),
+      }),
+      preventSubmitMessage: Joi.alternatives().conditional('logicType', {
+        is: LogicType.PreventSubmit,
+        then: Joi.string(),
+      }),
+      // Allow other field related key-values to be provided and let the model
+      // layer handle the validation.
+    }).unknown(true),
```
seaerchin, line 1731: same as above wrt `unknown`

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -717,6 +717,42 @@ export const updateFormSettings = (
...
+ * @returns ok(created logic dto) on success
+ * @returns err(database errors) if db error is thrown during logic update
+ */
+export const createFormLogic = (
+  form: IPopulatedForm,
+  createdLogic: LogicDto,
+): ResultAsync<ILogicSchema, DatabaseError | FormNotFoundError> => {
+  // Create new form logic
+  return ResultAsync.fromPromise(
+    FormModel.createFormLogic(form._id.toHexString(), createdLogic),
```
seaerchin, line 733: why do we invoke `toHexString()` here but not when logging?

src/app/modules/form/form.utils.ts
```diff
@@ -121,3 +121,18 @@ export const getLogicById = (
...
+ * @returns the last logic if exists, `null` otherwise
+ */
+export const getLastLogic = (
+  form_logics: IFormSchema['form_logics'],
+): ILogicSchema | null => {
+  if (!form_logics) {
+    return null
+  }
+
+  return form_logics[form_logics.length - 1] ?? null
```
seaerchin, line 137: i think this bugs when form_logics is an empty array. this is because an empty array is a truthy value (`![] === true`) so it wouldn't be caught by the first check.

when it goes to the second check, `[].length === 0`. this means that we are accessing `form_logics[-1]` which evaluates to `undefined` 

i think this could just be avoided using `_.isEmpty` for the initial check and omitting the ternary thereafter (as it's guaranteed to contain >=1 element)

tshuli: Thanks!

To clarify, if I understood your comment correctly
The function returns the last form logic only if it exists in form_logics and null otherwise

The first check is to check that form_logics is not undefined (could be empty array). It returns null if form_logics is undefined.
The second check is to return null if the last form logic is undefined. In this case, if form_logics is empty, form_logics[-1] evaluates to undefined and subsequently to null via the nullish coalescing operator `??`, which is what we want

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: Thanks! I've updated this along with the other PR on joi validation

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: I left `unknown` so that we don't block requests that have more keys than we expected, which are harmless (the validation is there to ensure that the keys which we expect are present in the expected shape)

src/app/modules/form/form.utils.ts
seaerchin: ah ok you're right on the second point, where it defaults to `null` due to the `??` operator. in that case, actually this whole chunk might be better off as `_.isEmpty(form_logics) ? form_logics[form_logics.length - 1] : null`, which would be clearer!

src/app/modules/form/admin-form/admin-form.controller.ts
seaerchin: ah ok, thanks for pointing this out! just as an aside, we're not expecting any more keys beyond `_id?` though, right?

src/app/modules/form/form.utils.ts
tshuli: Ah but we need the first check if form_logics is defined. This is because we are assuming that the form object returned from the db contains form_logics, but we need to check if this is true i.e. form_logics is defined, and return a DatabaseError otherwise

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: nope, not expecting any. for create form logic there should also not be _id in the request

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: thanks! I guess we could just not cast to string for both to be consistent, edited accordingly

src/app/modules/form/form.utils.ts
seaerchin: i think for `_.isEmpty()`, it returns `true` as long as collection is empty or it's not a collection type (which is also true for `null | undefined`), which would preserve existing behaviour isn't it?

tshuli: Thanks! You are right and I was about to edit, but then I realised that isEmpty() doesn't type guard for [undefined](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/45521), i.e. typescript will complain that `form_logics[form_logics.length - 1]` is invalid as `form_logics` may be undefined

So I think we will have to stick to the current implementation which explicitly checks that form_logic is defined

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -1606,6 +1608,111 @@ describe('admin-form.service', () => {
...
+        ...mockEmailForm,
+        ...mockFormLogicUpdated,
+      } as unknown) as IPopulatedForm
+      mockEncryptFormUpdated = ({
+        ...mockEncryptForm,
+        ...mockFormLogicUpdated,
+      } as unknown) as IPopulatedForm
+    })
+
+    it('should return ok(created logic) on successful form logic create for email mode form', async () => {
```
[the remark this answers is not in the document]
tshuli: added

tshuli:

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import { PresignedPost } from 'aws-sdk/clients/s3'
 import { ObjectId } from 'bson-ext'
-import { assignIn, cloneDeep, merge, omit, pick } from 'lodash'
+import _, { assignIn, cloneDeep, merge, omit, pick } from 'lodash'
```
karrui, line 4: nit: is it possible to stick with either importing the whole library or the separate functions instead of both? 

the new imports are used for `_.omit` and `_.last`, of which `omit` is already imported. should we just import `last` also and remove the `_` import for consistency?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1642,6 +1642,113 @@ export const _handleCreateFormField: RequestHandler<
...
+}
+
+/**
+ * Handler for POST /forms/:formId/logic
+ */
+export const handleCreateLogic = [
+  celebrate({
+    [Segments.BODY]: Joi.object({
+      ...joiLogicBody,
+      // Allow other field related key-values to be provided and let the model
```
karrui, line 1745: Since this is a new endpoint, we can (and should) strictly control the validation. Suggest removing `.unknown(true)`, then it can just be

```ts
[Segments.BODY]: joiLogicBody
```

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1850,38 +1957,7 @@ export const handleUpdateLogic = [
...
-          then: Joi.array().items(Joi.string()).required(),
-        }),
-        preventSubmitMessage: Joi.alternatives().conditional('logicType', {
-          is: LogicType.PreventSubmit,
-          then: Joi.string().required(),
-        }),
+        ...joiLogicBody,
         // Allow other field related key-values to be provided and let the model
         // layer handle the validation.
       }).unknown(true),
```
karrui, line 1963: did not spot this the last time, should remove `.unknown(true)` here too

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1,5 +1,5 @@
 import { PresignedPost } from 'aws-sdk/clients/s3'
-import { assignIn, last, omit } from 'lodash'
+import _, { assignIn, last, omit } from 'lodash'
```
karrui, line 2: nit: import destructured function instead of `_`

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
```diff
@@ -267,18 +265,25 @@ function EditLogicModalController(
...
 
+  vm.createNewLogic = function (newLogic) {
+    $q.when(AdminFormService.createFormLogic(vm.myform._id, newLogic))
+      .then((createdLogic) => {
+        vm.formLogics.push(createdLogic)
+        externalScope.myform.form_logics.push(createdLogic) // update global myform
+        $uibModalInstance.close()
+      })
+      .catch((logicCreateError) => {
+        console.error(logicCreateError)
```
karrui, line 282: remove console log?

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
tshuli: thanks! edited

src/app/modules/form/admin-form/admin-form.controller.ts

src/app/modules/form/admin-form/admin-form.controller.ts

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: edited

src/public/modules/forms/admin/controllers/edit-logic-modal.client.controller.js
tshuli: edited

src/public/modules/forms/admin/directives/edit-form.client.directive.js
```diff
@@ -98,13 +98,10 @@ function editFormController(
...
   let conditionFieldSet = new Set() // To monitor which fields are conditions in the form logic
   // On change of form logic (scope.myform.form_logics), update our logic monitors
-  $scope.$watch(
-    (scope) => scope.myform.form_logics,
-    function (_newVal, _oldVal) {
-      updateHiddenFieldSet()
-      updateConditionFieldSet()
-    },
-  )
+  $scope.$watchCollection('myform.form_logics', function (_newVal, _oldVal) {
```
karrui, line 101: nit: since both values are unused, i think we can just call

```ts
$scope.$watchCollection('myform.form_logics', function () {
  ...
}
```

However, this is also indicative of a code smell, but probably no need to fix it since React is coming

src/public/services/AdminFormService.ts
```diff
@@ -122,6 +122,15 @@ export const updateFormEndPage = async (
     .then(({ data }) => data)
 }
 
+export const createFormLogic = async (
+  formId: string,
+  createdLogic: LogicDto,
```
karrui, line 127: nit: rename `createdLogic` to `createLogicBody` for clarity? (Since the logic has not actually been inserted, and is actually just a body parameter to the endpoint)

src/types/form.ts
```diff
@@ -301,6 +301,10 @@ export interface IFormModel extends Model<IFormSchema> {
     formId: string,
     fields?: (keyof IPopulatedForm)[],
   ): Promise<IPopulatedForm | null>
+  createFormLogic(
+    formId: string,
+    createdLogic: LogicDto,
```
karrui, line 306: if fixing the nit, also remember to rename it here

karrui (approved):
lgtm with nits

karrui:
Maybe rebase on dev for the new Github Actions :P

src/public/services/AdminFormService.ts
tshuli: thanks! edited

src/types/form.ts
tshuli: edited

src/public/modules/forms/admin/directives/edit-form.client.directive.js
tshuli: Will leave this for now~

PR #1798, opened 2021-05-05
chore: gitignore intellij metadata directory

yong-jie:
IntelliJ IDEs generate a .idea directory for metadata.

Would like to add the entry to gitignore so I won't accidentally commit these metadata files.


mantariksh (approved):
good .idea, very intelliJnt

PR #1799, opened 2021-05-05
feat(admin-form): individual form field api 

### LINKED ISSUE #1498: [API] Extract individual form field view endpoint from GET /:formId/adminform

This issue tracks the creation of a new endpoint just to **retrieving a single field details**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
GET /admin/forms/:formId/fields/:fieldId
```

> The return shape of this view subset is up to the implementer, as long as the relevant details can still be displayed when rendering field details. This endpoint will be useful for retrieving the most updated field details when the admin wants to modify the given field.

seaerchin:
## Problem
Individual form fields require their own endpoint to disentangle their logic from other unrelated logic in forms, such as the settings/collaborators.

Closes #1498

## Solution
This adds a new GET endpoint for individual fields

## Tests
Added unit and integration tests for `/fields/:fieldId` but there are no manual tests. This is because the frontend is still using the whole admin form object to update the UI


mantariksh (approved):

seaerchin:
just leaving this comment here to remind myself that this is my first PR without any comments HAHAHHA

PR #1800, opened 2021-05-05
fix: align joi validation types to schema

### LINKED ISSUE #1810: LogicDto is typed incorrectly
In `src/types/form_logic.ts`, `LogicDto` is typed as

```
LogicDto = ConditionalPick<LeanDocument<ILogicSchema>, Primitive>
```

This was initially typed to be in line with `FormFieldDto`, but this is wrong because Logic contains nested objects which are not primitives, resulting in `LogicDto` only containing `_id` and `logicType`.

tshuli:
## Problem

- This PR builds off #1695 and replaces the typings for joi validation for logic endpoint to align to the schema
- Also closes #1810

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -1811,7 +1811,14 @@ export const handleUpdateLogic = [
               state: Joi.string()
                 .valid(...Object.values(LogicConditionState))
                 .required(),
-              value: Joi.any().required(),
+              value: Joi.alternatives()
+                .try(
+                  Joi.number(),
+                  Joi.string(),
+                  Joi.array().items(Joi.string()),
+                  Joi.array().items(Joi.number()),
+                )
+                .required(),
               ifValueType: Joi.valid(...Object.values(LogicIfValue)),
             }).unknown(true),
           )
@@ -1822,7 +1829,7 @@ export const handleUpdateLogic = [
         }),
         preventSubmitMessage: Joi.alternatives().conditional('logicType', {
           is: LogicType.PreventSubmit,
-          then: Joi.string(),
+          then: Joi.string().required(),
         }),
         // Allow other field related key-values to be provided and let the model
         // layer handle the validation.
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -1819,7 +1819,9 @@ export const handleUpdateLogic = [
                   Joi.array().items(Joi.number()),
                 )
                 .required(),
-              ifValueType: Joi.valid(...Object.values(LogicIfValue)),
+              ifValueType: Joi.string()
+                .valid(...Object.values(LogicIfValue))
+                .required(),
             }).unknown(true),
           )
           .required(),
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -16,7 +16,9 @@ import {
   IForm,
   IFormDocument,
   IPopulatedForm,
+  LogicConditionState,
   LogicDto,
+  LogicIfValue,
   LogicType,
   ResponseMode,
 } from '../../../../types'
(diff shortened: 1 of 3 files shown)


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1806,9 +1808,11 @@ export const handleUpdateLogic = [
           .items(
             Joi.object({
               field: Joi.string().required(),
-              state: Joi.string().required(),
-              value: Joi.string().required(),
-              ifValueType: Joi.string(),
+              state: Joi.string()
+                .valid(...Object.values(LogicConditionState))
+                .required(),
+              value: Joi.any().required(),
```
mantariksh, line 1814: can I check why this is `any`? would it be possible to experiment with the various logic field types to determine exactly what type is being returned here? imo we should always know the exact shape of our APIs

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1818,7 +1822,7 @@ export const handleUpdateLogic = [
         }),
         preventSubmitMessage: Joi.alternatives().conditional('logicType', {
           is: LogicType.PreventSubmit,
-          then: Joi.string().required(),
+          then: Joi.string(),
```
mantariksh, line 1825: can I check why this is not `required`?

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: Thanks! I typed this based on the schema in `form_logic.server.schema.ts`. From what I can see though, it's either a number or string, so I've tightened accordingly. I also left open the possibility for it to be a number array or string array because we allow for those types in `src/types/form_logic.ts`

```
export interface ICondition {
  field: IFieldSchema['_id']
  state: LogicConditionState
  value: string | number | string[] | number[]
  ifValueType?: LogicIfValue
}
```

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: As above, it was to align with the schema, but I've tightened since we do require this on the frontend

tshuli:
@mantariksh thanks for reviewing, for your re-review pls

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1806,9 +1808,18 @@ export const handleUpdateLogic = [
...
+                .required(),
+              value: Joi.alternatives()
+                .try(
+                  Joi.number(),
+                  Joi.string(),
+                  Joi.array().items(Joi.string()),
+                  Joi.array().items(Joi.number()),
+                )
+                .required(),
+              ifValueType: Joi.valid(...Object.values(LogicIfValue)),
```
mantariksh, line 1822: can I check why this is not required, and should it be? also is the behaviour different if you exclude `Joi.string()`, as in the `state` key?

src/types/form_logic.ts
```diff
@@ -116,4 +111,4 @@ export type LogicCondition =
 /**
  * Logic POJO with functions removed
  */
-export type LogicDto = ConditionalPick<LeanDocument<ILogicSchema>, Primitive>
+export type LogicDto = ILogic & { _id?: IFieldSchema['_id'][] }
```
mantariksh, line 114: not that it makes a big difference, but why `IFieldSchema['_id']` instead of `ILogicSchema['_id']`?

tshuli: thanks! Actually I think the most correct one should be `Document['_id']`, since that is where all the Schemas inherit the `_id` property from

src/app/modules/form/admin-form/admin-form.controller.ts
tshuli: thanks, made it required. Excluding `string()` shouldn't make a difference because all the values in LogicIfValue are strings, but I think adding `string()` is more [canonical](https://joi.dev/api/?v=17.4.0#anyvalidvalues---aliases-equal). Edited accordingly

tshuli:
@mantariksh thanks for reviewing, for your re-review pls

PR #1813, opened 2021-05-06
feat(email-submission): separate error logging for db and state checks

yong-jie:
## Problem
Currently, the first two stages of the submission pipeline have their errors logged jointly.

This could be confusing if we have to debug error messages. It is also inconsistent from the rest of the pipeline.

## Solution
Introduce a `mapErr` after the fetching of the form which logs the database errors, and add another `mapErr` during the state checking step with its own error message.

## Reviewers

Would appreciate KR's help in checking if i'm using `neverthrow` properly in this case.


karrui (approved):
lgtm

PR #1815, opened 2021-05-06
feat(form-v2): rip out AngularJS client and related files

### LINKED ISSUE #1812: [React prep] Remove AngularJS application
React requires its own node_modules, and having webpack packages in the parent node_module was causing build warnings due to a conflict. The AngularJS app should be removed in its entirety in the React branch.

karrui:
Note that this is being merged into the new base branch for form-v2, `form-v2/develop`

## Problem

Inserting a React application into the current repository was creating conflicts due to differing `babel-loader` versions, a package used to run webpack to compile the AngularJS application. 

This PR removes the AngularJS files in preparation for insertion of the new React application, and removes the unnecessary `node_modules` packages so React will not have any compilation conflicts.

Closes #1812

## Solution

**Features**:

- build: remove angularjs packages
- chore: update travis to only run backend tests
- feat: rip out frontend tests and other mentions of /public 

## Dependencies
- Removed many dependencies


karrui:
Update: React app still returns a warning... might as well just ignore warning and slot React base app in

PR #1819, opened 2021-05-06
feat(react): add base react app directory in root of the application

karrui:
Note that SKIP_PREFLIGHT_CHECKS are enabled for all react scripts due to conflicting packages in the root node_modules :(

## Problem

This PR creates the base directory with the initial develop tooling for the React app migration.

**Dev tooling includes:**
- Storybook with Chromatic CI/CD pipeline, for
  - Visual regression testing (via Chromatic)
  - A11y testing
- Linter and formatter
- Aliases (say no more to `../../../../paths/to/hell`)

There is also a Button component story example to see the power of Storybook

## Solution

**Features**:

- lint: add linting and formatter tooling that does not conflict with the current application
- feat(Spinner): add Spinner custom component
- feat(Button): add Button component and story

**Design System**
- feat(design-system): add base colours to theme
- feat(design-system): add base typography to theme
- feat(design-system): add Button styling variants
- feat(design-system): add initial Input outline styling

## Deploy Notes

- chore(github-actions): add Chromatic CI/CD


mantariksh (commented):
???????????????????????????????????????????????????????????????????????????????????

frontend/public/index.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
```
mantariksh, line 1: can delete the logos also

frontend/public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
```
mantariksh, line 1: delete?

mantariksh: or add some docs saying that this should be replaced with our own stuff

frontend/src/components/Spinner/index.ts
```diff
@@ -0,0 +1,3 @@
+import { Spinner } from './Spinner'
```
karrui, line 1:
```ts
export { Spinner as default } from './Spinner'
```

mantariksh (approved):

frontend/config-overrides.js
```diff
@@ -0,0 +1,16 @@
...
+const {
+  aliasDangerous,
+  configPaths,
+  aliasJest,
+} = require('react-app-rewire-alias/lib/aliasDangerous')
+
+const aliasMap = configPaths('tsconfig.paths.json')
+
+module.exports = function override(config) {
+  aliasDangerous(aliasMap)(config)
```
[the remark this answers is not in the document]
karrui: This is for aliasing outside of root folder; i.e having a `/shared` folder that may be outside of the react app's `/src` folder. Since that is not the case yet, will go back to using `alias`. Will add a comment

frontend/public/manifest.json
karrui: hard to comment in a JSON file (since those don't allow comments). Will create an issue instead for this

karrui: added issue #1901

PR #1820, opened 2021-05-06
fix: speed up precommit hooks

karrui:
## Problem

We have been installing a fresh version of lint-staged every precommit due to the script in `husky/pre-commit`. This caused every commit to take ~1 second at least and we pretty bad dev-exp.

This PR instead runs an npm script to use the installed lint-staged module, thus greatly(?) speeding up the pre-commit hook.


liangyuanruo (approved):

liangyuanruo:
Good one! Explains why they took so long

PR #1833, opened 2021-05-10
refactor(admin-form): extract update form startpage

### LINKED ISSUE #1486: [API] Extract update form start page endpoint from PUT /:formId/adminform

This issue tracks the creation of a new endpoint just to **update form's start page**. The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

## New proposed endpoint
```
PATCH /api/v3/admin/forms/:formId/start-page
```

seaerchin:
## Problem
This pr extracts the start page from the`/adminform` endpoint. This keeps the responsibility of the endpoint small and allows us to separate different concerns out and away from teh `/adminform`endpoint, derisking react migration and making the application easier to maintain and reason about. 

Closes #1486 

## Solution
Extract out the start page from `/adminform`

## Tests
Adds unit/integration tests for the new endpoint 

## Manual Tests
(already tested on staging) 

- [ ] Create a new form. Go into the admin panel and click the start page. Edit without saving. This should be prohibited.
- [ ] Create a new form. Go into the admin panel and click the start page. Edit the instructions field and add text there. Click save - the changes should be persisted. Open the start page again and clear the instructions field. Click save - the changes should be persisted. 
- [ ] Create a new form and update the time taken **programmatically** to be 12345. This should return an error from the backend. 
- [ ] Create a new form and update the logo **programmatically** to have `state: NOT_A_STATE`. This should return an error from teh backend
- [ ] Create a new form and update the colour **programmatically** to have `colour: transparent`. This should return an error from teh backend
- [ ] Create a new form and update the logo to be an image and click save. this should work. thereafter, repeat this **programmatically** but change the `fileSizeInBtes` to be >2MB. This should be rejected

## Notes
This PR can't be rolled back due to the client calling new endpoint


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2074,3 +2075,73 @@ export const handleUpdateCollaborators = [
...
+/**
+ * Handler for PUT /forms/:formId/start-page
+ */
+export const handleUpdateStartPage = [
+  celebrate({
+    [Segments.BODY]: {
+      title: Joi.string(),
+      paragraph: Joi.string().allow(''),
+      buttonLink: Joi.string().uri().allow(''),
+      buttonText: Joi.string().allow(''),
```
tshuli: for validation, should these be required even though we allow empty string?
also to check, should this be the shape of end page instead of start page?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2140,3 +2143,77 @@ export const handleUpdateCollaborators = [
...
+  celebrate({
+    [Segments.BODY]: {
+      paragraph: Joi.string().allow('').optional(),
+      estTimeTaken: Joi.number().min(1).max(1000).required(),
+      colorTheme: Joi.string()
+        .valid(...Object.values(Colors))
+        .required(),
+      logo: Joi.object({
+        state: Joi.string().valid(...Object.values(FormLogoState)),
+      }).required(),
```
mantariksh, line 2215: does this account for custom logos?

mantariksh (changes requested):
I think this code doesn't account for adding custom form logos, and it also seems that `IFormSchema` and `IFormDocument` are typed incorrectly because they don't account for custom form logos either. would you be okay with fixing the latter as well? either in this PR or separately

src/app/modules/form/admin-form/admin-form.controller.ts
seaerchin: will address this (and the associated types in this PR) - thanks for the catch! on that note, i'll change this to be a conditional type s.t. the extra fields are validated only when there's a `logo` provided

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2140,3 +2146,93 @@ export const handleUpdateCollaborators = [
...
+        .required(),
+      logo: Joi.object({
+        state: Joi.string().valid(...Object.values(FormLogoState)),
+        fileId: Joi.when('state', {
+          is: FormLogoState.Custom,
+          then: Joi.string().required(),
+        }),
+        fileName: Joi.when('state', {
+          is: FormLogoState.Custom,
+          then: Joi.string()
```
karrui, line 2224: same here

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2140,3 +2146,93 @@ export const handleUpdateCollaborators = [
...
+            // Captures only the extensions below regardless of their case
+            // Refer to https://regex101.com/ with the below regex for a full explanation
+            .pattern(/\.(gif|png|jpeg|jpg)$/im)
+            .required(),
+        }),
+        fileSizeInBytes: Joi.when('state', {
+          is: FormLogoState.Custom,
+          then: Joi.number().max(MAX_UPLOAD_FILE_SIZE).required(),
```
karrui, line 2232: samesies

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2140,3 +2146,93 @@ export const handleUpdateCollaborators = [
...
+      paragraph: Joi.string().allow('').optional(),
+      estTimeTaken: Joi.number().min(1).max(1000).required(),
+      colorTheme: Joi.string()
+        .valid(...Object.values(Colors))
+        .required(),
+      logo: Joi.object({
+        state: Joi.string().valid(...Object.values(FormLogoState)),
+        fileId: Joi.when('state', {
+          is: FormLogoState.Custom,
+          then: Joi.string().required(),
```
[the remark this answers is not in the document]
seaerchin: oh dang didn't know about this; i thought it would default to the normal joi behaviour where it rejects extra keys @__@ TIL and i'll update this (and below)

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1445,6 +1446,87 @@ describe('Form Model', () => {
...
+        }
+
+        // Act
+        const actual = await Form.updateStartPageById(
+          form._id,
+          updatedStartPage,
+        )
+
+        // Assert
+        // Should have defaults populated but also replace the endpage with the new params
```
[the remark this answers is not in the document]
seaerchin: whoops, will fix! good catch!

src/app/modules/form/admin-form/admin-form.controller.ts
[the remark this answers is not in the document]
karrui: Idk actually, you can experiment!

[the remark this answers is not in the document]
seaerchin: you're right - need the condition to reject :(

mantariksh (approved):

karrui (changes requested):
to fix joi validator before merging

seaerchin:
@karrui  @mantariksh  i think there's an existing bug on prod, where after changing the icon type from 'CUSTOM' to any of the other two options, there is additional props (from the `CUSTOM` option). this prohibits the change from being made here but i'll raise an issue and attempt a fix for that, together with the additional validation

karrui:
> @karrui @mantariksh i think there's an existing bug on prod, where after changing the icon type from `CUSTOM` to any of the other two options, there is additional props (from the `CUSTOM` option). this prohibits the change from being made here but i'll raise an issue and attempt a fix for that, together with the additional validation

Not really a bug in the strictest sense, since the `Form` schema will throw away all extra keys if the logo type is not custom due to the use of [mongoose discriminators](https://mongoosejs.com/docs/discriminators.html)

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
```diff
@@ -1429,4 +1437,145 @@ describe('admin-form.form.routes', () => {
...
+      const resp = await request
+        .put(`/admin/forms/${form._id}/start-page`)
+        .send(MOCK_UPDATED_START_PAGE)
+
+      // Assert
+      expect(resp.status).toBe(403)
+      expect(resp.body).toEqual(jsonParseStringify(expectedResponse))
+    })
+
+    it('should  return 404 when the form canont be found', async () => {
```
mantariksh, line 1503:
```suggestion
    it('should  return 404 when the form cannot be found', async () => {
```

mantariksh (commented):
lgtm barring the comments re: Joi validation

karrui (approved):
lgtm

PR #1848, opened 2021-05-10
ref: remove usage of express-serve-static-core types (ParamsDictionary, Query)

### LINKED ISSUE #948: Use `never` instead of `ParamsDictionary` in middleware

karrui:
## Problem

This PR creates a new `ControllerHandler` interface that extends from `express#RequestHandler` interface but replaces all the `any` casts to `unknown`, for stricter typing. With that, the dependency `@types/express-serve-static-core` is no longer needed and removed from the application.

This PR is however blocked by #1854 as `submitEncryptModeForm` has no typings and will now throw a compiler error due to the stricter `unknown` type (from `any`). Will be merged in when that issue is resolved.

Closes #948

## Solution

**Features**:
- add ControllerHandler type to cast unknowns to RequestHandler
- replace RequestHandler in controllers with ControllerHandler

**Improvements**:
- loosen Request type used in `utils/request.ts`
- remove dependency on `@types/express-serve-static-core` by updating types used

**Dependencies**:
- remove `@types/express-serve-static-core`


tshuli (approved):
lgtm

src/app/modules/examples/examples.controller.ts
```diff
@@ -25,10 +24,10 @@ const logger = createLoggerWithLabel(module)
  * @returns 500 when error occurs whilst querying the database
  */
 export const handleGetExamples: RequestHandler<
-  ParamsDictionary,
+  unknown,
   ErrorDto | ExampleFormsResult,
   unknown,
-  Query & ExampleFormsQueryDto
+  Request['query'] & ExampleFormsQueryDto
```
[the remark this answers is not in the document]
mantariksh: same for the rest

src/app/modules/frontend/frontend.controller.ts
```diff
@@ -13,10 +12,10 @@ const logger = createLoggerWithLabel(module)
...
  * @param req - Express request object
  * @param res - Express response object
- * @returns {String} Templated Javascript code for the frontend to initialise Google Tag Manager
+ * @returns Templated Javascript code for the frontend to initialise Google Tag Manager
  */
 export const addGoogleAnalyticsData: RequestHandler<
-  ParamsDictionary,
+  unknown,
   string | { message: string }
 > = (req, res) => {
```
mantariksh, line 20: should we add a `unknown` for the body here? otherwise the body is `any` by default

mantariksh: same for the other RequestHandlers which don't have a 3rd argument

src/app/modules/examples/examples.controller.ts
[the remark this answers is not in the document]
karrui: oh the types seems to be updated to be looser now and we can actually ignore it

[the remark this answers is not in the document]
karrui: but this means the types in `utils/request.ts` need to be loosened as well

src/app/modules/frontend/frontend.controller.ts
karrui: Will create a new type `ControllerHandler` and use that type to replace all `RequestHandler`s, since interface declaration merging cannot override the `any` typings :(

karrui: Resolved by 3928a2a

mantariksh (approved):

PR #1849, opened 2021-05-10
feat: store only user ID in session

### LINKED ISSUE #212: Should only store `userId` in sessions instead of the entire user object
As per title, should we have a single source of truth for the state of the user?

It currently does not matter much since we have no real features to update the user object once it is created (no email changes, no password changes, etc), and thus we can use `req.session.user` without any thought.

However, with the ability to update the user document now (emergency contact number feature #13), the source of truth may diverge.

Thus, suggest just storing the userId in session and retrieving them whenever needed.

jia1:
Closes #212

### DIFF
--- a/src/types/vendor/express.d.ts
+++ b/src/types/vendor/express.d.ts
@@ -7,11 +7,11 @@ declare global {
     }
 
     export interface Session {
-      user?: IUserSchema
+      user?: IUserSchema['_id']
     }
 
     export interface AuthedSession extends Session {
-      user: IUserSchema
+      user: IUserSchema['_id']
     }
   }
 }
--- a/src/types/vendor/express.d.ts
+++ b/src/types/vendor/express.d.ts
@@ -7,11 +7,15 @@ declare global {
     }
 
     export interface Session {
-      user?: IUserSchema['_id']
+      user?: {
+        _id: IUserSchema['_id']
+      }
     }
 
     export interface AuthedSession extends Session {
-      user: IUserSchema['_id']
+      user: {
+        _id: IUserSchema['_id']
+      }
     }
   }
 }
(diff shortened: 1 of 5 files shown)


karrui (approved):
lgtm

karrui (changes requested):
actually, in `src/app/loaders/express/logging.ts`, we have a getter for session.user._id for logging and that may break if we don't update that too

src/types/vendor/express.d.ts
```diff
@@ -7,11 +7,11 @@ declare global {
     }
 
     export interface Session {
-      user?: IUserSchema
+      user?: IUserSchema['_id']
```
karrui, line 10: this should look like 

```ts
user?: {
  _id: IUserSchema['_id']
}
```

src/types/vendor/express.d.ts
```diff
@@ -7,11 +7,11 @@ declare global {
...
     }
 
     export interface AuthedSession extends Session {
-      user: IUserSchema
+      user: IUserSchema['_id']
```
karrui, line 14: Same here

karrui (changes requested):

karrui (commented):
should be ok now, just need to fix the build error (remove the user.email call (since session doesn't contain the full user anymore) and replace it with _id i guess))

also search the codebase for instances of `TODO(#212)` and remove them! :D

jia1:
@karrui Do you mean the line containing: `message: Successfully logged in user ${user._id},`? Reference to `.email` was removed but pipeline still fails. [Travis build](https://travis-ci.com/github/opengovsg/FormSG/builds/228306260) gives me an empty screen.

karrui:
> @karrui Do you mean the line containing: `message: Successfully logged in user ${user._id},`? Reference to `.email` was removed but pipeline still fails. [Travis build](https://travis-ci.com/github/opengovsg/FormSG/builds/228306260) gives me an empty screen.

I've fixed it, can check out my commit for more info!

karrui (approved):
lgtm

PR #1855, opened 2021-05-11
feat: add response ID to thank you page

### LINKED ISSUE #1409: Add response ID to thank you page and email confirmation
problem
- respondents dont have their submission Id unless admin attaches PDF to email confirmation
- this gets in the way of communicating with agency about their response in future 

todo: 
- add responseid to thank you page and email confirmation as in design below

design: https://www.figma.com/file/ta5tjjBqphML8LPRkmxD53/DesignMaster_v2?node-id=1166%3A246

------
PREVIOUS:

todo [design]
- competitive research on what other form managers refer to this as. this is backwards incompatible as it has to match what the admins refer to it as, and RPA on both admin and respondent emails might break if we change this around frivolously.
- ID vs reference number?
- response vs submission?

suggested design
- include "Response reference number: ###########" before the rest of email body in unformatted text
- show this but in a different colour in the preview box on admin panel. can use an obviously fake number like 123456789.

chowyiyin:
## Problem
This PR adds the response ID to the Thank You page. For details, see #1409.

Closes #1409

## Screenshots
Screenshot 2021-05-12 at 7 12 13 PM
Screenshot 2021-05-17 at 10 00 37 AM
Screenshot 2021-05-19 at 12 21 06 PM

## Manual Tests
- [x] Create email field and turn on email confirmation. Check that dummy response ID is present in email confirmation preview in edit field modal
- **Create email field in email mode and submit form**
- [x] Check that correct response ID (submission ID, not form ID) is displayed above time of submission
- [x] Check that response ID is present in email confirmation
- [x] Enter preview mode and repeat above checks
- **Create email field in storage mode and submit form**
- [x] Check that correct response ID (submission ID, not form ID) is displayed above time of submission
- [x] Check that response ID is present in email confirmation
- [x] Enter preview mode and repeat above checks

### DIFF
--- a/src/app/services/mail/mail.service.ts
+++ b/src/app/services/mail/mail.service.ts
@@ -241,10 +241,12 @@ export class MailService {
       autoReplyMailData.sender || form.admin.agency.fullName
     ).replace('(', '\\(')
 
+    const responseId = submission.id
     const defaultBody = `Dear Sir or Madam,\n\nThank you for submitting this form.\n\nRegards,\n${form.admin.agency.fullName}`
     const autoReplyBody = (autoReplyMailData.body || defaultBody).split('\n')
 
     const templateData = {
+      responseId,
       autoReplyBody,
       // Only destructure formSummaryRenderData if form summary is included.
       ...(autoReplyMailData.includeFormSummary && formSummaryRenderData),
--- a/src/app/services/mail/mail.service.ts
+++ b/src/app/services/mail/mail.service.ts
@@ -241,12 +241,11 @@ export class MailService {
       autoReplyMailData.sender || form.admin.agency.fullName
     ).replace('(', '\\(')
 
-    const responseId = submission.id
     const defaultBody = `Dear Sir or Madam,\n\nThank you for submitting this form.\n\nRegards,\n${form.admin.agency.fullName}`
     const autoReplyBody = (autoReplyMailData.body || defaultBody).split('\n')
 
     const templateData = {
-      responseId,
+      submissionId: submission.id,
       autoReplyBody,
       // Only destructure formSummaryRenderData if form summary is included.
       ...(autoReplyMailData.includeFormSummary && formSummaryRenderData),
--- a/src/app/services/mail/mail.types.ts
+++ b/src/app/services/mail/mail.types.ts
@@ -72,8 +72,11 @@ export type SubmissionToAdminHtmlData = {
 }
 
 export type AutoreplyHtmlData =
-  | ({ autoReplyBody: string[] } & AutoreplySummaryRenderData)
-  | { autoReplyBody: string[] }
+  | ({
+      responseId: string
+      autoReplyBody: string[]
+    } & AutoreplySummaryRenderData)
+  | { responseId: string; autoReplyBody: string[] }
 
 export type BounceNotificationHtmlData = {
   formTitle: string
--- a/src/app/services/mail/mail.types.ts
+++ b/src/app/services/mail/mail.types.ts
@@ -71,7 +71,7 @@ export type SubmissionToAdminHtmlData = {
   appName: string
 }
 
-export type AutoreplyHtmlDefaultBody = {
+type AutoreplyHtmlDefaultBody = {
   submissionId: string
   autoReplyBody: string[]
 }
--- a/src/app/services/mail/mail.types.ts
+++ b/src/app/services/mail/mail.types.ts
@@ -71,12 +71,14 @@ export type SubmissionToAdminHtmlData = {
   appName: string
 }
 
+export type AutoreplyHtmlDefaultBody = {
+  submissionId: string
+  autoReplyBody: string[]
+}
+
 export type AutoreplyHtmlData =
-  | ({
-      responseId: string
-      autoReplyBody: string[]
-    } & AutoreplySummaryRenderData)
-  | { responseId: string; autoReplyBody: string[] }
+  | (AutoreplyHtmlDefaultBody & AutoreplySummaryRenderData)
+  | AutoreplyHtmlDefaultBody
 
 export type BounceNotificationHtmlData = {
   formTitle: string
--- a/src/public/modules/forms/base/css/end-page.css
+++ b/src/public/modules/forms/base/css/end-page.css
@@ -7,10 +7,10 @@
   font-weight: bold;
 }
 
-#end-page-header #end-page-responseid {
+#end-page-header #end-page-submissionId {
   font-size: 12px;
 }
 
-#end-page-header #end-page-responseid #end-page-id {
+#end-page-header #end-page-submissionId #end-page-id {
   text-transform: lowercase;
 }
--- a/src/public/modules/forms/base/css/end-page.css
+++ b/src/public/modules/forms/base/css/end-page.css
@@ -6,3 +6,11 @@
 #end-page-header #end-page-timestamp {
   font-weight: bold;
 }
+
+#end-page-header #end-page-responseid {
+  font-size: 12px;
+}
+
+#end-page-header #end-page-responseid #end-page-id {
+  text-transform: lowercase;
+}
--- a/src/public/modules/forms/base/directives/submit-form.directive.js
+++ b/src/public/modules/forms/base/directives/submit-form.directive.js
@@ -59,6 +59,7 @@ function submitFormDirective(
       logoUrl: '<',
       myInfoError: '<',
       disableSubmitButton: '<',
+      responseId: '<',
     },
     link: function (scope, _element, _attrs, _ctrl) {
       const startDate = Date.now() // Used to calculate time spent on form
@@ -321,7 +322,7 @@ function submitFormDirective(
             responseMode: form.responseMode,
           },
           submissionContent, // POST body
-        ).then(handleSubmitSuccess, handleSubmitFailure)
+        ).then((response) => handleSubmitSuccess(response), handleSubmitFailure)
       }
 
       /**
@@ -334,7 +335,8 @@ function submitFormDirective(
        * Returns a callback for form submission success, which updates UI
        * state and Google Analytics.
        */
-      const handleSubmitSuccess = () => {
+      const handleSubmitSuccess = (response) => {
+        scope.responseId = response.id
         setFormState(FORM_STATES.SUBMITTED)
         GTag.submitFormSuccess(scope.form, startDate, Date.now())
         if (shouldTrackPersistentLoginUse()) {
--- a/src/public/modules/forms/base/directives/submit-form.directive.js
+++ b/src/public/modules/forms/base/directives/submit-form.directive.js
@@ -321,7 +321,10 @@ function submitFormDirective(
             responseMode: form.responseMode,
           },
           submissionContent, // POST body
-        ).then((response) => handleSubmitSuccess(response), handleSubmitFailure)
+        ).then(
+          (response) => handleSubmitSuccess(response),
+          (err, message) => handleSubmitFailure(err, message),
+        )
       }
 
       /**
--- a/src/public/modules/forms/base/directives/submit-form.directive.js
+++ b/src/public/modules/forms/base/directives/submit-form.directive.js
@@ -335,7 +335,7 @@ function submitFormDirective(
        * state and Google Analytics.
        */
       const handleSubmitSuccess = (response) => {
-        scope.responseId = response.submissionId
+        scope.submissionId = response.submissionId
         setFormState(FORM_STATES.SUBMITTED)
         GTag.submitFormSuccess(scope.form, startDate, Date.now())
         if (shouldTrackPersistentLoginUse()) {
--- a/src/public/modules/forms/base/directives/submit-form.directive.js
+++ b/src/public/modules/forms/base/directives/submit-form.directive.js
@@ -59,7 +59,6 @@ function submitFormDirective(
       logoUrl: '<',
       myInfoError: '<',
       disableSubmitButton: '<',
-      responseId: '<',
     },
     link: function (scope, _element, _attrs, _ctrl) {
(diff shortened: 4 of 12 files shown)


src/public/modules/forms/base/css/end-page.css
```diff
@@ -6,3 +6,11 @@
...
   font-weight: bold;
 }
+
+#end-page-header #end-page-responseid {
+  font-size: 12px;
+}
+
+#end-page-header #end-page-responseid #end-page-id {
+  text-transform: lowercase;
+}
```
seaerchin, line 16: correct me if i'm wrong but wouldn't this also apply styling to `end-page-header` where there wasn't previously?

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -321,7 +321,7 @@ function submitFormDirective(
             responseMode: form.responseMode,
           },
           submissionContent, // POST body
-        ).then(handleSubmitSuccess, handleSubmitFailure)
+        ).then((response) => handleSubmitSuccess(response), handleSubmitFailure)
```
seaerchin, line 324: would it be better to omit the parameter for consistency? (or add the error to `handleSubmitFailure` also)

seaerchin (commented):
looks good! but have some minor comments about small stuff

seaerchin:
btw i think cos you branched off another PR, you'd rebase off develop (once it's merged) so that old commits aren't seen or if it isn't, set the compared branch as the base branch!n

src/public/modules/forms/base/css/end-page.css
chowyiyin: If I'm not wrong, the style is applied to `end-page-responseid` which is a descendant of the `end-page-header` element. It only applies to `end-page-header` as well if I insert a comma between the two, which will then make them two separate selectors.

src/public/modules/forms/base/directives/submit-form.directive.js
chowyiyin: The parameter was added as the `handleSubmitSuccess` function now has to extract the response ID from the response. I can add the error parameter the `handleSubmitFailure`.

seaerchin: i think how it works is the parameters get passed implicitly actually! but yea, if it adds clarity then should add the error parameter to `handleSubmitFailure`

src/public/modules/forms/base/css/end-page.css
seaerchin: ah ok you're right i think how it works is through the [descendant combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator)!

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -133,8 +135,11 @@ function SubmissionsFactory(
         version: body.version,
       })
       .then(
-        function () {
-          deferred.resolve('Submission has finished.')
+        function (response) {
+          deferred.resolve({
+            message: 'Submission has finished.',
+            id: response.submissionId,
```
mantariksh, line 141: would it be better to name this key `submissionId` so it's clear exactly what ID it refers to?

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -93,18 +93,20 @@ function SubmissionsFactory(
...
+        let response = {}
+        try {
+          response = JSON.parse(xhr.responseText)
+          // eslint-disable-next-line no-empty
+        } catch (e) {}
         if (xhr.status === HttpStatus.OK) {
-          deferred.resolve('Submission has finished.')
+          deferred.resolve({
+            message: 'Submission has finished.',
+            id: response.submissionId || 'Not available',
```
mantariksh, line 107: I don't think we need the `|| 'Not available'` since the API shape should always be consistent

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -93,18 +93,20 @@ function SubmissionsFactory(
...
     xhr.onreadystatechange = function () {
-      // 4 DONE  The operation is complete.
       const OPERATION_DONE = 4
       if (xhr.readyState === OPERATION_DONE) {
         // waterfall is successful
+        let response = {}
+        try {
+          response = JSON.parse(xhr.responseText)
+          // eslint-disable-next-line no-empty
+        } catch (e) {}
```
mantariksh, line 103: we shouldn't ever leave errors unhandled (as the eslint rule suggests haha) - if the response cannot be parsed, we should do `deferred.reject`

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -353,6 +357,7 @@ function submitFormDirective(
        * @param {string?} toastMessage The toast message to display, if any.
        */
       const handleSubmitFailure = (error, toastMessage) => {
+        scope.responseId = ''
```
mantariksh, line 360: can I check why this is necessary?

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -321,7 +321,10 @@ function submitFormDirective(
             responseMode: form.responseMode,
           },
           submissionContent, // POST body
-        ).then(handleSubmitSuccess, handleSubmitFailure)
+        ).then(
+          (response) => handleSubmitSuccess(response),
+          (err, message) => handleSubmitFailure(err, message),
```
mantariksh, line 326: this is actually equivalent to the previous version (`.then(handleSubmitSuccess, handleSubmitFailure)`) - let's chat further if you're not sure why!

karrui:
can i request for some screenshots too since this is a client side change?

src/public/modules/forms/base/directives/submit-form.directive.js
chowyiyin: Oh right. I realised it is not necessary because the thank you page will not render if submission fails. I will remove this and also make the necessary changes for the comments above as well.

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -93,23 +93,24 @@ function SubmissionsFactory(
...
+          if (xhr.status === HttpStatus.OK) {
+            deferred.resolve({
+              message: 'Submission has finished.',
+              submissionId: response.submissionId,
+            })
+          } else {
+            deferred.reject(`${response.message}`)
+          }
+          // eslint-disable-next-line no-empty
+        } catch (e) {
```
seaerchin, line 111: nit: if this is unused, then we can omit. but should we default to the message on the error or this string?

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -93,23 +93,24 @@ function SubmissionsFactory(
...
+          response = JSON.parse(xhr.responseText)
+          if (xhr.status === HttpStatus.OK) {
+            deferred.resolve({
+              message: 'Submission has finished.',
+              submissionId: response.submissionId,
+            })
+          } else {
+            deferred.reject(`${response.message}`)
+          }
+          // eslint-disable-next-line no-empty
```
seaerchin, line 110: would it be better to remove this since it is unused?

src/public/modules/forms/services/submissions.client.factory.js
chowyiyin: If I'm not wrong, I think this error message will be shown to the user so I think it would be better to keep this original error message to inform the user of possible actions they can take, rather than the actual error message that a user might not understand?

seaerchin: oh just to be clear, i'm talking about the `e` parameter, which isn't used (i think), so maybe should remove!

src/app/services/mail/mail.types.ts
```diff
@@ -72,8 +72,11 @@ export type SubmissionToAdminHtmlData = {
 }
 
 export type AutoreplyHtmlData =
-  | ({ autoReplyBody: string[] } & AutoreplySummaryRenderData)
-  | { autoReplyBody: string[] }
+  | ({
+      responseId: string
+      autoReplyBody: string[]
+    } & AutoreplySummaryRenderData)
+  | { responseId: string; autoReplyBody: string[] }
```
mantariksh, line 79: would it be better to extract `{ responseId: string; autoReplyBody: string[] }` into a separate type? this is the DRY principle

mantariksh: also, let's try to always refer to the submission ID as `submissionId` in the code. I know we use the name `responseId` for user-facing communication, but let's keep it consistent in the code so devs always know what it refers to.

src/app/services/mail/mail.service.ts
```diff
@@ -241,10 +241,12 @@ export class MailService {
       autoReplyMailData.sender || form.admin.agency.fullName
     ).replace('(', '\\(')
 
+    const responseId = submission.id
     const defaultBody = `Dear Sir or Madam,\n\nThank you for submitting this form.\n\nRegards,\n${form.admin.agency.fullName}`
     const autoReplyBody = (autoReplyMailData.body || defaultBody).split('\n')
 
     const templateData = {
+      responseId,
```
mantariksh, line 249: instead of saving `responseId` to a variable, I would much rather just do:
```
const templateData = {
  responseId: submission.id
}
```
there are a couple of reasons for this:
1. it's easier to read. currently if I'm not sure what `responseId` refers to, I have to go and find where the variable was declared. if we avoid declaring the variable, it's immediately clear that it's the submissionId.
2. we generally try to avoid holding state in variables. this is because it's hard to tell if a variable (especially an object) has been modified along the way. if we just declare them inline, it's clear that they haven't been modified.

src/public/modules/forms/services/submissions.client.factory.js
mantariksh: yeah I think you can just delete the `e`, so you end up with `} catch () {`

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -93,23 +93,23 @@ function SubmissionsFactory(
     xhr.send(fd)
     // On Response
     xhr.onreadystatechange = function () {
-      // 4 DONE  The operation is complete.
       const OPERATION_DONE = 4
       if (xhr.readyState === OPERATION_DONE) {
         // waterfall is successful
```
mantariksh, line 98: Screenshot 2021-05-18 at 10 15 35 AM

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -93,23 +93,23 @@ function SubmissionsFactory(
...
+        let response = {}
+        try {
+          response = JSON.parse(xhr.responseText)
+          if (xhr.status === HttpStatus.OK) {
+            deferred.resolve({
+              message: 'Submission has finished.',
+              submissionId: response.submissionId,
+            })
+          } else {
+            deferred.reject(`${response.message}`)
```
mantariksh, line 108: let's keep the default message here as a backup, i.e. `deferred.reject(response.message || 'Please refresh...`

src/public/modules/forms/base/directives/submit-form.directive.js
```diff
@@ -334,7 +334,8 @@ function submitFormDirective(
        * Returns a callback for form submission success, which updates UI
        * state and Google Analytics.
        */
-      const handleSubmitSuccess = () => {
+      const handleSubmitSuccess = (response) => {
+        scope.responseId = response.submissionId
```
mantariksh, line 338: again, let's refer to the submission ID as `submissionId` in the code for consistency

mantariksh (changes requested):
is it possible to increase the padding between the reference number, the horizontal line and the body of the actual email confirmation?

src/public/modules/forms/services/submissions.client.factory.js
seaerchin: HAHAHAHAHHA

src/public/modules/forms/services/submissions.client.factory.js
chowyiyin: I think removing the `e` causes an error (`Identifier expected`).

seaerchin: from the [mdn docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch), i think you can just omit `()` entirely, which is what is subpect is causing the error!
Screenshot from 2021-05-18 17-39-09

src/app/services/mail/mail.types.ts
```diff
@@ -71,9 +71,14 @@ export type SubmissionToAdminHtmlData = {
   appName: string
 }
 
+export type AutoreplyHtmlDefaultBody = {
```
mantariksh, line 74: this type is not meant to be used outside right? so you don't need to export it. principle of encapsulation - hide the internal representation of data in this module; only expose what is necessary for other modules to work with it

mantariksh (commented):
lgtm, to test on staging before approval

mantariksh:
also could you add more manual tests to cover all the cases? eg
- storage mode
- email confirmation preview in edit field modal
- actual email confirmation
- submission in admin preview mode for both email and storage mode
- email confirmation in admin preview mode for both email and storage mode

src/public/modules/forms/services/submissions.client.factory.js
chowyiyin: Eslint complains about this somehow 
Screenshot 2021-05-19 at 11 08 53 AM

mantariksh: nvm just keep the `e` lmao

mantariksh (approved):

PR #1858, opened 2021-05-11
fix(collaborator-modal/client/controller): bugfix for spinner ui

seaerchin:
## Problem
fixes the infinite spinner on UI when adding collaborators. this was due to an erroneous change (by me) from changing the `then` block into a `catch`. this was made because i thought that it should still be a catch block (as there is an error) but this is not the case because it is an ordinary value now. 

## Solution
changes it back into a `then` block

### DIFF
--- a/src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
@@ -237,7 +237,7 @@ function CollaboratorModalController(
     )
 
     $scope.btnStatus = 2 // pressed; loading
-    $scope.updatePermissionList(permissionList).catch((err) => {
+    $scope.updatePermissionList(permissionList).then((err) => {
       if (err) {
         // Make the alert message correspond to the error code
         if (err.response.status === StatusCodes.BAD_REQUEST) {


src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -237,29 +237,34 @@ function CollaboratorModalController(
...
-        resetCollabForm()
-      }, 1000)
-    })
+        $timeout(() => {
+          resetCollabForm()
+        }, 1000)
+      })
+      .catch((err) => {
+        console.log('ERR', err)
+        if (err) {
```
mantariksh, line 252: there's no need for this `if` anymore right?

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -237,29 +237,34 @@ function CollaboratorModalController(
...
+            Toastr.error(
+              'Please ensure that the email entered is a valid government email. If the error still persists, refresh and try again later.',
+            )
+          } else if (err.response.status === StatusCodes.UNPROCESSABLE_ENTITY) {
+            Toastr.error(`${email} is not part of a whitelisted agency.`)
+          } else {
+            Toastr.error('Error adding collaborator.')
+          }
+          resetCollabForm()
+          return
```
mantariksh, line 264: the `return` is no longer necessary right?

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -237,29 +237,34 @@ function CollaboratorModalController(
...
-      $timeout(() => {
-        resetCollabForm()
-      }, 1000)
-    })
+        $timeout(() => {
+          resetCollabForm()
+        }, 1000)
+      })
+      .catch((err) => {
+        console.log('ERR', err)
```
karrui, line 252: remove console log?

karrui (changes requested):
remove debug logging

karrui (approved):

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
seaerchin: you're right - this is sloppy and i should've removed it

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
seaerchin: yeap - again, sloppy from me and should've removed it

Issue #1522, opened 2021-03-30
[API] Migrate verifed field related endpoints to new /api counterparts


karrui:
This issue tracks the migration of verified field related endpoints to their new `api/forms/:formId/fieldverifications` counterparts.

The old endpoint should not be deleted for backwards compatibility (at least for a few weeks after the new endpoint has been released).

The current to new mapping is shown below:

\# | Existing endpoints | New endpoints<br>(note that all routes are prefixed with /api/v3) | Description
-- | -- | --|--|
1 | POST /transaction  body: { formId } | POST /forms/:formId/fieldverifications | create verification transaction instance
2 | GET /transaction/:transactionId | GET /forms/:formId/fieldverifications/:id | retrieve transaction by id
3 | POST /transaction/:transactionId/reset<br>body: { fieldId } | POST /forms/:formId/fieldverifications/:id/fields/:fieldId/reset | reset field verification by id
4 | POST /transaction/:transactionId/otp<br>body: { fieldId, answer } | POST /forms/:formId/fieldverifications/:id/fields/:fieldId/otp/generate<br>body: { answer, fieldType } | generate OTP for field
5 | POST /transaction/:transactionId/otp/verify<br>body: { fieldId, otp} | POST /forms/:formId/fieldverifications/:id/fields/:fieldId/otp/verify<br>body: { otp } | verify OTP for field

seaerchin:
@karrui not migrating #2 as per the PR because it's unused

seaerchin:
@karrui what do you feel about changing the endpoint for 3-5 to the plural form

karrui:
> @karrui what do you feel about changing the endpoint for 3-5 to the plural form

pls do idk why i missed those

seaerchin:
@karrui are we fully trustring the frontend on the otp generation side? from what i can see, `answer` can be any string - should we instead discriminate on the `type` of answer? (otp/email) and then validate that too

seaerchin:
per discussion, will do the above

PR #1866, opened 2021-05-11
refactor(verification): migrate verified field

seaerchin:
## Problem
Part 1 of #1522 
This migrates the first `/POST` endpoint to the new API v3 route. This allows for us to shift our new routes to be grouped together and allows us to follow a RESTful api structure. 

## Solution
Shift the old endpoint to be rooted at `/api/v3` and change the structure of the route so that the form id is passed as a parameter in the URL

## Tests
Ported over existing tests and changed them so that the `formId` parameter exists on the request

## Manual Test
- [ ] create a form with a verified email field. go to the public form view; enter a email and click on verify. there should be a pop-up for the otp to be entered
- [ ] create a form with a verified mobile number field. go to the public form view; enter a number and click on verify. there should be a pop-up for the otp to be entered


src/app/modules/verification/verification.controller.ts
```diff
@@ -53,6 +54,45 @@ export const handleCreateTransaction: RequestHandler<
...
+/**
+ * NOTE: Private handler for POST /forms/:formId/fieldverifications
+ * When a form is loaded publicly, a transaction is created, and populated with the field ids of fields that are verifiable.
+ * If no fields are verifiable, then it did not create a transaction and returns an empty object.
+ * @param req
+ * @param res
+ * @returns 201 - transaction is created
+ * @returns 200 - transaction was not created as no fields were verifiable for the form
+ */
+export const handleCreateTransactionWithFieldId: RequestHandler<
```
karrui, line 66: the handler name does not seem to match its intended behaviour? should it be `handleCreateVerificationTransaction` instead?

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    })
+
+    it('should return 200 when form has no verifiable fields', async () => {
+      // Act
+      const response = await request
+        // ID of form with no verifiable fields
+        .post(`/forms/${mockEmptyFormId}/fieldverifications`)
+
+      // Assert
+      expect(response.status).toBe(StatusCodes.OK)
```
karrui, line 79: may want to assert on returned shape too

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  beforeAll(async () => await dbHandler.connect())
+
+  beforeEach(async () => {
+    request = session(verificationApp)
+    jest.clearAllMocks()
+    await dbHandler.clearDatabase()
+
+    // Form without verifiable fields
+    const { form: emptyForm } = await dbHandler.insertEmailForm()
```
karrui, line 41: may want to inline this since this is only used in a single test. Same for the verifiable form

src/public/services/FieldVerificationService.ts
```diff
@@ -16,6 +16,10 @@ type VerifiedFieldSignature = Opaque<string, 'VerifiedFieldSignature'>
 /** Exported for testing */
 export const TRANSACTION_ENDPOINT = '/transaction'
```
karrui, line 17: mark as deprecated?

src/public/services/FieldVerificationService.ts
```diff
@@ -16,6 +16,10 @@ type VerifiedFieldSignature = Opaque<string, 'VerifiedFieldSignature'>
...
 
+/** Exported for testing */
+export const FORM_API_PREFIX = '/api/v3/forms'
+export const VERIFICATION_ENDPOINT = 'fieldverifications'
```
karrui, line 21: nit: i don't think this endpoint needs to be a constant tbh

karrui (changes requested):

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
seaerchin: good point - will do this!

src/public/services/FieldVerificationService.ts
seaerchin: sure, will do so!

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
seaerchin: i didn't do this cos it'll be reused for the related endpoints in the issues @__@ if it's an issue, i can change it!

karrui (approved):
lgtm

Issue #1864, opened 2021-05-11
Migrate to Github Actions due to Travis memory limits


liangyuanruo:
**Problem**
Tests are frequently failing on Travis due to out-of-memory errors. Example log:

```text
 form_feedback.server.model › Schema › should create and save successfully
    spawn ENOMEM
      at Object.spawn [as default] (node_modules/mongodb-memory-server-core/node_modules/cross-spawn/index.js:12:24)
      at MongoInstance.Object.<anonymous>.MongoInstance._launchMongod (node_modules/mongodb-memory-server-core/src/util/MongoInstance.ts:228:36)
      at MongoInstance.<anonymous> (node_modules/mongodb-memory-server-core/src/util/MongoInstance.ts:140:30)
      at step (node_modules/mongodb-memory-server-core/lib/util/MongoInstance.js:44:23)
      at Object.next (node_modules/mongodb-memory-server-core/lib/util/MongoInstance.js:25:53)
      at fulfilled (node_modules/mongodb-memory-server-core/lib/util/MongoInstance.js:16:58)
```

**Attempted fixes**
1. [Using nproc to limit number of workers spawned](https://github.com/opengovsg/FormSG/pull/1860)
2. [Using --runInBand option to limit process spawn](https://github.com/opengovsg/FormSG/pull/1863)

**Solution**
Migrate to GitHub Actions, since the same tests appear to pass immediately.

liangyuanruo:
Closing as #1870 addressed the immediate issue

PR #1870, opened 2021-05-12
build: migrate unit tests to GitHub Actions

mantariksh:
## Problem

Unit tests on Travis CI have been intermittently failing due to memory issues with Jest. It does not seem like the issue is with the tests themselves, since running `jest --logHeapUsage` shows that the memory usage does not increase consistently as more tests pass. This suggests that the memory limitation is with Travis.

Part of #1864 

## Solution

This PR migrates only the frontend and backend unit tests to GitHub Actions (GA) to test whether the tests pass consistently. If they do, we can consider migrating the entire CI pipeline to GA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
       - name: Load Node.js modules
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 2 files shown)


karrui (approved):
EL GEE TEE EM!!!!!!!!!!!!!!!!!!!!!!!!!!!1111oneoen!1!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,66 @@
+name: CI
+
+on: [push, pull_request]
```
yong-jie, line 3: for the `pull_request` hook, the [three default events](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) enabled are the `opened`, `reopened`, and `synchronize`. 

The `synchronize` event happens when the HEAD gets updated in a PR, like when someone pushes to a branch that has a PR open. This is an overlap with the `push` hook. Consider omitting the `synchronize` event to prevent repeated computation.

```yaml
on:
  push:
  pull_request:
    types: [opened, reopened]

```

yong-jie (approved):
one point to consider, otherwise lgtm

PR #1871, opened 2021-05-12
refactor(verified-fields): migrate get transaction

seaerchin:
## Problem
The verification belongs to a `field`, which is in itself a part of a `form`. Hence, in order for our endpoints to be RESTful, the verification service should belong to `field` and by extension, `form`. 

Part 2 of #1522

## Solution
This PR implements a `/GET transactionMetadata` endpoint (currently unused) as part of a RESTful API hierarchy. 

## Tests
Adds more unit tests to `getTransactionMetadata` for error cases 

## Manual Tests
None here because it's currently unused

### DIFF
--- a/src/app/modules/verification/verification.controller.ts
+++ b/src/app/modules/verification/verification.controller.ts
@@ -97,6 +97,9 @@ export const handleCreateTransactionWithFieldId: RequestHandler<
  * Returns a transaction's id and expiry time if it exists
  * @param req
  * @param res
+ * @returns 200 with transactionId/formId and expiry time when transaction exists
+ * @returns 404 when the transaction could not be found
+ * @returns 500 when database error occurs
  */
 export const handleGetTransactionMetadata: RequestHandler<
   {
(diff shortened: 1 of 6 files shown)


karrui:
If the endpoint is unused, we should delete it. I just checked and the current GET specific transactionId endpoint is indeed not used. Maybe this PR can be repurposed to removing the entire endpoint.

@mantariksh you wrote the original TS implementation, can help to double check that it is indeed unused?

seaerchin:
@karrui already checked with @mantariksh that it's unused; i thought it might be used in react. which is why i did it. if there's no use for it, i'll delete

karrui:
> @karrui already checked with @mantariksh that it's unused; i thought it might be used in react, which is why i did it. if there's no use for it, i'll delete!

The consensus is not add dangling endpoints until its really needed, so yes pls delete it

seaerchin:
np, i'll close this PR and delete the unused stuff in the original PR!

PR #1872, opened 2021-05-12
[WIP] Encapsulate construction and representation of parsed responses into a class

### LINKED ISSUE #1104: Encapsulate the construction and representation of parsed responses into a class
Implementing #101 turned out to be exceeding difficult because submission data was being duplicated every time the codebase required a slightly different representation:

- `autoReplyData` for email auto-replies
- `jsonData` for the data collation tool
- `formData` for saving encrypted responses or emailing form admin

The responsibility for validation and formatting is also distributed into controllers such as `spcp.controller.ts`, which are free to modify the `parsedResponses` array at any time.

TODO

Encapsulate the submission data into a single class
- Move the validation and construction of `parsedResponses` into the class constructor
- Add helper methods to allow the addition of SP/CP data
- Add helper methods to generate different representations of the submission data

OUTCOME

Controllers should not modify responses directly. Instead, all interaction with submission data should be performed through this class.

This is a very sensitive refactor which may cause responses to be lost or incompatible with the data collation tool. Great care should be taken during the refactoring, and tests written for each step of the way to ensure the preservation of existing behavior.

yong-jie:
## Problem
Closes #1104 

# Solution
ParsedResponsesObject to be made as a dataclass, meaning its primary job is to store data with minimal business logic to handle the presentation of its data.

More complex projections of its data (like for example, emails to admins, autoreply, data collation tool) will be carried out using double dispatch; it will be the responsibility of other classes/functions to implement logic of building these projections.

### What's done
- [x] Encapsulate parsed responses into a class, with regular responses and NDI responses in separate fields
- [x] Static init method that constructs ParsedResponseObj after validation
- [x] Email submissions endpoint refactored to use ParsedResponseObj
- [x] Encrypt submissions endpoint refactored to use ParsedResponseObj
- [x] Admin-form endpoint refactored to use ParsedResponseObj

### TODO
- [ ] Think of a name that's better than `ParsedResponseObject`
- [ ] Consider the need to separate SP/CP/MyInfo responses into separate fields rather than one lumped up `ndiResponses` field
- [ ] Figure out a good way to present responses in function calls
- [ ] Figure out a way to integrate with SubmissionEmailObj
- [ ] Change existing tests on getProcessedResponses to target the new static init method
- [ ] Delete getProcessedResponses
- [ ] Fix Admin form controller tests
- [ ] Additional tests on new functionality added to ParsedResponseObj

## Solution

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Tests

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/app/modules/submission/email-submission/email-submission.controller.ts
+++ b/src/app/modules/submission/email-submission/email-submission.controller.ts
@@ -26,6 +26,7 @@ import { IPopulatedEmailFormWithResponsesAndHash } from './email-submission.type
 import {
   mapAttachmentsFromResponses,
   mapRouteError,
+  ParsedResponsesObject,
   SubmissionEmailObj,
 } from './email-submission.util'
 
@@ -119,10 +120,7 @@ const submitEmailModeForm: RequestHandler<
         // Validate responses
         EmailSubmissionService.validateAttachments(req.body.responses)
           .andThen(() =>
-            SubmissionService.ParsedResponsesObject.parseResponses(
-              form,
-              req.body.responses,
-            ),
+            ParsedResponsesObject.parseResponses(form, req.body.responses),
           )
           .map((parsedResponses) => ({ parsedResponses, form }))
           .mapErr((error) => {
--- a/src/app/modules/submission/email-submission/email-submission.controller.ts
+++ b/src/app/modules/submission/email-submission/email-submission.controller.ts
@@ -119,7 +119,10 @@ const submitEmailModeForm: RequestHandler<
         // Validate responses
         EmailSubmissionService.validateAttachments(req.body.responses)
           .andThen(() =>
-            SubmissionService.getProcessedResponses(form, req.body.responses),
+            SubmissionService.ParsedResponsesObject.parseResponses(
+              form,
+              req.body.responses,
+            ),
           )
           .map((parsedResponses) => ({ parsedResponses, form }))
           .mapErr((error) => {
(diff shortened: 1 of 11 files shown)


src/app/modules/submission/submission.service.ts
```diff
@@ -166,6 +166,112 @@ export const getProcessedResponses = (
   return ok(processedResponses)
 }
 
+export class ParsedResponsesObject {
```
karrui, line 169: Should this be in its own file as a service? `ParsedResponseService` or something

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -257,7 +262,8 @@ const submitEmailModeForm: RequestHandler<
         // This is why sendSubmissionToAdmin is separated from sendEmailConfirmations in 2 blocks
         return MailService.sendSubmissionToAdmin({
           replyToEmails: EmailSubmissionService.extractEmailAnswers(
-            parsedResponses,
+            // TODO: A better solution might exist
+            [...parsedResponses.responses, ...parsedResponses.ndiResponses],
```
karrui, line 266: can always expose a function in the ParsedResponsesService for this with all the various checks you need

src/app/modules/submission/submission.service.ts
yong-jie: agreed that this object probably doesnt need to be in submission service anymore, but i was thinking that `ParsedResponsesObject` might actually be more of a dataclass than a service, in that its primary purpose is to store data and project it in different ways (similar to the `SubmissionsEmailObj` that will be replaced).

I think therefore, this object could go to `email-submission.util.ts`?

PR #1889, opened 2021-05-16
fix: add types to RequestHandler in submitEncryptModeForm

### LINKED ISSUE #1854: submitEncryptModeForm handler is not typed
The `submitEncryptModeForm` controller does not contain generic type arguments for Express's `RequestHandler`. This means that the body, route parameters and query parameters are typed as either `any` or `string`, even if they are `undefined` at runtime, so the entire controller is not type-safe.

Generic arguments should be added to `RequestHandler` in order to ensure type safety. Reference can be taken from `submitEmailModeForm` as the arguments will be quite similar.

frankchn:
## Problem
Closes #1854 

## Solution
Adds relevant types

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -1,6 +1,6 @@
 import JoiDate from '@joi/date'
 import { celebrate, Joi as BaseJoi, Segments } from 'celebrate'
-import { RequestHandler } from 'express'
+import { Request, RequestHandler } from 'express'
 import { Query } from 'express-serve-static-core'
 import { StatusCodes } from 'http-status-codes'
 import JSONStream from 'JSONStream'
@@ -10,6 +10,7 @@ import { RequireAtLeastOne, SetOptional } from 'type-fest'
 import {
   AuthType,
   EncryptedSubmissionDto,
+  FieldResponse,
   SubmissionMetadataList,
 } from '../../../../types'
 import { EncryptSubmissionDto, ErrorDto } from '../../../../types/api'
@@ -56,7 +57,24 @@ const EncryptSubmission = getEncryptSubmissionModel(mongoose)
 // NOTE: Refer to this for documentation: https://github.com/sideway/joi-date/blob/master/API.md
 const Joi = BaseJoi.extend(JoiDate)
 
-const submitEncryptModeForm: RequestHandler = async (req, res) => {
+const submitEncryptModeForm: RequestHandler<
+  { formId: string },
+  {
+    message: string
+    submissionId?: string
+    spcpSubmissionFailure?: boolean
+    isPageFound?: boolean
+    formTitle?: string
+  },
+  {
+    responses: FieldResponse[]
+    encryptedContent: string
+    isPreview: boolean
+    attachments: Record<string, unknown>
+    version: number
+  },
+  { captchaResponse?: unknown }
+> = async (req, res) => {
   const { formId } = req.params
 
   if ('isPreview' in req.body) {
@@ -72,7 +90,7 @@ const submitEncryptModeForm: RequestHandler = async (req, res) => {
 
   const logMeta = {
     action: 'submitEncryptModeForm',
-    ...createReqMeta(req),
+    ...createReqMeta(req as Request),
     formId,
   }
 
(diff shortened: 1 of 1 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -57,11 +58,28 @@ const EncryptSubmission = getEncryptSubmissionModel(mongoose)
...
 
-const submitEncryptModeForm: RequestHandler = async (req, res) => {
+const submitEncryptModeForm: RequestHandler<
+  { formId: string },
+  {
+    message: string
+    submissionId?: string
+    spcpSubmissionFailure?: boolean
+    isPageFound?: boolean
+    formTitle?: string
```
mantariksh, line 68: I think you can remove `isPageFound` and `formTitle`. they're not used by the frontend.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -57,11 +58,28 @@ const EncryptSubmission = getEncryptSubmissionModel(mongoose)
...
+  },
+  {
+    responses: FieldResponse[]
+    encryptedContent: string
+    isPreview: boolean
+    attachments?: Record<string, unknown>
+    version: number
+  },
```
mantariksh, line 76: can we reuse `EncryptSubmissionDto` instead?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -57,11 +58,28 @@ const EncryptSubmission = getEncryptSubmissionModel(mongoose)
...
+  { captchaResponse?: unknown }
```
liangyuanruo, line 77: I reckon it would help future developers if we gave meaningful names to these types, instead of anonymously inlining them here

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -57,11 +58,28 @@ const EncryptSubmission = getEncryptSubmissionModel(mongoose)
...
+> = async (req, res) => {
   const { formId } = req.params
   const logMeta = {
     action: 'handleEncryptedSubmission',
-    ...createReqMeta(req),
+    ...createReqMeta(req as Request),
```
liangyuanruo, line 82: can we avoid typecasts? this is related to the inlined types above, right?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
frankchn: I think it is still used in some bits of the code right? https://github.com/opengovsg/FormSG/blob/aecc1dbd44b4258584819a2b3e822347c1446be9/src/public/modules/forms/services/form-error.client.factory.js#L40-L44

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
frankchn: Hmm without this there is a type error -- we are doing the same thing in `submitEmailModeForm` so I think it is OK?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
frankchn: Any examples? I've changed part of this to use `EncryptSubmissionDto` instead. Should I do similar things for the rest of them?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
mantariksh: yup they're used elsewhere, but not in this particular case of when storage mode submissions failing.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
karrui: Should be ok for now; will be fixed when I reopen #1848

liangyuanruo: ok @karrui will leave to u to track this

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
liangyuanruo: yes that would help!

mantariksh (approved):

PR #1897, opened 2021-05-17
fix(start-page): remove extra logo props when logo type is not custom

### LINKED ISSUE #1885: Additional properties being sent over when logo options change
## Problem
Currently, when we change the logo type from `CUSTOM` to any of the other two, the first request sent over to the backend for this has additional properties from the `CUSTOM` logo type. 

This breaks the API contract between the backend and the frontend and prevents stricter validation on the shape of the request (because it is not uniform).

## Steps to reproduce 
1. Create a new form
2. Click on the welcome button
3. Edit the logo type to custom and save
4. Click on the welcome button again
5. Edit the logo type to any of the other two and save
6. Go to the network panel, rightclick the latest network request to /adminform 
7. Select copy as fetch 
8. Paste in console and verify that the logo property has additional properties beyond `state`

## Proposed solution
Fix the bug on the toggle and enforce stricter validation for the logo options 

## Images/Videos
recording

Screenshot 2021-05-14 at 2 19 42 PM

seaerchin:
## Problem
See #1885 for a description of the problem

Closes #1885 

## Solution
Added extra functionality to `vm.updateLogoUrl` to remove extra parameters on each button click because each button calls that upon change.

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -2218,6 +2218,7 @@ export const handleUpdateStartPage = [
         fileId: Joi.when('state', {
           is: FormLogoState.Custom,
           then: Joi.string().required(),
+          otherwise: Joi.any().forbidden(),
         }),
         fileName: Joi.when('state', {
           is: FormLogoState.Custom,
@@ -2226,10 +2227,12 @@ export const handleUpdateStartPage = [
             // Refer to https://regex101.com/ with the below regex for a full explanation
             .pattern(/\.(gif|png|jpeg|jpg)$/im)
             .required(),
+          otherwise: Joi.any().forbidden(),
         }),
         fileSizeInBytes: Joi.when('state', {
           is: FormLogoState.Custom,
           then: Joi.number().max(MAX_UPLOAD_FILE_SIZE).required(),
+          otherwise: Joi.any().forbidden(),
         }),
       }).required(),
     },
--- a/src/public/modules/forms/admin/controllers/edit-start-page-modal.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/edit-start-page-modal.client.controller.js
@@ -43,8 +43,14 @@ function EditStartPageController(
 
   vm.saveStartPage = function (isValid) {
     vm.hasClickedSave = true
+    const logoState = vm.myform.startPage.logo.state
 
     if (isValid) {
+      if (logoState !== FormLogoState.Custom) {
+        vm.myform.startPage.logo = {
+          state: logoState,
+        }
+      }
       $q.when(updateStartPage({ newStartPage: vm.myform.startPage })).then(
         (error) => {
           if (!error) {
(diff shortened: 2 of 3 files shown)


src/public/modules/forms/admin/views/edit-start-page.client.modal.html
```diff
@@ -18,7 +18,7 @@ <h2 class="modal-title">Edit Welcome</h2>
               <label class="col-xs-12">
                 <input
                   type="radio"
-                  ng-change="vm.updateLogoUrl()"
+                  ng-change="vm.updateLogoUrl(vm.myform.startPage.logo.state)"
```
karrui, line 21: since this happens when the user changes the radio, when they switches back to custom logo without saving, the uploaded file details will be cleared right?

i think the better place to clear the extra states should be just right before the endpoint is called

seaerchin: good catch, will fix!

tshuli (approved):
lgtm

src/public/modules/forms/admin/controllers/edit-start-page-modal.client.controller.js
```diff
@@ -43,8 +43,14 @@ function EditStartPageController(
 
   vm.saveStartPage = function (isValid) {
     vm.hasClickedSave = true
+    const logoState = vm.myform.startPage.logo.state
```
karrui, line 46: would it be better if we cloned the startPage before resetting the logo? So the parameter to updateStartPage is the cloned object, but will not affect the "real" start page in case of failure

seaerchin: good point! will change

src/public/modules/forms/admin/controllers/edit-start-page-modal.client.controller.js
```diff
@@ -43,8 +43,14 @@ function EditStartPageController(
...
 
     if (isValid) {
+      if (logoState !== FormLogoState.Custom) {
+        vm.myform.startPage.logo = {
+          state: logoState,
+        }
+      }
```
mantariksh, line 53: this should be done after the update is successful, right?

seaerchin: no, this should be done _prior_. This is because there's additional properties sent on the request, which prevents stricter joi validation.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2218,6 +2218,7 @@ export const handleUpdateStartPage = [
         fileId: Joi.when('state', {
           is: FormLogoState.Custom,
           then: Joi.string().required(),
+          otherwise: Joi.any().forbidden(),
```
karrui, line 2221: ensure that this PR goes out with the specifc start page update PR, or it won't be backwards compatible

karrui (approved):
lgtm

src/app/modules/form/admin-form/admin-form.controller.ts
seaerchin: yeap, the base branch for this PR is the `start-page` branch, will merge this into the dependent branch before finally merging into develop

PR #1900, opened 2021-05-17
refactor(verification): reset field verification state when field changes

seaerchin:
## Problem
Restructures the endpoints so that the our endpoints follow a RESTful hierarchy because the verification state is part of fields, which is in turn part of the form. 

Part of #1522 

## Manual Tests
tested on staging 
- [ ] Create a new form and set mobile field to require verification. Publish the form and navigate to it. Key in a valid email, then click verify. Click backspace (so that the dropdown to enter the otp disappears) then, reenter the email again. Verify the email with the **previous** otp. This should fail. Next, verify with the **new** otp. This should pass. Attempt to submit the form after verification. This should pass
- [ ] Repeat the above with email field
- [ ] Create a form with >1 fields that require verification. Attempt to submit without 0 verifications. Attempt to submit with less than the required amount of verifications. Both should fail. Submitting with the required amount of verifications should pass

### DIFF
--- a/src/app/modules/verification/__tests__/verification.controller.spec.ts
+++ b/src/app/modules/verification/__tests__/verification.controller.spec.ts
@@ -799,4 +799,125 @@ describe('Verification controller', () => {
       expect(mockRes.json).toHaveBeenCalledWith({ message: expect.any(String) })
     })
   })
+
+  describe('handleResetFieldVerification', () => {
+    const MOCK_REQ = expressHandler.mockRequest({
+      params: {
+        transactionId: MOCK_TRANSACTION_ID,
+        fieldId: MOCK_FIELD_ID,
+        formId: MOCK_FORM_ID,
+      },
+    })
+
+    it('should correctly call service when params are valid', async () => {
+      // Arrange
+      MockVerificationFactory.resetFieldForTransaction.mockReturnValueOnce(
+        okAsync(mockTransaction),
+      )
+
+      // Act
+      await VerificationController.handleResetFieldVerification(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      // Assert
+      expect(
+        MockVerificationFactory.resetFieldForTransaction,
+      ).toHaveBeenCalledWith(MOCK_TRANSACTION_ID, MOCK_FIELD_ID)
+      expect(mockRes.sendStatus).toHaveBeenCalledWith(StatusCodes.OK)
+    })
+
+    it('should return 400 when transaction has expired', async () => {
+      // Arrange
+      MockVerificationFactory.resetFieldForTransaction.mockReturnValueOnce(
+        errAsync(new TransactionExpiredError()),
+      )
+
+      // Act
+      await VerificationController.handleResetFieldVerification(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      // Assert
+      expect(
+        MockVerificationFactory.resetFieldForTransaction,
+      ).toHaveBeenCalledWith(MOCK_TRANSACTION_ID, MOCK_FIELD_ID)
+      expect(mockRes.status).toHaveBeenCalledWith(StatusCodes.BAD_REQUEST)
+      expect(mockRes.json).toHaveBeenCalledWith({
+        message: expect.any(String),
+      })
+    })
+
+    it('should return 404 when transaction is not found', async () => {
+      // Arrange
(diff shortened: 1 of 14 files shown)


src/app/routes/api/v3/forms/public-forms.verification.routes.ts
```diff
@@ -7,3 +7,14 @@ export const PublicFormsVerificationRouter = Router()
 PublicFormsVerificationRouter.route(
   '/:formId([a-fA-F0-9]{24})/fieldverifications',
 ).post(VerificationController.handleCreateVerificationTransaction)
+
+/**
+ * Route for resetting the verification of a given field
+ * @returns 400 when the transaction has expired
+ * @returns 404 when the transaction could not be found
+ * @returns 404 when the field could not be found
```
tshuli, line 16: On success, should the route return 200?

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -94,4 +138,82 @@ describe('public-forms.verification.routes', () => {
       })
     })
   })
+
+  describe('POST /forms/:formId/fieldverification/:transactionId/fields/:fieldId/reset', () => {
+    it('should return 200 when transactionId and fieldId for email field are valid', async () => {
```
tshuli, line 143: nit: when `formId`, transactionId and fieldId

src/app/routes/api/v3/forms/public-forms.verification.routes.ts
seaerchin: It does but I missed out on the docs @__@ good catch!

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
seaerchin: Will fix, thanks!

src/public/services/FieldVerificationService.ts
```diff
@@ -89,18 +89,21 @@ export const verifyOtp = async ({
 /**
  * Reset the field in the transaction, removing the previously saved signature.
  *
+ * @param formId The id of the form to reset the transaction for
  * @param transactionId The generated transaction id for the form
  * @param fieldId The id of the verification field to reset
  * @returns 200 OK status if successfully reset
```
tshuli, line 95: For clarity, should this be returning 200 or void

seaerchin: the http status code is 200 but the request itself carries no information beyond that (and hence, `promise` returns void)!

karrui: haven't reviewed yet besides this thread, but if 200 and nothing is returned, i think a better HTTP code is `201 Created`

karrui: Oh but nothing is created, nvm

karrui: the response code should be `204 ( No Content )` then!

src/app/modules/verification/__tests__/verification.controller.spec.ts
```diff
@@ -799,4 +802,152 @@ describe('Verification controller', () => {
...
+
+      // Act
+      await VerificationController.handleResetFieldVerification(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      // Assert
+      expect(
```
mantariksh, line 876: should we expect that MockFormService was called correctly?

mantariksh (commented):
lgtm other than the issue of the correct http status code

src/public/services/FieldVerificationService.ts
seaerchin: ah cool, didn't realise this! pasting the mdn links [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses) for future readers!

src/app/modules/verification/__tests__/verification.controller.spec.ts
seaerchin: i agree we should! fixed in [here](https://github.com/opengovsg/FormSG/pull/1900/commits/a2a7d2f0a51285312dd20fcf42bbcb54b12e73fc)

mantariksh (approved):

karrui (approved):
lgtm

PR #1916, opened 2021-05-18
refactor(verification): migrate otp generation to new api endpoint

seaerchin:
**integration tests will come in a future PR to prevent review fatigue**
## Problem
Migrates generation of otp to new endpoint for a restful hierarchy. 

Part 4 of #1522 

## Solution
Adds new route and controller for generation of top

## Manual Tests 
- [ ] Create a form with a verifiable mobile field and publish it. Go to the form, and verify with a valid mobile number. There should be an OTP generated
- [ ] Follow the above, but on successful otp generation, go to the networks tab in the developer panel and `copy as fetch` for the request. Change the `fieldType` to something else. This should fail. Change the `answer` property to be an invalid number. this should also fail
- [ ] Create a form with a verifiable email field and publish it. Go to the form, and verify with a valid email. There should be an OTP generated
- [ ] Follow the above, but on successful otp generation, go to the networks tab in the developer panel and `copy as fetch` for the request. Change the `fieldType` to something else. This should fail. Change the `answer` property to be an invalid email. this should also fail.

### DIFF
--- a/src/app/modules/verification/verification.controller.ts
+++ b/src/app/modules/verification/verification.controller.ts
@@ -183,19 +183,18 @@ export const handleGetOtp: RequestHandler<
  * @param transactionId The id of the transaction to verify
  * @param formId The id of the form to verify
  * @param fieldId The id of the field to verify
- * @returns 200 when otp generated successfully
+ * @returns 201 when otp generated successfully
  * @returns 400 when the parameters could not be parsed
  * @returns 400 when the transaction has expired
- * @returns 400 when the otp data could not be retrieved from the database
  * @returns 400 when the otp could not be sent via sms
  * @returns 400 when the otp could not be sent via email
  * @returns 400 when the provided phone number is not valid
  * @returns 400 when the field type is not supported for validation
  * @returns 404 when the requested form was not found
+ * @returns 404 when the transaction was not found
+ * @returns 404 when the field was not found
  * @returns 422 when the user requested for a new otp without waiting
  * @returns 500 when the otp could not be hashed
- * @returns 500 when the transaction could not be found
- * @returns 500 when the field could not be found in the transaction
  * @returns 500 when there is a database error
  */
 export const _handleGenerateOtp: RequestHandler<
--- a/src/app/modules/verification/verification.controller.ts
+++ b/src/app/modules/verification/verification.controller.ts
@@ -2,10 +2,7 @@ import { celebrate, Joi, Segments } from 'celebrate'
 import { RequestHandler } from 'express'
 import { StatusCodes } from 'http-status-codes'
 
-import {
-  SALT_ROUNDS,
-  VERIFIED_FIELDTYPES,
-} from '../../../shared/util/verification'
+import { SALT_ROUNDS } from '../../../shared/util/verification'
 import { ErrorDto } from '../../../types/api'
 import { createLoggerWithLabel } from '../../config/logger'
 import { generateOtpWithHash } from '../../utils/otp'
@@ -243,7 +240,6 @@ export const _handleGenerateOtp: RequestHandler<
 export const handleGenerateOtp = [
   celebrate({
     [Segments.BODY]: Joi.object({
-      fieldType: Joi.string().valid(...VERIFIED_FIELDTYPES),
       answer: Joi.string().required(),
     }),
   }),
--- a/src/app/modules/verification/verification.controller.ts
+++ b/src/app/modules/verification/verification.controller.ts
@@ -1,3 +1,4 @@
+import { celebrate, Joi, Segments } from 'celebrate'
 import { RequestHandler } from 'express'
 import { StatusCodes } from 'http-status-codes'
 
@@ -7,6 +8,7 @@ import { createLoggerWithLabel } from '../../config/logger'
 import * as FormService from '../../modules/form/form.service'
 import { generateOtpWithHash } from '../../utils/otp'
 import { createReqMeta } from '../../utils/request'
+import * as FormService from '../form/form.service'
 
 import { VerificationFactory } from './verification.factory'
 import { Transaction } from './verification.types'
@@ -133,6 +135,7 @@ export const handleResetField: RequestHandler<
  * The current answer is signed, and the signature is also saved in the transaction, with the field id as the key.
  * @param req
  * @param res
+ * @deprecated in favour of handleGenerateOtp
  */
 export const handleGetOtp: RequestHandler<
   { transactionId: string },
@@ -168,6 +171,61 @@ export const handleGetOtp: RequestHandler<
       return res.status(statusCode).json({ message: errorMessage })
     })
 }
+
+/**
+ * NOTE: This is exported solely for testing
+ * Generates an otp when a user requests to verify a field.
+ * The current answer is signed, and the signature is also saved in the transaction, with the field id as the key.
+ * @param req
+ * @param res
+ */
+export const _handleGenerateOtp: RequestHandler<
+  { transactionId: string; formId: string; fieldId: string },
+  ErrorDto,
+  { answer: string }
+> = async (req, res) => {
+  const { transactionId, formId, fieldId } = req.params
+  const { answer } = req.body
+  const logMeta = {
+    action: 'handleGenerateOtp',
+    transactionId,
+    fieldId,
+    ...createReqMeta(req),
+  }
+  return FormService.retrieveFormById(formId)
+    .andThen(() => generateOtpWithHash(logMeta, SALT_ROUNDS))
+    .andThen(({ otp, hashedOtp }) =>
+      VerificationFactory.sendNewOtp({
+        fieldId,
+        hashedOtp,
+        otp,
+        recipient: answer,
+        transactionId,
+      }),
+    )
+    .map(() => res.sendStatus(StatusCodes.CREATED))
+    .mapErr((error) => {
+      logger.error({
+        message: 'Error creating new OTP',
+        meta: logMeta,
+        error,
+      })
+      const { errorMessage, statusCode } = mapRouteError(error)
+      return res.status(statusCode).json({ message: errorMessage })
+    })
+}
+
+/**
+ * Handler for the POST /forms/:formId/fieldverifications/:transactiond/fields/:fieldId/otp/generate endpoint
+ */
+export const handleGenerateOtp = [
+  celebrate({
+    [Segments.BODY]: Joi.object({
+      answer: Joi.string().required(),
+    }),
+  }),
+  _handleGenerateOtp,
+] as RequestHandler[]
 
 /**
  * When user submits their otp for the field, the otp is validated.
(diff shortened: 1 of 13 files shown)


src/app/modules/verification/verification.controller.ts
```diff
@@ -168,6 +175,81 @@ export const handleGetOtp: RequestHandler<
...
+    })
+}
+
+/**
+ * Handler for the POST /forms/:formId/fieldverifications/:transactionId/fields/:fieldId/otp/generate endpoint
+ */
+export const handleGenerateOtp = [
+  celebrate({
+    [Segments.BODY]: Joi.object({
+      fieldType: Joi.string().valid(...VERIFIED_FIELDTYPES),
```
mantariksh, line 242: this is not used in the controller right? doesn't seem necessary

src/app/modules/verification/verification.controller.ts
```diff
@@ -168,6 +175,81 @@ export const handleGetOtp: RequestHandler<
...
+      answer: Joi.string().when('fieldType', {
+        is: VerifiableFieldType.Email,
+        then: Joi.string().email(),
+        otherwise: Joi.string().pattern(/\+(65)\d{8}/m),
```
mantariksh, line 246: I think we do allow OTPs to international numbers right? so this might be too strict

src/public/modules/forms/base/components/verifiable-field.client.component.js
```diff
@@ -47,7 +48,11 @@ function verifiableFieldController($q, $timeout, $interval) {
         throw new Error('No transaction id')
       }
 
+      console.log(vm.fieldType)
```
mantariksh, line 51: remove

src/public/services/FieldVerificationService.ts
```diff
@@ -39,24 +41,33 @@ export const createTransactionForForm = async (
...
+  fieldType: VerifiableFieldType
 }): Promise<void> => {
-  return axios.post(`${TRANSACTION_ENDPOINT}/${transactionId}/otp`, {
-    fieldId,
-    answer,
-  })
+  return axios.post(
+    `${FORM_API_PREFIX}/${formId}/${VERIFICATION_ENDPOINT}/${transactionId}/fields/${fieldId}/otp/generate`,
+    {
+      fieldType,
```
mantariksh, line 67: again, fieldType doesn't seem necessary

src/types/field/fieldTypes.ts
```diff
@@ -52,3 +52,8 @@ export enum SPCPFieldTitle {
   CpUid = 'CorpPass Validated UID',
   CpUen = 'CorpPass Validated UEN',
 }
+
+export enum VerifiableFieldType {
+  Email = BasicField.Email,
+  Mobile = BasicField.Mobile,
+}
```
mantariksh, line 59: nit: is it possible to `Pick` from the enum?

mantariksh (changes requested):

src/app/modules/verification/verification.controller.ts
seaerchin: this is needed for stronger validation - `fieldType` is added as a discriminant because prior to this, we were doing validation but expecting any string to come through. 

since we're migrating to a new endpoint, can add stronger validation so that invalid emails/phone numbers are caught and prevented

seaerchin: this isn't present on the controller because it's strictly validation only and hence, is beyond the bounds of what the controller has to know

src/app/modules/verification/verification.controller.ts
seaerchin: oh dang never knew that - it's for all international numbers right?

src/public/modules/forms/base/components/verifiable-field.client.component.js
seaerchin: got careless - thanks for the spot!

src/types/field/fieldTypes.ts
seaerchin: that's alot better, it slipped my mind ._.

src/app/modules/verification/verification.controller.ts
mantariksh: 1. that's circular logic - you're adding something to validate and then validating it. this doesn't mean that the validation is important
2. what do you mean by "discriminant" and where were we doing the validation on fieldType before this?
3. if there is genuinely an application-level need for this validation, imo it should happen in the controller. we shouldn't implicitly delegate application logic to the Joi validators. they exist purely so that we can be sure of the types in the controllers.

mantariksh: note that I am referring to `fieldType`, not `answer`

src/app/modules/verification/verification.controller.ts
mantariksh: I think so, @karrui can confirm?

src/app/modules/verification/verification.controller.ts
mantariksh: I see what you mean by "discriminant" now. I'm still a bit hesitant to introduce new validation on public routes, given the many issues we've faced before re: phone number validation. bear in mind that the covid form uses verified fields, and literally the whole population could be using it, so lives are literally at stake. but if you've done your research on the reliability and universality of Joi validation, I'll trust your judgement.

seaerchin: i think given what you've said, the best (for now), would be to make the phone number validation as permissive as possible (i.e., defaulting to the previous implementation where any string is allowed), especially because yknow, people are counting on this @___@ but with a TODO / issue raised so that we're cognizant going forward that once this period ends (yes, i still believe covid will end and we'll go back to normalcy), we can correct for stronger phone number validation

src/types/field/fieldTypes.ts
seaerchin: oddly enough, `Pick` [doesn't actually work](https://github.com/microsoft/TypeScript/issues/33308). I think this is because enums are not actually types and are more akin to `const` declarations (they are values).

changing it to type and using `Extract` works but i want to refer to the sub-fields (cos of validation) rather than typing it and possibly making an error @__@

src/app/modules/verification/verification.controller.ts
karrui: ya, OTPs can be any number

src/types/field/fieldTypes.ts
karrui: can't it just be 
```ts
export type VerifiableFieldType = BasicField.Email | BasicField.Mobile
```
?

It will only still allow for only the two basic fields

src/app/modules/verification/verification.controller.ts
karrui, line 242: Not a fan of having extra keys that the controller does not actually need, just let the controller handle malformed data in this case, so we remove all the complexity of the conditional. We can still only allow the answer to be either an email or phone number string

src/app/modules/verification/verification.controller.ts
seaerchin: kk, won't raise the issue then but will change validation to all strings

src/app/modules/verification/verification.controller.ts
seaerchin: can i clarify on this point? my understanding of using `joi` is to ensure that the shape of the request adheres to what we expect. 

wouldn't this be better on the validator itself so that the controller can be ignorant of this (i.e., that the respective field type is valid) and avoid bringing in validation of the _request_ into the controller?

src/types/field/fieldTypes.ts
seaerchin: yeppers, the reason is cos the joi validator takes string and idw to type wrongly @__@

src/app/modules/verification/verification.controller.ts
karrui: fair enough, but i'd argue that this is just a guard right, since for example

lets say `fieldIdA` is a `BasicField.Mobile`

but you pass in `fieldType: BasicField.Email` and a valid email string into the POST body.

the controller will still need to validate that the fieldId is indeed a mobile field and thus the arguments are invalid. Joi validation does nothing for this case, and thus what is the point?

src/app/modules/verification/verification.controller.ts
```diff
@@ -168,6 +175,81 @@ export const handleGetOtp: RequestHandler<
...
+> = async (req, res) => {
+  const { transactionId, formId, fieldId } = req.params
+  const { answer } = req.body
+  const logMeta = {
+    action: 'handleGenerateOtp',
+    transactionId,
+    fieldId,
+    ...createReqMeta(req),
+  }
+  return FormService.retrieveFormById(formId)
```
mantariksh, line 213: could you add a comment here about why we retrieve the form and then don't do anything with it?

src/app/modules/verification/verification.controller.ts
seaerchin: @karrui @mantariksh agreed, i think i'll change it so that the validation takes place on the controller rather than within the service itself.

this allows us to introduce [nominal typing](answer) in our codebase with the help of [ts-brand](answer), as valid contact numbers/emails aren't just any string but rather, special instances of strings that have passed the validation upfront. this also allows us to parse rather than validate in that we are preserving information in the type system and have it be correct _by definition_ rather than validate it at the call site

mantariksh: is that necessary for this PR? might it be better to preserve existing functionality in this PR and introduce your suggested changes separately, so we can separately evaluate the impact of the new validation without coupling it to this PR (which is meant to be a refactor)?

seaerchin: @mantariksh good point, that slipped my mind - will make the appropriate changes here and introduce suggestions in another PR!

src/app/modules/verification/verification.controller.ts
```diff
@@ -240,11 +239,7 @@ export const handleGenerateOtp = [
   celebrate({
     [Segments.BODY]: Joi.object({
       fieldType: Joi.string().valid(...VERIFIED_FIELDTYPES),
```
karrui, line 246: this means we don't need this fieldType param anymore right?

src/app/modules/verification/verification.controller.ts
```diff
@@ -168,6 +174,82 @@ export const handleGetOtp: RequestHandler<
...
+  )
+}
+
+/**
+ * Handler for the POST /forms/:formId/fieldverifications/:transactionId/fields/:fieldId/otp/generate endpoint
+ */
+export const handleGenerateOtp = [
+  celebrate({
+    [Segments.BODY]: Joi.object({
+      fieldType: Joi.string().valid(...VERIFIED_FIELDTYPES),
```
mantariksh, line 246: this is no longer required right? otherwise lgtm

karrui (approved):
lgtm

mantariksh (approved):

PR #1924, opened 2021-05-19
test(integration): integration tests for generate otp

seaerchin:
## Problem
#1916 had no integration tests to prevent PR from being too big 

## Solution 
Adds integration tests for #1916 

This also adds a new test helper for twilio. This is because twilio uses a function to initialise and we (might) need to dynamically change the result of an api call.

as hoisting with `jest` using `mock<someName>` doesn't work even with the fix [here](https://github.com/kulshekhar/ts-jest/pull/1937), the alternative was to shift this out and make it its own file so that callers (in our test env) are actually calling our mock twilio constructor, and within the test, we can alter the returned result dynamically


src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +239,322 @@ describe('public-forms.verification.routes', () => {
...
+        buildCelebrateError({
+          body: {
+            key: 'answer',
+            message: 'Please ensure that the phone number provided is valid.',
+          },
+        }),
+      )
+    })
+
+    it('should return 400 when the parameters are malformed', async () => {
```
chowyiyin, line 330: Would it be better to compress this test and the expired transaction test into one test since they are testing the same thing? The description could be 'should return 400 when `getOtpData` returns null' because the malformed parameter and expired transaction is detected in `getOtpData` right?

seaerchin: this is an error on my end, should've created an expired transaction in the db. good catch!

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +239,325 @@ describe('public-forms.verification.routes', () => {
...
+    })
+
+    it('should return 201 when parameters for email field are valid', async () => {
+      // Act
+      const response = await request
+        .post(
+          `/forms/${mockVerifiableFormId}/fieldverifications/${mockTransactionId}/fields/${mockEmailFieldId}/otp/generate`,
+        )
+        .send({
+          fieldType: VerifiableFieldType.Email,
```
mantariksh, line 257: is this correct given the discussions on `fieldType` in the other PR?

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +239,325 @@ describe('public-forms.verification.routes', () => {
...
+          // NOTE: a valid mobile number is +65 followed by ANY 8 digits
+          answer: '+6512345678',
+        })
+
+      // Assert
+      expect(response.status).toBe(StatusCodes.CREATED)
+      expect(response.text).toBe(getReasonPhrase(StatusCodes.CREATED))
+    })
+
+    it('should return 400 when fieldType is email but the provided email is not valid', async () => {
```
mantariksh, line 283: again, is this correct given our discussion on the reliability of Joi validation? we don't want a situation where people are preventing from submitting because Joi hasn't updated their validation types

mantariksh (changes requested):
validation is too strict as per discussion on #1916

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +238,306 @@ describe('public-forms.verification.routes', () => {
...
+          // 7 digits after +65 instead of 8
+          answer: '+651234567',
+        })
+
+      // Assert
+      expect(response.status).toBe(StatusCodes.BAD_REQUEST)
+      expect(response.body).toEqual(expectedResponse)
+    })
+
+    it('should return 400 when the parameters are malformed', async () => {
```
mantariksh, line 321: can we describe this test more specifically pls? not really sure what "parameters are malformed" means

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +238,306 @@ describe('public-forms.verification.routes', () => {
...
+
+      // Assert
+      expect(response.status).toBe(StatusCodes.CREATED)
+      expect(response.text).toBe(getReasonPhrase(StatusCodes.CREATED))
+    })
+
+    it('should return 400 when fieldType is email but the provided email is not valid', async () => {
+      // Arrange
+      const expectedResponse = {
+        message: 'Sorry, something went wrong. Please refresh and try again.',
```
mantariksh, line 282: this is for the main PR, but should this have a more specific error message?

mantariksh (approved):
couple of suggested wording changes, otherwise lgtm

PR #1925, opened 2021-05-19
feat: upgrade myinfo-gov-client to 4.0.0

mantariksh:
Upgrades `@opengovsg/myinfo-gov-client` to the latest version, and makes the necessary update to accommodate the breaking changes.

The latest version includes a breaking change which makes nested scope data optional (see [PR](https://github.com/opengovsg/myinfo-gov-client/pull/30)). This is because there is no way of knowing whether a given `MyInfoPerson` data object contains the given nested scopes without knowing which scopes were requested in advance.

In FormSG's case, `vehicles.vehicleno` is the only nested scope we deal with, and this property is only accessed if the `vehicleno` field is requested, so the property access is safe. Hence the additional `?` optional chain was added purely to accommodate the new types from `myinfo-gov-client`.

## Manual tests
- [ ] Add a Vehicle Number field to a MyInfo form. Test that the form can be submitted for a MyInfo profile with no prefilled vehicle number.
- [ ] Use a profile with a prefilled vehicle number and check that the form can be submitted without editing the field.
- [ ] Use a profile with a prefilled vehicle number and edit the field. Check that the form can still be submitted.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4724,9 +4724,9 @@
       }
     },
     "@opengovsg/myinfo-gov-client": {
-      "version": "4.0.0-beta.0",
-      "resolved": "https://registry.npmjs.org/@opengovsg/myinfo-gov-client/-/myinfo-gov-client-4.0.0-beta.0.tgz",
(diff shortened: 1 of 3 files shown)


seaerchin (approved):
lgtm

karrui (approved):
lgtm

chowyiyin (approved):
lgtm

PR #1926, opened 2021-05-19
refactor(verification): migrate otp verification to new v3 routes

seaerchin:
**integration tests will come in a future PR to avoid review fatigue**
**another PR will be raised for tidying up the (now unused) routes in `FieldVerificationService`**

## Problem
Migrates verification of otp to new endpoint for a restful hierarchy.

Closes #1522

## Solution
Adds new route (`POST /forms/:formId/fieldverifications/:id/fields/:fieldId/otp/verify`) and controller for verification of otp 

**Breaking Changes** 

- [x] Yes - this PR contains breaking changes
    - This PR cannot be rolled back due to new endpoints being called on the client. 
- [ ] No - this PR is backwards compatible

PR #1928, opened 2021-05-19
test(verification): integration tests for otp verification

seaerchin:
## Problem
#1926 has no integration tests to prevent review fatigue. This PR is a follow up on that

## Solution
This adds integration tests for #1926


src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +235,249 @@ describe('public-forms.verification.routes', () => {
...
+          `/forms/${mockVerifiableFormId}/fieldverifications/${mockTransactionId}/fields/${mockMobileFieldId}/otp/verify`,
+        )
+        .send({ otp: MOCK_OTP })
+
+      // Assert
+      expect(response.status).toBe(StatusCodes.NOT_FOUND)
+      expect(response.body).toEqual(expectedResponse)
+    })
+
+    it('should return 404 when the the form could not be found', async () => {
```
chowyiyin, line 315: Small thing, double "the" typo in description

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +235,249 @@ describe('public-forms.verification.routes', () => {
...
+          `/forms/${new ObjectId().toHexString()}/fieldverifications/${mockTransactionId}/fields/${mockMobileFieldId}/otp/verify`,
+        )
+        .send({ otp: MOCK_OTP })
+
+      // Assert
+      expect(response.status).toBe(StatusCodes.NOT_FOUND)
+      expect(response.body).toEqual(expectedResponse)
+    })
+
+    it('should return 404 when the the transaction could not be found', async () => {
```
chowyiyin, line 333: Same as above.

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +235,249 @@ describe('public-forms.verification.routes', () => {
...
+      // Assert
+      expect(response.status).toBe(StatusCodes.INTERNAL_SERVER_ERROR)
+      expect(response.body).toEqual(expectedResponse)
+    })
+
+    it('should return 500 when database error occurs', async () => {
+      // Arrange
+      jest
+        .spyOn(VerificationModel, 'findById')
+        .mockReturnValue({ exec: jest.fn().mockRejectedValue('no') })
```
chowyiyin, line 466: Not sure, but is it okay that the `mockRejectedValue` is 'no' instead of an error?

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
seaerchin: good catch! will fix, thanks for spotting!

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
seaerchin: yeap! the reason this is ok (or not, because of `neverthrow` hehehe) is due to `neverthrow`'s  `ResultAsync.fromPromise`.

our calls to the database essentially returns `Promise<ModelSchema | null>`,  where the null is used to indicate an _expected_ error when something is off with the query and the requested information could not be found (eg: no model matching the specified options). 

however, promises and throwable functions actually have the possibility of an _exception_, which is an unexpected error that the called function isn't qualified to handle, and returns to teh caller. this is **not** encoded in the type signature and hence, is a mental overhead that engineers have to keep in the when calling functions, lest it crashes their system. In light of this, promises are actually typed as such: `Promise<your_return_type | exception>`

The `ResultAsync.fromPromise` function in `neverthrow` essentially allows us to sidestep this problem for promises (the corresponding function for throwables is `Result.fromThrowable`. This essentially allows us to catch a rejecting promise (or an error) and map it back using a function. Because the _type_ of exception is not encoded in ts' type system, this has to transform an unknown, back into a proper error type, which is done by `transformMongoError`

tl; dr: rejecting the promise causes it to go into error branch, and then a default error is returned from `transformMongoError` hence it's fine. thx for coming to my ted talk.

chowyiyin: Ohh okay thanks for sharing!

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
```diff
@@ -231,4 +235,249 @@ describe('public-forms.verification.routes', () => {
       expect(response.body).toEqual(expectedResponse)
     })
   })
+
+  describe('POST /forms/:formId/fieldverifications/:transactionId/fields/:fieldId/otp/verify', () => {
+    beforeAll(() => {
+      jest.spyOn(bcrypt, 'compare').mockResolvedValue(true)
```
karrui, line 241: does this need to be mocked? we should try to only mock things we cannot simulate in integration tests; we should instead  call the generate end point for real and see if verify returns success.

see our [auth integration tests](https://github.com/opengovsg/FormSG/blob/develop/src/app/routes/api/v3/auth/__tests__/auth.routes.spec.ts#L451)

karrui (changes requested):
update tests to not mock the generation of hash so it is as close to real world as possible

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
mantariksh: @chowyiyin fyi the team decided a while ago that integration tests only need to cover Joi validation and success cases, since the controller unit tests should already test for the various error cases. @seaerchin has put in a lot of effort to mock lower layers even in integration tests so that error cases can be tested, which is great, but requires quite a bit of additional work.

src/app/routes/api/v3/forms/__tests__/public-forms.verification.routes.spec.ts
seaerchin: good point! will do! will wait for otp generation to be merged first because this change is dependent on that

PR #1929, opened 2021-05-19
fix(encrypt-submission): shift encrypt mode form guard higher up the pipeline

yong-jie:
## Problem
The guard to prevent submisison of non-storage mode forms in storage mode submission endpoint is located very low in the pipeline, when it should have been one of the first checks made.

Reordering this above will also help reflect the typing of the `form` object accurately, earlier in the pipeline.

## Solution
Copy paste, with changes in wording of the log messaging

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -156,12 +156,12 @@ const submitEncryptModeForm: RequestHandler = async (req, res) => {
   if (!isFormEncryptMode(form)) {
     logger.error({
       message:
-        'Trying to encrypt verified SpCp fields on non-encrypt mode form',
+        'Trying to submit non-encrypt mode submission on encrypt-form submission endpoint',
       meta: logMeta,
     })
     return res.status(StatusCodes.UNPROCESSABLE_ENTITY).json({
       message:
-        'Unable to encrypt verified SPCP fields on non storage mode forms',
+        'Unable to process non-encrypt mode submission on encrypt-form submission endpoint',
     })
   }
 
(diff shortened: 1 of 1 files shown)


mantariksh (commented):
actually might it be better to use `EncryptSubmissionService.checkFormIsEncryptMode`, and use `mapRouteError` to get the status code and error message?

advantages are
1. a step towards keeping all the status codes and error messages in one place
2. easier to chain `Result`s in future if someone wants to clean up this controller, like @seaerchin did for `submitEmailModeForm`

yong-jie:
good idea, i just switched the approach and moved it higher up the pipeline so that the `form` variable has the correct type

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -76,7 +75,22 @@ const submitEncryptModeForm: RequestHandler = async (req, res) => {
...
+      meta: logMeta,
+    })
+    const { statusCode, errorMessage } = mapRouteError(
+      checkFormIsEncryptModeResult.error,
+    )
+    return res.status(statusCode).json({
+      message: errorMessage,
+    })
+  }
+  const form = checkFormIsEncryptModeResult.value
```
seaerchin, line 93: actually i think neverthrow solves this problem of repeated error checking - we can just chain this into the next method and be confident that the form is already encrypted!

then, the `mapErr` will be applied to the result of the whole expression (inclusive) of this! we could do a pass through on the err (using `mapErr`) so that the type is preserved but logs are still present too. 

idk if it's beyond the scope of this PR but yea, just something to think about, especially because lines 70 - 186 is essentially just checking the form

yong-jie: thanks for the review! yup the end goal will definitely to refactor the whole pipeline to use the same map/mapErr pattern as the email-submission endpoint. but that should be left for later after all the upcoming work on the encrypt-submission pipeline is completed

mantariksh (approved):

karrui (approved):
lgtm

PR #1936, opened 2021-05-20
feat: webhook retries (WIP)

mantariksh:
Opening a temporary PR just for sharing.

### DIFF
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -9,6 +9,7 @@ import {
   IEmailSubmissionSchema,
   IEncryptedSubmissionSchema,
   IEncryptSubmissionModel,
+  IPopulatedWebhookSubmission,
   ISubmissionModel,
   ISubmissionSchema,
   IWebhookResponse,
@@ -17,6 +18,7 @@ import {
   SubmissionCursorData,
   SubmissionMetadata,
   SubmissionType,
+  SubmissionWebhookInfo,
   WebhookData,
   WebhookView,
 } from '../../types'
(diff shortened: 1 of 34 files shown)


src/app/models/submission.server.model.ts
```diff
@@ -182,8 +184,11 @@ const EncryptSubmissionSchema = new Schema<
 EncryptSubmissionSchema.methods.getWebhookView = function (
   this: IEncryptedSubmissionSchema,
 ): WebhookView {
+  const formId = this.populated('form')
+    ? String(this.form._id)
+    : String(this.form)
```
liangyuanruo, line 189: why do we need this ternary op?

mantariksh: In the current `develop`, `getWebhookView` is only called in the encrypt submissions pipeline. This means the form has already been retrieved, so there is no need to make an additional network call to retrieve the webhook URL.

For webhook retries, at the point where the message is consumed, the only information we have is the `submissionId`. In order to minimise the network calls required, I implemented `EncryptSubmissionSchema.statics.retrieveWebhookInfoById`, which retrieves all the relevant info for the webhook, including the webhook URL from the form.

However, this means that `getWebhookView` can be called on two types of `EncryptSubmission` documents: unpopulated and populated (with `form.webhook.url` and `form.webhook.isRetryEnabled`). The two types have different shapes; in particular, the formId is `this.form` in the unpopulated document, but `this.form._id` in the populated one. That's how this ternary op came about.

mantariksh:
Closing in favour of smaller incremental PRs

PR #1937, opened 2021-05-20
chore: add a11y checks to test pipeline

karrui:
## Problem

We can currently see a11y compliance via [Storybook's a11y addon](https://storybook.js.org/addons/@storybook/addon-a11y), but the addon itself  will not throw any errors even if there are a11y violations. When multiple developers are working on several components, it is easy to overlook those a11y warnings (or even ignore them).

Thus, this PR adds an a11y check to the frontend tests pipeline to ensure that our rendered components are somewhat accessible, using `@storybook/addon-storyshots-puppeteer` which uses [axe](https://www.deque.com/axe/) under the hood to catch the most common a11y violations.

Also clean up Storybook versions to use the stable version and clean up s ome storybook configuration.

(Also also to experiment with this for the a11y talk I'm about to start writing)

## Solution

**Features**:

- set up a11y testing with storyshots-puppeteer

**Improvements**:

- remove all generated ChakraUI props docs in components stories

### DIFF
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -1,6 +1,7 @@
 {
   "root": true,
   "plugins": ["import", "simple-import-sort", "prettier"],
+  "extends": ["eslint:recommended", "plugin:prettier/recommended"],
   "overrides": [
     {
       "files": ["*.test.ts"],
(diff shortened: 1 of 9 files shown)


mantariksh (approved):
image

Issue #1492, opened 2021-03-30
Automatic webhook retries feature


liangyuanruo:
> This issue is a catch-all for the automatic webhook retries feature

The ability to automatically attempt retries is a useful feature that can greatly reduce the amount of recovery work necessary for downstream webhook consumers. This can generally be implemented using a message queue with a visibility timeout, such as AWS SQS.

## Design document
The design document should address the following:

1. Determine the queue message format, keeping an eye on future extensibility, such as retries with attachments
2. Identify and addressing risks to the system, such as potential for cascade failure
3. Clear communication to users of expectations and conditions under which retries will take place, to be spelled out in guide.form.gov.sg
4. Describe the configuration of both the messaging queue as well as the dead letter queue (DLQ)
5. Key metrics for monitoring for alerts, such as the number of messages in the DLQ
6. Recovery procedures for managing dropped messages

syan-syan:
is this closed? @mantariksh

mantariksh:
@syan-syan yup but there isn't an issue open for the implementation yet, so we can either convert this issue to include implementation or open a new one

karrui:
Mentioned the relevant PRs to this issue for future reference

PR #1939, opened 2021-05-20
feat(webhook-retries): add isRetryEnabled to model (2)

mantariksh:
Second in a series of PRs to enable webhook retries. This PR adds a `webhook.isRetryEnabled` key to the `Form` model, which determines whether retries are enabled for a form. It also makes the necessary changes to ensure that this key can be updated as part of the current form settings update APIs.

### DIFF
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -353,6 +353,10 @@ const compileFormModel = (db: Mongoose): IFormModel => {
               'Webhook must be a valid URL over HTTPS and point to a public IP.',
           },
         },
+        isRetryEnabled: {
+          type: Boolean,
+          default: false,
+        },
       },
 
       msgSrvcName: {
--- a/src/app/routes/api/v3/admin/forms/admin-forms.settings.routes.ts
+++ b/src/app/routes/api/v3/admin/forms/admin-forms.settings.routes.ts
@@ -24,8 +24,9 @@ const updateSettingsValidator = celebrate({
     submissionLimit: Joi.number().allow(null),
     title: Joi.string(),
     webhook: Joi.object({
-      url: Joi.string().uri().required().allow(''),
-    }),
+      url: Joi.string().uri().allow(''),
+      isRetryEnabled: Joi.boolean(),
+    }).min(1),
   }).min(1),
 })
 
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -68,6 +68,7 @@ const FORM_DEFAULTS = {
   permissionList: [],
   webhook: {
     url: '',
+    isRetryEnabled: false,
   },
   status: 'PRIVATE',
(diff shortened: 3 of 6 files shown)


src/app/models/form.server.model.ts
```diff
@@ -353,6 +353,10 @@ const compileFormModel = (db: Mongoose): IFormModel => {
               'Webhook must be a valid URL over HTTPS and point to a public IP.',
           },
         },
+        isRetryEnabled: {
```
karrui, line 356: Worried this turns into the `hasAllowedEmailDomains` and `allowedEmailDomains` thingamagic. Do you foresee us adding some sort of `retryCount` parameter to the model? If not then this is gtg

src/app/routes/api/v3/admin/forms/admin-forms.settings.routes.ts
```diff
@@ -24,8 +24,9 @@ const updateSettingsValidator = celebrate({
     submissionLimit: Joi.number().allow(null),
     title: Joi.string(),
     webhook: Joi.object({
-      url: Joi.string().uri().required().allow(''),
-    }),
+      url: Joi.string().uri().allow(''),
+      isRetryEnabled: Joi.boolean(),
+    }).min(1),
```
karrui, line 29: do we need `min` for webhooks? should it be more specific? no `url` means no `isRetryEnabled`?

EDIT: I see both have defaults, so i don't think `min` is needed

seaerchin: maybe it's for subsequent updates? so that at least one of the two keys are set on subsequent updates if they do want to update webhook config

src/app/models/form.server.model.ts
liangyuanruo: i think it should be fine because we don't intend the feature to be a form-specific parameter

mantariksh: nope, I don't foresee having to add such a thing. I don't see a use case for allowing admins to customise the retry format.

src/app/routes/api/v3/admin/forms/admin-forms.settings.routes.ts
mantariksh: > EDIT: I see both have defaults, so i don't think min is needed

haha but if the frontend sends the update `webhook: {}`, what does that mean? what should be the expected behaviour? and how does mongoose even behave when we do `findByIdAndUpdate(formId, { webhook: {} })`? does that conform to the behaviour expected when the frontend sends this update?

to be honest I don't really want to find out any of the answers, which is why the `.min(1)` is there hahahaha

karrui: yea, requiring at least one sub-key if `webhooks` object is sent to the backend makes sense oki

karrui (approved):
lgtm

karrui:
Related to #1492

PR #1940, opened 2021-05-20
feat(webhook-retries): produce and consume retries (3)

mantariksh:
Third in a series of PRs to enable webhook retries. This PR contains the bulk of the application logic for the production and consumption of messages from the queue.

Can be reviewed by commit.

### DIFF
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -217,18 +217,16 @@ EncryptSubmissionSchema.statics.retrieveWebhookInfoById = function (
   this: IEncryptSubmissionModel,
   submissionId: string,
 ): Promise<SubmissionWebhookInfo | null> {
-  return (this.findById(submissionId)
+  return this.findById(submissionId)
     .populate('form', 'webhook')
-    .exec() as Promise<IPopulatedWebhookSubmission | null>).then(
-    (populatedSubmission) => {
+    .then((populatedSubmission: IPopulatedWebhookSubmission | null) => {
       if (!populatedSubmission) return null
       return {
         webhookUrl: populatedSubmission.form.webhook?.url ?? '',
         isRetryEnabled: !!populatedSubmission.form.webhook?.isRetryEnabled,
         webhookView: populatedSubmission.getWebhookView(),
       }
-    },
-  )
+    })
 }
 
 EncryptSubmissionSchema.statics.findSingleMetadata = function (
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -9,6 +9,7 @@ import {
   IEmailSubmissionSchema,
   IEncryptedSubmissionSchema,
   IEncryptSubmissionModel,
+  IPopulatedWebhookSubmission,
   ISubmissionModel,
   ISubmissionSchema,
   IWebhookResponse,
@@ -17,6 +18,7 @@ import {
   SubmissionCursorData,
   SubmissionMetadata,
   SubmissionType,
+  SubmissionWebhookInfo,
   WebhookData,
   WebhookView,
 } from '../../types'
@@ -182,8 +184,11 @@ const EncryptSubmissionSchema = new Schema<
 EncryptSubmissionSchema.methods.getWebhookView = function (
   this: IEncryptedSubmissionSchema,
 ): WebhookView {
+  const formId = this.populated('form')
+    ? String(this.form._id)
+    : String(this.form)
   const webhookData: WebhookData = {
-    formId: String(this.form),
+    formId,
     submissionId: String(this._id),
     encryptedContent: this.encryptedContent,
     verifiedContent: this.verifiedContent,
@@ -206,6 +211,24 @@ EncryptSubmissionSchema.statics.addWebhookResponse = function (
     { $push: { webhookResponses: webhookResponse } },
     { new: true, setDefaultsOnInsert: true, runValidators: true },
   ).exec()
+}
+
+EncryptSubmissionSchema.statics.retrieveWebhookInfoById = function (
+  this: IEncryptSubmissionModel,
+  submissionId: string,
+): Promise<SubmissionWebhookInfo | null> {
+  return (this.findById(submissionId)
+    .populate('form', 'webhook')
+    .exec() as Promise<IPopulatedWebhookSubmission | null>).then(
+    (populatedSubmission) => {
+      if (!populatedSubmission) return null
+      return {
+        webhookUrl: populatedSubmission.form.webhook?.url ?? '',
+        isRetryEnabled: !!populatedSubmission.form.webhook?.isRetryEnabled,
+        webhookView: populatedSubmission.getWebhookView(),
+      }
+    },
+  )
 }
 
 EncryptSubmissionSchema.statics.findSingleMetadata = function (
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -182,7 +182,7 @@ const EncryptSubmissionSchema = new Schema<
  * which will be posted to the webhook URL.
  */
 EncryptSubmissionSchema.methods.getWebhookView = function (
-  this: IEncryptedSubmissionSchema,
+  this: IEncryptedSubmissionSchema | IPopulatedWebhookSubmission,
 ): WebhookView {
   const formId = this.populated('form')
     ? String(this.form._id)
--- a/src/app/modules/webhook/webhook.consumer.ts
+++ b/src/app/modules/webhook/webhook.consumer.ts
@@ -91,20 +91,19 @@ export const startWebhookConsumer = (
 const createWebhookQueueHandler = (producer: WebhookProducer) => async (
   sqsMessage: aws.SQS.Message,
 ): Promise<void> => {
+  const { Body, MessageId } = sqsMessage
+  const logMeta = {
+    action: 'createWebhookQueueHandler',
+    MessageId,
+  }
   logger.info({
     message: 'Consumed message from webhook queue',
-    meta: {
-      action: 'createWebhookQueueHandler',
-    },
+    meta: logMeta,
   })
-  const { Body } = sqsMessage
   if (!Body) {
     logger.error({
       message: 'Webhook queue message contained undefined body',
-      meta: {
-        action: 'createWebhookQueueHandler',
-        sqsMessage,
-      },
+      meta: logMeta,
     })
     // Malformed message will be retried until redrive policy is exceeded,
     // upon which it will be moved to dead-letter queue
@@ -116,9 +115,7 @@ const createWebhookQueueHandler = (producer: WebhookProducer) => async (
   if (webhookMessageResult.isErr()) {
     logger.error({
       message: 'Webhook queue message could not be parsed',
-      meta: {
-        action: 'createWebhookQueueHandler',
-      },
+      meta: logMeta,
       error: webhookMessageResult.error,
     })
     return Promise.reject()
@@ -129,16 +126,14 @@ const createWebhookQueueHandler = (producer: WebhookProducer) => async (
   if (!webhookMessage.isDue()) {
     logger.info({
       message: 'Webhook not due yet, requeueing',
-      meta: {
-        action: 'createWebhookQueueHandler',
-      },
+      meta: logMeta,
     })
     const requeueResult = await producer.sendMessage(webhookMessage)
     if (requeueResult.isErr()) {
       logger.error({
         message: 'Webhook queue message could not be requeued',
         meta: {
-          action: 'createWebhookQueueHandler',
+          ...logMeta,
           webhookMessage: webhookMessage.prettify(),
         },
         error: requeueResult.error,
--- a/src/app/modules/webhook/webhook.consumer.ts
+++ b/src/app/modules/webhook/webhook.consumer.ts
@@ -227,8 +227,8 @@ const retrieveWebhookInfo = (
 ): ResultAsync<
   SubmissionWebhookInfo,
   SubmissionNotFoundError | PossibleDatabaseError
-> =>
-  ResultAsync.fromPromise(
+> => {
+  return ResultAsync.fromPromise(
     EncryptSubmission.retrieveWebhookInfoById(submissionId),
     (error) => {
       logger.error({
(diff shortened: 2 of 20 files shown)


src/app/models/submission.server.model.ts
```diff
@@ -208,6 +213,24 @@ EncryptSubmissionSchema.statics.addWebhookResponse = function (
   ).exec()
 }
 
+EncryptSubmissionSchema.statics.retrieveWebhookInfoById = function (
+  this: IEncryptSubmissionModel,
+  submissionId: string,
+): Promise<SubmissionWebhookInfo | null> {
+  return (this.findById(submissionId)
+    .populate('form', 'webhook')
+    .exec() as Promise<IPopulatedWebhookSubmission | null>).then(
```
karrui, line 222: nit: should we remove the `exec()` call and move the cast into the returned promise?

```ts
EncryptSubmissionSchema.statics.retrieveWebhookInfoById = function (
  this: IEncryptSubmissionModel,
  submissionId: string,
): Promise<SubmissionWebhookInfo | null> {
  return this.findById(submissionId)
    .populate('form', 'webhook')
    .then((populatedSubmission: IPopulatedWebhookSubmission | null) => {
      if (!populatedSubmission) return null
      return {
        webhookUrl: populatedSubmission.form.webhook?.url ?? '',
        isRetryEnabled: !!populatedSubmission.form.webhook?.isRetryEnabled,
        webhookView: populatedSubmission.getWebhookView(),
      }
    })
}
```

src/app/modules/webhook/webhook.utils.ts
```diff
@@ -14,3 +20,30 @@ export const formatWebhookResponse = (
...
+/**
+ * Computes epoch of next webhook attempt based on previous attempts.
+ * @param previousAttempts Array of epochs of previous attempts
+ * @returns ok(epoch of next attempt) if there are valid retries remaining
+ * @returns err(WebhookNoMoreRetriesError) if there are no more retries remaining
+ */
+export const getNextAttempt = (
+  previousAttempts: number[],
+): Result<number, WebhookNoMoreRetriesError> => {
+  if (previousAttempts.length >= RETRY_INTERVALS.length)
```
karrui, line 34: nit: if return is already on the next line, should we add braces for better visual blocks

src/app/modules/webhook/webhook.message.ts
```diff
@@ -0,0 +1,174 @@
...
+   * retried as yet. This function succeeds as long as
+   * the retry policy allows for at least one retry.
+   * @param submissionId
+   * @returns ok(encapsulated message) if retry policy exists
+   * @returns err if the retry policy does not allow any retries
+   */
+  static fromSubmissionId(
+    submissionId: string,
+  ): Result<WebhookQueueMessage, WebhookNoMoreRetriesError> {
+    return getNextAttempt([]).map(
```
karrui, line 88: should we add the parameter name to make it clearer (i was confused at why an array was given):

```ts
return getNextAttempt(/** previousAttempts= */ [])
```

or should [] be set as default?
or should there be another helper method `getFirstNextAttempt` that just calls `getNextAttempt([])`?

src/app/modules/webhook/webhook.message.ts
```diff
@@ -0,0 +1,174 @@
...
+        }),
+    )
+  }
+
+  /**
+   * Serialises for enqueueing.
+   * @returns Serialised message
+   */
+  serialise(): string {
+    return JSON.stringify(this.message)
```
karrui, line 104: do we want to use our custom stringify-safe instead?

src/app/modules/webhook/webhook.utils.ts
```diff
@@ -14,3 +20,30 @@ export const formatWebhookResponse = (
...
+  return ok(Date.now() + nextAttemptWaitTimeSeconds * 1000)
+}
+
+/**
+ * Converts an epoch to a readable format.
+ * @param epoch
+ * @returns the epoch represented as a readable string
+ */
+export const prettifyEpoch = (epoch: number): string =>
+  moment(epoch).tz('Asia/Singapore').format('D MMM YYYY, h:mm:ssa z')
```
karrui, line 71: offtopic: we should really have a timezone env var or something lol

src/app/utils/random-uniform.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Generates a random integer between min and max (both inclusive).
+ * If min/max are not integers, the ceiling and floor are taken respectively.
+ * @param min
+ * @param max
+ * @returns Integer generated uniformly in interval
+ */
+export const randomUniform = (min: number, max: number): number => {
```
karrui, line 8: nit: rename to `randomUniformInt` for clarity?

src/app/modules/webhook/webhook.consumer.ts
```diff
@@ -0,0 +1,247 @@
...
+      return webhookMessage
+        .incrementAttempts()
+        .asyncAndThen((newMessage) => producer.sendMessage(newMessage))
+    })
+  })
+
+  if (retryResult.isOk()) return Promise.resolve()
+  // Error cases
+  // Special handling for max retries exceeded - log a separate message
+  // and resolve Promise so that message is removed from queue
```
karrui, line 195: Should there also be a handler for `WebhookRetriesNotEnabledError`?

src/app/modules/webhook/webhook.consumer.ts
```diff
@@ -0,0 +1,247 @@
...
+ * @returns ok(webhook information) if database retrieval succeeds
+ * @returns err if submission ID does not exist or database retrieval errors
+ */
+const retrieveWebhookInfo = (
+  submissionId: string,
+): ResultAsync<
+  SubmissionWebhookInfo,
+  SubmissionNotFoundError | PossibleDatabaseError
+> =>
+  ResultAsync.fromPromise(
```
karrui, line 231: nit: wrap in braces and `return` for better visual flow?

karrui (changes requested):

seaerchin (commented):
this is a really well thought out PR and i learnt alot just from reading through it! just have a question wrt how a sqs consumer works

src/app/modules/webhook/webhook.consumer.ts
```diff
@@ -0,0 +1,247 @@
...
+    // Malformed message will be retried until redrive policy is exceeded,
+    // upon which it will be moved to dead-letter queue
+    return Promise.reject()
+  }
+
+  // Parse message
+  const webhookMessageResult = WebhookQueueMessage.deserialise(Body)
+  if (webhookMessageResult.isErr()) {
+    logger.error({
+      message: 'Webhook queue message could not be parsed',
```
liangyuanruo, line 118: SQS should provide a globally unique identifier for every message. Can we include that for every log message, so that there's no ambiguity if we need to track down a message?

src/app/modules/webhook/webhook.utils.ts
liangyuanruo: maybe the day we become FormSEA or FormASEAN

liangyuanruo: on a more serious note, maybe we can hardcode a `const TZ='Asia/Singapore'` variable somewhere

src/app/modules/webhook/webhook.consumer.ts
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * Starts polling a queue for webhook messages.
+ * @param queueUrl URL of queue from which to consume messages
+ * @param producer Producer which can be used to enqueue messages
+ */
+export const startWebhookConsumer = (
+  queueUrl: string,
+  producer: WebhookProducer,
+): void => {
+  const app = Consumer.create({
```
chowyiyin, line 38: Should we be specifying a duration for the `pollingWaitTimeMs` parameter to enable long polling?

src/app/models/submission.server.model.ts
```diff
@@ -182,8 +184,11 @@ const EncryptSubmissionSchema = new Schema<
 EncryptSubmissionSchema.methods.getWebhookView = function (
   this: IEncryptedSubmissionSchema,
 ): WebhookView {
+  const formId = this.populated('form')
```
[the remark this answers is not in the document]
mantariksh: yes!

src/app/modules/webhook/webhook.message.ts
mantariksh: will do, but I actually don't think it's necessary, as we are in control of what we are stringifying here. `stringify-safe` is more for cases where the object we're stringifying comes from an external source.

mantariksh: actually if it's okay with you, I'd rather just stick to `JSON.stringify`. `stringifySafe` can return `undefined` if the argument is `undefined`, and I'd rather not work around that unless there's a good reason to use `stringifySafe`.

src/app/modules/webhook/webhook.consumer.ts
mantariksh: idts, since retries can be re-enabled halfway. wdyt?

src/app/modules/webhook/webhook.consumer.ts
```diff
@@ -0,0 +1,247 @@
...
+        'Webhook consumer encountered error while interacting with queue',
+      meta: {
+        action: 'startWebhookConsumer',
+        message,
+      },
+      error,
+    })
+  })
+
+  app.start()
```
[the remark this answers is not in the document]
mantariksh: if you check out the [implementation](https://github.com/bbc/sqs-consumer/blob/master/src/consumer.ts), this works because the consumer calls a `setTimeout` when we do `app.start()`, and again whenever it polls the queue. so this kind of works like a promise in that the next action has already been scheduled in the event loop when the function returns.

src/app/modules/webhook/webhook.consumer.ts
mantariksh: I don't think so. long polling is a feature of the queue, not of the consumer. we enable long polling on the AWS console by specifying the queue's `ReceiveMessage` wait time, so SQS knows how long to wait before sending a response back to the consumer.

on the other hand, `pollingWaitTimeMs` specifies how long the _consumer_ should wait in between each poll (see [implementation](https://github.com/bbc/sqs-consumer/blob/master/src/consumer.ts)). so e.g. if `ReceiveMessage` (on the **queue**) is 20 seconds and `pollingWaitTimeMs` (on the **consumer**) is 30s, then the following happens:
- Consumer polls the queue
- Assuming queue is empty, queue waits for 20 seconds then sends back an empty response
- Consumer waits for 30 seconds, then polls the queue again

is this consistent with your understanding?

src/app/modules/webhook/webhook.message.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * Determines whether the message is currently due to be sent.
+   * @returns true if webhook is currently due to be sent, false otherwise
+   */
+  isDue(): boolean {
+    // Allow tolerance for clock drift
+    return (
+      Math.abs(differenceInSeconds(Date.now(), this.message.nextAttempt)) <=
+      DUE_TIME_TOLERANCE_SECONDS
```
mantariksh, line 115: TODO: this is a bug. should account for `nextAttempt` to be further in the past than DUE_TIME_TOLERANCE_SECONDS, otherwise such a message will never be due.

src/app/modules/webhook/webhook.consumer.ts
chowyiyin: Oh yes that's right. Is there any reason why using the `ReceiveMessage` wait time is more preferable than the consumer's wait time? The main objective of long polling is to minimise the number of API calls to the queue right? It seems like `pollingWaitTimeMs` allows us to set an arbitrary wait time. If this is so, we can adjust according to an approximate of how crowded the queue is. I think we can get an approximation of the number of messages in the queue by using `getQueueAttributes`, which has a `ApproximateNumberOfMessages` attribute. Will this be useful?

mantariksh: I think `ReceiveMessage` makes more sense because the queue is aware of whether it contains messages, whereas the consumer is not. tbh it seems to me that using a combination of `pollingWaitTimeMs` and `ApproximateNumberOfMessages` is just a workaround to achieve the same effect as `ReceiveMessageWaitTime`.

let's say the queue is empty for a long time. using `ReceiveMessageWaitTime` (let's say 10s) means that the queue will wait for 20s before responding to a poll. using `pollingWaitTimeMs` (let's say 20s) gives roughly the same effect.

but now let's say the queue is populated. `ReceiveMessageWaitTime` doesn't matter here, as the queue will respond immediately to polls as long as it contains messages. using `pollingWaitTimeMs` means the consumer will wait between polls regardless of whether the queue is populated, meaning messages may get bunched up in the queue. using `pollingWaitTimeMs` + `ApproximateNumberOfMessages` allows us to get around this, but with additional complexity as well as an additional API call per poll (to get the approx number of messages), just to achieve the same effect as `ReceiveMessageWaitTime`.

wdyt?

src/app/modules/webhook/webhook.consumer.ts
karrui: I see; yea ok. Maybe add a documentation as to why the other errors are not handled then?

src/app/modules/webhook/webhook.consumer.ts
chowyiyin: Okay that's true! I was thinking more along the lines of checking the approximate number of messages every few minutes  and not before every poll. The main motivation for my suggestion is that the maximum `ReceiveMessageWaitTime` is only 20 seconds if I remember correctly, so using `pollingWaitTimeMs` can potentially help us increase the duration in between calls. But yep I agree with you too that using `ReceiveMessageWaitTime` achieves almost the same effect especially if our queues are seldom empty for long periods of time. Thanks for the discussion :)

mantariksh: sure, I agree that's a good solution in case we ever need to increase the waiting time beyond 20 seconds one day!

src/app/modules/webhook/webhook.consumer.ts
mantariksh: looking at this again and I think you're right that we should add special handling here. if the admin purposefully disabled retries halfway then there's no reason to move the message to the DLQ. thanks for the catch!

karrui (approved):
lgtm

karrui:
Related to #1492

PR #1941, opened 2021-05-20
feat(webhook-retries): enable toggling retries on frontend (4)

mantariksh:
Last in a series of PRs to enable webhook retries. This PR adds a toggle on the frontend to enable or disable retries.

### Webhooks disabled (toggle hidden)
Screenshot 2021-06-07 at 12 42 17 PM

### Webhooks enabled, retries disabled
Screenshot 2021-06-07 at 12 46 47 PM

### Webhooks and retries both enabled
Screenshot 2021-06-07 at 12 46 57 PM

### Process of adding URL, enabling retries, disabling retries, removing URL
webhook-retries-frontend

### DIFF
--- a/src/public/modules/forms/admin/directiveViews/settings-form.client.view.html
+++ b/src/public/modules/forms/admin/directiveViews/settings-form.client.view.html
@@ -430,16 +430,23 @@
           </div>
         </div>
         <div class="settings-input">
+          <div class="alert-custom alert-info alert-margin">
+            <i class="bx bx-error-circle bx-md icon-spacing"></i>
+            <span class="alert-msg">
+              Your system must meet certain requirements before retries can be
+              safely enabled.
+              <a
+                translate-attr="{ href: 'LINKS.WEBHOOK_RETRIES' }"
+                target="_blank"
+                >Learn more</a
+              >
+            </span>
+          </div>
           <div class="row">
             <div class="col-xs-9">
               <span class="label-custom label-medium label-bottom"
                 >Enable retries</span
               >
-              (<a
-                translate-attr="{ href: 'LINKS.WEBHOOK_RETRIES' }"
-                target="_blank"
-                >What are retries?</a
-              >)
             </div>
             <div class="col-xs-3">
               <label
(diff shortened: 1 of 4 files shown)


karrui (approved):
lgtm

src/public/modules/forms/admin/directiveViews/settings-form.client.view.html
```diff
@@ -428,6 +429,39 @@
...
+        <div class="settings-input">
+          <div class="row">
+            <div class="col-xs-9">
+              <span class="label-custom label-medium label-bottom"
+                >Enable retries</span
+              >
+              (<a
+                translate-attr="{ href: 'LINKS.WEBHOOK_RETRIES' }"
+                target="_blank"
+                >What are retries?</a
```
liangyuanruo, line 441: should we use stronger language, given the propensity of users to muck around with webhooks?

liangyuanruo: hopefully we can avoid the situation where a form admin were to happily enable it without considering the requirements on the HTTP status returned or considerations around idempotency

karrui:
Related to #1492

liangyuanruo:
I know you consulted the designers, but should we emphasize more forcefully that admins are to read the guide before enabling the feature?

mantariksh:
in the midst of updating this based on comments from @pearlyong, will push changes and request another review soon!

mantariksh:
@karrui @liangyuanruo requesting re-review after the following changes suggested by @pearlyong:
1. Hide the retries toggle if there is no valid webhook URL, to reduce the total number of info boxes on the page
2. Place "webhook URL cannot be changed if form is public" infobox immediately below webhook URL input
3. Place infobox on receiving system requirements for retries under the "Enable retries toggle"

In addition, I allowed retries to be toggled on/off even when the form is public. imo there are no downsides for us, and this allows form admins to take advantage of this feature without taking down their forms.

karrui (approved):
lgtm

Issue #1768, opened 2021-04-30
Public form endpoint has unnecessary query parameters


mantariksh:
When accessing the new `/api/v3/forms/:formId` endpoint, the frontend automatically adds a `accessMode=adminform` query parameter as this is the default in the FormApi AngularJS service. This does not affect functionality, but is unnecessary technical debt which is easy to tackle.

PR #1947, opened 2021-05-21
refactor: replace $resource in angularjs form-api.client.factory.js with typescript FormService 

### LINKED ISSUE #1835: Convert form-api.client.factory.js to TypeScript

chowyiyin:
## Problem
This PR replaces the `$resource` object in `form-api.client.factory.js` with API call services provided by`FormService.ts`. Due to coupling between the interceptors and `form-fields.client.factory`, interceptors are kept intact in `form-api.client.factory.js`. Previously, the `accessMode` query parameter was also added to all API calls. This migration removes this from any API calls in which it is irrelevant. 

Closes #1835, #1768

## Solution
`getInterceptor` is called in `form-api.client.factory.js` and the relevant interceptor is passed to functions in `FormService.ts`, that handle all API calls.

### Unit Tests
* `FormService.test.ts`: add unit tests for functions in `FormService.ts`
### Manual Tests
- [x] Load forms dashboard page. Check that all available forms are present. (query)
- [x] Click on any form. Should be redirected to edit page. Check that previous fields (if any) are present. (getAdmin)
- [x] Click on `open in new window` in share tab **without** activating form. Should be redirected to error page. (getPublic)
- [x] Click on `open in new window` in share tab after activating form. Should be redirected to public view of form. (getPublic)
- [x] Duplicate a field in a form. Check that the field is duplicated correctly i.e. contents of the field are correct (update)
- [x] Duplicate any form. Ensure duplicate form modal appears and duplicated form appears on the dashboard upon completion (save) 
- [x] Delete a form. Ensure that the form is no longer visible on the dashboard (delete)
- [x] Create a form from scratch in storage mode and email mode. Should be redirected to edit page with no existing fields. Check that form appears on dashboard. (create)
- [x] Select create new form and select any template. Modal with template form in public view should be displayed. (template)
- [x] Select use template (following from previous check). Perform same checks done for create form.
- [x] Select preview mode of any existing form. Should be redirected to preview page. (preview)
- [x] Add a collaborator and change role to owner. Ownership should have changed. (transferOwner)
- [x] Create myinfo fields. Check that myinfo fields are correctly stripped and injected. 
- [x] Duplicate myinfo fields. Check that myinfo fields are correctly stripped and injected.
- [x] Duplicate a form with myinfo fields. Check that myinfo fields are correctly stripped and injected.
- [x] Check that myinfo fields are correctly injected into public form view
- [x] Check that forms can be createdfrom the examples page
- [x] Check that forms can be created from the "Share Template" link under the "Share" tab


liangyuanruo:
@mantariksh @chowyiyin Thanks for working on this - I'm thinking it could help with the React migration to also take a look at organising the different services better - `FormApi`, `FormFields`, `FormService` etc, probably in a separate PR.

src/public/modules/forms/admin/controllers/list-forms.client.controller.js
```diff
@@ -193,9 +195,9 @@ function ListFormsController(
       resolve: {
         FormToDuplicate: () => {
           // Retrieve the form so that we can populate the modal with any existing email recipients
-          return FormApi.preview({
-            formId: vm.myforms[formIndex]._id,
-          }).$promise.then((res) => res.form)
+          return FormApi.preview(vm.myforms[formIndex]._id).then(
```
mantariksh, line 198: does this need a `$q.when`?

src/public/modules/forms/admin/controllers/admin-form.client.controller.js
```diff
@@ -275,8 +275,8 @@ function AdminFormController(
           .catch(handleUpdateError)
       }
       default:
-        return FormApi.update({ formId: $scope.myform._id }, { form: update })
-          .$promise.then((savedForm) => {
+        return FormApi.update($scope.myform._id, { form: update })
```
mantariksh, line 278: I think just wrap all the service calls in `$q.when` to be safe

mantariksh (commented):
This was a tough one to disentangle, great job!

When thinking about the implementation of the services, our first priority is to ensure that they are reusable in React. The point of the frontend services migration is to maximise the amount of code which is battle-tested in production in order to minimise the risk of the final cutover to React.

The FormService is currently doing quite a lot: injecting/stripping MyInfo data, converting POJOs into classes, handling error redirects etc. The issue with this is that it's entirely possible that we will handle all these things completely differently in React. For example, it's unlikely that we will port over the existing viewmodel classes (e.g. `Form.class.js`) into React.

I think a better approach might be to make the service dumb - it takes in the relevant data to call the correct endpoint, and returns the response. That's it. Any additional formatting of the request and response must be done by the caller. Then in the existing AngularJS FormAPI service, we perform all the necessary transformations on the request and response before and after calling the new TypeScript service. This way we can be sure that the new service is reusable - it's completely framework-agnostic as it only makes API calls.

Wdyt? Looping in @karrui and @tshuli too

chowyiyin:
> This was a tough one to disentangle, great job!
Thank you! :)

> When thinking about the implementation of the services, our first priority is to ensure that they are reusable in React. The point of the frontend services migration is to maximise the amount of code which is battle-tested in production in order to minimise the risk of the final cutover to React.
> 
> The FormService is currently doing quite a lot: injecting/stripping MyInfo data, converting POJOs into classes, handling error redirects etc. The issue with this is that it's entirely possible that we will handle all these things completely differently in React. For example, it's unlikely that we will port over the existing viewmodel classes (e.g. `Form.class.js`) into React.
> 
> I think a better approach might be to make the service dumb - it takes in the relevant data to call the correct endpoint, and returns the response. That's it. Any additional formatting of the request and response must be done by the caller. Then in the existing AngularJS FormAPI service, we perform all the necessary transformations on the request and response before and after calling the new TypeScript service. This way we can be sure that the new service is reusable - it's completely framework-agnostic as it only makes API calls.

Do correct me if I am wrong, but I think this service is, in a way, framework agnostic since the things mentioned above (e.g. injecting and stripping MyInfo data) are actually handled in the interceptor? `FormService` simply uses the interceptor that is passed in. The way the interceptor transforms the response and request is up to the caller to decide (in this case, `form-api.client.factory.js`.

mantariksh:
>Do correct me if I am wrong, but I think this service is, in a way, framework agnostic since the things mentioned above (e.g. injecting and stripping MyInfo data) are actually handled in the interceptor?

This is true! But the reasons for needing interceptors are not framework-agnostic, eg the fact that a Form POJO has to be converted into a Form class. This reflects itself in the code too, e.g. in how the `Interceptor` type requires a `responseError`, which is needed purely because of how we handle redirects in AngularJS.

Furthermore, if the service just trusts the interceptors passed to it, then ultimately the caller has to implement the interceptor correctly, meaning there isn't much value in terms of code reuse.

tshuli:
Agree with @mantariksh on scoping the service to calling the correct API and returning the response - I think this is also consistent with how we've written the other services!

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+): Promise<IPopulatedForm> => {
+  const instance = axios.create()
+  return instance
+    .put<IPopulatedForm>(`${formId}/${accessMode}`, update)
+    .then(({ data }) => data)
+}
+
+export const saveForm = async (
+  formId: string,
+  formToSave: FormFieldDto[],
```
mantariksh, line 64: not sure this typing is correct. if you refer to the backend /duplicate endpoint, the type should be `DuplicateFormBody`, which means the argument should be named similarly too

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+    .get<PublicFormViewDto>(`${PUBLIC_FORM_ENDPOINT}/${formId}`, {
+      headers: { 'If-Modified-Since': '0' },
+    })
+    .then(({ data }) => data)
+}
+
+export const updateForm = async (
+  formId: string,
+  update: { form: FormUpdateParams },
+  accessMode = 'adminform',
```
mantariksh, line 54: again this endpoint is form admins only anyway, so there's no need to allow the caller to change the `accessMode`

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+    .then(({ data }) => data)
+}
+
+export const deleteForm = async (formId: string): Promise<void> => {
+  const instance = axios.create()
+  return instance.delete(`${ADMIN_FORM_ENDPOINT}/${formId}`)
+}
+
+// createForm is called without formId, so the endpoint is just /adminform
+export const createForm = async (newForm: {
```
mantariksh, line 81: doesn't seem like there's a need to wrap the argument in an object, this service should just be called directly with the `form` object

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+  form: FormFieldDto[]
```
mantariksh, line 82: don't think this type is correct - check out the `createForm` flow in the backend to see what it should be

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+  const instance = axios.create()
+  return instance
+    .post<IFormSchema>(`${ADMIN_FORM_ENDPOINT}`, newForm)
+    .then(({ data }) => data)
+}
+
+// Used for viewing templates with use-template or examples listing. Any logged in officer is authorized.
+export const queryTemplate = async (
+  formId: string,
+  accessMode = 'adminform',
```
mantariksh, line 93: no need, only admins can get form templates

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+  form: FormFieldDto[]
+}): Promise<IFormSchema> => {
+  const instance = axios.create()
+  return instance
+    .post<IFormSchema>(`${ADMIN_FORM_ENDPOINT}`, newForm)
+    .then(({ data }) => data)
+}
+
+// Used for viewing templates with use-template or examples listing. Any logged in officer is authorized.
+export const queryTemplate = async (
```
mantariksh, line 91: in the jsdocs, could you flag which service functions are used purely by the examples page? this is to make it easier for us to know which functions can be deleted in React, as we don't plan to maintain the Examples page in v3

mantariksh (changes requested):
first round of review, let's address these first before the next round!

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+  IPopulatedForm,
+  PublicForm,
+} from 'src/types'
+import { FormFieldDto } from 'src/types/api'
+
+// endpoints exported for testing
+export const ADMIN_FORM_ENDPOINT = '/api/v3/admin/forms'
+export const PUBLIC_FORM_ENDPOINT = '/api/v3/forms'
+
+export const queryForm = async (): Promise<FormMetaView[]> => {
```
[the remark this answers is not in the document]
chowyiyin: I think I'll just go with `getDashboardView` too. I think this name is okay since the type `FormMetaView` is only specified for forms displayed on the dashboard.

src/public/services/FormService.ts
```diff
@@ -0,0 +1,134 @@
...
+  const instance = axios.create()
+  return instance
+    .get<FormMetaView[]>(`${ADMIN_FORM_ENDPOINT}`, {
+      headers: { 'If-Modified-Since': '0' },
```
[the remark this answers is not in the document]
mantariksh: ok confirmed with yuan, this can be dropped

src/public/modules/forms/config/forms.client.routes.js
```diff
@@ -21,8 +21,9 @@ angular.module('forms').config([
           FormData: [
             'FormApi',
             '$transition$',
-            function (FormApi, $transition$) {
-              return FormApi.getPublic($transition$.params()).$promise
+            '$q',
+            function (FormApi, $transition$, $q) {
+              return $q.when(FormApi.getPublic($transition$.params().formId))
```
mantariksh, line 26: I don't think we need `$q.when` in this file. you can see that the original code returned Promises via `.$promise` - this is because the routes are designed to load when the Promise resolves. can see the `state` documentation [here](https://ui-router.github.io/ng1/docs/1.0.0-beta.1/classes/state.stateprovider.html)

src/public/modules/forms/services/form-api.client.factory.js
```diff
@@ -25,129 +24,142 @@ const extractFormId = (path) => {
...
-  // accessMode is either adminForm or publicForm
-  let resourceUrl = '/:formId/:accessMode'
-  const V3_PUBLICFORM_URL = '/api/v3/forms/:formId'
+  const generateService = (service, errorParams, ...inputs) => {
+    return service(...inputs.map((input) => handleInput(input)))
+      .then((data) => {
+        return handleResponse(data)
+      })
+      .catch((err) => handleError(err, errorParams))
+  }
```
mantariksh, line 93: while I see how this reduces repeated code, I would actually rather call the Myinfo injection/stripping code explicitly in each function and only when it is relevant. this would be more verbose, but would make the pipeline for each API call very explicit according to the shape. rn it's hard for me to tell what transformations are being applied to the data for each API call, and this might also break silently if e.g. we change the param for a service from `{ form: someObject }` to just `someObject`. wdyt?

mantariksh (changes requested):

src/public/services/FormService.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * Transfers ownership of form to another user with the given email.
+ * @param newOwner Object with email of the new owner.
+ * @returns Updated form with new ownership.
+ */
+export const transferOwner = async (
+  formId: string,
+  newOwner: { email: string },
```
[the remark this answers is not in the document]
chowyiyin: I think the backend does expect the the email to be wrapped in an object. Similarly for `updateForm` and actually `createForm` as well.

src/public/services/FormService.ts
```diff
@@ -0,0 +1,170 @@
+import axios from 'axios'
```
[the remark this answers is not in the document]
chowyiyin: Yep sounds good!

src/public/modules/forms/services/form-api.client.factory.js
chowyiyin: Oh yes thats true. Initially I rewrote it as such because this actually mimics the behaviour of the previous request and response interceptors when we were using the `$resource` object. I think it does makes sense to call the injection/stripping code explicitly now since we are refactoring. However, would it be alright if I keep this function and just pass in  the functions `handleInput` and `handleResponse` so that it is specific to each API call? I personally think that this `generateService` function keeps the code neater.

src/public/services/FormService.ts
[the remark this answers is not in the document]
mantariksh: agree, but the caller of the service shouldn't have to worry about the shape of the API right? the service can do the work of wrapping the string in the correct key

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -98,7 +95,10 @@ function CollaboratorModalController(
   $scope.updatePermissionList = (permissionList) => {
```
mantariksh, line 95: could you update the name in the jsdocs for this function too?

src/public/modules/forms/services/form-api.client.factory.js
mantariksh: sure that sounds good

src/public/modules/forms/config/forms.client.routes.js
```diff
@@ -36,14 +36,15 @@ angular.module('forms').config([
           FormData: [
             'FormApi',
             '$transition$',
-            function (FormApi, $transition$) {
-              return FormApi.preview($transition$.params()).$promise.then(
-                (FormData) => {
+            '$q',
+            function (FormApi, $transition$, $q) {
+              return $q
+                .when(FormApi.previewForm($transition$.params().formId))
```
mantariksh, line 42: same comment as previous review - `resolve` expects a Promise, so this doesn't need to be wrapped in `$q.when`

src/public/modules/forms/config/forms.client.routes.js
```diff
@@ -57,13 +58,14 @@ angular.module('forms').config([
           FormData: [
             'FormApi',
             '$transition$',
-            function (FormApi, $transition$) {
-              return FormApi.template($transition$.params()).$promise.then(
-                (FormData) => {
+            '$q',
+            function (FormApi, $transition$, $q) {
+              return $q
+                .when(FormApi.template($transition$.params()))
```
mantariksh, line 64: same here, no need for $q

src/public/modules/users/controllers/examples-card.client.directive.js
```diff
@@ -99,9 +101,9 @@ function examplesCardController(
       controllerAs: 'vm',
       resolve: {
         FormToDuplicate: () => {
-          return FormApi.template({
-            formId: $scope.form._id,
-          }).$promise.then((res) => res.form)
+          return $q
+            .when(FormApi.queryTemplate($scope.form._id))
```
mantariksh, line 105: I don't think you need $q.when here either. the uibmodal [docs](https://angular-ui.github.io/bootstrap/) say that this works the same way as `resolve` in ui-router.

src/public/services/CreateFormService.ts
```diff
@@ -0,0 +1,50 @@
...
+  return axios
+    .post<IFormSchema>(`${ADMIN_FORM_ENDPOINT}`, { form: newForm })
+    .then(({ data }) => data)
+}
+
+/**
+ * Deletes the form with the corresponding formId
+ * @param formId formId of form to delete
+ */
+export const deleteForm = async (
```
mantariksh, line 44: I think this belongs in `UpdateFormService` actually, wdyt? I would think `CreateFormService` relate to actions that can create forms which don't yet exist, whereas deleting is an action to update the state of an existing form

src/public/services/AdminViewFormService.ts
```diff
@@ -0,0 +1,43 @@
+import axios from 'axios'
+
+import { FormMetaView, PublicForm } from 'src/types'
+import { FormViewDto } from 'src/types/api'
```
mantariksh, line 4: use relative imports for consistency

src/public/services/CreateFormService.ts
```diff
@@ -0,0 +1,50 @@
+import axios from 'axios'
+
+import { DuplicateFormBody } from 'src/app/modules/form/admin-form/admin-form.types'
```
mantariksh, line 3: this type should be moved into `src/types/api`. the frontend should never have to import anything from `src/app`. the dependencies should be kept clean such that `app` and `public` import from `shared` and `types` only, and not from each other.

src/public/services/ExamplesService.ts
```diff
@@ -1,12 +1,16 @@
 import axios from 'axios'
 
+import { DuplicateFormBody } from 'src/app/modules/form/admin-form/admin-form.types'
```
mantariksh, line 3: move into `src/types/api`

src/public/services/PublicFormService.ts
```diff
@@ -1,5 +1,7 @@
 import axios from 'axios'
 
+import { PublicFormViewDto } from 'src/app/modules/form/public-form/public-form.types'
```
mantariksh, line 3: move into `src/types/api`

src/public/services/UpdateFormService.ts
```diff
@@ -1,21 +1,25 @@
 import axios from 'axios'
 
-import { FormSettings, LogicDto } from '../../types'
+import { FormUpdateParams } from 'src/app/modules/form/admin-form/admin-form.types'
```
mantariksh, line 3: move into `src/types/api`

src/public/services/CreateFormService.ts
chowyiyin: That makes sense! Initially I wanted the service to be in charge of handling the existence of a form i.e. creating and deleting forms but I couldn't think of a good name for it. I'll shift it into `UpdateFormService` then.

src/public/services/CreateFormService.ts
chowyiyin: Will do this in another PR

mantariksh:
Blocked by #2061

src/public/modules/forms/services/form-api.client.factory.js
```diff
@@ -25,129 +29,170 @@ const extractFormId = (path) => {
...
-        return FormErrorService.redirect({
-          response,
-          targetState: errorTargetState,
-          targetFormId: extractFormId(get(response, 'config.url')),
-        })
-      }
+      FormErrorService.redirect({
+        response: err.response,
+        targetState: errorTargetState,
+        targetFormId: extractFormId(get(err.response, 'config.url')),
```
mantariksh, line 83: I actually think the current implementation of `extractFormId` is wrong because we don't always have the form ID at the start of the path, but that can be fixed separately

mantariksh (commented):
lgtm, really well done! let's test this super thoroughly on staging because it touches nearly every single important API. you can add the following checks to your manual tests:
- myinfo fields are correctly stripped and injected when creating myinfo fields
- myinfo fields are correctly stripped and injected when duplicating myinfo fields
- myinfo fields are correctly stripped and injected when duplicating a form with myinfo fields
- myinfo fields are correctly injected into public form view
- forms can be created from the examples page
- forms can be created from the "Share Template" link under the "Share" tab

mantariksh (approved):
thanks for the thorough testing! I know it's a lot of work but I hope the importance is clear, since this PR touches a lot of core functionality.

PR #1955, opened 2021-05-24
refactor: convert rating.client.service.js to typescript enum

### LINKED ISSUE #1836: Convert form-factory.client.service.js to TypeScript
This service offers a `generateForm` function that calls `FormApi` to perform one of create, save, or use-template.

### LINKED ISSUE #1839: Convert rating.client.service.js to TypeScript enums

chowyiyin:
This PR replaces `rating.client.service.js` with a new Typescript enum `Rating` in `types/form.ts` and the existing `RatingShape` enum in `types/field/ratingField.ts`.

Closes #1839

## Tests
- [ ] Create new rating field. Ensure that all ratings (1-10) and rating shapes (heart and star) are present and check that preview and subsequent form has the correct number of rating options.


src/types/form.ts
```diff
@@ -36,6 +36,19 @@ export enum Colors {
   Grey = 'grey',
 }
 
+export enum Rating {
+  One = 1,
```
mantariksh, line 40: according to the [docs](https://www.typescriptlang.org/docs/handbook/enums.html), if you were to continue using a numeric enum, you could actually just set `One = 1` and the rest of the numbers would auto-increment

src/types/form.ts
```diff
@@ -36,6 +36,19 @@ export enum Colors {
   Grey = 'grey',
 }
 
+export enum Rating {
```
mantariksh, line 39: I'm not sure an enum is the best way to model the data here. let's say one day we decide that ratings can have increments of 0.5 - would we expand the enum with `OnePointFive = 1.5`, `TwoPointFive = 2.5` etc? doesn't seem maintainable.

furthermore, since this is a numeric enum instead of a string enum, `Object.values` actually gives you the wrong thing, as you can see in this playground demo:
Screenshot 2021-05-24 at 12 49 15 PM

this is because numeric enums are two-way, meaning you can do both `Rating.One` to get `1` and `Rating[1]` to get `'One'`.

I think it might be better to just generate and export a list of strings, perhaps using lodash's `range` function.

src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
```diff
@@ -13,7 +13,8 @@ const { uploadImage } = require('../../../../services/FileHandlerService')
 const {
   DateSelectedValidation: DateValidationOptions,
 } = require('../../../../../shared/constants')
-
+const { Rating } = require('../../../../../types/form')
+const { RatingShape } = require('../../../../../types/field/ratingField')
```
mantariksh, line 17: you can actually just import from `types`, since `types` contains an `index.ts` which exports everything

mantariksh (changes requested):

src/types/form.ts
```diff
@@ -1,3 +1,4 @@
+import _ from 'lodash'
```
mantariksh, line 1: prefer importing just the specific function so the frontend doesn't have to load the entire `lodash` library, which is huge. I _think_ this is the syntax:
```suggestion
import range from 'lodash/range'
```

src/types/form.ts
```diff
@@ -41,6 +42,20 @@ export enum ResponseMode {
...
+  Two,
+  Three,
+  Four,
+  Five,
+  Six,
+  Seven,
+  Eight,
+  Nine,
+  Ten,
+}
```
mantariksh, line 58: can delete right?

src/types/form.ts
```diff
@@ -41,6 +42,20 @@ export enum ResponseMode {
   Email = 'email',
 }
 
+export const Rating = _.range(1, 11)
```
mantariksh, line 45: since the original array was an array of strings, should we convert these to strings?

src/types/form.ts
```diff
@@ -41,6 +42,8 @@ export enum ResponseMode {
   Email = 'email',
 }
 
+export const Rating = range(1, 11).map(String)
```
mantariksh, line 45: nice, very functional

mantariksh (approved):

PR #1956, opened 2021-05-24
feat(spcp-session.client.factory): extract public form JWT related fns into non AngularJS dependent files

### LINKED ISSUE #1840: Convert spcp-session.client.factory.js to TypeScript
AngularJS service for user data management when logging into Singpass or Corppass.

karrui:
Ready for review 

## Problem

`spcp-session.client.factory` is responsible for handling the login sessions of SP/CP forms. However, it has a lot of AngularJS-coupled functions that this PR attempts to extract out.

The main piece of responsibility this PR has extracted would be
- `services/PublicFormAuthService.ts`; use `js-cookies` functions to 
  - retrieve SPCP specific cookies depending on auth type of form
  - decode the SPCP JWT depending on auth type of form
  - handle logout depending on auth type of form

Closes #1840 

## Solution

**Breaking Changes** 

  - Purely client side refactor

**Features**:

- feat(cookieUtils): add util functions for manipulating cookies
- feat(PublicFormAuthService): add functions for spcp jwt handling
- ref: use PublicFormAuthService fns in spcp-session.client.factory

**Improvements**:

- feat: update calls to spcp-session functions to include auth type
- fix: add error handling when error occurs when parsing jwt

## Tests

- unit tests have been added for all the new written functions

### Manual tests
- [ ] Log in to SP form
  - [ ] in Chrome dev console should see jwtSp cookie. 
  - [ ] Submit the form. Should submit properly and see SP credentials in submission
  - [ ] Logout. jwtSp cookie should be cleared
  - [ ] Log back into SP, then delete cookie manually. Should refresh and "logout" on the client after some seconds.
- [ ] Log in to CP form
  - [ ] in Chrome dev console should see jwtCp cookie. 
  - [ ] Submit the form. Should submit properly and see CP credentials in submission
  - [ ] Logout. jwtCp cookie should be cleared
  - [ ] Log back into CP, then delete cookie manually. Should refresh and "logout" on the client after some seconds.
- [ ] Log in to MyInfo form should work as per normal since this is not touched
  - [ ] Refresh the form. Should be logged out of MyInfo

**New dependencies**:

- `zod`: shape validation and type guards
- `js-cookies`: cookie manipulation


karrui:
Might actually use `js-cookie` package instead, exploring

karrui:
@mantariksh this PR should also fix the weird logout from SP/CP bug when one logs out from MyInfo.

mantariksh (commented):
re: approach, do you think this is a good time to change the frontend implementation so that it uses the `spcpSession.userName` returned by the backend to render the NRIC, instead of decoding the JWT? advantage is that the frontend doesn't have to worry about decoding the JWT, so this brings us closer to the ideal "render-only" frontend. disadvantage is that the frontend doesn't know `rememberMe` and `issuedAt` without decoding the JWT, but it seems those are only used for some very specific GA event which I'm not sure we need anymore.

karrui:
Closing in favour of removing decoding part entirely, tracked by #2065

PR #1957, opened 2021-05-24
fix(verification): loosen OTP waiting time by 2 seconds

### LINKED ISSUE #1923: Verified field OTP waiting time check is susceptible to clock drift
Server logs contain some instances of "OTP requested before waiting time elapsed" errors. The client UI enforces that OTPs can only be requested every 30 seconds, so this is likely to be a bug.

A possible diagnosis is server clock drift. The servers enforce a 30-second waiting time between requesting new OTPs for verified fields. However, this may be susceptible to clock drift between servers, as follows:

1. Server A is 1 second ahead of Server B.
2. Client requests OTP at 9:00:00am (Server A time). Server A saves the OTP hash to the database, created at 9:00:00am.
3. Client waits for 30 seconds and requests another OTP at 9:00:30am (Server A time).
4. OTP request goes to Server B. Server B reads from the database that the hash was created at 9:00:00am. However, the current time on server B is 9:00:29, so Server B rejects the request as only 29 seconds have elapsed. 

Consider adding a few seconds (maybe 10% = 3 seconds) of tolerance when checking whether the OTP waiting time has elapsed.

mantariksh:
## Problem

Closes #1923, see issue for detailed explanation of problem

## Solution

Add a tolerance of 2 seconds to the OTP waiting time check. This means the server allows OTPs to be requested every 28 seconds.

**Breaking Changes** 

    - Details ...
- [x] No - this PR is backwards compatible, only server-side changes

## Tests

No tests to be updated. Current service tests check that the OTP requests are rejected if made within 25 seconds of each other, which is still true _(pats self on the back)._

## Manual tests
- [ ] Request for an OTP for a verified field. When the 30 second timer on the UI elapses, _immediately_ request a new one. The request should be valid and you should receive the new OTP.

### DIFF
--- a/src/app/modules/verification/verification.util.ts
+++ b/src/app/modules/verification/verification.util.ts
@@ -4,6 +4,7 @@ import {
   HASH_EXPIRE_AFTER_SECONDS,
   VERIFIED_FIELDTYPES,
   WAIT_FOR_OTP_SECONDS,
+  WAIT_FOR_OTP_TOLERANCE_SECONDS,
 } from '../../../shared/util/verification'
 import {
   IFieldSchema,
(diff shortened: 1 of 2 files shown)


chowyiyin (approved):
lgtm

karrui (approved):
lgtm

Issue #1222, opened 2021-02-23
Convert Mongoose "set" hooks to validators


mantariksh:
Some keys in Mongoose models have validators and some have "set" hooks. This makes behaviour inconsistent when documents are invalid. Examples: `authType` in form model and `includeFormSummary` in `emailField`model.

seaerchin:
Another thing to note is that this affects the order of assignment. In `authType` for `form.server.model`, if you set status to be public before assigning authType, this defaults the authType to NIL instead of the expected behaviour where authType is set successfully

PR #1971, opened 2021-05-25
refactor: replace set hook with validator hook in emailField model

chowyiyin:
Refer to issue #1222 for details. This PR only replaces the set hook in `emailField`'s `includeFormSummary`. Two set hooks remain in `form.server.model.ts`. Conversion of these hooks is slightly more complicated due to the use of functions such as `findByIdAndUpdate`, `findOneAndUpdate` amongst others, which write directly to the database. The validator is therefore not able to access the object to perform the needed validation.

## Tests
Ensure existing unit tests still pass. 
### Manual Tests
- [ ] Create email field directly in storage mode via API call by **flipping `includePdfSummary` to `true`**. Ensure that new email field is not created upon refresh.
- [ ] Create email field in email mode. Ensure that including PDF summary is allowed.

### DIFF
--- a/src/app/models/field/emailField.ts
+++ b/src/app/models/field/emailField.ts
@@ -29,9 +29,15 @@ const createEmailFieldSchema = (): Schema<IEmailFieldSchema> => {
       includeFormSummary: {
         type: Boolean,
         default: false,
-        set: function (this: IEmailFieldSchema, v: boolean) {
-          // PDF response not allowed for encrypt forms
-          return this.parent().responseMode === ResponseMode.Encrypt ? false : v
+        validate: {
+          validator: function (this: IEmailFieldSchema) {
+            // PDF response not allowed for encrypt forms but ignore if no autoreply
+            return (
+              this.parent().responseMode !== ResponseMode.Encrypt ||
+              !this.autoReplyOptions.hasAutoReply
+            )
+          },
+          message: 'Autoreply PDF is not allowed for storage mode forms',
         },
       },
     },
--- a/src/app/models/field/emailField.ts
+++ b/src/app/models/field/emailField.ts
@@ -37,7 +37,8 @@ const createEmailFieldSchema = (): Schema<IEmailFieldSchema> => {
               !this.autoReplyOptions.hasAutoReply
             )
           },
-          message: 'Autoreply PDF is not allowed for storage mode forms',
+          message:
+            'PDF response summaries are not allowed for email confirmations in storage mode forms',
         },
       },
     },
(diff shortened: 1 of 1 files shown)


src/app/models/field/emailField.ts
```diff
@@ -29,9 +29,15 @@ const createEmailFieldSchema = (): Schema<IEmailFieldSchema> => {
       includeFormSummary: {
         type: Boolean,
         default: false,
-        set: function (this: IEmailFieldSchema, v: boolean) {
-          // PDF response not allowed for encrypt forms
-          return this.parent().responseMode === ResponseMode.Encrypt ? false : v
+        validate: {
```
mantariksh, line 32: there's a pre-validate hook below in the code which seems to do the same thing. that can be deleted, right?

src/app/models/field/emailField.ts
```diff
@@ -29,9 +29,15 @@ const createEmailFieldSchema = (): Schema<IEmailFieldSchema> => {
...
+          validator: function (this: IEmailFieldSchema) {
+            // PDF response not allowed for encrypt forms but ignore if no autoreply
+            return (
+              this.parent().responseMode !== ResponseMode.Encrypt ||
+              !this.autoReplyOptions.hasAutoReply
+            )
+          },
+          message: 'Autoreply PDF is not allowed for storage mode forms',
```
mantariksh, line 40: we often use "autoreply" to refer to "email confirmation", but this terminology makes no sense to a user. let's stick to "email confirmation" for error messages, especially user-facing ones.

mantariksh (changes requested):

mantariksh (approved):
could you also add in a manual test to ensure that attempting to manually circumvent this validation via API doesn't work?

src/app/models/field/emailField.ts
```diff
@@ -29,9 +29,16 @@ const createEmailFieldSchema = (): Schema<IEmailFieldSchema> => {
...
-        set: function (this: IEmailFieldSchema, v: boolean) {
-          // PDF response not allowed for encrypt forms
-          return this.parent().responseMode === ResponseMode.Encrypt ? false : v
+        validate: {
+          validator: function (this: IEmailFieldSchema, v: boolean) {
+            // PDF response not allowed for encrypt forms but ignore if no autoreply
+            return v
+              ? this.parent().responseMode !== ResponseMode.Encrypt ||
+                  !this.autoReplyOptions.hasAutoReply
+              : true
```
mantariksh, line 38: can we make this a bit clearer by explaining all the cases? e.g.
```
// always ok to set to false
!v ||
// either true or false is okay unless in storage mode
this.parent().responseMode !== ResponseMode.Encrypt
// in storage mode, we can ignore this field if email confirmation is not enabled anyway
!this.autoReplyOptions.hasAutoReply
```

mantariksh (changes requested):

mantariksh (approved):

Issue #1874, opened 2021-05-12
Extract admin mode submissions related functions from submissions.client.factory.js


karrui:
Issue extracted from #1841

See related issues #1873

This is a key client-side service that determines the performance and safety of many key features, but seems to have two separate areas of concern - the public form facing submissions and the admin form retrieval of submissions.

Whilst migrating to Typescript, should also take the opportunity to move the related functions into separate services each with their respective responsibilities.

The other responsibility of the `submissions.client.factory` seems to be related to the admin-facing portion for retrieving form's submissions. This should be extracted out into its own `AdminSubmissionsService` file, that handles

- [ ]  Retrieve submission counts
- [ ] Retrieve encrypt mode responses list
- [ ] Retrieve individual response
- [ ] Download encrypt mode form responses
  - Without attachments
  - With attachments
  - Individual attachment

**Design decisions**
Decided against putting this in the `AdminFormService` as that file should only be dealing with the "form" object and not the "submissions" object relating to a form. This follows the server-side convention.

karrui:
Will be pair programming some parts with chin, otherwise leaving him alone

karrui:
Seems to be done, barring decrypt service worker that requires a webpack rehaul to work with typescript for some reason.

Will close

PR #1983, opened 2021-05-27
refactor(submissions.client.factory): extract admin form features

seaerchin:
## Problem
`submissions.client.factory` client-side service that determines the performance and safety of many key features, but has two separate areas of concern - the public form facing submissions and the admin form retrieval of submissions. 

Part 1 of #1874 

## Solution
This extracts out all the straight forward client submission methods to `AdminFormService`, and replaces old method calls with the new one.

The download related methods will be in a separate PR for closer scrutiny!

**Breaking Changes** 

    - Details ...

## Tests
This adds unit tests for all new methods added.

## Manual Tests 
This has been already tested on staging
Create a storage mode form with any field and activate it. Submit the form >= 11 times. Go to the data section and unlock responses. 

- [ ] There should be the correct number of responses displayed
- [ ] Click on any page that is not 1. The relevant responses should be displayed
- [ ] Select a time period without responses and attempt to download (both options). This should **not** work
- [ ] Select a time period with responses and attempt to download (both options). This should work
- [ ] Search for a response using a valid `referenceId`. this should work and clicking into the response should display all related data. 
- [ ] Search for a response using an invalid/partial `referenceId`. There should be no responses displayed.
- [ ] Clicking into a single response should work and display all data correctly.

### DIFF
--- a/src/public/modules/forms/admin/directives/settings-form.client.directive.js
+++ b/src/public/modules/forms/admin/directives/settings-form.client.directive.js
@@ -1,6 +1,7 @@
 'use strict'
 const dedent = require('dedent-js')
 const { get, set, isEqual } = require('lodash')
+const AdminFormService = require('../../../../services/AdminFormService')
 
 const SETTINGS_PATH = [
   'title',
@@ -28,6 +29,7 @@ const createTempSettings = (myform) => {
 angular
   .module('forms')
   .directive('settingsFormDirective', [
+    '$q',
     'Toastr',
     '$timeout',
     'responseModeEnum',
@@ -38,12 +40,12 @@ angular
   ])
 
 function settingsFormDirective(
+  $q,
   Toastr,
   $timeout,
   responseModeEnum,
   $uibModal,
   Auth,
-  Submissions,
 ) {
   return {
     templateUrl:
@@ -64,11 +66,20 @@ function settingsFormDirective(
         $scope.submissionLimitUnlimited = null
 
         $scope.currentResponsesCount = 0
-        Submissions.count({ formId: $scope.myform._id }).then(
-          (responsesCount) => {
-            $scope.currentResponsesCount = responsesCount
-          },
+        $q.when(
+          AdminFormService.countFormSubmissions({
+            formId: $scope.myform._id,
+          }),
         )
+          .then((responsesCount) => {
+            $scope.currentResponsesCount = responsesCount
+          })
+          .catch((error) => {
+            Toastr.error(
+              `Unfortunately, there was an error in displaying the number of responses. Please refresh and try again!`,
+            )
+            console.log(error)
+          })
 
         const getCurrentSettings = () => {
           // Detect difference between the new form (tempForm) and the old form (myform),
--- a/src/public/modules/forms/admin/directives/settings-form.client.directive.js
+++ b/src/public/modules/forms/admin/directives/settings-form.client.directive.js
@@ -76,7 +76,7 @@ function settingsFormDirective(
           })
           .catch((error) => {
             Toastr.error(
-              `Unfortunately, there was an error in displaying the number of responses. Please refresh and try again!`,
+              `There was an error in displaying the number of responses. Please refresh and try again.`,
             )
             console.log(error)
           })
--- a/src/public/modules/forms/admin/directives/settings-form.client.directive.js
+++ b/src/public/modules/forms/admin/directives/settings-form.client.directive.js
@@ -1,7 +1,7 @@
 'use strict'
 const dedent = require('dedent-js')
 const { get, set, isEqual } = require('lodash')
-const AdminFormService = require('../../../../services/AdminFormService')
+const AdminSubmissionsService = require('../../../../services/AdminSubmissionsService')
 
 const SETTINGS_PATH = [
   'title',
@@ -67,7 +67,7 @@ function settingsFormDirective(
 
         $scope.currentResponsesCount = 0
         $q.when(
-          AdminFormService.countFormSubmissions({
+          AdminSubmissionsService.countFormSubmissions({
             formId: $scope.myform._id,
           }),
         )
--- a/src/public/modules/forms/services/submissions.client.factory.js
+++ b/src/public/modules/forms/services/submissions.client.factory.js
@@ -18,7 +18,6 @@ angular
     '$timeout',
     '$window',
     'GTag',
-    'responseModeEnum',
     'FormSgSdk',
     SubmissionsFactory,
   ])
@@ -31,15 +30,7 @@ function killWorkers(pool) {
   pool.forEach((worker) => worker.terminate())
 }
 
-function SubmissionsFactory(
-  $q,
-  $http,
-  $timeout,
-  $window,
-  GTag,
-  responseModeEnum,
-  FormSgSdk,
-) {
+function SubmissionsFactory($q, $http, $timeout, $window, GTag, FormSgSdk) {
   const ADMIN_FORMS_PREFIX = '/api/v3/admin/forms'
 
   const generateDownloadUrl = (params, downloadAttachments) => {
@@ -78,6 +69,7 @@ function SubmissionsFactory(
       )
       return deferred.promise
     },
+
     getMetadata: function (params) {
       const deferred = $q.defer()
       let resUrl = `${fixParamsToUrl(
--- a/src/public/modules/forms/services/submissions.client.factory.js
+++ b/src/public/modules/forms/services/submissions.client.factory.js
@@ -49,48 +49,6 @@ function SubmissionsFactory($q, $http, $timeout, $window, GTag, FormSgSdk) {
   }
 
   const submissionService = {
-    count: function (params) {
-      const deferred = $q.defer()
-      let resUrl = fixParamsToUrl(
-        params,
-        `${ADMIN_FORMS_PREFIX}/:formId/submissions/count`,
-      )
-      if (params.startDate && params.endDate) {
-        resUrl += `?startDate=${params.startDate}&endDate=${params.endDate}`
-      }
-
-      $http.get(resUrl).then(
-        function (response) {
-          deferred.resolve(response.data)
-        },
-        function () {
-          deferred.reject('Submissions count cannot be obtained.')
-        },
-      )
-      return deferred.promise
-    },
-
-    getMetadata: function (params) {
-      const deferred = $q.defer()
-      let resUrl = `${fixParamsToUrl(
-        params,
-        `${ADMIN_FORMS_PREFIX}/:formId/submissions/metadata`,
-      )}?page=${params.page}`
-
-      if (params.filterBySubmissionRefId) {
-        resUrl += `&submissionId=${params.filterBySubmissionRefId}`
-      }
-
-      $http.get(resUrl).then(
-        function (response) {
-          deferred.resolve(response.data)
-        },
-        function () {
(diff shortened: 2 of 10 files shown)


src/types/api/submission.ts
```diff
@@ -6,3 +8,21 @@ export type SubmissionResponseDto = {
...
+
+export type SubmissionCountQueryDto = {
+  formId: string
+  startDate?: Date
+  endDate?: Date
+}
+
+export type SubmissionMetadataQueryDto = {
+  formId: string
+} & RequireAtLeastOne<{
```
chowyiyin, line 20: Just wanted to check, why don't we use `RequireExactlyOne` instead? Since it seems like we only use either `submissionId` or `pageNum`.

seaerchin: you're right! i set to `RequireAtLeastOne` because the FE callsite passed in both parameters and i didn't want to change it yet so i made it be more permissive. but because our BE handler uses `RequireExactlyOne`, it still works the same way!

src/public/modules/forms/admin/directives/settings-form.client.directive.js
```diff
@@ -64,11 +66,20 @@ function settingsFormDirective(
...
+          }),
         )
+          .then((responsesCount) => {
+            $scope.currentResponsesCount = responsesCount
+          })
+          .catch((error) => {
+            Toastr.error(
+              `Unfortunately, there was an error in displaying the number of responses. Please refresh and try again!`,
+            )
+            console.log(error)
```
karrui, line 81: remove console log?

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -184,209 +118,225 @@ function SubmissionsFactory(
       workerPool = []
       // Creates a new AbortController for every request
       downloadAbortController = new AbortController()
+      const { formId, startDate, endDate } = params
```
karrui, line 121: does this need to be destructured?

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -184,209 +118,225 @@ function SubmissionsFactory(
...
                     }
-                  }
 
-                  checkComplete()
-                })
-            })
+                    checkComplete()
+                  })
+              })
+          })
```
karrui, line 338: assume the whole chunk is untouched

src/public/services/AdminSubmissionsService.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @param submissionId The id of the specified submission to retrieve
+ * @param pageNum The page number of the responses
+ * @returns The metadata of the form
+ */
+export const getFormsMetadata = async ({
+  formId,
+  submissionId,
+  pageNum,
+}: SubmissionMetadataQueryDto): Promise<SubmissionMetadataList> => {
+  const params = submissionId ? { submissionId } : { page: pageNum }
```
karrui, line 45: add comment about how this is mutually exclusive, or split this into two functions so there is no ambiguity?

src/types/api/submission.ts
```diff
@@ -6,3 +8,21 @@ export type SubmissionResponseDto = {
 }
 
 export type SubmissionErrorDto = ErrorDto & { spcpSubmissionFailure?: true }
+
+export type SubmissionCountQueryDto = {
+  formId: string
+  startDate?: Date
```
karrui, line 14: should update the type to require both startDate and endDate or none of them

src/types/api/submission.ts
karrui: like what i said above, can split into two separate functions to remove this ambiguity

src/public/modules/forms/admin/directives/settings-form.client.directive.js
```diff
@@ -64,11 +66,20 @@ function settingsFormDirective(
...
+          AdminSubmissionsService.countFormSubmissions({
+            formId: $scope.myform._id,
+          }),
         )
+          .then((responsesCount) => {
+            $scope.currentResponsesCount = responsesCount
+          })
+          .catch((error) => {
+            Toastr.error(
+              `Unfortunately, there was an error in displaying the number of responses. Please refresh and try again!`,
```
mantariksh, line 79:
```suggestion
              `There was an error in displaying the number of responses. Please refresh and try again.`,
```

src/public/services/__tests__/AdminSubmissionsService.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      // Act
+      const actual = countFormSubmissions({
+        formId: MOCK_FORM_ID,
+        startDate: MOCK_START_DATE,
+        endDate: MOCK_END_DATE,
+      })
+      MockAxios.mockResponse({ data: 123 })
+
+      // Assert
+      await expect(actual).resolves.toEqual(123)
```
mantariksh, line 31: I would much rather structure this as follows:
```suggestion
      // Arrange
      MockAxios.mockResponse({ data: 123 })
      
      // Act
      const actual = await countFormSubmissions({
        formId: MOCK_FORM_ID,
        startDate: MOCK_START_DATE,
        endDate: MOCK_END_DATE,
      })

      // Assert
      expect(actual).toEqual(123)
```

advantages:
1. more understandable flow. you clearly set up the test then call the tested function, rather than jumping back and forth between set up and function calls.
2. takes advantage of the clarity of `async` `await` syntax. no need to take the extra step of understanding "oh so the Promise is created in one place and awaited on in another place".
3. `.resolves` is quite Jest-specific, whereas `expect(actual).toEqual(123)` is valid syntax for many frameworks. it's a small thing but I have a slight preference for framework-agnostic code, which will be helpful if we migrate test frameworks in 2-3 years.

chowyiyin: Oh actually I referred to Chin's code for the tests in my $resource PR so would just like to join this discussion. I think the documentation for [mockResponse](https://github.com/knee-cola/jest-mock-axios#axiosmockresponseresponse-requestinfo-silentmode) requires it to be called after the `axios` call. But maybe we use to `MockAxios.get.mockResolvedValueOnce` instead to make the flow clearer

karrui: MockAxios calls can't be done before (or after) the await... cause it's hijacking the promise lol

mantariksh: that's damn stupid

src/public/modules/forms/admin/directives/settings-form.client.directive.js
seaerchin: i think the console log here has value - in angular, we log errors because we want to be able to see the stack trace. if we remove the console log here, we see generic error messages without context, which isn't helpful for troubleshooting.

src/public/services/__tests__/AdminSubmissionsService.test.ts
seaerchin: we can't actually place it before the call because, as @chowyiyin pointed out, it needs to be after. i think this is a good place to put a comment because it's very specific! i'll try it with  `MockAxios.get.mockResolvedValueOnce` and if that fails, add a comment along with a link to docs explaining this

src/types/api/submission.ts
seaerchin: thanks for pointing this out and suggesting using the XOR type! i'll refactor so it uses XOR!

src/public/services/__tests__/AdminSubmissionsService.test.ts
mantariksh: what I don't understand is why elsewhere in the codebase we stick to the arrange/act/assert pattern with near-religious zeal, but suddenly we're happy to make accommodations here for the `MockAxios` library which doesn't even provide any utility above Jest's native mocking? I don't get it, and I would much rather just use `jest.mock` if `MockAxios` is forcing us to contort the structure of our tests

src/public/modules/forms/services/submissions.client.factory.js
seaerchin: thanks for pointing this out! i preferred this method because i wanted to be explicit about what we were passing into the newly written ts function; however, there were some other functions (eg: `generateDownloadUrl`) that rightfully expect a dict and hence, i did it this way rather than passing in `params` to the ts service

src/public/services/AdminSubmissionsService.ts
seaerchin: good point - will do!

src/types/api/submission.ts
mantariksh: how would you use the XOR type here? XOR is "either one or the other, not both and not none", whereas here you need "either both or none, not one or the other"

this would make more sense to me:
```
{
  formId: string
  dates?: {
    startDate: Date
    endDate: Date
  }
}
```

src/public/services/__tests__/AdminSubmissionsService.test.ts
karrui: Yea, maybe we can switch back to just mocking axios without a third party matcher!

seaerchin: yea, i agree that the structure you proposed is better and i'm going to try (and report back) if it works! i think i should've been more mindful of using MockAxios without considering its utility

seaerchin: yeap it works! changed imports in this file to use remove mockAxios but idt it's a big enough deal to open a new issue for it right? (or we open but yknow p4)

src/public/services/AdminSubmissionsService.ts
```diff
@@ -0,0 +1,84 @@
...
+  return axios.get(queryUrl).then(({ data }) => data)
+}
+
+/**
+ * Retrieves the metadata for a page of submissions
+ * @param formId The id of the form to retrieve submission for
+ * @param pageNum The page number of the responses
+ * @returns The metadata of the page of forms
+ */
+export const getFormsMetadata = async ({
```
karrui, line 39: i know the server is named like that (a regret), but should we name this more explicitly like `getFormsMetadataByPage`, very easy to mix up getFormMetadata and getFormsMetadata

src/public/services/__tests__/AdminSubmissionsService.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      // Arrange
+      MockAxios.get.mockResolvedValueOnce({ data: 123 })
+
+      // Act
+      const actual = countFormSubmissions({
+        formId: MOCK_FORM_ID,
+      })
+
+      // Assert
+      await expect(actual).resolves.toEqual(123)
```
karrui, line 56: like what anti has mentioned, since we don't use mockaxios anymore, we can await in the Act block instead of in the assert block

karrui: update all the tests with the await in the assert block pattern too

karrui (approved):
lgtm, only with last few changes

src/public/services/AdminSubmissionsService.ts
mantariksh: oml yes pls rename this

seaerchin: sure, will rename the other one to `getFormMetadataById` - couldn't think of a good name during that point in time @___@

PR #1986, opened 2021-05-28
feat(auth): support sgID for form submissions

LoneRifle:
## Problem
FormSG is to acquire an additional authentication means for form submitters, ie, sgID. This PR will provide these integrations

## Solution
Fully flesh out sgID integrations with FormSG

[config]
- Add needed configuration to configure sgID authentication, driving
  most of these using env vars

[modules]
- add an sgid module, taking inspiration from spcp
- add sgid-related entries alongside spcp equivalents in public-form
  and email-submission controllers
- do likewise for frontend components

**Breaking Changes** 

**Logistics**
-  This PR has been blocked by #1987; once that PR is merged to develop, this PR must be rebased on the branch

## Screenshots

https://user-images.githubusercontent.com/10572368/119929313-d80f5e80-bfaf-11eb-99c0-459b92baf9cf.mov

## Tests
- Find an sgID developer who has access to a special build of the Singpass App that talks to sgID staging
- Create a betaFlag - `sgid` - for a user
- With this user, create a form with auth type set to Singpass (Lite)
- Make the form public and attempt to log into the form

## Deploy Notes

  - `SGID_ENDPOINT` - The sgID OIDC endpoint
  - `SGID_CLIENT_ID` - The client id registered with sgID
  - `SGID_CLIENT_SECRET` - The client secret registered with sgID
  - `SGID_PRIVATE_KEY` - The path to the private key to decrypt payloads from sgID. Also used for JWT signing
  - `SGID_PUBLIC_KEY` - The path to the public key given to sgID. Also used to verify JWTs created by FormSG
  - `SGID_REDIRECT_URI` - The callback uri that sgID will pass the authorization code and state to
  - `SGID_COOKIE_MAX_AGE` - Max sgID cookie age with remember me unchecked
  - `SGID_COOKIE_MAX_AGE_PRESERVED` - Max sgID cookie age with remember me checked
  - `SGID_COOKIE_DOMAIN` - Domain name set on cookie that holds the sgID jwt

**New dependencies**:

- `@opengovsg/sgid-client` : SDK for sgID based on openid-client


seaerchin (commented):
cool stuff! cos this is a draft pr, not going to focus on code style etc but just wanted to check if my understanding is correct!

as i understand it, we hit the `submit-form-directive` and what happens is that this issues a GET request over to the backend. This isn't rendered because of the `ng-if` (except the `start-page`) component.

next, when we click login, there is a request issued to `/auth/redirect` to get the redirect url and subsequently, redirect there. Next, the sgID client sets the cookie on the requests and routes back. This then triggers a `setUser` call to set the `SpcpSession.userName`, which allows the 
form to be rendered. 

On submission, we verify and check for existence of a username field then attach it to the email response.

this should be roughly what the PR implements right?

LoneRifle:
> cool stuff! cos this is a draft pr, not going to focus on code style etc but just wanted to check if my understanding is correct!
> 
> as i understand it, we hit the `submit-form-directive` and what happens is that this issues a GET request over to the backend. This isn't rendered because of the `ng-if` (except the `start-page`) component.
> 
> next, when we click login, there is a request issued to `/auth/redirect` to get the redirect url and subsequently, redirect there. Next, the sgID client sets the cookie on the requests and routes back. This then triggers a `setUser` call to set the `SpcpSession.userName`, which allows the
> form to be rendered.
> 
> On submission, we verify and check for existence of a username field then attach it to the email response.
> 
> this should be roughly what the PR implements right?

Indeed. This was basically almost a drop-in replacement for Singpass SAML, and a lot of code was re-purposed from the SAML integration.

src/app/modules/sgid/sgid.errors.ts
```diff
@@ -0,0 +1,43 @@
+import { ApplicationError } from '../../modules/core/core.errors'
+
+export class CreateRedirectUrlError extends ApplicationError {
```
liangyuanruo, line 3: suggest prefixing names with `Sgid` or similar for ease of interpretation against other errors in the logs

docker-compose.yml
```diff
@@ -70,6 +70,12 @@ services:
       - CORPPASS_IDP_ID
       - IS_SP_MAINTENANCE
       - IS_CP_MAINTENANCE
+      - SGID_ENDPOINT
+      - SGID_CLIENT_ID
+      - SGID_CLIENT_SECRET
+      - SGID_REDIRECT_URI
+      - SGID_PRIVATE_KEY
+      - SGID_PUBLIC_KEY
```
liangyuanruo, line 78: heads up - I _think_ we are running against the maximum length of 4096 characters for environment variables in Elastic Beanstalk. have you tried deploying to staging?

LoneRifle: we haven't, given that there's still a bunch of setup to undertake. Thanks for notifying!

LoneRifle: ... do we have a viable solution to the extensive configuration that FormSG needs? I wouldn't want to compete with the [Java runtime for number of env vars supported](http://bash.org/?14351)

mantariksh:
Blocked by #1987

src/app/config/schema.ts
```diff
@@ -238,6 +241,62 @@ export const optionalVarsSchema: Schema<IOptionalVarsSchema> = {
       env: 'CUSTOM_CLOUDWATCH_LOG_GROUP',
     },
   },
+  sgid: {
```
mantariksh, line 244: might it be better to put these in a separate file like in `spcp-myinfo.config`? might be easier to keep the slices of config more manageable

mantariksh (commented):
I noticed the implementation depends on the frontend being able to decode the JWT and display the NRIC. we're actually planning on removing that behaviour in the frontend, since we return the NRIC via the `spcpSession` key from the backend anyway. this way we can increase the security of the cookie by setting `httpOnly` to true, since the frontend javascript has no need to read it. would this work for SGID?

src/app/config/schema.ts
LoneRifle: I don't mind, let me rework this

LoneRifle:
> we're actually planning on removing that behaviour in the frontend, since we return the NRIC via the `spcpSession` key from the backend anyway. this way we can increase the security of the cookie by setting `httpOnly` to true, since the frontend javascript has no need to read it. would this work for SGID?

should be okay, this module is patterned after spcp anyway. should I wait for that feature introduction too?

src/app/config/config.ts
```diff
@@ -33,6 +34,12 @@ const compulsoryVars = convict(compulsoryVarsSchema)
   .validate({ allowed: 'strict' })
   .getProperties()
 
+// Load and validate sgid configuration values
+// If environment variables are not present, an error will be thrown
+const sgid = convict(sgidVarsSchema)
```
karrui, line 39: nit: prefer if this instantiation resides in `app/config/features`, following the other features

src/app/config/config.ts
```diff
@@ -206,6 +213,7 @@ const config: Config = {
   db: dbConfig,
   aws: awsConfig,
   mail: mailConfig,
+  sgid,
```
karrui, line 216: if we move the `sgid` instantiation to its own file, then we will not need to add it into the exported `config` constant and `sgid.service` can import directly

src/app/config/sgid.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import { ISgidVarsSchema } from '../../types'
+
+const HOUR_IN_MILLIS = 1000 * 60 * 60
+const DAY_IN_MILLIS = 24 * HOUR_IN_MILLIS
+
+export const sgidVarsSchema: Schema<ISgidVarsSchema> = {
+  endpoint: {
+    doc: 'The sgID OIDC endpoint',
+    format: '*',
```
karrui, line 11: should the format be `url` instead? should we also set the default to `null` for clarity

src/app/config/sgid.ts
```diff
@@ -0,0 +1,63 @@
...
+    default: '',
+    env: 'SGID_ENDPOINT',
+  },
+  clientId: {
+    doc: 'The client id registered with sgID',
+    format: '*',
```
karrui, line 17: should the format be `String` with a `null` default?

src/app/config/sgid.ts
```diff
@@ -0,0 +1,63 @@
...
+    default: '',
+    env: 'SGID_CLIENT_ID',
+  },
+  clientSecret: {
+    doc: 'The client secret registered with sgID',
+    format: '*',
```
karrui, line 23: should the format be `String` with a `null` default?

src/app/config/sgid.ts
```diff
@@ -0,0 +1,63 @@
...
+    default: '',
+    env: 'SGID_CLIENT_SECRET',
+  },
+  privateKey: {
+    doc: 'The path to the private key to decrypt payloads from sgID. Also used for JWT signing',
+    format: '*',
```
karrui, line 29: should the format be `String` with a `null` default?

should we also rename this to `privateKeyPath` for clarity?

src/app/config/sgid.ts
```diff
@@ -0,0 +1,63 @@
...
+    default: '',
+    env: 'SGID_PRIVATE_KEY',
+  },
+  publicKey: {
+    doc: 'The path to the public key given to sgID. Also used to verify JWTs created by FormSG',
+    format: '*',
```
karrui, line 35: should the format be `String` with a `null` default?

should we also rename this to `publicKeyPath` for clarity?

src/app/config/sgid.ts
```diff
@@ -0,0 +1,63 @@
...
+    default: '',
+    env: 'SGID_PUBLIC_KEY',
+  },
+  redirectUri: {
+    doc: 'The callback uri that sgID will pass the authorization code and state to',
+    format: '*',
```
karrui, line 41: should the format be `"url"` with a `null` default?

src/app/loaders/express/index.ts
```diff
@@ -159,6 +160,8 @@ const loadExpressApp = async (connection: Connection) => {
   // Registered routes with the Singpass/Corppass servers
   app.use('/singpass/login', SingpassLoginRouter)
   app.use('/corppass/login', CorppassLoginRouter)
+  // Registered routes with sgID
+  app.use('/sgid', SgidRouter)
```
karrui, line 164: the singpass and corppass routes are stuck as they are due to a hardcoded NDI whitelist.

i don't think this is the case for `/sgid` right? Should we move it to a more semantically correct API endpoint `/api/v3/forms/:formId/auth/sgid` instead? 

the new router can reside in `PublicFormAuthRouter` :P

src/app/modules/sgid/sgid.service.ts
```diff
@@ -0,0 +1,194 @@
...
+      ...sgidOptions,
+      privateKey: fs.readFileSync(privateKey),
+    })
+    this.publicKey = fs.readFileSync(publicKey)
+    this.cookieDomain = cookieDomain
+    this.cookieMaxAge = cookieMaxAge
+    this.cookieMaxAgePreserved = cookieMaxAgePreserved
+  }
+
+  createRedirectUrl(state: string): Result<string, SgidCreateRedirectUrlError> {
```
karrui, line 58: can we make it more explicit on what `state` is? From the call in the controller it seems to be made up of `formId` and `isPersistentLogin`; should we make those arguments of this function and create the state string in this function instead?

There will also be a benefit where the controller does not need to know how to create a state and leave that to this function.

src/app/modules/sgid/__tests__/sgid.controller.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+  MOCK_SGID_FORM,
+  MOCK_SP_FORM,
+  MOCK_STATE,
+  MOCK_TARGET,
+  MOCK_TOKEN_RESULT,
+  MOCK_USER_INFO,
+} from './sgid.test.constants'
+
+jest.mock('../sgid.service')
+const sgidService = mocked(realSgidService, true)
```
karrui, line 35: nit: PascalCase the service name (`SgidService`)?

src/app/modules/sgid/__tests__/sgid.controller.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+const MOCK_LOGIN_REQ = expressHandler.mockRequest({
+  query: { code: MOCK_AUTH_CODE, state: MOCK_STATE },
+})
+
+describe('sgid.controller', () => {
+  beforeEach(() => jest.clearAllMocks())
+  afterAll(() => jest.clearAllMocks())
+
+  describe('handleLogin', () => {
+    const { handleLogin } = SgidController
```
karrui, line 52: nit: should we invoke the full `SgidController.handleLogin` instead of destructuring it here? Allows the reader to know where the function originates amongst all the other functions being called :P

src/app/modules/sgid/__tests__/sgid.routes.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+      expect(FormService.retrieveFullFormById).toHaveBeenCalledWith(MOCK_TARGET)
+      expect(sgidService.token).not.toHaveBeenCalled()
+      expect(sgidService.userInfo).not.toHaveBeenCalled()
+      expect(sgidService.createJWT).not.toHaveBeenCalled()
+      expect(sgidService.getCookieSettings).not.toHaveBeenCalled()
+    })
+
+    it('should set isLoginError cookie and redirect when form has wrong auth type', async () => {
+      FormService.retrieveFullFormById.mockReturnValue(
+        // Note that this is a CorpPass form
```
karrui, line 143: wrong comment? looks like an SP form instead :P

src/app/modules/sgid/__tests__/sgid.service.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+      expect(url._unsafeUnwrap()).toEqual(MOCK_REDIRECT_URL)
+      expect(sgidClient.authorizationUrl).toHaveBeenCalledWith(MOCK_STATE)
+    })
+    it('should return error if not ok', () => {
+      const sgidService = new SgidService(MOCK_OPTIONS)
+      const sgidClient = mocked(MockSgidClient.mock.instances[0], true)
+      sgidClient.authorizationUrl.mockReturnValue({
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        url: undefined,
```
karrui, line 74: 🤔

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -0,0 +1,82 @@
+import { ParamsDictionary, RequestHandler } from 'express-serve-static-core'
```
karrui, line 1: we've removed `express-serve-static-core` from the app, use `ControllerHandler` from `'core/core.types'` instead? :P

src/app/modules/sgid/sgid.service.ts
karrui: can we also have JSDocs for all the exported functions in this file

src/app/modules/sgid/sgid.service.ts
```diff
@@ -0,0 +1,194 @@
...
+    state: string,
+  ): Result<{ formId: string; rememberMe: boolean }, SgidInvalidStateError> {
+    const [formId, rememberMeStr] = state.split(',')
+    const rememberMe = rememberMeStr === 'true'
+    return formId
+      ? ok({ formId, rememberMe })
+      : err(new SgidInvalidStateError())
+  }
+
+  token(
```
karrui, line 86: nit: increase the clarity of what this function does with a more explicit name? Should it be called `retrieveAccessToken` instead?

src/app/modules/sgid/sgid.service.ts
```diff
@@ -0,0 +1,194 @@
...
+          action: 'token',
+          code,
+        },
+        error,
+      })
+      return new SgidFetchAccessTokenError()
+    })
+  }
+
+  userInfo({
```
karrui, line 105: same ^, should we call this `retrieveUserInfo`?

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+      error: jwtResult.error,
+    })
+    res.cookie('isLoginError', true)
+    return res.redirect(`/${formId}`)
+  }
+
+  const { maxAge, jwt } = jwtResult.value
+  res.cookie('jwtSgid', jwt, {
+    maxAge,
+    httpOnly: false, // the JWT needs to be read by client-side JS
```
karrui, line 76: as noted by @mantariksh, @tshuli is working on #2193 which may necessitate some changes on this PR :P FYI

karrui: OR we can ignore it for now and update it as part of #2193 instead. See who is faster :P

src/app/modules/sgid/sgid.util.ts
```diff
@@ -0,0 +1,63 @@
...
+export const validateSgidForm = <T extends IFormSchema>(
+  form: T,
+): Result<SgidForm<T>, AuthTypeMismatchError> => {
+  return isSgidForm(form)
+    ? ok(form)
+    : err(new AuthTypeMismatchError(AuthType.SGID, form.authType))
+}
+
+// Typeguard to ensure that form has eserviceId and correct authType
+const isSgidForm = <F extends IFormSchema>(form: F): form is SgidForm<F> => {
```
karrui, line 28: the typeguard comment seems to insinuate that this checks for the existence of an eserviceId too. Should that be part of the checks in this typeguard?

EDIT: just saw that SGID doesn't have an esrvcId. Update the comment then?

src/app/modules/verified-content/verified-content.service.ts
```diff
@@ -36,6 +36,12 @@ export const getVerifiedContent = ({
       return getSpVerifiedContent(data)
     case AuthType.CP:
       return getCpVerifiedContent(data)
+    case AuthType.SGID:
```
karrui, line 39: iffy on this returning an error. Should we just constrain the type to be back to the previous implementation; i.e. only allow SP/CP auth types? 

Only add in here when SGID is ready to support verified fields?

EDIT: I see that there is a new `SgidUinFin` key being added? Does that not mean SGID has verified content now?

src/public/modules/forms/admin/directives/settings-form.client.directive.js
```diff
@@ -2,6 +2,7 @@
 const dedent = require('dedent-js')
 const { get, set, isEqual } = require('lodash')
 const AdminSubmissionsService = require('../../../../services/AdminSubmissionsService')
+const BetaService = require('../../../../services/BetaService')
```
karrui, line 5: welcome back ol' BetaService 😆

karrui (changes requested):
sorry for the deluge of comments

src/app/modules/verified-content/verified-content.service.ts
LoneRifle: It does... but only for e-mail mode. @liangyuanruo has disallowed sgID for providing verified content in storage mode until we have e2e encryption in-place for fields obtained from eg MyInfo and sgID

src/app/modules/sgid/__tests__/sgid.controller.spec.ts
LoneRifle: It isn't a class definition though, but okay.

src/app/loaders/express/index.ts
LoneRifle: The route was decided by @liangyuanruo . I have already raised the issue of placing the endpoint at the appropriate path.

src/app/modules/sgid/__tests__/sgid.service.spec.ts
LoneRifle: I need this there to induce SgidService to throw up

src/app/config/sgid.ts
LoneRifle: I give up:
```
    'endpoint' uses an unknown format type: url
```

src/app/modules/sgid/sgid.service.ts
```diff
@@ -31,31 +31,41 @@ export class SgidService {
...
+    this.publicKeyPath = fs.readFileSync(publicKeyPath)
     this.cookieDomain = cookieDomain
     this.cookieMaxAge = cookieMaxAge
     this.cookieMaxAgePreserved = cookieMaxAgePreserved
   }
 
-  createRedirectUrl(state: string): Result<string, SgidCreateRedirectUrlError> {
+  /**
+   * Create a URL to sgID which is used to redirect the user for authentication
+   * @param {String} formId - the form id to redirect to after authentication
```
karrui, line 60: nit: don't need types in JSDoc anymore, the Typescript typings should be the source of truth?

src/app/modules/sgid/sgid.service.ts
```diff
@@ -129,6 +158,11 @@ export class SgidService {
     )
   }
 
+  /**
+   * Create a JWT based on the userinfo from sgID
+   * @param {{ 'myinfo.nric_number': string }} data - the userinfo as obtained from sgID
+   * @param {Boolean} rememberMe - determines how long the JWT is valid for
+   */
   createJWT(
```
karrui, line 166: nit: change name to `createJwt` instead to follow the current (unspoken and loosely enforced 😢 ) naming conventions in this repo?

src/app/modules/sgid/sgid.service.ts
```diff
@@ -144,7 +178,7 @@ export class SgidService {
 
   /**
    * Verifies a sgID JWT and extracts its payload.
-   * @param jwtSgid The contents of the JWT cookie
+   * @param {String} jwtSgid The contents of the JWT cookie
    */
   extractJWTInfo(
```
karrui, line 183: nit: here too, change name to `createJwtInfo` instead?

karrui (approved):
lgtm with nits

src/app/config/config.ts
```diff
@@ -14,6 +14,7 @@ import {
   MailConfig,
 } from '../../types'
 
+import { sgid } from './features/sgid.config'
```
mantariksh, line 17: we don't need to export this from the central config right? it can just be imported directly from the SGID config file when needed

src/public/modules/forms/admin/directives/settings-form.client.directive.js
```diff
@@ -143,9 +144,11 @@ function settingsFormDirective(
           )
         }
 
-        $scope.isDisableAuthType = () => {
+        $scope.isDisableAuthType = (authType) => {
           return (
             $scope.isFormPublic() ||
+            (authType.val === 'SGID' &&
+              !BetaService.userHasAccessToFieldType($scope.user, 'sgid')) ||
```
mantariksh, line 151: this implementation looks like if the beta flag is not enabled, the user can see the SGID option but it is disabled for them right? might this be confusing for users who see a new option but don't know why it's disabled for them?

LoneRifle: It's possible, but this may be deliberate; if they send in a support query about it, FormSG can then direct that query to sgID.

src/app/config/config.ts
LoneRifle: This is done to be consistent with the other feature config

src/app/modules/sgid/sgid.service.ts
LoneRifle: We're not creating anything here; we are extracting information out of a JWT

src/app/config/config.ts
mantariksh: no leh, none of the other feature config is exported both from the config file and from the main config. e.g. `spcp-myinfo.config` is not.

mantariksh:
awaiting confirmation on design and copy

LoneRifle:
> awaiting confirmation on design and copy

Confirmation sought. .. I have noticed however that #2193 will be merged soon, and this branch will have to be rebased (again) to bring in-line with norms.

PR #1987, opened 2021-05-28
refactor(encrypt-submission): introduce IncomingEncryptSubmission

yong-jie:
## Problem
This is the first part of a series of refactors to remove the notion of parsed responses.

Historically, we have been injecting metadata and performing operations on the responses array in different stages of the submissions pipeline. This creates a huge coupling to not just the data structure/format but also the exact position of each response in the parsed responses array.

See #1104

## Solution
Create a base IncomingSubmission class that holds at the very least a reference to the form and responses. All other metadata and additional fields should be tracked by this class rather than the `responses` array.

I am opening the PR for the encrypt submission pipeline because the changes will be less intrusive so everyone can review the approach taken here before committing to the same pattern in the email-mode submission pipeline.

Changing the submissions pipeline requires the untangling of many dependencies that were built on top of the the old approach. An example is the `sendEmailConfirmation` function which used to take in parsed responses. This PR also includes the refactoring of this function to demonstrate one way we can go about doing this.

As for other functions that are much more tightly-coupled to a parsedResponse object, I've created a static utility that generates a parsedResponse from a response and some metadata data structures. This is only there in this transition period as we gradually remove the dependency on parsedResponses.

**Breaking Changes** 
    - Details ...

**Features**:

- Nominal typing: Some types are now nominally typed as they are used together with structurally-similar variables. Nominal typing will prevent bugs arising from such easy-to-miss mistakes when such variables together.

**Improvements**:

- sendEmailConfirmation function previously took in parsed responses, but it really didnt need to have the full object inside. I moved out the extract email confirmation data functionality outside of the function and have controllers use it before passing in auto reply data into the new function signature.
- getFilteredResponses: Moved to utils rather than service, as it is a standalone pure function that does not call other services/repositories used purely for `getProcessedResponses` and the new `IncomingSubmission` class.

**Bug Fixes**:

- Typo fix in the admin form controller handler

## Tests

This is a sensitive refactor that modifies the behaviour of the storage mode submissions/preview pipeline. This PR also touches the code responsible for sending email confirmations, so checks need to be made to test that they all function as intended.

Do the following tests **with the network tab open** so you can copy the request body for testing.

- [ ] Duplicate [this form](https://staging.form.gov.sg/60bda996c76a3c0011a314ba/use-template). Submit with all optional fields left empty. No errors should occur.
- [ ] Use the same form and fill everything up. No errors should occur. You should get an auto-reply email.
- [ ] Go to the network tab and copy the request as cURL. paste in terminal, modify the responses field by setting the email field to be a non-email. Should fail and tell you that there's something wrong with the submission
- [ ] Paste the same request you copied earlier into the terminal, modify the responses array by deleting one element. Should fail and tell you that the form has been updated.
- [ ] I'm sorry but you'll have to repeat the above 4 again, but this time in the preview page.
- [ ] Decrypt responses and ensure that none were lost.


src/app/modules/submission/submission.types.ts
```diff
@@ -1,18 +1,37 @@
...
 
 export type ProcessedResponse = {
   question: string
   isVisible?: boolean
   isUserVerified?: boolean
 }
 
+/**
+ * Represents a field map that is guaranteed to contain the id of
+ * ALL responses in an incoming response.
```
chowyiyin, line 21: I think it might be clearer to say "the id of ALL **field** responses in an incoming response"?

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -210,3 +219,104 @@ export const createEncryptedSubmissionDto = (
     attachmentMetadata: attachmentPresignedUrls,
   }
 }
+
+export class IncomingEncryptSubmission extends IncomingSubmission {
+  constructor(
```
chowyiyin, line 224: Would it be good to make this constructor private to make sure that `.init` is called i.e. prevent direct calls to the constructor by accident?

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -210,3 +219,104 @@ export const createEncryptedSubmissionDto = (
...
+      .andThen((filteredResponses) =>
+        this.getFieldMap(form, filteredResponses).map((fieldMap) => ({
+          filteredResponses,
+          fieldMap,
+        })),
+      )
+      .map(({ filteredResponses, fieldMap }) => ({
+        filteredResponses,
+        fieldMap,
+        visibleFieldIds: getVisibleFieldIds(filteredResponses, form),
```
chowyiyin, line 290: Small thing but maybe use `this.getVisibleFieldIds` for consistency?

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -210,3 +219,104 @@ export const createEncryptedSubmissionDto = (
...
+        ),
+        visibleResponseIds: this.getVisibleResponseIds(
+          filteredResponses,
+          fieldMap,
+          (response: FieldResponse) =>
+            'answer' in response &&
+            typeof response.answer === 'string' &&
+            response.answer.trim() !== '',
+        ),
+      }))
```
karrui, line 303: suggest creating this block into a smaller readable function like
```ts
const transformFilteredResponses = ({filteredResponses, form, fieldMap}) => { ... }
```

then the code would have a better visual flow:

```ts
.andThen((filteredResponses) =>
  this.getFieldMap(form, filteredResponses).map((fieldMap) => 
    transformFilteredResponses({filteredResponses, fieldMap, form})
)
```

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -210,3 +219,104 @@ export const createEncryptedSubmissionDto = (
...
+          filteredResponses,
+          fieldMap,
+        ),
+        visibleResponseIds: this.getVisibleResponseIds(
+          filteredResponses,
+          fieldMap,
+          (response: FieldResponse) =>
+            'answer' in response &&
+            typeof response.answer === 'string' &&
+            response.answer.trim() !== '',
```
karrui, line 301: should this be extracted into a helper method?

src/app/modules/submission/submission.utils.ts
```diff
@@ -68,3 +92,237 @@ export const extractEmailConfirmationData = (
...
+  }
+
+  protected static validateResponses(
+    responses: FieldResponse[],
+    form: IFormDocument,
+    fieldMap: ValidatedFieldMap,
+    visibleFieldIds: FieldIdSet,
+    visibleResponseIds: VisibleResponseIdSet,
+    verifiableResponseIds: VerifiableResponseIdSet,
+  ): Result<true, ProcessingError | ValidateFieldError> {
```
karrui, line 299: suggest making this an object instead of positional arguments as discussed on slack

src/app/modules/submission/submission.utils.ts
```diff
@@ -1,11 +1,34 @@
-import { keyBy } from 'lodash'
+import _, { keyBy } from 'lodash'
```
karrui, line 1: for consistency, should we exclusively either use `_` or the destructured function? I'm leaning towards destructured but you can choose :P

src/app/modules/submission/submission.utils.ts
```diff
@@ -68,3 +92,237 @@ export const extractEmailConfirmationData = (
...
+        formId: form._id,
+        onlyInForm,
+      }),
+    )
+  }
+  return ok(results)
+}
+
+export class IncomingSubmission {
+  constructor(
```
karrui, line 149: privatise if not used?

src/app/modules/submission/submission.utils.ts
```diff
@@ -68,3 +92,237 @@ export const extractEmailConfirmationData = (
...
+   * @param fieldMap
+   * @param visibilityPredicate - Predicate that determines if a response
+   * is visible.
+   * @returns visibleResponseIds - Generated set of response IDs.
+   * @protected
+   */
+  protected static getVisibleResponseIds(
+    responses: FieldResponse[],
+    fieldMap: ValidatedFieldMap,
+    visibilityPredicate: (r: FieldResponse) => boolean,
```
karrui, line 214: suggest using object args for readability

src/app/modules/submission/submission.utils.ts
```diff
@@ -68,3 +92,237 @@ export const extractEmailConfirmationData = (
...
+   * @param visibleResponseIds
+   * @param verifiableResponseIds
+   * @returns processedFieldResponse - The ProcessedFieldResponse
+   * @protected
+   */
+  protected static getLegacyProcessedFieldResponse(
+    response: FieldResponse,
+    fieldMap: ValidatedFieldMap,
+    visibleResponseIds: VisibleResponseIdSet,
+    verifiableResponseIds: VerifiableResponseIdSet,
```
karrui, line 270: suggest making this an object instead of positional arguments as discussed on slack

karrui (commented):
initial comments

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -210,3 +219,104 @@ export const createEncryptedSubmissionDto = (
     attachmentMetadata: attachmentPresignedUrls,
   }
 }
+
+export class IncomingEncryptSubmission extends IncomingSubmission {
+  constructor(
+    responses: FieldResponse[],
+    public readonly form: IPopulatedEncryptedForm,
+    public readonly encryptedContent: string,
```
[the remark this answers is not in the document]
yong-jie: It is stored as a variable and [used here](https://github.com/opengovsg/FormSG/blob/adc24ce058ab1e573640bd9826549e50f2778b13/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts#L308). The variable might not look like it's actually being saved, but it's cus I used this [syntactic sugar](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -210,3 +219,104 @@ export const createEncryptedSubmissionDto = (
...
+  ): Result<
+    IncomingEncryptSubmission,
+    ProcessingError | ConflictError | ValidateFieldError
+  > {
+    const logMeta = {
+      action: 'IncomingEncryptSubmission initialisation',
+      formId: form._id,
+    }
+    return checkIsEncryptedEncoding(encryptedContent)
+      .mapErr((error) => {
```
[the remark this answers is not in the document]
yong-jie: 👌 will remove these log statements. The underlying functions themselves dont log, but the caller (submission controllers) do. My original intention of these log statements was to aid debugging so we'd know exactly at which stage did something fail at. 

Then again the error messages would still surface to the top at the controller's `mapErr` and so we should be good without logging anything here

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
yong-jie: thanks for the review! good idea, will make this private

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
yong-jie: putting this as a static helper method. Other child classes that inherit `IncomingSubmission`, like `IncomingEmailSubmission` will also implement its own variant of a visibility predicate, which will complete our sharding TODO in `getProcessedResponses`

src/app/modules/submission/submission.utils.ts
yong-jie: cant make private cus it's needed for child class construction. but marking it protected works, so i'll do that

src/app/modules/submission/submission.utils.ts
```diff
@@ -68,3 +92,253 @@ export const extractEmailConfirmationData = (
...
+      new ConflictError('Some form fields are missing', {
+        formId: form._id,
+        onlyInForm,
+      }),
+    )
+  }
+  return ok(results)
+}
+
+export class IncomingSubmission {
```
karrui, line 148: can this be a separate file too? :X

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
```diff
@@ -0,0 +1,90 @@
...
+    responses: FieldResponse[],
+    form: IPopulatedEncryptedForm,
+  ): Result<
+    FieldResponse[],
+    ProcessingError | ConflictError | ValidateFieldError
+  > {
+    const transformFilteredResponses = ({
+      filteredResponses,
+      fieldMap,
+    }: {
```
karrui, line 53: can this be explicitly typed too

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private static processResponses(
+    responses: FieldResponse[],
+    form: IPopulatedEncryptedForm,
+  ): Result<
+    FieldResponse[],
+    ProcessingError | ConflictError | ValidateFieldError
+  > {
+    const transformFilteredResponses = ({
```
karrui, line 50: can this function be moved out of `processResponses`?

src/app/modules/submission/submission.utils.ts
```diff
@@ -68,3 +92,253 @@ export const extractEmailConfirmationData = (
...
   }, [])
 }
+
+/**
+ * Extracts response data to be sent in email confirmations
+ * @param responses Responses from form filler
+ * @param formFields Fields from form object
+ * @returns Array of data for email confirmations
+ */
+export const extractEmailConfirmationDataFromIncomingSubmission = (
```
karrui, line 102: any reason why this function is not a class method of `IncomingSubmission`? Separate responsibilities?

`incomingSubmission.extractEmailConfirmationData` seems to make sense

karrui (commented):
lgtm with questions

src/app/modules/submission/submission.utils.ts
mantariksh: ^ I had the exact same thought

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
```diff
@@ -0,0 +1,109 @@
...
+  }: {
+    form: IPopulatedEncryptedForm
+    filteredResponses: FieldResponse[]
+    fieldMap: ValidatedFieldMap
+  }): {
+    responses: FieldResponse[]
+    fieldMap: ValidatedFieldMap
+    visibleFieldIds: FieldIdSet
+    verifiableResponseIds: VerifiableResponseIdSet
+    visibleResponseIds: VisibleResponseIdSet
```
mantariksh, line 85: help me understand something - why take the trouble to pass large amounts of data between static functions when we can just have things like `visibleFieldIds` and `verifiableResponseIds` as class members, and declare functions like `transformFilteredResponses` as class members too? that way you get an encapsulated class which can perform processing and validation internally without having to keep track of so many params

src/app/modules/submission/submission.service.ts
```diff
@@ -221,43 +182,39 @@ export const getFormSubmissionsCount = (
...
   attachments,
+  autoReplyData,
 }: {
   form: IPopulatedForm
   submission: S
-  parsedResponses: ProcessedFieldResponse[]
-  autoReplyData?: EmailRespondentConfirmationField[]
+  responsesData?: EmailRespondentConfirmationField[]
   attachments?: IAttachmentInfo[]
+  autoReplyData: AutoReplyMailData[]
```
mantariksh, line 202: the variable naming here has always been quite confusing. would `recipientData: AutoReplyMailData` be more reflective of the purpose of this param?

mantariksh (commented):
I thought the purpose of this refactor was to avoid having to append singpass/corppass/myinfo data at the end of the responses array. will that be done in a separate PR which builds on this?

src/app/modules/submission/submission.utils.ts
yong-jie: you're right, my intent was to minimise the amount of logic related to the representation of data within this data class itself.

Eventually an `IncomingSubmission` will need to emit various projections of its internal data for different purposes. Some would be for the data collation tool etc, and others like this for extracting email data/recipients. Having all of these different representations baked into class methods will work, albeit messily especially as we scale and increase the feature set (like sgid).

the idea i was thinking of was incorporating aspects of the strategy pattern (like in sendEmailConfirmation, which uses either this function or the old`extractEmailConfirmationData` depending on the data structure), and double-dispatch (which can hopefully help to cleanly project data into `SubmissionEmailObject`).

will definitely need your input whether this is a premature optimisation or sth worth keeping/working towards

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
yong-jie: we dont want 1) constructors to throw, and 2) instantiated objects to hold invalid data. This means that we need a separate validation (in this case `init`) method that does the validation for us before instantiating. The problem here is that although it these metadata will be useful as instance fields (and will be saved as such in `IncomingEmailSubmission`), we can't rely on them during the `init` step since instantiation can't happen yet

mantariksh: (1) won't happen anyway due to neverthrow, and doesn't having a private constructor address (2)? `init` can create an object and call methods on it to validate itself, and return an error if the validation fails. so any consumer of the class can only receive valid objects.

imo the advantage of having all these data as class members is that we can call methods on these objects in future.

yong-jie: constructors cant return neverthrow objects, so sadly we still need to rely on an external validation method prior instantiation.

When i picture the flow of data in a scenario that uses these class members for validation, the less-than-ideal events i see are

1) Potentially invalid `IncomingSubmission` instances circulate in memory: Even if we eventually discard and return an error, ideally we don't instantiate them in the first place. i draw similarities to saving an invalid submission to the DB first, and then deleting if it fails validation.

2) Metadata that is used purely during validation (for now) ends up getting tracked for the lifetime of the instance.

I do I agree with the advantage you mentioned though. When we add features to the storage mode submission pipeline these metadata could come in handy.

Given that these metadata will eventually be saved in `IncomingEmailSubmission`, I can always just shift it up to the parent `IncomingSubmission` class so both child classes track metadata.

Let me revise the `init` pipeline to use the instantiate-then-validate pattern and revisit this discussion. maybe the code cleanliness benefit might just outweigh everything

mantariksh: >constructors cant return neverthrow objects, so sadly we still need to rely on an external validation method prior instantiation.

yeah, I think what I meant was that if the constructor is just meant to instantiate class members (eg `verifiableFieldIds`, `visibleFieldIds`etc) then it shouldn't throw anyway

mantariksh: >Potentially invalid IncomingSubmission instances circulate in memory

I'm not sure this is how software works leh. it's not like a virus where things can circulate in an uncontrolled manner right? imo very different from saving invalid documents to the db, where the invalid state is clearly persisted, unlike having an object in memory where it's very clear when it goes out of scope.

yong-jie: it's not a 100% accurate analogy, but it's also a non-ideal behaviour especially if in the future, devs build on the init pipeline and incorrectly assume that all IncomingEncryptSubmission objects are pre-validated

src/app/modules/submission/submission.service.ts
yong-jie: 👌 made the change

src/app/modules/submission/IncomingSubmission.class.ts
```diff
@@ -0,0 +1,219 @@
...
+    })
+
+    const validationResultCombined = combine(validationResultList)
+    if (validationResultCombined.isErr()) {
+      return err(
+        new ProcessingError('Response ID does not match form field IDs'),
+      )
+    }
+
+    return ok(fieldMap as ValidatedFieldMap)
```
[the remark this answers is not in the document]
yong-jie: didnt think of that, thanks!

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
yong-jie: i made changes to the init pipeline, lmk what you think!

src/app/modules/submission/IncomingSubmission.class.ts
```diff
@@ -0,0 +1,199 @@
...
+
+export abstract class IncomingSubmission {
+  private readonly visibleFieldIds: FieldIdSet
+  private readonly visibleResponseIds: VisibleResponseIdSet
+  private readonly verifiableResponseIds: VerifiableResponseIdSet
+  protected constructor(
+    public readonly responses: FieldResponse[],
+    public readonly form: IPopulatedForm,
+    private fieldMap: ValidatedFieldMap,
+  ) {
```
mantariksh, line 32: it would be good to add a note here stating that the subclass should be instantiated using the filtered responses from `getModeFilter`, not the raw responses from the frontend. hopefully that's a temporary thing, I suspect more refactors will fall out once we implement `IncomingEmailSubmission`

src/app/modules/submission/IncomingSubmission.class.ts
```diff
@@ -0,0 +1,199 @@
...
+        this.form._id,
+        formField,
+        this.getLegacyProcessedFieldResponse(response),
+      )
+    })
+
+    const validationResultCombined = combine(validationResultList)
+    if (validationResultCombined.isErr()) {
+      return err(validationResultCombined.error)
+    }
```
mantariksh, line 187: I noticed the behaviour here has changed, since we now validate everything rather than returning early once we find the first invalid answer. since we are doing this anyway, is there value in using `combineWithAllErrors` so we retrieve all the errors?

mantariksh (commented):
imo the instantiate-then-validate code is much more understandable, thank you for incorporating that suggestion!

src/app/modules/submission/IncomingSubmission.class.ts
yong-jie: let me take this one step further and encode this requirement as a nominal type. the compiler will crash when instantiating with unfiltered responses 👍

src/app/modules/submission/submission.types.ts
```diff
@@ -1,18 +1,37 @@
...
+
 import {
   IAttachmentResponse,
   ICheckboxResponse,
   ISingleAnswerResponse,
   ITableResponse,
 } from 'src/types/response'
 
-import { BasicField } from '../../../types/field'
+import { FieldIdSet } from '../../../shared/util/logic'
```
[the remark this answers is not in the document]
yong-jie: you're right, having `any` turned off the typecheck surrounding ids. Changed to string!

src/app/modules/submission/IncomingSubmission.class.ts
```diff
@@ -0,0 +1,199 @@
...
+    const processedResponse: ProcessedFieldResponse = {
+      ...response,
+      isVisible: this.visibleResponseIds.has(responseId),
+      question: formField.getQuestion(),
+    }
+
+    /**
+     * This pattern of mutating processedResponse is copied from the original
+     * getProcessedResponse function. TODO: Check if can just move this into
+     * the object instantiation stage, preserving the possible undefined value.
```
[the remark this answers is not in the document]
yong-jie: thanks for the help with this! changes made

src/app/modules/submission/IncomingSubmission.class.ts
yong-jie: we cant go wrong with being more verbose with our error logging! i just checked that winston logger allows lists of error messages in the `meta` field so this should work out fine

yong-jie: realized that `mapRouteError` doesnt expect errors to be lists. Gonna create a generic error-mapper wrapper that takes the first element of an error list to pass into the error-mapper, with appropriate checks for anomaly cases when the list is empty, etc

mantariksh: okay! it's not a big deal though so you can also just keep it as is haha

src/app/modules/submission/submission.errors.ts
```diff
@@ -44,6 +44,16 @@ export class ValidateFieldError extends ApplicationError {
   }
 }
 
+/**
+ * A custom error class returned when a method explicitly returns a list of errors
+ * but the list itself is empty.
+ */
+export class EmptyErrorFieldError extends ApplicationError {
```
mantariksh, line 51: this seems like a pretty general error type, so you can consider putting it in `core.errors.ts` so it can be shared for similar use cases in future

mantariksh (approved):
lgtm, thank you for completing this odyssey!

src/app/modules/submission/submission.errors.ts
yong-jie: 👌  moved!

PR #1996, opened 2021-05-31
feat: Deduplicate Mongodb Error Messages

### LINKED ISSUE #1694: Joining of Mongo error messages results in walls of text
When a user attempts to edit a form and encounters one or more validation errors, the server combines all the validation error messages into a single comma-separated string and returns it to the client. This normally works fine because a client would normally encounter only one validation error at the time. However, in rare edge cases, they might end up with something like this:

Screenshot 2021-04-21 at 10 17 16 AM

Error messages should be made more readable in case of multiple validation errors.

frankchn:
## Problem
MongoDB validation error messages may be duplicated in specific cases. This deduplicates those error messages.

Closes #1694 

## Solution
Adds a simple de-duplication to the error messages.

### DIFF
--- a/src/app/utils/handle-mongo-error.ts
+++ b/src/app/utils/handle-mongo-error.ts
@@ -47,7 +47,7 @@ export const getMongoErrorMessage = (
     const joinedMessage = dedupMessages.join(', ')
 
     return formatErrorRecoveryMessage(
-      joinedMessage ?? err.message ?? defaultErrorMessage,
+      joinedMessage || err.message ?? defaultErrorMessage,
     )
   }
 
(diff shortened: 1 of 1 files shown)


src/app/utils/handle-mongo-error.ts
```diff
@@ -41,10 +41,10 @@ export const getMongoErrorMessage = (
...
-    const joinedMessage = Object.values(err.errors)
-      .map((err) => err.message)
-      .join(', ')
+    // Deduplicate then Join all error messages into a single message if available.
+    const messages = Object.values(err.errors).map((error) => error.message)
+    const dedupMessages = [...new Set(messages)]
+    const joinedMessage = dedupMessages.join(', ')
 
     return formatErrorRecoveryMessage(
       joinedMessage ?? err.message ?? defaultErrorMessage,
```
liangyuanruo, line 50: If `error.message` is `undefined` then `joinedMessage` could be an empty string?
```suggestion
      joinedMessage || err.message ?? defaultErrorMessage,Write<br class="Apple-interchange-newline">
```

liangyuanruo (approved):
minor suggestion - lgtm otherwise!

PR #2021, opened 2021-05-31
refactor: nominal typing

seaerchin:
**note: this is marked as draft to ensure that the design is sound before fixing up the tests/methods in the controller layer**

## Problem
Currently, our types for email and mobile phone number are `string` - this results in possibly incorrect definitions where we want to use a `Email` type but we pass in a normal `string`, which is not caught by the type system.

a more detailed discussion can be found [here](https://github.com/catamphetamine/libphonenumber-js/issues/170#issuecomment-363156068)!

## Solution
1. use `ts-brand` to get pseudo nominal typing through the use of a brand.
2. create a companion object with the same name as the type so that conversion to/from the type is painless

## Design Considerations
1. `zod` was not used due to its limitations - `z.infer` associates input types with output types of the schema. Hence, even though we validate it, the output is still string, which doesn't help.


src/types/verification_modes.ts
```diff
@@ -0,0 +1,65 @@
+import { identity } from 'lodash'
```
karrui, line 1: i think you are overcomplicating this. why can't we just use type-fest's `Opaque` type instead of a relatively unused and not-updated library?

to create an opaque type, we just need a creator method, e.g.

```ts
const createEmail = (email: string): Email => {
  // Can even add validation here and throw error or something
  return email as Email;
}
```

you really don't need something so complex like assert, etc.

everything else should stick to the same doesn't it? Only functions that absolutely need the correct type to declare it

seaerchin: the reason why we can't use the `Opaque` type is because it's meant as a black box - once we use it, there's no way for us to extract the type out from `Opaque` because it's set as `Type & {readonly __opaque__: Token}` where there isn't a witness to the base type. 

wrt `assert` and `parse` i just wanted to add utility methods for ppl to use AHAH so that the api is sensible and easy to use

karrui: we should not need to extract the type out. any arguments of 'string' will be able to be used since Opaque is a superset.

src/app/services/mail/mail.service.ts
```diff
@@ -434,15 +432,21 @@ export class MailService {
     formData,
   }: {
     replyToEmails?: string[]
```
karrui, line 434: should this also be of Email[] type

PR #2028, opened 2021-05-31
feat: sort responses for csv download according to creation date

### LINKED ISSUE #1721: Investigate: sort responses by date (oldest to newest) before download
Natural order in which responses are sorted on most if not all form managers (many users are confused this isn't the default). Our csv download should sort by date by default instead of requiring form admins to sort as an additional step. This makes it easier to copy paste just the newest responses into another spreadsheet that the admin may already be maintaining or have made other edits / analyses on.

chowyiyin:
## Problem
Currently, when a CSV of responses are downloaded (in storage mode), the responses are not sorted according to date. Refer to issue #1721

Closes #1721

## Solution
The naive approach was adopted to solve this problem. Specifically `Array.prototype.sort()` was used to sort the responses. This solution seems satisfactory even for a large number of responses. When tested with 100,000 responses, the sorting duration ranged from 1 to 2 seconds. With approximately 500,000 responses, the sorting duration ranged from 9 to 10 seconds, which is not significant compared to the decryption duration of approximately 3 minutes. Throughout the download, the following modal is displayed and no obvious freezes or slowdowns in the spinner were observed. The short clip below shows the condition of the spinner before and during sorting. 

https://user-images.githubusercontent.com/54089291/120165348-1023d500-c22e-11eb-8dec-acd5e62a2633.mov

## Tests
- [ ] Submit responses to storage mode form and download CSV. Ensure downloaded CSV is sorted (oldest to newest)

### DIFF
--- a/src/public/modules/forms/helpers/CsvMergedHeadersGenerator.js
+++ b/src/public/modules/forms/helpers/CsvMergedHeadersGenerator.js
@@ -107,21 +107,25 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
 
     // Craft a new csv row for each unprocessed record
     // O(qn), where q = number of unique questions, n = number of submissions.
-    this.unprocessed.forEach((up) => {
-      let createdAt = moment(up.created).tz('Asia/Singapore')
-      createdAt = createdAt.isValid()
-        ? createdAt.format('DD MMM YYYY hh:mm:ss A')
-        : createdAt
-      let row = [up.submissionId, createdAt]
-      for (let [fieldId] of this.fieldIdToQuestion) {
-        const numCols = this.fieldIdToNumCols[fieldId]
-        for (let colIndex = 0; colIndex < numCols; colIndex++) {
-          row.push(this._extractAnswer(up.record, fieldId, colIndex))
+    this.unprocessed
+      .map((up) => {
+        const createdAt = moment(up.created).tz('Asia/Singapore')
+        return { up, createdAt }
+      })
+      .sort((a, b) => this._dateComparator(a.createdAt, b.createdAt)) // responses are first sorted
+      .forEach(({ up, createdAt }) => {
+        createdAt = createdAt.isValid()
+          ? createdAt.format('DD MMM YYYY hh:mm:ss A')
+          : createdAt
+        let row = [up.submissionId, createdAt]
+        for (let [fieldId] of this.fieldIdToQuestion) {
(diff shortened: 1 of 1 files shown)


src/public/modules/forms/helpers/CsvMergedHeadersGenerator.js
```diff
@@ -107,21 +107,25 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
-        : createdAt
-      let row = [up.submissionId, createdAt]
-      for (let [fieldId] of this.fieldIdToQuestion) {
-        const numCols = this.fieldIdToNumCols[fieldId]
-        for (let colIndex = 0; colIndex < numCols; colIndex++) {
-          row.push(this._extractAnswer(up.record, fieldId, colIndex))
+    this.unprocessed
+      .map((up) => {
+        const createdAt = moment(up.created).tz('Asia/Singapore')
+        return { up, createdAt }
```
liangyuanruo, line 113: suggest moving this parsing into the `_dateComparator` instead, so that the __dateComparator_ can compare instances of `a` and `b` directly.

also - what is the type of `up.created`? is it possible that the cast to `Asia/Singapore` is unnecessary?

liangyuanruo (changes requested):
As commented. It might also make sense to decouple the code into a standalone, mutable `sort()` instance method, instead of embedding it in `process()`.

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.js
chowyiyin: I've made the changes! However, I only passed in `up.created` into `_dateComparator` instead of `up`, as I believe `_dateComparator` should only be handling dates. 

`up.created` is a `string` so I think the cast is still necessary.

mantariksh (changes requested):
also to do the following:
- make a more conscious decision about what the upper limit of testing should be, and test on staging with this upper limit. 500,000 is a lot but if we expect that some users may download even more, we should ensure that the implementation is stable.
- test on staging with multiple browsers

liangyuanruo:
> * test on staging with multiple browsers

I think we can get away with testing on localhost, since code changes are client-side

liangyuanruo:
Sure, we can test on GSIB to be safe.

mantariksh:
to populate a test form with 500k submissions and test performance on GSIB

mantariksh (approved):

liangyuanruo (approved):

PR #2040, opened 2021-06-01
chore: remove redundant ValidationOption object properties for short text, long text and number fields

### LINKED ISSUE #408: Redundant ValidationOption object properties for short text, long text and number fields
The `ValidationOption` object defines the required validation rules for the short text and long text fields. An example for Short Text is shown below:

```typescript
enum ShortTextSelectedValidation {
  Max = 'Maximum',
  Min = 'Minimum',
  Exact = 'Exact',
  Range = 'Range',
}

{
  customMax: number | null
  customMin: number | null
  customVal: number | null
  selectedValidation: ShortTextSelectedValidation | null
}
```

The type model is potentially inconsistent for the following reasons:

- Short text and long text fields do not offer range validation, so the type system is not reflective of application functionality today.
- The key `selectedValidation` defines the validation rule being applied, while the `customVal` property defines the number of characters to validate against. This makes the `customMin` and `customMax` keys redundant and runs the risk of inconsistent application state.

TODOs:
1. Verify that the database is internally consistent today for all short text and long text fields:
  a. `customVal === customMin` when `selectedValidation === 'Minimum'`
  b. `customVal === customMax` when `selectedValidation === 'Maximum'`
  c. `customVal === customMin === customMax` when `selectedValidation === 'Exact'`
  d. `selectedValidation === 'Range'` should not exist
2. Simplify the codebase by removing references to `customMin`, `customMax` and `Range`
3. Write a database migration script and/or mongoose hooks to fix inconsistencies discovered

tshuli:
## Problem
Closes #408

## Solution

- Modify codebase (both frontend and backend) to use only `customVal` for text and number field validation, and remove references to `Range`
- For backward compatibility, use mongoose virtuals to return `customMin` and `customMax` properties using the value of `customVal` when reading the form document so that old frontend clients can continue to function normally. Existing frontend code already keeps the value of `customVal` consistent with `customMin` and `customMax` so there is no need to add `pre save` hooks to ensure consistent state when saving the form document using an old client. 

**Breaking Changes** 
- [x] Yes - this PR contains breaking changes. Change in schema for number and text fields.

## Tests
- [ ] Create a form with the following fields
  - Short text, with minimum 3 chars
  - Short text, with exactly 3 chars
  - Long text, with minimum 3 chars
  - Long text, with exactly 3 chars
  - Number, with minimum 3 chars
  - Number, with exactly 3 chars
- [ ] Test that the validation works correctly and the form can be submitted
- [ ] Edit the validation by replacing minimum with maximum. Check that the validation works correctly and form can be submitted.

**Backward compatibility checks**
- [ ] Repeat the above tests but using an old client for the public form user. Check that the validation works correctly and form can be submitted.
- [ ] Repeat the above tests but using an old client for the admin. Check that the form can be modified correctly and that the modified changes persist.

## Deploy Notes

**New scripts**:

New script added at `scripts/20210517_validation-object/script.js`

Part 1 of the script checks that database state is currently consistent - should be run before this PR is approved
- Tested against prod on 8 Jun 2021 and database state is found to be consistent

Part 2 of the script unsets customMin and customMax for text and number fields - should be run after code is deployed to production
- Tested on staging on 8 Jun 2021


scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,283 @@
...
+    $project: {
+      form_fields: 1,
+      selectedValidation: '$form_fields.ValidationOptions.selectedValidation',
+    },
+  },
+  {
+    $match: {
+      selectedValidation: 'Range',
+    },
+  },
```
mantariksh, line 136: I have a feeling we can skip all the previous steps and just do `{ $match: { 'form_fields.ValidationOptions.selectedValidation': 'Range' } }`, since we shouldn't have such a key regardless of fieldType

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,283 @@
...
+// Scripts for #408
+
+/**
+ * Part 1: Verify that the database is internally consistent today
+ * for all short text, long text and number fields.
+ * Expect 0 records for all scripts
+ */
+
+// a. customVal === customMin when selectedValidation === 'Minimum'
+// Expect 0 records
```
mantariksh, line 11: do we need a fallback plan in case any of these are not 0? just tryna plan ahead so we don't get stuck when running this db migration on a weekend night. maybe we should decide what the source of truth should be for each case (short text, long text, number) - if it's always `customVal` then actually we don't even need to verify consistency, but if it's `customMin` or `customMax` then maybe we need to plan for the case where they're not equal

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,283 @@
...
+/**
+ * Part 2: Unset form_fields.ValidationOptions.customMin and
+ * form_fields.ValidationOptions.customMax
+ * for all short text, long text and number fields.
+ */
+
+// ********** BEFORE **********
+
+// Count number of forms with (short text, long text and number fields) and
+// with ValidationOptions.customMin
```
mantariksh, line 148: maybe we should do a count for `ValidationOptions.customMin` and `customMax` for all other field types as well, as a sanity check to make sure that we haven't affected this key in decimal and checkbox fields

src/app/models/field/numberField.ts
```diff
@@ -1,21 +1,19 @@
...
-      customMax: {
-        type: Number,
-        default: null,
-      },
-      customMin: {
-        type: Number,
-        default: null,
-      },
+  const ValidationOptionsSchema = new Schema<
+    WithCustomMinMax<NumberValidationOptions>
```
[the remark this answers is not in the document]
tshuli: thanks! I added a todo for the type itself at `src/types/field/utils/virtuals.ts`, so I think it is ok to leave out here because once we delete the type, we would remove references to the type as well

scripts/20210517_validation-object/script.js
tshuli: I don't think this will happen - it is only legacy forms that we potentially need to worry about. Anyway this is something which should be done before the deployment itself - i'll test on prod secondary and update the status above

scripts/20210517_validation-object/script.js
tshuli: that is true with an `elemMatch`... but I guess I will keep the current implementation because it's more 'precise' :)

src/app/utils/field-validation/validators/textValidator.ts
```diff
@@ -20,8 +20,8 @@ type TextFieldValidatorConstructor = (
 const minLengthValidator: TextFieldValidatorConstructor = (textField) => (
   response,
 ) => {
-  const { customMin } = textField.ValidationOptions
-  const min = customMin !== null ? Number(customMin) : null
+  const { customVal } = textField.ValidationOptions
+  const min = customVal !== null ? Number(customVal) : null
```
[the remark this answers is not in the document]
tshuli: thanks! have edited

src/app/models/field/numberField.ts
```diff
@@ -26,6 +24,39 @@ const createNumberFieldSchema = () => {
         default: null,
       },
     },
+    {
+      toJSON: {
+        virtuals: true,
+      },
+    },
```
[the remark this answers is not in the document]
tshuli: added

src/app/models/field/numberField.ts
```diff
@@ -26,6 +24,39 @@ const createNumberFieldSchema = () => {
...
+  )
+
+  // Virtuals to allow for backwards compatibility after customMin and customMax were removed as part of #408
+  // TODO: Remove virtuals (#2039)
+  ValidationOptionsSchema.virtual('customMin').get(function (
+    this: WithCustomMinMax<NumberValidationOptions>,
```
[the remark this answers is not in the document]
tshuli: thanks, have edited

src/app/models/field/shortTextField.ts
```diff
@@ -1,30 +1,56 @@
...
-        type: Number,
-        default: null,
-      },
-      customMin: {
-        type: Number,
-        default: null,
-      },
+  const ValidationOptionsSchema = new Schema<
+    WithCustomMinMax<ShortTextValidationOptions>
+  >(
```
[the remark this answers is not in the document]
tshuli: extracted

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,283 @@
...
+// a. customVal === customMin when selectedValidation === 'Minimum'
+// Expect 0 records
+
+db.getCollection('forms').aggregate([
+  { $unwind: '$form_fields' },
+  {
+    $match: {
+      'form_fields.fieldType': { $in: ['textfield', 'textarea', 'number'] },
+    },
+  },
```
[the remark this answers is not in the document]
tshuli: cool, thanks! have added in. then it would be an `$elemMatch` :)

scripts/20210517_validation-object/script.js
tshuli: thanks! i've added in

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,394 @@
...
+ * Expect 0 records for all scripts
+ */
+
+// a. customVal === customMin when selectedValidation === 'Minimum'
+// Expect 0 records
+
+db.getCollection('forms').aggregate([
+  {
+    $match: {
+      'form_fields': {$elemMatch: {'fieldType': { $in: ['textfield', 'textarea', 'number'] }}},
```
mantariksh, line 16: I actually don't think you need `$elemMatch` here. from the [docs](https://docs.mongodb.com/manual/reference/operator/query/elemMatch/#std-label-elemmatch-single-query-condition): 
> If you specify a single query predicate in the $elemMatch expression, and are not using the $not or $ne operators inside of $elemMatch, $elemMatch can be omitted.

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,394 @@
...
+    },
+  },
+  {
+    $match: {
+      $and: [
+        {
+          selectedValidation: 'Minimum',
+        },
+        {
+          $expr: { $ne: ['$customMin', '$customVal'] },
```
mantariksh, line 41: it would be good to sanity-check this locally by deliberately inserting a field where these are not equal and ensuring that the query finds the correct result

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,394 @@
...
+  {
+    $match: {
+      $and: [
+        {
+          selectedValidation: 'Exact',
+        },
+        {
+          $or: [
+            {
+              $expr: { $ne: ['$customMax', '$customVal'] },
```
mantariksh, line 119: same here re: sanity check

scripts/20210517_validation-object/script.js
mantariksh: okay sure! I don't think you need elemMatch here either though

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,394 @@
...
+// ********** BEFORE **********
+
+// Count number of forms with (short text, long text and number fields) and
+// with ValidationOptions.customMin
+// Should match number of updated records later
+
+db.forms.count({
+  form_fields: {
+    $elemMatch: {
+      $and: [
```
mantariksh, line 174: I don't think you need the $and when using $elemMatch

scripts/20210517_validation-object/script.js
```diff
@@ -0,0 +1,394 @@
...
+
+// ********** UPDATE **********
+
+// Remove customMin
+
+db.forms.update(
+  {
+    form_fields: {
+      $elemMatch: {
+        $and: [
```
mantariksh, line 243: don't think you need $and with $elemMatch

src/app/models/field/util/textValidationOptionsSchema.ts
```diff
@@ -0,0 +1,39 @@
+import { Schema } from 'mongoose'
```
mantariksh, line 1: this isn't a util since it contains a schema definition. perhaps `common` might be better for the folder name?

mantariksh (approved):
a few nits on query syntax and naming but otherwise lgtm. thanks for the cool implementation! though since this is our first usage of mongoose virtuals, we should test super thoroughly on staging (eg check that the network call contains customMin and customMax with old client retrieving a field which has had customMin and customMax removed)

PR #2045, opened 2021-06-01
refactor(ts-migration): migrates process-decrypted-content to ts

seaerchin:
**note: this PR is part 1 of extracting downloadEncryptedResponse**

## Problem
previously, this file was written in javascript and lacked type safety. this made it hard to assert the results of calling functions in this file and led us to many possible errors, such as when `null` isn't explicitly handled or when the shape of a result isn't known.

## Solution
This PR ports over the file and associated functions to typescript. This PR also does some minor refactoring so that the functions in the file are written more cleanly. 

**Breaking Changes** 

    - Details ...


src/public/modules/forms/helpers/process-decrypted-content.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+    default:
+      return null
+  }
+}
+
+const isVerifiedFormField = (
+  possibleFormField: unknown,
+): possibleFormField is VerifiedFormField => {
+  return !!possibleFormField && !!(possibleFormField as VerifiedFormField)._id
```
karrui, line 56: may want to use lodash's `get` just in case we are unable to destructure, instead of casting.

```ts
return !!possibleFormField && !!get(possibleFormField, _id)
```

can we also do stronger checks for what exactly constitutes a verified form field?

I see the typing in the sdk is

```ts

    _id: string;
    question: string;
    fieldType: FieldType;
    isHeader?: boolean;
    signature?: string;
} & ({
    answer: string;
    answerArray?: never;
} | {
    answer?: never;
    answerArray: string[] | string[][];
});
```

can i also query on why we need this typeguard? if we are using SDK types, should we not be assured that if the decrypted response is not null, then it will be a "VerifiedFormField"?

seaerchin: we need the typeguard because we are trying to transform the type from `VerifiedField | null -> VerifiedField`. Oddly enough, ts can't infer the type if i do `.filter(field => !!field)` and complains that `Signature '(field: FormField | null): boolean' must be a type predicate`

wrt the checks, you're right in that if the response is non-null then it'll be a `VerifiedFormField`; the reason why i did it as such as cos i set the initial type to unknown but i think it's a premature optimization and i'll just type it according to the sdk HAHAH

mantariksh (approved):
mglt

src/public/modules/forms/helpers/process-decrypted-content.ts
```diff
@@ -0,0 +1,95 @@
...
+const convertToResponseArray = (
+  verifiedObj: Record<string, string>,
+): VerifiedFormField[] => {
+  return CURRENT_VERIFIED_FIELDS.filter((fieldType) =>
+    has(verifiedObj, fieldType),
+  )
+    .map((fieldType) =>
+      getResponseFromVerifiedField(fieldType, verifiedObj[fieldType]),
+    )
+    .filter<VerifiedFormField>(isVerifiedFormField)
```
tshuli, line 76: how about defining the typeguard inline instead of in a separate function? looks neater

```suggestion
    .filter((field): field is VerifiedFormField => {
      return !!field
    })
```

src/public/modules/forms/helpers/process-decrypted-content.ts
```diff
@@ -0,0 +1,95 @@
...
+    default:
+      return null
+  }
+}
+
+const isVerifiedFormField = (
+  possibleFormField: VerifiedFormField | null,
+): possibleFormField is VerifiedFormField => {
+  return !!possibleFormField
+}
```
tshuli, line 57: then no need for this (see comments on typeguard in L76)

tshuli (approved):
lgtm except for a nit on typeguard

liangyuanruo (commented):
Might it be cleaner to define a utility function for using type guards like this? Suggesting this because I find using an anonymous function slightly misleading. It's also possible that typefest or some other library already has such a utility function implemented.

```ts
function isNotNull<T>(argument: T | null): argument is T {
    return argument !== null
}
```

PR #2046, opened 2021-06-01
fix(deps): update mongoose to 5.12.12, update model types

karrui:
## Problem

This PR updates `mongoose` to 5.12.12 and updates the types of our models due to having better mongoose types available now.

### DIFF
--- a/src/types/submission.ts
+++ b/src/types/submission.ts
@@ -83,7 +83,7 @@ export interface IEmailSubmission extends ISubmission {
   getWebhookView(): null
 }
 
-export type IEmailSubmissionSchema = IEmailSubmission & ISubmissionSchema
+export interface IEmailSubmissionSchema extends IEmailSubmission, Document {}
 
 export interface IEncryptedSubmission extends ISubmission {
   recipientEmails: never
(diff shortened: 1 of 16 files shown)


src/types/submission.ts
```diff
@@ -83,7 +83,7 @@ export interface IEmailSubmission extends ISubmission {
   getWebhookView(): null
 }
 
-export type IEmailSubmissionSchema = IEmailSubmission & ISubmissionSchema
+export interface IEmailSubmissionSchema extends IEmailSubmission, Document {}
```
mantariksh, line 86: how does this make a difference?

mantariksh (approved):

src/types/submission.ts
karrui: no real difference, could be the same as 

```ts
export type IEmailSubmissionSchema = IEmailSubmission & Document
```

Just needed to ensure no double typing i guess (since `IEmailSubmission` already extends `ISubmission`, but `ISubmissionSchema` extends `ISubmission` and `Document` again)

(But since it was prefixed with `I` I thought to make it an interface (and also all our schemas are interfaces for adding of instance and model methods) for consistency

PR #2053, opened 2021-06-02
refactor(ts-migration): migrates ndjsonstream to typescript

seaerchin:
**Part 2 of extracting admin downloads**

## Problem
`ndjsonstream` was previously written in javascript, which lacked type-safety. this meant that consumers did not know the concrete types which makes it hard to reliably use the stream and to troubleshoot.

## Solution
1. refactored the file to use typescript rather than javascript for type-safety
2. rewrote the code so that the underlying intention is clear. 

**Notes** 
1. the original authors utilize a `try/catch` block for ndjsonstream so that `enqueue` does not fail. however, `enqueue` only throws on `TypeError`, where the controller is not a `ReadableDefaultStream`. i'm currently unsure if this would ever happen (for once, the mozilla [docs](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController) are unclear about where this controller originates from) but i left it in because it's additional safety 

**Breaking Changes** 

    - Details ...


src/public/modules/forms/helpers/ndjsonStream.ts
```diff
@@ -0,0 +1,68 @@
...
+      maybeReader = reader
+      const decoder = new TextDecoder()
+      let data_buf = ''
+
+      return reader
+        .read()
+        .then(function processResult(result): Promise<void> | undefined | void {
+          if (shouldCancel) {
+            return
+          }
```
mantariksh, line 25: noticed you've shifted this block outside the `if (result.done)` - why?

src/public/modules/forms/helpers/ndjsonStream.ts
```diff
@@ -0,0 +1,68 @@
...
+          })
+
+          data_buf = lines[lines.length - 1]
+          return reader.read().then(processResult)
+        })
+    },
+    cancel: function (reason) {
+      console.log('Cancel registered due to ', reason)
+      shouldCancel = true
+      maybeReader && maybeReader.cancel()
```
mantariksh, line 65: consider sticking to `if (maybeReader) maybeRead.cancel()` for style consistency

src/public/modules/forms/helpers/ndjsonStream.ts
seaerchin: i just found it really hard to read with the nested if conditions @__@ i think it's clearer this way but do feel free to disagree!

src/public/modules/forms/helpers/ndjsonStream.ts
seaerchin: ok sure, this works too!

src/public/modules/forms/helpers/ndjsonStream.ts
mantariksh: but the behaviour is different right - previously we would return early only if `result.done && shouldCancel`, but now you're returning early if `shouldCancel` is true, regardless of `result.done`?

seaerchin: oh shit you're absolutely right. great catch and i'll edit the initial condition!

src/public/modules/forms/helpers/ndjsonStream.ts
```diff
@@ -0,0 +1,72 @@
...
+              }
+            }
+            return controller.close()
+          }
+
+          // Read the input in as a stream and split by newline and trim
+          data_buf += decoder.decode(result.value, { stream: true })
+          const lines = data_buf.split('\n')
+          const readableLines = lines.slice(0, -1).map((line) => line.trim())
+          const bufferStore = lines[lines.length - 1]
```
karrui, line 44: Should we keep this part the same? We are conscious of performance in this file, and slicing then mapping incurs an additional `O(n) + O(n)` cost. Should we literally just copy and paste the original file and add types? All these refactors are actually reducing the optimization already done in the original library.

karrui (changes requested):

src/public/modules/forms/helpers/ndjsonStream.ts
seaerchin: I understand that performance is a key factor in this file and given that this is the case, maybe it's reasonable to change it back to the old version.

i'll add comments so that this part is easier to parse and understand!

src/public/modules/forms/helpers/ndjsonStream.ts
```diff
@@ -0,0 +1,71 @@
...
+            data_buf = data_buf.trim()
+            if (data_buf.length !== 0) {
+              try {
+                controller.enqueue(data_buf)
+              } catch (e) {
+                controller.error(e)
+                return
+              }
+            }
+            return controller.close()
```
karrui, line 37: still some differences between the old js and this file. Used to be 

```ts
controller.close()
return
```

I don't think there's an issue here, but perhaps we should stick to the old implementation since there is still a possible regression (in that the initial implementation returns `undefined` but this returns the return type of `controller.close()`

src/public/modules/forms/helpers/ndjsonStream.ts
```diff
@@ -0,0 +1,71 @@
...
+          // Trims the line and queues it in the controller if there is content in the line
+          for (let i = 0; i < lines.length - 1; ++i) {
+            const l = lines[i].trim()
+            if (l.length > 0) {
+              try {
+                controller.enqueue(l)
+              } catch (e) {
+                controller.error(e)
+                shouldCancel = true
+                return reader.cancel()
```
karrui, line 54: same here

karrui (approved):
lgtm

mantariksh (approved):

karrui:
Will be reopening into `develop` branch since this is standalone

PR #2058, opened 2021-06-02
refactor(beta): migrate to TypeScript

### LINKED ISSUE #1506: Migrate betas.client.factory.js to TypeScript
The _Betas_ service is used to manage user access to hidden features in the codebase based on the beta flag permissions within the _user_ object. Originally slated for removal, this has found new use with sgID and should be migrated to TS.

LoneRifle:
## Problem

Closes #1506 

## Solution

- Provide test coverage via jest, crowbar-ing a mock into the angular
  global at runtime, to minimise changes to the module under test
  - Restructure `betas.client.factory` so that its functions are
    accessible both via module import and `angular.module().factory()`
- Allow the functions to take in other beta feature fields for test
  purposes, defaulting to `BETA_FEATURES_FIELD`
- Migrate betas.client.factory to the TypeScript-based BetaService


src/public/services/BetaService.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export const isBetaField = (
+  fieldType: string,
+  betaFeaturesField = BETA_FEATURES_FIELD,
+): BetaFeature | undefined => {
+  return betaFeaturesField[fieldType]
+}
+
+export const userHasAccessToFieldType = (
+  user: HasBetaFlags,
```
liangyuanruo, line 68: does the frontend already have a `User` type? `HasBetaFlags` barely provides any type safety as the only property it contains is optional.

liangyuanruo (approved):
lgtm otherwise!

src/public/services/BetaService.ts
karrui: not yet, will have once I refactor auth controller in the client

LoneRifle: Hang on, isn't there an `IUser` type?

PR #2061, opened 2021-06-03
refactor(formApiClientFactory): rearrange types

chowyiyin:
This PR is a continuation of the FormApi refactor. This PR rearranges the directory of types to remove dependencies across `src/app` and `src/public`.

## Solution
- `DuplicateFormBody`, `UpdateFormParams` and `PublicFormViewDto` were shifted to `types/api/form.ts`
- `IPossiblyPrefilledField` was shifted to `types/field/myinfoField.ts`
- `EditFormFieldParams` was shifted to `types/api/field.ts`


src/public/services/CreateFormService.ts
```diff
@@ -0,0 +1,38 @@
+import axios from 'axios'
```
mantariksh, line 1: this file and the corresponding tests are not supposed to be in here right?

mantariksh (changes requested):

mantariksh (approved):

PR #2063, opened 2021-06-03
feat: log form updates

### LINKED ISSUE #2056: Log all updates to forms

tshuli:
## Problem

Closes #2056

## Tests

- [ ] Create a form with short text field. Edit the field. Check that the edit action shows up in the logs with details of the edit under `body` meta.

### DIFF
--- a/src/app/modules/auth/auth.middlewares.ts
+++ b/src/app/modules/auth/auth.middlewares.ts
@@ -34,11 +34,11 @@ const DENIED_DOMAINS = ['myrp.edu.sg', 'ichat.sp.edu.sg']
  * @returns 400 if user in session is from a disallowed domain and
  * HTTP method changes database state; next otherwise
  */
-export const denyRpSpStudentEmails: ControllerHandler<
-  unknown,
-  unknown,
-  unknown
-> = async (req, res, next) => {
+export const denyRpSpStudentEmails: ControllerHandler = async (
+  req,
+  res,
+  next,
+) => {
   const userId = (req.session as Express.AuthedSession).user._id
   return UserService.findUserById(userId)
     .map((user) => {
@@ -62,14 +62,15 @@ export const denyRpSpStudentEmails: ControllerHandler<
  * Logs all admin actions which change database state (i.e. non-GET requests)
  * @returns next
  */
-export const logAdminAction: ControllerHandler<
-  unknown,
-  unknown,
-  unknown
-> = async (req, res, next) => {
+export const logAdminAction: ControllerHandler<{ formId: string }> = async (
+  req,
+  res,
+  next,
+) => {
   const sessionUserId = (req.session as Express.AuthedSession).user._id
   const body = req.body
   const method = req.method
+  const { formId } = req.params
 
   if (req.method.toLowerCase() !== 'get') {
     logger.info({
@@ -79,6 +80,7 @@ export const logAdminAction: ControllerHandler<
         method,
         ...createReqMeta(req),
         sessionUserId,
+        formId,
         body,
       },
     })
--- a/src/app/modules/auth/auth.middlewares.ts
+++ b/src/app/modules/auth/auth.middlewares.ts
@@ -1,9 +1,13 @@
 import { StatusCodes } from 'http-status-codes'
 
+import { createLoggerWithLabel } from '../../config/logger'
+import { createReqMeta } from '../../utils/request'
 import { ControllerHandler } from '../core/core.types'
 import * as UserService from '../user/user.service'
 
 import { isUserInSession } from './auth.utils'
+
+const logger = createLoggerWithLabel(module)
 
 /**
  * Middleware that only allows authenticated users to pass through to the next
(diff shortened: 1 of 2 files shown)


liangyuanruo (changes requested):
our logging setup is complicated enough as is - I think you should extend `loggingMiddleware`'s `dynamicMeta` function in `logging.ts` instead.

Bonus points if you can spot quick refactors at the same time.

tshuli:
@liangyuanruo thoughts for your consideration

logging it via loggingMiddleware would capture body data from all requests - even submission responses, which we don't want to store in logs

We could set a condition to log the body only if the url corresponds to an admin form route, but then isn't it neater to do it at the admin form router level?

src/app/modules/auth/auth.middlewares.ts
```diff
@@ -53,3 +57,32 @@ export const denyRpSpStudentEmails: ControllerHandler<
...
+
+/**
+ * Logs all admin actions which change database state (i.e. non-GET requests)
+ * @returns next
+ */
+export const logAdminAction: ControllerHandler<
+  unknown,
+  unknown,
+  unknown
+> = async (req, res, next) => {
```
mantariksh, line 69: actually we don't need the `unknown` types since our custom `ControllerHandler` defaults these args to `unknown`

src/app/modules/auth/auth.middlewares.ts
```diff
@@ -53,3 +57,32 @@ export const denyRpSpStudentEmails: ControllerHandler<
...
+
+  if (req.method.toLowerCase() !== 'get') {
+    logger.info({
+      message: 'Admin attempting to make changes',
+      meta: {
+        action: 'logAdminAction',
+        method,
+        ...createReqMeta(req),
+        sessionUserId,
+        body,
```
mantariksh, line 82: a couple of additional things we should log:
1. the form ID, necessary bc the idea is to be able to see a trail of changes for each form. most of the time the formId is in `req.params`, but we can't assume that it is, so you might need to allow for it to be `undefined`. actually maybe you can just log `req.params` so we can query the form ID as `meta.params.formId`
2. the query params. although we do log `urlWithQueryParams` via `createReqMeta`, it might be easier to see and query it in object form.

mantariksh:
>We could set a condition to log the body only if the url corresponds to an admin form route, but then isn't it neater to do it at the admin form router level?

I agree w this

liangyuanruo (approved):
Ok, agree with you and Antariksh. will approve first

src/app/modules/auth/auth.middlewares.ts
tshuli: thanks! removed them; also for denyRpSpStudentEmails :)

src/app/modules/auth/auth.middlewares.ts
tshuli: added in logging for `formId`, but for the other query params, they are only defined in the specific subroutes, so we would need to apply logging at the level of each subroute in order to capture them (which is what I wanted to avoid as it would be quite messy). I think logging by `formId` should be sufficient for investigative purposes - some untangling of the various requests based on URL will be needed but we don't expect this to be too onerous?

tshuli:
@mantariksh thanks for reviewing, pls see if proposed logging of query params is ok

src/app/modules/auth/auth.middlewares.ts
```diff
@@ -53,3 +57,34 @@ export const denyRpSpStudentEmails: ControllerHandler<
         .json({ message: 'User not found' }),
     )
 }
+
+/**
+ * Logs all admin actions which change database state (i.e. non-GET requests)
+ * @returns next
+ */
+export const logAdminAction: ControllerHandler<{ formId: string }> = async (
```
mantariksh, line 65: I'm not sure it's safe to always assume that `formId` will always be a route parameter. that may be the case currently (I'm not even sure) but idt we should couple the implementation to this assumption.

src/app/modules/auth/auth.middlewares.ts
mantariksh: can't we just log `req.query` and `req.params`?

mantariksh: no need to type them, just log only

src/app/routes/api/v3/admin/forms/admin-forms.routes.ts
```diff
@@ -19,6 +20,9 @@ export const AdminFormsRouter = Router()
 AdminFormsRouter.use(withUserAuthentication)
 AdminFormsRouter.use(denyRpSpStudentEmails)
 
+// Log all non-get admin form actions
+AdminFormsRouter.use('/:formId([a-fA-F0-9]{24})/', logAdminAction)
```
mantariksh, line 24: nit for style consistency
```suggestion
AdminFormsRouter.use('/:formId([a-fA-F0-9]{24})', logAdminAction)
```

src/app/modules/auth/auth.middlewares.ts
mantariksh: as discussed, this makes sense to log form-specific actions!

tshuli:
thanks @mantariksh  :)

PR #2067, opened 2021-06-03
fix: restore typings to some model static methods

karrui:
## Problem

Some typings were accidentally removed in #2046 when replacing some static methods' declaration from assigning a function to inlining their methods. This PR fixes that.

## Solution

**Bug Fixes**:

- fix: restore typings to some model static methods


karrui:
Failing tests, setting as draft

liangyuanruo (approved):

PR #2079, opened 2021-06-07
refactor: ensure consistent filesize const

### LINKED ISSUE #690: Ensure that calculation of megabyte is consistent throughout the code base
Megabyte should be consistently defined as 1 million bytes instead of 1048576 bytes, to avoid inconsistency in file size validation

jia1:
Closes #690

### DIFF
--- a/src/app/constants/filesize.ts
+++ b/src/app/constants/filesize.ts
@@ -1,2 +1,0 @@
-// 1 megabyte in bytes
-export const MB = 1048576
--- a/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
@@ -7,6 +7,7 @@ const cloneDeep = require('lodash/cloneDeep')
 const {
   VALID_UPLOAD_FILE_TYPES,
   MAX_UPLOAD_FILE_SIZE,
+  MB,
 } = require('shared/constants')
 const { UPDATE_FORM_TYPES } = require('../constants/update-form-types')
 const { uploadImage } = require('../../../../services/FileHandlerService')
@@ -559,8 +560,8 @@ function EditFieldsModalController(
       field.uploadedFile = ''
       switch (ngfError.$error) {
         case 'maxSize':
-          vm.uploadError = `${(ngfError.size / 1000000).toFixed(2)} MB / ${
-            vm.maxImageSize / 1000000
+          vm.uploadError = `${(ngfError.size / MB).toFixed(2)} MB / ${
+            vm.maxImageSize / MB
           } MB: File size exceeded`
           break
         case 'resize':
@@ -585,7 +586,7 @@ function EditFieldsModalController(
           field.url = result.url
           field.fileMd5Hash = result.fileMd5Hash
           field.name = result.name
-          field.size = `${(result.size / 1000000).toFixed(2)} MB`
+          field.size = `${(result.size / MB).toFixed(2)} MB`
         })
         .catch((uploadError) => {
           // This is a reference to the ng-model of the upload button, which points to the uploaded file
--- a/src/public/modules/forms/admin/controllers/edit-start-page-modal.client.controller.js
+++ b/src/public/modules/forms/admin/controllers/edit-start-page-modal.client.controller.js
@@ -4,6 +4,7 @@ const axios = require('axios').default
 const {
   MAX_UPLOAD_FILE_SIZE,
   VALID_UPLOAD_FILE_TYPES,
+  MB,
 } = require('shared/constants')
 const { uploadLogo } = require('../../../../services/FileHandlerService')
 const { FormLogoState } = require('../../../../../types')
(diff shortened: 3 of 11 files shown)


src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
```diff
@@ -9,6 +9,7 @@ const {
   MAX_UPLOAD_FILE_SIZE,
 } = require('shared/constants')
 const { UPDATE_FORM_TYPES } = require('../constants/update-form-types')
+const { bytesToMegabytes } = require('../../../../../app/constants/filesize')
```
karrui, line 12: since this is the client, I don't think it can access `/app`, which may be why the CI/CD is failing. May want to move the constant to the `/shared` folder.

src/app/constants/filesize.ts
```diff
@@ -1,2 +1,3 @@
-// 1 megabyte in bytes
-export const MB = 1048576
+export const bytesToKilobytes = (bytes: number) => (bytes / 1000).toFixed(2)
+export const bytesToMegabytes = (bytes: number) => (bytes / 1000000).toFixed(2)
+export const megabytesToBytes = (megabytes: number) => megabytes * 1000000
```
karrui, line 3: Should we only have constants instead of a function? The file referencing the constant can do the appropriate calculations.

```ts
export const KB = 1000
export const MB = 1000 * KB
```

src/app/constants/filesize.ts
```diff
@@ -1,2 +1,3 @@
-// 1 megabyte in bytes
-export const MB = 1048576
+export const bytesToKilobytes = (bytes: number) => (bytes / 1000).toFixed(2)
```
karrui, line 1: Should this be in src/shared/constants.ts

karrui (changes requested):

jia1:
GitHub is laggy for me right now, can't check to see what's going on with the build pipelines and other checks

karrui:
Yeah, Fastly is down, bringing half the internet down 😂

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

karrui:
I've rebased your PR to make the commit history cleaner and easier to review~

karrui (approved):
lgtm, @tshuli can you take a looksee to see if anything got missed?

src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
```diff
@@ -559,8 +560,8 @@ function EditFieldsModalController(
       field.uploadedFile = ''
       switch (ngfError.$error) {
         case 'maxSize':
-          vm.uploadError = `${(ngfError.size / 1000000).toFixed(2)} MB / ${
-            vm.maxImageSize / 1000000
+          vm.uploadError = `${(ngfError.size / MB).toFixed(2)} MB / ${
+            vm.maxImageSize / MB
```
mantariksh, line 564: should we do a `toFixed` here as well so we don't show a large decimal to the user?

src/public/modules/forms/admin/controllers/edit-start-page-modal.client.controller.js
```diff
@@ -116,8 +118,8 @@ function EditStartPageController(
       vm.uploaded.file = ''
       switch (ngfError.$error) {
         case 'maxSize':
-          vm.uploadError = `${(ngfError.size / 1000000).toFixed(2)} MB / ${
-            vm.maxLogoSize / 1000000
+          vm.uploadError = `${(ngfError.size / MB).toFixed(2)} MB / ${
+            vm.maxLogoSize / MB
```
mantariksh, line 122: same comment re: toFixed

src/public/modules/forms/base/components/field-attachment.client.component.js
```diff
@@ -22,6 +22,8 @@ angular.module('forms').component('attachmentFieldComponent', {
 function attachmentFieldComponentController($timeout) {
   const vm = this
 
+  vm.MB = MB
```
mantariksh, line 25: why do we need this?

tshuli (approved):
nothing further from me!

src/public/modules/forms/base/components/field-attachment.client.component.js
karrui: some calculations done in the html counterpart

```
ngf-max-size="{{ vm.field.attachmentSize * vm.MB }}"
```

We can theoretically also precalculate this in the JS file but that will probably require watching etc

mantariksh: ah I see it now, thanks!

src/public/modules/forms/admin/controllers/edit-fields-modal.client.controller.js
jia1: While I can't be fully certain why the original author did not include `toFixed`, my best guess is that we expect `maxImageSize` to be divisible by `MB` (and it is). Please let me know if I have to add `toFixed` ultimately.

mantariksh: nope no need, I'm okay to leave this as is

mantariksh (approved):

PR #2080, opened 2021-06-07
refactor: convert CsvMergedHeadersGenerator to typescript

### LINKED ISSUE #2030: Convert CsvMergedHeadersGenerator class and associated functions to TypeScript
To follow the parent `CsvGenerator` class, which was already converted to TS in #1899 . Do also look into migrating related code if it makes sense.

chowyiyin:
## Problem
This PR converts the `CsvMergedHeadersGenerator` to typescript.

Closes #2030

## Solution
Apart from converting the `CsvMergedHeadersGenerator` class to typescript, the response classes in `csv-response-classes` were also converted to typescript. Also, the `Response` class was converted to an abstract class, which cannot be instantiated. All subclass will have to implement the functions `getAnswer` and `numCols`. This was done as the existing implementation in `CsvMergedHeadersGenerator` implies the need for all subclasses of `Response` to have the two aforementioned functions.

## Manual Tests
- [ ] Create a storage mode form with a table field (to check `TableResponse`), a checkbox field (to check `ArrayAnswerResponse`) and an email field (to check `SingleAnswerResponse`). Submit a few responses and download the csv. Ensure that all answers to each form field are present in the csv and the metadata headers are also present.


src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -1,21 +1,32 @@
...
- *    answer?: string,
- *    answerArray?: string[],
- *    fieldType: string,
- *    isHeader?: boolean,
- * }} DisplayedResponse
- */
-
-class CsvMergedHeadersGenerator extends CsvGenerator {
-  constructor(expectedNumberOfRecords, numOfMetaDataRows) {
+import { keyBy } from 'lodash'
```
mantariksh, line 1: maintain the import of the specific library 'lodash/keyBy' so we minimise the bundle size of the frontend

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -27,25 +38,32 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
    */
-  length() {
+  length(): number {
     return this.unprocessed.length
   }
 
   /**
-   *
+   * Adds an UnprocessedRecord to this.unprocessed
    * @param {Object} decryptedContent
```
mantariksh, line 47: we should remove the JSdocs types, since typescript code already contains more specific types. no point maintaining the type documentation in two places

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -1,21 +1,32 @@
...
+import { DisplayedResponse } from 'src/types/response'
+
+import { Response } from './csv-response-classes'
+import { CsvGenerator } from './CsvGenerator'
+import { getResponseInstance } from './response-factory'
+
+type UnprocessedRecord = {
+  created: string
+  submissionId: string
+  record: ReturnType<typeof keyBy>
```
mantariksh, line 14: haha we're going to have to be more specific than this - have you checked out the documentation for `keyBy`? if you actually hover your mouse over `ReturnType<typeof keyBy>`, you'll notice that it's `any` :P

src/public/modules/forms/helpers/csv-response-classes/TableResponse.class.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(responseData: TwoDimResponse) {
+    super(responseData)
+  }
+
+  getAnswer(colIndex: number): string {
+    // Leave cell empty if number of rows is fewer than the index
+    if (colIndex >= this._data.answerArray.length) {
+      return ''
+    }
+    return (this._data as TwoDimResponse).answerArray[colIndex].join(';')
```
mantariksh, line 19: `this._data` should be typed as a class member. I think that should work if you get its type to inherit from `DisplayedResponseWithoutAnswer`

src/public/modules/forms/helpers/csv-response-classes/SingleAnswerResponse.class.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { Response } from './Response.class'
+
+export class SingleAnswerResponse extends Response {
+  constructor(responseData: DisplayedResponse) {
+    super(responseData)
+  }
+
+  getAnswer(): string {
+    return this._data.answer ?? ''
```
mantariksh, line 11: again, this class member should be typed more specifically

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -108,17 +131,20 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
-        const numCols = this.fieldIdToNumCols[fieldId]
-        for (let colIndex = 0; colIndex < numCols; colIndex++) {
-          row.push(this._extractAnswer(up.record, fieldId, colIndex))
-        }
-      }
+        : createdAt.toString() // just convert to string if given date is not valid
+      const row = [up.submissionId, formattedDate]
+
+      this.fieldIdToQuestion.forEach(
+        (_question: { created: string; question: string }, fieldId: string) => {
```
mantariksh, line 141: you don't actually need to specify these types. since `fieldIdToQuestion` is a typed Map, its `forEach` is also automatically typed.

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -80,9 +98,14 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
    * Extracts the string representation from an unprocessed record.
    * @param {Object} unprocessedRecord
    * @param {string} fieldId
+   * @param {number} colIndex
    * @returns {string}
    */
-  _extractAnswer(unprocessedRecord, fieldId, colIndex) {
+  _extractAnswer(
+    unprocessedRecord: Dictionary<Response>,
```
mantariksh, line 105: the problem with `Dictionary` is that it assumes that every string is a valid key. for example, if you do `const value = unprocessedRecord['some crap']`, `value` will be typed as a `Response`. so instead of using `Dictionary` it would be better to declare our own specific type.

mantariksh (changes requested):
I think we should type `getResponseInstance` as part of this as well, wdyt?

src/public/modules/forms/helpers/csv-response-classes/Response.class.ts
```diff
@@ -0,0 +1,29 @@
+import { DisplayedResponse } from '../../../../../types/response'
+
+export abstract class Response {
+  _data: DisplayedResponse
```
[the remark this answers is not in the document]
chowyiyin: By defining the type of `_data` as `DisplayedResponseWithoutAnswer` in `Response.class.ts`, `_data` is typed as `DisplayedResponseWithoutAnswer` and therefore doesn't have the properties `answerArray` or `answer`. Is there a way to get pass this problem?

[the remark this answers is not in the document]
mantariksh: does it work if you type the `_data` member more specifically in the subclasses? eg can you do `_data: DisplayedResponseWithoutAnswer` in `Response.class` and `_data: NestedArrayResponse` in `TableResponse.class`?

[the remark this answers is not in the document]
chowyiyin: oh yes it does thank you!

src/types/response/index.ts
```diff
@@ -56,3 +56,21 @@ export interface IClientEncryptSubmission extends IClientSubmission {
...
+}
+
+export interface ArrayResponse extends DisplayedResponseWithoutAnswer {
+  answerArray: string[]
+}
+
+export interface NestedResponse extends DisplayedResponseWithoutAnswer {
+  answerArray: string[][]
+}
+export interface SingleResponse extends DisplayedResponseWithoutAnswer {
```
mantariksh, line 74: nit but let's stick to either `type` or `interface`, no point mixing the two.

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,54 @@
...
+    default:
+      if (isSingleResponse(fieldRecordData)) {
+        return new SingleAnswerResponse(fieldRecordData)
+      }
+      break
+  }
+  return new ErrorResponse(fieldRecordData)
+}
+
+function isNestedResponse(
```
mantariksh, line 39: let's stick to arrow syntax instead of `function` syntax for style consistency

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,54 @@
...
+  response: DisplayedResponseWithoutAnswer,
+): response is NestedResponse {
+  return 'answerArray' in response
```
mantariksh, line 42: this is not quite enough to determine that the response is a `NestedResponse`, right? we should also check that the array is in fact nested. the `hasProp` utility in the backend might come in useful here; you can consider moving that into the `shared` folder so the frontend can make use of it.

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+function isArrayResponse(
+  response: DisplayedResponseWithoutAnswer,
+): response is ArrayResponse {
+  return 'answerArray' in response
```
mantariksh, line 47: similarly here we should check the type of `answerArray`, not just the presence of the key. e.g. what if it's undefined?

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function isSingleResponse(
+  response: DisplayedResponseWithoutAnswer,
+): response is SingleResponse {
+  return 'answer' in response
```
mantariksh, line 53: same here - we expect `answer` to be of a specific type, in this case I think it's `string` (but maybe also `number`? you'll have to check)

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,54 @@
...
+        return new ArrayAnswerResponse(fieldRecordData)
+      }
+      break
+    default:
+      if (isSingleResponse(fieldRecordData)) {
+        return new SingleAnswerResponse(fieldRecordData)
+      }
+      break
+  }
+  return new ErrorResponse(fieldRecordData)
```
mantariksh, line 36: interesting approach! my concern here is that this will fail silently at the point where we actually discover the error, and then unexpectedly break something else down the line. I _think_ if you simply throw an error here, it will bubble up and get caught by `worker.on('error'` - could you help to verify this? if that's the case then I think we should rely on existing error handling mechanisms (i.e. `worker.on('error'` and `errorCount` in `submissions.client.factory`)

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,54 @@
...
+    case 'table':
+      if (isNestedResponse(fieldRecordData)) {
+        return new TableResponse(fieldRecordData)
+      }
+      break
+    case 'checkbox':
+      if (isArrayResponse(fieldRecordData)) {
+        return new ArrayAnswerResponse(fieldRecordData)
+      }
+      break
```
mantariksh, line 29: this mixing of `switch` and `if` is a little confusing. I'd actually much rather stick to just if-else.

mantariksh (changes requested):

src/public/modules/forms/helpers/response-factory.ts
chowyiyin: I just tried and the error doesn't get caught in `worker.on('error'`. I think it will only get caught if the error occurred in the decryption worker. But I do agree this approach would be better so that its clearer that an error occurred. I'll just wrap the call to `addRecord` with a try catch block and increment `errorCount when there's an error.

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,65 @@
...
+    // eslint-disable-next-line typesafe/no-throw-sync-func
+    throw new Error('Response did not match any known type') // should be caught in submissions client factory
+  }
+}
+
+const isNestedResponse = (
+  response: DisplayedResponseWithoutAnswer,
+): response is NestedResponse => {
+  return (
+    hasProp(response, 'answerArray') &&
```
seaerchin, line 42: wdyt about extracting 

```ts
hasProp(response, 'answerArray') &&
Array.isArray(response.answerArray)
```

into its own function (maybe `isAnswerArray`) and then having the additional subtyping checks?

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -79,11 +97,16 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
-   * @param {Object} unprocessedRecord
-   * @param {string} fieldId
-   * @returns {string}
+   * @param unprocessedRecord
+   * @param fieldId
+   * @param colIndex
+   * @returns string representation of unprocessed record
    */
-  _extractAnswer(unprocessedRecord, fieldId, colIndex) {
+  _extractAnswer(
```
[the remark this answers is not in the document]
chowyiyin: Correct me if I'm wrong but I think `#extractAnswer` is not allowed?

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -28,25 +39,32 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
    */
-  addRecord({ record, created, submissionId }) {
+  addRecord({
+    record,
+    created,
+    submissionId,
+  }: {
+    record: DisplayedResponseWithoutAnswer[]
+    created: string
+    submissionId: string
```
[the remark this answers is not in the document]
chowyiyin: Nope. The `record` in `UnprocessedRecord` is of another type: `{ [fieldId: string]: Response }`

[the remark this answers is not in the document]
seaerchin: Ah ok! Wdyt about making this a type alias for readability

src/public/modules/forms/helpers/response-factory.ts
chowyiyin: Hm.. I know I won't be able to do this with a normal `isAnswerArray` function that returns a boolean as typescript will complain that `answerArray` may not be defined. Do you know of something similar to a type guard that can help to overcome this issue?

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,65 @@
...
+} from '../../../../types/response'
+
+import {
+  ArrayAnswerResponse,
+  Response,
+  SingleAnswerResponse,
+  TableResponse,
+} from './csv-response-classes'
+
+export const getResponseInstance = (
```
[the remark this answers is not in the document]
chowyiyin: I did think about how to add more clarity to the type signature so that it would be clear that an error could be created here. But the benefit of the current approach is that the error can just bubble up to `submissions.client.factory` and it doesn't have to be handled directly in `addRecord`. I did add the comments to the relevant portions of the code to indicate the possibility of an error. Do you think this is sufficient?

src/public/modules/forms/helpers/response-factory.ts
seaerchin: i think you'd need to create a base type! so the base function is a type predicate for that then the other nested ones just extend the base type so ts can infer properly!

src/public/modules/forms/helpers/response-factory.ts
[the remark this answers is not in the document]
seaerchin: i think not handling here is ok but the comments have to be in the jsdoc, because otherwise they won't be reflected on mouse hover @__@ 

the main problem is that errors if not stated in the signature, should be plainly obvious to the caller! adding the comment is good but cos it's not in the JSDoc, callers won't see on hover, which defeats the purpose of the comment

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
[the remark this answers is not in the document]
seaerchin: i think you'd have to declare it in the class then initialize it in the constructor! refer to myinfo.service.ts but it should be possible!

src/public/modules/forms/helpers/response-factory.ts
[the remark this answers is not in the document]
chowyiyin: Oh right... I'll add it in thank you!!

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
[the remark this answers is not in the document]
chowyiyin: I would prefer not to haha :) since this type is only referenced once

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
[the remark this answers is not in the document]
chowyiyin: Not possible due to ECMAScript version

mantariksh (commented):
pinging @karrui for another pair of eyes

tests/unit/frontend/forms/helpers/CsvMergedHeadersGenerator.test.js
```diff
@@ -173,30 +165,6 @@ describe('CsvMergedHeadersGenerator', () => {
       expect(Object.values(generator.fieldIdToNumCols)).toEqual([2])
     })
 
-    it('should handle adding of single header record', () => {
```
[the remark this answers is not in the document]
chowyiyin: This test no longer passes with the new implementation because we can't instantiate a `Response` object, and any object that doesn't fit the three response types will result in an error. I'm not completely sure and do correct me if I'm wrong, but I don't think there's a case where we will try to do this?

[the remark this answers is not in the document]
mantariksh: I'm pretty sure this is important because it seems to be checking for the case where a form contains a single statement field. @karrui can correct me if I'm wrong. does the code work for that case?

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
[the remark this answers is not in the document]
karrui: this should be a type. even though its only used once in this file, functions calling this may need the type.

for example, i see that in the current code this type is actually a CsvRecord.submissionData from the decryption worker (which has not been translated to Typescript yet),

Ideally, the type would be exported from the decryption worker and imported and used here, but since that does not exist yet we should type it here.

tests/unit/frontend/forms/helpers/CsvMergedHeadersGenerator.test.js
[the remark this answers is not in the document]
chowyiyin: I tested and it still works. I printed the record for statement fields and the `answer` has value `Success` so a `SingleAnswerResponse` object will still be created

src/public/modules/forms/helpers/csv-response-classes/ArrayAnswerResponse.class.ts
```diff
@@ -0,0 +1,20 @@
+import { ArrayResponse } from '../../../../../types/response'
+
+import { Response } from './Response.class'
+
+export class ArrayAnswerResponse extends Response {
+  _data: ArrayResponse
```
karrui, line 6: can probably mark this as private with javascript's private modifier [`#`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) and remove the underscore hehe

src/public/modules/forms/helpers/csv-response-classes/Response.class.ts
```diff
@@ -0,0 +1,29 @@
+import { DisplayedResponseWithoutAnswer } from '../../../../../types/response'
+
+export abstract class Response {
+  _data: DisplayedResponseWithoutAnswer
```
karrui, line 4: same here

src/public/modules/forms/helpers/csv-response-classes/SingleAnswerResponse.class.ts
```diff
@@ -0,0 +1,20 @@
+import { SingleResponse } from '../../../../../types/response'
+
+import { Response } from './Response.class'
+
+export class SingleAnswerResponse extends Response {
+  _data: SingleResponse
```
karrui, line 6: same here

src/public/modules/forms/helpers/csv-response-classes/TableResponse.class.ts
```diff
@@ -0,0 +1,24 @@
+import { NestedResponse } from '../../../../../types/response'
+
+import { Response } from './Response.class'
+
+export class TableResponse extends Response {
+  _data: NestedResponse
```
karrui, line 6: same here

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -1,6 +1,8 @@
 'use strict'
 
-const CsvMHGenerator = require('../helpers/CsvMergedHeadersGenerator')
+const {
+  CsvMergedHeadersGenerator: CsvMHGenerator,
```
karrui, line 4: i think we can use the full name now, dont need to rename it anymore :P

src/public/modules/forms/services/submissions.client.factory.js
```diff
@@ -179,8 +181,13 @@ function SubmissionsFactory($q, $http, $timeout, $window, GTag, FormSgSdk) {
                 }
 
                 if (csvRecord.submissionData) {
-                  // accumulate dataset if it exists, since we may have status columns available
-                  experimentalCsvGenerator.addRecord(csvRecord.submissionData)
+                  try {
+                    // accumulate dataset if it exists, since we may have status columns available
+                    experimentalCsvGenerator.addRecord(csvRecord.submissionData)
```
karrui, line 186: should we rename experimentalCsvGenerator now that it has already been in the application for a while

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    hasAnswerArray(response) &&
+    response.answerArray.every((value) => typeof value === 'string')
+  )
+}
+
+const isSingleResponse = (
+  response: DisplayedResponseWithoutAnswer,
+): response is SingleResponse => {
+  return hasProp(response, 'answer') && typeof response.answer === 'string'
```
karrui, line 68: i think there may be issues with this check; are we certain that all answers must be string? What about number fields? They might be of `number` type instead?

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,77 @@
...
+    )
+  )
+}
+
+const isArrayResponse = (
+  response: DisplayedResponseWithoutAnswer,
+): response is ArrayResponse => {
+  return (
+    hasAnswerArray(response) &&
+    response.answerArray.every((value) => typeof value === 'string')
```
karrui, line 61: same here, might be too strict

tests/unit/frontend/forms/helpers/CsvMergedHeadersGenerator.test.js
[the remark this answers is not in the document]
karrui: we shouldn't remove this test case. Why would it fail? If it fails does it not mean the code is not working? This test is checking that indeed fields with `isHeader = true` does not get added to the `fieldIdToQuestion` map

karrui (changes requested):

karrui:
Can we also add more test cases with the additional response type checks?

src/public/modules/forms/helpers/response-factory.ts
chowyiyin: Responses are decrypted via the `decrypt` function in the crypto SDK which returns a `DecryptedObject` which has properties `responses: FormField[]` and `verified?: Record<string, any>`. Answers in `FormField` are supposed to be of type 
```
{ answer: string; answerArray?: never }
  | { answer?: never; answerArray: string[] | string[][] }
```
`responses` and `verified` are then combined after processing the verified fields and form the csv records. So I think number field answers are also strings after decryption. I'm not sure if we should be removing the check `typeof response.answer === 'string'` because verified field answers could be of type `any`. But I think it still works because if I'm not wrong, all verified field answers are also strings at the moment.

Could you confirm if my understanding is correct and whether you think we should remove the check I mentioned?

karrui: Should be fine then, make sure to test on all form field types just in case :P

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -109,26 +114,44 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
   }
 
+  /**
+   * Extracts the string representation from a field response
+   * @param unprocessedRecord
+   * @param fieldId
+   * @param colIndex
+   * @returns string representation of unprocessed record
+   */
+  _extractAnswer(
```
karrui, line 141: missed this, but if this is private we can use the `#` modifier too

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -151,20 +174,20 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
 
   /**
    * Comparator for dates
-   * @param string firstDate
-   * @param string secondDate
+   * @param firstDate
+   * @param secondDate
    */
-  _dateComparator(firstDate, secondDate) {
+  _dateComparator(firstDate: string, secondDate: string): number {
```
karrui, line 190: and here

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -28,25 +45,25 @@ class CsvMergedHeadersGenerator extends CsvGenerator {
...
+   * @param decryptedContent.created
+   * @param decryptedContent.submissionId
+   * @throws Error when trying to convert record into a response instance. Should be caught in submissions client factory.
    */
-  addRecord({ record, created, submissionId }) {
+  addRecord({ record, created, submissionId }: SubmissionRecord): void {
     // First pass, create object with { [fieldId]: question } from
     // decryptedContent to get all the questions.
     const fieldRecords = record.map((content) => {
       const fieldRecord = getResponseInstance(content)
```
karrui, line 64: This won't fail with non-answer fields like Headers and Statements right? The old way returned a SingleAnswer record by default but now we throw an error.

src/public/modules/forms/helpers/csv-response-classes/ArrayAnswerResponse.class.ts
chowyiyin: Not possible due to ECMA version :(
```
ERROR in /home/travis/build/opengovsg/FormSG/src/public/modules/forms/helpers/csv-response-classes/SingleAnswerResponse.class.ts(6,3)
          TS18028: Private identifiers are only available when targeting ECMAScript 2015 and higher.
```

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
chowyiyin: We can't do this because of the ECMA version :(
```
ERROR in /home/travis/build/opengovsg/FormSG/src/public/modules/forms/helpers/csv-response-classes/SingleAnswerResponse.class.ts(6,3)
          TS18028: Private identifiers are only available when targeting ECMAScript 2015 and higher.
```

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
chowyiyin: Yup I checked, it won't fail. They have `answer=""`, so should be fine!

src/public/modules/forms/helpers/csv-response-classes/ArrayAnswerResponse.class.ts
karrui: ehhhhh we are targeting ES6 (i.e. ECMAScript 2015) in our TSConfig....

karrui: use Typescript's [`private`](https://www.typescriptlang.org/docs/handbook/classes.html#public-private-and-protected-modifiers) modifier then...

src/public/modules/forms/helpers/response-factory.ts
chowyiyin: I think the e2e tests do check the downloaded csv content for the case when all fields are added. That should be enough right?

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
karrui: Somehow my comment didn't go through, our TSConfig is indeed targeting ES6 though... but probably not worth it to debug. 

We can use Typescript's own [`private`](https://www.typescriptlang.org/docs/handbook/2/classes.html#private) modifier in that case

src/public/modules/forms/helpers/response-factory.ts
karrui: yea, don't think there's a way to programatically test them correctly since we will be mocking everything anyways

karrui (approved):
lgtm but wait for @mantariksh's approval too

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -1,21 +1,38 @@
...
+import { DisplayedResponseWithoutAnswer } from '../../../../types/response'
+
+import { Response } from './csv-response-classes'
+import { CsvGenerator } from './CsvGenerator'
+import { getResponseInstance } from './response-factory'
+
+type UnprocessedRecord = {
+  created: string
+  submissionId: string
+  record: { [fieldId: string]: Response }
```
mantariksh, line 13: noticed the `{ [fieldId: string]: Response }` shape is repeated in a couple of places. we can extract that into a type for clarity

mantariksh (approved):

src/public/modules/forms/helpers/response-factory.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @param fieldRecordData Field record
+ * @returns Response instance
+ * @throws Error when data does not fit any known response type
+ */
+export const getResponseInstance = (
+  fieldRecordData: DisplayedResponseWithoutAnswer,
+): Response => {
+  if (
+    isNestedResponse(fieldRecordData) &&
+    fieldRecordData.fieldType === 'table'
```
mantariksh, line 27: we can use the enums here - `BasicField.Table`, `BasicField.Checkbox`

PR #2081, opened 2021-06-07
refactor(decryption.worker): refactors decryption worker to ts 

seaerchin:
## Problem
the decryption worker was written in js previously and hence, untyped. this led to unreliability because of no type-checking and resulted in fragile code that was difficult to maintain. 

## Solution
This PR migrates the decryption worker to typescript and edits the `worker.common.js` file to ensure that the worker is transpiled correctly.

As there were multiple try/catches in the original version, a design choice has been made to use `neverthrow` here to simplify the code flow and also, distinguish each error uniquely.

Using `neverthrow` also allows us to extract each step out as a pure function and ensure that its responsibility is limited and focused, allowing for easier trouble shooting of bugs in the future.

**note**: there are 2 duplicated methods (`downloadAndDecryptAttachmentsAsZip` and `downloadAndDecryptAttachments`) because `AdminSubmissionService` utilizes the initialized sdk, which requires `window` - will update depending on discussion in previous PRs. 

**Breaking Changes** 

    - Details ...


seaerchin:
Due to `node_modules` having an error in transpilation with the rewritten web worker, this PR will be closed until the react rewrite (and rebased off it) so that this is no longer a concern.

Issue #2057, opened 2021-06-02
Migrate auth.client.service.js to TypeScript


mantariksh:
The API calls in this service should be migrated into a separate TypeScript service, while the setting of `localStorage` can remain in the current service, as the storage of user details in `localStorage` may not be migrated to React.

PR #2084, opened 2021-06-07
refactor(auth.client): (1) extract email validation and send login otp flow to Typescript

karrui:
## Problem

This PR is part 1 of migrating auth.client.js to Typescript. This PR extracts out the first two functions
- `checkIsEmailAllowed` for verifying whether email domain is whitelisted, and
- `sendLoginOtp` for sending the login OTP

Related to #2057

## Solution

**Breaking Changes** 

**Features**:
- add `AuthService` file containing the first two service functions for admin login flow
- replace old `auth.client.service.js` function calls with new `AuthService` functions.

## Tests

Unit tests have been added for the new `AuthService` functions.

### Manual Tests
- [ ] Login with invalid email. Should return and display not-whitelisted error message below the input
- [ ] Login with disconnected internet (chrome console > network > offline). Should display error message below the input
- [ ] Login with valid email. Should proceed to verify OTP page
- [ ] Enter invalid OTP. Should reject.
- [ ] Click resend OTP, then use previous valid OTP. Should reject. Should receive new OTP.

(Verification of OTP not tested because not in this PR)

### DIFF
--- a/src/public/modules/users/controllers/authentication.client.controller.js
+++ b/src/public/modules/users/controllers/authentication.client.controller.js
@@ -1,10 +1,12 @@
 'use strict'
 
 const HttpStatus = require('http-status-codes')
+const AdminAuthService = require('../../../services/AdminAuthService')
 
 angular
   .module('users')
   .controller('AuthenticationController', [
+    '$q',
     '$scope',
     '$state',
     '$timeout',
@@ -15,6 +17,7 @@ angular
   ])
 
 function AuthenticationController(
+  $q,
   $scope,
   $state,
   $timeout,
@@ -54,7 +57,7 @@ function AuthenticationController(
 
   vm.handleEmailKeyUp = function (e) {
     if (e.keyCode == 13) {
-      vm.isSubmitEmailDisabled === false && vm.checkEmail()
+      vm.isSubmitEmailDisabled === false && vm.login()
       // condition vm.isSubmitEmailDisabled == false is necessary to prevent retries using enter key
       // when submit button is disabled
     } else {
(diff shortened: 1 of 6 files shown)


seaerchin:
lgtm! cool usage of the opaque type!

src/public/modules/users/controllers/authentication.client.controller.js
```diff
@@ -195,21 +193,22 @@ function AuthenticationController(
           vm.signInMsg.isMsg = false
           vm.showOtpDelayNotification = true
         }, 20000)
-      },
-      function (error) {
+      })
+      .catch((error) => {
+        const errorMsg = get(
+          error,
+          'response.data',
```
mantariksh, line 200: this endpoint returns `{ message: string }` if there's an error right, so this should be `'response.data.message'`?

karrui: Good catch, introduced when catching verify email domain's response... we should have fixed that in v3 API call T_T

Fixed in f429926

karrui:
Ready for re-review

src/public/services/AuthService.ts
```diff
@@ -0,0 +1,35 @@
+import axios from 'axios'
+import { Opaque } from 'type-fest'
+
+// Exported for testing.
+export const AUTH_ENDPOINT = '/api/v3/auth'
+
+type Email = Opaque<string, 'Email'>
+
+/**
+ * Check whether the given email string is from an email domain.
```
tshuli, line 10: minor nit:

```suggestion
 * Check whether the given email string is from a whitelisted email domain.
```

src/public/services/AuthService.ts
```diff
@@ -0,0 +1,35 @@
...
+    })
+    .then(() => email as Email)
+}
+
+/**
+ * Sends login OTP to given email
+ * @param email email to send login OTP to
+ * @returns success string if login OTP is sent successfully
+ * @throws Error on invalid email
+ * @throws Error on non 2xx response
```
tshuli, line 27: To check - seems like no error thrown?

src/public/services/__tests__/AuthService.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      expect(MockAxios.post).toHaveBeenCalledWith(EXPECTED_POST_ENDPOINT, {
+        email: mockEmail.toLowerCase(),
+      })
+    })
+  })
+
+  describe('sendLoginOtp', () => {
+    const EXPECTED_POST_ENDPOINT = `${AUTH_ENDPOINT}/otp/generate`
+
+    it('should return success string when OTP is successfully generated', async () => {
```
tshuli, line 42: how about a test case for failure? :)

src/public/services/__tests__/AuthService.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+// Duplicated here instead of exporting from AuthService to prevent production
+// code from casting to Email type without going through a type guard/validator.
+type TestEmail = Opaque<string, 'Email'>
+
+describe('AuthService', () => {
+  describe('checkIsEmailAllowed', () => {
+    const EXPECTED_POST_ENDPOINT = `${AUTH_ENDPOINT}/email/validate`
+
+    it('should return given email argument when email is allowed', async () => {
```
tshuli, line 23: how about a test case when email domain is not whitelisted?

src/public/services/__tests__/AuthService.test.ts
karrui: testing failure is probably unneeded since we don't do anything on failure for this test

src/public/services/__tests__/AuthService.test.ts
karrui: that test case is already tested on the server, these tests are just testing that the endpoint is called

src/public/services/AuthService.ts
karrui: correcto, fixed documentation in 3a2c1c4

mantariksh (approved):

PR #2086, opened 2021-06-07
refactor(auth.client): (2) extract verify otp flow to Typescript

karrui:
Chain: [`sendOtp`](https://github.com/opengovsg/FormSG/pull/2084) <--- `verifyOtp` (you are here)

## Problem

This PR is part 2 of migrating auth.client.js to Typescript. This PR extracts out the verify OTP function.

Related to #2057

## Solution

**Breaking Changes** 

**Features**:

- feat(AuthService): add verifyLoginOtp fn
- ref(AuthService): move User related fns to UserService
- ref: rename Admin prefix to User prefix to better fit the backend API

## Tests

- add unit tests for `verifyLoginOtp`
### Manual Tests
- [ ] Reach verify OTP flow. Invalid OTP should be rejected. Valid OTP should succeed
- [ ] On successful log in, the user object should be stored in localStorage

## Deploy Notes

**New dependencies**:
- `zod`: for writing agency and user base schemas
- moved `date-fns` from devDependencies to dependencies
- ~~downgrade `typescript` to `4.2.4`~~ Not relevant in this PR, already merged in #2092 

**New dev dependencies**:
- `jest-localstorage-mock`: mock localstorage in frontend jest tests


karrui:
Failing on Typescript 4.3.x, see
https://github.com/colinhacks/zod/issues/473
https://github.com/microsoft/TypeScript/issues/44299

Will figure out alternatives or a TS downgrade

src/public/modules/users/controllers/authentication.client.controller.js
```diff
@@ -239,25 +244,32 @@ function AuthenticationController(
...
+      .catch((error) => {
+        let errorMsg = get(
+          error,
+          'response.data',
+          'Failed to send login OTP. Please try again later and if the problem persists, contact us.',
+        )
+        const errorStatus = get(error, 'response.status')
+        if (errorStatus >= StatusCodes.INTERNAL_SERVER_ERROR) {
+          errorMsg = 'An unknown error occurred'
+        }
```
seaerchin, line 257: would a ternary operator be better here or would it be too long to be readable? 

```ts
const errorMsg = (errorStatus >= StatusCodes.INTERNAL_SERVER_ERROR) ? 'An unknown error occurred' : get(
          error,
          'response.data',
          'Failed to send login OTP. Please try again later and if the problem persists, contact us.',
        )
```

src/public/services/UserService.ts
```diff
@@ -0,0 +1,14 @@
...
+export const STORAGE_USER_KEY = 'user'
+
+/**
+ * Save logged in user to localStorage.
+ * May not be needed in React depending on implementation.
+ *
+ * @param user the user to save to local storage
+ */
+export const saveUserToLocalStorage = (user: User | null): void => {
+  localStorage.setItem(STORAGE_USER_KEY, JSON.stringify(user))
```
mantariksh, line 13: if `user` is `null`, do we really want to set localstorage to the string `'null'`?

mantariksh: seems it might be better to just use `clearUserFromLocalStorage` from part 4, which means the 'user' key is either set correctly or not, without this intermediate `'null'` state

mantariksh: ok I now see that's already done

.vscode/settings.json
```diff
@@ -16,5 +16,6 @@
       "source.fixAll.eslint": false
     },
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib"
```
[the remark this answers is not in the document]
karrui: not if you have your own Typescript global/ VSCode bundled typescript. This setting enforces that when opening Forms on VSCode that Form's typescript version is used

src/types/api/user.ts
```diff
@@ -0,0 +1,25 @@
...
+
+type UserId = Opaque<string, 'UserId'>
+type UserContact = Opaque<string, 'UserContact'>
+
+const DateString = z.string().refine(
+  (val) => isDate(parseISO(val)),
+  (val) => ({ message: `${val} is not a valid date` }),
+)
+
+export const User = z.object({
```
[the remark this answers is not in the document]
karrui: I think you can treat this as a class, where the types and the implementation are the same. But since you've raised it up a couple of times I'll change it 😆

src/public/services/UserService.ts
karrui: I'll move them into this PR for completeness

src/public/modules/users/controllers/authentication.client.controller.js
karrui: too many brackets and commas, poor readability, prefer the current implementation.

src/public/services/AuthService.ts
```diff
@@ -33,3 +37,23 @@ export const sendLoginOtp = async (email: Email): Promise<string> => {
...
+ * @throws Error on non 2xx response
+ */
+export const verifyLoginOtp = async (params: {
+  otp: string
+  email: string
+}): Promise<User> => {
+  return axios
+    .post<User>(`${AUTH_ENDPOINT}/otp/verify`, params)
+    .then(({ data }) => {
+      saveUserToLocalStorage(data)
```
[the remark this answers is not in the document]
karrui: good point, will do

src/public/services/UserService.ts
karrui: Oh clearing of user is not in this PR yet, I'll change the argument in this PR to only accept user

karrui: Done in 5a03fa6

src/public/services/AuthService.ts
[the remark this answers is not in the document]
karrui: Done in 5a03fa6

karrui:
Ready for re-review

src/public/modules/users/controllers/authentication.client.controller.js
```diff
@@ -239,25 +246,32 @@ function AuthenticationController(
...
```
mantariksh, line 259: idk why we override the message returned by the server, but we can just preserve existing behaviour

mantariksh (approved):

src/public/modules/users/controllers/authentication.client.controller.js
karrui: Yea, didn't want to touch existing behaviour (for now), can be fixed in rEaCt

PR #2093, opened 2021-06-07
feat: enable retries for webhooks

### LINKED ISSUE #1492: Automatic webhook retries feature

mantariksh:
## Problem

Webhooks are currently only attempted once at the point of form submission. This means that if the receiving system is temporarily down, the form admin has to download the response and trigger their webhook workflow manually for all submissions made during the downtime. This is far from an ideal user experience, especially for forms with critical use cases and many submissions.

Closes #1492 

## Solution

Implement a system of retries using AWS SQS. Further details can be found in the design document.

### For reviewers
Every commit has been reviewed in an individual PR except for the following:

**https://github.com/opengovsg/FormSG/pull/2093/commits/ba0c923d6b6ab5372b3cf6a9e557e09272217fee**
This commit increases the jitter as the size of the base interval increases. This makes sense because it helps to spread load spikes more evenly over the 24 hours.

For example, say there's a spike in responses over a period of 1 hour, and the downstream system cannot handle the load. If the jitter is a constant +- 30min, then it will continue to receive spikes over 1-hour periods due to retries. On the other hand, if the jitter increases with time, then spikes in load can be spread over longer time periods.

Credit to @waynetee for suggesting this.

**https://github.com/opengovsg/FormSG/pull/2093/commits/e6f19626b0faf80a5a07497484a07d1df5f90eb7**
This commit changes the implementation to calculate the time of the next attempt from the time of the _initial attempt_, rather than the time of the last attempt.

For instance, let's say a form was submitted at 9am, and the initial webhook attempts failed at 9am and 9.03am. The initial implementation was to calculate the subsequent attempt 1h +- 15min from 9.03am. The new implementation is to calculate the subsequent attempt 1h +- 15min from 9am.

This makes the timing of the webhook retries more predictable for form admins, since they can predict the time that a particular retry will be attempted based on the time of form submission or the time of the initial attempt.

**https://github.com/opengovsg/FormSG/pull/2093/commits/eca2bcff9beede9b771ba612fac4a4ca44d6b7c6**
This commit mocks `Date.now()` in the producer tests. Without this mocking, the producer tests may fail due to differences in the output of `Date.now()` between the test and the producer.

**Breaking Changes** 

    - Can be rolled back. Existing messages in the queue will be processed once the code is released and any fixes have been made. Receiving systems will just receive the retries later than expected.

## Tests

Tests were added for all new webhook modules and functions.

## Deploy Notes

- `WEBHOOK_SQS_URL`: URL of the SQS queue to use for webhook retries

**New dependencies**:

- [zod](https://www.npmjs.com/package/zod): for parsing webhook queue messages and outputting a typed object
- [sqs-producer](https://www.npmjs.com/package/sqs-producer): for writing messages to the queue
- [sqs-consumer](https://www.npmjs.com/package/sqs-consumer): for consuming messages from the queue

## Manual tests
- [ ] `WEBHOOK_SQS_URL` env var should be added in production.
- [ ] [Guide PR](https://github.com/opendocsg/opendoc-formsg-faq/pull/82) should be merged and go.gov.sg/form-webhook-retries should be updated to point to the correct section
- [ ] Spin up a webhook server and set it up such that it returns a non-200 response for the first request, and subsequently returns 200. Enable retries on the form. Check in the logs that when the initial webhook fails, the webhook is added to the queue (message: "Pushed webhook to queue"). Check that the subsequent retry is received by your server, and the FormSG server logs "Webhook POST succeeded".
- [ ] Without deactivating the form, disable retries. Ensure that your server is returning only non-200 responses, and submit the form. Check in the logs that the webhook fails but the webhook is not added to the queue (there is no "Pushed webhook to queue" log message).

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -26,9 +26,12 @@ RUN apk update && apk upgrade && \
     ttf-freefont \
     tini \
     # Localstack - these are necessary in order to initialise local S3 buckets
+    # jq is a package for easily parsing Localstack health endpoint's JSON output
+    jq \
     py-pip && \
     npm install --quiet node-gyp -g && \
-    pip install awscli-local
+    # [ver1] ensures that the underlying AWS CLI version is also installed
+    pip install awscli-local[ver1]
 
 # Chinese fonts
 RUN echo @edge http://nl.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories && apk add wqy-zenhei@edge
@@ -41,5 +44,5 @@ EXPOSE 5000
 # tini is the init process that will adopt orphaned zombie processes
 # e.g. chromium when launched to create a new PDF
 ENTRYPOINT [ "tini", "--" ]
-# Create local S3 buckets before building the app
-CMD npm run docker-dev
+# Create local AWS resources before building the app
+CMD sh init-localstack.sh && npm run docker-dev
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,7 @@ services:
       - MYINFO_CERT_PATH=./node_modules/@opengovsg/mockpass/static/certs/spcp.crt
       - MYINFO_CLIENT_ID=mockClientId
       - MYINFO_CLIENT_SECRET=mockClientSecret
+      - WEBHOOK_SQS_URL=http://localhost:4566/000000000000/local-webhooks-sqs-main
       - GA_TRACKING_ID
       - SENTRY_CONFIG_URL
       - TWILIO_ACCOUNT_SID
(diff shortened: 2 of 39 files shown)


karrui (approved):
only reviewed the 3 new commits, lgtm

Issue #1935, opened 2021-05-19
SMS verifications free tier limit


syan-syan:
## why

today, verified sms has a free tier limit of 10k responses per form. product ops (POPS) enforces this free tier by manually checking twice a month to see if any form has crossed the 5k mark and emailing form admins to give them a heads up. 

given currently heavy workload on POPS and the increasing number of forms that fall in this category, it is worth automating this to free up POPS to focus on other pressing matters.

### 1. on-enable modal

update this modal to make free tier limit and behaviour clear

header: OTP verification will be disabled at 10K responses

We provide 10,000 free SMS OTP verifications per account, only counting owned forms. Once this limit is reached, SMS OTP verification will be automatically disabled for all owned forms. If you are a collaborator, ensure the form's owner has enough free verifications. If you require more than 10K verifications, please [arrange advance billing with us](https://go.gov.sg/form-twilio-setup).

(smaller text) Free verifications used: (insert count)/10,000

change button to our red button with text 'Accept' 

### 2. warning emails

send email at response count = 2.5k, 5k and 7.5k marks to remind form admin to prepare twilio switch. 

### 3. disable sms verifications on owned forms

at account level 10k free verifications sent 

### 4. update verified SMS enabling behaviour

when admin tries to enable verified SMS, check if form owner's account has 10k free verifications.

(a) if so show an error message.

Error: Form owner has reached free tier limit for SMS verification. To continue using SMS verification, please [arrange billing](https://go.gov.sg/form-twilio-setup). 

(b) if not, show the default modal in #1. 

### 4. disable notification 

send an email to notify admins when SMS verification disabled for an account/form.

syan-syan:
before launch
- change modal button to blue that says I understand
- log email sends, log disables
- warning & disable emails for admins & collaborators 
- add one-liner quota

can do after launch
- #2531

pearlyong:
[mock](https://www.figma.com/file/ta5tjjBqphML8LPRkmxD53/DesignMaster_v2?node-id=1801%3A49461) for one-liner displaying number of free SMSes used

PR #2104, opened 2021-06-08
feat(sms limit): sms warning for different limits

seaerchin:
## Problem
This PR is part 1 of #1935 - warning emails should be sent out at certain pre-defined threshold

## Solution
1. Adds a method to mail service to send templated warning emails out to users. 
    - the method **does not** care about the current count - that part will be done by verification service in an upcoming PR
   
## Screenshots
Screenshot 2021-06-08 at 5 49 27 PM

### DIFF
--- a/src/app/services/mail/mail.service.ts
+++ b/src/app/services/mail/mail.service.ts
@@ -28,6 +28,7 @@ import {
   SendMailOptions,
   SendSingleAutoreplyMailArgs,
   SmsVerificationDisabledData,
+  SmsVerificationWarningData,
 } from './mail.types'
 import {
   generateAutoreplyHtml,
@@ -35,6 +36,7 @@ import {
   generateBounceNotificationHtml,
   generateLoginOtpHtml,
   generateSmsVerificationDisabledHtml,
+  generateSmsVerificationWarningHtml,
   generateSubmissionToAdminHtml,
   generateVerificationOtpHtml,
   isToFieldValid,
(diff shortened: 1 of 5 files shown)


src/app/services/mail/mail.service.ts
```diff
@@ -572,6 +575,42 @@ export class MailService {
       }),
     )
   }
+
+  /**
+   * Sends a warning email to the admin and collaborators of the form when their current verified sms counts hits a limit
+   * @param form The form whose admin and collaborators will be issued a warning
+   * @param remainingSmsCounts The current total responses to the form
```
chowyiyin, line 582: The current total responses isn't the remaining sms count right?

chowyiyin (commented):
lgtm

src/app/views/templates/sms-verification-warning.view.html
```diff
@@ -0,0 +1,29 @@
...
+      You are receiving this as you have enabled SMS OTP verification on a
+      Mobile Number field of your form:
+      <a href="<%= formLink %>"> '<%= formTitle %>' </a>.
+    </p>
+    <p>
+      You can receive <b><%= numAvailable %></b> more responses until free SMS
+      OTP verification is automatically disabled.
+    </p>
+    <p>
+      If you require SMS OTP verification for more than 10,000 responses, please
```
chowyiyin, line 16: Should this be `smsVerificationLimit` instead of 10,000?

seaerchin: thx, good catch!

src/app/services/mail/mail.service.ts
seaerchin: changed teh api (and updated the comment @__@)

karrui (approved):
lgtm

mantariksh (approved):

PR #2111, opened 2021-06-09
refactor(ts-migration): ndjsonstream and process-decrypted-content

seaerchin:
The previous two branches were merged into their dependent branches instead of `develop`. As these 2 changes are self-contained, they are reopened here to be properly merged into `develop` instead

Context from the previous 2 PRs
## Problem
#2045
previously, this file was written in javascript and lacked type safety. this made it hard to assert the results of calling functions in this file and led us to many possible errors, such as when `null` isn't explicitly handled or when the shape of a result isn't known.

## Solution
This PR ports over the file and associated functions to typescript. This PR also does some minor refactoring so that the functions in the file are written more cleanly. 

**Breaking Changes** 

    - Details ...

## Problem
#2053
`ndjsonstream` was previously written in javascript, which lacked type-safety. this meant that consumers did not know the concrete types which makes it hard to reliably use the stream and to troubleshoot.

## Solution
1. refactored the file to use typescript rather than javascript for type-safety
2. rewrote the code so that the underlying intention is clear. 

**Notes** 
1. the original authors utilize a `try/catch` block for ndjsonstream so that `enqueue` does not fail. however, `enqueue` only throws on `TypeError`, where the controller is not a `ReadableDefaultStream`. i'm currently unsure if this would ever happen (for once, the mozilla [docs](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController) are unclear about where this controller originates from) but i left it in because it's additional safety 

**Breaking Changes** 

    - Details ...


karrui (approved):
lgtm, checked that the content is the same as #2045 and #2053

PR #2112, opened 2021-06-09
fix: return storage mode submission version when when retrieving from server

### LINKED ISSUE #2029: SubmissionModel#getSubmissionCursorByFormId and SubmissionModel#findEncryptedSubmissionById should return submission version
currently, the encrypted submission object returned does not return the version number, whilst `formsg-javascript-sdk#decrypt` requires the version number in the param (but is unused):

```ts
EncryptSubmissionSchema.statics.getSubmissionCursorByFormId = function (
  formId,
  dateRange = {},
) {
  const streamQuery = {
    form: formId,
    ...createQueryWithDateParam(dateRange?.startDate, dateRange?.endDate),
  }
  return (
    this.find(streamQuery)
      .select({
        encryptedContent: 1,
        verifiedContent: 1,
        attachmentMetadata: 1,
        created: 1,
        id: 1,
        //  <-- no `version` key returned
      })
      .batchSize(2000)
      .read('secondary')
      .lean()
      // Override typing as Map is converted to Object once passed through `lean()`.
      .cursor() as QueryCursor<SubmissionCursorData>
  )
}
```

```ts
EncryptSubmissionSchema.statics.findEncryptedSubmissionById = function (
  formId: string,
  submissionId: string,
) {
  return this.findOne({
    _id: submissionId,
    form: formId,
    submissionType: SubmissionType.Encrypt,
  })
    .select({
      encryptedContent: 1,
      verifiedContent: 1,
      attachmentMetadata: 1,
      created: 1,
      // <-- no `version` key returned
    })
    .exec()
}
```

this results in us having to hardcode a version number to use the parameter instead of using the submission's version number.

There is currently no runtime errors as the version is not used by the SDK's decrypt method, but this may change in the future

karrui:
## Problem

Currently, the encrypted submission object returned does not return the version number, whilst `formsg-javascript-sdk#decrypt` requires the version number in the param (but not used since we only have 1 version so far).

**This bug does not affect prod**, since the `version` parameter is actually unused by the SDK. This is to prevent future updates from breaking and to correctly return expected required parameters when retrieving storage mode responses.

This PR returns the version number when retrieving submission (meta)data regardless so that the decrypt call doesn't need a mock version number parameter.

Closes #2029

## Solution

**Breaking Changes** 

**Bug Fixes**:

- add version key when retrieving storage mode submission data
- pass version on client when decrypting storage mode responses

## Tests

Tests have been updated to account for the new `version` key returned.
### Manual test
- [ ] View individual storage mode response.  Should decrypt and display correctly
- [ ] Download storage mode response (without attachment). Should download correctly and correct responses decrypted.
- [ ] Download storage mode response. same as above.

### DIFF
--- a/src/app/models/__tests__/encrypt-submission.server.model.spec.ts
+++ b/src/app/models/__tests__/encrypt-submission.server.model.spec.ts
@@ -4,12 +4,11 @@ import moment from 'moment-timezone'
 import mongoose from 'mongoose'
 
 import getSubmissionModel, {
+  getEmailSubmissionModel,
   getEncryptSubmissionModel,
 } from 'src/app/models/submission.server.model'
 import {
-  IEmailSubmissionSchema,
   IEncryptedSubmissionSchema,
-  ISubmissionSchema,
   SubmissionMetadata,
   SubmissionType,
 } from 'src/types'
@@ -17,6 +16,7 @@ import {
 import dbHandler from 'tests/unit/backend/helpers/jest-db'
 
 const Submission = getSubmissionModel(mongoose)
+const EmailSubmission = getEmailSubmissionModel(mongoose)
 const EncryptSubmission = getEncryptSubmissionModel(mongoose)
 
 describe('Encrypt Submission Model', () => {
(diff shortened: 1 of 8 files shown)


karrui:
@seaerchin this fixes the need to hardcode the version number in a Typescript file calling the SDK. However, this PR may also introduce conflicts. See which one gets merged first.

mantariksh (approved):

PR #2114, opened 2021-06-09
chore: remove blocking of SP and RP admin updates

mantariksh:
Removes the temporary fix introduced in #2047 to block form updates from admins who had been erroneously onboarded.


karrui (approved):
elgeeteeem

Issue #1842, opened 2021-05-10
Remove feature manager functionality


liangyuanruo:
Benefits
- Accelerate the React migration by removing the need to doubly test complicated features and their interactions
- Reduce maintenance costs by no longer maintaining code that is not needed
- Free up the need for environment variables
- Save infrastructure costs

Blocked by #1104

PR #2118, opened 2021-06-09
feat(feature-manager): remove feature toggles from frontend

mantariksh:
## Problem

The feature manager, which allows features to be toggled on and off based on environment variables, was a premature optimisation which has not been useful for users and slows development work unnecessarily. It should be removed.

The frontend relies on the `/client/features` API to determine which features to activate, and this reliance must be eliminated first so that the API can subsequently be removed in a backwards-compatible manner.

Part of #1842 

## Solution

Remove all the frontend feature toggles as well as API calls to `/client/features`.

**Breaking Changes** 

    - Details ...
    - Client-side changes only


chowyiyin (commented):
lgtm

karrui (approved):
lgtm

PR #2120, opened 2021-06-09
feat(feature-manager): tear out AggregateStats feature

mantariksh:
## Problem

As part of removing the feature manager, the `AggregateStats` feature needs to be removed. This feature controls whether the Examples page relies on the Submissions collection or the FormStatisticsTotal collection for its queries.

Part of #1842 

## Solution

The approach was to:
1. Tear out the Examples service's reliance on the `RetrievalType` enum, which was in turn determined by the `AGGREGATE_COLLECTION` env var. Since there is now only one `RetrievalType`, the higher-order functions which used to take in a `RetrievalType` and return a service function were converted into regular functions.
2. Tear out the Examples factory and link the controllers directly to the service.
3. Remove mentions of `AGGREGATE_COLLECTION` throughout the documentation and dev environment.

### For reviewers
If you wish to review the test files in detail, the changes may make more sense by commit. https://github.com/opengovsg/FormSG/pull/2120/commits/cc207db84a89bc5b10fc80cabc7b2e2e9185f803 does the actual work of updating the tests, whereas https://github.com/opengovsg/FormSG/pull/2120/commits/e821d0b0abd317da25e1d81f7d64437d71e499f4 just removes certain `describe` blocks which are no longer relevant due to mentions of the AggregateStats feature.

**Breaking Changes** 

    - Details ...
    - Server-side only. Note that this PR can be merged independently of #2118 in a backwards-compatible manner, since the frontend does not rely on this env var.

**Deprecated environment variables**:

- `AGGREGATE_COLLECTION`: previously used to toggle between using the Submissions and FormStatisticsTotal collections for calculating examples page statistics


karrui (approved):
lgtm

PR #2125, opened 2021-06-10
fix(auth): make login emails case-insensitive

mantariksh:
## Problem

#2084 introduced a bug where non-lowercase emails could not be used to log in to the admin dashboard.

The frontend used to mutate `vm.credentials.email` to lowercase at the point of sending the OTP, then use the lowercased string when subsequently verifying the OTP. This is important because the email domains in the Agencies collection are all in lowercase, so only lowercase email domains are considered to be valid. However, #2084 removed this mutation, resulting in OTP verification breaking for non-lowercase emails.

This exposes a source of bugs in the authorisation flow, namely that the backend implicitly relies on the frontend to send email strings in the correct case.

## Solution

For all authorisation endpoints, lowercase email addresses at the edge of the application using Joi validation. This ensures that there is no ambiguity about the case of the string at any layer of the backend.

**Breaking Changes** 

    - Details ...

**Bug Fixes**:

Login to admin dashboard is now case-insensitive for email addresses.

## Tests

Integration tests were added to test the case-insensitivity of the authorisation endpoints.

## Manual tests
- [ ] Ensure that you can log in successfully to the dashboard when typing your email in uppercase.

### DIFF
--- a/src/app/modules/auth/__tests__/auth.routes.spec.ts
+++ b/src/app/modules/auth/__tests__/auth.routes.spec.ts
@@ -97,6 +97,23 @@ describe('auth.routes', () => {
       expect(response.text).toEqual('OK')
     })
 
+    it('should return 200 when domain of body.email has a case-insensitive match in Agency collection', async () => {
+      // Arrange
+      // Insert agency
+      const validDomain = 'example.com'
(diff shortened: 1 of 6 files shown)


karrui (approved):
lgtm, thanks for the catch!

PR #2127, opened 2021-06-10
feat(feature-manager): remove GoogleAnalytics feature

mantariksh:
## Problem

Part of #1842 

## Solution

Removes the `GoogleAnalytics` feature from the feature manager.

**Breaking Changes** 

    - Details ...
    - Server-side only

## Deploy Notes

**Note that the server can no longer start up successfully without the presence of the `GA_TRACKING_ID` environment variable.**

A mock default for `GA_TRACKING_ID` has been adding in `docker-compose.yml` to minimise disruption to developers. However, this means that this env var has to be removed from `docker-compose.yml` in order to use Google Analytics on localhost. This tradeoff was made because developers rarely need to use GA in localhost, but always need to have their dev environment configured correctly.

## Tests

The end-to-end tests used to distinguish between "basic environment" and "full environment" tests to ensure that the open-source version of the product (without any optional env vars) works correctly. Since all env vars are now required and the server now breaks without them, this distinction was removed in https://github.com/opengovsg/FormSG/pull/2127/commits/2ca7deaccd8f0d6590b2ff7ead3e1e45e0415248.


mantariksh:
e2e tests are doing their job, need to update them to use full-env

mantariksh:
actually I think I'll add the test commit separately

karrui (approved):

PR #2129, opened 2021-06-10
test(adminsubmissionservice): unit tests for download methods

seaerchin:
## Problem
#2054 had no unit tests for the newly added method

## Solution
Adds unit tests to download related methods for AdminSubmissionService

**Breaking Changes** 

    - Details ...

### DIFF
--- a/src/public/services/__tests__/AdminSubmissionsService.test.ts
+++ b/src/public/services/__tests__/AdminSubmissionsService.test.ts
@@ -1,18 +1,31 @@
+import { decode } from '@stablelib/base64'
 import axios from 'axios'
+import JSZip from 'jszip'
 import { mocked } from 'ts-jest/utils'
 
 import { SubmissionMetadataList } from 'src/types'
 
-import {
-  countFormSubmissions,
-  getEncryptedResponse,
-  getSubmissionMetadataById,
-  getSubmissionsMetadataByPage,
-} from '../AdminSubmissionsService'
+import * as AdminSubmissionService from '../AdminSubmissionsService'
+import { FormSgSdk } from '../FormSgSdkService'
 import { ADMIN_FORM_ENDPOINT } from '../UpdateFormService'
 
 jest.mock('axios')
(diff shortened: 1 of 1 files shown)


src/public/services/__tests__/AdminSubmissionsService.test.ts
```diff
@@ -162,4 +175,102 @@ describe('AdminSubmissionsService', () => {
...
+
+  describe('downloadAndDecryptAttachment', () => {
+    const MOCK_URL = 'www.decryptme.com'
+    const MOCK_SECRET_KEY = 'this is a secret'
+    const MOCK_PUBLIC_KEY = 'some public key'
+    const MOCK_NONCE = 'use me once and throw away pls'
+    const MOCK_BINARY =
+      '01101001 00100000 01100001 01101101 00100000 01100001 00100000 01101000 01110101 01101101 01100001 01101110'
+
+    it('should decrypt successfully when there is data', async () => {
```
mantariksh, line 187: should we test for case where the sdk returns null?

mantariksh (approved):

src/public/services/__tests__/AdminSubmissionsService.test.ts
seaerchin: yes we really should - but because the sdk just uses `nacl.box.open`, this requires that the prior step of `decryptBase64` is not mocked. i can't seem to figure this out, which is why i've omitted this test. 

any hints towards figuring this out would be greatly appreciated!!!

karrui: mock the sdk returning null

mantariksh: yea just mock out the sdk

seaerchin: very good point - we're just testing that the service method works, not the sdk. thanks for the help! will add it in!

PR #2131, opened 2021-06-10
feat(feature-manager): remove Intranet from feature manager

mantariksh:
## Problem

Part of #1842 

## Solution

1. Remove the Intranet feature from the feature manager.
2. Remove the IntranetFactory and link the service directly to its consumers. The IntranetFactory previously served simply to wrap the service in a `Result` in case the feature was not activated, so it is no longer necessary.
3. Allow the intranet IP list path to default to an empty string, which means that no IPs are detected as intranet IPs.

**Breaking Changes** 

    - Details ...
    - Server-side only

## Deploy Notes

**Note that the server will now crash in the absence of the `INTRANET_IP_LIST_PATH` environment variable.**


docker-compose.yml
```diff
@@ -39,6 +39,7 @@ services:
       - MYINFO_CLIENT_ID=mockClientId
       - MYINFO_CLIENT_SECRET=mockClientSecret
       - WEBHOOK_SQS_URL=http://localhost:4566/000000000000/local-webhooks-sqs-main
+      - INTRANET_IP_LIST_PATH=tests/mock-intranet-ips.txt
```
karrui, line 42: does this need to be added to docker?

karrui (approved):
lgtm with question

docker-compose.yml
mantariksh: ? this is the docker-compose file

mantariksh: as discussed, see https://github.com/opengovsg/FormSG/pull/2131/commits/849a406d944d06ff91b945542351eaddbba31d30

karrui (approved):
lgtm

PR #2133, opened 2021-06-10
feat(sms-limiting): email for disabling sms verification

seaerchin:
## Problem
Admins who have breached their limit of allowable verified SMS should have an eamil sent out to them 

## Solution
1. Adds a method to mail service to send templated disabled emails out to users.

## Screenshots
Screenshot 2021-06-10 at 3 45 07 PM

### DIFF
--- a/src/app/modules/verification/verification.constants.ts
+++ b/src/app/modules/verification/verification.constants.ts
@@ -1,0 +1,1 @@
+export const SMS_VERIFICATION_LIMIT = 10000
--- a/src/app/services/mail/mail.service.ts
+++ b/src/app/services/mail/mail.service.ts
@@ -10,10 +10,12 @@ import {
   BounceType,
   EmailAdminDataField,
   IEmailFormSchema,
+  IPopulatedForm,
   ISubmissionSchema,
 } from '../../../types'
 import config from '../../config/config'
 import { createLoggerWithLabel } from '../../config/logger'
+import { SMS_VERIFICATION_LIMIT } from '../../modules/verification/verification.constants'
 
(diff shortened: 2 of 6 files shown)


src/app/services/mail/mail.service.ts
```diff
@@ -572,6 +576,41 @@ export class MailService {
...
+      const mailOptions: MailOptions = {
+        to: emailRecipients,
+        from: this.#senderFromString,
+        html: mailHtml,
+        subject: '[FormSG] SMS Verification - Free Tier Limit Reached',
+        replyTo: this.#senderMail,
+        bcc: this.#senderMail,
+      }
+
+      return this.#sendNodeMail(mailOptions, { formId: form._id })
```
chowyiyin, line 611: Would it be good to extract this portion? I think this is repeated in `sendSmsVerificationWarningEmail` right?

seaerchin: yeap! the reason is cos i want them to be independent and avoid rebasing ;--; but ya, i'll add it in after they get merged into the base branch!

karrui (approved):
lgtm

src/app/modules/verification/verification.constants.ts
```diff
@@ -0,0 +1 @@
+export const SMS_VERIFICATION_LIMIT = 10000
```
mantariksh, line 1: we can put this in `shared/util/verification` to keep all the numbers in one place

seaerchin: Thanks! Will update in base branch

PR #2149, opened 2021-06-11
feat(feature-manager): hardcode /features API response

mantariksh:
## Problem

Old clients make API calls to the /features endpoint in order to retrieve the list of features currently enabled in the application. However, since the feature manager is being deprecated, all features are always enabled. This means the dependence of this API endpoint on the feature manager is blocking the deprecation of the feature manager.

Part of #1842 

## Solution

Hardcode the API response from the /features endpoint until old clients stop calling it. This allows the rest of feature manager to be safely deprecated.

### DIFF
--- a/src/app/modules/frontend/frontend.controller.ts
+++ b/src/app/modules/frontend/frontend.controller.ts
@@ -141,12 +141,12 @@ export const showFeaturesStates: ControllerHandler<
   Record<FeatureNames, boolean>
 > = (_req, res) => {
   return res.json({
-    captcha: true,
-    sms: true,
-    'spcp-myinfo': true,
-    'verified-fields': true,
-    'google-analytics': true,
-    'webhook-verified-content': true,
-    sentry: true,
+    [FeatureNames.Captcha]: true,
+    [FeatureNames.Sms]: true,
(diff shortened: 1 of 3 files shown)


src/app/modules/frontend/frontend.controller.ts
```diff
@@ -116,15 +115,38 @@ export const generateRedirectUrl: ControllerHandler<
...
+ * TODO (#2147): delete this
  */
 export const showFeaturesStates: ControllerHandler<
   unknown,
-  typeof featureManager.states
+  Record<FeatureNames, boolean>
 > = (_req, res) => {
-  return res.json(featureManager.states)
+  return res.json({
+    captcha: true,
```
karrui, line 144: Should we reference the enum here?

```ts
return res.json({
  [FeatureNames.Captcha]: true,
  [FeatureNames.GoogleAnalytics]: true,
  ...
})
```

karrui (approved):
lgtm

src/app/modules/frontend/frontend.controller.ts
mantariksh: sure can also! the keys are restricted by the controller type but okay

PR #2154, opened 2021-06-14
refactor(angularjs): remove angular-moment

LoneRifle:
## Problem

As we prepare for the React migration, we should minimise the number of 
AngularJS-specific libraries that are currently in use.

## Solution
As we have no need for the AngularJS filters and directives that
angular-moment provides, remove it completely, referencing moment
as a CommonJS module instead.

## Testing
Verify that date formatting remains correct for:
- Admin dashboard form list
- Form builder heading
- Viewing responses in storage mode
- On form submission

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7306,14 +7306,6 @@
       "resolved": "https://registry.npmjs.org/angular-messages/-/angular-messages-1.8.2.tgz",
       "integrity": "sha512-M1qNh/30cLJi4yJJ+3YB8saPonRcavz5Dquqz0T/aUySKJhIkUoeCkmF+BcLH4SJ5PBp04yy4CZUUeNRVi7jZA=="
     },
-    "angular-moment": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/angular-moment/-/angular-moment-1.3.0.tgz",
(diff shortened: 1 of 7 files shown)


mantariksh:
@LoneRifle could you explain more about this? I’m actually not sure this is helpful for React migration because it doesn’t give us any code which we can battle-test in production and then reuse in React. regardless of whether we use `moment-timezone` or `angular-moment` in AngularJS, we're going to throw out the code anyway.

LoneRifle:
> @LoneRifle could you explain more about this? I’m actually not sure this is helpful for React migration because it doesn’t give us any code which we can battle-test in production and then reuse in React. regardless of whether we use `moment-timezone` or `angular-moment` in AngularJS, we're going to throw out the code anyway.

I agree - however, I think that we ought to throw out code earlier if we think we can, so that there is less to figure out later on.

mantariksh:
>This PR would also help to better identify dependencies on moment-timezone within components so that moving these out of AngularJS would be easier.

ok! agree here

mantariksh:
## Manual tests
- [ ] Modify a form and check that the "last modified" under the form title in the admin dashboard is formatted the same way as in production
- [ ] Check that the "last modified" date for forms in the form dashboard is formatted the same way as in production
- [ ] Storage mode CSV download works correctly
- [ ] Formatting of submission date in end page (page after submitting form) works correctly

mantariksh (approved):

PR #2100, opened 2021-06-08
feat: add UEN field

aniruddha-adhikary:
## Problem
It adds a UEN Field

Closes #1273 

## Solution
I added a new field in the backend, frontend admin and public frontend.

**Breaking Changes** 

- [ ] Yes - this PR contains breaking changes
    - Details ...
- [x] No - this PR is backwards compatible  

**Features**:
- New Field for UEN
- Validates UEN number syntax/format

## Before & After Screenshots

**BEFORE**:
[Imagine no UEN Fields]

**AFTER**:
Screenshot 2021-06-08 at 2 00 56 PM
Screenshot 2021-06-08 at 2 02 07 PM

## Deploy Notes
I am not sure if adding a new field would require any sort of migration. Maintainers, please do edit if you think it needs to.

PR #2156, opened 2021-06-14
refactor: simplify isUenValid logic

mantariksh:
Simplifies the logic to validate UENs. Implementation can be cross-referenced against the [docs](https://www.uen.gov.sg/ueninternet/faces/pages/admin/aboutUEN.jspx).

Manual tests as per #2100.

### DIFF
--- a/src/shared/util/uen-validation.ts
+++ b/src/shared/util/uen-validation.ts
@@ -1,164 +1,124 @@
+import validator from 'validator'
+
+/**
+ * Validate entity-type indicators, as per
+ * https://www.uen.gov.sg/ueninternet/faces/pages/admin/aboutUEN.jspx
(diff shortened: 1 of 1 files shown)


yong-jie:
watch out!

```js
// Length 9
isUenValid('1       A') // true
isUenValid('        A') // true
isUenValid('   1    A') // true

// Length 10
isUenValid('   1+1   A') // true
isUenValid(' 0.1+0.1 A') // true
isUenValid('1       1A') // true
isUenValid('-1      1A') // true


```

mantariksh:
@yong-jie thanks for the catch! fixed in 7c66b8b by using regex matching instead

karrui (approved):
lgtm

PR #2159, opened 2021-06-14
feat(feature-manager): remove webhooks, verified content

mantariksh:
Removes webhooks and verified content from feature manager.

Note that unlike many of the other modules, the `.factory` file in the webhooks module was not deleted. This is to make the dependencies between the various webhook modules (service, producer, consumer) clearer by putting all the initialisation in one place.

Part of #1842 

**Breaking Changes** 

## Tests
- [x] Webhooks work in development environment

## Manual tests
- [ ] Create a storage mode form with verified fields. Submit and check that decrypted responses works, with the "[verified]" prefix for verified fields.
- [ ] Create a SP-authenticated storage mode form. Submit and check that the NRIC can be decrypted.
- [ ] Create a webhook server. Set the server to return a non-200 status the first time, then subsequently return 200. Enable retries and check that the webhook is received and decrypted correctly for both the initial attempt (which returns non-200) and the retry (which returns 200).

## Deployment changes
**Note that the server can no longer start without the presence of the `SIGNING_SECRET_KEY` env var.** A default has been added here and in the FormSG SDK to make onboarding easier for developers.


mantariksh:
Converting to draft while figuring out a better way to ensure backwards compatibility for dev environment

karrui (approved):
lgtm

PR #2176, opened 2021-06-15
build: release v5.15.0 - webhook attachments

mantariksh:
This release deploys the changes from #1713 in isolation. These changes were reverted in #2169 in order to minimise risk for release v5.14.0, and are now being redeployed.


mantariksh:
## Manual tests
- [ ] Create a webhook receiving server which decrypts and prints the contents of attachments. Configure it to return 400 for the first request, and subsequently 200. Create a storage mode form with attachment fields, point the webhook URL to your receiving server and enable webhook retries. Check that the attachments are downloaded and decrypted correctly in both webhook attempts (the initial attempt which returns 400 and the first retry which returns 200).

yong-jie (approved):

PR #2186, opened 2021-06-16
chore: improve logging for webhook retries

### LINKED ISSUE #2180: Improve logging for webhook retries
1. Current logging in `webhook.consumer.ts` contains `warn`-level logs for the cases where the form no longer has retries enabled, and where the maximum number of retries has been exceeded. In reality, the latter should be an `error`-level log because widespread failing retries could imply a bug in the signature generation.
2. The form ID should be logged for easier monitoring.

mantariksh:
## Problem

Closes #2180 

## Solution

Makes the following changes:
- When a webhook is no longer retried, log as an error instead of warning. This is for better monitoring, since widespread retry failures could be a symptom of a bug.
- Log the form ID in error logs in the webhook consumer wherever possible.
- Log the message content in the webhook producer.

### DIFF
--- a/src/app/config/logger.ts
+++ b/src/app/config/logger.ts
@@ -13,7 +13,7 @@ import { Environment } from '../../types'
 import { aws, customCloudWatchGroup, isDev, nodeEnv } from './config'
 
 // Params to enforce the logging format.
-type CustomLoggerParams = {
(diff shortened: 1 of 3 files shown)


karrui (approved):
lgtm

Issue #2177, opened 2021-06-15
Remove deprecated form.permissionsList.read key


karrui:
Old forms (created before March 2020) has a form.permissionsList.read key, but Joi validation is blocking this due to new schema shape. This deprecated key should be removed from the collection so Joi validation can be made stricter (again)

syan-syan:
@karrui can close if no outstanding tasks?

karrui:
we need to revert the loosening validation hotfix first before this can be closed. i will open a PR for that

PR #2190, opened 2021-06-16
docs(script): add script to remove permissionList.read key from the db

karrui:
## Problem

Script to remove permissionList.read key from the database.
The key is already removed from the schema, so old/new clients should not matter. 
Any updates by old forms (with the deprecated key already saved in client) will not affect the database.

Related to #2177

## Solution

**Breaking Changes** 

## Deploy Notes

**New scripts**:

- add script to remove permissionList.read key from the database.

### DIFF
--- a/scripts/20210616_remove-permissionList-read/remove-permissionList-read.js
+++ b/scripts/20210616_remove-permissionList-read/remove-permissionList-read.js
@@ -1,0 +1,54 @@
+/* eslint-disable */
+
+// Number of forms with permissionList.read key - expect 0 after running update
+db.forms.count({
+  permissionList: { $elemMatch: { read: { $exists: true } } }
+})
+
+{
+  // Number of objects in permissionList with read key - expect 0 after running update
+  const permissionListWithReadKey = [
+    {
+      $match: {
+        permissionList: { $elemMatch: { read: { $exists: true } } }
+      },
+    },
+    { $project: { permissionList: 1 } },
+    { $unwind: '$permissionList' },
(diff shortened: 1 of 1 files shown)


scripts/20210616_remove-permissionList-read/remove-permissionList-read.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable */
+
+// Number of forms with permissionList.read key - expect 0 after running update
+db.forms.count({
+  permissionList: { $elemMatch: { read: { $exists: true } } }
```
mantariksh, line 5: no need to change but this is equivalent to `'permissionList.read': { $exists: true }`

scripts/20210616_remove-permissionList-read/remove-permissionList-read.js
```diff
@@ -0,0 +1,54 @@
...
+  ]
+
+  db.getCollection('forms').aggregate(permissionListWithReadKey)
+}
+
+// Update - Remove read key
+db.forms.update(
+  { 'permissionList.read': { $exists: true } },
+  { $unset: { 'permissionList.$[elem].read': 1 } },
+  { arrayFilters: [{ 'elem.read': { $exists: true } }], multi: true }
```
mantariksh, line 29: again no need to change but this is equivalent to this:
```suggestion
  { $unset: { 'permissionList.$[].read': 1 } },
  { multi: true }
```

mantariksh (approved):

PR #2193, opened 2021-06-17
feat: Set SP/CP JWT cookie to HttpOnly

### LINKED ISSUE #2065: Set SP/CP JWT cookie to HttpOnly when public users sign in to SP/CP forms
This issue proposes that we set the SP/CP JWT cookies that is being returned from the server when a public user logs in to SP/CP form to be set with the [`HttpOnly`](https://owasp.org/www-community/HttpOnly) header, preventing the possibility of XSS attacks (albeit slim, as the server has various protections set in place by middlewares such as `helmet`).

## Current behaviour
When the public users sign in to SPCP forms, a JWT cookie is returned from the server and set on the browser. This is currently not `HttpOnly` as the JWT is currently being decoded on the client to retrieve information such as the user's NRIC (in SP cookies) and CP user IDs (in CP cookies) so that the logout button can be populated with the ID:

Screenshot 2021-06-03 at 4 09 44 PM

Other related JWT information such as whether the user has selected "remember me" or the token issuing time is also retrieved, but this information is only used for GA logging which does not seem to have been useful.

The user's SP/CP ids from decoding the JWT is also not required as retrieving a public form already returns the needed ID:

```
form: <form object>
spcpSession: {
  userName: <NRIC or CP UID>
}
```

## Proposed behaviour
Set the `HttpOnly` cookie header on SP/CP JWT cookies

## Changes required
1. We will no longer be able to logout users by deleting the cookie in the browser when the user clicks the logout button (shown above), as Javascript will no longer be able to access the cookies.
2. We will also not be able to logout the user when the cookie is cleared, which can happen when the user deletes their cookies, or when the cookie expires.

However, this can be mitigated with the following proposed changes:

1. A new endpoint (probably `/api/v3/forms/:authType/logout`) will need to be created so that the server can remove the cookie when called.
2. Instead of checking whether the cookie exists every 5 seconds (current behaviour), we include a new subkey the returned form object when a user retrieves a form with the HttpOnly JWT:

```
form: <form object>
spcpSession: {
  userName: <NRIC or CP UID>
  msToExpiry: 123456 <-- new value
}
```

The server already knows the information from the JWT and should be able to send it forward to the client. The client will now only need to have a single `setTimeout` to refresh the page using `msToExpiry`.

Note that this does not handle the case when the user clears the cookie on their own, but that should not be the responsibility the application to handle anyways.

tshuli:
## Problem

Closes #2065

## Solution

- App is updated to use `/logout` endpoint to delete sp/cp jwt cookie.
 
**Breaking Changes** 

    - New frontend client expects `/logout` endpoint to be available and will not be able to logout from sp/cp upon rollback

## Tests
- [ ] Create form with singpass. Log in without selecting 'remember me'. Check that form can be submitted and logout works, with jwtSp cookie deleted upon logout
  - [ ] Repeat the above with 'remember me' checked
- [ ] Create form with corppass. Log in. Check that form can be submitted and logout works, with jwtCp cookie deleted upon logout
- [ ] Create form with myinfo and check that myinfo form submission works normally

### DIFF
--- a/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
+++ b/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
@@ -39,7 +39,9 @@ import {
   LoginPageValidationError,
   MissingJwtError,
 } from '../../../spcp/spcp.errors'
+import { SpcpFactory } from '../../../spcp/spcp.factory'
 import { SpcpService } from '../../../spcp/spcp.service'
+import { JwtName } from '../../../spcp/spcp.types'
 import {
   AuthTypeMismatchError,
   FormAuthNoEsrvcIdError,
@@ -1485,6 +1487,71 @@ describe('public-form.controller', () => {
       expect(mockRes.json).toBeCalledWith({
         message: 'Sorry, something went wrong. Please try again.',
       })
+    })
+  })
+
+  describe('handleFormAuthLogout', () => {
+    it('should return 200 if authType is SP and call clearCookie()', async () => {
+      const authType = AuthType.SP
+      const MOCK_REQ = expressHandler.mockRequest({
+        params: {
+          authType,
+        },
+      })
+      const mockRes = expressHandler.mockResponse({ clearCookie: jest.fn() })
+
+      await PublicFormController._handleFormAuthLogout(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.status).toBeCalledWith(200)
+      expect(mockRes.clearCookie).toHaveBeenCalledWith(JwtName[authType])
+      expect(mockRes.json).toBeCalledWith({
+        message: 'Successfully logged out.',
+      })
+    })
+
+    it('should return 200 if authType is CP and call clearCookie()', async () => {
+      const authType = AuthType.CP
+      const MOCK_REQ = expressHandler.mockRequest({
+        params: {
+          authType,
+        },
+      })
+      const mockRes = expressHandler.mockResponse({ clearCookie: jest.fn() })
+
+      await PublicFormController._handleFormAuthLogout(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.status).toBeCalledWith(200)
+      expect(mockRes.clearCookie).toHaveBeenCalledWith(JwtName[authType])
+      expect(mockRes.json).toBeCalledWith({
+        message: 'Successfully logged out.',
+      })
+    })
+
+    it('should return 400 if authType is invalid and not clear cookie', async () => {
+      const MOCK_REQ = expressHandler.mockRequest({
+        params: {
+          authType: AuthType.NIL,
+        },
+      })
+      const mockRes = expressHandler.mockResponse({ clearCookie: jest.fn() })
+
+      await PublicFormController._handleFormAuthLogout(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.status).toBeCalledWith(400)
+      expect(mockRes.json).toBeCalledWith({ message: 'Invalid authType.' })
+      expect(mockRes.clearCookie).not.toHaveBeenCalled()
     })
   })
 
--- a/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
+++ b/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
@@ -515,7 +515,12 @@ describe('public-form.controller', () => {
 
       it('should return 200 when client authenticates using SP', async () => {
         // Arrange
-        const MOCK_SPCP_SESSION = { userName: MOCK_JWT_PAYLOAD.userName }
+        const MOCK_SPCP_SESSION = {
+          userName: MOCK_JWT_PAYLOAD.userName,
+          exp: 1000000000,
+          iat: 100000000,
+          rememberMe: false,
+        }
         const MOCK_SP_AUTH_FORM = {
           ...BASE_FORM,
           authType: AuthType.SP,
@@ -543,13 +548,23 @@ describe('public-form.controller', () => {
         expect(mockRes.json).toHaveBeenCalledWith({
           form: MOCK_SP_AUTH_FORM.getPublicView(),
           isIntranetUser: false,
-          spcpSession: MOCK_SPCP_SESSION,
+          spcpSession: {
+            userName: MOCK_SPCP_SESSION.userName,
+            iat: MOCK_SPCP_SESSION.iat,
+            rememberMe: MOCK_SPCP_SESSION.rememberMe,
+            msToExpiry: expect.toBeNumber(),
+          },
         })
       })
 
       it('should return 200 when client authenticates using CP', async () => {
         // Arrange
-        const MOCK_SPCP_SESSION = { userName: MOCK_JWT_PAYLOAD.userName }
+        const MOCK_SPCP_SESSION = {
+          userName: MOCK_JWT_PAYLOAD.userName,
+          exp: 1000000000,
+          iat: 100000000,
+          rememberMe: false,
+        }
         const MOCK_CP_AUTH_FORM = {
           ...BASE_FORM,
           authType: AuthType.CP,
@@ -576,7 +591,12 @@ describe('public-form.controller', () => {
         expect(mockRes.json).toHaveBeenCalledWith({
           form: MOCK_CP_AUTH_FORM.getPublicView(),
           isIntranetUser: false,
-          spcpSession: MOCK_SPCP_SESSION,
+          spcpSession: {
+            userName: MOCK_SPCP_SESSION.userName,
+            iat: MOCK_SPCP_SESSION.iat,
+            rememberMe: MOCK_SPCP_SESSION.rememberMe,
+            msToExpiry: expect.toBeNumber(),
+          },
         })
       })
 
@@ -989,6 +1009,9 @@ describe('public-form.controller', () => {
     describe('errors in form access', () => {
       const MOCK_SPCP_SESSION = {
         userName: 'mock',
+        exp: 1000000000,
+        iat: 100000000,
+        rememberMe: false,
       }
 
       it('should return 200 with isIntranetUser set to false when a user accesses a form from outside intranet', async () => {
@@ -1051,7 +1074,12 @@ describe('public-form.controller', () => {
         // Assert
         expect(mockRes.json).toHaveBeenCalledWith({
           form: MOCK_SP_AUTH_FORM.getPublicView(),
-          spcpSession: MOCK_SPCP_SESSION,
+          spcpSession: {
+            userName: MOCK_SPCP_SESSION.userName,
+            iat: MOCK_SPCP_SESSION.iat,
+            rememberMe: MOCK_SPCP_SESSION.rememberMe,
+            msToExpiry: expect.toBeNumber(),
+          },
           isIntranetUser: true,
         })
       })
@@ -1086,7 +1114,12 @@ describe('public-form.controller', () => {
         // Assert
         expect(mockRes.json).toHaveBeenCalledWith({
           form: MOCK_CP_AUTH_FORM.getPublicView(),
-          spcpSession: MOCK_SPCP_SESSION,
+          spcpSession: {
+            userName: MOCK_SPCP_SESSION.userName,
+            iat: MOCK_SPCP_SESSION.iat,
+            rememberMe: MOCK_SPCP_SESSION.rememberMe,
+            msToExpiry: expect.toBeNumber(),
+          },
           isIntranetUser: true,
         })
       })
--- a/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
+++ b/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
@@ -1477,7 +1477,9 @@ describe('public-form.controller', () => {
           authType,
         },
       })
-      const mockRes = expressHandler.mockResponse({ clearCookie: jest.fn() })
+      const mockRes = expressHandler.mockResponse({
+        clearCookie: jest.fn().mockReturnThis(),
+      })
 
       await PublicFormController._handleSpcpLogout(MOCK_REQ, mockRes, jest.fn())
 
@@ -1495,7 +1497,9 @@ describe('public-form.controller', () => {
           authType,
         },
       })
-      const mockRes = expressHandler.mockResponse({ clearCookie: jest.fn() })
+      const mockRes = expressHandler.mockResponse({
+        clearCookie: jest.fn().mockReturnThis(),
+      })
 
       await PublicFormController._handleSpcpLogout(MOCK_REQ, mockRes, jest.fn())
 
--- a/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
+++ b/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
@@ -549,12 +549,7 @@ describe('public-form.controller', () => {
         expect(mockRes.json).toHaveBeenCalledWith({
           form: MOCK_SP_AUTH_FORM.getPublicView(),
           isIntranetUser: false,
-          spcpSession: {
-            userName: MOCK_SPCP_SESSION.userName,
-            iat: MOCK_SPCP_SESSION.iat,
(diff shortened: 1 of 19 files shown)


src/public/services/PublicFormAuthService.ts
```diff
@@ -31,3 +33,11 @@ export const validateEsrvcId = (
     )
     .then(({ data }) => data)
 }
+
+export const spcpLogout = (
```
mantariksh, line 37: nit: we usually try to name functions with verbs, e.g. `logoutOfSpcpSession`

src/app/routes/api/v3/forms/public-forms.auth.routes.ts
```diff
@@ -22,6 +22,17 @@ PublicFormsAuthRouter.route('/:formId([a-fA-F0-9]{24})/auth/redirect').get(
...
 )
 
+/**
+ * Removes SP/CP JWT cookie when called to logout user from SP/CP
+ * @route /:authType/logout
+ *
+ * @returns 200 with success message when user logs out successfully
+ * @returns 400 if authType is invalid
+ */
+PublicFormsAuthRouter.route('/:authType/logout').get(
```
mantariksh, line 32: I was thinking `/auth/:authType/logout` might be better, @karrui wdyt?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -271,13 +273,27 @@ export const handleGetPublicForm: ControllerHandler<
     case AuthType.SP:
     case AuthType.CP:
       return SpcpFactory.extractJwtPayloadFromRequest(authType, req.cookies)
-        .map(({ userName }) =>
-          res.json({
+        .map(({ userName, exp, iat, rememberMe }) => {
+          if (!exp) {
+            {
```
mantariksh, line 278: why the extra braces here?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -271,13 +273,27 @@ export const handleGetPublicForm: ControllerHandler<
     case AuthType.SP:
     case AuthType.CP:
       return SpcpFactory.extractJwtPayloadFromRequest(authType, req.cookies)
-        .map(({ userName }) =>
-          res.json({
+        .map(({ userName, exp, iat, rememberMe }) => {
+          if (!exp) {
```
mantariksh, line 277: why would `exp` not be defined? is it not consistent?

mantariksh: and if it is consistent should we account for it in the `isSingpassJwtPayload` and `isCorppassJwtPayload` type guards?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -271,13 +273,27 @@ export const handleGetPublicForm: ControllerHandler<
...
             form: publicForm,
             isIntranetUser,
-            spcpSession: { userName },
-          }),
-        )
+            spcpSession: {
+              userName,
+              iat,
+              rememberMe,
+              msToExpiry: exp * 1000 - Date.now(), // Used to tell browser when to refresh page upon cookie expiry
```
mantariksh, line 293: do we need to perform this transformation on the backend? seems to be only relevant to the UI. would it be cleaner if we just returned the relevant contents of the jwt as-is, and the frontend can perform these calculations which are specific to the UI?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -451,6 +467,42 @@ export const handleFormAuthRedirect = [
...
+ */
+export const _handleFormAuthLogout: ControllerHandler<
+  { authType: AuthType },
+  PublicFormAuthLogoutDto
+> = (req, res) => {
+  const { authType } = req.params
+
+  if (authType !== AuthType.SP && authType !== AuthType.CP) {
+    return res.status(400).json({ message: 'Invalid authType.' })
+  }
```
mantariksh, line 486: the Joi validation already guarantees that the authType is SP or CP right? so we can type this as `{ authType: AuthType.SP | AuthType.CP }` and remove the unnecessary `if` statement

src/app/modules/spcp/spcp.types.ts
```diff
@@ -10,12 +10,16 @@ export type SpcpCookies = Partial<Record<JwtName, string>>
 export type SingpassJwtPayload = {
   userName: string
   rememberMe: boolean
+  iat?: number
+  exp?: number
```
mantariksh, line 14: why are these optional?

src/public/modules/forms/services/spcp-session.client.factory.js
```diff
@@ -36,20 +36,11 @@ function SpcpSession($window, $cookies) {
     clearUserName: function () {
       session.userName = undefined
     },
-    logout: function () {
-      $cookies.remove(
-        session.cookieName,
-        $window.spcpCookieDomain ? { domain: $window.spcpCookieDomain } : {},
-      )
+    logout: function (authType) {
+      PublicFormAuthService.spcpLogout(authType)
```
mantariksh, line 40: do we need to use `$q.when` and a `catch` here? what if the logout is unsuccessful?

src/public/modules/forms/services/spcp-session.client.factory.js
```diff
@@ -16,18 +16,18 @@ function SpcpSession($window, $cookies) {
...
+      session.userName = userName
+      session.rememberMe = rememberMe
+      session.issuedAt = iat
+      if (!rememberMe) {
+        $timeout(function () {
+          $window.location.reload()
+        }, msToExpiry)
+        // Refresh page after cookie expiry time
+        // Timeout is not set when rememberMe === true because cookie expiry is 30 days
+        // i.e. 2592000000 ms which exceeds the maximum delay value of
```
mantariksh, line 29: if we returned the date of expiry as suggested in another comment, would it be possible to use `setInterval` (or better still, the angularJS `$interval`) to check if the date of expiry has exceeded?

src/public/modules/forms/base/controllers/submit-form.client.controller.js
```diff
@@ -27,6 +27,15 @@ function SubmitFormController(
   // The form attribute of the FormData object contains the form fields, logic etc
   vm.myform = FormData.form
 
+  // For SP / CP forms, also include the spcpSession details
+  // This allows the log out button to be correctly populated with the UID
+  // Also provides time to cookie expiry so that client can refresh page
+  if (['SP', 'CP'].includes(vm.myform.authType)) {
+    if (FormData.spcpSession && FormData.spcpSession.userName) {
```
mantariksh, line 34: would it be cleaner to combine these conditionals instead of splitting them into two `if` statements?

src/app/routes/api/v3/forms/public-forms.auth.routes.ts
karrui: agree

src/public/services/PublicFormAuthService.ts
tshuli: thanks! edited

src/app/routes/api/v3/forms/public-forms.auth.routes.ts
tshuli: thanks! edited

src/app/modules/form/public-form/public-form.controller.ts
tshuli: thanks, I cleaned up the types. exp is defined.

src/app/modules/form/public-form/public-form.controller.ts
tshuli: thanks, edited to return exp directly to frontend

src/app/modules/form/public-form/public-form.controller.ts
tshuli: edited

src/app/modules/spcp/spcp.types.ts
tshuli: cleaned up the types

src/public/modules/forms/services/spcp-session.client.factory.js
tshuli: edited

src/public/modules/forms/services/spcp-session.client.factory.js
tshuli: thanks, edited to use $interval

src/public/modules/forms/base/controllers/submit-form.client.controller.js
tshuli: thanks! combined

src/app/modules/form/public-form/public-form.controller.ts
tshuli: Actually on review, exp is defined when the payload is received from the client because it is set with the cookie. When the payload is first generated by the server, exp is undefined. So I think it should be correct to indicate in the types that exp is optional, and to include the defensive check - see if ok?

mantariksh: >when the payload is received from the client

sorry what dyou mean by this? it's always the server which determines the cookie payload right?

if `exp` only appears after we do `res.cookie`, then it seems what's going on here is we have two different types, one pre-`res.cookie` and one post-`res.cookie`. if that's the case then we should separate the types, rather than artificially combining them through an optional key (which in reality is mandatory in one and absent in the other).

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -271,13 +273,27 @@ export const handleGetPublicForm: ControllerHandler<
...
-        .map(({ userName }) =>
-          res.json({
+        .map(({ userName, exp, iat, rememberMe }) => {
+          if (!exp) {
+            {
+              logger.error({
+                message: 'Invalid expiry time for cookie',
+                meta: logMeta,
+              })
+            }
```
mantariksh, line 283: think I left a comment abt this previously but why is there an extra pair of braces here?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -271,13 +273,27 @@ export const handleGetPublicForm: ControllerHandler<
...
             isIntranetUser,
-            spcpSession: { userName },
-          }),
-        )
+            spcpSession: {
+              userName,
+              iat,
+              exp,
+              rememberMe,
+            },
```
mantariksh, line 294: nit: instead of destructuring and restructuring everything, we can just keep this as one `spcpSession` object

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -451,6 +467,38 @@ export const handleFormAuthRedirect = [
...
+  res.clearCookie(JwtName[authType])
+
+  return res.status(200).json({ message: 'Successfully logged out.' })
+}
+
+/**
+ * Handler for /forms/auth/:authType/logout
+ * Valid AuthTypes are SP or CP
+ */
+export const handleFormAuthLogout = [
```
mantariksh, line 493: nit: strictly speaking this only handles SPCP logout, so this name will become less appropriate as we add more auth types (e.g. SGID). maybe `handleSpcpLogout`?

tshuli: thanks! edited

src/app/modules/form/public-form/public-form.controller.ts
tshuli: I thought about it earlier, and since the `spcpSession` object might potentially contain secrets (which is partly why we wanted to change to using httponly), we should be explicit about destructuring and returning the properties in the form object, but I guess in this case no harm. edited

src/app/modules/form/public-form/public-form.controller.ts
tshuli: thanks! sorry left it out when i reinstated the commit. it's been removed

src/app/modules/form/public-form/__tests__/public-form.routes.spec.ts
```diff
@@ -96,13 +97,13 @@ describe('public-form.routes', () => {
...
       // NOTE: This is needed to inject admin info into the form
       const fullForm = await dbHandler.getFullFormById(formId)
-      const expectedResponseBody = JSON.parse(
-        JSON.stringify({
-          form: fullForm?.getPublicView(),
-          spcpSession: { userName: MOCK_COOKIE_PAYLOAD.userName },
-          isIntranetUser: false,
+      const expectedResponseBody = {
+        form: JSON.parse(JSON.stringify(fullForm?.getPublicView())),
+        spcpSession: expect.objectContaining({
```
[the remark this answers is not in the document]
tshuli: thanks, i've added in

src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
```diff
@@ -543,13 +549,23 @@ describe('public-form.controller', () => {
         expect(mockRes.json).toHaveBeenCalledWith({
           form: MOCK_SP_AUTH_FORM.getPublicView(),
           isIntranetUser: false,
-          spcpSession: MOCK_SPCP_SESSION,
+          spcpSession: {
+            userName: MOCK_SPCP_SESSION.userName,
+            iat: MOCK_SPCP_SESSION.iat,
+            rememberMe: MOCK_SPCP_SESSION.rememberMe,
+            exp: MOCK_SPCP_SESSION.exp,
+          },
```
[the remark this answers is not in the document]
tshuli: edited

tshuli:
thanks! edited the types as well

src/app/modules/form/public-form/public-form.controller.ts
tshuli: yes, I meant that `exp` and `iat` are only set upon `res.cookie`. separated the types

karrui (approved):
lgtm, 
but are we certain it is backwards compatible? old clients won't be able to access the cookie since backend is now setting it to HTTP-only, do we care about mitigating this or do we have any way to force them to refresh?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -451,6 +453,38 @@ export const handleFormAuthRedirect = [
...
+ */
+export const _handleSpcpLogout: ControllerHandler<
+  { authType: AuthType.SP | AuthType.CP },
+  PublicFormAuthLogoutDto
+> = (req, res) => {
+  const { authType } = req.params
+
+  res.clearCookie(JwtName[authType])
+
+  return res.status(200).json({ message: 'Successfully logged out.' })
```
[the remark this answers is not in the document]
tshuli: thanks! edited

src/app/modules/spcp/spcp.types.ts
```diff
@@ -20,6 +20,18 @@ export type CorppassJwtPayload = {
 
 export type JwtPayload = SingpassJwtPayload | CorppassJwtPayload
 
+type Timestamp = {
```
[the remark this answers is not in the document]
tshuli: edited

src/app/routes/api/v3/forms/__tests__/public-forms.form.routes.spec.ts
```diff
@@ -105,13 +106,13 @@ describe('public-form.form.routes', () => {
...
       const fullForm = await dbHandler.getFullFormById(formId)
-      const expectedResponseBody = JSON.parse(
-        JSON.stringify({
-          form: fullForm?.getPublicView(),
-          spcpSession: { userName: MOCK_COOKIE_PAYLOAD.userName },
-          isIntranetUser: false,
+      const expectedResponseBody = {
+        form: JSON.parse(JSON.stringify(fullForm?.getPublicView())),
+        spcpSession: expect.objectContaining({
+          userName: MOCK_COOKIE_PAYLOAD.userName,
```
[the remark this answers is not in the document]
tshuli: added

tshuli:
thanks @karrui for comments! I did consider the same problem of backward compatibility for old clients (see #2221). if the user was already logged in to sp/cp before this change, then the existing sp/cp cookie on the browser remains `httponly: false` i.e. client can still delete it. for new logins to sp/cp after the change, the browser receives a 302 redirect to the form after the login is successful. 302 redirects are supposed to be non-cacheable hence the client will reload the page (see [link](https://github.com/opengovsg/FormSG/issues/2221)). I tested with with chrome, firefox, ie and edge that the browser does reload the form upon redirect, so backwards compatibility should not be an issue

PR #2194, opened 2021-06-17
feat(config): support config via dotenv, use EB to create file

### LINKED ISSUE #2126: Address environment variable limit of 4096 bytes
Elastic Beanstalk deployments have a [limit of 4,096 bytes](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environments-cfg-softwaresettings.html#environments-cfg-softwaresettings-console) for environment variables, which is likely to block SGID integration.

One possible way to solve this issue is to use the [AWS Systems Manager Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html) to save the environment variables instead, and inject them into the EC2 environment during EB initialisation via `.ebextensions` configuration files. This [article](https://www.honeybadger.io/blog/configuration-with-ssm-parameter-store/) hints at the shell command that can do this.

The implementer should also update relevant internal documentation once the integration is complete.

LoneRifle:
## Problem

Closes #2126 

## Solution
Allow Form to be configured using a .env file, read by dotenv and
populated into `process.env`.

- `npm install dotenv`
- pick up the `.env` config using dotenv just before we invoke
  `convict`
- remove `.env` entry from `.dockerignore`
- remove env var entries from formsg in `docker-compose.yml` since
  we now rely on the .env file copied into the container instead
- change README given that we no longer declare some env vars in
  `docker-compose.yml`

- use an .ebextensions config to create an .env from Parameter Store
  - determine the environment name by looking at container config files
    directly, [an undocumented technique](https://serverfault.com/a/989037)
- mount the .env into /opt/formsg/.env

**Breaking Changes** 

**Features**:

- Engineers will experience no change to their dev flow, however, the backend will
  now be able to populate `process.env` with the contents of `.env` if present

## Deploy Notes

- Parameter stores as specified in the `.ebextensions` config will have to be created,
  using staging equivalents as a model for production
- To minimise disruption, this feature can be released without making changes to the
  config in Elastic Beanstalk; env variables can be removed from that environment later
  to switch configuration from via Elastic Beanstalk itself to via SSM Parameter Store

**New dependencies**:

- `dotenv` : reads `.env` at runtime


.ebextensions/env-file-creation.config
```diff
@@ -0,0 +1,41 @@
...
+commands:
+  01-create-env:
+    command: "/tmp/create-env.sh"
+
+files:
+  "/tmp/create-env.sh":
+      mode: "000755"
+      content : |
+        #!/bin/bash
+        # Reach into the undocumented container config
```
mantariksh, line 12: what do you mean by "undocumented container config"?

.ebextensions/env-file-creation.config
```diff
@@ -0,0 +1,41 @@
...
+            mkdir -p ${TARGET_DIR}
+            if [ $? -ne 0 ]; then
+                echo 'ERROR: Directory creation failed!'
+                exit 1
+            fi
+        else
+            echo "Directory ${TARGET_DIR} already exists!"
+        fi
+        echo "Creating config for ${ENV_NAME} in ${AWS_REGION}"
+        aws ssm get-parameter --name "${ENV_NAME}-general" --with-decryption --region $AWS_REGION | jq -r '.Parameter.Value' > $TARGET_DIR/.env
```
mantariksh, line 29: is it safe to run this in production when these parameter stores do not exist yet?

docker-compose.yml
```diff
@@ -49,34 +49,6 @@ services:
       # https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
       - GOOGLE_CAPTCHA=6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
       - GOOGLE_CAPTCHA_PUBLIC=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
-      - TWILIO_ACCOUNT_SID
```
mantariksh, line 52: actually do we really need to delete these? even if we leave it in, env vars declared in `.env` will be injected correctly right? I'm asking because ultimately all of these will have local env defaults once we finish removing feature manager

.ebextensions/env-file-creation.config
LoneRifle: See the commit message; tldr - AWS did not document reading directly from this particular file

.ebextensions/env-file-creation.config
LoneRifle: Indeed it is, but I will create the parameter store ahead of time.

docker-compose.yml
LoneRifle: They won't actually; dotenv will not overwrite anything in `process.env` that is already set. I assume that stating the env var in docker-compose will implicitly set it.

.ebextensions/env-file-creation.config
mantariksh: wow this is even more sus than I initially thought

LoneRifle: Actually wait, not undocumented -
https://aws.amazon.com/premiumsupport/knowledge-center/view-environment-properties-elastic-beanstalk/

.ebextensions/env-file-creation.config
liangyuanruo: make no mistake; this is a migration that we will have to perform carefully - recommend releasing standalone.

docker-compose.yml
liangyuanruo: seems to me this isn't a big loss, apart from serving as documentation?

src/app/config/config.ts
```diff
@@ -1,3 +1,13 @@
+// On bootstrapping the application, deliberately
+// import dotenv first and invoke it, so that we
+// pick up all env vars from .env before configuration
+
+/* eslint-disable import/first */
+// eslint-disable-next-line simple-import-sort/imports
+import dotenv from 'dotenv'
+
+dotenv.config()
+
```
liangyuanruo, line 10: how do you feel about using the [preload](https://github.com/motdotla/dotenv#preload) feature instead of polluting the application code?

.ebextensions/env-file-creation.config
liangyuanruo: we could also spin up an identical EB environment, then do a DNS cutover after testing that the deploy works as intended

liangyuanruo (changes requested):
@LoneRifle thanks for undertaking this - could you show the Forms team the benefits of SSM to get some buy-in so that we can manage these ourselves in the future? i think we also have to beef up documentation, both in this repo and in the internal procedures doc.

.ebextensions/env-file-creation.config
liangyuanruo: @LoneRifle could we discuss with the team at one of Form's sync-ups? i think we need greater clarity around these files so that we can assess whether it is wise to rely on them in production

docker-compose.yml
mantariksh: yup not a big loss, actually mainly because I'm adding nearly all these env vars back anyway - see #2218 and #2222

.ebextensions/env-file-creation.config
```diff
@@ -0,0 +1,41 @@
...
+    command: "/tmp/create-env.sh"
+
+files:
+  "/tmp/create-env.sh":
+      mode: "000755"
+      content : |
+        #!/bin/bash
+        # Reach into the undocumented container config
+        AWS_REGION='`{"Ref": "AWS::Region"}`'
+        ENV_NAME=$(jq .system.environment_name /opt/elasticbeanstalk/deploy/configuration/containerconfiguration | sed 's/"//g')
```
mantariksh, line 14: @liangyuanruo and I had a chat abt this and we were thinking maybe there's a better way to do this. instead of coupling the ELB env name to the SSM parameters, can we just add an env var in ELB which indicates the prefix of the parameter stores? eg it can be `staging` or `prod`, and the staging parameter stores can be prefixed with `staging` and prod stores with `prod`

mantariksh (commented):
lgtm but a couple of remaining questions:
- As far as I understand, a parameter store is just a string. This means there is a possibility that we might accidentally format it incorrectly. Is it possible to add safeguards to prevent this possibility, e.g. validating the format of the store before piping it into `.env`?
- If the same env var is present in both ELB and a parameter store, which one takes precedence? If we accidentally repeat an env var in multiple parameter stores, which one will take precedence?

Also, could you give the team a clearer idea of the migration plan before approval pls? e.g.:
- Are we creating the parameter stores ahead of time? When?
- How do we check that all the env vars have been migrated correctly? Do we need to get all hands on deck so we can quickly test that all functionality works as expected?
- What will be the procedure to change env vars once this goes out?
- Are we doing an environment cutover like @liangyuanruo suggested?

LoneRifle:
> lgtm but a couple of remaining questions:
> 
> * As far as I understand, a parameter store is just a string. This means there is a possibility that we might accidentally format it incorrectly. Is it possible to add safeguards to prevent this possibility, e.g. validating the format of the store before piping it into `.env`?

I could look into this, but no guarantees on this front

> * If the same env var is present in both ELB and a parameter store, which one takes precedence? 

ELB env vars take precedence over parameter store, since `dotenv` respects prevailing env vars at the time.

> * If we accidentally repeat an env var in multiple parameter stores, which one will take precedence?
If an env var is repeated, the latest entry in the generated .env file wins

> Also, could you give the team a clearer idea of the migration plan before approval pls? e.g.:
> 
> * Are we creating the parameter stores ahead of time? When?

Parameter stores for staging and uat already created. prod stores will be created on approval of this PR but _before_ merging

> * How do we check that all the env vars have been migrated correctly? Do we need to get all hands on deck so we can quickly test that all functionality works as expected?

All hands on deck to quickly test would be needed.

> * What will be the procedure to change env vars once this goes out?

Make changes in SSM, then trigger a deploy in Elastic Beanstalk through insertion of dummy env vars into config
or redeploying application version


> * Are we doing an environment cutover like @liangyuanruo suggested?
May not be necessary. The change is backward compatible, so we could do this in two phases:

- Push codebase to production, let it run for a while
- Save EB config (which will contain the env vars), then remove all env vars to trigger a deploy that relies on SSM.

Roll back would be to do the steps in reverse.

mantariksh (approved):

liangyuanruo (approved):

PR #2196, opened 2021-06-17
feat(design-system): add Checkbox and Radio Components and stories

### LINKED ISSUE #2008: [React] Add Field/Checkbox & Radio component
**Time estimation: 3-4d**

- [ ] Radio field
- [ ] Checkbox field

Some things to consider:

chowyiyin:
## Problem
This PR implements the `Checkbox` and `Radio` components. `ChakraUI`'s original `Checkbox` and `Radio` components's default hover and focus styles are implemented on the `Control` sub-component. However, we want the these styles to be applied on the entire checkbox and radio rows. Hence, these styles are defined in an additional `part` (`Container`) in the theme for `Checkbox` and is applied explicitly on `ChakraCheckbox` instead to achieve the design specified on Figma. The `Radio` component already has a `Container` part and hence the definition of these styles are done there.

As mentioned in the issue, it could be possible to use different types of components apart from input. Hence, the others input component is introduced as prop, which defaults to an input component (now a `ChakraUI` `input` component). This should be replaced with the custom input component when it is completed.

Closes #2008

### DIFF
--- a/frontend/src/components/Checkbox/Checkbox.stories.tsx
+++ b/frontend/src/components/Checkbox/Checkbox.stories.tsx
@@ -16,17 +16,18 @@ Default.args = { options: ['Option 1', 'Option 2', 'Option 3'] }
 export const OthersInput = Template.bind({})
 OthersInput.args = {
   options: ['Option 1', 'Option 2', 'Option 3'],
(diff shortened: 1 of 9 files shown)


karrui:
@chowyiyin can you check why chromatic is not building on your branches?

PR #2201, opened 2021-06-17
feat: restore v3 API for retrieving individual admin form

mantariksh:
Restores changes from #2113, which were reverted in #2169.

    - New client incompatible with old server due to new API endpoint


karrui (approved):
ownself approve ownself

PR #2206, opened 2021-06-17
refactor(email-submission): encapsulate parsedResponses

yong-jie:
This is a quick refactor that encapsulates the `parsedResponses` object in the email-form submission flow.

## Problem
Parsed responses array gets passed into functions which can freely modify it. Certain methods also rely on an implicit ordering of the array in order to function properly, creating a tight coupling to the sequence in which middlewares are chained.

See #1104

## Solution

The `getProcessedResponses` function has been moved into the `parseResponses` method of the `ParsedResponsesObject`. This object also tracks a separate array of NDI responses, which used to be directly appended to the end of the `parsedResponses` array.

**Breaking Changes** 

## Tests

Copy this [email-mode form](https://staging.form.gov.sg/60d0086fcf8dc0001197b48d/use-template).

- [ ] Submit the form with valid inputs. Submission should go through.
- [ ] Copy the request as a CURL, and modify the responses to have invalid information (like invalid email, phone number). The error you get is `There is something wrong with your form submission. Please check your responses and try again.`
- [ ] Paste that same request you copied as CURL, this time deleting a compulsory field (like the decimal field). The error message is `The form has been updated. Please refresh and submit again.`

### DIFF
--- a/src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
+++ b/src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
@@ -1,0 +1,126 @@
+import { err, ok, Result } from 'neverthrow'
+
+import {
+  getLogicUnitPreventingSubmit,
+  getVisibleFieldIds,
+} from '../../../../shared/util/logic'
+import {
+  FieldResponse,
+  IFieldSchema,
+  IFormDocument,
+  ResponseMode,
+} from '../../../../types'
+import { validateField } from '../../../utils/field-validation'
+import {
+  ConflictError,
+  ProcessingError,
+  ValidateFieldError,
+} from '../submission.errors'
+import { ProcessedFieldResponse } from '../submission.types'
+import { getFilteredResponses } from '../submission.utils'
+
+export default class ParsedResponsesObject {
+  public ndiResponses: ProcessedFieldResponse[] = []
+  constructor(public responses: ProcessedFieldResponse[]) {}
+
+  addNdiResponses(
+    ndiResponses: ProcessedFieldResponse[],
+  ): ParsedResponsesObject {
+    this.ndiResponses = ndiResponses
+    return this
+  }
+
+  /**
+   * Injects response metadata such as the question, visibility state. In
+   * addition, validation such as input validation or signature validation on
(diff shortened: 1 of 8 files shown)


mantariksh (approved):
lgtm, minor suggestion about signature of `addNdiResponses`

src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
```diff
@@ -0,0 +1,130 @@
...
+import { getFilteredResponses } from '../submission.utils'
+
+export default class ParsedResponsesObject {
+  public ndiResponses: ProcessedFieldResponse[] = []
+  private constructor(public responses: ProcessedFieldResponse[]) {}
+
+  addNdiResponses(
+    ndiResponses: ProcessedFieldResponse[],
+  ): ParsedResponsesObject {
+    this.ndiResponses = ndiResponses
```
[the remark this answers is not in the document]
yong-jie: makes sense, i just pushed a commit that addresses this, wdyt?

src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
```diff
@@ -23,11 +28,38 @@ export default class ParsedResponsesObject {
...
-    this.ndiResponses = ndiResponses
-    return this
+  addNdiResponses({
+    authType,
+    uinFin,
+    userInfo,
+  }: {
+    authType: AuthType
+    uinFin: string
+    userInfo?: string
```
mantariksh, line 38: might it be better to type these more strictly? perhaps something like
```
type NdiResponseInfo = 
  | { authType: AuthType.SP | AuthType.MyInfo; uinFin: string }
  | { authType: AuthType.CP; uinFin: string; userInfo: string }
```
so we don't end up e.g. accidentally adding a Corppass response without a `userInfo`

yong-jie: oh wow didnt think of that! I've just incorporated this change, now I don't have to use neverthrow since the types guarantee the correct permutation is always passed in.

mantariksh (approved):

PR #2210, opened 2021-06-18
feat(design-system): add Field/YesNo component and stories, update react-hook-form, add more breakpoints

### LINKED ISSUE #2009: [React] Add Field/YesNo component
**Time estimation: 1d**

karrui:
## Problem

This PR adds the YesNo field component alongside with their stories. A Playground story is also created to check how it would work with React hook form.

Whilst working on this PR, also add breakpoints to the application according to design system so the component is mobile responsive.

Closes #2009 

## Solution

**Features**:

- add breakpoints according to design system
- add YesNo presentation component
- add composite field playground component story

## Before & After Screenshots
In storybook

## Tests

**New dependencies**:

- `react-hook-form` : updated to `7.8.8`

### DIFF
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from 'react'
+import { KeyboardEvent, useCallback, useMemo } from 'react'
 import { BiCheck, BiX } from 'react-icons/bi'
 import {
   Box,
@@ -66,9 +66,20 @@ const YesNoOption = forwardRef<YesNoOptionProps, 'input'>(
       }
     }, [input.onChange, props.isChecked])
 
+    const handleSpacebar = useCallback(
+      (e: KeyboardEvent<HTMLInputElement>) => {
+        if (e.key !== ' ') return
+        if (props.isChecked) {
+          e.preventDefault()
+          handleSelect()
+        }
+      },
+      [handleSelect, props.isChecked],
+    )
+
     return (
       <Box as="label" w="100%" zIndex={props.isChecked ? 1 : 'initial'}>
-        <input {...input} onClick={handleSelect} />
+        <input {...input} onClick={handleSelect} onKeyDown={handleSpacebar} />
         <Flex
           {...checkbox}
           aria-hidden={false}
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -104,12 +104,14 @@ const YesNoOption = forwardRef<YesNoOptionProps, 'input'>(
               bg: 'neutral.200',
             },
             _checked: {
-              boxShadow: 'none',
+              bg: 'neutral.300',
+              boxShadow: `0 0 0 2px ${neutral500}`,
+              borderColor: 'neutral.500',
               _hover: {
-                bg: 'primary.200',
+                bg: 'neutral.300',
               },
               _active: {
-                borderColor: 'primary.500',
+                borderColor: 'neutral.500',
               },
             },
           }}
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -11,6 +11,7 @@ import {
   UseRadioGroupReturn,
   UseRadioProps,
   useStyleConfig,
+  VisuallyHidden,
 } from '@chakra-ui/react'
 
 import { YESNO_THEME_KEY } from '~theme/components/Field/YesNo'
@@ -100,7 +101,10 @@ const YesNoOption = forwardRef<YesNoOptionProps, 'input'>(
         _focusWithin={{ zIndex: 1 }}
       >
         <input {...input} onClick={handleSelect} onKeyDown={handleSpacebar} />
-        <Box {...checkbox} aria-hidden={false} __css={style}>
+        <VisuallyHidden>
+          "{children}" option {props.isChecked ? 'selected' : 'unselected'}
+        </VisuallyHidden>
+        <Box {...checkbox} __css={style}>
           {children}
         </Box>
       </Box>
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -59,6 +59,7 @@ export const YesNo = forwardRef<YesNoProps, 'input'>(
         enterKeyHint: '',
         isDisabled: props.isDisabled,
       }
+
       return [
         getRadioProps({
           value: 'no',
@@ -78,13 +79,19 @@ export const YesNo = forwardRef<YesNoProps, 'input'>(
           side="left"
           colorScheme={colorScheme}
           {...noProps}
-          // Ref is set here so any errors can focus this input
+          // Ref is set here so any errors can focus this input.
           ref={ref}
         >
           <Icon as={BiX} __css={styles.icon} />
           No
         </YesNoOption>
-        <YesNoOption side="right" colorScheme={colorScheme} {...yesProps}>
+        <YesNoOption
+          side="right"
+          colorScheme={colorScheme}
+          {...yesProps}
+          // Ref is set here so any errors can focus this input.
+          ref={ref}
+        >
           <Icon as={BiCheck} __css={styles.icon} />
           Yes
         </YesNoOption>
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -116,11 +116,22 @@ export const YesNo = forwardRef<YesNoProps, 'input'>(
 
     const groupProps = getRootProps()
     const [noProps, yesProps] = useMemo(() => {
+      const baseProps = {
+        enterKeyHint: '',
+        disabled: isDisabled,
+        id: props.name,
+      }
       return [
-        getRadioProps({ value: 'no', enterKeyHint: '', disabled: isDisabled }),
-        getRadioProps({ value: 'yes', enterKeyHint: '', disabled: isDisabled }),
+        getRadioProps({
+          value: 'no',
+          ...baseProps,
+        }),
+        getRadioProps({
+          value: 'yes',
+          ...baseProps,
+        }),
       ]
-    }, [getRadioProps, isDisabled])
+    }, [getRadioProps, isDisabled, props.name])
 
     return (
       <HStack spacing="-px" {...groupProps}>
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -60,6 +60,13 @@ interface YesNoOptionProps extends UseRadioProps {
    * Color scheme of the component to render. Defaults to `primary`.
    */
   colorScheme?: FieldColorScheme
+
+  /**
+   * Callback to be invoked when selection changes.
+   * @note Overridden since this component must be encapsulated by a
+   * `RadioGroup` component and will pass in its `onChange` callback as a prop.
+   */
+  onChange?: UseRadioGroupReturn['onChange']
 }
 
 const YesNoOption = forwardRef<YesNoOptionProps, 'input'>(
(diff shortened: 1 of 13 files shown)


karrui:
Converting to draft pending design tweaks

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+          <Icon
+            display={['none', 'none', 'initial']}
+            as={BiX}
+            fontSize="1.5rem"
+            mr="0.5rem"
+            aria-hidden
+          />
+          No
+        </YesNoOption>
+        <YesNoOption variant="right" colorScheme={colorScheme} {...yesProps}>
```
chowyiyin, line 156: Just wondering, why is the `ref` only passed to the `no` option but not the `yes` option?

karrui: Since this is a radio component, we only want the first ref to be acted on by react hook form. If we pass the same ref to both components, on any form error, the last ref (the "yes" option) instead of the "no" option. You can try it in the Playground story, set isRequired to true and submit without selecting any option

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+  UseRadioGroupReturn,
+  UseRadioProps,
+  useStyleConfig,
+  VisuallyHidden,
+} from '@chakra-ui/react'
+
+import { YESNO_THEME_KEY } from '~theme/components/Field/YesNo'
+import { FieldColorScheme } from '~theme/foundations/colours'
+
+export interface YesNoProps {
```
chowyiyin, line 20: Is there a reason why we don't extend `UseRadioGroupProps`? Is it to make explicit what is passed into this component?

karrui: Yes, we don't actually need those props. We want to be slightly stricter with this component

frontend/src/components/Field/YesNo/YesNo.tsx
chowyiyin: ah okay thanks!

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+  colorScheme?: FieldColorScheme
+}
+
+const YesNoOption = forwardRef<YesNoOptionProps, 'input'>(
+  ({ children, variant, colorScheme = 'primary', ...props }, ref) => {
+    const style = useStyleConfig(YESNO_THEME_KEY, { variant, colorScheme })
+
+    const { getInputProps, getCheckboxProps } = useRadio({
+      ...props,
+      isDisabled: props.disabled,
```
mantariksh, line 71: why do we need to rely on the html `disabled` here instead of the `isDisabled` from `UseRadioProps`?

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+    })
+    const input = getInputProps({}, ref)
```
mantariksh, line 73: what happens if we don't pass the empty object and ref? can't find any docs about this in Chakra. could you add a comment here to explain?

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    const checkbox = getCheckboxProps({})
```
mantariksh, line 75: similarly, why do we need the empty object here? can't find any documentation on this

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    const handleSelect = useCallback(() => {
+      if (props.isChecked && input.onChange) {
+        // eslint-disable-next-line @typescript-eslint/no-extra-semi
+        ;(input.onChange as UseRadioGroupReturn['onChange'])('')
```
mantariksh, line 80: why is this casting necessary and safe? technically onChange takes a HTML form event right?

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }, [input.onChange, props.isChecked])
+
+    const handleSpacebar = useCallback(
+      (e: KeyboardEvent<HTMLInputElement>) => {
+        if (e.key !== ' ') return
+        if (props.isChecked) {
+          e.preventDefault()
+          handleSelect()
```
mantariksh, line 89: why use two nested callbacks instead of combining them and setting `input.onChange` as the dependency for both?

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    return (
+      <Box
+        as="label"
+        w="100%"
+        zIndex={props.isChecked ? 1 : 'initial'}
+        _active={{ zIndex: 1 }}
+        _focusWithin={{ zIndex: 1 }}
+      >
+        <input {...input} onClick={handleSelect} onKeyDown={handleSpacebar} />
```
mantariksh, line 103: I know the variable naming is in line with the chakra docs but can we name this `inputProps` instead? it does get confusing

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+        }),
+        getRadioProps({
+          value: 'yes',
+          ...baseProps,
+        }),
+      ]
+    }, [getRadioProps, isDisabled, props.name])
+
+    return (
+      <HStack spacing="-px" {...groupProps}>
```
mantariksh, line 139: what does `-px` do?

frontend/src/components/Field/YesNo/YesNo.tsx
mantariksh: @karrui can we document this pls? I don't understand even after you've explained. also based on your explanation, is this implementation coupled to react-hook-form?

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+   */
+  variant: 'left' | 'right'
+
+  /**
+   * Color scheme of the component to render. Defaults to `primary`.
+   */
+  colorScheme?: FieldColorScheme
+}
+
+const YesNoOption = forwardRef<YesNoOptionProps, 'input'>(
```
mantariksh, line 65: might be neater to put this in a separate file

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -0,0 +1,169 @@
...
+      [handleSelect, props.isChecked],
+    )
+
+    return (
+      <Box
+        as="label"
+        w="100%"
+        zIndex={props.isChecked ? 1 : 'initial'}
+        _active={{ zIndex: 1 }}
+        _focusWithin={{ zIndex: 1 }}
```
mantariksh, line 101: why do we have these custom styles here rather than in the theme?

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: Was from experimentation before I fully understood how useRadio worked, oops! 
Changed it to use isDisabled prop that is passed into YesNoOption

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: We want to assign a ref (if it exists) to this option, so we pass in the forwarded ref. Since ref is the second argument, just added an empty object to the first. Will add comment and use undefined for the first option instead

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: One is a keyboard event and the other is a click event, won't be compatible and it'll be clearer this way (i hope)?

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: Thanks for all the questions, after changes and experimentation, the same refs can be passed to both options and any errors will still target the first option (i.e. the "No" option).

re @mantariksh: after the change, the implementation is the standard implementation for radio groups (afaik)

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: removed

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: Use -1px instead, same thing with less voodoo

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: Done

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: I've overridden the prop to allow for both event and values.
yes, technically onChange takes a HTML form event, but this onChange prop is actually retrieved from useRadioProps, which accepts input.

We can't use the event version as this is an onClick event instead of the standard onChange event that will be invoking props.onChange, and thus require a placeholder value of `""` instead.

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: Done

frontend/src/components/Field/YesNo/YesNo.tsx
karrui: Inlined them now as per discussion.

frontend/src/utils/storybook.tsx
```diff
@@ -1,6 +1,20 @@
...
+/**
+ * Viewports mapping viewport key to their width in (pixel) number.
+ * Used for Chromatic viewpoint snapshots which requires the numbers in pixels.
+ */
+export const viewports = {
+  xs: parseInt(theme.breakpoints['xs']) * 16,
+  sm: parseInt(theme.breakpoints['sm']) * 16,
+  md: parseInt(theme.breakpoints['md']) * 16,
+  lg: parseInt(theme.breakpoints['lg']) * 16,
+  xl: parseInt(theme.breakpoints['xl']) * 16,
```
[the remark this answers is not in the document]
karrui: Done.

frontend/src/components/Field/YesNo/YesNo.tsx
mantariksh: yup makes more sense now

frontend/src/components/Field/YesNo/YesNoOption.tsx
```diff
@@ -0,0 +1,84 @@
...
+      ...props,
+    })
+    // Empty object needed here as ref is the second argument,
+    // and ref is required so that any refs passed in gets forwarded.
+    const inputProps = getInputProps(undefined, ref)
+    const checkboxProps = getCheckboxProps()
+
+    const handleSelect = useCallback(() => {
+      if (props.isChecked) {
+        props.onChange?.('')
```
mantariksh, line 53: can we leave a comment here or a link to documentation on why we need to do this pls?

frontend/src/components/Field/YesNo/YesNoOption.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    return (
+      <Box as="label" __css={styles.container}>
+        <input
+          {...inputProps}
+          onClick={handleSelect}
+          onKeyDown={handleSpacebar}
+        />
+        <VisuallyHidden>
+          "{children}" option {props.isChecked ? 'selected' : 'unselected'}
```
mantariksh, line 76: how does this behave on a screenreader? might it be better to just have the option ("yes" or "no")?

mantariksh (approved):
thanks for moving the styles into the theme, the components are so much cleaner now!

frontend/src/components/Field/YesNo/YesNoOption.tsx
karrui: This will voice out "Yes option selected" on selection, and "No option selected" on deselection. The double quotes are not voiced out. I think this is clearer than "Yes selected" or "No selected"

frontend/src/components/Field/YesNo/YesNoOption.tsx
karrui: Done. Please tell me if it's not clear enough!

mantariksh (approved):

PR #2212, opened 2021-06-20
feat: Remove self from collaborator list

### LINKED ISSUE #33: Remove self from collaborator list
**Give users the ability to remove themselves from collaborator lists of forms**

Currently only the form owner can remove a collaborator; this is slightly more inconvenient during events such as handing over administration of a form than if collaborators had the ability to remove themselves.

**Proposed solution** to add a bin next to own name for Editors and Read-only to remove oneself.

Confirmation modal: "You will be removed as a collaborator and will lose access to this form. You cannot undo this." buttons: "Remove me" and "Cancel"

frankchn:
## Problem
Enables writers and readers to remove themselves from the collaborator list.

Closes #33

## Solution
Added a new `DELETE /:formId([a-fA-F0-9]{24})/collaborators/self` route that removes the current user from the collaborator list, and wired up the appropriate parts of the front-end for this.

**Breaking Changes** 

## Screenshots
remove-self

## Tests
Create a new form with collaborators, then login as the collaborators and manually remove themselves. It should succeed.

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -2319,7 +2319,7 @@ export const handleRemoveSelfFromCollaborators: ControllerHandler<
           level: PermissionLevel.Read,
         })
       })
-      // Step 2: Update the form collaborators
+      // Step 3: Update the form collaborators
       .andThen((form) => {
         const updatedPermissionList = form.permissionList.filter(
           (user) => user.email.toLowerCase() !== currentUserEmail.toLowerCase(),
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -2289,6 +2289,68 @@ export const handleUpdateCollaborators = [
 ] as ControllerHandler[]
 
 /**
+ * Handler for DELETE /api/v3/admin/forms/:formId/collaborators/self
+ * @precondition Must be preceded by request validation
+ * @security session
+ *
+ * @returns 200 with updated collaborators and permissions
+ * @returns 403 when current user does not have permissions to remove themselves from the collaborators list
+ * @returns 404 when form cannot be found
+ * @returns 410 when updating collaborators for an archived form
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database error occurs
+ */
+export const handleRemoveSelfFromCollaborators: ControllerHandler<
+  { formId: string },
+  PermissionsUpdateDto | ErrorDto
+> = (req, res) => {
+  const { formId } = req.params
+  const sessionUserId = (req.session as Express.AuthedSession).user._id
+  let currentUserEmail = ''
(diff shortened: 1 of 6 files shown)


src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -136,6 +142,33 @@ function CollaboratorModalController(
...
 
+  /**
+   * Removes the current user from the collaborator list and bounces the user back to the forms select page
+   */
+  $scope.removeSelfFromCollab = () => {
+    return $q
+      .when(UpdateFormService.removeSelfFromCollaborators($scope.myform._id))
+      .then(() => {
+        // redirects the user back to the home page now that they have lost access to this form
+        $window.location.href = $window.location.origin
```
frankchn, line 153: @mantariksh @liangyuanruo I am pretty sure this is not the best way to redirect the user to the home page or the front page but I am not sure what the canonical way is.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2277,6 +2277,68 @@ export const handleUpdateCollaborators = [
...
+          user,
+          formId,
+          level: PermissionLevel.Read,
+        })
+      })
+      // Step 2: Update the form collaborators
+      .andThen((form) => {
+        const updatedPermissionList = form.permissionList.filter(
+          (user) => user.email.toLowerCase() !== currentUserEmail.toLowerCase(),
+        )
```
mantariksh, line 2315: I'm thinking we should return a non-200 status code (perhaps 404) if the logged in user is not found in the permission list of the given form, wdyt?

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
```diff
@@ -7,10 +7,12 @@ const UpdateFormService = require('../../../../services/UpdateFormService')
 angular
   .module('forms')
   .controller('CollaboratorModalController', [
+    '$location',
```
mantariksh, line 10: the `$location` service seems to be unused, so you can remove it

src/public/modules/forms/admin/controllers/collaborator-modal.client.controller.js
mantariksh: I think redirecting the user back to the dashboard containing their list of forms would be more intuitive. I tested this and it seems to work (you'll need to replace the `$window` service with `$state`):
```
$scope.toggleRemoveSelfFromCollabModal()
$scope.closeModal()
$state.go('listForms')
```

src/app/modules/form/admin-form/admin-form.controller.ts
frankchn: I think this is redundant because `AuthService.getFormAfterPermissionChecks` should already check for this?

mantariksh: yea true

mantariksh (approved):
testing on staging

PR #2218, opened 2021-06-21
feat(feature-manager): remove sms from feature manager

mantariksh:
Removes the SMS feature from feature manager.

Part of #1842

## Solution

The implementation is along the same lines as all previous PRs related to #1842. Note that the `.factory` file in the `sms` service was maintained due to the function signatures in `sms.service` requiring a default Twilio config.

However, there was a slight wrinkle when deciding how to mock Twilio credentials in the dev environment. Twilio does provide [test credentials](https://www.twilio.com/docs/iam/test-credentials) which are not linked to the data in the production account. However, the test credentials are specific to each account, and the Twilio console still says to keep the test credentials secure. Hence the test credentials were not added as dev environment defaults.

**Breaking Changes** 

    - Details ...

## Tests

- [x] Verified fields can be added in dev environment, but attempts to send SMSes fail

## Manual tests
- [ ] Submit a form with a verified SMS field

## Deploy Notes

**Note that the server will now fail to start if any of the required Twilio env vars are missing.** Mock defaults have been added in `docker-compose` in order to avoid disruption to the dev environment.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,10 +55,11 @@ services:
       # Keep in sync with the development key in
       # https://github.com/opengovsg/formsg-javascript-sdk/blob/develop/src/resource/signing-keys.ts
       - SIGNING_SECRET_KEY=HDBXpu+2/gu10bLHpy8HjpN89xbA6boH9GwibPGJA8BOXmB+zOUpxCP33/S5p8vBWlPokC7gLR0ca8urVwfMUQ==
-      - TWILIO_ACCOUNT_SID
-      - TWILIO_API_KEY
-      - TWILIO_API_SECRET
-      - TWILIO_MESSAGING_SERVICE_SID
+      # Mock Twilio credentials. SMSes do not work in dev environment.
+      - TWILIO_ACCOUNT_SID=ACmockTwilioAccountSid
+      - TWILIO_API_KEY=mockTwilioApiKey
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -55,10 +55,11 @@ services:
       # Keep in sync with the development key in
       # https://github.com/opengovsg/formsg-javascript-sdk/blob/develop/src/resource/signing-keys.ts
       - SIGNING_SECRET_KEY=HDBXpu+2/gu10bLHpy8HjpN89xbA6boH9GwibPGJA8BOXmB+zOUpxCP33/S5p8vBWlPokC7gLR0ca8urVwfMUQ==
-      - TWILIO_ACCOUNT_SID
-      - TWILIO_API_KEY
-      - TWILIO_API_SECRET
-      - TWILIO_MESSAGING_SERVICE_SID
+      # Mock Twilio credentials. SMSes do not work in dev environment.
```
karrui, line 58: Sad, this means we need to override this on our docker-compose with the appropriate credentials if we want smsing in dev? Increases the risk of secrets being pushed if we accidentally forget too :(

karrui (approved):
lgtm

docker-compose.yml
mantariksh: tbh I'm not sure there's a good solution which actually allows SMSes in dev. even with twilio test credentials, the SMSes don't actually get sent.

PR #2219, opened 2021-06-21
fix: Clean-up and right-align home page statistics

frankchn:
## Problem
The current numbers on the home page is hard to read (too many digits). This right-aligns numbers and adds commas per locale settings.

## Solution
Fixed the CSS and added Angular's `DecimalPipe`.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
before-commas

**AFTER**:
after-commas

## Tests
Verify that the home page loads and looks as expected.

### DIFF
--- a/src/public/modules/core/css/landing.css
+++ b/src/public/modules/core/css/landing.css
@@ -162,7 +162,18 @@
   margin-left: 20px;
 }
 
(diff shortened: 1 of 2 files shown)


mantariksh (approved):
lgtm, ran it by @pearlyong also

PR #2220, opened 2021-06-21
feat(form-v2): toast component

### LINKED ISSUE #2022: [React] Add Toast component
**Time estimation: 1d**

seaerchin:
## Problem
Form requires our own toasting system, with different design considerations compared to the toast built by chakra ui. This PR solves that issue by adding a custom toast component.

This PR also adds additional viewports to test desktop vs mobile views (base views only went up tablet size).

Closes #2022 

## Solution
1. Adds an (unexposed) `Toast` variant for `Alert`. This controls the base UI theming (borders, colours etc) for the component
2. Adds a `Toast` component. This composes `Alert`  and other various sub-components. There is a media query here to dynamically resize the toast component and determine its margin. This component receives an [`onClose` prop](https://github.com/chakra-ui/chakra-ui/blob/ae42755583e38fe04c2c19685e096d1800e22cca/packages/toast/src/use-toast.tsx#L36-L39) that Chakra ui uses to close the toast with animation. 
    1. we can't use a ref and the `toast.close` method as described [here](https://chakra-ui.com/docs/feedback/toast#closing-toasts) because it only closes the last opened toast. 
3.  Adds a custom `useToast` hook, with sensible defaults set for the toast. 

**Breaking Changes** 

    - Details ...

## Screenshots
screen shots are available in the storybook preview

### DIFF
--- a/frontend/src/hooks/useToast.ts
+++ b/frontend/src/hooks/useToast.ts
@@ -1,0 +1,38 @@
+import { PropsWithChildren } from 'react'
+import {
+  RenderProps,
+  useToast as useChakraToast,
+  UseToastOptions,
+} from '@chakra-ui/react'
+
+import { Toast, ToastProps } from '~/components/Toast/Toast'
+
+type ToastBehaviourProps = Pick<
+  UseToastOptions,
+  'duration' | 'position' | 'render'
+>
+
+// onClose is provided by the chakra hook and should not be exposed to clients
+export type UseToastProps = Omit<ToastBehaviourProps & ToastProps, 'onClose'>
+
+type UseToast = () => (props: UseToastProps) => void
+
+export const useToast: UseToast = () => {
+  const toast = useChakraToast()
+
+  return ({
+    duration = 6000,
+    position = 'top',
+    render,
+    ...rest
+  }: UseToastProps) => {
+    toast({
+      duration,
+      position,
+      render:
+        render ??
+        ((props: PropsWithChildren<RenderProps>) =>
+          Toast({ ...rest, ...props })),
+    })
+  }
+}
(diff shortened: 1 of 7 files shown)


karrui:
Can you enable chromatic and reopen this PR so the storybook will build? ask @chowyiyin how she did it for her PR

karrui (commented):
I think we should just create a [custom toast](https://chakra-ui.com/docs/feedback/toast#custom-component) component, then create a new useToast hook that internally calls Chakra's useToast function that is actually calling the render function.

Some example of what I mean

```ts
// useToast.tsx
import { useToast as useChakraToast, UseToastOptions } from '@chakra-ui/toast'

export const useToast = (initialOptions: UseToastOptions) => {
  const toast = useChakraToast({
    ...initialOptions,
    render: () => <SomeDefaultToastAccordingToInitialOptions />
  })

  const realUseToastFunction = (props?: UseToastOptions) => {
    // Remember, props can be anything you want. But UseToastOptions contains
    // stuff like `variant`, `title`, `description`, etc. Seems like the stuff
    // you would want to render in your own custom toast

    return toast({
      ...props,
      render: () => <CustomToast {...props} />,
    })
  }
  return realUseToastFunction
}

```

using it:
```ts
const SomeComponent: FC = () => {
  const toast = useToast()

  return <Button onClick={() => toast({...yourMiscPropsJustLikeChakrasUseToast)}>AAAA</Button>
}

```

Storybook should also only render the displayed Toast component by itself, with documentation on how to use it with the new useToast hook.

frontend/src/hooks/useToast.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { Toast, ToastProps } from '~/components/Toast/Toast'
+
+type ToastBehaviourProps = Pick<
+  UseToastOptions,
+  'duration' | 'position' | 'render'
+>
+
+// onClose is provided by the chakra hook and should not be exposed to clients
+export type UseToastProps = Omit<ToastBehaviourProps & ToastProps, 'onClose'>
```
karrui, line 16: This can just be `UseToastOptions` too right? They take the same thing

seaerchin:
closing this PR because chromatic only runs on branches that are preefiexd with `form-v2`; will reopen another PR to ensure chromatic runs

PR #2233, opened 2021-06-22
feat(design-system): add Rating field component

### LINKED ISSUE #2011: [React] Add Field/RatingField component
**Time estimation: 2d**

Ensure a11y, lots of variants, including the new number system

karrui:
## Problem

This PR contains the new Rating field component. A11y has been checked using MacOS voiceover. 

Closes #2011 

## Solution

**Features**:

- Add `Rating` component in `components/Field/Rating`
- Add related stories

## Screenshots
In generated storybook.


mantariksh:
would be useful if you could rebase pls thank you!!

karrui:
> would be useful if you could rebase pls thank you!!

Done. Sorry for not catching the need to rebase earlier!

karrui:
@mantariksh i have not moved the style props for this component into theme yet, don't need review yet (unless you are already in the midst of doing it)

karrui:
@mantariksh i opted to leave the wrapper style props in the Rating component itself due to its responsive nature. Base styles cannot have responsive props... maybe we should have guidelines on when to use theme vs when to use style props :P

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -0,0 +1,186 @@
...
+    })
+
+    // Generate options from maxRating given.
+    const options = useMemo(
+      () => Array.from({ length: numberOfRatings }, (_, i) => i + 1),
+      [numberOfRatings],
+    )
+
+    const ratingLayout = useMemo(() => {
+      switch (variant) {
```
mantariksh, line 109: Screenshot 2021-07-06 at 4 36 08 PM

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -0,0 +1,186 @@
...
+    }, [variant])
+
+    return (
+      <Grid
+        rowGap="0.5rem"
+        templateAreas={[
+          `'caption' 'rating'`,
+          `'caption' 'rating'`,
+          `'caption' 'rating'`,
+          `'rating' 'caption'`,
```
mantariksh, line 124: tbh even after looking at the design idk why the order is reversed for the last row, could you add a comment to explain pls?

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -0,0 +1,186 @@
...
+        ]}
+      >
+        <Stack
+          gridArea="rating"
+          as="fieldset"
+          direction={['column', 'column', 'column', 'row']}
+          spacing={['0.5rem', '0.5rem', '0.5rem', '1rem']}
+          align={['flex-start', 'flex-start', 'flex-start', 'center']}
```
mantariksh, line 132: in general it seems like the last element for each one is unique, why is that the case?

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+export interface RatingProps {
+  /**
+   * The `name` attribute forwarded to each rating `radio` element
+   */
+  name: string
+  /**
+   * Number of rating options to render.
+   */
+  numberOfRatings: number
```
mantariksh, line 25: would `maxRating` be a more appropriate name?

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    const [currentValue, setCurrentValue] = useState<number | undefined>(
+      processDefaultValue(defaultValue),
+    )
+
+    // Call onChange everytime currentValue changes.
+    useEffect(() => {
+      onChange?.(currentValue)
+    }, [currentValue, onChange])
```
mantariksh, line 89: instead of relying on `setCurrentValue` to set the value and calling `onChange` as a side effect, why not declare a wrapper function which calls both `setCurrentValue` and `onChange`, and pass that as the `onChange` prop to `RatingOption`? doesn't that make the change process more explicit?

frontend/src/components/Field/Rating/RatingOption.tsx
```diff
@@ -0,0 +1,206 @@
...
+}
+
+export const RatingOption = forwardRef<RatingOptionProps, 'input'>(
+  (
+    { colorScheme = 'primary', name, onChange, selectedValue, value, variant },
+    ref,
+  ) => {
+    const handleSelect = useCallback(() => {
+      if (selectedValue === value) {
+        onChange?.(undefined)
```
mantariksh, line 145: maybe just add a comment here indicating that this is an "unselect"

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -0,0 +1,186 @@
...
+  wrapComponentsPerRow?: number
+  /**
+   * The value of the rating to be `checked` initially.
+   */
+  defaultValue?: number
+
+  /**
+   * Variant of rating field to render
+   */
+  variant: 'Heart' | 'Star' | 'Number'
```
mantariksh, line 48: we can consider making this lowercase so that it's not coupled to our formsg enums

frontend/src/components/Field/Rating/Rating.tsx
karrui: resolved irl

frontend/src/components/Field/Rating/Rating.tsx
karrui: I've thought about that naming, but `maxRating` may cause the user to infer that the component can render less than the maxRatings?

Instead, `numberOfRatings` explicitly declare the number of ratings declared. wdyt?

frontend/src/components/Field/Rating/Rating.tsx
karrui: you right

frontend/src/components/Field/Rating/Rating.tsx
mantariksh: sounds good

mantariksh (approved):

PR #2234, opened 2021-06-22
docs(script): add script to sync (has)AllowedEmailDomains

karrui:
## Problem

As mentioned in https://github.com/opengovsg/FormSG/pull/1698#issuecomment-864894628, we will be sunsetting the hasAllowedEmailDomains key in email form fields in 3 stages. This PR is stage 1 of 3.

This PR will handle
1. Correcting state inconsistency in DB that will arise when we switch from using `hasAllowedEmailDomains` as the source of truth to `allowedEmailDomains`.
2. After DB migration is complete, possible inconsistencies should be fixed.
  - all `hasAllowedEmailDomains = false` will have `allowedEmailDomains = []`
  - all `hasAllowedEmailDomain = true` must have non-empty `allowedEmailDomains`

See #1312.

## Solution

**Breaking Changes** 

  - DB migration, not backwards compatible, though will not have any issues on a roll back since there is still a source of truth.

## Deploy Notes

**New scripts**:

docs(script): add script to sync (has)AllowedEmailDomains

### DIFF
--- a/scripts/20210622_sync-allowedEmailDomains-with-hasAllowedEmailDomains/script.js
+++ b/scripts/20210622_sync-allowedEmailDomains-with-hasAllowedEmailDomains/script.js
@@ -1,0 +1,169 @@
+/* eslint-disable */
+
+// Optional: Retrieve forms with fields that has a populated allowedEmailDomains but has not allowed email domains.
+// For sampling to check if form has been fixed properly after migration.
+db.getCollection('forms').aggregate([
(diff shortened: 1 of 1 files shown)


karrui:
e2eeeeeeeeeeeeeeeeeeeeeeeee tests are eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

mantariksh (approved):

PR #2241, opened 2021-06-23
test: fix flaky form feedback test again

mantariksh:
## Problem

One of the unit tests for form feedback is still flaky despite many attempts to fix it (#1620, #2217).

## Solution

I finally realised that the array of form feedback returned is inherently non-deterministic. This is because form feedback with identical timestamps can be returned in any order, but the service being tested subsequently adds an `index` key depending on the order. This means there is no way for the test to exactly predict what the contents of the form feedback array should look like.

The solution was to omit the `index` key when checking for the shape of the form feedback array.


karrui (approved):
😢

PR #2246, opened 2021-06-24
feat(design-system): add Input, Textarea component and stories

### LINKED ISSUE #2002: [React] Add Field/Input component
**Time estimation: 1d**

karrui:
## Problem

This PR adds the `components/Input` component alongside with their stories. A Playground story is also created to check how it could work with React hook form.

Closes #2002 

## Solution

**Features**:

- add Input component and stories
- add Textarea component and stories

## Before & After Screenshots
In storybook


frontend/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const inputStyles = useMultiStyleConfig('Input', props)
+
+  // Omit extra props so they will not be passed into the DOM and trigger
+  // React warnings.
+  const inputProps = omit(props, ['isSuccess', 'isPrefilled'])
+
+  // Return normal input component if not success state.
+  if (!props.isSuccess) {
+    return <ChakraInput ref={ref} {...inputProps} sx={inputStyles.field} />
+  }
```
mantariksh, line 35: sorry why are we doing this?

karrui: As per the comments in the non-success case, `InputGroup` is required for `InputRightElement` to retrieve the correct style props and will crash if not included. 

However, since we just display the normal input on success, we do not need that special requirement and can just return the input as per normal

frontend/src/theme/components/Input.ts
```diff
@@ -1,23 +1,76 @@
...
-        h: '44px',
-        // textStyle: 'body-1',
-        borderColor: 'neutral.400',
-        border: '1px solid',
-        borderRadius: '4px',
-        _focus: {
-          boxShadow: 'none',
-          borderColor: 'primary.500',
-          borderWidth: '2px',
+    outline: (props: Record<string, any>) => {
```
mantariksh, line 18: lol

mantariksh (approved):

frontend/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    // InputGroup is required for InputRightElement to retrieve the correct
+    // style props. Will crash if not included.
+    <InputGroup>
+      <ChakraInput ref={ref} {...inputProps} sx={inputStyles.field} />
+      <InputRightElement sx={inputStyles.success}>
+        <Icon as={BxsCheckCircle} />
+      </InputRightElement>
+    </InputGroup>
```
seaerchin, line 45: Could i clarify on why we chose to do this? Testing on chakra's website seems to suggest that there are styling conflicts and consumers of our component might mistakenly think that this is a base `Input` in success state and compose it as part of an `InputGroup` only to discover that it's actually not.

Issue #1759, opened 2021-04-29
Visiting archived form URL shows uninformative error message, wrong form ID


mantariksh:
Visiting an archived form URL shows the following:

![Screenshot 2021-04-29 at 11 37 55 AM](https://user-images.githubusercontent.com/29480346/116499694-59a39c00-a8df-11eb-8b6b-18c5cf72fef5.png)

This has a couple of problems:
1. The "Gone" error message is uninformative. This is a regression because the error message used to be something like "This form has been deleted."
2. The target form ID is cut off, probably because of a bug in slicing the string.

syan-syan:
todo: change copy to 'This form is no longer active' and 'Form ID: (correct form id)'

aniruddha-adhikary:
<img width="889" alt="Screenshot 2021-07-07 at 1 31 57 AM" src="https://user-images.githubusercontent.com/932949/124643040-25da8900-dec3-11eb-9d48-455511e23d55.png">

This appears to be fixed.

aniruddha-adhikary:
#2320 fixes it

PR #2261, opened 2021-06-28
fix: correctly retrieve targetFormId for redirect state

### LINKED ISSUE #2254: Login sometimes fails even with correct OTP
To replicate:
1. Go to an admin form URL, e.g. form.gov.sg/\<formId\>/adminform without being signed in
2. You should be redirected to the sign in page
3. Enter the correct OTP
4. The "OTP verified!" success message is shown, but then will not subsequently redirect to the dashboard. User is then subsequently stuck; if they enter the OTP again it will show that the OTP is incorrect, but they have no other way to move forward. Refreshing will bring the user to the admin dashboard.

karrui:
## Problem

this fixes the bug where `FormErrorService.redirect`'s targetFormId was incorrectly passed the first 24 characters of the URL, which will fail as the API URLs have changed to be more semantic (first 24 digits is `api/v3/xxxxx` instead of the formId itself).

instead, take in the targetFormId directly as we already know the formId.
Closes #2254

## Solution

**Breaking Changes** 

    - Details ...

**Bug Fixes**:

- correctly retrieve targetFormId for redirect state 

## Before & After Screenshots

**BEFORE**:

Incorrect `targetFormId`
Screenshot 2021-06-28 at 11 31 44 AM

**AFTER**:

Correctly navigates back.

### Manual tests
Pre:
- [ ] Grab an URL of a form you own, and another URL for a form you don't own (probably can be rubbish formId)
Test:
- [ ] Clear cookies. Go to the URL of the form you own again. Should be kicked out to the login page. Sign in should redirect you correctly back into the view admin form page of that form.
- [ ] Clear cookies. Go to the URL of the form you don't own. Should be kicked out to an 404 error page, with a correct target form ID this time, e.g.


mantariksh (approved):
also addresses part of #1759

PR #2262, opened 2021-06-28
feat(verified-sms): disabling sms verifications

seaerchin:
## Problem
The backend needs a way to disable sms verifications easily. This should also be done in a single query, so that we avoid overloading the database with requests, which can be non-ideal under load. 

## Solution
This PR accomplishes the above through using an `updateMany` function defined on the schema. Some points are noted below for discussion

1. avoid updating timestamps on disabling. this is **intentional** and done to avoid confusion on the admin side. the reason for doing so is because admins expect their own actions to alter the last modified state of the form, and not something beyond their control. (informed @syan-syan on this already)
2. Returning `true` on service level - done because the service has no idea of the actual count it's supposed to update (and it should not). unable to return `IFormDocument` because `updateMany` from mongoose doesn't return the documents :(

**Breaking Changes** 

    - Details ...

### DIFF
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import { ObjectId } from 'bson-ext'
-import _, { cloneDeep, map, merge, omit, orderBy, pick } from 'lodash'
+import { cloneDeep, map, merge, omit, orderBy, pick, range } from 'lodash'
 import mongoose, { Types } from 'mongoose'
 
 import getFormModel, {
@@ -1682,7 +1682,7 @@ describe('Form Model', () => {
     describe('disableSmsVerificationsForUser', () => {
       it('should disable sms verifications for all forms belonging to a user successfully', async () => {
         // Arrange
-        const mockFormPromises = _.range(3).map(() => {
+        const mockFormPromises = range(3).map(() => {
           return Form.create({
             admin: populatedAdmin._id,
             responseMode: ResponseMode.Email,
@@ -1710,7 +1710,7 @@ describe('Form Model', () => {
 
       it('should not disable non mobile fields for a user', async () => {
         // Arrange
-        const mockFormPromises = _.range(3).map(() => {
+        const mockFormPromises = range(3).map(() => {
           return Form.create({
             admin: populatedAdmin._id,
             responseMode: ResponseMode.Email,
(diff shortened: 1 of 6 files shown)


src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import { ObjectId } from 'bson-ext'
-import { cloneDeep, map, merge, omit, orderBy, pick } from 'lodash'
+import _, { cloneDeep, map, merge, omit, orderBy, pick } from 'lodash'
```
karrui, line 3: nit: stick to either destructured imports since the rest are already done that way? or change everything to `_.`

src/app/models/form.server.model.ts
```diff
@@ -751,6 +751,26 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     ).exec()
   }
 
+  FormSchema.statics.disableSmsVerificationsForUser = async function (
+    userId: IUserSchema['_id'],
+  ) {
+    return this.updateMany(
+      // Filter the collection so that only speciified user is selected
```
karrui, line 758: nit: typo in specified

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1044,3 +1044,28 @@ export const updateStartPage = (
...
+ * Disables sms verifications for all forms belonging to the specified user
+ * @param userId the id of the user whose sms verifications should be disabled
+ * @returns ok(true) when the forms have been successfully disabled
+ * @returns err(PossibleDatabaseError) when an error occurred while attempting to disable sms verifications
+ */
+export const disableSmsVerificationsForUser = (
+  userId: string,
+): ResultAsync<true, PossibleDatabaseError> =>
+  ResultAsync.fromPromise(
+    FormModel.disableSmsVerificationsForUser(userId).then(() => true),
```
karrui, line 1058: should the `.then` part be chained from ResultAsync instead?

```ts
ResultAsync.fromPromise(
  FormModel.disableSmsverificationsForUser(userId),
  (error) => { ... }
).map(() => true)
```

karrui (approved):
lgtm with nits

src/app/models/__tests__/form.server.model.spec.ts
seaerchin: gotcha, will switch to named imports

src/app/modules/form/admin-form/admin-form.service.ts
seaerchin: thanks for suggestion, will change!

PR #2263, opened 2021-06-28
fix: allow creation of storage form when emails are invalid

### LINKED ISSUE #2189: Form creation modal is disabled when switching response mode
To reproduce:
1. Click on "Create new form" and select Email mode
2. Delete all the emails under "Emails where responses will be sent"
3. Select Storage mode
4. Observe that "Get Started" button is disabled even though inputs for storage mode are valid

mantariksh:
## Problem

Closes #2189 

## Solution

- https://github.com/opengovsg/FormSG/commit/dd7bd161a89beb93884156749f63b0f924e8268d: refactor the form emails input to use `ng-message` with a validation directive instead of manually validating on `ng-change`
- https://github.com/opengovsg/FormSG/commit/aa4d6423968336c401616fe034d408960e7f34a1: update the validation directive to return true for all validators as long as the response mode is not `'email'`
- https://github.com/opengovsg/FormSG/pull/2263/commits/f55660530e269cac30b148b55f316e221bef5824: update Joi validation on all form creation (including duplication and copy) endpoints to allow empty strings or arrays for the `emails` key if the response mode is not email mode

## Breaking changes

## Manual tests
- [ ] Go to the create form modal and select "Email mode". Make the email list input invalid, e.g. by deleting all content in the input. Switch back to "Storage mode". You should still be able to create the form as long as the form title is valid. Make sure to go all the way through and create the form.
- [ ] Select "Storage mode" and type an invalid form title. You should be prevented from creating the form.
- [ ] Select "Email mode" and ensure that you are prevented from creating the form with the following types of inputs for the email list, with the correct error message in each case:
    - No input
    - Any invalid email present
    - Duplicate emails present
    - More than 30 non-duplicate emails present
- [ ] Repeat previous test for email list in the Settings tab for an email mode form. If the input is invalid, it should revert to its previous value on blur.

### DIFF
--- a/src/public/modules/forms/admin/directives/validate-form-emails-input.directive.js
+++ b/src/public/modules/forms/admin/directives/validate-form-emails-input.directive.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const validator = require('validator')
+const get = require('lodash/get')
 
 angular
   .module('forms')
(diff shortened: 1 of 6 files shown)


src/public/modules/forms/admin/directives/validate-form-emails-input.directive.js
```diff
@@ -0,0 +1,39 @@
...
+
+function validateFormEmailsInput() {
+  return {
+    restrict: 'A',
+    require: 'ngModel',
+    link: (scope, _elem, _attr, ngModel) => {
+      ngModel.$validators.validEmails = (rawInput) => {
+        const emails = String(rawInput).split(',')
+        return (
+          get(scope, 'vm.formData.responseMode') !== 'email' ||
```
karrui, line 18: not sure about this, but since `get(scope, 'vm.formData.responseMode') !== 'email'` is repeated, can we initiate it once outside of the validators, or do we need them at the point of evaluation?

karrui (approved):
lgtm with 1 q

src/public/modules/forms/admin/directives/validate-form-emails-input.directive.js
mantariksh: haha I didn't initiate outside because I wasn't sure, but let me try this now

mantariksh: edit: we can fix this in React migration

PR #2265, opened 2021-06-28
feat(design-system): add checkbox and radio components and stories

### LINKED ISSUE #2008: [React] Add Field/Checkbox & Radio component

chowyiyin:
## Problem
This PR implements the checkbox and radio component.

Closes #2008

## Solution
The compound components pattern was adopted to avoid prop drilling and also inverts the control, giving clients more freedom in the way they use the checkbox component. The `Others` component was also extracted as they work the same way for both `Checkbox` and `Radio`. Note that the base components (i.e. `Checkbox` and `Radio` components) are generated in the `Others` component. This is because `React.cloneElement(child, { ref: 'newRef' })` overrides the original `ref` and hence, it is still not possible for two parents to have a `ref` to the same child. Using `cloneElement` therefore does not produce the desired result when trying to pass the merged `ref` to `Checkbox` or `Radio` component.

## Screenshots
In Storybook


karrui:
> This is because `React.cloneElement(child, { ref: 'newRef' })` overrides the original `ref`

Is it not possible to use the `useMergedRefs` hook to bypass this?

frontend/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+  value: 'Option',
+}
+
+export const Group: Story<CheckboxProps> = (args) => {
+  return (
+    <VStack align="left">
+      <Checkbox value="Option 1" />
+      <Checkbox value="Option 2" />
+      <Checkbox value="Option 3" />
+      <Others {...args} value="Others" base="checkbox">
```
karrui, line 33: I would actually prefer duplicating code instead of having the `base` parameter so the code remains isolated and we know that it works for that specific field

chowyiyin: Yep I agree. If for some reason we want to include the "Other" option for any other fields as well, this `Other` component may also not work well for them as the Chakra UI implementation of different base component may be different too. Will change!

frontend/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { Meta, Story } from '@storybook/react'
+import { omit } from 'lodash'
+
+import Button from '~components/Button'
+
+import { Checkbox } from './Checkbox'
+import { CheckboxOthers } from './CheckboxOthers'
+
+export default {
+  title: 'Components/Checkbox',
```
karrui, line 17: should the story be nested in `Components/Fields/Checkbox` for consistency?

frontend/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+    <form onSubmit={handleSubmit(onSubmit)} noValidate>
+      <FormControl
+        isRequired={isRequired}
+        isDisabled={isDisabled}
+        isInvalid={!!errors[name]}
+        mb={6}
+      >
+        <FormLabel htmlFor={name}>{label}</FormLabel>
+        <VStack align="left">
+          <CheckboxGroup {...omit(field, 'ref')}>
```
karrui, line 73: why is ref omitted here?

frontend/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,13 @@
+import {
+  Checkbox as ChakraCheckbox,
+  CheckboxProps,
+  forwardRef,
+} from '@chakra-ui/react'
+
+export const Checkbox = forwardRef<CheckboxProps, 'input'>((props, ref) => {
+  return (
+    <ChakraCheckbox {...props} ref={ref}>
+      {props.value}
```
karrui, line 10: Shouldn't we pass in props.children instead?

karrui: and if that's the case, we can use Checkbox directly, from Chakra without doing anything more right?

```ts
export const Checkbox = ChakraCheckbox
```

frontend/src/components/Checkbox/CheckboxOthers.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const mergedRef = useMergeRefs(ref, inputRef)
+
+    const handleInputChange = () => {
+      if (!inputRef?.current?.checked) {
+        inputRef?.current?.click()
+      }
+    }
+
+    return (
+      <OthersProvider value={{ onInputChange: handleInputChange }}>
```
karrui, line 26: we probably don't need `OthersProvider` for this, just pass in `onInputChange` directly to `OthersWrapper` :P

```ts
<OthersWrapper onInputChange={handleInputChange}>
  ...
</OthersWrapper>
```

and have OthersWrapper take in an onInputChange prop in addition to other props

frontend/src/components/Checkbox/Checkbox.stories.tsx
chowyiyin: `CheckboxGroup` doesn't have `forwardRef` so passing in a ref causes an error.

frontend/src/components/Checkbox/Checkbox.tsx
chowyiyin: I wanted the way we specify the value/child of `Checkbox` to be consistent for `Checkbox` and `CheckboxOthers`. For `CheckboxOthers`, it will be difficult to distinguish the text from the others component if the value is passed in through the child so I standardised it by always using the "value" prop. There's a tradeoff here though, because we can't decouple the submission value and the value that is displayed. Let me know if there's a better way of doing this :)

karrui: Its not mutually exclusive right? The checkbox just needs a `value` prop, the child does not matter afaik.

e.g.

```ts
<Checkbox value="a">
  Can be anything here
</Checkbox>
```

Also doesn't matter for checkbox others right?

```ts
```

frontend/src/components/Checkbox/Checkbox.stories.tsx
karrui: Then where is the field's ref being passed into? This would be used to focus on the field when any errors occur. 

Maybe we don't even need to use a controller for this. We can just use `register` directly on Checkbox and Checkbox others

frontend/src/components/Checkbox/CheckboxOthers.tsx
```diff
@@ -0,0 +1,62 @@
...
+const OthersWrapper = ({
+  children,
+  onInputChange,
+  onClick,
+}: {
+  children: React.ReactNode
+  onInputChange: () => void
+  onClick?: (e: ChangeEvent<HTMLInputElement>) => void
+}) => {
+  const style = useMultiStyleConfig('Checkbox', {}).others
```
karrui, line 48: nit: prefer not assigning with `.others` so it is obvious that the style being used is `style.others`

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
+import { ComponentStyleConfig } from '@chakra-ui/react'
+import { getColor } from '@chakra-ui/theme-tools'
+
+export const Checkbox: ComponentStyleConfig = {
+  parts: ['others'],
+  baseStyle: ({ theme }) => ({
+    control: {
+      borderRadius: '0.19rem',
```
karrui, line 8: Any reason for this weird rem value? Looked at the figma and it's a 2px radius (i.e 0.125rem)

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+      borderColor: 'primary.500',
+      _checked: {
+        bg: 'primary.500',
+        borderColor: 'primary.500',
+        _hover: {
+          bg: 'primary.500',
+          borderColor: 'primary.500',
+        },
+      },
+      _focus: {
```
karrui, line 19: `_disabled` styling seems to be missing. Figma seems to not have the _disabled:_check state, so maybe ask @pearlyong  about it

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+        boxShadow: 'none',
+      },
+    },
+    label: {
+      textColor: 'secondary.700',
+      mx: '1rem',
```
karrui, line 25: I think this is supposed to be `ml`? There does not seem to be a right margin on figma

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    container: {
+      h: '44px',
```
karrui, line 28: nit: Should this be in size? Probably no need since we only have a single size for now, but good to think about?

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+      _hover: {
+        bg: 'primary.100',
+      },
+      _focusWithin: {
+        borderColor: 'primary.500',
+        boxShadow: `0 0 0 2px ${getColor(theme, 'primary.500')}`,
+      },
+    },
+    others: {
+      pl: '48px',
```
karrui, line 41: nit: use rem

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+      mt: '2px',
```
karrui, line 42: Figma does not seem to have margin, any reason why this is here?

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+  }),
+  sizes: {
+    // override md size, which is the default size
+    md: {
+      control: { w: '1.5rem', h: '1.5rem' },
+      label: { textStyle: 'body-1' },
```
karrui, line 49: Don't think textStyle belongs in sizes

frontend/src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+      icon: { fontSize: '0.75rem' },
```
karrui, line 50: Actually I just realized that we need to override Chakra's default icon with the icons used in our design system (use `boxicon`)

See [ChakraUI's docs](https://chakra-ui.com/docs/form/checkbox#icon)

karrui (commented):
sorry for the delay in reviewing!

frontend/src/components/Checkbox/CheckboxOthers.tsx
```diff
@@ -0,0 +1,62 @@
+import { ChangeEvent, cloneElement, isValidElement, useRef } from 'react'
```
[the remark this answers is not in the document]
chowyiyin: We can’t do `Checkbox.Others` because `Checkbox` is of type `ComponentWithAs<CheckboxProps, ‘input’> (because we use `forwardRef`) I found a suggested solution to introduce a type `ComponentWithAs<CheckboxProps, ‘input’> & {Others?: typeof Others}` but this doesn’t really work because `Checkbox.Others` could be undefined. I think a workaround is to export an object `{ CheckboxOption: Checkbox, CheckboxOthers: Others }`, which we can then use as `Checkbox.CheckboxOption` and `Checkbox.CheckboxOthers`. Wdyt?

frontend/src/theme/components/Checkbox.ts
chowyiyin: have added styles for this but I need a little help with styling cursor for disabled states for radio containers! im trying to set the cursor to 'default' when disabled but the styles aren't applying

frontend/src/components/Checkbox/CheckboxOthers.tsx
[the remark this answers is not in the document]
karrui: Possible to sidestep with a typescript cast: https://stackoverflow.com/a/63454836

frontend/src/theme/components/Checkbox.ts
karrui: the cursor should be `not-allowed`

frontend/src/components/Field/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,80 @@
...
+  useMultiStyleConfig,
+} from '@chakra-ui/react'
+
+type CheckboxComponent = ComponentWithAs<'input', CheckboxProps> & {
+  Others: typeof CheckboxOthers
+}
+
+// forward ref to be consistent with chakra's checkbox
+export const Checkbox = forwardRef<CheckboxProps, 'input'>((props, ref) => {
+  return <ChakraCheckbox ref={ref} {...props} icon={<Icon as={BiCheck} />} />
```
karrui, line 27: Changing the icon may have affected the look and size and animation of the check.
Chakra has an [animated checkbox component](https://github.com/chakra-ui/chakra-ui/blob/main/packages/checkbox/src/checkbox-icon.tsx), we might want to do the same for our replaced icon. 
(However may not be worth it, I think the Chakra default checkbox looks fine as is. Double check with pearly to see if we really need to use `BiCheck`?) 

If not I think we can revert instead of spending time to get the check to look and feel as before (with BiCheck)

frontend/src/theme/components/Checkbox.ts
karrui: disabled + check state seems to be wrong too, should add stories for all of the combinations.

karrui (changes requested):
radio looks fine, some more work on the checkbox side

pearlyong:
@chowyiyin replying here since i can't comment on this in Chromatic! one fix - on overflow can it wrap to next line like this? https://www.figma.com/file/u00dlQLsaQfvqv6ot5ZrJZ/Form-Design-System?node-id=1843%3A780

mantariksh:
Closing in favour of #2525 and its successors

PR #2266, opened 2021-06-28
chore(db): db script to create index on smscounts

seaerchin:
## Problem
querying smscounts is slow, even when we index on `formAdmin` and `msgSrvcId`. This PR aims to solve that by creating a boolean key to index on, so that we can speed up the query.

## Solution
Creates a boolean key in `smscounts` that tracks whether the `msgSrvcSid` used in the verification is formSG's own.

### DIFF
--- a/scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
+++ b/scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
@@ -51,7 +51,7 @@ db.getCollection('smscounts').updateMany(
 // Count of forms using our twilio acc
 db.getCollection('smscounts').count({
   smsType: 'VERIFICATION',
-msgSrvcSid: {$eq: formTwilioId},
+  msgSrvcSid: { $eq: formTwilioId },
 })
 
 // Count of forms using their own twilio acc
--- a/scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
+++ b/scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
@@ -1,0 +1,45 @@
+/* eslint-disable */
+
+/*
+This script creates a new key, isOnboardedAccount, on the smscounts db which indexes the msgSrvcId variable
+*/
+
+let formTwilioId = 'insert the form twilio id here'
+
+// == PRE-UPDATE CHECKS ==
+// Count the number of verifications
+db.getCollection('smscounts').count({ smsType: 'VERIFICATION' })
+
+// == UPDATE ==
+// Update verifications which have message service id equal to form twilio id
+db.getCollection('smscounts').updateMany(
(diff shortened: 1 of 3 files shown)


scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
```diff
@@ -0,0 +1,45 @@
...
+
+/*
+This script creates a new key, isOnboardedAccount, on the smscounts db which indexes the msgSrvcId variable
+*/
+
+let formTwilioId = 'insert the form twilio id here'
+
+// == PRE-UPDATE CHECKS ==
+// Count the number of verifications
+db.getCollection('smscounts').count({ smsType: 'VERIFICATION' })
```
karrui, line 11: we should get a count of the documents with/without our own twilio ID too, so we can check if the updated amounts are exactly the same too

src/app/services/sms/sms_count.server.model.ts
```diff
@@ -34,6 +34,9 @@ const VerificationSmsCountSchema = new Schema<IVerificationSmsCountSchema>({
       required: true,
```
karrui, line 34: this PR should be purely the script and not contain anything else.

I think adding a new schema key (and possibly the programatic setting of whether `isOnboardedAccount`) should be done in a separate PR.

karrui (changes requested):

src/app/services/sms/sms_count.server.model.ts
seaerchin: agreed, will remove this and add it in the follow up PR

scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
seaerchin: sure, will do

scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
```diff
@@ -0,0 +1,61 @@
...
+let formTwilioId = 'insert the form twilio id here'
+
+// == PRE-UPDATE CHECKS ==
+// Count the number of verifications (A)
+db.getCollection('smscounts').count({ smsType: 'VERIFICATION' })
+
+// Count of forms using our twilio acc (B)
+db.getCollection('smscounts').count({
+  smsType: 'VERIFICATION',
+msgSrvcSid: {$eq: formTwilioId},
```
karrui, line 16: nit: spacing

karrui (approved):
lgtm

scripts/20210628_create-isOnboardedAcc/create-isOnboardedAcc.js
seaerchin: thx 4 catch

PR #2267, opened 2021-06-28
feat(design-system): add NumberInput component and stories

### LINKED ISSUE #2004: [React] Add Field/Number component
**Time estimation: 1d**

Blocked by #2002, might want to be attempted by the same person

karrui:
> This PR builds on the input component PR #2246 

## Problem

This PR adds the `NumberInput` component, as well as stories depicting its various states and how to use it with `react-hook-form`.

Closes #2004 

## Solution

**Features**:

- feat(theme): add NumberInput theme that extends from Input theme
- feat: add NumberInput component
- feat: add NumberInput stories

## Before & After Screenshots
In generated storybook.


mantariksh (approved):

frontend/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+  isSuccess?: boolean
+}
+
+export const NumberInput = forwardRef<NumberInputProps, 'input'>(
+  (props, ref) => {
+    const { isSuccess } = props
+    const inputStyles = useMultiStyleConfig('NumberInput', props)
+    // Omit extra props so they will not be passed into the DOM and trigger
+    // React warnings.
+    const inputProps = omit(props, ['isSuccess', 'isPrefilled'])
```
seaerchin, line 35: would it be better to remove the prop totally as the `isPrefilled` prop is omitted here and not used anywhere

frontend/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <ChakraNumberInput {...inputProps} ref={ref}>
+        <NumberInputField sx={inputStyles.field} />
+        {isSuccess && (
+          <InputRightElement sx={inputStyles.success}>
+            <Icon as={BxsCheckCircle} />
+          </InputRightElement>
+        )}
+        <NumberInputStepper>
+          <NumberIncrementStepper>
+            <Icon as={BiCaretUp} transform="translate(0, 3px)" />
```
seaerchin, line 47: would it be better to convert the px to rem?

frontend/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+          </NumberIncrementStepper>
+          <NumberDecrementStepper>
+            <Icon as={BiCaretDown} transform="translate(0, -3px)" />
```
seaerchin, line 50: would it be better to convert the px to rem?

seaerchin:
i saw that there was a [lock on disabled ](https://www.figma.com/file/u00dlQLsaQfvqv6ot5ZrJZ/Form-Design-System?node-id=292%3A674) but it's not present in the storybook - should we add it in?

karrui:
Design changes, moving to draft again

frontend/src/components/NumberInput/NumberInput.tsx
karrui: inside the theme :P

frontend/src/components/NumberInput/NumberInput.tsx
karrui: not for adhoc changes like this, imagine 3px is 0.1875rem :P

karrui: This is moot anyways since the design is changing lul

frontend/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const incProps = getIncrementButtonProps()
+    const decProps = getDecrementButtonProps()
+
+    const inputEndPadding = showSteppers
+      ? stepperWrapperRef.current?.offsetWidth
+      : undefined
+
+    return (
+      <Box {...htmlProps} ref={ref} __css={styles.root}>
+        <chakra.input
```
mantariksh, line 68: why are we using chakra's internal `input` here?

frontend/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    const stepperWrapperRef = useRef<HTMLDivElement | null>(null)
+
+    const controlProps = useFormControlProps(props)
+    const {
+      htmlProps,
+      getInputProps,
+      getIncrementButtonProps,
+      getDecrementButtonProps,
+    } = useNumberInput({ ...controlProps, clampValueOnBlur })
```
mantariksh, line 56: I find it EXTREMELY frustrating that `chakra` doesn't have any documentation on this. how do you know to use `useFormControlProps`? how dyou know what args it takes, and that the return value can be passed back into `useNumberInput`?

karrui: I looked at the source code for their own NumberInput implementation.

Reason for using `useFormControlProps` is so that this component can be wrapped with a `FormControl` component and be able to access their relevant props, such as whether the field is disabled, readOnly, invalid, etc.

frontend/src/components/NumberInput/NumberInput.tsx
karrui: the `Input` component performs more transformation of `inputProps`, removing some important variables such as `aria-invalid`, etc. Resulting in some styles not being provided. I'll add a comment

karrui: Possible to patch upstream, i will open an issue if I have time....

frontend/src/components/NumberInput/NumberInput.tsx
karrui: Possible to patch upstream, i will open an issue if I have time....

mantariksh (approved):

PR #2276, opened 2021-06-30
feat(sms-limiting): sms counts retrieval

seaerchin:
## Problem
Part of allowing sms limiting is to have a way to count the number of free sms that an agency user has sent out thus far. Because the frontend also requires this count to determine the appropriate state on the modal, there is also a new endpoint exposed in this PR

## Solution
Relevant technical overview (per the design doc) is [here](https://docs.google.com/document/d/12z_2ZcSV-HiPCYVUmMtwCQC0CBMWCBfTPdhDL_JmmqY/edit#heading=h.uqcwwe63dqzz)

Some relevant technical choices are listed below for discussion
1. `isOnboardedAccount` set using a pre-save hook. This is done so that the updating/setting is automatic, and developers don't have to be concerned with this in the future.  
2. Exposing the endpoint to count is done on `SmsController` - decided on this because other controllers don't seem too concerned with the sms counts. Arguably, this could be done on `AdminFormController` or `UserController` but those seem to be more concerned with the properties of the form/user respectively. (am ok w/ changing if needed tho)

**Breaking Changes** 

    - Details ...

**Manual Tests** 
Because this endpoint (and method) will be used subsequently to determine whether the form should be disabled or not on the frontend, this needs to be tested for speed.

the test plan is as follows on staging: 

1. Populate the staging db with 2.5k sms counts. 
2. next, query the endpoint at `/:formId([a-fA-F0-9]{24})/verified-sms/count/free`
3. log the time taken for this operation
4. Repeat the above 3 steps for the various warning limits (2.5k, 5k, 7.5k; 10k for disabled, with different emails)


karrui:
> isOnboardedAccount set using a pre-save hook. This is done so that the updating/setting is automatic, and developers don't have to be concerned with this in the future.

Just take note that our current way of manually onboarding users will not trigger this pre-save hook; will need to add another manual check when we do so until we have an automated way to do this on the GUI

seaerchin:
@karrui if my understanding is correct, don't we save the credentials in AWS secrets manager and subsequently, update all forms of the specified user to have the specified message service id? if that's the case, wouldn't this check work because we are checking the form's `msgsrvcid`, which would already have been updated?

karrui:
> @karrui could you help me understand this? my understanding is that we save the credentials in AWS secrets manager and subsequently, update all forms of the specified user to have the specified message service id. if that's the case, wouldn't this check work because we are checking the form's `msgsrvcid`, which would already have been updated?

Pre-save hooks don't run when we update using an external GUI such as Robo3T

src/app/routes/api/v3/sms/sms.routes.ts
```diff
@@ -0,0 +1,9 @@
+import { Router } from 'express'
```
karrui, line 1: I don't think we should add a new `/sms` endpoint just for this. Would the endpoint not be better suited in `/api/v3/forms/:formId/verified-sms/count/free`?

tests/unit/backend/helpers/jest-db.ts
```diff
@@ -222,13 +230,45 @@ const getFullFormById = async (
 ): Promise<IPopulatedForm | null> =>
   await getEmailFormModel(mongoose).getFullFormById(formId)
 
+const insertVerifiedSms = async ({
```
karrui, line 233: i don't think this needs to be a dbHandler function, should it just be exposed as a test function?

src/app/services/sms/sms_count.server.model.ts
```diff
@@ -34,8 +34,20 @@ const VerificationSmsCountSchema = new Schema<IVerificationSmsCountSchema>({
...
   },
+  isOnboardedAccount: {
+    type: Boolean,
+  },
 })
 
+VerificationSmsCountSchema.pre<IVerificationSmsCountSchema>(
+  'save',
+  function (next) {
+    const formTwilioId = process.env.TWILIO_MESSAGING_SERVICE_SID
```
karrui, line 45: we should be using the variable in `config`, not directly on process.env

karrui: This also allows for mocking during tests

tests/unit/backend/helpers/jest-db.ts
```diff
@@ -222,13 +230,45 @@ const getFullFormById = async (
...
+  isOnboarded?: boolean
+}): Promise<ISmsCountSchema> => {
+  const SmsModel = getSmsCountModel(mongoose)
+  const sms = await SmsModel.create({
+    smsType: SmsType.Verification,
+    logType,
+    msgSrvcSid,
+    form: formId,
+    formAdmin,
+    isOnboarded,
```
karrui, line 253: Should is `isOnboarded` be skipped in the create parameter? Seems to do nothing. Was this put in erroneously?

tests/unit/backend/helpers/jest-db.ts
```diff
@@ -222,13 +230,45 @@ const getFullFormById = async (
...
+  })
+
+  // Workaround because the isOnboarded property is set onSave.
```
karrui, line 256: If put this function in the model tests directly, and mock `config`'s esrvcMsgId or something to be the `msgSrvcSid` we use to create, then we should get the default behaviour without needed this workaround.

src/app/services/sms/__tests__/sms.controller.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  let mockUser: IUserSchema
+  let mockForm: IFormSchema
+  const VERIFICATION_SMS_COUNT = 3
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  afterAll(async () => {
+    await dbHandler.clearDatabase()
```
karrui, line 25: I don't think we should be calling the db layer in controller tests. Should we mock the db models instead?

src/app/services/sms/__tests__/sms.service.spec.ts
```diff
@@ -351,6 +352,88 @@ describe('sms.service', () => {
     })
   })
 
+  describe('retrieveFreeSmsCounts', () => {
+    let mockOtpData: FormOtpData
+    let mockForm: IFormSchema
+    const VERIFICATION_SMS_COUNT = 3
+
+    beforeEach(async () => {
+      mockForm = await FormModel.create({
```
karrui, line 361: database layer should be mocked in service tests too

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
```diff
@@ -278,7 +284,40 @@ describe('SmsCount', () => {
...
+    it('should save successfully and set isOnboarded to false when sms is sent using default credentials', async () => {
+      // Arrange
+      // NOTE: As the isOnboardedAccount property is set based on TWILIO_MESSAGING_SERVICE_SID,
+      // even if the key is absent, this will still pass
+      const verificationParams = merge(
+        createVerificationSmsCountParams({
+          logType: LogType.success,
+          smsType: SmsType.Verification,
+        }),
+        { msgSrvcSid: process.env.TWILIO_MESSAGING_SERVICE_SID },
```
karrui, line 299: can we mock config etc here so the test is not testing nothing

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
```diff
@@ -278,7 +284,40 @@ describe('SmsCount', () => {
         'createdAt',
         '__v',
       ])
-      expect(actualSavedObject).toEqual(omit(smsCountParamsWithExtra, 'extra'))
+      expect(actualSavedObject).toEqual(expected)
+    })
+
+    it('should save successfully and set isOnboarded to false when sms is sent using default credentials', async () => {
```
karrui, line 290: should we have test for isOnboarded being true

src/app/services/sms/sms.controller.ts
```diff
@@ -0,0 +1,72 @@
...
+import * as SmsService from '../../services/sms/sms.service'
+import { createReqMeta } from '../../utils/request'
+
+import { SmsCountsMeta } from './sms.types'
+import { mapRouteError } from './sms.util'
+
+const logger = createLoggerWithLabel(module)
+
+/**
+ * Private handler to retrieve the free sms counts remaining for a user and a form belonging to the user
```
karrui, line 18: nit: should it not be the free sms counts remaining for a form by querying counts remaining for the admin?

src/app/services/sms/sms.controller.ts
```diff
@@ -0,0 +1,72 @@
...
+ * This is the controller for GET /:userId/:formId endpoint
```
karrui, line 19: the endpoint is wrong

src/app/services/sms/sms.controller.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  // Step 1: Check that the form exists
+  return (
+    FormService.retrieveFormById(formId)
+      // Step 2: Retrieve the free sms count
+      .andThen(({ msgSrvcName, admin }) => {
+        // This casting is required because no type parameter is specified on the document.
+        // Hence, this defaults to any but on the schema, this is defined as an ObjectId
+        const adminId = (admin as Types.ObjectId).toHexString()
```
karrui, line 45: can just call `String(admin)` for safety instead of casting. The `String` constructor will attempt to call the object's `toString()` method

src/app/services/sms/sms.controller.ts
```diff
@@ -0,0 +1,72 @@
...
+        return SmsService.retrieveFreeSmsCounts(adminId).map(
+          (freeSmsCount) => ({
+            msgSrvcSid: msgSrvcName,
```
karrui, line 48: This looks wrong to me. We are mapping the form's msgSrvcName to free sms counts remaining? What if the form has its own onboarded service name?

src/app/services/sms/sms.controller.ts
```diff
@@ -0,0 +1,72 @@
...
+            freeSmsCount,
```
karrui, line 49: nit: should we rename this to `freeSmsCountsRemaining`, this is very vague as to what the count is

src/app/services/sms/sms.service.ts
```diff
@@ -461,3 +465,32 @@ export const sendBouncedSubmissionSms = (
...
+ */
+export const retrieveFreeSmsCounts = (
+  userId: string,
+): ResultAsync<number, PossibleDatabaseError> => {
+  return ResultAsync.fromPromise(
+    SmsCount.countDocuments({
+      'formAdmin.userId': userId,
+      smsType: SmsType.Verification,
+      isOnboardedAccount: false,
+    }).exec(),
```
karrui, line 483: should this be wrapped as a SmsCount's model instance method instead?

karrui (changes requested):

src/app/routes/api/v3/sms/sms.routes.ts
seaerchin: sure, i just thought that it made more sense to be in sms cos the `form` or `/admin/form` endpoints were concerned with form properties but i'll change it! 

will change to `/admin/form` because it's regarding privileged information and also because ti should be protected by authentication (checked wtih sonjia on this)

src/app/services/sms/sms_count.server.model.ts
seaerchin: great point, i'll change accordingly!

tests/unit/backend/helpers/jest-db.ts
seaerchin: made a mistake here, will change in the create portion! thanks for the catch!

src/app/services/sms/sms.controller.ts
seaerchin: good point, will change!

src/app/services/sms/sms.service.ts
seaerchin: TIL: statics define a method on the model rather than on the schema

tests/unit/backend/helpers/jest-db.ts
seaerchin: yep, will change as idt it'll be reused in integration tests as we alreday have service methods to perform such functions. thanks for suggestion!

seaerchin:
i think i'll change this so that the endpoint is rooted at `/api/v3/admin/forms/:formId/verified-sms/count/free`. However, not going to change the controller to sms controller due to point #2 in the PR write up (unless you think it should change also) 

also, i'll rebase this off `develop` (sorry kr) because i didn't take into account `config` initially, and it's been changed due to feature manager removal.

src/app/services/sms/sms.controller.ts
seaerchin: paraphrasing what you told me: this endpoint has only 1 role and it should not expose more information (msgSrvcId) than is necessary. rather, it should only return the count and let the frontend obtain msgSrvcId from the form's properties

tests/unit/backend/helpers/jest-db.ts
seaerchin: i removed this function because when we mock the behaviour of stuff 1 level below, we don't require this function anymore.

src/app/services/sms/__tests__/sms.controller.spec.ts
seaerchin: mocked the service layer a level below instead but yea, you're right! paraphrasing what you told me, even though we don't mock here, our tests are brittle because we are relying on behaviour that is external to this function for our tests. 

what should happen si that we mock the function's dependencies then see if the results are what we expect

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
seaerchin: could i clarify this point - i edited the [initial test](https://github.com/opengovsg/FormSG/pull/2276/files#diff-8bc32fb6c3b559254ada4bcadce8e406624db2a61d3f24d5e85cf93b36ae662aR266-R287) so that `isOnboarded` is tested as being part of the success case. were you thinking of something else?

src/app/services/sms/sms.controller.ts
seaerchin: changed because your description is better

src/app/services/sms/sms.controller.ts
seaerchin: i shifted the return type for the controller to be in `/api` and renamed this; however, the dto itself has a general naming because it's supposed to just represent a transfer of `smsCount` and what this count specifically refers to should be determined by the caller

src/app/routes/api/v3/admin/forms/admin-forms.form.routes.ts
```diff
@@ -197,3 +198,17 @@ AdminFormsFormRouter.put(
...
+ * @security session
+ *
+ * @returns 200 with the free sms counts
+ * @returns 401 when user does not exist in session
+ * @returns 404 when the formId is not found in the database
+ * @returns 500 when a database error occurs during retrieval
+ */
+AdminFormsFormRouter.get(
+  '/:formId([a-fA-F0-9]{24})/verified-sms/count/free',
+  SmsController.handleGetFreeSmsCountForFormAdmin,
```
karrui, line 213: Now that it's nested in AdminForm router, should the controller function be in AdminFormController instead?

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
```diff
@@ -230,9 +231,22 @@ describe('SmsCount', () => {
...
+
+    afterAll(() => {
+      smsConfig.twilioMsgSrvcSid = twilioMsgSrvcSid
+    })
+
     it('should create and save successfully', async () => {
       // Arrange
       const smsCountParams = createVerificationSmsCountParams()
+      const expected = merge(smsCountParams, {
+        isOnboardedAccount: true,
```
karrui, line 248: Why would this be true? Do you think this test should test for the default state, which is `isOnboardedAccount === false`

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
```diff
@@ -260,6 +274,9 @@ describe('SmsCount', () => {
           extra: 'somethingExtra',
         },
       )
+      const expected = merge(omit(smsCountParamsWithExtra, 'extra'), {
+        isOnboardedAccount: true,
```
karrui, line 278: same here

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
karrui: I think we should separate the tests into default -> false, custom -> true tests

src/app/services/sms/sms.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+  // Step 1: Check that the form exists
+  return (
+    FormService.retrieveFormById(formId)
+      // Step 2: Retrieve the free sms count
+      .andThen(({ admin }) => {
+        return SmsService.retrieveFreeSmsCounts(String(admin))
+      })
+      // Step 3: Map/MapErr accordingly
+      .map((freeSmsCountForAdmin) =>
+        res.status(StatusCodes.OK).json({ smsCounts: freeSmsCountForAdmin }),
```
karrui, line 44: nit: given the specificity of the endpoint, could just returning a number suffice? i.e.

`res...json(freeSmsCountForAdmin)`

src/app/services/sms/sms.util.ts
```diff
@@ -18,3 +32,57 @@ export const renderVerificationSms = (
 ): string => dedent`Use the OTP ${otp} to submit on ${appHost}.
 
   Never share your OTP with anyone else. If you did not request this OTP, you can safely ignore this SMS.`
+
+/**
```
karrui, line 36: if the retrieval handler is in AdminFormController, then we can reuse admin-form.utils too

src/app/services/sms/sms.types.ts
```diff
@@ -88,6 +91,7 @@ export interface IBouncedSubmissionSmsCountSchema
 
 export interface ISmsCountModel extends Model<ISmsCountSchema> {
   logSms: (logParams: LogSmsParams) => Promise<ISmsCountSchema>
+  retrieveFreeSmsCounts: (userId: string) => Promise<number>
```
karrui, line 94: can we add JSDocs for this? (realized logSms has missing JSDocs, but yeah just add for this one to prevent PR from going out of scope)

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
seaerchin: could i clarify this - do you mean another `describe` block nested so that the true/false behaviour is for that block or to have the tests such that the default behaviour is `false` and then we test for `true` (thinking the latter one right)

src/app/routes/api/v3/admin/forms/admin-forms.form.routes.ts
seaerchin: sure! i agree with your point that it's easier to find also so will make the required changes

src/app/services/sms/sms.util.ts
seaerchin: sure, will shift this over to `admin-form.utils`

src/app/services/sms/__tests__/sms_count.server.model.spec.ts
seaerchin: agreed - our tests should have sensible defaults rather than having exceptional cases as being the default. changed accordingly!

src/app/services/sms/sms.controller.ts
seaerchin: changed accordingly!

src/app/services/sms/sms.types.ts
seaerchin: sure, added!

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -9798,4 +9801,139 @@ describe('admin-form.controller', () => {
...
+    beforeAll(() => {
+      MockFormService.retrieveFormById.mockReturnValue(okAsync(mockForm))
+      MockSmsService.retrieveFreeSmsCounts.mockReturnValue(
+        okAsync(VERIFICATION_SMS_COUNT),
+      )
+    })
+
+    it('should retrieve counts and msgSrvcId when the user and the form exist', async () => {
+      // Arrange
+      const MOCK_REQ = expressHandler.mockRequest({
```
karrui, line 9820: nit: if its not a globally defined constant, this should be in camelCase.

Same applies for the rest of invocations in this test

src/app/services/sms/sms.service.ts
```diff
@@ -461,3 +465,28 @@ export const sendBouncedSubmissionSms = (
...
+): ResultAsync<number, PossibleDatabaseError> => {
+  return ResultAsync.fromPromise(
+    SmsCount.retrieveFreeSmsCounts(userId),
+    (error) => {
+      logger.error({
+        message: `Retrieving free sms counts failed for ${userId}`,
+        meta: {
+          action: 'retrieveFreeSmsCounts',
+          userId,
+        },
```
karrui, line 486: error should also be passed into the logger for logging

karrui (approved):
lgtm, just need to log that one extra error

seaerchin:
Tested on staging - counts retrieval on staging takes ~70ms for 10k sms counts. 

Screenshot 2021-07-14 at 10 35 58 AM

PR #2278, opened 2021-07-01
refactor: revert the revert of encapsulate parsedResponses

karrui:
This reverts #2248, where it temporarily reverted commit 6512a84 to reduce risk for the next release.


mantariksh:
waiting until after response ID is released

yong-jie (approved):

Issue #2005, opened 2021-05-31
[React] Add Field/PhoneNumber component


karrui:
**Time estimation: 2d**

Blocked by completion of #2002

Complex component, considerations include:
- A11y
- Handling international vs national numbers
- Which 3rd party packages to use
- Styling may also need overriding if 3rd party package has own styles

PR #2279, opened 2021-07-01
feat(design-system): add PhoneNumberInput (intl) component and stories

karrui:
## Problem

Note that this PR builds upon the Input component PR #2246 as they share styles. Will rebase as needed if there are changes in the base branch.

This PR only concerns itself with the international phone number component. A future PR will add the non-intl version (without country select)

Related to #2005 

## Solution

**Features**:

- add PhoneNumberInput component and stories
- add PhoneNumberInput theme

**Improvements**:

- allow `sx` param override on input component so other components can override styles of Input

## Before & After Screenshots
In storybook

## Deploy Notes

**New dependencies**:
- add `country-flag-icons` package and type definitions for rendering of flag icons in the custom input
- add libphonenumber-js package for parsing and formatting of phone numbers


frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -0,0 +1,304 @@
...
+          const appended = newValue.slice(inputValue.length)
+          setInputValue(formatter.input(appended))
+
+          const number = formatter.getNumber()
+          if (number?.country && number.country !== country) {
+            setCountry(number.country)
+          }
+        } else {
+          // Reset the formatter, but do not reformat.
+          // Doing so now will cause the user to loose their cursor position
```
seaerchin, line 149: nit: typo on loose

mantariksh (approved):

frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -0,0 +1,304 @@
...
+          newValue.slice(0, inputValue.length) === inputValue
+
+        if (isAppend) {
+          const appended = newValue.slice(inputValue.length)
+          setInputValue(formatter.input(appended))
+
+          const number = formatter.getNumber()
+          if (number?.country && number.country !== country) {
+            setCountry(number.country)
+          }
```
seaerchin, line 146: could we add a comment here fpr what this does?

frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -0,0 +1,304 @@
...
+    <InputLeftAddon
+      aria-disabled={inputProps.disabled}
+      title={COUNTRY_CODE_TO_NAME[value]}
+      as="label"
+      sx={styles.country}
+    >
+      <Flex>
+        <Icon
+          aria-disabled={inputProps.disabled}
+          // Show FLags if available. If value does not exist for any reason,
```
seaerchin, line 280: nit: casing

frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -0,0 +1,304 @@
...
+        innerInputRef?.current?.focus()
+      },
+      [country, onInputChange],
+    )
+
+    const onInputBlur = useCallback(() => {
+      const number = formatter.getNumber()
+      onBlur?.()
+
+      onChange(number?.number as string | undefined)
```
seaerchin, line 182: could i clarify on why this `onChange` is needed, as we would have called it up in the `onInputChange` block already?

seaerchin:
could i ask why the country/input are set separately rather than using the formatter as the source of truth because it seems like we're always combining the updates to input/country + formatter

frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
karrui: this is just setting the country state with the processed number~

frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
karrui: The number may be formatted again.

This can happen in the below scenario:

1. `onChange` gets called when user types for example "65aabvcd123"
2. `formatter.getNumber().number` will transform that into "65" and cut out the remaining characters since the remaining string is not a valid number
3. Will need to call onChange on this new number.

I will add a new comment for clarity

karrui:
> could i ask why the country/input are set separately rather than using the formatter as the source of truth because it seems like we're always combining the updates to input/country + formatter

We want the user to also select country independently without having to type a valid number to trigger country changes.

PR #2280, opened 2021-07-01
feat(sms-limiting): admin facing frontend

seaerchin:
## Problem
This PR adds the admin facing frontend for the sms limiting feature. This feature is dependent on [this PR](https://github.com/opengovsg/FormSG/pull/2276).

The frontend in this PR can be split into 3 portions, which will be covered below: 
1. loading state 
2. success state
3. error state 

## Solution
The approach adopted tries to be as non-intrusive as possible and has minimal behavioural changes wrt old behaviour. 

**loading state**
Only the toggle is disabled during the loading state, when we are fetching data from the backend. This is to prevent the admin from toggling the state and showing the modal when there isn't enough data to determine the admin verificaiton state. 

**success state**
On fetching counts successfully from the backend, we toggle and set the verification state based on the counts + the form's message service id. 

Because `$scope.field` is also read by the verified button component, a design choice was made to set shared props on this. Alternatively, we could create a new property on `$scope` and share it through sharing the property using 2 way data binding but it'll serve much of the same purpose and there is very little benefit to be gained because of the shift to react.   

**error state**
Main concerns with the display during error state is 
1. how to handle the toggle
2. keeping sms toggling consistent 

a few different flows were considered: 
1. don't disable the toggle - this was rejected because we do not want to mislead users who have exceeded into thinking that they can toggle on (only to be rejected on save). 
2. Toaster vs displaying infobox - felt that a toaster offered better user experience compared to an infobox because the toaster is displayed in response to an action.

**Breaking Changes** 

- [x] Yes - this PR contains breaking changes (dependent)
    - This PR queries a new endpoint to obtain the counts for the admin, which will break on rollback of server
    - If the frontend change (this PR) is rollbacked, this will be ok as it just means that the new endpoint won't be queried

## Manual Tests
1. Create a form 
2. Create a mobile field for the form
3. Make the form public
4. Duplicate the form

- [ ] Attempt to toggle the mobile field. A modal should show up, displaying the sms counts for the admin of the form 
- [ ] Go to the other form and request OTP verification on the public user facing frontend. On the first form, disable otp verification then turn it on again. An updated sms counts should be shown 
- [ ] Transfer ownership of the first form to another user. Turn mobile otp verification off then on again. The counts should be for the new form admin. 

**The below tests require db write permissions on staging**
- [ ] Edit the form so that it has a message service id. Attempt to toggle otp verification. This should succeed without any modal displaying
- [ ] Insert >10k documents sent under the admin into the sms counts table (see the block below for the commands). Open the mobile field. The OTP verification should be disabled.  
```
db.getCollection('smscounts').insertMany([10,000 of fake sms count document])
```

**fake sms count document**
```
 {
          "smsType" : "VERIFICATION",
          "form" : <formId here>,
          "formAdmin" : {
              "email" : <your email here>,
              "userId" : ObjectId(<your id here>)
          },
          "msgSrvcSid" : <form's twilio sid here>,
          "logType" : "SUCCESS",
          "createdAt" : ISODate("2021-06-30T07:41:53.768Z"),
          "isOnboardedAccount" : false,
          "__v" : 0
        }
```

- [ ] Edit the form in db and add a own custom message service id. Go to that form and attempt to toggle sms verification on. This should succeed without any modal showing.

## Screen recordings
**Initial loading state**
recording (4)

**Admin verification under limit**
success

**Admin over limit**
over limit

**Form has message service id**
has messageid

**Fetching sms counts errored**
recording (1)


src/public/modules/forms/admin/directives/configure-mobile.client.directive.js
```diff
@@ -10,24 +18,79 @@ function configureMobileDirective() {
...
     },
     controller: [
+      '$q',
       '$uibModal',
       '$scope',
       '$translate',
-      function ($uibModal, $scope, $translate) {
+      'Toastr',
+      function ($q, $uibModal, $scope, $translate, Toastr) {
         // Get support form link from translation json.
```
chowyiyin, line 34: Maybe we can update this comment?

src/public/modules/forms/admin/directives/configure-mobile.client.directive.js
```diff
@@ -10,24 +18,79 @@ function configureMobileDirective() {
...
+            $scope.verifiedSmsCount = smsCounts
+            $scope.adminVerifiedSmsState = getAdminVerifiedSmsState(
+              smsCounts,
+              $scope.form.msgSrvcName,
+            )
+            // NOTE: This links into the verifiable field component and hence, is used by both email and mobile
+            $scope.field.hasAdminExceededSmsLimit =
+              $scope.adminVerifiedSmsState ===
+              ADMIN_VERIFIED_SMS_STATES.limitExceeded
+            $scope.field.hasRetrievalError = false
```
chowyiyin, line 66: Probably don't need to set `hasRetrievalError` to `false` since its already set above?

src/public/modules/forms/admin/directiveViews/configure-mobile.client.view.html
```diff
@@ -20,10 +27,30 @@
...
   <br />
+  <div
+    class="alert-custom alert-error inline-padding"
+    ng-show="field.hasAdminExceededSmsLimit"
+  >
+    <i class="bx bx-exclamation bx-md icon-spacing"></i>
+    <span class="alert-msg">
+      You have reached the free tier limit for SMS verification. To continue
+      using SMS verification,
+      <a href="https://go.gov.sg/form-twilio-setup" target="_blank">
```
chowyiyin, line 51: I think we can use `ng-href="{{twilioSetupLink}}` here!

src/public/translations/en-SG/main.json
```diff
@@ -18,6 +18,7 @@
     "GO_GOV": "https://go.gov.sg",
     "POSTMAN_GOV": "https://postman.gov.sg",
     "SUPPORT_FORM_LINK": "https://go.gov.sg/formsg-support",
+    "TWILIO_SETUP_LINK": "https://go.gov.sg/form-twilio-setup",
```
chowyiyin, line 21: Should we name this "VERIFIED_SMS_SETUP_LINK` instead to decouple the naming from the service?

src/public/modules/forms/admin/directives/configure-mobile.client.directive.js
seaerchin: agreed, will change!

src/public/translations/en-SG/main.json
seaerchin: yeap, good point - will change

src/public/modules/forms/admin/directives/configure-mobile.client.directive.js
seaerchin: thanks for comment - changed!

karrui:
> [ ] Insert >10k documents sent under the admin into the sms counts table (see the block below for the commands). Open the mobile field. The OTP verification should be disabled.

Can we set the limit as an env var so we can test it much easier instead of doing this, which may get rolled back often

src/public/modules/forms/admin/directiveViews/configure-mobile.client.view.html
```diff
@@ -9,9 +12,13 @@
       ></i>
     </div>
     <div class="col-xs-4 field-input">
+      <div ng-show="isLoading" class="loading-spinner">
```
karrui, line 15: can there be a screenshot of the loading state in the PR

src/shared/util/verification.ts
```diff
@@ -21,3 +21,11 @@ export enum VfnErrors {
...
   InvalidMobileNumber = 'INVALID_MOBILE_NUMBER',
 }
+
+export enum ADMIN_VERIFIED_SMS_STATES {
+  limitExceeded = 'LIMIT_EXCEEDED',
+  belowLimit = 'BELOW_LIMIT',
+  hasMessageServiceId = 'MESSAGE_SERVICE_ID_OBTAINED',
+}
+
+export const SMS_VERIFICATION_LIMIT = 10000
```
karrui, line 31: can this be an env var (that is then sent to the frontend the usual env-var-ry way)

karrui (approved):
lgtm with comments

pearlyong (approved):

src/public/modules/forms/admin/directiveViews/configure-mobile.client.view.html
seaerchin: sure, will add in!

src/shared/util/verification.ts
seaerchin: sure, could i ask for clarification on when we decide something should be an env-var vs a constant?

seaerchin:
Tested retrieval speed on staging - 10k documents takes ~70ms. For all real world use-cases, the toggle won't be shown ;--;

Screenshot 2021-07-14 at 10 35 58 AM

seaerchin:
Tested retrieval speed on staging **(slow 3g using chrome throttling)** -  10k documents takes ~2s. Toggle works! 

Screenshot 2021-07-14 at 11 06 40 AM

PR #2300, opened 2021-07-05
feat(sms-limiting): disable sms verifications toggling

seaerchin:
## Problem
This PR addresses the edge case when the admin wishes to perform some operation on the form field but should be prohibited from doing so because their sms limit has been exceeded. This PR is dependent on #2276 

## Solution
The solution consists of 2 portions. 

**Checking if the operation should be prohibited**
- this is done though a service method called `shouldUpdateFormField`, where the applicable scenario is checked - the field is of type Mobile and it's verifiable. When this occurs, the free sms counts used of the admin is retrieved and checked to see if the limit has been exceeded, and a `Result` is returned accordingly. 

**Preventing operation if it should be prohibited**
- because the type returned in the error case is `Err<some base type>`, the error returned will be propagated and short-circuited, preventing the operation from taking place

## Technical Choices
1. Placing of `shouldUpdateFormField` in controller rather than service.
    - the choice to update should be clearly visible and not abstracted by the service
    - choice to prohibit the update should be a controller level responsibility; service should just perform a single function

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
@@ -34,6 +34,7 @@ import {
 import * as SubmissionService from 'src/app/modules/submission/submission.service'
 import * as SubmissionUtils from 'src/app/modules/submission/submission.utils'
 import { MissingUserError } from 'src/app/modules/user/user.errors'
+import { SmsLimitExceededError } from 'src/app/modules/verification/verification.errors'
 import {
   MailGenerationError,
   MailSendError,
@@ -6740,10 +6741,10 @@ describe('admin-form.controller', () => {
       _id: MOCK_USER_ID,
       email: 'somerandom@example.com',
     } as IPopulatedUser
-    const MOCK_FIELD = generateDefaultField(BasicField.Rating)
+    const MOCK_FIELD = generateDefaultField(BasicField.Mobile)
     const MOCK_UPDATED_FIELD = {
       ...MOCK_FIELD,
-      title: 'some new title',
+      isVerifiable: true,
     } as FieldUpdateDto
 
     const MOCK_FORM = {
@@ -6771,7 +6772,9 @@ describe('admin-form.controller', () => {
       MockAuthService.getFormAfterPermissionChecks.mockReturnValue(
         okAsync(MOCK_FORM),
       )
-
+      MockAdminFormService.shouldUpdateFormField.mockReturnValue(
+        okAsync(MOCK_FORM),
+      )
       MockAdminFormService.updateFormField.mockReturnValue(
         okAsync(MOCK_UPDATED_FIELD as IFieldSchema),
       )
@@ -6868,6 +6871,29 @@ describe('admin-form.controller', () => {
         message: expectedErrorString,
       })
       expect(MockAdminFormService.updateFormField).not.toHaveBeenCalled()
+    })
+
+    it('should return 409 when the field could not be updated due to a sms limit exceeded error', async () => {
+      // Arrange
+      MockAdminFormService.shouldUpdateFormField.mockReturnValueOnce(
+        errAsync(new SmsLimitExceededError()),
+      )
+      const expected = {
+        message:
+          'You have exceeded the free sms limit. Please refresh and try again.',
+      }
+      const mockRes = expressHandler.mockResponse()
+
+      // Act
+      await AdminFormController._handleUpdateFormField(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      // Assert
+      expect(mockRes.status).toBeCalledWith(409)
+      expect(mockRes.json).toBeCalledWith(expected)
     })
 
     it('should return 410 when form to update form field for is already archived', async () => {
@@ -7019,10 +7045,12 @@ describe('admin-form.controller', () => {
       title: 'mock title',
     } as IPopulatedForm
 
-    const MOCK_RETURNED_FIELD = generateDefaultField(BasicField.Nric)
+    const MOCK_RETURNED_FIELD = generateDefaultField(BasicField.Mobile, {
+      isVerifiable: true,
+    })
     const MOCK_CREATE_FIELD_BODY = pick(MOCK_RETURNED_FIELD, [
       'fieldType',
-      'title',
+      'isVerifiable',
     ]) as FieldCreateDto
     const MOCK_REQ = expressHandler.mockRequest({
       session: {
@@ -7038,6 +7066,9 @@ describe('admin-form.controller', () => {
     beforeEach(() => {
       MockUserService.getPopulatedUserById.mockReturnValue(okAsync(MOCK_USER))
       MockAuthService.getFormAfterPermissionChecks.mockReturnValue(
+        okAsync(MOCK_FORM),
+      )
+      MockAdminFormService.shouldUpdateFormField.mockReturnValue(
         okAsync(MOCK_FORM),
       )
       MockAdminFormService.createFormField.mockReturnValue(
@@ -7109,6 +7140,29 @@ describe('admin-form.controller', () => {
         message: expectedErrorString,
       })
       expect(MockAdminFormService.createFormField).not.toHaveBeenCalled()
+    })
+
+    it('should return 409 when the field could not be created due to a sms limit exceeded error', async () => {
+      // Arrange
+      MockAdminFormService.shouldUpdateFormField.mockReturnValueOnce(
+        errAsync(new SmsLimitExceededError()),
+      )
+      const expected = {
+        message:
+          'You have exceeded the free sms limit. Please refresh and try again.',
+      }
+      const mockRes = expressHandler.mockResponse()
+
+      // Act
+      await AdminFormController._handleUpdateFormField(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      // Assert
+      expect(mockRes.status).toBeCalledWith(409)
+      expect(mockRes.json).toBeCalledWith(expected)
     })
 
     it('should return 410 when attempting to create a form field for an archived form', async () => {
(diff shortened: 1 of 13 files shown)


src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -7108,6 +7139,29 @@ describe('admin-form.controller', () => {
...
+        errAsync(new SmsLimitExceededError()),
+      )
+      const expected = {
+        message:
+          'You have exceeded the free sms limit. Please refresh and try again.',
+      }
+      const mockRes = expressHandler.mockResponse()
+
+      // Act
+      await AdminFormController._handleUpdateFormField(
```
karrui, line 7154: this is the exact same test as above. did you mean to call `_handleCreateFormField` instead?

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -7464,6 +7525,28 @@ describe('admin-form.controller', () => {
...
```
karrui, line 7540: same here...

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -2242,4 +2246,106 @@ describe('admin-form.service', () => {
...
+      BasicField.Mobile,
+    )
+    const MOCK_VERIFIABLE_MOBILE_FIELD = generateDefaultField(
+      BasicField.Mobile,
+      { isVerifiable: true },
+    )
+
+    const countSpy = jest.spyOn(SmsService, 'retrieveFreeSmsCounts')
+
+    it('should return the form when admin is under the limit', async () => {
```
karrui, line 2267: should we increase the specificity of this test description? we are after all in AdminFormService, and it's not obvious what this is checking.

suggestion "should return the given form when form field to update is a mobile field and admin is still under the free SMS quota limit"

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -2242,4 +2246,106 @@ describe('admin-form.service', () => {
...
+      const actual = await shouldUpdateFormField(
+        MOCK_FORM,
+        MOCK_VERIFIABLE_MOBILE_FIELD,
+      )
+
+      // Assert
+      expect(actual._unsafeUnwrap()).toEqual(MOCK_FORM)
+    })
+
+    it('should return the form when the form has a message service name', async () => {
```
karrui, line 2281: same here; test description should be clearer

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
karrui: we can even have a deeper describe block of 
```ts
describe('when update form field is of BasicField.Mobile type', () => {
  ...
})
```

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -2242,4 +2246,106 @@ describe('admin-form.service', () => {
...
+      // Act
+      const actual = await shouldUpdateFormField(MOCK_FORM, MOCK_NUMBER_FIELD)
+
+      // Assert
+      // The sms counts check is only executed when it is a mobile field
+      expect(countSpy).not.toHaveBeenCalled()
+      expect(actual._unsafeUnwrap()).toEqual(MOCK_FORM)
+    })
+
+    it('should return sms retrieval error when sms limit exceeded and admin is attempting to toggle sms verification for mobile field on', async () => {
```
karrui, line 2321: this test description should not be "coupled" to the actions of the client, i think the first part is enough

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1069,3 +1074,38 @@ export const disableSmsVerificationsForUser = (
...
+ * while attempting to toggle verification on for a mobile form field
+ */
+export const shouldUpdateFormField = (
+  form: IPopulatedForm,
+  formField: IField,
+): ResultAsync<
+  IPopulatedForm,
+  PossibleDatabaseError | SmsLimitExceededError
+> => {
+  const formAdminId = String(form.admin._id)
```
karrui, line 1095: nit: this can be instantiated after the `if` block below since it's not used until then.

src/app/modules/form/form.utils.ts
```diff
@@ -121,3 +123,8 @@ export const getLogicById = (
 
   return form_logics.find((logic) => logicId === String(logic._id)) ?? null
 }
+
+// Typeguard to check if a form has a message service id
+export const isOnboardedForm = (form: IForm): form is IOnboardedForm<IForm> => {
```
karrui, line 128: nit: make arg more generic to allow better typing for forms that extend from `IForm`:

```ts
export const isOnboardedForm = <T extends IForm = IForm>(form: T): form is IOnboardedForm<T> => {
   ...
}
```

src/shared/util/verification.ts
```diff
@@ -21,3 +21,5 @@ export enum VfnErrors {
   TransactionNotFound = 'TRANSACTION_NOT_FOUND',
   InvalidMobileNumber = 'INVALID_MOBILE_NUMBER',
 }
+
+export const SMS_VERIFICATION_LIMIT = 10000
```
karrui, line 25: should we make this an env var with a default and instantiate in config?

We can still send the variable to the frontend just like how we send other variables currently

karrui (changes requested):

src/shared/util/verification.ts
seaerchin: sure, changed it and updated default to 10,000!

src/app/modules/form/admin-form/admin-form.service.ts
seaerchin: sure!

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
seaerchin: i think having a deeper `describe` block makes sense because it's a special case so i'll adjust it accordingly!

src/app/modules/form/form.utils.ts
seaerchin: yep, good point; changed!

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1069,3 +1074,38 @@ export const disableSmsVerificationsForUser = (
...
+ * Checks if the given form field should be updated.
+ * This currently checks if the admin has exceeded their free sms limit.
+ * @param form The form which the specified field belongs to
+ * @param formField The field which we should perform the update for
+ * @returns ok(form) If the field can be updated
+ * @return err(PossibleDatabaseError) if an error occurred while performing the required checks
+ * @return err(SmsLimitExceededError) if the form admin went over the free sm limit
+ * while attempting to toggle verification on for a mobile form field
+ */
+export const shouldUpdateFormField = (
```
karrui, line 1088: now that I'm reading this function signature again, should we make it clearer so ppl reading only the controller can understand what this function does?

Can we further split this function up so it can be extended if needed?

```ts
const shouldUpdateFormField = (...) => {
  switch (formField.fieldType) {
    case BasicField.Mobile:
      return assessMobileFieldUpdateAllowed(...)
    default:
      return okAsync(form)
}
```
`assessMobileFieldUpdateAllowed` can still be a function in this admin-form service since it's still acting on form fields.

If not, we can also move this field type check into the controller and call `assessMobileFieldUpdateAllowed`. This makes the pipelines clearer

src/app/modules/form/admin-form/admin-form.utils.ts
```diff
@@ -57,6 +58,11 @@ export const mapRouteError = (
   coreErrorMessage?: string,
 ): ErrorResponseData => {
   switch (error.constructor) {
+    case SmsLimitExceededError:
+      return {
+        statusCode: StatusCodes.CONFLICT,
```
karrui, line 63: rethinking this status code. Would returning 400 be better?

src/app/modules/verification/verification.errors.ts
```diff
@@ -82,3 +82,14 @@ export class NonVerifiedFieldTypeError extends ApplicationError {
     )
   }
 }
+
+/**
+ * Agency user has sent too many SMSes using default Twilio credentials
+ */
+export class SmsLimitExceededError extends ApplicationError {
+  constructor(
+    message = 'You have exceeded the free sms limit. Please refresh and try again.',
```
karrui, line 91: This message will be shown to the user right? I don't think refreshing the page would change anything if the sms limit is reached. Maybe consult with @syan-syan

src/app/modules/form/admin-form/admin-form.service.ts
seaerchin: i was thinking about doing the split-up but was wary about premature optimization actually HAHA now that you've raised it, i'll split it up!

src/app/modules/form/admin-form/admin-form.utils.ts
seaerchin: to give context, this error is thrown when sms limit is exceeded but the UI does not reflect it yet; hence, i thought `409 conflict` would be better as it represents a [difference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) between the UI state and BE state. 

could i understand why you think 400 is better?

src/app/modules/verification/verification.errors.ts
seaerchin: will clarify with sonjia wrt this! what changes on page refresh is that now sms toggling would be disabled tgt with link shown to address billing

seaerchin: confirmed that this is ok; will leave it as is

src/app/modules/form/admin-form/admin-form.utils.ts
karrui: > represents a difference between the UI state and BE state.

This is not a particularly correct justification; all PUT requests inherently have a difference between client and BE before the request is resolved (that's the whole point of a PUT, to sync backend and client state).

But yea, no big deal, 409 works too

src/app/modules/form/admin-form/admin-form.service.ts
seaerchin: done [here](https://github.com/opengovsg/FormSG/pull/2300/commits/8b5d20c5df364df3daed5e9e41fba411be333248)

karrui (approved):

PR #2301, opened 2021-07-05
fix(admin-form.service): fixed logical error in duplicate form field

seaerchin:
## Problem
`AdminFormService.duplicateFormField` occasionally returns the wrong error type. This is because the underlying schema method returns null when either the field or the form was not found.

## Solution
Insert a check after calling `form.duplicateFormFieldById` to see if there is a field and return the error appropriately.

**Breaking Changes** 

    - Details ...

## Tests
This also fixes an erroneous integration test where the property being tested was when teh field wasn't found but the `expect` block checks that the _form_ was not found.

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.service.ts
@@ -499,8 +499,9 @@ export const duplicateFormField = (
     },
   ).andThen((updatedForm) => {
     if (!updatedForm) {
-      // Null return could be due to form or field not existing.
-      // Hence, check if field exists in form and return the appropriate error
+      // Success means field is in initial form object but query still returned null.
+      // Return best guess error that form is now not found in the DB.
+      // Otherwise, err(FieldNotFoundError) will be returned by the function.
       return getFormField(form, fieldId).asyncAndThen(() =>
         errAsync(new FormNotFoundError()),
       )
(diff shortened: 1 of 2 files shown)


chowyiyin (commented):
lgtm

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -499,7 +499,11 @@ export const duplicateFormField = (
     },
   ).andThen((updatedForm) => {
     if (!updatedForm) {
-      return errAsync(new FormNotFoundError())
+      // Null return could be due to form or field not existing.
+      // Hence, check if field exists in form and return the appropriate error
+      return getFormField(form, fieldId).asyncAndThen(() =>
```
karrui, line 504: i think better comments might make this clearer. 

```ts
return getFormField(form, fieldId)
  // Success means field is in initial form object; but query still returned null.
  // Return best guess error that form is now not found in the DB.
  .asyncAndThen(() => errAsync(new FormNotFoundError()))
  // Otherwise, err(FieldNotFoundError) will be returned by the function.
```

karrui (approved):
lgtm with suggestion on clearer comments

src/app/modules/form/admin-form/admin-form.service.ts
seaerchin: fixed, thanks!

Issue #2208, opened 2021-06-18
Logic model validation is loose


mantariksh:
In form logic, specific field types are only compatible with certain specific "condition states", e.g. dropdown fields can only be used with "is equals to" or "is either" conditions.

However, the backend does not enforce this coupling between field types and condition states. It simply accepts any field ID linked to any possible condition state. For example, nothing is stopping the frontend from mistakenly saving a logic unit where an "is greater than" condition is applied to a dropdown field. This introduces risks for the React frontend migration, since sending API calls with the wrong logic unit shape could result in corruption of the database state.

This should be tackled in a few steps:
1. Check that the current database does not contain any mismatched logic conditions.
2. Enforce the correct matching of logic conditions to field types in the model layer.
3. Enforce the correct shape of logic API calls through Joi validation.

liangyuanruo:
Note: to also watch out for data pollution

mantariksh:
on second thoughts, not clear if Joi validation can be used here, since the Joi validator only has access to the field ID and not the field type, and hence may not be able to check if the corresponding conditions are valid

PR #2302, opened 2021-07-05
feat(form-logic): introduce form logic model validation

yong-jie:
## Problem

See #2208 

In form logic, specific field types are only compatible with certain specific "condition states", e.g. dropdown fields can only be used with "is equals to" or "is either" conditions.

However, the backend does not enforce this coupling between field types and condition states. It simply accepts any field ID linked to any possible condition state. For example, nothing is stopping the frontend from mistakenly saving a logic unit where an "is greater than" condition is applied to a dropdown field. This introduces risks for the React frontend migration, since sending API calls with the wrong logic unit shape could result in corruption of the database state.

## Solution

1) Refactor createFormLogic and updateFormLogic to use .save(), because mongoose does not grant access to the underlying form document for the update operations.

2) Move `LogicMap` from being a frontend logic to a shared util.

3) Add validators on the form_logics array

**Breaking Changes** 

**Features**:

Form logic now does validation on the field type and state. This ensures that things like yes/no fields don't have invalid condition states like `less than`.

## Manual tests

- [ ] Via API, attempt to save mismatched form logic (e.g. logic state `''is less than or equal to'` for a dropdown field) and check that model validation disallows it.
- [ ] For each valid logic field type (dropdown, radio, number, yes/no, rating, decimal), create a logic unit involving that field type to make sure valid logic is not affected.


src/app/models/form.server.model.ts
```diff
@@ -208,7 +211,69 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+            ): conditions is ValidatableCondition[] =>
+              !conditions.some((condition) => !condition.fieldType)
+
+            const conditions = v.flatMap((logic) => {
+              return logic.conditions.map<IncompleteValidatableCondition>(
+                (condition) => {
+                  const {
+                    field,
+                    state,
+                  }: { field: ObjectId; state: LogicConditionState } = condition
```
mantariksh, line 253: ooo I'm a bit wary of asserting that `field` is an `ObjectId`. it's not inconceivable that some of them in the database have been saved as strings, so calling `toHexString` on them would result in a null pointer exception. can we use the `String` constructor instead so we can deal with both ObjectIds and strings?

src/app/models/form.server.model.ts
```diff
@@ -208,7 +211,69 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+                    state,
+                    fieldType: form.form_fields?.find(
+                      (f: IFieldSchema) =>
+                        f._id.toHexString() === field.toHexString(),
+                    )?.fieldType,
+                  }
+                },
+              )
+            })
+            if (!isAllConditionsReferenceExistingFields(conditions))
```
mantariksh, line 264: nit: can add the braces around the `return` statement since it's spilling to the next line anyway

src/app/models/form.server.model.ts
```diff
@@ -208,7 +211,69 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+              return false
+
+            const validationResults = conditions.map((condition) => {
+              const { fieldType, state } = condition
+              const applicableIfStates = getApplicableIfStates(fieldType)
+              return applicableIfStates.includes(state)
+            })
+            return validationResults.every((r) => r)
```
mantariksh, line 272:
```suggestion
            return conditions.every((condition) => {
              const { fieldType, state } = condition
              const applicableIfStates = getApplicableIfStates(fieldType)
              return applicableIfStates.includes(state)
            })
```

src/app/models/form.server.model.ts
```diff
@@ -208,7 +211,69 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+                    fieldType: form.form_fields?.find(
+                      (f: IFieldSchema) =>
+                        f._id.toHexString() === field.toHexString(),
+                    )?.fieldType,
+                  }
+                },
+              )
+            })
+            if (!isAllConditionsReferenceExistingFields(conditions))
+              return false
```
mantariksh, line 265: I think this is a bug actually. we don't prevent admins from saving forms with invalid logic (i.e. logic referencing non-existent fields), we merely warn them that their form has invalid logic. I think you may need to loosen the validation to accept conditions referencing non-existent fields.

src/app/models/form.server.model.ts
```diff
@@ -718,17 +782,15 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
-      {
-        arrayFilters: [{ 'object._id': logicId }],
-        new: true,
-        runValidators: true,
-      },
-    ).exec()
+    let form = await this.findById(formId).exec()
+    if (!form?.form_logics) return null
+    const index = form.form_logics.findIndex(
+      (logic) => logic._id.toHexString() === logicId,
```
mantariksh, line 788: again I think it would be safer to use the `String` constructor instead of `toHexString`

src/app/models/form.server.model.ts
mantariksh: would be good to add model validation tests for this case also, it's important to make sure this doesn't regress as quite a few forms have invalid logic (we see the logs all the time)

mantariksh (changes requested):
main concern is a regression where we are disallowing logic referencing non-existent fields

src/app/models/form.server.model.ts
yong-jie: will be deleting this line instead, since we won't enforce conditions no reference existing fields

src/app/models/form.server.model.ts
```diff
@@ -208,7 +211,62 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+                    fieldType: this.form_fields?.find(
+                      (f: IFieldSchema) => String(f._id) === String(field),
+                    )?.fieldType,
+                  }
+                },
+              )
+            })
+
+            return conditions.every((condition) => {
+              // Conditions don't necessarily need to reference existing fields
```
mantariksh, line 259: would be good if we can add a bit more explanation here to explain that we allow logic referencing non-existent fields to be saved, so we don't make life difficult for form admins when they delete fields

mantariksh (commented):
final request - could you pls help add tests in `form.server.model.spec.ts` to ensure that the logic validation works as expected? in particular under the `'Base Schema'` suite, it would be good to add a test to ensure that we do not accidentally disallow logic which references non-existent fields

src/app/models/form.server.model.ts
yong-jie: added more documentation, lmk what you think!

mantariksh (approved):

src/app/models/form.server.model.ts
mantariksh: lgtm!

mantariksh (commented):
I just realised there's another logic condition field called `ifValueType`. do we need to check that these are valid for each fieldType too?

yong-jie:
`ifValueType` is a frontend-only value that's for some reason being tracked in the database. this is a tech debt, but afaik no validation is needed/possible

mantariksh (approved):

PR #2304, opened 2021-07-05
feat(design-system): add tabs style and stories

### LINKED ISSUE #2020: [React] Add Tab component
**Time estimation: 1d**

chowyiyin:
## Problem
This PR implements styles for the tab component. Chakra UI's tab component can be used directly. 

Closes #2020
## Screenshots
In storybook

### DIFF
--- a/frontend/src/components/Tabs/Tabs.stories.tsx
+++ b/frontend/src/components/Tabs/Tabs.stories.tsx
@@ -1,0 +1,29 @@
+import { Tab, TabList, Tabs, TabsProps } from '@chakra-ui/react'
+import { Meta, Story } from '@storybook/react'
+
+export default {
+  title: 'Components/Tabs',
+  component: Tabs,
+} as Meta
+
+const TabTemplate: Story<TabsProps> = (args) => {
+  return (
+    <Tabs {...args}>
+      <TabList>
+        <Tab>Create</Tab>
+        <Tab>Settings</Tab>
+        <Tab>Results</Tab>
+      </TabList>
+    </Tabs>
+  )
+}
+
+export const LightTab = TabTemplate.bind({})
+LightTab.args = {
+  variant: 'line',
+}
+
+export const DarkTab = TabTemplate.bind({})
+DarkTab.args = {
+  variant: 'enclosed',
+}
--- a/frontend/src/theme/components/Tabs.ts
+++ b/frontend/src/theme/components/Tabs.ts
@@ -1,0 +1,105 @@
+import {
+  ChakraTheme,
+  ComponentMultiStyleConfig,
+  SystemStyleObjectRecord,
+  ThemingPropsThunk,
+} from '@chakra-ui/react'
+import { merge } from 'lodash'
+
+const variantLine: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
+  props,
+) => {
+  const { orientation } = props
+  const isVertical = orientation === 'vertical'
+  const marginProp = isVertical ? 'marginStart' : 'marginBottom'
+  return {
+    root: {
+      w: 'auto',
+      p: '24px',
+      mx: '-16px', // ignore parent padding and extend full width of screen
+    },
+    tablist: {
+      overflowX: 'scroll',
+      borderColor: 'transparent',
+      /* Hide scrollbar for IE and Edge */
+      '&::MsOverflowStyle': 'none',
+      /* Hide scrollbar for Firefox */
+      '&::scrollbarWidth': 'none',
+      /* Hide scrollbar for Chrome, Safari and Opera */
+      '&::-webkit-scrollbar': {
+        display: 'none',
+      },
+    },
+    tabPanel: {
+      p: '4px',
+    },
+    tab: {
+      textTransform: 'uppercase',
+      p: '4px',
+      [marginProp]: '4px',
+    },
+  }
+}
+
+const variantLight: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
+  props,
+) => {
+  return merge(variantLine(props), {
+    root: {
+      backgroundColor: 'white',
+    },
+    tab: {
+      _hover: {
+        color: 'primary.400',
+      },
+      _selected: {
+        color: 'primary.500',
+      },
+      color: 'primary.300',
+    },
+  })
+}
+
+const variantDark: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
+  props,
+) => {
+  return merge(variantLine(props), {
+    root: {
+      backgroundColor: 'secondary.500',
+    },
+    tab: {
+      _hover: {
+        color: 'secondary.200',
+      },
+      _selected: {
+        color: 'white',
+      },
+      color: 'secondary.300',
+    },
+  })
+}
+
+export const Tabs: ComponentMultiStyleConfig = {
+  parts: [],
+  baseStyle: {
+    tab: {
+      mr: '6%',
+      px: '3%',
+    },
+  },
+  sizes: {
+    md: {
+      tab: {
+        textStyle: 'subhead-3',
+        p: '0px',
+      },
+    },
+  },
+  variants: {
+    line: variantLight,
+    enclosed: variantDark,
+  },
+  defaultProps: {
+    colorScheme: 'primary',
+  },
+}
--- a/frontend/src/theme/components/Tabs.ts
+++ b/frontend/src/theme/components/Tabs.ts
@@ -34,7 +34,7 @@ const variantLine: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
       p: '4px',
     },
     tab: {
-      textTransform: 'uppercase',
+      textTransform: 'capitalise',
       p: '4px',
       [marginProp]: '4px',
     },
@@ -46,7 +46,7 @@ const variantLight: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
 ) => {
   return merge(variantLine(props), {
     root: {
-      backgroundColor: 'white',
+      bg: 'white',
     },
     tab: {
       _hover: {
@@ -55,6 +55,10 @@ const variantLight: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
       _selected: {
         color: 'primary.500',
       },
+      _active: {
+        bg: 'white',
+      },
+      transitionProperty: 'common',
       color: 'primary.300',
     },
   })
@@ -65,15 +69,19 @@ const variantDark: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
 ) => {
   return merge(variantLine(props), {
     root: {
-      backgroundColor: 'secondary.500',
+      bg: 'secondary.500',
     },
     tab: {
+      _active: {
+        bg: 'secondary.500',
+      },
       _hover: {
         color: 'secondary.200',
       },
       _selected: {
         color: 'white',
       },
+      transitionProperty: 'common',
       color: 'secondary.300',
     },
   })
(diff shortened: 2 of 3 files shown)


frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
...
+  sizes: {
+    md: {
+      tab: {
+        textStyle: 'subhead-3',
+        p: '0px',
+      },
+    },
+  },
+  variants: {
+    line: variantLight,
```
karrui, line 110: I don't think we should override Chakra's base variants but create our own with better names. Line works but enclosed does not seem to be as clear as we want for our own variants.

Suggest `light-line` and `dark-line` variants, and maybe add a comment on how Chakra already has `line` variant but we have two variants over it.

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
...
+      color: 'secondary.300',
+    },
+  })
+}
+
+export const Tabs: ComponentMultiStyleConfig = {
+  parts: [],
+  baseStyle: {
+    tab: {
+      mr: '6%',
```
karrui, line 97: This seems arbitrary, is it really what the design is?

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
+import {
```
karrui, line 1: Can we convert all `px` to `rem`s

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
...
+  ChakraTheme,
+  ComponentMultiStyleConfig,
+  SystemStyleObjectRecord,
+  ThemingPropsThunk,
+} from '@chakra-ui/react'
+import { merge } from 'lodash'
```
karrui, line 7: suggest using `lodash/merge` instead to facilitate tree shaking

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
...
+  props,
+) => {
+  const { orientation } = props
+  const isVertical = orientation === 'vertical'
+  const marginProp = isVertical ? 'marginStart' : 'marginBottom'
+  return {
+    root: {
+      w: 'auto',
+      p: '24px',
+      mx: '-16px', // ignore parent padding and extend full width of screen
```
karrui, line 19: Is this because we are overriding `line` styles? Will probably not be needed with our own variant

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
...
+    },
+    tablist: {
+      overflowX: 'scroll',
```
karrui, line 22: ensure that overflow scrolling work. May want to have a story to test this with

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
+import {
+  ChakraTheme,
+  ComponentMultiStyleConfig,
```
karrui, line 3: Focus styles seem to be missing, might want to ask @pearlyong to update them

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,116 @@
...
+  ThemingPropsThunk,
+} from '@chakra-ui/react'
+import { merge } from 'lodash'
+
+const variantLine: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> = (
+  props,
+) => {
+  const { orientation } = props
+  const isVertical = orientation === 'vertical'
+  const marginProp = isVertical ? 'marginStart' : 'marginBottom'
```
karrui, line 14: Why do we need this? Design only has horizontal orientation right?

karrui (changes requested):

frontend/src/theme/components/Tabs.ts
chowyiyin: I made an estimation using the distance between the tabs and the total length of the `_IndividualTabs` component in figma. I realised this value is wrong, it should be around '15%'. Is it okay to use this estimation?

frontend/src/theme/components/Tabs.ts
chowyiyin: Realised the padding is from storybook so this line is not needed

frontend/src/theme/components/Tabs.ts
chowyiyin: Okay, I've left a comment in figma!

frontend/src/theme/components/Tabs.ts
karrui: looking at the figma design, the desktop gap between tab items is 160px and the mobile gap is 32px. you can check the distance between components by selecting the component, then holding the cmd key.

do double check with @pearlyong on the real gap and if it changes

karrui:
Scrolling is not working

chowyiyin:
Scrolling works using mouse pad or touch and drag (use chrome dev tools) and also works if press shift and scroll with mouse. I will have to figure out how to do horizontal scroll directly on mouse wheel. Will update!

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,134 @@
+import {
+  ChakraTheme,
+  ComponentMultiStyleConfig,
+  SystemStyleObjectRecord,
+  ThemingPropsThunk,
+} from '@chakra-ui/react'
+import { getColor } from '@chakra-ui/theme-tools'
+import merge from 'lodash/merge'
+
+export const DRAG_SCROLL_SPEED = 0.8
```
karrui, line 10: I think the speed should be 0.9-1, it feels a little sluggish now

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,134 @@
...
+
+const variantLine: ThemingPropsThunk<SystemStyleObjectRecord, ChakraTheme> =
+  () => {
+    return {
+      root: {
+        w: '100%',
+      },
+      tablist: {
+        w: '100%',
```
karrui, line 19: May need `white-space: nowrap` properties too (following instructions from the use-drag-scroll package)

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,134 @@
...
+        overflowX: 'scroll',
+        borderColor: 'transparent',
+        /* Hide scrollbar for IE and Edge */
+        '&::MsOverflowStyle': 'none',
```
karrui, line 23: I think it is ok to show the scrollbar (since it will only show on scroll)

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,134 @@
...
+        /* Hide scrollbar for Firefox */
+        '&::scrollbarWidth': 'none',
+        /* Hide scrollbar for Chrome, Safari and Opera */
+        '&::-webkit-scrollbar': {
+          display: 'none',
+        },
+        p: '1.5625rem',
```
karrui, line 30: This number seems a little off. We use a 4 point grid (even numbers), and looking at the figma mock the padding seems to be 28px top 22px bottom. Can you double check all the margin figures again?

The total height of this tab bar is also 80px, but 84px in your story.

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,134 @@
...
+  parts: [],
+  baseStyle: {
+    tab: {
+      mr: '2rem',
+    },
+  },
+  sizes: {
+    md: {
+      tab: {
+        textStyle: 'subhead-3',
```
karrui, line 121: I think this should be in the baseStyle instead of textStyle

frontend/src/theme/components/Tabs.ts
```diff
@@ -0,0 +1,134 @@
...
+      _hover: {
+        color: 'secondary.200',
+      },
+      _selected: {
+        color: 'white',
+        _focus: {
+          boxShadow: '0 0 0 2px white',
+        },
+      },
+      transitionProperty: 'common',
```
karrui, line 105: nit: should we add a transitionDuration: 'common' prop too for the delay before animating?

karrui (changes requested):
double check measurements;

a couple things to also look into:
- move focused tab into the view if its partially cut off when focused (especially with keyboard)
  - might want to use [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)
- prevent focus ring from showing up if clicked with a mouse instead of tab-targeted with a keyboard (this one may not be worth the time, you decide)

frontend/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+  variant: 'line-dark',
+}
+
+export const WithScrolling = TabTemplate.bind({})
+WithScrolling.args = {
+  variant: 'line-dark',
+}
+WithScrolling.parameters = {
+  viewport: {
+    defaultViewport: 'mobile1',
```
karrui, line 39: May also want to add Chromatic [viewport parameter](https://www.chromatic.com/docs/viewports) so it renders correctly on Chromatic.

See frontend/src/utils/storybook and YesNo story for example

frontend/src/theme/components/Tabs.ts
chowyiyin: i changed `overflowX` to `auto` but the scrollbar is shown as long as there is overflow. Is that okay?

frontend/src/theme/components/Tabs.ts
chowyiyin: sorry why is it baseStyle?

karrui: textStyle has nothing to do with sizing

frontend/src/theme/components/Tabs.ts
karrui: I think that's fine. Clarify with pearly if unclear!

chowyiyin: I've styled the custom scrollbar! But customisation is limited on Firefox browsers so I just did the best I can. Also, I discussed the padding with pearly and we've decided to use 24px padding so the total height is (24 * 2 + 24 + 4)px = 76px

karrui:
Seems to be some jitter when transitioning from focused state back to unfocused state (click into section and click outside). 
Also nit: possible to only show focus state on keyboard effect? Not sure if the `focus-visible` package can do it.

karrui (approved):
lgtm

pearlyong:
@chowyiyin 
- i'm seeing an a11y violation in storybook "Ensures all ARIA attributes have valid values"
- I tried to keyboard tab them but it stops at the first tab?
- on mobile, after i tap on a tab, it doesn't seem to transition to the Active state? looks like there is some opacity on it, and it only transitions to Active after I tap away from it

mantariksh:
@karrui requesting another review due to latest changes pls!

frontend/src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  // Ensures that when navigating horizontally through tabs,
+  // the focused tab is scrolled horizontally into view
+  const handleFocus = () => {
+    tabRef.current?.scrollIntoView({
+      // center horizontally
+      inline: 'center',
+      // prevent unnecessary vertical scrolling
+      block: 'nearest',
+    })
```
karrui, line 15: This block is causing the display to "jump" when you click on a tab that is already in the viewport; maybe it's due to `use-drag-scroll`?

mantariksh: isn't this the behaviour you requested in [this comment](https://github.com/opengovsg/FormSG/pull/2304#pullrequestreview-705795355)?

karrui: yea, but it's jumping even when the thing is in the viewport, but does not happen all the time, weirdly enough. maybe remove the `inline` prop so it doesn't jump to the center unnecessarily once it's in the viewport?

mantariksh: sure, done in 3c8229d

karrui (approved):
lgtm, wait for @pearlyong's review

PR #2314, opened 2021-07-06
fix(types): recognise User.betaFlags in zod

LoneRifle:
## Context

Beta flags are returned by the backend, but the Zod types used by the frontend
does not recognise them. Make the appropriate change to do so.


karrui (approved):
lgtm

PR #2326, opened 2021-07-07
feat(sms-limiting): check admin sms count on public submission

seaerchin:
## Problem
When the public chooses to generate their OTP, each sms costs formSG money if the sms is sent using formSG credentials. This PR checks if the admin of the form has a certain # of smses sent using formSG's credentials and takes the appropriate action (either warning the admin or disabling the sms verifications totally).

## Solution
This PR chooses to implement the check at the controller level (per OTP verification) rather than on the db schema itself. This was chosen because preventing the toggle on the db schema would forcibly invalidate form duplications if the admin has exceeded sms limits, which should not be the case. 

## Technical Considerations

**Merging of checks and action of disabling**
This was done because of considerations on db querying, where we wanted to avoid repeated db queries. A better API would be to separate the check and the action, but this requires the count, which makes the API clunky, as we need to dynamically decide at the controller what to do. 

This presents an API like so: obtain count -> check based on form + count -> get action based on form + count -> perform action. 

In light of this concern, i felt that merging them into a single method and exposing it on the service makes sense, especially given how specific this method is (should be used only in a single scenario) 

**Breaking Changes** 

    - Details ...


src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -53,42 +55,25 @@ import {
...
-  transferFormOwnership,
-  updateEndPage,
-  updateForm,
-  updateFormCollaborators,
-  updateFormField,
-  updateFormLogic,
-  updateFormSettings,
-  updateStartPage,
-} from '../admin-form.service'
+import * as AdminFormService from '../admin-form.service'
```
seaerchin, line 76: actual tests are at the bottom of this file; this change was made to allow the usage of `jest.spyOn`

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -53,42 +55,23 @@ import {
...
```
seaerchin, line 74: actual tests are at the bottom of this file; this change was made to allow the usage of `jest.spyOn`

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1069,3 +1077,62 @@ export const disableSmsVerificationsForUser = (
...
+
+/**
+ * Checks the number of free smses sent by the admin of the form and deactivates verification or sends mail as required
+ * @param form The form whose admin's sms counts needs to be checked
+ * @returns ok(true) when the verification has been deactivated successfully or no action is required
+ * @returns err(MailGenerationError) when an error occurred on creating the HTML template for the email
+ * @returns err(MailSendError) when an error occurred on sending the email
+ * @returns err(PossibleDatabaseError) when an error occurred while retrieving the counts from the database
+ */
+export const checkFreeSmsSentByAdminAndDeactivateVerification = (
```
karrui, line 1089: Maybe move this into VerificationService since this is only used when verification controller calls this fn

I know this is probably following the deactivate form function signature but this does not really need to be in a single function (just like deactivate form does not need to 😢 ). 
Maybe we can split this into two functions and have the (only) controller call in VerificationController call them both in order.

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1069,3 +1077,62 @@ export const disableSmsVerificationsForUser = (
...
+
+/**
+ * Checks the number of free smses sent by the admin of a form and performs the appropriate action
+ * @param form The form whose admin's sms counts needs to be checked
+ * @returns ok(true) when the action has been performed successfully
+ * @returns err(MailGenerationError) when an error occurred on creating the HTML template for the email
+ * @returns err(MailSendError) when an error occurred on sending the email
+ * @returns err(PossibleDatabaseError) when an error occurred while retrieving the counts from the database
+ */
+export const checkSmsCountAndPerformAction = (
```
karrui, line 1115: Doesn't need to be exported since this is a helper function

if for testing, can just test the exported function. exporting helper functions to test is a small code smell

src/app/modules/form/admin-form/admin-form.service.ts
karrui: Rereading the code, it does not even deactivate the form, only possibly deactivate the form. So the function name is already a misnomer :P.

Idk name suggestions sorry

src/app/modules/verification/__tests__/verification.controller.spec.ts
```diff
@@ -651,6 +670,7 @@ describe('Verification controller', () => {
         recipient: MOCK_ANSWER,
       })
       expect(mockRes.sendStatus).toHaveBeenCalledWith(StatusCodes.CREATED)
+      expect(disableSpy).toBeCalledWith(MOCK_FORM)
     })
 
     it('should return 400 when form SMS parameters are malformed', async () => {
```
karrui, line 676: need add test cases for when `checkFreeSmsSentByAdminAndDeactivateVerification` returns an error

src/app/modules/form/admin-form/admin-form.service.ts
seaerchin: done [here](https://github.com/opengovsg/FormSG/pull/2326/commits/301d20ee1e4a16f06c518f6cf5ee5ff0588ca8a8)

src/app/modules/verification/__tests__/verification.controller.spec.ts
seaerchin: yep, i thought about this but the only observable behaviour which we could test is the logging - is that what you meant?

src/app/modules/form/admin-form/admin-form.service.ts
seaerchin: wrt this, per what we discussed irl, will change the naming but keep the same behaviour for neatness! thinking of `processAdminSmsCounts`, wdyt?

karrui: agree

karrui (approved):
lgtm with exception of long name lol

Issue #2066, opened 2021-06-03
Type DTOs using domain oriented design that client/server types extend from


karrui:
Currently, our DTO types are too coupled with server-side implementation, making the client dependent on server-side modules (such as mongo) to get the appropriate types.

This issue proposes DTO types to be completely free of such dependencies and instead be client/server agnostic. Server side types should instead extend from the DTO types to generate their own typings. 

This change will properly allow types to be shared between frontend and backend without any (possibly incompatible - between commonjs and node implementations, for example) dependencies

## Current implementation
```
        ILogic & Document
            /
        LogicDto <-- depending on Mongo (server) related types 
          /
      client
```

## Proposed implementation
```
       LogicDto
        /   \
       /     \
  client      \
               \
         ILogic & Document
```

karrui:
Accidentally closed by commit message

PR #2355, opened 2021-07-08
ref(shared-types-refactor): move form field types to root shared folder

karrui:
## Problem

Continuation from #2322, this PR moves the form field types to the new shared folder.

Related to #2066

## Solution

**Breaking Changes** 

    - Details ...

**Improvements**:

- move form field types to shared/types/field
- use relocated shared types when instantiating backend field types


karrui:
Will fix the e2e test dependencies...

shared/types/field/tableField.ts
```diff
@@ -0,0 +1,20 @@
+import { FieldBase, BasicField } from './base'
+import { DropdownFieldBase } from './dropdownField'
+import { ShortTextFieldBase } from './shortTextField'
+
+export interface ShortTextColumnBase extends ShortTextFieldBase {
```
mantariksh, line 5: I'm not sure this is right, we don't have `fieldType` for table columns right?

src/types/field/baseField.ts
```diff
@@ -1,32 +1,20 @@
...
-export interface IField {
-  globalId?: string
-  title: string
-  description: string
-  required: boolean
-  disabled: boolean
-  fieldType: BasicField
-  myInfo?: IMyInfo
-}
+export type IField = FieldBase
```
mantariksh, line 14: is it really necessary to reexport all these types as aliases of the `Base` type? also haven't we always been strict about only naming interfaces starting with `I` previously?

src/types/field/attachmentField.ts
```diff
@@ -1,18 +1,14 @@
...
-  SevenMb = '7',
-  TenMb = '10',
-  TwentyMb = '20',
-}
+import { IFieldSchema } from './baseField'
 
-export interface IAttachmentField extends IField {
-  attachmentSize: AttachmentSize
-}
+export { AttachmentSize }
```
mantariksh, line 9: is this really necessary? seems to be complicating our import/exports

src/types/field/attachmentField.ts
```diff
@@ -1,18 +1,14 @@
...
 
-export interface IAttachmentFieldSchema
-  extends IAttachmentField,
-    IFieldSchema {}
+export type IAttachmentField = AttachmentFieldBase
+export interface IAttachmentFieldSchema extends IAttachmentField, IFieldSchema {
+  fieldType: BasicField.Attachment
```
mantariksh, line 13: this is unnecessary because the `Base` type already has the specific `fieldType` right?

src/types/field/emailField.ts
```diff
@@ -1,20 +1,15 @@
...
-  isVerifiable: boolean
-  hasAllowedEmailDomains: boolean
-  allowedEmailDomains: string[]
-}
+export { AutoReplyOptions }
 
+export type IEmailField = EmailFieldBase
 export interface IEmailFieldSchema extends IEmailField, IFieldSchema {
+  fieldType: BasicField.Email
   isVerifiable: boolean
```
mantariksh, line 14: doesn't the base type already contain `isVerifiable`?

src/types/field/index.ts
```diff
@@ -1,25 +1,45 @@
...
+  SpNric = 'SingPass Validated NRIC',
+  CpUid = 'CorpPass Validated UID',
+  CpUen = 'CorpPass Validated UEN',
+}
+
+export enum SgidFieldTitle {
+  SgidNric = 'sgID Validated NRIC',
+}
+
+export { BasicField, MyInfoAttribute, FormField }
```
mantariksh, line 41: again is it really necessary to re-export? is confusing if we can import the same thing from two places

src/types/field/baseField.ts
karrui: I did not want to rename everything in the backend whilst doing this refactor. There should hopefully be a final PR where all these aliases are removed and only the shared names are used. I can also do it in this PR if needed

src/types/field/attachmentField.ts
karrui: This is done to keep all imports at parity, since many `AttachmentSize` imports are imported from this file in the backend. As above, I can re-export everything via a central location in `src/types` and correct all import paths if desired

src/types/field/attachmentField.ts
karrui: Needed, since IFieldSchema has `fieldType: BasicField`. This is reasserting that we want to use IAttachmentField's `fieldType`

src/types/field/emailField.ts
karrui: Same reasoning, IFieldSchema has `isVerifiable?: boolean` and we want to assert that it is no longer optional (following IEmailField)

src/types/field/baseField.ts
mantariksh: ah ok sounds good!

mantariksh: separate PR is good, as long as we do it eventually

shared/types/field/tableField.ts
karrui: Good catch, I only saw the discriminator function, grrr and forgot that the fieldType is discriminated against and not in the value itself. Will fix

karrui: Fixed in 395149a

karrui: Fixed in ca42de9

mantariksh (approved):

PR #2359, opened 2021-07-09
build: release v5.19.2 - fix form sharing

mantariksh:
Fixes bug where frontend was seeing all forms as having an SGID beta feature, resulting in form template sharing breaking across the app.


mantariksh:
temporarily closing in favour of individual fix PRs, will reopen when ready

PR #2362, opened 2021-07-11
fix: allow myinfo authtype for logout

### LINKED ISSUE #2360: Attempting to log out of MyInfo form results in 400 error
This is because logging out with any AuthType results in a call to the /logout endpoint, which disallows MyInfo as a valid authType.

Possible solutions:
1. Allow `MyInfo` authType at /logout endpoint. Also need to ensure the correct cookie name is cleared, as the MyInfo cookie name is not present in the `JwtName` enum. This will probably work, but need to test because we do set some custom settings on the MyInfo cookie, so there's a chance that simply calling `res.clearCookie` may not work.
2. In the frontend, when user clicks the logout button for a MyInfo form, don't call the /logout endpoint. Instead simply force a reload, upon which the /publicform endpoint will clear the cookie.

Prefer solution 1 as it does not couple logout behaviour to the /publicform endpoint by implicitly relying on that endpoint to clear the cookie. Also because a server-side fix will work for old clients.

tshuli:
## Problem

Closes #2360

## Solution
- Allow MyInfo authtype for `/logout` endpoint and update `JwtName` type to include MyInfoCookie

## Tests
**(tested on staging)**
- [ ] Create myinfo form and publish. Login. Check that you can logout before submission and `MyInfoCookie` is deleted
- [ ] Login again. Check that you can submit, and you are auto logged out after submission

### DIFF
--- a/src/app/modules/form/public-form/public-form.controller.ts
+++ b/src/app/modules/form/public-form/public-form.controller.ts
@@ -492,7 +492,7 @@ export const handleFormAuthRedirect = [
  * @returns 200 with success message when user logs out successfully
  * @returns 400 if authType is invalid
  */
-export const _handleSpcpLogout: ControllerHandler<
+export const _handlePublicAuthLogout: ControllerHandler<
   { authType: AuthType.SP | AuthType.CP | AuthType.MyInfo | AuthType.SGID },
   PublicFormAuthLogoutDto
 > = (req, res) => {
@@ -521,7 +521,7 @@ export const _handleSpcpLogout: ControllerHandler<
  * Handler for /forms/auth/:authType/logout
  * Valid AuthTypes are SP / CP / MyInfo / SGID
  */
-export const handleSpcpLogout = [
+export const handlePublicAuthLogout = [
   celebrate({
     [Segments.PARAMS]: Joi.object({
       authType: Joi.string()
@@ -529,7 +529,7 @@ export const handleSpcpLogout = [
         .required(),
     }),
   }),
-  _handleSpcpLogout,
+  _handlePublicAuthLogout,
 ] as ControllerHandler[]
 
 /**
--- a/src/app/modules/form/public-form/public-form.controller.ts
+++ b/src/app/modules/form/public-form/public-form.controller.ts
@@ -498,21 +498,22 @@ export const _handleSpcpLogout: ControllerHandler<
 > = (req, res) => {
   const { authType } = req.params
 
-  if (authType === AuthType.MyInfo) {
-    return res
-      .clearCookie(MYINFO_COOKIE_NAME)
-      .status(200)
-      .json({ message: 'Successfully logged out.' })
-  } else if (authType === AuthType.SGID) {
-    return res
-      .clearCookie(SGID_COOKIE_NAME)
-      .status(200)
-      .json({ message: 'Successfully logged out.' })
-  } else {
-    return res
-      .clearCookie(JwtName[authType])
-      .status(200)
-      .json({ message: 'Successfully logged out.' })
+  switch (authType) {
+    case AuthType.MyInfo:
+      return res
+        .clearCookie(MYINFO_COOKIE_NAME)
+        .status(200)
+        .json({ message: 'Successfully logged out.' })
+    case AuthType.SGID:
+      return res
+        .clearCookie(SGID_COOKIE_NAME)
+        .status(200)
+        .json({ message: 'Successfully logged out.' })
+    default:
+      return res
+        .clearCookie(JwtName[authType])
+        .status(200)
+        .json({ message: 'Successfully logged out.' })
   }
 }
 
--- a/src/app/modules/form/public-form/public-form.controller.ts
+++ b/src/app/modules/form/public-form/public-form.controller.ts
@@ -498,17 +498,22 @@ export const _handleSpcpLogout: ControllerHandler<
 > = (req, res) => {
   const { authType } = req.params
 
-  const cookieName =
-    authType === AuthType.MyInfo
-      ? MYINFO_COOKIE_NAME
-      : authType === AuthType.SGID
-      ? SGID_COOKIE_NAME
-      : JwtName[authType]
-
-  return res
-    .clearCookie(cookieName)
-    .status(200)
-    .json({ message: 'Successfully logged out.' })
+  if (authType === AuthType.MyInfo) {
+    return res
+      .clearCookie(MYINFO_COOKIE_NAME)
+      .status(200)
+      .json({ message: 'Successfully logged out.' })
+  } else if (authType === AuthType.SGID) {
+    return res
+      .clearCookie(SGID_COOKIE_NAME)
+      .status(200)
+      .json({ message: 'Successfully logged out.' })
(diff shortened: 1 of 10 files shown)


src/app/modules/spcp/spcp.types.ts
```diff
@@ -4,6 +4,7 @@ export enum JwtName {
   SP = 'jwtSp',
   CP = 'jwtCp',
   SGID = 'jwtSgid',
+  MyInfo = 'MyInfoCookie',
```
mantariksh, line 7: I'm not sure this is the right fix. notice that the enum is defined in the SPCP module, because we deliberately want to keep SPCP and MyInfo decoupled (actually SGID shouldn't be here either). we should avoid a situation where we accidentally loosen the types in the SPCP module just for this fix. what do you think about just adding a switch statement in the controller to determine what cookie name to clear?

tshuli: thanks! agree, have cleaned up

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -485,33 +486,40 @@ export const handleFormAuthRedirect = [
...
   PublicFormAuthLogoutDto
 > = (req, res) => {
   const { authType } = req.params
 
+  const cookieName =
+    authType === AuthType.MyInfo
+      ? MYINFO_COOKIE_NAME
+      : authType === AuthType.SGID
+      ? SGID_COOKIE_NAME
+      : JwtName[authType]
```
mantariksh, line 506: ok for now but I generally don't favour nested ternary operators. hard to read and parse the logic. I would prefer it if we use `let` with a `switch` statement, or just repeat the `res.clearCookie` statement as necessary

src/app/modules/sgid/sgid.constants.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Name of cookie which contains state of sgid login, and access
+ * token if login was successful.
+ */
+export const SGID_COOKIE_NAME = 'jwtSgid'
```
mantariksh, line 5: I noticed this string is hardcoded elsewhere in the backend; if we are making this change in this PR, can we use this constant there too?

tshuli: thanks, made the changes

src/app/modules/form/public-form/public-form.controller.ts
tshuli: not keen on `let` so I went with `if` + repeat, see if ok

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -485,33 +486,45 @@ export const handleFormAuthRedirect = [
...
+      .clearCookie(MYINFO_COOKIE_NAME)
+      .status(200)
+      .json({ message: 'Successfully logged out.' })
+  } else if (authType === AuthType.SGID) {
+    return res
+      .clearCookie(SGID_COOKIE_NAME)
+      .status(200)
+      .json({ message: 'Successfully logged out.' })
+  } else {
+    return res
```
mantariksh, line 512: if you're really not keen on a switch statement, can we at least take this out of the `else` block so that there is a return guaranteed? I still a switch is cleaner since you're repeating the return statements anyway:
```
switch (authType) {
  case AuthType.MyInfo:
    return res
      .clearCookie(MYINFO_COOKIE_NAME)
      .status(200)
      .json({ message: 'Successfully logged out.' })
  case AuthType.SGID:
    return res
      .clearCookie(SGID_COOKIE_NAME)
      .status(200)
      .json({ message: 'Successfully logged out.' })
  default:
    return res
      .clearCookie(JwtName[authType])
      .status(200)
      .json({ message: 'Successfully logged out.' })
}
```

mantariksh: ........................ok la but this doesn't guarantee a return statement also LOL

mantariksh (approved):

tshuli:
switched :)

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -485,33 +486,46 @@ export const handleFormAuthRedirect = [
...
-  return res
-    .clearCookie(JwtName[authType])
-    .status(200)
-    .json({ message: 'Successfully logged out.' })
+  switch (authType) {
+    case AuthType.MyInfo:
+      return res
+        .clearCookie(MYINFO_COOKIE_NAME)
+        .status(200)
+        .json({ message: 'Successfully logged out.' })
```
karrui, line 506: Should we create a new service function in PublicFormService to retrieve the cookie name so the controller does not need to know these business logic? For example:

```ts
// public-form.service.ts
const getCookieNameByAuthType = (authType: AuthType) => {
  switch (authType):
    case AuthType.MyInfo:
      return MYINFO_COOKIE_NAME
   ...
}

// public-form.controller.ts
// here
return res.clearCookie(getCookieNameByAuthType(authType).status(200).json({...})
```

Cleans up the repetitive nature

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -485,33 +486,46 @@ export const handleFormAuthRedirect = [
...
 /**
  * NOTE: This is exported only for testing
- * Logs user out of SP / CP By deleting cookie
+ * Logs user out of SP / CP / MyInfo / SGID by deleting cookie
  * @param authType type of authentication
  *
  * @returns 200 with success message when user logs out successfully
  * @returns 400 if authType is invalid
  */
 export const _handleSpcpLogout: ControllerHandler<
```
karrui, line 495: Should we rename this function now that it handles all auth types? :X

tshuli: renamed, how about `handlePublicAuthLogout`?

src/app/modules/form/public-form/public-form.controller.ts
tshuli: nice, thanks! edited

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -31,17 +31,16 @@ import {
...
 import { validateSgidForm } from '../../sgid/sgid.util'
 import { InvalidJwtError, VerifyJwtError } from '../../spcp/spcp.errors'
 import { SpcpService } from '../../spcp/spcp.service'
-import { JwtName } from '../../spcp/spcp.types'
 import { getRedirectTarget, validateSpcpForm } from '../../spcp/spcp.util'
 import { AuthTypeMismatchError, PrivateFormError } from '../form.errors'
 import * as FormService from '../form.service'
 
 import * as PublicFormService from './public-form.service'
+import { getCookieNameByAuthType } from './public-form.service'
```
karrui, line 43: Duplicate imports?

karrui (approved):
lgtm with one last change

src/app/modules/form/public-form/public-form.controller.ts
tshuli: thanks! edited

PR #2369, opened 2021-07-12
fix(BetaService): make SGID form-level beta feature

### LINKED ISSUE #2370: Creation of form from template is broken
Any attempt to create form from template results in the following error:

image (3)

mantariksh:
## Problem

#2370 happened because the frontend `BetaService` is coupled to beta _fields_, as opposed to form-level beta features such as SGID.

Closes #2370 

## Solution

Generalise `BetaService` to be able to apply to any _form_-level feature. Also, return the form `authType` from the examples page queries so we are able to check the auth type when seeing whether the user has permissions to duplicate the form.

## Tests
- [x] Creating non-SGID form from Examples page works
- [x] Creating non-SGID form from template works
- [x] Duplicating non-SGID form works
- [x] Creating SGID form from template **does not work** (assuming user does not have SGID beta flag)


karrui (approved):

karrui (approved):
lgtm

PR #2380, opened 2021-07-13
feat(form-v2): tiles component

### LINKED ISSUE #2024: [React] Add Tiles component
**Time estimation: 1d**

seaerchin:
## Problem
We require a tile component to display info compactly. This PR adds the `Tile` component as specified in the Figma design [here](https://www.figma.com/file/u00dlQLsaQfvqv6ot5ZrJZ/Form-Design-System?node-id=2153%3A4121).

Closes #2024 

## Solution
The tile component is a _wrapper_ component - i.e, its styles are decided by us and the people who use it are not given (as much) control over the underlying styling of the component itself. The general approach towards designing this component will be laid out below

**Tag**
Allow users to pass in tag as a component rather than specifying the tag's styles as props for the user to pass in. This gives the user more control of how they want the tag to look and makes the tiles component simpler also. 

**Description**
Instead of having a specific description prop (like tag), the more general `children` prop was chosen. This is because the `Tile` component is more of a _wrapper_ component, and it makes more sense to have children as the description rather than separately defining description and then allowing a children prop, which does not make sense. 

**Variant**
As variants are determined based on whether there is a description, we can determine this automatically instead of adding extra props. A type is defined for safety and clarity.

**Questions for reviewers**
1. How much stylistic power should be exposed on the component? The gauge i used here was that the figma did not seem to allow for much stylistic deviation and i thought that the tiles component was to be used 'as is' rather than having extensive customization done hence, did not expose any style props. 
2. Is the approach taken towards designing this component logical?

**Breaking Changes** 

    - Details ...

**Screenshots**
This is available in the generated storybook snapshot

**TODO**
add more stories


frontend/src/components/Tile/Tile.tsx
```diff
@@ -0,0 +1,67 @@
...
+type TileVariant = 'complex' | 'simple'
+export const Tile = forwardRef<TileProps, 'div'>(
+  ({ tag, icon, title, subtitle, children }, ref) => {
+    const variant: TileVariant = children ? 'complex' : 'simple'
+    const styles = useMultiStyleConfig('Tile', { variant })
+
+    return (
+      // Ref passed into the component as a whole so that it can be focused
+      <Box sx={styles.container} ref={ref}>
+        <VStack spacing="1rem" alignItems="flex-start">
```
chowyiyin, line 50: Should we be putting all styles into the theme?

frontend/src/components/Tile/Tile.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <HStack spacing="1rem">
+            <Icon sx={styles.icon} as={icon}></Icon>
+            {tag}
+          </HStack>
+          <Box>
+            <Text sx={styles.title} as="h4">
```
chowyiyin, line 56: Do we need to put `as="h4"` here?

chowyiyin (commented):
The text spills out of the tile when its too long
Screenshot 2021-07-14 at 11 06 18 AM
Maybe we can add `overflow-wrap` to the theme?

frontend/src/components/Tile/Tile.tsx
seaerchin: i'm not too sure wym! i think it's present alr @__@

frontend/src/components/Tile/Tile.tsx
```diff
@@ -0,0 +1,67 @@
...
+              {title}
+            </Text>
+            <Text sx={styles.subtitle}>{subtitle}</Text>
+          </Box>
```
karrui, line 60: Can export the Description, Title and Subtitle wrappers as a subcomponent so outside users can build their own tag with their own styling (if needed)

frontend/src/theme/components/Tile.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const Tile: ComponentMultiStyleConfig = {
+  parts: ['container', 'title', 'icon', 'subtitle'],
+  baseStyle: {
+    container: {
+      borderRadius: '0.25rem',
+      padding: '1.5rem',
+      _active: {
+        bgColor: 'primary.200',
+        border: '3px solid var(--chakra-colors-primary-500)',
```
karrui, line 11: Since the box sizing has not changed, this causes the component to "squish" when activated.

frontend/src/components/Tile/Tile.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+type TileVariant = 'complex' | 'simple'
+export const Tile = forwardRef<TileProps, 'div'>(
+  ({ tag, icon, title, subtitle, children }, ref) => {
+    const variant: TileVariant = children ? 'complex' : 'simple'
+    const styles = useMultiStyleConfig('Tile', { variant })
+
+    return (
+      // Ref passed into the component as a whole so that it can be focused
+      <Box sx={styles.container} ref={ref}>
```
karrui, line 49: Should this be a html button (with all the a11y stuff) instead? It is a selectable component.

frontend/src/components/Tile/Tile.tsx
seaerchin: i agree with putting _parts_ of the component into the theme; however, i think this is more of a utility component (thus, not a part) and should be left as is, cos otherwise our themes run the risk of being overloaded @__@

i'll experiment with using `VStack` as the outer component and removing teh box so that we can put it in themes!

seaerchin:
> The text spills out of the tile when its too long
> Screenshot 2021-07-14 at 11 06 18 AM
> Maybe we can add `overflow-wrap` to the theme?

pepehands ;--; will do!!

karrui (changes requested):

frontend/src/components/Tile/Tile.tsx
karrui: this causes the text to be a h4 html component. I personally don't think it's needed haha

frontend/src/components/Tile/Tile.tsx
karrui: Can always use the `_first` selector to only apply the styling to the first child

frontend/src/components/Tile/Tile.tsx
seaerchin: great point; changed

frontend/src/theme/components/Tile.ts
seaerchin: fixed using the `boxShadow` trick!

frontend/src/components/Tile/Tile.tsx
seaerchin: thanks for suggestion! i removed the tag; h4 defines a [header](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) and perhaps this might not be suited for this after all!

frontend/src/components/Tile/Tile.tsx
seaerchin: could i clarify on this point - i actually decided _against_ doing this for the reasons stated in the PR write-up and i wanted to know when this is preferred (or is this all the time b/c we r building our own design system)

karrui: > This is because the Tile component is more of a wrapper component

Precisely what you said in your PR description? So you expose `Tile.Tile`, `Tile.Title`, `Title.Description`, in addition to the main `Tile` component that uses all of those things precisely so users can modify any part of the component if they want instead of building another one, since all it is is a wrapper containing various styling

seaerchin: gotcha - to confirm this: wrapper components should export their wrapped components because it helps people to style components if needed; but wrt this, we shouldn't export  a `Description` component directly right? 

primarily because this is a wrapper and doesn't contribute anything of value beyond the chakra tags so we should export the components that are different (i took this approach)

seaerchin:
> > The text spills out of the tile when its too long
> > Screenshot 2021-07-14 at 11 06 18 AM
> > Maybe we can add `overflow-wrap` to the theme?
> 
> pepehands ;--; will do!!

@chowyiyin this is cursed... i tried what's documented [here](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/) as well as reading [mdn docs](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) for overflow-wrap.

i'm not too sure what's causing the issue actually and any help would be appreciated ;--; 

The main issue is because of `maxWidth` being set which, if removed, will not have overflow because the container stretches to fit.

frontend/src/theme/components/Tile.ts
```diff
@@ -0,0 +1,65 @@
...
+    borderColor: 'transparent',
+    // NOTE: For the boxShadow styling, due to conflicts with the focus-visible package,
+    // the !important is required to display the boxShadow styling correctly.
+    boxShadow: '0 0 0 2px var(--chakra-colors-primary-500) !important',
+  },
+  bgColor: 'white',
+  border: '1px solid',
+  borderColor: 'neutral.300',
+  flexDir: 'column',
+  alignItems: 'flex-start',
```
chowyiyin, line 30:
```suggestion
  whiteSpace: 'normal',
  display: 'inline-block',
  wordWrap: 'break-word',
  maxWidth: '20.75rem',
```

chowyiyin: I think this solves the overflow problem! Just add `textAlign: left` to `title`, `subtitle` and probably add another part for `TileText` to define `textAlign: left` and the color that you've defined inline in the `TileText` component :)

seaerchin: ok i think i understand how this happens! from poring over mdn docs, `maxWidth` means exactly that. after the component's `width` exceeds the `maxWidth`, it is deem to have overflowed. 

This then ties into the `whiteSpace` [behaviour](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space), which determines how to layout the elements wrt their whitespace. 

thanks for your help!

frontend/src/theme/components/Tile.ts
```diff
@@ -0,0 +1,67 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+const activeContainerStyle = {
+  bgColor: 'primary.200',
+  borderColor: 'transparent',
+  // NOTE: For the boxShadow styling, due to conflicts with the focus-visible package,
+  // the !important is required to display the boxShadow styling correctly.
+  boxShadow: '0 0 0 3px var(--chakra-colors-primary-500) !important',
```
karrui, line 8: Should we add an `_hover` state for active that has `bg: 'primary.200'`  so hovering a selected tile won't have any jarring color changes?

frontend/src/theme/components/Tile.ts
```diff
@@ -0,0 +1,67 @@
...
+  alignItems: 'flex-start',
+  maxWidth: 'inherit',
+  textAlign: 'left',
+}
+
+export const Tile: ComponentMultiStyleConfig = {
+  parts: ['container', 'title', 'icon', 'subtitle'],
+  baseStyle: ({ isActive }) => ({
+    container: isActive
+      ? { ...baseContainerStyle, ...activeContainerStyle }
```
karrui, line 39: nit: do we really need two constants for this? should we inline them instead for something that is easier to read?

i.e.

```ts
container: {
  bg: isActive? 'primary.200' : 'white',
  ...
}
```

This allows the reader to know exactly what the differences between active and non-active states are without having to scroll up and down between the two constants and see what is being replaced

frontend/src/components/Tile/Tile.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+Simple.args = {
+  variant: 'simple',
+  title: 'Simple',
+  subtitle: 'Receive responses in forms',
+  icon: BiMailSend,
+}
+
+export const Playground: Story = () => {
+  const [isStorageClicked, setIsStorageClicked] = useState<boolean>(false)
+  const [isEmailClicked, setIsEmailClicked] = useState<boolean>(false)
```
karrui, line 98: nit: can be simplified

```tsx
const [selected, setSelected] = useState('')

...
<StorageTile
  onClick={() => setSelected('storage')
  isActive={selected === 'storage'}
/>
```

frontend/src/components/Tile/Tile.stories.tsx
```diff
@@ -0,0 +1,162 @@
+import { useState } from 'react'
+import { BiLockAlt, BiMailSend } from 'react-icons/bi'
+import { Stack, UnorderedList } from '@chakra-ui/react'
+import { Meta, Story } from '@storybook/react'
+import _ from 'lodash'
```
karrui, line 5: nit: import specific function instead of the whole library?

karrui (approved):
some nits, fix the hover styling on active state and we are gtg!

frontend/src/components/Tile/Tile.stories.tsx
seaerchin: sure!

frontend/src/components/Tile/Tile.stories.tsx
seaerchin: good catch!! idk why i didn't think of this actually .-.

frontend/src/theme/components/Tile.ts
seaerchin: let me check back with pearly on this!

frontend/src/theme/components/Tile.ts
seaerchin: i did this because the activeState is shared between `_active` and when `isActive` is true! i think this is worth the overhead but lmk if this is impt to change!

frontend/src/theme/components/Tile.ts
seaerchin: clarified n changed!

Issue #30, opened 2020-08-03
Logic for checkboxes


r00dgirl:
**Background**

Checkbox is the next most popular limited-choice field after those supported by logic - at 6.8% of all occurrences it is about as popular as Yes/No. The checkbox field serves unique use cases where admins want respondents to be able to select up to _n_ choices, instead of just one choice. 

Currently it is not feasible to set a logic condition for a respondent's choices having included a particular value. 
- the logic-based workaround is very unwieldy, you would have to ask about each choice as its own individual question, at the expense of respondent UX
- more likely you would just do away with logic, and just issue a text-based prompt to respondents to answer a further question if it is applicable to them, especially if you have lots of checkbox options

**Discuss**

'includes' enum for checkbox fieldType

mantariksh:
bumping priority as engineering design is already complete, so this is an easy win which we can get while the engineering design is still up to date

mantariksh:
design doc:
https://docs.google.com/document/d/1ggEb9UyHjK3EipQjXpCmOwaenM6_f1AqcE46VxBkIZI/edit

closed AngularJS PR:
https://github.com/opengovsg/FormSG/pull/2389

tshuli:
Moved to linear

PR #2381, opened 2021-07-13
feat: part one checkbox logic - logic module

chowyiyin:
## Problem
This PR implements the extension of the logic feature to logic fields. This first part defines the types needed for this logic extension and implements the logic checks for checkboxes.

Part 1/3 of #30

## Solution
Handling of checkbox fields in logic is done separately from existing logic fields and hence should not affect correctness of existing logic. 

### Unit tests
Added tests to `logic.spec.ts`

### DIFF
--- a/src/shared/util/logic-utils/checkbox/checkbox-response-value-transformer.ts
+++ b/src/shared/util/logic-utils/checkbox/checkbox-response-value-transformer.ts
@@ -1,0 +1,139 @@
+/* eslint-disable typesafe/no-throw-sync-func */
+import { cloneDeep, omit } from 'lodash'
+
+import {
+  BasicField,
+  FieldSchemaOrResponse,
+  ICheckboxField,
+  ICheckboxResponse,
+  IField,
+  ILogicCheckboxResponse,
+  ILogicClientFieldSchema,
+  ILogicInputClientSchema,
+  IPopulatedForm,
+} from '../../../../types'
+import { isCheckboxField } from '../../../../types/field/utils/guards'
+import { hasProp } from '../../has-prop'
+import { LogicFieldSchemaOrResponse } from '../../logic'
+
+/**
+ * Adaptor function to transform checkbox field response values into checkbox condition shape
+ * @param field Logic field from either the backend or frontend
+ * @param form
+ * @returns Checkbox field response in checkbox condition shape
+ * @throws Error if checkbox value has invalid shape or if field cannot be found in the form
+ */
+export const transformCheckboxForLogic = (
+  field: FieldSchemaOrResponse,
+  formFields: IPopulatedForm['form_fields'],
+): LogicFieldSchemaOrResponse => {
+  field = cloneDeep(field) // clone field to prevent changes to original
+  if (isClientCheckboxValue(field)) {
+    return convertClientCheckboxValue(field, formFields)
+  } else if (isServerCheckboxValue(field)) {
+    return convertServerCheckboxValue(field, formFields)
+  } else {
+    // eslint-disable-next-line typesafe/no-throw-sync-func
+    throw new Error('Checkbox value shape is invalid')
+  }
+}
+
+type ClientCheckboxField = Omit<ILogicInputClientSchema, 'fieldValue'> & {
+  fieldValue: boolean[]
+}
+
+type TransformedClientCheckboxField = ILogicClientFieldSchema
+
+const isClientCheckboxValue = (
+  field: FieldSchemaOrResponse,
+): field is ClientCheckboxField => {
+  if (field.fieldType !== BasicField.Checkbox) {
+    return false
+  }
+  return (
+    hasProp(field, 'fieldValue') &&
+    Array.isArray(field.fieldValue) &&
+    (field.fieldValue as unknown[]).every((value) => typeof value === 'boolean')
+  )
+}
+
+const convertClientCheckboxValue = (
+  field: ClientCheckboxField,
+  formFields: IPopulatedForm['form_fields'],
+): TransformedClientCheckboxField => {
+  const completeField = getCheckboxField(field, formFields)
+  const others = field.fieldValue[completeField.fieldOptions.length]
+  const options = completeField.fieldOptions.filter(
+    (_, i) => field.fieldValue[i],
+  )
+
+  return {
+    ...omit(field, ['fieldValue']),
+    fieldValue: { options, others },
+  }
+}
+
+type ServerCheckboxField = ICheckboxResponse
+
+type TransformedServerCheckboxField = ILogicCheckboxResponse
+
+const isServerCheckboxValue = (
+  field: FieldSchemaOrResponse,
+): field is ServerCheckboxField => {
+  if (field.fieldType !== BasicField.Checkbox) {
+    return false
+  }
+  return (
+    hasProp(field, 'answerArray') &&
+    Array.isArray(field.answerArray) &&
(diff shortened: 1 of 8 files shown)


src/types/field/baseField.ts
```diff
@@ -47,7 +47,7 @@ export interface IFieldSchema extends IField, Document {
 // needs it as a variable to store the client's answer to a field.
 // Hence we need this interface for client-side fields.
 export interface IClientFieldSchema extends IFieldSchema {
-  fieldValue: string
+  fieldValue: string | string[] | boolean[]
```
mantariksh, line 50: it seems we are trying to generalise this to all field types, but clearly `IClientFieldSchema` is only used for logic. maybe we just need to generalise it to logic fields.

src/types/form_logic.ts
```diff
@@ -107,6 +131,57 @@ export type LogicCondition =
...
+export type ILogicClientFieldSchema =
+  | IClientFieldSchema
+  | ClientLogicCheckboxResponse
+
+// Type for backend fields after being transformed and passed into the logic module
+export type LogicFieldResponse =
+  | ISingleAnswerResponse
+  | ILogicCheckboxResponse
+  | ITableResponse
+  | IAttachmentResponse
```
mantariksh, line 184: does it make things easier if we leave these out? since tables and attachments can't be used in logic anyway

src/shared/util/logic.ts
```diff
@@ -334,20 +358,34 @@ const isConditionFulfilled = (
...
+      field.fieldType === BasicField.Checkbox &&
+      isLogicCheckboxCondition(condition) &&
+      isCheckboxConditionValue(currentValue)
+    ) {
+      condition.value.forEach((obj) => {
+        obj.options = obj.options.sort()
+      })
+      currentValue.options = currentValue.options.sort()
+      return condition.value.some((val) =>
+        isEqual(JSON.stringify(currentValue), JSON.stringify(val)),
```
mantariksh, line 381: seems a bit sus to just stringify, makes this susceptible to bugs where keys/arrays are not sorted, or there are extra keys etc. can we either use a lodash function to check equality or write a helper function to check equality in the keys that we care about?

src/shared/util/logic-utils/checkbox/checkbox-response-value-transformer.ts
```diff
@@ -0,0 +1,139 @@
...
+  formFields: IPopulatedForm['form_fields'],
+): LogicFieldSchemaOrResponse => {
+  field = cloneDeep(field) // clone field to prevent changes to original
+  if (isClientCheckboxValue(field)) {
+    return convertClientCheckboxValue(field, formFields)
+  } else if (isServerCheckboxValue(field)) {
+    return convertServerCheckboxValue(field, formFields)
+  } else {
+    // eslint-disable-next-line typesafe/no-throw-sync-func
+    throw new Error('Checkbox value shape is invalid')
```
mantariksh, line 37: where does this error propagate up to? does it get caught safely and without affecting the functionality of the form?

src/shared/util/logic-utils/checkbox/checkbox-response-value-transformer.ts
```diff
@@ -0,0 +1,139 @@
...
+const getCheckboxField = (
+  field: FieldSchemaOrResponse,
+  formFields: IPopulatedForm['form_fields'],
+): ICheckboxField => {
+  const completeField = findField(field, formFields)
+  if (!completeField) {
+    throw new Error('Field cannot be found')
+  }
+  if (!isCheckboxField(completeField)) {
+    throw new Error(`${completeField} is not a checkbox field`)
```
mantariksh, line 136: where are these errors caught?

src/shared/util/logic-utils/checkbox/checkbox-response-value-transformer.ts
```diff
@@ -0,0 +1,139 @@
...
+    // eslint-disable-next-line typesafe/no-throw-sync-func
+    throw new Error('Checkbox value shape is invalid')
+  }
+}
+
+type ClientCheckboxField = Omit<ILogicInputClientSchema, 'fieldValue'> & {
+  fieldValue: boolean[]
+}
+
+type TransformedClientCheckboxField = ILogicClientFieldSchema
```
mantariksh, line 45: this type aliasing is pretty confusing, can we just stick to one name for one type?

mantariksh (changes requested):
1. combine the 3 PRs into one. I think you've split them by stage of development (each PR is sort of like a commit) instead of splitting them into multiple parts which can be understood independently. one symptom of this is that I have to click across all 3 PRs in order to see what `formatFieldsForLogic` is doing. I understand the intention to make the PRs more bite-sized, and that's absolutely the right intention, but that approach only makes sense if the individual PRs can be understood in isolation.
2. I think we're taking on a significant amount of complexity because the adapter layer is inside the logic module. this isn't quite in the spirit of the adapter design pattern - the idea of that pattern is that the data is already transformed by the time it reaches the layer which needs to read it (in this case, the logic module). I think it would be good to separate the adapter layers for frontend and backend and transform the data into the correct shape before it reaches the logic module itself. this has the additional advantages of allowing the server and client adapters to evolve independently and also allowing the reader to only keep one transformation in their head at once.
3. the proliferation of nearly-identical-but-slightly-different types is quite unmaintainable because it's very hard to keep track of the purpose of each type. the difficulty of naming the types clearly exacerbates this. we can address this in a few ways: (a) move types closer to where they are used so they're easier to keep track of (b) when a transformation applies to only one key in an object (e.g. `fieldValue`, `value` etc), write functions which mould only that particular key into its desired shape rather than functions which return an entirely new object with that key changed.

src/types/form_logic.ts
```diff
@@ -107,6 +125,61 @@ export type LogicCondition =
...
+
+// Type for fields that are passed into the logic module
+export type FieldSchemaOrResponse = ILogicInputClientSchema | FieldResponse
+
+// Type for client logic fields after transformation
+export interface ILogicClientFieldSchema
+  extends Omit<ILogicInputClientSchema, 'fieldValue'> {
+  // Use omit instead of directly extending IFieldSchema
+  // to prevent typescript from complaining about return type in adaptor function
+  fieldValue: string | CheckboxConditionValue
```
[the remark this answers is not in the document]
chowyiyin: I don’t think this will solve the problem because we will still have to assign the value to the field object in the end, which will ultimately require the definition of a type for it.

src/shared/util/logic-utils/checkbox/checkbox-response-value-transformer.ts
chowyiyin: I've added a try catch block to the places where the adaptors are called (i.e. in `ParsedResponsesObject.class.ts`, `submit-form.directive.js`and `IncomingEncryptSubmission.class.ts`

src/shared/util/logic.ts
```diff
@@ -276,14 +289,24 @@ const isLogicUnitSatisfied = (
...
 
 const getCurrentValue = (
-  field: LogicFieldSchemaOrResponse,
-): string | null | undefined | string[] => {
+  field: FieldSchemaOrResponse,
+):
+  | string
+  | string[]
+  | boolean[]
+  | ITableRow[]
```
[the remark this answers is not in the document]
chowyiyin: Done! I’ve added a function to check that the fields in `isConditionFulfilled` are logic fields. This is called before calling `getCurrentValue` so that we can narrow the types there.

PR #2384, opened 2021-07-13
feat(form-v2): badge component

### LINKED ISSUE #2015: [React] Add Tag component
**Time estimation: 0.5d**

seaerchin:
Kar Rui EDIT: Our Tag figma design is more closely aligned to a Badge in Chakra -- theming Badge instead

## Problem
This PR implements the `Tag` component as described in the figma [here](https://www.figma.com/file/u00dlQLsaQfvqv6ot5ZrJZ/Form-Design-System?node-id=2177%3A8102)

Closes #2015 

## Solution
The `Tag` component as presented in the figma only requires thematic changes. Hence, only theming is changed with regard to the chakra component.

Additionally, because Chakra's `Tag` components take additional properties that determine the colour/fill etc of the tag, these are omitted in the exported interface. 

**Theming**
Chakra's `extendTheme` recursively merges keys with the object provided. Hence, even though `parts` is empty in the theme, it will use the inbuilt `parts` property for chakra's own `Tag`.

**Breaking Changes** 

    - Details ...

## Screenshots
available in the generated storybook


frontend/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { Tag as ChakraTag, TagProps as ChakraTagProps } from '@chakra-ui/react'
+
+type TagVariants = 'primary' | 'secondary'
```
karrui, line 4: this should be exported from the theme file.

also, primary and secondary are current color schemes and thus may be confusing. Suggest "solid" and "light" variant

karrui: Please do clarify with pearly as to whether we allow tags to be of other colors too. I think there is value in creating other colors (see current AngularJS beta tag)

karrui (changes requested):

frontend/src/components/Tag/Tag.tsx
seaerchin: pearly mentioned 2 for now; might be more in the future. fwiw, she said it'll be in our colour palette so i'm tempted to leave it as is for now and when the additional colors come in, add them to the tag theming. wdyt?

karrui: primary and secondary are not variants though, should still rename the variants into non-colorscheme related strings

seaerchin: ps ya sorry for ambiguity; i agree with the theme export and i was just asking about colours!

frontend/src/theme/components/Tag.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+export type TagVariants = 'solid' | 'light'
+
+export const Tag: ComponentMultiStyleConfig = {
+  parts: [],
+  variants: {
```
karrui, line 7: looking at the figma, there is a 4px padding on all sides.
the textStyle should also be `caption-1`

karrui (changes requested):

frontend/src/theme/components/Tag.ts
seaerchin: changed! there's a point to note wrt the 4px padding - chakra's internal padding means that this isn't consistent throughout.

the 2 ways which we can go about doing this is to either add a custom key onto the `size` property within the styling or to override chakra's `size` theming uniformly from `sm` to `lg`. 

because our figma only shows a single size, i've opted for the former (preserves chakra functionality wrt sizing) but depending on what pearly says wrt sizing, i might opt for the second one (will confirm with her if the padding should change/be constant if this is chosen).

karrui: just add a custom key, we only style for `md` right now

seaerchin: yeppers, did that!

frontend/src/theme/components/Tag.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+    light: {
+      container: {
+        bgColor: 'primary.100',
+        textColor: 'primary.400',
+      },
+    },
+  },
+  sizes: {
+    uni: {
```
karrui, line 27: leave as default `md` key is fine, don't need add new key; this keeps the application sizes consistent

karrui (changes requested):

frontend/src/theme/components/Tag.ts
```diff
@@ -0,0 +1,38 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+export type TagVariants = 'solid' | 'light'
+
+export const Tag: ComponentMultiStyleConfig = {
+  parts: [],
+  baseStyle: {
+    label: {
+      textStyle: 'caption-1',
```
karrui, line 9: textStyle cannot be used in this way; you might need to destructure:

```ts
import { textStyles } from './textStyles' // I forgot the real path/import
...
label: {
  ...textStyles['caption-1']
}
```

frontend/src/theme/components/Tag.ts
seaerchin: thanks for input, changed!

frontend/src/theme/components/Tag.ts
seaerchin: wdyt about putting it on the component? not sure what's the appropriate way to do it but because it reads properly from the component level, i felt like it's better to have it there

karrui: don't put it in the component. The example I gave was in the component theme. Just spread out the textStyle props in this file

frontend/src/components/Tag/Tag.tsx
```diff
@@ -12,5 +12,5 @@ export interface TagProps extends Omit<ChakraTagProps, 'colorScheme'> {
 }
 
 export const Tag = (props: TagProps): JSX.Element => {
-  return <ChakraTag {...props} />
+  return <ChakraTag textStyle="caption-1" {...props} />
```
karrui: I think this should go into the theme file

karrui (changes requested):

frontend/src/components/Tag/Tag.tsx
seaerchin: i think you reviweed an old commit @__@  this was changed [here
](https://github.com/opengovsg/FormSG/pull/2384/commits/cb111d26df467e7e19a92e1624a21d081314643b)

frontend/src/theme/components/Tag.ts
```diff
@@ -0,0 +1,40 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+import { textStyles } from '../textStyles'
+
+export type TagVariants = 'solid' | 'light'
+
+export const Tag: ComponentMultiStyleConfig = {
+  parts: [],
```
karrui, line 8: nit: can ignore parts if we are not extending from chakra's defaults

karrui (approved):
lgtm, waiting for @pearlyong

pearlyong (approved):

PR #2388, opened 2021-07-14
feat: part two of checkbox logic - adaptor functions

chowyiyin:
## Problem
This PR implements the extension of the logic feature to logic fields. This second part implements adaptor functions that transform:
* Frontend/Backend checkbox field responses to same shape as checkbox condition values
* Checkbox condition values between the frontend and backend

Part 2/3 of #30

## Solution

### Unit tests
 - `checkbox-response-value-transformer.spec`  
 - `form-logic-checkbox.client.service.spec`  
 -  Add tests to `form-logic.client.service.spec`

### DIFF
--- a/src/shared/util/logic-utils/logic-formatter.ts
+++ b/src/shared/util/logic-utils/logic-formatter.ts
@@ -1,0 +1,30 @@
+import { cloneDeep } from 'lodash'
+
+import {
+  BasicField,
+  FieldSchemaOrResponse,
+  IPopulatedForm,
+} from '../../../types'
+import { LogicFieldSchemaOrResponse } from '../logic'
+
+import { transformCheckboxForLogic } from './checkbox/checkbox-response-value-transformer'
+
+/**
+ * Transforms fields into the correct shape for logic module
+ * @param fields Form fields
+ * @param form
+ * @returns Transformed fields to be input into logic module
+ * @throws Error if any transformation fails
+ */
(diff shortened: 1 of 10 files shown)


mantariksh:
preliminary thought: might be better to encapsulate all this complexity in a class, akin to `SubmissionEmailObj`

src/shared/util/logic-utils/logic-formatter.ts
```diff
@@ -0,0 +1,30 @@
...
+export const formatFieldsForLogic = (
+  fields: FieldSchemaOrResponse[],
+  formFields: IPopulatedForm['form_fields'],
+): LogicFieldSchemaOrResponse[] => {
+  return fields.map((field) => {
+    if (field.fieldType === BasicField.Checkbox) {
+      return transformCheckboxForLogic(field, formFields)
+    } else {
+      return cloneDeep(field) as LogicFieldSchemaOrResponse
+    }
```
mantariksh, line 28: can we do something like this so we don't have to include an unnecessary `throw` in `transformCheckboxForLogic`?
```
if (isClientCheckboxValue(field)) {
    return convertClientCheckboxValue(field, formFields)
  } else if (isServerCheckboxValue(field)) {
    return convertServerCheckboxValue(field, formFields)
  } else {
    return cloneDeep(field) as LogicFieldSchemaOrResponse
  }
```

PR #2389, opened 2021-07-14
feat: checkbox logic

chowyiyin:
## Problem
This PR implements the extension of the logic feature to logic fields. This last part of the solution includes rendering of logic specification for checkboxes in the logic tab and edit logic modal. It also adds the checkbox condition value shape to the backend `Joi` validator and calls all the implemented transformers before and after relevant function calls.

Part 3/3 of #30

## Solution

## Screen Recordings

https://user-images.githubusercontent.com/54089291/127587529-5c49f42d-9bf7-4b90-8d10-712d526b3b3e.mov

https://user-images.githubusercontent.com/54089291/127590410-4140fef4-2f98-4be2-914d-b79b16cf4ea1.mov

### Manual Tests
Create a checkbox field with options "Option 1", "Option 2", "Option 3" and with "Others" toggled on
Create any two other fields
- [ ] Create a logic unit `IF checkbox STATE is any one of the following VALUE Option 1 then show a field` - open public view and check that field is shown when those combinations are selected
- [ ]  Create a logic unit `IF checkbox STATE is any one of the following VALUE Option 1 then disable submission` - open public view and check that submission is blocked when those combinations are selected
- [ ] Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others) or (Option 2 and Option 3) then show a field` - open public view and check that field is shown when those combinations are selected
- [ ]  Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others) or (Option 2 and Option 3) then disable submission` - open public view and check that submission is blocked when those combinations are selected
- [ ] Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others) or (Option 2 and Option 3) **and add another logic unit within this condition** then show a field` - open public view and check that field is shown when those combinations are selected
- [ ]  Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others) or (Option 2 and Option 3) **and add another logic unit within this condition** then disable submission` - open public view and check that submission is blocked when those combinations are selected
- [ ] Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others) or (Option 2 and Option 3) **and add another condition** then show a field` - open public view and check that field is shown when those combinations are selected
- [ ]  Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others) or (Option 2 and Option 3) **and add another condition** then disable submission` - open public view and check that submission is blocked when those combinations are selectedAdd a user-defined "Others" option to checkbox
- [ ] Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others [in-built one]) or (Option 2 and Others [user-defined])  then show a field` - open public view and check that field is shown when those combinations are selected
- [ ] Create a logic unit `IF checkbox STATE is any one of the following VALUE (Option 1 and Others [in-built one]) or (Option 2 and Others [user-defined])  then disable submission` - open public view and check that submission is blocked when those combinations are selected
- [ ] Delete `Option 1` from checkbox options and go to the logic tab - make sure the error message "Your form's logic has errors. Scroll down to fix them before sharing your form." appears at the top. 
  - [ ] Click on the condition with checkbox and make sure the error message "One or more of Checkbox's values above are invalid." appears at the bottom of the left panel. Return to the logic tab and refresh. Check that error message "Some values are no longer valid" appears below the checkbox field. 
  - [ ] Remove `Option 1` from the condition and check that the errors are no longer displayed.


src/public/modules/forms/admin/controllers/admin-form.client.controller.js
```diff
@@ -390,4 +398,72 @@ function AdminFormController(
...
+        $scope.myform.form_fields.find(
+          (field) => field._id === condition.field,
+        ), // must use the myform.form_fields and not condition.fieldInfo because condition.fieldInfo has not been updated
+        condition.state,
+      )
+    })
+  }
+
+  $scope.checkIfHasInvalidValues = function (values, field, state) {
+    if (
```
mantariksh, line 419: can we extract this functionality in the logic module so that we can reuse it in React?

src/shared/util/logic.ts
```diff
@@ -201,7 +202,7 @@ export const getLogicUnitPreventingSubmit = (
   const preventSubmitConditions = getPreventSubmitConditions(form)
   return preventSubmitConditions.find((logicUnit) =>
     isLogicUnitSatisfied(
-      submission,
+      formatFieldsForLogic(submission, form.form_fields),
```
mantariksh, line 205: it seems we are taking on a lot of additional complexity because the adapter layer is in the logic module itself, and as a result the adapter functions have to account for both client and server shapes. might it be better to shift the adapter layers outside instead, and have a separate adapter for both client and server so that they can evolve independently?

src/app/utils/logic-adaptor.ts
```diff
@@ -0,0 +1,95 @@
...
+): IPreventSubmitLogicSchema | undefined => {
+  const transformedSubmission = transformSubmission(submission, form)
+  return Logic.getLogicUnitPreventingSubmit(
+    transformedSubmission,
+    form,
+    visibleFieldIds,
+  )
+}
+
+export * from '../../shared/util/logic'
```
mantariksh, line 95: think about the principle of encapsulation here - the idea of this adapter is to hide the underlying logic module and only expose what's necessary. is it possible to just export what's necessary instead of freely exposing the underlying module?

src/shared/util/logic-utils/logic-values-checker.ts
```diff
@@ -0,0 +1,99 @@
...
+} from './checkbox'
+
+/**
+ * Checks if any of the values specified in logic are invalid.
+ * @param values Condition values.
+ * @param field Field with corresponding condition values.
+ * @param state Condition state
+ * @returns If the values are invalid
+ */
+export const checkIfHasInvalidValues = (
```
mantariksh, line 26: don't really understand why this is in `shared`, since the implementation looks specific to the client-side data representation. would it make more sense to move this (and all the associated types and helpers) into a client-side service?

mantariksh (changes requested):
well done, looks much better and it's starting to come together!

src/app/utils/logic-adaptor.ts
chowyiyin: But I am only re-exposing things that the logic module is currently exposing right? The adaptor should have the same API as the underlying logic module because the adaptor should act as the bridge between the frontend/backend and the logic module. The frontend/backend shouldn't have to access the underlying module at all.

src/app/utils/logic-adaptor.ts
```diff
@@ -0,0 +1,95 @@
...
+  field: ICheckboxResponse,
+  formFields: IPopulatedForm['form_fields'],
+): Logic.LogicFieldResponse => {
+  const completeField = formFields.find(
+    (formField) => String(formField._id) === String(field._id),
+  )
+  if (!isCheckboxField(completeField)) {
+    // eslint-disable-next-line typesafe/no-throw-sync-func
+    throw new Error(`${completeField} is not a Checkbox field`)
+  }
```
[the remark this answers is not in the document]
chowyiyin: Understood but using `neverthrow` changes the signatures of these functions which shouldn't be the case for adaptors. If this error is ever thrown, it means that theres inconsistency in our data because the response and field have the same `_id` but have different `fieldTypes` so I thought it would be good to fail early. What do you think? Do you think it would be better to just log the error and return a dummy transformed response? (i.e. empty `options` and false `others`)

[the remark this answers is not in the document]
mantariksh: >so I thought it would be good to fail early

fair enough! let's stick to this then

> using neverthrow changes the signatures of these functions which shouldn't be the case for adaptors

actually isn't the whole point of adapters to wrap underlying functionality in more suitable signatures? so we could add `neverthrow` to the backend adapter? (this is the advantage of separating the adapters btw, looser coupling)

src/app/utils/logic-adaptor.ts
mantariksh: >The adaptor should have the same API as the underlying logic module

not sure what you mean by this - isn't the whole point of an adapter to bridge incompatible interfaces by reshaping data?

anyway what's the behaviour if you export multiple functions with the same name from a single module (in this case, `getVisibleFieldIds` and `getLogicUnitPreventingSubmit`?

src/app/utils/logic-adaptor.ts
```diff
@@ -1,32 +1,126 @@
-import { ok, Result } from 'neverthrow'
+import { cloneDeep, omit } from 'lodash'
+import { err, fromThrowable, ok, Result } from 'neverthrow'
 
+import { hasProp } from '../../shared/util/has-prop'
 import {
   FieldIdSet,
   getLogicUnitPreventingSubmit as logicGetLogicUnitPreventingSubmit,
   getVisibleFieldIds as logicGetVisibleFieldIds,
```
mantariksh, line 8: hmm this is a very non-intuitive way of namespacing the logic module. appending the word `logic` in front doesn't make it clear that this comes from the logic module. can we do an `import *` instead?

src/public/modules/forms/config/forms.client.routes.js
```diff
@@ -139,7 +139,7 @@ angular.module('forms').config([
             template: `<edit-logic-component 
                 myform="myform"
                 update-form="updateForm(update)"
-                is-logic-error="isLogicError">
+                is-logic-error="isLogicError"
```
mantariksh, line 142: this looks like a mistake!

mantariksh:
I'm closing this given that checkbox logic is a P4 and that React migration is on the horizon. The backend implementation can be reused with React in future.

Unfortunately this was put on the back burner because of Zendemic then Hackathon, and now React migration has progressed to a point where it doesn't make sense to maintain the AngularJS frontend for this feature.

@chowyiyin we'll revisit this after React :)

PR #2399, opened 2021-07-15
feat(shared-types): relocate (admin-related) Submission types to new shared folder

karrui:
## Problem

This PR moved admin-related Submission types to the new shared folder and uses them between the client and the server.

Public-facing submission types (to type public submissions) are not typed yet, as they rely on `Field` types and can be done in a separate PR for ease of review.

Related to #2066 

## Solution

**Breaking Changes** 

**Improvements**:

- move `src/shared/submission.ts` to new rooted `/shared` folder
- use relocated shared types in client services
- alias and use new shared types in backend types


karrui:
EDIT: nvm... i should not have rebased. Forgot that this PR also relies on form, field types.

shared/types/submission.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export type SubmissionResponseDto = {
+  message: string
+  submissionId: string
+}
+
+export type SubmissionErrorDto = ErrorDto & { spcpSubmissionFailure?: true }
+
+export type SubmissionCountQueryDto =
```
mantariksh, line 76: why did you choose to make this type exclude `formId`? doesn't that just make it harder to type the object which includes the `formId`?

shared/types/submission.ts
```diff
@@ -0,0 +1,89 @@
...
+      startDate: DateString
+      endDate: DateString
+    }
+  | undefined
+
+export type FormSubmissionMetadataQueryDto = RequireAtLeastOne<
+  {
+    page: number
+    submissionId: string
+  },
```
mantariksh, line 87: does this still function as intended when the keys are required? in the example in the [docs](https://github.com/sindresorhus/type-fest/blob/main/source/require-at-least-one.d.ts), the RequireAtLeastOne properties are optional

src/public/services/__tests__/AdminSubmissionsService.test.ts
```diff
@@ -178,7 +180,11 @@ describe('AdminSubmissionsService', () => {
 
     it('should decrypt successfully when there is data', async () => {
       // Arrange
-      mockFormSgSdk.crypto.decryptFile.mockReturnValueOnce('great decryption')
+      // Do not really want to mock an Uint8Array since it does not matter for
```
mantariksh, line 183:
```suggestion
      // Dowan mock an Uint8Array since it does not matter for
```

shared/types/submission.ts
karrui: this is just the query dto, which does not include the formId right? the formId is encoded in the API endpoint instead

shared/types/submission.ts
karrui: Yea, they set all other keys to be optional when wrapped with `RequireAtLeastOne`. Just tested locally too (by creating a random constant with the type), will only error if the object is empty

src/public/services/__tests__/AdminSubmissionsService.test.ts
karrui: NO

mantariksh (approved):

PR #2400, opened 2021-07-15
feat(shared-types): move billing related types to shared folder

karrui:
> Note that this PR builds on the typings exposed by #2385

## Problem

This PR extracts out billing related types to the root shared folder

Related to #2066

## Solution

**Breaking Changes** 

  - Only type changes

**Features**:

- move billing related types to root shared folder
- use relocated types in client and server

### DIFF
--- a/src/types/login.ts
+++ b/src/types/login.ts
@@ -1,27 +1,22 @@
 import { Document, Model } from 'mongoose'
 
+import { FormBillingStatistic, LoginBase } from '../../shared/types/billing'
+
 import { IAgencySchema } from './agency'
-import { AuthType, IFormSchema, IPopulatedForm } from './form'
+import { IFormSchema, IPopulatedForm } from './form'
 import { IUserSchema } from './user'
 
-export interface ILogin {
+export interface ILogin extends LoginBase {
   admin: IUserSchema['_id']
   form: IFormSchema['_id']
   agency: IAgencySchema['_id']
-  authType: AuthType
-  esrvcId: string
+}
+
+export interface ILoginSchema extends ILogin, Document {
   created?: Date
 }
 
-export interface ILoginSchema extends ILogin, Document {}
-
-export type LoginStatistic = {
-  adminEmail: IUserSchema['email']
-  formName: IFormSchema['title']
-  total: number
-  formId: IFormSchema['_id']
-  authType: ILoginSchema['authType']
-}
+export type LoginStatistic = FormBillingStatistic
 
 export interface ILoginModel extends Model<ILoginSchema> {
   aggregateLoginStats: (
--- a/src/types/login.ts
+++ b/src/types/login.ts
@@ -6,7 +6,7 @@ import { IAgencySchema } from './agency'
 import { IFormSchema, IPopulatedForm } from './form'
 import { IUserSchema } from './user'
 
-export interface ILogin extends LoginBase {
+interface ILogin extends LoginBase {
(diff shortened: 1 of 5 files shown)


mantariksh (approved):

src/types/login.ts
```diff
@@ -1,28 +1,23 @@
...
 
+import { FormBillingStatistic, LoginBase } from '../../shared/types/billing'
+
 import { IAgencySchema } from './agency'
-import { AuthType, IFormSchema, IPopulatedForm } from './form'
+import { IFormSchema, IPopulatedForm } from './form'
 import { IUserSchema } from './user'
 
-export interface ILogin {
+export interface ILogin extends LoginBase {
```
timotheeg, line 9: Does `ILogin` still need to be exported? Looking at the rest of the codebase, it looks like the expectation is to only use `ILoginSchema`? (I only see `ILogin` being used in test, not in app code)

src/types/login.ts
```diff
@@ -1,28 +1,23 @@
...
-export interface ILoginSchema extends ILogin, Document {}
-
-export type LoginStatistic = {
-  adminEmail: IUserSchema['email']
-  formName: IFormSchema['title']
-  total: number
-  formId: IFormSchema['_id']
-  authType: ILoginSchema['authType']
+export interface ILoginSchema extends ILogin, Document {
+  created?: Date
```
timotheeg, line 16: By moving `created` to `ILoginSchema`, is it still optional?

timotheeg (commented):
I find the new organisation a little strange now (i.e. login depending on billing to extend some of the types). Why would `LoginBase` be in billing for example? 🤔 

I'd think login is a "more fundamental" blocks other modules depend on.

Similar comment for `LoginStatistic` being a copy of `FormBillingStatistic`

I wonder if this indicates there should be a `types/login.ts` in the shared folder instead?

karrui:
> I find the new organisation a little strange now (i.e. login depending on billing to extend some of the types). Why would `LoginBase` be in billing for example? 🤔
> 
> I'd think login is a "more fundamental" blocks other modules depend on.
> 
> Similar comment for `LoginStatistic` being a copy of `FormBillingStatistic`
> 
> I wonder if this indicates there should be a `types/login.ts` in the shared folder instead?

`Login` is kind of a misnomer (but immortalized as a collection in the database 😢 ) since it's actually tracking Corppass/Singpass logins for billing purposes, and since the types are to be shared I thought to make it clearer. Should I change it back to `login`?

src/types/login.ts
karrui: We can probably stop exporting that; seems like the newer mongoose versions automatically type with a lean version of `xxxSchema` types when creating a new document. It used to be the case that we needed a separate typing for creating documents, as `xxxSchema` types contain extra props that is not needed to create a new document.

src/types/login.ts
karrui: Yes, since this typing is used for document creation, and we don't want to enforce a `created` prop to be assigned for document creation.

We should actually create a `LoginDocument` typing for the object that is retrieved from the database (with all the defaults populated, and have the optional types converted to required), but there are currently no cases where that is needed and thus not done yet.

timotheeg (approved):
Sorry I didn't mean to be blocking this PR 😱  

I think with the comment in `billing.ts`, it's probably good enough as is. And we can reorg again later if needed.

Let's go!

PR #2444, opened 2021-07-27
feat(design-system): add Tooltip component, styles and stories

### LINKED ISSUE #2023: [React] Add Tooltip component
**Time estimation: 2d**

Research on the various libraries is needed for a11y, etc.

chowyiyin:
## Problem
This PR implements the Tooltip component.
To test the displaying of the tooltip on tap on mobile devices, please use Chrome's developer tools to trigger a touch event.

Closes #2023

## Screenshots
In storybook

### DIFF
--- a/frontend/src/components/Tooltip/Tooltip.stories.tsx
+++ b/frontend/src/components/Tooltip/Tooltip.stories.tsx
@@ -15,21 +15,21 @@ const Template: Story<TooltipProps> = (args) => {
   return (
     // padding to create space for tooltips to appear
     <VStack p="350px">
-      <Tooltip {...args} label="Display tooltip content top" placement="top" />
       <Tooltip
         {...args}
-        label="Display tooltip content bottom"
-        placement="bottom"
+        label="Display tooltip content right"
+        placement="right"
       />
       <Tooltip
         {...args}
         label="Display tooltip content left"
         placement="left"
       />
+      <Tooltip {...args} label="Display tooltip content top" placement="top" />
       <Tooltip
         {...args}
-        label="Display tooltip content right"
-        placement="right"
+        label="Display tooltip content bottom"
+        placement="bottom"
       />
     </VStack>
   )
--- a/frontend/src/components/Tooltip/Tooltip.stories.tsx
+++ b/frontend/src/components/Tooltip/Tooltip.stories.tsx
@@ -48,6 +48,13 @@ const SingleToolTip: Story<TooltipProps> = (args) => {
   )
 }
 
+export const OpenTooltip = SingleToolTip.bind({})
+OpenTooltip.args = {
+  label: 'This tooltip is open',
+  isOpen: true,
+  children: <Icon name="Question" />,
+}
+
 export const MobileExample = SingleToolTip.bind({})
 MobileExample.args = {
   label:
--- a/frontend/src/components/Tooltip/Tooltip.stories.tsx
+++ b/frontend/src/components/Tooltip/Tooltip.stories.tsx
@@ -1,0 +1,62 @@
+import { HStack, Icon, TooltipProps, VStack } from '@chakra-ui/react'
+import { Meta, Story } from '@storybook/react'
+
+import { viewports } from '~utils/storybook'
+
+import { Tooltip } from './Tooltip'
+
+export default {
+  title: 'Components/Tooltip',
+  component: Tooltip,
+  decorators: [],
+} as Meta
+
+const Template: Story<TooltipProps> = (args) => {
+  return (
+    // padding to create space for tooltips to appear
+    <VStack p="350px">
+      <Tooltip {...args} label="Display tooltip content top" placement="top" />
+      <Tooltip
+        {...args}
+        label="Display tooltip content bottom"
+        placement="bottom"
+      />
+      <Tooltip
+        {...args}
+        label="Display tooltip content left"
+        placement="left"
+      />
+      <Tooltip
+        {...args}
+        label="Display tooltip content right"
+        placement="right"
+      />
+    </VStack>
+  )
+}
+export const TooltipOnHover = Template.bind({})
+TooltipOnHover.args = {
+  children: <Icon name="Question" />,
+}
+
+const SingleToolTip: Story<TooltipProps> = (args) => {
+  return (
+    <HStack>
+      <div>OTP Verification Long Message</div>
+      <Tooltip {...args} />
+    </HStack>
+  )
+}
+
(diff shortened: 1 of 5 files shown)


frontend/src/theme/components/Tooltip.ts
```diff
@@ -0,0 +1,26 @@
...
+    tooltip: {
+      '--tooltip-bg': 'var(--chakra-colors-secondary-700)',
+      px: '0.75rem',
+      py: '0.5rem',
+      color: 'white',
+      textStyle: 'body-2',
+      borderRadius: '0.25rem',
+      textAlign: 'left',
+      margin: '0.25rem',
+      maxWidth: '19.5rem',
```
karrui, line 16: Is there a need for maxWidth?

frontend/src/theme/components/Tooltip.ts
```diff
@@ -0,0 +1,26 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+export const TOOLTIP_THEME_KEY = 'CustomTooltip'
+export const CustomTooltip: ComponentMultiStyleConfig = {
+  parts: ['tooltip', 'wrapper'],
+  baseStyle: {
+    tooltip: {
+      '--tooltip-bg': 'var(--chakra-colors-secondary-700)',
```
karrui, line 8: I see that chakra's own tooltip has a `bg` prop in addition to this. Is there a reason `bg` is not included here?

frontend/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,42 @@
...
+  Box,
+  Tooltip as ChakraTooltip,
+  TooltipProps,
+  useMultiStyleConfig,
+} from '@chakra-ui/react'
+
+import { TOOLTIP_THEME_KEY } from '~/theme/components/Tooltip'
+
+export const Tooltip = ({ children, ...props }: TooltipProps): JSX.Element => {
+  // component is controlled in mobile mode
```
karrui, line 12: Do we need special handling in mobile? I tried Chakra's own tooltip on mobile and it seems to work fine

frontend/src/theme/components/Tooltip.ts
```diff
@@ -0,0 +1,26 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+export const TOOLTIP_THEME_KEY = 'CustomTooltip'
+export const CustomTooltip: ComponentMultiStyleConfig = {
+  parts: ['tooltip', 'wrapper'],
```
karrui, line 5: I think we can just override the original tooltip styling if we don't actually need the special mobile handling code

frontend/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const SingleToolTip: Story<TooltipProps> = (args) => {
+  return (
+    <HStack>
+      <div>OTP Verification Long Message</div>
+      <Tooltip {...args} />
+    </HStack>
+  )
+}
+
```
karrui, line 50: Should we also have a story where the tooltip is already open (with `isOpen` prop)? For automated visual regression testing.

karrui (changes requested):
can we just use the default tooltip without our own special mobile handling logic? I think there is already mobile handling by default

frontend/src/components/Tooltip/Tooltip.tsx
chowyiyin: oh noi tried tapping with chrome's dev tool device emulator and it didn't work :( didn't think of trying on my phone. Will remove!

frontend/src/theme/components/Tooltip.ts
chowyiyin: The default tooltip also has maxWidth, so I asked Pearly about this and she said we could use 312px. I think this is to prevent the tooltip from becoming too big.

frontend/src/theme/components/Tooltip.ts
chowyiyin: I think `bg` uses the `--tooltip-bg` value so I think overriding `--tooltip-bg` is sufficient?

frontend/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -48,6 +48,13 @@ const SingleToolTip: Story<TooltipProps> = (args) => {
   )
 }
 
+export const OpenTooltip = SingleToolTip.bind({})
```
karrui, line 51: nit: should this story use the `Template` tooltip story? So we can show all 4 directions in a single story

karrui (approved):
one last story suggestion

pearlyong:
- i tried on mobile but tooltip doesn't display?
- can help to check the font, it looks like a heavier weight than it should be? Can use body-2 for this
- can help check the bg colour for the tooltip as well? i could be wrong but the desktop one looks darker than mobile? should be secondary-700

sorry I can't seem to inspect these myself^

karrui:
> * Help icon should also be the solid one, not outline

Haha the icon is a placeholder to display the tooltip, this is not part of the component! :P

mantariksh:
@karrui requesting another review due to extensive changes here too!

frontend/src/theme/components/Tooltip.ts
```diff
@@ -0,0 +1,22 @@
+import { ComponentStyleConfig } from '@chakra-ui/theme'
+
+import { textStyles } from '../textStyles'
+
+export const Tooltip: ComponentStyleConfig = {
+  baseStyle: {
+    // overriding --tooltip-big since Chakra UI does it this way -
```
karrui, line 7: typo: big -> bg

karrui (approved):
lgtm with nits

mantariksh:
apologies, had to force push because chromatic was not picking up the changeset correctly

karrui:
May want to force push again so the stories are updated (again)

The open tooltip story snapshot gets cut off, may need extra bottom margin

karrui (commented):
still can't seem to focus on the tooltip, might have missed out that commit

mantariksh:
@karrui you can't focus by tabbing (as with all other text content) but you can use ctrl-shift-option and arrow keys to get voiceover to read the tooltips

mantariksh:
>The open tooltip story snapshot gets cut off, may need extra bottom margin

unfortunately bottom margin doesn't seem to help, chromatic still cuts it off

PR #2454, opened 2021-07-28
feat(design-system): add Pagination component and stories

### LINKED ISSUE #2016: [React] Add Pagination component
**Time estimation: 2d**

Research on various libraries needed, or write own pagination logic.

karrui:
## Problem

This PR adds the Pagination component, along with its stories. Uses a `usePagination` hook to generate the correct buttons.

Closes #2016 

## Solution

**Features**:
- add Pagination component and stories

**Improvements**
- chore: move root path alias to bottom
  - in a bid to let other paths resolve first before trying to use the root import, not sure if it works

## Before & After Screenshots
In storybook

## Dependencies
- "lodash": "^4.17.21"

### DIFF
--- a/frontend/src/components/Pagination/Pagination.tsx
+++ b/frontend/src/components/Pagination/Pagination.tsx
@@ -1,9 +1,15 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
-import { useMemo } from 'react'
-import { Box, Button, ButtonProps, Flex, HStack, Text } from '@chakra-ui/react'
-import range from 'lodash/range'
+import { BiChevronLeft, BiChevronRight } from 'react-icons/bi'
+import {
+  Button,
+  ButtonProps,
+  Flex,
+  Icon,
+  Text,
+  useMultiStyleConfig,
+} from '@chakra-ui/react'
 
-import { usePaginationRange } from '~/hooks/usePaginationRange'
+import { PAGINATION_THEME_KEY } from '~theme/components/Pagination'
+import { usePaginationRange } from '~hooks/usePaginationRange'
 
 // Separate constant to denote a separator in the pagination component.
 const SEPARATOR = '\u2026'
@@ -37,30 +43,22 @@ export interface PaginationProps {
 }
 
 interface PageButtonProps {
-  activePage: PaginationProps['currentPage']
-  page: number
+  selectedPage: PaginationProps['currentPage']
+  page: number | typeof SEPARATOR
   onClick: PaginationProps['onPageChange']
 }
 
-const PageButton = ({ activePage, page, onClick }: PageButtonProps) => {
-  const isActive = page === activePage
+const PageButton = ({ selectedPage, page, onClick }: PageButtonProps) => {
+  const isSelected = page === selectedPage
+
+  const styles = useMultiStyleConfig(PAGINATION_THEME_KEY, { isSelected })
+
+  if (page === SEPARATOR) {
+    return <Text sx={styles.separator}>{page}</Text>
+  }
+
   return (
-    <Button
-      p="0.25rem"
-      minH="2rem"
-      minW="2rem"
-      border="none"
-      textStyle="body-2"
-      bg={isActive ? 'secondary.500' : 'transparent'}
-      _active={{
-        bg: isActive ? 'secondary.700' : 'secondary.200',
-      }}
-      _hover={{
-        bg: isActive ? 'secondary.600' : 'secondary.100',
-      }}
-      color={isActive ? 'white' : 'secondary.500'}
-      onClick={() => onClick(page)}
-    >
+    <Button sx={styles.button} onClick={() => onClick(page)}>
       {page}
     </Button>
   )
@@ -68,45 +66,12 @@ const PageButton = ({ activePage, page, onClick }: PageButtonProps) => {
 
 interface StepButtonProps extends ButtonProps {
   onClick: () => void
-  isDisabled?: boolean
   children: React.ReactNode
 }
 
-const StepButton = ({
-  onClick,
-  isDisabled,
-  children,
-  ...props
-}: StepButtonProps) => {
+const StepButton = ({ onClick, children, ...props }: StepButtonProps) => {
   return (
-    <Button
-      bg="transparent"
-      _hover={{
-        bg: 'secondary.100',
-        color: 'secondary.600',
-      }}
-      _active={{
-        bg: 'secondary.200',
-      }}
-      _disabled={{
-        bg: 'transparent',
-        cursor: 'not-allowed',
-        color: 'secondary.300',
-        _hover: {
-          bg: 'transparent',
-          color: 'secondary.300',
-        },
-      }}
-      isDisabled={isDisabled}
-      p="0.25rem"
-      minH="2rem"
-      minW="2rem"
-      border="none"
-      textStyle="body-2"
-      color="secondary.500"
-      onClick={onClick}
-      {...props}
-    >
+    <Button onClick={onClick} {...props}>
       {children}
     </Button>
   )
@@ -129,32 +94,33 @@ export const Pagination = ({
 
   const totalPageCount = Math.ceil(totalCount / pageSize)
 
+  const styles = useMultiStyleConfig(PAGINATION_THEME_KEY, {})
+
   return (
-    <Flex>
+    <Flex __css={styles.container}>
       <StepButton
-        mr="1rem"
+        sx={styles.stepperback}
         isDisabled={currentPage === 1}
         onClick={() => onPageChange(currentPage - 1)}
       >
+        <Icon fontSize="1.5rem" as={BiChevronLeft} />
         Back
       </StepButton>
-      {paginationRange.map((p, i) => {
-        if (p === SEPARATOR) return <Text key={i}>{p}</Text>
-        return (
-          <PageButton
-            key={i}
-            page={p}
-            activePage={currentPage}
-            onClick={onPageChange}
-          />
-        )
-      })}
+      {paginationRange.map((p, i) => (
+        <PageButton
+          key={i}
+          page={p}
+          selectedPage={currentPage}
+          onClick={onPageChange}
+        />
+      ))}
       <StepButton
-        ml="1rem"
+        sx={styles.steppernext}
         isDisabled={currentPage === totalPageCount}
         onClick={() => onPageChange(currentPage + 1)}
       >
         Next
+        <Icon fontSize="1.5rem" as={BiChevronRight} />
       </StepButton>
     </Flex>
   )
--- a/frontend/src/components/Pagination/Pagination.tsx
+++ b/frontend/src/components/Pagination/Pagination.tsx
@@ -1,0 +1,161 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import { useMemo } from 'react'
+import { Box, Button, ButtonProps, Flex, HStack, Text } from '@chakra-ui/react'
+import range from 'lodash/range'
+
+import { usePaginationRange } from '~/hooks/usePaginationRange'
+
+// Separate constant to denote a separator in the pagination component.
+const SEPARATOR = '\u2026'
+
+export interface PaginationProps {
+  /**
+   * Number of pages to display to left and right of current page.
+   * Defaults to `1`.
+   */
+  siblingCount?: number
+
+  /**
+   * Total number of elements to paginate.
+   */
+  totalCount: number
+
+  /**
+   * Size of each page. Determines the number of rendered page counts.
+   */
+  pageSize: number
+
+  /**
+   * Callback function invoked with the updated page when the page is changed.
+   */
+  onPageChange: (page: number) => void
+
+  /**
+   * Represents the current active page. Note that a `1-based index` is used.
+   */
+  currentPage: number
+}
+
+interface PageButtonProps {
+  activePage: PaginationProps['currentPage']
+  page: number
+  onClick: PaginationProps['onPageChange']
+}
+
+const PageButton = ({ activePage, page, onClick }: PageButtonProps) => {
+  const isActive = page === activePage
+  return (
+    <Button
+      p="0.25rem"
+      minH="2rem"
+      minW="2rem"
+      border="none"
+      textStyle="body-2"
+      bg={isActive ? 'secondary.500' : 'transparent'}
+      _active={{
+        bg: isActive ? 'secondary.700' : 'secondary.200',
+      }}
+      _hover={{
+        bg: isActive ? 'secondary.600' : 'secondary.100',
+      }}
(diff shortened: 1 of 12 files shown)


frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+   */
+  currentPage: number
+
+  /**
+   * A constant to denote a separator in the pagination component.
+   */
+  separator: T
+}
+
+export const usePaginationRange = <T extends unknown = string>({
```
seaerchin, line 37: would it be a good idea to add [this?](https://material-ui.com/components/pagination/#pagination-ranges) (the boundary range)

frontend/src/components/Pagination/PaginationDesktop.tsx
```diff
@@ -0,0 +1,118 @@
...
+    onPageChange(currentPage - 1)
+  }, [currentPage, isDisablePrevPage, onPageChange])
+
+  const handlePageNext = useCallback(() => {
+    if (isDisableNextPage) return
+    onPageChange(currentPage + 1)
+  }, [currentPage, isDisableNextPage, onPageChange])
+
+  return (
+    <Box __css={styles.container}>
```
seaerchin, line 90: could i ask why we aren't using the compound component pattern here? it would allow users to style the buttons/boundary separators, esp w/ exposing style props!

seaerchin (commented):
some minor nits, along with questions wrt design choice on why not compound components because it feels like we're boxing the user into our style

frontend/src/components/Pagination/PaginationDesktop.tsx
karrui: As discussed on Slack, this component is not a layout component but its own component, don't think there is a real need to expose child components

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+    const leftSiblingIndex = Math.max(currentPage - siblingCount, 1)
+    const rightSiblingIndex = Math.min(
+      currentPage + siblingCount,
+      totalPageCount,
+    )
+
+    // Do not show separator just when there is just one page number to be
+    // inserted between the extremes of sibling and the page limits
+    // i.e 1 and totalPageCount.
+    const shouldShowLeftSeparator = leftSiblingIndex - 1 > 2
```
mantariksh, line 70: I think `leftSiblingIndex > 3` would be clearer here, so it's obvious that we only show the left separator if the left sibling is at least 4

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+    const shouldShowRightSeparator = rightSiblingIndex < totalPageCount - 2
+
+    const firstPageIndex = 1
+    const lastPageIndex = totalPageCount
+
+    /**
+     * Case 2: No left separator to show, but rights separator to be shown.
```
mantariksh, line 77:
```suggestion
     * Case 2: No left separator to show, but [rights](https://www.un.org/en/about-us/universal-declaration-of-human-rights) separator to be shown.
```

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    // Do not show separator just when there is just one page number to be
+    // inserted between the extremes of sibling and the page limits
+    // i.e 1 and totalPageCount.
+    const shouldShowLeftSeparator = leftSiblingIndex - 1 > 2
+    const shouldShowRightSeparator = rightSiblingIndex < totalPageCount - 2
+
+    const firstPageIndex = 1
+    const lastPageIndex = totalPageCount
```
mantariksh, line 74: hmm do we really need to have this alias for a `const` variable?

frontend/src/hooks/usePaginationRange.ts
mantariksh: actually, rather than constantly using 3 as a magic number, it might be good to declare it as a constant or even a prop. something like `MIN_ITEMS_WITHOUT_SEP` or `minItemsWithoutSep` if it's a prop, since it refers to the minimum number of items on a side which doesn't have a separator.

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Case 2: No left separator to show, but rights separator to be shown.
+     */
+    if (!shouldShowLeftSeparator && shouldShowRightSeparator) {
+      const leftItemCount = 3 + 2 * siblingCount
```
mantariksh, line 80:
```suggestion
      const leftItemCount = MIN_ITEMS_WITHOUT_SEP + 2 * siblingCount
```

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+      const leftRange = range(1, leftItemCount + 1)
+
+      return [...leftRange, separator, totalPageCount]
+    }
+
+    /**
+     * Case 3: No right separator to show, but left separator to be shown.
+     */
+    if (shouldShowLeftSeparator && !shouldShowRightSeparator) {
+      const rightItemCount = 3 + 2 * siblingCount
```
mantariksh, line 90:
```suggestion
      const rightItemCount = MIN_ITEMS_WITHOUT_SEP + 2 * siblingCount
```

frontend/src/components/Pagination/PaginationDesktop.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const styles = useMultiStyleConfig(PAGINATION_THEME_KEY, {})
+
+  const totalPageCount = Math.ceil(totalCount / pageSize)
+  const isDisableNextPage = isDisabled || currentPage >= totalPageCount
+  const isDisablePrevPage = isDisabled || currentPage <= 1
+
+  const handlePageBack = useCallback(() => {
+    if (isDisablePrevPage) return
+    onPageChange(currentPage - 1)
```
mantariksh, line 81: the component seems to rely on the parent keeping the state of `currentPage` consistent, i.e. there might be a case where the currently focused page is different from the `currentPage` prop. am I understanding that correctly, and might this be a source of bugs in future? an alternative might be to rename `currentPage` to `initialPage`, and the component to keep track of `currentPage` in its internal state (perhaps exposed to the parent via a ref?)

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+  pageSize,
+  siblingCount,
+  currentPage,
+  separator,
+}: UsePaginationRangeProps<T>): (T | number)[] => {
+  const paginationRange = useMemo(() => {
+    const totalPageCount = Math.ceil(totalCount / pageSize)
+
+    // Pages count is determined as siblingCount + firstPage + lastPage + currentPage + 2*separator
+    const totalPageNumbers = siblingCount + 5
```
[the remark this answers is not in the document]
timotheeg: Agreed on both counts.

I think the siblingCount formula is just accidentally missing `2*`
```suggestion
    const totalPageNumbers = 2 * siblingCount + 5
```

frontend/src/hooks/usePaginationRange.ts
timotheeg: I'm thinking 3 is not really the value the code cares about, rather the logic the code tries to convey is that a separator represents a gap of 2 or above.

I imagine that's why the formula is showing the `> 2` above, but then it is unfortunate the formula for the right separator below  is not following the same formula format.

I reckon it would be clearer if both formulas were consistent, like this:
```javascript
const shouldShowLeftSeparator = leftSiblingIndex - 1 > 2
const shouldShowRightSeparator = totalPageCount - rightSiblingIndex > 2
```

And then we can replace the `2` with a named variable, perhaps named `minGapSize`, or such like.

frontend/src/hooks/usePaginationRange.ts
timotheeg: I'm thinking these alias are actually nice, but they should be used consistently in the rest of the code in this function. Even the formula above would make more using them

```javascript
const firstPageIndex = 1
const lastPageIndex = totalPageCount

const shouldShowLeftSeparator = leftSiblingIndex - firstPageIndex > 2
const shouldShowRightSeparator = lastPageIndex - rightSiblingIndex > 2
```

frontend/src/hooks/usePaginationRange.ts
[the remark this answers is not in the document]
timotheeg: This makes me think the storybook should have additional reference cases to show 1 before and 1 after separators are introduced.

With current code, if we had a storybook case for "exactly 7 pages" (with current page = 1). then I think the display would be incorrect like this:

image

So to make sure we match the design, I think we need cases like these:
* less than 7 (shows 6 pages, no separators
* exactly 7 (shows 7 pages no separators)
* exactly 8 (shows 5 pages, separator, and page 8)

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * Custom hook to generate page ranges for any pagination component.
+ * Referenced from
+ * https://www.freecodecamp.org/news/build-a-custom-pagination-component-in-react/
+ */
+
+import { useMemo } from 'react'
+import range from 'lodash/range'
```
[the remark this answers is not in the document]
timotheeg: I'm not too worried about this 🤔. Use-cases would be to have sibling count as 1 in most cases, 2, in some cases, and maybe 3 or 4 in rare cases (I expect?), so we're talking about invoking a range on less than 10, which is not worth optimizing for.

frontend/src/hooks/usePaginationRange.ts
timotheeg: LOL 😂

frontend/src/hooks/usePaginationRange.ts
timotheeg: /nit: just suggestion: would it be easier to read if the formula was like this?
```javascript
leftItemCount = totalButtonSlots - 2
```
(2 is basically separator and last page)

Same for rightItemCount below

frontend/src/hooks/usePaginationRange.ts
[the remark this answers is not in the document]
karrui: Thanks for the catch, will add the stories and update the logic as described.

frontend/src/hooks/usePaginationRange.ts
karrui: correct; we want to only show if the differences between the ends are larger than a gap of 2; will create a named variable.

frontend/src/hooks/usePaginationRange.ts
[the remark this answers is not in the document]
karrui: @mantariksh regarding the variable naming; it's not really the total number of page slots too, since it's actually counting the total number of pages to display; i.e 70 items, 10 items per page = 7 pages. 

Maybe I'll continue to call it totalPageCount, but change totalPageNumbers to numPageButtonSlots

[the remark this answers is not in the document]
mantariksh: https://www.youtube.com/watch?v=nQY4dIxY1H4

frontend/src/hooks/usePaginationRange.ts
karrui: I've opted to remove those aliases, reads a little easier imho

frontend/src/hooks/usePaginationRange.ts
karrui: Makes sense. Updated.

frontend/src/components/Pagination/PaginationDesktop.tsx
karrui: This components follows how controlled components are usually used though; only rendering what the parent passes in. If onPageChange does not change the currentPage the currentPage will remain as is -- the parent component controls that.

(Similar to how `value` props and `onChange` props must be passed in for the value to get updated)

I think I will stick with the current implementation if there are no big objections.

frontend/src/hooks/usePaginationRange.ts
karrui: Updated to use 
```ts
// 2 indicates first page number and the separator.
const rightItemCount = numPageDisplaySlots - 2
```

instead as suggested by Tim

frontend/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,57 @@
...
+    xl: false,
+  })
+
+  return isShowMobileVariant ? (
+    <Pagination.Mobile {...props} />
+  ) : (
+    <Pagination.Desktop {...props} />
+  )
+}
+
```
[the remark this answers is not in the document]
karrui: I think this is fine, I don't think there is even a risk. We want the mobile variant to show up in mobile view and the desktop variant to show up in desktop view

frontend/src/hooks/usePaginationRange.ts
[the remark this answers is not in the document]
timotheeg: https://martinfowler.com/bliki/TwoHardThings.html

timotheeg (approved):
lgtm 👍

frontend/src/hooks/usePaginationRange.ts
```diff
@@ -0,0 +1,118 @@
...
+     */
+    if (shouldShowLeftSeparator && !shouldShowRightSeparator) {
+      // 2 indicates first page number and the separator.
+      const rightItemCount = numPageDisplaySlots - 2
+      const rightRange = range(
+        totalPageCount - rightItemCount + 1,
+        totalPageCount + 1,
+      )
+      return [1, separator, ...rightRange]
+    }
```
timotheeg, line 102: /over-engineering (not asking for a change, just putting down the thought 😅)

If we wanted to additionally remove the hardcoded `2` and the comment that indicates what `2` represents, we could turn the code like below:
```suggestion
    /**
     * Case 2: No left separator to show, but right separator to be shown.
     */
    if (!shouldShowLeftSeparator && shouldShowRightSeparator) {
      const pageItems: (T | number)[] = [separator, totalPageCount]
      const leftItemCount = numPageDisplaySlots - pageItems.length
      
      pageItems.unshift(...range(1, leftItemCount + 1))

      return pageItems
    }

    /**
     * Case 3: No right separator to show, but left separator to be shown.
     */
    if (shouldShowLeftSeparator && !shouldShowRightSeparator) {
      const pageItems: (T | number)[] = [1, separator]
      const rightItemCount = numPageDisplaySlots - pageItems.length
      
      pageItems.push(...range(
        totalPageCount - rightItemCount + 1,
        totalPageCount + 1)
      )

      return pageItems
    }
```

PR #2463, opened 2021-07-29
fix: Fix express session types based on latest definitions

timotheeg:
## Context
The package [`@types/express-session`](https://www.npmjs.com/package/@types/express-session) introduced a breaking change between `1.17.0` and `1.17.1` when it cleaned up its type definition. See discussion [here](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/d1259ee31e6b17c31a5a86e27a0d12f3ec7c5a19#r44080605).

Although our `package.json` declaration is `^1.17.0`, we have been protected from typescript compilation errors thanks to the version being locked in `package-lock.json`.

Doing a fresh install from scratch based on package.json alone however yields typescript compilation errors on type isses like these:
```
src/app/modules/auth/auth.controller.ts:163:21 - error TS2339: Property 'user' does not exist on type 'Session & Partial<SessionData>'.
163         req.session.user = { _id }
                        ~~~~
src/app/modules/auth/auth.middlewares.ts:18:23 - error TS2345: Argument of type 'Session & Partial<SessionData>' is not assignable to parameter of type 'Session | undefined'.
  Type 'Session & Partial<SessionData>' has no properties in common with type 'Session'.
18   if (isUserInSession(req.session)) {
                         ~~~~~~~~~~~
src/app/modules/auth/auth.middlewares.ts:36:26 - error TS2352: Conversion of type 'Session & Partial<SessionData>' to type 'AuthedSession' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'user' is missing in type 'Session & Partial<SessionData>' but required in type 'AuthedSession'.
36   const sessionUserId = (req.session as Express.AuthedSession).user._id
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  src/types/vendor/express.d.ts:16:7
    16       user: {
             ~~~~
    'user' is declared here.
```

## Approach

1. Upgrade explicitly to package `@types/express-session@1.17.4`
2. Fix augmentation of the types for express session to be compatible with [latest declarations](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-session/index.d.ts)

Note in particular that:
* The global Express namespace no longer include `Session` and `SessionData`
* Augmentation of Session must be done via extension on `SessionData`, not `Session`

Still TODO: 
- [X] Verify tests pass
- [x] Test locally

## Risks
Impact: **High**: making code changes affecting authenticated sessions could break user access
Likelihood: **Low**: Functionality of code itself is not changed, only type declarations

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4917,13 +4917,12 @@
       }
     },
     "@types/express-session": {
-      "version": "1.17.0",
-      "resolved": "https://registry.npmjs.org/@types/express-session/-/express-session-1.17.0.tgz",
-      "integrity": "sha512-OQEHeBFE1UhChVIBhRh9qElHUvTp4BzKKHxMDkGHT7WuYk5eL93hPG7D8YAIkoBSbhNEY0RjreF15zn+U0eLjA==",
+      "version": "1.17.4",
+      "resolved": "https://registry.npmjs.org/@types/express-session/-/express-session-1.17.4.tgz",
+      "integrity": "sha512-7cNlSI8+oOBUHTfPXMwDxF/Lchx5aJ3ho7+p9jJZYVg9dVDJFh3qdMXmJtRsysnvS+C6x46k9DRYmrmCkE+MVg==",
       "dev": true,
       "requires": {
-        "@types/express": "*",
-        "@types/node": "*"
+        "@types/express": "*"
       }
     },
     "@types/glob": {
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     "@types/express": "^4.17.13",
     "@types/express-rate-limit": "^5.1.3",
     "@types/express-request-id": "^1.4.2",
-    "@types/express-session": "^1.17.0",
+    "@types/express-session": "^1.17.4",
     "@types/has-ansi": "^3.0.0",
     "@types/helmet": "4.0.0",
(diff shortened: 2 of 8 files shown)


timotheeg:
Hmmmmmmm, Looks like I could make a much smaller changeset by only updating `express.d.ts` to this:

```typescript
import { SessionData } from 'express-session'
import { IUserSchema } from 'src/types'

declare module 'express-session' {
  export interface SessionData {
    user?: {
      _id: IUserSchema['_id']
    }
  }
}

declare global {
  namespace Express {
    export interface Request {
      id?: string
    }

    export interface Session extends SessionData {}

    export interface AuthedSession extends Session {
      user: {
        _id: IUserSchema['_id']
      }
    }
  }
}
```

But that would be exposing the Session type into Express, which is different from the Native type, so am not sure if that is desirable 🤔

mantariksh (approved):
fwiw personally I'm okay if we don't test this on staging since changes are pretty small and confined to types only!

timotheeg:
For tracking purposes, I have open this issue, which will get closed when this is merged and deployed.
https://github.com/opengovsg/FormSG/issues/2464

timotheeg:
Tested build in docker, seems to be working OK 👍

PR #2467, opened 2021-07-29
feat(model): remove esrvcId validation in form model

### LINKED ISSUE #2466: Update form eservice id validation
new eservice ids do not conform to validated format

karrui:
## Problem

esrvcId validation has changed to also include underscores. Instead of updating our validation every time this happens (unannounced), remove the validation entirely.

This should still not allow invalid esrvcIds as there is a client side check when activating the form with an invalid esrvcId

Closes #2466

## Solution

**Breaking Changes** 

    - Schema changes, if new forms have esrvcId that were invalid in the old schema, forms may fail to save.

**Features**:

- feat(model): remove esrvcId validation in form model

### DIFF
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -383,10 +383,6 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       esrvcId: {
         type: String,
         required: false,
-        validate: [
-          /^([a-zA-Z0-9-]){1,25}$/i,
-          'e-service ID must be alphanumeric, dashes are allowed',
-        ],
       },
 
       webhook: {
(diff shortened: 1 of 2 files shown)


tshuli (approved):
lgtm

mantariksh (changes requested):
we have to update this pattern in the frontend too:
```
ng-pattern="/^([a-zA-Z0-9\-]){1,25}$/i"
```

mantariksh:
a deeper fix might be to introduce a `shared` validation function (like we have for NRIC, UEN etc) which can be reused by the frontend and backend

karrui:
> a deeper fix might be to introduce a `shared` validation function (like we have for NRIC, UEN etc) which can be reused by the frontend and backend

Good call on the frontend, but should we even ahve a validation function? Could we just use our current validation method (hacky as it may be) to ascertain whether the esrvcId is valid on form activation?

PR #2468, opened 2021-07-29
fix(SGID): disallow SGID authentication in storage mode

### LINKED ISSUE #2390: SGID forms can be duplicated into storage mode
Model-level validation should disallow this as SGID is not supported in storage mode.

mantariksh:
## Problem

SGID is currently not enabled in storage mode, and the settings tab UI prevents users from enabling SGID on a storage mode form. However, SGID email-mode forms can still be duplicated into storage mode while preserving the SGID auth type, resulting in invalid forms which cannot be submitted.

Closes #2390 

## Solution

Add model-level validation which sets the `authType` to `NIL` when attempting to create an SGID form in storage mode.

## Manual tests
- [ ] Duplicate an SGID email-mode form into storage mode. The authentication type of the new duplicated form should be NIL.

### DIFF
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -12,6 +12,7 @@ import getFormModel, {
   getEncryptedFormModel,
 } from 'src/app/models/form.server.model'
 import {
+  AuthType,
   BasicField,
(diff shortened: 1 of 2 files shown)


karrui (approved):
lgtm

PR #2475, opened 2021-07-30
chore: Regenerate package-lock with fresh install

timotheeg:
## Context
[package-lock](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json) is a fantastic system to lock versions of dependencies and sub-dependencies we use, and ensures consistency between different installs.

Unlike package.json, where we specify semver rules of dependencies, package-lock pins down explicit versions for everything, at the time they were looked up. `npm install ci` respects these explicit versions, and tools like Snyk check these for vulnerability. 

A package-lock file may therefore fall behind on patch or minor upgrades that are considered acceptable by project owners, as specified by the dependency rules in package.json.

Dependencies and their sub-dependencies are updated when a new dependencies is added or modified, but this is done within a limited scope of the action (untouched package do not get their versions inspected and upgraded).

Because of this, package-lock in FormSG has been left behind multiple updates of acceptable patch and minor upgrades.

This PR is a general upgrade of all dependencies and sub-dependencies by regenerating a fresh package-lock file.

NB: every upgrades (even patch!) introduces risk or behaviour change, breakages and regressions. A balance needs to be striked between relying on the consistency provided by package-lock, and deliberate efforts to upgrades packages on regular basis. The PR is opened because it looks like our package-lock has fallen behind by quite a bit.

## Approach
Generate a fresh package-lock by running the following commands:
```bash
rm -rf package-lock.json node_modules
npm install
```

## Risk
Impact: High: Many dependencies are updated, which could have subtle behaviour and/or performance changes
Likelihood: Medium: We are only following the default dependency acceptance rules, which (by right) should prevent breaking changes. The PR includes no code change and no change to package.jon.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,18 +5,18 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.10.1",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.1.tgz",
-      "integrity": "sha512-IGhtTmpjGbYzcEDOw7DcQtbQSXcG9ftmAXtWTu9V936vDye4xjjekktFAtgZsWpzTj/X01jocB46mTywm/4SZw==",
+      "version": "7.14.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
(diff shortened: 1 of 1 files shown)


timotheeg:
Apologies, I should have opened this as draft until I had fixed the mockingoose tests 😔

timotheeg:
Closing this for now. The mockinggoose failure is annoying, and I should focus on actual Snyke vulnearibilities before doing troubleshooting on general deps upgrades.

Will re-open the PR when the P0s are all done.

PR #2478, opened 2021-07-30
feat(design-system): Add Searchbar component, stories (and relevant hook)

### LINKED ISSUE #2019: [React] Add Searchbar component
**Time estimation: 2d**

~~Possibly blocked by input component theme styles in #2002~~ Unblocked

Research on various libraries needed for a11y, etc.

karrui:
## Problem

This PR implements the Searchbar component.

Closes #2019 

## Solution

**Features**:

- Add searchbar component and stories
- add `useSearchbar` hook that can be used with searchbar component

## Screenshots
In Storybook


frontend/src/hooks/useSearchbar.ts
```diff
@@ -0,0 +1,51 @@
+/**
```
mantariksh, line 1: what's the value of placing this hook in a centralised `hooks` folder when it's only ever going to be used in the SearchBar component? or is it actually reusable?

karrui: Technically you require it to use the searchbar, but the searchbar component itself does not use it. I'll move it into the component folder anyways and export it!

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -0,0 +1,89 @@
...
+  InputProps,
+  useMergeRefs,
+  useMultiStyleConfig,
+} from '@chakra-ui/react'
+
+import { SEARCHBAR_THEME_KEY } from '~/theme/components/Searchbar'
+
+import IconButton from '../IconButton'
+
+export interface SearchbarProps extends InputProps {
```
tshuli, line 19: Is it helpful to explicitly pass in the variant as one of the `SearchbarProps` and set a default value?

frontend/src/theme/components/Searchbar.ts
```diff
@@ -0,0 +1,34 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+import { Input } from './Input'
+
+export const SEARCHBAR_THEME_KEY = 'Searchbar'
+
+export const Searchbar: ComponentMultiStyleConfig = {
+  parts: ['icon', 'field'],
+  baseStyle: {},
+  variants: {
```
tshuli, line 10: Would it be helpful to declare the variant types?

frontend/src/components/Searchbar/useSearchbar.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * If `true`, the searchbar will be expanded on initial render.
+   */
+  isInitiallyExpanded?: boolean
+  /**
+   * If `true`, the searchbar will be focused whenever the searchbar is expanded.
+   */
+  isFocusOnExpand?: boolean
+}): UseSearchbarReturn => {
+  const [isExpanded, onExpansionChange] = useState(isInitiallyExpanded)
```
mantariksh, line 33: isn't the convention to use `is...` and `setIs....`?

frontend/src/components/Searchbar/useSearchbar.ts
```diff
@@ -0,0 +1,51 @@
...
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  useLayoutEffect(() => {
```
mantariksh, line 36: do we really need `useLayoutEffect` here? the effect seems to depend on the component props and state, so it doesn't seem to need the DOM to be mutated before running. tbh unnecessarily using `useLayoutEffect` instead of `useEffect` smells like an antipattern since it blocks visual updates

frontend/src/components/Searchbar/useSearchbar.ts
karrui: good catch! probably accidentally renamed

PR #2498, opened 2021-08-02
chore: remove express-device

timotheeg:
## Context
We have a vulnerability report on express-device [here](https://app.snyk.io/org/formsg/project/7d2653ec-310a-4c60-a94a-eca85308e725#issue-SNYK-JS-USERAGENT-174737)

[express-device](https://www.npmjs.com/package/express-device) has not been upgraded for some time, but diving deeper in code, it looks like we do not actually use the capabilities provided by express-device ? 🤔 

The following command yields nothing:
```bash
git grep -E 'is_(desktop|phone|tablet|mobile|tv|bot|car)|device_(type|name)'
```

## Approach
Remove express-device and all the references.


mantariksh:
we should probably check that the user agent string still appears as expected in the Cloudwatch logs

timotheeg:
True. I would think that user-agent logging is a built-in functionality of express itself. `express-device's job is to provide device type flags, and I see no code where we're using these flags or where we add them to log.

timotheeg:
Tested on staging with User-Agent still showing in log 👍 

If someone has extra context on why we were using `express-device` in the first place, and a reason we shouldn't remove it, please speak now :)

@mantariksh @karrui @tshuli

mantariksh (approved):
>If someone has extra context on why we were using express-device in the first place, and a reason we shouldn't remove it, please speak now :)

¯\_(ツ)_/¯ was before our time lol

timotheeg:
Right 😄 ! Well, I'm just going to merge then. @liangyuanruo if you remember something, let us know :)

PR #2499, opened 2021-08-02
fix(login): show Joi validation error correctly

### LINKED ISSUE #2455: Unhelpful error message on OTP validation failure
## Problem
FormSG uses a OTP to log in. This OTP can only be comprised of numbers and should be 6 digit long. Although the frontend prevents characters from being inputted, this does **not** prevent copy-pasting errors. Hence, a user can paste an erroneous otp into the `input` and request to sign in. 

This results in the `celebrate` validation failing (as it should be number and 6 digits only) and the error is shown in the info box, which is extremely jarring and unhelpful to the user. 

## Steps to reproduce
1. Attempt to sign into form
2. at the otp page, paste any string containing alphabets (eg: `password`)
3. click sign in

## Suggested Fix 
Set default message on celebrate

## Screenshots 
Screenshot 2021-07-28 at 2 58 35 PM

mantariksh:
## Problem

Closes #2455 

## Solution

- Correct casing in Joi OTP error message
- In the frontend, search for a Joi validation error and prioritise rendering it over the API response. This is because the presence of a Joi error message implies the absence of an API response.

### Before
Screenshot 2021-07-28 at 2 58 35 PM

### After
Screenshot 2021-08-02 at 11 41 32 AM

### DIFF
--- a/src/app/modules/auth/__tests__/auth.routes.spec.ts
+++ b/src/app/modules/auth/__tests__/auth.routes.spec.ts
@@ -357,7 +357,7 @@ describe('auth.routes', () => {
       expect(response.status).toEqual(400)
       expect(response.body).toEqual(
         buildCelebrateError({
-          body: { key: 'otp', message: 'Please enter a valid otp' },
+          body: { key: 'otp', message: 'Please enter a valid OTP' },
         }),
       )
     })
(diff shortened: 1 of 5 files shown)


timotheeg (approved):
LGTM 👍

PR #2500, opened 2021-08-02
feat(sms-limiting): prerelease

seaerchin:
## Problem
This PR is part 1 of `sms-limiting`; this PR does a few things. 

1. Adds in methods for sending warning/disabling mail
2. Adds in a new endpoint for count retrieval (`/:formId/verified-sms/count/free`) 
3. Adds in a db script for updating the `smscounts` table to have the `isOnboardedAcc` key. 
4. Adds in service/db method to disabling sms verifications for admin (this is **unused** here)

## Tests
This PR should not have any observable interface by which we can perform manual testing, except for `/count/free`. 

**preconditions**
1. run this query to obtain your sms counts: 
```
db.getCollection('smscounts').countDocuments({
    "formAdmin.userId": ObjectId("your userId here"),
    smsType: 'VERIFICATION', 
    isOnboardedAccount: false, 
})
```

2. the db should already be populated with the `isOnboardedAccount` key

- [ ] Do not login and query the route at `http://localhost:5000/api/v3/admin/forms/609a520ead304e002c5a362f/verified-sms/count/free`. This should fail with `401 unauthorized`
- [ ] Subsequently, login and query the route again. This should succeed and return the sms count of the current logged in user and the sms quota (currently 10k). This sms count should be equivalent to the one obtained from the query.

### DIFF
--- a/src/app/loaders/express/locals.ts
+++ b/src/app/loaders/express/locals.ts
@@ -4,7 +4,6 @@ import config from '../../config/config'
 import { captchaConfig } from '../../config/features/captcha.config'
 import { googleAnalyticsConfig } from '../../config/features/google-analytics.config'
 import { sentryConfig } from '../../config/features/sentry.config'
-import { smsConfig } from '../../config/features/sms.config'
 import { spcpMyInfoConfig } from '../../config/features/spcp-myinfo.config'
 
 // Construct js with environment variables needed by frontend
@@ -23,7 +22,6 @@ const frontendVars = {
   GATrackingID: googleAnalyticsConfig.GATrackingID,
   spcpCookieDomain: spcpMyInfoConfig.spcpCookieDomain, // Cookie domain used for removing spcp cookies
   oldSpcpCookieDomain: spcpMyInfoConfig.oldSpcpCookieDomain, // Old cookie domain used for backward compatibility. TODO (#2329): Delete env var
-  smsVerificationLimit: smsConfig.smsVerificationLimit,
 }
 const environment = ejs.render(
   `
@@ -49,8 +47,6 @@ const environment = ejs.render(
     var spcpCookieDomain = "<%= spcpCookieDomain%>"
     // Old SPCP Cookie
     var oldSpcpCookieDomain = "<%= oldSpcpCookieDomain%>"
-    // Sms verification limit
-    var smsVerificationLimit = "<%= smsVerificationLimit%>"
   `,
   frontendVars,
 )
--- a/src/app/services/sms/sms.types.ts
+++ b/src/app/services/sms/sms.types.ts
@@ -63,7 +63,7 @@ export interface IVerificationSmsCount extends ISmsCount {
   isOnboardedAccount: boolean
 }
 
-export type IVerificationSmsCountSchema = ISmsCountSchema & {
+export interface IVerificationSmsCountSchema extends ISmsCountSchema {
   isOnboardedAccount: boolean
 }
 
--- a/src/types/database.ts
+++ b/src/types/database.ts
@@ -3,11 +3,3 @@
 export interface PublicView<T> {
   getPublicView(): T
 }
-
-// The returned object from updateMany
-// Refer here: https://mongoosejs.com/docs/api/model.html#model_Model.updateMany
-export interface UpdateManyMeta {
-  n: number
-  nModified: number
-  ok: number
-}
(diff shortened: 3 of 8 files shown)


karrui (commented):
sorry, but can i suggest changing the shape of /count/free instead of passing an env var to the client to know the available quota?

two pros to this approach:
1. single source of truth; the quota endpoint denotes the current quota, prevents desync between old clients and new server
2. reduces the stuff sent by passing env vars to the client from the server, which I think is an anti pattern

karrui:
The rest of my comments disappeared... hold on

src/app/loaders/express/locals.ts
```diff
@@ -47,6 +49,8 @@ const environment = ejs.render(
     var spcpCookieDomain = "<%= spcpCookieDomain%>"
     // Old SPCP Cookie
     var oldSpcpCookieDomain = "<%= oldSpcpCookieDomain%>"
+    // Sms verification limit
+    var smsVerificationLimit = "<%= smsVerificationLimit%>"
```
karrui, line 53: Looking at this again, I don't think we should send this env var to the client and ahve the client use it. We should use the `/count/free` endpoint to transmit this information. Reasoning is expanded on in the review...

With that said, we should update the endpoint to return an object instead of a number
```jsonc
// GET /count/free
{
  remainingFree: // number of free smses remaining?
  quota: // sms verification limit env var, but retrieved from the server side and transmitted to client instead
}
```

seaerchin:
> sorry, but can i suggest changing the shape of /count/free instead of passing an env var to the client to know the available quota?
> 
> two pros to this approach:
> 
> 1. single source of truth; the quota endpoint denotes the current quota, prevents desync between old clients and new server
> 2. reduces the stuff sent by passing env vars to the client from the server, which I think is an anti pattern

i agree with point 2! wrt point 1, could i ask for further clarification on how there's a desync as the env var is refreshed on page fetch from the BE.

will change b/c i agree that mass sending env-vars is an anti-pattern due to locality issues (we lose track of where/what the env var is for)

src/app/loaders/express/locals.ts
seaerchin: done [here](https://github.com/opengovsg/FormSG/pull/2500/commits/f41056aaa14081db5463d8878a0161626d7d37a6)

src/types/database.ts
```diff
@@ -3,3 +3,11 @@
 export interface PublicView<T> {
   getPublicView(): T
 }
+
+// The returned object from updateMany
+// Refer here: https://mongoosejs.com/docs/api/model.html#model_Model.updateMany
+export interface UpdateManyMeta {
```
karrui, line 9: should use mongoose's own type instead: `mongoose.UpdateWriteOpResult`

src/app/services/sms/sms.types.ts
```diff
@@ -60,9 +60,12 @@ export interface IVerificationSmsCount extends ISmsCount {
     email: string
     userId: IUserSchema['_id']
   }
+  isOnboardedAccount: boolean
 }
 
-export type IVerificationSmsCountSchema = ISmsCountSchema
+export type IVerificationSmsCountSchema = ISmsCountSchema & {
```
karrui, line 66: this seems to be a leftover bug. it should extend from `IVerificationSmsCount` instead.

src/app/services/sms/sms_count.server.model.ts
```diff
@@ -125,6 +138,16 @@ const compileSmsCountModel = (db: Mongoose) => {
...
+  SmsCountSchema.statics.retrieveFreeSmsCounts = async function (
+    userId: string,
+  ) {
+    return this.countDocuments({
+      'formAdmin.userId': userId,
+      smsType: SmsType.Verification,
+      isOnboardedAccount: false,
+    }).exec()
+  }
+
```
karrui, line 150: Should we also add an `index` block here (even though it does nothing on Mongo Atlas) so people who fork this repository at least gets the index added.

```
SmsCountSchema.index({
      'formAdmin.userId': 1,
      smsType: 1,
      isOnboardedAccount: 1,
})
```

karrui (approved):
lgtm pending some last changes

src/types/database.ts
seaerchin: fixed!

src/app/services/sms/sms_count.server.model.ts
seaerchin: mongoose [docs](https://mongoosejs.com/docs/guide.html#indexes) says that this incurs a perf hit and recommends disablign it using `autoIndex: false` in production. 

would this negatively impact our app as well or does atlas sidestep this?

seaerchin: discussed offline; not too sure how atlas behaves w/ this so erring on safe side and not doing it.

PR #2509, opened 2021-08-03
ref: migrate public-facing submission types to shared folder

karrui:
## Problem

This PR moves public-facing submission types to the shared root folder and extends them in the client and server.

See #2066

## Solution

**Breaking Changes** 

    - Details ...
  - Only moving types

**Features**:

- ref: add and use public storage mode submission types

**Improvements**:

- feat: remove unused ClientSubmission interface
- clean up now unused response types


shared/types/response.ts
```diff
@@ -1,7 +1,10 @@
+import { Opaque } from 'type-fest'
 import { z } from 'zod'
 import { BasicField, MyInfoAttribute } from './field'
 
 const ResponseBase = z.object({
+  myInfo: z.never().optional(),
+  signature: z.never().optional(),
```
mantariksh, line 7: just making sure - does this work when merging with `MyInfoResponseBase` and `VerifiableResponseBase`?

shared/types/submission.ts
```diff
@@ -87,3 +94,31 @@ export type FormSubmissionMetadataQueryDto = RequireAtLeastOne<
...
 >
+
+type SubmissionContentBase = {
+  responses: FieldResponse[]
+}
+
+/**
+ * Shape of email form submissions
+ */
+export type EmailModeSubmissionContentDto = SubmissionContentBase
```
mantariksh, line 105: tiny nit but this is really a mouthful - any issue with the old `EmailSubmissionDto` naming?

src/types/api/email_submission.ts
```diff
@@ -1,5 +1,28 @@
...
-export type EmailSubmissionDto = {
-  responses: ({ question: string } & FieldResponse)[]
+import {
+  AttachmentResponse,
+  FieldResponse,
+} from '../../../shared/types/response'
+import { EmailModeSubmissionContentDto } from '../../../shared/types/submission'
+
+/**
+ * AttachmentResponses with additional server injected metadata on email forms..
```
mantariksh, line 10: two full stops means twice as final issit

mantariksh (approved):

shared/types/response.ts
karrui: Yea checked, the base type only accepts undefined, but extended types accept the correct items

shared/types/submission.ts
karrui: EmailSubmissionDto gives the wrong impression that that is the entire shape of the DTO, when in actuality the content has to be transformed into a `FormData` object prior to calling the endpoint. Just thought to make that distinction. Your thoughts?

karrui: Will fix in another PR if you think we should still change :P

PR #2522, opened 2021-08-03
feat(design-system): add dropdown menu component and stories

### LINKED ISSUE #1999: [React] Add Menu component
Button's dropdown menu

**Time estimation: 2d**

For ChakraUI's implementation: see https://chakra-ui.com/docs/overlay/menu

tshuli:
**Features**:

- add dropdown menu component and stories. closes #1999 

## Before & After Screenshots
In storybook


frontend/src/components/Menu/DropdownMenu.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+  variant: DropdownMenuVariant
+  size: DropdownMenuSize
+  children: string
+}
+
+type DropdownMenuGroupTemplateProps = {
+  variant: DropdownMenuVariant
+}
+
+const DropdownMenuTemplate: Story<DropdownMenuTemplateProps> = ({
```
seaerchin, line 41: should we also have an interactive playground story with integration w/ react hook forms to demonstrate how the component is used? 

also, should we have a default `active` state (where the dropdown is open) to catch visual regressions easily?

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+  children,
+  isActive,
+  size,
+}: DropdownMenuButtonProps): JSX.Element => {
+  const styles = useMultiStyleConfig('DropdownMenu', { variant, size })
+  return (
+    <ChakraMenuButton
+      sx={styles.outerBox}
+      as={Box}
+      bgColor={isActive ? 'secondary.300' : 'clear'} //  For rounded outline
```
seaerchin, line 54: i think the second argument of `useMultiStyleConfig` accepts any arbitrary object, so we could add the `isActive` prop there and keep styling fully in themes!

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+    >
+      <Box sx={styles.innerBox}>
```
seaerchin, line 56: i'm not sure if this component is the one causing the issue, but there's no hover styling/cursor: pointer attribute, which comes inbuilt with the chakra version of the menu, as it's no longer a button. 

this also means that the menu itself is no longer tab-targettable, which creates issues for screen readers @_@

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+        {children}
+        {isActive ? <BxsChevronUp /> : <BxsChevronDown />}
```
seaerchin, line 58: should we set `aria-role: presentation` and `aria-hidden: true`? as this whole element is a button, the icon is purely presentational and hiding it from keyboard events/screen readers would avoid the problem of this being accidentally tab-targeted. 

a good article on `aria-role: presentation` is [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role) one!

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,63 @@
...
+      justifyContent: 'space-between',
+      alignItems: 'center',
+      padding: '8px 16px',
+      height: '44px',
+      background: 'white',
+      boxSizing: 'border-box',
+      borderStyle: 'solid',
+      borderWidth: '1px',
+      borderRadius: '4px',
+      borderColor: 'white',
```
seaerchin, line 26: not sure about this but i think inlining it into `outerBox` might be clearer!

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,63 @@
...
+        },
+        _active: {
+          borderWidth: '2px',
+          outline: '4px solid',
+          outlineColor: 'secondary.300',
+        },
+        _focus: {
+          outline: '4px solid',
+          outlineColor: 'secondary.300',
+        },
```
seaerchin, line 55: should we swap the order of these two styles? pseudo styles should activate in order of user interaction and i think `_focus` has precedence over `_active`. (learnt this from [this article](https://bitsofco.de/when-do-the-hover-focus-and-active-pseudo-classes-apply/))

tshuli: thanks! edited

frontend/src/theme/components/DropdownMenu.ts
tshuli: I kept the current implementation because I am using the outerBox padding background to provide the curved effect

frontend/src/components/Menu/DropdownMenu.tsx
tshuli: thanks! added

frontend/src/components/Menu/DropdownMenu.tsx
tshuli: yes good point, changed it back to button

frontend/src/components/Menu/DropdownMenu.tsx
tshuli: thanks! edited accordingly

frontend/src/components/Menu/DropdownMenu.stories.tsx
tshuli: added, see if that works!

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
...
+    option: {
+      _hover: {
+        background: 'primary.100',
+      },
+      _focus: {
+        background: 'white',
+        borderWidth: '2px',
+        borderStyle: 'solid',
+        borderColor: 'primary.500',
+      },
```
seaerchin, line 49: i'm not entirely sure this is correct for 2 reasons.

1. setting `borderWidth` here will cause the element to resize. this leads to it shrinking.
2. next, setting background color doesn't seem correct because css pseudo classes are allowed to overlap (focus/active being the most obvious ones). 

i think the focus styling is actually only the border, and a trick to work around the resizing might be to use `boxShadow`. tbh, i'm not entirely sure that's the best way also. because the item doesn't have an actual border, another way could be to have a 2px transparent border and only show colour on focus (will avoid resizing)

frontend/src/components/Menu/DropdownMenu.stories.tsx
```diff
@@ -0,0 +1,204 @@
...
+              </DropdownMenuItem>
+              <DropdownMenuItem
+                onClick={() => {
+                  toast({
+                    status: 'success',
+                    description: 'Successfully downloaded',
+                    isClosable: true,
+                    duration: 2000,
+                  })
+                }}
```
seaerchin, line 162: need to wrap in chakra theme provider

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/react'
+
+export type DropdownMenuVariant = 'outline' | 'clear'
+export type DropdownMenuSize = 'normal' | 'stretch'
+
+export const DropdownMenu: ComponentMultiStyleConfig = {
+  parts: ['outerbox', 'innerbox', 'option'],
+  baseStyle: ({ isActive }) => ({
+    outerBox: {
+      padding: '4px',
```
seaerchin, line 10: i think we aim to use rem because it's based on element size and hence, more responsive (read more [here](https://chiamakaikeanyi.dev/sizing-in-css-px-vs-em-vs-rem/)). for reference, 16px = 1rem.

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
...
+      background: 'white',
+      borderRadius: '8px',
+      border: '0px',
+      width: '177px',
```
seaerchin, line 14: need double check if fixed width or dynamic width

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
...
+export type DropdownMenuSize = 'normal' | 'stretch'
+
+export const DropdownMenu: ComponentMultiStyleConfig = {
+  parts: ['outerbox', 'innerbox', 'option'],
+  baseStyle: ({ isActive }) => ({
+    outerBox: {
+      padding: '4px',
+      background: 'white',
+      borderRadius: '8px',
+      border: '0px',
```
seaerchin, line 13: should this be removed?

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
...
+        borderColor: 'secondary.500',
+        _hover: {
+          borderColor: 'secondary.700',
+        },
+      },
+    },
+  },
+  defaultProps: {
+    variant: 'outline',
+    size: 'normal',
```
seaerchin, line 74: i'm not too sure that this is how chakra works actually. the `size` prop defaults to our theme sizing, which is `sm/md/lg` etc. i think setting it to `normal` implies an invalid key, which defaults the component to `md` (the actual chakra default)

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
...
+      background: 'primary.200',
+    },
+  }),
+  sizes: {
+    stretch: {
+      outerBox: {
+        width: '304px',
+      },
+    },
+  },
```
seaerchin, line 61: i think this works but perhaps we should restyle the `lg` size instead? i think this allows for less cognitive overhead (name of sizes is consistent throughout instead of having custom size names for different components)

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
...
+      },
+      _active: {
+        background: 'white',
+      },
+    },
+    innerBox: {
+      display: 'flex',
+      flexDirection: 'row',
+      justifyContent: 'space-between',
+      alignItems: 'center',
```
seaerchin, line 30: should we use `Flex` here?

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,76 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/react'
+
+export type DropdownMenuVariant = 'outline' | 'clear'
+export type DropdownMenuSize = 'normal' | 'stretch'
+
+export const DropdownMenu: ComponentMultiStyleConfig = {
+  parts: ['outerbox', 'innerbox', 'option'],
```
seaerchin, line 7: the casing here is inconsistent (outerBox vs outerbox) - does the component still function as expected?

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+   */
+  variant: DropdownMenuVariant
+  /**
+   * Whether the menu has been opened
+   */
+  isActive: boolean
+  /**
+   * The size of the dropdown menu - stretched or not.
+   */
+  size: DropdownMenuSize
```
seaerchin, line 33: should we check with pearly on how the component determines the sizing? it could be that on focus, it stretches to the width (some inputs do this) or on large screens, it automatically becomes stretched. 

if that's the case, it would make more sense to determine the sizing internally and not expose it.

seaerchin (commented):
i think there's a slight error on behaviour - both the menu button + options "shrink" when you hover over them. i think this is because there's a change in border size, which leads to the component resizing.

seaerchin:
also, i can't seem to tab target the options + they have border on `_hover`, which seems to differ from the figma

frontend/src/theme/components/DropdownMenu.ts
tshuli: apparently yes still - but I've fixed the casing

frontend/src/theme/components/DropdownMenu.ts
tshuli: thanks! have used `lg` instead

frontend/src/theme/components/DropdownMenu.ts
tshuli: as below - used `md` instead

frontend/src/theme/components/DropdownMenu.ts
tshuli: thanks for highlighting - changed to dynamic width as per discussion on figma

frontend/src/theme/components/DropdownMenu.ts
tshuli: thanks for highlighting - changed to rem

frontend/src/components/Menu/DropdownMenu.stories.tsx
tshuli: thanks - to simplify, i've used alert instead

frontend/src/components/Menu/DropdownMenu.tsx
tshuli: thanks! changed to dynamic sizing as per discussion on figma, and similar to button - two options are available (instead of determining it internally)

tshuli:
for options, they can be selected using arrow keys instead of tab - as per aria [specs](https://www.w3.org/TR/wai-aria-practices/#menu)

also fixed the css for options `_hover`

tshuli:
Note that theming for dropdown menu is different from dropdown field (see form design system)

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,91 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/react'
+
+export type DropdownMenuVariant = 'outline' | 'clear'
+export type DropdownMenuSize = 'md' | 'lg'
+
+export const DropdownMenu: ComponentMultiStyleConfig = {
```
karrui, line 6: is there a reason why we are not just overriding Chakra's base `Menu` theme and instead using our own? 
Overriding only allows users to use Chakra's base Menu item without having to import our own (if they wanted)

That also gives us a baseline of things to change.

For example, in your implementation, the focus state seems a little different from the design. Do you want a quick pair program for this?

tshuli: Thanks! edited, see if the revised implementation works

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const styles = useMultiStyleConfig('Menu', {
+    isActive,
+  })
+  return (
+    <ChakraMenuButton
+      padding={isActive ? '0.4375rem 0.9375rem' : '0.5rem 1rem'}
+      borderWidth={isActive ? '0.125rem' : '0.0625rem'}
+      borderColor={isClear ? 'white' : 'secondary.500'}
+      minWidth={isStretch ? '100%' : 'max-content'}
+      _hover={{ borderColor: isClear ? 'white' : 'secondary.700' }}
```
karrui, line 55: These should all be in the theme, shouldn't it?

karrui: Looking at the design and the [ChakraUI documentation](https://chakra-ui.com/docs/overlay/menu#usage), I don't actually think we need to have custom styles at all for the menu button at all. Looks like the theming could just be:

```tsx
...
// Outline variant
const DropdownMenuButton = (props) => {
  const ChevronIcon = isActive ? BxsChevronUp : BxsChevronDown
  return (
    <ChakraMenuButton 
      as={Button} 
      variant="outline" 
      colorScheme="secondary" 
      rightIcon={<ChevronIcon fontSize="1.5rem" />}
      {...props}
    />
  )
}
```

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * @preconditions Must be a child of DropdownMenu component
+ * after DropdownMenuButton, and returned using a render prop
+ * (see implementation in DropdownMenu.stories).
+ *
+ * Used to wrap MenuItem component
+ */
+export const DropdownMenuList = (props: MenuListProps): JSX.Element => {
```
karrui, line 86: nit: can simplify to 

```tsx
export const DropdownMenuList = ChakraMenuList
```

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+ * Used to wrap DropdownMenuButton and DropdownMenuList components
+ */
+export const DropdownMenu = (props: MenuProps): JSX.Element => {
+  return <ChakraMenu {...props} matchWidth={true} gutter={4}></ChakraMenu>
+}
+
+/**
+ * Item in DropdownMenuList
+ */
+export const DropdownMenuItem = (props: MenuItemProps): JSX.Element => {
```
karrui, line 100: Same here, can simplify

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return <ChakraMenuItem {...props}></ChakraMenuItem>
+}
```
karrui, line 102: Should we use the subcomponent pattern and reexport all the various components as Menu's subcomponents for clarity?

For example, 
```tsx
export const DropdownMenu = ... // your implementation

DropdownMenu.Button = DropdownMenuButton
DropdownMenu.List = DropdownMenuList
...

```

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,68 @@
...
+    'groupTitle',
+    'command',
+    'divider',
+    'icon',
+  ],
+  baseStyle: {
+    icon: {
+      ml: '1.5rem',
+    },
+    button: {
```
karrui, line 19: i don't think we need theming for `baseStyle.button`; we should just be using Button's own themes

karrui (changes requested):

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+} from '@chakra-ui/react'
+
+import { BxsChevronDown } from '~/assets/icons/BxsChevronDown'
+import { BxsChevronUp } from '~/assets/icons/BxsChevronUp'
+
+export interface DropdownMenuButtonProps {
+  /**
+   * If true, the button will have a clear border
+   */
+  isClear?: boolean
```
mantariksh, line 21: wondering if we should use theme variants here instead of props

tshuli:
Thanks both! I've edited it to use the existing Button theming

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,33 @@
...
+        padding: '0.625rem 0.875rem',
+        background: 'white',
+        borderWidth: '0.125rem',
+        borderStyle: 'solid',
+        borderColor: 'primary.500',
+        borderRadius: '0.125rem',
+      },
+      _active: {
+        padding: '0.75rem 1rem',
+        borderWidth: '0rem',
```
karrui, line 24: as above, should we use outline instead of border so the widths don't change

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Used to wrap DropdownMenuButton, DropdownMenuItem and DropdownMenuList components
+ */
+export const DropdownMenu = (props: MenuProps): JSX.Element => {
+  return <ChakraMenu {...props} matchWidth={true} gutter={4}></ChakraMenu>
+}
+
+DropdownMenu.DropdownMenuButton = DropdownMenuButton
+DropdownMenu.DropdownMenuList = DropdownMenuList
+DropdownMenu.DropdownMenuItem = DropdownMenuItem
```
karrui, line 63: should we update the naming to remove redundancies? e.g.

```
DropdownMenu.Button = DropdownMenuButton,
DropdownMenu.List = ...
```
etc

frontend/src/components/Menu/DropdownMenu.stories.tsx
```diff
@@ -0,0 +1,171 @@
...
+  isOpen?: boolean
+  variant: ThemeButtonVariant
+  children: string
+}
+
+type DropdownMenuGroupTemplateProps = {
+  variant: ThemeButtonVariant
+}
+
+const { DropdownMenuButton, DropdownMenuItem, DropdownMenuList } = DropdownMenu
```
karrui, line 34: nit: we should not need to destructure from the main component, should just use `DropdownMenu.Button` to instantiate the component directly, e.g.

```tsx
<DropdownMenu>
  <DropdownMenu.Button />
  <DropdownMenu.List>
    <DropdownMenu.Item/>
    <DropdownMenu.Item/>
  </DropdownMenu.List>  
</DropdownMenu>

**Review by karrui — CHANGES_REQUESTED** [review]

File: frontend/src/theme/components/DropdownMenu.ts [hunk 27]
```diff
@@ -0,0 +1,33 @@
...
+export const Menu: ComponentMultiStyleConfig = {
+  parts: ['list', 'item'],
+  baseStyle: {
+    item: {
+      padding: '0.75rem 1rem',
+      fontWeight: '400',
+      _hover: {
+        padding: '0.75rem 1rem',
+        background: 'primary.100',
+        borderWidth: '0rem',
```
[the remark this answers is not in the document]
tshuli: so the chakra implementation of `_hover` seems slightly buggy, and the `_focus` border is displayed on hover. So `borderWidth: '0rem'` is to override and remove the focus border on hover only

But i've removed the `borderWidth: '0rem'` override for active since (as per discussion) it should retain the focus border if applicable

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,33 @@
...
+      },
+      _focus: {
+        padding: '0.625rem 0.875rem',
+        background: 'white',
+        borderWidth: '0.125rem',
+        borderStyle: 'solid',
+        borderColor: 'primary.500',
+        borderRadius: '0.125rem',
```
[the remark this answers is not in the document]
tshuli: thanks, I removed it

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,33 @@
...
+      padding: '0.75rem 1rem',
+      fontWeight: '400',
+      _hover: {
+        padding: '0.75rem 1rem',
+        background: 'primary.100',
+        borderWidth: '0rem',
+      },
+      _focus: {
+        padding: '0.625rem 0.875rem',
+        background: 'white',
```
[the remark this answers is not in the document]
tshuli: pending clarification

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,33 @@
...
+        borderWidth: '0.125rem',
+        borderStyle: 'solid',
```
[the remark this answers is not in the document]
tshuli: thanks, changed to outline instead

frontend/src/components/Menu/DropdownMenu.stories.tsx
tshuli: thanks! edited

frontend/src/theme/components/DropdownMenu.ts
```diff
@@ -0,0 +1,33 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/react'
+
+export const Menu: ComponentMultiStyleConfig = {
+  parts: ['list', 'item'],
+  baseStyle: {
+    item: {
+      padding: '0.75rem 1rem',
+      fontWeight: '400',
+      _hover: {
+        padding: '0.75rem 1rem',
```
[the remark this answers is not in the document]
tshuli: removed; it was to override the focus padding (see discussion below) but now no longer necessary

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Item in DropdownMenuList
+ */
+const DropdownMenuItem = ChakraMenuItem
+
+/**
+ * Used to wrap DropdownMenuButton, DropdownMenuItem and DropdownMenuList components
+ */
+export const DropdownMenu = (props: MenuProps): JSX.Element => {
+  return <ChakraMenu {...props} matchWidth={true} gutter={4}></ChakraMenu>
```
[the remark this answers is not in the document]
tshuli: changed

frontend/src/components/Menu/DropdownMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+      colorScheme="secondary"
+      textAlign="left"
+      rightIcon={ChevronIcon}
+      iconSpacing="1.5rem"
+      _hover={{ bgColor: 'white' }}
+      _active={{
+        bgColor: 'white',
+        borderWidth: '0.125rem',
+        padding: '0.4375rem 0.9375rem',
+      }}
```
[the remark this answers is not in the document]
tshuli: pending clarification

frontend/src/theme/components/DropdownMenu.ts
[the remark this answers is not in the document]
tshuli: clarified and removed background prop

frontend/src/components/Menu/DropdownMenu.tsx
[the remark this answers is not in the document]
tshuli: pearly confirmed that styling is correct (i.e. slight differences in border and background behavior)

tshuli:
@karrui have confirmed the designs with pearly, for your re-review!

karrui (approved):
some styling still looks off, but i'll leave that to @pearlyong, codewide looks fine

mantariksh (commented):
can we also stick to a single naming convention pls - either `Menu` or `DropdownMenu`? it's confusing that everything is named after the parent folder except this. (noticed because I'm using this for Datepicker haha)

tshuli:
Ok, I've changed everything to `Menu`

frontend/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,67 @@
...
+        bgColor: 'white',
+        borderColor: isVariantOutline ? 'secondary.900' : '',
+        color: 'secondary.900',
+      }}
+      _active={{
+        bgColor: 'white',
+        boxShadow: isVariantOutline ? '0 0 0 0.0625rem black' : '',
+      }}
+      {...props}
+    ></ChakraMenuButton>
```
karrui, line 40: nit: use "/>" closing tag shorthand

e.g.

```tsx
<ChakraMenuButton
   // props
/> // <-- this shorthand, don't have to repeat the closing tag for cleanliness
```

frontend/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Item in MenuList
+ */
+const MenuItem = ChakraMenuItem
+
+/**
+ * Used to wrap MenuButton, MenuItem and MenuList components
+ */
+export const Menu = (props: MenuProps): JSX.Element => {
+  return <ChakraMenu matchWidth={true} gutter={4} {...props}></ChakraMenu>
```
karrui, line 62: nit: use "/>" shorthand

karrui (approved):
lgtm with nits

PR #2525, opened 2021-08-04
feat(Checkbox): add individual checkbox and stories

mantariksh:
Part of #2008

Opening purely for code review first, will request design review when the rest of #2008 is complete.

### DIFF
--- a/frontend/src/components/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/Checkbox/Checkbox.tsx
@@ -1,0 +1,40 @@
+import {
+  Checkbox as ChakraCheckbox,
+  CheckboxProps as ChakraCheckboxProps,
+  Icon,
+  useMultiStyleConfig,
+} from '@chakra-ui/react'
+
+import { BxCheckAnimated } from '~/assets/icons'
+import { CHECKBOX_THEME_KEY } from '~/theme/components/Checkbox'
+import { FieldColorScheme } from '~/theme/foundations/colours'
+
+export interface CheckboxProps extends ChakraCheckboxProps {
+  colorScheme?: FieldColorScheme
+}
+
+export const Checkbox = ({
+  children,
+  colorScheme = 'primary',
+  ...props
+}: CheckboxProps): JSX.Element => {
+  const { icon: iconStyles } = useMultiStyleConfig(CHECKBOX_THEME_KEY, {
+    size: props.size,
+  })
+  return (
+    <ChakraCheckbox
+      icon={
+        <Icon
+          as={BxCheckAnimated}
+          __css={iconStyles}
+          // This prop needs to be passed explicitly for animations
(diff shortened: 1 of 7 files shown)


frontend/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,33 @@
...
+  children,
+  colorScheme = 'primary',
+  ...props
+}: CheckboxProps): JSX.Element => {
+  const { icon: iconStyles } = useMultiStyleConfig(CHECKBOX_THEME_KEY, {
+    size: props.size,
+  })
+  return (
+    <ChakraCheckbox
+      icon={<Icon as={BxCheck} __css={iconStyles} />}
```
karrui, line 26: We lose the animation because we changed the icon. Do we want to be extra and animate our checkbox icon too?

See [Chakra's implementation](https://github.com/chakra-ui/chakra-ui/blob/main/packages/checkbox/src/checkbox-icon.tsx) for reference

frontend/src/theme/components/Field/Checkbox.ts
```diff
@@ -0,0 +1,78 @@
...
+          bg: getColor(theme, `neutral.500`),
+        },
+      },
+    },
+    // Container for the checkbox as well as label
+    container: {
+      w: '100%',
+      px: '0.25rem',
+      py: '0.5rem',
+      _hover: {
```
karrui, line 46: should we disable bg on hover on disabled state?

frontend/src/theme/components/Field/Checkbox.ts
```diff
@@ -0,0 +1,78 @@
...
+        borderColor: getColor(theme, `neutral.500`),
+        bg: 'white',
+        _checked: {
+          borderColor: getColor(theme, `neutral.500`),
+          bg: getColor(theme, `neutral.500`),
+        },
+      },
+    },
+    // Container for the checkbox as well as label
+    container: {
```
karrui, line 42: not in the design, but should we duplicate the _active state to be the same as _focusWithin state? Active state can be triggered by double click or long hold left click(or in devtools). Maybe ask @pearlyong ?

karrui (commented):
Code wise ok, some nits. Some design comments too~
Assume styling for "others" component will come later

frontend/src/theme/components/Field/Checkbox.ts
mantariksh: if you say so hahaha

frontend/src/theme/components/Field/Checkbox.ts
mantariksh: I don't think so because this leads to confusing behaviour where two options can be highlighted at once, e.g. if you click one option such that it's currently focused, then click on another option so it's currently active

PR #2527, opened 2021-08-04
feat: clean up all naming discrepancies between shared and backend types

### LINKED ISSUE #2066: Type DTOs using domain oriented design that client/server types extend from

karrui:
## Problem

Final PR????
Remove all aliases, reexport as same names as they are imported from the root shared folder.

Closes #2066 

## Solution

**Breaking Changes** 

    - Details ...
  - Type refactoring in the backend

**Improvements**:

- Remove all aliased types in the backend. Re-export as is from shared folder

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -39,14 +39,13 @@ import { FormLogicSchema, LogicDto } from './form_logic'
 import { IPopulatedUser, IUserSchema, PublicUser } from './user'
 
 export {
-  FormAuthType as AuthType,
-  FormColorTheme as ColorTheme,
-  FormEndPage as EndPage,
-  FormStartPage as StartPage,
-  FormStatus as Status,
-  FormColorTheme as Colors,
-  FormResponseMode as ResponseMode,
-  FormWebhook as Webhook,
+  FormAuthType,
+  FormColorTheme,
+  FormEndPage,
+  FormStartPage,
+  FormStatus,
+  FormResponseMode,
+  FormWebhook,
   FormSettings,
   StorageFormSettings,
   EmailFormSettings,
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -53,6 +53,7 @@ export {
   PublicFormDto,
   PublicStorageFormDto,
   PublicEmailFormDto,
+  FormPermission,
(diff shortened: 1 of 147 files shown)


src/types/form.ts
```diff
@@ -39,20 +39,20 @@ import { FormLogicSchema, LogicDto } from './form_logic'
 import { IPopulatedUser, IUserSchema, PublicUser } from './user'
 
 export {
```
mantariksh, line 41: hmm I thought we were going to clean up this importing and re-exporting? can't everything just import from the same place? I feel like this complicates our import/export dependencies

karrui: I think we should still have a central backend types that reexports shared typed, cause sometimes backend will have their own overrides of shared types, and its cleaner to have all type imports from backend instead of some from shared and some from the backend type folder?

actually kind of a moot point though... since we have instances of importing from both shared and types T_T.

I'll change it to import directly from shared types then in that case if there is no need to override types in the backend

karrui:
Sorry for letting this languish; will update this PR today

karrui:
@mantariksh sorry for the large PR, didn't seem practical to split them into smaller PRs as so many files are changed. I just updated the types and didn't update any code code so probably can just skim

karrui:
Oops.. forgot about updating for TESTS, will do so now

src/types/form.ts
karrui: DONE

karrui:
Tested on staging; everything seems to work....

PR #2532, opened 2021-08-05
feat(sms-limiting): disabled email

seaerchin:
## Problem
We want to inform admins on the email sent out about forms that would be affected by SMS verification being disabled so that they are able to accurately assess the impact.

## Solution
This solution has 3 parts. 

1. write a db query + wrapper to interface with the database and retrieve the forms of the admin that will be impacted.
2. Refactor disabled mail sending into 2 portions 
    a. Send mail to admin - this includes all forms that will be impacted
    b. Send mail to collaborators - only the forms that they are a part of 
3. Refactor warning mail sending - same as point 2a. 

## Screenshots 
**Disabled mail for admin**
Screenshot 2021-08-05 at 4 59 13 PM

**Disabled mail for collaborator**
Screenshot 2021-08-05 at 5 00 16 PM

**Warning mail for admin**
Screenshot 2021-08-05 at 5 01 02 PM

### DIFF
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -828,6 +828,22 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     ).exec()
   }
 
+  /**
+   * Retrieves all the public forms for a user which has sms verifications enabled
+   * @param userId The userId to retrieve the forms for
+   * @returns All public forms that have sms verifications enabled
+   */
+  FormSchema.statics.retrievePublicFormsWithSmsVerification = async function (
+    userId: IUserSchema['_id'],
+  ) {
+    return this.find({
+      admin: userId,
+      'form_fields.fieldType': BasicField.Mobile,
+      'form_fields.isVerifiable': true,
+      status: Status.Public,
+    }).exec()
+  }
+
   // Hooks
   FormSchema.pre<IFormSchema>('validate', function (next) {
     // Reject save if form document is too large
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -830,6 +830,7 @@ const compileFormModel = (db: Mongoose): IFormModel => {
 
   /**
    * Retrieves all the public forms for a user which has sms verifications enabled
+   * This only retrieves forms that are using FormSG credentials
    * @param userId The userId to retrieve the forms for
    * @returns All public forms that have sms verifications enabled
    */
@@ -841,6 +842,9 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       'form_fields.fieldType': BasicField.Mobile,
       'form_fields.isVerifiable': true,
       status: Status.Public,
+      msgSrvcName: {
+        $exists: false,
+      },
     })
       .read('secondary')
       .exec()
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -841,7 +841,9 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       'form_fields.fieldType': BasicField.Mobile,
       'form_fields.isVerifiable': true,
       status: Status.Public,
-    }).exec()
+    })
+      .read('secondary')
+      .exec()
   }
 
   // Hooks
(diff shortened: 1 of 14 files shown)


src/app/models/form.server.model.ts
```diff
@@ -828,6 +828,22 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+   * @returns All public forms that have sms verifications enabled
+   */
+  FormSchema.statics.retrievePublicFormsWithSmsVerification = async function (
+    userId: IUserSchema['_id'],
+  ) {
+    return this.find({
+      admin: userId,
+      'form_fields.fieldType': BasicField.Mobile,
+      'form_fields.isVerifiable': true,
+      status: Status.Public,
```
karrui, line 843: The problem with this query is that it may be expensive. Has this been benchmarked on staging?

src/app/services/mail/mail.service.ts
```diff
@@ -581,71 +596,154 @@ export class MailService {
...
+  // Sms verifications being disabled for the form.
+  // Note that this method also emails the admin to notify them that the collaborators have been informed.
+  sendDisabledMailForCollab = (
+    form: IPopulatedForm,
+    admin: IPopulatedUser,
+  ): ResultAsync<true, MailGenerationError | MailSendError> => {
+    const htmlData: CollabSmsDisabledData = {
+      form: extractFormLinkView(form, this.#appUrl),
+      smsVerificationLimit:
+        smsConfig.smsVerificationLimit.toLocaleString('en-US'),
```
karrui, line 644: If we are going to pass in "en-US" to the parameter, might as well just use `String(smsConfig.smsVerificationLimit)`?

src/app/services/mail/mail.service.ts
```diff
@@ -581,71 +596,154 @@ export class MailService {
...
+  // Helper method to send an email to a form admin which contains a summary of
+  // which forms would be impacted by sms verifications being removed.
+  sendDisabledMailForAdmin = (
+    forms: IPopulatedForm[],
+    admin: IPopulatedUser,
+  ): ResultAsync<true, MailGenerationError | MailSendError> => {
+    const htmlData: AdminSmsDisabledData = {
+      forms: forms.map((f) => extractFormLinkView(f, this.#appUrl)),
+      smsVerificationLimit:
+        smsConfig.smsVerificationLimit.toLocaleString('en-US'),
```
karrui, line 678: samesies

src/app/services/mail/mail.service.ts
```diff
@@ -581,71 +596,154 @@ export class MailService {
...
+        // Step 3: Send to each individual form
+        return combine(
+          forms.map((f) =>
+            // If there are no collaborators, do not send out the email.
+            // Admin would already have received a summary email from Step 2.
+            f.permissionList.length
+              ? this.sendDisabledMailForCollab(f, form.admin)
+              : okAsync<true, never>(true),
+          ),
+        )
```
karrui, line 629: nit: Should we extract this to its own function?

src/app/models/form.server.model.ts
karrui: If this is benched to be slow, we have a minimum of 2 options:

1. Add index
2. Use current index on the `form` collection `admin_1_lastModified_-1` and programatically filter.

karrui (commented):
will need to check for efficiency of the db query first

src/app/services/mail/mail.service.ts
seaerchin: there's a (small) difference - using locale string formats big numbers for us, so numbers like 10000 turns into 10,000, which isn't present for using `String` (will give 10000 instead). 

I think i'd keep this cos admin will see this number and i want to make it look professional

src/app/services/mail/mail.service.ts
seaerchin: see above explanation

karrui: add this rationale to jsdoc?

seaerchin: sure!

src/app/models/form.server.model.ts
seaerchin: resolved - worst case at present took 0.01s

karrui:
As discussed on Slack, may need updated copy  and logic to only disable forms which are not onboarded

PR #2540, opened 2021-08-06
feat(Radio): add individual radio and stories

mantariksh:
Part of #2008

Opening purely for code review first, will request design review when the rest of #2008 is complete.

### DIFF
--- a/frontend/src/components/Radio/Radio.tsx
+++ b/frontend/src/components/Radio/Radio.tsx
@@ -32,6 +32,8 @@ import {
 } from '@chakra-ui/react'
 import { callAll, split } from '@chakra-ui/utils'
 
+import { RADIO_THEME_KEY } from '~/theme/components/Radio'
+
 type Omitted = 'onChange' | 'defaultChecked' | 'checked'
 type BaseControlProps = Omit<HTMLChakraProps<'div'>, Omitted>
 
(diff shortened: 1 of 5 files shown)


frontend/src/theme/components/Radio.ts
```diff
@@ -0,0 +1,89 @@
...
+      textStyle: 'body-1',
+      ml: '1rem',
+    },
+  }),
+  sizes: {
+    // md is the default and we only have one size, so override it
+    md: {
+      control: { w: '1.5rem', h: '1.5rem' },
+    },
+  },
```
karrui, line 88: should we add a `defaultProps` object here with `colorScheme: "primary"` so the default radio has the appropriate colour scheme that is consistent with our other components?

frontend/src/theme/components/index.ts
```diff
@@ -24,6 +25,7 @@ export const components = {
   PhoneNumberInput,
   Textarea,
   [CHECKBOX_THEME_KEY]: Checkbox,
+  [RADIO_THEME_KEY]: Radio,
```
karrui, line 28: regretting the theme key now, since they are technically only used once haha.
but since this is now exported, should it also be used in the component itself?

frontend/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,152 @@
...
+ * Radio component is used in forms when a user needs to select a single value from
+ * several options.
+ *
+ * @see Docs https://chakra-ui.com/radio
+ */
+export const Radio = forwardRef<RadioProps, 'input'>((props, ref) => {
+  const { onChange: onChangeProp, value: valueProp } = props
+
+  const group = useRadioGroupContext()
+  const styles = useMultiStyleConfig('Radio', { ...group, ...props })
```
karrui, line 68: use theme key here? but i think we can just do away with the theme key and use these strings instead :P

mantariksh: this is actually a copypasta mistake haha oops. I actually like the theme keys because it's not immediately obvious to new devs that the string must be consistent with the one in the Chakra repo.

PR #2541, opened 2021-08-06
feat(sms-limiting): ui changes for modal

seaerchin:
## Description
This changes the UI to be more polished.

This takes place on 3 fronts: 
1. adding a quota beneath the toggle so that users are able to see their sms counts. 
2. changing the sms counts modal button from red -> blue.
3. updating the sms counts modal description.

## Screenshots
**Quota**
Screenshot 2021-08-06 at 12 16 54 PM

**Modal**
Screenshot 2021-08-06 at 12 17 39 PM

### DIFF
--- a/src/public/modules/forms/admin/directiveViews/configure-mobile.client.view.html
+++ b/src/public/modules/forms/admin/directiveViews/configure-mobile.client.view.html
@@ -1,15 +1,22 @@
 <div class="row">
-  <div class="toggle-option">
+  <div class="toggle-option label-bottom">
     <div
       ng-class="{'toggle-option--disabled' : field.hasAdminExceededSmsLimit || field.hasRetrievalError}"
-      class="col-xs-8 label-custom label-medium label-bottom"
+      class="col-xs-8 justify-content-start"
     >
-      OTP verification
-      <i
-        class="glyphicon glyphicon-question-sign"
-        uib-tooltip="When enabled, respondents must verify by entering a code sent to their mobile number"
-        tooltip-trigger="'click mouseenter'"
-      ></i>
+      <div class="label-custom label-medium">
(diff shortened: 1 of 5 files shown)


src/public/modules/forms/admin/directiveViews/configure-mobile.client.view.html
```diff
@@ -1,15 +1,22 @@
 <div class="row">
-  <div class="toggle-option">
+  <div class="toggle-option label-bottom">
     <div
       ng-class="{'toggle-option--disabled' : field.hasAdminExceededSmsLimit || field.hasRetrievalError}"
-      class="col-xs-8 label-custom label-medium label-bottom"
+      class="col-xs-8 justify-content-start"
```
karrui, line 5: Can you double check if `justify-content-start` does anything? We are still using bootstrap 3, I don't see any documentation on this css

karrui (approved):
small comment, otherwise lgtm

src/public/modules/forms/admin/directiveViews/configure-mobile.client.view.html
seaerchin: I referenced the newer version of bootstrap [here](https://getbootstrap.com/docs/4.0/layout/grid/#horizontal-alignment) where it aligned horizontally as I wasn't aware of the exact version of bootstrap used. 

Will double check if this does anything

karrui: The current version of bootstrap being used can be found in our package.json

seaerchin: checked; this is extra, will remove

PR #2576, opened 2021-08-12
feat(design-system): add Footer component

### LINKED ISSUE #2479: [React] Add Navigation/Footer pattern component
**Time estimation: 1d**

**Depends on**
- Link

karrui:
~~This PR builds on top of #2574 as it requires links to be colourable (is that even a word)~~ #2574 has been merged

## Problem

This is less of a design system component and more of a Form specific footer component. Nevertheless, some care has been taken to ensure parts of the components can be used by themselves.

The theme was also not moved into the global design system theme as there are just too many subcomponents and makes more sense to be on the component themselves since they are building off smaller components.

In atomic design, this is more of a template rather than an atom, and we should only add specific themes to the atomic components imho.

Closes #2479 

## Solution

**Features**:

- feat: add footer assets like ogp brand svgs and social media icons
- feat(Footer): add responsive Footer component
- feat(Footer): add responsive stories

## Before & After Screenshots
In storybook. (Not working right now, will give screenshots insteaddddd)

Screenshot 2021-08-12 at 2 18 20 PM
Screenshot 2021-08-12 at 2 19 02 PM
Screenshot 2021-08-12 at 2 19 18 PM


karrui:
Grrr, I've emailed Chromatic to enable it and they've responded with the set up instructions... Guess they have not enabled it yet

frontend/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { BxlYoutube } from '~assets/icons/BxlYoutube'
+import { ReactComponent as OgpLogoSvg } from '~assets/svgs/brand/ogp-logo.svg'
+import { ReactComponent as OgpLogoFullSvg } from '~assets/svgs/brand/ogp-logo-full.svg'
+import Link from '~components/Link'
+
+const OgpLogoFull = chakra(OgpLogoFullSvg)
+const OgpLogo = chakra(OgpLogoSvg)
+
+const SOCIAL_MEDIA_LINKS = {
+  facebook: 'https://www.facebook.com/opengovsg',
```
mantariksh, line 26: hmmmmmmmmmm is there a way to do this which will make i18n easier in future? also generally I'm thinking it might be neater to put all the links in one place so it's easy to check that all the URLs in the frontend are up to date

mantariksh (approved):

frontend/src/components/Footer/Footer.tsx
karrui: I'll make a PR to incorp i18n after this then

karrui: Actually for links they can probably just live in a constant file

mantariksh: maybe we all live in a constant file

PR #2584, opened 2021-08-13
fix: Add 10 MB max length checks to form field editor

### LINKED ISSUE #1814: Pre-validate hooks not running
Due to the new usage of `findByIdAndUpdate`, certain pre-validate and pre-save hooks are not running as `findByIdAndUpdate` does not activate them. For example, the Form model has a check to ensure that the total form size does not exceed 10MB, but this check does not run when the Settings page is updated as the settings update uses `findByIdAndUpdate`. The Verification model also has a check for duplicate field IDs which is not run when Verification documents are updated.

The following fixes should be made to ensure that the functionality in the hooks is maintained:
1. For the Form model, the check for object size should be moved in to the Dropdown and Checkbox fields, which have a risk of becoming too large as the user is allowed to copy and paste multiple options into a textarea.
2. For the Form model, the checks for the admin and permissionList should be moved into the admin and permissionList keys.
3. For the Verification model, the check for duplicate field IDs should be moved into the `fields` key.
4. All other instances of `findByIdAndUpdate` and `findOneAndUpdate` should be checked to ensure that the models do not rely on Mongoose hooks to maintain the correct state.

frankchn:
## Problem
Closes #1814 

## Solution
As discussed, adds relevant checks for field size on the form field editor itself.

### DIFF
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -1,4 +1,4 @@
-import BSON, { ObjectId } from 'bson-ext'
+import { ObjectId } from 'bson-ext'
 import { compact, omit, pick, uniq } from 'lodash'
 import mongoose, {
   Mongoose,
@@ -9,7 +9,6 @@ import mongoose, {
 } from 'mongoose'
 import validator from 'validator'
 
-import { MB } from '../../../shared/constants/file'
 import {
   ADMIN_FORM_META_FIELDS,
   EMAIL_FORM_SETTINGS_FIELDS,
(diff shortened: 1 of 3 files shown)


mantariksh:
frontend code tested and lgtm, but I'm rethinking our decision to remove the backend check. while the check doesn't run on every single update to the form (due to the nuances of `findByIdAndUpdate` as mentioned in the issue), I think it's still better to have it than to omit it, so it at least runs on the form updates which don't use `findByIdAndUpdate`. wdyt?

frankchn:
@mantariksh yeah I think that is fair, undoing the removal.

mantariksh (approved):

PR #2585, opened 2021-08-13
fix(sms-limiting): hide counts when forms are onboarded

seaerchin:
## Problem
1. Modal shows the counts when teh form is onboarded. 

## Solution 
1. Hide the count


karrui (approved):

karrui:
Will force push to fix the testcafe thingamagic

PR #2586, opened 2021-08-13
fix(verification): prevents otp request when limit exceeded

### LINKED ISSUE #2531: Fix generate OTP issue
after 11 aug

seaerchin:
## Problem
At the present moment, users that have already loaded the page can continue to verify their mobile number, even when the form then disables OTP verification 
Closes #2531

## Solution
1. check if the form is onboarded 
2. check if the form has verifiable mobile fields

**Notes** 
users can request OTP from 2 endpoints - `handleGetOtp` or `handleGenerateOtp`. this check is only added on `handleGenerateOtp` because there are very few calls to `handleGetOtp` and we would have to query for actual form level data from the db to establish if the otp generation is valid.


karrui (approved):
lgtm

PR #2592, opened 2021-08-16
chore: freeze test-cafe package version to 1.15.1

karrui:
Also reverts opengovsg/FormSG#2572

The latest version is causing testcafe to be extremely slow, causing test failures due to tests running before the page renders. Revert and freeze to previous version. There is almost no point in updating testcafe as e2e tests are slowly becoming less and less useful due to the impending React migration. We will probably re-evaluate the e2e testing framework used then.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1447,9 +1447,9 @@
       }
     },
     "@babel/plugin-proposal-async-generator-functions": {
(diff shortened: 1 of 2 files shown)


mantariksh (approved):
Screenshot 2021-08-16 at 10 37 42 AM

Issue #2578, opened 2021-08-12
Email domain validation should be case-insensitive


mantariksh:
e.g. if an email field is restricted to `@open.gov.sg` emails, then `@OPEN.GOV.SG` should still be valid.

mantariksh:
Closed via #2682

PR #2622, opened 2021-08-17
fix: Email domain validation should be case-insensitive (#2578)

daffodilistic:
## Problem
Allow submission of a form from the public if the email domain is in all caps (`JOHN_DOE@FOO.BAR`) and mixed case.(`xXx_joHn_DoE_xXx@fOo.bAr`)

Closes [#2578]

## Solution
Validate against lower-cased email string from client by converting the email domain portion to lowercase.

**Breaking Changes** 
- [X] No - this PR is backwards compatible.

**Features**:

N/A

**Improvements**:

Users can now enter emails with either all caps or mixed case.

**Bug Fixes**:

A HTTP 400 response will not be sent to client if user enters an email with either all caps or mixed case.

## Before & After Screenshots

**BEFORE**:
N/A

**AFTER**:
N/A

## Tests
Should test on both client & server, but this is the best I can do for now (~2 hours spent with minimal Angular/TypeScript experience 😅)

## Deploy Notes
Not required - application code fix

None

**New scripts**:

None

**New dependencies**:

None

**New dev dependencies**:

None


LoneRifle:
```
> FormSG@5.23.0 build-backend /home/travis/build/opengovsg/FormSG
> tsc -p tsconfig.build.json
src/app/utils/field-validation/validators/emailValidator.ts:41:31 - error TS2532: Object is possibly 'undefined'.
41     const emailDomain = '@' + emailAddress.split('@').pop().toLowerCase()
                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/app/modules/submission/email-submission/email-submission.middleware.ts
```diff
@@ -64,8 +64,8 @@ export const validateResponseParams = celebrate({
             fieldType: Joi.string()
               .required()
               .valid(...Object.values(BasicField)),
-            answer: Joi.string().allow(''),
-            answerArray: Joi.array(),
+            answer: Joi.string().allow('').insensitive(),
+            answerArray: Joi.array().items(Joi.string().insensitive()),
```
karrui, line 68: nit: I don't think the `.insensitive()` part is necessary, since we don't restrict the types of strings we allow

karrui (approved):

mantariksh (commented):
do we need to deal with the converse case as well, i.e. convert the `allowedEmailDomains` list to lowercase before comparing?

mantariksh (changes requested):
requesting changes due to my previous review!

mantariksh:
@daffodilistic hey just letting you know, we're probably going to make the required changes in the next few days since this is a bug! feel free to make the changes if you have time in the meanwhile, but if not no worries, we'll pick it up.

timotheeg:
> do we need to deal with the converse case as well, i.e. convert the allowedEmailDomains list to lowercase before comparing?

Yes we should 👍 . It would have been nice to save the domains lowercased in the field definition itself at save time, but since it's not been done, we should lowercase the values in allowedDomains upon reading them from store. That or lowercase both operands of the comparisons as we go.

daffodilistic:
> @daffodilistic hey just letting you know, we're probably going to make the required changes in the next few days since this is a bug! feel free to make the changes if you have time in the meanwhile, but if not no worries, we'll pick it up.

Unfortunately I don't have the time to contribute on this for the next few working days, so please go ahead.

mantariksh:
Closing in favour of #2682

PR #2627, opened 2021-08-18
fix: Release 5.24.1 hotfix

tshuli:
## Problem
- Fixes a bug with #2586, which checked for disabled verified mobile field using falseyness of empty array instead of length
- Also fixes missing error message routing on `OtpRequestError` so that the correct error message is shown to user


timotheeg (approved):

tshuli:
Closing in favour of PR to push to dev branch to preserve changelog

PR #2628, opened 2021-08-18
feat(Radio): implement Others option

mantariksh:
Part of #2008, purely for code review. A separate PR will be opened for the final design review.

### Placement of context
Unlike in the Checkbox field, selection of the "Others" option in the radio button is not independent of the remaining options, since selecting "Others" means that all the other radio buttons are unchecked. This means it made sense for the context to wrap the entire radio field, rather than just the "Others" portion. This also allows the wrapper to check the state of the "Others" button and return the correct state to the parent form.

### Shape of returned value
For a Radio field, the parent form needs to know:
- Which option is selected?
- Is the selected option the in-built "Others" option?
- What text is in the "Others" text input?

This needs to cover edge cases such as:
- Text is present in Others input, but "Others" radio button is not selected (answer should be whichever radio button is selected)
- Text is present in Others input, "Others" radio button is selected and text is identical to one of the non-Others option (parent should be able to know that the answer is from the text input, not one of the radio options)

All this information can be represented with the shape
```
value: string
isOthers: boolean
```
If `isOthers` is true, the parent knows that the `value` comes from the text input. If not, the parent knows that the `value` is one of the radio options.

### Encapsulation of Others component
The "Others" component is exposed in a single component and the text input is not directly registered in the form individually, which allows for a simpler API with the parent. However, one drawback of this approach is that when "Others" is selected without text in the text input, `react-hook-form` is unable to return focus directly to the text input.


frontend/src/components/Radio/Radio.stories.tsx
```diff
@@ -93,3 +102,80 @@ const TemplateGroup: Story<RadioProps> = (args) => {
...
+    <form onSubmit={handleSubmit(onSubmit)}>
+      <FormControl isInvalid={!isEmpty(errors)} mb={6}>
+        <FormLabel isRequired>{label}</FormLabel>
+        <Controller
+          control={control}
+          name={name}
+          render={({ field }) => (
+            <Radio.RadioGroup {...field}>
+              {options.map((o, idx) => (
+                <Radio key={idx} value={o} {...args}>
```
karrui, line 147: i think field.ref may need to be passed into the first radio for react-hook-form to focus on the radio on error. do you want to try that?

frontend/src/components/Radio/useRadioGroup.ts
```diff
@@ -0,0 +1,22 @@
+import { createContext, RefObject, useContext } from 'react'
```
karrui, line 1: chakra-ui itself has a `useRadioGroup` hook. Would this naming cause confusion?

frontend/src/components/Radio/Radio.stories.tsx
mantariksh: hmm that seems to break everything, probably because we're passing the same `ref` to both `RadioGroup` and `Radio`

mantariksh: by "break everything" I mean it shows the "Answer is required" error even when there is an answer given

frontend/src/components/Radio/useRadioGroup.ts
mantariksh: sure, renamed to `useRadioGroupWithOthers` in https://github.com/opengovsg/FormSG/pull/2628/commits/d5e266b58793355cc0cab8684821649c3d695793

frontend/src/components/Radio/Radio.stories.tsx
mantariksh: resolved in https://github.com/opengovsg/FormSG/pull/2628/commits/ee75b15edfebbef1cf8c3f68547a8ffa71b2011a

mantariksh:
cherry-picked the relevant commits into #2669

PR #2629, opened 2021-08-18
fix(MailService): correct type, pass in original form admin as param

karrui:
## Problem

This PR fixes the bug where collab warning/disabled emails fail to send.

This occurs because the `form` param types passed into the functions to send the disabled/warning emails was not actually `IPopulatedForm`, and thus form.admin.email would have been undefined, resulting in a mail error where the emails failed to send.


tshuli (approved):
lgtm

PR #2630, opened 2021-08-18
fix: prevents otp request when limit exceeded

tshuli:
## Problem
- Fixes bugs with #2586
  - Checks for disabled verified mobile field using array length instead of falsyness
  - Disallows OTP for fields which are not verifiable even if `msgSrvcName` is set
  - Also fixes missing error message routing on `OtpRequestError` so that the correct error message is shown to user
  - Reimplements check at the field level instead of form level because `handleGenerateOtp`  is used by both email and sms fields, and also because there may be more than one verified mobile field
  - Also adds additional unit tests to prevent code regression

## Tests

- [ ] Create form with mobile verification. Open in public form view. Disable mobile verification. Attempt to verify and check that otp is not sent, and an error message appears in OTP box
- [ ] Now enable mobile verification. Check that OTP is sent normally and form can be submitted
- [ ] Add `msgSrvcName` to form and repeat the first test. Check that OTP cannot be sent.
- [ ] Now enable mobile verification. Add 10000 OTPs. Check that OTPs can still be sent and form can be submitted
- [ ] Check that email otp verification works normally
- [ ] Create form with two verified mobile fields and open in public view. Disable mobile verification for one field. Attempt to verify and check that otp is not sent, and an error message appears in OTP box for non-verified field, but otp is sent for verified field

### DIFF
--- a/src/app/modules/verification/__tests__/verification.service.spec.ts
+++ b/src/app/modules/verification/__tests__/verification.service.spec.ts
@@ -73,7 +73,8 @@ jest.mock('src/app/utils/hash')
 const MockHashUtils = mocked(HashUtils, true)
 
 describe('Verification service', () => {
-  const mockFieldId = new ObjectId().toHexString()
+  const mockFieldIdObj = new ObjectId()
+  const mockFieldId = mockFieldIdObj.toHexString()
   const mockField = { ...generateFieldParams(), _id: mockFieldId }
   const mockTransactionId = new ObjectId().toHexString()
   const mockFormId = new ObjectId().toHexString()
@@ -298,7 +299,12 @@ describe('Verification service', () => {
       const mockForm = {
         _id: new ObjectId(),
         title: 'mockForm',
-        form_fields: [],
+        form_fields: [
+          generateDefaultField(BasicField.Mobile, {
+            isVerifiable: true,
+            _id: mockFieldIdObj as unknown as string,
+          }),
+        ],
         msgSrvcName: 'abc',
       } as unknown as IFormSchema
 
@@ -462,7 +468,12 @@ describe('Verification service', () => {
       const mockForm = {
         _id: new ObjectId(),
         title: 'mockForm',
-        form_fields: [],
+        form_fields: [
+          generateDefaultField(BasicField.Mobile, {
+            isVerifiable: true,
+            _id: mockFieldIdObj as unknown as string,
+          }),
+        ],
         msgSrvcName: 'abc',
       } as unknown as IFormSchema
 
@@ -473,6 +484,7 @@ describe('Verification service', () => {
       MockSmsFactory.sendVerificationOtp.mockReturnValueOnce(errAsync(error))
       const field = generateFieldParams({
         fieldType: BasicField.Mobile,
+        _id: mockFieldIdObj as unknown as string,
       })
       const transaction = await VerificationModel.create({
         formId: mockFormId,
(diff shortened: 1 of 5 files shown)


src/app/modules/verification/verification.service.ts
```diff
@@ -322,7 +324,13 @@ export const sendNewOtp = ({
           return errAsync(new WaitForOtpError())
         }
 
-        return sendOtpForField(transaction.formId, field, recipient, otp)
+        return sendOtpForField(
+          transaction.formId,
+          fieldId,
+          field,
```
karrui, line 330: Do we need to pass in the fieldId? Won't the fieldId be present in given field parameter already?

tshuli: thanks for highlighting, edited accordingly

tshuli:
@karrui thanks for reviewing! please check in particular that `997d5f9` makes sense

src/app/modules/verification/verification.service.ts
```diff
@@ -477,12 +479,21 @@ const sendOtpForField = (
   | InvalidNumberError
   | MailSendError
   | NonVerifiedFieldTypeError
+  | SmsLimitExceededError
+  | OtpRequestError
```
timotheeg, line 483: Shouldn't `otpRequestError` be swallowed in this method rather than surfaced to caller?

As in, with my (probably very broken) understanding, an SMS limit error should be surfaced because the form is not behaving as a user would expect, but `otpRequestError` means the field didn't need an opt anyway, and so the form is performing correctly if no otp is sent

Or more likely I probably need a bit of help to understand what a field being not verifiable means here?

src/app/modules/verification/__tests__/verification.service.spec.ts
```diff
@@ -295,6 +296,20 @@ describe('Verification service', () => {
     })
 
     it('should send OTP and update hashes when parameters are valid', async () => {
+      const mockForm = {
```
karrui, line 299: this seems to be repeated a bunch of times, should we make it a constant just under describe block? (line 281?)

src/app/modules/verification/verification.service.ts
karrui: we surface all errors to the controllers anyways, and the controller will have a error handler that returns the correct error code depending on the error

karrui (approved):
lgtm

src/app/modules/verification/verification.service.ts
tshuli: Thanks! as what Kar Rui said. the error is surfaced to the controller so that it can be returned and then displayed on the frontend

In this case, the error is triggered when client still has the old form where OTP verification is enabled. the form (from the client's perspective) is not performing correctly if no otp is sent, hence the need for error message

src/app/modules/verification/__tests__/verification.service.spec.ts
tshuli: thanks for highlighting, edited!

src/app/modules/verification/verification.service.ts
timotheeg: Thanks guys! Just for extra info for me: what does `isVerifiable` on the field represent?

karrui: Whether the Mobile Number field has SMS verification enabled or not:

Screenshot 2021-08-19 at 9 59 23 AM
Screenshot 2021-08-19 at 9 59 27 AM

timotheeg: Hmm, but so why is there an error here?

tshuli: as explained together with karrui over huddle :)

timotheeg: Understood the flow. I had the complete incorrect assumption of the starting point 😑😰.

timotheeg (approved):
👍

PR #2632, opened 2021-08-18
fix(react/FormLabel): remove extra height when optional indicator is shown

karrui:
## Problem

Styling bug where height of label with optional text is different from one without the optional text. See gif of before and after.

## Before & After Screenshots

**BEFORE**:
recording (21)

**AFTER**:

recording (22)


timotheeg (approved):
👍

I love the PR description!

PR #2641, opened 2021-08-19
feat(react): add Public and Private Route components

karrui:
## Problem

This PR adds the PrivateRoute and PublicRoute components as the base for any new routes in the future.

Additionally, an example AuthContext is created to allow a mock generated logged in or logged out states

### DIFF
--- a/frontend/src/app/AppRouter.tsx
+++ b/frontend/src/app/AppRouter.tsx
@@ -1,0 +1,28 @@
+import { lazy, Suspense } from 'react'
+import { Route, Switch } from 'react-router-dom'
+
+import { LOGIN_ROUTE, ROOT_ROUTE } from '~constants/routes'
+
+import { PrivateRoute } from './PrivateRoute'
+import { PublicRoute } from './PublicRoute'
+
+const WorkspacePage = lazy(() => import('~pages/workspace'))
+const LoginPage = lazy(() => import('~pages/login'))
+
+export const AppRouter = (): JSX.Element => {
+  return (
+    <Suspense fallback={<div>Loading...</div>}>
+      <Switch>
+        <PublicRoute exact path={LOGIN_ROUTE}>
+          <LoginPage />
+        </PublicRoute>
+        <PrivateRoute exact path={ROOT_ROUTE}>
+          <WorkspacePage />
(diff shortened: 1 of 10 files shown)


frontend/src/app/AppRouter.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { PrivateRoute } from './PrivateRoute'
+import { PublicRoute } from './PublicRoute'
+
+const WorkspacePage = lazy(() => import('~pages/workspace'))
+const LoginPage = lazy(() => import('~pages/login'))
+
+export const AppRouter = (): JSX.Element => {
+  return (
+    <Suspense fallback={<div>Loading...</div>}>
```
mantariksh, line 14: is it wise to put a single suspense boundary so high up in the app? doesn't this mean that if we fail to wrap a lower-level lazy component in its own suspense boundary, the entire app will fallback to this div?

also don't we need a more deliberate design for the loading state?

also what suspense integration do we plan to use?

mantariksh (approved):
approving because temporary mocks

frontend/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,60 @@
+import { createContext, FC, useCallback, useContext, useState } from 'react'
+
+type AuthContextProps = {
+  user?: any
+  isAuthenticated?: boolean
```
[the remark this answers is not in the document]
karrui: mocked for now, will expand on it in the real implementation. The user object will probably be out of this context in the final implementation.

> can't we infer isAuthenticated based on whether user is defined or not? 

Yes we can; in the real implementation we actually store a `is-logged-in` string token in localStorage instead of the user object since we always retrieve the user from our /user endpoint

frontend/src/app/AppRouter.tsx
karrui: I think it's fine, since this lazy-load distinguishes between the authenticated and non-authenticated parts of the application.
But yes, you are correct if we use another lazy loaded component it will fallback to this div. This div is also temporary, probably need @pearlyong to design a catch-all loading screen :P

PR #2626, opened 2021-08-18
build: Release 5.24.0

tshuli:
**Improvements**
- feat(sms-limiting): changes for mail/modal [`#2565`](https://github.com/opengovsg/FormSG/pull/2565)

**Fixes**
- fix: disable eserviceid check modal for sgid forms [`#2593`](https://github.com/opengovsg/FormSG/pull/2593)
- fix(verification): prevents otp request when limit exceeded [`#2586`](https://github.com/opengovsg/FormSG/pull/2586)
- fix: prevents otp request when limit exceeded [`#2630`](https://github.com/opengovsg/FormSG/pull/2630)
- fix(MailService): correct type, pass in original form admin as param [`#2629`](https://github.com/opengovsg/FormSG/pull/2629)

**Dependency Updates**
- chore: freeze test-cafe package version to 1.15.1 [`#2592`](https://github.com/opengovsg/FormSG/pull/2592)
- chore(deps-dev): bump @types/jest from 26.0.24 to 27.0.0 [`#2573`](https://github.com/opengovsg/FormSG/pull/2573)
- chore(deps-dev): bump @types/jest from 27.0.0 to 27.0.1 [`#2589`](https://github.com/opengovsg/FormSG/pull/2589)
- chore(deps-dev): bump @typescript-eslint/eslint-plugin [`#2597`](https://github.com/opengovsg/FormSG/pull/2597)
- chore(deps-dev): bump @typescript-eslint/parser from 4.29.1 to 4.29.2 [`#2595`](https://github.com/opengovsg/FormSG/pull/2595)
- chore(deps-dev): bump axios-mock-adapter from 1.19.0 to 1.20.0 [`#2591`](https://github.com/opengovsg/FormSG/pull/2591)
- chore(deps-dev): bump core-js from 3.16.1 to 3.16.2 [`#2625`](https://github.com/opengovsg/FormSG/pull/2625)
- chore(deps-dev): bump jest-localstorage-mock from 2.4.14 to 2.4.16 [`#2581`](https://github.com/opengovsg/FormSG/pull/2581)
- chore(deps-dev): bump supertest from 6.1.4 to 6.1.5 [`#2571`](https://github.com/opengovsg/FormSG/pull/2571)
- chore(deps-dev): bump testcafe from 1.15.1 to 1.15.2 [`#2572`](https://github.com/opengovsg/FormSG/pull/2572)
[...]

PR #2566, opened 2021-08-11
build: release v5.23.0

karrui:
~~Pre-release for now and not updated the version due to having one last copy change before this release is pushed.~~ Copywriting changes next week

# Release v5.23.0

New week, new features. 
This week's release includes:
- automatically removing verification on verified mobile fields if free sms quota has been hit (10k per user)
- moving `shared` folder to root

## Features
- feat: remove frontend code to delete   spcp cookie [`#2551`](https://github.com/opengovsg/FormSG/pull/2551)
- feat: sms limiting   [`#2504`](https://github.com/opengovsg/FormSG/pull/2504)

## Improvements
- ref: migrate public-facing submission   types to shared folder   [`#2509`](https://github.com/opengovsg/FormSG/pull/2509)
- refactor: move src/types/api/auth to   shared/types/form/form_auth   [`#2510`](https://github.com/opengovsg/FormSG/pull/2510)
- feat: loosen esrvcId validation to only   restrict whitespace [`#2521`](https://github.com/opengovsg/FormSG/pull/2521)
- feat: move most of src/shared to root   shared folder [`#2519`](https://github.com/opengovsg/FormSG/pull/2519)

## Fixes
- fix: update response limit error   message for storage mode submissions   [`#2448`](https://github.com/opengovsg/FormSG/pull/2448)

## Dependency updates
- chore: update bcrypt dependencies in   package-lock [`#2530`](https://github.com/opengovsg/FormSG/pull/2530)
- fix(deps): bump   @opengovsg/spcp-auth-client from 1.4.8 to 1.4.9   [`#2544`](https://github.com/opengovsg/FormSG/pull/2544)
- fix(deps): bump @sentry/browser from   6.10.0 to 6.11.0 [`#2546`](https://github.com/opengovsg/FormSG/pull/2546)
- fix(deps): bump @sentry/integrations   from 6.10.0 to 6.11.0   [`#2550`](https://github.com/opengovsg/FormSG/pull/2550)
- fix(deps): bump aws-sdk from 2.958.0 to   2.959.0 [`#2523`](https://github.com/opengovsg/FormSG/pull/2523)
- fix(deps): bump aws-sdk from 2.959.0 to   2.960.0 [`#2528`](https://github.com/opengovsg/FormSG/pull/2528)
- fix(deps): bump aws-sdk from 2.960.0 to   2.961.0 [`#2535`](https://github.com/opengovsg/FormSG/pull/2535)
[...]

PR #2646, opened 2021-08-20
build: hotfix release v5.24.1

karrui:
## Hotfix release
This PR fixes a bug where sending of verified SMSes would be blocked if the form admin's free SMS quota has been hit. This prevents  the vicious cycle where
1. Automatically disabling verified sms fields only happen after a verification sms is sent, and the admin has exceeded their quota.
2. sms is blocked from being sent, resulting in (1) not happening, resulting in the verified sms field not being disabled, resulting in sms being blocked... and the cycle continues.

By only checking whether the current field to send SMS is verifiable, it solves the above cycle.
- If user has hit the free quota but still has a verifiable sms field in their form, then
  - when public user invokes sms verification, the sms will be sent and server will disable mobile verification on all their free forms
  - when the next user (on the old verifiable form) invokes the sms verification, the same field will now not be verifiable and correctly return an error

### More context:
Release v5.24.0 (#2626) added a check to prevent verification SMSes from being sent out if the user’s free sms limit has exceeded.
This revealed a bug. If a form's admin has exceeded the free sms verification limit prior to the automation of the disabling of the verifiability of the mobile field (in release v5.23.0 (#2566)), then their verified mobile field (that was not automatically disabled) would stop working due to the check.

Currently, If a user’s free sms limit is hit, all verified mobile fields of forms that have not been onboarded (still using the free tier) will automatically have its verifiable status removed (which reverts the field to a normal phone number field). Works as intended when the user hits their limit.

However, this toggling of verifiable state is only done when the admin account exceeds the free limit tier (10k smses).
This means that users who have already exceeded the free tier prior to the sms limiting feature release will not have their verified sms fields toggled off, and since the limit has exceeded, any SMS sending will be rejected by the server. The public respondent will then see an error message that prompts them to refresh (but that would not fix the issue).


karrui:
## Tests
- [ ] Set the admin's current sms count to 9999 (1 below limit) (can do this by deleting/creating a new user and populating counts)
- [ ] Create an email form with 2 verifiable mobile fields. Send the first sms. Should succeed. Disabled email should be sent to admins and collaborators.
- [ ] Send second sms. Should now return an error since limit is hit. Refresh the form and the sms fields should not be verifiable.
- [ ] With the admin now at limit (10k), manually toggle on a field's isVerifiable status in the database. 
  - [ ] Refresh the form. The sms field should now be verifiable again. Verify sms. Should receive disabled email again. The field should go back to being unverifiable after a refresh.

tshuli (dismissed):
lgtm with minor nit - should we also remove `SmsLimitExceededError` as error type for `sendNewOtp` in `verification.service.ts`?

tshuli:
Tests ok too

tshuli:
forced push to adjust the error typings as mentioned in comment above

src/app/modules/verification/verification.service.ts
```diff
@@ -479,16 +477,15 @@ const sendOtpForField = (
   | InvalidNumberError
   | MailSendError
   | NonVerifiedFieldTypeError
-  | SmsLimitExceededError
   | OtpRequestError
 > => {
-  const { fieldType, _id } = field
+  const { fieldType, _id: fieldId } = field
```
timotheeg, line 482: 👍 reads much better!

timotheeg (approved):
lgtm 👍

tshuli (approved):

PR #2648, opened 2021-08-20
feat(design-system): add Toggle component and stories

mantariksh:
Implements Toggle component.

### Inclusion of icon in thumb
The original Chakra UI Switch [implementation](https://github.com/chakra-ui/chakra-ui/blob/main/packages/switch/src/switch.tsx) did not have the option to include an icon within the thumb (circular part of the Switch). Since we need to have icons in the thumb, I copied the implementation and adapted it to include the correct icon (checked/unchecked/disabled).


frontend/src/theme/components/Toggle.ts
```diff
@@ -0,0 +1,109 @@
...
+    },
+    track: {
+      bg: 'neutral.400',
+      _checked: {
+        bg: 'success.500',
+      },
+      p: 0,
+      _focus: {
+        // disable Chakra UI's default shadow
+        boxShadow: `none`,
```
karrui, line 48: Could we not have used a double box shadow?

```
boxShadow: `0 0 0 2px white, 0 0 0 3px var(--chakra-colors-primary-500)`;
```

mantariksh:
need to fix a11y, label is not linked to toggle

mantariksh:
> need to fix a11y, label is not linked to toggle

done

frontend/src/theme/components/Toggle.ts
mantariksh: excuse me while I go drown myself

karrui (approved):
lgtm

PR #2583, opened 2021-08-13
feat(Checkbox): implement Others option

mantariksh:
- Fixes Checkbox component to use `forwardRef`
- Implements "Others" option for checkbox
- Omits unsupported SVG props from `BxCheckAnimated` to get rid of React warnings in console

Part of #2008, purely code review for now. Design review will be done on `form-v2/checkbox-radio` branch when it is complete.

PR #2669, opened 2021-08-24
feat(design-system): add Checkbox and Radio components

### LINKED ISSUE #2008: [React] Add Field/Checkbox & Radio component

mantariksh:
Implements individual Checkbox and Radio components as well as options to add an "Others" text input for each.

Culmination of #2525, #2540, #2583, #2628.

Closes #2008


karrui (approved):
leggo

PR #2675, opened 2021-08-25
build(react): set up Docker image for prod and dev

### LINKED ISSUE #1904: [React] Set up CI/CD pipeline for building of React application

karrui:
## Problem

This PRs sets up the Docker containers for both prod and dev.

### Development mode
In development mode, docker-compose only builds the backend dependencies such as the server, localstack, and the database.
Since client should (ideally) be completely separate from the server, we can start the frontend independently from the server. This allows for better DX in terms of hot reloading, not needing to rebuild the Docker image when we update npm packages, etc.

To start the server, run Docker as per normal with `npm run dev` in **the root folder** (or just docker-compose up)
To start the client, go into `/frontend` and run `npm start`.

To view the client, go to `localhost:3000`. Even though our server will be running on `localhost:5000`, there is a `proxy` parameter in `frontend/package.json` (courtesy of create-react-app) that will proxy all API calls on 3000 to port 5000 automatically. This only happens on development mode.

### Production mode
Dockerfile.production has been modified to work and build correctly. 
Whilst doing so, encountered a couple of obstacles, such as the styleSrc CSP headers needing to be `'unsafe-inline'` due to the usage of emotion. This may be temporary if we want to run a postbuild script that adds a template variable for a nonce into the built index.html that create-react-app creates.

Closes #1904


.travis.yml
```diff
@@ -0,0 +1,86 @@
+os: linux
```
mantariksh, line 1: actually why not take this opportunity to use GA for deployment? if not we'll have to migrate again in future

karrui: this is just copy pasted from the old code; still waiting on @tshuli for the GA portion, didn't want to supersede his work

src/app/loaders/express/index.ts
```diff
@@ -108,40 +107,16 @@ const loadExpressApp = async (connection: Connection) => {
...
-  app.use(nocache()) // Add headers to prevent browser caching front-end code
-
   // Generate UUID for request and add it to X-Request-Id header
   app.use(addRequestId())
 
-  // Setting the app static folder
-  app.use('/public', express.static(path.resolve('./dist/frontend')))
-
-  // Point crawlers to our robots.txt
-  app.use(
```
mantariksh, line 132: do we not need to keep the functionality in this line?

src/app/loaders/express/index.ts
```diff
@@ -166,6 +141,16 @@ const loadExpressApp = async (connection: Connection) => {
   // New routes in preparation for API refactor.
   app.use('/api', ApiRouter)
 
+  // Serve static client files only in prod
+  if (config.nodeEnv === Environment.Prod) {
+    const frontendPath = path.resolve('dist/frontend')
+    app.use(express.static(frontendPath))
+
+    app.get('*', (_req, res) => {
```
mantariksh, line 149: the order matters here, right? if we do `app.get('*')` before `app.use('/api`, that'll mess up our APIs as far as I can tell. would be good to add a comment here stating that it's important for this block to be below all our other `app.use`.

package.json
```diff
@@ -19,14 +19,13 @@
...
-    "build-frontend": "webpack --config webpack.prod.js",
-    "build-frontend-dev": "webpack --config webpack.dev.js",
-    "build-frontend-dev:watch": "webpack --config webpack.dev.js --watch",
+    "build": "npm run build:backend && npm run build:frontend",
+    "build:frontend": "npm run --prefix frontend build",
+    "build:backend": "tsc -p tsconfig.build.json",
     "start": "node -r dotenv/config dist/backend/src/app/server.js",
     "dev": "docker-compose up --build",
-    "docker-dev": "npm run build-frontend-dev:watch & ts-node-dev --respawn --transpile-only --inspect=0.0.0.0 --exit-child -r dotenv/config -- src/app/server.ts",
+    "dev:backend": "tsnd --respawn --transpile-only --inspect=0.0.0.0 --exit-child -- src/app/server.ts",
```
mantariksh, line 27: are the `dotenv` options no longer required?

package.json
```diff
@@ -19,14 +19,13 @@
...
-    "build": "npm run build-backend && npm run build-frontend",
-    "build-backend": "tsc -p tsconfig.build.json",
-    "build-frontend": "webpack --config webpack.prod.js",
-    "build-frontend-dev": "webpack --config webpack.dev.js",
-    "build-frontend-dev:watch": "webpack --config webpack.dev.js --watch",
+    "build": "npm run build:backend && npm run build:frontend",
+    "build:frontend": "npm run --prefix frontend build",
+    "build:backend": "tsc -p tsconfig.build.json",
     "start": "node -r dotenv/config dist/backend/src/app/server.js",
     "dev": "docker-compose up --build",
```
mantariksh, line 26: I think it might not be very intuitive DX if `npm run dev` doesn't start up everything we need. is there a way to arrange the scripts to make that possible? i.e. to make this something like `docker-compose up --build & npm run --prefix frontend start`?

frontend/src/pages/workspace/WorkspacePage.tsx
```diff
@@ -1,13 +1,21 @@
...
+import { Divider } from '@chakra-ui/react'
+
 import { useAuth } from '~contexts/AuthContext'
+import { getLandingPageStatistics } from '~services/AnalyticsService'
 import Button from '~components/Button'
 
 export const WorkspacePage = (): JSX.Element => {
   const { user, logout } = useAuth()
 
+  const { data, isLoading } = useQuery('stats', getLandingPageStatistics)
```
mantariksh, line 11: assuming this is temporary and we'll wrap this query in a custom hook when we actually implement it

frontend/package.json
```diff
@@ -27,7 +29,7 @@
     "gen:theme-typings": "chakra-cli tokens src/theme/index.ts",
     "postinstall": "npm run gen:theme-typings",
     "start": "cross-env SKIP_PREFLIGHT_CHECK=true react-app-rewired start",
-    "build": "cross-env SKIP_PREFLIGHT_CHECK=true react-app-rewired build",
+    "build": "cross-env CI=false BUILD_PATH='../dist/frontend' INLINE_RUNTIME_CHUNK=false SKIP_PREFLIGHT_CHECK=true react-app-rewired build",
```
mantariksh, line 32: lol can we add a comment here on why these options are required pls? we will definitely forget

src/app/loaders/express/index.ts
```diff
@@ -166,6 +141,16 @@ const loadExpressApp = async (connection: Connection) => {
   // New routes in preparation for API refactor.
   app.use('/api', ApiRouter)
 
+  // Serve static client files only in prod
+  if (config.nodeEnv === Environment.Prod) {
+    const frontendPath = path.resolve('dist/frontend')
```
mantariksh, line 146: where in the React config do we specify that the React build files should output to `dist/frontend`?

frontend/src/pages/workspace/WorkspacePage.tsx
karrui: correct, this is just to test whether the frontend is correctly calling backend APIs

src/app/loaders/express/index.ts
karrui: in the build path flag in package.json `BUILD_PATH='../dist/frontend'`

frontend/package.json
karrui: I wish we can add comments to package.json... I guess we can go with a hacky way as demonstrated in this [SO post](https://stackoverflow.com/a/45815391) and add a `scriptComments` key?

src/app/loaders/express/index.ts
karrui: We do! Opened an issue #2729 to track

package.json
karrui: Thanks for the catch, probably missed from the merging of develop back into this branch

frontend/package.json
karrui: Done in e2aacb7

package.json
karrui: Done in dcd3aab

src/app/loaders/express/index.ts
karrui: Done

mantariksh (approved):

PR #2683, opened 2021-08-26
feat(react): sign in feature

### LINKED ISSUE #2489: [React] Sign in flow feature
**Time estimation: 1w**

This issue tracks the completion of the sign in flow.

Sign in flow mocks can be found at 
Figma > DesignMaster_v3 > Landing,Login,Examples,Empty,Error

karrui:
## Problem

This PR implements the sign in feature, with mocked API calls using `msw`. This allows for the storybook to be rendered and working almost as close to the real thing (calling server endpoints) as possible.

Closes #2489 

## Solution

**Features**:

- Add LoginPage component and related stories. Subcomponent consists of
  - OtpForm, 
  - LoginForm, 
  - ResendOtpButton. This component implements a cooldown timer of 60 seconds using the a newly implemented `useInterval` hook.
- Add a bunch of new hooks to facilitate this feature:
  - `useLocalStorage` used for storing logged in state (in favour of storying the user object, since the user object will be stale and can be fetched from the /user endpoint)
  - `useInterval` used for rendering updates in an interval
  - `useUser` basically a react-query hook to retrieve the current logged in user
  - `AuthContext/useAuth` hook providing a bunch of authentication related functions
- Add a bunch of services (copy-pasted from AngularJS folder) to facilitate this feature
  - UserService
  - AuthService
  - new ApiService (using `axios`, but may switch to `wretched`?)

- Add `msw` package for mocking API calls. Used inside storybook (and locally, can be disabled when the server is up)

## Before & After Screenshots
Working flow in storybook

No need to scroll on smaller laptop screen resolutions

1366px x 768px desktop layout
localhost_6006_iframe html_id=pages-loginpage--desktop args= viewMode=story(13_ Laptop 1366px)

1280px x 800px desktop layout
localhost_6006_iframe html_id=pages-loginpage--desktop args= viewMode=story(13_ Laptop 1280px)

## Deploy Notes

**New dependencies**:

- `react-query` for making api calls

**New dev dependencies**:
- `msw` for mocking network calls in app
- `msw-storybook-addon` for mocking network calls in storybook


frontend/public/mockServiceWorker.js
```diff
@@ -0,0 +1,338 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker (0.35.0).
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ * - Please do NOT serve this file on production.
```
mantariksh, line 8: so I ran `npm run build` and saw that `mockServiceWorker.js` is included in `frontend/build` (which presumably will change to `dist/frontend` once we configure it correctly). if we use `app.use(express.static(path.resolve('dist/frontend')))` (as per #2675) will that mean someone would be able to get this file from `form.gov.sg/mockServiceWorker.js`?

frontend/src/hooks/useUser.ts
```diff
@@ -0,0 +1,29 @@
...
+type UseUserReturn = {
+  user: UserDto | undefined
+  isLoading: boolean
+}
+
+export const useUser = (): UseUserReturn => {
+  const [isLoggedIn] = useLocalStorage<boolean>(LOGGED_IN_KEY)
+
+  const { data: user, isLoading } = useQuery<UserDto>(
+    'currentUser',
```
mantariksh, line 19: how are we going to manage query keys when we have many many custom hooks? any reliable way to prevent duplication?

frontend/src/contexts/AuthContext.tsx
```diff
@@ -1,10 +1,15 @@
...
+import { useLocalStorage } from '~hooks/useLocalStorage'
+import * as AuthService from '~services/AuthService'
 
 type AuthContextProps = {
-  user?: any
   isAuthenticated?: boolean
-  login: () => void
-  logout: () => void
+  sendLoginOtp: typeof AuthService.sendLoginOtp
+  verifyLoginOtp: (params: { otp: string; email: string }) => Promise<void>
```
mantariksh, line 11: do we really need to couple the auth _state_ (isAuthenticated, logout, login) with the auth _mechanism_ (OTP)? intuitively I'd think the auth context should be agnostic to auth mechanism, and just have the single responsibility of keeping track of the authentication state

frontend/src/index.tsx
```diff
@@ -8,6 +8,10 @@ import { App } from './app/App'
 import reportWebVitals from './reportWebVitals'
 import * as serviceWorker from './serviceWorker'
 
+if (process.env.NODE_ENV === 'development') {
```
mantariksh, line 11: how does this env var get injected into the frontend? currently the setup is such that `docker-compose` will inject all the required env vars in the dev environment even if there's no `.env` - will this behaviour be preserved?

also should this be `test` instead of `development`? we don't want API calls to be mocked for development right?

frontend/src/typings/vendor/msw-storybook-addon.d.ts
```diff
@@ -0,0 +1,12 @@
+declare module 'msw-storybook-addon' {
```
mantariksh, line 1: was this autogenerated / why is it necessary?

frontend/src/mocks/msw/handlers/auth.ts
```diff
@@ -0,0 +1,71 @@
+import { rest } from 'msw'
```
mantariksh, line 1: is there a way to look at the behaviour of components in Storybook for different API responses? e.g. change whether the call to `/otp/generate` succeeds or fails, and see how the component responds to each?

frontend/src/pages/login/LoginPage.tsx
```diff
@@ -1,14 +1,212 @@
...
+import { OtpForm, OtpFormInputs } from './components/OtpForm'
+
+export type LoginOtpData = {
+  email: string
+}
+
+const BrandLogo = chakra(BrandLogoSvg)
+const LoginImage = chakra(LoginImageSvg)
+
+export const LoginPage: FC = () => {
```
mantariksh, line 32: what happened to our `JSX.Element` convention haha

frontend/src/pages/login/LoginPage.tsx
```diff
@@ -1,14 +1,212 @@
...
+          minH={{ base: 'initial', lg: '100vh' }}
+          maxW="90rem"
+          margin="auto"
+          templateAreas={{
+            base: `'login'`,
+            md: `'sidebar' 'login'`,
+            lg: `'sidebar login' 'copy links'`,
+          }}
+          templateRows={{ lg: '1fr auto' }}
+          templateColumns={{ lg: '5fr 7fr' }}
```
mantariksh, line 98: do you think there might be a better way to manage all these styles instead of inlining them? the component is damn hard to read rn, there's so much CSS that it's hard to get a sense of what the general structure of the component is and how the component state interacts with the DOM. might it be worth extracting all this CSS into a separate `ts` file so we can just spread in the component?

frontend/src/pages/login/components/LoginForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+    return isGovDomain || 'Please log in with a gov.sg email address.'
+  }, [])
+
+  const onSubmitForm = async (inputs: LoginFormInputs) => {
+    return onSubmit(inputs).catch((e) => {
+      setError('email', { type: 'server', message: e.message })
+    })
+  }
+
+  const isMobile = useBreakpointValue({ base: true, xs: true, lg: false })
```
mantariksh, line 41: the behaviour of this hook is not clear to me from the docs - does `md` default to `true` or `false` in this case?

frontend/src/services/ApiService.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    if (typeof e.response.data === 'string') {
+      return new HttpError(e.response.data, statusCode)
+    }
+    if (e.response.data?.message) {
+      return new HttpError(e.response.data.message, statusCode)
+    }
+    if (e.response.statusText) {
+      return new HttpError(e.response.statusText, statusCode)
+    }
```
mantariksh, line 33: I'm not sure we want to do this - the status text ("Forbidden", "Unauthorized") is rarely useful info for a user. personally I'd rather default to a generic "contact us" error message at this point

frontend/src/pages/login/LoginPage.tsx
karrui: BETRAYED BY MY OLD SNIPPET

frontend/public/mockServiceWorker.js
karrui: Yes they would, but it would be useless because the service worker would not be registered anyways. I guess we can add a .dockerignore entry and it'll not be in the image

frontend/src/contexts/AuthContext.tsx
karrui: You make a great point. Will move the functions out

frontend/src/hooks/useUser.ts
karrui: Seems like a great article about this, wdyt: https://tkdodo.eu/blog/effective-react-query-keys

frontend/src/typings/vendor/msw-storybook-addon.d.ts
karrui: There are no types for that package, so I had to create one. These particular types are retrieved from a languished PR in that repository

frontend/src/pages/login/LoginPage.tsx
karrui: We'll lose autosuggestions if that's the case. However, I can do it styled components style and create new components that are more indicative of what they do

frontend/src/mocks/msw/handlers/auth.ts
karrui: yea, it is possible. We just need to create alternative handlers for the error cases. I'll add one non-whitelisted test case then.

frontend/src/services/ApiService.ts
karrui: Out of scope of this PR? (even though I created it here :X) We can revisit this in the future?

frontend/src/pages/login/components/LoginForm.tsx
karrui: `md` will follow `base`, so `true`. But I've removed this in favour of just hiding the display

frontend/src/pages/login/LoginPage.tsx
karrui: Done in 99371c9, is this more or less what you were looking for?

frontend/src/contexts/AuthContext.tsx
karrui: Done

frontend/src/hooks/useUser.ts
karrui: Implemented in 12ed93f

frontend/src/index.tsx
karrui: Moved to `test` environment by default

frontend/public/mockServiceWorker.js
mantariksh: sounds good

frontend/src/pages/login/LoginPage.tsx
```diff
@@ -28,16 +28,109 @@ export type LoginOtpData = {
...
+    pt="0.5rem"
+    pb="4rem"
+    display={{ base: 'none', lg: 'flex' }}
+    gridArea="links"
+    children={children}
+  />
+)
+
+// Grid area styling for the left sidebar that only displays on tablet and desktop breakpoints.
+const NonMobileSidebarGridArea: FC = ({ children }) => (
```
mantariksh, line 113: JSX.Element? haha

mantariksh (approved):

frontend/src/mocks/msw/handlers/auth.ts
karrui: added in 9363659

Issue #2611, opened 2021-08-17
[React] Public form MobileField component


karrui:
Used to render a mobile number field field on public forms.

Should take in a MobileField schema object and generate the appropriate validations using react-hook-form

karrui:
Closed by #3460

PR #2704, opened 2021-08-30
feat(design-system): add single country phone number input render variant

karrui:
Blocks #2611 

## Problem

The current phone number input is the international version, which allows for country selection. This PR adds a single country phone number input variant where country cannot be changed once the default country is set.

The implementation ended up having a lot of duplicate code, which has resolved by creating a `PhoneNumberInputContext` component, and sharing the relevant code amongst the newly implemented `SingleCountryPhoneNumberInput` and `IntlPhoneNumberInput` components.

## Solution

**Features**:

- feat: extract logic from PhoneNumberInput to PhoneNumberInputContext
- feat: add SingleCountryPhoneNumberInput component
- feat: add IntlPhoneNumberInput component that uses new context
- feat: update PhoneNumberInput to distinguish between single country and intl component to render


karrui:
Converting to draft, found a bug where the autoformat fails

timotheeg:
Still (slowly) reviewing the PR, but just to note: the storybook Playground for International PhoneNumberInput is broken for me 🤔 . I select my country and enter my number, but I see the country code isbeing prepended, and the validation fails (tested with valid SG and FR numbers):
image
image

karrui:
> Still (slowly) reviewing the PR, but just to note: the storybook Playground for International PhoneNumberInput is broken for me 🤔 . I select my country and enter my number, but I see the country code is being prepended, and the validation fails (tested with valid SG and FR numbers):

Validation is not in this PR's scope, the playground is only a "sneak peek" per se for the validation function.
However, the validation may be failing due the the validation function checking if the number given is a valid mobile phone number; the first number you used seems to be a landline number. Not sure about the French number. 

Nevertheless, the real validation function will probably be much looser (like in current implementation, using isPossibleNumber instead of isValidNumber)

frontend/src/components/PhoneNumberInput/IntlPhoneNumberInput.stories.tsx
```diff
@@ -92,7 +92,9 @@ export const Playground: Story = ({
             required: isRequired
               ? { value: true, message: 'Required field' }
               : false,
-            validate: (val) => isValidPhoneNumber(val) || 'Invalid number',
+            validate: (val) => {
+              return isValidPhoneNumber(val) || 'Invalid number'
```
karrui, line 96: @timotheeg here is the relevant line; however note that this is a story-only function and not really part of the component, and can be replaced with another validation function depending on the caller

timotheeg:
Oh wait no! The validation only works for mobile numbers, and I had entered landline numbers in both cases 🤦 😓 .

It wasn't clear to me this was mobile only. Is it intended?

image
image

frontend/src/components/PhoneNumberInput/IntlPhoneNumberInput.stories.tsx
timotheeg: Roger that. I spotted the import for mobile-only verification:
```typescript
import { isValidPhoneNumber } from 'libphonenumber-js/mobile'
```

All good 👍

timotheeg:
Thanks for the explanation @karrui!

Since the component lets caller specify the validation strategy, I reckon the stories should reflect that somehow. Maybe having 2 stories per phone input types, illustrating 2 different validations 🤔 . And also the playground should mention what validation it is using to prevent surprises like I just had 😅

karrui:
@timotheeg 

Will probably only do so when the real form components are being created; the current `PhoneNumberInput` is just a design system component for now, with no real need for even the Playground story -- currently only being provided so a11y can be tested as a form input

timotheeg:
Fair enough 👍 . We can revisit this later.

frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -79,231 +59,49 @@ export interface PhoneNumberInputProps
...
    */
   examples?: { [country in CountryCode]: NationalNumber }
+
+  /**
+   * Whether international phone numbers are allowed. Defaults to `true`.
+   * If allowed, the phone number input will be prefixed with the country code,
+   * and the selected country will be displayed in the input's left add-on, and
+   * autoformatting will be enabled.
+   */
+  isAllowInternational?: boolean
```
timotheeg, line 69: /supernit: `allowSomething` is already a `true|false` naming. The `is` prefix may not have been needed for the variable name.

timotheeg (approved):
lgtm 👍

PR #2710, opened 2021-08-31
fix: only check for sms quota on verified mobile field OTP verifications instead of all verified fields

karrui:
## Problem

Some admins are receiving spammy emails about their free SMS OTP tier quota being hit when the public verifies their email in a verified email field. This is due to the sms check being applied on every OTP generation, regardless of whether the field to verify is a mobile or email field.

This PR fixes the bug by only processing any sms checking action on verified mobile fields.

## Solution

**Breaking Changes** 

    - Details ...

**Bug Fixes**:

- Check mobile fieldType before processing admin sms email

### DIFF
--- a/src/app/modules/verification/verification.service.ts
+++ b/src/app/modules/verification/verification.service.ts
@@ -363,6 +363,35 @@ export const sendNewOtp = ({
   })
 }
 
+export const checkShouldDisableVerifiedFields = (
+  form: IPopulatedForm,
+  transaction: IVerificationSchema,
+  fieldId: string,
+): ResultAsync<boolean, void> => {
+  return getFieldFromTransaction(transaction, fieldId)
+    .asyncAndThen((field) => {
+      switch (field.fieldType) {
+        case BasicField.Mobile:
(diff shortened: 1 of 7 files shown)


src/app/modules/verification/verification.service.ts
```diff
@@ -363,6 +363,35 @@ export const sendNewOtp = ({
   })
 }
 
+export const checkShouldDisableVerifiedFields = (
```
mantariksh, line 366: I think the function name is a bit misleading here, because it's not just a check right? it performs a whole bunch of actions including updating potentially multiple forms and contacting admins. maybe we can rename to `disableVerifiedFieldsIfRequired`? or something like that

mantariksh (approved):

src/app/modules/verification/verification.service.ts
karrui: > there are 2 hard things in software engineering
your name is better; will change

karrui: Resolved in 765e5ed

PR #2712, opened 2021-08-31
feat(design-system): use FormHelperText for label description so SRs are aware

karrui:
## Problem

When focusing on the input field, only the label is being spoken out loud. The description should also be spoken out loud. 
This PR uses Chakra's FormHelperText component instead of Text to render so the input's aria-describedby prop contains the description id too.

### DIFF
--- a/frontend/src/components/FormControl/FormLabel/FormLabel.tsx
+++ b/frontend/src/components/FormControl/FormLabel/FormLabel.tsx
@@ -85,6 +85,12 @@ const FormLabelDescription = ({
   children,
   ...props
 }: TextProps): JSX.Element => {
+  // useFormControlContext is a ChakraUI hook that returns props passed down
+  // from a parent ChakraUI's `FormControl` component.
+  // The return object is used to determine whether FormHelperText or Text is
+  // used.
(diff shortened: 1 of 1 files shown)


frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -79,13 +81,30 @@ export const FormLabel = ({
 
 FormLabel.Label = ChakraFormLabel
 
-FormLabel.Description = ({ children, ...props }: TextProps): JSX.Element => {
+const FormLabelDescription = ({
+  children,
+  ...props
+}: TextProps): JSX.Element => {
+  const field = useFormControlContext()
```
mantariksh, line 88: this hook is not documented, can we add a comment here to say what it does pls?

mantariksh (approved):
approving first since just documentation requested

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
karrui: Resolved in 2547093. Sorry for the delay.

PR #2726, opened 2021-09-01
chore(deps): Update winston-cloudwatch

timotheeg:
## Context
Snyk reports a vulnerability in dependencies of [winston-cloudwatch@2.5.2](https://www.npmjs.com/package/winston-cloudwatch). See report [here](https://app.snyk.io/org/formsg/project/7d2653ec-310a-4c60-a94a-eca85308e725#issue-SNYK-JS-PACRESOLVER-1564857).

## Approach
Upgrade winston-cloudwatch to 3.0.2

Instantiation signature stays the same so upgrade should be backward compatible.

winston-cloudwatch@3 no longer lists aws-sdk as a direct dependency (set as a peer dependency instead), but since Form [does list `aws-sdk` as a dependency](https://github.com/opengovsg/FormSG/blob/5016b6d68e0ca21ec705a0ae378b6a291d2ae641/package.json#L85), we're good to go.

cc @mantariksh @karrui

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9742,9 +9742,9 @@
           "integrity": "sha512-eGuFFw7Upda+g4p+QHvnW0RyTX/SVeJBDM/gCtMARO0cLuT2HcEKnTPvhjV6aGeqrCB/sbNop0Kszm0jsaWU4A=="
         },
         "tslib": {
-          "version": "2.2.0",
-          "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.2.0.tgz",
-          "integrity": "sha512-gS9GVHRU+RGn5KQM2rllAlR3dU6m7AcpJKdtH8gFvQiC4Otgk98XnmMU+nZenHt/+VhnBPWwgrJsyrdcw6i23w=="
+          "version": "2.3.1",
+          "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.3.1.tgz",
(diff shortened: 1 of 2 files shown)


timotheeg:
Sigh, there is a type upgrade to do... Doing it now

timotheeg:
Looks like the type upgrade might be wrong in the typescript definition... 😢  I've opened an issue upstream [here](https://github.com/lazywithclass/winston-cloudwatch/issues/159)

But for now, to unblock this PR, I have added a dummy `name` into the constructor options with a TODO comment for future removal.

mantariksh (commented):
just to be sure, we should probably test this on staging by making sure that public-facing email notifications (i.e. for email confirmations and verified email fields) go to the custom Cloudwatch log group rather than the main log group. we do this so that we can make email data for public users expire after 1 week instead of the standard 1 year for logs, so we don't keep public user data around for unnecessarily long (the 1 week is useful e.g. when there are outages with SGmail causing submissions to be lost because they don't reach admins' email inboxes, upon which we can check that custom log group to find email addresses of respondents).

timotheeg:
Thanks for the advice, I'll push this to staging and try it out. I hesitated to provide an empty string in `name` this the constructor has a default value of 'CloudWatch' 🤔 

See here:
https://github.com/lazywithclass/winston-cloudwatch/blob/d77124fbba1a5986605e805c62938549aa7e1ad3/index.js#L17

karrui (approved):
lgtm

timotheeg:
Verified that logs still end up in cloudwatch log below:
```
/aws/elasticbeanstalk/formsg-staging-alb/var/log/eb-docker/containers/eb-current-app/stdouterr.log
```

Tested with a dummy staging form.
image


I'll be merging  now

PR #2735, opened 2021-09-03
wip(react): add (verifiable) Mobile field component

karrui:
> Note that this builds on top of #2704 

Still WIP, just want to put it out here so we can check if the approach makes sense. Will probably split up the verifiable component into its own wrapper so it can be reused for email verified fields too (kind of like the current AngularJS way)

TODO:
- [ ] Tests
- [ ] Split up verified component from the field itself
- [ ] More stories

A couple commits have been cherry picked into this branch from the sign in feature branch (#2683), such as the useInterval and ResendOtpButton implementations.

## Problem

See #2611 

## Solution

**Features**:

- Add verifiable mobile field components and stories.

**Fixes**:
- chore: add installation of shared folder packages in postinstall


karrui:
Will create a MobileField schema that does not care whether it is verifiable or not yet in a new PR to reduce review complexity

PR #2736, opened 2021-09-03
fix(radio): Remove unnecessary nbsp from option text

timotheeg:
## Context

We append a `&nbsp;` (non-breakable space) to **all** radio fields, which can cause a rendering error on mobile when the radio text just about covers the available space, as reported by a user for [this form](https://form.gov.sg/#!/611d22357f4b5f0012537c4a) on mobile.

image

## Solution
The empty line is caused by a non-breakable character forcefully added in code to **all** radio fields.

The code had the comment: "Empty space is to prevent visual bug when radio button is empty".

I am assuming that:
* that comment might no longer be applicable? (checkbox fields don't have the same `&nbsp;` for example)
* There shouldn't be empty options anyway?

~~A better option is probably to make the rendering conditional where `&nbsp;` is not appended to the value, it is instead the default when the value is falsy. I might look at this later, but I'm opening the PR first anyway to get some early feedback.~~

~~@liangyuanruo might remember how old the `&nbsp;` is and if it is still relevant?~~

The current form definition really cannot have empty option values for radios, but the comment still scares me that some old form entries in DB could have empty fields?

I have added a commit to conditionally show the option OR use an `&nbsp;`, instead of forcefully appending `&nbsp;` to **all** entries.

Tested locally with valid radio values

### DIFF
--- a/src/public/modules/forms/base/componentViews/field-radiobutton.client.view.html
+++ b/src/public/modules/forms/base/componentViews/field-radiobutton.client.view.html
@@ -41,7 +41,8 @@
               aria-labelledby="label-{{ vm.field._id || 'defaultID'}}"
               aria-describedby="description-{{ vm.field._id || 'defaultID'}}"
             />
-            {{ option }}
+            <span ng-if="option">{{ option }}</span>
+            <span ng-if="!option">&nbsp;</span>
             <span
               class="radiomark"
               ng-class="vm.field.fieldValue === option ? '{{ vm.colortheme }}-border' : ''"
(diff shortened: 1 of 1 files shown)


karrui:
> The code had the comment: "Empty space is to prevent visual bug when radio button is empty".

This can still happen during the field preview when an admin is creating a new radio option. Can you check if that breaks anything in the field preview?

An alternative is to only add the `nbsp;` character if the field value is an empty string.

timotheeg:
> An alternative is to only add the nbsp; character if the field value is an empty string.

Yup 👍 The current change in the PR does that already: if the field value is falsy (including empty string), then `&nbsp;` is used. It doesn't need to be added (since there's nothing to add to), rather, it's used in place of the input.

I think I tested locally, but I'll do it one more time and send a confirmation message.

karrui (approved):

timotheeg:
Tested again locally and preview confirmed to work 🎉 .

Radios get checked when there's one or multiple blank entries (I assume that's part of the error indicator), but rendering is correct 👍 
image
image

The forms shows correctly as well:
image

I'll merge now.

PR #2761, opened 2021-09-08
feat(design-system): add avatar component and stories

### LINKED ISSUE #2014: [React] Add Avatar component
**Time estimation: 2d**

**"Sub" component needed**
- [ ] Avatar
- [ ] AvatarMenu (for dropdown). <-- may want to wait for Dropdown field to be completed to share the theming. Depends on which is done first, see #1999

tshuli:
**Features**:

- add avatar component and stories. closes #2014
- note that as clarified with design, avatar does not have its own hover / active / focus styles, as it is wrapped in clear button

## Before & After Screenshots
In storybook

### DIFF
--- a/frontend/src/components/Avatar/Avatar.stories.tsx
+++ b/frontend/src/components/Avatar/Avatar.stories.tsx
@@ -10,13 +10,21 @@ export default {
 type AvatarTemplateProps = {
   hasNotification?: boolean
   name?: string
+  showBorder?: boolean
 }
 
 const AvatarTemplate: Story<AvatarTemplateProps> = ({
   hasNotification,
   name,
+  showBorder,
 }) => {
-  return <Avatar hasNotification={hasNotification} name={name}></Avatar>
+  return (
+    <Avatar
+      hasNotification={hasNotification}
+      name={name}
+      showBorder={showBorder}
+    ></Avatar>
+  )
 }
 
 export const Default = AvatarTemplate.bind({})
@@ -24,3 +32,6 @@ Default.args = { hasNotification: false, name: 'ABC' }
 
 export const WithNotification = AvatarTemplate.bind({})
 WithNotification.args = { hasNotification: true, name: 'ABC' }
+
+export const WithBorder = AvatarTemplate.bind({})
+WithBorder.args = { showBorder: true, name: 'ABC' }
--- a/frontend/src/components/Avatar/Avatar.stories.tsx
+++ b/frontend/src/components/Avatar/Avatar.stories.tsx
@@ -1,0 +1,26 @@
+import { Meta, Story } from '@storybook/react'
+
+import { Avatar } from './Avatar'
+
+export default {
+  title: 'Components/Avatar',
+  component: Avatar,
+} as Meta
+
+type AvatarTemplateProps = {
+  hasNotification?: boolean
+  name?: string
+}
+
+const AvatarTemplate: Story<AvatarTemplateProps> = ({
+  hasNotification,
+  name,
+}) => {
+  return <Avatar hasNotification={hasNotification} name={name}></Avatar>
+}
+
+export const Default = AvatarTemplate.bind({})
+Default.args = { hasNotification: false, name: 'ABC' }
+
+export const WithNotification = AvatarTemplate.bind({})
+WithNotification.args = { hasNotification: true, name: 'ABC' }
--- a/frontend/src/components/Avatar/Avatar.tsx
+++ b/frontend/src/components/Avatar/Avatar.tsx
@@ -1,0 +1,31 @@
+import {
+  Avatar as ChakraAvatar,
+  AvatarBadge as ChakraAvatarBadge,
+  AvatarProps as ChakraAvatarProps,
+} from '@chakra-ui/react'
+
+export interface AvatarProps extends ChakraAvatarProps {
+  hasNotification?: boolean
+}
+
+/*
+ * Avatar component
+ * May be wrapped in AvatarMenuButton
+ */
+export const Avatar = ({
+  hasNotification,
+  ...rest
+}: AvatarProps): JSX.Element => {
+  return (
+    <ChakraAvatar
+      getInitials={
+        // Extract first letter of avatar name.
+        // Default method extracts first two letters.
+        rest.getInitials ? rest.getInitials : (fullName) => fullName[0]
+      }
+      {...rest}
+    >
+      {hasNotification ? <ChakraAvatarBadge /> : null}
+    </ChakraAvatar>
+  )
+}
--- a/frontend/src/components/Avatar/Avatar.tsx
+++ b/frontend/src/components/Avatar/Avatar.tsx
@@ -31,6 +31,9 @@ export const Avatar = ({
           }
         : {})}
       {...rest}
+      aria-label={
+        hasNotification ? 'Avatar with unread notifications' : 'Avatar '
+      }
     >
       {hasNotification ? <ChakraAvatarBadge /> : null}
     </ChakraAvatar>
--- a/frontend/src/components/Avatar/Avatar.tsx
+++ b/frontend/src/components/Avatar/Avatar.tsx
@@ -6,6 +6,7 @@ import {
 
 export interface AvatarProps extends ChakraAvatarProps {
   hasNotification?: boolean
+  showBorder?: boolean
 }
 
 /*
@@ -14,6 +15,7 @@ export interface AvatarProps extends ChakraAvatarProps {
  */
 export const Avatar = ({
   hasNotification,
+  showBorder,
   ...rest
 }: AvatarProps): JSX.Element => {
   return (
@@ -23,6 +25,11 @@ export const Avatar = ({
         // Default method extracts first two letters.
         rest.getInitials ? rest.getInitials : (fullName) => fullName[0]
       }
+      {...(showBorder
+        ? {
+            boxShadow: `0 0 0 4px var(--chakra-colors-primary-300)`,
+          }
+        : {})}
       {...rest}
     >
       {hasNotification ? <ChakraAvatarBadge /> : null}
--- a/frontend/src/templates/AvatarMenu/AvatarMenu.stories.tsx
+++ b/frontend/src/templates/AvatarMenu/AvatarMenu.stories.tsx
@@ -1,0 +1,76 @@
+import { MenuDivider } from '@chakra-ui/react'
+import { Meta, Story } from '@storybook/react'
+
+import { Avatar } from '../../components/Avatar/Avatar'
+import Menu from '../../components/Menu'
+
+import { AvatarMenuProps } from './AvatarMenu'
+import AvatarMenu from '.'
+
+export default {
+  title: 'Templates/AvatarMenu',
+  component: Avatar,
+} as Meta
+
+type AvatarTemplateProps = AvatarMenuProps
+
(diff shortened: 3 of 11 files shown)


frontend/src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type AvatarTemplateProps = {
+  fullName?: string
+  userName?: string
+  hasNotification?: boolean
+  isOpen?: boolean
+}
+
+const AvatarTemplate: Story<AvatarTemplateProps> = ({
```
karrui, line 26: I don't think we need this for the Avatar story. The base avatar story should only be the Avatar component itself instead of all the bells and whistles

frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export interface AvatarProps extends ChakraAvatarProps {
+  hasNotification?: boolean
+}
+
+/*
+ * Utility to extract first letter of avatar name
+ * Override default getInitials Avatar method
+ */
+const getInitials = (fullName: string) => fullName[0]
```
karrui, line 24: nit: should this be inlined into the Avatar component (and only used if props.getInitials are not provided?

frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+/*
+ * Avatar component
+ * May be wrapped in AvatarMenuButton
```
karrui, line 28: I think avatar being in an dropdown menu is out of scope of this implementation. We can have a template component that uses Avatar, dropdown menu button, etc to compose a new AvatarMenuButton instead of having it be in this file

frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+ */
+export const Avatar = (props: AvatarProps): JSX.Element => {
+  const avatarBadge = props.hasNotification ? <ChakraAvatarBadge /> : ''
+
+  return (
+    <ChakraAvatar getInitials={getInitials} {...props}>
```
karrui, line 34: Would there not be a warning from React that some unknown `hasNotification` prop is being passed into the dom? We should destructure if we do not want to pass it into ChakraAvatar:

```tsx
const Avatar = ({hasNotification, ...rest}) => {
...
}
```

frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+  )
+}
+
+/**
+ * DropdownMenuButton styled for avatar
+ * Used to wrap Avatar component
+ * @preconditions Must be a child of DropdownMenu component,
+ * and returned using a render prop (see implementation in Avatar.stories).
+ */
+const AvatarMenuButton = (props: ButtonProps): JSX.Element => {
```
karrui, line 46: as above, I don't think this belongs in the Avatar design component, but maybe we can create a template that uses Avatar component, with separate stories

frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+      {...props}
+    ></DropdownMenu.Button>
+  )
+}
+
+/**
+ * DropdownMenuItem styled for avatar username
+ * @preconditions Must be a child of DropdownMenu component,
+ */
+const AvatarUsername = (props: MenuItemProps): JSX.Element => {
```
karrui, line 62: This name is slightly confusing, because it is actually part of the dropdown and not the avatar itself. As such, I think it would make more sense to move these related components into an AvatarDropdown component separate from the Avatar component

karrui (changes requested):

frontend/src/components/Avatar/Avatar.stories.tsx
tshuli: edited

frontend/src/components/Avatar/Avatar.tsx
tshuli: edited

frontend/src/components/Avatar/Avatar.tsx
tshuli: thanks, shifted to template, see if that works

frontend/src/components/Avatar/Avatar.tsx
tshuli: thanks, excluded hasNotification

frontend/src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type AvatarTemplateProps = {
+  hasNotification?: boolean
+}
+
+const AvatarTemplate: Story<AvatarTemplateProps> = ({ hasNotification }) => {
+  return <Avatar hasNotification={hasNotification}></Avatar>
+}
+
+const AvatarGroupTemplate: Story = () => {
```
karrui, line 19: should we split the story into two separate stories? Makes it easier to see what are the available props

WithNotification
WithoutNotification

frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const Avatar = ({
+  hasNotification,
+  ...rest
+}: AvatarProps): JSX.Element => {
+  const avatarBadge = hasNotification ? <ChakraAvatarBadge /> : ''
+
+  return (
+    <ChakraAvatar
+      getInitials={
+        rest.getInitials ? rest.getInitials : (fullName) => fullName[0] // Extract first letter of avatar name
```
karrui, line 24: I think this is done by default? I don't think we need to override this behaviour

frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,31 @@
...
+      }
+      {...rest}
+    >
+      {avatarBadge}
```
karrui, line 28: nit: should we inline this directly so it's easier to read? also prefer `null` to `''`

```tsx
<ChakraAvatar>
  {hasNotification ? <ChakraAvatarBadge /> : null}
</ChakraAvatar>
```

frontend/src/templates/AvatarMenu/AvatarMenu.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+}) => {
+  return (
+    <SimpleGrid
+      columns={2}
+      spacing={'200px'}
+      alignItems="center"
+      templateColumns="min-content min-content"
+    >
+      <Box>
+        Closed
```
karrui, line 67: prefer individual stories rather than grouping up the states so the controls can be used to toggle state and better regression visibility. When the `AvatarMenu/Default` story fails regression testing, it is not obvious at first glance which state is failing.

suggest

"Open Menu" and "Closed Menu" individual stories

frontend/src/templates/AvatarMenu/AvatarMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const AvatarMenuButton = (props: ButtonProps): JSX.Element => {
+  return (
+    <Menu.Button
+      variant="clear"
+      px="0"
+      iconSpacing="0.5rem"
+      color="secondary.300"
+      {...props}
+    ></Menu.Button>
```
karrui, line 26: nit: use `/>` shorthand rather than repeating the closing tag.

```tsx
<Menu.Button
  variant="clear"
  ...
/>
```

frontend/src/templates/AvatarMenu/AvatarMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * MenuItem styled for avatar username
+ * @preconditions Must be a child of Menu component,
+ */
+export const AvatarUsername = (props: MenuItemProps): JSX.Element => {
+  const styles = useMultiStyleConfig('AvatarMenu', {})
+
+  const userIcon = <Icon as={BxsUser} sx={styles.usernameIcon}></Icon>
```
karrui, line 37: nit: use `/>` shorthand rather than repeating the closing tag.

frontend/src/theme/components/Avatar.ts
```diff
@@ -0,0 +1,25 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/react'
+
+export const Avatar: ComponentMultiStyleConfig = {
+  parts: ['container', 'badge', 'usernameItem', 'usernameIcon'],
+  baseStyle: {
+    container: {
+      backgroundColor: 'primary.500',
+      fontSize: '0.875rem',
+      fontWeight: '500',
```
karrui, line 9: This seems incorrect. I think we should use `textStyle: 'subhead-2'` (according to figma design) instead of redeclaring the text style here so the design changes if we update subhead-2. If the textStyle does not hold, we can import textStyle and spread its props directly in this object.

frontend/src/templates/AvatarMenu/AvatarMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { BxsUser } from '~/assets/icons/BxsUser'
+import Menu from '~/components/Menu'
+
+/**
+ * MenuButton styled for avatar
+ * Used to wrap Avatar component
+ * @preconditions Must be a child of Menu component,
+ * and returned using a render prop (see implementation in AvatarMenu.stories).
+ */
+export const AvatarMenuButton = (props: ButtonProps): JSX.Element => {
```
karrui, line 18: since this file is an AvatarMenu, should we make this a subcomponent of the main AvatarMenu component (which is absent)?

There seems to warrant an AvatarMenu component which is basically the entire AvatarTemplate story in AvatarMenu.stories.tsx? 

Basically rule of thumb: the stories themselves should contain a "Default" story that is literally 

```tsx
const Template: Story = (args) => <Component {...args} />
```
which is literally rendering the component. If it requires more than that, the story is used incorrectly.

karrui (changes requested):

frontend/src/templates/AvatarMenu/AvatarMenu.tsx
tshuli: thanks for advice! I extracted the AvatarMenu out, see if ok

frontend/src/theme/components/Avatar.ts
tshuli: thanks, edited

frontend/src/templates/AvatarMenu/AvatarMenu.stories.tsx
tshuli: thanks! edited

frontend/src/components/Avatar/Avatar.tsx
tshuli: thanks! edited

frontend/src/components/Avatar/Avatar.tsx
tshuli: Need, because default is to extract two characters

tshuli: Added docs for clarity

frontend/src/components/Avatar/Avatar.stories.tsx
tshuli: thanks! edited

karrui (approved):
lgtm, now await @pearlyong's design review :D

PR #2764, opened 2021-09-09
feat(react): allow markdown usage, other useToast functions in toasts

karrui:
## Problem

Some toasts require something like links to be appended. This PR adds the capability to Toasts, much like on Banners and InlineMessage components

ChakraUI's default useToast hook also has quite a few toast-related functions, which should also be exposed by our custom hook. This PR also adds the relevant hook functions into our custom hooks.

A new story has been added for markdown usage in toasts

### DIFF
--- a/frontend/src/components/Toast/Toast.tsx
+++ b/frontend/src/components/Toast/Toast.tsx
@@ -15,7 +15,7 @@ import {
 import { BxsCheckCircle, BxsErrorCircle } from '~assets/icons'
 import { useMdComponents } from '~hooks/useMdComponents'
 
-type ToastStatus = 'danger' | 'success' | 'warning'
+export type ToastStatus = 'danger' | 'success' | 'warning'
 
 export interface ToastProps
   extends Omit<
--- a/frontend/src/components/Toast/Toast.tsx
+++ b/frontend/src/components/Toast/Toast.tsx
@@ -1,5 +1,6 @@
-import React from 'react'
+import { useMemo } from 'react'
 import { BiX } from 'react-icons/bi'
+import ReactMarkdown from 'react-markdown'
 import {
   Alert,
   AlertDescription,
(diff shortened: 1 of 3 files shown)


frontend/src/components/Toast/Toast.tsx
```diff
@@ -44,13 +52,38 @@ export const Toast = ({
...
+  const mdComponents = useMdComponents(styles)
+
+  const descriptionComponent = useMemo(() => {
+    // ReactMarkdown requires children to be of string type. If description is
+    // not a string, return description as is.
+    if (!description || !useMarkdown || typeof description !== 'string')
+      return description
+    return (
+      <ReactMarkdown components={mdComponents}>{description}</ReactMarkdown>
+    )
```
mantariksh, line 69: wondering if we should extract this into its own component?

frontend/src/hooks/useToast.ts
```diff
@@ -1,41 +1,71 @@
...
 
 // onClose is provided by the chakra hook and should not be exposed to clients
-export type UseToastProps = Omit<ToastBehaviourProps & ToastProps, 'onClose'>
+export type UseToastProps = Omit<
+  ToastBehaviourProps & ToastProps,
+  'onClose' | 'status'
+> & { status?: ToastStatus }
 
-type UseToast = (
-  // NOTE: Chakra's toast status is different from ours, hence we need to omit it for type compatibility
```
mantariksh, line 19: should we retain this comment for `UseToastProps`?

frontend/src/components/Toast/Toast.tsx
karrui: I think its fine, since mdComponents has the possibility to be unique with its given styles

frontend/src/hooks/useToast.ts
karrui: resolved in 92281db

tshuli (approved):
lgtm

PR #2775, opened 2021-09-10
fix(otp): Ensure uniform distribution of digits

### LINKED ISSUE #2768: Validate OTP digit distribution (and fix if needed)
We've been notified that digit distribution for OTP generation might not be entirely uniform, due to the randomization, scaling,  and clipping system used.

While this is not an urgent issue, considering we have a rate limit on OTP submissions, we should still verify whether our algorithm does match our design intention of uniform distribution, and if it doesn't, we should fix it.

This is an interesting math exercise, where we can easily generate a huge amount of samples and verify the distribution empirically. One of FormSG engineer could take the task on a learning Friday and share his/her findings with the rest of the team.

See FormSG working document for 2021-09-09 for details.

timotheeg:
## Problem
We have received a report that our OTP generation method does not have a uniform distribution of digits. While is not a serious security issue, it is clearly not an intended behaviour.

Closes #2768 (issue contains more details description of the distribution problem)

## Solution
Use the native API [`crypto.randomInt()`](https://nodejs.org/api/crypto.html#crypto_crypto_randomint_min_max_callback), which was introduced in node 14.10 / 12.19, to avoid quantisation issues.

This is **not** a breaking change and is fully backward compatible

**Bug Fixes**
Uniform distribution brings back the guarantee that a random OTP guess only has 1/10e6 of being right.

## Tests
- Log in: Verify OTP flow is working

### DIFF
--- a/src/app/utils/otp.ts
+++ b/src/app/utils/otp.ts
@@ -10,15 +10,11 @@ const DEFAULT_SALT_ROUNDS = 10
  * @returns 6 digit OTP string
  */
 export const generateOtp = (): string => {
-  const length = 6
-  const chars = '0123456789'
   // Generates cryptographically strong pseudo-random data.
-  // The size argument is a number indicating the number of bytes to generate.
-  const digits = new Array(length)
-  for (let i = 0; i < length; i++) {
-    digits[i] = chars[crypto.randomInt(0, chars.length)]
-  }
-  return digits.join('')
+  return Array(6)
+    .fill(0)
+    .map(() => crypto.randomInt(0, 10))
+    .join('')
 }
 
 /**
--- a/src/app/utils/otp.ts
+++ b/src/app/utils/otp.ts
@@ -14,11 +14,9 @@ export const generateOtp = (): string => {
   const chars = '0123456789'
   // Generates cryptographically strong pseudo-random data.
   // The size argument is a number indicating the number of bytes to generate.
-  const rnd = crypto.randomBytes(length)
-  const d = chars.length / 256
   const digits = new Array(length)
   for (let i = 0; i < length; i++) {
-    digits[i] = chars[Math.floor(rnd[i] * d)]
+    digits[i] = chars[crypto.randomInt(0, chars.length)]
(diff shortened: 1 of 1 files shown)


liangyuanruo (approved):
Thanks, found the PR from the issue. This is kinda nitpick, but I think we can lose the `chars` and `length` variables by simply doing this:

```typescript
return [...new Array(6)]
  .map(() => crypto.randomInt(0,10))
  .join('')
```

Unfortunately, the array destructuring is necessary because even though `new Array(6)` is an array of length 6, `map()` only works if explicit values have been been assigned.

timotheeg:
Aye, my first version was like this:
```javascript
return Array(6)
  .fill(0)
  .map(() => crypto.randomInt(0,10))
  .join('')
```

But I decided against it to make it super obvious what little actually changed, and the fill(0) was annoying me a bit, same as the restructuring you have I guess :)

But yeah, if we're OK with a bit of extra functional processing for readability, I'll update the PR.

timotheeg:
Also, I liked the idea of separating the alphabet from the picking, in case we wanted to use [Crockford's base32](https://en.wikipedia.org/wiki/Base32#Crockford.27s_Base32) or a variant thereof. 

But I suppose, we can revisit this later if we want. As long as we just use digits, we can indeed simplify the function with just `randomInt()` 👍

timotheeg:
Updated! @liangyuanruo

timotheeg:
It's strange: your prior approval still seem to apply and I can merge now, even though I pushed new code 🤔  I thought pushes would dismiss approvals. I'll double check the project settings later.

liangyuanruo:
> But yeah, if we're OK with a bit of extra functional processing for readability, I'll update the PR.

Yes I think this is very much to be encouraged; there is already an emphasis on strongly-typed code, and the team has been encouraged to write functional, stateless code as much as possible during the backend TypeScript refactor. This shows up in the use of _neverthrow_ and _fp-ts_, and I think is a major reason why production incidents have been fairly low.

PR #2784, opened 2021-09-13
fix(v2): correctly copy html files to /dist build folder

karrui:
## Problem

`tsc` literally only touches `.ts` and `.js` files, and there is no way to make it process other extensions. This means other files that the server depends on such as the `.html` server view files are not moved to the `/dist` build folder.

Since our Docker build file had been updated to only keep the necessary files (in #2675) (only /dist folder is copied over), this means the view files will not be accessible by the express server (since they are not available in the Docker image).

That results in errors like 

```
error.message | ENOENT: no such file or directory, open '/opt/formsg/src/app/views/templates/otp-email.server.view.html'
```

This PR updates the build pipeline to copy any HTML files in the backend source folder into the build folder using the `copyfiles` tool.  This is achieved by adding/updating 4 scripts in root's pacakge.json:

```diff
+"build": "npm run clean && npm run build:backend && npm run build:frontend",
+"build:backend": "tsc -p tsconfig.build.json && npm run copyfiles:backend",
+"copyfiles:backend": "copyfiles -e src/public/**/*.html -u 1 src/**/*.html dist/backend/src",
+"clean": "rimraf dist/",
```

New stuff include 
- clearing the entire `/dist` directory before running `npm run build`,
- running `npm run copyfiles:backend` script after tsc in `npm run build:backend`,
- `copyfiles:backend` script to copy all html files in backend src directory to its appropriate dist/backend/src directory,
- `clean` script to clear dist folder (if needed)

## Deploy Notes

**New dev dependencies**:

- [rimraf](https://www.npmjs.com/package/rimraf): `rm -rf` for node
- [copyfiles](https://www.npmjs.com/package/copyfiles): `cp` for node

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8956,6 +8956,134 @@
         }
       }
     },
+    "copyfiles": {
+      "version": "2.4.1",
+      "resolved": "https://registry.npmjs.org/copyfiles/-/copyfiles-2.4.1.tgz",
+      "integrity": "sha512-fereAvAvxDrQDOXybk3Qu3dPbOoKoysFMWtkY3mv5BsL8//OSZVL5DCLYqgRfY5cWirgRzlC+WSrxp6Bo3eNZg==",
+      "dev": true,
+      "requires": {
+        "glob": "^7.0.5",
+        "minimatch": "^3.0.3",
+        "mkdirp": "^1.0.4",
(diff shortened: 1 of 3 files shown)


tshuli (approved):
lgtm

PR #2785, opened 2021-09-13
refactor: use modern JS constructs to validate NRIC

timotheeg:
## Context

While following code from the NIRC component, I ended up on the nirc validation where I thought several minor  improvements could be made. Below are some (very minor) issues I thought about when seeing the code
* The regex is using unnecessary constructs like `{1}`
* Regexp specifies both lowercase and uppercase matches, while it would be simpler to uppercase the input prior to running the test
* The regexp captures groups, but these are not used because the format validation and checksum validation are disjointed
* The digits string is exploded to iterate over the characters, but that is unnecessary since the coefficients is an array of the same format that could be used instead
* We can us the string array operator to read digits from the input, and to read checksum markers (i.e. there's no need to create local arrays.)
* the `reduce()` function does an unnecessary assignment to a local variable, which looks confusing because of the shadow variable name `sum`.

The code was working perfectly fine, so the refactor is unnecessary. I initially only wanted to remove the`{1}` and add a `i` flag to the regex, but I ended up refactoring the whole file 😑 

I hesitated to open the PR since it is unplanned and unnecessary, and will consume precious reviewer time, but in the end, since I did the changes, I thought I'd open it up anyway for some discussions, like:
* does using for named groups in regexes affect our browser support matrix?
* does using array operator for string character access affect our browser support matrix? (do we have precedent in code?)

## Approach
I've made changes that addresses all the points above. The code reduces significantly. I am obviously really biased, but I think it is more readable now.

## Testing
- [ ] Create a form with a NIRC component
- [ ] Verify validation works as expected in major browsers (Chrome, FF, safari, Edge)

### DIFF
--- a/shared/utils/nric-validation.ts
+++ b/shared/utils/nric-validation.ts
@@ -18,16 +18,16 @@ export const isNricValid = (value: string): boolean => {
 
   const { prefix, digits, checksum } = parsed.groups as NricParts
 
-  // http://www.ngiam.net/NRIC/NRIC_numbers.pdf
-  const coefficients = [2, 7, 6, 5, 4, 3, 2]
+  // Specifications at: http://www.ngiam.net/NRIC/NRIC_numbers.pdf
+  const weights = [2, 7, 6, 5, 4, 3, 2]
   const startConstant = prefix === 'S' || prefix === 'F' ? 0 : 4
   const checksumEncoding =
     prefix === 'S' || prefix === 'T' ? 'JZIHGFEDCBA' : 'XWUTRQPNMLK'
 
-  const sum = coefficients.reduce(
-    (acc, coef, idx) => acc + coef * parseInt(digits[idx]),
+  const weightedSum = weights.reduce(
+    (acc, weight, idx) => acc + weight * parseInt(digits[idx]),
     startConstant,
   )
 
-  return checksum === checksumEncoding[sum % 11]
+  return checksum === checksumEncoding[weightedSum % 11]
 }
--- a/shared/utils/nric-validation.ts
+++ b/shared/utils/nric-validation.ts
@@ -1,41 +1,33 @@
+type NricParts = {
+  prefix: string
+  digits: string
+  checksum: string
+}
+
+const NIRC_FORMAT = /^(?<prefix>[STFG])(?<digits>\d{7})(?<checksum>[A-Z])$/
+
 /**
  * Validates whether a provided string value adheres to the UIN/FIN format
  * as provided on the Singapore Government's National Registration Identity Card.
  * @param value The value to be validated
  */
 export const isNricValid = (value: string): boolean => {
-  return isFormatValid(value) && isChecksumValid(value)
-}
+  const parsed = value?.toUpperCase().match(NIRC_FORMAT)
 
-/**
- * Tests whether a provided string value obeys a simple format check
- * @param value The value to be validated
- */
-const isFormatValid = (value: string): boolean => {
-  return /^([STFGstfg]{1})([0-9]{7})([A-Za-z]{1})$/.test(value)
-}
+  if (!parsed) return false
 
-/**
- * Algorithm to test whether the NRIC checksum is valid
- * @param value The value to be validated
- */
-const isChecksumValid = (value: string): boolean => {
+  const { prefix, digits, checksum } = parsed.groups as NricParts
+
   // http://www.ngiam.net/NRIC/NRIC_numbers.pdf
-  value = value.toUpperCase()
-  const prefix = value.charAt(0)
-  const suffix = value.charAt(value.length - 1)
   const coefficients = [2, 7, 6, 5, 4, 3, 2]
-  const constant = prefix === 'S' || prefix === 'F' ? 0 : 4
-  const coding =
-    prefix === 'S' || prefix === 'T'
-      ? ['J', 'Z', 'I', 'H', 'G', 'F', 'E', 'D', 'C', 'B', 'A']
-      : ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K']
-  const sum = value
-    .substring(1, value.length - 1)
-    .split('')
-    .reduce(function (sum, digit, idx) {
-      sum += parseInt(digit) * coefficients[idx]
-      return sum
-    }, constant)
-  return suffix === coding[sum % 11]
+  const start_constant = prefix === 'S' || prefix === 'F' ? 0 : 4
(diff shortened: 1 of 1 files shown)


shared/utils/nric-validation.ts
```diff
@@ -1,41 +1,33 @@
+type NricParts = {
+  prefix: string
+  digits: string
+  checksum: string
+}
+
+const NIRC_FORMAT = /^(?<prefix>[STFG])(?<digits>\d{7})(?<checksum>[A-Z])$/
```
mantariksh, line 7:
```suggestion
const NRIC_FORMAT = /^(?<prefix>[STFG])(?<digits>\d{7})(?<checksum>[A-Z])$/
```

shared/utils/nric-validation.ts
```diff
@@ -1,41 +1,33 @@
...
+
 /**
  * Validates whether a provided string value adheres to the UIN/FIN format
  * as provided on the Singapore Government's National Registration Identity Card.
  * @param value The value to be validated
  */
 export const isNricValid = (value: string): boolean => {
-  return isFormatValid(value) && isChecksumValid(value)
-}
+  const parsed = value?.toUpperCase().match(NIRC_FORMAT)
```
mantariksh, line 15:
```suggestion
  const parsed = value?.toUpperCase().match(NRIC_FORMAT)
```

shared/utils/nric-validation.ts
```diff
@@ -1,41 +1,33 @@
...
-      : ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K']
-  const sum = value
-    .substring(1, value.length - 1)
-    .split('')
-    .reduce(function (sum, digit, idx) {
-      sum += parseInt(digit) * coefficients[idx]
-      return sum
-    }, constant)
-  return suffix === coding[sum % 11]
+  const start_constant = prefix === 'S' || prefix === 'F' ? 0 : 4
```
mantariksh, line 23:
```suggestion
  const startConstant = prefix === 'S' || prefix === 'F' ? 0 : 4
```

shared/utils/nric-validation.ts
```diff
@@ -1,41 +1,33 @@
...
+  const checksum_encoding =
```
mantariksh, line 24:
```suggestion
  const checksumEncoding =
```

shared/utils/nric-validation.ts
```diff
@@ -1,41 +1,33 @@
...
+    prefix === 'S' || prefix === 'T' ? 'JZIHGFEDCBA' : 'XWUTRQPNMLK'
+
+  const sum = coefficients.reduce(
+    (acc, coef, idx) => acc + coef * parseInt(digits[idx]),
+    start_constant,
```
mantariksh, line 29:
```suggestion
    startConstant,
```

shared/utils/nric-validation.ts
```diff
@@ -1,41 +1,33 @@
...
+  )
+
+  return checksum === checksum_encoding[sum % 11]
```
mantariksh, line 32:
```suggestion
  return checksum === checksumEncoding[sum % 11]
```

mantariksh (commented):
a few typo/casing suggestions first, will review the logic soon!

shared/utils/nric-validation.ts
timotheeg: OMG code dislexia 😭 Thanks for the correction!

shared/utils/nric-validation.ts
```diff
@@ -1,41 +1,33 @@
...
-    .reduce(function (sum, digit, idx) {
-      sum += parseInt(digit) * coefficients[idx]
-      return sum
-    }, constant)
-  return suffix === coding[sum % 11]
+  const startConstant = prefix === 'S' || prefix === 'F' ? 0 : 4
+  const checksumEncoding =
+    prefix === 'S' || prefix === 'T' ? 'JZIHGFEDCBA' : 'XWUTRQPNMLK'
+
+  const sum = coefficients.reduce(
```
tshuli, line 27: Could I suggest adding a comment so its clear to future devs what this line does?

```suggestion
  // Multiply coefficients and digits element-wise and then sum the result 
  const sum = coefficients.reduce(
```

tshuli (approved):
lgtm!

shared/utils/nric-validation.ts
timotheeg: In this case, I'm thinking the code is very short using simple well-understood operations, so I reckon it's rather easy to understand on its own, and a comment that duplicates what the code does is probably unnecessary.

That said, the code itself maybe could be made a little bit more self-explanatory 🤔 . Renaming `coefficients` into `weights`, and `sum` into `weightedSum` might make clearer that it's just that: simple code to compute a weighted sum 😅 . I'll add a commit to do that!

timotheeg: Added commit to rename variables! I think the code reads pretty clearly now 👍 😃

timotheeg:
On my 2 questions from the PR description:

> does using for named groups in regexes affect our browser support matrix?

[Support matrix for named character groups](https://caniuse.com/mdn-javascript_builtins_regexp_named_capture_groups)
image


> does using array operator for string character access affect our browser support matrix? (do we have precedent in code?)

String array-like character accessed was introduced in EcmaScript 5 (see doc [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#character_access)).

which has [this support matrix](https://caniuse.com/?search=ecmascript%205):
image

I guess the biggest question I have now is whether we support IE11? 

I just checked the transpiled (tsc+babel) result to check if by chance the code would be transformed to use older constructs like `.charAt()` to be more backward compatible, but it doesn't look like it does:
```javascript
function(e,t,n){"use strict";Object.defineProperty(t,"__esModule",{value:!0}),t.isNricValid=void 0;var r=/^(?<prefix>[STFG])(?<digits>\d{7})(?<checksum>[A-Z])$/;t.isNricValid=function(e){var t=null==e?void 0:e.toUpperCase().match(r);if(!t)return!1;var n=t.groups,i=n.prefix,o=n.digits;return n.checksum===("S"===i||"T"===i?"JZIHGFEDCBA":"XWUTRQPNMLK")[[2,7,6,5,4,3,2].reduce((function(e,t,n){return e+t*parseInt(o[n])}),"S"===i||"F"===i?0:4)%11]}}
```

timotheeg:
Confirmed with Sonjia we do not support IE11, so I'm just going to merge this now.

PR #2793, opened 2021-09-13
fix(v2): correctly assert that public route location shape may be undefined

karrui:
## Problem

The application accesses the need to redirect from a public route to a private route if the user is already authenticated. The link to redirect to is done by checking the URL that could possibly be stored in `location.state.from`. However, `location.state` could be undefined if the page had not been redirected from the application, resulting in an error boundary being hit.

This PR correctly types the `location.state` variable as possibly undefined.


tshuli (approved):
lgtm

PR #2794, opened 2021-09-13
feat(react): allow markdown in FormLabel description, allow `as` prop in custom Link component

karrui:
## Problem

Some designs in Figma contains Form label descriptions with links. This seemed easier to achieve in markdown and thus the ability for markdown in FormLabel.Description is implemented.

See 
Screenshot 2021-09-13 at 10 50 34 PM

Whilst doing so, realize the `Link` component cannot be re-morphed into another type using the `as` prop which is a common in ChakraUI.  This commit was cherry picked from commit 22faa66 in #2683.

## Solution

**Features**:

- Add `useMarkdown` prop in FormLabel.Description and `useMarkdownForDescription` prop in the main FormLabel component to allow markdown in form label descriptions.
- Allow `as` prop in custom Link component

## Before & After Screenshots
New story added in Storybook for markdown in form label description.

### DIFF
--- a/frontend/src/components/FormControl/FormLabel/FormLabel.tsx
+++ b/frontend/src/components/FormControl/FormLabel/FormLabel.tsx
@@ -1,4 +1,5 @@
 import { useMemo } from 'react'
+import ReactMarkdown from 'react-markdown'
 import {
   Box,
   FormHelperText,
@@ -13,6 +14,7 @@ import {
 } from '@chakra-ui/react'
 
 import { BxsHelpCircle } from '~assets/icons/BxsHelpCircle'
+import { useMdComponents } from '~hooks/useMdComponents'
 
 export interface FormLabelProps extends ChakraFormLabelProps {
   /**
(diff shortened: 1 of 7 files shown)


frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -52,6 +59,7 @@ export const FormLabel = ({
   tooltipText,
   questionNumber,
   description,
+  useMarkdownForDescription,
```
mantariksh, line 62: should this have a default?

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -73,18 +81,25 @@ export const FormLabel = ({
...
 
 FormLabel.Label = ChakraFormLabel
 
+interface FormLabelDescriptionProps extends TextProps {
+  useMarkdown?: boolean
+  children: string
+}
 const FormLabelDescription = ({
   children,
+  useMarkdown,
```
mantariksh, line 100: should this have a default?

mantariksh (approved):

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
karrui: add default of `false`, which is the same effect but more explicit

PR #2806, opened 2021-09-14
feat(design-system): add vertical line variant to Tabs

karrui:
## Problem

This PR implements the figma component known as Side Navigation, used in Settings page.

## Solution

**Features**:

- feat(theme): add styling for vertical tab component
- feat: add new story for vertical line variant

**Improvements**:

- style(Tabs): styling of tabs to only apply to overridden variants
- feat(theme): add defined sizing only to custom tab variants
- refactor(theme): update variant line-light theme function names

## Before & After Screenshots
Add new vertical line story to Storybook

### DIFF
--- a/frontend/src/components/Tabs/Tabs.stories.tsx
+++ b/frontend/src/components/Tabs/Tabs.stories.tsx
@@ -21,7 +21,7 @@ export default {
 } as Meta
 
 const TabTemplate: Story<TabsProps> = (args) => {
-  const contentFontColour = args.variant === 'line-light' ? 'black' : 'white'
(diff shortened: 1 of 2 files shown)


tshuli:
Just to check, the tab width changes in active state resulting in the page shifting around - is this intended behavior?
recording

karrui:
> Just to check, the tab width changes in active state resulting in the page shifting around - is this intended behavior?

Yeah, the caller can add a min width attribute if needed; I don't think it's the responsibility of this component to do so

tshuli (approved):
thanks! lgtm otherwise

PR #2819, opened 2021-09-19
refactor: use combined route validators and handlers

### LINKED ISSUE #1642: Combine validators with handler functions and export them atomically
## Problem 
Currently, in our `routes.ts`, we have the validator set up within the route itself and the handler imported from the controller. For some handlers, the validation might be a precondition in that the handler _requires_ that the incoming data be valid. 

Hence, the best practice would be to combine the handling function together with the validator in the controller itself and export them together as an array such as `export const handleSomething = [validateSomething, doSomething]`. This would prevent human errors where there are no validations due to forgetting them.

## Notes
The validation should be defined in the same file (preferably close to the handler). It is ok to redefine it because validation differs on a case by case basis.

tehtea:
## Problem
Prevent developers from forgetting the use of validators when using a controller handler.

Closes #1642 

## Solution
As suggested in the issue, put the original handler with a validation middleware in an array, and expose the final array for usage rather than the original handler

**Breaking Changes** 

    - Details ...

**Improvements**:

- Details ...

## Tests
Ran a regression test using `npm run test-backend`, and ran happy-path testing locally with the changed routes

### DIFF
--- a/src/app/modules/user/user.middleware.ts
+++ b/src/app/modules/user/user.middleware.ts
@@ -1,0 +1,14 @@
+import { celebrate, Joi } from 'celebrate'
+
+/**
+ * Celebrate validation for the contact OTP verification endpoint.
+ */
+export const validateContactOtpVerificationParams = celebrate({
+  body: Joi.object({
+    userId: Joi.string().required(),
+    otp: Joi.string()
+      .required()
+      .regex(/^\d{6}$/),
+    contact: Joi.string().required(),
+  }),
+})
--- a/src/app/modules/user/user.middleware.ts
+++ b/src/app/modules/user/user.middleware.ts
@@ -1,6 +1,16 @@
 import { celebrate, Joi, Segments } from 'celebrate'
 
 /**
+ * Celebrate validation for the contact OTP sending endpoint.
+ */
+export const validateContactSendOtpParams = celebrate({
+  [Segments.BODY]: Joi.object().keys({
+    contact: Joi.string().required(),
+    userId: Joi.string().required(),
+  }),
+})
+
+/**
  * Celebrate validation for the contact OTP verification endpoint.
  */
 export const validateContactOtpVerificationParams = celebrate({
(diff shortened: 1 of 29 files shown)


tehtea:
Only added the change for the contact verification OTP for now, to verify the approach and also to check if there are any particular routes to not put this change on.

src/app/modules/user/user.middleware.ts
```diff
@@ -0,0 +1,14 @@
+import { celebrate, Joi } from 'celebrate'
+
+/**
+ * Celebrate validation for the contact OTP verification endpoint.
+ */
+export const validateContactOtpVerificationParams = celebrate({
+  body: Joi.object({
```
mantariksh, line 7: we normally use the `Segments` enum exposed by `celebrate` to avoid accidentally misnaming the keys. see `user.routes.ts` for reference

mantariksh (commented):
thank you @tehtea for the contribution, the approach is correct! in addition, you can delete the validation middleware for this route from `user.routes.ts` so that the middleware does not run twice.

tehtea:
Done! Nah, thank you all for open-sourcing this project, I learned a lot of good practices from here! Applied the changes to other routes I could find, feel free to take a look.

karrui:
Some checks seem to not be starting, @tehtea do you mind pushing an empty commit to see if that would retrigger the CI/CD?

karrui:
@tehtea can you try merging develop into this PR and see if the CI/CD gets triggered?  If that still does not work, as a last resort, I will pull your branch and reopen this PR

tehtea:
@karrui Okay, just rebased the commits and force-pushed, let's see how it goes

karrui:
Sorry for the wait, we should be able to trigger our CI/CD once #3175 is merged in.

karrui:
@tehtea should work now if you rebase and push again. Thank you so much for being patient and we're so sorry that it took so long to get this handled :(

tehtea:
@karrui No worries, these things happen :) pushed some minor changes, here's hoping it works.

tshuli:
merged in #3428

PR #2824, opened 2021-09-20
Datadog agent integration

timotheeg:
## Context

We are implementing DataDog as a central monitoring system. The FormSg AWS account has been integrated to DataDog already, but more needs to be done:
* Run DataDog Agent on our servers
* Install various DataDog connectors (e.g. for express and winston)
* Report custom data as needed

This PR only install the DataDog agent on the FormSG servers.

## Approach
Run the DataDog agent on the FormSG server. It will report host system metrics.

The EB environment needs 2 new environment variables: `DD_API_KEY` (see [here](https://app.datadoghq.com/account/settings#api)) and `DD_ENV` (`staging`|`production`).

Tested on staging, sample data here:
https://app.datadoghq.com/dashboard/b3u-egb-apu/formsg?tpl_var_env=staging&from_ts=1632260114510&to_ts=1632274514510&live=true

Note: The DD agent is configured to accept non-local traffic to allow host-Docker communication. Care should be taken to never expose the DogStatsD port 8125/8126(?) to other hosts, or the internet.

## Deploy Notes
- [ ]  Early (e.g. 730am or 8am) add the 2 environment variables below (cause EB redeploy, but new env vars are not used yet)
- [ ] Deploy new as normal 

- `DD_API_KEY` : FormSG specific API key for DataDog
- `DD_ENV`: `staging` or `production`

**New dependencies**:

- [`hot-shots`](https://www.npmjs.com/package/hot-shots) : client for StatsD/Datadog Agent
- [`connect-datadog`](https://www.npmjs.com/package/connect-datadog): integrations for Connect/Express


mantariksh:
>  Early (e.g. 730am or 8am) add the 2 environment variables below (cause EB redeploy, but new env vars are not used yet)

@timotheeg can I check why we need to do this? adding environment variables is safe, we do it regularly (e.g. to update site banners)

timotheeg:
> > Early (e.g. 730am or 8am) add the 2 environment variables below (cause EB redeploy, but new env vars are not used yet)
> 
> @timotheeg can I check why we need to do this? adding environment variables is safe, we do it regularly (e.g. to update site banners)

As we discussed earlier, this was me being paranoidically careful. Changing env vars causes an EB deployment, which means old instances are put in drain mode. That is safe, provided we do not have long running requests which exceed the drain timeout (I haven't checked what's the drain timeout for our setup btw). Such requests can break, so doing an early change is just reducing risks. Since the majority of our requests are short, that note may indeed be overkill. But ideally, we predictable response times for all requests (or built-in retries mechanisms for known long requests), guaranteeing that changing env vars is indeed completely safe at any time.

liangyuanruo:
I would also note that enabling non-local traffic to allow host-Docker communication also means that we should never expose the DogStatsD port 8125/8126(?) to the internet.

timotheeg:
> I would also note that enabling non-local traffic to allow host-Docker communication also means that we should never expose the DogStatsD port 8125/8126(?) to the internet.

That is correct. I don't think think we would ever open these ports (?). In fact they should not even be opened to any other hosts in the network.

src/app/loaders/express/index.ts
```diff
@@ -1,13 +1,16 @@
...
 import addRequestId from 'express-request-id'
+import { StatsD } from 'hot-shots'
 import http from 'http'
 import { Connection } from 'mongoose'
 import nocache from 'nocache'
 import path from 'path'
 import url from 'url'
 
 import config from '../../config/config'
+// import { datadogConfig } from '../../config/features/datadog.config'
```
tshuli, line 13: Is this commented import needed?

tshuli (approved):
lgtm except one nit

src/app/loaders/express/index.ts
timotheeg: Nope, that's an accident! Good catch! :)

timotheeg: Pushed commit to fix! Thanks!

src/app/loaders/express/index.ts
```diff
@@ -141,6 +143,21 @@ const loadExpressApp = async (connection: Connection) => {
...
+  app.use(
+    connectDatadog({
+      method: true, // i.e. normalized path, to keep cardinality in check
+      response_code: true,
+      path: false, // !! Important: do not turn this true or the tag cardinality will explode
+      dogstatsd: new StatsD({
+        useDefaultRoute: true,
+        errorHandler: (error) => {
+          // eslint-disable-next-line no-console
+          console.log(error)
```
mantariksh, line 155: under what circumstances will we encounter errors here? should we be logging a specific string here so we can set alerts for it?

mantariksh (approved):
lgtm, one question

src/app/loaders/express/index.ts
timotheeg: Actually, I'm not sure and I was even thinking of removing this error handler entirely 🤔 . I added it yesterday because I had trouble getting data, and I was suspecting some app-level issue, but I had just misconfigured the agent (shame on me, I didn't read Alwyn's sample properly! 😑 )

Now that you mention it I think I will remove it. being unable to send metrics would likely be reflected in the metrics themselves (no data or less data?), but logging every metric push error would just pollute logs.

timotheeg: removed

src/app/loaders/express/index.ts
```diff
@@ -141,6 +143,17 @@ const loadExpressApp = async (connection: Connection) => {
   // Log intranet usage
   app.use(IntranetMiddleware.logIntranetUsage)
 
+  app.use(
+    connectDatadog({
+      method: true, // i.e. normalized path, to keep cardinality in check
```
timotheeg, line 148: This comment is not correct 😑

PR #2836, opened 2021-09-21
Revert "refactor: use modern JS constructs to validate NRIC"

timotheeg:
Reverts opengovsg/FormSG#2785

The PR was not compatible with IE11, which is not acceptable for end users.

Reverting now, to keep `develop:HEAD` release-clean. The PR will be re-introduced once it has been made compatible.

@mantariksh @tshuli

### DIFF
--- a/shared/utils/nric-validation.ts
+++ b/shared/utils/nric-validation.ts
@@ -1,33 +1,41 @@
-type NricParts = {
-  prefix: string
-  digits: string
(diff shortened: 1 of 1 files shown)


mantariksh (approved):

mantariksh:
@timotheeg think you might need to add a `refactor: ` in front of the commit message to make the semantic PR check pass!

Issue #2539, opened 2021-08-06
Improve validation of Singpass/Corppass relay state


mantariksh:
1. At the point where the relay state is handed back to the /singpass/login and /corppass/login endpoints, the `formId` should be validated to be a valid MongoDB ObjectID the moment it is parsed. Currently the validation is very loose, and it only happens to be validated when we attempt to retrieve the form by its ID.
2. The `DESTINATION_REGEX` should be used with `.match` instead of `.exec`, since `.exec` mutates the state of the regex and the same regex is shared across multiple functions.

timotheeg:
Just to add a small clarification on `exec()` vs `match()` . The mutation is only an issue when regexes has a `g` flag, which is not the case here, but still: because such cases are really not obvious at all, I recommend using `.match()` all the time where the intent is to do a single extraction (or just `.test()`),  and only use `exec()` when you need to traverse a source string with multiple matches.

Sample script to illustrate the issue:
```
s1 = "foobar";
s2 = "foobarbaz";

re = /foo/g;

console.log( !! re.exec(s1) ); // true
console.log( !! re.exec(s2) ); // false
```

frankchn:
Fixed in #2869

PR #2869, opened 2021-09-27
fix: constrain the mongodb objectid regex format and use match

frankchn:
## Problem
See #2539, which this closes.

## Solution
1. Applies a more intense regex check to the SPCP return path (making sure that the ObjectID in the path is actually a valid ObjectID)
2. Use `str.match(regex)` instead of `regex.exec(str)` as part of best practices.

**Breaking Changes** 

## Tests
Make sure that SPCP login still works.


karrui (approved):
lgtm

PR #2878, opened 2021-09-28
feat(form-v2): add `/client/env` route for sending client env vars as JSON

karrui:
## Problem

The old route was `/client/environment`, which sent back the appropriate env vars as a javascript file to be merged into some header in the html. This seems unnecessary in React and we can just send the environment variables as an object instead.

For backwards compatibility, create new service and endpoint.

## Solution

**Breaking Changes** 

    - Details ...

**Features**:

- feat: add `api/v3/client/env` endpoint for sending client related env vars
- feat: add env service and query to retrieve env vars from server


tshuli (approved):
lgtm

PR #2892, opened 2021-09-30
chore: add postinstall step for including shared folder in frontend

karrui:
Some Chromatic builds (see #2888, #2891 before the fix commit) are failing due to having dependencies on some shared types and node modules.
This PR fixes that by installing shared dependencies in the Chromatic workflow (by adding installation of shared dependencies in the frontend/postinstall script)


tshuli (approved):
lgtmmm

PR #2950, opened 2021-10-19
chore(deps): bump vm2 to 3.9.4 in package-lock

tshuli:
## Problem
- bumps vm2 to 3.9.4 in package-lock to fix high severity vulnerability

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25218,9 +25218,9 @@
       "dev": true
     },
     "vm2": {
-      "version": "3.9.3",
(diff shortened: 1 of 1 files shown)


tshuli:
as discussed, it could, but in this case `vm2` is a dependency of `degenerator@3.0.1`, which is a nested depedency itself. `3.9.4` is still compatible with the versioning
    `"vm2": "^3.9.3"`

as a further check, snyk would pick it up if there's any reversion

PR #3030, opened 2021-11-05
chore: add workflow_dispatch event handler for ci.yml github action

karrui:
Sometimes we want to manually trigger CI from the CLI or on Github itself, and adding the `workflow_dispatch` event to the CI config is a way to do that.


karrui:
Closing for now

PR #3047, opened 2021-11-15
fix(v2): attempt deflaking Menu and Tooltip stories with chromatic param

karrui:
## Problem

Tooltip and Menu stories have flaking, possibly due to device width changing and the opened Tooltips/Menu components having to readjust. This PR attempts to fix that issue by requesting chromatic to wait for all "animations" to end before taking a snapshot. (if it persists, will try a delay)

## Solution

**Bug Fixes**:

- attempt deflaking Menu and Tooltip stories with chromatic param

## Before & After Screenshots
Hopefully the snapshots have been fixed.


karrui:
Probably doesn't work... will figure out other ways, such as setting set width

PR #3109, opened 2021-11-29
chore(ci): add semantic.yml to modify CI behaviour

karrui:
## Problem

This PR adds config for our semantic PR checker github action.

The changes are to only validate titles since we only squash commits, and add ref type for refactor shorthand


tshuli (approved):
lgtm

PR #3110, opened 2021-11-29
feat: Add twilio webhook endpoint for failed SMS deliveries

### LINKED ISSUE #3103: Create endpoint for Twilio webhooks
We currently catch the errors Twilio returns on creation of sms, but we do not have a way to catch errors when Twilio fails to send some SMSes.

We should expose an API Twilio can use to ping our servers on errors.

See https://www.twilio.com/docs/usage/webhooks

zatkiller:
## Problem

Currently, there is no error handling for failed delivery of Twilio SMSes.

Closes #3103

## Solution

Add a new endpoint which can be exposed to the Twilio Webhook status updates and logs any SMS failures

**Features**:
- New endpoint at `v3/notifications/twilio` which logs the errors if SMS delivery status is `failed` or `undelivered`
- Validation of request from Twilio using Twilio SDK

## Tests
- Update sms.service tests to include `statusCallback`
- Add `twilio.controller.spec.ts`
- Add `notifications.route.spec.ts`


src/app/modules/twilio/twilio.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param req Express request object
+ * @param res - Express response object
+ */
+export const twilioSmsUpdates: ControllerHandler<
+  unknown,
+  never,
+  ITwilioSmsWebhookBody
+> = async (req, res) => {
+  if (smsDeliveryFailedStatus.includes(req.body.MessageStatus))
+    logFailedSmsDelivery(req.body)
```
karrui, line 29: nit: add braces in `if` block for readability

src/app/modules/twilio/twilio.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+  res.sendStatus(StatusCodes.OK)
```
karrui, line 30: should we add the `return` keyword here? We've had cases of multiple `res.sendXXX` calls in the past due to forgetting to return

src/app/modules/twilio/twilio.service.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @param body body from Twilio Webhook with the failed sms details
+ */
+export const logFailedSmsDelivery = (body: ITwilioSmsWebhookBody): void => {
+  const { To, MessageSid, MessageStatus, ErrorCode, ErrorMessage } = body
+
+  logger.error({
+    message: 'Error occurred when attempting to send SMS on twillio',
+    meta: {
+      action: 'sendingTwilioSmsFailed',
```
karrui, line 17: `meta.action` seems to be different from the service name. We should keep it congruent

src/app/modules/twilio/twilio.service.ts
```diff
@@ -0,0 +1,25 @@
...
+      To,
+      MessageSid,
+      MessageStatus,
+      ErrorCode,
+      ErrorMessage,
```
karrui, line 22: Should we just log the entire webhook body?

src/app/modules/twilio/twilio.controller.ts
```diff
@@ -0,0 +1,33 @@
+import { StatusCodes } from 'http-status-codes'
+import Twilio from 'twilio'
+
+import { ITwilioSmsWebhookBody } from 'src/types/twilio'
+
+import { ControllerHandler } from '../core/core.types'
+
+import { logFailedSmsDelivery } from './twilio.service'
+
+const smsDeliveryFailedStatus = ['undelivered', 'failed']
```
karrui, line 10: I think we can just inline this; are there any stronger types too? We should also link to the documentation.

src/app/modules/twilio/twilio.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param req Express request object
+ * @param res - Express response object
+ */
+export const twilioSmsUpdates: ControllerHandler<
+  unknown,
+  never,
+  ITwilioSmsWebhookBody
+> = async (req, res) => {
+  if (smsDeliveryFailedStatus.includes(req.body.MessageStatus))
```
karrui, line 28: should we also log every other message status, using `logger.info` instead of `logger.error`?

src/app/services/sms/sms.service.ts
```diff
@@ -248,7 +251,7 @@ const sendSms = (
         logType: LogType.success,
       })
 
-      return result
+      return result as true
```
karrui, line 254: do we need to cast here? did it not infer the types automatically? Might want to debug why it is not if that's the case

(checked on my machine on my current react branch, it is automatically inferring)

src/app/services/sms/sms.service.ts
```diff
@@ -259,7 +262,7 @@ const sendSms = (
         logType: LogType.failure,
       })
 
-      return error
+      return error as SmsSendError | InvalidNumberError
```
karrui, line 265: ^

src/types/twilio.ts
```diff
@@ -0,0 +1,13 @@
+// Based off Twilio documentation - https://www.twilio.com/docs/usage/webhooks/sms-webhooks
+export interface ITwilioSmsWebhookBody {
+  SmsSid: string
+  SmsStatus: string
+  MessageStatus: string
```
karrui, line 5: should we use  a string union of all known status types here instead of the generic `string` type?  to make it clearer

karrui (changes requested):
some changes requested. Can we also add tests for the new API?

src/app/modules/twilio/twilio.controller.ts
zatkiller: How does the multple res.sendxxx call happen? Can't really visualize it in my head :sweat_smile:

karrui:
```ts
const someFn = () => {
  if (someCondition) {
    res.sendStatus(404)
  }
  // ... a bunch of extra work that happened because a return was not called
  res.sendStatus(200) // Double send error will occur here
}
```

Not likely in your code, but still good to be consistent across the codebase

src/app/modules/twilio/twilio.controller.ts
zatkiller: Currently, it seem like its provided only as string values from the example seen here: https://www.twilio.com/docs/usage/webhooks/sms-webhooks. 

Checked the twilio typescript support, dosen't seem to have stronger types for it either

src/app/modules/twilio/twilio.controller.ts
zatkiller: If this is the case, should I remove twilio.service and have logging of both error and non-errors on the controller level?

src/app/modules/twilio/twilio.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const logger = createLoggerWithLabel(module)
+
+/**
+ * Middleware which validates that a request came from Twilio Webhook
+ */
+const validateTwilioWebhook = Twilio.webhook({ validate: !isDev })
+
+/**
+ * Process the Webhook requests if they are failed or unsuccessful, ignoring the rest
```
karrui, line 18: nit: this JSDoc seems outdated since we log everything now

src/app/services/sms/sms.service.ts
```diff
@@ -186,12 +186,15 @@ const sendSms = (
...
 
+  const statusCallbackRoute = '/api/v3/notifications/twilio'
+
   return ResultAsync.fromPromise(
     client.messages.create({
       to: recipient,
       body: message,
       from: msgSrvcSid,
       forceDelivery: true,
+      statusCallback: config.app.appUrl + statusCallbackRoute,
```
karrui, line 197: to prevent the URL from desyncing, should we export a constant from the router itself so we have a higher chance of remembering to update it when we update the routes

e.g

```ts
// src/app/routes/api/v3/notifications/notifications.routes.ts
export const ABSOLUTE_TWILIO_NOTIF_ROUTE = '/api/v3/notifications/twilio'

// src/app/services/sms/sms.service.ts
statusCallback: `${config.app.appUrl}${ABSOLUTE_TWILIO_NOTIF_ROUTE}`
```

karrui (approved):
lgtm, a couple nits

src/app/services/sms/sms.service.ts
zatkiller: I tried adding this const into `notifications.routes` and imported it into `sms.service` but I get a `Route.post() requires callback functions but got a "object Undefined" error from `routes/.../notifications/bounce.routes`,  which should not be the case since POST '/email` under `bounce.routes` takes in `handleSns`

Seems like this error is due to a circular dependency being created. I traced the dependencies, and there seems to be a dependency in this order `notifications.routes` -> `bounce.routes` -> `bounce.controller` -> `bounce.service` -> `SmsFactory` -> `sms.service` -> `notification.routes`.
Also, managed to console log and find out that `handleSns` becomes `undefined` once I import`ABSOLUTE_TWILIO_NOTIF_ROUTE` into `sms.service

Should we move the const `ABSOLUTE_TWILIO_NOTIF_ROUTE` elsewhere or leave it in `sms.service`

karrui: i guess we can leave it. Just add a comment in sms routes in the handler to tell future changes to sync to the given file

Issue #3099, opened 2021-11-25
[React] set up web workers for decryption 


karrui:
This issue tracks setting up of web/service workers in the React application to prepare to be used for decrypting responses in Storage Mode without blocking UI.

Things we may want to explore:
- Web vs service worker
- Streaming download vs storing in memory
- Progress bar UX?

karrui:
Closed by #3125

PR #3125, opened 2021-12-02
feat(v2): Set up web workers on React

zatkiller:
## Problem

Currently in the AngularJS version, Web Workers are used to decrypt the storage mode submissions without blocking the UI. We have to reimplement this for React.

Related to  #3099
## Solution

- Set up worker-loader (webpack library) to spin up web workers
- Used comlink to communicate with webworkers in a rpc implementation

**Breaking Changes** 

**Features**:
- Barebone `ResponsesPage` with button and input to test our web workers creation and teardown
- Creation of worker pool and worker pool teardown under `downloadEncryptedResponses`
- `getFormSubmissionsMetadata`  and `useFormResponses` to retrieve form submissions meta data

- `decryption.worker.ts` : script run by web workers for decryption (Implementation will be done in follow up PRs)

**New dependencies**:
- [comlink](https://github.com/GoogleChromeLabs/comlink): Library to communicate with web workers in RPC style


frontend/src/features/admin-form/responses/ResponsesPage.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { useFormResponses } from './queries'
+
+const ResponsesPage = (): JSX.Element => {
+  const { data: settings } = useAdminFormSettings()
+  const { data, isLoading } = useFormResponses()
+  const { formId } = useParams()
+  const [secretKey, setSecretKey] = useState<string>('')
+
+  const handleCsvExport = async () => {
+    await downloadEncryptedResponses(formId!, settings!.title, secretKey)
```
karrui, line 19: does this not block the web ui?

frontend/src/features/admin-form/responses/queries.ts
```diff
@@ -1,13 +1,19 @@
...
-import { countFormSubmissions } from './AdminSubmissionsService'
+import {
+  countFormSubmissions,
+  getFormSubmissionsMetadata,
+} from './AdminSubmissionsService'
 
 export const adminFormResponsesKeys = {
   base: [...adminFormKeys.base, 'responses'] as const,
   id: (id: string) => [...adminFormResponsesKeys.base, id] as const,
+  count: (id: string) => [...adminFormResponsesKeys.base, id, 'count'] as const,
```
karrui, line 16: nit: this can be simplified to 
```tsx
count: (id) => [...adminFormResponsesKeys.id(id), 'count'] as const
```

frontend/src/features/admin-form/responses/ResponsesPage.tsx
zatkiller: Oh yes my bad, there shouldn't be an await here!

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -24,3 +31,54 @@ export const countFormSubmissions = async ({
...
+}
+
+interface Worker {
+  worker: DecryptionWorker
+  workerApi: Remote<{
+    log: (id: number) => string
+  }>
+}
+
+export const downloadEncryptedResponses = async (
```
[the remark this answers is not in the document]
zatkiller: Ah ok. So I should implement the singleton class within the component also if I were to go with that approach?

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -1,4 +1,11 @@
-import { SubmissionCountQueryDto } from '~shared/types/submission'
+import { Remote, wrap } from 'comlink'
+// eslint-disable-next-line import/no-webpack-loader-syntax
+import DecryptionWorker from 'worker-loader!./workers/decryption.worker'
```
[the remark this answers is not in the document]
zatkiller: Just to confirm, so I'm supposed to update the worker-loader rule in `craco.config.js` instead of webpack configs?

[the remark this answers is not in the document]
karrui: yes

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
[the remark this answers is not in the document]
karrui: I think as a start we can expose it as a React hook that cancels/kills all web worker activity if the function is called again.

```tsx
const useEncryptionWorkers = () => {
  const [workers, setWorkers] = useState<Worker[]>([])

  const downloadEncryptedResponses = (...) => {
     if (!isEmpty(workers)) {
      // kill workers
    }
    // Do the usual
  }

  return downloadEncryptedResponses
}
```

frontend/src/features/admin-form/responses/ResponsesPage.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const numWorkers = window.navigator.hardwareConcurrency || 4
+
+    const workerPool: WorkerInterface[] = []
+
+    // Workerpool sample setup
+    for (let i = 0; i < numWorkers; i++) {
+      const worker = new DecryptionWorker()
+      const workerApi =
+        wrap<import('./workers/decryption.worker').DecryptionWorker>(worker)
+      workerPool.push({ worker, workerApi })
```
karrui, line 41: Is there a point to also adding the `worker` constant into the map? Feels unnecessary and never gets used at all.
I think it would be cleaner if we just have an array of workers?

suggest
```ts
const worker = wrap<DecryptionWorker>(new DecryptionWorker())
workerPool.push(worker)
```

frontend/src/features/admin-form/responses/ResponsesPage.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+import { WorkerInterface } from './workers/type/workerinterface'
+import { killWorkers } from './AdminSubmissionsService'
+//import { downloadEncryptedResponses } from './AdminSubmissionsService'
+import { useFormResponses } from './queries'
+
+const useDecryptionWorkers = () => {
+  const [workers, setWorkers] = useState<WorkerInterface[]>([])
+
+  const downloadEncryptedResponses = (
```
karrui, line 25: nit: memoize this function by using `useCallback`

frontend/src/features/admin-form/responses/ResponsesPage.tsx
```diff
@@ -0,0 +1,100 @@
...
+import Button from '~components/Button'
+
+import { useAdminFormSettings } from '../settings/queries'
+
+import { WorkerInterface } from './workers/type/workerinterface'
+import { killWorkers } from './AdminSubmissionsService'
+//import { downloadEncryptedResponses } from './AdminSubmissionsService'
+import { useFormResponses } from './queries'
+
+const useDecryptionWorkers = () => {
```
karrui, line 22: nit: prefer this to be its own hook file

frontend/src/features/admin-form/responses/workers/type/workerinterface.ts
```diff
@@ -0,0 +1,13 @@
+import { Remote } from 'comlink'
```
karrui, line 1: don't think there is a need for this interace

package.json
```diff
@@ -97,6 +97,7 @@
     "bson-ext": "^2.0.5",
     "busboy": "^0.3.1",
     "celebrate": "^15.0.0",
+    "comlink": "^4.3.1",
```
karrui, line 100: as communicated on slack, this is installed in the wrong folder; should be installed in `/frontend`, and that is why chromatic build is failing :P

karrui (changes requested):

frontend/src/features/admin-form/responses/ResponsesPage.tsx
zatkiller: The worker constant needs to be kept in order to call `terminate` on the workers

karrui: I see, missed that part. Can't we then just store worker in workerPool and call wrap on the worker when we actually access the functions?

zatkiller: I think that will work!

frontend/package.json
```diff
@@ -8,9 +8,10 @@
     "@emotion/react": "^11.7.0",
     "@emotion/styled": "^11.6.0",
     "@opengovsg/formsg-sdk": "^0.9.0",
-    "axios": "^0.24.0",
-    "country-flag-icons": "^1.4.19",
-    "date-fns": "^2.27.0",
+    "axios": "^0.21.4",
```
karrui, line 11: regression here

frontend/package.json
```diff
@@ -8,9 +8,10 @@
...
+    "comlink": "^4.3.1",
+    "country-flag-icons": "^1.2.10",
+    "date-fns": "^2.23.0",
```
karrui, line 14: these two package versions regressed too

frontend/src/features/admin-form/responses/ResponsesPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+        Enter secret key:
+        <input
+          style={{ backgroundColor: 'grey' }}
+          type="text"
+          value={secretKey}
+          onChange={(e) => setSecretKey(e.target.value)}
+        />
+        <Button
+          onClick={() =>
+            downloadEncryptedResponses(formId!, settings!.title, secretKey)
```
karrui, line 33: I think we should extract a new `handleExportCsv` function that first checks whether the formId and settings exist instead of using `!` to forcibly assert truthiness; this essentially defeats the point of typescript;

```tsx
const handleExportCsv = useCallback(() => {
  if (!formId || !settings?.title) return
  return downloadEncryptedResponses(formId, settings.title, secretKey)
}, [])
```

frontend/src/features/admin-form/responses/ResponsesPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+          }
+        >
+          Export csv{' '}
```
karrui, line 36: nit: extra unneeded space

frontend/src/features/admin-form/responses/ResponsesPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+        </Button>
+        <Button>Export csv and attachments</Button>
+        {!!data &&
+          data.metadata.map((submission: StorageModeSubmissionMetadata) => {
```
karrui, line 40: nit: we can use optional chaining here instead of a truthiness check: `data?.metadata.map`

package.json
karrui: please remove this package even though it's also in /frontend directory haha

karrui (changes requested):

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -24,3 +33,22 @@ export const countFormSubmissions = async ({
...
+
+/**
+ * Get of submissions metadata for a given form
+ * @param formId fomID to retrieve metadata on
+ * @returns The metadata of the form
+ */
+export const getFormSubmissionsMetadata = async (
+  formId: string,
+): Promise<StorageModeSubmissionMetadataList> => {
+  const queryUrl = `${ADMIN_FORM_ENDPOINT}/${formId}/submissions/metadata?page=1` // To be generalized in later PRs
```
karrui, line 45: we should generalize this now, it may be missed in later PRs. there is an existing method that should be a copy-paste (and changing axios to ApiService) in `src/public/services/AdminSubmissionsService.ts#getSubmissionsMetadataByPage`

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -24,3 +33,22 @@ export const countFormSubmissions = async ({
...
+  return ApiService.get(queryUrl).then(({ data }) => data)
+}
+
+/**
+ * Terminates all workers
+ * @param workers an array of workers
+ */
+export const killWorkers = (workers: DecryptionWorker[]): void =>
```
karrui, line 53: should this function be in the hook itself since that's the only location its used?

karrui (commented):
some final changes :P

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
zatkiller: Sure, i think that will be better too. Will generalize this function and give it a default value = 1 in the params since there is no pagination yet

karrui (approved):
lgtm

PR #3139, opened 2021-12-03
feat: Add prefill support for singpass forms with redirects

### LINKED ISSUE #2985: Prefill to support singpass forms
Currently, prefill does not support singpass forms. 

TODO: Enhance prefill to support singpass (+ myinfo and sgid) forms. The prefill query params will have to be included in the redirect url.

frankchn:
## Problem
The prefilled forms currently do not support Singpass logins since we lose the query string parameters when we are redirected to Singpass.

Closes #2985

## Solution
This encodes the query string in base64 then transports it to Singpass and back via `relayState`. It then parses the `relayState` then redirects the user back to their form with the query string containing the prefilled form elements still intact.

**Breaking Changes** 

## Screenshot
image

## Tests
Verify that Singpass logins work with a standard form as well as a form with pre-filled parameters.


src/app/modules/spcp/spcp.service.ts
```diff
@@ -324,15 +324,23 @@ export class SpcpServiceClass {
...
       })
       return err(new InvalidOOBParamsError())
     }
-    const destination = payloads[0]
+
     const rememberMe = payloads[1] === 'true'
+    const encodedQuery = payloads.length === 3 ? payloads[2] : ''
+    const decodedQuery =
+      encodedQuery.length > 0
+        ? `?${Buffer.from(encodedQuery, 'base64').toString('utf8')}`
```
karrui, line 339: this may throw, we might want to catch it somehow

src/app/modules/spcp/spcp.util.ts
```diff
@@ -299,10 +295,11 @@ export const getRedirectTarget = (
...
   isPersistentLogin?: boolean,
+  encodedQuery?: string,
 ): string =>
   `/${formId},${
     // Need to cast to boolean because undefined is allowed as a valid value
     // We are not following corppass's official spec for
     // the target parameter
     authType === FormAuthType.SP ? !!isPersistentLogin : false
-  }`
+  },${encodedQuery ?? ''}`
```
karrui, line 305: nit: should we not add encoded query if it does not exist at all instead of an empty string?

src/app/modules/spcp/spcp.service.ts
```diff
@@ -324,15 +324,23 @@ export class SpcpServiceClass {
...
         meta: logMeta,
       })
       return err(new InvalidOOBParamsError())
     }
-    const destination = payloads[0]
+
     const rememberMe = payloads[1] === 'true'
+    const encodedQuery = payloads.length === 3 ? payloads[2] : ''
+    const decodedQuery =
+      encodedQuery.length > 0
```
karrui, line 338: nit: can be simplified to 

```
encodedQuery ? `?${Buffer....}` : ''
```

src/app/modules/spcp/spcp.util.ts
```diff
@@ -65,11 +65,7 @@ export const isValidAuthenticationQuery = (
   destination: string,
   idpPartnerEntityId: string,
 ): boolean => {
-  return (
-    !!destination &&
-    isArtifactValid(idpPartnerEntityId, samlArt) &&
-    DESTINATION_REGEX.test(destination)
-  )
+  return !!destination && isArtifactValid(idpPartnerEntityId, samlArt)
```
karrui, line 68: should we update the regex instead of completely removing it? Or is the regex unneeded after all?

src/app/modules/spcp/spcp.util.ts
```diff
@@ -299,10 +295,11 @@ export const getRedirectTarget = (
   formId: string,
   authType: FormAuthType.SP | FormAuthType.CP,
   isPersistentLogin?: boolean,
+  encodedQuery?: string,
 ): string =>
```
karrui, line 299: since we're here... should we update the types to make it stricter?

```ts
type RedirectTarget = `${string},${boolean},${string}`
```

src/public/services/PublicFormAuthService.ts
```diff
@@ -13,12 +13,13 @@ export const PUBLIC_FORMS_ENDPOINT = '/api/v3/forms'
 export const createRedirectURL = (
   formId: string,
   isPersistentLogin = false,
+  encodedQuery = '',
```
karrui, line 16: should we allow this to be `undefined` instead of setting it to an empty string?

```
encodedQuery?: string
```

src/public/services/__tests__/PublicFormAuthService.test.ts
```diff
@@ -28,7 +28,7 @@ describe('PublicFormAuthService', () => {
       expect(MockAxios.get).toHaveBeenCalledWith(
         `${PublicFormAuthService.PUBLIC_FORMS_ENDPOINT}/${MOCK_FORM_ID}/auth/redirect`,
         {
-          params: { isPersistentLogin: false },
+          params: { isPersistentLogin: false, encodedQuery: '' },
```
karrui, line 31: tests can then be simplified to remove the encodedQuery parameter.

should we also add a test to check if the API is called with an expected `encodedQuery`?

karrui (changes requested):
we should catch the Buffer.from call in case it throws

src/app/modules/spcp/spcp.service.ts
```diff
@@ -334,10 +334,21 @@ export class SpcpServiceClass {
...
         : ''
+    } catch (err) {
+      logger.warn({
+        message: 'Unable to decode encodedQuery',
+        meta: {
+          encodedQuery,
+        },
+        error: err,
+      })
+    }
```
karrui, line 352: i don't think we should let this be a silent failure; should we return an appropriate error (`InvalidOOBParamsError`) using neverthrow instead so the calling function can handle it?

```ts
return err(new InvalidOOBParamsError())
```

src/app/modules/spcp/spcp.util.ts
```diff
@@ -53,6 +53,8 @@ const isArtifactValid = function (
   )
 }
 
+export type RedirectTarget = `${string},${boolean},${string}`
```
karrui: this may cause the build to fail to compile, suggest
```ts
export type RedirectTarget = `${string},${boolean},${string}` | `${string},${boolean}` 
```
since there are 2 variants

src/app/modules/spcp/spcp.service.ts
frankchn: yeah this is fair too

karrui (approved):
lgtm

PR #3147, opened 2021-12-06
fix(ci/deploy): set truncated description during ELB deployment

karrui:
## Problem

We use git commit messages as the description for our ELB deployment.
AWS has a hard limit of 200 characters, which may exceed when a commit message is longer (like this) and throw an error. This commit attempts to fix that by truncating the commit message if it is longer than 200 characters.


tshuli (approved):
lgtm

PR #3156, opened 2021-12-07
Fix/twilio webhook

### LINKED ISSUE #3154: Manually validate Twilio webhooks to prevent HTTP 403 errors
The new Twilio webhooks endpoints to receive Sms delivery status is getting 403 error on new release (but did not trigger any 403 errors on staging). 

We should manually verify the signature by using the methodology as described from https://www.twilio.com/docs/usage/security instead of using the abstracted `Twilio.webhook` middleware from the twilio nodejs sdk package.

Related issue (Twilio webhooks endpoint): #3103

zatkiller:
## Problem

Current, twilio webhook endpoint to receive sms delivery status updates returns 403 in new release 5.37.0. 

This is because during the validation step, the auth token used by the twilio validation middleware is different from the auth token used to send sms (users own token).

Closes #3154

# Solution

Switch to validate Twilio webhook requests by using `celebrate` to check for the presence of twilio request headers and request body parameters

### DIFF
--- a/src/app/modules/twilio/twilio.controller.ts
+++ b/src/app/modules/twilio/twilio.controller.ts
@@ -1,9 +1,8 @@
+import { celebrate, Joi, Segments } from 'celebrate'
 import { StatusCodes } from 'http-status-codes'
-import Twilio from 'twilio'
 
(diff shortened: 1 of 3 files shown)


zatkiller:
Opened new PR #3157

PR #3162, opened 2021-12-08
feat: enable M series FIN validation

### LINKED ISSUE #3155: Use M series FIN/NRIC validation function
We already have the function, we just need to use it. This validation function should be used ASAP and does not need to wait until 1 Jan 2022.

zatkiller:
## Problem

Currently, M series FIN/NRIC validation has been implemented but is not being used.

Closes #3155

## Manual Tests
- Create a NRIC field on a form
- Type in M1234567K (or any other valid M series FIN value), there should be no warning displayed in the browser
- Submit a response to the form with M1234567K as the value in the NRIC field, no error should occur


karrui (approved):
lgtm, please add manual tests for release testing

PR #3178, opened 2021-12-13
feat(v2): update Menu styling to better fit design

karrui:
## Problem

There were slight variances in the Menu component as compared to the design. This PR attempts to correct those variances.

There are also some attempts to fix Chromatic snapshot flakiness on Menu and Tooltip stories by adding a font preload using Google fonts so the snapshot will not move that much.

## Solution

**Improvements**:

- feat(Button): allow forwardRef
- style(Menu): update theme styling to better fit design


tshuli (approved):
lgtm!

PR #3191, opened 2021-12-15
chore: enable dependabot for /shared directory

karrui:
Dependabot has not been running on the `/shared` directory ever since it was shared out. This PR adds that in


tshuli (approved):
lgtm

PR #3207, opened 2021-12-20
build: hotfix release 5.38.1 - update datadog agent to 7.32.1

karrui:
## Hotfix release 5.38.1

This hotfix release updates the datadog agent used in our ebextensions to a patched version that mitigates the current Log4j vulnerability. 

This does not mean our current application is vulnerable to the exploit. **We do not use Log4j in our application and the datadog agent used in `.ebextensions` is also not affected.**

Hotfixing only out of an abundance of caution.

Further reading: https://www.datadoghq.com/log4j-vulnerability/

### Fixes
- fix(ebextensions): update datadog agent to 7.32.3 to fix log4j vuln


tshuli (approved):
lgtm

Issue #2992, opened 2021-10-26
[React] Public form auth feature


karrui:
**Time estimation: 1w**

This issue tracks the completion of adding form auth to forms

karrui:
Closed by #3210

PR #3210, opened 2021-12-20
feat(v2): add initial public form authentication feature

karrui:
## Problem

This PR adds the public form log in feature, with a few known issues that are not resolved to be completed in future PRs

Related to #2992

## Solution

**Features**:
- feat(react-query): disable retry if error code is 404 globally
- feat: add AuthImageSvgr component
- feat: add FormAuth component and business logic to log in/log out
  - also center form field div when FormAuth component is shown
- add `usePublicAuthMutations` hook for logging in/out of public forms with spcp/sgid

**Improvements**:

- ref: add colorTheme prop to all field components
  - This will help in the future when we need to theme all (or some) form fields
- ref(SectionField): move dividerColor extrapolation to component and use colorTheme prop instead
- ref: extract all possible field render state into separate components 

**Bug Fixes**:

- fix(Button): remove auto height styling from Button
  - causing some wonky loading spinner heights
- fix(Button): remove fontWeight from baseProp in theme

## Known issues
- [ ] Does not automatically log out when JWT expires
- [ ] Does not correctly redirect to React version and insteads go to AngularJS version

## Before & After Screenshots
New stories in Storybook

## Tests

SOON

### DIFF
--- a/frontend/src/app/App.tsx
+++ b/frontend/src/app/App.tsx
@@ -5,6 +5,7 @@ import { ChakraProvider } from '@chakra-ui/react'
 
 import { theme } from '~theme/index'
 import { AuthProvider } from '~contexts/AuthContext'
+import { HttpError } from '~services/ApiService'
 
 import { AppRouter } from './AppRouter'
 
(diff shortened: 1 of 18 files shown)


karrui:
Will split out Button changes in a new PR, seems to have lots of snapshot changes

PR #3223, opened 2021-12-22
fix: mongoose v6 replica set second node setup

zatkiller:
## Problem
When mongoose is upgraded to v6, it seems that the server selection runs into issues when using a single node replica set for local testing.

## Solution
This PR introduces a second mongo node for the local replica set with some changes in the replica set configuration


karrui:
Leaving on draft as v6 is rolled back

karrui:
Superseded(ish) by https://github.com/opengovsg/FormSG/pull/3752, where another image with better replicaSet support is used.

PR #3286, opened 2022-01-15
feat: pre-fill to support myinfo forms

### LINKED ISSUE #3183: Prefill to support MyInfo forms
Currently, prefill supports singpass forms completed in #3139, but MyInfo (and SGID) forms were missed due to using a different service.

TODO: Enhance prefill to support myinfo and sgid forms. The prefill query params will have to be included in the redirect url.

frankchn:
## Problem
Adds the relevant routes to pass relayState through to MyInfo forms.

Closes #3183 

## Solution
Added handling for `encodedQuery` into the MyInfo controllers.

**Breaking Changes** 

**Features**:
- Adds support for prefilled text for MyInfo forms.

## Tests
Create a MyInfo form with a pre-fill textbox and ensure that the pre-fill text is carried over when the user logs in.

### DIFF
--- a/src/app/modules/myinfo/myinfo.controller.ts
+++ b/src/app/modules/myinfo/myinfo.controller.ts
@@ -196,9 +196,20 @@ export const loginToMyInfo: ControllerHandler<
     return res.sendStatus(StatusCodes.BAD_REQUEST)
   }
   const { formId, cookieDuration, encodedQuery } = parseStateResult.value
-  const redirectDestination = encodedQuery
-    ? `/${formId}?${Buffer.from(encodedQuery, 'base64').toString('utf8')}`
-    : `/${formId}`
+
+  let redirectDestination = `/${formId}`
+
+  if (encodedQuery) {
+    try {
+      redirectDestination = `/${formId}?${Buffer.from(
+        encodedQuery,
+        'base64',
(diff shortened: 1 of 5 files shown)


src/app/modules/myinfo/myinfo.controller.ts
```diff
@@ -194,8 +195,10 @@ export const loginToMyInfo: ControllerHandler<
     })
     return res.sendStatus(StatusCodes.BAD_REQUEST)
   }
-  const { formId, cookieDuration } = parseStateResult.value
-  const redirectDestination = `/${formId}`
+  const { formId, cookieDuration, encodedQuery } = parseStateResult.value
+  const redirectDestination = encodedQuery
+    ? `/${formId}?${Buffer.from(encodedQuery, 'base64').toString('utf8')}`
```
karrui, line 200: may need a try-catch here if the Buffer is wrongly encoded

karrui (changes requested):
sorry for the late review

src/app/modules/myinfo/myinfo.controller.ts
frankchn: good point, changed

karrui (approved):

PR #3303, opened 2022-01-19
fix: correct busboy filename extraction when utf8 filenames are used

### LINKED ISSUE #3299: Downgrade BusBoy package back to 0.3.1
Filenames seem to be double-encoded, causing some differences in the file names in submitted attachments.
See relevant issue opened in https://github.com/mscdex/busboy/issues/274

karrui:
This PR fixes the file names being double/wrongly encoded when outside of latin1 encoding.

Closes #3299, since we didn't have to downgrade after all

Relevant issue where this change was suggested https://github.com/mscdex/busboy/issues/274

## Tests
- Add unit test for verifying utf8 files are not being encoded wrongly

Also tested on staging
Screenshot 2022-01-19 at 2 34 10 PM


tshuli (approved):
lgtm

PR #3387, opened 2022-02-08
fix: DNS resolution error causes 5xx on form deletion

### LINKED ISSUE #3380: DNS resolution error causes 5xx on form deletion
**Describe the bug**
The Form model validation contains a check that the DNS for the webhook URL resolves correctly. However, this means that when the admin attempts to update the form when the webhook URL is no longer valid (e.g. when the admin's webhook server has been taken down), the model-level validation fails, resulting in a 5xx error.

So far, 5xx errors have only occurred at the DELETE form endpoint. Further investigation is required to determine whether other endpoints are also affected.

**To Reproduce**
1. Add a valid webhook URL to a form
2. Change the webhook URL such that the DNS resolution for the URL returns an error
3. Attempt to delete the form

**Expected behavior**
For DELETE form endpoint, form should continue to be deleted.

For all other endpoints, should return 400 with a relevant error message instead of 5xx.

tshuli:
## Problem
- Webhook url validation is done as part of the form model. This includes form archiving. As described in #3380, this means that if the webhook URL is invalid (i.e. DNS resolution fails), the document update cannot be saved - this is expected behavior
- However, for form archiving failure, previously the errors were not correctly mapped, resulting in a default `500` being returned from the server (instead of `422` in this case). The frontend also failed silently without any prompt to the admin

Closes #3380

## Solution
- I considered if we should simply skip model validation for DELETE form endpoint. However, Delete does not remove the document permanently, it simply updates the `status` field, and the form could be revived in future. Therefore, model validation should continue to be applied as this is a document update, as a safeguard to prevent the DB from entering into an invalid state 
- The remedy taken is 
  - Fix the error routing so that the correct errors are returned when form archiving fails
  - Display error message on the frontend so admin knows what to do to resolve the issue
- Note that the same issue does not occur for other endpoints (e.g. create / update form field, create form logic) which  trigger form validation as the errors are correctly mapped

- Separately, this PR adds in pre save validation to ensure only storage mode forms can have webhook url (resolves a TODO in the codebase) 

**Breaking Changes** 

    - Details ...

## Screenshot

Screenshot 2022-02-08 at 12 19 46 PM

## Tests
- Create storage mode form. Enter valid webhook url.
- In the DB, change the webhook url to an invalid one
- Attempt to delete form. Toastr should show up telling user to update the url. In network tab, check that the status code is `422`

### DIFF
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -885,8 +885,7 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     // Webhooks only allowed if encrypt mode
     if (
       this.responseMode !== FormResponseMode.Encrypt &&
-      this.webhook?.url &&
-      this.webhook.url.length > 0
+      this.webhook?.url?.length > 0
     ) {
       const validationError = this.invalidate(
         'webhook',
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -885,7 +885,7 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     // Webhooks only allowed if encrypt mode
     if (
       this.responseMode !== FormResponseMode.Encrypt &&
-      this.webhook?.url?.length > 0
+      (this.webhook?.url?.length ?? 0) > 0
     ) {
       const validationError = this.invalidate(
         'webhook',
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -397,7 +397,6 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       },
 
       webhook: {
-        // TODO: URL validation, encrypt mode validation
         url: {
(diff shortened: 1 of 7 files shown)


src/app/models/form.server.model.ts
```diff
@@ -883,6 +882,19 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       return next(err)
     }
 
+    // Webhooks only allowed if encrypt mode
+    if (
+      this.responseMode !== FormResponseMode.Encrypt &&
+      this.webhook?.url &&
+      this.webhook.url.length > 0
```
karrui, line 889: nit: can collapse into `this.webhook?.url?.length > 0`

karrui (approved):
lgtm

src/app/models/form.server.model.ts
tshuli: thanks! done

PR #3393, opened 2022-02-09
feat(v2): add (and render) public form date field

### LINKED ISSUE #2618: [React] Public form DateField component
Used to render a date field on public forms.

Blocked by #2006

Should take in a DateField schema object and generate the appropriate validations using react-hook-form

karrui:
## Problem

This PR adds the `DateField` component and renders it in the public form page.
The relevant validations have been added.

Closes #2618 

## Solution

**Features**:
- Add `DateField` component

**Improvements**:
- [feat: allow closing of datepicker when tapping out in mobile](https://github.com/opengovsg/FormSG/commit/84cff69e69b64d0beac2f067caad224f454c3a71)

## Before & After Screenshots
In storybook

## Tests

Unit tests for the field component has been added. 
> Note that we are unable to add date picker selection tests... I think the rendered test block is too big. Jest will timeout. Probably worth it just to do the input validation tests


frontend/src/templates/Field/Date/DateField.tsx
```diff
@@ -0,0 +1,69 @@
...
+    (date: Date) => {
+      const { selectedDateValidation } = schema.dateValidation
+      // All dates available.
+      if (!selectedDateValidation) return false
+
+      switch (selectedDateValidation) {
+        case DateSelectedValidation.NoPast:
+          return isBefore(date, startOfToday())
+        case DateSelectedValidation.NoFuture:
+          return isBefore(startOfToday(), date)
```
mantariksh, line 43: what if somehow the date given is during the day today? then today's date would be made unavailable, when actually it should be available.

would `isAfter(endOfToday(), date)` be more accurate?

frontend/src/utils/fieldValidation.ts
```diff
@@ -285,6 +296,62 @@ export const createCheckboxValidationRules: ValidationRuleFn<
...
+      noFuture: (val) => {
+        if (
+          !val ||
+          schema.dateValidation.selectedDateValidation !==
+            DateSelectedValidation.NoFuture
+        ) {
+          return true
+        }
+        return (
+          isBefore(parseISO(val), startOfTomorrow()) ||
```
mantariksh, line 316: should we make this consistent with `isDateUnavailable`?

frontend/src/utils/fieldValidation.ts
```diff
@@ -285,6 +296,62 @@ export const createCheckboxValidationRules: ValidationRuleFn<
...
+      noPast: (val) => {
+        if (
+          !val ||
+          schema.dateValidation.selectedDateValidation !==
+            DateSelectedValidation.NoPast
+        ) {
+          return true
+        }
+        return (
+          isAfter(parseISO(val), endOfYesterday()) ||
```
mantariksh, line 329: should we also make this consistent with `isDateUnavailable`?

mantariksh: perhaps we can extract this into a helper so future devs don't have to keep wrapping their head around "is `isAfter(parseISO(val), endOfYesterday())` equivalent to `!isBefore(date, startOfToday())`"

mantariksh (approved):

frontend/src/components/DatePicker/DateInput.tsx
```diff
@@ -76,9 +73,6 @@ export const DateInput = forwardRef<DateInputProps, 'input'>(
         <Popover
           placement="bottom-start"
           initialFocusRef={initialFocusRef}
-          // Prevent mobile taps to close popover when doing something like
-          // changing months in the selector.
-          closeOnBlur={!isMobile}
```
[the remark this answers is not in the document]
karrui: was due to a rerender issue the last time, but i have somehow fixed it and now it's not closing 😆

frontend/src/templates/Field/Date/DateField.tsx
karrui: You're right; even though the date that is given is always only in the ISO `YYYY-MM-DD` format, its better to be safe than sorry. Will update.

karrui: Changed it to `isAfter(date, endOfToday())`

PR #3428, opened 2022-02-14
refactor: use combined route validators and handlers

### LINKED ISSUE #1642: Combine validators with handler functions and export them atomically

tshuli:
This PR is a copy of #2819. Reopening to trigger CI/CD 
## Problem
Prevent developers from forgetting the use of validators when using a controller handler.

Closes #1642 

## Solution
As suggested in the issue, put the original handler with a validation middleware in an array, and expose the final array for usage rather than the original handler

**Breaking Changes** 

    - Details ...

**Improvements**:

- Details ...

## Tests
Ran a regression test using `npm run test-backend`, and ran happy-path testing locally with the changed routes

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ npm run dev
 
 After the Docker image has finished building, the application can be accessed at [localhost:5000](localhost:5000).
 
-If there have been no dependency changes in `package.json` or changes in the
+If there are no dependency changes in `package.json` or changes in the
 `src/app/server.ts` file, you can run
 
 ```bash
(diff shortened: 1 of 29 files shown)


karrui (approved):
lgtm, thanks @tehtea again! Sorry we had to rebranch in order to get the CI/CD working 👀

PR #3446, opened 2022-02-16
feat: add script to archive multiple forms

mantariksh:
## Problem

Product ops has to spend time manually archiving multiple forms.

## Solution

Add a function to archive an array of strings containing form IDs. Note that this function is nearly identical to `unlistFormArrayFromExamples` which already exists in the docs. Only the update query is different.

Tested on staging:
Screenshot 2022-02-16 at 1 47 57 PM


karrui (approved):

tshuli (approved):
lgtm

PR #3485, opened 2022-02-23
chore: Update README file for Mac OS Monterey and Apple Silicon Macs

frankchn:
## Problem
README documentation is not updated for Apple Silicon Macs on Mac OS Monterey for the v2 branch.

## Solution
Update README file in the FormV2 branch for Mac OS Monterey and Apple Silicon. Also updates the `nan` package to 2.15 to resolve a build error on Apple Silicon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,14 @@ The current product roadmap includes:
 
 Install [docker and docker-compose](https://docs.docker.com/get-docker/).
 
+### First Setup
+
+Run the following shell command to install relevant npm packages.
+
+```bash
+npm install
+```
+
 ### Running Locally
 
 Run the following shell command to build the Docker image from scratch. This will usually take 10 or so minutes.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -61,7 +69,7 @@ Run the following shell command to build the Docker image from scratch. This wil
 npm run dev
 ```
 
-After the Docker image has finished building, the application can be accessed at [localhost:5000](localhost:5000).
+After the Docker image has finished building, the React application can be accessed at [localhost:3000](localhost:3000).
```
karrui, line 72: May want to also add a note on the port for the server:

```
The server can be accessed at `localhost:5000`.
```

karrui (commented):
you may have used `npm install` with `npm` version > 6, which means the package-lock is now using version 2. May want to use npm v6 in this repo (until we get to upgrading the lockfile version in the repo)

frankchn:
@karrui ah you are right, updated to install with node 14 / npm 6. PTAL again :)

README.md
frankchn: done

karrui (approved):
lgtm

PR #3501, opened 2022-02-24
build: necessary adjustments to allow building in AWS ELB AL2 platform

karrui:
This PR adds the necessary adjustments for AWS ELB AL2 platform.


karrui:
## Release checklist
- [ ] ensure webhook retries can be written and consumed from SQS
- [ ] ensure SP and MyInfo logins work to test EFS, NAT and route tables (cannot test corppass as idt we have production credentials anymore)
- [ ] ensure email notifications are being received from SES (we might need to wait until after DNS cutover to test this as I think we have the full url registered with SES)

timotheeg (approved):

PR #3502, opened 2022-02-24
feat(v2): create droppable empty form placeholder

mantariksh:
Currently the form builder renders an empty page when the form has no fields. This PR adds the placeholder for fields to be dragged in when the form is empty:
empty-form-placeholder

Moreover, previously, background colour of existing fields would change to `neutral.100` when a new field was being dragged over, even though they are not responsive to new fields being dropped over them. This is inconsistent with the [design system](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=2584%3A107188). This PR fixes the issue by preventing the background colour of existing fields from changing when new fields are being dragged over.

## Other improvements
- Mocks added for POST request to create field and PUT request to update field

## Known bugs
These will be fixed in a future PR:
- After creating a new field, the `onSuccess` handler in the `Edit*` component does not fire, because the component unmounts and re-mounts due to its `key` in the parent changing.
- After updating a field and exiting the edit field drawer, the field reverts back to its previous state. When attempting to create a new field, the field flips into the new state again.

### DIFF
--- a/frontend/src/features/admin-form/AdminFormCreatePage.stories.tsx
+++ b/frontend/src/features/admin-form/AdminFormCreatePage.stories.tsx
@@ -5,6 +5,7 @@ import { Meta, Story } from '@storybook/react'
 import {
   createSingleField,
   getAdminFormResponse,
+  updateSingleField,
 } from '~/mocks/msw/handlers/admin-form'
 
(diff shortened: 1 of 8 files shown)


karrui (approved):
LGTM

Issue #3472, opened 2022-02-21
Fix server-side divergences between `form-v2/develop` and `develop` branch


karrui:
This issue keeps track of the various **server-side** divergences between the form-v2/develop and develop branch, so we hopefully only need to migrate infra + react related code instead of also having to check for server-side variances.

- [x] shared/constants/form.ts
![Screenshot 2022-02-21 at 6 58 12 PM](https://user-images.githubusercontent.com/22133008/154944655-63c10adf-44cc-4636-96ac-46a7b2c5bbf2.png)

- [x] shared/types/core.ts
![Screenshot 2022-02-21 at 6 58 40 PM](https://user-images.githubusercontent.com/22133008/154944668-db81e013-a535-4068-baef-efbc6be792d1.png)

- [x] shared/types/form/form.ts
![Screenshot 2022-02-21 at 6 59 21 PM](https://user-images.githubusercontent.com/22133008/154944675-de5d3261-a1dc-4b01-ad6f-878950e759cb.png)

- [x] shared/types/user.ts
![Screenshot 2022-02-21 at 6 59 39 PM](https://user-images.githubusercontent.com/22133008/154944714-5b7549cf-02cb-4220-adb3-e8d125e060b9.png)

- [x] shared/utils/file-validation.ts
export const VALID_EXTENSIONS

- [ ] ~~shared/utils/immutable-array-fns.ts~~
  - Not used in server, can keep it in react branch for now
Add 
```ts
export const insertAt = <T>(
  array: ExtractTypeFromArray<T>[],
  index: number,
  valueToInsert: ExtractTypeFromArray<T>,
): ExtractTypeFromArray<T>[] => {
  return [...array.slice(0, index), valueToInsert, ...array.slice(index)]
}
```

- [x] Add positional API (documented in #3135 
  - src/app/models/form.server.model.ts
  - src/app/modules/form/admin-form/admin-form.controller.ts
  - src/app/modules/form/admin-form/admin-form.service.ts
  - src/types/form.ts

- [x] Add env var route (documented in #2878)
  - src/app/modules/frontend/frontend.controller.ts
  - src/app/modules/frontend/frontend.routes.ts
  - src/app/modules/frontend/frontend.service.ts
  - src/app/routes/api/v3/client/client.routes.ts

![Screenshot 2022-02-21 at 7 08 17 PM](https://user-images.githubusercontent.com/22133008/154944844-fadbf81a-5f05-410d-91f0-4d3591f2f303.png)
[...]

PR #3504, opened 2022-02-24
feat: export shared types for use in client/backend, update user route error to return new status code, update SettingsDto return

karrui:
## Problem

This PR is part of #3472, where types that can be shared with client/server are exported.
In addition, update user route's `InvalidOtpError` to return `404` instead of `401` status code, due to React app logging user out automatically when 401 is retrieved.

Also also updates the fields SettingsDto returns to:
- Include `responseMode` in returned JSON
- include `publicKey` field in storage mode forms


timotheeg (approved):
lgtm 👍

PR #3513, opened 2022-02-25
feat: disable field creation on useAdminForm error

mantariksh:
Currently, fields in the field creation drawer are disabled only as long as the admin form is still loading. This means that if there is an error while loading the admin form, the fields can be dragged. We should only let the form be editable if it has already loaded successfully.


karrui (approved):
Lgtm

PR #3515, opened 2022-02-25
feat(v2): render EmptyWorkspace page when admin has no forms

karrui:
## Problem

This PR adds the EmptyWorkspace component that will be shown when admin has no forms.

## Solution

**Features**:

- feat: render EmptyWorkspace component when admin has no forms

**Improvements**:

- ref(WorkspacePage): move all logic into useWorkspaceForms hook
- feat: move CreateFormModal out to WorkspacePage

## Before & After Screenshots
New empty workspace stories have been added


timotheeg (approved):
lgtm 👍

PR #3519, opened 2022-02-25
wip: add figma storybook addon

karrui:
@pearlyong wanted to play around with linking Storybook with Figma frames. This is a WIP PR for her to look and play with


karrui:
Will do so in [opengovsg/design-system](https://github.com/opengovsg/design-system) instead

PR #3524, opened 2022-02-28
fix(ci/deploy): truncate deploy description to 180 chars

mantariksh:
## Problem

Elastic Beanstalk application versions have a character limit of 200 for descriptions. Our deploy pipeline contains a line to grab the latest commit message and truncate it to 180 characters (#3147) but it uses the `cut` command which only cuts the specified number of bytes for each line. This means there can still be [deploy failures](https://github.com/opengovsg/FormSG/runs/5353731644?check_suite_focus=true) if the total number of characters across lines exceeds 200.

## Solution

Change to `head -c 180` which gets the first 180 characters from the specified text.


karrui (approved):
lgtm. thanks for catching this!

Issue #3528, opened 2022-03-01
Update release workflows to work on release and staging branches


karrui:
For releasing production or staging builds on AWS ELB AL2 platforms currently, we need to push specifically to `release-al2` and `staging-al2` branches.

When this is stable, we should update the workflows so we can push to `staging` and `release` branches on Github and remove the branched deployment branches.

PR #3529, opened 2022-03-01
build: merge release-al2 branch back into develop

karrui:
This branch merges `release-al2` back into `develop`.

In the future, a PR will be opened so we release from the `release` branch instead. This is tracked by #3528


karrui:
@mantariksh not sure if we want to keep the hotfix in

mantariksh:
@karrui I think it might be better to revert the hotfix in `release-al2` then merge the whole thing back into `develop` so the logging doesn't get persisted but there's a record of it, wdyt?

karrui:
Will do that now.

karrui:
As discussed, will merge this into develop, then revert on develop for next release.

timotheeg (approved):

karrui:
also discussed with @timotheeg, no need to revert ba; the check is cheap and this prevents us from having to do this same thing in the future

PR #3532, opened 2022-03-01
chore: Add a new sentence about giving docker more RAM on Mac OS

frankchn:
Add a new sentence about giving docker more RAM on Mac OS or else it may not build the existing Angular based frontend.


karrui (approved):
did it work for you after the change, @frankchn?

frankchn:
@karrui Works now :) Thanks!

PR #3219, opened 2021-12-22
fix: revert mongoose v6.0.13 and related dependency update, update mongoose to v5.13.13

karrui:
Small breakage in Email mode form's setters and getters that caused a rollback to v5.38.0 on prod, will fix in future PR

This reverts commit d5ada10edcfe4c03bdf001ea5b75137ba80307fb.

This should be merged in to push a hotfix

PR #3541, opened 2022-03-03
fix: revert how we initiate mongo database in docker-compose local dev

karrui:
Since we've rolled back to v5 in #3219, this is unneeded and actually causes fresh builds to fail.
First introduced in https://github.com/opengovsg/FormSG/pull/3039

Also update the mongo image we use in docker-compose to 4.2 to achieve consistent with production version


docker-compose.yml
```diff
@@ -107,25 +107,18 @@ services:
     network_mode: 'service:formsg' # reuse formsg service's network stack so that it can resolve localhost:5156 to mockpass:5156
 
   database:
-    image: 'mongo:4.0'
+    image: 'mongo:4.2'
```
timotheeg, line 110: Is something stopping us from jumping to 4.4?

karrui: you mean on prod? nope; just need to schedule a time...

timotheeg (approved):

PR #3561, opened 2022-03-09
fix(v2/Dropdown): allow controlled selectedItems, render icon in MultiSelectItem

karrui:
## Problem

This PR cherry picks a bunch of `SingleSelect`/`MultiSelect` fixes that were encountered whilst working on #3556.

## Solution

**Features**:

- feat(Select): allow setting of empty placeholder
- feat(MultiSelectItem): show icon in tag if it exists

**Bug Fixes**:

- fix(ComboboxClearButton): show active state if item has been selected
  - Previously, the clear button will remain in inactive styling even if item has been selected
- fix(Single/MultiSelectProvider)!: allow control of selectedItem, refilter items whenever `items` prop changes
  - Instead of using `initialSelectedItem` prop, allow users of the component to pass in current value for selection.
- fix(SingleSelectProvider): use label string as aria-live announcement 
  - the value was used previously, which may cause screenreaders to read something different that what is being rendered.

## Before & After Screenshots
Most stories should not have a change. The only changes should be the ones for rendering an icon in the selected items in `MultiSelect`.

### DIFF
--- a/frontend/src/components/Dropdown/SingleSelect/SingleSelectProvider.tsx
+++ b/frontend/src/components/Dropdown/SingleSelect/SingleSelectProvider.tsx
@@ -36,7 +36,7 @@ export const SingleSelectProvider = ({
   name,
   filter = defaultFilter,
   nothingFoundLabel = 'No matching results',
-  placeholder = 'Select an option',
+  placeholder: placeholderProp,
   clearButtonLabel = 'Clear dropdown input',
   isClearable = true,
   isSearchable = true,
@@ -51,6 +51,11 @@ export const SingleSelectProvider = ({
 }: SingleSelectProviderProps): JSX.Element => {
   const { items, getItemByValue } = useItems({ rawItems })
   const [isFocused, setIsFocused] = useState(false)
+
+  const placeholder = useMemo(() => {
+    if (placeholderProp === null) return ''
+    return placeholderProp ?? 'Select an option'
+  }, [placeholderProp])
 
   const getFilteredItems = useCallback(
     (filterValue?: string) =>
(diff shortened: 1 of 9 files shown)


frontend/src/components/Dropdown/components/SelectCombobox/ComboboxClearButton.tsx
```diff
@@ -26,7 +27,7 @@ export const ComboboxClearButton = (): JSX.Element | null => {
       aria-label={clearButtonLabel}
       onClick={handleClearSelection}
       __css={styles.clearbutton}
-      color={inputValue ? 'secondary.500' : undefined}
+      color={inputValue || selectedItem ? 'secondary.500' : undefined}
```
tshuli, line 30: how about
```suggestion
      {...(inputValue ? { color: 'secondary.500' } : {})}
```

instead of passing in `undefined` to `color` prop

frontend/src/components/Dropdown/SingleSelect/SingleSelectProvider.tsx
```diff
@@ -171,7 +171,7 @@ export const SingleSelectProvider = ({
     if (inputAriaProp) return inputAriaProp
     let label = 'No option selected'
     if (selectedItem) {
-      label = `Option ${itemToValue(selectedItem)}, selected`
+      label = `Option ${itemToLabelString(selectedItem)}, selected`
```
tshuli, line 174: Should we remove the comma?

frontend/src/components/Dropdown/SingleSelect/SingleSelectProvider.tsx
```diff
@@ -148,7 +171,7 @@ export const SingleSelectProvider = ({
...
```
tshuli, line 174: Should we remove the comma?

```suggestion
      label = `Option ${itemToLabelString(selectedItem)} selected`
```

tshuli (approved):
minor nits but lgtm otherwise!

frontend/src/components/Dropdown/SingleSelect/SingleSelectProvider.tsx
karrui: I think comma adds a pause in screen reader so we should keep it

frontend/src/components/Dropdown/components/SelectCombobox/ComboboxClearButton.tsx
karrui: No difference and `undefined` makes it easier to read imho

PR #3576, opened 2022-03-14
perf(v2): massively optimize public form filling responsiveness

karrui:
## Problem

Perf was HORRENDOUS due to global formState mutating on literally every single error, causing each keystroke to render every single field again.

This PR updates the performance by only monitoring a subslice of the form state, so each field does not rerender from other field changes.

Also adds rendering for Dropdown and Date fields in public form view, must have been removed again due to merge conflicts.

See the comparison videos below for the massive responsiveness boost.

## Solution

**Improvements**:

- perf: use `useFormState` and `useWatch` over contextual counterparts
- perf: remove `dep` in validation, only trigger on OtherCheckbox/OtherRadio option change

**Bug Fixes**:

- fix(checkboxValidation): do not run length validation if empty value

## Before & After Screenshots

**BEFORE**:
horrendous public form rendering perf

**AFTER**:

Recording 2022-03-15 at 04 07 48

## Tests

Should have no changes in tests

## Deploy Notes

**New dependencies**:

- fix(deps): update react-hook-form to v7.28.0
  - needed to use the new `useFormState` hook, which allows monitoring substates of form state to gain the much better rendering speeds.

### DIFF
--- a/frontend/src/features/public-form/components/FormFields/FieldFactory.tsx
+++ b/frontend/src/features/public-form/components/FormFields/FieldFactory.tsx
@@ -7,7 +7,9 @@ import { FormColorTheme } from '~shared/types/form'
 import {
   AttachmentField,
   CheckboxField,
+  DateField,
   DecimalField,
+  DropdownField,
   EmailField,
   HomeNoField,
   ImageField,
@@ -61,6 +63,10 @@ export const FieldFactory = memo(
         return <LongTextField schema={field} colorTheme={colorTheme} />
       case BasicField.YesNo:
         return <YesNoField schema={field} colorTheme={colorTheme} />
+      case BasicField.Dropdown:
+        return <DropdownField schema={field} colorTheme={colorTheme} />
+      case BasicField.Date:
+        return <DateField schema={field} colorTheme={colorTheme} />
       case BasicField.Uen:
         return <UenField schema={field} colorTheme={colorTheme} />
       case BasicField.Attachment:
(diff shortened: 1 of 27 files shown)


frontend/src/templates/Field/Checkbox/CheckboxField.tsx
```diff
@@ -34,43 +38,35 @@ export const CheckboxField = ({
...
           !!value ||
           'Please specify a value for the "others" option'
         )
       },
     }),
-    [isOthersSelected],
+    [checkboxInputName, getValues],
   )
 
-  const othersInputError: FieldError | undefined = get(errors, othersInputName)
```
timotheeg, line 72: How does removing this reference help here?

frontend/src/features/public-form/components/FormFields/FieldFactory.tsx
```diff
@@ -61,6 +63,10 @@ export const FieldFactory = memo(
         return <LongTextField schema={field} colorTheme={colorTheme} />
       case BasicField.YesNo:
         return <YesNoField schema={field} colorTheme={colorTheme} />
+      case BasicField.Dropdown:
```
timotheeg, line 66: >  must have been removed again due to merge conflicts

Let's discuss this during next sync. We need to be very careful with merge conflicts 😰

timotheeg (commented):
Thanks for making a big perf improvement! I do not understand how not using a local const helps with the perf changes 🤔 . Can I book you later to do a quick pair review and discuss it?

timotheeg (approved):
lgtm

frontend/src/templates/Field/Checkbox/CheckboxField.tsx
karrui: As discussed, it does nothing ><. Just moved it in since we get nothing from declaring it outside

PR #3577, opened 2022-03-15
feat(v2/TableField): render Table field dropdown column variant

karrui:
## Problem

This PR adds the dropdown column variant to Table fields.
Had to update dropdown menu's popper configuration to prevent the dropdown options from being clipped by the TableField container due to the presence of `overflowX`.

## Solution

**Features**:

- feat(SelectPopover): add fixed strategy, constrain width to ref elem
- feat: render dropdown column in Table fields

## Before & After Screenshots
Public form stories should be updated to display dropdown columns.


karrui:
Merging merge conflicts before reopening for review

frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -139,6 +142,13 @@ export const TableField = ({
           </Tbody>
         </Table>
       </Box>
+      {isMobile ? null : (
```
mantariksh, line 145: do we not show the error in mobile?

frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -139,6 +142,13 @@ export const TableField = ({
...
+        // Error will only render if table field has errors. Since the only possible errors are
+        // for required fields (for now), just render a hardcoded error message.
+        <FormErrorMessage my="0.75rem">
+          Please fill in the required fields.
```
mantariksh, line 149: what might be a more sustainable solution here? perhaps `unique` the error messages and display all?

frontend/src/templates/Field/Table/TableField.tsx
mantariksh: saw the code in `ColumnCell` - would be good to add a comment here to explain!

mantariksh (approved):

frontend/src/components/Dropdown/components/SelectPopover/SelectPopover.tsx
```diff
@@ -12,11 +12,57 @@ export const SelectPopoverProvider: FC = ({ children }): JSX.Element => {
...
+            state.placement.startsWith('right')
+              ? 'height'
+              : 'width'
+
+          if (!popperElement) return
+
+          const popperSize =
+            popperElement[
+              `offset${
+                widthOrHeight[0].toUpperCase() + widthOrHeight.slice(1)
```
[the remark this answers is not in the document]
karrui: Indeed. The perils of copy-pasting code on the INTERWEBS. Will update.

frontend/src/templates/Field/Table/TableField.tsx
karrui: I fear that would be too confusing. I will unique and show the first error I guess. Will eventually allow users to fix all the errors.

frontend/src/templates/Field/Table/TableField.tsx
karrui: 2da28ab

frontend/src/templates/Field/Table/TableField.tsx
karrui: [2da28ab](https://github.com/opengovsg/FormSG/pull/3577/commits/2da28ab18d00018a43e933e7138daf792b52a956)

frontend/src/components/Dropdown/components/SelectPopover/SelectPopover.tsx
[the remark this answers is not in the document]
karrui: Didn't need this in the end, found a performance bug that caused too many rerenders when viewport width changed. Used[ ChakraUI's variant](https://github.com/chakra-ui/chakra-ui/blob/main/packages/popper/src/modifiers.ts#L9) instead. 

(Reason why we didn't use ChakraUI's own usePopper was due to their warning that the hook is undocumented and is subject to change...)
(also realize we ignored that when using FormControlProps LOL)

PR #3581, opened 2022-03-16
feat(v2): edit checkbox field in form builder

mantariksh:
This PR reimplements the checkbox field editor originally implemented in #3131, with the refactors from #3547 included. There are a few significant differences from the original implementation:
- The design of the checkbox editor has changed from two tabs to a single tab containing all the fields.
- The original `ValidationOptions` validation did not allow one of `customMin` or `customMax` to be set without setting the other. This PR fixes this.

## Known issues
1. ~When the sidebar drawer is closed, clicking on a field results in the drawer opening to the field list, rather than directly to the editing page for that field. This will be addressed separately.~ fixed in #3583 

## Next steps
Field duplication and deletion will be implemented together with the relevant mocks before submitting the form builder for design review.

While design review is happening, I will concurrently work on the edit pages for the other fields.

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
@@ -37,8 +37,8 @@ type EditCheckboxKeys = typeof EDIT_CHECKBOX_FIELD_KEYS[number]
 type EditCheckboxInputs = Pick<CheckboxFieldBase, EditCheckboxKeys> & {
   fieldOptions: string
   ValidationOptions: {
-    customMin?: number
-    customMax?: number
+    customMin?: number | string
+    customMax?: number | string
   }
 }
 
@@ -47,10 +47,10 @@ const transformCheckboxFieldToEditForm = (
 ): EditCheckboxInputs => {
   const nextValidationOptions = field.validateByValue
     ? {
-        customMin: field.ValidationOptions.customMin || undefined,
-        customMax: field.ValidationOptions.customMax || undefined,
+        customMin: field.ValidationOptions.customMin || '',
+        customMax: field.ValidationOptions.customMax || '',
       }
-    : { customMin: undefined, customMax: undefined }
+    : { customMin: '', customMax: '' }
   return {
     ...pick(field, EDIT_CHECKBOX_FIELD_KEYS),
     fieldOptions: SPLIT_TEXTAREA_TRANSFORM.input(field.fieldOptions),
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
@@ -1,0 +1,279 @@
+import { useEffect, useMemo } from 'react'
+import { Controller, RegisterOptions } from 'react-hook-form'
+import { Box, FormControl, Stack } from '@chakra-ui/react'
+import { extend, isEmpty, pick } from 'lodash'
+
+import { CheckboxFieldBase } from '~shared/types/field'
+
+import { createBaseValidationRules } from '~utils/fieldValidation'
+import FormErrorMessage from '~components/FormControl/FormErrorMessage'
+import FormLabel from '~components/FormControl/FormLabel'
+import Input from '~components/Input'
+import NumberInput from '~components/NumberInput'
+import Textarea from '~components/Textarea'
+import Toggle from '~components/Toggle'
+
+import { DrawerContentContainer } from './common/DrawerContentContainer'
+import { FormFieldDrawerActions } from './common/FormFieldDrawerActions'
+import { EditFieldProps } from './common/types'
+import { useEditFieldForm } from './common/useEditFieldForm'
+import {
+  SPLIT_TEXTAREA_TRANSFORM,
+  SPLIT_TEXTAREA_VALIDATION,
+} from './common/utils'
+
+type EditCheckboxProps = EditFieldProps<CheckboxFieldBase>
+
+const EDIT_CHECKBOX_FIELD_KEYS = [
+  'title',
+  'description',
+  'required',
+  'othersRadioButton',
+  'validateByValue',
+] as const
+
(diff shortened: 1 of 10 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
```diff
@@ -0,0 +1,279 @@
...
+    }),
+    [watchedInputs],
+  )
+
+  // Effect to clear validation option errors when selection limit is toggled off.
+  useEffect(() => {
+    if (!watchedInputs.validateByValue) {
+      clearErrors('ValidationOptions')
+    }
+  }, [clearErrors, watchedInputs.validateByValue])
```
karrui, line 184: nit: maybe we can use the [`deps`](https://react-hook-form.com/api/useform/register#:~:text=Validation%20will%20be%20triggered%20for%20the%20dependent%20inputs) prop in `register` instead of a useEffect to achieve the same result

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/useEditFieldForm.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  useDebounce(
+    () => handleChange(transform.output(watchedInputs as FormShape, field)),
+    300,
+    Object.values(watchedInputs),
+  )
+
+  const handleUpdateField = editForm.handleSubmit((inputs) => {
+    const updatedFormField: FieldShape = transform.output(
+      inputs as FormShape,
```
karrui, line 59: do you need to cast `FormShape` seeing how that type is already passed into `useForm` as a generic?

Same for the other casts

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/useEditFieldForm.ts
```diff
@@ -0,0 +1,93 @@
...
+    )
+    return handleSave(updatedFormField, {
+      onSuccess: (newField) => {
+        editForm.reset(
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          transform.input(newField),
+          // react-hook-form has strange behaviour where setting a field
+          // to undefined results in the form displaying the field in
+          // a previous state. Hence don't reset the form values.
```
karrui, line 70: `reset` resets the form to its initial state with `defaultValues`, which has been populated above; which may be why the field is seemingly set to the previous state. May just need to reset everything to empty strings instead of `undefined`?.

karrui (approved):
lgtm, much cleaner!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
mantariksh: tried this, unfortunately it didn't work, I think because the input value remains the same. even if the validation is re-triggered. e.g. if I put 0 as the min selection limit (invalid because less than 1) then I deselect the "Selection limits" toggle, the input will still be invalid.

I also tried using `Controller` and a custom `onChange` to unset the `customMin` and `customMax` when Selection Limits is toggled off, but ran into the same bug where setting values to `undefined` results in form getting set back to previous state. then tried setting to empty string instead of `undefined`, but ran into type errors (since these are supposed to be numbers) and also the errors still didn't get cleared even after running `trigger`, possibly because the `rules` only get updated on the following render.

think `useEffect` might be the best solution for now! wdyt?

karrui: Yea, can continue using useEffect then! But note that it may be hiding some bugs in that case haha

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/useEditFieldForm.ts
mantariksh: yeah so `inputs` is of type `UnpackNestedValue<FormShape>`, not `FormShape`. I tried looking into what the `UnpackNestedValue` type does, but figured it was probably too much effort to try to understand. do you know what's going on here?

````typescript
export declare type NestedValue<TValue extends object = object> = {
    [$NestedValue]: never;
} & TValue;
export declare type UnpackNestedValue<T> = T extends NestedValue<infer U> ? U : T extends Date | FileList | File ? T : T extends object ? {
    [K in keyof T]: UnpackNestedValue<T[K]>;
} : T;
```

karrui: This could be because of your defaultValues type in line 42. Any reason why it's casted to UnpackNestedValue?

karrui: This could be because of your defaultValues type in line 42. Any reason why it's casted to UnpackNestedValue?

mantariksh: yeah it's casted because the type required for `defaultValues` is also `UnpackNestedValue`

mantariksh: managed to remove the casts by changing the types in `UseEditFieldFormProps`

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/useEditFieldForm.ts
mantariksh: thanks, empty string seems to work! [1ea6867](https://github.com/opengovsg/FormSG/pull/3581/commits/1ea686727ac12b6d0ec500bed7c812a7852f74e1)

PR #3585, opened 2022-03-16
feat(v2): add editors for Yes/No, NRIC, UEN

mantariksh:
Adds edit drawers for Yes/No, NRIC and UEN fields. These fields were all done at the same time as they require no additional configuration other than setting the "Required" toggle, so the UI is exactly the same for all 3.

I chose to copy-paste the code instead of having a common component for all 3 as this allows the types to be tighter, i.e. tied to each `fieldType`. 

Depends on #3581


karrui (approved):
lgtm

tshuli (approved):

PR #3593, opened 2022-03-17
fix(v2/shared): update type shape for TableFieldDto

karrui:
## Problem

Previously, `TableFieldDto` was equivalent to `FormWithWithId<TableFieldBase>`, which only added the `_id` prop to table fields. However, `TableField#columns` was also a schema with an `_id`, which was not augmented by `FormFieldWithId<F>`.

> We technically had a hint of type incompatibility when I had to cast `FormFieldWithId<TableFieldBase>` with a new schema type. Just fixing it here now.

This PR updates `TableFieldDto` to include `column#_id`. Whilst doing so, `useBuilderFields` had to be updated to create a temporary object with column `_id`s populated  when attempting to create a new table field. This is done using `cuid`.

## Solution

**Bug Fixes**:

- fix: update FormFieldWithId type to assign id to TableField columns

## Dependencies
- add cuid package to generate collision resistant ids for table field column ids.


tshuli (approved):
lgtm

Issue #3570, opened 2022-03-14
[React/PublicForm] enable form submission feature


karrui:
Currently submitting public forms only print the submission object in console. We should transform the object and call the backend API for submission

Sub features
- [x] #3593
- [x] #3595
- [x] #3596
- [x] #3598
- [x] #3434
- [x] #3602
- [x] #3603
- [x] #3630
- [x] #3670

PR #3594, opened 2022-03-17
refactor(v2/submit-1): pass input types to react-hook-form generics for better validation

karrui:
> Note that this PR is built upon #3593

## Problem

Various form fields in our public form were not strongly typechecked prior to this PR, which may result in unforeseen type errors.

This PR adds the correct types to various form fields so type errors can be mitigated.

Related to #3570

## Solution

**Features**:

- feat(types): add field input types for better type checking in rhf
- feat: pass field input types into react-hook-form functions (that accept generic types) for stronger type checks.

## Before & After Screenshots
Should have no difference.


tshuli (approved):
lgtm

karrui:
Waiting for top down squash (rest of the chain to complete) before I merge this in

mantariksh (approved):

PR #3597, opened 2022-03-17
refactor(v2/submit-1): transform form inputs to FieldResponse shape

### LINKED ISSUE #3595: [React/PublicForm] Transform rhf inputs to expected FieldResponse output

karrui:
## Problem

Backend expects each individual form field response to be of `FieldResponse` shape, and thus a transformation layer is required.
This PR adds that transformation layer.
Related to #3570, closes #3595

## Solution

**Features**:

- feat(shared): add `removeAt` immutable array fn util
- feat(inputValidation): add fns to validate react hook form inputs
- feat(inputTransformatio): add input transformation to field responses functions
- feat: use transformInputsToOutputs function to process submission

**Improvements**:

- ref: move all field schema types to types file
- feat(shared): add generic to FormFieldDto type

## Before & After Screenshots
No UI changes


frontend/src/features/public-form/utils/inputValidation.ts
```diff
@@ -0,0 +1,94 @@
...
+  TableFieldValues,
+  VerifiableFieldValues,
+} from '~templates/Field/types'
+
+/**
+ * Regex to validate yyyy-mm-dd strings.
+ * @note this does not validate that the date is valid, just that the format is correct.
+ */
+const ISO_8601_DATE_ONLY_REGEX =
+  /^\d{4}-([0][1-9]|1[0-2])-([0-2][1-9]|[1-3]0|3[01])$/
```
tshuli, line 15: If this regex was taken from somewhere else, could we add the source as a comment, just for future reference please

tshuli (approved):
minor comments but lgtm

frontend/src/features/public-form/utils/inputValidation.ts
```diff
@@ -0,0 +1,94 @@
...
+export const validateTableInput = (
+  input: unknown,
+): input is TableFieldValues => {
+  return (
+    Array.isArray(input) &&
+    input.every((rowData) => {
+      if (typeof rowData !== 'object') return false
+      return Object.keys(rowData).every(
+        (colId) => typeof rowData[colId] === 'string',
+      )
```
mantariksh, line 32: this will return `true` even if `rowData` is an array right?

frontend/src/features/public-form/utils/inputValidation.ts
```diff
@@ -0,0 +1,94 @@
+import {
```
mantariksh, line 1: what's the value of this validation layer over what we're doing in `validateResponses` in #3600? it seems rather complicated to validate the answers in the form, then validate them again on submission, then reshape them, then validate them again in `validateResponses`.

I think `EncryptionService` made sense as a final check before end-to-end encryption, but personally I'm not sure this validation adds much value - we're just re-validating outputs which are already in our control.

frontend/src/features/public-form/utils/inputTransformation.ts
```diff
@@ -0,0 +1,237 @@
...
+): VerifiableAnswerOutput<F> => {
+  if (
+    !validateVerifiableInput(input) ||
+    (schema.isVerifiable && !input.signature)
+  ) {
+    throw new InputValidationError(schema._id, input)
+  }
+  return {
+    ...pickBaseOutputFromSchema(schema),
+    answer: input.value,
```
mantariksh, line 72: nit: maybe we can consider this as "inheriting" from `SingleAnswer`?
```suggestion
    ...transformToSingleAnswerOutput(schema, input),
    answer: input.value,
```

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -106,19 +109,40 @@ export const PublicFormProvider = ({
...
 
+  const handleSubmitForm = useCallback(
+    (formInputs: Record<FormFieldDto['_id'], unknown>) => {
+      if (!cachedDto?.form) return
+      try {
+        const responses = cachedDto.form.form_fields
+          .map((ff) => transformInputsToOutputs(ff, formInputs[ff._id]))
+          .filter((output): output is FieldResponse => output !== undefined)
+
+        return console.log(responses)
```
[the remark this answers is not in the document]
karrui: This is a temporary log to simulate submission; will remove when we call the API for realsies

frontend/src/features/public-form/utils/inputValidation.ts
karrui: the original intention of this validation layer is meant to act as a typeguard so the correct types are inferred after we pass it through this layer. Since a typeguard is going to be written, I thought we might as well correctly validate it.

But now that you say that, we can also create a proper union type that maps `BasicField` to their react-hook-form `FieldValues`. I will experiment to see if I can drop inputValidation layer completely.

frontend/src/features/public-form/utils/inputValidation.ts
karrui: yea, will add a guard

frontend/src/features/public-form/utils/inputTransformation.ts
karrui: not possible, since `SingleAnswerInput === string`, which is not the case for `VerifiableInput`, which has the shape 
```ts
{ value: string, signature: string }
```

frontend/src/features/public-form/utils/inputValidation.ts
karrui: I think the only way is via casting, do you think that's ok?

```ts
/// inputTransformation.ts
export const transformInputsToOutputs = (
  field: FormFieldDto,
  input: unknown,
): FieldResponse | undefined => {
  switch (field.fieldType) {
    case BasicField.Section:
      return transformToSectionOutput(field)
    case BasicField.Checkbox:
      return transformToCheckboxOutput(field, input as CheckboxFieldValues) // <--
...
}
```

Then we can remove the validation layer. But I feel like it would not be that big of a deal to pass it through the layer as well.

mantariksh: would it help if we pass generic types as per this comment? https://github.com/opengovsg/FormSG/pull/3623#discussion_r841678908

rn it seems like we are doing validation 4 times on the frontend:
1. `react-hook-form` validators
2. validate every single field for form logic upon every single change to the form
3. validate every single field upon form submission
4. validate every single response after it gets transformed from a field

it seems like step 2 and 3 are redundant, and (not sure about the magnitude but) might cause a performance hit on large forms bc we'll be validating every single form field for every change (e.g. character typed).

karrui: Done. Using type casting instead. 5e1d5dd

frontend/src/features/public-form/utils/inputValidation.ts
karrui: inputValidation utils deleted.

frontend/src/features/public-form/utils/inputValidation.ts
karrui: `inputValidation` util deleted.

PR #3611, opened 2022-03-18
feat(v2): update styling of admin logic content page

karrui:
## Problem

Logic design has changed. Update client to look like design.

Storybook should reflect new design


tshuli (approved):
lgtm

Issue #3603, opened 2022-03-17
[React/PublicForm] add logic support for public forms


karrui:
Requires both logic types to be supported:
- [x] Show fields (completed in #3623)
- [x] Prevent submission (tracked by #3813)

PR #3623, opened 2022-03-21
feat(v2/submit-5): show or hide fields based on current form logic in public form

karrui:
> This PR builds on #3600, which is part of public form submission epic

## Problem

This PR adds partial logic handling in public forms, specifically show/hide fields logic. The other logic type (prevent submission logic) will come in a separate PR.

To achieve that, a new `features/logic` folder is created in the react app directory to hold all logic related components and utilities.

Related to #3603

## Solution

**Features**:

- feat(YesNoField): update output of field to be "Yes" or "No" strings
  - This is because logic criteria is hardcoded to "Yes" or "No" strings, and it's probably easier to standardize in Yes/No field than change both client/serverside logic logic.
  - From this, also add `YesNoField`-specific validation
- allow undefined inputs to pass validation
  - some fields may not have any corresponding value, and should still pass validation so the output can be correctly transformed.
- feat(logic): add `groupLogicUnitByFields` and `getVisibleFieldIds` utility functions.
  - Cannot reuse the "shared" util fns as the input argument shapes are different. Will probably be able to extract common sub-util functions in the future.
- feat(FormFields): unregister form fields when they are unmounted
  - So previously entered field values will not appear in form submission if the fields are hidden due to logic.
- feat: add VisibleFormFields to render visible fields based on logic

**Improvements**:
- ref(TableField/utils): extract usage of createTableRow to own fn
- ref: move isShowFieldsLogic typeguard to logic feature directory

**Bug Fixes**:

- Details ...

## Before & After Screenshots
No difference yet, no logic in stories for now.


frontend/src/features/logic/utils/getVisibleFieldIds.ts
```diff
@@ -0,0 +1,205 @@
...
+} from '~features/public-form/utils'
+
+import { FieldIdSet, FieldIdToType } from '../types'
+
+import { groupLogicUnitsByField } from './groupLogicUnitsByField'
+
+const getCurrentFieldValue = (input: unknown, fieldType: BasicField) => {
+  switch (fieldType) {
+    case BasicField.Checkbox:
+      return validateCheckboxInput(input) ? input : null
```
mantariksh, line 25: do we really need to validate in this function? doesn't this mean we are running validation every single time any change is made to the form?

mantariksh: not just validation but we are re-validating outputs which are completely in our own control

frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { FormFieldDto } from '~shared/types/field'
+import { FormColorTheme, LogicDto } from '~shared/types/form'
+
+import { getVisibleFieldIds } from '~features/logic/utils'
+
+import { FieldFactory } from './FieldFactory'
+
+interface VisibleFormFieldsProps {
+  control: Control<FieldValues>
```
mantariksh, line 12: would it be possible to avoid the repeated validations by passing a more specific type here?

frontend/src/features/public-form/utils/inputTransformation.ts
```diff
@@ -62,41 +66,55 @@ const transformToVerifiableOutput = <
   input: unknown,
 ): VerifiableAnswerOutput<F> => {
   if (
-    !validateVerifiableInput(input) ||
-    (schema.isVerifiable && !input.signature)
+    input !== undefined &&
```
mantariksh, line 69: why do you need this here when `validateVerifiableInput` already has a check for `typeof input !== 'object'`?

frontend/src/features/public-form/utils/inputTransformation.ts
```diff
@@ -62,41 +66,55 @@ const transformToVerifiableOutput = <
...
+    signature: input?.signature,
   }
 }
 
 const transformToSingleAnswerOutput = <F extends FormFieldDto>(
   schema: F,
   input: unknown,
 ): SingleAnswerOutput<F> => {
-  if (!validateSingleAnswerInput(input)) {
+  if (input !== undefined && !validateSingleAnswerInput(input)) {
```
mantariksh, line 86: why do you need the undefined check when `validateSingleAnswerInput` already checked for `typeof input === 'string'`?

mantariksh: same for all the other `undefined` checks

frontend/src/features/logic/utils/groupLogicUnitsByField.ts
```diff
@@ -0,0 +1,118 @@
+import { FormCondition, LogicDto } from '~shared/types/form'
```
mantariksh, line 1: I see we're not reusing the shared logic module. this means we're going to have two nearly identical but slightly different logic modules, one each for the frontend and backend.

is this really necessary or is it possible to reuse the existing logic module? if we want to refactor (e.g. by using `reduce` in `groupLogicUnitsByField`) then it will be easy to do later.

frontend/src/features/logic/utils/getVisibleFieldIds.ts
```diff
@@ -0,0 +1,205 @@
...
+ * Gets the IDs of visible fields in a form according to its responses.
+ * This function loops through all the form fields until the set of visible
+ * fields no longer changes. The first loop adds all the fields with no
+ * conditions attached, the second adds fields which are made visible due to fields added in the previous loop, and so on.
+ * @param formInputs the field responses to retrieve logic units for.
+ * @param formProps the form fields and form logics to determine visible field ids.
+ * @returns a set of IDs of visible fields in the submission
+ */
+export const getVisibleFieldIds = (
+  formInputs: Record<FormFieldDto['_id'], unknown>,
```
mantariksh, line 163: would it not be possible to reuse the existing logic module, perhaps with an adapter layer?

frontend/src/templates/Field/YesNo/YesNoField.tsx
```diff
@@ -9,12 +9,20 @@ import { createBaseValidationRules } from '~/utils/fieldValidation'
...
 export interface YesNoFieldProps extends BaseFieldProps {
   schema: YesNoFieldSchema
 }
 
+const transform = {
+  fieldToInput: (value?: YesNoFieldValue) => {
+    if (value === undefined) return
+    return value === 'Yes' ? 'yes' : 'no'
+  },
+  inputToField: (value: 'yes' | 'no') => (value === 'yes' ? 'Yes' : 'No'),
```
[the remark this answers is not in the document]
karrui: We can, I just didn't want to change the base field... but I guess we can haha

frontend/src/features/logic/utils/getVisibleFieldIds.ts
```diff
@@ -0,0 +1,205 @@
...
+    case BasicField.Nric:
+    case BasicField.Uen:
+    case BasicField.YesNo:
+      return validateSingleAnswerInput(input) ? input : null
+    case BasicField.Attachment:
+    case BasicField.Statement:
+    case BasicField.Section:
+    case BasicField.Image:
+    case BasicField.Table:
+      // Field types not used for logic.
```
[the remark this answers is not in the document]
karrui: You're right; more like those fields have no value to be used for determining logic fulfilment. Will update doc

frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
karrui: i don't think so; since the key is just `fieldId`, which has nothing to do with the field values...

frontend/src/features/logic/utils/groupLogicUnitsByField.ts
karrui: the shared logic is not really shared; we will need to move functions in `src/shared/logic` folder into `/shared/logic`. 
I guess I can move all `features/logic/utils` utility functions into `shared/logic` directly first.

Was trying to keep the PR size small

frontend/src/features/logic/utils/getVisibleFieldIds.ts
karrui: I'll see what I can do

frontend/src/features/logic/utils/getVisibleFieldIds.ts
karrui: As mentioned in https://github.com/opengovsg/FormSG/pull/3597#discussion_r843505083, will refactor this to use a cast instead.

frontend/src/features/logic/utils/groupLogicUnitsByField.ts
```diff
@@ -0,0 +1,118 @@
...
+        logicUnit.conditions,
+        idToFieldTypeMap,
+      )
+      if (!hasAllConditions) {
+        acc.hasInvalidLogic = true
+        return acc
+      }
+      logicUnit.show.forEach((fieldId) => {
+        if (idToFieldTypeMap[fieldId]) {
+          acc.groupedLogic[fieldId] = acc.groupedLogic[fieldId]
```
karrui, line 107: `acc.groupedLogic[fieldId] = acc.groupLogic[fieldId] || []`

timotheeg (approved):

PR #3624, opened 2022-03-21
feat(v2): add question numbers in form builder

mantariksh:
## Problem

Form builder currently does not contain question numbers.

## Solution

Add question numbers, calculated by adding an `isSubmitted` metadata key to each field and counting only the fields which are submitted.

Other improvements:
- Example form fields added to Create tab mock for design review
- `SectionFieldRow` component removed as it was an unnecessary wrapper around `BaseSectionField`
- Padding between fields in the builder reduced to match the design


frontend/src/features/admin-form/create/builder-and-design/utils/questionNumbers.ts
```diff
@@ -0,0 +1,16 @@
+import { FormFieldDto } from '~shared/types/field'
+
+import { BASICFIELD_TO_DRAWER_META } from '../../constants'
+
+export const getBuilderQuestionNumbers = (
```
karrui, line 5: I am also working on generating field numbers (since I'm on working on public form submission epic), and I think the solution I came up with is a little cleaner and reusable:

```tsx
// frontend/src/features/form/types.ts
import { FormField, FormFieldDto } from '~shared/types/field'

/**
 * Partial question number, depends on field type. Some field types do not
 * require a question number
 */
export type FormFieldWithQuestionNo =
  FormFieldDto & {
    questionNumber?: number
  }
```

```tsx
// frontend/src/features/form/utils/augmentWithQuestionNo.ts
import { BasicField, FormFieldDto } from '~shared/types/field'

import { FormFieldWithQuestionNo } from '../types'

type AugmentedFieldAccumulator = {
  fields: FormFieldWithQuestionNo[]
  nonResponseFieldsCount: number
}

const NON_RESPONSE_FIELD_SET = new Set([
  BasicField.Section,
  BasicField.Statement,
  BasicField.Image,
])

export const augmentWithQuestionNo = (formFields: FormFieldDto[]) => {
  const { fields } = formFields.reduce<AugmentedFieldAccumulator>(
    (acc, field, index) => {
      if (NON_RESPONSE_FIELD_SET.has(field.fieldType)) {
        acc.nonResponseFieldsCount += 1
        acc.fields.push(field)
      } else {
        acc.fields.push({
          ...field,
          questionNumber: index + 1 - acc.nonResponseFieldsCount,
        })
      }
      return acc
    },
    { fields: [], nonResponseFieldsCount: 0 },
  )

  return fields
}
```

Then, since the schema now has the `questionNumber` prop, we can then use it directly and can remove the `questionNumber` prop in Field components.
I can't push my branch yet since it's dependent on the TableFieldDto PR (#3593) being merged haha, and probably quite annoying to make it work with base branch instead of just waiting for the PR to be merged.

Wdyt?

karrui: Opened a PR (https://github.com/opengovsg/FormSG/pull/3625) for easier reference

mantariksh: ok! will request a design review on this one first, then will make the necessary code changes when your PR is merged

mantariksh:
Superseded by #3692

PR #3644, opened 2022-03-24
feat: render govt masthead component

### LINKED ISSUE #3642: [React] Show GovtMasthead on public pages
Public pages:

- [x] public form page
- [ ] landing page (to be implemented)
- [ ] privacy + terms of use page

tshuli:
## Problem
Closes #3642 

### Screenshots

**Public View**
Screenshot 2022-04-13 at 4 11 40 PM

Screenshot 2022-04-13 at 4 11 52 PM
Screenshot 2022-04-13 at 4 12 16 PM
Screenshot 2022-04-13 at 4 12 23 PM


karrui:
Have you checked that all screens work correctly?I think some screens have a screen height of 100vh, which means unnecessary scrolling with the inclusion of a masthead

tshuli:
Thanks for pointing out! fixed it and also included screenshots

frontend/src/app/App.tsx
```diff
@@ -31,6 +32,10 @@ export const App = (): JSX.Element => (
     <BrowserRouter>
       <ChakraProvider theme={theme} resetCSS>
         <AuthProvider>
+          <Box height="2rem" width="100vw" position="fixed" top="0" zIndex="11">
+            <GovtMasthead />
+          </Box>
+          <Box height="2rem" width="100vw" />
```
karrui, line 38: Should we move this into a component for cleanliness, say `AppMasthead`? May also want to explore using a `Portal` instead of using `zIndex`, see

https://chakra-ui.com/docs/components/recipes/z-index

frontend/src/components/GovtMasthead/GovtMasthead.tsx
```diff
@@ -52,6 +52,7 @@ const HeaderBar = ({
     textStyle: { base: 'legal', md: 'caption-2' },
     display: 'flex',
     width: '100%',
+    alignItems: 'center',
```
karrui: this shouldn't be needed I think?

tshuli: thanks! yes, this has been reverted in `436736b`

frontend/src/app/App.tsx
tshuli: thanks! moved to `AppMasthead` as suggested. Also as discussed, investigated using `Portal` but will just stick to `zIndex` for now

frontend/src/components/GovtMasthead/GovtMasthead.tsx
```diff
@@ -211,3 +211,15 @@ export const GovtMasthead = ({
...
+
+// Used for rendering in app with fixed position
+export const AppMasthead = () => {
+  return (
+    <>
+      <Box height="2rem" width="100vw" position="fixed" top="0" zIndex="11">
+        <GovtMasthead />
+      </Box>
+      <Box height="2rem" width="100vw" />
+    </>
```
karrui, line 223: I think this can be encapsulated in a single child instead of using fragments. Prefer non-fragments due to possibility of layout being messed up (i.e. parent expecting single child when using `AppMasthead`, but actually there were 2 children).

Should we also change the zIndex to an exposed constant `banner`? See https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/foundations/z-index.ts.

Try:

```tsx
<Box h="2rem" w="100vw" pos="relative">
  <Box h="2rem" w="100vw" pos="absolute" top="0" zIndex="banner">
    <GovtMasthead />
  </Box>
</Box>
```

karrui (changes requested):

frontend/src/components/GovtMasthead/GovtMasthead.tsx
tshuli: thanks! incorporated these changes

karrui (approved):

tshuli:
Modified to show only on public form view, after clarification by design

timotheeg (approved):

tshuli:
restyled so that masthead does not overlap content when open

Issue #3656, opened 2022-03-28
[React/Static] Add Privacy policy page


karrui:
Should be using lazy load for these since they are probably going to be rarely accessed

karrui:
Closed by #3667

PR #3667, opened 2022-03-29
feat(v2): add (partial) Privacy Policy Page

karrui:
## Problem

This PR adds the Privacy Policy page. It is partial due to the lack of a navigation header as per design, pending that implementation.

The text in privacy policy is the exact same as the current one located in https://form.gov.sg/#!/privacy (as of 29 March 2022).

Related to #3656 

## Solution

**Features**:

- add and lazy-load `PrivacyPolicyPage` component in `/privacy` route.

**Improvements**:
- extract `AppFooter` component for sharing between app screens.

## Before & After Screenshots
New `PrivacyPolicyPage` story added


frontend/src/app/AppRouter.tsx
```diff
@@ -6,6 +6,7 @@ import {
   ADMINFORM_ROUTE,
   ADMINFORM_SETTINGS_SUBROUTE,
   LOGIN_ROUTE,
+  PRIVACY_ROUTE,
```
mantariksh, line 9: nit: can probably rename to `PRIVACY_POLICY_ROUTE` to make it clear what this is for haha

mantariksh (approved):

frontend/src/app/AppRouter.tsx
karrui: 62968cb

PR #3684, opened 2022-03-31
feat(v2): add admin form feedback tab

### LINKED ISSUE #3794: [React/Feedback] View form feedback table in admin view
Did not realize this is not an issue yet

tshuli:
## Problem
This PR implements the feedback responses tab in admin form view to view form feedback.

Note that the feedback export should call backend `/feedback/download` endpoint instead of generating it on browser currently. This will be updated in a future PR.

Closes #3794

## Screenshots
New stories added for rendering of form feedback table.

**New dependencies**:

- `react-csv` : To download form feedback as csv (KR: this does not seem to be the best library to use, i think there are better libraries that handle background streaming in a better way, such as react-papaparse. Will evaluate.)


frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+      : []
+  }, [feedbackData])
+
+  const { prepareRow, headerGroups, page, gotoPage } = useTable<Data>(
+    {
+      columns,
+      data,
+      initialState: { pageIndex: currentPage, pageSize: 9 },
+    },
+    useSortBy,
```
karrui, line 77: nit: how would sort work? Would it give the users the perception that it's sorting on the entire feedback instead of just a single page? Maybe it's a question to ask @pearlyong hehe

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+                      width: column.width,
+                    },
+                  }),
+                )}
+                color="white"
+                pl="1rem"
+                textStyle="subhead-2"
+                fontWeight="500"
+                fontSize="0.875rem"
+                textTransform="none"
```
karrui, line 104: I think the reason why you need to override `fontWeight` etc even with `textStyle` is due to the defaults in `Th` in the theme. Do you think adding a specific table variant would be a better call?

See `frontend/src/theme/components/Table.ts` for an example variant used for Table field; I think you can co-opt that and add a `solid` theme or something.

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+                  <Icon pb="0.2rem">
+                    {column.isSorted ? (
+                      column.isSortedDesc ? (
+                        <BxsChevronDown fontSize="2rem" />
+                      ) : (
+                        <BxsChevronUp fontSize="2rem" />
+                      )
+                    ) : (
+                      ''
+                    )}
```
karrui, line 117: I think we can simplify this:

```tsx
{column.isSorted
    ? <Icon as={isSortedDesc ? BxsChevronDown : BxsChevronUp} fontSize="2rem" /> 
    : null
}
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+          return (
+            <Tr>
+              {row.cells.map((cell) => {
+                return (
+                  <Td
+                    py="0.625rem"
+                    pl="1rem"
+                    color="secondary.500"
+                    fontWeight="regular"
+                    fontSize="0.875rem"
```
karrui, line 137: Same as above comment; we should use `textStyle` prop over manually listing the font props

frontend/src/typings/react-table-config.d.ts
```diff
@@ -0,0 +1,132 @@
+// Since usePagination and useSortBy are considered plugins, it is necessary to declare the types explicitly.
```
karrui, line 1: This means the types are global even though different instances of `react-table` may not use the above plugins. Suggest creating explicit interfaces that extend the plugins we need instead of a `.d.ts` file:

```ts
interface FeedbackTableOptions<D extends Record<string, unknown>>
  extends UsePaginationOptions<D>,
    UseSortByOptions<D> {}
```

and use that in useTable calls:

```tsx
const { prepareRow, headerGroups, page, gotoPage } = useTable<Data>(
    {
      columns,
      data,
      initialState: { pageIndex: currentPage, pageSize: 9 },
    } as FeedbackTableOptions,
    useSortBy,
    usePagination
)
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackService.ts
```diff
@@ -0,0 +1,15 @@
+import { FormFeedbackMetaDto } from '~shared/types/form/'
+
+import { ApiService } from '~services/ApiService'
+
+import { ADMIN_FORM_ENDPOINT } from '../../common/AdminViewFormService'
+
+export const getFormFeedback = async ({
```
karrui, line 7: nit: since there is only 1 param, prefer just using formId directly instead of being nested in an object:

```ts
export const getFormFeedback = (formId: string) => void
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+        width: '5rem',
+      },
+      {
+        Header: 'Date',
+        accessor: 'date',
+        sortType: (rowA, rowB) => {
+          const dateA = new Date(rowA.values.id)
+          const dateB = new Date(rowB.values.id)
+          return dateA > dateB ? 1 : -1
+        },
```
karrui, line 39: Is values.id a date string? What does it look like?

Also suggest extracting this column constant to outside of the component instead since we do not need any dependent data and it is truly a constant.

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,149 @@
+import React, { useEffect } from 'react'
```
karrui, line 1: nit: I don't think we need the `React` import, we should destructure the imports we need directly, like `useMemo`, `useCallback`, etc.

karrui (changes requested):

karrui:
There is also a screen for empty feedback, can be seen in our figma file in the `Empty, Error, Edge` page

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
tshuli: Actually it does sort on the entire feedback
e.g. if I am now on page 3 and I click sort by rating, then I will see the 19th - 27th entries based on the new sorted order
@pearlyong to confirm this is desired behavior

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
tshuli: yes, it is a date string and added documentation (actually it should be `values[id]` and  where `id` is a parameter for columnId, i've fixed it)

also moves constant out of component

karrui:
Taking over this PR since ShuLi is on vacation!

timotheeg:
Need to handle empty feedback

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,188 @@
...
+    maxWidth: 100, // maxWidth is only used as a limit for resizing
+  },
+  {
+    Header: 'Date',
+    accessor: 'date',
+    sortType: (rowA, rowB, columnId) => {
+      const dateA = new Date(rowA.values[columnId]) //rowA.values[columnId] is a date string
+      const dateB = new Date(rowB.values[columnId])
+      return dateA > dateB ? 1 : -1
+    },
```
timotheeg, line 40: Use `timestamp` as accessor and compare with a plain difference a - b . 

And reformat date in frontend.

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const data = useMemo(() => {
+    return (
+      feedbackData?.map((feedback) => {
+        return {
+          index: feedback.index,
+          date: feedback.date,
+          feedback: feedback.comment,
+          rating: feedback.rating,
+        }
+      }) ?? []
```
timotheeg, line 79: This mapping is not necessary. the accessor and title can be set in the header definition above

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
```diff
@@ -0,0 +1,188 @@
...
+    getTableProps,
+    getTableBodyProps,
+    headerGroups,
+    page,
+    gotoPage,
+  } = useTable<FeedbackColumnData>(
+    {
+      columns: FEEDBACK_TABLE_COLUMNS,
+      data,
+      initialState: { pageIndex: currentPage, pageSize: 9 },
```
timotheeg, line 94: Why 9? This needs to be synced with Pagination component.

frontend/src/features/admin-form/responses/FeedbackPage/queries.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @precondition Must be wrapped in a Router as `useParam` is used.
+ */
+export const useFormFeedback = (): UseQueryResult<FormFeedbackMetaDto> => {
+  const { formId } = useParams()
+  if (!formId) throw new Error('No formId provided')
+
+  return useQuery(
+    adminFormFeedbackKeys.id(formId),
+    () => getFormFeedback(formId),
+    { staleTime: 10 * 60 * 1000 },
```
timotheeg, line 25: Probably no need to prematurely optimize with caching

karrui: 0e643c3

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
karrui: 6eb78c1

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
karrui: 6eb78c1

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackTable.tsx
karrui: 6eb78c1

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackDownloadButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  formId: string | undefined
+  feedback: ProcessedFeedbackMeta[] | undefined
+}
+
+export const FeedbackDownloadButton = ({
+  feedback,
+  isDisabled,
+  formId,
+}: FeedbackDownloadButtonProps) => {
+  const data = useMemo(() => {
```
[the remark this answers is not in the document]
karrui: Shape is not consistent, and I realize we have a specific streaming endpoint which is not used here. I will add a new PR to use the streaming endpoint instead.

timotheeg (approved):
Let's go!

karrui (approved):
ownself lgtm ownself

PR #3691, opened 2022-04-04
feat: append new field to form on click

mantariksh:
## Problem

Currently, fields can only be added through drag and drop. We also want users to be able to simply click on fields to append them to the end of the form.

## Solution

Implement field creation on click, and also scroll the new field into view when it is created.

### Before (no action on click)
field-option-before

### After
field-option-after

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
@@ -88,10 +88,10 @@ export const FieldRowContainer = ({
     return false
   }, [stateData, field])
 
-  const activeFieldRef = useRef<HTMLDivElement | null>(null)
+  const ref = useRef<HTMLDivElement | null>(null)
   useEffect(() => {
     if (isActive) {
-      activeFieldRef.current?.scrollIntoView({
+      ref.current?.scrollIntoView({
         // Avoid sudden jump when field is clicked
         block: 'nearest',
         // Also avoid behavior: 'smooth' as scrolling may take very long
@@ -182,7 +182,7 @@ export const FieldRowContainer = ({
             align="center"
             onClick={handleFieldClick}
             onKeyDown={handleKeydown}
-            ref={isActive ? activeFieldRef : undefined}
+            ref={ref}
           >
             <Fade in={isActive}>
               <chakra.button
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/FieldListOption.tsx
```diff
@@ -62,6 +69,21 @@ export const FieldListOption = forwardRef<FieldOptionProps, 'button'>(
       () => BASICFIELD_TO_DRAWER_META[fieldType],
       [fieldType],
     )
+    const { data: form } = useCreateTabForm()
+    const numFields = useMemo(() => form?.form_fields?.length ?? 0, [form])
```
timotheeg, line 73: Just checking: is doing memoization based on `form` sufficient? As in: could `form` as a reference be constant, but `form.form_fields.length` have changed?

(Sorry, I know we use `useMemo()` everywhere, but this still looks a little funny to me 😅)

mantariksh: I can't find documentation on this anywhere, but I'm pretty sure `react-query` changes the data reference when the data changes - the alternative doesn't make any sense! the only situation when the reference remains the same should be when the data hasn't change, due to [structural sharing](https://react-query.tanstack.com/guides/important-defaults#_top):

> Query results by default are structurally shared to detect if data has actually changed and if not, the data reference remains unchanged to better help with value stabilization with regards to useMemo and useCallback. If this concept sounds foreign, then don't worry about it! 99.9% of the time you will not need to disable this and it makes your app more performant at zero cost to you.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -169,6 +182,7 @@ export const FieldRowContainer = ({
             align="center"
             onClick={handleFieldClick}
             onKeyDown={handleKeydown}
+            ref={isActive ? activeFieldRef : undefined}
```
karrui, line 185: is there a reason why `ref` is only assigned on isActive state? We can also just assign the ref regardless, and call scrollIntoView when active.

karrui (approved):
lgtm

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
mantariksh: sure! 992ae61c

PR #3692, opened 2022-04-04
feat: add question numbers in form builder

mantariksh:
## Problem

Form builder currently does not contain question numbers.

## Solution

Calculate and render question numbers. However, note that this approach is only temporary so that the form builder can go through design review. It will be subsequently be refactored based on #3625 (the approach in #3625 requires fairly extensive changes, so I avoided duplicating it here to avoid merge conflicts).

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/utils/questionNumbers.ts
+++ b/frontend/src/features/admin-form/create/builder-and-design/utils/questionNumbers.ts
@@ -8,8 +8,7 @@ export const getBuilderQuestionNumbers = (
   let questionCounter = 0
   return fields.map((field) => {
     if (BASICFIELD_TO_DRAWER_META[field.fieldType].isSubmitted) {
-      questionCounter += 1
-      return `${questionCounter}.`
+      return `${++questionCounter}.`
     }
     return undefined
   })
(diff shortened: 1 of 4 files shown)


frontend/src/features/admin-form/create/builder-and-design/utils/questionNumbers.ts
```diff
@@ -0,0 +1,16 @@
...
+import { BASICFIELD_TO_DRAWER_META } from '../../constants'
+
+export const getBuilderQuestionNumbers = (
+  fields: FormFieldDto[],
+): (string | undefined)[] => {
+  let questionCounter = 0
+  return fields.map((field) => {
+    if (BASICFIELD_TO_DRAWER_META[field.fieldType].isSubmitted) {
+      questionCounter += 1
+      return `${questionCounter}.`
```
timotheeg, line 12: I dunno if it's too late to give this comment, but I'd recommend keeping this as number (i.e. meaningful data), and do the eventual transformation to string later, as close to the point of rendering as possible. That would leave the actual renderer in control of how to display, whether with a dot-suffix, or paren-suffix, or whatever else the renderer wants to do.

Also super nit 😅 
```javascript
return ++questionCounter;
```

mantariksh: made the `++` change in [f3d04a0](https://github.com/opengovsg/FormSG/pull/3692/commits/f3d04a09ae7d5be1c86cfda056a1259064cb8629), but would you be okay with keeping the data flow as-is for now? I'm going to refactor towards the approach in #3625 once it's merged anyway, and modifying the data flow will cause merge conflicts

timotheeg: Yup, yup, let's go!

timotheeg (approved):

PR #3694, opened 2022-04-04
feat: implement field deletion confirmation modal

mantariksh:
_Depends on #3693_

## Problem

Currently, clicking the delete button on a field causes it to be deleted immediately. However, as per the [design](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=2584%3A107021), there should be a confirmation modal before the field is deleted.

## Solution

Add a confirmation modal before actually deleting the field.

### Before
delete-modal-before

### After
delete-modal-after


timotheeg (approved):

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContext.ts
```diff
@@ -3,6 +3,7 @@ import { UseDisclosureReturn } from '@chakra-ui/react'
 
 type BuilderAndDesignContextProps = {
   mobileCreateEditModal: UseDisclosureReturn
+  deleteFieldModal: UseDisclosureReturn
```
karrui, line 6: nit: rename to a clearer variable such as `xxxModalDisclosure`?

frontend/src/features/admin-form/create/builder-and-design/DeleteFieldModal/DeleteFieldModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+      })
+    }
+  }, [deleteFieldMutation, onClose, stateData])
+
+  return (
+    <Modal isOpen={isOpen} onClose={onClose}>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalCloseButton />
+        <ModalHeader>{`Delete ${fieldTypeLabel} field`}</ModalHeader>
```
karrui, line 52: nit: do we need a string template here? should we use direct reference instead?: `Delete {fieldTypeLabel} field`

karrui (approved):

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContext.ts
mantariksh: [b26f1a0](https://github.com/opengovsg/FormSG/pull/3694/commits/b26f1a0aac9bf5bafec7f27c64eb20334f0cbdfe)

frontend/src/features/admin-form/create/builder-and-design/DeleteFieldModal/DeleteFieldModal.tsx
mantariksh: [4eb2272](https://github.com/opengovsg/FormSG/pull/3694/commits/4eb227265425a4359ad945b9bb4e97189d314226)

PR #3702, opened 2022-04-05
fix(EditCheckbox): validate numbers and set types correctly

mantariksh:
## Problem

1. Number inputs for `customMin` and `customMax` in `EditCheckbox` currently allow decimal points as well as `e` (e.g. `1e5`).
2. The output type for the number inputs is set to `number | string | undefined` when in fact it is always `string`.

## Solution

Tighten the number input validation and fix the types.


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
```diff
@@ -37,8 +37,8 @@ type EditCheckboxKeys = typeof EDIT_CHECKBOX_FIELD_KEYS[number]
 type EditCheckboxInputs = Pick<CheckboxFieldBase, EditCheckboxKeys> & {
   fieldOptions: string
   ValidationOptions: {
-    customMin?: number | string
-    customMax?: number | string
+    customMin: string
```
karrui, line 40: should we only allow `number` instead of string so we do not need to do all the string transformations?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
```diff
@@ -238,11 +239,17 @@ export const EditCheckbox = (props: EditCheckboxProps): JSX.Element => {
...
+              render={({ field: { onChange, ...rest } }) => (
                 <NumberInput
+                  inputMode="numeric"
+                  precision={0}
                   flex={1}
                   showSteppers={false}
-                  {...field}
+                  onChange={(val) => {
+                    // Only allow numeric inputs
+                    onChange(val.replace(/\D/g, ''))
```
karrui, line 250: should we transform output of NumberInput to be `number` type directly? This better fits with the `Number` type in the schema.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
```diff
@@ -47,8 +47,8 @@ const transformCheckboxFieldToEditForm = (
 ): EditCheckboxInputs => {
   const nextValidationOptions = field.validateByValue
     ? {
-        customMin: field.ValidationOptions.customMin || '',
-        customMax: field.ValidationOptions.customMax || '',
+        customMin: field.ValidationOptions.customMin?.toString() || '',
```
karrui, line 50: This means we do not need this toString call anymore, nor the parseInt call below.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
mantariksh: sure!

mantariksh: [6744778](https://github.com/opengovsg/FormSG/pull/3702/commits/6744778bc42caf15ca8ddd6ffcf75568869db583)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
mantariksh: [6744778](https://github.com/opengovsg/FormSG/pull/3702/commits/6744778bc42caf15ca8ddd6ffcf75568869db583)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
mantariksh: [6744778](https://github.com/opengovsg/FormSG/pull/3702/commits/6744778bc42caf15ca8ddd6ffcf75568869db583)

mantariksh:
found out this doesn't actually work lol will fix

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
mantariksh: further fixed in [2b16210](https://github.com/opengovsg/FormSG/pull/3702/commits/2b16210db1245956dc3a3e31af1876b2131f0fd2)

mantariksh:
@karrui so in the end I decided to revert back to the string input because:
- Using `number` resulted in behaviour where typing large numbers would eventually result in users seeing something like `1000e+21`, upon which typing more numbers would result in the `e+` getting stripped away by `.replace` and the number varying unpredictably. Users should never type numbers that big anyway, but it was still weird behaviour.
- Simply limiting the number of characters via `onChange` caused a situation where users would type but not see the input changing, with no error message to inform them of what was happening.
- We can ensure that the `toString` conversions are safe by simply limiting the inputs to some sane number (currently 10,000).

This PR is yours now though haha so feel free to take whichever approach you feel is best!

karrui (approved):

PR #3710, opened 2022-04-06
fix: update ports for maildev to 1080 and 1025

timotheeg:
## Context
[maildev](https://hub.docker.com/r/maildev/maildev) has released a new docker image on 2022-03-30, which changed the internal app ports.

The smtp server now runs on port 1025 (was 25), the web interface runs in port 1080 (was 80). This was done on purpose in [this PR](https://github.com/maildev/maildev/pull/281).

This change in port causes the docker setup for FormSG to break since the app can't connect to the mail server, and so the ports need to be updated.

## Approach

Update both the web app and smtp port for maildev.

The PR targets branch `develop`, and we can merge `develop` into `form-v2/develop` later.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,7 +32,7 @@ services:
       - SECRET_ENV=development
       - SUBMISSIONS_RATE_LIMIT=200
       - SEND_AUTH_OTP_RATE_LIMIT=60
-      - SES_PORT=25
+      - SES_PORT=1025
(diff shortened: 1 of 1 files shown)


karrui (approved):
lgtm. 

We can also explore adding command args to set the port in compose so it is clear what the port mappings are instead of relying on the defaults (which may change as it did now)

PR #3716, opened 2022-04-07
feat(v2): add Settings Webhooks tab

karrui:
## Problem

This PR adds the missing Webhooks tab in Admin form settings page.
Also removes all unused tabs that will not make the first React migration cut.
Also also prevents activation of form if e-serviceId is not set.

## Solution

**Features**:
- feat: add WebhooksSection directory for all webhook settings related components
  - call API when url/ retry state is changed.

**Improvements**:

- feat: prevent form activation when form has authType but no esrvcId
- feat: remove unused settings tab panels
- feat(Toggle): add useMarkdown flag for markdown usage 
  - this was required since design wanted to link a URL to toggle description.

## Before & After Screenshots
New stories for SettingsWebhooksPage has been added


frontend/src/features/admin-form/settings/components/WebhooksSection/WebhookUrlInput.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (!isValid) {
+      return resetField('url')
+    }
+    return handleUpdateWebhook()
+  }, [handleUpdateWebhook, isValid, resetField])
+
+  const handleWebhookUrlEnterKeyDown: KeyboardEventHandler = useCallback(
+    (e) => {
+      if (!isValid || e.key !== 'Enter') return
+      return handleUpdateWebhook()
```
karrui, line 61: if we blur here instead of calling the function

karrui: 538bc44

timotheeg (approved):
lgtm

PR #3750, opened 2022-04-13
feat: conditional react routing

timotheeg:
## Context

We need a way to control rollout of React (and stop it if needed). 

We have several constraints:
* We should have an illegibility criteria where some form are excluded from the react rollout
* Domain must stay constants as `form.gov.sg`
* Form URLs must be constant, so whatever URLs were shared before, during, and after rollout will work without us having to maintain deprecated custom routes
* Respondent taking SP/CP/MyInfo forms should have the same environment before and after login, so they are not confused
* Admin can chose an environment (react/angular) and have it served consistently across multiple sessions

We considered several approach to do routing in infra, but the constraints above make it hard to do this outside application (which can have knowledge on admin/respondent and eligibility criteria based on form definitions.

## Approach
* Update old routes to have a consistent param `formId` in URL
* Conditional routing is done in app code with a new controller `reactMigrationController`
    * new config block `reactMigration populated from new env vars
    * static assets are loaded first so the react controller can catch all urls later
* Admins choice is saved in a cookie with 1 year expiry
    * new API endpoint `/api/v3/admin/environment/(react|angular)` to allow setting of the cookie (only needed if `http-only`)
* Respondent choice of react/angular is saved in a session cookie. Same user may get different ui for forms during the rollout period, but a single interaction with a form, including refreshes should be consistent within one ui.

## TODOs / TODISCUSS
- [ ] Add Eligibility criteria (forms with SP/CP/MyInfo) are excluded from first phase of rollout
- [ ] Test for all common cases (pairing session wearing hacker hats -> can we break it?)
- [ ] Build UI in BOTH angular and React to allow switching the environment (simple api call to set cookie and redirect)
- [ ] Discuss whether frontend needs access to the environment cookies to take decisions (cookies do not contain secure data and do not necessarily need to be `http-only`

## Deploy Notes

-  `REACT_MIGRATION_RESPONDENT_ROLLOUT`: number between 0 and 100
-  `REACT_MIGRATION_ADMIN_ROLLOUT`: number between 0 and 100


src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export const RESPONDENT_COOKIE_OPTIONS = {
+  httpOnly: true,
+  sameSite: 'strict' as const,
+  secure: !config.isDev,
+}
+
+export const ADMIN_COOKIE_OPTIONS = {
+  httpOnly: true,
+  maxAge: 365 * 24 * 60 * 60, // 1 year
```
timotheeg, line 26: Reduce to 2 months

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+  const threshold = isSPCP
+    ? config.reactMigrationConfig.respondentRolloutSPCP
+    : config.reactMigrationConfig.respondentRolloutNoSPCP
+
+  if (threshold <= 0) {
+    // Check the rollout value first, if it's 0, react is DISABLED
+    // And we ignore cookies entirely!
+    showReact = false
+  } else if (req.cookies) {
+    if (ADMIN_COOKIE_NAME in req.cookies) {
```
timotheeg, line 72: maybe add comment here about priorityl let admins be dogfooder!

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+  Record<string, string>
+> = async (req, res, next) => {
+  const formResult = await FormService.retrieveFormKeysById(req.params.formId, [
+    'authType',
+  ])
+  let showReact: boolean | undefined = undefined
+  let isSPCP = true
+
+  if (!formResult.isErr()) {
+    isSPCP = [FormAuthType.SP, FormAuthType.CP, FormAuthType.MyInfo].includes(
```
timotheeg, line 58: change this conditional to isSPCP = formResult.value.authType === FormAuthType.Nil ?

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+import { FormAuthType } from '../../../../shared/types'
+import config from '../../config/config'
+import { ControllerHandler } from '../core/core.types'
+import * as FormService from '../form/form.service'
+import * as PublicFormController from '../form/public-form/public-form.controller'
+import { RedirectParams } from '../form/public-form/public-form.types'
+import * as HomeController from '../home/home.controller'
+
+const RESPONDENT_COOKIE_NAME = 'v2-respondent-ui'
+const ADMIN_COOKIE_NAME = 'v2-admin-ui'
```
timotheeg, line 12: Make them env-var so we can reset anytime

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+export const ADMIN_COOKIE_OPTIONS = {
+  httpOnly: true,
+  maxAge: 365 * 24 * 60 * 60, // 1 year
+  sameSite: 'strict' as const,
+  secure: !config.isDev,
+}
+
+const serveFormReact: ControllerHandler = (_req, res) => {
+  const reactFrontendPath = path.resolve('dist/frontend')
+  res.sendFile(path.join(reactFrontendPath, 'index.html'))
```
timotheeg, line 33: Add return before all `res.whatever()`

karrui:
I've fixed the Chromatic failure in #3751

src/app/modules/react-migration/react-migration.controller.ts
timotheeg: #bb1239ef545269503c02ee7ff2caa988492e6b16

Just to note: adding the returns is to ensure we don't have accidental multiple data sending on `res`, because that can cause runtime exceptions.

In this file, it feels a little unnecessary because the code is structured in such a way that it should never happen. All the functions which return data do so at the very end, and even in the case of conditionals, the conditional evaluation is itself done at the very end, which makes it clear nothing should go after it.

```javascript
if (condition) {
    res.send(/*... */)
} else {
    res.send(/*... */)
}
```

src/app/modules/react-migration/react-migration.controller.ts
timotheeg: https://github.com/opengovsg/FormSG/pull/3750/commits/67e50620156ad0c043c62834ae9ac4fb18d5f415

src/app/modules/react-migration/react-migration.controller.ts
timotheeg: https://github.com/opengovsg/FormSG/pull/3750/commits/b5ceec2aa436830782f93e84ebed30d295fc434a

src/app/modules/react-migration/react-migration.controller.ts
timotheeg: https://github.com/opengovsg/FormSG/pull/3750/commits/3c5321d41c45bda994f562165d5e5f80634e9737

src/app/modules/react-migration/react-migration.controller.ts
timotheeg: https://github.com/opengovsg/FormSG/pull/3750/commits/2b2147ac222bf91ec70de5368f51ea2df56757c0

src/app/loaders/express/logging.ts
```diff
@@ -61,6 +67,20 @@ const loggingMiddleware = () => {
       if (transactionId) {
         meta.transactionId = transactionId
       }
+
+      // Temporary: cookies are blacklisted, but we to track the state of the rollout for this particular request
```
timotheeg, line 71:
```suggestion
      // Temporary: cookies are blacklisted, but we want to track the state of the rollout for this particular request
```

karrui (approved):

Dockerfile.production
```diff
@@ -10,24 +10,36 @@ WORKDIR /build
...
+# TODO: npm install ci --production instead of copying the entire node_modules from the build stage -_-
 
 # Install build from backend-build
 COPY --from=build /build/node_modules /opt/formsg/node_modules
 COPY --from=build /build/shared/node_modules /opt/formsg/shared/node_modules
 COPY --from=build /build/dist /opt/formsg/dist
 COPY --from=build /build/package.json /opt/formsg/package.json
 
+# Make sure the views
+# COPY --from=build /build/src/app/views /opt/formsg/src/app/views
```
timotheeg, line 41: Not correct for email rendering 😢

PR #3774, opened 2022-04-20
feat(v2/PublicForm): add handling of form color theme, update design

karrui:
## Problem

This PR implements handling of the current form's color theme and updates components to render their expected color theme. A couple of components have also been updated to accept theming of color.

Also update mobile and desktop styling to better fit design, most notably padding and the display of submission button.

Stories have been added for the various color themes and mobile views and snapshots should reflect those changes.


timotheeg (approved):
lgtm 👍

PR #3783, opened 2022-04-24
fix: Update TROUBLESHOOTING.md with MongoDB not primary error

frankchn:


karrui (approved):
lgtm. thanks for the troubleshooting steps! can you change the branch to one that is prefixed by `form-v2` so the cicd passes?

PS: i would have thought the bitnami package introduced in #3752 would have automatically set up replicasets properly; is that not the case? :(

frankchn:
> lgtm. thanks for the troubleshooting steps! can you change the branch to one that is prefixed by `form-v2` so the cicd passes?
> 
> PS: i would have thought the bitnami package introduced in #3752 would have automatically set up replicasets properly; is that not the case? :(

Yeah it doesn't set it properly for me at least half the time unfortunately :(

PR #3771, opened 2022-04-19
feat(v2): add admin form collaborator modal

karrui:
## Problem

This PR adds the initial admin form collaboration feature, where admins can add/remove collaborators.

Note that there is still no transfer ownership feature, or viewer-only modal. Those will come in later PRs.
Meaning there are no client-side guards if a user with view-only rights attempts to modify collaborator permissions (backend will correctly block the action).

Related to #2487 

## Solution

**Features**:

- add CollaboratorModal component and stories, and use in admin nav bar
  - includes `useCollaboratorModal` hook to control modal open close
  - add various skeleton states

## Before & After Screenshots
New stories in storybook

Issue #2487, opened 2021-07-30
[React/AdminForm] Edit collaborators feature


karrui:
**Time estimation: 1w**

Flow mocks can be found at 
Figma > DesignMaster_v3 > Collaborators

This issue tracks the completion of all collaborator related functions, like editing their permissions, or transferring ownership

- [x] #3771
- [x] #3788
- [x] #3793
- [x] #3802

PR #3788, opened 2022-04-25
feat(v2/collab): Add transfer ownership feature

karrui:
> Note that this PR builds on top of #3771.

## Problem

This PR implements the ability for form owners to transfer ownership of their forms.

Same caveats as the base PR apply, no client-side guards if a user with view-only rights attempts to transfer permissions (backend will correctly block the action).

Related to #2487

## Solution

**Features**:

- feat: add TransferOwnershipScreen component
- feat(PermissionDropdown): only allow transfer owner option if current logged in user is the form owner

**Improvements**:

- feat: move current modal into CollabatorListScreen so there can be a distinction between transfer ownership and default list view.
- feat: extract MotionBox and XMotionBox template components
  - duplicate code, also used in CreateFormModal
- feat(CollaboratorListScreen): add bottom divider and fix padding

## Before & After Screenshots
Stories for collaborator modal should have been updated to also include the transfer ownership feature.
Will add more stories if I have time...

### DIFF
--- a/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/PermissionDropdown.tsx
+++ b/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/PermissionDropdown.tsx
@@ -2,7 +2,7 @@ import { Text } from '@chakra-ui/react'
 
 import Menu from '~components/Menu'
 
-import { DropdownRole } from './AddCollaboratorInput'
+import { DropdownRole } from '../constants'
 
 export interface PermissionDropdownProps {
   value: DropdownRole
--- a/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/PermissionDropdown.tsx
+++ b/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/PermissionDropdown.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from 'react'
 import { Text } from '@chakra-ui/react'
 
 import Menu from '~components/Menu'
@@ -8,7 +9,7 @@ export interface PermissionDropdownProps {
   value: DropdownRole
   onChange: (role: DropdownRole) => void
   isLoading: boolean
-
+  allowTransferOwnership: boolean
   buttonVariant?: 'outline' | 'clear'
 }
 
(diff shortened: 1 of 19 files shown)


frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormModalContent.tsx
```diff
@@ -11,53 +8,19 @@ import {
...
-      custom={direction}
-      variants={SCREEN_ANIMATION_VARIANT}
-      initial={isFirstLoad ? 'center' : 'enter'}
-      animate="center"
-      transition={{
-        x: { type: 'spring', stiffness: 300, damping: 30 },
-        opacity: { duration: 0.2 },
-      }}
-    >
+    <XMotionBox keyProp={currentStep} custom={direction}>
```
timotheeg, line 19: Great job extracting that to its own base component! 👏

frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/PermissionDropdown.tsx
```diff
@@ -1,23 +1,32 @@
...
   onChange,
   isLoading,
+  allowTransferOwnership,
   buttonVariant = 'outline',
 }: PermissionDropdownProps): JSX.Element => {
+  const availableRoles = useMemo(() => {
+    return Object.values(DropdownRole).filter((role) => {
+      // Either not owner role, or owner role and allowTransferOwnership is true.
+      return role !== DropdownRole.Owner || allowTransferOwnership
+    })
```
timotheeg, line 27: /nit Would typescript allow something like this: 
```javascript
const roles = { ...DropdownRole }

if (!allowTransferOwnership) {
    delete roles[DropdownRole.Owner]
}

return Object.values(roles);
```
(I think it makes clearer what this is doing)

timotheeg (approved):
lgtm 👍

frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/PermissionDropdown.tsx
karrui: A little troublesome due to types (destructuring enums result in an object with all enums marked as required, meaning it will take some casting to allow `delete` keyword).

Will keep it like this for now since it's still understandable

Issue #3792, opened 2022-04-26
[React] Add i18n support for application


karrui:
This is tangentially related to #1906, but that issue is specifically for form admins to optionally add i18n translations to their own form fields.

This issue tracks the i18n of the entire application and its default text.

LinHuiqing:
Migrated to Linear: FRM-1410

PR #3823, opened 2022-05-04
feat(v2): add react-i18next and associated boilerplate for interface …

frankchn:
## Problem
We need internationalization support for the new React-based frontend.

Part of #3792

## Solution
This adds react-i18n and related packages, adds boilerplate code, and translates one single sentence within the new site to demonstrate that it works.

**Breaking Changes** 

## Tests
All tests should run exactly the same as before. Verify that the "Build secure government forms in minutes" slogan still appears correctly on the login page.

**New dependencies**:

- i18next: ^21.6.16 - Main package
- i18next-browser-languagedetector: ^6.1.4 - Autodetects browser languages
- i18next-icu: ^2.0.3 - Backend support for ICU message format
- intl-messageformat: ^9.13.0" - Required for i18next-icu
- react-i18next: ^11.16.7 - React wrapper for i18next


frontend/src/i18n/locales/en-sg.ts
```diff
@@ -0,0 +1,11 @@
+import Translation from './translation.interface'
+
+const enSG: Translation = {
```
karrui, line 3: supernit: prefer  `export const` over `export default`, but am ok! (rationale is in case we rename things, but also cognisant that translation files probably will never be renamed haha)

frontend/src/i18n/locales/index.ts
```diff
@@ -0,0 +1,9 @@
+import { ResourceLanguage } from 'i18next'
+
+import enSG from './en-sg'
+
+const locales = {
```
karrui, line 5: supernit: prefer  `export const` over `export default`, but am ok! (rationale is in case we rename things, but also cognisant that translation files probably will never be renamed haha)

frontend/src/i18n/locales/translation.interface.ts
```diff
@@ -0,0 +1,9 @@
+interface Translation {
```
karrui, line 1: supernit: prefer `types.ts` over `translation.interface.ts` for consistency with rest of frontend. If we really want to use dot notation, maybe `translation.types.ts`

karrui (approved):
lgtm, thanks for setting this up.

karrui (changes requested):
Ah, realize that our snapshot tests regressed. Might need to add the i18n pipeline into storybook: https://storybook.js.org/addons/storybook-react-i18next

frontend/src/i18n/locales/en-sg.ts
frankchn: done

frontend/src/i18n/locales/index.ts
frankchn: done

frontend/src/i18n/locales/translation.interface.ts
frankchn: done

frankchn:
> Ah, realize that our snapshot tests regressed. Might need to add the i18n pipeline into storybook: https://storybook.js.org/addons/storybook-react-i18next

thanks! adding

frankchn:
Thanks @karrui -- updated :)

karrui (approved):
lgtm!

PR #3824, opened 2022-05-04
feat: track Twilio metrics to DataDog

timotheeg:
## Problem
We need to be able to diagnose Twilio errors early by having a clear view into Twilio API usage, callbacks, and success/errors in both cases.

## Solution
Report real time counters of Twilio stats under the metrics:
* `vendor.twilio.sms.send`
* `vendor.twilio.sms.update`

Note that this is a PR for develop to be released to production asap.

### DIFF
--- a/src/app/services/sms/sms.service.ts
+++ b/src/app/services/sms/sms.service.ts
@@ -7,6 +7,7 @@ import Twilio from 'twilio'
 import { isPhoneNumber } from '../../../../shared/utils/phone-num-validation'
 import { AdminContactOtpData, FormOtpData } from '../../../types'
 import config from '../../config/config'
+import { statsdClient } from '../../config/datadog-statsd-client'
 import { createLoggerWithLabel } from '../../config/logger'
 import getFormModel from '../../models/form.server.model'
 import {
@@ -49,6 +50,9 @@ const secretsManager = new SecretsManager({ region: config.aws.region })
 export const twilioClientCache = new NodeCache({
   deleteOnExpire: true,
   stdTTL: 10,
+})
+const ddClient = statsdClient.childClient({
+  prefix: 'vendor.twilio.',
 })
 
 /**
@@ -212,7 +216,17 @@ const sendSms = (
     },
   )
     .andThen(({ status, sid, errorCode, errorMessage }) => {
+      const ddTags = {
+        // msgsrvcsid: msgSrvcSid,
+        smsstatus: status,
+        errorcode: '0',
+      }
+
       if (!sid || errorCode) {
+        if (errorCode) {
+          ddTags.errorcode = `${errorCode}`
+        }
+
         logger.error({
           message: 'Encountered error code or missing sid after sending SMS',
           meta: {
@@ -222,6 +236,8 @@ const sendSms = (
             errorMessage,
           },
         })
+
+        ddClient.increment('sms.send', 1, 1, ddTags)
 
         // Invalid number error code, throw a more reasonable error for error
         // handling.
@@ -236,6 +252,8 @@ const sendSms = (
               }),
         )
       }
+
+      ddClient.increment('sms.send', 1, 1, ddTags)
 
       // No errors.
       logger.info({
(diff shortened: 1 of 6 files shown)


src/app/services/sms/sms.service.ts
```diff
@@ -50,6 +51,9 @@ export const twilioClientCache = new NodeCache({
   deleteOnExpire: true,
   stdTTL: 10,
 })
+const ddClient = statsdClient.childClient({
+  prefix: 'vendor.twilio.',
```
karrui, line 55: nit: should we extract this? wary of magic strings. Also suggest renaming to `DD_CLIENT` since it's a top-level constant

src/app/services/sms/sms.service.ts
```diff
@@ -212,7 +216,17 @@ const sendSms = (
     },
   )
     .andThen(({ status, sid, errorCode, errorMessage }) => {
+      const ddTags = {
+        // msgsrvcsid: msgSrvcSid,
```
karrui, line 220: any reason why this is commented out?

src/app/services/sms/sms.service.ts
```diff
@@ -212,7 +216,17 @@ const sendSms = (
     },
   )
     .andThen(({ status, sid, errorCode, errorMessage }) => {
+      const ddTags = {
```
karrui, line 219: should we extract these tags to a centralized location? I'm wary that duplicate tags may occur down the road, resulting in fragmented logs

src/app/services/sms/sms.service.ts
```diff
@@ -237,6 +253,8 @@ const sendSms = (
         )
       }
 
+      ddClient.increment('sms.send', 1, 1, ddTags)
```
karrui, line 256: did we abandon using winston as a transport? If this PR not using winston/centralized logging is just due to the urgent nature, we may want to add a TODO here so it's more obvious

karrui (approved):

src/app/services/sms/sms.service.ts
timotheeg: This is not logging, it's metric tracking. the parent ddclient is the centralized statsd tracker

src/app/services/sms/sms.service.ts
timotheeg: I'll add a comment for this. Tags should only contain values with limited cardinality, and I don't have a good sense of the cardinality of msgSrvcSid. I included it because I thought we might be able to see patterns of failures for a specific Message Service provider, but I reckon, it's not that urgent, or can be investigated when we see a raise of errors.

src/app/services/sms/sms.service.ts
timotheeg: > should we extract these tags to a centralized location?

Probably not, tags are specific to a chunk of code we want to instrument. I expect that as we instrument different part of the codebase, we will get different tags, and that's OK. If we see a need for uniformalizxation, we can refactor later.

src/app/services/sms/sms.service.ts
timotheeg: Right, let me do that. I'll extract a twilioDDClient somewhere that include the fixed prefix so we can avoit the repetition.

I hesitated to capitalize the name but I didn't because I typically only capitalize primitive constants, not objects or instances. And I think we do the same in other parts of the code? Like `logger` or  `secretsManager` are not capitalized either?

timotheeg: https://github.com/opengovsg/FormSG/pull/3824/commits/22335bccba9c58f8b91685a45bf12a7b62f0c28b

src/app/services/sms/sms.service.ts
timotheeg: https://github.com/opengovsg/FormSG/pull/3824/commits/2b72767f95f59beb67d66c1855a03bc0bde914a2

src/app/services/sms/sms.service.ts
timotheeg: That said, since, the same tag shape is used in 2 places, I've made a shared type definition for it.
https://github.com/opengovsg/FormSG/pull/3824/commits/91634045838ad35a7fd05457906855cc2a75b1d0

Still, because these are different cases (async response vs webhook callback), it's entirely possible we track different tags as needed.

PR #3826, opened 2022-05-04
feat: bookmark compatibility routing for react and angular

timotheeg:
## Problem

People may have bookmarked URLS into forms.

Angular JS uses Hash based routing (e.g. path like `/#!/examples`), where the URL hash is NOT sent to the server. 

Because the server doesn't get to see the URL, the frontend needs to be able to identify the old URLs and redirect to the new URL scheme.

For example:

`/#!/:formid/admin` -> `/admin/forms/:formid`

## Solution
Implement a router in react that will check the hash on the base route and redirect as needed.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
+    <base href="/" />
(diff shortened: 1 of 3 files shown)


timotheeg:
Will reopen with v2-name branch

Issue #3830, opened 2022-05-05
Use official sns message validator package


karrui:
We currently check for sns validity using a bunch of heuristics. We should use the official package.

https://www.npmjs.com/package/sns-validator

mantariksh:
Closed via #3832

PR #3832, opened 2022-05-06
feat: validate SNS notification with AWS lib rather than custom implementation

timotheeg:
## Problem
We had implemented SNS validation with our custom implementation code, which unfortunately turned out to have a vulnerability (which we patched yesterday).

Instead of rolling our own validation, we should be using a library provided by AWS themselves.

Fixes https://github.com/opengovsg/FormSG/issues/3830

## Solution
Use the [sns-validator](https://github.com/aws/aws-js-sns-message-validator) library


src/app/modules/bounce/bounce.service.ts
```diff
@@ -50,109 +47,36 @@ const logger = createLoggerWithLabel(module)
...
-    !hasRequiredKeys(body) ||
-    body.SignatureVersion !== '1' ||
-    !isValidCertUrl(body.SigningCertURL)
+): ResultAsync<true, InvalidNotificationError> => {
+  return ResultAsync.fromPromise(
+    new Promise((resolve, reject) => {
+      const snsValidator = new SNSMessageValidator()
+
+      snsValidator.validate(
+        body as unknown as Record<string, unknown>,
```
timotheeg, line 65: types... 😢😑

karrui (approved):

PR #3852, opened 2022-05-10
refactor(theme): use apply keyword in theme instead of spreading

karrui:
## Problem

Recently discovered the `apply` keyword in Chakra theme props, which allows for spreading of theme properties without having to import the property itself.


karrui:
Will need chromatic to see if everything still works as expected

karrui:
NOPE DOESN"T WORK

PR #3857, opened 2022-05-11
feat(v2): landing page

### LINKED ISSUE #3655: [React/Static] Add landing page
Should be using lazy load for these since this is a distinct page from the rest of the app

karrui:
> ~~!! Still WIP, for design to check out first~~

## Problem

This PR adds the landing page of the application
Closes #3655 

## Solution

**Features**:

- Add landing page 

## Before & After Screenshots
New stories have been added for LandingPage


karrui:
animated pictures will come in a later PR once pearly is done with them

karrui:
i think this PR can just merge once design review passes; there is no real app logic in the landing page lol

timotheeg (approved):

frontend/src/app/AppRouter.tsx
```diff
@@ -41,6 +43,10 @@ export const AppRouter = (): JSX.Element => {
   return (
     <WithSuspense>
       <Routes>
+        <Route
+          path={LANDING_ROUTE}
+          element={<PublicElement element={<LandingPage />} />}
+        />
         <Route
           path={ROOT_ROUTE}
           element={<PrivateElement element={<WorkspacePage />} />}
```
timotheeg, line 52: Bloody github is not showing a correct code 😢 this line should be `HashRouterElement`.

frontend/src/constants/routes.ts
```diff
@@ -1,6 +1,5 @@
-// TODO: Update landing route to correct route.
 export const LANDING_ROUTE = '/'
-export const ROOT_ROUTE = '/'
```
timotheeg, line 3: This change breaks the bookmark routing I have added, since the hashes will never match the `/workspace` prefix 🤔

frontend/src/app/HashRouterElement.tsx
```diff
@@ -57,9 +73,14 @@ export const HashRouterElement = ({
...
-    <Navigate replace to={redirectTo} state={{ from: location }} />
+  if (isAuthenticated && strict) {
+    return <Navigate to={state?.from.pathname ?? ROOT_ROUTE} replace />
+  }
+
+  return (
+    <Flex flexDir="column" height="100vh" pos="relative">
+      <GovtMasthead />
+      {element}
+    </Flex>
```
timotheeg, line 84: /torefactor: Probably better to have `HashRouterElement` return a `PublicElement` rather than duplicate `PublicElement`'s code 🤔

frontend/src/app/HashRouterElement.tsx
```diff
@@ -54,33 +47,21 @@ const hashRouteMapper = [
...
-    return <Navigate to={state?.from.pathname ?? ROOT_ROUTE} replace />
-  }
-
-  return (
-    <Flex flexDir="column" height="100vh" pos="relative">
-      <GovtMasthead />
-      {element}
-    </Flex>
-  )
+  return <PublicElement element={element} strict={strict} />
```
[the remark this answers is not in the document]
timotheeg: Right! So when HashRouter was first introduced the `ROOT_ROUTE` was `/` and was private so `HashRouter` replicated the code of `PrivateElement`.

With the route change that this landing PR introduces, the landing zone is `/` and is public. I initially changed `HashRouter` to now replicate `PublicElement` instead if Private (in [this commit](https://github.com/opengovsg/FormSG/pull/3857/commits/6c4411220d470b404fe4313d7a88c90f437c5e79)), but I figured `HashRouter` should return the `PublicElement` instead of duplicating its code. So this is what [this latest commit](https://github.com/opengovsg/FormSG/pull/3857/commits/455fa9a4f9843ad33d5c5f2a9dc515c962f765d6) is doing. Should all work 🤞 

Some things might still warrant some discussions:
* It's a bit funny that the routes are now defined like this:
```javascript
export const LANDING_ROUTE = '/'
export const ROOT_ROUTE = '/workspace'
```
 Perhaps we should refactor the names like this:
```javascript
export const ROOT_ROUTE = '/'
export const ADMIN_ROUTE = '/workspace'
```
(just a suggestion)

* I was now a little confused about the `/workspace` url namespace 😅 . Are we now using `/workspace/admin/forms` or just `/workspace//forms`? The redirection rules in HashRouter will need to correctly reflect whatever final decision we take.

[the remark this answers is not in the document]
karrui: /workspace is its own route, once an admin goes into their form it becomes /admin/forms/:formId

i see where the confusion is, yeah, should rename `ROOT_ROUTE` for sure... maybe the workspace page can be `/admin` too, though i think `/workspace` works better.

PR #3858, opened 2022-05-11
fix: restore copyfiles

timotheeg:
We accidentally removed the copyfiles dependency, this PR restores it


tshuli (approved):

tshuli:
Need to rename branch to start with form-v2/ to avoid angular CI

PR #3861, opened 2022-05-12
feat(v2): add GovtMasthead to PublicElement component, update PP and TOU routes

karrui:
## Problem

General cleanup PR. See changes below.

## Solution

**Breaking Changes** 

**Features**:

- feat: add `GovtMasthead` component to all public form elements
- feat: add `AppPublicHeader` to TOU and PP pages
- feat: add routing to PP and TOU routes in public form footer

**Bug Fixes**:

- fix: remove strictness from Privacy and Tou page routing
  - this was erroneously preventing logged in users from accessing the privacy/tou pages

### DIFF
--- a/frontend/src/app/AppRouter.tsx
+++ b/frontend/src/app/AppRouter.tsx
@@ -51,11 +51,11 @@ export const AppRouter = (): JSX.Element => {
         />
         <Route
           path={PRIVACY_POLICY_ROUTE}
-          element={<PublicElement strict element={<PrivacyPolicyPage />} />}
+          element={<PublicElement element={<PrivacyPolicyPage />} />}
         />
         <Route
           path={TOU_ROUTE}
-          element={<PublicElement strict element={<TermsOfUsePage />} />}
+          element={<PublicElement element={<TermsOfUsePage />} />}
         />
         <Route
           path={PUBLICFORM_ROUTE}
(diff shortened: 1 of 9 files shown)


frontend/src/app/AppRouter.tsx
```diff
@@ -51,11 +51,11 @@ export const AppRouter = (): JSX.Element => {
         />
         <Route
           path={PRIVACY_POLICY_ROUTE}
-          element={<PublicElement strict element={<PrivacyPolicyPage />} />}
```
timotheeg, line 54: Just for my reference, what does `strict` controls? Is it [this](https://reactjs.org/docs/strict-mode.html)?

frontend/src/features/login/LoginPage.tsx
```diff
@@ -175,7 +175,7 @@ export const LoginPage = (): JSX.Element => {
   }
 
   return (
-    <Flex flexDir="column" minH="100vh">
+    <Flex flexDir="column" h="100%">
```
timotheeg, line 178: Sorry, some UI noob question here: Is `h` the same as `height` in the Flex component? When do we use one over the other?

karrui: yes. they are just all aliases

frontend/src/app/AppRouter.tsx
karrui: No sorry, it's this: https://github.com/opengovsg/FormSG/blob/form-v2/develop/frontend/src/app/PublicElement.tsx#L15

on hindsight `strict` is a bad name for that prop haha

frontend/src/features/login/LoginPage.tsx
timotheeg: Right I was asking because just above in this PR, we use `height` (in `PublicElement.tsx`) but with a different unit `100vh`, so I was wondering if there are subtle differences 😅

timotheeg (approved):

Issue #3439, opened 2022-02-15
[React] form 404 page


karrui:
This screen is not implemented yet

- [x] Unauthorized viewing of admin routes
- [x] Missing form
- [x] Private form (technically 401 page)
- [x] Catchall 404

karrui:
Closed by #3871

PR #3871, opened 2022-05-17
feat(v2): add 404 screens for the application, update login page design

karrui:
## Problem

This PR adds the missing 404 screens for both invalid public form and forbidden pages.
Also updates login page to updated design, and remove govt mastheads on pages that were already covered by the `PublicElement` component.

Related to #3439

## Solution

**Features**:

- feat: add FormNotFound screen for public forms
- feat: add NotFound404 page for invalid routes
- feat: add AdminForbidden403 page for forbidden form admin routes

## Before & After Screenshots
New stories for above pages and screens added


frontend/.storybook/preview.tsx
```diff
@@ -5,6 +5,7 @@ import { HelmetProvider } from 'react-helmet-async'
 import { QueryClient, QueryClientProvider } from 'react-query'
```
mantariksh, line 5: would be helpful if we can add a comment in this file saying that it's used for global storybook config, and link to the docs. I know the docs are explicit about this but helps to save a few seconds/minutes of confusion for future devs since this file isn't imported anywhere

frontend/.storybook/preview.tsx
```diff
@@ -20,7 +21,19 @@ dayjsUtils.init()
 
 const withReactQuery: DecoratorFn = (storyFn) => {
   const queryClient = new QueryClient({
-    defaultOptions: { queries: { staleTime: Infinity } },
+    defaultOptions: {
```
mantariksh, line 24: would it make sense to have a constant for the query client config in the app, and import and reuse it here? to make sure storybook react-query behaviour matches the app

frontend/src/features/admin-form/common/AdminFormLayout.tsx
```diff
@@ -1,12 +1,29 @@
-import { Outlet } from 'react-router-dom'
+import { Outlet, useParams } from 'react-router-dom'
 import { Flex } from '@chakra-ui/react'
+import { get } from 'lodash'
+
+import AdminForbidden403 from '~pages/AdminForbidden403'
```
mantariksh, line 5: small nit but might be better to name these after the purpose of the page rather than coupling to a specific status code, e.g. `NoPermissionToAdminForm` (not the best name but you get the idea)

mantariksh (approved):

frontend/.storybook/preview.tsx
karrui: there are actually some differences between the config in storybook and the actual app (we actually do not want them to be the same!)

for example, the staleness should be `infinite` in storybook. now that i'm looking at it, i should also set to zero retries in storybook (done in 3d432ab)

frontend/.storybook/preview.tsx
karrui: 921b87c

frontend/src/features/admin-form/common/AdminFormLayout.tsx
karrui: will change to AdminForbiddenErrorPage?

PR #3872, opened 2022-05-17
feat(v2): i18n - add storage of user preferred language and test string interpolations

frankchn:
## Problem
Part of #3792

## Solution
This adds support for detecting languages from a user's localStorage (key is `formsg-language`), and also adds some Chinese translations and an interpolated string to test out those functionalities.

**Breaking Changes** 

**Features**:

- Adds user preferred language detector (via configuring the `i18next-browser-languagedetector` package)
- Adds some Chinese translations
- Adds an interpolated string to translate

## Screenshots

image

### DIFF
--- a/frontend/src/i18n/locales/types.ts
+++ b/frontend/src/i18n/locales/types.ts
@@ -2,6 +2,14 @@ interface Translation {
   translation: {
     general: {
       slogan: string
+      copyright: string
+    }
+    loginPage: {
+      onlyAvailableForPublicOfficers: string
+      email: string
+      emailEmptyErrorMsg: string
+      login: string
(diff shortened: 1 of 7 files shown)


timotheeg (approved):
Looking good 👍

frontend/src/i18n/locales/types.ts
```diff
@@ -1,7 +1,21 @@
 interface Translation {
   translation: {
-    general: {
-      slogan: string
+    features: {
```
karrui, line 3: wonder if there is a way to automate this... seems like it could be a PITA to update

karrui (approved):
lgtm

PR #3875, opened 2022-05-17
fix(v2): fix some uncaught a11y issues in public form, login page

karrui:
## Problem

Fixes some a11y issues raised in design review of #3861 ([chromatic link](https://www.chromatic.com/build?appId=60911f69d4403b0039b266a2&number=1279)) 

## Solution

**Bug Fixes**:

- fix(LoginPage): remove extra htmlFor prop 
- fix(ColumnCell): use VisuallyHidden rather than display none 
- feat(AuthImageSvgr): update auth svg to a more neutral color

## Before & After Screenshots
Public form and login page stories should have (less) a11y violations


timotheeg (approved):
Needs to resolve conflicts

PR #3881, opened 2022-05-18
feat(v2): update footer component to new design

karrui:
## Problem

This PR updates the footer component to the updated design, which includes the removal of the tagline and the instagram and OGP social media links and icons.

took the chance to also make the props more generic, which allows for injection of own logos and social media links into the Footer primitive template, with OGP related defaults.


frontend/src/components/Footer/Footer.stories.tsx
```diff
@@ -2,46 +2,57 @@ import { Meta, Story } from '@storybook/react'
...
-} as Meta
-
-const defaultArgs: FooterProps = {
+const DEFAULT_ARGS: FooterProps = {
   appName: 'Form',
-  tagline: 'Build secure government forms in minutes',
   footerLinks: [
     {
-      label: 'User Guide',
+      label: 'User guide',
```
mantariksh, line 12: would it make sense to import this from the app rather than keeping it in two places? to make sure storybook reflects what's on the app

frontend/src/components/Footer/common/constants.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const DEFAULT_FOOTER_ICON_LINK: FooterLinkWithIcon = {
+  href: 'https://open.gov.sg',
+  label: 'Open Government Products homepage',
+  Icon: OgpLogoFull,
+}
+
+export const DEFAULT_SOCIAL_MEDIA_LINKS: FooterLinkWithIcon[] = [
+  {
+    href: 'https://sg.linkedin.com/company/open-government-products',
+
```
mantariksh, line 17: what's with this empty line

mantariksh (approved):

frontend/src/components/Footer/Footer.stories.tsx
karrui: This is the general Footer story, not actually the app's footer. Can be kept separate bah

PR #3894, opened 2022-05-24
feat: add a priority UI selector cookie for react rollout QA

timotheeg:
## Problem
We need the ability to rollback from angular to react if we detect something really wrong. 

One mechanism in place to do that is to make sure the router serves angular if the rollout rate is 0 (fastest way to release).

BUT rollout also has a QA phase, where we need the rollout rate to be ZERO for public while we (OGP) are still able to access forms internally.

## Solution
This PR introduces an additional priority cookie for QA that will allow us (formSG dev team) to access the react UI for forms in production, while members of the public will still 100% be on react.

Note: this adds some complexity to the dev team, the QA cookie will always takes priority if it is present, so it might confuse product ops and devs, when they try to load the UI.

### DIFF
--- a/src/app/modules/react-migration/react-migration.controller.ts
+++ b/src/app/modules/react-migration/react-migration.controller.ts
@@ -10,8 +10,8 @@ import { RedirectParams } from '../form/public-form/public-form.types'
 import * as HomeController from '../home/home.controller'
 
 export enum UiCookieValues {
-  react = 'react',
-  angular = 'angular',
+  React = 'react',
+  Angular = 'angular',
 }
 
 export type SetEnvironmentParams = {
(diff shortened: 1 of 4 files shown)


mantariksh (approved):
lgtm, small nit is it might be better to have `'react'` and `'angular'` as constants rather than hardcoding them repeatedly!

timotheeg:
I've added a commit to drop the repeated magic strings, please help me have one more look @mantariksh 🙏 
https://github.com/opengovsg/FormSG/pull/3894/commits/c3b2429f9a4de9ed69b2d4c960f3c09cf2dcfcc7

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -9,8 +9,13 @@ import * as PublicFormController from '../form/public-form/public-form.controlle
 import { RedirectParams } from '../form/public-form/public-form.types'
 import * as HomeController from '../home/home.controller'
 
+export enum UiCookieValues {
+  react = 'react',
+  angular = 'angular',
```
mantariksh, line 14: usually we use this casing by convention for enums:
```suggestion
  React = 'react',
  Angular = 'angular',
```

timotheeg: Fixed in https://github.com/opengovsg/FormSG/pull/3894/commits/ae988d85b4ef7b3d2fef96ac00dc11ad0a2e78c6

PR #3903, opened 2022-05-25
fix: prevent webhook url from pointing to app subdomain

### LINKED ISSUE #3902: Prevent webhook URLs from pointing to subdomains of app URL
Currently we prevent webhook URLs from pointing to private IP addresses or back to the app URL (`form.gov.sg` in the case of production). However, admins can still point their webhooks to subdomains, e.g. `www.form.gov.sg`. This should be blocked through webhook URL validation.

tshuli:
Closes #3902 

**Tests**
- [ ] Create storage mode form. Enter subdomain of app domain in webhook url. it should fail with the following message.

Screenshot 2022-05-25 at 9 29 20 PM

### DIFF
--- a/src/app/modules/webhook/webhook.validation.ts
+++ b/src/app/modules/webhook/webhook.validation.ts
@@ -21,20 +21,17 @@ export const validateWebhookUrl = (webhookUrl: string): Promise<void> => {
     }
     const webhookUrlParsed = new URL(webhookUrl)
     const appUrlParsed = new URL(config.app.appUrl)
-    if (webhookUrlParsed.hostname === appUrlParsed.hostname) {
+    if (
+      webhookUrlParsed.hostname === appUrlParsed.hostname ||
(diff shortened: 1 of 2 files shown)


src/app/modules/webhook/webhook.validation.ts
```diff
@@ -28,6 +28,13 @@ export const validateWebhookUrl = (webhookUrl: string): Promise<void> => {
         ),
       )
     }
+    if (webhookUrlParsed.hostname.endsWith(appUrlParsed.hostname)) {
```
mantariksh, line 31: the hostname is `form.gov.sg` right, so technically this should be `'.' + appUrlParsed.hostname` right? plus (nit) we can probably merge this and the above block with a `||`, and just use a common error message

tshuli: that is correct, thanks for the catch! I've made edits

mantariksh (approved):

PR #3909, opened 2022-05-26
chore: update Storybook to v6.5

karrui:
This PR runs `npx sb upgrade` to upgrade all storybook related packages to 6.5.


karrui:
quick note, i'll be going on an upgrading spree... might try to go react 18 so we can move to chakra v2

tshuli (approved):
lgtm

PR #3910, opened 2022-05-26
feat: React 18 and chakra v2 upgrade

karrui:
## Problem

This PR upgrades the react version to 18. Also updates packages who now support react 18. Does not really have to be merged I guess; this PR is merely an exercise in seeing whether it is possible.

I have not updated the test suites to React 18, since they depend on storybook rendering and that may not be all there yet. The fallback to React 17 functionality in tests seems to still work well.


karrui:
closing, will try again next time :)

PR #3915, opened 2022-05-27
feat(v2): trim email when logging in

karrui:


siddarth2824 (approved):
LGTM!

mantariksh:
@siddarth2824 @karrui I feel like we should be doing this on the server also

karrui:
> @siddarth2824 @karrui I feel like we should be doing this on the server also

Should we create an issue for this?

PR #3919, opened 2022-05-30
feat(form-v2): add myInfo types and constants

seaerchin:
Part 1 of #3926

## Problem
Forms in react currently does not have any myinfo fields implemented, despite the angular version of the application supporting myinfo. This leads to a feature gap between the two versions of the application

Related to #3360 

## Solution
This PR adds MyInfo related constants and typings. These are as follows: 
1. drag/drop related constants
2. form preview related constants (the rendered form in the right panel) - this includes the label/icon etc 
3. adding a `MyInfoImplemented/MyInfoUnimplemented` type so that we encode type-safety into the methods and no callers can mistakenly call a myinfo method wrongly (this is because the figma has more myinfo fields than currently exists)

Issue #3360, opened 2022-02-03
[React/Builder] MyInfo fields integration


karrui:
Should look the same as the Basic field tab, but when dragged and dropped should have prefilled text. Will ask @pearlyong for mocks

karrui:
Closed by #3926, #3948, #4028

PR #3926, opened 2022-05-31
feat(form-v2): myinfo implementation 

seaerchin:
Part 2 of #3919
## Problem
see #3919 for an overview of the problem 

Related to #3360 

## Solution
This PR implements the myinfo formfields in the create tab of the admin form page. This is done through
1. adding a method `getMyInfoFieldCreationMeta` so that on drag/drop end, there can be a form field returned.
    - this method groups myinfo fields together by their basic field types as only a few basic fields are myinfoable. this allows for resharing of common props apart from dropdown types, where the `fieldOptions` have to be generated based on the `MyInfoAttribute`
2. add a new `MyInfoFieldOption` and extending the draggable functionality to it. not sure if htere's a way to write this better but the meta of the field is based on `MyInfoAttribute` and not the basic field type and hence, there wasn't a "quick" way to reuse code from `BasicFieldOption`
3. add a new `EditMyInfo` component that displays visual information about the myinfo field type selected


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/EditFieldDrawer.tsx
```diff
@@ -60,8 +62,10 @@ export const EditFieldDrawer = (): JSX.Element | null => {
 
   const basicFieldText = useMemo(() => {
     if (!fieldToEdit?.fieldType) return ''
+    if (isMyInfo(fieldToEdit))
+      return MYINFO_FIELD_CONSTANTS[fieldToEdit.myInfo.attr].value
```
karrui, line 66: nit: wrap with braces since its multiline for cleanliness.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignTab.tsx
```diff
@@ -75,12 +86,50 @@ export const BuilderAndDesignTab = (): JSX.Element => {
...
+        }
+
+        case CREATE_MYINFO_MARRIAGE_DROP_ID: {
+          return setToCreating(
+            getMyInfoFieldCreationMeta(
+              CREATE_MYINFO_MARRIAGE_FIELDS_ORDERED[source.index],
+            ),
+            destination.index,
+          )
+        }
```
karrui, line 131: do we actually need all these separate ids? can we not group them in one like how we do it for basic fields?

frontend/src/features/admin-form/create/builder-and-design/constants.ts
```diff
@@ -95,3 +89,10 @@ export const CREATE_MYINFO_PARTICULARS_DROP_ID = 'create-myinfo-particulars'
 export const CREATE_MYINFO_MARRIAGE_DROP_ID = 'create-myinfo-marriage'
 export const FIELD_LIST_DROP_ID = 'formFieldList'
 export const PENDING_CREATE_FIELD_ID = 'FIELD-PENDING-CREATION'
+
+export const MOCK_MYINFO_IMPLEMENTED_TYPES: MyInfoImplementedTypes[] = [
```
karrui, line 93: Mocks should not be located in app code, should we move this?

frontend/src/features/admin-form/create/builder-and-design/utils/fieldCreation.ts
```diff
@@ -160,3 +175,87 @@ export const getFieldCreationMeta = (fieldType: BasicField): FieldCreateDto => {
...
+    case MyInfoAttribute.HousingType:
+    case MyInfoAttribute.HdbType:
+    case MyInfoAttribute.Occupation:
+    case MyInfoAttribute.WorkpassStatus:
+    case MyInfoAttribute.Marital:
+    case MyInfoAttribute.CountryOfMarriage: {
+      return {
+        ...baseMeta,
+        fieldType: BasicField.Dropdown,
+        ...getMyInfoDropdownFieldMeta(myInfoAttribute),
```
karrui, line 236: i don't think we should inject any dropdown fields in field creation. we don't actually want to save the fields in the database (unless i am misunderstanding what this function does)

frontend/src/features/admin-form/create/builder-and-design/utils/fieldCreation.ts
```diff
@@ -160,3 +175,87 @@ export const getFieldCreationMeta = (fieldType: BasicField): FieldCreateDto => {
...
+      }
+    }
+
+    case MyInfoAttribute.MarriageCertNo: {
+      return {
+        ...baseMeta,
+        fieldType: BasicField.Number,
```
karrui, line 243: i think this should still be of shorttext type, doesn't make sense for the number to be incrementable or decrementable haha

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/EditFieldDrawer.tsx
seaerchin: sounds good, will do!

frontend/src/features/admin-form/create/builder-and-design/constants.ts
seaerchin: thanks for the catch! i think my brain short-circuited and i forgot that `mocks/msw` existed even though this is only used there oof

frontend/src/features/admin-form/create/builder-and-design/utils/fieldCreation.ts
seaerchin: good catch, didn't realize this - fixed!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignTab.tsx
seaerchin: sorry, could i get some more clarity on this ah cos i think it's implemented in a similar way ._.

karrui: oh nevermind, it is indeed implemented in a similar way in basicfields haha. ignore me

frontend/src/features/admin-form/create/builder-and-design/utils/fieldCreation.ts
karrui: even if they are number-only, makes sense to be used as shorttext regardless just in case haha.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/useBuilderFields.ts
```diff
@@ -64,5 +66,7 @@ export const useBuilderFields = () => {
     return existingFields
   }, [formData, stateData])
 
-  return { builderFields }
+  return {
+    builderFields: builderFields?.map((field) => augmentWithMyInfo(field)),
```
karrui, line 70: should we map this inside the memoized builderfields constant instead of here?

frontend/src/features/admin-form/create/builder-and-design/utils/fieldCreation.ts
```diff
@@ -9,9 +9,9 @@ import {
   RatingShape,
 } from '~shared/types/field'
 import {
-  getMyInfoDropdownFieldMeta,
   MYINFO_FIELD_CONSTANTS,
   myInfoDateFieldMeta,
+  myInfoDropdownFieldMeta,
   myInfoMobileFieldMeta,
   myInfoTextFieldMeta,
```
karrui, line 16: i suspect these meta constants do not actually need to be shared. should we move them into the client side? 
nit: should also be in UPPER_CASE

seaerchin: good point, changed and relocated them!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/useBuilderFields.ts
seaerchin: good catch, changed accordingly

seaerchin:
i'm getting an error on myInfo field creation: 
image

this seems to be because the `useEffect` has an extra dependency of the field id when the field is created. i can't tell where this is coming from as the there's no `useEffect` added as far as i can tell ._. any ideas?

karrui:
> i'm getting an error on myInfo field creation: 
> 
> image
> 
> 
> 
> this seems to be because the `useEffect` has an extra dependency of the field id when the field is created. i can't tell where this is coming from as the there's no `useEffect` added as far as i can tell ._. any ideas?

This is from react-hook-form; some dynamically generated field does not have a default value; add it during register or in the field creation step

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/useBuilderFields.ts
```diff
@@ -53,7 +55,9 @@ export const useBuilderFields = () => {
   const { data: formData } = useCreateTabForm()
   const stateData = useBuilderAndDesignStore(stateDataSelector)
   const builderFields = useMemo(() => {
-    const existingFields = formData?.form_fields
+    const existingFields = formData?.form_fields?.map((field) =>
+      augmentWithMyInfo(field),
+    )
```
karrui, line 60: nit: can simplify 

```ts
const existingFields = formData?.form_fields?.map(augmentWithMyInfo)
```

karrui (approved):
lgtm

can we have another PR after this to add MyInfo field preview like the current AngularJS app?

seaerchin:
will update types in the original PR, rebase and then address the warning before merge! can i clarify @karrui wym by the preview?

karrui:
> will update types in the original PR, rebase and then address the warning before merge! can i clarify @karrui wym by the preview?

you know, like PHUA CHU KANG in the preview in the admin form builder side haha

karrui:
stories seem to have unexpected changes

karrui:
@seaerchin any updates on the story regressions?

PR #3927, opened 2022-05-31
fix: server to randomize image and logo filenames before generating presigned url

tshuli:
## Problem
Fixes [#128](https://github.com/datagovsg/formsg-private/issues/128)

## Solution
- Backend prefixes random string to image and logo filenames before returning presigned url, hence ensuring filename in bucket is unpredictable to client when `/presign` endpoint is first called
- Backward compatibility is ensured by setting `isNewClient` flag for the frontend, which will be removed at a later date

## Tests
- [ ] Create storage mode form. Test that logo and image upload works and can be displayed in public form. 
- [ ] (safety check, should not be affected by this PR) Add attachment field. Test that file upload works and can be retrieved from storage mode  
- [ ] Repeat with old client
(this can be done by rolling back to commit `d979d0c1`)

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
@@ -4906,9 +4906,17 @@ describe('admin-form.routes', () => {
       fileId: 'some file id',
       fileMd5Hash: SparkMD5.hash('test file name'),
       fileType: VALID_UPLOAD_FILE_TYPES[0],
+      isNewClient: true, // TODO (#128): Flag for server to know whether to append random object ID in front. To remove 2 weeks after release.
     }
 
-    it('should return 200 with presigned POST URL object', async () => {
+    const DEFAULT_POST_PARAMS_OLD_CLIENT = {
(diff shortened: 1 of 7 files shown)


src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
```diff
@@ -4929,7 +4937,43 @@ describe('admin-form.routes', () => {
...
+          acl: 'public-read',
+          bucket: expect.any(String),
+        }),
+      })
+      expect(response.body.fields.key).toEqual(
+        expect.stringContaining(DEFAULT_POST_PARAMS.fileId),
+      )
+      expect(DEFAULT_POST_PARAMS.fileId.length).toBeLessThan(
+        response.body.fields.key.length,
+      )
```
mantariksh, line 4951: any way to make this stricter? maybe do a regex check on the first 24 chars for the ObjectId

mantariksh (approved):

src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
tshuli: thanks! done that

PR #3937, opened 2022-06-02
feat: log ip address on callback from twilio

tshuli:
## Problem
- Currently, we are unable to match the sender IP address to the twilio sms delivery report

## Solution
- pass sender IP address to twilio as part of query param for callback

## Tests

- [ ] Create form. Test that email and mobile field OTP verification and submission works
- [ ] Check that twilio webhooks contain sender IP address after OTP has been sent
- [ ] Check that admin contact number adding works normally

### DIFF
--- a/src/app/modules/twilio/twilio.controller.ts
+++ b/src/app/modules/twilio/twilio.controller.ts
@@ -54,6 +54,12 @@ export const twilioSmsUpdates: ControllerHandler<
    * Example: https://www.twilio.com/docs/usage/webhooks/sms-webhooks.
    */
 
+  // Extract public sender's ip address which was passed to twilio as a query param in the status callback
+  const url = req.url.split('?')
+  const paramsString = url.length > 1 ? url[1] : ''
+  const queryParams = new URLSearchParams(paramsString)
+  const senderIp = queryParams.get('senderIp')
+
   const ddTags: TwilioSmsStatsdTags = {
     // msgSrvcSid not included to limit tag cardinality (for now?)
     smsstatus: req.body.SmsStatus,
@@ -70,6 +76,7 @@ export const twilioSmsUpdates: ControllerHandler<
       meta: {
         action: 'twilioSmsUpdates',
         body: req.body,
+        senderIp,
       },
     })
   } else {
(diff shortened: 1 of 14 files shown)


hanstirtaputra (dismissed):
LGTM

mantariksh (commented):
as discussed, would be better to pass the IP as a param to Twilio

tshuli:
> as discussed, would be better to pass the IP as a param to Twilio

Thanks! I changed the implementation to pass it as a query param in the callback

src/app/modules/twilio/twilio.controller.ts
```diff
@@ -54,6 +54,12 @@ export const twilioSmsUpdates: ControllerHandler<
    * Example: https://www.twilio.com/docs/usage/webhooks/sms-webhooks.
    */
 
+  // Extract public sender's ip address which was passed to twilio as a query param in the status callback
+  const url = req.url.split('?')
```
mantariksh, line 58: hmm is it necessary to do this manually? can't we use the NodeJS [URL API](https://nodejs.org/api/url.html)?

src/app/modules/twilio/twilio.controller.ts
```diff
@@ -54,6 +54,12 @@ export const twilioSmsUpdates: ControllerHandler<
...
+  const paramsString = url.length > 1 ? url[1] : ''
+  const queryParams = new URLSearchParams(paramsString)
+  const senderIp = queryParams.get('senderIp')
```
mantariksh, line 61: according to the [docs](https://nodejs.org/api/url.html#urlsearchparamsgetname) this can be null right? do we need to deal with that case?

src/app/modules/twilio/twilio.controller.ts
tshuli: thanks! have done so

src/app/modules/twilio/twilio.controller.ts
tshuli: yes - it will be null if there is no senderIp (e.g. if it is a bounce sms notification from our server). thereafter we pass it to the logger

mantariksh (approved):

karrui:
will need to fix tests

PR #3941, opened 2022-06-02
feat: add backend validation for form feedback submission

hanstirtaputra:
## Problem
This enforces backend validation form feedback submissions. It's in line with the UI behaviour where we can only submit feedback once per submission

## Solution
- Add new `/api/v3/forms/{formId}/submissions/{submissionId}/feedback` endpoint to support this
- Link Angular FE to the new feedback endpoint

Created new controller and route instead of replacing the pre-existing one as we still want to preserve the old `/api/v3/forms/{formId}/feedback` route

**Breaking Changes** 

## Tests
- [x] Test that submitting a feedback with `submissionId` for a form without a prior form submission with the same `submissionId` returns `404`
- [x] Test that submitting feedback twice for the same `submissionId` and `formId` returns `422`
- [x] Test that feedback using only endpoint of `/api/v3/forms/{formId}/feedback` still works
- [x] Test that old client works with new server both locally and on staging (load the submit feedback page on the old client, deploy the new version of the code, then submit feedback to see that feedback submission still works)


shared/types/form/form_feedback.ts
```diff
@@ -17,6 +17,7 @@ export type FormFeedbackBase = {
   formId: FormDto['_id']
   created?: Date
   lastModified?: Date
+  formSubmissionId: string
```
mantariksh, line 20: can rename to `submissionId` in line with the rest of the codebase. at runtime, is this actually a string or is it an `ObjectId` object?

src/app/models/form_feedback.server.model.ts
```diff
@@ -14,6 +15,12 @@ const FormFeedbackSchema = new Schema<IFormFeedbackSchema, IFormFeedbackModel>(
       ref: FORM_SCHEMA_ID,
       required: true,
     },
+    formSubmissionId: {
+      type: Schema.Types.ObjectId,
+      ref: SUBMISSION_SCHEMA_ID,
+      // TODO: Update to true once we fully migrate to /submissions/{submissionId}/feedback endpoint
+      required: false,
```
mantariksh, line 22: don't think it'll be possible to do this. even after we deprecate the old endpoint, there will still be old feedback documents in the database which don't have this key. how does `mongoose` behave if you state that a property is `required` then attempt to retrieve a document which doesn't have the property?

src/app/modules/feedback/__tests__/feedback.service.spec.ts
```diff
@@ -248,4 +248,64 @@ describe('feedback.service', () => {
...
+  describe('checkHasPreviousFeedback', () => {
+    const MOCK_FORM_ID = new ObjectId().toHexString()
+    const MOCK_SUBMISSION_ID = new ObjectId().toHexString()
+
+    beforeEach(async () => {
+      await dbHandler.clearCollection(FormFeedback.collection.name)
+    })
+    afterEach(() => jest.clearAllMocks())
+
+    it('should return true when previous feedback exists', async () => {
```
mantariksh, line 261: the test description isn't very clear - maybe something like `should return true when feedback already exists for given submision ID`?

src/app/modules/feedback/__tests__/feedback.service.spec.ts
```diff
@@ -248,4 +248,64 @@ describe('feedback.service', () => {
...
+      const actualResult = await FeedbackService.checkHasPreviousFeedback(
+        MOCK_FORM_ID,
+        MOCK_SUBMISSION_ID,
+      )
+
+      expect(actualResult.isOk()).toEqual(true)
+      expect(actualResult._unsafeUnwrap()).toEqual(true)
+    })
+
+    it('should return false previous feedback does not exist', async () => {
```
mantariksh, line 278: similarly test description not really clear, maybe `should return false if there is no existing feedback for given submission ID`

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -0,0 +1,143 @@
...
+  [Segments.BODY]: Joi.object()
+    .keys({
+      rating: Joi.number().min(1).max(5).cast('string').required(),
+      comment: Joi.string().allow('').required(),
+    })
+    // Allow other keys for backwards compability as frontend might put extra keys in the body.
+    .unknown(true),
+})
+
+export const submitFormFeedback: ControllerHandler<
```
mantariksh, line 28: can we add jsdocs for this and also differentiate its name from the existing controller, even it's just something like `submitFormFeedbackV2`? will be super confusing to have two functions with the same name in different modules. we can also mark the old one as `@deprecated`

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  const isSubmissionIdValid = checkDoesSubmissionIdExistRes.value
+  if (!isSubmissionIdValid) {
+    return res
+      .status(StatusCodes.NOT_FOUND)
+      .json({ message: 'SubmissionId is not valid' })
+  }
+
+  const checkHasPreviousFeedbackRes =
+    await FeedbackService.checkHasPreviousFeedback(formId, submissionId)
```
mantariksh, line 63: would it be possible to chain these async calls with `andThen` for better readability?

mantariksh: hopefully that would allow us to log and return errors just once, all the way at the end

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -0,0 +1,143 @@
...
+    })
+    const { errorMessage, statusCode } = mapRouteError(error)
+    return res.status(statusCode).json({ message: errorMessage })
+  }
+
+  const isSubmissionIdValid = checkDoesSubmissionIdExistRes.value
+  if (!isSubmissionIdValid) {
+    return res
+      .status(StatusCodes.NOT_FOUND)
+      .json({ message: 'SubmissionId is not valid' })
```
mantariksh, line 59: can we create a custom error class for this, like we do in the rest of the app? then we can add the custom error class as a case in `mapRouteError`. we do this so that we can see all the error messages being returned to the user in one place.

mantariksh: in addition, `SubmissionId is not valid` is not a very informative error message for an end user. we don't expect this case to happen for legit users, but it still might if our code happens to have a bug. as a general rule all error messages returned to the user should be informative, even if it's something generic like "Sorry, something went wrong. Please try again."

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -0,0 +1,143 @@
...
+    })
+    const { errorMessage, statusCode } = mapRouteError(error)
+    return res.status(statusCode).json({ message: errorMessage })
+  }
+
+  const hasPreviousFeedback = checkHasPreviousFeedbackRes.value
+  if (hasPreviousFeedback) {
+    return res
+      .status(StatusCodes.UNPROCESSABLE_ENTITY)
+      .json({ message: 'Multiple feedbacks has already been submitted' })
```
mantariksh, line 80: similarly not a helpful error message for the user to see. maybe something like 'You have already submitted feedback for this submission. To provide additional feedback, please contact the form administrator."

src/app/modules/feedback/feedback.service.ts
```diff
@@ -111,3 +111,27 @@ export const getFormFeedbacks = (
     }
   })
 }
+
+export const checkHasPreviousFeedback = (
```
mantariksh, line 115: can we add jsdocs for all new functions? also I think this function name conveys exactly what this function is doing. maybe something like `hasPrevFeedbackForSubmission`? usually we prefix functions which return booleans with `has` or `does`

src/app/modules/feedback/feedback.service.ts
```diff
@@ -111,3 +111,27 @@ export const getFormFeedbacks = (
...
+  formId: string,
+  submissionId: string,
+): ResultAsync<boolean, DatabaseError> =>
+  ResultAsync.fromPromise(
+    FormFeedbackModel.exists({
```
mantariksh, line 120: good use of the `mongoose` API!

src/app/modules/form/public-form/public-form.service.ts
```diff
@@ -42,6 +44,7 @@ export const insertFormFeedback = ({
   return ResultAsync.fromPromise(
     FormFeedbackModel.create({
       formId,
+      formSubmissionId,
```
mantariksh, line 47: what's the behaviour if this is `undefined` (i.e. if the old endpoint is used)? does the property not get created in the db at all, or does it get created with an `undefined` value?

src/app/modules/submission/__tests__/submission/submission.service.spec.ts
```diff
@@ -721,4 +721,66 @@ describe('submission.service', () => {
...
+
+  describe('checkDoesSubmissionIdExist', () => {
+    const MOCK_SUBMISSION_ID = MOCK_SUBMISSION._id
+
+    beforeEach(async () => {
+      await dbHandler.clearCollection(Submission.collection.name)
+    })
+    afterEach(() => jest.clearAllMocks())
+
+    it('should return true with valid submissionId', async () => {
```
mantariksh, line 733: maybe `should return true if submissionId already exists`?

src/app/modules/submission/__tests__/submission/submission.service.spec.ts
```diff
@@ -721,4 +721,66 @@ describe('submission.service', () => {
...
+
+      const actualResult = await SubmissionService.checkDoesSubmissionIdExist(
+        MOCK_SUBMISSION_ID,
+      )
+
+      expect(actualResult.isOk()).toEqual(true)
+      expect(actualResult._unsafeUnwrap()).toEqual(true)
+    })
+
+    it('should return false with invalid submissionId', async () => {
```
mantariksh, line 752: `should return false if submissionId does not exist`

src/app/routes/api/v3/forms/__tests__/public-forms.submissions.feedback.routes.spec.ts
```diff
@@ -0,0 +1,219 @@
+import { ObjectId } from 'mongodb'
```
mantariksh, line 1: for the file naming, I don't think we need to nest it so deeply. we can just add this to the feedback integration tests right?

src/app/routes/api/v3/forms/public-forms.routes.ts
```diff
@@ -13,3 +14,5 @@ PublicFormsRouter.use(PublicFormsFeedbackRouter)
 PublicFormsRouter.use(PublicFormsFormRouter)
 PublicFormsRouter.use(PublicFormsAuthRouter)
 PublicFormsRouter.use(PublicFormsVerificationRouter)
+// TODO: Cleanup PublicFormsFeedbackRouter once it's fully migrated to PublicFormSubmissionsFeedbackRouter
```
mantariksh, line 17: for all TODOs, create a github issue with more context, assign yourself and link to it (e.g. `// TODO #3980`). if every engineer creates loose TODOs, the codebase will just accumulate them and no one will get to them

src/app/routes/api/v3/forms/public-forms.submissions.feedback.routes.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @produces application/json
+ * @returns 200 if feedback was successfully saved
+ * @returns 400 if form feedback was malformed and hence cannot be saved
+ * @returns 404 if form with formId or submissionId does not exist, or form is private
+ * @returns 410 if form has been archived
+ * @returns 422 if form feedback for the submissionId has already been submitted
+ * @returns 500 if database error occurs
+ */
+PublicFormSubmissionsFeedbackRouter.route(
+  '/:formId([a-fA-F0-9]{24})/submissions/:submissionId([a-fA-F0-9]{24})/feedback',
```
mantariksh, line 24: route looks good

tests/unit/backend/helpers/jest-db.ts
```diff
@@ -269,6 +273,58 @@ const getFullFormById = async (
...
+  return {
+    submission: submission as ISubmissionSchema,
+  }
+}
+
+const insertFormFeedback = async ({
+  formId,
+  formSubmissionId,
+  rating = '5',
+  comment = 'FormSG rocks!',
```
mantariksh, line 306: does it though

mantariksh (commented):
a few more thoughts:
1. have you tested that the old client works with the new server? would be good to do so both locally and on staging (load the submit feedback page on the old client, deploy the new version of the code, then submit feedback and make sure it works as expected)
2. do you think we need to create an index on the new `submissionId` field?
3. what are the pros and cons of this approach vs creating a [unique constraint](https://www.mongodb.com/docs/v4.4/core/index-unique/) on the `submissionId`?

src/app/models/form_feedback.server.model.ts
hanstirtaputra: I think there won't be an issue when retrieving a document which doesn't have a `required` property. The behaviour is that the documents will still be fetched.

How I tested:
1. Populate DB with `formfeedback` documents without the submissionId by using the old endpoint (and `submissionId` still not `required`)
2. Update `submissionId` to be `required`
3. Fetch all form feedbacks using `api/v3/admin/forms/{formId}/feedback` endpoint, also tested from FE by going to the feedback page. Outcome is that `formfeedback`s are successfully retrieved

shared/types/form/form_feedback.ts
hanstirtaputra: I think it's a `string`, but I'll update it to use `SubmissionResponseDto['submissionId']` instead of `string` so that it's directly linked to the `SubmissionResponseDto`

src/app/modules/feedback/feedback.controller.ts
hanstirtaputra: Right okay, I named it the same because I thought we would remove the old one soon enough, but now I think your suggestion is better, thanks!

src/app/modules/form/public-form/public-form.service.ts
hanstirtaputra: The property does not get created in the db at all. This doesn't cause an issue with adding feedbacks and also getting feedbacks using both the new and old endpoints

src/app/modules/feedback/feedback.controller.ts
hanstirtaputra: Updated the PR to chain `andThen` with `mapErr` for better readability

src/app/modules/feedback/feedback.controller.ts
hanstirtaputra: Okay, updated the PR to add custom error classes, with more informative error messages!

src/app/modules/feedback/feedback.service.ts
hanstirtaputra: Updated the PR to add jsdocs, and rename the functions

src/app/routes/api/v3/forms/__tests__/public-forms.submissions.feedback.routes.spec.ts
hanstirtaputra: Yep we can! Updated the PR to move the tests to `public-forms.feedback.routes.spec.ts`.

Initially separated them because of the different route file names. The tests are mostly the same, one for the old API and one for the new one

src/app/routes/api/v3/forms/public-forms.routes.ts
hanstirtaputra: Got it, I created an [issue](https://github.com/opengovsg/FormSG/issues/3964) and linked it. Thanks

hanstirtaputra:
> have you tested that the old client works with the new server? would be good to do so both locally and on staging (load the submit feedback page on the old client, deploy the new version of the code, then submit feedback and make sure it works as expected)

Yep tested both locally and on staging and it works as expected - feedbacks are still submitted and reflected in the feedback data. Upon reload it switches to the new endpoint and continues to work with the same behaviour.

> do you think we need to create an index on the new submissionId field?

Yes I think so since we will be searching by `submissionId` in our validation, I've updated the PR to add an index on `submissionId` field for `formfeedback` collection

tests/unit/backend/helpers/jest-db.ts
```diff
@@ -269,6 +273,58 @@ const getFullFormById = async (
...
+  encryptedContent = 'encryptedContent',
+}: {
+  formId?: ObjectID
+  submissionType?: string
+  version?: string
+  encryptedContent?: string
+} = {}): Promise<{
+  submission: ISubmissionSchema
+}> => {
+  const SubmissionFormModel = getSubmissionModel(mongoose)
```
mantariksh, line 289: the `Form` in the name isn't necessary, plus it makes sense for `getSubmissionModel` to return something called `SubmissionModel` haha
```suggestion
  const SubmissionModel = getSubmissionModel(mongoose)
```

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -59,15 +59,7 @@ const validateSubmitFormFeedbackParams = celebrate({
...
- * NOTE: This is exported solely for unit testing
- * Handler for POST /:formId/feedback endpoint
- * @precondition formId should be present in req.params.
- * @precondition Joi validation should enforce shape of req.body before this handler is invoked.
- *
- * @returns 200 if feedback was successfully saved
- * @returns 404 if form with formId does not exist or is private
- * @returns 410 if form has been archived
- * @returns 500 if database error occurs
+ * @deprecated use submitFormFeedbackV2 instead
```
mantariksh, line 62: we can add a TODO comment here linking to the cleanup issue you created, so we can just search for the issue number and see all the places where we need to clean up

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @precondition formId and submissionId should be present in req.params.
+ * @precondition Joi validation should enforce shape of req.body before this handler is invoked.
+ *
+ * @returns 200 if feedback was successfully saved
+ * @returns 404 if form with formId does not exist or is private, or submissionId does not exist
+ * @returns 422 if duplicate feedback with the same submissionId and formId exists
+ * @returns 410 if form has been archived
+ * @returns 500 if database error occurs
+ */
+const submitFormFeedbackV2: ControllerHandler<
```
mantariksh, line 39: can add a TODO here to remove the V2 when we do the cleanup

src/app/modules/feedback/feedback.error.ts
```diff
@@ -0,0 +1,17 @@
+import { ApplicationError } from '../core/core.errors'
```
mantariksh, line 1: nit on file naming: the convention is `.errors.ts`

src/app/modules/feedback/feedback.error.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class InvalidSubmissionIdError extends ApplicationError {
+  constructor(
+    message = 'Feedback submissionId can not be found in the form submissions. Please submit a form before you submit a feedback.',
```
mantariksh, line 5: the user doesn't know what a `submissionId` is, so people who see this would likely be confused. perhaps we can just use a generic error message, like "Sorry, something went wrong. Please try again."?

src/app/modules/feedback/feedback.util.ts
```diff
@@ -0,0 +1,63 @@
...
+import {
+  DuplicateFeedbackSubmissionError,
+  InvalidSubmissionIdError,
+} from './feedback.error'
+
+const logger = createLoggerWithLabel(module)
+
+export const mapRouteError: MapRouteError = (
+  error: ApplicationError,
+  coreErrorMessage?: string,
```
mantariksh, line 17: can we provide a default here e.g. `'Sorry, something went wrong. Please try again.'` so it can be used for unexpected cases like `DatabaseError`?

src/app/modules/feedback/feedback.util.ts
```diff
@@ -0,0 +1,63 @@
...
+) => {
+  switch (error.constructor) {
+    case FormErrors.FormNotFoundError:
+      return {
+        statusCode: StatusCodes.NOT_FOUND,
+        errorMessage: error.message,
```
mantariksh, line 23: can we go through each error message to ensure that it makes sense to the user in the context of form feedback? e.g. in this case it's just "Form not found". if a user submits the form, and while they're typing feedback the form admin makes it private, then the user tries to submit the feedback, they will see "Form not found" and be confused. maybe we can make this error message specific to this context, e.g. "This form has been made private, so feedback submissions are no longer accepted."

src/app/modules/feedback/feedback.util.ts
```diff
@@ -0,0 +1,63 @@
...
+      }
+    case FormErrors.FormDeletedError:
+      return {
+        statusCode: StatusCodes.GONE,
+        errorMessage: error.message,
```
mantariksh, line 28: similar to the previous comment, would be good to go through all these error messages and make sure that they make sense in the context of form feedback.

src/app/modules/feedback/feedback.util.ts
hanstirtaputra: Updated the PR for clearer error messages in the context of feedback, thanks!

src/app/modules/feedback/feedback.service.ts
```diff
@@ -111,3 +113,44 @@ export const getFormFeedbacks = (
...
+ * @returns err(InvalidSubmissionIdError) if submissionId does not exist amongst all the form submissions
+ * @returns err(DatabaseError) if database query errors
+ */
+export const hasNoPreviousFeedback = (
+  formId: string,
+  submissionId: string,
+): ResultAsync<true, DuplicateFeedbackSubmissionError | DatabaseError> =>
+  ResultAsync.fromPromise(
+    FormFeedbackModel.exists({
+      formId: formId,
```
mantariksh, line 135: why do we need to search for the formId here?

mantariksh:
>Test that a non-existent submissionId returns 404
can we make this test clearer pls? other engineers should be able to carry out the tests without having read your code changes

src/app/models/form_feedback.server.model.ts
```diff
@@ -14,6 +15,13 @@ const FormFeedbackSchema = new Schema<IFormFeedbackSchema, IFormFeedbackModel>(
       ref: FORM_SCHEMA_ID,
       required: true,
     },
+    submissionId: {
+      type: Schema.Types.ObjectId,
+      ref: SUBMISSION_SCHEMA_ID,
+      // TODO #3964: Update to true once we fully migrate to /submissions/{submissionId}/feedback endpoint
+      required: false,
+      index: true,
```
mantariksh, line 23: just for consistency across the codebase, can we use the `.index` syntax for this? can look at the other model files for examples

src/app/modules/feedback/feedback.service.ts
hanstirtaputra: Oh, we don't need actually since `submissionId` is probably unique, updated PR to remove `formId` from the parameters

src/app/routes/api/v3/forms/__tests__/public-forms.feedback.routes.spec.ts
```diff
@@ -157,3 +166,209 @@ describe('public-form.feedback.routes', () => {
     })
   })
 })
+
+describe('public-form.submissions.feedback.routes', () => {
```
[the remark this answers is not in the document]
hanstirtaputra: The two `describe`s are testing two different endpoints now (`/forms/:formId/feedback` and `/forms/:formId/submissions/:submissionId/feedback`), and the `public-form.submissions.feedback.routes` is named appropriately according to the route file name. I'm thinking to update the path in the next PR for the [cleanup](https://github.com/opengovsg/FormSG/issues/3964)! I've added a TODO comment here to make it clear

shared/types/form/form_feedback.ts
```diff
@@ -17,6 +18,7 @@ export type FormFeedbackBase = {
   formId: FormDto['_id']
   created?: Date
   lastModified?: Date
+  submissionId: SubmissionResponseDto['submissionId']
```
mantariksh, line 21: this should be optional right, because old feedback documents won't contain the `submissionId` key?

src/app/routes/api/v3/forms/__tests__/public-forms.feedback.routes.spec.ts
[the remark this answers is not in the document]
mantariksh: the file naming is more of a way to keep the files small than to mirror the routes exactly. actually I think we should combine the current `public-form.submissions.feedback.routes.ts` into `public-form.feedback.routes.ts`. beyond a certain point nesting the filename just becomes more confusing, and I think `public-form.feedback.routes.ts` makes it easier for other devs to find the routes related to feedback.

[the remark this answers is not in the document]
mantariksh: the file naming is more of a way to keep the files small than to mirror the routes exactly. actually I think we should combine the current `public-form.submissions.feedback.routes.ts` into `public-form.feedback.routes.ts`. beyond a certain point nesting the filename just becomes more confusing, and I think `public-form.feedback.routes.ts` makes it easier for other devs to find the routes related to feedback.

src/app/routes/api/v3/forms/public-forms.submissions.feedback.routes.ts
```diff
@@ -0,0 +1,25 @@
+import { Router } from 'express'
```
mantariksh, line 1: I think we should add this to the existing `public-forms.feedback.routes.ts`

mantariksh, line 1: I think we should add this to the existing `public-forms.feedback.routes.ts`

mantariksh (commented):
almost there, just a couple of nits!

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+      }).map(() =>
+        res
+          .status(StatusCodes.OK)
+          .json({ message: 'Successfully submitted feedback' }),
+      )
+    })
+    .mapErr((error) => {
+      const { errorMessage, statusCode } = mapRouteError(error)
+      logger.error({
+        message: errorMessage,
```
mantariksh: the error message in the log is meant for us, not the user. this should be some string like `'Error while submitting form feedback'`

mantariksh (approved):
lgtm other than one nit

PR #3942, opened 2022-06-02
feat(v2/SingleSelect): prevent clearing of selected item when isClearable is true, prevent clearing of EditRating select fields

### LINKED ISSUE #3932: Remove ability to clear dropdowns in field editing forms
e.g. Clicking the "X" in the "Number of steps" dropdown while editing Rating fields results in weird behaviour where the field disappears:
Screenshot 2022-06-01 at 10 27 38 AM

karrui:
## Problem

Previously, the isClearable prop only controls whether the clear button is shown. However, when one presses the escape key, the item can still be cleared.

This PR fixes that issue and prevents clearing of the item when isClearable is true.

Also sets EditRating dropdown selection to not be clearable.
 
Closes #3932


mantariksh (approved):
Looks Good To Me

PR #3952, opened 2022-06-03
feat: enforce max 10 OTP requests for verified field

mantariksh:
## Problem

There is no limit to the number of OTPs users can request for a given transaction ID. This allows users to continually request for OTPs, which can be costly in the case of SMSes.

## Solution

Enforce a limit of maximum 10 OTPs before user needs to refresh.

**Breaking Changes** 

## Screenshots

Screenshot 2022-06-03 at 3 51 20 PM

## Tests

- [x] Requesting OTP for the 11th time results in the error seen in above screenshot
- [x] Refreshing page clears error, allowing user to verify the field
- [x] Entering a wrong OTP results in `hashRetries` incrementing but not `otpRequests`
- [x] Requesting for a new OTP results in `otpRequests` incrementing and `hashRetries` being reset to 0
- [x] Request an OTP on the old server, then switch over to the new server. Requesting another OTP results in the `otpRequests` key being added to the `field` object in the database, with a value of 1.

### DIFF
--- a/src/app/modules/verification/verification.util.ts
+++ b/src/app/modules/verification/verification.util.ts
@@ -2,6 +2,7 @@ import { StatusCodes } from 'http-status-codes'
 
 import {
   HASH_EXPIRE_AFTER_SECONDS,
+  MAX_OTP_REQUESTS,
   VERIFIED_FIELDTYPES,
   WAIT_FOR_OTP_SECONDS,
   WAIT_FOR_OTP_TOLERANCE_SECONDS,
@@ -14,6 +15,7 @@ import {
 import { smsConfig } from '../../config/features/sms.config'
 import { createLoggerWithLabel } from '../../config/logger'
 import {
+  OtpRequestCountExceededError,
   OtpRequestError,
   SmsLimitExceededError,
 } from '../../modules/verification/verification.errors'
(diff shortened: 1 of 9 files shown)


src/app/modules/verification/verification.util.ts
```diff
@@ -155,6 +167,11 @@ export const mapRouteError: MapRouteError = (
         errorMessage: `You must wait for ${WAIT_FOR_OTP_SECONDS} seconds between each OTP request.`,
         statusCode: StatusCodes.UNPROCESSABLE_ENTITY,
       }
+    case OtpRequestCountExceededError:
+      return {
+        errorMessage: `You have requested for too many OTPs. Please refresh and try again.`,
```
karrui, line 172: reads a little weird haha, maybe
```suggestion
        errorMessage: `You have requested too many OTPs. Please refresh and try again.`,
```

karrui (approved):
lgtm wah anti writing code again how does this feel

hanstirtaputra (approved):
LGTM

src/app/modules/verification/verification.util.ts
mantariksh: updated in 0f5705b4

PR #3976, opened 2022-06-09
feat: wrap csv cell contents in quotes

wanlingt:
## Problem
See [#133](https://github.com/datagovsg/formsg-private/issues/133)

## Solution
Storage mode responses, Form feedback: For the generated CSV files, wrap cell contents in quotes unless they are pure numbers

Data collation tool: Answer fields in the JSON data collation tool section sent in emails are wrapped in quotes, unless they are pure numbers

Pure numbers are left as they are so that excel is able to parse them as numbers

Closes [#133](https://github.com/datagovsg/formsg-private/issues/133)

**Breaking Changes** 

## Tests
- [x] Test that cell contents in downloaded storage mode responses CSV are wrapped in quotes unless they are pure numbers
- [x] Test that comments in downloaded form feedback CSV are wrapped in quotes unless they are pure numbers
- [x] Test that the answer fields for the data collation tool section in emails are wrapped in quotes unless they are pure numbers
- [x] Submit a form using email responses with command specified in [#133](https://github.com/datagovsg/formsg-private/issues/133) in one of the fields. Convert to excel using the data collation tool. Open the excel file with macro enabled, the app should not launch

### DIFF
--- a/src/app/modules/submission/email-submission/email-submission.util.ts
+++ b/src/app/modules/submission/email-submission/email-submission.util.ts
@@ -651,7 +651,7 @@ const getDataCollationFormattedResponse = (
   if (fieldType !== BasicField.Section) {
     return {
       question: getJsonPrefixedQuestion(response),
-      answer,
+      answer: isNaN(Number(answer)) ? `'${answer}'` : answer,
     }
   }
   return undefined
--- a/src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
+++ b/src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
@@ -148,7 +148,7 @@ export class CsvMergedHeadersGenerator extends CsvGenerator {
     const fieldRecord = unprocessedRecord[fieldId]
     if (!fieldRecord) return ''
     const answer = fieldRecord.getAnswer(colIndex)
-    // wrap answer in quotes if answer is not a number. 
+    // wrap answer in quotes if answer is not a number.
     // for number, do not wrap in quotes so that excel parses it as number
     return isNaN(Number(answer)) ? `'${answer}'` : answer
   }
--- a/src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
+++ b/src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
@@ -147,7 +147,9 @@ export class CsvMergedHeadersGenerator extends CsvGenerator {
   ): string {
     const fieldRecord = unprocessedRecord[fieldId]
     if (!fieldRecord) return ''
-    return fieldRecord.getAnswer(colIndex)
+    const answer = fieldRecord.getAnswer(colIndex)
+    // wrap answer in quotes if answer is not a number
+    return isNaN(Number(answer)) ? `'${answer}'` : answer
   }
 
   /**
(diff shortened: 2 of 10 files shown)


src/app/modules/submission/email-submission/email-submission.util.ts
```diff
@@ -651,7 +651,7 @@ const getDataCollationFormattedResponse = (
   if (fieldType !== BasicField.Section) {
     return {
       question: getJsonPrefixedQuestion(response),
-      answer,
+      answer: isNaN(Number(answer)) ? `'${answer}'` : answer,
```
karrui, line 654: may also want to do this processing when decrypting in the client side too.
see `decryption.worker` and the `processDecryptedContent` utility

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
```diff
@@ -147,7 +147,9 @@ export class CsvMergedHeadersGenerator extends CsvGenerator {
   ): string {
     const fieldRecord = unprocessedRecord[fieldId]
     if (!fieldRecord) return ''
-    return fieldRecord.getAnswer(colIndex)
+    const answer = fieldRecord.getAnswer(colIndex)
+    // wrap answer in quotes if answer is not a number
+    return isNaN(Number(answer)) ? `'${answer}'` : answer
```
tshuli, line 152: Could I please check - why can't we wrap in quotes even if it is a number?

tests/end-to-end/helpers/encrypt-mode.js
```diff
@@ -30,33 +30,57 @@ const WEBHOOK_PORT = process.env.MOCK_WEBHOOK_PORT
...
+          }
+        } else {
+          answers.push(`'Others: ${field.val}'`)
+        }
+      }
+      break
     default:
-      answers.push(...getResponseArray(field))
+      if (isNaN(Number(responseArray))) {
+        answers.push(`'${responseArray}'`)
```
tshuli, line 80: likewise - what if we wrap the string containing numbers in quotes?

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
wanlingt: thought I would follow the suggestion in the [issue](https://github.com/datagovsg/formsg-private/issues/133) and leave numbers as they are, so that they can be manipulated in excel if the users wish. Also thought it might be possible that current users run excel macros/formulas on existing forms. it would definitely be more straightforward if we wrapped everything in quotes though! wdyt? :)

src/app/modules/submission/email-submission/email-submission.util.ts
wanlingt: Processing on the client side is done in `CsvMergedHeadersGenerator`

src/public/modules/forms/helpers/CsvMergedHeadersGenerator.ts
tshuli: thanks for highlighting! i think it is fine then. just a minor suggestion for documentation
```suggestion
    const answer = fieldRecord.getAnswer(colIndex)
    // wrap answer in quotes if answer is not a number. 
    // for number, do not wrap in quotes so that excel parses it as number
    return isNaN(Number(answer)) ? `'${answer}'` : answer
```

tshuli (approved):
lgtm

PR #3980, opened 2022-06-09
feat: add SPCP validation before generating OTP

siddarth2824:
## Problem
Currently users can call the OTP generation endpoint via API even for forms protected by SPCP/MyInfo. 

## Solution

Before generating an OTP when the API is called, verify users if the form is protected by SPCP/MyInfo.

**Breaking Changes** 

## Tests

- [x] Test that if a form requires users to authenticate themselves via Singpass, calling the endpoint to generate OTP without successfully verifying Singpass will return `400 Bad Request` error
- [x] Test that if a form requires users to authenticate themselves via Corpass, calling the endpoint to generate OTP without successfully verifying Corpass will return `400 Bad Request` error
- [x] Test that if a form requires users to authenticate themselves via MyInfo, calling the endpoint to generate OTP without successfully verifying MyInfo will return `400 Bad Request` error
- [x] Test that if a form requires users to authenticate themselves via SGID, calling the endpoint to generate OTP without successfully verifying SGID will return `400 Bad Request` error

### DIFF
--- a/src/app/modules/verification/__tests__/verification.controller.spec.ts
+++ b/src/app/modules/verification/__tests__/verification.controller.spec.ts
@@ -87,7 +87,7 @@ describe('Verification controller', () => {
   const MOCK_OTP = 'otp'
   let mockTransaction: IVerificationSchema
   let mockRes: Response
-  const expectedParamsForSendingOtp = {
+  const EXPECTED_PARAMS_FOR_SENDING_OTP = {
     transactionId: MOCK_TRANSACTION_ID,
     fieldId: MOCK_FIELD_ID,
     otp: MOCK_OTP,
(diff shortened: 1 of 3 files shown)


src/app/modules/verification/__tests__/verification.controller.spec.ts
```diff
@@ -60,6 +87,14 @@ describe('Verification controller', () => {
   const MOCK_OTP = 'otp'
   let mockTransaction: IVerificationSchema
   let mockRes: Response
+  const expectedParamsForSendingOtp = {
```
karrui, line 90: constants should be upper snake cased

karrui (approved):
lgtm

Issue #3361, opened 2022-02-03
[React/Builder] Render and edit end page on builder


karrui:
Mocks in "Builder:Thank You" figma page
![Screenshot 2022-02-03 at 11 59 29 AM](https://user-images.githubusercontent.com/22133008/152278990-c09ec746-86db-4b0b-ab8d-faa70a5ef4fa.png)

![End Page From Settings](https://user-images.githubusercontent.com/24553546/172576995-469cffc5-2f46-49aa-809d-10ef0f6a9024.png)

PR #3983, opened 2022-06-10
feat(v2/end-page-builder-1) end page builder from settings

hanstirtaputra:
## Problem
First part of end page builder PR.

This PR adds the functionality to build the end/thank you page through settings

Tracked by #3361 

## Solution

**Features**:

- Add end page builder from settings page

## Tests
Below are tests for the behaviour of the form fields to update the end page:
- [x] Test that `buttonLink` input only accepts URLs starting with `https://`
- [x] Test that  when `buttonLink` value is invalid, we are unable to save the end page settings

Test functionality of the builder:
- [x] Test that changes made to end page in the builder gets reflected in the end page of the form when it is shared

## Before & After Screenshots

Storybook added for settings end page builder

### DIFF
--- a/frontend/src/features/admin-form/common/AdminFormPageService.ts
+++ b/frontend/src/features/admin-form/common/AdminFormPageService.ts
@@ -1,8 +1,8 @@
-import axios from 'axios'
-
 import { EndPageUpdateDto } from '~shared/types'
 
-const ADMIN_FORM_ENDPOINT = '/api/v3/admin/forms'
+import { ApiService } from '~services/ApiService'
+
+const ADMIN_FORM_ENDPOINT = 'admin/forms'
 
 /**
  * Updates the end page for the given form referenced by its id
@@ -15,10 +15,8 @@ export const updateFormEndPage = async (
   formId: string,
   newEndPage: EndPageUpdateDto,
 ): Promise<EndPageUpdateDto> => {
-  return axios
-    .put<EndPageUpdateDto>(
-      `${ADMIN_FORM_ENDPOINT}/${formId}/end-page`,
-      newEndPage,
-    )
-    .then(({ data }) => data)
+  return ApiService.put<EndPageUpdateDto>(
+    `${ADMIN_FORM_ENDPOINT}/${formId}/end-page`,
+    newEndPage,
+  ).then(({ data }) => data)
 }
--- a/frontend/src/features/admin-form/common/AdminFormPageService.ts
+++ b/frontend/src/features/admin-form/common/AdminFormPageService.ts
@@ -1,0 +1,24 @@
+import axios from 'axios'
+
+import { EndPageUpdateDto } from '~shared/types'
+
+const ADMIN_FORM_ENDPOINT = '/api/v3/admin/forms'
+
+/**
+ * Updates the end page for the given form referenced by its id
+ *
+ * @param formId the id of the form to update end page for
+ * @param newEndPage the new endpage to replace with
+ * @returns the updated end page on success
+ */
+export const updateFormEndPage = async (
+  formId: string,
+  newEndPage: EndPageUpdateDto,
+): Promise<EndPageUpdateDto> => {
+  return axios
+    .put<EndPageUpdateDto>(
+      `${ADMIN_FORM_ENDPOINT}/${formId}/end-page`,
+      newEndPage,
+    )
+    .then(({ data }) => data)
+}
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
@@ -7,7 +7,11 @@ import {
   patchAdminFormSettings,
 } from '~/mocks/msw/handlers/admin-form'
 
-import { StoryRouter, viewports } from '~utils/storybook'
+import {
+  getMobileViewParameters,
+  getTabletViewParameters,
+  StoryRouter,
+} from '~utils/storybook'
 
 import { SettingsEndPage } from './SettingsEndPage'
 
@@ -38,17 +42,7 @@ const Template: Story = () => <SettingsEndPage />
 export const Desktop = Template.bind({})
 
 export const Tablet = Template.bind({})
-Tablet.parameters = {
-  viewport: {
-    defaultViewport: 'tablet',
-  },
-  chromatic: { viewports: [viewports.md] },
-}
+Tablet.parameters = getTabletViewParameters()
 
 export const Mobile = Template.bind({})
-Mobile.parameters = {
-  viewport: {
-    defaultViewport: 'mobile1',
-  },
-  chromatic: { viewports: [viewports.xs] },
-}
+Mobile.parameters = getMobileViewParameters()
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
@@ -1,0 +1,55 @@
+import { Meta, Story } from '@storybook/react'
+
+import { FormSettings } from '~shared/types'
+
+import {
+  getAdminFormSettings,
+  patchAdminFormSettings,
+} from '~/mocks/msw/handlers/admin-form'
+
+import { StoryRouter, viewports } from '~utils/storybook'
+
+import { SettingsEndPage } from './SettingsEndPage'
+
+const buildMswRoutes = ({
+  overrides,
+  delay,
+}: {
+  overrides?: Partial<FormSettings>
+  delay?: number | 'infinite'
+} = {}) => [
+  getAdminFormSettings({ overrides, delay }),
+  patchAdminFormSettings({ overrides }),
+]
+
+export default {
+  title: 'Pages/AdminFormPage/Settings/EndPage',
+  component: SettingsEndPage,
+  decorators: [StoryRouter({ initialEntries: ['/12345'], path: '/:formId' })],
+  parameters: {
+    // Required so skeleton "animation" does not hide content.
+    chromatic: { pauseAnimationAtEnd: true },
+    layout: 'fullscreen',
+    msw: buildMswRoutes(),
+  },
+} as Meta
+
+const Template: Story = () => <SettingsEndPage />
+
+export const Desktop = Template.bind({})
+
+export const Tablet = Template.bind({})
+Tablet.parameters = {
+  viewport: {
+    defaultViewport: 'tablet',
+  },
+  chromatic: { viewports: [viewports.md] },
+}
+
+export const Mobile = Template.bind({})
+Mobile.parameters = {
+  viewport: {
+    defaultViewport: 'mobile1',
+  },
+  chromatic: { viewports: [viewports.xs] },
+}
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
@@ -29,7 +29,6 @@ export default {
   parameters: {
     // Required so skeleton "animation" does not hide content.
     chromatic: { pauseAnimationAtEnd: true },
-    layout: 'fullscreen',
     msw: buildMswRoutes(),
   },
 } as Meta
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
@@ -1,6 +1,6 @@
 import { Meta, Story } from '@storybook/react'
 
-import { getAdminForm } from '~/mocks/msw/handlers/admin-form'
+import { createFormBuilderMocks } from '~/mocks/msw/handlers/admin-form'
 
 import {
   getMobileViewParameters,
@@ -14,7 +14,7 @@ const buildMswRoutes = ({
   delay,
 }: {
   delay?: number | 'infinite'
-} = {}) => [getAdminForm({}, delay)]
+} = {}) => [...createFormBuilderMocks({}, delay)]
 
 export default {
   title: 'Pages/AdminFormPage/Settings/EndPage',
@@ -23,7 +23,7 @@ export default {
   parameters: {
     // Required so skeleton "animation" does not hide content.
     chromatic: { pauseAnimationAtEnd: true },
-    msw: buildMswRoutes(),
+    msw: buildMswRoutes({ delay: 0 }),
   },
 } as Meta
 
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
@@ -1,11 +1,6 @@
 import { Meta, Story } from '@storybook/react'
 
-import { FormSettings } from '~shared/types'
-
-import {
-  getAdminFormSettings,
-  patchAdminFormSettings,
-} from '~/mocks/msw/handlers/admin-form'
+import { getAdminForm } from '~/mocks/msw/handlers/admin-form'
 
 import {
   getMobileViewParameters,
@@ -16,15 +11,10 @@ import {
 import { SettingsEndPage } from './SettingsEndPage'
 
 const buildMswRoutes = ({
-  overrides,
   delay,
 }: {
-  overrides?: Partial<FormSettings>
   delay?: number | 'infinite'
-} = {}) => [
-  getAdminFormSettings({ overrides, delay }),
-  patchAdminFormSettings({ overrides }),
-]
+} = {}) => [getAdminForm({}, delay)]
 
 export default {
   title: 'Pages/AdminFormPage/Settings/EndPage',
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.tsx
@@ -9,9 +9,7 @@ export const SettingsEndPage = (): JSX.Element => {
 
   return (
     <>
-      <CategoryHeader mb={0} mr="2rem">
-        Customise Thank You page
-      </CategoryHeader>
+      <CategoryHeader mb={0}>Customise Thank You page</CategoryHeader>
       <Skeleton isLoaded={!isLoading && !!settings}>
         {settings ? <EndPageSettingsInput settings={settings.endPage} /> : null}
       </Skeleton>
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.tsx
@@ -1,19 +1,15 @@
-import { Skeleton } from '@chakra-ui/react'
-
 import { useAdminForm } from '../common/queries'
 
 import { CategoryHeader } from './components/CategoryHeader'
 import { EndPageSettingsInput } from './components/EndPageSettingsSection/EndPageSettingsInput'
 
 export const SettingsEndPage = (): JSX.Element => {
-  const { data: form, isLoading } = useAdminForm()
+  const { data: form } = useAdminForm()
 
   return (
     <>
       <CategoryHeader mb={0}>Customise Thank You page</CategoryHeader>
-      <Skeleton isLoaded={!isLoading && !!form}>
-        {form && <EndPageSettingsInput endPage={form.endPage} />}
-      </Skeleton>
+      {form && <EndPageSettingsInput endPage={form.endPage} />}
     </>
   )
 }
--- a/frontend/src/features/admin-form/settings/SettingsEndPage.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsEndPage.tsx
@@ -1,0 +1,101 @@
+import { useForm } from 'react-hook-form'
+import { Flex } from '@chakra-ui/layout'
+import { FormControl, Skeleton, Stack } from '@chakra-ui/react'
+
+import FormErrorMessage from '~components/FormControl/FormErrorMessage'
+import FormLabel from '~components/FormControl/FormLabel'
+import Input from '~components/Input'
+import Textarea from '~components/Textarea'
+
+import { useAdminForm } from '../common/queries'
+
+import { CategoryHeader } from './components/CategoryHeader'
+import { useMutateFormSettings } from './mutations'
+
+type SettingsInput = {
+  title: string
+  paragraph: string
+  buttonText: string
+  buttonLink: string
+}
+
+export const SettingsEndPage = (): JSX.Element => {
+  const { data: form, isLoading } = useAdminForm()
+  const { mutateFormEndPage } = useMutateFormSettings()
+
+  const {
+    register,
+    formState: { errors },
+    handleSubmit,
+  } = useForm<SettingsInput>({
+    mode: 'onBlur',
+    defaultValues: {
+      title: form?.endPage.title ?? 'Thank you for filling out the form.',
+      paragraph: form?.endPage.paragraph ?? '',
+      buttonText: form?.endPage.buttonText ?? 'Submit another response',
+      buttonLink: form?.endPage.buttonLink ?? '',
+    },
+  })
+
+  const handleUpdateEndPage = handleSubmit((endPage) =>
+    mutateFormEndPage.mutate(endPage),
+  )
+
+  return (
+    <>
+      <CategoryHeader mb={0} mr="2rem">
+        Customise Thank You page
+      </CategoryHeader>
+
+      <Stack gap="2rem" paddingTop="2.5rem">
+        <FormControl isInvalid={false}>
+          <FormLabel isRequired>Title</FormLabel>
+          <Skeleton isLoaded={!isLoading}>
+            <Input
+              // TODO (hans): Update Validation rules
+              {...register('title')}
+              onBlur={handleUpdateEndPage}
+            />
+          </Skeleton>
+          <FormErrorMessage>{errors.title?.message}</FormErrorMessage>
+        </FormControl>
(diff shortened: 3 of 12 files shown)


src/app/modules/form/admin-form/admin-form.middlewares.ts
```diff
@@ -26,5 +26,11 @@ export const updateSettingsValidator = celebrate({
       url: Joi.string().uri().allow(''),
       isRetryEnabled: Joi.boolean(),
     }).min(1),
+    endPage: Joi.object({
+      title: Joi.string(),
+      paragraph: Joi.string().allow(''),
+      buttonLink: Joi.string().uri().allow(''),
+      buttonText: Joi.string(),
+    }).min(1),
```
mantariksh, line 34: hmm. so I think our internal mental model should still be that the endPage is separate from the rest of the settings. it's going to get very messy if the end page is editable both via its own endpoint and via the settings update endpoint.

we want the end page to be editable on both the settings page and the build tab but this is purely a UX enhancement. the way we think about the data model shouldn't change.

shared/constants/form.ts
```diff
@@ -29,6 +29,7 @@ const FORM_SETTINGS_FIELDS = <const>[
   'submissionLimit',
   'title',
   'webhook',
+  'endPage',
```
mantariksh, line 32: same here, don't think we should think about the endPage as part of the settings

frontend/src/features/admin-form/settings/SettingsService.ts
```diff
@@ -104,6 +104,13 @@ export const updateFormWebhookRetries = async (
   })
 }
 
+export const updateFormEndPage: UpdateFormFn<'endPage'> = async (
```
mantariksh, line 107: let's move this out of the `settings` folder because it's going to be shared across both the Create and Settings tabs. maybe `admin-form/common`?

frontend/src/features/admin-form/settings/mutations.ts
```diff
@@ -278,6 +280,19 @@ export const useMutateFormSettings = () => {
     },
   )
 
+  const mutateFormEndPage = useMutation(
```
mantariksh, line 283: similarly let's move this out of the `settings` folder as it's going to be shared across Create and Settings

frontend/src/features/admin-form/settings/SettingsEndPage.tsx
```diff
@@ -0,0 +1,18 @@
+import { Skeleton } from '@chakra-ui/react'
+
+import { CategoryHeader } from './components/CategoryHeader'
+import { EndPageSettingsInput } from './components/EndPageSettingsSection/EndPageSettingsInput'
+import { useAdminFormSettings } from './queries'
+
+export const SettingsEndPage = (): JSX.Element => {
+  const { data: settings, isLoading } = useAdminFormSettings()
```
mantariksh, line 8: as mentioned we shouldn't think of endPage as part of settings, so maybe just use `useAdminForm`?

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+import FormErrorMessage from '~components/FormControl/FormErrorMessage'
+import FormLabel from '~components/FormControl/FormLabel'
+import Input from '~components/Input'
+import Textarea from '~components/Textarea'
+
+import { useMutateFormSettings } from '../../mutations'
+
+interface EndPageSettingsInputProps {
+  settings: {
```
mantariksh, line 14: why nest the info under a settings key?

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+    title: string
+    paragraph?: string
+    buttonLink?: string
+    buttonText: string
```
mantariksh, line 18: can we reuse `FormEndPage` from `~shared`?

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+        endPage.title === title &&
+        endPage.buttonText === buttonText &&
+        endPage.buttonLink === defaultButtonLink &&
+        endPage.paragraph === defaultParagraph
+      ) {
+        return
+      }
+
+      return mutateFormEndPage.mutate(endPage)
+    })()
```
mantariksh, line 58: why is this invoked immediately?

mantariksh (commented):
some prelim comments!

frontend/src/features/admin-form/settings/SettingsService.ts
hanstirtaputra: Okay, created a separate `AdminFormPageService` for this

frontend/src/features/admin-form/settings/SettingsEndPage.tsx
hanstirtaputra: Sounds good. Updated the PR to use `useAdminForm`

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: Brain fart... I updated the PR directly include the info instead of nesting it. Thanks!

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+  })
+
+  const handleUpdateEndPage = useCallback(() => {
+    return handleSubmit((endPage) => {
+      if (
+        endPage.title === title &&
+        endPage.buttonText === buttonText &&
+        endPage.buttonLink === defaultButtonLink &&
+        endPage.paragraph === defaultParagraph
+      ) {
```
karrui, line 48: these conditional checks looks a little off. the benefit of `react-hook-form` is that once `handleSubmit` is called, we know that validation has passed. Only valid inputs should be passed in (and thus has no need for these checks).

but i see that there are no validation on the fields yet, will flag in later comments.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
karrui: no need to invoke mutation immediately. can look at other examples in the code base as to how this is normally handled.

most of the time you return the mutation function and keep track of state using the mutation function's exported state (i.e. `mutateFormEndPage.isLoading`)

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+        require_protocol: true,
+      }) ||
+      'Please enter a valid URL (starting with https://)',
+  })
+
+  return (
+    <Stack gap="2rem" paddingTop="2.5rem">
+      <FormControl isInvalid={!!errors.title}>
+        <FormLabel isRequired>Title</FormLabel>
+        <Input {...register('title')} onBlur={handleUpdateEndPage} />
```
karrui, line 86: we should add a required validation on this registered field.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <FormErrorMessage>{errors.title?.message}</FormErrorMessage>
+      </FormControl>
+      <FormControl isInvalid={!!errors.paragraph}>
+        <FormLabel isRequired>Follow-up paragraph</FormLabel>
+        <Textarea {...register('paragraph')} onBlur={handleUpdateEndPage} />
```
karrui, line 91: we should add a required validation on this registered field.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Stack gap="2rem" paddingTop="2.5rem">
+      <FormControl isInvalid={!!errors.title}>
+        <FormLabel isRequired>Title</FormLabel>
+        <Input {...register('title')} onBlur={handleUpdateEndPage} />
+        <FormErrorMessage>{errors.title?.message}</FormErrorMessage>
+      </FormControl>
+      <FormControl isInvalid={!!errors.paragraph}>
+        <FormLabel isRequired>Follow-up paragraph</FormLabel>
```
karrui, line 90: seems to be a typo? design seems to be `Follow-up instructions` instead

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const buttonLinkRegister = register('buttonLink', {
+    onBlur: handleUpdateButtonLink,
+    validate: (url) =>
+      !url ||
+      validator.isURL(url, {
+        protocols: ['https'],
+        require_protocol: true,
+      }) ||
+      'Please enter a valid URL (starting with https://)',
+  })
```
karrui, line 80: don't think we need to virtualize the registration of the field here, we could memoize the validation rules instead.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+        gap={{ base: '2rem', md: '1rem' }}
+      >
+        <FormControl isInvalid={!!errors.buttonText}>
+          <FormLabel isRequired>Button text</FormLabel>
+          <Input {...register('buttonText')} onBlur={handleUpdateEndPage} />
+          <FormErrorMessage>{errors.buttonText?.message}</FormErrorMessage>
+        </FormControl>
+        <FormControl isInvalid={!!errors.buttonLink}>
+          <FormLabel isRequired>Button redirect link</FormLabel>
+          <Input placeholder="Default form link" {...buttonLinkRegister} />
```
karrui, line 105: we should invoke `register` call in the render function instead of virtualizing it.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
+import { useCallback, useMemo } from 'react'
```
karrui, line 1: suggest reclarifying with designers on whether we should save values on blur for this screen, i think design may not be updated.

i think we moved on to a submit button for saving instead of autosaving for group inputs.

if we have a submit button on this component, we can remove all the onBlur calls (which looks dangerous to me as they could hold old state for other parts of the endpage inputs)

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+export const EndPageSettingsInput = ({
+  title,
+  paragraph,
+  buttonLink,
+  buttonText,
+}: FormEndPage): JSX.Element => {
+  const { mutateFormEndPage } = useMutateFormPage()
+
+  const defaultParagraph = useMemo(() => paragraph ?? '', [paragraph])
+  const defaultButtonLink = useMemo(() => buttonLink ?? '', [buttonLink])
```
karrui, line 24: i don't think this is needed.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+    title,
+  ])
+
+  const handleUpdateButtonLink = useCallback(() => {
+    if (errors.buttonLink) {
+      return resetField('buttonLink')
+    }
+
+    return handleUpdateEndPage()
+  }, [errors.buttonLink, handleUpdateEndPage, resetField])
```
karrui, line 69: suggest using a submit button so we can remove this

karrui (changes requested):

frontend/src/features/admin-form/settings/SettingsEndPage.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { EndPageSettingsInput } from './components/EndPageSettingsSection/EndPageSettingsInput'
+
+export const SettingsEndPage = (): JSX.Element => {
+  const { data: form, isLoading } = useAdminForm()
+
+  return (
+    <>
+      <CategoryHeader mb={0}>Customise Thank You page</CategoryHeader>
+      <Skeleton isLoaded={!isLoading && !!form}>
+        {form ? <EndPageSettingsInput {...form.endPage} /> : null}
```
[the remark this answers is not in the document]
hanstirtaputra: Ahh yes, skeleton won't appear if form isn't loaded. I didn't want to use `useAdminForm` in `EndPageSettingsInput` and inline skeleton states because I didn't want to deal with `undefined` `form` in `EndPageSettingsInput`, thought it would be simpler to deal with the loading in a parent component and assume that `endPage` is defined in `EndPageSettingsInput`

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: Oh yes, thanks for pointing it out! Will update

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: Okay, checking with the designers to use a submit button sintead

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: This check here prevents the case of no changes in any `endPage` data when a user clicks between the different inputs still causing a `PUT` request to still be sent. It's also because of the `onBlur` updates. Will clarify if it can be changed to a submit button

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+import Input from '~components/Input'
+import Textarea from '~components/Textarea'
+
+import { useMutateFormPage } from '~features/admin-form/common/mutations'
+
+export const EndPageSettingsInput = ({
+  title,
+  paragraph,
+  buttonLink,
+  buttonText,
```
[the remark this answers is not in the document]
hanstirtaputra: Okay thanks for the suggestion!

frontend/src/features/admin-form/common/AdminFormPageService.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param formId the id of the form to update end page for
+ * @param newEndPage the new endpage to replace with
+ * @returns the updated end page on success
+ */
+export const updateFormEndPage = async (
+  formId: string,
+  newEndPage: EndPageUpdateDto,
+): Promise<EndPageUpdateDto> => {
+  return axios
```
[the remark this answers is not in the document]
hanstirtaputra: Okay updated to use `ApiService`! I used `axios` and `/api/v3/...` for the base url before that

frontend/src/features/admin-form/settings/SettingsEndPage.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const SettingsEndPage = (): JSX.Element => {
+  const { data: form, isLoading } = useAdminForm()
+
+  return (
+    <>
+      <CategoryHeader mb={0}>Customise Thank You page</CategoryHeader>
+      <Skeleton isLoaded={!isLoading && !!form}>
+        {form && <EndPageSettingsInput endPage={form.endPage} />}
+      </Skeleton>
```
karrui: skeleton still doesn't show anything, might not even want to wrap if skeleton if not actually using a skeleton

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const buttonLinkValidation = useCallback(
+    (url) =>
+      !url ||
+      validator.isURL(url, {
+        protocols: ['https'],
+        require_protocol: true,
+      }) ||
+      'Please enter a valid URL (starting with https://)',
+    [],
+  )
```
karrui: the whole validation object can actually be memoized; don't have to memoize just the function haha:

```tsx
const buttonLinkRules = useMemo(() => {
  return { 
    validate: (url) =>
      !url ||
      validator.isURL(url, {
        protocols: ['https'],
        require_protocol: true,
      }) ||
      'Please enter a valid URL (starting with https://)'
  }
}, [])
```


in fact, we can even instantiate it outside of this component as a constant.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const {
+    register,
+    formState: { errors },
+    handleSubmit,
+  } = useForm<FormEndPage>({
+    mode: 'onChange',
```
karrui: i think this can now be `'onBlur'` since we have a submit button now

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <FormLabel isRequired>Title</FormLabel>
+        <Input
+          {...register('title', {
+            required: 'Title is required for the Thank You page',
+          })}
+        />
+        <FormErrorMessage>{errors.title?.message}</FormErrorMessage>
+      </FormControl>
+      <FormControl isInvalid={!!errors.paragraph}>
+        <FormLabel isRequired>Follow-up instructions</FormLabel>
```
karrui, line 56: looking at the current behavior, follow up instructions are not required. seems like only the title is required!

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+    mutateFormEndPage.mutate(endPage),
+  )
+
+  return (
+    <Stack gap="2rem" paddingTop="2.5rem">
+      <FormControl isInvalid={!!errors.title}>
+        <FormLabel isRequired>Title</FormLabel>
+        <Input
+          {...register('title', {
+            required: 'Title is required for the Thank You page',
```
karrui: nit: i think we can just use the `REQUIRED_ERROR` constant instead of a special one

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+        />
+        <FormErrorMessage>{errors.paragraph?.message}</FormErrorMessage>
+      </FormControl>
+      <Stack
+        direction={{ base: 'column', md: 'row' }}
+        gap={{ base: '2rem', md: '1rem' }}
+      >
+        <FormControl isInvalid={!!errors.buttonText}>
+          <FormLabel isRequired>Button text</FormLabel>
+          <Input {...register('buttonText')} />
```
karrui: might need a placeholder here too, the current one seems to be `'Submit another form'`

frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const Template: Story = () => <SettingsEndPage />
+
+export const Desktop = Template.bind({})
+
+export const Tablet = Template.bind({})
+Tablet.parameters = getTabletViewParameters()
+
+export const Mobile = Template.bind({})
+Mobile.parameters = getMobileViewParameters()
```
karrui, line 38: these stories don't seem to work

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: Alright thanks! Can't we only use `useMemo` within a React component?

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: I think "Submit another form" is a default value instead of a placeholder though. Currently it already defaults to "Submit another form" for the `buttonText`

karrui: the db schema doesn't actually require a value for `buttonText`, and instead the client defaults to "Submit another form" if the value does not exist.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
karrui: ya, if we instantiate outside the component don't need use `useMemo` la haha

frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
hanstirtaputra: Updated the PR to fix this, thanks!

frontend/src/features/admin-form/settings/SettingsEndPage.tsx
hanstirtaputra: Okay, I'll remove the Skeleton for this then!

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: Okay thanks hahaha I'll instantiate outside

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
hanstirtaputra: Okay then I'll add the placeholder

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -42,6 +31,18 @@ export const EndPageSettingsInput = ({
     defaultValues: endPage,
   })
 
+  const buttonLinkRules = useMemo(() => {
```
karrui: nit

```suggestion
 const buttonLinkRules: RegisterOptions<FormEndPage, 'buttonLink'> = useMemo(() => {
```

frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
```diff
@@ -16,15 +11,10 @@ import {
...
-  overrides,
   delay,
 }: {
-  overrides?: Partial<FormSettings>
   delay?: number | 'infinite'
-} = {}) => [
-  getAdminFormSettings({ overrides, delay }),
-  patchAdminFormSettings({ overrides }),
-]
+} = {}) => [getAdminForm({}, delay)]
```
karrui: should we just  use `createFormBuilderMocks` instead? it already contains a handler for this new function you created.

frontend/src/features/admin-form/settings/components/EndPageSettingsSection/EndPageSettingsInput.tsx
```diff
@@ -14,6 +13,16 @@ import Textarea from '~components/Textarea'
 
 import { useMutateFormPage } from '~features/admin-form/common/mutations'
 
+const buttonLinkRules = {
```
karrui: nit: 

```suggestion
const buttonLinkRules: RegisterOptions<FormEndPage, 'buttonLink'> = {
```

karrui (approved):
lgtm otherwise, please get designers to review!

frontend/src/features/admin-form/settings/SettingsEndPage.stories.tsx
hanstirtaputra: Okay, updated PR to use `createFormBuilderMocks`. I removed the `getAdminForm` implementation as well

PR #3990, opened 2022-06-10
feat: add last seen feature update date attribute to user collection

### LINKED ISSUE #3985: [React/What's New] Add flag to User Document to encapsulate last seen feature update date attribute
Form admins will be able to see a notification on the “What’s New” tab on the navigation bar whenever there is a new feature update. For this to happen, some form of information should be stored either on the client or on the backend, to know whether the form admin has seen the latest feature update in the list of feature updates.

siddarth2824:
## Problem

This PR adds attribute to user collection to store user's last seen feature update date which will be used in future PRs to decide whether or not to notify users on the latest feature updates.

Closes #3985 

## Solution

**Breaking Changes** 

## Tests

- [x] Test to ensure that user with last seen feature update date attribute is returned appropriately when added to collection

### DIFF
--- a/src/app/models/user.server.model.ts
+++ b/src/app/models/user.server.model.ts
@@ -71,6 +71,9 @@ const compileUserModel = (db: Mongoose) => {
       betaFlags: {
         sgid: Boolean,
       },
+      flags: {
+        lastSeenFeatureUpdateDate: String,
+      },
     },
     {
       timestamps: {
--- a/src/app/models/user.server.model.ts
+++ b/src/app/models/user.server.model.ts
@@ -72,7 +72,7 @@ const compileUserModel = (db: Mongoose) => {
         sgid: Boolean,
       },
       flags: {
-        lastSeenFeatureUpdateDate: String,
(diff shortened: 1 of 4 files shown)


src/app/models/user.server.model.ts
```diff
@@ -71,6 +71,9 @@ const compileUserModel = (db: Mongoose) => {
       betaFlags: {
         sgid: Boolean,
       },
+      flags: {
+        lastSeenFeatureUpdateDate: String,
```
karrui, line 75: should this be a date instead? or even number (for timestamp?) string type may be lossy

siddarth2824: Initially I thought it would be easier to store it as string since the date stored in the JSON file would be as a string as well. But now that I think about it, if we need to do comparison of dates, we would need to use an external date library like `moment` to handle the comparison. Since the date type in Mongo is represented as UTC datetime we can handle the standardisation of date with `moment` as well. I will update it to date type instead.

karrui (approved):
lgtm

PR #3991, opened 2022-06-13
feat: add end-page uri scheme validation

wanlingt:
## Problem
No checks are done on the URI scheme in the end page's Thank You message button link

Closes [#134](https://github.com/datagovsg/formsg-private/issues/134)

## Solution
Using [Joi's inbuilt settings](https://github.com/sideway/joi/blob/master/API.md#stringurioptions), only allow URIs with the schemes `http` or `https`

**Breaking Changes** 

## Tests
- [x] Change the end-page's thank you button link to a scheme that's not http or https. Submit the HTTP request. You should get a 400 status error and an error message which states "Validation failed"
Screenshot 2022-07-04 at 10 11 01 AM

- [x] Change the end-page's thank you button link to a scheme that's http or https. Submit the HTTP request. Request should succeed.

### DIFF
--- a/src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
+++ b/src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
@@ -2134,7 +2134,55 @@ describe('admin-form.form.routes', () => {
       buttonText: 'end page button',
     }
 
-    it('should return 400 when button link does not use valid URI scheme', async () => {
+    it('should return 200 when button link is updated with a valid HTTPS URI scheme', async () => {
+      //Arrange
+      const form = await EmailFormModel.create({
+        emails: [defaultUser.email],
+        title: 'email me',
+        admin: defaultUser._id,
+        endpage: MOCK_END_PAGE,
+      })
+
+      //Act
+      const validUriScheme = 'https://valid.scheme'
+      const response = await request
+        .put(`/admin/forms/${form._id}/end-page`)
+        .send({
+          buttonLink: validUriScheme,
+        })
+
+      //Assert
+      expect(response.status).toBe(200)
+      expect(response.body).toEqual(
+        expect.objectContaining({ buttonLink: validUriScheme }),
+      )
+    })
+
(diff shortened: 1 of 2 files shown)


tshuli (commented):
implementation looks good! could we maybe add in a couple of unit tests as well?

mantariksh (commented):
have we checked if there are any existing non-http/https links in the DB?

also what do you think about adding model-level validation as well?

wanlingt:
> have we checked if there are any existing non-http/https links in the DB?

The non-http/https links are all formHashes (I understand these were the default for old forms and have been replaced by empty strings?), except for 1 which was entered by me when I was testing the issue.

> also what do you think about adding model-level validation as well?

I think this would make sense since we're sure we only want to allow http/https links, will add it in. This means we would have to change the formHashes to empty strings in the DB?

wanlingt:
> have we checked if there are any existing non-http/https links in the DB?
> also what do you think about adding model-level validation as well?

As discussed, we'll go without model-level validation given that some old forms in the DB have links which are not in http/https format

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
```diff
@@ -2126,4 +2127,39 @@ describe('admin-form.form.routes', () => {
...
     })
   })
+
+  describe('PUT /admin/forms/:formId/end-page', () => {
+    const MOCK_END_PAGE: Partial<FormEndPage> = {
+      title: 'end page title',
+      buttonText: 'end page button',
+    }
+
+    it('should return 400 when button link does not use valid URI scheme', async () => {
```
karrui, line 2137: can we have a positive test for this route too?

karrui (approved):
lgtm

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
wanlingt: added!

PR #3997, opened 2022-06-16
feat: add new user endpoint to update user last seen feature update date attribute

### LINKED ISSUE #3986: [React/What's New] Add endpoint to update user's last seen feature update date information
Currently there is no exposed endpoint to update user's information in User's collection. A new endpoint will have to be created to update user's last seen feature update date information.

siddarth2824:
## Problem
Currently there is no endpoint to update user's last seen feature update date attribute which will be used to handle logic on whether or not to notify form admin users when there is a new feature update.

Closes #3986 

## Solution

**Breaking Changes** 

## Tests

- [x] Should return 200 if update is done successfully
- [x] Should return 401 if form admin user is not logged in 
- [x] Should return 422 if form admin user id does not exist in the User Collection
- [x] Should return 500 if there is a database error

### DIFF
--- a/src/app/modules/user/__tests__/user.controller.spec.ts
+++ b/src/app/modules/user/__tests__/user.controller.spec.ts
@@ -16,6 +16,7 @@ import { IPopulatedUser } from 'src/types'
 import expressHandler from 'tests/unit/backend/helpers/jest-express'
 
 import { DatabaseError } from '../../core/core.errors'
+import { UNAUTHORIZED_USER_MESSAGE } from '../user.constant'
 
 jest.mock('src/app/modules/user/user.service')
 jest.mock('src/app/services/sms/sms.factory')
@@ -89,7 +90,7 @@ describe('user.controller', () => {
       // Assert
       // Should trigger unauthorized response.
       expect(mockRes.status).toBeCalledWith(401)
-      expect(mockRes.json).toBeCalledWith('User is unauthorized.')
+      expect(mockRes.json).toBeCalledWith(UNAUTHORIZED_USER_MESSAGE)
       // Service functions should not be called.
       expect(MockUserService.verifyContactOtp).not.toHaveBeenCalled()
       expect(MockUserService.updateUserContact).not.toHaveBeenCalled()
@@ -120,7 +121,7 @@ describe('user.controller', () => {
       // Assert
       // Should trigger unauthorized response.
       expect(mockRes.status).toBeCalledWith(401)
-      expect(mockRes.json).toBeCalledWith('User is unauthorized.')
+      expect(mockRes.json).toBeCalledWith(UNAUTHORIZED_USER_MESSAGE)
       // Service functions should not be called.
       expect(MockUserService.verifyContactOtp).not.toHaveBeenCalled()
       expect(MockUserService.updateUserContact).not.toHaveBeenCalled()
@@ -240,7 +241,7 @@ describe('user.controller', () => {
       // Assert
       // Should trigger unauthorized response.
       expect(mockRes.status).toBeCalledWith(401)
-      expect(mockRes.json).toBeCalledWith('User is unauthorized.')
+      expect(mockRes.json).toBeCalledWith(UNAUTHORIZED_USER_MESSAGE)
       // Service functions should not be called.
       expect(MockUserService.verifyContactOtp).not.toHaveBeenCalled()
       expect(MockUserService.updateUserContact).not.toHaveBeenCalled()
@@ -272,7 +273,7 @@ describe('user.controller', () => {
       // Assert
       // Should trigger unauthorized response.
       expect(mockRes.status).toBeCalledWith(401)
-      expect(mockRes.json).toBeCalledWith('User is unauthorized.')
+      expect(mockRes.json).toBeCalledWith(UNAUTHORIZED_USER_MESSAGE)
       // Service functions should not be called.
       expect(MockUserService.verifyContactOtp).not.toHaveBeenCalled()
       expect(MockUserService.updateUserContact).not.toHaveBeenCalled()
@@ -409,7 +410,7 @@ describe('user.controller', () => {
       // Should trigger unauthorized response.
       expect(mockRes.status).toBeCalledWith(401)
       expect(mockRes.json).toBeCalledWith({
-        message: 'User is unauthorized.',
+        message: UNAUTHORIZED_USER_MESSAGE,
       })
     })
 
@@ -439,12 +440,6 @@ describe('user.controller', () => {
         },
       },
     })
-    // Mock implementaion of Date.now() due to slight difference in Date.now() execution in test and when
-    // Date.now() in updateUserLastSeenFeatureUpdateDate function is executed
-    jest
-      .spyOn(global.Date, 'now')
-      .mockReturnValue(new Date('2022-07-14T11:01:58.135Z').valueOf())
-    const MOCK_DATE = new Date('2022-07-14T11:01:58.135Z')
 
     it('should return 200 when successful', async () => {
       // Arrange
@@ -471,7 +466,7 @@ describe('user.controller', () => {
       // Expect services to be called with correct arguments.
       expect(
         MockUserService.updateUserLastSeenFeatureUpdateDate,
-      ).toBeCalledWith(MOCK_REQ.session.user?._id, MOCK_DATE)
+      ).toBeCalledWith(MOCK_REQ.session.user?._id)
       expect(mockRes.status).toBeCalledWith(200)
       expect(mockRes.json).toBeCalledWith(mockPopulatedUser)
     })
@@ -479,7 +474,6 @@ describe('user.controller', () => {
     it('should return 401 if session does not contain user id', async () => {
       // Arrange
       const MOCK_REQ_WITH_NO_USER_ID_IN_SESSION = expressHandler.mockRequest({
-        body: { date: MOCK_DATE },
         session: {},
       })
       const mockRes = expressHandler.mockResponse()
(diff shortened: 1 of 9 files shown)


src/app/modules/user/user.controller.ts
```diff
@@ -206,3 +214,51 @@ export const handleFetchUser: ControllerHandler = async (req, res) => {
       return res.status(statusCode).json({ message: errorMessage })
     })
 }
+
+export const _handleUpdateUserLastSeenFeatureUpdateDate: ControllerHandler<
+  unknown,
+  string,
+  UpdateUserLastFeatureUpdateDateDto
+> = async (req, res) => {
+  const { latestLastSeenFeatureUpdateDate } = req.body
```
karrui, line 223: nit: seems unnecessarily verbose; suggest just `date`, since the endpoint is already indicative of what is being updated

src/app/modules/user/user.middleware.ts
```diff
@@ -22,3 +25,15 @@ export const validateContactOtpVerificationParams = celebrate({
...
 })
+
+/**
+ * Celebrate validation for the update user's last seen feature update date endpoint.
+ */
+export const validateUpdateUserLastSeenFeatureUpdateParams = celebrate({
+  [Segments.BODY]: Joi.object({
+    latestLastSeenFeatureUpdateDate: Joi.date()
+      .format('YYYY-MM-DD')
+      .raw()
```
karrui, line 36: have you checked the format of the date created here? is it utc or some local time? might result in inaccurate comparisons. do you think we should use `Date.now()` timestamp instead?

src/app/modules/user/user.controller.ts
```diff
@@ -206,3 +214,51 @@ export const handleFetchUser: ControllerHandler = async (req, res) => {
...
+  }
+
+  return updateUserLastSeenFeatureUpdateDate(
+    sessionUserId,
+    latestLastSeenFeatureUpdateDate,
+  )
+    .map(() => {
+      return res
+        .status(StatusCodes.OK)
+        .json(UPDATE_LAST_SEEN_FEATURE_DATE.SUCCESS)
```
karrui, line 237: do you think we should return the updated user back to the client instead? So the user can be updated on the client using data from the server.

src/app/routes/api/v3/user/user.routes.ts
```diff
@@ -42,4 +43,19 @@ UserRouter.post('/contact/otp/generate', UserController.handleContactSendOtp)
...
+ * @param body.latestLastSeenFeatureUpdateDate the latest seen feature update date to update user with
+ * @returns 200 when user last seen feature update updates sucessfully
+ * @returns 401 if user is not currently logged in
+ * @returns 422 when userId does not exist in the database
+ * @returns 500 when database errors occurs
+ */
+UserRouter.post(
+  '/whats-new/last-seen',
+  UserController.handleUpdateUserLastSeenFeatureUpdateDate,
+)
```
karrui, line 59: wondering if we should rename this route to `/flags/last-seen-feature` instead? since it's not really operating on the any `what's new` collection anymore. We can even add a UserFlagsRouter subrouter to contain it if we want to.

src/app/modules/user/user.controller.ts
siddarth2824: Now that I think about it, I think we would need to because the User object is cached on the client side because of `useQuery`. Hence if there is an update being done to the user and it's not returned, the last seen feature update date attribute on the client side will be stale for a period of time and this might cause incorrect behaviour when deciding whether or not to notify the user on the feature updates.

src/app/modules/user/user.middleware.ts
siddarth2824: The format of date here is consistent with the format specified in the `format` function and also moment.js format. It will be in UTC format only if the `utc()` function is called.

https://joi.dev/module/joi-date/api/?v=2.1.0

siddarth2824: I think we can use Date.now() timestamp instead of passing in the date in the request because as long as the viewed date (date at which the update function is called) is later than the latest feature update date, we won't need to notify the form admin users.

src/app/routes/api/v3/user/user.routes.ts
siddarth2824: I am not really sure what the purpose of a subrouter is in this context. Could you please elaborate as to why we would need to add a subrouter?

src/app/modules/user/user.middleware.ts
karrui: I prefer a time stamp too; it's less ambiguous. Should update the body prop to timestamp over date then!

siddarth2824: I changed my implementation to just generate the current date on the server instead of sending it over from the client. This is because it doesn't matter if I update the attribute using the date specified in the JSON file or generate the current date when the update API call was called since both of them will be equal to or later than the date specified in the JSON file.

karrui: that actually makes much more sense lmao. cheers!

src/app/modules/user/__tests__/user.controller.spec.ts
```diff
@@ -436,27 +432,36 @@ describe('user.controller', () => {
...
           _id: VALID_SESSION_USER_ID,
         },
       },
     })
+    // Mock implementaion of Date.now() due to slight difference in Date.now() execution in test and when
+    // Date.now() in updateUserLastSeenFeatureUpdateDate function is executed
+    jest
+      .spyOn(global.Date, 'now')
+      .mockReturnValue(new Date('2022-07-14T11:01:58.135Z').valueOf())
+    const MOCK_DATE = new Date('2022-07-14T11:01:58.135Z')
```
karrui, line 447: can probably create `MOCK_DATE` first then use that in the spy haha

src/app/modules/user/user.service.ts
```diff
@@ -166,27 +166,32 @@ export const updateUserContact = (
...
+ * @param date the latest last seen feature update date to update with
  * @returns ok(true) if update was successful
  * @returns err(MissingUserError) if user document cannot be found
  * @returns err(DatabaseError) if any error occurs whilst querying the database
  */
 export const updateUserLastSeenFeatureUpdateDate = (
   userId: IUserSchema['_id'],
-  latestLastSeenFeatureUpdateDate: Date,
-): ResultAsync<true, MissingUserError | DatabaseError> => {
+  date: Date,
```
karrui, line 176: instead of just passing in a date, should we just call `Date.now()` in this function?

src/app/modules/user/__tests__/user.controller.spec.ts
karrui: we also have a package specifically for mocking all `Date` constructor calls in tests called `MockDate`, might want to use it over this implementation

src/app/routes/api/v3/user/user.routes.ts
```diff
@@ -46,15 +46,15 @@ UserRouter.post('/contact/otp/verify', UserController.handleContactVerifyOtp)
...
+ * @route POST /user/flag/last-seen
+ * @param body.date the latest seen feature update date to update user with
  * @returns 200 when user last seen feature update updates sucessfully
  * @returns 401 if user is not currently logged in
  * @returns 422 when userId does not exist in the database
  * @returns 500 when database errors occurs
  */
 UserRouter.post(
-  '/whats-new/last-seen',
+  '/flag/last-seen',
```
karrui, line 57: suggest a more verbose endpoint like `new-features-last-seen`, since last-seen may be a little too ambiguous, like is this to track user's last login (we do track that in our schema btw hehe)?

src/app/modules/user/user.controller.ts
```diff
@@ -215,50 +209,38 @@ export const handleFetchUser: ControllerHandler = async (req, res) => {
...
  * @returns 422 when user id does not exist in the database
  * @returns 500 database errors occurs
  */
-export const handleUpdateUserLastSeenFeatureUpdateDate = [
-  validateUpdateUserLastSeenFeatureUpdateParams,
-  _handleUpdateUserLastSeenFeatureUpdateDate,
-] as ControllerHandler[]
+export const handleUpdateUserLastSeenFeatureUpdateDate: ControllerHandler =
+  async (req, res) => {
+    const currentDateTime = new Date(Date.now())
```
karrui, line 221: should this just be `new Date()`? i think its the same thing

src/app/modules/user/user.service.ts
siddarth2824: Yeap I think I should have done that earlier.

src/app/modules/user/user.controller.ts
```diff
@@ -44,7 +38,7 @@ export const _handleContactSendOtp: ControllerHandler<
   // Guard against user updating for a different user, or if user is not logged
   // in.
   if (!sessionUserId || sessionUserId !== userId) {
-    return res.status(StatusCodes.UNAUTHORIZED).json(UNAUTHORIZED_USER_MESSAGE)
+    return res.status(StatusCodes.UNAUTHORIZED).json('User is unauthorized.')
```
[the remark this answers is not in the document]
siddarth2824: HAHAHA I thought having a file with one constant was a bit unnecessary

src/app/modules/user/__tests__/user.controller.spec.ts
siddarth2824: I will check it and update accordingly

src/app/modules/user/user.controller.ts
siddarth2824: Yeap they are the same thing. I will make the appropriate change

src/app/modules/user/__tests__/user.controller.spec.ts
siddarth2824: When I checked through the codebase, I was not able to find any util file that mocks the Date constructor. All the spec files that mock the Date constructor calls did it by calling `jest.spyOn(Date, now).mockReturnValue()`. Could I ask where I can find the package which mocks the Date constructor calls?

karrui: https://github.com/opengovsg/FormSG/blob/develop/src/app/models/__tests__/user.server.model.spec.ts#L1

karrui (approved):
lgtm

PR #3998, opened 2022-06-16
feat: check if field allows intl numbers before sending intl OTP (#3973)

### LINKED ISSUE #3973: Verified field enhancements
See https://github.com/datagovsg/formsg-private/issues/141

justynoh:
## Problem
Currently, SMS OTP is generated for international mobile numbers even though the field from which it is requested did not allow intl numbers to be entered.

Closes #3973  

## Solution
Prior to sending the SMS OTP to an intl mobile number, check if the field it was requested from allows intl numbers to be added.

**Breaking Changes** 

## Tests

- [x] Should return OtpRequestError if SMS OTP is requested for an international number when the field does not allow it

### DIFF
--- a/src/app/modules/verification/__tests__/verification.service.spec.ts
+++ b/src/app/modules/verification/__tests__/verification.service.spec.ts
@@ -44,7 +44,7 @@ import dbHandler from 'tests/unit/backend/helpers/jest-db'
 
 import { BasicField } from '../../../../../shared/types'
 import { SMS_WARNING_TIERS } from '../../../../../shared/utils/verification'
-import { DatabaseError, MalformedParametersError } from '../../core/core.errors'
+import { DatabaseError } from '../../core/core.errors'
 import * as AdminFormService from '../../form/admin-form/admin-form.service'
 import { FormNotFoundError } from '../../form/form.errors'
 import * as FormUtils from '../../form/form.utils'
@@ -1043,7 +1043,7 @@ describe('Verification service', () => {
       expect(actual._unsafeUnwrapErr()).toBeInstanceOf(OtpRequestError)
     })
 
-    it('should return MalformedParametersError if there are no matching form fields with the correct id', async () => {
+    it('should return OtpRequestError if there are no matching form fields with the correct id', async () => {
       // Arrange
       const mockForm = {
         form_fields: [
@@ -1064,10 +1064,10 @@ describe('Verification service', () => {
       )
 
       // Assert
-      expect(actual._unsafeUnwrapErr()).toBeInstanceOf(MalformedParametersError)
+      expect(actual._unsafeUnwrapErr()).toBeInstanceOf(OtpRequestError)
     })
 
-    it('should return MalformedParametersError if form_fields is empty', async () => {
+    it('should return OtpRequestError if form_fields is empty', async () => {
       // Arrange
       const mockForm = {
         form_fields: [],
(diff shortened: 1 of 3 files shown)


src/app/modules/verification/verification.service.ts
```diff
@@ -611,20 +614,33 @@ const checkSmsCountAndPerformAction = (
...
+ * Check whether we should generate an OTP according to the requirements:
+ * 1. the field in the form is verifiable, and
+ * 2. if the recipient is within the allowed countries set by the field
+ * If these conditions are not met, then don't generate an OTP.
  */
 export const shouldGenerateMobileOtp = (
   { form_fields }: Pick<IFormSchema, 'form_fields'>,
   fieldId: string,
+  recipient: string,
 ): ResultAsync<true, OtpRequestError> => {
```
karrui, line 626: update the return type to also return `MalformedParametersError`?
Any reason why this error is used instead of reusing `OtpRequestError`? I would think all errors that occur in this function qualifies as a request error

src/app/modules/verification/verification.service.ts
```diff
@@ -611,20 +614,33 @@ const checkSmsCountAndPerformAction = (
...
-    ).length > 0
+  // Get the field with this fieldId
+  const field = form_fields?.find(({ _id }) => fieldId === String(_id))
+
+  if (!field)
+    return errAsync(new MalformedParametersError('Field id not found'))
+
+  if (field.fieldType !== BasicField.Mobile) {
+    return errAsync(
+      new MalformedParametersError('Field id is not a mobile field'),
```
karrui, line 635: does returning `MalformedParametersError` change any error mapping in the controller?? do we return the expected error message when that happens?

src/app/modules/verification/verification.service.ts
justynoh: Seems that if anything goes wrong in those checks it's an issue with how the args of the call to `shouldGenerateMobileOtp` was formed, and its similar to how `MalformedParametersError` is used in `sendNewOtp`. Changing it to `OtpRequestError` is fine too, just that these errors are more of the type where "we can't find the info we need to send it" rather than "we are not going to send it" so i thought this was more appropriate. 

Noted on the updated return type tho - thanks for catching that.

justynoh: the second error is also `MalformedParametersError` because we need to access `allowIntlNumbers` which only exists in the mobile field

src/app/modules/verification/verification.service.ts
tshuli: Minor nit that you should never see this case on the frontend (it's not possible for the field type to change). So maybe `OtpRequestError` is more appropriate

`if (field.fieldType !== BasicField.Mobile)`

tshuli: Or update the mapping to change the error message

justynoh: Sure, will make that change!

src/app/modules/verification/verification.service.ts
justynoh: cancel last, i'll change it to `OtpRequestError` :)

src/app/modules/verification/__tests__/verification.service.spec.ts
```diff
@@ -1047,29 +1053,58 @@ describe('Verification service', () => {
...
+        recipient,
       )
 
       // Assert
-      expect(actual._unsafeUnwrapErr()).toBeInstanceOf(OtpRequestError)
+      expect(actual._unsafeUnwrapErr()).toBeInstanceOf(MalformedParametersError)
     })
 
-    it('should return OtpRequestError if form_fields is empty', async () => {
+    it('should return MalformedParametersError if form_fields is empty', async () => {
```
karrui, line 1070: did these tests not fail :O

karrui: ah the ci pipeline bugged out; i've cancelled them as they most definitely will fail haha

justynoh: uh oh

tshuli (approved):
lgtm

PR #3965, opened 2022-06-07
feat(v2/storage-4): add storage mode response download modals

karrui:
## Problem

This PR adds storage mode response download modals.

TODOs to be done in future PRs (too keep this PR smaller):
- [ ] add summary modal when download has completed, or errored.

Related to #3129

## Solution

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Features**:
- progress modal will show up 5 seconds after download (if download is not yet complete) for non-attachment downloads
- show confirmation modal on downloading responses with attachments, which will segue to progress modal
- allow cancellation of decryption
- add prompt when user attempts to leave the page
- add Progress component theme

## Before & After Screenshots
New stories have been added for `DownloadWithAttachmentModal` and `ProgressModal` component.

Issue #3129, opened 2021-12-02
[React/StorageMode] Download storage mode form responses


karrui:
Note that even if #2791 is not completed we can mock something light to allow for the download of responses

karrui:
Closed by the above linked PRs

PR #3999, opened 2022-06-16
feat(v2/storage-5): allow download of responses with date range

karrui:
Builds on #3965

## Problem

This PR allows for downloading of responses by entering a date range. Also updates the method of calculating the number of responses to download by retrieving responses count by date on date range selection instead of only during decryption.

Related to #3129

## Solution

**Features**:

- feat: pass in selected date range to storage download
- feat: pass separate `dateRangeResponsesCount` prop via context
- feat: use `dateRangeResponseCount` for rendering download progress

**Improvements**:

- ref(storage): rename `responsesCount` prop passed down from context to `totalResponsesCount`

## Before & After Screenshots
None needed i think...

### DIFF
--- a/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
+++ b/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
@@ -23,9 +23,13 @@ export const countFormSubmissions = async ({
   dates,
 }: {
   formId: string
-  dates?: SubmissionCountQueryDto
+  dates?: Partial<SubmissionCountQueryDto>
 }): Promise<number> => {
   const queryUrl = `${ADMIN_FORM_ENDPOINT}/${formId}/submissions/count`
+  if (dates && (!dates.startDate || !dates.endDate)) {
+    delete dates.startDate
+    delete dates.endDate
+  }
   if (dates) {
     return ApiService.get(queryUrl, {
       params: { ...dates },
--- a/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
+++ b/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
@@ -26,13 +26,9 @@ export const countFormSubmissions = async ({
   dates?: Partial<SubmissionCountQueryDto>
 }): Promise<number> => {
   const queryUrl = `${ADMIN_FORM_ENDPOINT}/${formId}/submissions/count`
-  if (dates && (!dates.startDate || !dates.endDate)) {
-    delete dates.startDate
-    delete dates.endDate
-  }
-  if (dates) {
+  if (dates && dates.startDate && dates.endDate) {
     return ApiService.get(queryUrl, {
-      params: { ...dates },
+      params: dates,
     }).then(({ data }) => data)
   }
   return ApiService.get(queryUrl).then(({ data }) => data)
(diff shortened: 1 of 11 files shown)


frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -18,9 +18,13 @@ export const countFormSubmissions = async ({
   dates,
 }: {
   formId: string
-  dates?: SubmissionCountQueryDto
+  dates?: Partial<SubmissionCountQueryDto>
 }): Promise<number> => {
   const queryUrl = `${ADMIN_FORM_ENDPOINT}/${formId}/submissions/count`
+  if (dates && (!dates.startDate || !dates.endDate)) {
+    delete dates.startDate
+    delete dates.endDate
```
tshuli: Why is this necessary; what happens if we simply pass in undefined startDate and endDate? 

Also should we allow a case where only the startDate or endDate is specified? Right now both must be specified to pass the test

karrui: following current implementation tbh, server will reject if only one date is supplied

frontend/src/features/admin-form/responses/ResponsesPage/storage/StorageResponsesProvider.tsx
```diff
@@ -29,25 +38,31 @@ export const StorageResponsesProvider = ({
...
+    if (!secretKey || !dateRangeResponsesCount) return null
 
     return {
       secretKey,
-      // TODO: Add selector for start and end dates.
-      endDate: undefined,
-      startDate: undefined,
+      responsesCount: dateRangeResponsesCount,
+      startDate: dateRange[0],
+      endDate: dateRange[1],
```
timotheeg, line 47: Just to confirm my understanding: if `dateRangeResponsesCount` is greater than 0, it means for sure `dateRange` is not `undefined`, is that correct?

I'm asking because lines 30+31 indicate `dateRange` _could_ be `undefined`, but by the time we reach this code here, `dateRange` is guaranteed to be defined, and therefore we don't need the optional chaining syntax any more, correct?

Extension to this question: Is there a guarantee that `setDateRange()` will never be called with `undefined`, such that we could change line 30+31 to remove the optional chaining?

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -18,9 +18,13 @@ export const countFormSubmissions = async ({
...
+  }
   if (dates) {
```
timotheeg: /nit suggestion

```suggestion
  if (dates && dates.startDate && dates.endDate) {
```

frontend/src/features/admin-form/responses/ResponsesPage/storage/StorageResponsesProvider.tsx
karrui: ah yes, brain fart. will remove optional chain.

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
karrui: done.

timotheeg (approved):
lgtm 👍

PR #4000, opened 2022-06-16
feat(storage-responses/1): render storage mode responses metadata in table

karrui:
No panik at number of lines changed... most come from storybook and msw setup.

edit: wow we are at #4000... i think even the old formsg repo before we open sourced had have this amount... that's awesome.

## Problem

This PR implements the table to display metadata of a storage-mode form's responses.

There are also some groundwork done to prepare for navigation of individual responses when clicking into the table row, though that full implementation will come in a later PR. 

Closes #3127

## Solution

**Features**:

- feat: add ResponsesTable component to render metadata of storage-mode responses.
  - some prerendering of the next +/- 1 pages of metadata is done to facilitate individual response navigation, may be removed if it adds too much strain to the server. 

**Improvements**:

- feat: move StorageResponsesProvider into admin form layout 
  - so navigating throughout the form will not reset the secret key once entered 
- feat: update app router to use ResponsesLayout for consistent layout inside the responses subroute.

- Details ...

## Before & After Screenshots
New stories have been added to render unlocked storage mode responses view.

Issue #3128, opened 2021-12-02
[React/StorageMode] View individual storage mode responses


karrui:
Related to #3127, this issue concerns itself with the decryption of individual responses

~~This is marked KIV, pending further user testing~~
Design now available

karrui:
Closed by #4001

PR #4001, opened 2022-06-16
feat(storage-responses/2): render individual storage mode form response

karrui:
> Builds on top of PR to add ResponseTable, #4000.

## Problem

This PR implements individual storage mode form response feature. 
It is available on its own endpoint, meaning form admins can navigate to the individual response if they wanted to.

Some metadata is lost if an individual response is navigated to, such as its respondent index from the table, or the ability to go to the next/previous response.
If navigated via the response table, moving to the next/previous respondent will move the page number of the table backlink accordingly.

Allows for downloading of individual attachment in this view, though not the entire attachments of the individual response (pending design).

Related to #3128

TODO: (will be done in future PRs)
- [ ] add feature to download all attachments in submission to achieve feature parity with current AngularJS client.
- [ ] add hover/active styling when clicking in from response table

## Solution

**Features**:

- feat: add `IndividualResponsePage` and route
- feat: add `DecryptedRow` component to render responses inside the decrypted submission
- feat: allow nav between individual responses when metadata exists

**Bug Fixes**:

> Bug fix logic will also be moved to the `develop` branch for backwards compatibility

This PR also contains a bug fix for the bug where the server will reject form submissions due to perceived missing fields due to the change in  processing of mobile and email fields in storage mode:
- previously (and the current behaviour in `develop`), all mobile and email field answers are sent to the server in storage mode due to the possibility of requiring the data to email users or to verify fields. 
- in  #3600, the behaviour of the fields to send to the server in storage mode was tightened to only send the minimum data required. This conflicts with the fields the server expects to receive and would return a conflict error.

## Before & After Screenshots
TODO: add stories

### DIFF
--- a/frontend/src/app/AppRouter.tsx
+++ b/frontend/src/app/AppRouter.tsx
@@ -21,6 +21,7 @@ import { CreatePage } from '~features/admin-form/create/CreatePage'
 import {
   FeedbackPage,
   FormResultsLayout,
+  IndividualResponsePage,
   ResponsesLayout,
   ResponsesPage,
 } from '~features/admin-form/responses'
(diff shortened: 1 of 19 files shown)


timotheeg (approved):
lgtm

PR #4004, opened 2022-06-17
fix(v2): prevent everchanging snapshots from triggering chromatic regressions

karrui:
Some stories always trigger changes that result in false positives such as lottie animations, text rendering depending on current url, or popper positioning.

This PR adds code to tell Chromatic to ignore those specific elements in the story.


frontend/src/mocks/msw/handlers/user.ts
```diff
@@ -34,18 +32,6 @@ export const getUser = ({
...
-          ctx.delay(),
-          ctx.status(403),
-          ctx.json({
-            message: 'Not authorized',
-          }),
-        )
-      }
       // If authenticated, return a mocked user details
       return res(ctx.delay(delay), ctx.status(200), ctx.json(mockUser))
     },
```
siddarth2824, line 37: If the mock getUser function no longer handles the case where the user is not authenticated, won't there be no necessity for the return type to be `UserDto | ErrorDto` because it will always be `UserDto`?

frontend/src/mocks/msw/handlers/user.ts
```diff
@@ -34,18 +32,6 @@ export const getUser = ({
...
-      if (!isAuthenticated) {
-        // If not authenticated, respond with a 403 error
-        return res(
-          ctx.delay(),
-          ctx.status(403),
-          ctx.json({
-            message: 'Not authorized',
-          }),
-        )
-      }
```
siddarth2824, line 48: If the case where the user is not authenticated is not handled in this mock handler function, is there a necessity for the return type to be `UserDto | ErrorDto` instead of just `UserDto`? This is because the function will always return a `UserDto` object.

tshuli (approved):

siddarth2824 (approved):
LGTM

PR #4011, opened 2022-06-19
feat: Singpass OIDC migration

### LINKED ISSUE #2600: OIDC migration
## Context

SingPass (SP) is to migrating away from SAML into OIDC fully by year end, and do the same for CorpPass (CP) next year. 

FormSG integrates to SPCP as an integrator. When end-users authenticate over SP, FormSG does the authentication on behalf of the government officer who administers the form. This is achieved by officers getting an identifier (named `eServiceId`), that FormSG passes along the SMAL auth.

This custom carrying of `eServiceID` eventually allows SP to bill the correct form owners, and not FormSG itself.

The new proposed OIDC flow does not have provision for `eServiceID`. Instead, OIDC authentication requests are only identified by a client ID representing the application making the call (In our case FormSG). This single identifier system thus threatens to break our billing operations.

A meeting took place mid June with several solutions proposed, but the issue has fallen silent since. During that meeting, some solution were looked at:
1. SingPass to update their OIDC flow to accept the `eServiceID` as an additional parameter
    * This option represents SingPass continuing support of the existing "integrator client" capability
2. OGP to be the central bearer of all SingPass login costs
    * How many SingPass logins per month do we see? Where will budget come from?
3. OGP to handle identity management for officers
    * OGP to somehow (?) work with officers to create ClientIDs
    * OGP to store such clientID (and secret if needed?)
    * OGP make OIDC calls to Singpass with the clientID of the officer who owns the form.

Option 1 places the burden of implementation on Singpass.

Option 2 fastest to implement, but clearly has running cost implications that need to be sorted out.

Option 3 is the hardest for OGP to build, and creates onboarding friction for officers, which reduces the appeal of the FormSg service.

## TODOs

* Revive discussions with Singpass
* Pick a definitive solution
* Implement before SAML auth flow gets retired entirely

The baseline implementation with a single clientID representing FormSG itself should still be undertaken asap:
* Offer public API to expose our public key
* Integrate the OIDC flow with FormSG's own clientID

## References
* [Singpass OIDC Design Document](https://docs.google.com/document/d/155Bv7T-duB202Dd8JHtjQJTNq7EFqv56CNwWjr4fNdg/edit)
* [Singpass OIDC specs v1.1](https://drive.google.com/file/d/1u0faFKeWl3pQ8esPjDxqpAgqw8OE95q1/view?usp=sharing)
* [FormSG Client Submission](https://docs.google.com/spreadsheets/d/1gnQ7Gtrd2FF9FkIllRt6Cqp1SxXCDDnV/edit?usp=sharing&ouid=114367138436779320097&rtpof=true&sd=true)
* [JSON web key Generator](https://github.com/mitreid-connect/json-web-key-generator)
* [OpenID Connect specs](https://openid.net/developers/specs/)

tshuli:
**Problem**
This PR implements Singpass integration using the OIDC protocol.

Details on key generation and key change procedure can be found in the internal procedures.

Closes #2600

**Breaking Changes** 

- [x] No - this PR is backwards compatible. 

- However, note that the new JWK keys will be used for signing and verifying singpass JWTs for the singpass session. Similar to any migration of SP keys, any existing JWTs will not be valid and users will need to re-login in order to regenerate the signed JWTs.
- In the event of a rollback, users with a JWT generated using the new keys will likewise be required to re-login to Singpass.

**Additional Details**

- As the Singpass authentication flow for OIDC can be completed even with an invalid eservice Id, there is no way for us to check for eservice ID validity at the point of form activation. Therefore, this check will be disabled.

## Tests

**Form Submission**
- [ ] Check that Singpass, CP, Myinfo, Sgid forms can be created and submitted normally
- [ ] Check that jwt validity is correct (1 month) if `Keep Me Logged In` is selected for singpass form (extract jwt from cookie and decode using `token.dev`)

**EserviceId Check**
- [ ] Check that on form activation, SP and Myinfo now no longer perform eserviceid check

**Prefill**
- [ ] Check that prefill works with singpass form

## Deploy Notes
- New JWKS endpoint should be deployed ahead of the main body of code, to ensure availability across all servers during the rolling deployment process
- New JWKS keysets should be loaded onto EFS and env vars updated ahead of deployment

- `env var` : env var details
| `SP_OIDC_NDI_DISCOVERY_ENDPOINT` | NDI's Singpass OIDC Discovery Endpoint                                                                                                                                                          |
| `SP_OIDC_NDI_JWKS_ENDPOINT`      | NDI's Singpass OIDC JWKS Endpoint                                                                                                                                                               |
| `SP_OIDC_RP_CLIENT_ID`           | The Relying Party's Client ID as registered with NDI                                                                                                                                            |
| `SP_OIDC_RP_REDIRECT_URL`        | The Relying Party's Redirect URL                                                                                                                                                                |
| `SP_OIDC_RP_JWKS_PUBLIC_PATH`    | Path to the Relying Party's Public Json Web Key Set used for Singpass-related communication with NDI. This will be hosted at /singpass/.well-known/jwks.json endpoint.                          |
| `SP_OIDC_RP_JWKS_SECRET_PATH`    | Path to the Relying Party's Secret Json Web Key Set used for Singpass-related communication with NDI                                                                                            |

**New dependencies**:

`openid-connect`, `jose`, `jwk-to-pem`, `promise-timeout`, `ulid`


src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+
+      if (!isECPrivate(jwk)) {
+        throw new JwkError()
+      }
+
+      const cryptoKeySet = {
+        kid: jwk.kid,
+        use: jwk.use,
+        alg: jwk.alg,
+        key: createPrivateKey(jwkToPem(jwk, { private: true })), // Conversion to pem is necessary because in node 14, crypto does not support import of JWK directly
```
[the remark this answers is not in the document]
karrui: are we affected by this issue? https://github.com/Brightspace/node-jwk-to-pem/issues/107

src/app/modules/spcp/sp.oidc.client.types.ts
```diff
@@ -0,0 +1,102 @@
...
+    hasProp(jwk, 'crv') &&
+    typeof jwk.crv === 'string' &&
+    hasProp(jwk, 'd') &&
+    typeof jwk.d === 'string'
+  )
+}
+
+export const isSigningKey = (key: CryptoKey): key is SigningKey => {
+  return !!key.alg && key.use === 'sig'
+}
```
karrui, line 102: these functions should be extracted to an utils directory/file imho, they are not really type declarations

src/app/modules/spcp/sp.oidc.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      spOidcNdiDiscoveryEndpoint: props.spOidcNdiDiscoveryEndpoint,
+      spOidcNdiJwksEndpoint: props.spOidcNdiJwksEndpoint,
+      spOidcRpPublicJwks: JSON.parse(
+        fs.readFileSync(props.spOidcRpJwksPublicPath).toString(),
+      ),
+      spOidcRpSecretJwks: JSON.parse(
+        fs.readFileSync(props.spOidcRpJwksSecretPath).toString(),
+      ),
+    })
+    this.#spOidcProps = props
```
karrui, line 56: nit: i think there is no need to pass in all the props right? i feel like we should only give the props the class needs

from a quick scan, only 3 subprops are used: 
`spCookieMaxAgePreserved`, `spCookieMaxAge` and `spcpCookieDomain`

src/app/modules/spcp/sp.oidc.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      spOidcRpClientId: props.spOidcRpClientId,
+      spOidcRpRedirectUrl: props.spOidcRpRedirectUrl,
+      spOidcNdiDiscoveryEndpoint: props.spOidcNdiDiscoveryEndpoint,
+      spOidcNdiJwksEndpoint: props.spOidcNdiJwksEndpoint,
+      spOidcRpPublicJwks: JSON.parse(
+        fs.readFileSync(props.spOidcRpJwksPublicPath).toString(),
+      ),
+      spOidcRpSecretJwks: JSON.parse(
+        fs.readFileSync(props.spOidcRpJwksSecretPath).toString(),
+      ),
```
karrui, line 54: these functions may throw, will they be caught if so?

src/app/modules/spcp/sp.oidc.service.ts
```diff
@@ -0,0 +1,310 @@
...
+
+  /**
+   * Extracts the SP JWT from an object containing cookies
+   * @param cookies Object containing cookies
+   * @returns ok(cookie)
+   * @returns err(missingJwtError) if the SP JWT does not exist
+   */
+  extractJwt(cookies: SpcpCookies): Result<string, MissingJwtError> {
+    const jwtName = JwtName.SP
+    const cookie = cookies[jwtName]
```
karrui, line 97: nit: simplify?

```ts
const cookie = cookies[JwtName.SP]
```

src/app/modules/spcp/sp.oidc.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      this.#spOidcClient
+        .exchangeAuthCodeAndDecodeVerifyToken(code)
+        .then((decodedVerifiedToken) => {
+          return this.#spOidcClient.extractNricFromIdToken(decodedVerifiedToken)
+        })
+        .then((result) => {
+          if (result instanceof Error) {
+            return Promise.reject(result)
+          }
+          return result
```
karrui, line 191: comment: should we just use neverthrow in `extractNricFromIdToken` directly instead of this?

src/app/modules/spcp/sp.oidc.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      decodedQuery = encodedQuery
+        ? `?${Buffer.from(encodedQuery, 'base64').toString('utf8')}`
+        : ''
+    } catch (e) {
+      logger.error({
+        message: 'Unable to decode encodedQuery',
+        meta: {
+          action: 'parseOOBParams',
+          encodedQuery,
+        },
```
karrui, line 239: seems like incorrect `action`, should we use `logMeta` here too?

src/app/modules/spcp/spcp.controller.ts
```diff
@@ -184,3 +186,112 @@ export const handleLogin: (
...
+ */
+export const handleSpOidcLogin: ControllerHandler<
+  unknown,
+  unknown,
+  unknown,
+  { state: string; code: string }
+> = async (req, res) => {
+  const { state, code } = req.query
+  const logMeta = {
+    action: 'handleLogin',
```
karrui, line 214: typo: action should match function name?

src/app/loaders/express/index.ts
```diff
@@ -169,6 +170,8 @@ const loadExpressApp = async (connection: Connection) => {
   // Registered routes with the Singpass/Corppass servers
   app.use('/singpass/login', SingpassLoginRouter)
   app.use('/corppass/login', CorppassLoginRouter)
+  // jwks endpoint for SP OIDC
+  app.use('/singpass/.well-known/jwks.json', SpOidcJwksRouter)
```
karrui, line 174: nit: i feel like this should have been in the /api/v3/singpass router but i guess its too late :(

src/app/modules/spcp/spcp.routes.ts
```diff
@@ -95,3 +95,14 @@ CorppassLoginRouter.get(
   loginParamsMiddleware,
   SpcpController.handleLogin(FormAuthType.CP),
 )
+
+// Handles SingPass JWKS requests
+export const SpOidcJwksRouter = Router()
```
karrui, line 100: should this be its own file in our root `/routes` directory? to better follow the mapping of routes in our application.

karrui (commented):
preliminary comments... i now know how people feel when they review my monster PRs

src/app/modules/spcp/__tests__/sp.oidc.client.spec.ts
```diff
@@ -0,0 +1,685 @@
+import axios from 'axios'
+import { createPublicKey } from 'crypto'
+import fs from 'fs'
+import jwkToPem, { EC } from 'jwk-to-pem'
```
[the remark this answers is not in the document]
tshuli: we did! in [82ba7b4](https://github.com/opengovsg/FormSG/pull/4011/commits/82ba7b40892131c1c1e062cfb2375e4373db2aaa)

src/app/modules/spcp/__tests__/spcp.test.constants.ts
```diff
@@ -40,6 +40,13 @@ export const MOCK_SERVICE_PARAMS: ISpcpMyInfo = {
   myInfoCertPath: 'myInfoCertPath',
   myInfoClientId: 'myInfoClientId',
   myInfoClientSecret: 'myInfoClientSecret',
+  spOidcNdiDiscoveryEndpoint:
+    'https://stg-id.singpass.gov.sg/.well-known/openid-configuration',
+  spOidcNdiJwksEndpoint: 'https://stg-id.singpass.gov.sg/.well-known/keys',
+  spOidcRpClientId: 'spOidcRpClientId',
+  spOidcRpRedirectUrl: 'https://staging.form.gov.sg/singpass/login',
```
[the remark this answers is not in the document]
tshuli: nope, not called in the tests - will replace with dummy urls

src/app/modules/spcp/sp.oidc.client.errors.ts
```diff
@@ -0,0 +1,80 @@
+/**
+ * Error while creating Authorisation URL
+ */
+export class CreateAuthorisationUrlError extends Error {
```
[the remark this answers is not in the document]
tshuli: Yes I thought about this (same for use of neverthrow below) - at some point (similar to SpcpAuthClient / MyInfoGovClient) we could publish the SpcpOidcClient as a standalone package. So the client is now written to be standalone to facilitate that. Thoughts?

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+    const spOidcNdiPublicJwks = await getJwksWithRetries
+
+    return spOidcNdiPublicJwks.keys.map((jwk) => {
+      if (!isEC(jwk) || !jwk.kid || !jwk.use) {
+        throw new JwkError()
+      }
+      return {
+        kid: jwk.kid,
+        use: jwk.use,
+        key: createPublicKey(jwkToPem(jwk)), // Conversion to pem is necessary because in node 14, crypto does not support import of JWK directly
```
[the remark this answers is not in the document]
tshuli: Ok, will add

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+  ): Promise<string> {
+    const possibleSigningKeys = this.#spOidcRpSecretKeys.filter(
+      (key): key is SigningKey => isSigningKey(key),
+    )
+
+    if (possibleSigningKeys.length === 0) {
+      throw new CreateJwtError('Create JWT failed. No signing keys found.')
+    }
+
+    const signingKey = possibleSigningKeys[0]
```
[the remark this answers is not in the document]
tshuli: Yes. In the event of key rotation, we only need to expose the old + new public signing keys for 1h (to allow NDI to refresh cache), and then load the new secret signing key on our servers.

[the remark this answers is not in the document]
tshuli: Will add comment

src/app/modules/spcp/sp.oidc.client.types.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @param obj Object
+ * @param prop Property to check
+ */
+
+const hasProp = <K extends string>(
+  obj: unknown,
+  prop: K,
+): obj is Record<K, unknown> => {
+  return typeof obj === 'object' && obj !== null && prop in obj
+}
```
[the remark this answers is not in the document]
tshuli: Yes - same as above reasons, I intended for the client to be standalone, hence duplicated it

src/app/modules/spcp/sp.oidc.client.ts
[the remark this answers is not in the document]
tshuli: Hmm possibly. Testing locally the conversion takes 370ms. But we only incur this once when the RP's private keys are first loaded, so it doesn't affect the performance thereafter
The issue will be resolved when we upgrade to Node >=15.12 and can use the jwk directly

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+    this.#spOidcNdiDiscoveryEndpoint = spOidcNdiDiscoveryEndpoint
+    this.#spOidcNdiJwksEndpoint = spOidcNdiJwksEndpoint
+    this.#spOidcRpClientId = spOidcRpClientId
+    this.#spOidcRpRedirectUrl = spOidcRpRedirectUrl
+    this.#spOidcRpSecretJwks = spOidcRpSecretJwks
+
+    // On expiry, refresh cache. If fail to refresh, log but do not throw error.
+    this.on('expired', () =>
+      this.refresh().catch((err) =>
+        console.warn(
```
timotheeg, line 75: No logger for this?

src/app/modules/spcp/spcp.controller.ts
```diff
@@ -184,3 +186,112 @@ export const handleLogin: (
...
+ * Handler to return the public JWKS hosted on the app's well known end point
+ * @returns public JWKS in json format
+ */
+export const handleGetWellKnown: ControllerHandler<
+  unknown,
+  Record<string, unknown>,
+  unknown,
+  unknown
+> = async (_req, res) => {
+  return res.json(SpOidcService.getRpPublicJwks())
```
timotheeg, line 200: /nit: json-encoding at every request feels a little wasted since the json data is actually static and extracted from a json blob. Perhaps we could return the file from `spOidcRpJwksPublicPath` instead.

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+          `Attempted but failed to refresh sp oidc cache on expiry. Reason: ${err}`,
+        ),
+      ),
+    )
+
+    // Trigger refresh on instantiation to populate cache. If fail to refresh, log but do not throw error.
+    void this.refresh().catch((err) =>
```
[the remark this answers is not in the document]
tshuli: > Ideally, the implementation should only have a single in-flight refresh per node process at a time

Can look into this

> The app is not considered ready till we have fetched the data

This approach might be problematic, because the data fetch could fail for multiple reasons (including downtime by NDI). Singpass integration is only one feature of the forms app, so we should not couple the availability of the entire form app to this. Hence, in the current implementation, if the data is not available (whether in cache or from NDI), an error is thrown and handled appropriately

src/app/modules/spcp/sp.oidc.client.ts
tshuli: No for now, because the intent was to keep the client standalone as much as possible - see [discussion](https://github.com/opengovsg/FormSG/pull/4011#discussion_r902153742), but welcome thoughts on it

src/app/modules/spcp/spcp.controller.ts
tshuli: Possible too. Right now the approach is to load the public key into the client on instantiation, and then retrieve it from there for the jwks endpoint - this ensures that it is always in sync. But ok to return the file directly too in the controller

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+  SpOidcClientCacheConstructorParams,
+  SpOidcClientConstructorParams,
+} from './sp.oidc.client.types'
+
+/**
+ * Cache class which provides read-through capability and refresh-ahead before expiry
+ * Handles discovery and retrieval of NDI's public jwks
+ * Exported for testing
+ */
+export class SpOidcClientCache extends NodeCache {
```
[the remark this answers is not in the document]
tshuli: thanks, good point, will modify

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+   * @returns object {ndiPublicKeys, baseClient}
+   */
+  async refresh(): Promise<Refresh> {
+    const [ndiPublicKeys, baseClient] = await Promise.all([
+      this.retrievePublicKeysFromNdi(),
+      this.retrieveBaseClientFromNdi(),
+    ])
+
+    this.set('ndiPublicKeys', ndiPublicKeys, 3600) // TTL of 60 minutes
+    this.set('baseClient', baseClient, 3600) // TTL of 60 minutes
```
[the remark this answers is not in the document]
tshuli: To explain the thinking here - if the refresh ahead fails (max 3 attempts based on specs), then that means we will keep the keys indefinitely until our interaction with NDI fails subsequently (which indicates keys no longer valid). 

Hence, the `ndiPublicKeys` and `baseClient` keys have a 1h TTL so that at that point, another attempt is made to refresh the keys. If the keys are still not available, then we return an error (but try to refetch the keys on every request, which re-sets the TTL).

An alternative I can think of is to not set expiry on the 2 keys, but keep on attempting to refresh indefinitely from the 55min mark until success, with exponential backoff for interval between refresh. Thoughts?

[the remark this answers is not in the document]
timotheeg: Thanks for that ShuLi! The alternative you propose sounds good to me 👍 

To clarify a little further, I had 2 considerations with my comment:
1. As long as we have managed to fetch data, let's optimistically keep it around and reuse it, under the assumption that NDI data just doesn't change all that often anyway.
2. Let's be really careful with parallel refreshes

For 2. With the current code, if the fetch-ahead fails, and the keys then expire, parallel refreshes can be triggered by inbound calls to the app, which is not ideal.

Keeping the keys alive ensures that can't happen (except at startup), and indeed we should have an independent process running to continuously attempt to (cheaply) refresh the keys.

If the code implements an additional guarantees that there can only be a single refresh in-flight at a time. That's even better.

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+
+  /**
+   * Method to retrieve NDI's public keys from cache
+   * @async
+   * @returns NDI's public keys
+   */
+  async getNdiPublicKeys(): Promise<CryptoKeySet> {
+    const ndiPublicKeys = this.get<CryptoKeySet>('ndiPublicKeys')
+    if (!ndiPublicKeys) {
+      const { ndiPublicKeys } = await this.refresh()
```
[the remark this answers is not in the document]
tshuli: thanks - added in jsdoc (explained that not using neverthrow to keep this module standalone)

src/app/modules/spcp/sp.oidc.client.ts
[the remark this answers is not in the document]
tshuli: added issue #4021

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+   */
+  constructor(config: SpOidcClientConstructorParams) {
+    const {
+      spOidcRpClientId,
+      spOidcRpRedirectUrl,
+      spOidcNdiDiscoveryEndpoint,
+      spOidcNdiJwksEndpoint,
+      spOidcRpSecretJwks,
+      spOidcRpPublicJwks,
+    } = config
```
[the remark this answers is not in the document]
tshuli: thanks! edited

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    const baseClient = await this.getBaseClientFromCache()
+
+    const authorisationUrl = baseClient.authorizationUrl({
+      scope: 'openid',
+      response_type: 'code',
+      state: state,
+      esrvc: esrvcId,
+      nonce: String(new ObjectId()), // Not used - nonce is a required parameter for SPCP's OIDC implementation although it is optional in OIDC specs
```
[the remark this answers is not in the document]
tshuli: haha ok ulid then~

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+  }
+}
+
+/**
+ * Wrapper around the openid-client library to carry out authentication related tasks with Singpass NDI,
+ * and provides methods for decryption and verification of JWE/JWS returned by NDI after authorisation code exchange
+ */
+export class SpOidcClient {
+  #spOidcRpSecretKeys: CryptoKeySet
+  #spOidcRpPublicKeys: CryptoKeySet
```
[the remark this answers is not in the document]
tshuli: actually no, it was not meant to be a set. I've renamed it to CryptoKeys.

It is not a set because I need to filter on `kid` and `use` values. There may be multiple keys available, but not all of them may be usable.

It is not an object because I have more than one index (`kid` and `use`)

src/app/modules/spcp/sp.oidc.client.types.ts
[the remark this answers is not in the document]
tshuli: actually on further thought I will import it. can disentangle later if necessary

src/app/modules/spcp/sp.oidc.service.ts
tshuli: thanks! as above, would not use neverthrow, in order to keep the client standalone

src/app/modules/spcp/sp.oidc.client.ts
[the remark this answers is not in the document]
tshuli: update: implemented single in-flight refresh per node process, by storing the promise on a refresh attempt until it is fulfilled, and returning that promise to future refresh attempts

src/app/modules/spcp/spcp.routes.ts
tshuli: yes, good idea!

src/app/modules/spcp/sp.oidc.client.ts
```diff
@@ -0,0 +1,583 @@
...
+   * @returns nric string
+   * @returns InvalidIdTokenError is nric not found in idToken
+   */
+  extractNricFromIdToken(
+    idToken: JWTVerifyResult,
+  ): string | InvalidIdTokenError {
+    if (!idToken.payload.sub) {
+      return new InvalidIdTokenError('sub attribute missing in idToken payload')
+    }
+    const nricMatches = idToken.payload.sub.match(/s=([STFG])(\d{7})([A-Z])/g)
```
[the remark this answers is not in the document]
tshuli: thanks - i used the validation functions directly, instead of exporting the regex

src/app/modules/spcp/sp.oidc.service.ts
tshuli: No - if the functions throw (e.g. file does not exist at the specified path), the `SpOidcServiceClass` constructor will throw an error, the app will crash and the instances will be terminated (before they are deployed to accept traffic). I think this is better than trying to catch and handle the error because there is no way that the class can be instantiated without a valid jwks.

Is also similar to the existing SpcpServiceClass.

src/app/modules/spcp/sp.oidc.client.ts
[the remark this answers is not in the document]
tshuli: Thanks! Originally used `console` to keep it standalone, but actually on review will use logger for better monitoring. We can think of a way to disentangle next time if we publish it as a standalone package e.g. pass in the logger as a callback to the constructor params

tshuli: On review will use logger for better monitoring

src/app/modules/spcp/sp.oidc.client.cache.ts
```diff
@@ -0,0 +1,233 @@
...
+    this.#spOidcNdiDiscoveryEndpoint = spOidcNdiDiscoveryEndpoint
+    this.#spOidcNdiJwksEndpoint = spOidcNdiJwksEndpoint
+    this.#spOidcRpClientId = spOidcRpClientId
+    this.#spOidcRpRedirectUrl = spOidcRpRedirectUrl
+    this.#spOidcRpSecretJwks = spOidcRpSecretJwks
+    this._cache = new NodeCache(options)
+
+    // On expiry, refresh cache. If fail to refresh, log but do not throw error.
+    this._cache.on('expired', () =>
+      this.refresh().catch((err) =>
```
timotheeg: Just checking: `refresh()` no longer has a fail mode, so the `catch()` here and on line 90 will never run, is that correct?

src/app/modules/spcp/sp.oidc.client.cache.ts
```diff
@@ -0,0 +1,233 @@
...
+   * Method to retrieve NDI's public keys from cache
+   * @async
+   * @returns NDI's public keys
+   * @throws error if refresh fails
+   */
+  async getNdiPublicKeys(): Promise<CryptoKeys> {
+    const ndiPublicKeys = this._cache.get<CryptoKeys>(NDI_PUBLIC_KEY_NAME)
+    if (!ndiPublicKeys) {
+      const { ndiPublicKeys } = await this.refresh()
+      return ndiPublicKeys
```
timotheeg: Warning: since `refresh()` can be a never ending promise, the `await` here (and the equivalent `await refresh()` call in `getBaseClient()` below) can cause memory leaks. As in, each of the callers and the resources they hold can be kept in memory.

Refreshes should be initiated within a single process. In the accessors, if the data is not present, we can consider  raising an error, or do an independent call (that may fail) to just get the specific data.

timotheeg (commented):
Thanks for making all the updates ShuLi!

I think there is still a problem with the cache refresh mechanism. I've booked you for this afternoon to go over this synchronously. Hopefully I think it'll be faster to get a conclusion.

src/app/modules/spcp/sp.oidc.client.errors.ts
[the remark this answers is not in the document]
karrui: i think we can cross that bridge when we get to it haha.

src/app/modules/spcp/sp.oidc.client.cache.ts
tshuli: thanks! I've added a 10s timeout to prevent this from happening

src/app/modules/spcp/sp.oidc.client.cache.ts
```diff
@@ -0,0 +1,211 @@
...
+   * Method to retrieve base client from cache
+   * Triggers a refresh
+   * @async
+   * @returns Base client
+   * @throws error if refresh does not complete within 10s
+   */
+  async getBaseClient(): Promise<BaseClient> {
+    const baseClient = this._cache.get<BaseClient>(BASE_CLIENT_NAME)
+    if (!baseClient) {
+      const { baseClient } = await timeout(this.refresh(), 10000)
```
timotheeg: Let's have a constant for the `10000`

src/app/modules/spcp/sp.oidc.client.cache.ts
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * Method to make network call to retrieve public JWKS from NDI
+   * Max of 3 back-to-back attemps with timeout of 3s per attempt as per NDI specs
+   * @async
+   * @returns NDI's public keys
+   * @throws JwkError if keys are not the correct shape
+   */
+  async retrievePublicKeysFromNdi(): Promise<CryptoKeys> {
+    const getJwksWithRetries = retryPromiseThreeAttempts(
+      () => axios.get<PublicJwks>(this.#spOidcNdiJwksEndpoint),
```
timotheeg: Since the axios call is the actual interaction to the NDI service, the timeout specs should be specified on it (rather than on the `retryPromiseThreeAttempts()` wrapper). It will allow Axios to do the necessary cleanup.

Same comment for `retrieveBaseClientFromNdi()` below.

src/app/modules/spcp/sp.oidc.client.cache.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.#spOidcRpClientId = spOidcRpClientId
+    this.#spOidcRpRedirectUrl = spOidcRpRedirectUrl
+    this.#spOidcRpSecretJwks = spOidcRpSecretJwks
+    this._cache = new NodeCache(options)
+
+    // On expiry, refresh cache.
+    this._cache.on('expired', () => this.refresh())
+
+    // Trigger refresh on instantiation to populate cache.
+    void this.refresh()
```
timotheeg, line 85: Even though `refresh()` is a never-failing promise (infinite retry), continuous failures on startup are problematic because the Singpass functionality would be broken. Is it possible to have some custom logging for the startup failures anyway?

Hmm, or perhaps we do have enough information after all, since we will have the rejection logs, AND we should see independently the SingPass login errors, correct? 🤔

src/app/modules/spcp/sp.oidc.client.cache.ts
tshuli: added the 3s timeout for the axios call in for `retrievePublicKeysFromNdi`. However, in `retrieveBaseClientFromNdi`, `Issuer.discover()` uses  `fetch` in its underlying implementation, which doesn't provide a timeout. So I've kept the 3s timeout in `retryPromiseThreeAttempts()`

src/app/modules/spcp/sp.oidc.client.cache.ts
tshuli: Thanks! I think no harm to make additional logs to facilitate troubleshooting, I've included it

timotheeg: Thanks Shuli!

The current login implementation might be a little noisy (one or 2 per singpass login attempts, rather than one per refresh attempt), but I think that's OK for now 👍

timotheeg (approved):

PR #4020, opened 2022-06-21
feat: add question numbers to show in form builder tool

### LINKED ISSUE #3812: [React] Add question numbers to admin form builder
Previously added in #3692 but schema changes in #3625 removed it

justynoh:
## Problem
Question numbers are not shown in the admin form builder because the generated question numbers are not passed down to the component that generates the form field label. Also, the original question number generation function produced a `string|undefined`.

Closes #3812 

## Solution
The generated question numbers are passed to the component that generates the form field label. The question number generation function type has also been changed to `number|undefined`.

**Breaking Changes** 

## Screenshots

**AFTER**:
image
image

## Tests
- [X] App does not break when there are no questions. 
- [X] Question numbers should change when submittable fields are dragged around.
- [X] Question numbers should not change when non-submittable fields are added and/or dragged around.
- [X] Question numbers should change for user when toggling a field with show field logic.


justynoh:
(See also `frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx` line 318)

Possible improvement Q: The generation of question numbers is done way upstream in `BuilderFields.tsx` and produces a `number|undefined`. Seems like there is no type guarantee that `fieldType. isSubmitted` iff `questionNumber: number` (I might be wrong about this). Is it worth refactoring to get this guarantee?

karrui:
> Seems like there is no type guarantee that `fieldType.isSubmitted` iff `questionNumber: number` (I might be wrong about this). Is it worth refactoring to get this guarantee?

isSubmitted is only used in the constant to denote submittability, so probably no need, but you can try if you want to. The lower components should only care about whether there is a question number or not in their props

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -310,50 +310,63 @@ export const FieldRowContainer = ({
...
-      return <YesNoField schema={field} {...rest} />
-    case BasicField.Image:
-      return <ImageField schema={field} {...rest} />
-    case BasicField.Table:
-      return <TableField schema={field} {...rest} />
-  }
-})
+const MemoFieldRow = memo(
+  ({ field, questionNumber, ...rest }: MemoFieldRowProps) => {
+    // Q: The generation of question numbers is done way upstream and produces a
```
karrui, line 318: nit: should we remove this comment, since it may add confusion in the future?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -310,50 +310,63 @@ export const FieldRowContainer = ({
...
+      case BasicField.Section:
+        return <SectionFieldRow field={field} {...rest} />
+      case BasicField.Image:
+        return <ImageField schema={field} {...rest} />
+      case BasicField.Statement:
+        return <ParagraphField schema={field} {...rest} />
+      // User fields with question numbers
+      case BasicField.Attachment:
+        return (
+          <AttachmentField schema={{ ...field, questionNumber }} {...rest} />
```
karrui, line 334: instead of recreating a new object with questionNumber in this `MemoFieldRow` component, should we pass in the field that is already augmented with question numbers using `frontend/src/features/form/utils/augmentWithQuestionNo.ts`?

traced the creation of the fields all the way up to `frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/useBuilderFields.ts`, that could be a good place to augment them.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
justynoh: I'll rephrase the comment to be a note for possible improvement rather than a question!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
justynoh: o yes you're right! hmm, `frontend/src/features/form/utils/augmentWithQuestionNo.ts` and `frontend/src/features/admin-form/create/builder-and-design/utils/questionNumbers.ts` both seem to be doing the same thing. Will get rid of the latter to clean up the code. Thanks for pointing that out!

justynoh: Note: this also takes care of the earlier comment about a possible improvement.

karrui (approved):
lgtm

PR #4028, opened 2022-06-22
feat(form-v2): MyInfo prefill values

seaerchin:
## Problem
Current formSG FE displays prefilled values for PCK for myInfo fields. The react FE has no such ability. This PR adds prefilled values back into the displayed formfields.

## Solution
1. Add prefilled values into the constants for myinfo
2. Hook the individual fields up to the prefilled values. ~This is done through a typeguard (necessary as `MyInfoFormField` overlaps `FormFieldWithId`) + then reading the `fieldValue`. Didn't do a specialised variant of `MemoFieldRow` as the underlying `MobileField/<MyInfo>Field` would still need to read the `fieldValue` + result in alot of repeated code~
    - this is done through setting the default values in the initial `useForm` call. a few types/utility methods were defined to make clear the guarantees (ie, the field **must** be prefilled to have the default value set) and to smooth the differing value injection over (as mobile field injects default values differently from other fields)

### DIFF
--- a/frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
+++ b/frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
@@ -3,31 +3,22 @@ import { keyBy } from 'lodash'
 import { types as myInfoTypeArray } from '~shared/constants/field/myinfo'
 import { BasicField, FormFieldDto, MyInfoFormField } from '~shared/types/field'
 
-import { isMyInfo } from './isMyInfo'
-
 const MAP_ATTR_TO_NAME = keyBy(myInfoTypeArray, 'name')
 
 // Making a copy by destructuring so original object does not get affected.
 export const augmentWithMyInfo = ({
   ...field
 }: FormFieldDto): MyInfoFormField => {
-  if (!isMyInfo(field)) return field
-
-  const myInfoBlock = MAP_ATTR_TO_NAME[field.myInfo.attr]
-
+  // Only dropdown fields have augmented options for now.
   switch (field.fieldType) {
     case BasicField.Dropdown: {
-      return {
-        ...field,
-        fieldValue: myInfoBlock.prefilledValue,
-        fieldOptions: myInfoBlock.fieldOptions ?? [],
-      }
+      // No need to augment if no MyInfo attribute
+      if (!field.myInfo?.attr) return field
+      const myInfoBlock = MAP_ATTR_TO_NAME[field.myInfo.attr]
+      field.fieldOptions = myInfoBlock.fieldOptions ?? []
+      return field
     }
-    default: {
-      return {
-        ...field,
-        fieldValue: myInfoBlock.prefilledValue,
-      }
-    }
+    default:
+      return field
   }
 }
--- a/frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
+++ b/frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
@@ -3,22 +3,37 @@ import { keyBy } from 'lodash'
 import { types as myInfoTypeArray } from '~shared/constants/field/myinfo'
 import { BasicField, FormFieldDto, MyInfoFormField } from '~shared/types/field'
 
+import { isMyInfo } from './isMyInfo'
+
 const MAP_ATTR_TO_NAME = keyBy(myInfoTypeArray, 'name')
+
+export const isMyInfoFormField = (
+  field: FormFieldDto,
+): field is MyInfoFormField => {
+  return isMyInfo(field) && (field as MyInfoFormField).fieldValue !== undefined
+}
 
 // Making a copy by destructuring so original object does not get affected.
 export const augmentWithMyInfo = ({
   ...field
 }: FormFieldDto): MyInfoFormField => {
-  // Only dropdown fields have augmented options for now.
+  if (!isMyInfo(field)) return field
+
+  const myInfoBlock = MAP_ATTR_TO_NAME[field.myInfo.attr]
+
   switch (field.fieldType) {
     case BasicField.Dropdown: {
-      // No need to augment if no MyInfo attribute
-      if (!field.myInfo?.attr) return field
-      const myInfoBlock = MAP_ATTR_TO_NAME[field.myInfo.attr]
-      field.fieldOptions = myInfoBlock.fieldOptions ?? []
-      return field
+      return {
+        ...field,
+        fieldValue: myInfoBlock.prefilledValue,
+        fieldOptions: myInfoBlock.fieldOptions ?? [],
+      }
     }
-    default:
-      return field
+    default: {
+      return {
+        ...field,
+        fieldValue: myInfoBlock.prefilledValue,
+      }
+    }
   }
 }
--- a/frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
+++ b/frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
@@ -7,12 +7,6 @@ import { isMyInfo } from './isMyInfo'
 
 const MAP_ATTR_TO_NAME = keyBy(myInfoTypeArray, 'name')
 
-export const isMyInfoFormField = (
-  field: FormFieldDto,
-): field is MyInfoFormField => {
-  return isMyInfo(field) && (field as MyInfoFormField).fieldValue !== undefined
-}
-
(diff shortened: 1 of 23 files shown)


frontend/src/templates/Field/Date/DateField.tsx
```diff
@@ -61,6 +63,7 @@ export const DateField = ({
         control={control}
         name={schema._id}
         rules={validationRules}
+        defaultValue={isMyInfoFormField(schema) ? schema.fieldValue : ''}
```
karrui: i don't think we should inject `defaultValues` into the fields like this. This may result in having to update these fields whenever something else changes, especially if this change is solely for admin form builder (and fields are also used in public forms). 

Since all fields are wrapped by `react-hook-form`'s `FormProvider` by design, should we update the `defaultValue`s via its initial `useForm` call instead?

The form methods are passed down in [`FieldRowContainer.tsx`](https://github.com/opengovsg/FormSG/blob/form-v2/develop/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx#L99), and you may find out how public form fields augment their MyInfo values in [`FormFields.tsx`](https://github.com/opengovsg/FormSG/blob/form-v2/develop/frontend/src/features/public-form/components/FormFields/FormFields.tsx#L36)

frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
```diff
@@ -3,22 +3,31 @@ import { keyBy } from 'lodash'
...
   switch (field.fieldType) {
     case BasicField.Dropdown: {
-      // No need to augment if no MyInfo attribute
-      if (!field.myInfo?.attr) return field
-      const myInfoBlock = MAP_ATTR_TO_NAME[field.myInfo.attr]
-      field.fieldOptions = myInfoBlock.fieldOptions ?? []
-      return field
+      return {
+        ...field,
+        fieldValue: myInfoBlock.prefilledValue,
```
karrui: suggest not changing this function, since it may be confusing as to when `prefilledValue` may exist on public forms.

karrui (changes requested):

frontend/src/templates/Field/Date/DateField.tsx
seaerchin: good call - this seems way superior overall. will change to initializing the values in `FieldRowContainer` rather than hacking around

frontend/src/features/myinfo/utils/augmentWithMyInfo.ts
seaerchin: yep, i think changing to injecting the default values thru the initial `useForm` call would eliminate the need for this

frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -27,18 +35,18 @@ export const FormFields = ({
   colorTheme,
   onSubmit,
 }: FormFieldsProps): JSX.Element => {
-  // TODO: Inject default values if field is also prefilled.
   const augmentedFormFields = useMemo(
-    () => formFields.map(augmentWithMyInfo),
+    () => formFields.map(augmentWithMyInfo).map(augmentWithMyInfoPrefill),
```
karrui: do we need this here? public forms already have their fieldValue properly injected from the server. The TODO you removed is more for prefilled fields haha

frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -27,18 +35,18 @@ export const FormFields = ({
...
     [formFields],
   )
 
   const defaultFormValues = useMemo(() => {
     return augmentedFormFields.reduce<FormFieldValues>((acc, field) => {
-      if (field.fieldValue !== undefined) {
-        acc[field._id] = field.fieldValue
+      if (isMyInfoPrefilledFormField(field)) {
+        acc[field._id] = extractPrefilledValue(field)
```
karrui: i think this should be reverted, previous code was correctly setting fieldValue (if it exists from the data the server returned)

edit: scrolled down and saw what `isMyInfoPrefilledFormField` actually does. maybe rename it haha. Good bug fix on the `extractPrefilledValue` function!

shared/constants/field/myinfo/index.ts
```diff
@@ -17,6 +17,7 @@ export type MyInfoFieldBlock = {
   fieldType: MyInfoField['fieldType']
   fieldOptions?: string[]
   ValidationOptions?: Record<string, unknown>
+  prefilledValue: string
```
karrui: nit: should we add documentation to what prefilledValue means? Actually should we rename it to `previewValue` instead? make it more clear

shared/types/field/utils/myInfo.ts
```diff
@@ -0,0 +1,7 @@
+import { MyInfoPrefilledFormField, MyInfoFormField } from '../index'
+
+export const isMyInfoPrefilledFormField = (
```
karrui: ah, should we rename this function to `hasExistingFieldValue` instead? I know currently the prop is only used by myinfo but it confused me. can probably also put in a more generic place

shared/constants/field/myinfo/index.ts
seaerchin: actually that makes alot of sense - was thinking of it as `prefilled` as the values are supposed to be pre-filled but these will only be seen in the preview ._.

shared/types/field/utils/myInfo.ts
seaerchin: ~i could shift it up 1 level into `field/utils` - wdyt?~ it's already there oops; not sure where else to put this, i'll shift it into `src/features/myInof/utils` instead (also, renamed)

frontend/src/features/public-form/components/FormFields/FormFields.tsx
seaerchin: oh yea ps i forgot how the forms work - just to double check, we should just delete this cos the reducer below will inject the (actual) `prefilledValue` right?

frontend/src/features/myinfo/utils/augmentWithMyInfoPrefill.ts
```diff
@@ -0,0 +1,17 @@
+import { keyBy } from 'lodash'
+
+import { types as myInfoTypeArray } from '~shared/constants/field/myinfo'
+import { FormFieldDto, MyInfoFormField } from '~shared/types'
+
+import { isMyInfo } from './isMyInfo'
+
+const MAP_ATTR_TO_NAME = keyBy(myInfoTypeArray, 'name')
+
+export const augmentWithMyInfoPrefill = (
```
karrui: nit: still think this name might be misleading, what about augmentWithMyInfoPreview?

frontend/src/features/myinfo/utils/extractPreviewValue.ts
```diff
@@ -0,0 +1,18 @@
+import { BasicField, MyInfoPrefilledFormField } from '~shared/types'
+
+type PrefilledMyInfoValue = string | { value: string }
+
+export const extractPreviewValue = ({
```
karrui, line 5: nit: this one could or could not be a previewValue, i think this can continue to be called `extractPrefilledValue` haha

karrui (approved):
lgtm

frontend/src/features/myinfo/utils/augmentWithMyInfoPrefill.ts
seaerchin: can, will do!

frontend/src/features/myinfo/utils/extractPreviewValue.ts
seaerchin: sounds good!

seaerchin:
fyi @karrui, the story for all logic seems to be working fine on my local but there's a delay before the values appear. this could be the reason why the chromatic story is showing the empty state.

is there a workaround/can you verify this behaviour also just to check that i've not broken any stories and if good, i'll merge tmrw

karrui:
> is there a workaround/can you verify this behaviour also just to check that i've not broken any stories and if good, i'll merge tmrw

ah... try adding a [chromatic delay](https://www.chromatic.com/docs/delay) to the story maybe!

```tsx
parameters: {
    layout: 'fullscreen',
    // Prevent flaky tests due to modal animating in.
    chromatic: { delay: 200 },
  },
```

Issue #4027, opened 2022-06-22
[React] bug: form builder field option inputs lacking validation


justynoh:
**Describe the bug**

For short and long answer fields in the form builder, 
1. Exact matching does not work (always reports an error)
2. "Number of characters allowed" field allows decimals, nonpositive numbers and numbers in "e" notation.

**Expected behavior**

1. Exact matching should not error when the number of characters entered in the field is equal to the allowed number.
2. "Number of characters allowed" box should only allow positive integers (Q: is there any reason we are allowing "e" notation? especially if the max number allowed in the box is 10,000 which is at most a 5 digit number and is not big enough to warrant allowing "e" notation. to me we should only allow that box to contain characters '0' through '9')

**To Reproduce**

See screenshots below

**Screenshots**

Point 1
<img width="1161" alt="image" src="https://user-images.githubusercontent.com/25571626/174967262-f2a025a2-5d34-4ef8-bd8f-0008c214f5f9.png">

Point 2
<img width="1161" alt="image" src="https://user-images.githubusercontent.com/25571626/174967760-67bd59d0-9b20-417d-a332-bdcacae72120.png">

justynoh:
resolved

PR #4030, opened 2022-06-23
fix(form-v2): perform input validation for field validation options in text, number, table and checkbox fields

justynoh:
## Problem

For short and long answer fields, number field and table (max and min row) in the form builder,

1. Exact matching does not work (always reports an error)
2. "Number of characters allowed" field allows decimals, nonpositive numbers and numbers in "e" notation.

Root cause of point 1 is that the value being passed up from the form builder is a `string` whereas the form field validation checks expect a `number` for this value. For point 2, we just need to add input validation checks.

Closes issue #4027 

## Solution

Added input validation checks to ensure only numerals [0-9] can be input into those fields. Also modified `onChange` to parse the user input from a `string` to a `number` prior to assigning it to the `schema` property.

**Breaking Changes** 

## Tests

For **Edit [Short answer, Long answer, Number, Checkbox]** > **Number of characters allowed**:

- [X] Unable to enter characters `e`, `-` and `.`
- [X] Appropriate error message appears when "Exact _x_" and field does not contain _x_ characters
- [X] No error message appears when "Exact _x_" and field contains _x_ characters

and for **Edit Table** > **[Minimum rows, Maximum rows allowed]**:

- [X] Unable to enter characters `e`, `-` and `.`
- [X] Unable to delete rows in form builder when respondent is allowed to add rows and the number of rows is already equal to minimum rows
- [X] Unable to add more rows when respondent is allowed to add rows and the number of rows is already equal to maximum rows

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
@@ -247,10 +247,8 @@ export const EditCheckbox = ({ field }: EditCheckboxProps): JSX.Element => {
                   precision={0}
                   flex={1}
                   showSteppers={false}
-                  onChange={(val) => {
-                    // Only allow numeric inputs and return it as a number
-                    const numericValue = parseInt(val.replace(/\D/g, ''))
-                    onChange(isNaN(numericValue) ? val : numericValue)
+                  onChange={(_, valNum) => {
+                    onChange(isNaN(valNum) ? '' : valNum)
                   }}
                   {...rest}
                   placeholder="Minimum"
@@ -267,10 +265,8 @@ export const EditCheckbox = ({ field }: EditCheckboxProps): JSX.Element => {
                   precision={0}
                   flex={1}
                   showSteppers={false}
-                  onChange={(val) => {
-                    // Only allow numeric inputs and return it as a number
-                    const numericValue = parseInt(val.replace(/\D/g, ''))
-                    onChange(isNaN(numericValue) ? val : numericValue)
+                  onChange={(_, valNum) => {
+                    onChange(isNaN(valNum) ? '' : valNum)
                   }}
                   {...rest}
                   placeholder="Maximum"
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
@@ -13,6 +13,8 @@ import NumberInput from '~components/NumberInput'
 import Textarea from '~components/Textarea'
 import Toggle from '~components/Toggle'
 
+import { validateNumberInput } from '../../../utils/validateNumberInput'
+
 import {
   SPLIT_TEXTAREA_TRANSFORM,
   SPLIT_TEXTAREA_VALIDATION,
@@ -247,8 +249,8 @@ export const EditCheckbox = ({ field }: EditCheckboxProps): JSX.Element => {
                   precision={0}
                   flex={1}
                   showSteppers={false}
-                  onChange={(_, valNum) => {
-                    onChange(isNaN(valNum) ? '' : valNum)
+                  onChange={(valStr, valNum) => {
+                    validateNumberInput(onChange, valStr, valNum)
                   }}
                   {...rest}
                   placeholder="Minimum"
@@ -265,8 +267,8 @@ export const EditCheckbox = ({ field }: EditCheckboxProps): JSX.Element => {
                   precision={0}
                   flex={1}
                   showSteppers={false}
-                  onChange={(_, valNum) => {
-                    onChange(isNaN(valNum) ? '' : valNum)
+                  onChange={(valStr, valNum) => {
+                    validateNumberInput(onChange, valStr, valNum)
                   }}
                   {...rest}
                   placeholder="Maximum"
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
@@ -249,9 +249,7 @@ export const EditCheckbox = ({ field }: EditCheckboxProps): JSX.Element => {
                   precision={0}
                   flex={1}
                   showSteppers={false}
-                  onChange={(valStr, valNum) => {
-                    validateNumberInput(onChange, valStr, valNum)
-                  }}
+                  onChange={validateNumberInput(onChange)}
                   {...rest}
                   placeholder="Minimum"
                 />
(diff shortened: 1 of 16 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditLongText/EditLongText.tsx
```diff
@@ -168,13 +168,19 @@ export const EditLongText = ({ field }: EditLongTextProps): JSX.Element => {
...
                 flex={1}
+                inputMode="numeric"
                 showSteppers={false}
-                {...field}
                 placeholder="Number of characters"
                 isDisabled={!watchedSelectedValidation}
+                onChange={(val) => {
+                  // Only allow numeric inputs and return it as a number
+                  const numericValue = parseInt(val.replace(/\D/g, ''))
+                  onChange(isNaN(numericValue) ? 0 : numericValue)
```
karrui: should we just not return if the parsed `numericValue` is `NaN`? this way the input will not suddenly change to 0 when non numbers are entered.

karrui (commented):
do we also need to update other fields such as `EditCheckbox`, etc?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditLongText/EditLongText.tsx
justynoh: Non numbers can't be entered anymore, the only allowed characters in the field is 0 to 9. The NaN case only occurs when the input is empty (so you get `parseInt('')` which is where the NaN comes from), which naturally corresponds to 0.

EDIT: WAIT i misunderstood give me a min

justynoh: u right, but I think in `frontend/src/utils/fieldValidation.ts`, `customVal` is declared to be of type `number|null`. To be sure, I'll change the type declaration as well so that we deal with that case too

justynoh:
added for checkbox, I don't think I saw any other field that requires numeric validation (correct me if i'm wrong pls!)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
```diff
@@ -248,8 +248,9 @@ export const EditCheckbox = ({ field }: EditCheckboxProps): JSX.Element => {
                   flex={1}
                   showSteppers={false}
                   onChange={(val) => {
-                    // Only allow numeric inputs
-                    onChange(val.replace(/\D/g, ''))
+                    // Only allow numeric inputs and return it as a number
+                    const numericValue = parseInt(val.replace(/\D/g, ''))
+                    onChange(isNaN(numericValue) ? val : numericValue)
```
karrui: now that i'm seeing it, should we inline this into `NumberInput` since it's always repeated (for integer-only inputs)? 

We can only enable this if the inputType is numeric, or hide this inside a prop flag. Feel free to hit me up on Slack if you need clarifications

justynoh: discussed on slack

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditCheckbox.tsx
```diff
@@ -247,10 +247,8 @@ export const EditCheckbox = ({ field }: EditCheckboxProps): JSX.Element => {
                   precision={0}
                   flex={1}
                   showSteppers={false}
-                  onChange={(val) => {
-                    // Only allow numeric inputs and return it as a number
-                    const numericValue = parseInt(val.replace(/\D/g, ''))
-                    onChange(isNaN(numericValue) ? val : numericValue)
+                  onChange={(_, valNum) => {
+                    onChange(isNaN(valNum) ? '' : valNum)
```
karrui: i think we should still not trigger onChange if value is not valid:

```tsx
 onChange={(_, valNum) => {
  if (!isNaN(valNum)) return
  return onChange(valNum)
```

justynoh: currently it's still only possible to enter digits 0-9, so the only way it will be NaN is if the input value is `''`. is there any particular reason to prefer either?

karrui: we won't suddenly clear the input if an invalid character is entered, and instead just do nothing.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
```diff
@@ -175,13 +175,19 @@ export const EditNumber = ({ field }: EditNumberProps): JSX.Element => {
...
+                inputMode="numeric"
                 showSteppers={false}
-                {...field}
                 placeholder="Number of characters"
                 isDisabled={!watchedSelectedValidation}
+                onChange={(val) => {
+                  // Only allow numeric inputs and return it as a number
+                  const numericValue = parseInt(val.replace(/\D/g, ''))
+                  onChange(isNaN(numericValue) ? val : numericValue)
+                }}
```
karrui: missed one here :P

frontend/src/features/admin-form/create/builder-and-design/utils/validateNumberInput.ts
```diff
@@ -0,0 +1,22 @@
...
+  valStr: string,
+  valNum: number,
+): void => {
+  if (!valStr) {
+    onChange('')
+    return
+  }
+  if (isNaN(valNum)) return
+  onChange(valNum)
+}
```
karrui: supernit: simplify

```ts
if (!valStr) {
  return onChange('')
}

if (!isNaN(valNum)) {
  return onChange(valNum)
}
```

karrui (commented):
missed one, lgtm otherwise

frontend/src/features/admin-form/create/builder-and-design/utils/validateNumberInput.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Validates the input given to Chakra's NumberInput component. Pass this to the
+ * onChange parameter for NumberInput. Calls onChange('') if the input is empty,
+ * onChange(number) if the input is a number, and does nothing if the input is
+ * invalid.
+ * @param onChange react Component onChange
+ * @param valStr input value as string
+ * @param valNum input value as number (can be NaN!)
+ * @returns void
+ */
+export const validateNumberInput = (
```
karrui: nit: seeing this again, should we make this a higher order function so it can be directly instantiated?

```tsx
export const validateNumberInput = (onChange) => (valStr, valNum) => {
  ...
}

// usage
render={({ field: { onChange, ...rest } }) => (
                <NumberInput
                  flex={1}
                  onChange={validateNumberInput(onChange)}
                  {...rest}
                />
              )}

```

karrui (approved):
lgtm

frontend/src/features/admin-form/create/builder-and-design/utils/validateNumberInput.ts
justynoh: ok!

Issue #4014, opened 2022-06-20
[React] Dragging new field before saving triggers error


justynoh:
**Describe the bug**
A form admin is able to drag a newly-created field in the form builder to reorder it prior to saving the field. Attempting this, the form admin will see that the field returns back to its original position and an error message appears "Cannot read properties of undefined (reading '_id')".

**To Reproduce**
1. Go to form builder and create a couple fields
2. Drag any new field from the sidebar into the form
3. Before clicking "Create field" on the sidebar, drag the field to reorder
4. Experience the error!

**Expected behavior**
Either 1. Don't give users the option to reorder prior to saving, or 2. they should be able to drag and drop.

**Screenshots**
<img width="1490" alt="image" src="https://user-images.githubusercontent.com/25571626/174569175-9a3272e1-2f78-4edd-b823-627a5b6674d7.png">

justynoh:
resolved

PR #4032, opened 2022-06-23
fix(form-v2): disallow dragging fields before creation

justynoh:
## Problem
See issue #4014 

## Solution
Add a check to disallow dragging if the field is still in the process of being created.

**Breaking Changes** 

## Tests
- [X] Fields are not draggable before they are created
- [X] Fields are still draggable in edit mode


karrui (approved):
lgtm

should prevent any fields from being dragged as long as any field is being edited too (has dirty state)?

can do in another PR.

PR #4041, opened 2022-06-23
feat(form-workspaces/1): add form workspaces side menu

### LINKED ISSUE #4034: [React/Form Workspaces] Add workspaces side menu desktop view
Add desktop view for form workspaces side menu

### LINKED ISSUE #4035: [React/Form Workspaces] Add workspaces side menu mobile view
Add mobile view for form workspaces side menu

hanstirtaputra:
## Problem
We need a side menu for form workspaces to categorise forms

Closes #4034
Closes #4035 

## Solution
Add side menu for form workspaces in both desktop and mobile views. It currently uses mock data. 

**Known Bugs**:
- Since we split the views for mobile and desktop, and `useState` for the `currWorkspaceId` in each of the views, when we change from desktop to mobile view, the `currWorkspaceId` is reset. 
    - This can be fixed by using `zustand` state or passing state from parent `WorkspacePage`, but for this PR I want to focus on the views first

**Breaking Changes** 

## Before & After Screenshots
Storybook for `WorkspacePage` will show the new view


frontend/src/features/workspace/components/WorkspaceSideMenu/WorkspaceMenuTabs.tsx
```diff
@@ -0,0 +1,77 @@
...
+      _hover={{
+        textColor: 'var(--chakra-colors-primary-500)',
+      }}
+      onClick={onClick}
+      aria-label={`${label} workspace tab`}
+      {...styles}
+      {...props}
+    >
+      <Text textStyle="body-2" whiteSpace="nowrap">
+        {truncateLongTextWithEllipsis(label)}
```
timotheeg: Couldn't this be done with a [text-overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) css rule instead of custom code?

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -122,63 +138,19 @@ export const WorkspacePage = (): JSX.Element => {
     [isUserLoading, hasSeenAnnouncement],
   )
 
+  // TODO (hans): Get totalFormCount and isLoading from GET workspaces API when it is implemented
+  const totalFormCount = Math.round(Math.random())
```
timotheeg: /nit even for a mock, get a bigger range of number, maybe? the current code only gives 0 or 1 😅 (or is the 0/1 on purpose?)

```suggestion
  const totalFormCount = Math.round(Math.random() * 1000)
```

frontend/src/features/workspace/WorkspaceContent.tsx
```diff
@@ -0,0 +1,158 @@
...
+const useWorkspaceForms = () => {
+  const { data: dashboardForms, isLoading } = useWorkspace()
+  const [searchParams, setSearchParams] = useSearchParams()
+  const [sortedForms, setSortedForms] = useState(dashboardForms)
+  const [isManipulating, setIsManipulating] = useState(false)
+
+  const createFormModalDisclosure = useDisclosure()
+
+  const currentPage = Number(
+    searchParams.get('page') ?? PAGE_DEFAULTS.pageNumber,
```
timotheeg: I've seen getting page numbers from a few PRs recently, and I am a little uncomfortable grabbing query string values without validating them 🤔 . The code can cause currentPage to be `NaN`.

Can we validate that the key contains a number before attempting to convert it to a number (same for `size` below).

Perhaps we need a thin wrapper utility to `useSearchParams()`, where we can specify the data type we want and the default value.

Just for illustration, we could have something like this here:
```javascript
const VALID_NUM_ARG_RE = /^[1-9]\d*$/

const pageParam = searchParams.get('page')
const sizeParam = searchParams.get('size')

const currentPage = VALID_NUM_ARG_RE.test(pageParam)
    ? parseInt(pageParam)
    : PAGE_DEFAULTS.pageNumber
const pageSize = VALID_NUM_ARG_RE.test(sizeParam)
    ? parseInt(sizeParam)
    : PAGE_DEFAULTS.size
```

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -1,113 +1,129 @@
...
+    _id: '3',
+    title: 'Public sentiment',
+    numForms: 12,
+  },
+  {
+    _id: '4',
+    title: 'Very long number of forms',
+    numForms: 531159214021,
+  },
+]
```
karrui: we should mock workspace responses with `msw` so code does not get polluted with mock data. if needed, for local testing we can disable all other endpoints and only mock a specific workspaces endpoint so all other api calls still work

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -1,113 +1,129 @@
...
+
+const WorkspaceMobilePage = (): JSX.Element => {
```
karrui: looking at the figma, i don't think we need a separate page for workspaces. I am wary of maintaining two separate components for the same data -- it may lead to maintainability issues.

can we try to minimise area of which mobile components are used? i think only the drawer needs a separate component, and even then should take in the same data.

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -1,113 +1,129 @@
...
+              workspaces={MOCK_WORKSPACES_DATA}
+              currWorkspace={currWorkspaceId}
+              onClick={(id) => {
+                setCurrWorkspaceId(id)
+                onClose()
+              }}
+            />
+          </DrawerBody>
+        </DrawerContent>
+      </Drawer>
```
karrui: as above, i think only the drawer needs to be its own mobile component. the rest of the workspace page should be able to be styled responsively.

can hmu on slack if you need guidance.

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -122,63 +138,19 @@ export const WorkspacePage = (): JSX.Element => {
     [isUserLoading, hasSeenAnnouncement],
   )
 
+  // TODO (hans): Get totalFormCount and isLoading from GET workspaces API when it is implemented
+  const totalFormCount = Math.round(Math.random())
+  const isLoading = false
+  const isMobile = useIsMobile()
+
+  // TODO (hans): Add <EmptyWorkspace/> if totalFormCount === 0
```
karrui: use msw to mock api requests instead of hardcoding.

frontend/src/features/workspace/components/WorkspaceSideMenu/WorkspaceMenuTabs.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <Flex
+      as="button"
+      justifyContent="space-between"
+      w="100%"
+      pl="2rem"
+      pr="1.5rem"
+      h="3.5rem"
+      alignItems="center"
+      _hover={{
+        textColor: 'var(--chakra-colors-primary-500)',
```
karrui: nit: can just use `"primary.500"`, since this is a chakra component.

frontend/src/features/workspace/components/WorkspaceSideMenu/WorkspaceMenuTabs.tsx
```diff
@@ -0,0 +1,77 @@
...
+  numForms,
+  isSelected,
+  onClick,
+  ...props
+}: WorkspaceTabProps): JSX.Element => {
+  const styles = isSelected
+    ? {
+        borderLeft: '2px',
+        borderLeftColor: 'var(--chakra-colors-primary-500)',
+        textColor: 'var(--chakra-colors-primary-500)',
```
karrui: same here, can use the colors directly instead of --var

frontend/src/features/workspace/components/WorkspaceSideMenu/WorkspaceMenuTabs.tsx
karrui: we don't need this util function, check out Text's [`noOfLines`](https://chakra-ui.com/docs/components/text#truncate-text) prop.

frontend/src/features/workspace/WorkspacePage.tsx
hanstirtaputra: Okay this is better, thanks. Will update to mock using `msw`

frontend/src/features/workspace/WorkspacePage.tsx
hanstirtaputra: Okay, I'll update the views to combine mobile and desktop, and only use drawer as separate component

frontend/src/features/workspace/WorkspacePage.tsx
hanstirtaputra: Okay thanks, I've updated to use Chakra's `Grid` to do that

frontend/src/features/workspace/components/WorkspaceSideMenu/WorkspaceMenuTabs.tsx
hanstirtaputra: Yes... I forgot that existed. Anyway I've updated this to use Chakra `Text`'s `isTruncated`

frontend/src/features/workspace/WorkspacePage.tsx
hanstirtaputra: Updated this to use `msw` to mock, this is not required anymore!

frontend/src/features/workspace/WorkspaceContent.tsx
hanstirtaputra: Sounds good. I've updated the PR to validate the query string values first. Anyway this page and size param for pagination might not be used in workspaces. We might not even have pagination for workspaces for now

karrui:
i'll wait to review this PR until @syan-syan's admin header PR #4129 is merged in, suspect the addition of admin header will change the layout slightly.

karrui (approved):
lgtm

PR #4042, opened 2022-06-23
fix(form-v2): automatically return to field selection after creating/saving field

### LINKED ISSUE #4018: [React] Field builder does not close after saving
image

It is intended for it to close so that user can continue to build other fields without having to press back or close it manually.

justynoh:
## Problem

Closes #4018 

## Solution
Set the build state to be inactive once saved.

**Breaking Changes** 

## Tests
- [X] Return to field drawer after creating a completely new field
- [X] Return to field drawer after saving a change to an existing field


karrui (approved):
lgtm

PR #4043, opened 2022-06-23
feat(storage-responses/3): allow searching of individual using full submission id

### LINKED ISSUE #3128: [React/StorageMode] View individual storage mode responses

karrui:
Builds on #3001

## Problem

This PR adds the ability for form admins to search for an individual form response with its response id.

Closes #3128 

## Solution

**Features**:

- feat: update usePageSearchParams to also return subId search state
- feat: allow searching by id and navigation from searched response
- feat: allow reset of searched submission id

**Improvements**:

- feat: use fetching state instead of loading state to discern loading state of components

## Before & After Screenshots
Unlocked story should show new search bar.

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -67,7 +67,7 @@ export const IndividualResponsePage = (): JSX.Element => {
     getPreviousSubmissionId,
     onNavNextSubmissionId,
     onNavPreviousSubmissionId,
-    isAnyLoading,
+    isAnyFetching,
   } = useUnlockedResponses()
   const { data, isLoading } = useIndividualSubmission()
 
@@ -152,13 +152,13 @@ export const IndividualResponsePage = (): JSX.Element => {
           </Skeleton>
           <ButtonGroup>
             <IconButton
-              isDisabled={!prevSubmissionId || isAnyLoading}
+              isDisabled={!prevSubmissionId || isAnyFetching}
               onClick={handleNavigatePrev}
               icon={<BiChevronLeft />}
               aria-label="Previous submission"
             />
             <IconButton
-              isDisabled={!nextSubmissionId || isAnyLoading}
+              isDisabled={!nextSubmissionId || isAnyFetching}
               onClick={handleNavigateNext}
               icon={<BiChevronRight />}
               aria-label="Next submission"
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -63,6 +63,7 @@ export const IndividualResponsePage = (): JSX.Element => {
   const { secretKey } = useStorageResponsesContext()
   const {
     lastNavPage,
+    lastNavSubmissionId,
     getNextSubmissionId,
     getPreviousSubmissionId,
     onNavNextSubmissionId,
@@ -117,11 +118,19 @@ export const IndividualResponsePage = (): JSX.Element => {
   ])
 
   const backLink = useMemo(() => {
-    if (lastNavPage) {
-      return `..?page=${lastNavPage}`
+    const searchParams = new URLSearchParams()
+    if (!lastNavPage && !lastNavSubmissionId) {
+      return `..`
     }
-    return `..`
-  }, [lastNavPage])
+
+    if (lastNavPage) {
+      searchParams.append('page', lastNavPage.toString())
+    }
+    if (lastNavSubmissionId) {
+      searchParams.append('submissionId', lastNavSubmissionId)
+    }
+    return `..?${searchParams}`
+  }, [lastNavPage, lastNavSubmissionId])
 
   if (!secretKey) {
     return <SecretKeyVerification />
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -118,16 +118,16 @@ export const IndividualResponsePage = (): JSX.Element => {
   ])
 
   const backLink = useMemo(() => {
-    const searchParams = new URLSearchParams()
     if (!lastNavPage && !lastNavSubmissionId) {
       return `..`
     }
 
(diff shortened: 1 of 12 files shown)


frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -117,11 +118,19 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+    if (!lastNavPage && !lastNavSubmissionId) {
+      return `..`
+    }
+
     if (lastNavPage) {
-      return `..?page=${lastNavPage}`
+      searchParams.append('page', lastNavPage.toString())
+    }
+    if (lastNavSubmissionId) {
+      searchParams.append('submissionId', lastNavSubmissionId)
```
timotheeg: /nit Use `set()` rather than append, to show this has nothing to do with multi-value args.

```suggestion
      searchParams.set('page', lastNavPage.toString())
    }
    if (lastNavSubmissionId) {
      searchParams.set('submissionId', lastNavSubmissionId)
```

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/UnlockedResponses.tsx
```diff
@@ -1,21 +1,45 @@
...
+  }, [filteredCount, count, submissionId])
 
   const prettifiedResponsesCount = useMemo(() => {
-    if (!count) return
-    return simplur` ${[count]}response[|s] to date`
-  }, [count])
+    if (filteredCount !== undefined) {
+      return simplur` ${[filteredCount]}result[|s] found`
+    } else if (count !== undefined) {
+      return simplur` ${[count]}response[|s] to date`
```
timotheeg, line 36: Just checking: is the spacing correct?
```suggestion
      return simplur`${[filteredCount]} result[|s] found`
    } else if (count !== undefined) {
      return simplur`${[count]} response[|s] to date`
```

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/hooks/usePageSearchParams.ts
```diff
@@ -17,12 +19,35 @@ export const usePageSearchParams = () => {
       if (page < 0) {
         page = 1
       }
-      setParams({
-        [PAGE_KEY]: String(page),
-      })
+      params.set(PAGE_KEY, page.toString())
+      setParams(params)
```
timotheeg, line 23: Why do we no longer clear the whole query string and only leave `PAGE_KEY`, like was done previously?

timotheeg (commented):
Looking good, just some minor questions from me.

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
karrui: this does have something to do with multivalue args (but also not really since one response search can only have one result, but does future proof if we ever allow partial id searching in the future)

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/UnlockedResponses.tsx
karrui: it is, there is a number before this string haha

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/hooks/usePageSearchParams.ts
karrui: in case there are other values in the query, just in case they get cleared accidentally. This hook does not actually care about the other params, just that the page param is set properly.

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
timotheeg: Sorry, what does partial id searching mean in this context?

`.append()` is used if you want to have 2 or more `pages` key  in the query string, or 2 or more `submissionId` in the query string. I assume we will not have that, and so `.set()` should be used instead.

Of course functionally it doesn't matter here, since the code starts with a fresh `URLSearchParams` anyway, but semantically it's clearer to use `.set()` to indicate intent.

```javascript
p = new URLSearchParams()
p.set('foo', 1)
p.set('bar', 2)

p.toString() // foo=1&bar=2

p.append('foo', 3)

p.toString() // foo=1&bar=2&foo=3

p.set('foo', 4)

p.toString() // foo=4&bar=2
```

Personally, I'm really not fan of multi-value keys in query string btw (e.g. `foo=1&foo=2`), I much prefer having a singular list key that includes multiple entries with some encode/parse convention, like `fooList=1,2`

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/UnlockedResponses.tsx
timotheeg: Oh duh! Sorry, I thought `simplur` was including the number in the output as well. Thanks for the clarification.

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/hooks/usePageSearchParams.ts
timotheeg: Got it. Thanks!

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
karrui: fuzzy search, but i'll just reset it as you suggested. We can reimplement multi-arg in the future if needed.

karrui: ah i got you. yes, we should use `set` instead of `append`. that was a bug!!!

karrui: fixed in f3e05e5

timotheeg (approved):
lgtm 👍

PR #4062, opened 2022-06-29
feat: add new feature update notification logic

### LINKED ISSUE #3988: [React/What's New] Add logic to notify form admins when there is a new feature update
Add logic to handle when to notify form admins when there is a new feature update.

### LINKED ISSUE #3989: [React/What's New] Add logic to update form admins last seen feature update information
When form admins view the list of feature updates, it should update the form admins last seen feature update information in the backend.

siddarth2824:
## Problem

There is currently no logic to check whether or not form admin users should be notified on a new feature update and when to update form admin user's last seen feature update date which is used in the notification logic.

Closes #3988
Closes #3989

## Solution

As the current header in the Figma design has not been implemented, a solid button signifies that there is a new feature update that the form admin user has not seen while a outlined button signifies that the form admin user is up to date with all the latest feature updates.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/whats-new/FeatureUpdateList.ts
+++ b/frontend/src/features/whats-new/FeatureUpdateList.ts
@@ -7,17 +7,17 @@ export interface FeatureUpdateImage {
 }
 export interface FeatureUpdate {
   id: number
-  date: string
+  date: Date
   title: string
   description: string
   image?: FeatureUpdateImage
 }
 
 // TODO: Confirm the actual features which will be showcased in the What's New section
-export const FeatureUpdateList: FeatureUpdate[] = [
+export const FEATURE_UPDATE_LIST: FeatureUpdate[] = [
   {
     id: 1,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'UI Improvements',
     description: `
   * Cool new datepicker with day, month, and year views
@@ -27,7 +27,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 2,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'No more hangups',
     description:
       'With the new Phone Number field, you can quickly collect people’s digits. Local and international validation available.',
@@ -38,7 +38,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 3,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'Introducing Workspaces',
     description:
       "A workspace that's never messy. The first thing you see when you log into FormSG is your Workspace. This is where you store and organize all of your forms. Create multiple Workspaces and move your forms into them to stay organised.",
@@ -49,7 +49,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 4,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'UI Improvements',
     description: `
 * Cool new datepicker with day, month, and year views
@@ -59,7 +59,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 5,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'No more hangups',
     description:
       'With the new Phone Number field, you can quickly collect people’s digits. Local and international validation available.',
@@ -70,7 +70,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 6,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'Introducing Workspaces',
     description:
       "A workspace that's never messy. The first thing you see when you log into FormSG is your Workspace. This is where you store and organize all of your forms. Create multiple Workspaces and move your forms into them to stay organised.",
@@ -81,7 +81,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 7,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'UI Improvements',
     description: `
 * Cool new datepicker with day, month, and year views
@@ -91,7 +91,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 8,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'No more hangups',
     description:
       'With the new Phone Number field, you can quickly collect people’s digits. Local and international validation available.',
@@ -102,7 +102,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 9,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'Introducing Workspaces',
     description:
       "A workspace that's never messy. The first thing you see when you log into FormSG is your Workspace. This is where you store and organize all of your forms. Create multiple Workspaces and move your forms into them to stay organised.",
@@ -113,7 +113,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 10,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'UI Improvements',
     description: `
 * Cool new datepicker with day, month, and year views
@@ -123,7 +123,7 @@ export const FeatureUpdateList: FeatureUpdate[] = [
   },
   {
     id: 11,
-    date: '17 June 2022',
+    date: new Date('17 June 2022 GMT+8'),
     title: 'No more hangups',
     description:
       'With the new Phone Number field, you can quickly collect people’s digits. Local and international validation available.',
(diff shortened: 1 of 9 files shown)


frontend/src/features/whats-new/FeatureUpdateList.ts
```diff
@@ -16,7 +16,7 @@ export interface FeatureUpdate {
 export const FeatureUpdateList: FeatureUpdate[] = [
```
karrui: not this PR, but we'll probably need to update this before launch. can you make an issue and add a TODO here?

frontend/src/features/whats-new/utils/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  user,
+  isLoading,
+}: shouldShowLatestFeatureUpdateNotificationProps): boolean => {
+  if (isLoading) {
+    return false
+  }
+  const sortedFeatureUpdateList = FeatureUpdateList.sort(
+    (featureUpdateA, featureUpdateB) =>
+      Date.parse(featureUpdateB.date) - Date.parse(featureUpdateA.date),
+  )
```
karrui: suspect we don't need to sort. We should always just check the first feature list's date and it should be fine.

frontend/src/features/whats-new/utils/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  const lastSeenUserFeatureUpdateDate = new Date(
+    user?.flags?.lastSeenFeatureUpdateDate ??
+      sortedFeatureUpdateList[sortedFeatureUpdateList.length - 1].date,
+  )
```
karrui: do we need a default? if there is no last seen feature date i think we can return `true` immediately.

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -108,6 +111,24 @@ export const WorkspacePage = (): JSX.Element => {
...
+    useState(false)
+  const shouldShowFeatureUpdateNotification =
+    shouldShowLatestFeatureUpdateNotification({
+      user,
+      isLoading: isUserLoading,
+    })
+
+  useEffect(() => {
+    setShouldShowWhatsNewNotification(shouldShowFeatureUpdateNotification)
+  }, [shouldShowFeatureUpdateNotification])
```
karrui: feels a little weird to combine an util fn, effect, and use state. Can't we just memoize this?
also feels a little weird to pass in loading states to the utility function.

suggest just memoizing:
```tsx
const shouldShowFeatureUpdateNotification = useMemo(() => {
  if (isUserLoading || !user) return false
  return getShowLatestFeatureUpdateNotification(user)
}, [user, isUserLoading, getShowLatestFeatureUpdateNotification])
```

frontend/src/features/whats-new/utils/utils.ts
```diff
@@ -0,0 +1,31 @@
+import { isAfter } from 'date-fns'
+
+import { UserDto } from '~shared/types/user'
+
+import { FeatureUpdateList } from '../FeatureUpdateList'
+
+export interface shouldShowLatestFeatureUpdateNotificationProps {
```
karrui: nit: name of function sounds like a variable instead of a function.
suggest `getShowLatestFeatureUpdateNotificationState`?

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -108,6 +111,24 @@ export const WorkspacePage = (): JSX.Element => {
...
+
+  const onWhatsNewDrawerOpen = () => {
```
karrui: nit: memoize with `useCallback`

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -67,7 +71,11 @@ export const WorkspaceHeader = ({
         >
           Create form
         </Button>
-        <Button isFullWidth={isMobile} onClick={handleOpenWhatsNewDrawer}>
+        <Button
+          isFullWidth={isMobile}
+          onClick={handleOpenWhatsNewDrawer}
+          variant={whatsNewButtonVariant}
+        >
```
karrui, line 80: question: this is temporary right (until workspace header is done)? any reason why we need the `variant` prop?

siddarth2824: Yes this is temporary because the admin header is not up yet for me to add the notification icon. Hence I used the button variant to represent whether or not there is a new feature update that the form admin user has not seen yet.

karrui: gotcha. maybe need a TODO to signify that it is temporary

frontend/src/features/whats-new/utils/utils.ts
siddarth2824: I added this function just in case someone updated the list wrongly. Hence by sorting, we will always get the latest date at the top of the list.

frontend/src/features/whats-new/FeatureUpdateList.ts
siddarth2824: Can I clarify what needs to be updated here?

karrui: the dates, and the actual features we want to showcase. i suspect the figma text are placeholders. maybe sync with PMs/Kenneth

frontend/src/features/whats-new/utils/utils.ts
siddarth2824: Yeap you are right. I thought it would be neater this way but just returning true is much neater.

frontend/src/features/whats-new/FeatureUpdateList.ts
siddarth2824: Ahhh I see. I will clarify with Pearly once she is back.

frontend/src/features/whats-new/utils/utils.ts
```diff
@@ -4,28 +4,22 @@ import { UserDto } from '~shared/types/user'
...
+  user: UserDto | undefined,
+): boolean => {
+  if (user?.flags?.lastSeenFeatureUpdateDate) {
+    const sortedFeatureUpdateList = FeatureUpdateList.sort(
+      (featureUpdateA, featureUpdateB) =>
+        Date.parse(featureUpdateB.date) - Date.parse(featureUpdateA.date),
+    )
+    const lastSeenUserFeatureUpdateDate = new Date(
+      user?.flags?.lastSeenFeatureUpdateDate,
+    )
```
karrui: is there a need for this conversion? is the flag not stored as a date type? if this is due to the serialization due to client-server boundary, we have a [`transformAllIsoStringsToDate`](https://github.com/opengovsg/FormSG/blob/bc928f713e717d9137a55f2acff5f025fde42d9d/frontend/src/utils/date.ts#L125) transformation function.

frontend/src/features/whats-new/utils/utils.ts
```diff
@@ -4,28 +4,22 @@ import { UserDto } from '~shared/types/user'
...
-      sortedFeatureUpdateList[sortedFeatureUpdateList.length - 1].date,
-  )
+    const latestFeatureUpdateDate = new Date(sortedFeatureUpdateList[0].date)
 
-  const latestFeatureUpdateDate = new Date(sortedFeatureUpdateList[0].date)
+    return isAfter(latestFeatureUpdateDate, lastSeenUserFeatureUpdateDate)
+  }
 
-  return isAfter(latestFeatureUpdateDate, lastSeenUserFeatureUpdateDate)
+  return true
```
karrui: nit: prefer early return, reverse conditional check?
should we also use more date-fns?
e.g.

```tsx
if (!user?.flags?.lastSeenFeatureUpdateDate) {
  return true
}

const { date: firstDate } = FEATURE_UPDATE_LIST
  .sort((a, b) => compareDesc(a.date, b.date))
  .shift()

return isAfter(firstDate, user.flags.lastSeenFeatureUpdateDate)
```

frontend/src/features/whats-new/FeatureUpdateList.ts
```diff
@@ -13,10 +13,11 @@ export interface FeatureUpdate {
   image?: FeatureUpdateImage
 }
 
+// TODO: Confirm the actual features which will be showcased in the What's New section
 export const FeatureUpdateList: FeatureUpdate[] = [
```
karrui: also also, should we make this UPPER_CASE since this is a constant?

frontend/src/features/whats-new/FeatureUpdateList.ts
```diff
@@ -13,10 +13,11 @@ export interface FeatureUpdate {
...
   {
     id: 1,
-    date: '17 July',
+    date: '17 June 2022',
```
karrui: since this is now an object instead of JSON, should we make this into a `Date` instead? So we do not need to perform any conversion when using it. 

To also enforce SG timezone, we can include GMT information:
```ts
date: new Date('17 Jun 2022 GMT+8')
```

karrui (approved):
lgtm with some nits

frontend/src/features/whats-new/FeatureUpdateList.ts
siddarth2824: Yeap I think that makes sense. Initially, I didn't think of making it into Date type because it's only use was to display the date as a String for each feature update item. But since I am using it for sorting, it would be better if I just stored it as a date and use date-fns to format it into string when displaying it within the drawer.

frontend/src/features/whats-new/utils/utils.ts
siddarth2824: That is very neat. I did not think of using date-fns for comparison since I was always so used to using the Javascript sort function. Thanks for this suggestion.

frontend/src/features/whats-new/utils/utils.ts
siddarth2824: Nope there is no need for conversion

PR #4063, opened 2022-06-29
feat(form-v2): billing page

### LINKED ISSUE #4044: [React] Enable admins to view their SPCP billing
in figma responses, feedback, billing

justynoh:
## Problem
This PR implements a feature for form admins to see their SPCP billing history. 

Closes #4044 

## Solution
- Add `/billing` route to app routes
- Create `BillingPage` which consists of two subpages. `BillingForm` is a form for the user to enter their e-service ID. The request automatically defaults to the current year and month. `BillCharges` shows the table of charges for the selected month.

**Breaking Changes** 

## Screenshots

**Desktop**

image
image
image
image
image
image

**Mobile**
image
image
image
image
image
image

## Tests

- [X] New stories added for the `BillingPage` page with mocked data.

For an e-service ID with no records:
- [X] Enter an e-service ID with no records, no records should be shown, instead a message saying `No records found for <e-service ID> in <month> <year>` is displayed
- [X] Table title should display `0 logins found for <e-service ID>`
- [X] Export button should be disabled and nothing should download when clicked.

For an e-service ID with records:
- [X] Enter an e-service ID with records, records should be shown in pages of 10
- [X] Table title should display `<sum of login column> found for <e-service ID>`
- [X] Clicking table headers should sort all the records (even across pages) by the corresponding column.
- [X] Clicking export should download a file named as `<e-service ID>-charges-<month> <year>.csv` with the table contents.

In both cases:
- [X] Date should be able to be reselected in the dropdown menu.
- [X] Should be able to open and close the search bar, and use it to query other e-service IDs.

And finally:
- [X] All of the above, but now on mobile! 😄


frontend/src/features/user/billing/DateRange.ts
```diff
@@ -0,0 +1,36 @@
...
+]
+
+export const dateRangeToString = ({ yr, mth }: DateRange): string =>
+  MTH_TO_STRING[mth] + ' ' + yr.toString()
+
+export const stringToDateRange = (dateRangeString: string): DateRange => {
+  const dateRangePair = dateRangeString.split(' ')
+
+  if (dateRangePair.length !== 2)
+    throw new Error(`Date range string ${dateRangeString} does not split in 2`)
```
timotheeg: We could possibly throw a [SyntaxError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) for this.

Also, since we do a syntax check, we could be much stricter on catching errors: 

```javascript
const matches = dateRangeString.match(/^([a-z]+) ([12]\d{3})$/i)

if (matches) {
    const [_, mthstring, yrstring] = matches
    const mth = MTH_TO_STRING.findIndex((mth) => mth === mthstring)

    if (mth > -1) {
        return { mth, yr: parseInt(yrstring) }
    }
}

throw new Error(`Date range string [${dateRangeString}] does not have a valid format`)

File: frontend/src/components/Searchbar/Searchbar.tsx [hunk 2]
```diff
@@ -24,21 +25,44 @@ export interface SearchbarProps extends InputProps {
...
 
   /**
-   * Optional. Function to be invoked when the search icon is clicked.
-   * If provided, the search icon will be clickable.
+   * Whether the searchbar is initially expanded or not. Defaults to `false`.
+   * @note Set this to `true` if `isExpandable` is set to `false`, else the
+   * searchbar will not be usable.
    */
-  onSearchIconClick?: () => void
+  isInitiallyExpanded?: boolean
```
timotheeg: This reads a little weird and goes against typical markup property conventions (e.g. we don't have checkboxes with `initiallyChecked`, just `checked`).

So I'm thinking the introduction of the prop `isExpandable` doesn't justify changing `isExpanded` into `isInitiallyExpanded`. I wonder instead if we could keep the searchbar prop as `isExpanded`, while using  different name for the internal state variable, say `actuallyExpanded` maybe, or whatever nicer sounding name you can come up with!

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -24,21 +25,44 @@ export interface SearchbarProps extends InputProps {
...
+   * Whether the searchbar is expandable/collapsable or not. Defaults to `true`.
    */
-  isExpanded: boolean
+  isExpandable?: boolean
 
   /**
-   * Optional. Function to be invoked when the search icon is clicked.
-   * If provided, the search icon will be clickable.
+   * Whether the searchbar is initially expanded or not. Defaults to `false`.
+   * @note Set this to `true` if `isExpandable` is set to `false`, else the
```
timotheeg: From a prop input contract perspective, I reckon if `isExpandable` is `false`, the `searchbar` should always be usable, whether `isInitiallyExpanded` is `true`, `false`, or not supplied at all 🤔 (reduce issue with callers making a mistake).

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -24,21 +25,44 @@ export interface SearchbarProps extends InputProps {
...
+      isInitiallyExpanded = false,
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      onExpand,
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      onCollapse,
+      ...props
+    },
+    ref,
+  ) => {
+    const [isExpanded, setIsExpanded] = useState<boolean>(isInitiallyExpanded)
```
timotheeg: In, relation to my comment above, to show the priority and fallback, this could possibly be 
```suggestion
    const [isActuallyExpanded, setIsExpanded] = useState<boolean>(!isExpandable || isExpanded)
```

Again, I realize `isActuallyExpanded` doesn't read very nice. Perhaps, we can use a simple prefix to denote the internal state variable instead:
```suggestion
    const [_isExpanded, setIsExpanded] = useState<boolean>(!isExpandable || isExpanded)
```

timotheeg (commented):
I'm not done with review, but I thought I'd send this before I head out to lunch! :)

frontend/src/features/user/billing/BillCharges/BillCharges.tsx
```diff
@@ -0,0 +1,253 @@
...
+  const {
+    data: { loginStats } = { loginStats: [] },
+    isLoading,
+    isRefetching,
+    refetch,
+  } = useQuery(esrvcId, () =>
+    getBillingInfo({
+      esrvcId,
+      yr: dateRange.yr.toString(),
+      mth: dateRange.mth.toString(),
```
timotheeg, line 67: /ignore: not related to this PR, but just a note that it's a little strange that `BillingQueryDto` specifies year and month as string rather than numbers 🤔

frontend/src/features/user/billing/BillCharges/BillCharges.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+  const loginCount = useMemo(
+    () =>
+      loginStats
+        ?.map(({ total }) => {
+          return total
+        })
+        .reduce((acc, curr) => {
+          return acc + curr
+        }, 0) ?? 0,
```
timotheeg: The `.map()` step is not needed.
```suggestion
        ?.reduce((acc, { total }) => {
          return acc + total
        }, 0) ?? 0,
```

frontend/src/features/user/billing/BillCharges/components/BillingTable.tsx
```diff
@@ -0,0 +1,186 @@
...
+    prepareRow,
+    getTableProps,
+    getTableBodyProps,
+    headerGroups,
+    page,
+    gotoPage,
+  } = useTable<BillingColumnData>(
+    {
+      columns: useMemo(() => BILLING_TABLE_COLUMNS, []),
+      data: useMemo(() => loginStats, [loginStats]),
```
timotheeg: These 2 `useMemo()` are not needed
```suggestion
      columns: BILLING_TABLE_COLUMNS,
      data: loginStats,
```

timotheeg (commented):
PR looks good.

I only added a few extra more minor comments.

frontend/src/features/user/billing/BillCharges/BillCharges.tsx
justynoh: you're right! thanks for pointing that out :)

frontend/src/features/user/billing/BillCharges/BillCharges.tsx
justynoh: agreed! could be worth opening an issue to change this for `after-react`? since it would require some refactoring in both front and backend

frontend/src/features/user/billing/BillCharges/components/BillingTable.tsx
justynoh: that's what I thought too, but it turns out if you get rid of these `useMemo` calls, it causes an infinite render loop for some reason :(

see also: https://react-table-v7.tanstack.com/docs/quick-start#:~:text=It%27s%20important%20that%20we%27re%20using%20React.useMemo%20here%20to%20ensure%20that%20our%20data%20isn%27t%20recreated%20on%20every%20render.%20If%20we%20didn%27t%20use%20React.useMemo%2C%20the%20table%20would%20think%20it%20was%20receiving%20new%20data%20on%20every%20render%20and%20attempt%20to%20recalculate%20a%20lot%20of%20logic%20every%20single%20time.%20Not%20cool!

frontend/src/features/user/billing/DateRange.ts
justynoh: Actually, this was a workaround and is somewhat of an artifact of the `SingleSelect` component props.

`onChange: (value: string) => void`
`items: Item[]`

where `type Item = | string | {value: string; label?: string; <and a bunch of other optional fields>} | null`

So the `onChange` function has to transform the string into a `DateRange` object. However in reality, to me the `Dropdown` component should itself guarantee that any input passed to `onChange` is a valid entry item. In that case it would make more sense for `onChange` to also accept an argument `index: number` which is the index of the item selected in the `items` array. That way we won't actually have to do any string validation (since anyway, it doesn't make sense to throw an error on strings that we created!).

Any ideas for getting around this at all? Otherwise, aggressive string validation will have to stay :(

frontend/src/features/user/billing/BillCharges/components/BillingTable.tsx
timotheeg: In that reference link you pasted, the data and column definition are objects and arrays created in-line, so it's a specific case. Objects/Arrays created in-line will always have different references (basically theres one new structure created per function call), and that's what causes them to be seen as different even though the structure doesn't change. 

In the code you have here, `BILLING_TABLE_COLUMNS` is a module constant, so not only the structure is static, but the reference is static too, and there's no need for an additional `useMemo()` wrapper. 

The same applies to `loginStats`.

Basically a `useMemo()` like this:
```javascript
useMemo(() => someRef, [someRef])
```
is useless by definition 😅 . `useMemo()` returns the same data reference across multiple calls as long as the list of depencies hasn't changed. Obviously if `someRef` changes, `useMemo()` will re-memoize, which in this case will store and return `someRef` again. Using `someRef` directly should do the exact same thing more cheaply.

frontend/src/features/user/billing/DateRange.ts
timotheeg: Right!

I think this is an encapsulation issue. In this file, the functions are standalone utilities, and it's not obvious that they are used exclusively by the the select component of the billing page. As generic conversion utilities, having the functions validating their input is the right thing to do.

Well... maybe this is nitpicking 😅, the file location **is** a hint of the usage scope, but it's not sufficient to know if we really only use it within the context of the select component or not.

On the other hand, if the functions can be scoped down to not be generic conversion utilities, and instead be specific callback handlers to the billing page's select component, then we can probably omit the validation and just have a comment that we trust the input. That narrowing down of the scope can be made explicit with naming convention perhaps, or maybe by having the functions declared in the files they are used, and not exported 🤔.

From a performance perspective, the validation is probably cheap enough that it's OK to just perform it, even when it is unnecessary (given our knowledge of the full app). So I think it's fine to have the validation.

frontend/src/features/user/billing/BillCharges/components/BillingTable.tsx
justynoh: Actually, you're right - I just tested and it works! 

Previously (before I added the `useMemo`s) I was getting an infinite render loop and somehow adding those `useMemo`s fixed it... But now, it still works when I remove the `useMemo`s? Not too sure why that happened haha

Upon further reflection upon what you said above - said infinite render loop occured in the storybook, not on the app itself. Quite sure that back then, the mocked data I input was inline into the storybook args, but since then I've moved the mock data out of that file and into an msw file. That must be why it used to throw that loop. tagging @karrui for reference haha

Regardless, will make this change! Thanks for the explanation Tim 😅

frontend/src/features/user/billing/DateRange.ts
justynoh: I think that's reasonable! As you mentioned the functions could be declared in the files they're used but I think that might start to encroach modularity, so I'm ok with just doing the input validation and treating the `dateRange` type as its own standalone interface. Will implement this 👍

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: 👍

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: Good point! Will update the component to reflect this.

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: updated: `isNowExpanded`

timotheeg (approved):

frontend/src/features/user/billing/BillCharges/BillCharges.tsx
timotheeg: Yes please :)

frontend/src/features/user/billing/BillCharges/components/BillingTableSkeleton.tsx
```diff
@@ -0,0 +1,18 @@
...
+    <Skeleton w="38%" h="1rem" />
+    <Skeleton w="23%" h="1rem" />
+    <Skeleton w="16%" h="1rem" />
+    <Skeleton w="14%" h="1rem" />
+  </Box>
+)
+
+export const BillingTableSkeleton = (): JSX.Element => (
+  <Container maxW="69.5rem">
+    {[...Array(11)].map(() => (
```
timotheeg: /SuperNit: You could use this instead (it doesn't create a secondary array):
```suggestion
    Array(11).fill().map(() => (
```

frontend/src/features/user/billing/BillCharges/components/BillingTable.tsx
```diff
@@ -0,0 +1,186 @@
...
+          >
+            {headerGroup.headers.map((column) => (
+              <Th as="div" pos="relative" {...column.getHeaderProps()}>
+                <Flex align="center" {...column.getSortByToggleProps()}>
+                  {column.render('Header')}
+                  {column.isSorted ? (
+                    <Icon
+                      fontSize="1rem"
+                      as={
+                        column.isSorted && column.isSortedDesc
```
timotheeg: `column.isSorted` is already assessed above
```suggestion
                        column.isSortedDesc
```

timotheeg (approved):
lgtm, 2 minor nits comments

frontend/src/features/user/billing/BillCharges/components/BillingTableSkeleton.tsx
justynoh: 👍

frontend/src/features/user/billing/BillCharges/components/BillingTable.tsx
justynoh: 👍

timotheeg (approved):
👍

PR #4064, opened 2022-06-30
perf(v2): virtualize Select and MultiSelect components to reduce sluggish loads and filters

### LINKED ISSUE #4025: [React]: Add virtual rendering to dropdown fields
Current dropdown fields will definitely lag if there are a couple thousand options, which may be possible.

karrui:
## Problem

When a dropdown field has thousands of fields, rendering those options will slow to a crawl. This PR utilizes `react-virtuoso` to virtualize rendering of the options, resulting in a much more performant scroll for dropdowns with a large amount of options.

side note: we can probably use `react-virtuoso` to virtualize our fields too, though ctrl+f usage may be affected

Closes #4025 

## Solution

**Breaking Changes** 

**Features**:

- Details ...

**Improvements**:

- feat: add virtualisation to `Select` and `MultiSelect` components

## Before & After Screenshots
Stories have been updated to load 2000 options.

**BEFORE**:

Recording 2022-06-30 at 12 24 46

**AFTER**:
Recording 2022-06-30 at 12 23 03

## Deploy Notes

**New dependencies**:

- `react-virtuoso`: for virtualizing dropdown options.

### DIFF
--- a/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
+++ b/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
@@ -12,6 +12,7 @@ import {
   UseMultipleSelectionProps,
 } from 'downshift'
 
+import { VIRTUAL_LIST_MAX_HEIGHT } from '../constants'
 import { useItems } from '../hooks/useItems'
 import { MultiSelectContext } from '../MultiSelectContext'
 import { SelectContext, SharedSelectContextReturnProps } from '../SelectContext'
@@ -21,7 +22,10 @@ import { itemToLabelString, itemToValue } from '../utils/itemUtils'
 
 export interface MultiSelectProviderProps<
   Item extends ComboboxItem = ComboboxItem,
-> extends Omit<SharedSelectContextReturnProps<Item>, 'isClearable'>,
+> extends Omit<
+      SharedSelectContextReturnProps<Item>,
+      'isClearable' | 'virtualListRef' | 'virtualListHeight'
+    >,
     FormControlOptions {
   /** Controlled selected values */
   values: string[]
@@ -267,6 +271,13 @@ export const MultiSelectProvider = ({
     isEmpty: selectedItems.length === 0,
   })
 
+  const virtualListHeight = useMemo(() => {
+    const totalHeight = filteredItems.length * 48
+    // If the total height is less than the max height, just return the total height.
+    // Otherwise, return the max height.
+    return Math.min(totalHeight, VIRTUAL_LIST_MAX_HEIGHT)
+  }, [filteredItems.length])
+
   return (
     <SelectContext.Provider
       value={{
@@ -302,6 +313,7 @@ export const MultiSelectProvider = ({
         resetInputValue,
         inputAria,
         virtualListRef,
+        virtualListHeight,
       }}
     >
       <MultiSelectContext.Provider
--- a/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
+++ b/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { VirtuosoHandle } from 'react-virtuoso'
 import {
   FormControlOptions,
   useFormControlProps,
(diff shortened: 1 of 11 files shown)


frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
```diff
@@ -252,6 +271,13 @@ export const MultiSelectProvider = ({
     isEmpty: selectedItems.length === 0,
   })
 
+  const virtualListHeight = useMemo(() => {
+    const totalHeight = filteredItems.length * 48
```
justynoh, line 275: hardcoding this 48 causes issues in the case where there are a small number of options that are very long (i.e. when `filteredItems.length` is small => `totalHeight` is small => `virtualListHeight` is small, but the length of the options stretch over multiple rows), case in point

normal usage:
image

long option but not many of them
image

in this case it seems that we would like actually for `virtualListHeight` to be the `VIRTUAL_LIST_MAX_HEIGHT` (or at least, bigger than what it shows currently) 

One option to fix this is not to calculate the `virtualListHeight` here, but rather in `SelectMenu` and `MultiSelectMenu` itself by getting the actual heights of the `DropdownItems`?

https://stackoverflow.com/questions/35153599/reactjs-get-height-of-an-element

frontend/src/components/Dropdown/SingleSelect/SingleSelectProvider.tsx
```diff
@@ -205,6 +222,13 @@ export const SingleSelectProvider = ({
     }
   }, [inputAriaProp, name, selectedItem])
 
+  const virtualListHeight = useMemo(() => {
+    const totalHeight = filteredItems.length * 48
```
justynoh, line 226: see the same comment in `MultiSelectProvider.tsx`

frontend/src/components/Dropdown/components/MultiSelectMenu.tsx
```diff
@@ -40,14 +49,23 @@ export const MultiSelectMenu = (): JSX.Element => {
...
+                <MultiDropdownItem
+                  key={`${itemToValue(item)}${index}`}
+                  item={item}
+                  index={index}
+                />
+              )
+            }}
+          />
+        )}
         {isOpen && items.length === 0 ? (
```
justynoh, line 69: extreme nit: I might restructure this to be

```javascript
{ isOpen && (
  items.length === 0 
    ? <ListItem...> {nothingFoundLabel} </ListItem>
    : <Virtuoso ... /> 
)}

File: frontend/src/components/Dropdown/components/SelectMenu.tsx [hunk 4]
```diff
@@ -25,14 +34,23 @@ export const SelectMenu = (): JSX.Element => {
...
+                <DropdownItem
+                  key={`${itemToValue(item)}${index}`}
+                  item={item}
+                  index={index}
+                />
+              )
+            }}
+          />
+        )}
         {isOpen && items.length === 0 ? (
```
justynoh, line 54: see the same comment in `MultiSelectMenu.tsx`

justynoh (commented):
just one issue with hardcoded dropdown item height which may affect expected behavior

frontend/src/components/Dropdown/components/MultiSelectMenu.tsx
karrui: i find these nested tertiary conditionals hard to read and parse haha. Conscious decision for readability.

frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
karrui: conscious decision for performance. this is good enough for 99% of the cases tbh, and we do not need to do intensive calculations of dom height for dynamicism.

justynoh: fair point, just wanted to make sure we considered that case!

frontend/src/components/Dropdown/components/MultiSelectMenu.tsx
justynoh: haha its fine, for me it helps the train of logic more when I can see i've covered all cases (in this case, the reason I prefer it is because if the dropdown is open, then i know i've covered both the ===0 and >0 case. and its obvious that if the dropdown is closed, we don't want to render anything)

justynoh (approved):
lgtm

PR #4065, opened 2022-06-30
fix(v2): add aria-label to select components' chevron icons

karrui:
## Problem

Accessibility testing revealed that dropdown caret had no label. 

## Solution

**Improvements**:

- fix: add aria-label to select combobox chevron icon


wanlingt (approved):
lgtm!

PR #4069, opened 2022-07-01
fix(v2): fix overflow in workspace row

karrui:
## Problem

Extremely long form titles were resulting in incorrect overflow of the page. This PR fixes that bug.
Also replace deprecated usage of `isTruncated` prop and replace them with the `noOfLine` prop.

## Solution

**Breaking Changes** 

**Improvements**:

- ref: replace deprecated usage of isTruncated with noOfLines

**Bug Fixes**:

- fix: correctly truncate workspace form row on long title

## Before & After Screenshots

**BEFORE**:

Screenshot 2022-07-01 at 11 20 20 AM
**AFTER**:

Screenshot 2022-07-01 at 11 20 13 AM

### DIFF
--- a/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/CollaboratorRow.tsx
+++ b/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/CollaboratorRow.tsx
@@ -1,4 +1,3 @@
-import { FC } from 'react'
 import { Skeleton, Stack, Text } from '@chakra-ui/react'
 
 interface CollaboratorRowProps {
@@ -8,12 +7,13 @@ interface CollaboratorRowProps {
   children: React.ReactNode
 }
 
-export const CollaboratorText: FC = ({ children }) => {
+export const CollaboratorText = ({ children }: { children?: string }) => {
   return (
     <Text
       textStyle={{ base: 'subhead-1', md: 'body-2' }}
       color={{ base: 'secondary.700', md: 'secondary.500' }}
       noOfLines={{ base: 0, md: 1 }}
+      title={children}
     >
       {children}
     </Text>
--- a/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/CollaboratorRow.tsx
+++ b/frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/CollaboratorRow.tsx
@@ -13,7 +13,7 @@ export const CollaboratorText: FC = ({ children }) => {
     <Text
       textStyle={{ base: 'subhead-1', md: 'body-2' }}
       color={{ base: 'secondary.700', md: 'secondary.500' }}
-      isTruncated
+      noOfLines={{ base: 0, md: 1 }}
     >
       {children}
     </Text>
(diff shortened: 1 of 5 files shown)


frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
```diff
@@ -32,7 +32,7 @@ export const FieldLogicBadge = ({ field }: FieldLogicBadgeProps) => {
           <Icon as={fieldMeta.icon} fontSize="1rem" />
         </Tooltip>
         {field.questionNumber ? <Text>{field.questionNumber}.</Text> : null}
-        <Text isTruncated>{field.title}</Text>
+        <Text noOfLines={1}>{field.title}</Text>
```
justynoh, line 35: is there any reason to prefer `noOfLines={1}` here rather than `noOfLines={0}`?

if we do not truncate, it looks like this:
image

whereas truncating will look like this:
image

note also the delete button overlaps with the field title.

frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/CollaboratorRow.tsx
```diff
@@ -13,7 +13,7 @@ export const CollaboratorText: FC = ({ children }) => {
     <Text
       textStyle={{ base: 'subhead-1', md: 'body-2' }}
       color={{ base: 'secondary.700', md: 'secondary.500' }}
-      isTruncated
+      noOfLines={{ base: 0, md: 1 }}
```
wanlingt, line 15: If this is truncated, is there any other way to see the full emails of existing collaborators?

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
justynoh, line 35: why do we truncate the field name here? feels like it does not need to be truncated right haha

the current behavior of dropdown logic does not truncate the option selected, as shown below

image

justynoh, line 35: unrelated to this PR: should fix delete box overlapping with long field title, regardless of truncated or not. For reference:

truncated
image

not truncated
image

edit: created issue #4076

frontend/src/features/admin-form/common/components/CollaboratorModal/CollaboratorListScreen/CollaboratorRow.tsx
karrui: good point. added a `title`; hovering the row would now display the full email.

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
karrui: no clue, preserving current behavior. can always fix in a new issue.

wanlingt (approved):
lgtm!

Issue #3968, opened 2022-06-08
Explore alternate hashing mechanism for OTPs and MyInfo


mantariksh:
We currently use the `bcrypt` library for hashing OTPs. This hashing is highly CPU-intensive, so server CPU utilisation spikes when  there are many users requesting for OTPs for verified mobile/email fields.

We should switch to a different hashing mechanism which is less CPU-intensive, and apply this throughout the app where `bcrypt` isn't necessary. Non-exhaustive list of features where this can be applied:
- Admin login OTPs
- Verified field OTPs
- MyInfo verification hashes

This needs to be done in a backwards-compatible manner, i.e. existing hashes in the DB should still be valid. One way to do this could be to check for the hash prefix, as [`bcrypt` hashes always start with `$2$`](https://auth0.com/blog/hashing-in-action-understanding-bcrypt/).

wanlingt:
@timotheeg 's comments:

On the encryption topic, the brute force concern is coupled to a malicious user having access to the DB, or having breached the servers and executing code from them. Otherwise the rate limit on REST API endpoints + CF WAF should prevent REST based BF.

From that, my thinking is that we do not need cryptographically secure hashing for OTPs, so I’d rather be as light as possible, as long as we should not have a plain, buffet-style, read-and-get-usable-value system (meaning both clear text and straight hashes are no-go) that looks negligent.

Long-ish salt + fast hashing sounds good enough to me. And for “fast hashing”, even sha2 is probably too expensive. We could use [xxhash](https://github.com/Cyan4973/xxHash) instead.

LinHuiqing:
Migrated to Linear: FRM-1382

PR #4075, opened 2022-07-01
update: number of bcrypt salt rounds for otp and myinfo hashing

hanstirtaputra:
## Problem
Bcrypt hashing is highly CPU-intensive, we want to reduce the time taken in hashing for data that does not require cryptographic security

Related to #3968

## Solution
This is a quick fix to reduce the time taken to hash OTPs as well as MyInfo verification while we explore alternative hashing mechanisms and open another PR to update the hashing mechanism once that decision is finalised.

**Breaking Changes** 

## Tests
- [x] Test that we can request OTP and successfully verify the OTP on the new server
- [x] Test that we can request OTP on the old server, switch to the new server (with `salt rounds` reduced to 1), and verify the requested OTP on the new server

### DIFF
--- a/shared/utils/verification.ts
+++ b/shared/utils/verification.ts
@@ -1,7 +1,7 @@
 import { BasicField } from '../types/field'
 
 export const VERIFIED_FIELDTYPES = [BasicField.Email, BasicField.Mobile]
-export const SALT_ROUNDS = 10
+export const SALT_ROUNDS = 1
 export const TRANSACTION_EXPIRE_AFTER_SECONDS = 14400 // 4 hours
 export const HASH_EXPIRE_AFTER_SECONDS = 60 * 30 // 30 minutes
 export const WAIT_FOR_OTP_SECONDS = 30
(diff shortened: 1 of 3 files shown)


timotheeg:
~~Hey Hans, I just realized that whatever new method of hash computation we chose, we must plan for the transition and be backward compatible at least for the duration of the OTP TTLs, no matter what. This PR is not backward compatible because a hash comparison from 1 round will fail on the stored version computed on 10 rounds.~~

~~So although the code change here is simple (change from 10 rounds to 1), it cannot be released as-is.~~

~~My apologies for not realizing that during our chat. I got too excited by the idea we can use the same hashing algo for very little code change.~~

~~Please do add that consideration to the design doc (basically it's not a con specific to the non-bcrypt hashing options).~~

~~And now considering there is no "quick" win option for a fast release, let's pick the solution from your design doc asap, and run with it with a complete PR.~~

~~We can probably close this PR in the meanwhile then.~~

~~Apologies again for not realizing on the spot when we chatted 😓.~~

Scratch that!

The [bcrypt hash format](https://en.wikipedia.org/wiki/Bcrypt#Description) contains the number of round (on top of the salt and algorithm), so the bcrypt rehashing of data for comparison is independent on our constant in code, and this PR **IS** backward compatible 🥳

timotheeg (approved):
I updated my last comment, and this is good to go! 👍

PR #4107, opened 2022-07-06
fix: bounce notification

### LINKED ISSUE #4094: Users get spammed with bounce notifications when mail services are down
The `bounce` module has mechanisms to track when email mode responses are lost due to email bounces. The database contains the fields `hasAutoSmsed` and `hasAutoEmailed` to track whether emails and SMSes have already been sent to form admins, to ensure that they do not get spammed in the case of widespread email bounces due to mail server issues.

However, we still have multiple reports of users getting spammed with multiple emails and SMSes when mail servers have issues.

Based on a quick preliminary investigation, it's likely that during a [refactor](https://github.com/opengovsg/FormSG/pull/1591), [someone](https://github.com/mantariksh) accidentally omitted the check for `hasAutoSmsed` and `hasAutoEmailed` before the email/SMS is sent. Oops. We should reinstate that check (search the PR for `hasNotified` to see the check that was omitted).

wanlingt:
## Problem

Users are spammed with multiple smses and emails after email-mode responses bounce

Closes #4094 

## Solution
Re-introduce `hasNotified`, which checks if form admins and collaborators have already been sms-ed or emailed after an email response's critical bounce.

**Breaking Changes** 

## Tests

Same tests as those in [Bounce module refactor PR](https://github.com/opengovsg/FormSG/pull/1591)
- [x] Create an email mode form. Add a valid collaborator with a registered contact number, and add a non-existent email as one of the email recipients. Submit the form. The form should NOT be deactivated, and the valid email recipient should not receive any emails/SMSes. Check the document in the Bounces collection corresponding to the form ID. The bounces array should contain the correct hasBounced key for each email address.
- [x] Remove all valid email recipients and leave only non-existent emails. Submit the form. The form should be deactivated, and the form admin + collaborators should each receive only ONE email and TWO smses, informing them of the bounce and form deactivation.

### DIFF
--- a/src/app/modules/bounce/__tests__/bounce.controller.spec.ts
+++ b/src/app/modules/bounce/__tests__/bounce.controller.spec.ts
@@ -172,6 +172,7 @@ describe('handleSns', () => {
     expect(MockBounceService.getEditorsWithContactNumbers).toHaveBeenCalledWith(
       mockForm,
     )
+    expect(mockBounceDoc.hasNotified).toHaveBeenCalled()
     expect(MockBounceService.sendEmailBounceNotification).toHaveBeenCalledWith(
       mockBounceDoc,
       mockForm,
@@ -229,6 +230,7 @@ describe('handleSns', () => {
     expect(MockBounceService.getEditorsWithContactNumbers).toHaveBeenCalledWith(
       mockForm,
     )
+    expect(mockBounceDoc.hasNotified).toHaveBeenCalled()
     expect(MockBounceService.sendEmailBounceNotification).toHaveBeenCalledWith(
       mockBounceDoc,
       mockForm,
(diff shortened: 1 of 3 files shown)


timotheeg (commented):
Just to make reveiwing the changeset easier, load it with [ignore-whitespace enabled](https://github.com/opengovsg/FormSG/pull/4107/files?w=1).

Since the PR introduces a branching behaviour, a new test needs to be added where `hasNotified()` is set to return `true`, and the test verifies that `sendEmailBounceNotification()` is not called.

wanlingt:
> Just to make reveiwing the changeset easier, load it with [ignore-whitespace enabled](https://github.com/opengovsg/FormSG/pull/4107/files?w=1).
> 
> Since the PR introduces a branching behaviour, a new test needs to be added where `hasNotified()` is set to return `true`, and the test verifies that `sendEmailBounceNotification()` is not called.

Thanks Tim! Didn't know about the whitespace setting, its really useful. I've added new tests for `hasNotified()` set to `true`

timotheeg (approved):
lgtm 👍 

I think the unit tests could be refactored at some point to be a bit more granular, but we can revisit that another time. Let's get this out to help admins.

PR #4112, opened 2022-07-06
feat: add CountryRegion field to frontend

### LINKED ISSUE #481: Country field
basically a dropdown with country list filled in.

useful for
- citizenship / nationality
- place of birth
- component of address field

jia1:
## Problem

Resolves #481

## Solution

- Handled the switch statements when rendering a field

**Breaking Changes** 

    - Details ...

**Features**:

- Users can add CountryRegion field to their forms instead of specifying every country/region by themselves when using the Dropdown field.
- Top value for the CountryRegion field is "Singapore".

## Before & After Screenshots

**BEFORE**:

**AFTER**:


shared/constants/countries.ts
```diff
@@ -1,4 +1,5 @@
 export enum Country {
+  Singapore = 'SINGAPORE',
```
timotheeg: Hmm, I'm not sure we should do that 🤔. It's a quick hack, but it looks like a breach of concern.

The `Country` enum is a shared constant, and it's raw data. Changing the key order to achieve a rendering objective in one component specifically (the CountryField) is an overreach.

Presenting Singapore first in the country field is a display concern, and should probably be done in the `CountryField` component itself instead. There'll be a bit more code to achieve that goal, but it'll be more explicit, and `CountryField` could take a `defaultCountry` arg, that defaults to Singapore.

Also can I clarify: is the requirement to have Singapore **first** in the list, or is it to have Singapore **preselected** in the list (i.e. default value)? If the later, we can use the `selected` attribute on the option to preselect it, while maintaining the overall country list order.

karrui: No requirement, but for the expected use cases it should be one of the first few for ease of selection. Though i concede that this sorting should be done in the field itself and not this enum. Thanks for raising this!

karrui (approved):
lgtm after my commits, but let's wait for @timotheeg to review.

shared/constants/countries.ts
jia1: Will be resolving this thread as the necessary refactor has been done. Thank you!

shared/constants/countryRegion.ts
```diff
@@ -1,256 +1,256 @@
...
-  Vietnam = 'VIETNAM',
-  Wake_Island = 'WAKE ISLAND',
-  Wallis_and_Futuna = 'WALLIS AND FUTUNA',
-  Western_Sahara = 'WESTERN SAHARA',
-  Yemen = 'YEMEN',
-  Yugoslavia = 'YUGOSLAVIA',
-  Zaire = 'ZAIRE',
-  Zambia = 'ZAMBIA',
-  Zimbabwe = 'ZIMBABWE',
+  Afghanistan = 'Afghanistan',
```
karrui, line 2: should we leave the enum values in upper case (since it is shared and meant to be backwards compatible)? We can do the transformation on the client side just before.

it's also ok to leave it here like this. but note that validation for myinfo values might require processing in the future when they also use this type.

jia1: Hmm, if I leave the enum values in upper case, that would mean I need to produce the correct capitalisation on render. This means I need another set of values as there is no known way to get to the right capitalisation with just JS built-in functions. Does not seem DRY to me.

Since we plan to render countries/regions in its appropriate capitalisation, I would recommend having the corrected values here and call an upper or lower-case function on validation.

Issue #4119, opened 2022-07-07
[React] refactor: modify Searchbar component to meet design system specifications


justynoh:
also need to update code for all use cases of the component, including Billing and Storage Responses.

PR #4120, opened 2022-07-07
refactor(form-v2): update searchbar to match design system specification

justynoh:
## Problem
Original searchbar component did not match design spec. This PR implements the searchbar as per design spec and adjusts Storage responses and Billing page to match the new searchbar component props and behavior. 

Closes issue #4119 

## Solution
Incorporate `X` button into the searchbar component itself, which closes and clears the value in the searchbar. Remove `RightElement` prop.

**Breaking Changes** 

## Tests

- [X] Expandable searchbar should be able to collapse when X is clicked, and the original value should be cleared from the searchbar (when expanding the searchbar again, the input field should be empty).
- [X] Non-expandable searchbar should always be open, and no X should be seen.

### DIFF
--- a/frontend/src/components/Searchbar/Searchbar.tsx
+++ b/frontend/src/components/Searchbar/Searchbar.tsx
@@ -1,12 +1,12 @@
-import { KeyboardEvent, useCallback, useRef } from 'react'
-import { BiSearch } from 'react-icons/bi'
+import { KeyboardEvent, useCallback, useRef, useState } from 'react'
+import { BiSearch, BiX } from 'react-icons/bi'
 import {
   forwardRef,
   Input,
   InputGroup,
   InputLeftElement,
   InputProps,
-  useControllableState,
+  InputRightElement,
   useMergeRefs,
   useMultiStyleConfig,
 } from '@chakra-ui/react'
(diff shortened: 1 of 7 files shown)


frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -17,104 +17,157 @@ import IconButton from '../IconButton'
...
+   * Initial value in the searchbar input.
+   */
   value?: string
 
-  rightElement?: React.ReactElement
+  /**
+   * Function to be invoked when the value in the searchbar input changes (but
+   * the search button has not been clicked).
+   */
+  onValueChange?: (newValue: string) => void
```
karrui: this is fairly confusing, since there is already the `onChange` prop. What is the difference between `onValueChange` and `onChange`?

justynoh: there's no more `onChange` i believe. I renamed it because `onChange` feels too uninformative as a prop name - it could be a change to whether its expanded, or what the search term was when the user pressed enter

karrui: opinion: `onChange` is a standard input prop though, and is what developers using inputs would expect. I think we should stick with `onChange`, especially when internally we are assigning `onValueChange` to yet another internal `onChange` prop

justynoh: hmm, i think that's reasonable, can just keep the documentation for reference

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/SubmissionSearchbar.tsx
```diff
@@ -26,7 +26,7 @@ export const SubmissionSearchbar = ({
       ref={inputRef}
       value={inputValue}
       isExpanded={!!submissionId}
-      onValueChange={(nextValue) => setInputValue(nextValue)}
+      onChange={(nextValue) => setInputValue(nextValue)}
```
karrui: nit: simplify to

```tsx
onChange={setInputValue}
```

karrui (approved):
lgtm

Issue #4006, opened 2022-06-17
[React] Admin Header component


syan-syan:
Header when signed in

https://www.figma.com/file/u00dlQLsaQfvqv6ot5ZrJZ/Form-Design-System?node-id=2153%3A4186

PR #4125, opened 2022-07-08
feat(form-v2): show emergency contact popup in workspace page

### LINKED ISSUE #4110: [React] Show popup to prompt form admins to fill in their emergency contact number
Parity feature with AngularJS variant

wanlingt:
## Problem

Emergency contact modal does not pop up on the workspace page

Working towards #4110 

Somewhat blocked by #4006

## Solution
Emergency contact modal pops up on the workspace page if the user does not have an existing emergency contact. This appears after the rollout announcement modal.

**Breaking Changes** 

**Features**:

- Local storage is used to track if a user has already seen the emergency contact modal. 
  TODO: Clear `emergencyContactKey` in local storage upon logout, when the logout button is implemented (related to Issue #4006 )

## Before & After Screenshots

**AFTER**:
emergencyContactModal (1)

## Tests
- [ ] Log in. If rollout announcement and emergency contact keys exist in local storage, remove them. On the workspace page, the rollout announcement modal should appear first, then the emergency contact modal. Refresh the workspace page. Both modals should not pop up.

### DIFF
--- a/frontend/src/features/workspace/WorkspacePage.tsx
+++ b/frontend/src/features/workspace/WorkspacePage.tsx
@@ -126,13 +126,13 @@ export const WorkspacePage = (): JSX.Element => {
     [isUserLoading, hasSeenAnnouncement],
   )
 
-  const EMERGENCY_CONTACT_KEY = useMemo(
-    () => EMERGENCY_CONTACT_KEY_PREFIX + user?._id,
+  const emergencyContactKey = useMemo(
+    () => (user?._id ? EMERGENCY_CONTACT_KEY_PREFIX + user._id : null),
     [user],
   )
 
   const [hasSeenEmergencyContact, setHasSeenEmergencyContact] =
-    useLocalStorage<boolean>(EMERGENCY_CONTACT_KEY)
+    useLocalStorage<boolean>(emergencyContactKey)
 
   const isEmergencyContactModalOpen = useMemo(
     () =>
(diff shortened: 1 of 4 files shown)


frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -160,6 +182,10 @@ export const WorkspacePage = (): JSX.Element => {
               onClose={() => setHasSeenAnnouncement(true)}
               isOpen={isAnnouncementModalOpen}
             />
+            <EmergencyContactModal
+              onClose={() => setHasSeenEmergencyContact(true)}
+              isOpen={isEmergencyContactModalOpen}
+            />
```
karrui, line 188: probably not the correct place to put this, since the modal will probably be instantiated in the admin header (when it is complete) (and it is also now when i realize that is why this feature is not in the app yet)

Add a TODO i guess, to move the auto-instantiation of the modal into the admin header when that is done?

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -122,6 +126,24 @@ export const WorkspacePage = (): JSX.Element => {
     [isUserLoading, hasSeenAnnouncement],
   )
 
+  const EMERGENCY_CONTACT_KEY = useMemo(
```
karrui: nit: since this is an inlined constant variable, should this be in camelCase instead?

should we also return `null` if user._id does not exist yet? so we don't have cases of `has-emergency-contactundefined` being set as a key.

karrui (approved):
lgtm with some suggestions

frontend/src/features/workspace/WorkspacePage.tsx
wanlingt: gotcha. and this is because we want it to appear whenever an admin logs in, not just the workspace page right? will create this as a separate issue

wanlingt: Issue #4153

PR #4145, opened 2022-07-12
feat: add views to restrict particular days for date field

### LINKED ISSUE #4134: [React/Date Validation Extension/Restricting Particular Days of the Week] Add related views
Add views for form admin user to restrict particular days of the week.

Design image for reference: 
Screenshot 2022-07-12 at 11 15 35 AM

siddarth2824:
## Problem

Add views to allow form-admin users to restrict particular days for the date field

Closes #4134 

## Solution

**Breaking Changes** 

    - Details ...

## Before & After Screenshots
New stories added as `WithParticularDaysRestricted` under `Features/EditFieldDrawer/EditDate`

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
@@ -297,12 +297,7 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
               name="invalidDays"
               rules={{
                 validate: (val) => {
-                  const validDaysSet = new Set(val)
-                  return validDaysSet.has(
-                    InvalidDaysOptions.SingaporePublicHolidays,
-                  )
-                    ? val.length >= 2
-                    : val.length >= 1
+                  return !!val.length || 'Error placeholder'
                 },
               }}
               render={({ field: { ref, ...field } }) => (
@@ -330,6 +325,9 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
                 </CheckboxGroup>
               )}
             />
+            <FormErrorMessage>
+              {get(errors, 'invalidDays.message')}
+            </FormErrorMessage>
           </FormControl>
         ) : null}
       </Stack>
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
@@ -1,20 +1,14 @@
 import { useMemo } from 'react'
 import { Controller, RegisterOptions } from 'react-hook-form'
-import {
-  Box,
-  FormControl,
-  HStack,
-  SimpleGrid,
-  Stack,
-  Text,
-} from '@chakra-ui/react'
+import { Box, FormControl, SimpleGrid, Stack, Text } from '@chakra-ui/react'
 import { isBefore, isDate, isEqual } from 'date-fns'
-import { extend, get, isEmpty, pick } from 'lodash'
+import { conformsTo, extend, get, isEmpty, pick } from 'lodash'
 
 import {
   DateFieldBase,
   DateSelectedValidation,
   DateValidationOptions,
+  RestrictParticularDaysOption,
 } from '~shared/types/field'
 
 import {
@@ -36,11 +30,23 @@ import { FormFieldDrawerActions } from '../common/FormFieldDrawerActions'
 import { EditFieldProps } from '../common/types'
 import { useEditFieldForm } from '../common/useEditFieldForm'
 
-const PARTICULAR_DAYS_OPTIONS = [
-  { leftOption: 'Monday', rightOption: 'Tuesday' },
-  { leftOption: 'Wednesday', rightOption: 'Thursday' },
-  { leftOption: 'Friday', rightOption: 'Saturday' },
-  { leftOption: 'Sunday', rightOption: 'Singapore Public Holidays' },
+export interface DaysOptionType {
+  title: string
+  index: number
+}
+
+const PARTICULAR_DAYS_OPTIONS_LEFT: DaysOptionType[] = [
+  { title: 'Monday', index: 0 },
+  { title: 'Wednesday', index: 2 },
+  { title: 'Friday', index: 4 },
+  { title: 'Sunday', index: 6 },
+]
+
+const PARTICULAR_DAYS_OPTIONS_RIGHT: DaysOptionType[] = [
+  { title: 'Tuesday', index: 1 },
+  { title: 'Thursday', index: 3 },
+  { title: 'Saturday', index: 5 },
+  { title: 'Singapore Public Holidays', index: 7 },
 ]
 
 type EditDateProps = EditFieldProps<DateFieldBase>
@@ -92,6 +98,8 @@ const transformDateEditFormToField = (
   originalField: DateFieldBase,
 ): DateFieldBase => {
   let nextValidationOptions: DateValidationOptions
+  let nextParticularDayRestrictionOption: RestrictParticularDaysOption
+
   switch (inputs.dateValidation.selectedDateValidation) {
     case '':
       nextValidationOptions = {
@@ -121,8 +129,23 @@ const transformDateEditFormToField = (
     }
   }
 
+  if (inputs.restrictParticularDays.addParticularDayRestriction) {
+    nextParticularDayRestrictionOption = {
+      addParticularDayRestriction:
+        inputs.restrictParticularDays.addParticularDayRestriction,
+      invalidDaysOfTheWeek: inputs.restrictParticularDays.invalidDaysOfTheWeek,
+    }
+  } else {
+    nextParticularDayRestrictionOption = {
+      addParticularDayRestriction:
+        inputs.restrictParticularDays.addParticularDayRestriction,
+      invalidDaysOfTheWeek: [],
+    }
+  }
+
   return extend({}, originalField, inputs, {
     dateValidation: nextValidationOptions,
+    restrictParticularDays: nextParticularDayRestrictionOption,
   })
 }
 
@@ -250,24 +273,78 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
             label="Customize days of the week"
           />
           <Text textStyle="body-2" color="secondary.400">
-            Unchecking a day will disable all the same days in the calendar
+            Checking a day will disable all the same days in the calendar
           </Text>
         </FormControl>
         {getValues('restrictParticularDays.addParticularDayRestriction') ? (
           <FormControl isRequired isReadOnly={isLoading}>
-            <Stack direction="column">
-              {PARTICULAR_DAYS_OPTIONS.map((option, index) => {
-                return (
-                  <HStack spacing="0.75rem">
-                    <Box w="9.25rem">
-                      <Checkbox defaultChecked>{option.leftOption}</Checkbox>
+            <Stack direction="row" spacing="0.75rem">
+              <Stack direction="column">
+                {PARTICULAR_DAYS_OPTIONS_LEFT.map((option) => {
+                  return (
+                    <Box width="9.25rem" key={option.index}>
+                      <Controller
+                        name="restrictParticularDays.invalidDaysOfTheWeek"
+                        control={control}
+                        key={option.index}
+                        render={({ field: { onChange, value, ref } }) => {
+                          return (
+                            <Checkbox
+                              onChange={() => {
+                                const idx = value.indexOf(option.index)
+                                value =
+                                  idx > -1
+                                    ? value.filter(
+                                        (dayId) => dayId !== option.index,
+                                      )
+                                    : [...value, option.index]
+                                onChange(value)
+                              }}
+                              ref={ref}
+                              isChecked={value.includes(option.index)}
+                            >
+                              {option.title}
+                            </Checkbox>
+                          )
+                        }}
+                      />
                     </Box>
-                    <Box>
-                      <Checkbox defaultChecked>{option.rightOption}</Checkbox>
+                  )
+                })}
+              </Stack>
+              <Stack direction="column">
+                {PARTICULAR_DAYS_OPTIONS_RIGHT.map((option) => {
+                  return (
+                    <Box key={option.index}>
+                      <Controller
+                        name="restrictParticularDays.invalidDaysOfTheWeek"
+                        control={control}
+                        key={option.index}
+                        render={({ field: { onChange, value, ref } }) => {
+                          return (
+                            <Checkbox
+                              onChange={() => {
+                                const idx = value.indexOf(option.index)
+                                value =
+                                  idx > -1
+                                    ? value.filter(
+                                        (dayId) => dayId !== option.index,
+                                      )
+                                    : [...value, option.index]
+                                onChange(value)
+                              }}
+                              ref={ref}
+                              isChecked={value.includes(option.index)}
+                            >
+                              {option.title}
+                            </Checkbox>
+                          )
+                        }}
+                      />
                     </Box>
-                  </HStack>
-                )
-              })}
+                  )
+                })}
+              </Stack>
             </Stack>
           </FormControl>
         ) : null}
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
@@ -297,7 +297,12 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
               name="invalidDays"
               rules={{
                 validate: (val) => {
-                  return !!val.length || 'Error placeholder'
+                  const validDaysSet = new Set(val)
+                  return validDaysSet.has(
+                    InvalidDaysOptions.SingaporePublicHolidays,
+                  )
+                    ? val.length >= 2 || 'Error placeholder'
+                    : val.length >= 1 || 'Error placeholder'
                 },
               }}
               render={({ field: { ref, ...field } }) => (
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
@@ -44,13 +44,14 @@ export interface InvalidDaysCheckboxOptions {
 }
 
 const INVALID_DAYS_CHECKBOX_VALUES: InvalidDaysCheckboxOptions[] = [
-  { label: 'Monday', value: DaysOfTheWeek.Monday },
-  { label: 'Tuesday', value: DaysOfTheWeek.Tuesday },
-  { label: 'Wednesday', value: DaysOfTheWeek.Wednesday },
-  { label: 'Thursday', value: DaysOfTheWeek.Thursday },
-  { label: 'Friday', value: DaysOfTheWeek.Friday },
-  { label: 'Saturday', value: DaysOfTheWeek.Saturday },
-  { label: 'Sunday', value: DaysOfTheWeek.Sunday },
+  { label: DaysOfTheWeek.Monday, value: DaysOfTheWeek.Monday },
+  { label: DaysOfTheWeek.Tuesday, value: DaysOfTheWeek.Tuesday },
+  { label: DaysOfTheWeek.Wednesday, value: DaysOfTheWeek.Wednesday },
+  { label: DaysOfTheWeek.Thursday, value: DaysOfTheWeek.Thursday },
+  { label: DaysOfTheWeek.Friday, value: DaysOfTheWeek.Friday },
+  { label: DaysOfTheWeek.Saturday, value: DaysOfTheWeek.Saturday },
+  { label: DaysOfTheWeek.Sunday, value: DaysOfTheWeek.Sunday },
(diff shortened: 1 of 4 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -28,6 +30,25 @@ import { FormFieldDrawerActions } from '../common/FormFieldDrawerActions'
...
+  { title: 'Friday', index: 4 },
+  { title: 'Sunday', index: 6 },
+]
+
+const PARTICULAR_DAYS_OPTIONS_RIGHT: DaysOptionType[] = [
+  { title: 'Tuesday', index: 1 },
+  { title: 'Thursday', index: 3 },
+  { title: 'Saturday', index: 5 },
+  { title: 'Singapore Public Holidays', index: 7 },
+]
```
karrui: probably don't need 2 separate arrays, can be a single array. also suggest converting to an enum for consistency between client and server.

```tsx
// Add comment asking ppl never to change the order if not THINGS BREAK
enum DaysOfWeek {
  Sunday = 0,
  ...
}

// render data
const INVALID_DAYS_CHECKBOX_VALUES = [
{ label: 'Sunday', value: DaysOfWeek.Sunday },
{ label: 'Monday', value: DaysOfWeek.Monday },
...
]
```

then use something like Chakra's `Wrap` component for responsiveness

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -41,6 +62,10 @@ type EditDateInputs = Pick<
     customMaxDate: string
     customMinDate: string
   }
+  restrictParticularDays: {
+    addParticularDayRestriction: boolean
```
karrui: this one should either:
1. Be removed, have a single source of truth for whether there are invalid days selected, or
2. not be sent to the server, and only used to store state or generate error messages (if any) on the **client**.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -94,8 +129,23 @@ const transformDateEditFormToField = (
...
+        inputs.restrictParticularDays.addParticularDayRestriction,
+      invalidDaysOfTheWeek: inputs.restrictParticularDays.invalidDaysOfTheWeek,
+    }
+  } else {
+    nextParticularDayRestrictionOption = {
+      addParticularDayRestriction:
+        inputs.restrictParticularDays.addParticularDayRestriction,
+      invalidDaysOfTheWeek: [],
+    }
+  }
```
karrui: no more need to transform if we only have a single source of truth for invalid dates

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +266,90 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
       </FormControl>
+      <Stack>
+        <FormControl isReadOnly={isLoading}>
+          <Toggle
+            {...register('restrictParticularDays.addParticularDayRestriction')}
+            label="Customize days of the week"
+          />
+          <Text textStyle="body-2" color="secondary.400">
+            Checking a day will disable all the same days in the calendar
+          </Text>
```
karrui: should we use the `description` prop instead in `Toggle` component?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +266,90 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
+        </FormControl>
+        {getValues('restrictParticularDays.addParticularDayRestriction') ? (
+          <FormControl isRequired isReadOnly={isLoading}>
+            <Stack direction="row" spacing="0.75rem">
+              <Stack direction="column">
```
karrui: as mentioned, `Wrap` should suffice

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +266,90 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
+                {PARTICULAR_DAYS_OPTIONS_LEFT.map((option) => {
```
karrui: as discussed, rhf returning boolean is a bug when keys are nested due to the way rhf detects controlled checkboxes. 

either add and register a hidden checkbox with the same id, or use non-nested keys.

See https://github.com/react-hook-form/react-hook-form/issues/7834#issuecomment-1040735711.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +266,90 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
+                              isChecked={value.includes(option.index)}
+                            >
+                              {option.title}
+                            </Checkbox>
+                          )
+                        }}
+                      />
+                    </Box>
+                  )
+                })}
```
karrui: probably can use it as an uncontrolled component:

```tsx
invalidCheckboxValues.map((o, idx) => (
        <Checkbox
          value={o}
          {...register(checkboxInputName, validationRules)}
        >
          {o}
        </Checkbox>
```

karrui:
moving this to draft, reopen when this is ready ok @siddarth2824

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
siddarth2824: Updated in latest commit

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
siddarth2824: Decided to leave the value there but not sent it to the server so as to store local state when the option has been selected. Updated in the latest commit.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
siddarth2824: Updated in the latest commit.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
siddarth2824: As discussed previously, a controlled checkbox group component had to be used to store the checkbox values in an array.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -28,9 +40,33 @@ import { FormFieldDrawerActions } from '../common/FormFieldDrawerActions'
 import { EditFieldProps } from '../common/types'
 import { useEditFieldForm } from '../common/useEditFieldForm'
 
+export interface InvalidDaysCheckboxOptions {
+  label: string
+  value: string
+}
+
+const INVALID_DAYS_CHECKBOX_VALUES: InvalidDaysCheckboxOptions[] = [
```
karrui: it they are the same value, do you think we can get away with just the enums in an array? 

```tsx
const INVALID_DAYS_CHECKBOX_VALUES = [InvalidDaysOptions.Monday, ...]
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -54,9 +92,20 @@ const transformDateFieldToEditForm = (field: DateFieldBase): EditDateInputs => {
       ? transformDateToShortIsoString(field.dateValidation.customMinDate) ?? ''
       : ('' as const),
   }
+
+  const nextAddParticularDayRestriction = field.invalidDays
```
karrui: nit: simplify to 

```tsx
const nextAddParticularDayRestriction = !!field?.invalidDays.length
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -54,9 +92,20 @@ const transformDateFieldToEditForm = (field: DateFieldBase): EditDateInputs => {
...
+    ? field.invalidDays.length > 0
+    : false
+
+  const nextInvalidDayOptions = field.invalidDays
```
karrui, line 96: nit: should we add comments here as why this is so?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +276,50 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
       </FormControl>
+      <Stack>
+        <FormControl isReadOnly={isLoading}>
+          <Toggle
+            {...register('addParticularDayRestriction')}
+            label="Customize days of the week"
+            description="Checking a day will disable all the same days in the calendar"
+          />
+        </FormControl>
+        {getValues('addParticularDayRestriction') ? (
```
karrui: should we use `watch` over `getValues`? [Docs](https://react-hook-form.com/api/useform/getvalues#:~:text=An%20optimized%20helper%20for%20reading%20form%20values.%20The%20difference%20between%20watch%20and%20getValues%20is%20that%20getValues%20will%20not%20trigger%20re%2Drenders%20or%20subscribe%20to%20input%20changes.) seem to imply the value returned by `watch` should be used in the render conditional over `getValues`

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +276,50 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
+          <FormControl isRequired isReadOnly={isLoading}>
+            <Controller
+              control={control}
+              name="invalidDays"
+              render={({ field: { ref, ...field } }) => {
```
karrui: `ref` seems unused here, but i suspect we want to put it in the first Checkbox (so errors will transfer focus to the correct component)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +276,50 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
+                return (
```
karrui: nit: prefer direct return over `return` keyword in the render function (unless some preprocessing is needed) for ease of reading

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -216,6 +276,50 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
+                  <CheckboxGroup
+                    {...field}
+                    defaultValue={Object.values(InvalidDaysOptions)}
+                  >
+                    <Wrap spacing="0.75rem">
+                      {INVALID_DAYS_CHECKBOX_VALUES.map((invalidDayOption) => {
+                        return (
```
karrui: ^

frontend/src/utils/date.ts
```diff
@@ -140,3 +142,21 @@ export const transformShortIsoStringToDate = (
 export const transformDateToShortIsoString = (date: unknown): string | null => {
   return isDate(date) ? format(date as Date, 'yyyy-MM-dd') : null
 }
+
+/** Transforms the invalid days array to checkbox group value */
+export const transformInvalidDaysToCheckedBoxesValue = (
+  invalidDays: InvalidDaysOptions[],
+): InvalidDaysOptions[] => {
+  return Object.values(InvalidDaysOptions).filter(
```
karrui: nit/perf: should we optimize this function? 

```tsx
const ALL_INVALID_DAYS = new Set(Object.values(InvalidDaysOptions))

// inside function
export const transformInvalidDaysToCheckedBoxesValue = (
  invalidDays: InvalidDaysOptions[],
): InvalidDaysOptions[] => {
 return invalidDays.filter(day => !ALL_INVALID_DAYS.has(value));
}
```

frontend/src/utils/date.ts
```diff
@@ -140,3 +142,21 @@ export const transformShortIsoStringToDate = (
...
+    (invalidDaysOption) => !invalidDays.includes(invalidDaysOption),
+  )
+}
+
+/** Transforms the checkbox group value to invalid days array */
+export const transformCheckedBoxesValueToInvalidDays = (
+  validDays: InvalidDaysOptions[],
+): InvalidDaysOptions[] => {
+  return Object.values(InvalidDaysOptions).filter(
```
karrui: samesies

frontend/src/utils/date.ts
karrui: oh let me rewrite that real quick this grabs the difference from the wrong array lol

other way means its an ultranit liow lol, but probably can still reuse the enum array as a constant.

[perf metrics](https://dev.to/arnaud/using-array-prototype-includes-vs-set-prototype-has-to-filter-arrays-41fg)

edit:

```tsx

const ALL_INVALID_DAYS_ARR = Object.values(InvalidDaysOptions)

export const transformInvalidDaysToCheckedBoxesValue = (
  invalidDays: InvalidDaysOptions[],
): InvalidDaysOptions[] => {
 const invalidDaysSet = new Set(invalidDays)
 return ALL_INVALID_DAYS_ARR.filter(day => !invalidDaysSet.has(value));
}
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
siddarth2824: Yeap I guess we can just map over the values in the enum when creating the checkboxes

frontend/src/utils/date.ts
siddarth2824: Can I check how this is more optimised compared to the one that I wrote? Is it because of `Object.values(InvalidDaysOptions)`?

karrui: `Array.prototype.includes` is always gonna be slower than `Set.prototype.has`

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
karrui: probably need a constant to preserve order (current enum ordering is in order (for now) but it's not actually guaranteed)

frontend/src/utils/date.ts
siddarth2824: Oh I didn't know `includes` is an O(n) operation. Thanks for the suggestion!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -30,7 +42,23 @@ import { useEditFieldForm } from '../common/useEditFieldForm'
 
 type EditDateProps = EditFieldProps<DateFieldBase>
 
-const EDIT_DATE_FIELD_KEYS = ['title', 'description', 'required'] as const
+const INVALID_DAYS_OPTIONS: string[] = [
```
karrui: nit: type should be `InvalidDaysOptions[]`

karrui (approved):
lgtm, but check if the checkboxes gets focused on error. also check that invalid options are cleared if toggle is toggled off

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -291,10 +291,20 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
+              rules={{
+                validate: (val) => {
+                  const validDaysSet = new Set(val)
+                  return validDaysSet.has(
+                    InvalidDaysOptions.SingaporePublicHolidays,
+                  )
+                    ? val.length >= 2
+                    : val.length >= 1
+                },
+              }}
```
karrui: what is this validation for? should we not need to just validate that the `val` array is non-empty?

siddarth2824: As discussed offline. I will make the necessary changes.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -54,9 +84,24 @@ const transformDateFieldToEditForm = (field: DateFieldBase): EditDateInputs => {
       ? transformDateToShortIsoString(field.dateValidation.customMinDate) ?? ''
       : ('' as const),
   }
+
+  const nextAddParticularDayRestriction = !!field?.invalidDays?.length
```
karrui, line 88: will this cause the toggle to turn off when user unchecks all options in the UI? is that intended behaviour?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -118,6 +179,11 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
     },
   })
 
+  const watchAddParticularDayRestriction = useWatch({
+    control: control,
```
karrui: nit: 
```suggestion
    control,
```

nit: any reason why we chose `useWatch` hook over `useForm#watch`?

karrui (approved):
lgtm

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
siddarth2824: In the documentation, they mentioned that using `useWatch` hook is slightly more performant than use `watch`. This was their reasoning:
```Behaves similarly to the watch API, however, this will isolate re-rendering at the custom hook level and potentially result in better performance for your application.```
Link for reference: https://react-hook-form.com/api/usewatch

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
siddarth2824: If user unchecks all options, the `invalidDays` array will be of length > 0 which will assign `true` to `nextAddParticularDayRestriction` and the toggle will be turned on still. However if the user checks all the options, the toggle will still be turned on. I checked with Pearly, if this is the supposed intended behaviour and she said she will check with Shanty and get back to me.

karrui (approved):

Issue #3362, opened 2022-02-03
[React/Builder] Render and edit start page on builder


karrui:
Mocks in "Builder:Design" figma page
![Screenshot 2022-02-03 at 12 00 42 PM](https://user-images.githubusercontent.com/22133008/152279087-7d55fc11-b8bb-4d1b-804c-75173682ba4b.png)

PR #4155, opened 2022-07-13
feat(form-v2/design-1): add design drawer, enable users to edit estimated time and form instructions

justynoh:
## Problem
This PR implements the start page view in the form builder and enables users to edit and save their estimated time and instructions for the form start page. 

Makes progress on #3362

## Solution
**Key Changes**
- Added `DesignDrawer` and `useDesignStore`.
- Updated `FormBuilder.tsx` to display form's `startPage`.
- Additional mutations and queries added to `admin-form/common`. 

**Other Changes**
- Refactor `FormBuilder` and `BuilderFields` files (just moved some functions around for better clarity and modularity).
- Updated some `endPage` builder files (for consistency/symmetry and clarity between `startPage` and `endPage` builder code)

**Breaking Changes** 

**Future PRs**
- feat(form-v2/design-2): logo radio buttons and form color theme selection [#3362] 
- feat(form-v2/design-3): addition of title alignment and banner options _(more complex - will require model updates)_ [#3362]
- feat(form-v2/design-4): open design drawer when start page is clicked in builder [#4159] 

## Screenshots

image

## Tests
- [X] Open a new form and edit the design.
- [X] Open an existing form and edit the design.


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
```diff
@@ -34,3 +51,55 @@ export const BuilderFields = memo(
...
+    if (stateData.state === BuildFieldState.CreatingField) {
+      return getFormFieldsWhileCreating(existingFields, stateData)
+    }
+    return existingFields
+  }, [formData, stateData])
+
+  return {
+    builderFields,
+  }
+}
```
karrui: was there any reason why this is moved into BuilderFields.tsx? Seemed much easier to find as its own file.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/EndPageView.tsx
```diff
@@ -0,0 +1,76 @@
...
+export const EndPageView = ({ ...props }: FlexProps): JSX.Element => {
+  const { data: form } = useAdminForm()
+  const { endPageData, setState } = useEndPageBuilderStore((state) => ({
+    endPageData: endPageDataSelector(state),
+    setState: setStateSelector(state),
+  }))
+
+  useEffect(() => {
+    if (form) setState(form.endPage)
+  }, [form, setState])
```
karrui: this was not in the previous code, is this needed? what bug was this fixing? (also pls refrain from adding changes unrelated to the PR haha hard to review)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
```diff
@@ -0,0 +1,234 @@
...
+  setStateSelector,
+  startPageDataSelector,
+  useDesignStore,
+} from '../useDesignStore'
+
+import { EmptyFormPlaceholder } from './BuilderAndDesignPlaceholder/EmptyFormPlaceholder'
+import BuilderAndDesignPlaceholder from './BuilderAndDesignPlaceholder'
+import { BuilderFields, useBuilderFields } from './BuilderFields'
+
+const FormBuilderStartPage = () => {
```
karrui: should we move this to its own component?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
```diff
@@ -0,0 +1,234 @@
...
+    startPageData: startPageDataSelector(state),
+    setState: setStateSelector(state),
+  }))
+
+  useEffect(() => {
+    if (form) setState(form.startPage)
+  }, [form, setState])
+
+  // Color theme options and other design stuff, identical to public form
+
```
karrui: nit: extra new line

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
```diff
@@ -0,0 +1,234 @@
...
+      startPageData.colorTheme
+        ? `theme-${startPageData.colorTheme}.500`
+        : `neutral.200`,
+    [startPageData.colorTheme],
+  )
+
+  const estTimeString = useMemo(() => {
+    if (!startPageData.estTimeTaken) return ''
+    return simplur`${startPageData.estTimeTaken} min[|s] estimated time to complete`
+  }, [startPageData.estTimeTaken])
```
karrui: if these are the same, any reason why we did not just reuse our current useFormHeader and useFormBannerLogo hooks? 

we can generify that hook by allowing `form` to be passed into them:

```tsx
const useFormBannerLogo = (form) => {
  ....
}

// in public form hook using the generified hook
const usePublicFormBannerLogo = () => {
  const { form } = usePublicFormContext()
  const { ...originalHookStuff } = useFormBannerLogo(form)
  ...
}

// Our admin form hook can look like:
const useBuilderFormBannerLogo = () => { 
  const { form } = useAdminFormOrTheStoreWeGotThisFrom()
  ... you get the idea 
}
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+  //
```
karrui: nit: leftover comment? should be deleted.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
```diff
@@ -0,0 +1,234 @@
...
+    }
+  }, [form, startPageData, logoBucketUrl])
+
+  const logoImgAlt = useMemo(
+    () => (form ? `Logo for ${form.admin.agency.fullName}` : undefined),
+    [form],
+  )
+
+  return (
+    <>
```
karrui: Should we figure out if it is worth it to also reuse the existing FormHeader component? Form builder view is supposed to reflect what the current public form looks like - will also be useful to only need to maintain a single component.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditEndPageDrawer/EditEndPage.tsx
```diff
@@ -133,5 +132,5 @@ export const EndPageBuilderInput = ({
 export const EditEndPage = (): JSX.Element => {
   const { data: form } = useAdminForm()
 
-  return <>{form ? <EndPageBuilderInput endPage={form.endPage} /> : null}</>
+  return form ? <EndPageBuilderInput endPage={form.endPage} /> : <></>
```
karrui: nit: can probably return `null` in this case, we just need to update the return type of this component

```tsx
export const EditEndPage = (): JSX.Element | null => {
```

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
```diff
@@ -0,0 +1,33 @@
...
+    startPageData: {
+      logo: { state: FormLogoState.Default },
+      colorTheme: FormColorTheme.Blue,
+    },
+    setState: (startPage) => {
+      const current = get()
+      if (isEqual(current.startPageData, startPage)) return
+      set({
+        startPageData: startPage,
+      })
```
karrui: useDesignStore seems like its functions could be in useBuilderAndDesignStore (after all it's in the name... haha)

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
```diff
@@ -0,0 +1,33 @@
...
+  setState: (startPage: FormStartPage) => void
+}
+
+export const useDesignStore = create<DesignStore>(
+  devtools((set, get) => ({
+    startPageData: {
+      logo: { state: FormLogoState.Default },
+      colorTheme: FormColorTheme.Blue,
+    },
+    setState: (startPage) => {
```
karrui: nit: the name of this function could probably be updated for clarity? how about setStartPageData?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/EndPageView.tsx
justynoh: same thing as what we were talking about earlier - the old store state from a different form gets loaded first, before getting updated to the correct form state

this fixes the issue in the start page builder so i added the same to the end page builder that's all haha

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
justynoh: yeah i considered it. This was a conscious decision, and I was going to rename `useBuilderAndDesignStore` later on to just `useBuilderStore` in a separate PR. Reason for this is that in `BuilderAndDesignDrawer`, we decide which drawer to show based on the `DrawerTabs` active tab.

image

Adding the start page design code to this complicates it by adding a case for a state that a user should never get to in runtime - specifically, one in which `DrawerTabs.Builder` is active, but the BuilderState is `editingStartPage`. Of course, it's the tradeoff between having a bit more code vs not being statically guaranteed. I think the tradeoff is worth it. Feel free to opine though

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
justynoh: update the corresponding for endpage also? hahah

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
justynoh: lol oops

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
justynoh: oh lol i usually leave an extra line for comments relating to blocks but sure haha

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
justynoh: ... shoot. not sure why I did that. will revert

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditEndPageDrawer/EditEndPage.tsx
justynoh: errrr that would have other effects upstream i think? (in `BuilderAndDesignContent`)

we could do that in another smaller PR

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/EndPageView.tsx
karrui: feel like this is just fixing the symptom instead of the cause. like discussed, maybe we should reset the store on unmount instead of setting it on mount.

```tsx
useEffect(() => {
  return () => {
    resetStore()
  }
}, [])
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
justynoh: ok, agreed. but also, a lot of stuff in `features/admin-form` references stuff in `features/public-form` (not vice-versa though). we should probably move all of this stuff to `features/form/utils` in a separate PR?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
justynoh: I thought about it but i didn't think it was big enough yet to move. 

looking forward, it'll definitely become bigger so i think its worth to pre-emptively move in this PR bah. will do!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditEndPageDrawer/EditEndPage.tsx
karrui: won't have other effects. can you try? should work as per normal

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
karrui: ye

justynoh: okei

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditEndPageDrawer/EditEndPage.tsx
justynoh: oop you are right hahah

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/EndPageView.tsx
justynoh: It would make sense if we change the data fields to be possibly undefined. Can do!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
justynoh: ok, managed to figure this out. Additionally, i think we should do this for the end page component as well in a separate PR. welp.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
justynoh: See #4186

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
justynoh: See #4187

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/StartPageView.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const StartPageView = () => {
+  const { data: form } = useCreateTabForm()
+  const startPageFromStore = useDesignStore(startPageDataSelector)
+
+  // Switch over to store's start page once the store is populated (done when
+  // user opens the drawer)
+  const startPage = useMemo(
+    () => (startPageFromStore ? startPageFromStore : form?.startPage),
+    [startPageFromStore, form?.startPage],
+  )
```
karrui, line 26: same here

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/StartPageView.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <>
+      <FormBannerLogo {...formBannerLogoProps} />
+      <FormHeader
+        title={form?.title}
+        estTimeString={estTimeString}
+        titleBg={titleBg}
+        titleColor={titleColor}
+        showHeader={true}
```
karrui: nit: react props operate on presence of the prop. do you think it is cleaner to not assign `true`?
```tsx
<FormHeader
  showHeader // equivalent to showHeader={true}
  ...
/>
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderAndDesignContent.tsx
```diff
@@ -175,20 +33,12 @@ export const BuilderAndDesignContent = ({
...
-          }
-        />
-      </Flex>
-    </>
+    <Flex flex={1} bg="neutral.200" overflow="auto">
+      {stateData.state === BuildFieldState.EditingEndPage ? (
+        <EndPageView />
+      ) : (
+        <FormBuilder placeholderProps={placeholderProps} />
+      )}
```
karrui: should we revert this change? Hiding the form builder will make the transitions more responsive than rerendering.

frontend/src/features/admin-form/common/mutations.ts
```diff
@@ -316,7 +317,25 @@ export const useMutateFormPage = () => {
   const queryClient = useQueryClient()
   const toast = useToast({ status: 'success', isClosable: true })
 
-  const updateFormData = useCallback(
+  const updateStartPageFormData = useCallback(
+    (newData: StartPageUpdateDto) => {
+      queryClient.setQueryData(adminFormKeys.startPage(formId), newData)
```
karrui: as discussed, this does nothing. we only need to update the admin form.
should remove both the start and end page keys since they are not actually used for queries

frontend/src/features/admin-form/common/mutations.ts
```diff
@@ -334,7 +353,24 @@ export const useMutateFormPage = () => {
...
+      toastDescription: string
+    }) => {
+      toast.closeAll()
+      updateStartPageFormData(newData)
+      toast({
+        description: toastDescription,
+      })
+    },
+    [toast, updateStartPageFormData],
+  )
```
karrui: nit: suggest inlining since it is only used once

frontend/src/features/admin-form/common/queries.ts
```diff
@@ -26,6 +26,7 @@ export const adminFormKeys = {
     [...adminFormKeys.id(id), 'collaborators'] as const,
   previewForm: (id: string) =>
     [...adminFormKeys.id(id), 'previewForm'] as const,
+  startPage: (id: string) => [...adminFormKeys.id(id), 'startPage'] as const,
   endPage: (id: string) => [...adminFormKeys.id(id), 'endPage'] as const,
```
karrui: these keys should be deleted since they are not used

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderAndDesignContent.tsx
karrui: maybe should add reasoning as a comment

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/EndPageView.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const EndPageView = ({ ...props }: FlexProps): JSX.Element => {
+  const { data: form } = useAdminForm()
+  const endPageFromStore = useEndPageBuilderStore(endPageDataSelector)
+
+  // Switch over to store's end page once the store is populated (done when
+  // user opens customization page)
+  const endPage = useMemo(
+    () => (endPageFromStore ? endPageFromStore : form?.endPage),
+    [endPageFromStore, form?.endPage],
+  )
```
karrui: should we encapsulate this process of retrieving endPage (and startPage) into a custom hook ala `useBuilderFields`?

make it more obvious the relationship between what to show

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/BuilderAndDesignDrawer.tsx
```diff
@@ -47,21 +48,20 @@ export const BuilderAndDesignDrawer = (): JSX.Element | null => {
 
   const renderDrawerContent = useMemo(() => {
```
karrui: maybe add return type to ensure we cover all switch cases

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  resetStartPageDataSelector,
+  setStartPageDataSelector,
+  useDesignStore,
+} from '../../useDesignStore'
+import { validateNumberInput } from '../../utils/validateNumberInput'
+import { CreatePageDrawerCloseButton } from '../CreatePageDrawerCloseButton'
+import { DrawerContentContainer } from '../EditFieldDrawer/edit-fieldtype/common/DrawerContentContainer'
+
+interface DesignDrawerInputProps {
+  startPage: FormStartPage
```
karrui: do we need the prop if this is not reused?  can just invoke the hook in the DesignDrawer component

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
```diff
@@ -0,0 +1,175 @@
...
+}
+
+const DesignDrawerInput = ({
```
karrui: nit: probably can call this DesignDrawer instead? since this component is more than one input.

frontend/src/features/public-form/components/FormStartPage/useFormBannerLogo.tsx
```diff
@@ -0,0 +1,36 @@
+import { useMemo } from 'react'
+
+import { AdminFormDto, PublicFormDto } from '~shared/types'
+import { FormLogoState } from '~shared/types/form/form_logo'
+
+import { useEnv } from '~features/env/queries'
+
+export const useFormBannerLogo = (form?: AdminFormDto | PublicFormDto) => {
```
karrui: instead of the entire form object, maybe we should just pass in the needed props in so we do not need to update the entire form on changes

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/StartPageView.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { FormBannerLogo } from '~features/public-form/components/FormStartPage/FormBannerLogo'
+import { FormHeader } from '~features/public-form/components/FormStartPage/FormHeader'
+import { useFormBannerLogo } from '~features/public-form/components/FormStartPage/useFormBannerLogo'
+import { useFormHeader } from '~features/public-form/components/FormStartPage/useFormHeader'
+
+import { useCreateTabForm } from '../useCreateTabForm'
+import { startPageDataSelector, useDesignStore } from '../useDesignStore'
+
+export const StartPageView = () => {
```
karrui, line 14: we should probably reuse public form's start page component so to maintain single source of truth, and also so we can be sure the preview is accurate

frontend/src/features/admin-form/common/mutations.ts
```diff
@@ -313,80 +313,8 @@ export const useMutateCollaborators = () => {
...
   if (!formId) throw new Error('No formId provided')
-
-  const queryClient = useQueryClient()
   const toast = useToast({ status: 'success', isClosable: true })
 
-  const updateStartPageFormData = useCallback(
-    (newData: StartPageUpdateDto) => {
-      queryClient.setQueryData(adminFormKeys.startPage(formId), newData)
-      // Only update adminForm if it already has prior data.
-      queryClient.setQueryData<AdminFormDto | undefined>(
```
karrui: i think we will still need to update the query for the entire form, since this mutation will indeed update that portion

frontend/src/features/admin-form/common/mutations.ts
```diff
@@ -313,80 +313,8 @@ export const useMutateCollaborators = () => {
...
-      )
-    },
-    [formId, queryClient],
-  )
-
-  const updateEndPageFormData = useCallback(
-    (newData: EndPageUpdateDto) => {
-      queryClient.setQueryData(adminFormKeys.endPage(formId), newData)
-      // Only update adminForm if it already has prior data.
-      queryClient.setQueryData<AdminFormDto | undefined>(
```
karrui, line 323: same here.

frontend/src/features/public-form/components/FormStartPage/FormHeader.tsx
```diff
@@ -1,98 +1,101 @@
...
+              variant="solid"
+              colorScheme="primary"
+              aria-label="Mobile section sidebar"
+              fontSize="1.5rem"
+              icon={<BxMenuAltLeft />}
+              d={{ base: 'flex', md: 'none' }}
+              onClick={onMobileDrawerOpen}
+            />
+          ) : (
+            <></>
```
karrui: nit: not changed by your PR, but should be change this to `null` instead?

frontend/src/features/public-form/components/FormStartPage/useFormBannerLogo.tsx
```diff
@@ -1,35 +1,45 @@
...
-import { useEnv } from '~features/env/queries'
-
-export const useFormBannerLogo = (form?: AdminFormDto | PublicFormDto) => {
-  const { data: { logoBucketUrl } = {} } = useEnv(
-    form?.startPage.logo.state === FormLogoState.Custom,
-  )
+interface UseFormBannerLogoInputs {
+  logoBucketUrl?: string
+  logo?: FormLogo
+  agency?: AgencyBase
```
karrui, line 9: nit: use AgencyDto instead?

frontend/src/features/admin-form/common/mutations.ts
justynoh: oh yes, we should still do that

frontend/src/features/public-form/components/FormStartPage/useFormBannerLogo.tsx
justynoh: the type of `form.admin.agency` is `AgencyBase` though, and none of the extra info is needed from `AgencyDto`. is there a reason for using `AgencyDto` instead?

karrui: because that's the real field that is being used, but i guess that's fine too.

karrui (approved):
lgtm

PR #4158, opened 2022-07-13
chore: cleanup unused settings end page builder

### LINKED ISSUE #4154: [React] Remove thank you settings tab from settings page in builder

hanstirtaputra:
## Problem
We no longer need to the feature to edit the end page from the settings page

Closes #4154 

## Solution
Remove end page builder from settings

**Breaking Changes** 

## Tests
- [x] Test that we can still edit the end page from the form builder page. Edit the end page from the form builder, then share the form, submit a response and see that the end page is indeed updated.


justynoh (approved):
lgtm

PR #4173, opened 2022-07-13
fix: reject invalid utf8 encoded characters

tshuli:
## Problem
Closes [#144](https://github.com/datagovsg/formsg-private/issues/144)

## Explanation
- Unicode characters can be escaped as `\\u[0-9a-fA-F]{4}` in JS: see https://mathiasbynens.be/notes/javascript-escapes
- Not all 65536 combinations are valid unicode (see https://charbase.com/dbbb-unicode-invalid-character for some examples). 
- If the node14 application receives unicode-escaped characters are not valid unicode, node 14 treats the sequence of characters as a string e.g. \udbbb is treated as a 6-character string instead of an escaped unicode sequence (not found in documentation, but this is observed through testing)
- Although the 6-character string is considered as a valid UTF-8 encoded string by the application (because each individual character is correctly encoded), if this is saved into the db, an error is thrown when the mongodb driver attempts to read the db document, because the driver interprets the sequence as an invalid UTF8 encoded character
- The fix implements a regex check at the application level when form fields are created or updated, to reject unicode character sequences that exist as a 6-character string
  - Since valid unicode-escaped characters will be processed correctly (e.g. \u00ae is processed as ® instead of the 6-character string), they will not trigger an error. 
  - If the user intends to input a 6-character string of the same form e.g. \udbbb, the backslash will be escaped (i.e. double backslash `\\udbbb`) and hence this will also not trigger an error.
  
Sample code for nodejs:
```javascript
// valid utf8
console.log(JSON.stringify("\u4E09"))
// "三"

// invalid utf8
console.log(JSON.stringify("\udbbb"))
// "\udbbb"
```

## Tests

*Valid UTF8 encoding is accepted*
- [ ] (POST) Create dropdown form field. Clone the network request as cURL. Modify it such that the fieldoptions includes `\u00ae` e.g. `fieldOptions: ['Option1\u00ae']`. Send the network request. You should receive 200.
- [ ] (PUT) Modify the dropdown form field options. Clone the network request. Now modify the network request so that the fieldoptions includes `fieldOptions: ['Option1\u00ae']`. Send the network request. You should receive 200.

*String that _looks_ like invalid UTF8 encoding is accepted*
- [ ] (POST) Create dropdown form field. Clone the network request as cURL. Modify it such that the fieldoptions includes `\\udbbb` e.g. `fieldOptions: ['Option1\\udbbb']`. Send the network request. You should receive 200.
- [ ] (PUT) Modify the dropdown form field options. Clone the network request. Now modify the network request so that the fieldoptions includes `fieldOptions: ['Option1\\udbbb']`. Send the network request. You should receive 200.

*Invalid UTF8 encoding is rejected*
- [ ] (POST) Create dropdown form field. Clone the network request as cURL. Modify it such that the fieldoptions includes `\udbbb` e.g. `fieldOptions: ['Option1\udbbb']`. Send the network request. You should receive 400.
- [ ] (PUT) Modify the dropdown form field options. Clone the network request. Now modify the network request so that the fieldoptions includes `fieldOptions: ['Option1\udbbb']`. Send the network request. You should receive 200.

### DIFF
--- a/src/app/modules/form/form.utils.ts
+++ b/src/app/modules/form/form.utils.ts
@@ -149,3 +149,10 @@ export const extractFormLinkView = <T extends IFormDocument>(
     link: `${appUrl}/${_id}`,
   }
 }
+
+/**
+ * Regex to to detect invalid-encoded utf-8 characters in stringified form field input
+ * Matches any sequence which starts with a non-backslash, an odd number of backslashes, followed by unicode escape sequence
+ * See https://mathiasbynens.be/notes/javascript-escapes for regex on unicode escape sequences
+ */
+export const UNICODE_ESCAPED_REGEX = /[^\\](\\\\)*\\u[0-9a-fA-F]{4}/
--- a/src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
+++ b/src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
@@ -1610,6 +1610,93 @@ describe('admin-form.form.routes', () => {
     })
   })
 
+  describe('PUT /:formId/fields', () => {
+    it('should return 200 if form field contains valid utf8 encoded unicode character sequence', async () => {
+      // Arrange
+      const fieldToInsert = omit(
+        generateDefaultField(BasicField.Dropdown, {
+          fieldOptions: ['Option1'],
+        }),
+        ['_id', 'globalId', 'getQuestion'],
+      )
+
+      const formToUpdate = (await EmailFormModel.create({
+        title: 'Form to update',
+        emails: [defaultUser.email],
+        admin: defaultUser._id,
+        form_fields: [generateDefaultField(BasicField.Rating)],
+      })) as IPopulatedForm
+
+      // Act
+      await request
+        .post(`/admin/forms/${formToUpdate._id}/fields`)
+        // No positional argument
+        .send(fieldToInsert)
+
+      const formToCheck = await EmailFormModel.findById(formToUpdate._id).lean()
+
+      const insertedField = formToCheck!.form_fields![1]
+      const insertedFieldId = insertedField._id
+
+      const updatedField = { ...insertedField, fieldOptions: ['Option1\u00ae'] }
+
+      const fieldUpdateResponse = await request
+        .put(`/admin/forms/${formToUpdate._id}/fields/${insertedFieldId}`)
+        // No positional argument
+        .send(updatedField)
+
+      const formToCheckUpdated = await EmailFormModel.findById(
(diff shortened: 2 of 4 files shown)


src/app/modules/form/form.utils.ts
```diff
@@ -149,3 +149,10 @@ export const extractFormLinkView = <T extends IFormDocument>(
     link: `${appUrl}/${_id}`,
   }
 }
+
+/**
+ * Regex to to detect invalid-encoded utf-8 characters in stringified form field input
+ * Matches any sequence which starts with a non-backslash, an odd number of backslashes, followed by unicode escape sequence
+ * See https://mathiasbynens.be/notes/javascript-escapes for regex on unicode escape sequences
+ */
+export const UNICODE_ESCAPED_REGEX = /[^\\](\\\\)*\\u[0-9a-fA-F]{4}/
```
timotheeg, line 158: /justANote: I was about to comment that this regex won't capture sequences at the beginning of inputs (since it consumes at least one leading character), but of course when used exclusively on stringified input, a first character (the double quote `"`) is guaranteed to exists, so it should be fine.

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
```diff
@@ -1714,6 +1801,63 @@ describe('admin-form.form.routes', () => {
       expect(response.body).toMatchObject(fieldToInsert)
     })
 
+    it('should return 200 if form field contains valid utf8 encoded unicode character sequence', async () => {
```
timotheeg, line 1851: /nit: Since the handling behaviour of intentionally escaped sequences is mentioned in a comment, and the regexp is non-trivial, let's have one more test here to document that a valid escaped sequence will not be incorrectly rejected:
```
fieldOptions: ['Option1\\udbbb'],
```

timotheeg (approved):
LGTM! 

And I didn't know about this behaviour of JSON stringification, so I added it to the PR description too :) 

```javascript
// valid utf8
console.log(JSON.stringify("\u4E09"))
// "三"

// invalid utf8
console.log(JSON.stringify("\udbbb"))
// "\udbbb"
```

src/app/modules/form/form.utils.ts
tshuli: thanks! it's exactly what I had in mind as well :)

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
tshuli: thanks for this, I've added in the test

PR #4200, opened 2022-07-15
feat(form-workspaces-be/1): add workspace db model

### LINKED ISSUE #4038: [React/Form Workspaces] Add database models
Add database models related to form workspaces

hanstirtaputra:
## Problem
We need to model workspaces in the database

Closes #4038 

## Solution
Add `workspace` database schema, model, and tests. The model is not complete with all the functionality in this PR. Functionalities will be added incrementally in subsequent PRs so that each PR is smaller.

**Breaking Changes** 

**Features**:

- Add `workspace` database model
- Add `schema` level validators to ensure that `formId` added to `workspace` document is valid
- Add unit tests for `workspace` model

## Tests
Unit tests are added to confirm functionality

## Deployment
- [ ] Add index for `admin` field in `workspaces` collection


src/app/models/workspace.server.model.ts
```diff
@@ -0,0 +1,90 @@
...
+                isUserFormAdmin || isUserFormCollaborator
+              const isFormIdUnique = this.formIds.includes(formId)
+              const isFormIdValid = doesUserHaveFormAccess && isFormIdUnique
+
+              if (!isFormIdValid) return false
+            }
+
+            return true
+          },
+        },
```
karrui: i think we should properly distinguish what should be a service-level validation and what should be a model level validation.

model-level validation should be validation that does not rely on business logic, such as formId uniqueness.
service-level validation is probably everything else.

in this validation block, i'd think that checking of permissions should be a service-level validation, as putting that validation here could result in some edge case like when users stop having read permissions  to the form and the db would prevent deletion of that form from the workspace.

src/app/models/workspace.server.model.ts
```diff
@@ -0,0 +1,90 @@
...
+        message: "Failed to update workspace document's formIds",
+      },
+    },
+    {
+      timestamps: {
+        createdAt: 'created',
+        updatedAt: 'lastModified',
```
karrui: probably don't need to rename this ba... the renaming feels like tech debt from the past haha.

src/app/models/workspace.server.model.ts
hanstirtaputra: Right that makes sense. I've moved the check for permissions out of the schema, and only perform `formId` uniqueness check instead.

src/app/models/workspace.server.model.ts
hanstirtaputra: Okay removed this!

src/app/models/workspace.server.model.ts
```diff
@@ -0,0 +1,55 @@
...
+          return areFormIdsUnique
+        },
+      },
+      message: "Failed to update workspace document's formIds",
+    },
+  })
+
+  WorkspaceSchema.index({
+    admin: 1,
+  })
```
karrui, line 37: might also want to add that in your PR comment so we remember to also add an index in our DB provider

src/app/models/workspace.server.model.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    formIds: {
+      type: [Schema.Types.ObjectId],
+      validate: {
+        async validator(this: IWorkspaceSchema) {
+          const areFormIdsUnique =
+            new Set(this.formIds).size === this.formIds.length
+
+          return areFormIdsUnique
+        },
```
karrui: nit: simplify?

```ts
validate: v => new Set(v.formIds).size === v.formIds.length,
```

karrui (approved):
lgtm

shared/types/workspace.ts
```diff
@@ -0,0 +1,15 @@
+import { Opaque } from 'type-fest'
+import { FormId } from './form'
+import { UserId } from './user'
+
+export type WorkspaceId = Opaque<string, 'WorkspaceId'>
+
+export type Workspace = {
+  _id: WorkspaceId
+  title: string
+  count: number
```
mantariksh, line 10: what's this `count` for? doesn't seem to be in the design doc, in fact the doc says that the count is not needed

src/app/models/__tests__/workspace.server.model.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        '__v',
+      ])
+
+      expect(saved._id).toBeDefined()
+      expect(saved.created).toBeInstanceOf(Date)
+      expect(saved.lastModified).toBeInstanceOf(Date)
+      expect(actualSavedObject).toEqual(expectedWorkspaceObject)
+    })
+
+    it("should update workspace's formIds successfully because user is the form admin", async () => {
```
mantariksh: this doesn't describe the test case clearly - the action is not an update, and "because user is the form admin" is unclear. maybe "should save successfully when workspace admin is the same as form admin"?

src/app/models/__tests__/workspace.server.model.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        '_id',
+        'created',
+        'lastModified',
+        '__v',
+      ])
+
+      expect(actualSavedObject).toEqual(expectedWorkspaceObject)
+    })
+
+    it("should create and save successfully because user has read and write access in the form's permissionList", async () => {
```
mantariksh: we normally use "when" rather than "because", since we're talking about different situations rather than cause and effect

src/app/models/workspace.server.model.ts
```diff
@@ -0,0 +1,55 @@
...
+  const WorkspaceSchema = new Schema<IWorkspaceSchema, IWorkspaceModel>({
+    title: {
+      type: String,
+      validate: [
+        /^[a-zA-Z0-9_\-./() &`;'"]*$/,
+        'Workspace title cannot contain special characters',
+      ],
+      required: 'Workspace title cannot be blank',
+      minlength: [4, 'Workspace title must be at least 4 characters'],
+      maxlength: [200, 'Workspace title can have a maximum of 200 characters'],
```
mantariksh: I see you've followed the same format as form titles, but given the space constraints on the UI, maybe workspace titles should be shorter? what's the behaviour of the UI when the workspace title overflows?

this is a small but important decision because it's much much harder to make these limits more strict later on (what would we do for workspace titles that already exceed the limit?) than to make them more lenient

src/app/models/__tests__/workspace.server.model.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+const MOCK_FORM_ID = new ObjectId()
+const MOCK_WORKSPACE_ID = new ObjectId()
+const MOCK_WORKSPACE_FIELDS = {
+  _id: MOCK_WORKSPACE_ID,
+  title: 'Workspace1',
+  admin: MOCK_USER_ID,
+  formIds: [],
+}
+
+describe('Workspace Model', () => {
```
mantariksh, line 22: do we need tests on the format of the title as well? normally the way I think about model-level tests is that we should test the validation on each key

shared/types/workspace.ts
mantariksh: ah, I see it becomes a virtual later on

src/app/models/workspace.server.model.ts
hanstirtaputra: Yep I think workspace titles can be shorter. I'll discuss this with design! 

As for the behaviour, when the workspace title overflows, the the title will be truncated and appended with ellipsis in the UI

src/app/models/__tests__/workspace.server.model.spec.ts
hanstirtaputra: Okay, will add the tests for title validation as well

src/app/models/__tests__/workspace.server.model.spec.ts
hanstirtaputra: I see, I'll update the test descriptions, thanks!

src/app/models/workspace.server.model.ts
hanstirtaputra: Reduced maximum title length to 50 characters as discussed!

mantariksh (approved):

PR #4205, opened 2022-07-15
feat: expose corppass oidc public jwks

tshuli:
## Problem
- Exposes the corppass OIDC public jwks endpoint in preparation for corppass OIDC migration

## Tests
- [ ] Corppass jwks should be accessible on `/api/v3/corppass/.well-known/jwks.json`

## Deploy Notes
- Before deployment, SSH into one of the instances and load the public jwks file, and then update parameter store with the new env var

`CP_OIDC_RP_JWKS_PUBLIC_PATH` Path to the Relying Party's Public Json Web Key Set used for Corppass-related communication with NDI. This will be hosted at api/v3/corppass/.well-known/jwks.json endpoint.


timotheeg (approved):
LGTM 👍

PR #4212, opened 2022-07-18
chore: ts-essentials should stay a dep (not dev-dep)

timotheeg:
## Problem
[`ts-essential`](https://www.npmjs.com/package/ts-essentials) has been moved to dev-dependencies, but [an error class of it is used in a file](https://github.com/opengovsg/FormSG/blob/2e3c97d50dcde1f3eec87d5eae2335839fbc815a/src/app/modules/form/public-form/public-form.controller.ts#L5), causing a run-time error on startup when the module is not present

## Solution
Move `ts-essential` from dev-dependencies back to dependencies

Note: `ts-essential` is meant to be a dev-dependency, we should investigate if the error reference can be removed from our own prod code.


justynoh (approved):
lgtm

PR #4216, opened 2022-07-18
feat: stop tracking middleware spans with dd-trace

timotheeg:
## Problem
By default, dd-trace's express integration creates spans for all middlewares.

While this is a deep level of granularity, we do not actually expect middleware themselves to be a problem, and this creates a huge number of spans, which are not free!

image

## Solution

middleware spans can be disabled in dd-trace (see [documentation](https://datadoghq.dev/dd-trace-js/interfaces/plugins.express.html)), if we initialize the tracer manually. The PR does that.

Here is a sample tracking for the same call (otp verify) without the middleware spans.

image

### DIFF
--- a/src/app/loaders/datadog-tracer.ts
+++ b/src/app/loaders/datadog-tracer.ts
@@ -1,0 +1,12 @@
+import tracer from 'dd-trace'
+
+tracer.init()
+
+// setup express to not track middlewares as spans
+// see documentation: https://datadoghq.dev/dd-trace-js/interfaces/plugins.express.html
+tracer.use('express', {
+  enabled: true,
+  middleware: false,
+})
+
+export default tracer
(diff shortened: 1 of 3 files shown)


src/app/loaders/datadog-tracer.ts
```diff
@@ -0,0 +1,12 @@
+import tracer from 'dd-trace'
+
+tracer.init()
+
+// setup express to not track middlewares as spans
+// see documentation: https://datadoghq.dev/dd-trace-js/interfaces/plugins.express.html
+tracer.use('express', {
+  enabled: true,
```
tshuli: minor nit that since this defaults to `true`, might not be necessary?
```suggestion
```

from https://datadoghq.dev/dd-trace-js/interfaces/plugins.express.html

tshuli (approved):
lgtm, just one minor nit for consideration

src/app/loaders/datadog-tracer.ts
timotheeg: Good comment! I initially did not add `enabled:true`, but also did not add the final module export 😑🤦. And that didn't work, I added both at the same time to make it work, but indeed, now I don't know which one did the job.

Let me try to do it more granularly to get to the minimum required.

timotheeg: Confirmed that `enabled:true` wasn't needed, and removed it in c1698c51f28ac4b64ae067b1436267cd920d2661

Issue #4228, opened 2022-07-19
chore: tighten joi validation to remove isNewClient param


tshuli:
- #4213 removed the isNewClient flag from both the frontend and server
- For backward compability, joi validation on the backend still allows the client to continue sending `isNewClient` param for old clients (#4229). New clients no longer have isNewClient flag.
- To remove isNewClient param from joi validation 2 weeks after release

PR #4213, opened 2022-07-18
chore: remove isNewClient flag

tshuli:
## Problem
- Removes `isNewClient` flag
- follow up to #3927
- closes #4103 

## Tests

- [ ] Create storage mode form. Test that logo and image upload works and can be displayed in public form.
- [ ] (safety check, should not be affected by this PR) Add attachment field. Test that file upload works and can be retrieved from storage mode

PR #4229, opened 2022-07-19
fix: allow client to send isNewClient param

tshuli:
## Problem
- #4213 removed the isNewClient flag from both the frontend and server
- For backward compability, joi validation on the backend needs to allow the client to continue sending `isNewClient` param for old clients. New clients no longer have isNewClient flag.
- To remove 2 weeks after release (TODO in #4228)


timotheeg (approved):
LGTM 👍 Good catch!

PR #4230, opened 2022-07-19
feat(form-v2): add option for admins to switch from AngularJS to React

### LINKED ISSUE #4096: Admins should be able to switch from AngularJS -> React
When React is first released, In the AngularJS app, form admins should have the ability to opt in to use React instead of AngularJS.

[Design here](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=8540%3A116586)
Final copy for info box tbc, but we can start work on it first.

We will need to create a new endpoint which updates their cookies (see `REACT_MIGRATION_ADMIN_COOKIE_NAME` env var).

wanlingt:
## Problem
Admins are not able to opt in to use React instead of AngularJS

Closes #4096 

## Solution

Add info box for admins to click to switch to React. Upon clicking,
- `v2-admin-ui` cookie is set as `react`
- the page reloads

**Breaking Changes** 

**Features**:

- Add new `react-switch-banner` component
- Create new `AdminService` to access the setting-of-cookie API endpoint

**Improvements**:

- Move `UiCookieValues` to `shared/types` folder

**Bug Fixes**:

- update param in `admin.routes`'s environment route

## Before & After Screenshots

**AFTER**:
https://user-images.githubusercontent.com/22133008/179727750-7eb2b3f2-8deb-4779-b9f4-c56482747c9c.mov

179693604-a37c13f6-70d6-4256-8c64-0681a743085c

## Tests
- [x] Log in. On the landing page's info box, click 'Try it here!'. The page should reload and should display the React interface. Also, the value of `v2-admin-ui cookie` should change from `angular` to `react`.


src/public/modules/forms/admin/components/react-switch-banner.client.component.js
```diff
@@ -0,0 +1,29 @@
...
+angular.module('core').component('reactSwitchBannerComponent', {
+  templateUrl:
+    'modules/forms/admin/componentViews/react-switch-banner.client.view.html',
+  controller: ['$scope', '$q', '$window', reactSwitchBannerController],
+  controllerAs: 'vm',
+})
+
+function reactSwitchBannerController($scope, $q, $window) {
+  const vm = this
+  $scope.ui = UiCookieValues.React
```
karrui: i don't think this is needed

src/app/routes/api/v3/admin/admin.routes.ts
```diff
@@ -10,6 +10,6 @@ AdminRouter.use('/forms', AdminFormsRouter)
 
 // This endpoint doesn't reaaaallly need to be a verified admin to be used
 AdminRouter.get(
-  '/environment/:env(react|angular)',
+  '/environment/:ui(react|angular)',
```
karrui, line 13: this was a bug yeah?

karrui (approved):
lgtm.

src/app/routes/api/v3/admin/admin.routes.ts
wanlingt: yup!

src/public/modules/forms/admin/components/react-switch-banner.client.component.js
wanlingt: by this do you mean I can replace `$scope` with `vm`?

src/app/routes/api/v3/admin/admin.routes.ts
timotheeg: Oopsie my bad 😰 . Thanks for that @wanlingt 🙏

src/public/services/__tests__/AdminService.test.ts
```diff
@@ -0,0 +1,32 @@
...
+jest.mock('axios', () => MockAxios)
+
+describe('AdminService', () => {
+  describe('adminChooseEnvironment', () => {
+    const MOCK_UI = UiCookieValues.React
+    const MOCK_RESPONSE = UiCookieValues.React
+
+    it('should return environment name if GET request succeeds', async () => {
+      // Arrange
+      const mockUi = UiCookieValues.React
```
timotheeg: Why does this test need 2 consts `MOCK_UI` and `mockUi`?

src/public/services/__tests__/AdminService.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      const expectedResponse = UiCookieValues.React
+
+      // Act
+      const actual = adminChooseEnvironment(MOCK_UI)
+      MockAxios.mockResponse({ data: MOCK_RESPONSE })
+
+      // Assert
+      await expect(actual).resolves.toEqual(MOCK_RESPONSE)
```
timotheeg, line 19: Is this assertion meaningful to the test? It's looks like it's testing `MockAxios` rather than our own code.

src/public/services/__tests__/AdminService.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      expect(MockAxios.get).toHaveBeenCalledWith(
+        `/api/v3/admin/environment/${MOCK_UI}`,
+        expectedResponse,
+        { params: { data: mockUi } },
```
timotheeg: How are these arguments part of the axios `get()` call 🤔 ? The call in AdminService only supplies the URL, no?

timotheeg (commented):
Since the main behaviour is the reload, if we could have a test that bannerController does reload indeed after successfully calling `AdminService.adminChooseEnvironment()`, that would be great!

src/public/services/__tests__/AdminService.test.ts
wanlingt: you're right, have removed!

src/public/services/__tests__/AdminService.test.ts
wanlingt: removed `mockUi`

src/public/services/__tests__/AdminService.test.ts
wanlingt: removed `MockAxios.mockResponse`

wanlingt:
> Since the main behaviour is the reload, if we could have a test that bannerController does reload indeed after successfully calling `AdminService.adminChooseEnvironment()`, that would be great!

Decided not to go ahead with the automated test - will do it manually instead.

src/public/modules/forms/admin/components/react-switch-banner.client.component.js
wanlingt: replaced!

PR #4232, opened 2022-07-19
fix(form-fields): disable field creation when field is disabled

### LINKED ISSUE #4118: [React] Form admins are able to attempt creation of MyInfo fields even when they do not have it enabled
backend correctly rejects this, but we should still block it properly in the client.

Recording 2022-07-06 at 16 53 35

seaerchin:
## Problem
Previously, `FieldListOption`s were _visually_ disabled but users could still drag/click on them to create fields manually. This was due to the `Draggable`'s state not being tied to the `isDisabled` prop and the `handleClick` not being dependent on the prop also. 

This was most evident during: 
1.  loading states on initial form load 
2. when `myInfo` fields >= 30 

Closes #4118

## Solution
1. Thread the `isDisabled` prop through to `Draggable` and `FieldListOption`
2. add extra loading states for creation to check behaviour for `disabled` states

### DIFF
--- a/frontend/src/features/admin-form/AdminFormCreatePage.stories.tsx
+++ b/frontend/src/features/admin-form/AdminFormCreatePage.stories.tsx
@@ -66,6 +66,17 @@ DesktopAllFields.parameters = {
     responseMode: FormResponseMode.Email,
   }),
 }
+export const DesktopLoading = Template.bind({})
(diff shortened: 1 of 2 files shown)


seaerchin:
closing this to create a branch with the proper naming

PR #4244, opened 2022-07-20
feat: edit date field schema to add additional attribute for invalid days of the week

### LINKED ISSUE #4136: [React/Date Validation Extension/Restricting Particular Days of the Week] Add additional attribute in dateField
An additional attribute needs to be added to _**DateValidationOptions**_  in **_dateField_** to allow the server to validate that the dates picked by public users does not violate the days that the form admin users have restricted.

siddarth2824:
## Problem
Currently in the date field schema, there is no attribute which is able to encapsulate information regarding restricted days of the week.

Closes #4136 

## Solution

Add an additional attribute to date field schema to encapsulate information on restricted days of the week

**Breaking Changes** 

    - Details ...

### DIFF
--- a/shared/types/field/dateField.ts
+++ b/shared/types/field/dateField.ts
@@ -16,4 +16,5 @@ export type DateValidationOptions = {
 export interface DateFieldBase extends MyInfoableFieldBase {
   fieldType: BasicField.Date
   dateValidation: DateValidationOptions
+  invalidDaysOfTheWeek?: number[]
 }
--- a/shared/types/field/dateField.ts
+++ b/shared/types/field/dateField.ts
@@ -7,6 +7,16 @@ export enum DateSelectedValidation {
   Custom = 'Custom date range',
 }
 
+export enum DaysOfTheWeek {
+  Sunday = 0,
+  Monday = 1,
+  Tuesday = 2,
+  Wednesday = 3,
+  Thursday = 4,
+  Friday = 5,
+  Saturday = 6,
+}
+
 export type DateValidationOptions = {
   customMaxDate: Date | null
   customMinDate: Date | null
@@ -16,5 +26,5 @@ export type DateValidationOptions = {
 export interface DateFieldBase extends MyInfoableFieldBase {
   fieldType: BasicField.Date
   dateValidation: DateValidationOptions
-  invalidDaysOfTheWeek?: number[]
+  invalidDaysOfTheWeek?: DaysOfTheWeek[] | null
 }
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -158,53 +158,6 @@ describe('Form Model', () => {
         expect(actualSavedObject).toEqual(expectedObject)
       })
 
-      it('should create new date form field with default empty array of invalid days of the week and save successfully', async () => {
-        const formParams = merge({}, MOCK_ENCRYPTED_FORM_PARAMS, {
-          admin: MOCK_ADMIN_OBJ_ID,
-          form_fields: [generateDefaultField(BasicField.Date)],
-        })
-
-        const form = await Form.create(formParams)
-        const saved = await form.save()
-
-        expect(saved._id).toBeDefined()
-        expect(saved.created).toBeInstanceOf(Date)
-        expect(saved.lastModified).toBeInstanceOf(Date)
-
-        const savedObjectFormFields = (
-          saved.form_fields as Types.DocumentArray<IFieldSchema>
-        ).toObject()
-
-        expect(savedObjectFormFields[0].invalidDaysOfTheWeek).toEqual([])
-      })
-
-      it('should create new date form field with expected array of invalid days of the week and save successfully', async () => {
-        const expectedInvalidDaysOfTheWeek = [1, 2, 3]
-        const formParams = merge({}, MOCK_ENCRYPTED_FORM_PARAMS, {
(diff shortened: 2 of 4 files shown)


karrui (approved):
lgtm, pls add tests for that specific prop

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -157,6 +158,53 @@ describe('Form Model', () => {
         expect(actualSavedObject).toEqual(expectedObject)
       })
 
+      it('should create new date form field with default empty array of invalid days of the week and save successfully', async () => {
```
karrui: should we extract it to its own test file? can look at https://github.com/opengovsg/FormSG/blob/2854ed2ea6147d7190aac3da11ad3696e0aaf0c5/src/app/models/field/__tests__/emailField.spec.ts#L24 for example

of course, field model test files can be kept in this directory still

siddarth2824: OHHHH I did not see that folder. My bad. I will create my own test file in that directory

siddarth2824: Updated in latest commit

src/app/models/field/dateField.ts
```diff
@@ -23,6 +23,10 @@ const createDateFieldSchema = () => {
         default: null,
       },
     },
+    invalidDaysOfTheWeek: {
+      type: [Number],
+      default: [],
```
karrui: nit: don't need default for arrays, since mongoose [adds an array by default](https://mongoosejs.com/docs/faq.html#array-defaults)

src/app/models/field/dateField.ts
```diff
@@ -23,6 +23,10 @@ const createDateFieldSchema = () => {
         default: null,
       },
     },
+    invalidDaysOfTheWeek: {
+      type: [Number],
```
karrui: do we want to use the DaysOfWeek enum instead? Allows for better type and validation

src/app/models/field/__tests__/dateField.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    })
+
+    // Assert
+    const expected = merge(expectedDateField, mockDateField, {
+      _id: expect.anything(),
+    })
+    expect(actual.field.toObject()).toEqual(expected)
+  })
+
+  it('should set expected array for invalid days of the week when date field specifies', async () => {
```
karrui: should we add a negative test for rejection on invalid data? think back to equivalence partitioning in CS2103 😉

shared/types/field/dateField.ts
```diff
@@ -16,4 +16,5 @@ export type DateValidationOptions = {
 export interface DateFieldBase extends MyInfoableFieldBase {
   fieldType: BasicField.Date
   dateValidation: DateValidationOptions
+  invalidDaysOfTheWeek?: number[]
```
karrui: should we use an enum as discussed in a previous PR?

siddarth2824: Yes. Updated in latest commit. Forget about that because I changed it in the other PR.

src/app/models/field/dateField.ts
siddarth2824: Updated PR

src/app/models/field/__tests__/dateField.spec.ts
siddarth2824: Updated PR

src/app/models/field/dateField.ts
siddarth2824: Updated PR

src/app/models/field/dateField.ts
```diff
@@ -26,6 +26,23 @@ const createDateFieldSchema = () => {
...
+            (v) => !isNaN(Number(v)),
+          )
+          for (const daysOfTheWeek of this.invalidDaysOfTheWeek) {
+            if (!daysOfTheWeekValues.includes(daysOfTheWeek)) {
+              return false
+            }
+          }
+          return true
+        },
+      },
```
karrui: as discussed, could we use mongoose's `match` property instead? if not, suggest simplifying instead of doing the same work:

```ts
validate: v => Array.isArray(v) && !v.some(el => el === null),
```

src/app/models/field/__tests__/dateField.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    await expect(
+      MockParent.create({
+        responseMode: FormResponseMode.Encrypt,
+        field: mockDateField,
+      }),
+    ).rejects.toThrowError(mongoose.Error.ValidationError)
+  })
+
+  it('should throw an error when null value is added to invalidDaysOfTheWeek attribute array', async () => {
```
karrui, line 123: nit: should we expand this test case to [valid, invalid] values so we can be sure any combination would be correctly validated?

karrui (approved):
Lgtm

src/app/models/field/__tests__/dateField.spec.ts
siddarth2824: added this case in the latest commit

karrui (approved):
lgtm

karrui:
backend tests do not run on form-v2/develop branch (disabled last time so we can save time on CI/CD), but maybe we should bring it back as we begin to work on features on form-v2. Can you double check that it runs locally and maybe open a new PR to re-enable the backend CI pipeline?

Issue #4039, opened 2022-06-23
[React/Form Workspaces] Add backend logic for workspaces APIs


hanstirtaputra:
Add backend logic for the form workspaces related APIs

PR #4247, opened 2022-07-20
feat(form-workspaces-be/2): add get workspaces functionality

hanstirtaputra:
Part of  #4039 

## Solution
Add GET `api/v3/admin/workspaces` endpoint business logic

**Breaking Changes** 

## Tests
Unit tests are included for this endpoint

### DIFF
--- a/shared/types/workspace.ts
+++ b/shared/types/workspace.ts
@@ -5,7 +5,6 @@ import { UserId } from './user'
 export type WorkspaceId = Opaque<string, 'WorkspaceId'>
 
 export type Workspace = {
-  _id: WorkspaceId
   title: string
   count: number
   formIds: FormId[]
--- a/shared/types/workspace.ts
+++ b/shared/types/workspace.ts
@@ -7,7 +7,6 @@ export type WorkspaceId = Opaque<string, 'WorkspaceId'>
 export type Workspace = {
   _id: WorkspaceId
   title: string
-  count: number
   formIds: FormId[]
   admin: UserId
 }
--- a/shared/types/workspace.ts
+++ b/shared/types/workspace.ts
@@ -5,6 +5,7 @@ import { UserId } from './user'
 export type WorkspaceId = Opaque<string, 'WorkspaceId'>
 
 export type Workspace = {
+  _id: WorkspaceId
   title: string
   count: number
   formIds: FormId[]
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -1,6 +1,7 @@
 import { Mongoose, Schema } from 'mongoose'
+import { WorkspaceDto } from 'shared/types/workspace'
 
-import { IWorkspaceModel, IWorkspaceSchema } from '../../types'
+import { IUserSchema, IWorkspaceModel, IWorkspaceSchema } from '../../types'
 
 export const WORKSPACE_SCHEMA_ID = 'Workspace'
 
@@ -35,6 +36,16 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
   WorkspaceSchema.index({
     admin: 1,
   })
+
+  WorkspaceSchema.virtual('count').get(function (this: IWorkspaceSchema) {
+    return this.formIds.length
+  })
+
+  WorkspaceSchema.statics.getWorkspaces = async function (
+    admin: IUserSchema['_id'],
+  ): Promise<WorkspaceDto[]> {
+    return this.find({ admin: admin }).sort('title').exec()
+  }
 
   return db.model<IWorkspaceSchema, IWorkspaceModel>(
     WORKSPACE_SCHEMA_ID,
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -28,6 +28,7 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
         validate: (v: Schema.Types.ObjectId[]) => new Set(v).size === v.length,
         message: "Failed to update workspace document's formIds",
       },
+      message: "Failed to update workspace document's formIds",
     },
     { timestamps: true },
   )
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -28,7 +28,6 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
         validate: (v: Schema.Types.ObjectId[]) => new Set(v).size === v.length,
         message: "Failed to update workspace document's formIds",
       },
-      message: "Failed to update workspace document's formIds",
     },
     { timestamps: true },
   )
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -8,12 +8,6 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
   const schemaOptions = {
     id: false,
     timestamps: true,
-    toObject: {
-      virtuals: true,
-    },
-    toJSON: {
-      virtuals: true,
-    },
   }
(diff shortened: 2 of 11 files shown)


src/app/modules/workspace/__tests__/workspace.routes.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    await dbHandler.clearDatabase()
+    jest.restoreAllMocks()
+  })
+  afterAll(async () => await dbHandler.closeDatabase())
+
+  describe('GET /workspaces', () => {
+    const GET_WORKSPACES_ENDPOINT = '/workspaces'
+
+    it('should return 200 with an empty array when a user has no workspaces', async () => {
+      const response = await request.get('/workspaces')
```
karrui, line 54: nit: should this use the `GET_WORKSPACES_ENDPOINT` constant too?

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+const WorkspaceModel = getWorkspaceModel(mongoose)
+
+const app = setupApp('/workspaces', WorkspacesRouter, {
+  setupWithAuth: true,
+})
+
+const MOCK_USER_ID = new ObjectId()
+const MOCK_FORM_ID = new ObjectId()
+const MOCK_WORKSPACE_ID = new ObjectId()
+const MOCK_WORKSPACE_FIELDS = {
```
karrui: nit: should this be called `MOCK_WORKSPACE_DOC` instead?

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      const response = await request.get('/workspaces')
+
+      expect(response.status).toEqual(200)
+      expect(response.body).toEqual([])
+    })
+
+    it("should return 200 with an array of the user's workspaces", async () => {
+      await WorkspaceModel.create(MOCK_WORKSPACE_FIELDS)
+      const response = await request.get(GET_WORKSPACES_ENDPOINT)
+      const expected = await WorkspaceModel.find({ _id: MOCK_WORKSPACE_ID })
```
karrui: nit: would it be clearer for the reader if we used `MOCK_WORKSPACE_DOC._id` (even though they are the same thing, haha)

karrui (approved):
lgtm

shared/types/workspace.ts
```diff
@@ -5,7 +5,6 @@ import { UserId } from './user'
 export type WorkspaceId = Opaque<string, 'WorkspaceId'>
 
 export type Workspace = {
-  _id: WorkspaceId
```
mantariksh, line 8: why remove this?

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    })
+    request = await createAuthedSession(user.email, request)
+  })
+  afterEach(async () => {
+    await dbHandler.clearDatabase()
+    jest.restoreAllMocks()
+  })
+  afterAll(async () => await dbHandler.closeDatabase())
+
+  describe('GET /workspaces', () => {
```
mantariksh, line 50: should we test the ordering of workspaces as well?

mantariksh (commented):
do we need model-level tests for `WorkspaceSchema.statics.getWorkspaces`?

src/app/models/workspace.server.model.ts
```diff
@@ -40,6 +41,16 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     admin: 1,
   })
 
+  WorkspaceSchema.virtual('count').get(function (this: IWorkspaceSchema) {
```
mantariksh: btw, cool use of virtuals! this is a good use case for it (although I'm not sure what it's used for? but I'll probably find out in later PRs)

src/app/modules/workspace/workspace.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ErrorResponseData } from '../core/core.types'
+
+const logger = createLoggerWithLabel(module)
+
+export const mapRouteError = (
+  error: ApplicationError,
+  coreErrorMessage?: string,
+): ErrorResponseData => {
+  switch (error.constructor) {
+    case DatabaseError:
```
mantariksh, line 22: what about the other specific error classes returned by `transformMongoError`, like `DatabaseValidationError` or `DatabaseConflictError`? should those return different status codes?

hanstirtaputra: Will update to include the other MongoErrors. Yes I think those should return different status codes

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
hanstirtaputra: Yes it should, thanks

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
hanstirtaputra: Yep, I think so, will update to use `MOCK_WORKSPACE_DOC._id` instead

shared/types/workspace.ts
hanstirtaputra: Thought this was the wrong type to add the `_id` in. I've reverted this change

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
hanstirtaputra: Yep, I've added test for workspace ordering as well

src/app/models/workspace.server.model.ts
hanstirtaputra: It's used in the FE, it's not necessary as clients can just use the `formIds.length`, but I felt this was just a nice-to-have

mantariksh: I would say if it's not necessary then let's not add it in. creating derived properties makes it harder to keep everything in sync, and this doesn't give us any performance boost

hanstirtaputra: Okay then, I've updated the PR to remove the `count` virtual property!

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    it('should return 401 when user is not logged in', async () => {
+      await logoutSession(request)
+      const response = await request.get(GET_WORKSPACES_ENDPOINT)
+
+      expect(response.status).toEqual(401)
+      expect(response.body).toEqual({ message: 'User is unauthorized.' })
+    })
+
+    it('should return 500 when database errors occur', async () => {
```
mantariksh, line 100: we can test for the other types of database errors as well. probably sufficient to do it just in integration tests, no need to add them in every other layer

src/app/routes/api/v3/admin/workspaces/workspaces.routes.ts
```diff
@@ -24,7 +24,6 @@ WorkspacesRouter.route('/')
    *
    * @returns 200 with a list of workspaces owned by the user
    * @returns 401 when user is not logged in
```
mantariksh, line 26: can update this with the new status codes

mantariksh (approved):
couple of more nits, otherwise lgtm!

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
hanstirtaputra: Similar to the other testcases in the codebase, I added the test to the controller layer as we can mock the `DatabaseConflictError` there, whereas in the integration test it's harder to mock the difference in mongoose versions

PR #4248, opened 2022-07-20
feat: migration corppass to oidc

tshuli:
This PR migrates corppass to OIDC.

There will be a separate PR to remove deprecated endpoints for SP/CP SAML; the current SAML code is kept for backward compatibility.

## Tests

**Form Submission**
- [ ] Check that Singpass, CP, Myinfo, Sgid email mode forms can be created and submitted normally with verified sms and email fields
- [ ] Repeat with storage mode (except myinfo)

**Prefill**
- [ ] Check that prefill works with corppass form

**Backward compatibility**
- [ ] Sign in to corppass using saml. (eg on `uat`). Use the saml jwt to load corppass form in oidc. Form should load and submit correctly. 

## Deploy Notes
- New JWKS keysets should be loaded onto EFS and env vars updated ahead of deployment

- `env var` : env var details
| `CP_OIDC_NDI_DISCOVERY_ENDPOINT` | NDI's Corppass OIDC Discovery Endpoint                                                                                                                                                          |
| `CP_OIDC_NDI_JWKS_ENDPOINT`      | NDI's Corppass OIDC JWKS Endpoint                                                                                                                                                               |
| `CP_OIDC_RP_CLIENT_ID`           | The Relying Party's Client ID as registered with NDI                                                                                                                                            |
| `CP_OIDC_RP_REDIRECT_URL`        | The Relying Party's Redirect URL                                                                                                                                                                |
| `CP_OIDC_RP_JWKS_SECRET_PATH`    | Path to the Relying Party's Secret Json Web Key Set used for Corppass-related communication with NDI                                                                                            |


src/app/modules/spcp/spcp.oidc.client.types.ts
```diff
@@ -0,0 +1,94 @@
...
+export type CpOidcClientConstructorParams = {
+  cpOidcRpClientId: SpcpOidcBaseClientConstructorParams['rpClientId']
+  cpOidcRpRedirectUrl: SpcpOidcBaseClientConstructorParams['rpRedirectUrl']
+  cpOidcNdiDiscoveryEndpoint: SpcpOidcBaseClientConstructorParams['ndiDiscoveryEndpoint']
+  cpOidcNdiJwksEndpoint: SpcpOidcBaseClientConstructorParams['ndiJwksEndpoint']
+  cpOidcRpSecretJwks: SpcpOidcBaseClientConstructorParams['rpSecretJwks']
+  cpOidcRpPublicJwks: SpcpOidcBaseClientConstructorParams['rpPublicJwks']
+}
+
+export type SpcpOidcBaseCilentCacheConstructorParams = {
```
timotheeg: Typo here and all associated imports 😅 
```suggestion
export type SpcpOidcBaseClientCacheConstructorParams = {
```

src/app/modules/spcp/spcp.oidc.client.cache.ts
```diff
@@ -36,7 +36,7 @@ const INITIAL_REFRESH_TIMEOUT = 10000
  * Cache class which provides read-through capability and refresh-ahead before expiry
  * Handles discovery and retrieval of NDI's public jwks
  */
-export class SpOidcClientCache {
+export class SpcpOidcBaseCilentCache {
```
timotheeg: Typo in class name as well
```suggestion
export class SpcpOidcBaseClientCache {
```

src/app/modules/spcp/spcp.oidc.client.ts
```diff
@@ -8,77 +9,90 @@ import {
...
+  #rpRedirectUrl: string
+  #rpClientId: string
+  #authType: FormAuthType.SP | FormAuthType.CP
 
   /**
    * @private
    * accessible only for testing
    */
-  _spOidcClientCache: SpOidcClientCache
+  _spcpOidcBaseCilentCache: SpcpOidcBaseCilentCache
```
timotheeg: Typo in member as well (and in accesses below in the file)
```suggestion
  _spcpOidcBaseClientCache: SpcpOidcBaseClientCache
```

src/app/modules/spcp/spcp.oidc.client.ts
```diff
@@ -169,8 +183,10 @@ export class SpOidcClient {
       scope: 'openid',
       response_type: 'code',
       state: state,
-      esrvc: esrvcId,
       nonce: ulid(), // Not used - nonce is a required parameter for SPCP's OIDC implementation although it is optional in OIDC specs
+      ...(this.#authType === FormAuthType.SP
+        ? { esrvc: esrvcId }
+        : { esrvcID: esrvcId }), // Additional parameter for agencies' eservice Id. Labelled esrvc in Singpass OIDC and esrvcID in Corppass OIDC
```
timotheeg: This (being given inconsistent naming specs by NDI for corpPass and singPass) is so damn sad 😑 

/suggestionFeelFreeToIgnore (also not sure if Typescript will allow):
```suggestion

      // Additional parameter for agencies' eservice Id. Labelled esrvc in Singpass OIDC and esrvcID in Corppass OIDC
      [this.#authType === FormAuthType.SP ? 'esrvc' : 'esrvcID']: esrvcId,
```

src/app/modules/spcp/spcp.oidc.client.ts
```diff
@@ -261,18 +277,59 @@ export class SpOidcClient {
 
     const baseClient = await this.getBaseClientFromCache()
 
+    const signingKeyResult = this.getSigningKey()
+
+    if (signingKeyResult instanceof GetSigningKeyError) {
+      throw new ExchangeAuthTokenError(
+        'Failed to exchange Auth Code, no signing key found',
+      )
+    }
```
timotheeg: This reads a bit funny, why is the function returning an error instead of throwing it? 🤔 Considering nothing of the `GetSigningKeyError` is kept, `getSigningKey()` could return `null` instead when there's no key.

Also, why are we calling `this.getSigningKey()` here anyway?  `signingKeyResult` is not used in this function. If the assertion is necessary, please add a comment to indicate why.

src/app/modules/spcp/spcp.util.ts
```diff
@@ -317,14 +318,31 @@ export const getRedirectTarget = (
...
+// Typeguards
+
+export const isExtractedCorppassNDIPayload = (
+  payload: unknown,
+): payload is ExtractedCorppassNDIPayload => {
+  return (
+    typeof payload === 'object' &&
+    !!payload &&
+    hasProp(payload, 'userInfo') &&
+    hasProp(payload, 'userName') &&
```
timotheeg, line 344: These 2 lines could be removed, the type checks below are specific enough.

src/app/modules/spcp/spcp.oidc.service.ts
```diff
@@ -0,0 +1,511 @@
...
+   * Creates a payload of SP/CP user data based on attributes
+   * @param attributes user data returned by SP/CP from client
+   * @param rememberMe Whether to enable longer duration for SingPass cookies
+   * @param authType FormAuthType.SP | FormAuthType.CP
+   * @return The payload
+   */
+  createJWTPayload(
+    attributes: string | ExtractedCorppassNDIPayload,
+    rememberMe: boolean,
+    authType: FormAuthType.SP | FormAuthType.CP,
```
timotheeg: Having most (all?) methods take an `authType` and then do conditional checks doesn't feel like the cleanest OO 🤔 I wonder if `spcp.oidc.service` could export 2 services `spService` and `cPService` instead, while having an internal base class that holds the common structure.

I make the comment after seeing that the client of the service typically do their checks on the `authType` first, so it feels like we could avoid having to do the checks at multiple level. 

Admittedly, I haven't thought very deeply about how hard it'd be to refactor 😓 , but I'd be keen to try if you want to pair.

src/app/modules/spcp/spcp.oidc.client.types.ts
timotheeg: fixed in f7919072bf9508d1d36d8bd709977a4d2b036d2f

src/app/modules/spcp/spcp.oidc.client.ts
timotheeg: Fixed in f7919072bf9508d1d36d8bd709977a4d2b036d2f

src/app/modules/spcp/spcp.oidc.client.cache.ts
timotheeg: fixed in f7919072bf9508d1d36d8bd709977a4d2b036d2f

src/app/modules/spcp/spcp.oidc.service.ts
timotheeg: I think the refactor idea applies to the OIDC client class too btw. Since there are the subclasses by types already, the base class should not need to be aware of the authType. The base classes themselves should represent this awareness.

I'm going to give it a quick shot at refactoring from the 2 prior comments today for POC (also to have a sense if it's harder than I assumed here). Will let you know when done.

src/app/modules/spcp/spcp.util.ts
timotheeg: Alamak, typescript doesn't let you access an undefined property 😅 . Ignore this comment.

src/app/modules/spcp/spcp.oidc.client.ts
tshuli: Fixed in #4516

src/app/modules/spcp/spcp.oidc.client.ts
tshuli: Resolved in #4516. Also as spoken its our linter configuration that a synchronous should return (not throw) any error

timotheeg (approved):
LGTM 👍

PR #4251, opened 2022-07-20
feat(form-workspaces-be/3): add create workspace logic 

hanstirtaputra:
Builds on #4247 

Part of #4039 

## Solution
Add logic for POST `api/v3/admin/workspaces`endpoint

**Breaking Changes** 
- [x] No - this PR is backwards compatible

PR #4252, opened 2022-07-20
feat(form-workspaces-be/4): add update workspace title logic

hanstirtaputra:
Builds on #4251 

Part of #4039

## Solution
Add logic for PUT `api/v3/admin/workspaces/{workspaceId}/title` endpoint

**Breaking Changes** 

### DIFF
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -56,14 +56,12 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
   WorkspaceSchema.statics.updateWorkspaceTitle = async function ({
     title,
     workspaceId,
-    admin,
   }: {
     title: string
     workspaceId: IWorkspaceSchema['_id']
     admin: IUserSchema['_id']
   }) {
-    await this.updateOne({ _id: workspaceId, admin }, { title: title }).exec()
-    return this.findOne({ _id: workspaceId, admin }).exec()
+    return this.findOneAndUpdate({ _id: workspaceId }, { title }, { new: true })
   }
 
   return db.model<IWorkspaceSchema, IWorkspaceModel>(
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -53,6 +53,18 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     return this.create({ title, admin, formIds: [] })
   }
 
+  WorkspaceSchema.statics.updateWorkspaceTitle = async function (
+    title: string,
+    workspaceId: IWorkspaceSchema['_id'],
+    admin: IUserSchema['_id'],
+  ): Promise<WorkspaceDto | null> {
+    await this.updateOne(
+      { _id: workspaceId, admin: admin },
+      { title: title },
+    ).exec()
+    return this.findOne({ _id: workspaceId, admin: admin }).exec()
+  }
+
   return db.model<IWorkspaceSchema, IWorkspaceModel>(
     WORKSPACE_SCHEMA_ID,
     WorkspaceSchema,
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -40,6 +40,12 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     admin: 1,
   })
 
+  WorkspaceSchema.statics.getWorkspace = async function (
+    workspaceId: IWorkspaceSchema['_id'],
+  ) {
+    return this.findById(workspaceId).exec()
+  }
+
   WorkspaceSchema.statics.getWorkspaces = async function (
     admin: IUserSchema['_id'],
   ) {
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -57,12 +57,9 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     title: string,
     workspaceId: IWorkspaceSchema['_id'],
     admin: IUserSchema['_id'],
-  ): Promise<WorkspaceDto | null> {
-    await this.updateOne(
-      { _id: workspaceId, admin: admin },
-      { title: title },
-    ).exec()
-    return this.findOne({ _id: workspaceId, admin: admin }).exec()
+  ) {
+    await this.updateOne({ _id: workspaceId, admin }, { title: title }).exec()
+    return this.findOne({ _id: workspaceId, admin }).exec()
   }
 
   return db.model<IWorkspaceSchema, IWorkspaceModel>(
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -53,11 +53,15 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     return this.create({ title, admin, formIds: [] })
   }
 
-  WorkspaceSchema.statics.updateWorkspaceTitle = async function (
-    title: string,
-    workspaceId: IWorkspaceSchema['_id'],
-    admin: IUserSchema['_id'],
-  ) {
+  WorkspaceSchema.statics.updateWorkspaceTitle = async function ({
+    title,
+    workspaceId,
+    admin,
+  }: {
+    title: string
+    workspaceId: IWorkspaceSchema['_id']
+    admin: IUserSchema['_id']
+  }) {
     await this.updateOne({ _id: workspaceId, admin }, { title: title }).exec()
     return this.findOne({ _id: workspaceId, admin }).exec()
   }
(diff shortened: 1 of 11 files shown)


src/app/models/workspace.server.model.ts
```diff
@@ -93,6 +93,18 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+  WorkspaceSchema.statics.updateWorkspaceTitle = async function (
+    title: string,
+    workspaceId: IWorkspaceSchema['_id'],
+    admin: IUserSchema['_id'],
+  ): Promise<WorkspaceDto | null> {
+    await this.updateOne(
+      { _id: workspaceId, admin: admin },
+      { title: title },
+    ).exec()
+    return this.findOne({ _id: workspaceId, admin: admin }).exec()
```
karrui: nit: 
```suggestion
    return this.findOne({ _id: workspaceId, admin }).exec()
```

src/app/models/workspace.server.model.ts
```diff
@@ -93,6 +93,18 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     return this.create({ title, admin, formIds: [] })
   }
 
+  WorkspaceSchema.statics.updateWorkspaceTitle = async function (
+    title: string,
+    workspaceId: IWorkspaceSchema['_id'],
+    admin: IUserSchema['_id'],
+  ): Promise<WorkspaceDto | null> {
+    await this.updateOne(
+      { _id: workspaceId, admin: admin },
```
karrui: nit: 
```suggestion
      { _id: workspaceId, admin },
```

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
```diff
@@ -157,4 +159,91 @@ describe('workspaces.routes', () => {
...
+        buildCelebrateError({
+          body: {
+            key: 'title',
+            message: '"title" length must be at least 4 characters long',
+          },
+        }),
+      )
+    })
+
+    it('should return 401 when user is not logged in', async () => {
```
karrui, line 231: should we also add a test case for when the user does not have permissions to update the workspace title?

src/app/modules/workspace/workspace.controller.ts
```diff
@@ -98,24 +94,37 @@ export const createWorkspace = [
...
+  { workspaceId: string },
+  WorkspaceDto | ErrorDto,
   { title: string }
 > = async (req, res) => {
+  const { workspaceId } = req.params
   const { title } = req.body
-  return WorkspaceService.updateWorkspaceTitle('', title)
+  const userId = (req.session as AuthedSessionData).user._id
+
+  return WorkspaceService.updateWorkspaceTitle(workspaceId, title, userId)
```
karrui: should we also return an appropriate error instead of `res.json(null)` when the user does not own a workspace with `workspaceId`?

src/app/models/workspace.server.model.ts
```diff
@@ -93,6 +93,18 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     return this.create({ title, admin, formIds: [] })
   }
 
+  WorkspaceSchema.statics.updateWorkspaceTitle = async function (
+    title: string,
+    workspaceId: IWorkspaceSchema['_id'],
+    admin: IUserSchema['_id'],
```
karrui: nit: should we use an object arg instead since it seems increasingly likely that there will be more and more arg parameters? would be less likely to make mistakes when passing in params due to similar types (i know some types have been branded but imagine if all 3 arg types were of `string` type)

mantariksh (commented):
same comments as in prev PRs, e.g. splitting database errors and testing each case. will wait for this to be updated before reviewing again!

src/app/modules/workspace/__tests__/workspace.routes.spec.ts
hanstirtaputra: Added test for this!

src/app/modules/workspace/workspace.controller.ts
hanstirtaputra: Yep! Updated to return an error instead

src/app/models/workspace.server.model.ts
```diff
@@ -53,6 +53,19 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+    title,
+    workspaceId,
+    admin,
+  }: {
+    title: string
+    workspaceId: IWorkspaceSchema['_id']
+    admin: IUserSchema['_id']
+  }) {
+    await this.updateOne({ _id: workspaceId, admin }, { title: title }).exec()
+    return this.findOne({ _id: workspaceId, admin }).exec()
```
mantariksh: rather than using 2 db calls, just use `findOneAndUpdate` with the `new` argument which makes `mongoose` return the new document.

in a situation where you think a library is doing something unintuitive (like forcing you to make 2 db calls to update a document and get back the updated version), there's probably a better way to do things!

mantariksh: also why do we need to pass the `admin`? the `workspaceId` should be sufficient right?

hanstirtaputra: Okay will update to use `findOneAndUpdate`! 

Yeah `workspaceId` is sufficient. I thought is just "defensive" so that users can't update workspace that don't belong to them (though this check is already done in the controller)

hanstirtaputra: I'll remove the `admin` from the filter query!

mantariksh: hmm okay. the way I think about this is that by the time the query reaches the model layer, those business logic checks should already have been done. but then again we have such checks in the model validation layer...up to you!

src/app/models/workspace.server.model.ts
```diff
@@ -53,6 +59,17 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
     return this.create({ title, admin, formIds: [] })
   }
 
+  WorkspaceSchema.statics.updateWorkspaceTitle = async function ({
+    title,
+    workspaceId,
+  }: {
+    title: string
+    workspaceId: IWorkspaceSchema['_id']
+    admin: IUserSchema['_id']
```
mantariksh: this argument is no longer required

src/app/modules/workspace/workspace.utils.ts
```diff
@@ -39,6 +44,16 @@ export const mapRouteError = (
...
       }
+    case WorkspaceNotFoundError:
+      return {
+        statusCode: StatusCodes.NOT_FOUND,
+        errorMessage: errorMessage,
+      }
+    case ForbiddenWorkspaceError:
+      return {
+        statusCode: StatusCodes.FORBIDDEN,
+        errorMessage: errorMessage,
```
mantariksh:
```suggestion
        errorMessage,
      }
    case WorkspaceNotFoundError:
      return {
        statusCode: StatusCodes.NOT_FOUND,
        errorMessage,
      }
    case ForbiddenWorkspaceError:
      return {
        statusCode: StatusCodes.FORBIDDEN,
        errorMessage,
```

src/app/modules/workspace/workspace.controller.ts
```diff
@@ -99,25 +95,53 @@ export const createWorkspace = [
...
-  return WorkspaceService.updateWorkspaceTitle('', title)
-    .map((workspace) => res.status(StatusCodes.OK).json(workspace))
-    .mapErr((err) =>
-      res.status(StatusCodes.BAD_REQUEST).json({ message: err.message }),
+  const userId = (req.session as AuthedSessionData).user._id
+
+  return WorkspaceService.getWorkspace(workspaceId)
+    .andThen((workspace) =>
+      WorkspaceService.verifyWorkspaceAdmin(workspace, userId),
+    )
```
mantariksh, line 115: wow actually I realised we can cut out the above two steps entirely by just moving the admin check to the model layer? interesting. on one hand it would be one less call to the db, on the other hand separating the check is probably more maintainable if we have more checks in future. will leave it up to you!

src/app/modules/workspace/workspace.service.ts
```diff
@@ -91,3 +117,35 @@ export const moveForms = (
...
+  ).andThen((workspace: WorkspaceDto) =>
+    workspace ? okAsync(workspace) : errAsync(new WorkspaceNotFoundError()),
+  )
+}
+
+export const verifyWorkspaceAdmin = (
+  workspace: WorkspaceDto,
+  userId: string,
+): ResultAsync<true, ForbiddenWorkspaceError> => {
+  if (workspace.admin != userId) {
```
mantariksh:
```suggestion
  if (workspace.admin !== userId) {
```

src/app/models/workspace.server.model.ts
hanstirtaputra: > the way I think about this is that by the time the query reaches the model layer, those business logic checks should already have been done

Yep agreed, I've removed `admin` check from the model!

src/app/modules/workspace/workspace.controller.ts
hanstirtaputra: I think we can separate the checks! The other benefit is that we will know whether the error is due to `workspaceId` not found or a permissions issue (user is not admin of an existing workspace)

mantariksh (approved):

PR #4254, opened 2022-07-20
feat(form-workspaces-be/5): add delete workspace logic

hanstirtaputra:
Builds on #4252 

Part of #4039 

## Solution
Add logic for DELETE `api/v3/admin/workspaces/{workspaceId}` endpoint

**Breaking Changes** 

### DIFF
--- a/src/app/models/__tests__/workspace.server.model.spec.ts
+++ b/src/app/models/__tests__/workspace.server.model.spec.ts
@@ -208,7 +208,6 @@ describe('Workspace Model', () => {
       it('workspace should no loger be in database when deleted', async () => {
         await Workspace.deleteWorkspace({
           workspaceId: MOCK_WORKSPACE_ID,
-          admin: MOCK_USER_ID,
         })
 
         const actual = await Workspace.exists({ _id: MOCK_WORKSPACE_ID })
@@ -222,7 +221,6 @@ describe('Workspace Model', () => {
         )
         await Workspace.deleteWorkspace({
           workspaceId: MOCK_WORKSPACE_ID,
-          admin: MOCK_USER_ID,
         })
 
         const actual = await Workspace.exists({ _id: MOCK_WORKSPACE_ID })
--- a/src/app/models/__tests__/workspace.server.model.spec.ts
+++ b/src/app/models/__tests__/workspace.server.model.spec.ts
@@ -1,11 +1,14 @@
 import { ObjectId } from 'bson-ext'
 import { omit } from 'lodash'
 import mongoose from 'mongoose'
+import { FormStatus } from 'shared/types'
 
 import { getWorkspaceModel } from 'src/app/models/workspace.server.model'
 import { IUserSchema } from 'src/types'
 
 import dbHandler from 'tests/unit/backend/helpers/jest-db'
+
+import getFormModel from '../form.server.model'
 
 const Workspace = getWorkspaceModel(mongoose)
 
@@ -197,6 +200,101 @@ describe('Workspace Model', () => {
         })
 
         expect(actual).toBeNull()
+      })
+    })
+
+    describe('deleteWorkspace', () => {
+      it('should return 1 upon successful workspace deletion', async () => {
+        const shouldDeleteForms = true
+        const actual = await Workspace.deleteWorkspace(
+          MOCK_WORKSPACE_ID,
+          MOCK_USER_ID,
+          shouldDeleteForms,
+        )
+
+        expect(actual).toEqual(1)
+      })
+
+      it('forms in workspace should be archived upon successful workspace deletion', async () => {
+        const shouldDeleteForms = true
+        const FormModel = await getFormModel(mongoose)
+
+        const workspaceId = new ObjectId()
+        const workspaceObject = {
+          _id: workspaceId,
+          title: 'Workspace2',
+          admin: MOCK_USER_ID,
+          formIds: [MOCK_FORM_ID],
+        }
+        const validWorkspace = new Workspace(workspaceObject)
+        await validWorkspace.save()
+
+        const privateForm = await FormModel.findById(MOCK_FORM_ID)
+        expect(privateForm?.status).toEqual(FormStatus.Private)
+
+        const actual = await Workspace.deleteWorkspace(
+          workspaceObject._id,
+          workspaceObject.admin,
+          shouldDeleteForms,
+        )
+        expect(actual).toEqual(1)
+
+        const archivedForm = await FormModel.findById(MOCK_FORM_ID)
+        expect(archivedForm?.status).toEqual(FormStatus.Archived)
+      })
+
+      it('forms in workspace should not be archived upon successful workspace deletion', async () => {
+        const shouldDeleteForms = false
+        const FormModel = await getFormModel(mongoose)
+
+        const workspaceId = new ObjectId()
+        const workspaceObject = {
+          _id: workspaceId,
+          title: 'Workspace2',
+          admin: MOCK_USER_ID,
+          formIds: [MOCK_FORM_ID],
+        }
+        const validWorkspace = new Workspace(workspaceObject)
+        await validWorkspace.save()
+
+        const privateForm = await FormModel.findById(MOCK_FORM_ID)
+        expect(privateForm?.status).toEqual(FormStatus.Private)
+
+        const actual = await Workspace.deleteWorkspace(
+          workspaceObject._id,
+          workspaceObject.admin,
+          shouldDeleteForms,
+        )
+        expect(actual).toEqual(1)
+
+        const archivedForm = await FormModel.findById(MOCK_FORM_ID)
+        expect(archivedForm?.status).toEqual(FormStatus.Private)
+      })
+
+      it('should return 0 upon unsuccessful delete due to invalid workspace id', async () => {
+        const shouldDeleteForms = true
+        const invalidWorkspaceId = new ObjectId()
+
+        const actual = await Workspace.deleteWorkspace(
+          invalidWorkspaceId,
+          MOCK_USER_ID,
+          shouldDeleteForms,
+        )
+
+        expect(actual).toEqual(0)
+      })
+
+      it('should return 0 upon unsuccessful delete due to admin not owning workspace', async () => {
+        const shouldDeleteForms = true
+        const invalidUserId = new ObjectId()
+
+        const actual = await Workspace.deleteWorkspace(
+          MOCK_WORKSPACE_ID,
+          invalidUserId,
+          shouldDeleteForms,
+        )
+
+        expect(actual).toEqual(0)
       })
     })
   })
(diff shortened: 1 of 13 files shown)


src/app/modules/workspace/workspace.service.ts
```diff
@@ -80,12 +81,28 @@ export const updateWorkspaceTitle = (
 
 export const deleteWorkspace = (
```
mantariksh: would be good to use an object arg here as well

src/app/models/workspace.server.model.ts
```diff
@@ -70,6 +75,34 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+      admin: admin,
+    })
+    if (shouldDeleteForms) {
+      await Form.updateMany(
+        { _id: { $in: workspaceToDelete?.formIds } },
+        { $set: { status: FormStatus.Archived } },
+      )
+    }
+    await session.commitTransaction()
+    session.endSession()
```
mantariksh: this logic belongs in the service layer above, not in the model layer.

model methods are meant to deal with operations on that model itself, without knowledge of the business logic. services are meant to coordinate changes to the models according to the business logic.

if you check the other usage of transactions in the codebase (`createTwilioCredentials`) you can see that it takes place in the service layer as well.

hanstirtaputra: Updated to move the transaction logic to the service layer!

src/types/workspace.ts
```diff
@@ -29,4 +29,12 @@ export interface IWorkspaceModel extends Model<IWorkspaceSchema> {
     title: string
     workspaceId: IWorkspaceSchema['_id']
   }): Promise<WorkspaceDto | null>
+
+  deleteWorkspace({
+    workspaceId,
+    session,
+  }: {
+    workspaceId: IWorkspaceSchema['_id']
+    session?: ClientSession
```
mantariksh, line 43: if we expect this operation to always take place within a transaction, we should make this mandatory

src/app/modules/workspace/workspace.service.ts
```diff
@@ -83,14 +84,65 @@ export const updateWorkspaceTitle = ({
...
+  shouldDeleteForms,
+  session,
+}: {
+  workspaceId: string
+  shouldDeleteForms: boolean
+  session: ClientSession
+}): Promise<void> => {
+  const workspaceToDelete = await WorkspaceModel.findOne({
+    _id: workspaceId,
+  })
```
mantariksh: why do we retrieve the whole document even when `shouldDeleteForms` is false?

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1498,3 +1498,17 @@ const deleteTwilioTransaction = async (
...
+  session,
+}: {
+  formIds: string[]
+  session?: ClientSession
+}): Promise<void> => {
+  const canBeArchivedForms = await FormModel.find({
+    _id: { $in: formIds },
+  })
+
+  canBeArchivedForms.forEach(async (form) => await form.archive(session))
```
mantariksh: rather than archive each form in an independent async call, we should issue one database query to archive multiple form IDs. that's what you did when profiling the time taken for this query, correct?

src/app/models/__tests__/workspace.server.model.spec.ts
```diff
@@ -199,5 +203,37 @@ describe('Workspace Model', () => {
         expect(actual).toBeNull()
       })
     })
+
+    describe('deleteWorkspace', () => {
+      it('workspace should no loger be in database when deleted', async () => {
```
mantariksh: test descriptions normally use the "should...when" pattern, e.g. "should correctly remove workspace from database when deleted"

src/app/models/workspace.server.model.ts
```diff
@@ -70,6 +70,22 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+    workspaceId: IWorkspaceSchema['_id']
+    session?: ClientSession
+  }) {
+    const deleted = await this.deleteOne(
+      {
+        _id: workspaceId,
+      },
+      { session },
+    )
+    return deleted.deletedCount == 1
```
mantariksh:
```suggestion
    return deleted.deletedCount === 1
```

src/app/models/workspace.server.model.ts
```diff
@@ -70,6 +70,22 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
   }
 
+  WorkspaceSchema.statics.deleteWorkspace = async function ({
+    workspaceId,
+    session,
+  }: {
+    workspaceId: IWorkspaceSchema['_id']
+    session?: ClientSession
+  }) {
+    const deleted = await this.deleteOne(
```
mantariksh: why not use `findOneAndRemove` so we can return the deleted document, which is more informative for the caller?

src/app/modules/workspace/__tests__/workspace.service.spec.ts
```diff
@@ -275,4 +278,181 @@ describe('workspace.service', () => {
...
+    const mockFormId = new ObjectId()
+    const mockWorkspace = {
+      _id: new ObjectId(),
+      admin: new ObjectId(),
+      title: 'workspace1',
+      formIds: [mockFormId],
+      count: 0,
+    }
+
+    it('should return undefined when successfully deleted workspace', async () => {
```
mantariksh: usually we should think carefully about returning `undefined`, as it's an uninformative response type. one alternative could be to use `findOneAndRemove` in the db layer so as to be able to at least return the deleted document.

src/app/modules/workspace/workspace.service.ts
```diff
@@ -83,14 +84,65 @@ export const updateWorkspaceTitle = ({
...
-  return okAsync({
-    workspaceId: workspaceId,
-    shouldDeleteForms: shouldDeleteForms,
+export const deleteWorkspace = ({
+  workspaceId,
+  shouldDeleteForms,
+}: {
+  workspaceId: string
+  shouldDeleteForms: boolean
+}): ResultAsync<void, DatabaseError> => {
```
mantariksh: this should be `PossibleDatabaseError` as per the return type of `transformMongoError` right?

src/app/modules/workspace/__tests__/workspace.service.spec.ts
```diff
@@ -275,4 +278,181 @@ describe('workspace.service', () => {
...
+      const actual = await WorkspaceService.deleteWorkspace({
+        workspaceId: mockWorkspace._id.toHexString(),
+        shouldDeleteForms: false,
+      })
+
+      expect(actual.isOk()).toEqual(true)
+      expect(actual._unsafeUnwrap()).toEqual(undefined)
+    })
+
+    it('should return undefined when failed to delete workspace', async () => {
```
mantariksh: this illustrates the previous comment - we return the same thing regardless of the outcome of the deletion

src/app/modules/workspace/__tests__/workspace.service.spec.ts
```diff
@@ -275,4 +278,181 @@ describe('workspace.service', () => {
...
+
+      const actual = await WorkspaceService.deleteWorkspace({
+        workspaceId: mockWorkspace._id.toHexString(),
+        shouldDeleteForms: false,
+      })
+      const doesFormExist = await FormModel.exists({ _id: mockFormId })
+      const isFormArchived = await FormModel.exists({
+        _id: mockFormId,
+        status: FormStatus.Archived,
+      })
```
mantariksh: instead of making two async calls, we can just retrieve the form once and check that it is not null and that status is not archived

src/app/modules/workspace/__tests__/workspace.service.spec.ts
```diff
@@ -275,4 +278,181 @@ describe('workspace.service', () => {
...
+
+      expect(actual.isErr()).toEqual(true)
+      expect(actual._unsafeUnwrapErr()).toBeInstanceOf(DatabaseError)
+      expect(doesWorkspaceExist).toEqual(true)
+      expect(doesFormExist).toEqual(true)
+      expect(isFormArchived).toEqual(false)
+    })
+
+    it('should not archive forms or delete workspace when transaction fails at forms deletion', async () => {
```
mantariksh: well done on covering all these cases!

mantariksh:
how do we ensure that when the user opts to delete the forms within a workspace, only the forms that they own are deleted?

src/types/workspace.ts
hanstirtaputra: Currently the test environment database doesn't support transactions, so I made the session optional so that we can mock the implementation of this function in the test environment (to run without transaction).

I've added a comment to this to make it clearer!

Can see issue #4503 for details

src/app/modules/workspace/workspace.service.ts
hanstirtaputra: Updated this to use `findOneAndRemove` instead so we get the whole document when we delete a workspace

src/app/modules/form/admin-form/admin-form.service.ts
hanstirtaputra: > that's what you did when profiling the time taken for this query, correct?

Yes that's what I did.

I updated the implementation to use one db query!

src/app/modules/workspace/__tests__/workspace.service.spec.ts
hanstirtaputra: Updated to use `findOneAndRemove`!

src/app/modules/workspace/workspace.service.ts
hanstirtaputra: Yes, updated the type to `PossibleDatabaseError`

src/app/modules/workspace/__tests__/workspace.service.spec.ts
hanstirtaputra: Updated to do this instead! Thanks

hanstirtaputra:
> how do we ensure that when the user opts to delete the forms within a workspace, only the forms that they own are deleted?

Added a check in AdminFormService's `archiveForms` function to only archive forms the user is an admin of

mantariksh:
closing this to clean up the open PRs until we revisit it!

src/types/workspace.ts
mantariksh: hmm, we shouldn't compromise on type safety in production code. for the dev taking over this - is there a better way to do this, e.g. by creating a mock session object to use for tests?

foochifa: Agreed on this - and we might be able to use our current `mongodb-memory-server-core` [package](https://github.com/nodkz/mongodb-memory-server#replica-set-start) to create a replica set on our DB. Which may help us test for transaction.

However, this branch is too deprecated, will merge it into `form-v2/form-workspaces` first and work on this issue later. 

FRM-976

foochifa:
merging this to feature branch to rebase to current stack. Thanks @hanstirtaputra! Will also follow up on #4448

Issue #3630, opened 2022-03-22
[React/PublicForm] Add prefilled fields support for public form


karrui:
Two subtasks:
- [ ] Unauthed form prefill (will be closed by #4267)
- [ ] Authed form prefill after redirection

karrui:
@tshuli assigning you this issue since you implemented it in AngularJS; thought this would help see the difference between React and AngularJS!

tshuli:
yay more prefill :D

karrui:
Closed by #4267 and #4275

PR #4267, opened 2022-07-21
feat(v2): add ability to pre-fill public forms fields

karrui:
## Problem

This PR adds the implementation for allowing form fields to be prefilled. Note that SPCP authentication prefill will come in a later PR. 

Related to #3630, not closing since there is still the auth variant to implement.

## Solution

**Breaking Changes** 

**Features**:

- feat: add prefilled fields to default field values if they exist
- feat: add prefilled field banner if any fields are prefilled

## Before & After Screenshots
feat: add prefilled public form story

### DIFF
--- a/frontend/src/features/public-form/components/FormFields/FieldFactory.tsx
+++ b/frontend/src/features/public-form/components/FormFields/FieldFactory.tsx
@@ -38,68 +38,69 @@ import {
 
 interface FieldFactoryProps {
   field: FormFieldWithQuestionNo
+  isPrefilled?: boolean
   colorTheme?: FormColorTheme
 }
 
 export const FieldFactory = memo(
-  ({ field, colorTheme }: FieldFactoryProps) => {
+  ({ field, ...rest }: FieldFactoryProps) => {
     switch (field.fieldType) {
       case BasicField.Section:
-        return <SectionField schema={field} colorTheme={colorTheme} />
+        return <SectionField schema={field} {...rest} />
       case BasicField.Checkbox:
-        return <CheckboxField schema={field} colorTheme={colorTheme} />
+        return <CheckboxField schema={field} {...rest} />
       case BasicField.Radio:
-        return <RadioField schema={field} colorTheme={colorTheme} />
+        return <RadioField schema={field} {...rest} />
       case BasicField.Nric:
-        return <NricField schema={field} colorTheme={colorTheme} />
+        return <NricField schema={field} {...rest} />
       case BasicField.Number:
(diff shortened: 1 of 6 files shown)


frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -31,18 +33,44 @@ export const FormFields = ({
...
+    return formFields.reduce((acc, field) => {
+      if (
+        field.fieldType === BasicField.ShortText &&
+        field.allowPrefill &&
+        ids.has(field._id)
+      ) {
+        acc[field._id] = queryParams[field._id]
+      }
+      return acc
+    }, {} as Record<string, string>)
```
timotheeg: There's no need to parse the full searchParams and make local structures for them, [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) already has the APIs to do the job:
```suggestion
    // Return object containing field id and query param value only if id exists in form fields.
    return formFields.reduce((acc, field) => {
      if (
        field.fieldType === BasicField.ShortText &&
        field.allowPrefill &&
        searchParams.has(field._id)
      ) {
        acc[field._id] = searchParams.get(field._id)
      }
      return acc
    }, {} as Record<string, string>)
```

frontend/src/features/public-form/components/FormFields/FieldFactory.tsx
```diff
@@ -38,68 +38,69 @@ import {
 
 interface FieldFactoryProps {
   field: FormFieldWithQuestionNo
+  isPrefilled?: boolean
   colorTheme?: FormColorTheme
 }
 
 export const FieldFactory = memo(
-  ({ field, colorTheme }: FieldFactoryProps) => {
+  ({ field, ...rest }: FieldFactoryProps) => {
```
timotheeg, line 46: Just checking: does this syntax mean typescript will only inject `isPrefilled` and `colorTheme` into `rest`, or will any other supplied props also go through (e.g. `key`, which was previously ignored)? Assuming they do go through, any issue with letting `key` propagate?

karrui: technically anything can go in, but FieldFactoryProps will constrain the props to still just be `isPrefilled` and `colorTheme`

frontend/src/features/public-form/components/FormFields/FormFields.tsx
karrui: resolved in 5c33a5a

timotheeg (approved):
LGTM

Issue #4236, opened 2022-07-19
[React] Display instructions for form


justynoh:
Currently, instructions are not displayed. They should be displayed both in the admin builder and the public form. See Figma tab "Create-Design".

PR #4271, opened 2022-07-22
feat(form-v2/design-2): logo radio buttons and form color theme selection

justynoh:
## Problem
This PR implements the form logo uploader tool and color theme selection in the design drawer.

Makes progress towards #3362 and provides parity with angular.

## Solution

**Breaking Changes** 

**Features**:

`FileHandlerService.ts`, `EditImage.tsx`, `UploadImageInput.tsx`
- Updated type of `UploadedFileData.size` in `FileHandlerService.ts` from `string` to `number`. As a result, moved stringification of size into `EditImage.tsx` itself. (Necessary due to `form.startPage.logo.fileSizeInBytes` being of type `number`.)
- Separated `attachment` type into its own `UploadedImage` type for reusability with design store.

`useFormHeader.tsx`
- Separated color theme computation into a new function `getTitleBg` for reusability with design store.

`StartPageView.tsx`, `DesignDrawer.tsx`, `useDesignStore.tsx`
- Update design store types to include `customLogoMeta`.
- Add logo and color fields to design drawer.
- Start page now shows uploaded custom logo preview.
 
## Before & After Screenshots

image

## Notes

- Instructions still not shown in form itself - see issue [#4236]
- There is currently still a bug with edit image field and design drawer logo uploads: 1) blob preview returns error due to CSP, and 2) after the field is saved and image uploaded to S3, the image is not loaded in builder due to 403 when fetching from S3 (as tested on staging).


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditImage/UploadImageInput.tsx
```diff
@@ -13,6 +13,11 @@ import Attachment from '~components/Field/Attachment'
 
 import { EditImageInputs } from './EditImage'
 
+export type UploadedImage = {
+  file?: File
+  srcUrl?: string
+}
+
```
timotheeg, line 20: I wonder if this type is descriptive enough 🤔? The definition allows a completely empty object, but is that really valid? Or is `UploadedImage` a union of 2 types, where at least one of the field is present in each?

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
```diff
@@ -2,26 +2,66 @@ import { isEqual } from 'lodash'
...
+    setStartPageData: (startPageData: FormStartPageInput) => {
+      const current = get()
+      if (isEqual(current.startPageData, startPageData)) return
+      set({ startPageData })
+    },
+    setAttachment: (attachment: UploadedImage) => {
       const current = get()
-      if (isEqual(current.startPageData, startPage)) return
+      if (!current.startPageData) return
+      if (isEqual(current.startPageData?.attachment, attachment)) return
```
timotheeg: /nit: since `current.startPageData` is checked above, the conditional access is not needed here:
```suggestion
      if (isEqual(current.startPageData.attachment, attachment)) return
```

justynoh: You're right! thanks for catching that

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditImage/UploadImageInput.tsx
justynoh: Should we can fix this in a separate PR? The type was originally `Partial<{ file: File srcUrl: ImageFieldBase['url'] }>` which this is equivalent to. But I agree that this type should be more restrictive. This fix would require edits to the startpage and editimage.

timotheeg (approved):
lgtm

PR #4272, opened 2022-07-22
feat: edit dateField schema

### LINKED ISSUE #4136: [React/Date Validation Extension/Restricting Particular Days of the Week] Add additional attribute in dateField

siddarth2824:
## Problem

There was an issue where checkbox group would serialise the values and store it as a `String` which created inconsistencies as the current `invalidDaysOfTheWeek` array in the schema was being stored as a `Number` array.

closes #4136 

## Solution

 Edit `dateField` schema to store invalidDaysOfTheWeek as a `String` enum array instead of a `Number` enum array

**Breaking Changes** 

   - Details:
      - There might be a chance that the database might contain an array that does not match the same type as the current type of `invalidDaysOfTheWeek`. Hence this PR might cause a break in the database due to the mismatch in the attribute's type.

### DIFF
--- a/shared/types/field/dateField.ts
+++ b/shared/types/field/dateField.ts
@@ -8,13 +8,13 @@ export enum DateSelectedValidation {
 }
 
 export enum DaysOfTheWeek {
-  Sunday = 0,
-  Monday = 1,
-  Tuesday = 2,
-  Wednesday = 3,
-  Thursday = 4,
-  Friday = 5,
-  Saturday = 6,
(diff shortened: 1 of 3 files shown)


karrui (approved):
lgtm. technically not backwards compatible but nbd since this is not released yet

Issue #4246, opened 2022-07-20
Singpass authentication options are missing tooltips and additional info


mantariksh:
Authentication options in settings page should have the same tooltips as in AngularJS, as well as the "Contact Us" link for SGID.

justynoh:
is sgID login supposed to be disabled by default (like in the angular app)?

may be related to this todo in `frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx`

<img width="487" alt="image" src="https://user-images.githubusercontent.com/25571626/180148707-5d5e1d47-c673-4663-abef-d1f938d19db5.png">

mantariksh:
@justynoh yes, supposed to be disabled unless user has `betaFlags.sgid` enabled. can cross check with Angular app

PR #4274, opened 2022-07-22
fix(form-v2): disable sgid auth option in settings for non-beta users

justynoh:
## Problem
Currently, the SGID auth option is open for all to select. We need to disable it for non-beta users.

Closes [#4246 ]

## Solution

- `isDisabled` is now a function that takes the `authType` as an input, which checks for `betaFlags.sgid` when the `authtype` is `FormAuthType.SGID`.
- Add the `authType` input to all calls to `isDisabled`.

**Breaking Changes** 

## Screenshots

image

## Tests
- [X] Non-beta testers should not be able to select the SGID option.

### DIFF
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
@@ -16,7 +16,7 @@ import Link from '~components/Link'
 import Radio from '~components/Radio'
 import Tooltip from '~components/Tooltip'
 
-import { useAdminForm } from '~features/admin-form/common/queries'
+import { useUser } from '~features/user/queries'
 
 import { useMutateFormSettings } from '../../mutations'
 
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
```diff
@@ -48,6 +50,7 @@ export const AuthSettingsSection = ({
   settings,
 }: AuthSettingsSectionProps): JSX.Element => {
   const { mutateFormAuthType } = useMutateFormSettings()
+  const { data: form } = useAdminForm()
```
karrui: might want to retrieve beta flags from the user object itself (from `useUser` hook over `useAdminForm`

justynoh: ooo you're right. Thanks!

karrui (approved):
lgtm

PR #3887, opened 2022-05-20
Instrumentation with DD in react branch

timotheeg:
## Problem
We need to have deep instrumentation, tracing, and RUM prior to react launch

We don't want to setup DataDog RUM into angular, we can maintain sentry for that, but datadog RUM will be setup for React.

dd-trace is active, to work in containers, the EB environment **must** have the environment variable `DD_AGENT_HOST`.

## Solution
Activate datadog for everything we need:
* RUM on frontend
* APM on server
* Statsd

## References:
* [DD RUM intro and setup](https://docs.datadoghq.com/real_user_monitoring/browser/)
* [DD APM intro and setup for nodejs](https://docs.datadoghq.com/tracing/setup_overview/setup/nodejs/?tab=containers)

PR #4294, opened 2022-07-25
fix: add 'self' to workerSrc to allow storage mode downloads

karrui:
`helmet.workerSrc` was updated in #3887. However, this removed the default of `'self'` and storage mode downloads stopped working (due to disallowing web workers).

This PR fixes that bug.


karrui:
Tested on staging, it works now

timotheeg (approved):

Issue #4302, opened 2022-07-25
Browser title does not match tab content


mantariksh:
To replicate:
- Preview a form
- Click on "Edit Form" to go back to the builder
- Go back to workspaces
- Browser title will remain on form title

PR #4319, opened 2022-07-25
fix(form-v2): set document title on dismount

tohhengxing:
## Problem
Browser title does not match tab content after leaving form preview page

Closes [#4302]

## Solution
set document.title to FormSG after component "dismount" using useEffect. There is probably a way to do it using Helmet but I think that would introduce additional complexity for no beneficial reason.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
<details>

https://user-images.githubusercontent.com/102740235/180798440-39b21e67-62db-460e-9a86-f742802317cd.mp4
</details>

**AFTER**:
<details>

https://user-images.githubusercontent.com/102740235/180798502-4a74b1a2-f6e3-4fd2-91e2-2562368d6341.mp4
</details>


timotheeg (approved):
lgtm

Issue #4303, opened 2022-07-25
Feature tour does not move on to design/logic


mantariksh:
To replicate:
- Clear feature tour key in local storage
- Open a form
- Click "Next" in Build tab
- Feature tour closes

karrui:
closed by #4320, #4388

PR #4320, opened 2022-07-25
fix(form-v2) : remove invalid conditional to fix feature tour bug

tohhengxing:
* joyride action value for next button is 'close' because continuous prop is false for tooltip

## Problem

Feature tour does not move on to design/logic

Closes [#4303 ]

## Solution

If continuous prop for joyride is false, then the default value for next button action is 'close'. 

Original code used a condition `ACTION.close` to determine whether to set local storage key to true, but this will lead to buggy behaviour as the next button has value `ACTION.close` which will incorrectly trigger the `onClose` function in `handleJoyrideCallback`

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
<details>

https://user-images.githubusercontent.com/102740235/180828894-4a8a4df5-d662-40f9-b04e-935ba4fae88f.mp4
</details>

**AFTER**:

<details>

https://user-images.githubusercontent.com/102740235/180828971-766e1897-7ed9-4f0c-9fc1-d54b521a7048.mp4
</details>

**ADDITIONAL ISSUES DISCOVERED NOT IN SCOPE OF THIS TASK**
the html title is wrong for the joyride tooltip for all buttons
<details>
<summary>picture of invalid titles</summary>

image
</details>

### DIFF
--- a/frontend/src/features/admin-form/create/featureTour/FeatureTour.tsx
+++ b/frontend/src/features/admin-form/create/featureTour/FeatureTour.tsx
@@ -25,12 +25,7 @@ export const FeatureTour = ({ onClose }: FeatureTourProps): JSX.Element => {
       if (type === EVENTS.STEP_AFTER || type === EVENTS.TARGET_NOT_FOUND) {
         setStepIndex(index + 1)
       }
-
-      if (
-        status === STATUS.FINISHED ||
-        status === STATUS.SKIPPED ||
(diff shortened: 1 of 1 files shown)


karrui (approved):
Lgtm

karrui:
Thanks for the catch on invalid title prop. Will create a new issue.

PR #4323, opened 2022-07-26
feat(form-v2/form-instructions-2): pull PublicSwitchEnvMessage into public form wrapper

justynoh:
## Problem
The react message currently shows up immediately before the form fields. However, we want this to be at the top of the form, immediately after the header. Moving it into the public form wrapper allows the message, form instructions and form fields to be displayed in the correct order. 

Makes progress towards [#4236 ]

## Solution

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/public-form/components/PublicSwitchEnvMessage.tsx
+++ b/frontend/src/features/public-form/components/PublicSwitchEnvMessage.tsx
@@ -1,5 +1,5 @@
 // TODO #4279: Remove after React rollout is complete
-import { Text } from '@chakra-ui/react'
+import { Box, Flex, Text } from '@chakra-ui/react'
 
 import Button from '~components/Button'
 import InlineMessage from '~components/InlineMessage'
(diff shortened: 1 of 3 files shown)


wanlingt (approved):
lgtm!

frontend/src/features/public-form/components/PublicSwitchEnvMessage.tsx
```diff
@@ -10,14 +10,25 @@ export const PublicSwitchEnvMessage = (): JSX.Element => {
   const { publicSwitchEnvMutation } = useEnvMutations()
 
   return (
-    <InlineMessage variant="warning" mb="1.5rem" mt={{ base: '2rem', md: '0' }}>
-      <Text>
-        You’re filling this form on the new FormSG. If you have trouble
-        submitting,{' '}
```
timotheeg, line 16: Just checking: explicit space wasn't needed?

timotheeg (approved):
lgtm

frontend/src/features/public-form/components/PublicSwitchEnvMessage.tsx
wanlingt: oops that was my bad! I'll change it in my [react-to-angular PR](https://github.com/opengovsg/FormSG/pull/4352) since it should apply to both `PublicSwitchEnvMessage` and `AdminSwitchEnvMessage` :)

PR #4324, opened 2022-07-26
feat: edit dateField schema attribute name

siddarth2824:
## Problem

According to the Figma design, form admin users should be able to restrict Singapore public holidays on top of days of the week. Currently, the dateField schema attribute to encapsulate that information is named `invalidDaysOfTheWeek` which is not accurate since form admin users can restrict Singapore public holidays as well. 

Furthermore, since the Figma design shows that on default all the options will be checked, the default value of the attribute should be a string array containing all the options values instead of an empty array.

## Solution

Change the attribute in dateField to be `validDays` instead of `invalidDaysOfTheWeek`. Set the default value of the attribute to be a string array containing all the options values.

**Breaking Changes** 

    - Due to the change in one of the attributes in dateField schema, rolling back this PR might cause inconsistencies in the database values.

### DIFF
--- a/shared/types/field/dateField.ts
+++ b/shared/types/field/dateField.ts
@@ -7,7 +7,7 @@ export enum DateSelectedValidation {
   Custom = 'Custom date range',
 }
 
-export enum DaysOfTheWeek {
+export enum ValidDaysOptions {
   Sunday = 'Sunday',
   Monday = 'Monday',
   Tuesday = 'Tuesday',
@@ -15,6 +15,7 @@ export enum DaysOfTheWeek {
   Thursday = 'Thursday',
   Friday = 'Friday',
   Saturday = 'Saturday',
+  SingaporePublicHolidays = 'Singapore public holidays',
 }
 
 export type DateValidationOptions = {
@@ -26,5 +27,5 @@ export type DateValidationOptions = {
 export interface DateFieldBase extends MyInfoableFieldBase {
   fieldType: BasicField.Date
   dateValidation: DateValidationOptions
-  invalidDaysOfTheWeek?: DaysOfTheWeek[] | null
+  validDays?: ValidDaysOptions[]
 }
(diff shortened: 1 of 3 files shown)


src/app/models/field/dateField.ts
```diff
@@ -23,14 +26,15 @@ const createDateFieldSchema = () => {
...
+    validDays: {
       type: [
         {
           type: String,
           required: true,
         },
       ],
-      enum: [...Object.values(DaysOfTheWeek)],
+      enum: [...Object.values(ValidDaysOptions)],
+      default: Object.values(ValidDaysOptions),
```
karrui: i don't think we should inverse the relationship. This means that all old date fields without the validDays prop (because the docs may not have been modified yet) will suddenly disallow all dates. We should stick with invalid-if-present model.

karrui (changes requested):

karrui:
> Furthermore, since the Figma design shows that on default all the options will be checked, the default value of the attribute should be a string array containing all the options values instead of an empty array.

We can do this on the client side. No reason to change the backend for this

shared/types/field/dateField.ts
```diff
@@ -26,5 +27,5 @@ export type DateValidationOptions = {
 export interface DateFieldBase extends MyInfoableFieldBase {
   fieldType: BasicField.Date
   dateValidation: DateValidationOptions
-  invalidDaysOfTheWeek?: DaysOfTheWeek[] | null
+  invalidDays?: InvalidDaysOptions[]
```
karrui, line 30: comment: no more null? actually looks correct. can i have a refresh on why null was there before?

karrui (approved):
lgtm

shared/types/field/dateField.ts
siddarth2824: I added null because I thought previously if no checkboxes were selected, it would be a null value instead of empty array. But since we added a checkbox group to group all the checkboxes, the default value would be an empty array instead of null if no checkbox was selected.

PR #4325, opened 2022-07-26
fix(authTypes): disable changing of authtype with MyInfo, remove esrvcId box from SGID option

### LINKED ISSUE #4310: SGID auth option should not have e-service ID input

justynoh:
## Problem
Form creators can currently change the authtype from MyInfo to something else while they have MyInfo fields in the form.
Also, we currently show an esrvcId box under the SGID option, which should not occur. 

Closes #4310 and #4312 

## Solution

**Breaking Changes** 

## Screenshots

image

### DIFF
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
@@ -30,6 +30,17 @@ import {
 } from './constants'
 import { EsrvcIdBox } from './EsrvcIdBox'
 
+const esrvcidRequired = (authType: FormAuthType) => {
+  switch (authType) {
+    case FormAuthType.SP:
+    case FormAuthType.MyInfo:
+    case FormAuthType.CP:
+      return true
+    default:
+      return false
+  }
+}
+
 interface AuthSettingsSectionProps {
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
```diff
@@ -167,3 +187,13 @@ export const AuthSettingsSection = ({
     </Box>
   )
 }
+const esrvcidRequired = (authType: FormAuthType) => {
```
karrui: nit: should we move this into an utils folder or move it above the component declaration? this should be declared before being used in the component

karrui (approved):
lgtm

karrui:
can we also add a new AuthSettings story with the MyInfo state?

frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
justynoh: will move it above the component declaration! don't think we need to move it into a utils file unless somewhere else we might need this info? it isn't even enforced in `FormBase` lol

justynoh:
Added a new story for private form with MyInfo fields :)

Issue #4328, opened 2022-07-26
Refresh button on "The form has been modified" toast leads to landing page


mantariksh:
<img width="923" alt="Screenshot 2022-07-26 at 12 04 41 PM" src="https://user-images.githubusercontent.com/29480346/180920574-38b25ad9-4125-4c61-9474-4263760dddfd.png">

PR #4331, opened 2022-07-26
fix(v2): redirect users to current browser link instead of homepage when prompting a refresh

### LINKED ISSUE #4297: End page button link should point to endPage.buttonLink, not landing page

karrui:
## Problem

3 locations were directing users to the home page instead of the intended refresh:

1. When end page button link is not provided, the default behaviour should be to refresh the page to submit the form again.
2. When form has changed and the toast prompts the user to refresh to fetch the latest changes in public form
3. Above but for preview form (should not happen once #4296 is closed, but including here for completeness)

Closes #4297, #4328

## Solution

**Breaking Changes** 

**Bug Fixes**:

- feat: use current location as href instead of empty string for link `href` props when directing to homepage is not desired (it probably never is)

### DIFF
--- a/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
+++ b/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
@@ -74,7 +74,8 @@ export const PreviewFormProvider = ({
           ),
           description: (
             <Text as="span">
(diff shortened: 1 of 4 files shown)


wanlingt (approved):
lgtm!

PR #4263, opened 2022-07-21
feat(form-v2/react-to-angular/1): add switch to angular option

wanlingt:
## Problem
We need to give public respondents and admins a way to opt out of React and revert back to the AngularJS app.

Makes progress on #3933 

## Solution
This PR implements the inline messages for both public respondents and admins, giving them the option to switch from React to the AngularJS app.

**Key changes**:
- Add `PublicSwitchEnvMessage` for public respondents, and `AdminSwitchEnvMessage` for admins
- New API endpoint for public respondents `/api/v3/forms/environment/(react|angular)` to allow setting of the cookie
- Add queries to allow for switching of env cookies

**Other changes**:
- Set public respondent's cookie expiry to 2 months
 
**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Future PRs**
- [ ] Add feedback modal to allow users to submit feedback upon clicking to change
- [ ] Add floating action button to all admin pages
- [ ] Remove visibility of infoboxes after rollouts hit their respective % (10% for public respondents, 100% for admins)

## Before & After Screenshots

**AFTER**:
**Admin**
admin_AdobeExpress

**Public**
public_AdobeExpress

## Tests
- [x] [Admin] Log in. On the workspace page, click to switch to the original FormSG. The page should reload and should display the React interface. Also, the value of `v2-admin-ui cookie` should change from `angular` to `react`. This cookie should expire in 2 months time.
- [x] [Public] On a public form page, click to switch to the original FormSG. The page should reload and should display the React interface. Also, the value of `v2-respondent-ui cookie` should be `react`. This cookie should expire in 2 months time.

Issue #3933, opened 2022-06-01
Switching from React -> AngularJS


mantariksh:
While React is being rolled out, form admins and public respondents will be able to opt out of React and revert back to the AngularJS app. When they do so, we should collect feedback as to why they are switching.

**Public respondent flow**
- [Design here](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=8147%3A115201)
- Copy for info box tbc but we can start work first
- When user clicks on link in infobox, feedback modal opens
- When feedback is submitted, the following should happen:
1. A FormSG form (PM me for the link) should be submitted programmatically. The form has three fields: URL (this can be grabbed programmatically at the point of submission), email (optional field, leave it blank for public users) and the feedback content itself.
2. User's cookie should switch them to AngularJS so they see AngularJS when accessing forms in future. **This cookie expiry should be set to 2 months** so that they don't see React again until we are sure it is stable. We don't want the same user to encounter errors multiple times even after explicitly opting out.
3. Page should refresh and revert to AngularJS
- Note that we will remove this infobox once React rollout for public form respondents goes above 10%. This is because we expect the most issues to happen in the first 10%, and beyond that we are likely to get a lot of noise.

**Admin flow**
- [Design here](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=8540%3A116856)
- Again, exact copies tbc
- Floating action button is visible in all admin pages
- Clicking on button opens feedback modal. Upon submitting feedback, the same FormSG form should be submitted programmatically, together with the admin user's email which can be grabbed from session storage.
- After feedback is submitted, user has the option to switch to AngularJS or continue using React. Again, if they switch to AngularJS, their **cookie expiry should be set to 2 months** so they don't see React until we are sure it is stable.
- This button will be available for form admins until rollout hits 100%

PR #4352, opened 2022-07-26
feat(form-v2/react-to-angular/2): add feedback modal

### LINKED ISSUE #4165: [React] bug report when switching back from React -> AngularJS
we should ask users why they are switching back for more information

See #3933 for more info

wanlingt:
## Solution
This PR implements the bug report feedback modal that pops up when the user clicks to switch from React to the AngularJS app.

Builds on #4263 
Closes #4165 
Makes progress on https://github.com/opengovsg/FormSG/issues/3933

**Key changes:**
- `SwitchEnvFeedbackModal` is triggered by clicking on the link in the message banner, or on the floating action icon `SwitchEnvIcon` displayed on Admin pages
- The `formId` of the bug report feedback form changes depending on the staging/production environment, and we want to keep this hidden from the public. Hence, I created new API endpoints to **post** (`submitSwitchEnvFeedbackForm`) and **retrieve** (`getReactToAngularFeedbackFormView`) the feedback form programmatically without having to use `formId` as a param.
  - The env var `REACT_TO_ANGULAR_FEEDBACK_FORM_ID` is used as the `formId`
  - `REACT_TO_ANGULAR_FEEDBACK_FORM_ID` is injected into the respective routes through `public-form.middlewares`
- In the frontend,
  - `submitSwitchEnvFormFeedback` is used to submit the feedback form programmatically
      - `formData` has been hardcoded to match the format of the actual form
  - `getSwitchEnvFormView` is used to retrieve the feedback form fields

**Breaking Changes** 

**Future PRs**
- [x] Remove visibility of infoboxes after rollouts hit their respective % (10% for public respondents, 100% for admins) - #4514 

## Before & After Screenshots

**AFTER**:
**Admin**
Recording 2022-07-29 at 10 05 54
Recording 2022-07-29 at 10 07 11

**Public**
Recording 2022-07-29 at 10 04 43

## Tests
- [ ] [Admin] Click on the **action box** to switch to submit feedback. The feedback modal should pop up. Fill in the fields and submit the form. You should receive a form response in your email inbox.
- [ ] [Admin] On the workspace page, click the message banner's link to switch to the original FormSG. A feedback modal should pop up. **Leave the feedback blank and submit.** The email response should be recorded with the correct URL and email.
- [ ] [Admin] On the workspace page, click the message banner's link to switch to the original FormSG. A feedback modal should pop up. **Fill in the feedback field and submit.** The email response should be recorded with the correct URL, feedback and email.
- [ ] [Admin] After submitting the feedback, click 'Yes, please'. The modal should close and you should be brought to the same page in the AngularJS app.
- [ ] [Admin] After submitting the feedback, click 'No, don't switch'. The modal should close and you should remain on the same page in the React app.
 - [ ] [Public] On a public form page, click the message banner's link to switch to the original FormSG. A feedback modal should pop up. **Fill in the fields and submit the form.** The email response should be recorded with the correct URL and email.

## Deploy Notes

- `REACT_TO_ANGULAR_FEEDBACK_FORM_ID` : Form ID of the bug report feedback form
  - Note for local testing: Create a bug report feedback form with the fields `url` (short answer - required), `feedback` (long answer - optional), `email` (email - optional). Update `REACT_TO_ANGULAR_FEEDBACK_FORM_ID` in docker-compose.yml with that formId
- `REACT_MIGRATION_ANGULAR_END_DATE`: Date that the AngularJS app will be available until, shown on the switch environment banner on the Admin workspace page


frontend/src/features/env/EnvService.ts
```diff
@@ -1,7 +1,86 @@
...
+      question: key,
+      answer: value ?? '',
+      fieldType:
+        key === 'url'
+          ? feedbackForm?.form.form_fields[0].fieldType
+          : key === 'feedback'
+          ? feedbackForm?.form.form_fields[1].fieldType
+          : feedbackForm?.form.form_fields[2].fieldType,
+    }
+    responses.push(entry)
```
timotheeg: Suggestion to reduce code duplication.

```suggestion
    const fieldIndex = key === 'url' ? 0 : key === 'feedback' ? 1 : 2
    const { _id, fieldType } = feedbackForm?.form.form_fields[fieldIndex] ?? {}
    const entry = {
      _id,
      question: key,
      answer: value ?? '',
      fieldType,
    }
    responses.push(entry)
```

Also question: what does it mean for the entries when `feedbackForm` is `undefined`?

timotheeg (commented):
I'm not done reviewing but sending one comment first because I have to stop for now.

frontend/src/features/env/mutations.ts
```diff
@@ -1,28 +1,52 @@
...
     onSuccess: () => window.location.reload(),
   })
 
   const adminSwitchEnvMutation = useMutation(() => adminChooseEnvironment(), {
     onSuccess: () => {
-      navigate('/#!/forms')
-      window.location.reload()
+      // If on form builder page
+      if (window.location.href.indexOf('admin/form/') > -1) {
+        window.location.assign(`#!/${formId}/admin`)
```
timotheeg: Isn't this missing a leading `/` before the `#`?

frontend/src/features/env/mutations.ts
```diff
@@ -1,28 +1,52 @@
...
   const publicSwitchEnvMutation = useMutation(() => publicChooseEnvironment(), {
     onSuccess: () => window.location.reload(),
   })
 
   const adminSwitchEnvMutation = useMutation(() => adminChooseEnvironment(), {
     onSuccess: () => {
-      navigate('/#!/forms')
-      window.location.reload()
+      // If on form builder page
+      if (window.location.href.indexOf('admin/form/') > -1) {
```
timotheeg: I'm always worried of `href.indexOf()` checks, because they can capture stuff in query string or hash, which should have been inconsequential to the actual test. Better to be very specific on the source string (check the properties [`window.location`](https://developer.mozilla.org/en-US/docs/Web/API/Location) offers), and by using a regexp, preferably with an anchor if possible.

frontend/src/features/env/mutations.ts
```diff
@@ -1,28 +1,52 @@
...
+        window.location.assign(`#!/${formId}/admin`)
+        // If on admin workspace page
+      } else if (window.location.href.indexOf('workspace') > -1) {
```
timotheeg: Same comment on `href` check.

frontend/src/features/workspace/components/AdminSwitchEnvMessage.tsx
```diff
@@ -1,23 +1,26 @@
...
-      </Text>
-    </InlineMessage>
+    <>
+      <InlineMessage>
+        <Text>
+          Welcome to the new FormSG! You can still
+          <Button variant="link" onClick={onOpen}>
+            <Text as="u">switch to the original one,</Text>
+          </Button>
+          which is available until 28 May 2022.
```
timotheeg: > until 28 May 2022

LOL, probably need to change that to, hmmm, let's say `15 September 2022` for now? 😅

frontend/src/features/env/mutations.ts
wanlingt: yes thanks for spotting! have updated it

frontend/src/features/env/mutations.ts
wanlingt: ah I see! got it, I've changed it to `window.location.pathname` with a regexp match

frontend/src/features/workspace/components/AdminSwitchEnvMessage.tsx
wanlingt: hahahhaha I've changed the date to an env var in case we need to change it again in the future 😬

frontend/src/features/env/mutations.ts
```diff
@@ -1,28 +1,52 @@
...
     onSuccess: () => window.location.reload(),
   })
 
   const adminSwitchEnvMutation = useMutation(() => adminChooseEnvironment(), {
     onSuccess: () => {
-      navigate('/#!/forms')
-      window.location.reload()
+      // If on form builder page
+      if (window.location.pathname.match('^/admin/form/')) {
+        window.location.assign(`/#!/${formId}/admin`)
```
timotheeg: Interesting! I didn't know `.match()` could take a string to represent the pattern. Quick trip to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) shows string will be wrapped with a call to the constructor of the `RegExp` class.

> [Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#parameters)
> 
> **`(regexp)`**
>
> A regular expression object, or any object that has a [`Symbol.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/match) method.
> 
> If `regexp` is not a RegExp object and does not have a `Symbol.match` method, it is implicitly converted to a [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) by using `new RegExp(regexp)`.

frontend/src/features/env/EnvService.ts
```diff
@@ -1,7 +1,83 @@
...
+
+// TODO #4279: Remove after React rollout is complete
+const createFeedbackResponsesArray = (
+  formInputs: switchEnvFeedbackFormBodyDto,
+  feedbackForm: PublicFormViewDto,
+) => {
+  const responses = []
+  for (const [key, value] of Object.entries(formInputs)) {
+    const fieldIndex = key === 'url' ? 0 : key === 'feedback' ? 1 : 2
+    const { _id, fieldType } = feedbackForm?.form.form_fields[fieldIndex] ?? {}
```
timotheeg: Since `feedbackForm` can no longer be `undefined`, the conditional access is no longer needed. Additionally, since my previous suggestion of using `?? {}` was to address that `undefined`, if we know for a fact that the 3 possible indexes (`0`, `1`, `2`) are all always present, it could be dropped as well.
```suggestion
    const { _id, fieldType } = feedbackForm.form.form_fields[fieldIndex]
```

frontend/src/features/env/EnvService.ts
```diff
@@ -1,7 +1,83 @@
...
+    const entry = {
+      _id,
+      question: key,
+      answer: value ?? '',
+      fieldType,
+    }
+    responses.push(entry)
```
timotheeg: Re-reading the code, the structure of the loop reads a little "strange". The loop heading is generic (using Object.entries()`), but the loop body shows very explicit expectations of having just 3 entries, 2 of which with known names.

Clearly we're working on mapping the 2 structure `formInputs` and `feedbackForm`, with their shapes being well defined. Because of that, the extraction code could make that mapping explicit instead, to show what we expect to get at the end, maybe something like this:

```suggestion
  // [key_in_inputs:string, corresponding_index_in_form:number]
  return [
    ['url', 0],
    ['feedback', 1],
    ['email', 2]
  ]
    .map(([inputKey, formFieldIndex]) => {
      const { _id, fieldType } = feedbackForm.form.form_fields[formFieldIndex]
      const answer = formInputs[inputKey] ?? ''
  
      return {
        _id,
        question: inputKey,
        answer,
        fieldType,
      }
    })
```

Some note on the above:
1. untested (am not sure if typescript will be happy with it 😅)
2. It gives explicit order of the resultant response array
3. it _looks_ like `formFieldIndex` are not needed in the map definition because the index values match the array indexes anyway, but I think it's worth having them stated anyway, because the indexes represent different things. For example, this would work as a map definition, and show the indexes are not necessarily synced:
```javascript
  return [
    ['url', 0],
    ['email', 2]
    ['feedback', 1],
  ]
    .map( /* ... */ )
```

frontend/src/features/env/mutations.ts
wanlingt: I decided to use explicit RegExp objects to make it clearer, [here](https://github.com/opengovsg/FormSG/pull/4352/commits/2a9f26509286f149b2b44d012e33e518328e49df) :)

frontend/src/features/env/EnvService.ts
wanlingt: thanks for the extensive comment and explanation Tim! It does make it clearer to make the mapping explicit. I've taken your suggestion and added type definitions [here](https://github.com/opengovsg/FormSG/pull/4352/commits/cad0025747635dea10b5d6bedf5dfbe0dca7dfa8)

frontend/src/features/env/SwitchEnvFeedbackModal.tsx
```diff
@@ -0,0 +1,185 @@
...
+                  {user ? (
+                    <Input
+                      type="hidden"
+                      {...register('email')}
+                      value={user.email}
+                    />
+                  ) : (
+                    <></>
+                  )}
+                </FormControl>
```
timotheeg: /nit: is the `FormControl` wrapper necessary when `user` is falsy? Also, shouldn't we use `null` rather than the empty markup? (am not sure whats the recommended approach is)

```suggestion
                { user ? (
                  <FormControl>
                    <Input
                      type="hidden"
                      {...register('email')}
                      value={user.email}
                    />
                  </FormControl>
                  ) : null }
                
```

src/app/config/schema.ts
```diff
@@ -294,19 +294,19 @@ export const optionalVarsSchema: Schema<IOptionalVarsSchema> = {
     respondentRolloutEmail: {
       doc: 'Percentage threshold to serve React for respondents for Phase 1 (email mode forms)',
       format: 'int',
-      default: 0,
+      default: 1,
```
timotheeg: Why are we changing these to 1 again?

src/app/modules/form/public-form/public-form.middlewares.ts
```diff
@@ -0,0 +1,16 @@
...
+import { reactMigration } from '../../../config/config'
+import { ControllerHandler } from '../../core/core.types'
+
+export const injectFeedbackFormUrl: ControllerHandler = (req, res, next) => {
+  const formId = reactMigration.reactToAngularFeedbackFormId
+  if (!formId) {
+    return res
+      .status(StatusCodes.BAD_REQUEST)
+      .json({ message: 'No feedback form id provided' })
+  }
```
timotheeg: I don't think this error represent a "Bad Request" 🤔 . If `config.reactMigration.reactToAngularFeedbackFormId` is not supplied, it's **our** mistake (not the caller), and should be a 500 unexpected error instead.

On that note don't we have a mechanism to ensure a value for the ~form~ env var (🤦!) is provided when the config object is instantiated?

frontend/src/features/env/SwitchEnvFeedbackModal.tsx
wanlingt: you're right `FormControl` wrapper isn't needed when `user` is falsy, amended

src/app/config/schema.ts
wanlingt: this was my mistake - i meant to change `docker-compose.yml` instead so that the banner message shows up.  Changed this back to `0`!

src/app/modules/form/public-form/public-form.middlewares.ts
wanlingt: As discussed, the error that is thrown should not be a runtime error; rather the app shouldn't even work since this is a missing config. Hence I made the following changes
- remove the error handling from `public-form.middlewares`
- moved the `reactToAngularFeedbackFormId` env var from `IOptionalVarsSchema` to `ICompulsoryVarsSchema`
(sidenote: I've set `default` as `null`, because according to the [convict docs](https://github.com/mozilla/node-convict/tree/master/packages/convict) 'Every setting must have a default value'. However, ['you can set a default value to null and if your format doesn't accept null it will throw an error.'](https://github.com/mozilla/node-convict/tree/master/packages/convict#how-can-i-define-a-configuration-property-as-required-without-providing-a-default-value)

Tested the code on staging-alt2, it works! :) We get 502 Bad Gateway when `REACT_TO_ANGULAR_FEEDBACK_FORM_ID` is not supplied.

timotheeg (commented):
LGTM! Thanks @wanlingt 👍

src/app/modules/form/public-form/public-form.middlewares.ts
timotheeg: > We get 502 Bad Gateway when REACT_TO_ANGULAR_FEEDBACK_FORM_ID is not supplied.

Wait, that was not my expectation 😰. I though having the config entry missing should cause the deployment to fail, but the service should stay up thanks to the old live instance(s) not being retired and not give 502s (maybe a 404 if a new route didn't get deployed?). Let's have another chat when you are in to discuss further 🙂

wanlingt: Tested on Elastic Beanstalk using **rolling deployments** with >1 instance.
Upon deploying this, we get a _"Failed to deploy application"_ error and the website is still accessible.

timotheeg (approved):

PR #4386, opened 2022-07-27
chore: upgrade nodejs docker image to fermium 3.16

timotheeg:
## Problem
[Snyk reported](https://app.snyk.io/org/formsg) that the nodejs fermium image has several security issues reported against it. 

Most are not a concern (e.g. ssh vulnerability is not a concern to us since we do not enable ssh anyway), but still, we should keep dependency fresh and patched.

## Solution
Upgrade to latest fermium image 3.16 . See docker hub for available versions [here](https://hub.docker.com/_/node)


karrui (approved):
lgtm

PR #4390, opened 2022-07-27
fix: trim secret key before validating it

### LINKED ISSUE #4373: Trim secret key before checking validity

timotheeg:
## Problem
A Secret key could be accidentally supplied with leading or tailing whitespaces (by copy paste or by file). The code does trim the values, but the regexp fails the validation earlier than that

Closes #4373

## Solution
* Remove the pattern check and move it into the validator
* Trim the secretKey before doing validations


karrui (approved):
lgtm

Issue #4330, opened 2022-07-26
Create form modal copy changes


mantariksh:
Remove "Benefits" from both cards

**Storage Mode**
View or download responses in FormSG
 
Attachments: up to 20MB per form
Up to Restricted and Sensitive (Normal) data
Supports webhooks for responses

**Email Mode**
Receive responses in your inbox
 
Attachments: up to 7MB per form
Up to Restricted and Sensitive (High) data
Supports MyInfo fields

(cut out "PDF responses" as I couldn't find a way to state the feature clearly without adding another bullet point. "PDF responses" is too vague)

Checkbox: change to "I acknowledge that if I lose my Secret Key, all my responses will be lost permanently and I cannot activate my form"

timotheeg:
Closed by #4394

PR #4394, opened 2022-07-27
fix: copy update for create-form modal

timotheeg:
## Problem
Copy needs to be updated, per requirements in #4330

## Solution
Update copy per requirement 💪


karrui (approved):
lgtm

PR #4395, opened 2022-07-27
feat(sgid): enable storage mode for sgid

kwajiehao:
## Problem
sgID is actively trying to grow their product, and part of the growth strategy is to have more sgID FormSG forms. However, sgID is currently available only in email mode. This is a big deterrence for form admins who expect a large number of respondents for their forms. 

In particular, sgID has a big use case (expected form submissions in the range of the tens of thousands) looking to collect feedback related to National Day, so we're looking to enable sgID storage mode forms before then.

## Solution
This PR enables storage mode for sgID on both the frontend and the backend. The main change made was to add functions for getting verified sgID content. Other changes involved mainly flipping switches to handle sgID as a valid auth type for encrypt mode forms.

**Breaking Changes** 

## Alternative solutions considered
**Using email mode but managing the data collation for them**
If we manage data collation on their half, we don't need to do additional development work on the FormSG codebase.

**Not using authentication**
This would save us the effort of not having to do additional development work. Authentication is probably a good-to-have feature for form admins for now. However, form admins are keen to have authentication because it gives them a greater signal on the form data (prevents spam and fake users).

### sgID stance
sgID is keen to work on this because it would be useful to have a full working prototype to persuade other clients to onboard, instead of an ad-hoc workaround solution (managed data collation). Taking into account the fact that we have the capacity to work on this, the preference is to make the change on the FormSG codebase.

## Before & After Screenshots

**BEFORE**:
Screenshot 2022-07-27 at 5 17 07 PM

**AFTER**:
Screenshot 2022-07-27 at 5 16 55 PM

## Tests
Expected outcomes:
- [ ] sgID option appears in storage mode form settings
- [ ] User is able to login with sgID for storage mode form and submit the form
- [ ] Admin is able to successfully decrypt sgID storage mode form responses
- [ ] Check that end-to-end flow for sgid email mode is unaffected

## Questions
- The current text for the sgID radio button in the form settings is “Singpass App-only Login (Free)”. Are we okay with this, or do we want to change the copy for this button?

### DIFF
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -671,17 +671,18 @@ describe('Form Model', () => {
         await expect(invalidForm.authType).toBe(FormAuthType.NIL)
       })
 
-      it('should set authType to NIL when given authType is SGID', async () => {
+      // Ensure that encrypted sgID forms can be created since they could not before
+      it('should set authType to SGID when given authType is SGID', async () => {
         // Arrange
-        const malformedParams = merge({}, MOCK_ENCRYPTED_FORM_PARAMS, {
+        const encryptFormParams = merge({}, MOCK_ENCRYPTED_FORM_PARAMS, {
           authType: FormAuthType.SGID,
         })
 
         // Act
-        const invalidForm = await EncryptedForm.create(malformedParams)
+        const sgidForm = await EncryptedForm.create(encryptFormParams)
 
         // Assert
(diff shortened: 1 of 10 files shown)


karrui (approved):
lgtm!

karrui:
@timotheeg can you take a look at this PR too? If its fine we should release it next release

timotheeg (approved):
LGTM. Thanks @kwajiehao !

PR #4404, opened 2022-07-28
feat(form-workspaces-be/6): add move forms from a single source workspace functionality

hanstirtaputra:
Builds on #4254 

Part of #4039 

## Solution
Add endpoint logic for POST `api/v3/admin/workspaces/{workspaceId}/forms/move` for moving forms from one source workspace to one destination workspace

**Breaking Changes** 

### DIFF
--- a/src/app/modules/workspace/__tests__/workspace.controller.spec.ts
+++ b/src/app/modules/workspace/__tests__/workspace.controller.spec.ts
@@ -328,4 +328,120 @@ describe('workspace.controller', () => {
       expect(mockRes.json).toBeCalledWith({ message: mockErrorString })
     })
   })
+
+  describe('updateFormsWorkspace', () => {
+    const MOCK_REQ = expressHandler.mockRequest({
+      params: {
+        workspaceId: new ObjectId() as IWorkspaceSchema['_id'],
+      },
+      session: {
+        user: {
+          _id: 'exists',
+        },
+      },
+      body: {
+        formIds: [new ObjectId().toHexString()],
+        destWorkspaceId: new ObjectId() as IWorkspaceSchema['_id'],
+      },
+    })
+    const MOCK_WORKSPACE = {
+      _id: new ObjectId() as IWorkspaceSchema['_id'],
+      title: 'Workspace1',
+      admin: new ObjectId() as IUserSchema['_id'],
+      formIds: [],
+    }
+
+    it('should return 200 with success message', async () => {
+      const mockRes = expressHandler.mockResponse()
+      MockWorkspaceService.getWorkspace.mockReturnValue(okAsync(MOCK_WORKSPACE))
+      MockWorkspaceService.verifyWorkspaceAdmin.mockReturnValue(okAsync(true))
+      MockWorkspaceService.moveForms.mockReturnValueOnce(
+        okAsync(MOCK_WORKSPACE),
+      )
+
+      await WorkspaceController.updateFormsWorkspace(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.json).toHaveBeenCalledWith(MOCK_WORKSPACE)
+    })
+
+    it('should return 403 when workspace is not found', async () => {
+      const mockRes = expressHandler.mockResponse()
+      const mockErrorString = 'something went wrong'
+
+      MockWorkspaceService.getWorkspace.mockReturnValue(okAsync(MOCK_WORKSPACE))
+      MockWorkspaceService.verifyWorkspaceAdmin.mockReturnValue(
+        errAsync(new ForbiddenWorkspaceError(mockErrorString)),
+      )
+
+      await WorkspaceController.updateFormsWorkspace(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.status).toBeCalledWith(403)
+      expect(mockRes.json).toBeCalledWith({ message: mockErrorString })
+    })
+
+    it('should return 404 when workspace is not found', async () => {
+      const mockRes = expressHandler.mockResponse()
+      const mockErrorString = 'something went wrong'
+
+      MockWorkspaceService.getWorkspace.mockReturnValue(
+        errAsync(new WorkspaceNotFoundError(mockErrorString)),
+      )
+      MockWorkspaceService.verifyWorkspaceAdmin.mockReturnValue(okAsync(true))
+
+      await WorkspaceController.updateFormsWorkspace(
+        MOCK_REQ,
+        mockRes,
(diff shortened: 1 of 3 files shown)


src/app/models/workspace.server.model.ts
```diff
@@ -86,6 +91,38 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+    formIds,
+    session,
+  }: {
+    workspaceId: IWorkspaceSchema['_id']
+    formIds: IFormSchema['_id'][]
+    session?: ClientSession
+  }) {
+    await this.updateOne(
+      { _id: workspaceId },
+      { $pullAll: { formIds } },
```
mantariksh, line 105: cool!

src/app/models/workspace.server.model.ts
```diff
@@ -86,6 +91,38 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+    formIds,
+    session,
+  }: {
+    workspaceId: IWorkspaceSchema['_id']
+    formIds: IFormSchema['_id'][]
+    session?: ClientSession
+  }) {
+    await this.updateOne(
+      { _id: workspaceId },
+      { $addToSet: { formIds: { $each: formIds } } },
```
mantariksh, line 121: also cool!

src/app/modules/workspace/__tests__/workspace.controller.spec.ts
```diff
@@ -328,4 +328,120 @@ describe('workspace.controller', () => {
...
+      await WorkspaceController.updateFormsWorkspace(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.json).toHaveBeenCalledWith(MOCK_WORKSPACE)
+    })
+
+    it('should return 403 when workspace is not found', async () => {
```
mantariksh, line 371: looks like wrong test description?

src/app/modules/workspace/__tests__/workspace.controller.spec.ts
```diff
@@ -328,4 +328,120 @@ describe('workspace.controller', () => {
...
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.status).toBeCalledWith(403)
+      expect(mockRes.json).toBeCalledWith({ message: mockErrorString })
+    })
+
+    it('should return 404 when workspace is not found', async () => {
```
mantariksh, line 390: probably need to swap this with prev test

src/app/modules/workspace/workspace.controller.ts
```diff
@@ -247,22 +247,57 @@ export const deleteForms: ControllerHandler<
...
  * @returns 404 when the workspace does not exist or belong to the user
  * @returns 422 when user of given id cannnot be found in the database
  * @returns 500 when database errors occur
  */
-export const moveForms: ControllerHandler<
-  { sourceWorkspaceId: string },
-  unknown,
-  any | ErrorDto,
-  { formIds: any[]; destWorkspaceId: string }
+export const updateFormsWorkspace: ControllerHandler<
```
mantariksh, line 255: not a very clear function name, maybe `moveFormsToWorkspace`?

src/app/models/workspace.server.model.ts
```diff
@@ -86,6 +91,38 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
   }
 
+  WorkspaceSchema.statics.removeFormIdsFromWorkspace = async function ({
+    workspaceId,
+    formIds,
+    session,
+  }: {
+    workspaceId: IWorkspaceSchema['_id']
+    formIds: IFormSchema['_id'][]
+    session?: ClientSession
```
mantariksh, line 101: if we always expect this operation to take place within a transaction, we should make this param mandatory

src/app/modules/workspace/workspace.controller.ts
```diff
@@ -247,22 +247,57 @@ export const deleteForms: ControllerHandler<
...
+      workspace
+        ? res.status(StatusCodes.OK).json(workspace)
+        : res.status(StatusCodes.INTERNAL_SERVER_ERROR).json({
+            message:
+              'Sorry something went wrong, we are unable to move the forms to another worksapce',
+          }),
     )
+    .mapErr((error) => {
+      logger.error({
+        message: 'Error moving forms to anotehr workspace',
```
mantariksh, line 289:
```suggestion
        message: 'Error moving forms to another workspace',
```

src/app/models/workspace.server.model.ts
```diff
@@ -86,6 +91,38 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+  }) {
+    await this.updateOne(
```
mantariksh, line 103: why don't we use `findOneAndUpdate` so we can return the updated document to the caller?

src/app/models/workspace.server.model.ts
```diff
@@ -86,6 +91,38 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+  WorkspaceSchema.statics.addFormIdsToWorkspace = async function ({
+    workspaceId,
+    formIds,
+    session,
+  }: {
+    workspaceId: IWorkspaceSchema['_id']
+    formIds: IFormSchema['_id'][]
+    session?: ClientSession
+  }) {
+    await this.updateOne(
```
mantariksh, line 119: we can use `findOneAndUpdate`? then we don't have to make another db query to retrieve the updated destination workspace at the end

src/app/modules/workspace/workspace.controller.ts
```diff
@@ -247,22 +247,57 @@ export const deleteForms: ControllerHandler<
...
+      }),
+    )
+    .map((workspace) =>
+      workspace
+        ? res.status(StatusCodes.OK).json(workspace)
+        : res.status(StatusCodes.INTERNAL_SERVER_ERROR).json({
+            message:
+              'Sorry something went wrong, we are unable to move the forms to another worksapce',
+          }),
     )
```
mantariksh: hmm, this is messy. at this point we should be sure that the operation has succeeded, all errors should flow to the `mapErr` block below. can we get `WorkspaceService.moveForms` to return an error if the workspace can't be found? using `findOneAndUpdate` in the model layer will probably help

mantariksh:
how do we prevent form admins from submitting IDs of forms they don't own, which would result in their workspaces containing forms they shouldn't be able to access?

hanstirtaputra:
Closing this PR as it is a special case of `updateMany` in #4448 . Will just keep #4448 instead.

Issue #4403, opened 2022-07-28
Email verified field OTP expiry is inconsistent


justynoh:
Should be 30 min expiry. 

![image](https://user-images.githubusercontent.com/25571626/181411241-c1c5279a-b9ff-41cb-bffe-2d3a92eb9a7b.png)

![image](https://user-images.githubusercontent.com/25571626/181411252-0e15a376-e537-4ba7-bb30-1336b80b3217.png)

PR #4405, opened 2022-07-28
fix(form-v2): update OTP resend button to wait 60s on first send, update copy for OTP validity

### LINKED ISSUE #4344: "Resend OTP" should start countdown upon 1st OTP requested
Currently, users are allowed to request for another OTP immediately. And the countdown starts only after they request for another.

justynoh:
## Problem
1. Currently, users who request for an OTP for the first time can resend immediately. #4344
2. Also, the copy for OTP validity says 10 min whereas the actual validity (in prod and as sent via email) is 30 min. #4403

## Solution

## Screenshots

1. 
https://user-images.githubusercontent.com/25571626/181412848-6d044bd0-9c4e-4e52-8b93-0316003ce481.mov

2.
image


wanlingt (approved):
lgtm!

PR #4409, opened 2022-07-28
fix(docker): upgrade puppeteer and chromium

timotheeg:
## Problem
PR #4386 upgraded the docker base image to address some security vulnerabilities. Unfortunately, the PR broke the docker build because the chromium version that was pin was no longer available in the latest fermium image.

## Solution
Upgrade both puppeteer-core and the chromium version to fix the docker builds.

Confirmed docker build is successful by releasing to staging

## Tests
- [x] In an email-mode form, add an email field, select email confirmation, check "Include pdf response", submit the form, verify that the submissions goes to admin AND that the respondent gets the confirmation with correct pdf attachment


karrui (approved):
Lgtm

PR #4410, opened 2022-07-28
fix(form-v2): make table question display format in storage mode responses consistent with AngularJS

### LINKED ISSUE #4376: Make table question display format in storage mode responses consistent with AngularJS
Angular: Question Title (Text Field, Dropdown, Text Field)
React: Question Title

Angular
image

React
image

justynoh:
## Problem
In angular, individual responses on storage mode display table question titles with the their column names. This is not done in React.

Closes #4376 

## Solution
When the table input is transformed to output before, add the column names with the table question name.

**Breaking Changes** 


karrui (approved):
lgtm

PR #4411, opened 2022-07-28
fix(v2): enable previewing of authenticated forms

### LINKED ISSUE #4358: Authenticated form preview mock data
- For SPCP, MyInfo and SGID, preview should not require login; should be pre-logged in with mock NRIC/UEN as per AngularJS
- MyInfo should be prefilled with mock data as per AngularJS

### LINKED ISSUE #4374: Option to switch between React <-> AngularJS should not appear in Preview mode

karrui:
## Problem

Previously, when a form with authentication is previewed, they will still need to log in with singpass. This PR updates preview to consist of a mock login state, along with injecting MyInfo values if MyInfo fields are used.

Also fixed styling issue with previewing form where the form will not fill the entire viewport.

Closes #4358, closes #4374

## Solution

**Breaking Changes** 

**Features**:

- feat: inject preview myinfo values if previewing form with myinfo

**Improvements**:

- style: wrap preview form page with flex wrapper 

**Bug Fixes**:

- fix: correctly show preview fields if user is already authenticated


frontend/src/features/admin-form/common/AdminViewFormService.ts
```diff
@@ -47,7 +54,19 @@ export const previewForm = async (
 ): Promise<PreviewFormViewDto> => {
   return ApiService.get<PreviewFormViewDto>(
     `${ADMIN_FORM_ENDPOINT}/${formId}/preview`,
-  ).then(({ data }) => data)
+  ).then(({ data }) => {
+    const mockedSpcpSession: PreviewFormViewDto['spcpSession'] =
+      data.spcpSession ?? data.form.authType !== FormAuthType.NIL
+        ? { userName: PREVIEW_MOCK_UINFIN }
+        : undefined
```
timotheeg: This code is not assigning `data.spcpSession` to mockedSpcpSession when `data.spcpSession` is set, is that intentional?

I had some difficulty reading this statement btw, with contending with the 3 operators precedence (`??`, `!==`, and `?:`). Perhaps we could split this into multiple statements, or add braces to emphasize intent 🤔 

Like, I wasn't sure if the intent of the code was to write this behaviour:
```javascript
const spcpSessionIsSet = data.spcpSession !== undefined && data.spcpSession !== null

mockedSpcpSession = undefined

if (spcpSessionIsSet) {
    mockedSpcpSession = data.spcpSession
}
else if (data.form.authType !== FormAuthType.NIL) {
    mockedSpcpSession = { userName: PREVIEW_MOCK_UINFIN }
}
```

or this: (which is the current behaviour)

```javascript
const spcpSessionIsSet = data.spcpSession !== undefined && data.spcpSession !== null
const spcpSessionIsTruthy = !!data.spcpSession

mockedSpcpSession = undefined

if (spcpSessionIsSet && spcpSessionIsTruthy) {
    mockedSpcpSession = { userName: PREVIEW_MOCK_UINFIN }
}
else if (data.form.authType !== FormAuthType.NIL) {
    mockedSpcpSession = { userName: PREVIEW_MOCK_UINFIN }
}
```

Just as an additional side note, I was confused by the `??` because it make it seems like we have a concern for a case of defined but falsy, but since `data.spcpSession` is of type SpcpSession (an object), it's not a valid concern 🤔

karrui: yea, the data sent by the server should take precedence (but as of now nothing will be sent by the server since there is probably no need to).

will update to more it more readable

karrui: fixed in d8b3feb

timotheeg (approved):
LGTM! 👍 

New mutation is much clearer to read, thanks! 👏 🎩

PR #4414, opened 2022-07-28
chore: improve webpack bundling

darrelhong:
## Overview
this PR covers some low-hanging fruits to improve the bundling of the v2 react frontend. while it might not be that important as you can get away with a lot these days (especially for a country [pretty fast](https://www.speedtest.net/global-index) broadband and mobile speeds), its at least good to take note. for example, i think the landing page bundles almost all the dependencies

## Problem
current bundle for react frontend has duplicate dependencies and lots of modules in main/vendor bundle

## Solution
add bundle analyzer plugin
update webpack alias to bundle from frontend/node_modules instead of shared
lazy import some of the bigger modules/pages

**Breaking Changes** 
probably the latest commit on this branch needs a bit more checking/testing. Not sure why some pages are lazily imported and others are not, maybe there are some considerations i am unaware of. Also didn't run the full docker workflow and just ran the frontend project on native host, so I can't be sure that nothing has broken
- [x] this PR contains breaking changes - Maybe
   
## Before & After Screenshots

**BEFORE**:
Screenshot 2022-07-29 at 12 14 15 AM

**AFTER**:
Screenshot 2022-07-29 at 12 17 52 AM

definitely more opportunities for improvement with splitting the vendor bundle and it also looks like lodash is not tree-shaken

**New dev dependencies**:

- `dependency` : :[webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer)

Other things that might be considered could be updating to [CRA v5](https://github.com/dilanx/craco/issues/378) to see if there's any improvements, or maybe even ejecting from CRA.

### DIFF
--- a/frontend/craco.config.js
+++ b/frontend/craco.config.js
@@ -2,6 +2,7 @@
 const CracoAlias = require('craco-alias')
 const merge = require('lodash/merge')
 const path = require('path')
+const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer')
 
 const customJestConfig = require('./jest.config')
 
@@ -24,6 +25,9 @@ module.exports = {
         ],
       },
     },
+    plugins: process.env.ANALYZE && [
+      new BundleAnalyzerPlugin({ generateStatsFile: true }),
+    ],
   },
   plugins: [
     {
--- a/frontend/craco.config.js
+++ b/frontend/craco.config.js
@@ -55,6 +55,13 @@ module.exports = {
     {
       plugin: {
         overrideWebpackConfig: ({ webpackConfig }) => {
+          /**
+           * Utility to allow for referencing folders outside of create-react-app root. \
+           * See https://github.com/facebook/create-react-app/issues/9127.
+           *
+           * Retrieved from https://gist.github.com/stevemu/53c5006c5e66fc277dea1454eb6acdb4.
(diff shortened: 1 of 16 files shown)


timotheeg:
Thanks for the PR @darrelhong 🙏 . `webpack-bundle-analyzer` is really neat!

We are working on functional completion atm, so we might not get to this PR right away, but we will relook at our bundle strategy and optimizations as soon as we can. Thanks for bringing it to our attention!

darrelhong:
since form-v2 is in "beta", it might be a good idea to merge some of these changes in so that any issues would be caught early. The changes should not affect the functional side of things and I believe the unit tests still pass. not too sure if it breaks any pipelines on your side

darrelhong:
@timotheeg still looking into this? if not i will close the PR

darrelhong:
@timotheeg @justynoh any update on this? I could take another look at this and it probably needs some rework

timotheeg:
We are getting close to finally getting rid of angular in the codebase, and we'll be looking at upgrading react and the webpack bundling after thyat. 

Since the PR is outdated with conflicts now, I'm going to close it first. It's still a good problem to solve, so I recommend opening an issue first, and then opening the PR again when the code base has stabilized (another 2-3 weeks I reckon 🤞)

PR #4419, opened 2022-07-29
feat: add functionality for question fields in builder tool to reflect selected colortheme

### LINKED ISSUE #4327: [React] Color theme selected in design drawer should be propagated to form fields
When a form creator selects a different color theme for their form, the start page is updated but not the rest of the fields which make use of the color theme. We should ensure that the rest of the fields get updated appropriately in the builder as well, since these are themed correctly in the public form.

image

image

justynoh:
## Problem
Currently, the builder content only shows all questions in the default blue theme, even if a different theme is selected. This PR adds support for the builder content to display with the correct color scheme even in the builder content.

Closes #4327 

## Solution
- Add calls to `useDesignStore` and `useCreateTabForm` in `FieldRowContainer` to retrieve correct `colorTheme`, and pass it along to children.
- Updated section row components to accept color themes and display correctly.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/181669968-94c3011f-eb47-4408-9ee2-082340594b42.mov

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
@@ -14,6 +14,7 @@ import {
 } from '@chakra-ui/react'
 import { times } from 'lodash'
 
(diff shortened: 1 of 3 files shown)


timotheeg (approved):
lgtm

PR #4421, opened 2022-07-29
fix: avoid calling useDesignStore on every render

justynoh:
This PR introduces a small fix to useCallback on each call to useDesignStore to avoid making the same call on every render.


karrui (approved):
lgtm

PR #4422, opened 2022-07-29
fix: add check that there is at least 1 option for split text areas

justynoh:
## Problem
Radio, checkbox and dropdown currently allow 0 options. They shouldn't.

Closes [#4378 ]

## Solution
Add check to ensure that the textarea is nonempty. 

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/constants.ts
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/constants.ts
@@ -9,6 +9,7 @@ export const SPLIT_TEXTAREA_TRANSFORM = {
 
 export const SPLIT_TEXTAREA_VALIDATION = {
   validate: (opts: string) => {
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/constants.ts
```diff
@@ -9,6 +9,7 @@ export const SPLIT_TEXTAREA_TRANSFORM = {
 
 export const SPLIT_TEXTAREA_VALIDATION = {
   validate: (opts: string) => {
+    if (!opts) return 'Please provide at least 1 option.'
     const optsArr = SPLIT_TEXTAREA_TRANSFORM.output(opts)
```
timotheeg: Let's use the existing mechanism that decides what constitutes an option instead. With it, whitespaces string will also be handled.
```suggestion
    const optsArr = SPLIT_TEXTAREA_TRANSFORM.output(opts)
    if (!optsArr.length) return 'Please provide at least 1 option.'
```

timotheeg (approved):

PR #4427, opened 2022-07-29
feat: remove Singapore public holiday option from restricting particular days

siddarth2824:
## Solution

Split the implementation of restricting particular days into 2 separate big features instead of 1. Hence, removal of all instances of Singapore public holidays as an invalid day option.

**Breaking Changes** 

    - In the DB there might be existing date fields with Singapore public holidays as selected invalid days. During validation on the server side, date responses selected by public users might be Singapore public holidays and there would be inaccurate validations.


karrui (approved):
lgtm

PR #4429, opened 2022-07-29
feat(form-v2): percentage-based react rollout for admins

### LINKED ISSUE #4217: Percentage-based React rollout for form admins
We should be able to progressively roll out the React frontend for form admins (workspace, form editing etc) by randomly assigning them a React/Angular cookie based on a predefined rollout percentage.

Ideally, we should have independent rollout thresholds for the admins and respondents. This is because given that the admin side is much more complex than the respondent side, there may be issues which purely affect form admins and might require us to roll back the admin side. In this case, ideally we can let the respondent side continue being tested while we fix the admin side.

Our current assessment is that independent thresholds are not difficult to implement, but if they turn out to be more complex than expected, we can discuss the possibility of making the admin rollout percentage identical to the respondent (no auth) rollout percentage.

Note: the previous plan was to have 2 weeks of allowing admins to opt in to React, after which they would see React by default. However, the risk of this is that we might not have enough testers in the first two weeks, and might then see a spike in issues when React becomes the default. A percentage-based rollout helps us avoid this.

### Once on React, stay on React

We also don't want users to be flipped back and forth between React and AngularJS as it will be disruptive to flip back to AngularJS just as they are getting used to React. If they get assigned React, they should stay on React.

### Once opting for Angular, stay on Angular until React is stable

We don't want situations where users opt in to Angular because they are facing issues, then we forcibly put them back on React, then they face issues again.

### Ability to increase rollout percentage

If we go with the current implementation where the admin cookie expiry is 2 months, we won't be able to ramp up the percentage progressively. This is because a user who gets randomly assigned AngularJS the first time (e.g. when rollout percentage is 1%) will continue being assigned AngularJS even when the rollout percentage is increased.

However, it is okay to have the actual percentage of admin users on React be higher than the rollout percentage (e.g. our percentage is 35% but actually 45% of users are on React).

So an implementation like this would work: when randomly choosing React/AngularJS, if choosing React, set a 2-month expiry. If choosing AngularJS, set a session expiry. Then when the rollout percentage is increased, those previously on AngularJS would be randomly assigned again, while those on React would stay on React. This means that e.g. when we increase rollout percentage to 10%, in reality it will be 10% + whoever was previously assigned React, but this is fine.

wanlingt:
## Problem
Currently, the React rollout for admins is either for all or none.

Closes #4217 

## Solution

Implement percentage-based React rollout for admins
- We can change the rollout percentage by changing the `REACT_MIGRATION_ADMIN_ROLLOUT` env var
- Depending on the % threshold, a randomly chosen selection of admins will be served the React app
- Admins who are shown the React app, or actively choose to switch to React will **stay on React** (admin cookie expiry is 2 months long)
- Admins who are shown the AngularJS app will remain on it **for that session** (session cookie)

**Breaking Changes** 

## Tests
- [ ] Change the admin rollout percentage to 100. Delete the `v2-admin-ui` cookie if it exists. Navigate to the forms homepage, you should be shown the React app. The `v2-admin-ui` cookie should be set to `react`
- [ ] Change the admin rollout percentage to 0. Delete the `v2-admin-ui` cookie if it exists. Navigate to the forms homepage, you should be shown the AngularJS app. The` v2-admin-ui` cookie should be set to `angular`
 
## Deploy Notes

- `REACT_MIGRATION_ADMIN_ROLLOUT ` : number between 0 and 100

### DIFF
--- a/src/app/config/schema.ts
+++ b/src/app/config/schema.ts
@@ -303,6 +303,12 @@ export const optionalVarsSchema: Schema<IOptionalVarsSchema> = {
       default: 0,
       env: 'REACT_MIGRATION_RESP_ROLLOUT_AUTH',
     },
+    adminRollout: {
+      doc: 'Percentage threshold to serve React for admins',
+      format: 'int',
+      default: 0,
+      env: 'REACT_MIGRATION_ADMIN_ROLLOUT',
+    },
     respondentCookieName: {
       doc: "Name of the cookie that will store respondents' assigned environment.",
       format: String,
--- a/src/app/loaders/express/logging.ts
+++ b/src/app/loaders/express/logging.ts
@@ -17,6 +17,7 @@ type LogMeta = {
   reactMigration?: {
     respRolloutAuth: number
     respRolloutNoAuth: number
+    adminRollout: number
     qaCookie: string | undefined
     adminCookie: string | undefined
     respCookie: string | undefined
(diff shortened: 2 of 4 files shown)


karrui:
seems to still be a draft. will review when PR is ready for review.

wanlingt:
> seems to still be a draft. will review when PR is ready for review.

all good now sorry! wanted to double check sth

timotheeg (approved):
LGTM

Issue #4372, opened 2022-07-26
Trim answer inputs where they are trimmed in AngularJS


mantariksh:
Preserve existing trimming behaviour for answers

karrui:
closed by #4430 i hope

PR #4430, opened 2022-07-29
feat(v2): add stronger validation for esrvcId and Twilio credential inputs

karrui:
## Problem

Cleaned up esrvcId and Twilio credential inputs. Trim and add validation for them.

Related to #4372, not sure if I missed any tbh.

## Solution

**Features**:

- feat: use rhf to add proper validation to esrvcId input
- feat: correctly trim/clear/update twilio credentials on update/delete
- feat: add toggle to show/hide API secret during input

### DIFF
--- a/frontend/src/features/admin-form/settings/components/TwilioSettingsSection/TwilioDetailsInputs.tsx
+++ b/frontend/src/features/admin-form/settings/components/TwilioSettingsSection/TwilioDetailsInputs.tsx
@@ -1,13 +1,12 @@
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useMemo } from 'react'
 import {
   RegisterOptions,
   useForm,
   UseFormRegisterReturn,
 } from 'react-hook-form'
-import { BiHappyHeartEyes, BiHide, BiShow } from 'react-icons/bi'
+import { BiHide, BiShow } from 'react-icons/bi'
 import {
   FormControl,
-  Icon,
   InputGroup,
   InputRightElement,
   Skeleton,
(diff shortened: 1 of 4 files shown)


frontend/src/features/admin-form/settings/components/TwilioSettingsSection/TwilioDetailsInputs.tsx
```diff
@@ -26,48 +38,78 @@ const TWILIO_INPUT_RULES: Record<keyof TwilioCredentials, RegisterOptions> = {
...
   messagingServiceSid: {
     required: 'Messaging service SID is required',
     pattern: {
       value: /^MG/,
       message: 'Messaging service SID must start with MG',
     },
+    validate: {
+      noWhitespace: (value) =>
+        !value.trim().match(/\s/) ||
+        'Messaging service SID must not contain whitespace',
```
timotheeg: /nit: The `trim()` implies we want to allow accidental leading or trailing whitespaces, but the pattern check above disallows leading whitespace 🤔 . We could probably remove the pattern check to do it in the validator instead, or add optional leading whitespaces in the pattern (e.g. `/^\s*MG/`). And same comment for the checks above.

Also nit: all 4 validators are nearly identical pure functions (minus the error message), so we could have a generator in the module for them.

karrui: inlined whitespace check into pattern validator: 353ba39

timotheeg: Thanks!

Although, the regex updates shouldn't check for trailing whitespace, since there is no tail anchor.

timotheeg (commented):
Oh, I didn't realize you had dropped the validator to shift the whole check to a pattern. The patterns are not equivalent to the previous code in their current state, since they'd allow a string like this: `  MG. 1. A.  `, which even after `.trim()` wwould contain whitespaces (which the validator was previously disallowing).

We can make this work with patten alone, but I'd recommend using this then:
```javascript
/^\s*MG\S+\s*$/
```
(and similar change for the other regexes)

This guarantees:
1. only leading and trailing whitespaces allowed (they'll be trimmed later when grabbing the value)
2. at least one character of "value" beyond the mandatory prefix

If we know more about the IDs (numeric only? alphanum only? fixed number of characters?), we could replace `\S+` with something more specific too.

karrui:
> If we know more about the IDs (numeric only? alphanum only? fixed number of characters?), we could replace `\S+` with something more specific too.

unfortunately twilio does not give any guarantees on the shape of those inputs. 
add new regex in 83e4d44

timotheeg (approved):
lgtm

PR #4431, opened 2022-07-31
feat(form-v2/restrict-particular-days-of-the-week): edit date picker callback for invalid days

siddarth2824:
## Problem

Currently the date picker does not check if a day in the date picker is considered an invalid day. Hence public users will still be able to click days from the date picker that are considered invalid based on the days the form admin users have restricted.

## Solution

Edit callback function in `DateField` to also check if a day in the date picker is considered to be an invalid day.

**Breaking Changes** 

### DIFF
--- a/frontend/src/templates/Field/Date/DateField.tsx
+++ b/frontend/src/templates/Field/Date/DateField.tsx
@@ -38,11 +38,11 @@ export const DateField = ({
       const selectedInvalidDays = schema.invalidDays ?? []
 
       // All dates available.
-      if (!selectedDateValidation && selectedInvalidDays.length === 0)
+      if (!selectedDateValidation && selectedInvalidDays.length === 0) {
         return false
+      }
 
       let isDateUnavailable = false
-      let isDayUnavailable = false
 
       switch (selectedDateValidation) {
         case DateSelectedValidation.NoPast:
@@ -64,8 +64,9 @@ export const DateField = ({
           isDateUnavailable = false
       }
 
-      isDayUnavailable = isDateAnInvalidDay(date, selectedInvalidDays)
-      return isDateUnavailable || isDayUnavailable
+      if (isDateUnavailable) return true
+
+      return isDateAnInvalidDay(date, selectedInvalidDays)
     },
     [schema.dateValidation, schema.invalidDays],
   )
--- a/frontend/src/templates/Field/Date/DateField.tsx
+++ b/frontend/src/templates/Field/Date/DateField.tsx
@@ -6,6 +6,7 @@ import { DateSelectedValidation } from '~shared/types/field'
 
 import {
   isDateAfterToday,
+  isDateAnInvalidDay,
   isDateBeforeToday,
   isDateOutOfRange,
 } from '~utils/date'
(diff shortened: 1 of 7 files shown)


frontend/src/utils/date.ts
```diff
@@ -160,3 +160,39 @@ export const transformCheckedBoxesValueToInvalidDays = (
   const validDaysSet = new Set(validDays)
   return ALL_INVALID_DAYS_ARR.filter((day) => !validDaysSet.has(day))
 }
+
+const DAY_TO_NUMBER_MAP: Record<InvalidDaysOptions, number> = {
```
karrui: can we not use the constant previously created? this seems like a duplicate constant

frontend/src/utils/date.ts
```diff
@@ -160,3 +160,39 @@ export const transformCheckedBoxesValueToInvalidDays = (
...
+  [InvalidDaysOptions.Saturday]: 6,
+  [InvalidDaysOptions.Sunday]: 7,
+}
+
+/**
+ * Convert the days of the week in the invalidDays array
+ * to a number array representing the number representation
+ * of the corresponding day of the week
+ */
+export const convertInvalidDaysOfTheWeekToNumberSet = (
```
karrui: same here. maybe we should make it a shared util if it is not already one

frontend/src/utils/fieldValidation.ts
```diff
@@ -391,6 +396,16 @@ export const createDateValidationRules: ValidationRuleFn<DateFieldBase> = (
           'Selected date is not within the allowed date range'
         )
       },
+      invaliDays: (val) => {
```
karrui: nit: typo? 
```suggestion
      invalidDays: (val) => {
```

frontend/src/utils/fieldValidation.ts
```diff
@@ -391,6 +396,16 @@ export const createDateValidationRules: ValidationRuleFn<DateFieldBase> = (
...
+        if (!val || !schema.invalidDays || schema.invalidDays.length === 0) {
+          return true
+        }
+
+        return (
+          !isDateAnInvalidDay(parseISO(val), schema.invalidDays) ||
+          'Error placeholder'
```
karrui: maybe we should think of an error message to put here instead of a placeholder. suggest  
```suggestion
          'Selected date is not an allowed day of the week'
```

frontend/src/utils/date.ts
siddarth2824: I wanted to do it but couldn't find the shared file that both files could import from. But I found it now. I will abstract it out into a shared function.

frontend/src/utils/fieldValidation.ts
siddarth2824: I was waiting for Pearly to update me with the error message. Let me confirm with her and update it in this PR.

siddarth2824: Latest commit contains the confirmed error message from Pearly.

karrui: not in commit yet

siddarth2824: Broke it down into two commits due to tests I added for util functions

karrui: 3speedy5u

frontend/src/templates/Field/Date/DateField.tsx
```diff
@@ -34,23 +35,39 @@ export const DateField = ({
   const isDateUnavailable = useCallback(
     (date: Date) => {
       const { selectedDateValidation } = schema.dateValidation
+      const selectedInvalidDays = schema.invalidDays ?? []
+
       // All dates available.
-      if (!selectedDateValidation) return false
+      if (!selectedDateValidation && selectedInvalidDays.length === 0)
+        return false
```
karrui, line 42: nit: wrap with braces if on multiple lines

frontend/src/templates/Field/Date/DateField.tsx
```diff
@@ -34,23 +35,39 @@ export const DateField = ({
...
+          )
+          break
         }
         default:
-          return false
+          isDateUnavailable = false
       }
+
+      isDayUnavailable = isDateAnInvalidDay(date, selectedInvalidDays)
+      return isDateUnavailable || isDayUnavailable
```
karrui: perf: should we return first if `isDateUnavailable` is true so it is (ever so slightly) more efficient?

karrui (approved):
lgtm

frontend/src/templates/Field/Date/DateField.tsx
siddarth2824: Yes we could return first if `isDateUnavailable` is true.

Issue #4336, opened 2022-07-26
Submit another form button leads to landing page


tutntut:

karrui:
closed by #4434

PR #4434, opened 2022-07-31
fix(form-v2): submit another form now returns to form page

tohhengxing:
## Problem

Submit another form button leads to landing page

Closes [#4336 ]

## Solution

removed faulty ?? operator

**Breaking Changes** 

**Bug Fixes**:
- `a ?? b` will return a if a is not null or undefined. `endPage.buttonLink` was returning empty string if alternative link not provided hence giving unexpected behaviour as outlined under
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

## Before & After Screenshots

**BEFORE**:

<details>

https://user-images.githubusercontent.com/102740235/182036912-93025202-3e4c-4aea-9f1d-d3d94b1bf717.mp4
</details>

**AFTER**:

<details>

https://user-images.githubusercontent.com/102740235/182036914-3236d71e-36e9-4cc8-8138-fbb621c8e46d.mp4
</details>

### DIFF
--- a/frontend/src/features/public-form/components/FormEndPage/components/EndPageBlock.tsx
+++ b/frontend/src/features/public-form/components/FormEndPage/components/EndPageBlock.tsx
@@ -78,7 +78,7 @@ export const EndPageBlock = ({
       <Box mt="2.25rem">
         <Button
           as="a"
-          href={endPage.buttonLink ?? window.location.href}
+          href={window.location.href}
           variant="solid"
           colorScheme={`theme-${colorTheme}`}
         >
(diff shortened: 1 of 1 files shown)


frontend/src/features/public-form/components/FormEndPage/components/EndPageBlock.tsx
```diff
@@ -78,7 +78,11 @@ export const EndPageBlock = ({
       <Box mt="2.25rem">
         <Button
           as="a"
-          href={endPage.buttonLink ?? window.location.href}
+          href={
+            endPage.buttonLink?.length
+              ? endPage.buttonLink
+              : window.location.href
+          }
```
timotheeg: The null coalescing operator was too strict, but this case is well suited for the [logical or](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR) instead. (`a || b`, returns `b` is `a` is falsy, including being empty string)

```suggestion
          href={endPage.buttonLink || window.location.href}
```

timotheeg (approved):

PR #4440, opened 2022-08-01
feat(form-v2): add error handling for deleted logic fields

### LINKED ISSUE #4341: Logic page is not accessible when a field with logic has been deleted

justynoh:
## Problem
Builder logic breaks when a user deletes a field that was used in logic. This PR fixes that by adding error handling to the various components.

Closes #4341 

## Solution

`LogicFieldBadge.tsx`, `InactiveLogicBlock.tsx`
- Added error/info state props to `LogicFieldBadge` to display in case the field data cannot be found.
- `InactiveLogicBlock` provides these props to the `LogicFieldBadge` component.

`useAdminFormLogic.ts`, `LogicContent.tsx`
- Added computation to determine globally if any errors have occured within all the logics. Added an error box to display message in such case.

`EditConditionBlock.tsx`, `ThenShowBlock.tsx`
- Add effects to set form errors and values for deleted fields, as well as display info boxes where necessary.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/182070801-42e0a925-3225-41a1-9493-c953cc677113.mov

### DIFF
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
@@ -40,6 +40,7 @@ export const ThenShowBlock = ({
     resetField,
     setError,
     control,
+    trigger,
   } = formMethods
 
   const logicTypeValue = watch('logicType')
@@ -87,7 +88,7 @@ export const ThenShowBlock = ({
   useEffect(() => {
     if (
       logicTypeValue !== LogicType.ShowFields ||
-      !showValueWatch.value ||
+      !showValueWatch.value?.length ||
       !mapIdToField
     )
       return
@@ -109,6 +110,7 @@ export const ThenShowBlock = ({
       setValue('show', filteredShowFields)
       return
     }
+    trigger('show')
   }, [
     logicTypeValue,
     mapIdToField,
@@ -116,6 +118,7 @@ export const ThenShowBlock = ({
     setError,
     setValue,
     showValueWatch.value,
+    trigger,
   ])
 
   return (
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
@@ -94,8 +94,6 @@ export const ThenShowBlock = ({
     const filteredShowFields = showValueWatch.value.filter(
       (field) => field in mapIdToField,
     )
-    const deletedFieldsCount =
-      showValueWatch.value.length - filteredShowFields.length
     if (filteredShowFields.length === 0) {
       resetField('show')
       setError('show', {
@@ -104,6 +102,8 @@ export const ThenShowBlock = ({
       })
       return
     }
+    const deletedFieldsCount =
+      showValueWatch.value.length - filteredShowFields.length
     if (deletedFieldsCount > 0) {
       setDeletedFieldsCount(deletedFieldsCount)
       setValue('show', filteredShowFields)
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
@@ -74,11 +74,13 @@ export const ThenShowBlock = ({
       : 'show'
   }, [logicTypeValue])
 
-  const [deletedFieldsCount, setDeletedFieldsCount] = useState<number>(0)
+  const [deletedFieldsCount, setDeletedFieldsCount] = useState(0)
 
   /**
    * Compute whether any/all fields in the show fields are deleted, then run
    * effect to delete fields on render and show appropriate error/infobox.
+   * useWatch here to avoid infinite re-render (since if there are deleted
+   * fields, we always reset the value of the show).
    */
   const showValueWatch = useWatchDependency(watch, 'show')
 
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
@@ -105,7 +105,7 @@ export const ThenShowBlock = ({
     if (filteredShowFields.length === 0)
       setError('show', {
         type: 'manual',
-        message: 'All fields were deleted. Please select other field(s).',
+        message: 'All fields were deleted, please select at least one field',
       })
     else setDeletedFieldsCount(deletedFieldsCount)
   }, [
@@ -126,13 +126,11 @@ export const ThenShowBlock = ({
       px={{ base: '1.5rem', md: '2rem' }}
     >
       {deletedFieldsCount ? (
-        <InlineMessage variant="info">
+        <InlineMessage variant="info" p={0}>
           <Text>
             <strong>{simplur`${deletedFieldsCount} Show field[|s]`}</strong>{' '}
-            {simplur`${[
-              deletedFieldsCount,
-            ]}[was|were] deleted. [It has|They have]`}{' '}
-            been removed from your logic.
+            {simplur`${[deletedFieldsCount]}[was|were] deleted, and [has|have]`}{' '}
+            been removed from your logic
           </Text>
         </InlineMessage>
       ) : null}
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
@@ -1,8 +1,9 @@
-import { useEffect, useMemo } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import { Controller, UseFormReturn } from 'react-hook-form'
 import { BiShow, BiX } from 'react-icons/bi'
-import { FormControl, Stack } from '@chakra-ui/react'
+import { FormControl, Stack, Text } from '@chakra-ui/react'
 import get from 'lodash/get'
+import simplur from 'simplur'
 
 import { FormFieldDto } from '~shared/types/field'
 import { LogicType } from '~shared/types/form'
@@ -10,6 +11,7 @@ import { LogicType } from '~shared/types/form'
 import { useWatchDependency } from '~hooks/useWatchDependency'
 import { MultiSelect, SingleSelect } from '~components/Dropdown'
 import FormErrorMessage from '~components/FormControl/FormErrorMessage'
+import InlineMessage from '~components/InlineMessage'
 import Textarea from '~components/Textarea'
 
 import { BASICFIELD_TO_DRAWER_META } from '~features/admin-form/create/constants'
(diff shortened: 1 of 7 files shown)


frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
```diff
@@ -9,30 +10,73 @@ import { FormFieldWithQuestionNo } from '~features/form/types'
 import { LogicBadge } from './LogicBadge'
 
 interface FieldLogicBadgeProps {
-  field: FormFieldWithQuestionNo
+  field: FormFieldWithQuestionNo | null
+  errorType?: 'error' | 'info'
+  errorString?: string
 }
```
mantariksh: `'info'` is not really an `errorType`, maybe `'variant'`? optionally you could also type this more strongly:

```typescript
type FieldLogicBadgeProps = 
  | { type: 'info', field: FormFieldWithQuestionNo }
  | { type: 'error', message: string }
}
```

mantariksh: in fact if you look at it this way, the props required for the two cases are completely different, which is forcing us to use multiple `switch` statements and ternary operators all over the place. can consider splitting this into separate components. although it will result in a couple of lines of repeated code, it will reduce the case splits required everywhere else.

arguably in future we might need more types of `LogicBadge` and at some point it might make sense to refactor them into the same component, but for now imo these make more sense separately. up to you though!

frontend/src/features/admin-form/create/logic/hooks/useAdminFormLogic.ts
```diff
@@ -21,11 +23,29 @@ export const useAdminFormLogic = () => {
...
+        : form?.form_logics.some(
+            (logic) =>
+              // Logic is errored if some condition does not exist, or all the
+              // show fields do not exist.
+              logic.conditions.some(
+                (condition) => !(condition.field in mapIdToField),
+              ) ||
+              (logic.logicType === LogicType.ShowFields &&
+                logic.show.every((field) => !(field in mapIdToField))),
+          ),
```
mantariksh: double negatives are a bit confusing
```suggestion
    () => {
      if (!mapIdToField || !form) return false
      return form.form_logics.some(
         // etc
      )
    }
    }
```

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
```diff
@@ -70,13 +74,67 @@ export const ThenShowBlock = ({
       : 'show'
   }, [logicTypeValue])
 
+  const [deletedFieldsCount, setDeletedFieldsCount] = useState<number>(0)
```
mantariksh: is the `<number>` not inferred?

mantariksh: looks like we have a mix of `useState(0)` and `useState<number>(0)` haha, I prefer the former

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
```diff
@@ -70,13 +74,67 @@ export const ThenShowBlock = ({
...
+    if (
+      logicTypeValue !== LogicType.ShowFields ||
+      !showValueWatch.value ||
+      !mapIdToField
+    )
+      return
+    const filteredShowFields = showValueWatch.value.filter(
+      (field) => field in mapIdToField,
+    )
+    const deletedFieldsCount =
```
mantariksh, line 98: can move this line below, closer to where it's used

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
```diff
@@ -70,13 +74,67 @@ export const ThenShowBlock = ({
       : 'show'
   }, [logicTypeValue])
 
+  const [deletedFieldsCount, setDeletedFieldsCount] = useState<number>(0)
+
+  /**
+   * Compute whether any/all fields in the show fields are deleted, then run
+   * effect to delete fields on render and show appropriate error/infobox.
+   */
+  const showValueWatch = useWatchDependency(watch, 'show')
```
mantariksh, line 86: why do we need to `watch` this? does the component not automatically update when `formFields` updates?

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
justynoh: sure, i like types but type inference is also cool

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
justynoh: `formFields` only contains the fields in the form, not the logics. Also, we need to useWatch here since the effect right after sets the show value again, resulting in infinite rerender.

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
justynoh: ok!

frontend/src/features/admin-form/create/logic/hooks/useAdminFormLogic.ts
justynoh: will change!

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
justynoh: i might be wrong / there might be a better way to achieve this though!

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
justynoh: as discussed, `'error' | 'info'` are the default fallbacks when `field` is `null`. This type is not ideal but I think any other way introduces a lot of complexity into the code. For example, another possibility is that we can have another component that deals with error cases specifically (so `FieldLogicBadge` only takes in a field, and `ErrorLogicBadge` takes in the `{ variant: 'error' | 'info', message: string }`) but then the error-handling behavior is repeated everywhere you want a logic badge anyway, so it boils down to the same thing

justynoh: i'll change the names to `variant` and `message` though so its less confusing!

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
```diff
@@ -9,30 +10,74 @@ import { FormFieldWithQuestionNo } from '~features/form/types'
...
           wrapperStyles={{ display: 'flex' }}
         >
-          <Icon as={fieldMeta.icon} fontSize="1rem" />
+          {fieldMeta ? <Icon as={fieldMeta.icon} fontSize="1rem" /> : errorIcon}
         </Tooltip>
-        {field.questionNumber ? <Text>{field.questionNumber}.</Text> : null}
-        <Text noOfLines={1}>{field.title}</Text>
+        {field ? (
+          <>
+            {field.questionNumber ? <Text>{field.questionNumber}.</Text> : null}
```
timotheeg, line 69: Just checking: is that breaking the continuous numbering?

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
```diff
@@ -70,13 +74,69 @@ export const ThenShowBlock = ({
...
+   * effect to delete fields on render and show appropriate error/infobox.
+   * useWatch here to avoid infinite re-render (since if there are deleted
+   * fields, we always reset the value of the show).
+   */
+  const showValueWatch = useWatchDependency(watch, 'show')
+
+  useEffect(() => {
+    if (
+      logicTypeValue !== LogicType.ShowFields ||
+      !showValueWatch.value ||
```
timotheeg: /nit Since `showValueWatch.value` is an array, we could test the for its length here too and save the processing below:
```suggestion
      !showValueWatch.value || !showValueWatch.value.length ||
```

It's inconsequential btw, since an empty array will give the correct behaviour with `0-0=0` later on in the function.

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
```diff
@@ -9,30 +10,74 @@ import { FormFieldWithQuestionNo } from '~features/form/types'
...
+        )
+      case 'info':
+        return (
+          <Text textColor="primary.500" noOfLines={1}>
+            {defaults.message}
+          </Text>
+        )
+    }
+  }, [defaults])
+
```
timotheeg: Rather than duplicating the markup in these 2 blocks, how about storing the color as a local variable since it's the only thing that changes, and then place the jsx markup below?

Or is the current structure done on purpose with the possibility that 2 `useMemos()` get `undefined` in case of unexpected variants? Is that even possible?

Anyway, so I'm suggesting this:

```suggestion
  const errorColor = useMemo(
    () => defaults.variant === 'error' ? 'danger.500' : 'primary.500', 
    [defaults.variant]
  )
```

And then later in the markup, replace `errorIcon` by:
```javascript
<Icon as={BxsErrorCircle} fontSize="1rem" color={errorColor} />
```

and `errorText` by:
```javascript
<Text textColor={errorColor} noOfLines={1}>
  {defaults.message}
</Text>
```

justynoh: I see your point and I agree - i'll refactor to get a bit more modularity here!

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
justynoh: Nope, the question number is stored along with the field, so the question numbers will always match the numbering in the builder!

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/ThenShowBlock.tsx
justynoh: Actually I found a bug related to this, i didn't realize `[]` is truthy (come on js, really??) so actually the behavior was not what I intended. Changing to a length check fixed this!

timotheeg: Of course it's truthy! it's an object, all objects are truthy! :D

timotheeg (approved):

timotheeg (approved):
+1 again

PR #4445, opened 2022-08-01
feat(form-v2): disable edit field drawer fields when field is being saved

### LINKED ISSUE #4371: Builder should have loading state
When builder is loading, it appears like there are no form fields

justynoh:
## Problem
Currently, the builder drawer does not display that the form field is in a loading state, even though it is read only. This PR changes the edit field drawers to display that the builder is in a saving state by greying out the field.

## Solution
Use `isDisabled` everywhere instead of `isReadOnly`.

**Breaking Changes** 


mantariksh (approved):
lgtm but check with @karrui if there was a reason for using `isReadOnly` originally instead of `isDisabled`

karrui:
yes, there was a reason for `isReadOnly` vs `isDisabled`, mainly for a11y. Disabled elements are skipped by most screen readers. We should stick to read only, and update the stylings if needed.

karrui:
I would think the drawer action buttons being in a loading/disabled state would be sufficient hinting.

justynoh:
closing because unnecessary hahah

PR #4447, opened 2022-08-01
fix: augment myinfo data only after admin field state is added

### LINKED ISSUE #4363: Dropdown MyInfo fields do not display possible options before creation
To replicate: click a MyInfo field in the drawer to put it into the Creating state. Then, try and click the dropdown in the builder preview. It should say “No matching options”.

karrui:
## Problem

Previously, fields to be created will not be augmented with their myinfo dropdown options as the augmentation of myinfo dropdown options were already done prior to the creation of the new field.

This PR fixes that by only augmenting after all edit/creation data has been added to the builder fields.

Closes #4363.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- fix: augment myinfo data only after admin field state is added


justynoh (approved):
lgtm

PR #4448, opened 2022-08-01
feat(form-workspaces-be/7): add move forms from multiple source workspaces functionality

hanstirtaputra:
Builds on #4254 

Part of #4039 

## Solution
Add endpoint logic for POST `api/v3/admin/forms/move` for moving forms that could be in multiple different workspaces, to one destination workspace. This API will likely be used in the "All forms" workspace that could have forms from different workspaces

**Breaking Changes** 

### DIFF
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -81,7 +81,7 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
       {
         _id: workspaceId,
       },
-      { session },
+      { session, new: true },
     )
   }
 
--- a/src/app/models/workspace.server.model.ts
+++ b/src/app/models/workspace.server.model.ts
@@ -1,6 +1,11 @@
 import { ClientSession, Mongoose, Schema } from 'mongoose'
 
-import { IUserSchema, IWorkspaceModel, IWorkspaceSchema } from '../../types'
+import {
+  IFormSchema,
+  IUserSchema,
+  IWorkspaceModel,
+  IWorkspaceSchema,
+} from '../../types'
 
 export const WORKSPACE_SCHEMA_ID = 'Workspace'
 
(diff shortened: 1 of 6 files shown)


src/app/models/workspace.server.model.ts
```diff
@@ -91,6 +91,18 @@ const compileWorkspaceModel = (db: Mongoose): IWorkspaceModel => {
...
+  WorkspaceSchema.statics.removeFormIdsFromAllWorkspaces = async function ({
+    admin,
+    formIds,
+    session,
+  }: {
+    admin: IUserSchema['_id']
+    formIds: IFormSchema['_id'][]
+    session?: ClientSession
+  }) {
+    await this.updateMany({ admin }, { $pullAll: { formIds } }, { session })
```
mantariksh: why not `return` this so that the caller gets back information about the operation?

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -10855,4 +10864,117 @@ describe('admin-form.controller', () => {
...
+      await AdminFormController.updateFormsWorkspace(
+        MOCK_REQ,
+        mockRes,
+        jest.fn(),
+      )
+
+      expect(mockRes.json).toHaveBeenCalledWith(MOCK_WORKSPACE)
+    })
+
+    it('should return 403 when workspace is not found', async () => {
```
mantariksh: wrong test description?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2725,3 +2728,58 @@ export const handleUpdateTwilio = [
...
+ * Handler for POST /forms/move.
+ * @security session
+ *
+ * @returns 200 with the destination workspace
+ * @returns 401 when user does not exist in session
+ * @returns 403 when user does not have permissions to update the form's workspace
+ * @returns 404 when the destWorkspaceId is not found in the database
+ * @returns 500 when a database error occurs
+ */
+export const updateFormsWorkspace: ControllerHandler<
```
mantariksh: why is this file in `admin-form.controller` instead of `workspace.controller`? also the function name is the same as the function currently in `workspace.controller`, can we differentiate them better according to their respective purposes?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2725,3 +2728,58 @@ export const handleUpdateTwilio = [
...
+  const userId = (req.session as AuthedSessionData).user._id
+
+  return WorkspaceService.getWorkspace(destWorkspaceId)
+    .andThen((workspace) =>
+      WorkspaceService.verifyWorkspaceAdmin(workspace, userId),
+    )
+    .andThen(() =>
+      WorkspaceService.moveForms({
+        userId,
+        sourceWorkspaceId: '',
```
mantariksh: we are using the `moveForms` service in a way where it is clearly not intended to be used. when passing an empty string, an unnecessary database call gets made, and it only works because we know that the query won't find a workspace with an empty string as an ID.

instead, we should make it explicit that the `moveForms` function may or may not pull the forms from their current workspace (perhaps by passing a boolean flag to the function, or making it possible to pass `null` for the `sourceWorkspaceId`), and avoid making unnecessary database calls.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2725,3 +2728,58 @@ export const handleUpdateTwilio = [
...
+        destWorkspaceId,
+        formIds,
+      }),
+    )
+    .map((workspace) =>
+      workspace
```
mantariksh: same comment as in previous PR

mantariksh (commented):
since the `updateOne` in the previous PR is just a special case of the `updateMany` in this one, we can consider using `updateMany` for both. but we should pass the `workspaceId` when we have it so the db query doesn't search through multiple workspaces even when we know the specific one we want to update.

mantariksh:
closing this to clean up the open PRs until we revisit it!

PR #4451, opened 2022-08-01
fix: ensure admin builder page has overflow auto style for proper drawer scroll, update form view layout

### LINKED ISSUE #4305: Save button gets cut off
Screenshot 2022-07-25 at 7 00 03 PM

karrui:
## Problem

Something (something... i really can't find it in code) is injecting overflow='initial' into the component's styling, preventing the overflow state from being set...
This causes the 100% height drawer to be offset and thus will not be able to be scrolled to the bottom (since the real bottom is below the visible bottom) 

This PR fixes that, and the drawer can be correctly scrolled now. 

Closes #4305

## Solution

**Breaking Changes** 

**Bug Fixes**:

- fix: ensure admin builder drawer has overflow auto style


mantariksh (approved):
image

karrui:
reverting the change, will fix in future PR

timotheeg (approved):

karrui:
pearly caught a regression,.. the padding change had to go out after all hehehe

karrui:
Some styling regressions, will draft while i fix them

timotheeg (approved):
👍

Issue #4453, opened 2022-08-01
[React] Submission success page should not show sections and instructions


karrui:
Seems to be introduced in #4155, but there are few regressions in the way the app handles the rendering of various public form page states.

We should only show the form end page without other public form components such as sections and instructions. 

![Screenshot 2022-08-01 at 6 20 44 PM](https://user-images.githubusercontent.com/22133008/182128331-bd207dde-0b00-4ad5-a4f6-8dd6e99b63da.png)

Also related to #4443

timotheeg:
Done in #4455 4455

PR #4455, opened 2022-08-01
fix(form-v2): use publicform context to selectively render components in form end page

tohhengxing:
## Problem

We should only show the form end page without other public form components such as sections and instructions.

Closes [#4453]

## Solution

public form context provider has property `submissionData` which is not null when form is submitted successfully. utilized that to selectively render components at form end page.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

<details>

https://user-images.githubusercontent.com/102740235/182192377-49430351-6958-4189-a2e8-e2e3ee40a536.mp4
</details>

**AFTER**:

<details>

https://user-images.githubusercontent.com/102740235/182192398-16320d6f-65cc-4f7c-a31e-eacfa5dcafd8.mp4
</details>

### DIFF
--- a/frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
+++ b/frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
@@ -21,11 +21,12 @@ import { useFormSections } from '../FormFields/FormSectionsContext'
 
 import { SidebarLink } from './SidebarLink'
 
-export const SectionSidebar = (): JSX.Element => {
+export const SectionSidebar = (): JSX.Element | null => {
   const { activeSectionId, navigatedSectionTitle } = useFormSections()
   const {
     miniHeaderRef,
     sectionScrollData,
+    submissionData,
     isMobileDrawerOpen,
     onMobileDrawerClose,
   } = usePublicFormContext()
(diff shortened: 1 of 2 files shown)


frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
```diff
@@ -21,11 +21,12 @@ import { useFormSections } from '../FormFields/FormSectionsContext'
 
 import { SidebarLink } from './SidebarLink'
 
-export const SectionSidebar = (): JSX.Element => {
+export const SectionSidebar = (): JSX.Element | null => {
```
timotheeg: Why is this needed? It doesn't look like the changeset can make `SectionSidebar()` return `null`

timotheeg (commented):
Thanks for another contribution @tohhengxing

frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
tohhengxing: Oops that was when I used null initially and I forgot to remove it. Sincere apologies @timotheeg

timotheeg: No worries at all. Remove it in a new commit and we can merge!

tohhengxing: removed! thank you very much!

timotheeg (approved):

PR #4459, opened 2022-08-02
fix(form-v2): remove additional log in step when switching from AngularJS to React

### LINKED ISSUE #4287: User appears logged out (even though session still exists in cookies) when user switches between React/AngularJS
Due to how the app assigns logged in state (with a localStorage boolean), may need to assign on first load with a call to GET /user to ascertain logged in state instead so this edge case can be fixed

wanlingt:
## Problem
User appears logged out after switching from AngularJS to React

Closes #4287 

## Solution
- Instead of refreshing the `/#!/forms` page, navigate to `/workspace`
- Query for `user` to ascertain logged in state

**Breaking Changes** 

**Features**:

## Before & After Screenshots

**BEFORE**:
Recording 2022-08-02 at 12 16 09

**AFTER**:
Recording 2022-08-02 at 13 04 42

## Tests
- [x] Make sure `is-logged-in` is not present in localStorage. 
Log in to the AngularJS app. Click the banner message to switch to React. 
You should be brought to the Workspace page (`/workspace`) on the React app and remain logged in.
`is-logged-in` in localStorage should have the value `true`
`v2-admin-ui` cookie should have the value `react`.

### DIFF
--- a/frontend/src/contexts/AuthContext.tsx
+++ b/frontend/src/contexts/AuthContext.tsx
@@ -41,14 +41,18 @@ export const useAuth = (): AuthContextProps => {
 
 // Provider hook that creates auth object and handles state
 const useProvideAuth = () => {
+  const [isLocalStorageAuthenticated, setIsLocalStorageAuthenticated] =
+    useLocalStorage<boolean>(LOGGED_IN_KEY)
   // TODO #4279: Remove after React rollout is complete
   const { data: user } = useQuery<UserDto>(
     userKeys.base,
     () => fetchUser(),
     // 10 minutes staletime, do not need to retrieve so often.
-    { staleTime: 600000 },
+    {
+      staleTime: 600000,
+      onSuccess: () => setIsLocalStorageAuthenticated(true),
+    },
   )
-  const [isLocalStorageAuthenticated] = useLocalStorage<boolean>(LOGGED_IN_KEY)
   const isAuthenticated = isLocalStorageAuthenticated || !!user
 
   // Return the user object and auth methods
--- a/frontend/src/contexts/AuthContext.tsx
+++ b/frontend/src/contexts/AuthContext.tsx
@@ -1,7 +1,13 @@
 import { createContext, FC, useContext } from 'react'
+import { useQuery } from 'react-query'
+
+import { UserDto } from '~shared/types'
 
 import { LOGGED_IN_KEY } from '~constants/localStorage'
 import { useLocalStorage } from '~hooks/useLocalStorage'
+import { fetchUser } from '~services/UserService'
+
+import { userKeys } from '~features/user/queries'
 
 type AuthContextProps = {
   isAuthenticated?: boolean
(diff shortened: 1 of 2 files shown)


frontend/src/contexts/AuthContext.tsx
```diff
@@ -35,7 +41,15 @@ export const useAuth = (): AuthContextProps => {
 
 // Provider hook that creates auth object and handles state
 const useProvideAuth = () => {
-  const [isAuthenticated] = useLocalStorage<boolean>(LOGGED_IN_KEY)
+  // TODO #4279: Remove after React rollout is complete
+  const { data: user } = useQuery<UserDto>(
```
karrui, line 47: Comment: could we use the useUser hook directly? Or does the use user hook look at the localStorage value before enabling? Maybe we can add an override enabled prop if that is the case

karrui (approved):
Lgtm

frontend/src/contexts/AuthContext.tsx
wanlingt: >  Or does the use user hook look at the localStorage value before enabling? 

Yup correct

> Maybe we can add an override enabled prop if that is the case

ok i try

wanlingt: > Or does the use user hook look at the localStorage value before enabling? 

the `useUser` hook uses `useAuth` to get the localStorage value (so there's a circular reference). To get around this I replaced `useAuth` with `useLocalStorage` in `useUser`

> Maybe we can add an override enabled prop if that is the case

Ok! added this too

karrui: ah, now it looks messy. i reviewed it without completely looking at the hook. suggest reverting haha ><. The previous code was more readable, and considering this is "throwaway" code i think we can get away with some duplication.

or we can set the state of isAuthenticated to be contingent on return value of useUser instead of the other way round:

```tsx
export const useUser = (): UseUserReturn => {
  const [isAuthenticated, setIsAuthenticated] = useLocalStorage<boolean>(LOGGED_IN_KEY)

  const { data: user, isLoading } = useQuery<UserDto>(
    userKeys.base,
    () => fetchUser(),
    // 10 minutes staletime, do not need to retrieve so often.
    { 
      staleTime: 600000,
      // Removed enabled state, just retrieve every 10 minutes
      // new functions, update auth states
      onSuccess: () => setIsAuthenticated(true),
      onError: () => setIsAuthenticated(undefined) // remove login on error, may not be needed since we already have ApiService
    },
  )

  return {
    user: isAuthenticated ? user : undefined,
    isLoading,
  }
}

```

wanlingt: haha it's ok! I will revert :) I think that option's the least messy / easiest to remove after the rollout

frontend/src/contexts/AuthContext.tsx
```diff
@@ -35,7 +41,15 @@ export const useAuth = (): AuthContextProps => {
...
+    userKeys.base,
+    () => fetchUser(),
+    // 10 minutes staletime, do not need to retrieve so often.
+    { staleTime: 600000 },
```
karrui: should we still add an `onSuccess` to set localStorage to the authenticated state?

karrui (approved):

frontend/src/contexts/AuthContext.tsx
wanlingt: oh yes good point! added

PR #4462, opened 2022-08-02
feat(form-v2): always enable save in builder drawer

### LINKED ISSUE #4368: Save button should be enabled even when there are no changes
Otherwise after changing back and forth, user might not know why they cannot save the field

justynoh:
## Problem
We want to always enable save in builder drawer for start page and fields. 

Closes #4368 

## Solution
Remove the `isSaveEnabled` prop, and only disable saves while the form is loading. 

## Screenshots

https://user-images.githubusercontent.com/25571626/182326758-6712cc6d-3491-42b3-ac3c-d806fe7a18f5.mov


karrui (approved):
lgtm

timotheeg (approved):

wanlingt (approved):
lgtm!

PR #4476, opened 2022-08-03
fix(form-v2): replaced max title length constant with 24 

### LINKED ISSUE #4472: Long form titles should be wrapped
**Describe the bug**
Admin: Long titles cover the builder header
Public form: Long titles are truncated

**Expected behavior**
Form title should be wrapped

**Screenshots**
Admin
image

Public form
image

tohhengxing:
## Problem

Closes #4472 

## Solution
updated max length constant to 24.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

refer to #4472 

**AFTER**:

image

### DIFF
--- a/frontend/src/utils/formValidation.ts
+++ b/frontend/src/utils/formValidation.ts
@@ -3,7 +3,7 @@ import validator from 'validator'
 
 const MAX_EMAIL_LENGTH = 30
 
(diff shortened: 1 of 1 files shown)


karrui:
the issue was incorrectly referring to capping the form title length. It should be to truncate/update how we render the form title on the client. This PR may not be needed.

PR #4482, opened 2022-08-03
feat(form-v2): set opacity for fields hidden by logic in builder

### LINKED ISSUE #4442: Display hidden fields differently in builder
In Angular, any field initially hidden by logic is shown in grey. We should replicate this behavior in react.

image
image

justynoh:
## Problem
Fields hidden by default due to logic hiding are greyed out in the form builder in angular, but we don't show them as hidden in react.

Closes #4442 

## Solution
- Reuse `getVisibleFieldsId` from logic to determine which fields are displayed by default.
- Set opacity of field row to 30% when the field is not visible by default.
- Add tooltip for info on fields with lower opacity.

**Breaking Changes** 

## Screenshots

image

image


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -68,12 +68,14 @@ import { SectionFieldRow } from './SectionFieldRow'
...
   field: FormFieldDto
   index: number
+  isVisible: boolean
   isDraggingOver: boolean
 }
 
 export const FieldRowContainer = ({
   field,
   index,
+  isVisible,
```
timotheeg: I'm thinking this property at a JSX.Element level is a bit confusing, since `visibility` means something in CSS 🤔 .

Can we find a clearer name to represent what this is? Maybe something like `isHiddenByLogicRule` or similar?

justynoh: Alright! Will change to `isHiddenByLogic`.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -210,116 +213,123 @@ export const FieldRowContainer = ({
...
-            }}
-            flexDir="column"
-            align="center"
-            onClick={handleFieldClick}
-            onKeyDown={handleKeydown}
-            ref={ref}
+          <Tooltip
+            hidden={!isHiddenByLogic}
+            placement="top"
+            label="This field may be hidden by your form logic"
```
timotheeg, line 219: `may be` or `is`? Asking because the Tooltip's only control on its hidden state is `isHiddenByLogic` 🤔

justynoh: hahah. this was the copy that was agreed upon. In actual fact, the field is hidden by logic by default (i.e. on first load of the form). Could change the variable name to `isDefaultHiddenByLogic` if that feels better?

timotheeg: Nah, let's keep it. We can revisit later if needed.

timotheeg (approved):

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
karrui: may is correct, since a logic match will show the field yea?

Issue #4335, opened 2022-07-26
Email mode's recipients setting removes added user and space bar input does not work


tutntut:
- Space bar input cannot be used.

- Added users being removed happens 
Steps to replicate:
1) Go to settings in email mode
2) Type a new email
3) Click anywhere else on the UI

karrui:
closed by #4510

PR #4509, opened 2022-08-05
feat: add tag input component 

karrui:
## Problem

This PR adds the primitive `TagInput` component, which can also be added to the design system (later).
Mostly used for creating tags in an array format via an input. Keys that create a tag defaults to the Enter, Comma, and Space keys.

Part 1 of updating email mode recipient list for better input handling, and to be closer to the design.

Related to #4335

## Solution

**Breaking Changes** 

**Features**:

- Add `TagInput` component and theme.

## Before & After Screenshots
New stories have been added for the component

## Deploy Notes

**New dependencies**:

- `react-roving-tabindex` : for adding roving tab index to navigate between created tags in the input

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -34751,6 +34751,14 @@
         "react-router": "6.0.2"
       }
     },
+    "react-roving-tabindex": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/react-roving-tabindex/-/react-roving-tabindex-3.2.0.tgz",
+      "integrity": "sha512-WY5d46Y/QuS2zgXs4MgYJlmi3uIyq7M9UD4l3SWim8ClnyV1L3r84brxElLxEBXdKuWFUbll9mOBpxfcDZsCVQ==",
+      "requires": {
+        "warning": "^4.0.3"
+      }
+    },
     "react-scripts": {
       "version": "4.0.3",
       "resolved": "https://registry.npmjs.org/react-scripts/-/react-scripts-4.0.3.tgz",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -54,6 +54,7 @@
     "react-popper": "^2.3.0",
     "react-query": "^3.34.2",
     "react-router-dom": "^6.0.2",
+    "react-roving-tabindex": "^3.2.0",
     "react-swipeable": "^7.0.0",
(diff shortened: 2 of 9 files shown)


njenwei (approved):
lgtm! 

Not sure if you want to consider: -
- `backspace` for deleting tags when nothing is selected (could be something where the first keypress focuses on the last element, and the second keypress actually deletes it). Having to use the `delete` key seems quite unintuitive and also, I believe macs and iphones don't have a `delete` key.
- A prop to prevent users from entering duplicate tags? I'm not sure if there will be instances where the user would need to enter multiple tags - TagInput would seem like the wrong type of input for that case.

karrui:
> lgtm! 
> 
> 
> 
> Not sure if you want to consider: -
> 
> - `backspace` for deleting tags when nothing is selected (could be something where the first keypress focuses on the last element, and the second keypress actually deletes it). Having to use the `delete` key seems quite unintuitive and also, I believe macs and iphones don't have a `delete` key.
> 
> - A prop to prevent users from entering duplicate tags? I'm not sure if there will be instances where the user would need to enter multiple tags - TagInput would seem like the wrong type of input for that case.

Was following guidance from this [npm package](https://www.npmjs.com/package/@accessible-components/tag-input), which made quite a bit of sense for a11y. 

Noted on the duplicate, but maybe that can also be handled by the caller, hmmm.

justynoh (approved):
lgtm

1. I also think we should allow backspace for deleting tags purely because it's a feature that most other such tag inputs allow on other websites, and many people are likely to be used to that being a feature. Personally I don't think I would have tried delete - if backspace didn't work, I would probably just have resigned to clicking the X's individually without even trying delete. (Maybe also too used to the usual mental schema of "backspace = clear from left of cursor, delete = clear from right of cursor" from Word, for example?) Surely there must be a way to accommodate this without compromising a11y

2. When a user enters `,,` will result in a tag with content `,` being created, not sure if we want to code against that edge case.

karrui:
good points. i will just allow backspace for now. will also add guards for duplicates and empty tags.

karrui:
> 2\. When a user enters `,,` will result in a tag with content `,` being created, not sure if we want to code against that edge case.

i think this is still a valid case, what if someone wants to create a tag for comma hehe. Will leave this in.

PR #4510, opened 2022-08-05
feat: use new TagInput component for email respondents input

### LINKED ISSUE #4335: Email mode's recipients setting removes added user and space bar input does not work

karrui:
> This PR builds on #4509 

## Problem

Updates the email respondents input to use newly created `TagInput` component. 

Also closes #4335

## Solution

**Breaking Changes** 

**Features**:

- feat: use new TagInput component for email respondents input

## Before & After Screenshots
Stories should update

### DIFF
--- a/frontend/src/components/TagInput/TagInput.tsx
+++ b/frontend/src/components/TagInput/TagInput.tsx
@@ -39,9 +39,10 @@ export interface TagInputProps
   keyDownKeys?: string[]
   /**
    * Optional function to call to validate created tags.
-   * Should return `true` if tag is invalid
+   * Should return `true` if tag is valid.
+   * If no function is provided, all tags are considered valid.
    */
-  tagInvalidation?: (tag: string) => boolean
+  tagValidation?: (tag: string) => boolean
 
   /**
    * Optional function to call when input is blurred.
@@ -63,7 +64,7 @@ export const TagInput = forwardRef<TagInputProps, 'input'>(
       onBlur,
       keyDownKeys = ['Enter', ',', ' '],
       tagColorScheme = 'secondary',
-      tagInvalidation,
+      tagValidation = () => true,
       size,
       preventDuplicates = true,
       ...props
@@ -172,7 +173,7 @@ export const TagInput = forwardRef<TagInputProps, 'input'>(
                 isDisabled={inputProps.disabled}
                 key={index}
                 colorScheme={tagColorScheme}
-                isInvalid={tagInvalidation?.(tag)}
+                isInvalid={!tagValidation(tag)}
                 label={tag}
                 onClearTag={handleRemoveTag(index)}
                 onBlur={onBlur}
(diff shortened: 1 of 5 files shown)


frontend/src/utils/formValidation.ts
```diff
@@ -30,49 +30,31 @@ export const FORM_TITLE_VALIDATION_RULES: UseControllerProps['rules'] = {
...
+        new Set(emails).size === emails.length ||
+        'Please remove duplicate emails.'
+      )
+    },
+    maxLength: (emails: string[]) => {
+      return (
+        emails.length <= MAX_EMAIL_LENGTH ||
+        'Please limit number of emails to 30.'
+      )
+    },
```
timotheeg: I hadn't spotted this before, but since the 2 validators above filter the array for truthy entries, should we not do the same filtering here too too, to be consistent on what we accept?

```suggestion
    duplicate: (emails: string[]) => {
      const truthyEmails = emails.filter(Boolean)
      return (
        new Set(truthyEmails).size === truthyEmails.length ||
        'Please remove duplicate emails.'
      )
    },
    maxLength: (emails: string[]) => {
      return (
        emails.filter(Boolean).length <= MAX_EMAIL_LENGTH ||
        'Please limit number of emails to 30.'
      )
    },
```

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
```diff
@@ -90,12 +86,16 @@ const AdminEmailRecipientsInput = ({
...
+      rules={ADMIN_EMAIL_VALIDATION_RULES}
       render={({ field }) => (
-        <Input
-          value={transform.input(field.value)}
-          onChange={(e) => field.onChange(transform.output(e.target.value))}
-          onBlur={handleBlur}
+        <TagInput
+          placeholder="Separate emails with a comma"
+          {...field}
+          tagInvalidation={(tag) => !isEmail(tag)}
```
timotheeg: Double negative looks a little weird, why not in reverse?
```suggestion
          tagValidation={isEmail}
```

karrui: done. updated underlying prop for `TagInput` component 79746b4

frontend/src/utils/formValidation.ts
karrui: done ed08f8c

frontend/src/components/TagInput/TagInput.tsx
```diff
@@ -39,9 +39,10 @@ export interface TagInputProps
   keyDownKeys?: string[]
   /**
    * Optional function to call to validate created tags.
-   * Should return `true` if tag is invalid
+   * Should return `true` if tag is valid.
+   * If no function is provided, all tags are considered valid.
    */
-  tagInvalidation?: (tag: string) => boolean
+  tagValidation?: (tag: string) => boolean
```
mantariksh, line 45: why not rename this `isTagValid`? we normally use `is` or `has` prefix for functions which return boolean right

mantariksh (approved):
lgtm, minor point on prop name

frontend/src/components/TagInput/TagInput.tsx
karrui: Only for Boolean constants tbh, mixing them for both functions and constants might be confusing (even though I know we don't always follow that)

PR #4512, opened 2022-08-05
fix(v2): scroll to top of window on submission

### LINKED ISSUE #4337: Thank you page after submission leads to middle of the page rather than from the top

karrui:
## Problem

Closes #4337


wanlingt (approved):
looks goood

PR #4514, opened 2022-08-05
feat(form-v2/react-to-angular/3): remove infoboxes when rollout percentages hit thresholds

### LINKED ISSUE #3933: Switching from React -> AngularJS

wanlingt:
## Solution
Remove visibility of switch-back-to-AngularJS infoboxes after rollouts hit their respective % (10% for public respondents, 100% for admins)

Closes #3933 

**Details**
- Add 
`respondentRolloutEmail`, `respondentRolloutStorage`, `adminRollout` to frontend end vars
- Determine `showSwitchEnvMessage` boolean depending on whether the rollout % > the % threshold to remove the infobox

**Breaking Changes** 

## Tests
- [x] Set the rollout % for an email mode form below the % threshold for removal of the infobox. Infobox should be visible.
- [x] Set the rollout % for a storage mode form below the % threshold for removal of the infobox. Infobox should be visible.
- [x] Set the rollout % for admins below the % threshold for removal of the infobox. Infobox should be visible.
- [x] Repeat the 3 tests above, but set the rollout % to be above the % threshold for removal of the infobox. The infobox should no longer be seen.

### DIFF
--- a/frontend/src/features/public-form/components/PublicFormWrapper.tsx
+++ b/frontend/src/features/public-form/components/PublicFormWrapper.tsx
@@ -1,7 +1,9 @@
 import { useMemo } from 'react'
 import { Flex, Spacer } from '@chakra-ui/react'
 
-import { FormColorTheme } from '~shared/types'
+import { FormColorTheme, FormResponseMode } from '~shared/types'
+
+import { useEnv } from '~features/env/queries'
 
 import { usePublicFormContext } from '../PublicFormContext'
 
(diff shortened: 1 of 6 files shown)


timotheeg (approved):
lgtm

PR #4515, opened 2022-08-05
feat(CreatePageSidebarContext): store current field list tab

### LINKED ISSUE #4309: User should see MyInfo instead of Basic fields upon saving MyInfo field

karrui:
## Problem

So tab state is stored when user creates MyInfo field and returns

Closes #4309


wanlingt (approved):
lgtm!

PR #4516, opened 2022-08-05
[WIP] ref: OIDCClient classes

timotheeg:
## Problem
SPCP OIDC client are very similar and are a great use case for OO practices. The implementation makes the base classes for services and clients aware of the authType, which seems counter productive in a OO context.

This PR is opened as an example of what could be done, as basis for discussion.

## Solution

Use OO approaches to drop conditionals blocks on AuthType such that the subClasses just do the job (the subclass themselves representing the types and thus knowing what to do.

The PR is a draft intended to be a source material for discussion.

## Test

- [X] Update OIDCClient class and subclasses
- [x] Split OIDC service into 2 services
- [ ] Update all tests


tshuli:
@timotheeg finished up as discussed, with the exception that I chose to continue using abstract class for `createJwtPayload` so that the child class can implement their own attribute checks (ie for corppass, the child class needs the guarantee that `userInfo` exists, it cannot be an optional attribute)

timotheeg:
> finished up as discussed, with the exception that I chose to continue using abstract class for createJwtPayload so that the child class can implement their own attribute checks (ie for corppass, the child class needs the guarantee that userInfo exists, it cannot be an optional attribute)

Thank you 🙏! And that last decision makes perfect sense: the child classes should implement the requirements they need. My attempt at sharing code was misguided: it's much more important to keep the functional guarantees than to save some minor duplicate code.

PR #4528, opened 2022-08-08
feat(form-v2): hide headers in section sidebar if they are hidden by logic

### LINKED ISSUE #4506: Headers hidden by logic still show up in section sidebar
Create a form with a header hidden by logic

image

Go to form, header still shows up in sidebar

image

Clicking on the header link does nothing. Only when the header is visible, clicking on the header link brings the window to the header.

image

Should the header show up at all? Or should it be represented in the sidebar, in some other way (maybe as "Header _(hidden)_"?)? Tagging @pearlyong for design input.

Eng wise should not be too difficult - we have a function that creates the section refs and a function that determines the visible field IDs. We can just filter the section IDs with the visible IDs to determine which are visible and which are not in the sidebar.

justynoh:
## Problem
Currently, form headers are not hidden in the section sidebar if they are hidden in the form due to logic. This PR implements this feature. 

Closes #4506 

## Solution

`PreviewFormProvider`, `PublicFormContext`, `PublicFormProvider`, `FormSectionsContext`
- Move the scroll data computation into the Form Sections provider since it is shared between preview form and public form, and it is the same computation in either case.
- Add function to provider allowing for updating of scroll data based on visible field computation.
- Add check for visibility before adding to scroll data.

`SectionSidebar`
- Get scroll data from the form sections context rather than the form provider.

`VisibleFormFields`
- Update scroll data when visible fields are recalculated.

`SectionLink`
- Update reference to Sidebar type.

`FormFooter`
- Remove unneeded spacer component which references scroll data (was necessary before due to sidebar wrapping FormFields component, but we have since moved the sidebar to the form wrapper which also wraps the footer, thus now unnecessary).
 
**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/183372600-832c86e4-44fb-4e79-b35c-baceb797c937.mov

### DIFF
--- a/frontend/src/features/public-form/components/FormFields/FormSectionsContext.tsx
+++ b/frontend/src/features/public-form/components/FormFields/FormSectionsContext.tsx
@@ -19,7 +19,7 @@ export type SidebarSectionMeta = Pick<FormFieldDto, 'title' | '_id'>
 interface FormSectionsContextProps {
   /** Scroll data to allow form-fillers to scroll to a particular section. */
   sectionScrollData: SidebarSectionMeta[]
-  updateSectionScrollData: (visibleFieldIds: FieldIdSet) => void
+  setVisibleFieldIdsForScrollData: (visibleFieldIds: FieldIdSet) => void
   sectionRefs: Record<string, RefObject<HTMLDivElement>>
   activeSectionId?: string
   navigatedSectionTitle?: string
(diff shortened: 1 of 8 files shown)


frontend/src/features/public-form/components/FormFields/FormSectionsContext.tsx
```diff
@@ -73,6 +101,8 @@ export const FormSectionsProvider = ({
   return (
     <FormSectionsContext.Provider
       value={{
+        sectionScrollData,
+        updateSectionScrollData: setVisibleFieldIds,
```
karrui: should we set this name to what it actually does? for clarity instead of having to look at this context.

suggest `setVisibleFieldIdsForScrollData` haha

karrui (approved):
lgtm

frontend/src/features/public-form/components/FormFields/FormSectionsContext.tsx
justynoh: okie!

PR #4538, opened 2022-08-10
feat(form-v2): add MyInfo badge to form builder for MyInfo fields

### LINKED ISSUE #4441: MyInfo fields in the builder should be labelled
As they are in angular

image

justynoh:
## Problem
MyInfo fields currently are not distinguished from normal fields in the form builder. This PR adds the MyInfo badge to the relevant fields in the form builder.

Closes #4441 

## Solution
Add an additional base field prop, `showMyInfoBadge` (not called `isMyInfo` because we explicitly don't want to show the badge in the public form, only in the form builder). `FieldRowContainer` adds the badge itself. Also updates to props for Short Text, Dropdown and Date fields to accommodate additional field container props.

**Breaking Changes** 

## Screenshots

image

image

### DIFF
--- a/frontend/src/templates/Field/FieldContainer.tsx
+++ b/frontend/src/templates/Field/FieldContainer.tsx
@@ -5,11 +5,12 @@
  * provides.
  */
 import { FieldError, useFormState } from 'react-hook-form'
-import { FormControl } from '@chakra-ui/react'
+import { Box, FormControl, Grid } from '@chakra-ui/react'
 import { get } from 'lodash'
 
 import { FormColorTheme } from '~shared/types/form'
 
+import Badge from '~components/Badge'
 import FormErrorMessage from '~components/FormControl/FormErrorMessage'
 import FormLabel from '~components/FormControl/FormLabel'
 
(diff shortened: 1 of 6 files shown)


frontend/src/templates/Field/FieldContainer.tsx
```diff
@@ -57,14 +64,25 @@ export const FieldContainer = ({
       isInvalid={!!error}
       id={schema._id}
     >
-      <FormLabel
-        questionNumber={
-          schema.questionNumber ? `${schema.questionNumber}.` : undefined
-        }
-        description={schema.description}
+      <Grid
```
karrui, line 67: nit: i wonder if grid is needed. Is `Wrap` sufficient? then instead of hidden and adding grid areas we can just have conditional rendering ala

```tsx
{showMyInfoBadge && <Badge ... />}
```

karrui (approved):
lgtm with q

frontend/src/templates/Field/FieldContainer.tsx
justynoh: as discussed, we'll keep to using grid to avoid margin hell

PR #4545, opened 2022-08-11
feat: copy changes to label and error message

siddarth2824:
## Problem

Add copy changes to form section label and error message for restricting days of the week for form admin users.

For reference:
Label:
Screenshot 2022-08-11 at 11 17 31 AM

Error Message:
Screenshot 2022-08-11 at 11 18 34 AM

I only added the label and error message for days of the week and not for public holidays as the feature for restricting Singapore public holidays has not been implemented yet.

## Solution

**Breaking Changes** 

    - Previous `InvalidDaysOption` was days of the week that were singular. This change would make the days of the weeks plural. Hence previous values stored in the `InvalidDays` option would be invalid and cause an error during server side validation.


karrui (approved):
lgtm. can we also add a drawer story for when the day restriction toggle is on?

PR #4546, opened 2022-08-11
fix(form-v2): add default values to field templates to ensure optional fields can be submitted

### LINKED ISSUE #4508: Radio field cannot be submitted if made optional
Create a form with an optional Radio field. Attempt to submit without touching the radio field. This will fail with a `ZodError`.

image

All other fields are ok, only radio doesn't work!

justynoh:
## Problem
Currently, optional radio fields left empty cannot be submitted, resulting in a `ZodError`. However, if an option is selected then deselected, the form is submittable. This indicates an issue with the default value given to the radio fields before it is touched. In particular, the default value given is undefined, which is not accepted by zod.

Closes #4508 

## Solution
Provide a default value (empty string) for radio and checkboxes.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/184061294-bf8cdd70-3a75-4d5c-b4fa-a3e472c487e7.mov

### DIFF
--- a/frontend/src/templates/Field/Checkbox/CheckboxField.tsx
+++ b/frontend/src/templates/Field/Checkbox/CheckboxField.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, useEffect, useMemo } from 'react'
+import { forwardRef, useMemo } from 'react'
 import {
   get,
   useFormContext,
@@ -55,8 +55,7 @@ export const CheckboxField = ({
     [schema],
   )
 
-  const { register, getValues, resetField } =
-    useFormContext<CheckboxFieldInputs>()
+  const { register, getValues } = useFormContext<CheckboxFieldInputs>()
   const { isValid, isSubmitting, errors } = useFormState<CheckboxFieldInputs>({
     name: schema._id,
   })
@@ -76,11 +75,6 @@ export const CheckboxField = ({
       },
     }),
     [checkboxInputName, getValues],
-  )
-
-  useEffect(
-    () => resetField(checkboxInputName, { defaultValue: '' }),
-    [checkboxInputName, resetField],
   )
 
   return (
--- a/frontend/src/templates/Field/Checkbox/CheckboxField.tsx
+++ b/frontend/src/templates/Field/Checkbox/CheckboxField.tsx
@@ -84,6 +84,7 @@ export const CheckboxField = ({
           colorScheme={fieldColorScheme}
           key={idx}
           value={o}
+          defaultValue={[]}
           {...register(checkboxInputName, validationRules)}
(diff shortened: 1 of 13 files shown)


frontend/src/templates/Field/Checkbox/CheckboxField.tsx
```diff
@@ -77,6 +78,11 @@ export const CheckboxField = ({
     [checkboxInputName, getValues],
   )
 
+  useEffect(
+    () => resetField(checkboxInputName, { defaultValue: '' }),
+    [checkboxInputName, resetField],
+  )
```
karrui: could we achieve the same outcome without using useEffects? This seems overkill. Could we try assigning a `defaultValue` prop in the components instead?

```tsx 
<Checkbox
  colorScheme={fieldColorScheme}
  key={idx}
  value={o}
  defaultValue="" // <----
  {...register(checkboxInputName, validationRules)}
>
```

```tsx
<FieldContainer schema={schema} errorKey={radioInputName}>
  <Controller
    name={radioInputName}
    rules={validationRules}
    defaultValue="" // <------
    render={({ field: { ref, onChange, value, ...rest } }) => (
      <Radio.RadioGroup
        colorScheme={fieldColorScheme}
```

for safety, should we add a default value to all field templates?

karrui (changes requested):

frontend/src/templates/Field/Checkbox/CheckboxField.tsx
justynoh: yeah, adding the default value would definitely work haha

As for adding a default value to all fields, in some cases it is possible and some not. I'll add defaults to all possible cases.

frontend/src/templates/Field/Checkbox/CheckboxField.tsx
```diff
@@ -84,6 +84,7 @@ export const CheckboxField = ({
           colorScheme={fieldColorScheme}
           key={idx}
           value={o}
+          defaultValue={[]}
```
karrui: the `defaultValue` for this looks wrong. This is value for a single checkbox. Should it be `defaultValue=''` instead?

justynoh: fixed

justynoh:
errrrr rating field is broken, please hold

justynoh:
decided to leave rating without a default value. giving it any value other than `undefined` will cause other expected behavior to break.

karrui (approved):
lgtm

karrui:
just merge without UI checks being accepted whenever @justynoh

PR #4548, opened 2022-08-11
fix(v2): update rendering of start page theme radio selection

karrui:
## Problem

This PR simplifies the styling and semantics of the radio buttons used to select the current color theme of the form.


karrui:
changed design instead of simplifying booooo, closing PR

Issue #4547, opened 2022-08-11
Start page in admin preview resets when cache is invalidated


karrui:
We should not link the state during preview to the start page that was retrieved from the server. cloned data should be used instead.

![Recording 2022-08-11 at 11 44 23](https://user-images.githubusercontent.com/22133008/184062227-37638a99-ab89-44b7-8452-0e96d2fa9c6a.gif)

justynoh:
:thonk:

PR #4554, opened 2022-08-11
fix(form-v2): refactor to avoid repopulating design store when cache is invalidated

justynoh:
## Problem
In the current implementation, design store contents is reset to the original server state upon cache invalidation. Rather, we should retain the existing design store contents. 

The reason why this occurs in the current implementation is that the effect in `DesignDrawerContainer` is run once upon invalidation (with an empty form), and then once again once the form is refetched (thus repopulating the design store with the server data).

Closes [#4547 ]

## Solution
Move the population of design store with server state data from `DesignDrawerContainer` into `DesignDrawer`, which now wraps `DesignInput`. Thus, when the cache is invalidated and the form is refetching, we don't even render the design drawer, which means we also don't re-run the effect to repopulate the store.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/184094680-80bad1ce-aa6f-49f5-9099-7c7dab58a834.mov

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
@@ -21,7 +21,7 @@ import {
 } from '@chakra-ui/react'
 import { cloneDeep, get, isEmpty } from 'lodash'
 
-import { FormColorTheme, FormLogoState } from '~shared/types'
+import { FormColorTheme, FormLogoState, FormStartPage } from '~shared/types'
 
 import { useToast } from '~hooks/useToast'
 import { uploadLogo } from '~services/FileHandlerService'
@@ -31,6 +31,7 @@ import Radio from '~components/Radio'
 
 import { useMutateFormPage } from '~features/admin-form/common/mutations'
 import { useCreatePageSidebar } from '~features/admin-form/create/common/CreatePageSidebarContext'
+import { useEnv } from '~features/env/queries'
 import { getTitleBg } from '~features/public-form/components/FormStartPage/useFormHeader'
 
 import {
@@ -38,8 +39,11 @@ import {
   customLogoMetaSelector,
   DesignState,
   FormStartPageInput,
+  resetDesignStoreSelector,
+  setCustomLogoMetaSelector,
   setStartPageDataSelector,
   setStateSelector,
+  startPageDataSelector,
   stateSelector,
   useDesignStore,
 } from '../../useDesignStore'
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
```diff
@@ -53,31 +57,35 @@ import {
...
 
 type DesignDrawerProps = {
-  startPageData: FormStartPageInput
+  startPage: FormStartPage
 }
 
-export const DesignDrawer = ({
-  startPageData,
-}: DesignDrawerProps): JSX.Element | null => {
+export const DesignInput = (): JSX.Element | null => {
```
karrui, line 63: nit: should this be extracted to its own component?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
```diff
@@ -183,6 +191,203 @@ export const DesignDrawer = ({
 
   if (!startPageData) return null
 
+  return (
+    <DrawerContentContainer>
```
karrui, line 197: i assume this is a pure movement of code and has no changes from the previous implementation

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
```diff
@@ -183,6 +191,203 @@ export const DesignDrawer = ({
...
+                  type: 'image/jpeg',
+                }),
+                'size',
+                { value: startPage.logo.fileSizeInBytes },
+              ),
+              srcUrl: `${logoBucketUrl}/${startPage.logo.fileId}`,
+            },
+    })
+    if (startPage.logo.state === FormLogoState.Custom)
+      setCustomLogoMeta(startPage.logo)
```
karrui: nit: can multiline conditionals be wrapped in braces for consistency in the codebase plsssssss

karrui (approved):
lgtm

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
justynoh: yepp

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
justynoh: we could but i don't think it's really necessary, `DesignInput` isn't reusable in any way (unlike, for instance, `EditFieldContainer` which can take in various `Edit<X>` as children)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/DesignDrawer/DesignDrawer.tsx
justynoh: sorry senpai 🙏

PR #4561, opened 2022-08-12
fix(form-v2): disable some MyInfo fields in builder and preview

### LINKED ISSUE #4362: Answers for MyInfo fields in builder should not be editable
e.g. "PHUA CHU KANG" for Name should not be editable, to more easily differentiate MyInfo fields from regular fields

wanlingt:
## Problem

MyInfo fields appear editable in the form builder and preview. Some fields (e.g. name, DOB) should not appear editable to be consistent with public forms

Closes #4362

## Solution

Add `previewIsDisabled` prop to `MyInfoBlock`, which contains the values that are injected into the form builder's and preview's MyInfo fields. `previewIsDisabled` is determined using the mapping from `MyInfoData` class.

**Improvements**:
- Move `isDisabled` from `FieldContainer` to form components used by MyInfo. This is because `isDisabled` in `FieldContainer` is set in `FormControl`, which [disables both the form label and the form element](https://chakra-ui.com/docs/components/form-control/props). As per the [design system](https://www.figma.com/file/u00dlQLsaQfvqv6ot5ZrJZ/Form-Design-System?node-id=2153%3A4682), only the form element itself should be disabled.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
<details>

screencapture-localhost-3000-admin-form-62f366fe51036800a31c65ac-2022-08-12-10_25_58
</details>

**AFTER**:
<details>

screencapture-localhost-3000-admin-form-62f366fe51036800a31c65ac-2022-08-12-10_36_10
</details>

### DIFF
--- a/frontend/src/features/myinfo/utils/augmentWithMyInfoDisplayValue.ts
+++ b/frontend/src/features/myinfo/utils/augmentWithMyInfoDisplayValue.ts
@@ -13,5 +13,10 @@ export const augmentWithMyInfoDisplayValue = (
   if (!isMyInfo(field)) return field
 
   const myInfoBlock = MAP_ATTR_TO_NAME[field.myInfo.attr]
-  return { ...field, fieldValue: myInfoBlock.previewValue }
+  return {
+    ...field,
+    fieldValue: myInfoBlock.previewValue,
+    // Leave default as false, same as in MyInfoData class
+    disabled: myInfoBlock.previewIsDisabled ?? false,
+  }
 }
--- a/frontend/src/templates/Field/FieldContainer.tsx
+++ b/frontend/src/templates/Field/FieldContainer.tsx
@@ -59,7 +59,6 @@ export const FieldContainer = ({
   return (
     <FormControl
       isRequired={schema.required}
-      isDisabled={schema.disabled}
       isReadOnly={isValid && isSubmitting}
       isInvalid={!!error}
       id={schema._id}
--- a/shared/constants/field/myinfo/index.ts
+++ b/shared/constants/field/myinfo/index.ts
@@ -21,6 +21,9 @@ export type MyInfoFieldBlock = {
   // for MyInfo forms. The running joke is that this is the personal
   // details of Phua Chu Kang, a famous singaporean sitcom.
   previewValue: string
+  // Refers to whether the field should be editable. The mapping is derived
+  // from MyInfoData class
+  previewIsDisabled?: boolean
 }
 
 export const types: MyInfoFieldBlock[] = [
@@ -34,6 +37,7 @@ export const types: MyInfoFieldBlock[] = [
       'The registered name of the form-filler. This field is verified by ICA for Singaporeans/PRs & foreigners on Long-Term Visit Pass, and by MOM for Employment Pass holders.',
     fieldType: BasicField.ShortText,
     previewValue: 'PHUA CHU KANG',
+    previewIsDisabled: true,
   },
   {
     name: MyInfoAttribute.Sex,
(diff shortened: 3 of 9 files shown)


justynoh (approved):

frontend/src/templates/Field/FieldContainer.tsx
```diff
@@ -59,7 +59,6 @@ export const FieldContainer = ({
   return (
     <FormControl
       isRequired={schema.required}
-      isDisabled={schema.disabled}
```
karrui, line 62: any reason why we removed this prop from the base container? this container should be the ultimate controller of field state instead of being in the children imho

frontend/src/features/myinfo/utils/augmentWithMyInfoDisplayValue.ts
```diff
@@ -13,5 +13,10 @@ export const augmentWithMyInfoDisplayValue = (
   if (!isMyInfo(field)) return field
 
   const myInfoBlock = MAP_ATTR_TO_NAME[field.myInfo.attr]
-  return { ...field, fieldValue: myInfoBlock.previewValue }
+  return {
+    ...field,
+    fieldValue: myInfoBlock.previewValue,
+    // Leave default as false, same as in MyInfoData class
+    disabled: myInfoBlock.previewIsDisabled ?? false,
```
karrui, line 20: seems like this may have a bug. The original `field.disabled` prop is ignored and overwritten by `false`.

```suggestion
    disabled: myInfoBlock.previewIsDisabled ?? field.disabled,
```

depending on the priority, hehe

shared/constants/field/myinfo/index.ts
```diff
@@ -34,6 +37,7 @@ export const types: MyInfoFieldBlock[] = [
       'The registered name of the form-filler. This field is verified by ICA for Singaporeans/PRs & foreigners on Long-Term Visit Pass, and by MOM for Employment Pass holders.',
     fieldType: BasicField.ShortText,
     previewValue: 'PHUA CHU KANG',
+    previewIsDisabled: true,
```
karrui, line 40: i wonder if we should discern what is or is not disabled on runtime (like how public forms are displayed) based on the `verified` prop instead of hardcoding it here.

Also wondering if we should actually disable the state in the builder, if it's disabled won't the admins be unable to preview myinfo options properly?

karrui (changes requested):

frontend/src/templates/Field/FieldContainer.tsx
wanlingt: I initially used the base container's isDisabled, but I realised this also disables the form label e.g.
image
which is different from the form design system's [disabled field](https://www.figma.com/file/u00dlQLsaQfvqv6ot5ZrJZ/Form-Design-System?node-id=2153%3A4682), hence moved it to the child component

karrui: we could update the design for a disabled form label instead i think. This is changing logic to fulfil styling, which may introduce subtle bugs in the future

wanlingt: ah i see ok makes sense! will look into it

shared/constants/field/myinfo/index.ts
wanlingt: just discussed with @mantariksh and @pearlyong , tldr will leave the MyInfo fields editable so will close this PR. Have closed the [related issue](https://github.com/opengovsg/FormSG/issues/4362#issuecomment-1214579995)

PR #4562, opened 2022-08-12
refactor(v2): reuse public end page component in builder

justynoh:
## Problem
End page view recreates the public form end page from scratch. This PR refactors it to make use of the public end page component for better modularity.

## Solution
To account for the accordion being expanded, add an optional `isExpandable` to `EndPageBlock` which defaults to `true`. Other than that, some layout simplifications to improve readability while retaining the existing design.

**Breaking Changes** 


wanlingt (approved):
lgtm!

PR #4563, opened 2022-08-12
refactor(v2): rename `BuilderAndDesignStore` to `FieldBuilderStore`

justynoh:
## Problem
`BuilderAndDesignStore` does not accurately reflect its use in the code anymore. In particular, end page and design have their own stores, so we should rename this to `FieldBuilderStore`.

## Solution
Rename `BuilderAndDesignStore` to `FieldBuilderStore` everywhere.

**Breaking Changes** 


karrui (approved):
lgtm

PR #4569, opened 2022-08-14
fix(form-v2): change css value to align logic button

### LINKED ISSUE #4541: Move logic + button to align with first logic block
Currently the + is at the top of the page

image

In Figma it's beside the first block

image

tohhengxing:
## Problem

align logic button with first block as per figma

Closes #4541 

## Solution

change css value

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:

<details>
scrolling works too

https://user-images.githubusercontent.com/102740235/184540021-91ad8d1d-937f-4875-9736-e01a1a91939d.mp4
</details>

### DIFF
--- a/frontend/src/features/admin-form/create/logic/CreatePageLogicTab.tsx
+++ b/frontend/src/features/admin-form/create/logic/CreatePageLogicTab.tsx
@@ -52,7 +52,7 @@ export const CreatePageLogicTab = (): JSX.Element => {
             <IconButton
               zIndex="docked"
               pos={{ base: 'fixed', md: 'sticky' }}
-              top={{ md: '1rem' }}
+              top={{ md: '17.1rem' }}
               ml="1rem"
               bottom={{ base: '5rem', md: undefined }}
               right={{ base: '1rem', md: undefined }}


karrui:
Ah, this seems to not be the intended fix. The figma design was actually aligning it to the latest block instead of just the first block. Hardcoding the height may also not work for smaller viewports. Closing for now, might not be worth the effort imho, but you can try! :P

tohhengxing:
> Ah, this seems to not be the intended fix. The figma design was actually aligning it to the latest block instead of just the first block. Hardcoding the height may also not work for smaller viewports. Closing for now, might not be worth the effort imho, but you can try! :P

Hi @karrui . Just to confirm right, the button is supposed to stick to the latest block and not slide around when the viewport is too small? Cos i also noticed the sliding effect for for the original one when the viewport is too small but I thought that was an intentional feature :p so i didn't do anything about it.

karrui:
i'm checking the design, it's actually sticking to the active logic block. Which is not that easy to implement and may not be worth the logic needed to make it work.

Issue #4573, opened 2022-08-15
Loading screens for builder logic and responses


justynoh:
We should include a loading screen for the logic tab in `CreatePageLogicTab` and responses page in `ResponsesPage`. @pearlyong

pearlyong:
Loading for [Responses](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=11692%3A175667) (can use for both email and storage mode)
Loading for [Empty Logic](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=11692%3A175894)

PR #4579, opened 2022-08-15
feat(v2): add loading screens for logic tab and results page

justynoh:
## Problem
There are loading states in the logic tab and results page files that have not been created. This PR implements the loading states for those files in line with documentation on Figma (see the issue).

Closes [#4573 ]

## Solution
Add loading states and corresponding stories.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/admin-form/create/logic/components/EmptyLogic.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/EmptyLogic.tsx
@@ -1,6 +1,14 @@
 import { useCallback } from 'react'
 import { BiPlus } from 'react-icons/bi'
-import { Divider, Flex, Grid, Icon, Stack, Text } from '@chakra-ui/react'
+import {
+  Divider,
+  Flex,
+  Grid,
+  Icon,
+  Skeleton,
+  Stack,
+  Text,
+} from '@chakra-ui/react'
 
 import { GUIDE_FORM_LOGIC } from '~constants/links'
 import Button from '~components/Button'
@@ -11,7 +19,11 @@ import { ALLOWED_FIELDS_META } from '../constants'
 
 import { LogicSvgr } from './LogicSvgr'
 
-export const EmptyLogic = (): JSX.Element => {
+interface EmptyLogicProps {
+  isLoaded: boolean
+}
+
+export const EmptyLogic = ({ isLoaded }: EmptyLogicProps): JSX.Element => {
   const setToCreating = useAdminLogicStore(
     useCallback((state) => state.setToCreating, []),
   )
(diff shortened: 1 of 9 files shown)


frontend/src/features/admin-form/create/logic/components/EmptyLogic.tsx
```diff
@@ -24,27 +37,31 @@ export const EmptyLogic = (): JSX.Element => {
...
-      <Button
-        my="2.5rem"
-        leftIcon={<BiPlus fontSize="1.5rem" />}
-        onClick={setToCreating}
-      >
-        Add logic
-      </Button>
-      <LogicSvgr maxW="292px" />
-      <Divider my="2.5rem" />
+      <Skeleton isLoaded={isLoaded}>
```
mantariksh: lmao I see this component has both the empty and loading states. I suppose implementing the loading state is a loooot easier if you can just wrap the empty state. what dyou think about renaming the component to `EmptyOrLoadingLogic`? just because we normally think of "empty" and "loading" as separate categories, rather than "loading" being a subset of "empty"

mantariksh (approved):
minor comment

frontend/src/features/admin-form/create/logic/components/EmptyLogic.tsx
```diff
@@ -24,27 +37,31 @@ export const EmptyLogic = (): JSX.Element => {
...
+        <Text textStyle="h2" as="h2">
+          Start creating logic for your form
+        </Text>
+      </Skeleton>
+      <Skeleton isLoaded={isLoaded}>
```
karrui: imo loading state should be modelled after a "populated" state, meaning we might want to create an entirely separate loading state similar to admin form builder and/or public form page

frontend/src/features/admin-form/responses/ResponsesPage/ResponsesPageSkeleton.tsx
```diff
@@ -0,0 +1,15 @@
+import { Box, Container, Skeleton, Stack } from '@chakra-ui/react'
+
+export const ResponsesPageSkeleton = (): JSX.Element => (
```
karrui, line 3: any reason we used a fresh component instead of using `Skeleton` inside `SecretKeyVerification`?

justynoh: This skeleton is needed for when we are still loading the form in the first place. Thus we don't have any information on whether it is an email form or storage form yet. It wouldn't make sense to display the secret key verification input skeleton, so we will display a fully general skeleton instead.

(Actually, there are already `Skeleton`s in `SecretKeyVerification`, which are conditioned on `isLoading` from `StorageResponsesContext`. But they won't end up being used since `isLoading` in `StorageResponsesContext` is directly taken from `useAdminForm`. We can probably get rid of those.)

frontend/src/features/admin-form/create/logic/components/EmptyLogic.tsx
justynoh: The reason for doing it this way is because the design on Figma was modelled after the empty state, so it made sense to simply wrap the empty state. It's also possible to just create a new skeleton component though, depends on how we want the skeleton to be shown! We can discuss this a bit more at sync.

frontend/src/features/admin-form/create/logic/components/EmptyLogic.tsx
justynoh: Same as my reply to anty's comment :P

karrui: if its figma then its figma~ let designer review then

frontend/src/features/admin-form/responses/ResponsesPage/ResponsesPageSkeleton.tsx
karrui: if there already is a skeleton, for `SecretKeyVerification`, why do we need a skeleton for this page?

justynoh: cos `SecretKeyVerification` is rendered when form is loaded and determined to be in storage mode. But until the form is loaded, we don't know whether to show the storage mode or email mode version of the responses page. So as I mentioned, we should show a generic skeleton rather than having the secret key verification skeleton

karrui: gotcha

karrui (approved):
lgtm

frontend/src/features/admin-form/create/logic/components/EmptyLogic.tsx
justynoh: discussed at sync, loading screen has been redesigned!

PR #4582, opened 2022-08-15
feat(whats-new): update whats new API to use version flag instead of date flag

karrui:
## Problem

Date flag seemed a little too brittle and needed additional processing for the What's New feature. As discussed, we can probably add an identifier every time we update the hardcoded what's new object.

This PR updates the user's last seen feature flag to store the version instead of a Date.

## Solution

**Breaking Changes** 

    - Change in User model flag to store last seen state to `flags.lastSeenFeatureUpdateVersion`

**Features**:

- feat: update whats new API to use version flag instead of date flag


justynoh (approved):
lgtm

PR #4590, opened 2022-08-16
fix(v2): re-enable backend tests

### LINKED ISSUE #4581: Reenable (and fix) backend tests for `form-v2/*` branches
might want to reenable backend tests for our form-v2/* branches, we actually have a number of server-side changes... haha. i see a bunch of failing tests locally already oops

wanlingt:
## Problem

Failing backend tests are escaping our notice

Closes #4581 

## Solution

- Reinstate backend tests in `ci-react`
- Fix failing tests - `helmet.spec.ts`, `public-form.controller.spec.ts` and `submission.utils.spec.ts`

**Breaking Changes** 

**Bug Fixes**:
- `submission.util.ts`: remove redundant false check for `encryptFormFieldModeFilter`

### DIFF
--- a/src/app/modules/submission/submission.utils.ts
+++ b/src/app/modules/submission/submission.utils.ts
@@ -51,18 +51,9 @@ const encryptFormFieldModeFilter = <T extends FormField>(
   responses: T[] = [],
 ) => {
   // To filter for autoreply-able fields.
-  return responses.filter((response) => {
-    if ([BasicField.Mobile, BasicField.Email].includes(response.fieldType))
-      return false
-    switch (response.fieldType) {
-      case BasicField.Mobile:
-        return response.isVerifiable
-      case BasicField.Email:
-        return response.autoReplyOptions.hasAutoReply || response.isVerifiable
-      default:
-        return false
-    }
-  })
+  return responses.filter((response) =>
+    [BasicField.Mobile, BasicField.Email].includes(response.fieldType),
+  )
 }
 
 // Exported for testing.
--- a/src/app/modules/submission/submission.utils.ts
+++ b/src/app/modules/submission/submission.utils.ts
@@ -51,9 +51,16 @@ const encryptFormFieldModeFilter = <T extends FormField>(
   responses: T[] = [],
 ) => {
   // To filter for autoreply-able fields.
-  return responses.filter((response) =>
-    [BasicField.Mobile, BasicField.Email].includes(response.fieldType),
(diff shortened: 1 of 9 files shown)


src/app/modules/submission/submission.utils.ts
```diff
@@ -51,18 +51,9 @@ const encryptFormFieldModeFilter = <T extends FormField>(
...
-      return false
-    switch (response.fieldType) {
-      case BasicField.Mobile:
-        return response.isVerifiable
-      case BasicField.Email:
-        return response.autoReplyOptions.hasAutoReply || response.isVerifiable
-      default:
-        return false
-    }
-  })
```
karrui: this is an intended change in our app, we should update tests for this behaviour instead

karrui (changes requested):

src/app/modules/submission/submission.utils.ts
wanlingt: ok got it! should we remove this part though?

>` if ([BasicField.Mobile, BasicField.Email].includes(response.fieldType))
       return false
`

since including it means we won't get to the switch case statements for `BasicField.Mobile` and `BasicField.Email`, i.e.this function always returns an empty array

karrui: Wait a second was that a bug!!! 😭

wanlingt: yaaa 🐛 🐞 😢

karrui: can double check if behaviour changes from expected listed in https://github.com/opengovsg/FormSG/pull/4001 if we remove the false check?

wanlingt: commented out the false check, checked if I was able to

- view an individual storage mode form response
- navigate to the next/previous response
- download individual attachment

all of the above work!

~I'm not able to receive an email auto-reply though (on `form-v2/develop` as well), investigating further~ oops it just wasn't working on `form-v2/develop`, the email auto-reply works now after modifying `encryptFormFieldModeFilter`

karrui: try submitting verified and unverified mobile/email fields just to be safe

wanlingt: btw is autoreply for mobile implemented? i can't seem to find it heh

karrui (approved):
lgtm, thanks!!!

Issue #4593, opened 2022-08-16
Make response ID more discoverable on end page


mantariksh:
Some form admins tell their respondent to take note of the response ID for record-keeping. The response ID used to be visible by default on the end page. However, in React it is hidden under an accordion which is not very obvious.

The rationale for hiding under an accordion is to make the end page less busy. We can also achieve this by removing the form title and timestamp, and just including "Response ID: abcdef" (not hidden under accordion).

pearlyong:
updated design [here](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=8129%3A115758)

PR #4596, opened 2022-08-17
fix(v2): remove accordion from end page to make response ID more discoverable

justynoh:
## Problem
The response id is currently hidden under an accordion in the form end page. We want to make it more discoverable. 

Closes [#4593 ]

## Solution
Remove accordion from the end page as well as the information about form title and submission date and time. Also removed corresponding props from the end page type definitions.

**Breaking Changes** 

## Screenshots

image

image


wanlingt (approved):
lgtm!

PR #4600, opened 2022-08-18
fix: return correct statuses on submission routes when sgid errors occur

karrui:
## Problem

We are erroneously returning status `500`s due to default fallthrough due to unhandled SGID error mapping. This PR fixes that.

## Solution

**Breaking Changes** 

    - Details ...

**Bug Fixes**:

- fix: return correct statuses on submission routes when sgid errors occur

## Tests

Add route-level tests to test for the returning of correct error codes


kwajiehao (approved):
lgtm

Issue #4144, opened 2022-07-12
[1] prompt using usePrompt hook to prevent navigation does not work when cancelling


karrui:
users literally cannot cancel their navigation. cancelling just reopens the window alert. 
May need to recheck how usePrompt works and fix or discard usage of usePrompt.

~~Adding a further wrench in the works, the next version of `react-router` seems to remove usePrompt entirely without a workaround. See https://github.com/remix-run/react-router/issues/8139#issuecomment-1176523524~~

Will just stick with current version of router for now.

karrui:
This is already fixed by #4601, but going one step further to make it look like the design

PR #4601, opened 2022-08-18
feat(v2): add and use NavigationTabs template components for admin form tab navigation

karrui:
## Problem

This PR updates the components used for navigation from ChakraUI tabs to a faux tab component meant to be used for links.
This component fixes a bug where the blocking of navigation would fail (infinite loop) until user accepts the navigation block. That was due to ChakraUI's `Tab` component retriggering its `onChange` callback continuously (which retriggers navigation, which retriggers the block).

By using real links instead of the Tab component, no `react-router-dom#navigate` call is called and users can safely accept/reject the prompt.

Related to #4144

## Solution

**Breaking Changes** 

**Features**:

-  add and use NavigationTabs template components for admin form tab navigation

## Before & After Screenshots
Should have absolutely no visual changes


frontend/src/features/admin-form/common/components/AdminFormNavbar/AdminFormNavbar.tsx
```diff
@@ -60,6 +67,21 @@ export const AdminFormNavbar = ({
...
+  const tabResponsiveVariant = useBreakpointValue({
+    base: 'line-dark',
+    xs: 'line-dark',
+    lg: 'line-light',
+  })
+
+  const checkTabActive = useCallback(
+    (to: string) => {
+      const match = pathname.match(ACTIVE_ADMINFORM_BUILDER_ROUTE_REGEX)
+      return (match?.[2] ?? '/') === `/${to}`
```
justynoh, line 81: Clearly this should never happen, but maybe we should throw an error instead if we fail to match, to rule out unintended behavior?

```suggestion
      if (!match) throw new Error('Unexpected pathname')
      return (match[2] ?? '/') === `/${to}`
```

justynoh (approved):
lgtm, just a minor comment

frontend/src/features/admin-form/common/components/AdminFormNavbar/AdminFormNavbar.tsx
karrui: this might happen on a form link without a slash at the end, that is why a default is needed

PR #4567, opened 2022-08-13
Update to CRA 5

darrelhong:
Updates the form-v2 frontend to use react-scripts 5. Local dev, build, unit tests and storybook validated.

Pros:
All the benefits of CRA 5 and webpack 5
Cached dev start timings ~10s vs ~20s before

Cons:
Hot reload timings pretty much the same
Uses craco alpha
Storybook config must be updated separately until storybook-preset-craco is updated

PR #4603, opened 2022-08-19
fix: local mongo replicaset host mismatch

darrelhong:
this simple change should fix issue brought up in #3783. There should be no need to manually config mongo hostname everytime the container IP changes

Deleting the mongodb_data volume should and re-running docker should update the mongo config, else updating the hostname to `database:27017`, using the steps in TROUBLESHOOTING.md, should also work

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -124,6 +124,7 @@ services:
     container_name: 'formsg-db'
     environment:
       - MONGODB_DATABASE=formsg
+      - MONGODB_ADVERTISED_HOSTNAME=database
       - MONGODB_REPLICA_SET_MODE=primary
       - ALLOW_EMPTY_PASSWORD=yes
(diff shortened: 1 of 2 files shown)


karrui:
Thanks for this contribution! Can you also update `TROUBLESHOOTING.md` to note that the issue can be fixed by updating local hostname?

darrelhong:
doneded. as a sidenote, not sure if you've also seen #4567 #4611? would appreciate your comments on those as well

karrui:
> doneded. as a sidenote, not sure if you've also seen #4567 #4611? would appreciate your comments on those as well

yup, but since those PRs affect tooling/build/etc, will be deferred until we are done with our react migration to evaluate. Thanks for contributing!

karrui (approved):
lgtm

PR #4606, opened 2022-08-19
feat(v2): load SwitchEnvModal conditionally

karrui:
this prevents react-query from needlessly querying the switch env form until it is needed


wanlingt (approved):
lgtm thanks!

PR #4610, opened 2022-08-19
feat(v2): add env var banners to admin and public form pages

### LINKED ISSUE #4550: [1] Add banners to app when env vars are set
similar to angularjs app, we should show banners on login, admin, public, sp/cp forms accordingly when the env var are set.

justynoh:
## Problem
Currently we do not show banners for env vars according to our [banner specification](https://github.com/opengovsg/FormSG/blob/develop/docs/DEPLOYMENT_SETUP.md#banners) in the react app.

Closes #4550 

## Solution
Add env var banners using the `Banner` component. Replicates behavior of angular app.

**Breaking Changes** 

## Screenshots

image
image
image

### DIFF
--- a/frontend/src/features/admin-form/common/AdminFormLayout.tsx
+++ b/frontend/src/features/admin-form/common/AdminFormLayout.tsx
@@ -27,6 +27,7 @@ export const AdminFormLayout = (): JSX.Element => {
   const { data: { siteBannerContent, adminBannerContent } = {} } = useEnv()
 
   const bannerContent = useMemo(
+    // Use || instead of ?? so that we fall through even if previous banners are empty string.
     () => siteBannerContent || adminBannerContent,
     [adminBannerContent, siteBannerContent],
   )
--- a/frontend/src/features/admin-form/common/AdminFormLayout.tsx
+++ b/frontend/src/features/admin-form/common/AdminFormLayout.tsx
@@ -1,9 +1,14 @@
+import { useMemo } from 'react'
 import { Outlet, useParams } from 'react-router-dom'
 import { Flex } from '@chakra-ui/react'
 import { get } from 'lodash'
 
+import { getBannerProps } from '~utils/getBannerProps'
+import { Banner } from '~components/Banner'
+
 import AdminForbiddenErrorPage from '~pages/AdminForbiddenError'
 import NotFoundErrorPage from '~pages/NotFoundError'
+import { useEnv } from '~features/env/queries'
 // TODO #4279: Remove after React rollout is complete
 import { SwitchEnvIcon } from '~features/env/SwitchEnvIcon'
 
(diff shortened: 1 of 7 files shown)


justynoh:
ignore angular tests pls hehe

frontend/src/features/admin-form/common/AdminFormLayout.tsx
```diff
@@ -19,6 +24,18 @@ export const AdminFormLayout = (): JSX.Element => {
   const { formId } = useParams()
   if (!formId) throw new Error('No formId provided')
 
+  const { data: { siteBannerContent, adminBannerContent } = {} } = useEnv()
+
+  const bannerContent = useMemo(
+    () => siteBannerContent || adminBannerContent,
```
karrui, line 31: nit: should we use `??` instead of `||`?  even though there is no real difference in our use cases, but nullish coalesce is probably slightly more correct

frontend/src/utils/getBannerProps.ts
```diff
@@ -0,0 +1,32 @@
...
+  variant: BannerVariant
+  msg: string
+}
+
+export const getBannerProps = (bannerContent?: string): BannerProps | null => {
+  if (!bannerContent) return null
+  if (bannerContent.startsWith('info:')) {
+    return {
+      variant: 'info',
+      msg: bannerContent.slice(5),
```
karrui: should we trim the content after slicing, so the banner will not have leading/ending spaces

karrui (approved):
lgtm

frontend/src/features/admin-form/common/AdminFormLayout.tsx
justynoh: Can I clarify why is `??` slightly more correct? I think `||` is better because it will fall through to the next case if the first one is `undefined` or `''`, which i think is correct in both cases. (If we defined the first env var but left it blank then `??` would return `''` which I would think is is unintended.)

karrui: If the banner is a number (0), which is impossible in our case haha, then the wrong output will arise due to using a Boolean OR operator. Boolean OR works on all falsey values, whilst nullish coalescence only works on undefined or null, which is slightly more correct. Either way it's not wrong in this case, but something to keep in mind.

karrui: Ah, I see what you mean. There is a possibility of empty string too. Yeah, let's keep it and add a comment so it won't be "refactored" by others haha

justynoh: added comment in [369db25](https://github.com/opengovsg/FormSG/pull/4610/commits/369db25ac2586413b5d778e73e185a08acced32a)

frontend/src/utils/getBannerProps.ts
justynoh: yes! thanks for catching it. added in [369db25](https://github.com/opengovsg/FormSG/pull/4610/commits/369db25ac2586413b5d778e73e185a08acced32a)

Issue #3934, opened 2022-06-01
Improve dev environment


mantariksh:
Current dev environment has quite a few problems for developers, e.g. every time a new package is installed, the docker image needs to be deleted and rebuilt. Also it takes about 10-15min to build.

Once AngularJS frontend has been deprecated, investigate how we can improve this for developers.

LinHuiqing:
Migrated to Linear - [FRM-1403 : Improve dev environment](https://linear.app/ogp/issue/FRM-1403/improve-dev-environment)

LinHuiqing:
Migrated to Linear - [FRM-1403 : Improve dev environment](https://linear.app/ogp/issue/FRM-1403/improve-dev-environment)

PR #4611, opened 2022-08-20
chore: update docker tooling for local dev

darrelhong:
## Changes

1. npm run dev no longer rebuilds images
Images shouldn't need to be rebuilt for every time for development. For first time run, docker compose will build the necessary images if they do not exist

2. Update development Dockerfile to copy only package.json files before npm run install.
Take advantage of docker cached layers as explained [here](http://bitjudo.com/blog/2014/03/13/building-efficient-dockerfiles-node-dot-js/). If there are no dependency changes, docker will just used the cached layer from previous runs. Note that if you don't run docker compose with the --build flag, this change doesn't do much as npm install will not even be run.

3. Update documentation with info on how to add dependencies
docker compose up will not take into account dependency changes so, the node_modules volume needs to be updated.
--build will rebuild the backend image and run npm install, --renew-anon-volumes is also needed as docker will reuse the previous node_modules volume and dependencies will not get updated. This update should help with #3934

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -44,8 +44,10 @@ ENV PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium-browser
 # Chinese fonts
 RUN echo @edge http://nl.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories && apk add wqy-zenhei@edge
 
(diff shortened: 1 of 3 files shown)


karrui (approved):
lgtm, cheers.

PR #4615, opened 2022-08-22
feat(v2): add better error messages for submission and network errors 

### LINKED ISSUE #4316: Improve error message when user is offline
When user is offline, error message shown is always "Network Error". To replicate:
1. Go to Network tab and switch "No throttling" to "Offline"
2. Attempt to make any network call e.g. edit form field or request OTP
3. Error message is "Network Error"

We should change this to something more informative: "You may have a problem with your network. Please refresh and try again."

### LINKED ISSUE #4389: [1] Show informative error messages for submission errors
Currently we only show a single generic error message when submission errors happen. The server often returns a more helpful error message and we should render it.

To replicate:
1. Go to a public form
2. Use developer tools to turn off network
3. Attempt to submit

justynoh:
## Problem
1. Currently, for submission errors we do not display the error message returned by the server, but a generic error message. Closes #4389 
2. Also, our offline error message just says "Network Error" which is not very informative. Closes #4316 

## Solution
1. Modify the submission handler to use the network error message if it exists, and fallback to the generic error message otherwise.
2. Add a network error check a la https://stackabuse.com/handling-errors-with-axios/#:~:text=in%20Node.js.-,Other,-Errors

**Breaking Changes** 

## Screenshots

image

### DIFF
--- a/frontend/src/services/ApiService.ts
+++ b/frontend/src/services/ApiService.ts
@@ -39,7 +39,7 @@ export const transformAxiosError = (e: Error): ApiError => {
       return new Error(
         'There was a problem with your internet connection. Please check your network and try again.',
       )
-    }
+    } else return e
   }
   return e
 }
--- a/frontend/src/services/ApiService.ts
+++ b/frontend/src/services/ApiService.ts
@@ -19,22 +19,27 @@ export class HttpError extends Error {
  * @returns HttpError if AxiosError, else original error
  */
 export const transformAxiosError = (e: Error): ApiError => {
-  if (axios.isAxiosError(e) && e.response) {
-    const statusCode = e.response.status
-    if (statusCode === 429) {
-      return new HttpError('Please try again later.', statusCode)
+  if (axios.isAxiosError(e)) {
+    if (e.response) {
+      const statusCode = e.response.status
+      if (statusCode === 429) {
+        return new HttpError('Please try again later.', statusCode)
+      }
+      if (typeof e.response.data === 'string') {
+        return new HttpError(e.response.data, statusCode)
+      }
+      if (e.response.data?.message) {
+        return new HttpError(e.response.data.message, statusCode)
+      }
+      if (e.response.statusText) {
+        return new HttpError(e.response.statusText, statusCode)
+      }
+      return new HttpError(`Http ${statusCode} error`, statusCode)
+    } else if (e.request) {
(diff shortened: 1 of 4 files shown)


frontend/src/services/ApiService.ts
```diff
@@ -19,22 +19,27 @@ export class HttpError extends Error {
...
+      }
+      if (e.response.statusText) {
+        return new HttpError(e.response.statusText, statusCode)
+      }
+      return new HttpError(`Http ${statusCode} error`, statusCode)
+    } else if (e.request) {
+      return new Error(
+        'There was a problem with your internet connection. Please check your network and try again.',
+      )
+    } else return e
```
timotheeg: This `else` is not needed since line 44 does it already?

frontend/src/services/ApiService.ts
```diff
@@ -19,22 +19,27 @@ export class HttpError extends Error {
...
-    }
-    if (e.response.statusText) {
-      return new HttpError(e.response.statusText, statusCode)
-    }
-
-    return new HttpError(`Http ${statusCode} error`, statusCode)
+  if (axios.isAxiosError(e)) {
+    if (e.response) {
+      const statusCode = e.response.status
+      if (statusCode === 429) {
```
tshuli: minor nit - should we use `http-status-codes` instead of hardcoding 429?

tshuli (approved):

frontend/src/services/ApiService.ts
justynoh: its actually needed, since the `ifs` are now nested. without the `else`, if it's an axios error that has no request or response, we don't return anything in that case

frontend/src/services/ApiService.ts
justynoh: actually we could be doing this everywhere in the react app right haha i did a cursory search and we never do this anywhere in `frontend/` :P guess i'll be the first one!

justynoh: updated in [6f93a1b](https://github.com/opengovsg/FormSG/pull/4615/commits/6f93a1bb03d02a8452639b65ab31bd1735ec629e)

frontend/src/services/ApiService.ts
justynoh: SORRY LOL i brainfarted, will make this change.

PR #4665, opened 2022-08-23
feat(v2): migrate remaining Google Analytics 4 events

### LINKED ISSUE #4780: [5] Migrate from Universal Analytics to Google Analytics 4

wanlingt:
## Problem

Our implementation of GA is incomplete.

Closes #4780 

## Solution

Port over the remaining GA events from AngularJS to React. Also, migrate from Universal Analytics to **Google Analytics 4**

Note that the current events have not yet been fully optimised for GA4. The events can be extended further in later PRs once we are clear about how we want to use GA, and what information we wish to get out of it.

TODO #4786 after React Rollout: Remove occurrences of `GA_TRACKING_ID` (used in Angular)

**Breaking Changes** 

**Features**:

- [Map parameters](https://support.google.com/analytics/answer/11091025?hl=en&ref_topic=11091421#zippy=%2Cin-this-article) from Universal Analytics to GA4. 
**UA:** Event parameters are classified under `event_category`, `event_label` and `action`
**GA4:** Event parameters can take on any name. There is no limit on the number of parameters.
- Move `showErrorToast` out of `useCommonFormProvider` into `PublicFormProvider` and `PreviewFormProvider`. `showErrorToast` in `PublicFormProvider` logs the `trackSubmitFormFailure` GA4 event (so it should only apply to public forms)
- Update CSP with [GA4 directives](https://developers.google.com/tag-platform/tag-manager/web/csp)

**Other details**:
- To get around the CSP error (which disallowed the inline script from executing), move the GA4 inline script ([dataLayer](https://developers.google.com/tag-platform/devguides/datalayer) code) to `frontend/src/index`. As a result, `partytown` is no longer used and `REACT_APP_GA_TRACKING_ID` is now injected at build time.
  - for knowledge: using the [hash](https://developers.google.com/tag-platform/tag-manager/web/csp) of the inline script worked, but this changes depending on the `GA_TRACKING_ID`, so it was not ideal.  
- Build time env vars have been moved to `buildtime-env` for local dev

- `REACT_APP_GA_TRACKING_ID` : GA4 tracking ID, used in `frontend/src/index`

## Tests
- [x] Check that GA4 events are being captured in [Google Analytics](https://analytics.google.com/analytics/web/) reports / debug view


frontend/src/index.tsx
```diff
@@ -15,6 +15,21 @@ if (process.env.NODE_ENV === 'test') {
...
+  var dataLayer: unknown[]
+}
+
+window.dataLayer = window.dataLayer || []
+function gtag(...args: unknown[]) {
+  // eslint-disable-next-line prefer-rest-params
+  dataLayer.push(arguments)
+}
+gtag('js', new Date())
+gtag('config', process.env.REACT_APP_GA_TRACKING_ID || '', { debug_mode: true })
```
timotheeg: Should probably remove the debug flag?

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -2,6 +2,13 @@ import { useCallback, useEffect, useRef, useState } from 'react'
 import { useMutation, UseMutationOptions } from 'react-query'
 
 import { useAdminForm } from '~features/admin-form/common/queries'
+import {
+  trackDownloadNetworkFailure,
+  trackDownloadResponseFailure,
+  trackDownloadResponseStart,
+  trackDownloadResponseSuccess,
+  trackPartialDecryptionFailure,
+} from '~features/analytics/AnalyticsService'
```
timotheeg, line 11: Just checking: have you verified these events are being reported from the worker? I'm asking because the GA base setup is now in the main app, and some events might be black-holed with the noop function 🤔

frontend/src/index.tsx
wanlingt: oops yes, thanks for spotting

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
wanlingt: Yup, I verified that both `trackDownloadResponseStart` and `trackDownloadResponseSuccess` are logged in GA. Wasn't sure how to test the other events though, do you have any ideas? Or can we assume that the rest will be reported too, since `trackDownloadResponseStart` and `trackDownloadResponseSuccess` are working?

timotheeg: Yup, I think that's fine 👍

timotheeg (approved):
LGTM. Thanks @wanlingt !

As 2 extra notes of description here:

1) we are injecting the GA Tracking ID at build time (just like DD RUM ID), before building the deployment image. This means we cannot share images across environments (staging/production), and we must build for every release for every environment. That is not ideal, but that is what we've been doing now anyway, even without the GA and DD vars. We should revisit this when we look at optimizing our deployment pipelines.

2) One reason the GA and DD variables need to exist at serve-time, rather than being loaded from a separate HTTP call is that a) they track page load time and should thus be initialised as early as possible; and b) they can capture and report run-time errors, so again, they should be ready to go before the app proper starts.

karrui:
should we remove the partytown package entirely then? is there also no way to continue using it? Will speed up load times

timotheeg:
If it's no longer used anywhere, yes please!

wanlingt:
> should we remove the partytown package entirely then? is there also no way to continue using it? Will speed up load times

good catch, will remove! I couldn't find a way to use it while still injecting GA_TRACKING_ID at build time, but we can investigate again when we optimize our deployment pipelines..

PR #4670, opened 2022-08-23
feat: add navigation confirmation when builder fields are dirty

### LINKED ISSUE #4304: [1] Confirmation modal when navigating out of form field editing before saving
[design](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=7298%3A106394)

karrui:
## Problem

This PR adds a navigation confirmation on navigation and change of field when the field being currently edited is dirty. The possible navigation cases are covered below:

- start page
- end page
- field
- builder tabs
- drawer close button
- drawer back button

Any permutation (start page -> field, field -> field, end page -> start page) will trigger a render of a new `DirtyModal` component if the fields are dirty. This was achieved by storing pending states in the respective stores and context, subscribing to a newly exposed `isDirty` selector in `fieldBuilderStore`. 

Any route navigation change will also trigger a similar `NavigationPrompt` modal.

Closes #4304

## Solution

**Breaking Changes** 

**Features**:

- feat: add `NavigationPrompt` modal component to show navigation confirmation modal when condition is true
  - this is used in this PR when the `isDirty` constant is true.
- feat: update `useDesignStore`, `useFieldBuilderStore`, `CreatePageSidebarContext` to be responsive to dirty field state. 
- feat: show confirmation modal when navigating or changing edit fields when current field is dirty.

## Screenshots
CreatePage stories should now have the behaviour, but here is a gif in case that is hard to navigate.
Recording 2022-08-25 at 12 06 00


justynoh:
Regarding the UX itself, is there any particular reason to have different modals in each case? Since there are two sets of modals (one for navigation and one for changing the active state within the same route), I wonder if users would have the (incorrect) impression that "discarding changes" means different things for each modal. 

For instance (and this is just my intuition - feel free to comment if you don't think so), if i click out of a dirty field, I might think that the "You have unsaved changes" modal will discard my changes to the current active field. But if let's say i've been working on my form and I try to navigate away, seeing a different modal ("Discard changes?") I might be worried that there are other changes to the form that aren't yet saved.

karrui:
> Regarding the UX itself, is there any particular reason to have different modals in each case? Since there are two sets of modals (one for navigation and one for changing the active state within the same route), I wonder if users would have the (incorrect) impression that "discarding changes" means different things for each modal.
> 
> For instance (and this is just my intuition - feel free to comment if you don't think so), if i click out of a dirty field, I might think that the "You have unsaved changes" modal will discard my changes to the current active field. But if let's say i've been working on my form and I try to navigate away, seeing a different modal ("Discard changes?") I might be worried that there are other changes to the form that aren't yet saved.

@pearlyong?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderAndDesignContent.tsx
```diff
@@ -36,7 +36,9 @@ export const BuilderAndDesignContent = ({
       ),
     )
 
-  useEffect(() => setFieldsToInactive, [setFieldsToInactive])
+  useEffect(() => {
+    setFieldsToInactive()
+  }, [setFieldsToInactive])
```
justynoh: Should we also `setFieldsToInactive` on unmount like we did previously so that we keep the state clean when the admin exits the builder?

frontend/src/features/admin-form/create/builder-and-design/useFieldBuilderStore.tsx
```diff
@@ -11,64 +11,116 @@ export enum FieldBuilderState {
...
-        field: FormFieldDto
-      }
-    | { state: FieldBuilderState.EditingEndPage }
+    | FieldBuilderCreateEditStateData
     | { state: FieldBuilderState.Inactive }
+  // Used when there is a dirty state and we want to hold the next state to be set.
+  // Will be used to set stateData if user confirms discarding changes.
+  holdingStateData:
+    | FieldBuilderCreateEditStateData
+    | { state: FieldBuilderState.Inactive }
```
justynoh, line 42: what's the reason for keeping the inactive state separate from `FieldBuilderCreateEditStateData`? didn't seem like there's anywhere that requires them to be separate

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/useEditFieldForm.ts
```diff
@@ -65,18 +67,25 @@ export const useEditFieldForm = <FormShape, FieldShape extends FormField>({
...
+        stateData: stateDataSelector(state),
+        setToInactive: setToInactiveSelector(state),
+        updateEditState: updateEditStateSelector(state),
+        updateCreateState: updateCreateStateSelector(state),
+        isDirty: isDirtySelector(state),
+        setIsDirty: setIsDirtySelector(state),
+      }),
+      [],
+    ),
+  )
```
justynoh: seems like `isDirty` isn't included in the destructuring here, can we take that line out?

frontend/src/templates/NavigationPrompt/useNavigationPrompt.ts
```diff
@@ -0,0 +1,67 @@
...
+      navigator: Navigator
+    }
+
+export const useNavigationPrompt = (when?: boolean) => {
+  const navigationContext = useContext(
+    NavigationContext,
+  ) as NavigationContextWithBlock
+
+  const [isPromptShown, setIsPromptShown] = useState(false)
+  const [currentPath, setCurrentPath] = useState('')
```
justynoh: is `targetPath` a better name for this state?

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
```diff
@@ -9,7 +9,7 @@ import { UploadedImage } from './BuilderAndDesignDrawer/EditFieldDrawer/edit-fie
 export type CustomLogoMeta = Omit<CustomFormLogo, keyof FormLogoBase>
 
 export enum DesignState {
-  EditingHeader,
+  EditingHeader = 1,
```
justynoh, line 12: why is this `=1` needed?

frontend/src/features/admin-form/create/builder-and-design/useFieldBuilderStore.tsx
```diff
@@ -11,64 +11,116 @@ export enum FieldBuilderState {
...
+  updateCreateState: (
+    field: FieldCreateDto,
+    insertionIndex: number,
+    holding?: boolean,
+  ) => void
+  updateEditState: (field: FormFieldDto, holding?: boolean) => void
+  setEditEndPage: (holding?: boolean) => void
+  setToInactive: (holding?: boolean) => void
+  isDirty: boolean
+  setIsDirty: (isDirty: boolean) => void
```
justynoh: since these are used throughout design and endpage too, would it be better to put them somewhere above, perhaps in the Create page context or Admin builder context? Because the original intention for separating the design, builder and endpage stores was to separate the concerns, thus putting `isDirty` here feels a bit counter to that as its also used in design and endpage

frontend/src/features/admin-form/create/builder-and-design/useFieldBuilderStore.tsx
karrui: reusing `FieldBuilderCreateEditStateData` type elsewhere where we want to ensure inactive state cannot be set.

frontend/src/features/admin-form/create/builder-and-design/useDesignStore.tsx
karrui: We want the enum to start from `1` (so falsey footguns, like when value is `0` and you do a negate check is less likely to happen)

frontend/src/features/admin-form/create/builder-and-design/useFieldBuilderStore.tsx
karrui: didn't want to put something that changes so often into context, since that will trigger rerenders. Opted to put in this store first, we can always move it in the future. Seemed overkill to add yet another store just to hold one dirty state. do you think we should create a new store? actually ok let me do it

karrui: done in 4ecda0b

pearlyong:
agree it might be confusing, let's stick with the 'You have unsaved changes' modal, which is easier English to understand

justynoh (approved):
lgtm

PR #4672, opened 2022-08-23
feat(v2): download attachments as ZIP from individual response page

### LINKED ISSUE #4645: [1] Download attachments as ZIP in storage mode individual responses
Angular vs React parity in storage mode 
Download attachments as ZIP in the header box
DEFINITELY fix for rollout

image

justynoh:
## Problem
Download all attachments as ZIP was a feature in Angular which we missed out in React.

Closes #4645 

## Solution
Add button in header to download the attachments as ZIP

**Breaking Changes** 

## Screenshots

image

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -1,5 +1,11 @@
 import { memo, useCallback, useMemo } from 'react'
-import { BiChevronLeft, BiChevronRight, BiLeftArrowAlt } from 'react-icons/bi'
+import {
+  BiChevronLeft,
+  BiChevronRight,
+  BiDownload,
+  BiLeftArrowAlt,
+} from 'react-icons/bi'
+import { useMutation } from 'react-query'
 import {
   Link as ReactLink,
   useLocation,
@@ -17,14 +23,19 @@ import {
   StackDivider,
   Text,
 } from '@chakra-ui/react'
+import FileSaver from 'file-saver'
+import simplur from 'simplur'
 
+import Button from '~components/Button'
 import IconButton from '~components/IconButton'
 
 import {
   SecretKeyVerification,
   useStorageResponsesContext,
 } from '../ResponsesPage/storage'
+import { AttachmentsDownloadMap } from '../ResponsesPage/storage/types'
 import { useUnlockedResponses } from '../ResponsesPage/storage/UnlockedResponses/UnlockedResponsesProvider'
+import { downloadAndDecryptAttachmentsAsZip } from '../ResponsesPage/storage/utils/downloadAndDecryptAttachment'
 
 import { DecryptedRow } from './DecryptedRow'
 import { useIndividualSubmission } from './queries'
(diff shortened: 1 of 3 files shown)


frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -132,6 +143,43 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+        attachmentDownloadUrls,
+        secretKey,
+      )
+      if (!byteArray) throw new Error('Invalid file')
+      return FileSaver.saveAs(
+        new Blob([byteArray]),
+        `RefNo ${submissionId}.zip`,
+      )
+    },
+  )
```
karrui: should we move this out into its own mutation.ts file?

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -132,6 +143,43 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+
+  const handleDownload = useCallback(() => {
+    if (!secretKey) return
```
karrui: we can probably return if there is also no `data` yet

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -132,6 +143,43 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+    return handleDownloadMutation.mutate({ attachmentDownloadUrls, secretKey })
```
karrui: should we also add error and success toasts to be more consistent?

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -189,10 +237,31 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+              Timestamp:
+            </Text>
+            <Skeleton isLoaded={!isLoading && !isError}>
+              &nbsp;{data?.submissionTime ?? 'Loading...'}
+            </Skeleton>
+          </Box>
+          <Box
+            alignItems="center"
+            display="inline-flex"
+            hidden={attachmentDownloadUrls.size === 0}
```
karrui: i think we should not even render if there are no attachments instead of just hiding this section

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -189,10 +237,31 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+            <Text as="span" textStyle="subhead-1">
+              Attachments:
             </Text>
             <Skeleton isLoaded={!isLoading && !isError}>
-              {data?.submissionTime ?? 'Loading...'}
+              &nbsp;
+              <Button
+                variant="link"
+                aria-label="Download attachments as ZIP"
+                isDisabled={handleDownloadMutation.isLoading}
```
karrui: nit: should we use the `isLoading` key instead?

```tsx
isLoading={handleDownloadMutation.isLoading}
```

not sure how the loading spinner will look like though, lol (maybe we can change the right icon to the spinner on loading state if it looks weird)

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -189,10 +237,31 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+          >
+            <Text as="span" textStyle="subhead-1">
+              Attachments:
             </Text>
             <Skeleton isLoaded={!isLoading && !isError}>
-              {data?.submissionTime ?? 'Loading...'}
+              &nbsp;
+              <Button
+                variant="link"
+                aria-label="Download attachments as ZIP"
```
karrui: is there a need for an aria-label seeing how there is already button text inside the button?

karrui (approved):
lgtm with some suggestions

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
karrui: we should also add an onError state to handle errors

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
justynoh: I didn't add them because we don't have them for individual attachment downloads either. Can add if we would like to, should we add for the individual attachment download too?

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
justynoh: Used for consistency with individual attachment download buttons as well, because the individual attachment download buttons use isDisabled. This is how it looks like for isLoading, I feel like it looks a bit weird? thoughts? haha

image

justynoh: tagging @karrui for thoughts, haha

karrui: Can we change the right icon to spinner when loading then? Can continue using disabled prop over loading.

justynoh: okie, done in [4dfeeab](https://github.com/opengovsg/FormSG/pull/4672/commits/4dfeeab65852da07ee85a782b7495ae883a6d95d)

PR #4674, opened 2022-08-23
fix(v2): fix if/then show fields logic for number/decimal fields

### LINKED ISSUE #4617: [1] Form logic behaves wrongly for number/decimal
For form logic, fields shown by default are wrong if the IF condition is for number or decimal fields is “number/decimal field is less than or equals to”. E.g. if I add a condition saying “If Number is less than or equals to 2, then show Image”, then Image is visible by default before I fill the Number field. This shouldn’t be the case - the Image field should be hidden by default.

wanlingt:
## Problem

Hidden fields were showing even though the logic conditions were not being fulfilled

Closes #4617 

## Solution

`Number(currentValue)` returns 0 when `currentValue` is empty. Hence, the less than equals to logic condition was always returning `true`. To fix this, we do a check for `currentValue` first and return `false` if it's empty, before checking the logic condition.

**Breaking Changes** 

**AFTER**:

Recording 2022-08-23 at 17 01 35

### DIFF
--- a/frontend/src/features/logic/utils/isConditionFulfilled.ts
+++ b/frontend/src/features/logic/utils/isConditionFulfilled.ts
@@ -40,7 +40,8 @@ export const isConditionFulfilled = (
   if (
     currentValue === null ||
     currentValue === undefined ||
-    currentValue.toString().length === 0
+    // check if currentValue is empty for Decimal and Number fields
+    (typeof currentValue === 'string' && currentValue.length === 0)
   ) {
     return false
   }
--- a/frontend/src/features/logic/utils/isConditionFulfilled.ts
+++ b/frontend/src/features/logic/utils/isConditionFulfilled.ts
@@ -37,23 +37,19 @@ export const isConditionFulfilled = (
   if (!isLogicableField(args)) return false
 
   const currentValue = getCurrentFieldValue(args.input, args.fieldType)
-  if (currentValue === undefined) return false
+  if (
+    currentValue === null ||
+    currentValue === undefined ||
+    currentValue.toString().length === 0
+  ) {
+    return false
+  }
 
   switch (condition.state) {
     case LogicConditionState.Lte:
-      if (!currentValue)
-        // currentValue must be populated for the condition to be checked
-        return false
-      else {
-        return Number(currentValue) <= Number(condition.value)
-      }
+      return Number(currentValue) <= Number(condition.value)
     case LogicConditionState.Gte:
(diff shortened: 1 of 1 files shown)


frontend/src/features/logic/utils/isConditionFulfilled.ts
```diff
@@ -41,9 +41,19 @@ export const isConditionFulfilled = (
 
   switch (condition.state) {
     case LogicConditionState.Lte:
-      return Number(currentValue) <= Number(condition.value)
+      if (!currentValue)
```
mantariksh: what if `currentValue` is 0? then this check will always be true, which is not what we want right

mantariksh: also should we insert this check on line 40 instead? we're already checking for `undefined`, so perhaps we should be checking for `null` as well? can reference `isConditionFulfilled` in `src/shared/util/logic.ts` for the current battle-tested implementation - not sure why we're not reusing that tbh (@karrui any idea?)

wanlingt: > what if currentValue is 0? then this check will always be true, which is not what we want right

yes good catch will fix! and gotcha will reference `isConditionFulfilled`

karrui: > also should we insert this check on line 40 instead? we're already checking for `undefined`, so perhaps we should be checking for `null` as well? can reference `isConditionFulfilled` in `src/shared/util/logic.ts` for the current battle-tested implementation - not sure why we're not reusing that tbh (@karrui any idea?)

src/shared is using backend types, so I had to copy and paste the code. At least I thought I copied them properly...

frontend/src/features/logic/utils/isConditionFulfilled.ts
```diff
@@ -37,7 +37,14 @@ export const isConditionFulfilled = (
   if (!isLogicableField(args)) return false
 
   const currentValue = getCurrentFieldValue(args.input, args.fieldType)
-  if (currentValue === undefined) return false
+  if (
+    currentValue === null ||
+    currentValue === undefined ||
+    // check if currentValue is empty for Decimal and Number fields
+    (typeof currentValue === 'string' && currentValue.length === 0)
```
karrui: should we reduce this check to 
```ts
currentValue === ''
```
instead?

karrui (approved):
lgtm

PR #4677, opened 2022-08-24
feat: call feedback download endpoint

### LINKED ISSUE #3795: [1] Download form feedback table in admin view
The current implementation in #3684 does not actually call the correct download endpoint, and instead converts the table into a csv. We should use the download endpoint for the most up to date feedback from the server.

tshuli:
## Problem
- Current implementation of feedback download converts the table into a csv
- With this PR, feedback download will now use the download endpoint for the most up to date feedback from the server

Closes #3795

## Tests
- [ ] Create form and submit multiple feedback
- [ ] Download feedback. Check that the `feedback/download` has been called in network tab. Check that feedback downloaded successfully.


tshuli:
~fixing an issue with the frontend tests - pass locally but not on github actions, probably due to different timezone~

frontend/package.json
```diff
@@ -37,6 +37,7 @@
     "libphonenumber-js": "^1.9.44",
     "lodash": "^4.17.21",
     "lottie-web": "^5.9.4",
+    "moment-timezone": "0.5.34",
```
karrui: should we reuse `date-fns-tz` over adding another dep?

see https://github.com/opengovsg/FormSG/blob/form-v2/develop/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.test.ts#L3

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackDownloadButton.tsx
```diff
@@ -1,39 +1,30 @@
...
-      index: entry.index,
-      date: entry.date,
-      feedback: entry.comment,
-      rating: entry.rating,
-    }))
-  }, [feedback])
+  const handleClick = useCallback(
+    () => downloadFeedback(formId || '', formTitle || ''),
+    [formId, formTitle],
+  )
```
karrui: should we create a mutation for this so we can get `react-query` to handle all loading/error states? as of now no errors are handled. See [example mutation here](https://github.com/opengovsg/FormSG/blob/form-v2/develop/frontend/src/features/admin-form/common/mutations.ts#L261), where there are `onSuccess` and `onError` handlers. the mutation's `isLoading` state can also be used to add a loading state to the feedback download button.

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
```diff
@@ -20,6 +22,9 @@ import { FeedbackTable } from './FeedbackTable'
 export const FeedbackPage = (): JSX.Element => {
   const { data: { average, count, feedback } = {}, isLoading } =
     useFormFeedback()
+  const { data: form } = useAdminForm()
+  const formTitle = form?.title
```
karrui: probably no need to assign this? can probably be used directly 

```tsx
formTitle={form?.title}
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackService.ts
```diff
@@ -1,13 +1,55 @@
...
+export const downloadFeedback = async (
+  formId: string,
+  formTitle: string,
+): Promise<void> => {
+  const expectedNumResponses = await countFeedback(formId)
+
+  return ApiService.get<FormFeedbackDto[]>(
+    `${ADMIN_FORM_ENDPOINT}/${formId}/feedback/download`,
+  ).then(({ data }) => {
+    if (!data) {
```
karrui: from what i can see, this check is unneeded, since something will always be returned. i don't think there is a need to throw an error here

frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor(expectedNumberOfRecords: number) {
+    super(expectedNumberOfRecords, 0)
+    this.setHeader(['Date', 'Comment', 'Rating'])
+  }
+
+  /**
+   * Generate a string representing a form feedback CSV line record
+   */
+  addLineFromFeedback(feedback: FormFeedbackDto) {
+    const createdAt = moment(feedback.created)
```
karrui: i think we should use `date-fns-tz` instead of using `moment`

frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
```diff
@@ -0,0 +1,26 @@
...
+      .tz('Asia/Singapore')
+      .format('DD MMM YYYY hh:mm:ss A')
+
+    this.addLine([createdAt, feedback.comment || '', feedback.rating])
```
karrui: nit: should we use the nullish coalescence operator `??` instead? Using boolean OR may convert a feedback comment of `0` into an empty string.

```tsx
let comment = 0
comment || '' // ''
comment ?? '' // 0
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackService.ts
```diff
@@ -1,13 +1,55 @@
...
     `${ADMIN_FORM_ENDPOINT}/${formId}/feedback`,
   ).then(({ data }) => data)
 }
+
+/**
+ * Counts feedback for a given form.
+ * @param formId the id of the form to count feedback for
+ * @returns the count of the retrieved feedback
+ */
+export const countFeedback = async (formId: string): Promise<number> => {
```
karrui: nit: should we name this function `getFormFeedbackCount` for consistency?

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackService.ts
```diff
@@ -1,13 +1,55 @@
...
+  ).then(({ data }) => data)
+}
+
+/**
+ * Downloads feedback for a given form.
+ * @param formId the id of the form to download feedback for
+ * @param formTitle the title of the form
+ * @returns a stream of feedback
+ */
+export const downloadFeedback = async (
```
karrui: nit: should we name this `downloadFormFeedback` for consistency

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackDownloadButton.tsx
tshuli: thanks! made the edits. see screenshots below:

success state
ss1

error state
ss2

loading state
ss3

frontend/src/features/admin-form/common/mutations.ts
```diff
@@ -388,3 +389,32 @@ export const usePreviewFormMutations = (formId: string) => {
...
+      toast({
+        description: error.message,
+        status: 'danger',
+      })
+    },
+    [toast],
+  )
+
+  const downloadFormFeedbackMutation = useMutation(
+    () => downloadFormFeedback(formId ?? '', formTitle ?? ''),
```
karrui: should we pass in the form Id and formTitle as parameters here instead of the hook?

May require adding type hints to `useMutation`. This allows the mutation function to be instantiated even when form is still loading.

```tsx
const downloadFormFeedbackMutation = useMutation(
    (formId, formTitle) => downloadFormFeedback(formId, formTitle),
    ...
)
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackDownloadButton.tsx
```diff
@@ -1,39 +1,36 @@
...
-    return feedback?.map((entry) => ({
-      index: entry.index,
-      date: entry.date,
-      feedback: entry.comment,
-      rating: entry.rating,
-    }))
-  }, [feedback])
+  const { downloadFormFeedbackMutation } = useFormFeedbackMutations(
+    formId ?? '',
+    formTitle ?? '',
```
karrui: as above, probably no need to pass in args to this hook first

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackDownloadButton.tsx
```diff
@@ -1,39 +1,36 @@
...
+  )
+
+  const handleClick = useCallback(
+    () => downloadFormFeedbackMutation.mutate(),
+    [downloadFormFeedbackMutation],
+  )
```
karrui:
```suggestion
  const handleClick = useCallback(
    () => {
      if (!formId || !formTitle) return
      return downloadFormFeedbackMutation.mutate(formId, formTitle)
    },
    [downloadFormFeedbackMutation],
  )
```

this will prevent the current bug where it may be possible (also not la, since its disabled) to trigger API call with empty args

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackDownloadButton.tsx
```diff
@@ -1,39 +1,36 @@
...
 
   return (
     <Button
-      disabled={isDisabled}
-      as={CSVLink}
-      filename={`${formId}-feedback.csv`}
-      data={data}
-      target="_blank"
+      disabled={isDisabled || downloadFormFeedbackMutation.isLoading}
```
karrui: should we use the `isDisabled` prop instead? that chakra prop injects a bunch of a11y props to denote a disabled state.

karrui (changes requested):

tshuli:
thanks @karrui ! address comments

karrui (approved):
lgtm

Issue #4659, opened 2022-08-22
[1] Do not validate on field creation in builder for mobile and home number fields


justynoh:
On admin dashboard when adding mobile field and home number field, “This field is required” error shows immediately

<img width="1482" alt="image" src="https://user-images.githubusercontent.com/25571626/185896797-6f6f688b-370d-4938-a335-da6778c44190.png">

PR #4684, opened 2022-08-25
feat: remove default value on Mobile field admin form preview

karrui:
does not seem compatible with `PhoneNumberInput`, will trigger validation immediately in admin form builder preview.


justynoh (approved):

justynoh:
Resolves #4659

PR #4685, opened 2022-08-25
fix: add navbar to billing page, fix misc css issues, modal popout conditionals

### LINKED ISSUE #4646: [0.5] Admin navbar doesn’t show up in billing page
Screenshot 2022-08-22 at 5 31 13 PM

karrui:
## Problem

This PR adds the navbar to the admin billing page. Also updates admin navbar logo to link to the dashboard instead of the landing page.

Also fixes a bunch of react errors that were popping out on the console, such as
- invalid svg kebab-cased props
- invalid -webkit-font-smoothing kebab case prop in theme rules
- missing key in billing table skeleton

also updates conditional to determine when the emergency contact or rollout modals popped out.

Closes #4646


justynoh (approved):
lgtm, my new favorite type is `boolean | undefined` :smile_cry:

frontend/src/app/AdminNavBar/AdminNavBar.tsx
```diff
@@ -118,25 +120,30 @@ export const AdminNavBar = ({ isMenuOpen }: AdminNavBarProps): JSX.Element => {
 
   // Emergency contact modal appears after the rollout announcement modal
   useEffect(() => {
-    if (!hasSeenContactModal && user && !user?.contact && hasSeenAnnouncement) {
+    if (
+      hasSeenContactModal === false &&
+      user &&
+      !user.contact &&
+      hasSeenAnnouncement === true
```
wanlingt, line 127: (for my learning!) - do we use `hasSeenAnnouncement === true` instead of just `hasSeenAnnouncement` for readability?

karrui: no, we could probably use `hasSeenAnnouncement` shorthand, oops hehe

frontend/src/features/workspace/WorkspacePage.tsx
```diff
@@ -44,10 +44,10 @@ export const WorkspacePage = (): JSX.Element => {
     [user],
   )
   const [hasSeenAnnouncement, setHasSeenAnnouncement] =
-    useLocalStorage<boolean>(ROLLOUT_ANNOUNCEMENT_KEY)
+    useLocalStorage<boolean>(ROLLOUT_ANNOUNCEMENT_KEY, false)
 
   const isAnnouncementModalOpen = useMemo(
-    () => !isUserLoading && !hasSeenAnnouncement,
+    () => !isUserLoading && hasSeenAnnouncement === false,
```
qin-guan, line 50: We should follow the same convention as @karrui mentioned - `!hasSeenAnnouncement`?

karrui: falseness is dangerous, hasSeenAnnouncement can be `undefined`

qin-guan: Would recommend having a default value to clear up ambiguity but this works too 👍🏻

karrui: the localstorage call has a default value to be set, which is why false check is good. there are 4 scenarios that could happen and all requires slightly different behaviour;

fetching from localstorage: undefined
fetched from localstorage, no key: false
fetched from localstorage, key is false: false
fetched from localstorage, key is true: true

PR #4686, opened 2022-08-25
fix(v2): make emergency contact modal phone number editable again

### LINKED ISSUE #4651: [2] Unable to edit existing number or select other countries in Emergency Contact modal
Can’t actually select other countries in the mobile field even though it’s presented as an option

justynoh:
## Problem
When the user has an existing emergency contact, the phone number in the emergency contact modal was not deletable.

Closes #4651 

## Solution
The cause of this is that when the control value was undefined, the Controller would default to switch the input value back to the default value and call onChange again. But actually, we should not be passing a default value in the first place because we have to wait till the user is retrieved to set the value (which the useEffect right below it does anyway).

Also, wrapped a function that wasn't in a useCallback in a useCallback. hehe

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/user/emergency-contact/components/ContactNumberInput.tsx
+++ b/frontend/src/features/user/emergency-contact/components/ContactNumberInput.tsx
@@ -34,11 +34,7 @@ const useContactNumberInput = () => {
     handleSubmit,
     reset,
     formState: { isValid, isSubmitting, errors },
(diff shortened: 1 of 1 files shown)


karrui (approved):
lgtm

wanlingt (approved):
lgtm!

PR #4697, opened 2022-08-26
fix(v2): add leading zero validation to decimal fields

### LINKED ISSUE #4640: [1] Leading 0s in decimal field are not validated in the frontend
**Describe the bug**
Error for leading 0s only shows up when user tries to submit the form. This should be displayed as a frontend validation message similar to MIN and MAX, e.g. 
image

Current state:
image

karrui:
## Problem

Server validation will reject decimal fields with leading zeros. We should add client-side validation so users do not encounter cryptic submission errors.

**Test cases**
```ts
/^0[0-9]/.test('000') // true
/^0[0-9]/.test('00') // true
/^0[0-9]/.test('0') // false

/^0[0-9]/.test('001') // true
/^0[0-9]/.test('01') // true
/^0[0-9]/.test('1') // false

/^0[0-9]/.test('0010') // true
/^0[0-9]/.test('010') // true
/^0[0-9]/.test('10') // false

/^0[0-9]/.test('0.10') // false
/^0[0-9]/.test('0.010') // false
/^0[0-9]/.test('.10') // false
/^0[0-9]/.test('00.10') // true
```

Closes #4640

## Solution

**Breaking Changes** 

**Bug Fixes**:

- fix: add leading zero validation to decimal fields

## Before & After Screenshots
Should have no change

### DIFF
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -219,6 +219,11 @@ export const createDecimalValidationRules: ValidationRuleFn<
         return 'Please enter a valid decimal'
       }
 
+      // Validate leading zeros
+      if (val.match(/^0[0-9].*$/)) {
+        return 'Please enter a valid decimal without leading zeros'
+      }
+
       if (!validateByValue) return true
 
       if (
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -220,7 +220,7 @@ export const createDecimalValidationRules: ValidationRuleFn<
       }
 
       // Validate leading zeros
-      if (val.match(/^0[0-9].*$/)) {
+      if (/^0[0-9]/.test(val)) {
         return 'Please enter a valid decimal without leading zeros'
       }
 


wanlingt (approved):
lgtm

frontend/src/utils/fieldValidation.ts
```diff
@@ -219,6 +219,11 @@ export const createDecimalValidationRules: ValidationRuleFn<
         return 'Please enter a valid decimal'
       }
 
+      // Validate leading zeros
+      if (val.match(/^0[0-9].*$/)) {
```
timotheeg: ~~So, small correction was needed after all 😅~~

SCRATCH THAT, I WAS WRONG 😱 testing for patterns that are not allowed is a bit of a mind bender 😅 

```suggestion
      if (/^0+([1-9.]|0$)/.test(val)) {
```

Reasoning:
~~1) (_functional bug_): testing the front for `/^0[0-9]/` would allow a double `0` at the lead and accept entries like `001` (oops! 😅). We have to be able to reject any number of leading zeros followed either by any digit 1 to 9 and whatever else or a zero~~
2) (_optimization_) `String.match()` return a data structure of the matches. For boolean-situations, better to use `RegExp.test()` instead
3) (_micro-optimization_) ending a test regex pattern with `.*$` is useless, because it simply means: "anything till the end". In such cases, better limit the test pattern to only the front part, to clearly indicate the intent of the test

 test cases
```
/^0+([1-9]|0$)/.test('000') // true
/^0+([1-9]|0$)/.test('00') // true
/^0+([1-9]|0$)/.test('0') // false

/^0+([1-9]|0$)/.test('001') // true
/^0+([1-9]|0$)/.test('01') // true
/^0+([1-9]|0$)/.test('1') // false

/^0+([1-9]|0$)/.test('0010') // true
/^0+([1-9]|0$)/.test('010') // true
/^0+([1-9]|0$)/.test('10') // false
/^0+([1-9]|0$)/.test('000.1') // false
```

karrui: something seems off with your regex too, it will wait for decimals with leading zeros

```ts
/^0+([1-9]|0$)/.test('000000.0001') // false
```

karrui: fixed in c87dad4

timotheeg (approved):
LGTM!

PR #4698, opened 2022-08-26
feat: add logout to public form header on submission end page (if auth'd), pass colorscheme to endpage

### LINKED ISSUE #4597: [3] signed in ID should show up on authed public form header on submission end page
Currently does not

Screenshot 2022-08-17 at 1 58 57 PM

karrui:
## Problem

This PR adds a logout button on submission end page if user is logged in, to better be in line with design.

Closes #4597.

## Solution

**Features**:

- feat: add logout button to form banner endpage

### DIFF
--- a/frontend/src/features/public-form/components/FormStartPage/FormBannerLogo.tsx
+++ b/frontend/src/features/public-form/components/FormStartPage/FormBannerLogo.tsx
@@ -1,11 +1,29 @@
 import { useEffect, useMemo, useState } from 'react'
 import { BiImage } from 'react-icons/bi'
-import { Box, Flex, Icon, Image, Skeleton, Stack, Text } from '@chakra-ui/react'
+import {
+  Box,
+  Flex,
+  Grid,
+  Icon,
+  Image,
+  Skeleton,
+  Spacer,
+  Stack,
+  Text,
+} from '@chakra-ui/react'
 
-interface FormBannerLogoInputProps {
+import { FormLogoutButton } from './FormLogoutButton'
+
+export interface FormBannerLogoProps {
   hasLogo: boolean
   logoImgSrc?: string
   logoImgAlt?: string
+  /**
+   * id of currently logged in user.
+   * If not provided, logout button will be hidden.
+   */
+  loggedInId?: string
+  onLogout?: () => void
 }
 
 export const InvalidLogo = ({ message }: { message: string }): JSX.Element => {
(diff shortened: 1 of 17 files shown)


frontend/src/features/public-form/components/FormStartPage/FormBannerLogo.tsx
```diff
@@ -56,15 +80,23 @@ export const FormBannerLogo = ({
   if (!hasLogo) return null
```
justynoh: If there's no logo then we don't actually display this form logout button at all. is that intended? seems a bit strange...

frontend/src/features/public-form/components/FormStartPage/FormLogoutButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+        />
+      ) : (
+        <Button
+          colorScheme={colorScheme}
+          variant="clear"
+          rightIcon={<BiLogOutCircle fontSize="1.5rem" />}
+          onClick={onLogout}
+        >
+          {loggedInId} - Sign out
+        </Button>
```
justynoh: seems that we always show this as a button even if `onLogout` is not provided (specifically, we don't provide the `onLogout` prop in the `EndPageView`) 

Should we disable the button in this case? Also, it just seems a bit weird that we allow for "logging out" after the form is submitted anyway. Maybe we should just have something that says "You have been successfully logged out"...

frontend/src/features/public-form/components/FormStartPage/FormBannerLogo.tsx
karrui: good catch. will talk to @pearlyong

karrui:
Draft, working on new screens provided in the issue

karrui:
new designs implemented.

frontend/src/features/public-form/components/FormStartPage/FormBannerLogo.tsx
karrui: will always display a logo now on end page (defaults to form logo, follows pearly's new designs)

frontend/src/features/public-form/components/FormStartPage/FormLogoutButton.tsx
karrui: disabled button if onLogout is not provided

>Also, it just seems a bit weird that we allow for "logging out" after the form is submitted anyway.

There used to be a "remember me" checkbox in AngularJS that kept the session alive for super long, but we have removed it. However, session still lasts for 24 hours, we should allow for users to log out regardless!

justynoh:
no change to stories, but there seems to be a visual regression for custom logo start page D:

karrui:
> no change to stories, but there seems to be a visual regression for custom logo start page D:

aiya, forgot update locations for the other stuff. form start page story regression is expected since form logo is now decoupled from start page. will fix and add new stories regardless

justynoh (approved):
lgtm

justynoh:
my favorite story is "No Logo"

PR #4699, opened 2022-08-26
feat(v2): prevent attachment field duplication if size limit exceeded, add toasts for duplication

### LINKED ISSUE #4648: [3] Disallow attachment field duplication if total attachment size will be exceeded
Duplicating an attachment field works even when it causes the attachment limit to be exceeded - should have service-layer validation

image

justynoh:
## Problem
Attachment fields can be duplicated even if the size limit ends up being exceeded which is a bug. In Angular, a modal pops up showing that the attachment can't be duplicated.

Simultaneously, found a difference between angular and react in that we don't notify users when logiced fields are duplicated. Added as part of the toast, as agreed on slack.

Closes #4648 

## Solution
1. Improve language of all the field update toasts.
2. Prevent attachment dupe:
a. In field row container, add check prior to dupe.
b. Factor out function getting max attachment size.
3. Add `logicedFieldIdsSet` to admin form logic, and access it in order to determine whether to show logic sentence in dupe toast.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/186885100-d0890702-3ee8-40a5-a35e-061d84dc74fd.mov

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
@@ -18,6 +18,7 @@ import { FormColorTheme } from '~shared/types'
 import { BasicField, FormFieldDto } from '~shared/types/field'
 
 import { useIsMobile } from '~hooks/useIsMobile'
+import { useToast } from '~hooks/useToast'
 import IconButton from '~components/IconButton'
 import Tooltip from '~components/Tooltip'
 import {
@@ -56,6 +57,7 @@ import { useBuilderAndDesignContext } from '../../BuilderAndDesignContext'
 import { PENDING_CREATE_FIELD_ID } from '../../constants'
 import { useDeleteFormField } from '../../mutations/useDeleteFormField'
 import { useDuplicateFormField } from '../../mutations/useDuplicateFormField'
+import { useCreateTabForm } from '../../useCreateTabForm'
 import {
   DesignState,
   setStateSelector,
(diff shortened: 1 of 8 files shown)


justynoh:
Actually, our backend should also probably validate the dupe request as well to ensure that it's valid. Might want to implement that to `develop` though

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -191,12 +197,31 @@ export const FieldRowContainer = ({
...
   const handleDuplicateClick = useCallback(() => {
-    // Duplicate button should be hidden if field
-    // is not yet created, but guard here just in case
-    if (stateData.state !== FieldBuilderState.CreatingField) {
-      duplicateFieldMutation.mutate(field._id)
+    // Duplicate button should be hidden if field is not yet created, but guard here just in case
+    if (stateData.state === FieldBuilderState.CreatingField) return
+    // Disallow duplicating attachment fields if after the dupe, the filesize exceeds the limit
+    if (field.fieldType === BasicField.Attachment) {
+      if (!form) return
```
karrui: Should we put this at the top of the function? Ideally no duplication will happen if there is no form anyways

karrui (approved):
Lgtm

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
justynoh: yesss moved in [4bdcda2](https://github.com/opengovsg/FormSG/pull/4699/commits/4bdcda2b8c579eca2c630ce339e255bae3a9991d)

PR #4705, opened 2022-08-29
feat(v2): add question titles to delete field modal, conditionally render message based on logic

### LINKED ISSUE #4644: [0.5] Delete field modal copy change
Delete field modal should display question titles rather than the field type (more relevant with MyInfo fields, where users associate fields with their intended content rather than datatype)
Fix for rollout: match Form Deletion modal. 

Modal title is just "Delete field"
Show field title in modal content with icon of field type and name of field (where form icon and form name are in the delete form modal)

image

justynoh:
## Problem
Currently the delete field modal doesn't match the delete modal format.

Closes #4644 

## Solution
Use `logicedFieldIdsSet` from admin logic to determine if the field is in logic. Also access the question numbers and titles from `useAdminFormLogic`.

_As a side note, actually `mapIdToFields` should be in `useCreateTabForm`, not `useAdminFormLogic` right..._

**Breaking Changes** 

## Screenshots

Nonlogiced field
image

Logiced field
image

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/DeleteFieldModal/DeleteFieldModal.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/DeleteFieldModal/DeleteFieldModal.tsx
@@ -1,18 +1,23 @@
 import { useCallback, useMemo } from 'react'
 import {
   ButtonGroup,
+  Icon,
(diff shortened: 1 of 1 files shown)


wanlingt (approved):
lgtm!

PR #4769, opened 2022-08-30
feat: remove special touch handling on Tooltip component

### LINKED ISSUE #4728: [1] bug: tooltip wrapped buttons require double tap on iOS Safari
Need to double click on builder tab to navigate to builder fields, tooltip quite annoying since it does not close
we should just remove special handling code of tooltip for mobile, and use default ChakraUI tooltip component instead

karrui:
## Problem

Components wrapped with our `Tooltip` component was requiring doubletap on iOS devices, which is probably less ideal UX than allowing users to look at tooltips without a long hold. This PR removes that custom touch handling code.

Closes #4728 

## Solution

**Breaking Changes** 

**Features**:

- feat: remove special touch handling on Tooltip component 

## Before & After Screenshots
Should have no change

### DIFF
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
@@ -1,8 +1,7 @@
 import { useMemo } from 'react'
-import { Icon, Stack, Text } from '@chakra-ui/react'
+import { Box, Icon, Stack, Text, Tooltip } from '@chakra-ui/react'
 
 import { BxsErrorCircle, BxsInfoCircle } from '~assets/icons'
-import Tooltip from '~components/Tooltip'
 
 import { BASICFIELD_TO_DRAWER_META } from '~features/admin-form/create/constants'
 import { FormFieldWithQuestionNo } from '~features/form/types'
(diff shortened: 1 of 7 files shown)


karrui:
converting to draft, alignment seems off now hehe

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
```diff
@@ -56,13 +55,11 @@ export const FieldLogicBadge = ({
...
     <LogicBadge display="inline-flex">
-      <Stack direction="row" spacing="0.25rem" maxW="100%">
-        <Tooltip
-          placement="top"
-          label={tooltipLabel}
-          wrapperStyles={{ display: 'flex' }}
-        >
-          <Icon as={tooltipIcon} fontSize="1rem" color={textColor} />
+      <Stack direction="row" spacing="0.25rem" maxW="100%" align="center">
+        <Tooltip placement="top" label={tooltipLabel} hasArrow>
```
justynoh: is there a reason not to use the Tooltip from our components here? Then we can also abstract out the `hasArrow` prop since we want arrows everywhere anyway

justynoh (approved):
lgtm with a question

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
karrui: no reason, just missed this out when i went back and forth on whether to keep our own tooltip component (with default arrows) or not haha. Fixed in 0b431af

PR #4777, opened 2022-08-30
fix(v2): use Portals instead of setting arbitrary zIndices

karrui:
## Problem

Instead of hoping no sibling component's zIndex values are higher in their render stack, use the `Portal` component to guarantee always on top of the stacking layout.

This PR only affects two components that were using arbitrary zIndex values - public form `MiniHeader` and the `SwitchEnvIcon` components.

## Solution

**Breaking Changes** 

**Improvements**:

- feat: use Portal instead of zIndex values for `SwitchEnvIcon` and `FormHeader` components
  - fixes the bug where dropdown options may appear in front of the mini header.


timotheeg (approved):
lgtm

Issue #4724, opened 2022-08-29
Background colour changes halfway for long dropdown options


mantariksh:
<img width="351" alt="Screenshot 2022-08-29 at 5 52 36 PM" src="https://user-images.githubusercontent.com/29480346/187175057-3e9157ba-fc8b-44c8-b207-7db2d7d4794c.png">

karrui:
what should the behaviour be? @pearlyong

pearlyong:
the entire option should be the same colour. so if it's on hover, then the entire option should be the hovered colour

pearlyong:
omg misunderstood - the text should hyphenate and wrap to next line if too long! lmk if any issues

PR #4778, opened 2022-08-30
fix(v2): incorrect dropdown options positioning on iOS Safari when the keyboard is open

### LINKED ISSUE #4718: [5] Dropdown options has incorrect positioning on iOS Safari when the keyboard is open
Only happens on iOS safari, might be due to keyboard shifting viewport. Will need to be fixed.

See related (unresolved) issue https://github.com/floating-ui/floating-ui/issues/805

Further reading on how we may get the correct position, or tell popperjs to update:
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

karrui:
> Builds on top of #4782 so the tests are fixed

## Problem

This PR replaces `react-popper` package usage with `@floating-ui/react-dom-interactions`. After experimentation, the layout is properly being updated on iOS Safari, which closes #4718 properly. The dev experience of using the package was also much better compared to react-popper. It is also the official successor of popperjs, which means a more updated package.

Closes #4718

## Solution

**Breaking Changes** 

**Features**:

- feat: use `@floating-ui/react-dom-interactions` instead of `react-popper` for Select components

**Bug Fixes**:

- Fixes bug where dropdown options were out of sync with the current viewport on iOS Safari.

## Before & After Screenshots
Should have no changes tbh.

## Tests

No changes to tests

## Deploy Notes

**New dependencies**:

- `@floating-ui/react-dom-interactions`: replaces `react-popper` for rendering Select component's dropdown options

### DIFF
--- a/frontend/src/components/Dropdown/components/SelectPopover/SelectPopover.tsx
+++ b/frontend/src/components/Dropdown/components/SelectPopover/SelectPopover.tsx
@@ -1,48 +1,51 @@
-import { FC, useRef, useState } from 'react'
-import { Modifier, usePopper } from 'react-popper'
+import { FC, useLayoutEffect, useMemo, useRef } from 'react'
 import { Box, useMergeRefs, useOutsideClick } from '@chakra-ui/react'
+import {
+  autoUpdate,
+  flip,
+  hide,
+  offset,
+  useFloating,
+} from '@floating-ui/react-dom-interactions'
 
 import { useSelectContext } from '~components/Dropdown/SelectContext'
 
 import { SelectPopoverContext } from './SelectPopoverContext'
 
-export const matchWidth: Modifier<'matchWidth'> = {
-  name: 'matchWidth',
-  enabled: true,
-  phase: 'beforeWrite',
-  requires: ['computeStyles'],
-  fn: ({ state }) => {
-    state.styles.popper.width = `${state.rects.reference.width}px`
-  },
-  effect:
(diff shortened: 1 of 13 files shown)


karrui:
~~DropdownField tests are failing, will investigate~~
Fixed in #4782

karrui:
Before i forget tmr: long selected options are no longer truncated (will probably be cause by chromatic)

Oh, issue already created: #4724

frontend/src/components/Dropdown/components/SelectPopover/SelectPopover.tsx
```diff
@@ -1,48 +1,51 @@
...
+    }),
+    [strategy, x, y],
+  )
+
+  // Allows
+  useLayoutEffect(() => {
+    if (isOpen && refs.reference.current && refs.floating.current) {
+      return autoUpdate(refs.reference.current, refs.floating.current, update)
+    }
+  }, [isOpen, update, refs.floating, refs.reference])
```
timotheeg, line 48: /justNoting: I don't have a suggestion here, but reading this file on its own, it's a little hard to keep track what all the refs are about 😰 :
* `reference`
* `wrapperRef`
* `mergedReferenceRefs`
* `refs.reference`
* `refs.floating`

Reading [the doc](https://floating-ui.com/docs/react-dom) helped a bit!

timotheeg (approved):
lgtm

PR #4787, opened 2022-09-01
feat: handle caught body-parser errors

karrui:
## Problem

When `body-parser` errors occur, such as when malformed JSON bodies are sent to our server, our generic error handler incorrectly returns a `500` http status code.

This PR adds correct error handling to body-parser errors.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Return correct status codes when body-parser errors are caught.

## Tests

Add unit tests for catching invalid JSON bodies

## Deploy Notes

**New dependencies**:

- `http-errors`: used for checking and handling errors from `body-parser`. (`body-parser` uses http-errors package to generate errors, as documented [here](https://github.com/expressjs/body-parser#errors))


timotheeg (approved):
Thanks KarRui!

Code LGTM from the perspective of throwing 400, which indicate the query should **not** be retried as-is.

I think a deeper problem is still that we are even running the body parser at all, in routes we should not. We should still fix that later.

Issue #4620, opened 2022-08-22
Tooltip for form items


qin-guan:
**Describe the bug**
Tooltips are missing in form items when editing

**To Reproduce**
1. Create a form
2. Add a field
3. Hover over the duplicate/delete buttons 

**Expected behavior**
Tooltip should show on hover

**Screenshots**
<img width="144" alt="image" src="https://user-images.githubusercontent.com/10321883/185886244-0d6bd835-49c1-4891-a19a-c12bfbd4df87.png">

mantariksh:
Copy:

Duplicate field
Delete field

PR #4789, opened 2022-09-04
feat: add tooltip description to field actions

darrelhong:
a very small PR that adds tooltips to the form field actions row in the form builder. The trash and settings button is pretty clear what it does but the duplicate button not as much. Could have added a title attribute instead but since chakra makes it easy to add a tooltip, a tooltip might be visually better.

## Before & After Screenshots
**AFTER**:
Screen Shot 2022-09-04 at 10 17 50 PM

Screen Shot 2022-09-04 at 10 17 44 PM

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
@@ -372,23 +372,27 @@ export const FieldRowContainer = ({
                     {
                       // Fields which are not yet created cannot be duplicated
                       stateData.state !== FieldBuilderState.CreatingField && (
-                        <IconButton
-                          aria-label="Duplicate field"
-                          isDisabled={isAnyMutationLoading}
-                          onClick={handleDuplicateClick}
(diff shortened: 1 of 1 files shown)


qin-guan:
Hey @darrelhong, thanks for the PR (: lgtm 

Closes https://github.com/opengovsg/FormSG/issues/4620

fyi @justynoh

timotheeg (commented):
Thanks @darrelhong. Please drop the `openDelay` property, we are not using it in any instance of `Tooltip` as we rely on the default instead.

darrelhong:
I actually added delay because
1. it maintains the current behaviour (no tooltip unless hold)
2. the icons already convey meaning
3. the main purpose of the tooltip is to give additional hints (for people who might are unsure what the icon means) rather than immediate info

To better understand, can observe the tooltips on github icons buttons, which appear after a slight delay, vs the tooltips on profile avatars, which has a noticeably shorter delay.

Let me know if the points are valid, if not, I'll be happy to update the PR.

pearlyong:
think we don't need a delay on tooltip appearing - we don't currently have that behaviour on our other tooltips! thanks

darrelhong:
got it, rebased and updated

justynoh (approved):
lgtm, thanks for the contribution! 🙂

PR #4791, opened 2022-09-05
fix(v2): fix render bug in delete field modal, misc builder improvements

### LINKED ISSUE #4657: [0.5] Draggable icon should not be displayed if field is in Creating state
Six dots icon for field drag and drop doesn’t work while the field has not been created, should test this with more users / monitor, we make it work or prevent the six dots from rendering before the field has been created?

### LINKED ISSUE #4729: [2] bug: Pressing close button on create field in mobile should discard the field instead of keeping field around
Only on mobile!

Steps to replicate:
- Create any field
- Click close button on field
- See field that is pending creation still being in the builder preview

### LINKED ISSUE #4739: Display verification button for verified fields in the builder
We don’t display the “verified” button in the form builder when OTP verification is enabled for email/mobile fields. Admins probably expect the builder preview to be as identical to the public form as possible - we can possibly add a disabled "Verify" button there. (Maybe even with a tooltip that says "Verification will be enabled for your form respondents" or smth)

### LINKED ISSUE #4741: Remove "Save field" footer from MyInfo fields edit drawer
MyInfo edit field drawer in builder should not have drawer footer (i.e. we should remove the “Save field” and “cancel” buttons) since users can’t edit myinfo fields anyway. 

Note: they should still have the create field button so they actually can add the field!

### LINKED ISSUE #4751: Rename statement to paragraph
Screenshot 2022-08-29 at 6 15 12 PM

justynoh:
## Problem
1. Delete field modal had a render bug (question numbers were not calculated correctly due to modal rendering on builder open rather than delete field button onClick). Also, trying to create fields crashes. oops.
2. Hide draggable icon when field is in creating state, closes #4657
3. Rename statement to paragraph in Edit Paragraph drawer, closes #4751 
4. Remove "Save field" footer from MyInfo edit drawers, closes #4741 
5. Do not retain create fields in mobile, closes #4729 
6. Verify button should be displayed for verified fields in the builder, closes #4739 

## Solution
1. `DeleteFieldModal`, `BuilderAndDesignTab` modified to fix render bugs by rendering modal only when deleteFieldModalDisclosure is open.
2. `FieldRowContainer` now hides draggable icon
3. `EditParagraph` field name is now Paragraph
4. `EditMyInfo` only renders the drawer footer if creating, otherwise no footer is shown. Additionally, fixed a small bug with `DrawerContentContainer` which dividers to show in between `null` or `undefined` children (which was necessary in this case).
5. `CreatePageSidebarContext` > `setActiveOrPendingTab` always sets to inactive field if creation is cancelled, regardless of mobile or desktop.
6.  Add `VerifiedFieldBuilderContainer` which wraps verified input fields with the builder button (which will be always disabled, as confirmed by @pearlyong). Use it in `FieldRowContainer` to wrap mobile and email fields if verified, a la `FieldFactory`. 

**Breaking Changes** 


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/VerifiableFieldBuilderContainer.tsx
```diff
@@ -0,0 +1,33 @@
...
+  children,
+}: VerifiableFieldBuilderContainerProps): JSX.Element => {
+  return (
+    <FieldContainer schema={schema}>
+      <Stack spacing="0.5rem" direction={{ base: 'column', md: 'row' }}>
+        {children}
+        <Box>
+          <Button isDisabled colorScheme={`theme-${colorTheme}`}>
+            Verify
+          </Button>
```
karrui, line 28: comment: might need design input, but should we allow users to transition to the verified state instantly by clicking the verify button so admins can visualize how verified fields work?

karrui (approved):
lgtm

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/VerifiableFieldBuilderContainer.tsx
justynoh: confirmed this with @pearlyong earlier today, decided that it would just be disabled in the builder!

PR #4794, opened 2022-09-05
feat(v2): add navigation prompt modal when navigating away during submission download

### LINKED ISSUE #4144: [1] prompt using usePrompt hook to prevent navigation does not work when cancelling

karrui:
## Problem

This PR adds a modal that appears when a form admin attempts navigation (unmounting the download progress component) during a form download.

Also fixes a bug where the progress modal can be closed by the escape key. This should not happen and users should not be able to close the modal during downloads.

Closes #4144 

## Solution

**Breaking Changes** 

**Features**:

- feat: replace `usePrompt` hook with `NavigationPrompt` component

**Bug Fixes**:

- fix: prevent progress modal from being closed by escape key

## Before & After Screenshots
Screenshot 2022-09-05 at 3 39 53 PM


justynoh (approved):
lgtm

Issue #4717, opened 2022-08-29
[2] Table field - cannot add rows in public form


wanlingt:
Can add rows in admin preview, cannot add in public form. 
To replicate:
Click add another row - ok
Click again - error required field (not supposed to show here)
Click again - goes to an empty page

wanlingt:
@pearlyong could you help to add the exact steps to replicate?

pearlyong:
**GSIB issue**
To replicate:
- As an admin, add a Table field
- Select 'Allow respondents to add more rows'
- Set Maximum rows allowed to 5 (5 is arbitrary)
- Try adding more rows in Admin builder preview, it should work fine
- Go to public form view and add more rows
- If you add 1 row, it's ok
- If you add another row, there'll be an inline error 'This field is required' (this error is not supposed to appear now)
- If you add one more, the page will go blank

PR #4798, opened 2022-09-05
fix(v2): handle null errors for table fields

justynoh:
## Problem
Currently, if table fields have no error, the app crashes because we try to call `Object.values` on `null`.

Also, clicking Add more rows the first time works, but the second time it causes the required error to show up. This is because we autofocus on the new row on the first time we Add more rows. Then, the second time, the click blurs the input, triggering validation.

Closes [#4717]

## Solution
Add "error handling" for the case where the field has no errors (lmaoo)

Don't autofocus on Adding more rows to prevent triggering validation

**Breaking Changes** 

### DIFF
--- a/frontend/src/templates/Field/Table/TableField.tsx
+++ b/frontend/src/templates/Field/Table/TableField.tsx
@@ -59,9 +59,7 @@ export const TableField = ({
     // would not need to be shown in the table field itself.
     if (isMobile) return
     // Get first available error amongst all column cell errors.
-    return head(
-      uniq(tableErrors?.flatMap((err = {}) => Object.values(err)))
-    )
+    return head(uniq(tableErrors?.flatMap((err = {}) => Object.values(err))))
   }, [isMobile, tableErrors])
 
   const { fields, append, remove } = useFieldArray<TableFieldInputs>({
--- a/frontend/src/templates/Field/Table/TableField.tsx
+++ b/frontend/src/templates/Field/Table/TableField.tsx
@@ -59,7 +59,9 @@ export const TableField = ({
     // would not need to be shown in the table field itself.
     if (isMobile) return
     // Get first available error amongst all column cell errors.
-    return head(uniq(tableErrors?.flatMap(Object.values)))
+    return head(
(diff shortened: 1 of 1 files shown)


frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -82,15 +89,15 @@ export const TableField = ({
...
+  }, [appendTableRow, fields.length, hasMinRowsChanged, remove, schema])
 
   const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } =
     useTable({ columns: columnsData, data: fields })
 
   const handleAddRow = useCallback(() => {
     if (!schema.maximumRows || fields.length >= schema.maximumRows) return
-    return append(createTableRow(schema))
-  }, [append, fields.length, schema])
+    return appendTableRow()
```
karrui, line 97: this is a regression for UX. Should we only prevent focus if in admin form builder preview?

justynoh: the problem occurs in public form tho

justynoh: Maybe there is a way to prevent triggering on blur when clicking the button. But this then goes to the same issue that we are having with verified field buttons if we don't want the validation to trigger on click. The difference is that there, respondents type the input first but here they don't input anything, so they wouldn't expect to be greeted with an error instantly.

frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -59,21 +59,28 @@ export const TableField = ({
     // would not need to be shown in the table field itself.
     if (isMobile) return
     // Get first available error amongst all column cell errors.
-    return head(uniq(tableErrors?.flatMap(Object.values)))
+    return head(
+      uniq(tableErrors?.flatMap((errs) => (errs ? Object.values(errs) : []))),
```
karrui: nit:
```suggestion
      uniq(tableErrors?.flatMap((err = {}) => Object.values(err)))
```

karrui (approved):
lgtm

Issue #4711, opened 2022-08-29
[2] Decimal field logic validation is too strict, using strict string matching instead of number


karrui:
Decimal logic currently does not account for decimal points. It should be loosened as long as the numeric value is equal.

For example:
if Logic condition of Decimal field IS EQUAL TO `1`, value of 1.0 **will not** trigger logic, only `“1”` exactly. Same thing if `1.` is entered.

Both logic condition and field value should be normalized to a number instead of string matching

PR #4802, opened 2022-09-05
fix(v2): add numerical comparison for decimal equality in logic

justynoh:
## Problem
For logic blocks requiring equality on decimals, we use string comparison when angular uses numerical comparison. We should mimic angular behavior for backward compatibility.

Closes [#4711] 

## Solution
Cast values to numbers before checking equality for blocks relying on decimal conditions.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/logic/utils/isConditionFulfilled.ts
+++ b/frontend/src/features/logic/utils/isConditionFulfilled.ts
@@ -78,6 +78,11 @@ export const isConditionFulfilled = (
         }
         return condition.value === currentValue.value
       }
+      // In angular, number equality is string=== but decimal equality is number===.
+      // Need to replicate this behavior for backward-compatibility.
+      if (fieldType === BasicField.Decimal) {
+        return Number(currentValue) === Number(condition.value)
+      }
       return String(condition.value) === currentValue
     }
   }


wanlingt (commented):
should we do this validation in the backend too? believe it's in [`src/shared/util/logic.ts`](https://github.com/opengovsg/FormSG/blob/form-v2/develop/src/shared/util/logic.ts)

timotheeg:
> should we do this validation in the backend too? believe it's in [src/shared/util/logic.ts](https://github.com/opengovsg/FormSG/blob/form-v2/develop/src/shared/util/logic.ts)

Eventually yes, but we should reduce the server side changes. Can we track this as a to-do issue for later?

frontend/src/features/logic/utils/isConditionFulfilled.ts
```diff
@@ -78,6 +78,11 @@ export const isConditionFulfilled = (
         }
         return condition.value === currentValue.value
       }
+      // In angular, number equality is string=== but decimal equality is number===.
+      // Need to replicate this behavior for backward-compatibility.
+      if (fieldType === BasicField.Decimal) {
```
timotheeg, line 83: Not a comment on your PR, but why are `input` and `fieldType` accessed onto `args` in the uses above? 🤔

timotheeg (approved):
lgtm

frontend/src/features/logic/utils/isConditionFulfilled.ts
justynoh: i think `isLogicableField` is a type guard, and it does something to restrict the type of `args.input` and `args.fieldType`

timotheeg: Thanks! Can we verify that actually? It looks like the other uses are themselves in checks (e.g. `if (isRadioFormFieldValue(currentValue, args.fieldType))`, so I'm not sure an access from `args` really matters? 🤔 

And I guess my question was to highlight consistency. This check here is using `fieldType` rather than `args.fieldType`, and it's not clear whether there's purpose in the differing uses.

PR #4805, opened 2022-09-06
feat(v2): hide react switch banner on AngularJS

### LINKED ISSUE #4595: [0.5] 0% rollout should hide react switch banner on angularjs
not sure if this is the case, creating issue just in case

wanlingt:
## Problem

The banner that informs users a new FormSG is now available is still visible when the React admin rollout % is 0.

Closes #4595

## Solution

Show the banner if `reactMigrationAdminRollout > 0`

### DIFF
--- a/src/public/modules/forms/admin/views/list-forms.client.view.html
+++ b/src/public/modules/forms/admin/views/list-forms.client.view.html
@@ -5,9 +5,9 @@
   <div id="list-tab">
     <section data-ng-init="vm.listForms()">
       <div id="list-form" class="container" ng-if="vm.user">
-        <div id="react-switch-banner" ng-if="vm.reactAdminRollout > 0">
-          <react-switch-banner-component></react-switch-banner-component>
-        </div>
+        <react-switch-banner-component
+          ng-if="vm.reactAdminRollout > 0"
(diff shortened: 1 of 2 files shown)


src/public/modules/forms/admin/views/list-forms.client.view.html
```diff
@@ -5,7 +5,9 @@
   <div id="list-tab">
     <section data-ng-init="vm.listForms()">
       <div id="list-form" class="container" ng-if="vm.user">
-        <react-switch-banner-component></react-switch-banner-component>
+        <div id="react-switch-banner" ng-if="vm.reactAdminRollout > 0">
+          <react-switch-banner-component></react-switch-banner-component>
```
karrui: my angularjs is rusty, but i think we can put the `ng-if` directly on the component
```tsx
<react-switch-banner-component ng-if="vm.reactAdminRollout > 0"></react-switch-banner-component>
```

karrui (approved):
lgtm

src/public/modules/forms/admin/views/list-forms.client.view.html
wanlingt: just tried it, can it works!

PR #4808, opened 2022-09-06
feat(v2): highlight response table row on hover

### LINKED ISSUE #4128: [0.5] Make response table row highlight on hover
1. Currently, clicking on storage response table rows directs the user to a page with an individual response. However, this is not apparent to the user from the page itself. We can consider changing the cursor to a hand as well as highlighting the row when the cursor hovers over the row to make it more apparent that the row is clickable. Look for @pearlyong for design specifics on this :)

4. Additionally, right now, the reference ID in the table is not highlightable (clicking and dragging along the reference ID causes the individual response page to open instead). Lumping in with this issue because it involves the same component.

wanlingt:
## Problem

It's not apparent to users that they can click into each row of the storage mode responses table

Closes #4128

## Solution

- Add `_hover` and `_active` states
- Change pointer to `cursor`

## Screenshots
**After**
Recording 2022-09-06 at 14 35 46


justynoh (approved):
lgtm

Issue #4748, opened 2022-08-29
[1] [mobile] Individual responses page is very wonky


justynoh:
<img width="430" alt="image" src="https://user-images.githubusercontent.com/25571626/187178945-cf03b1f0-cc39-48af-b431-e41f323945d7.png">

wenhaogoh:
Failed QA:
- Components did not stack properly on iPhone 12 iOS Safari
- Large padding when response table is scrolled all the way to the right

![IMAGE 2022-09-26 14:46:38](https://user-images.githubusercontent.com/52305097/192210652-ea10d2b2-4849-4f8b-9ece-2fcefae32859.jpg)

Issue #4732, opened 2022-08-29
[1] Lock individual response page navigation bar


justynoh:
For viewing storage responses, we should lock the top bar with “<- back to list, Respondent #1  <  >” in place, because it locks in angular. If admin scrolls down far through a long response, they have to scroll all the way back up to get back to the responses screen (or click the “RESPONSES” button but that seems a bit less intuitive)

Issue #4730, opened 2022-08-29
[0.5] Storage mode individual responses for verified fields do not match Angular


justynoh:
Storage mode responses are inconsistent between angular and react for verified fields! It is missing the "[verified] " prefix (i only checked for mobile soz, whoever picks up this issue can check for email verification too)

<img width="456" alt="image" src="https://user-images.githubusercontent.com/25571626/187176205-f908af22-82df-40ea-b6f5-3ba535ae6f73.png">

PR #4812, opened 2022-09-07
fix(v2): lock individual responses navbar, fix mobile view, add verified labels

justynoh:
## Problem
1. #4730 
2. #4732 
3. #4748 

## Solution
1. [verified] label was under header where it would never get used. Moved it to question which is where it actually is needed. lol
2. Created individual response navbar component and locked it in place at the top of the navbar. (Also needed to change form navbar `boxShadow` to `borderBottom`, otherwise the individual response navbar's zindex will overlap it, that's why there's a change to that file) (No changes to the computations of any of the variables, just shifted things around. Only the returned JSX elements have changes)
3. Added mobile support to all components of the individual response page.

**Breaking Changes** 

## Screenshots

1. 
image

2. and 3.

https://user-images.githubusercontent.com/25571626/188803560-4b910462-9c3b-4ede-b3e5-59e3cd5e24f9.mov


karrui (approved):
lgtm

karrui:
can we add a mobile story (do we even have stories for individual responses hehe)

PR #4813, opened 2022-09-07
feat: block navigation on public form page when form is dirty

### LINKED ISSUE #4793: Prevent accidental navigation on public form page
If a user on the public form page accidentally navigates back or forwards on their browser (e.g. by swiping the trackpad) and returns to the public form page, they lose all information that they've already filled in

**Describe the solution you'd like**
Possible solution: A confirmation modal pops up to ask if they're sure they want to navigate away from the page

darrelhong:
Closes #4793

the existing useNavigationPrompt can be used to fulfill the goal of preventing accidental navigation after form is modified.

Currently the implementation simply uses the isDirty variable provided useForm.formState to determine if navigation should be blocked. When a text field is touched and unfocused with no input change, it is considered dirty already, so if that is not the desired behaviour, further logic is needed perhaps similar to holdingStateDate in the form builder. One very important consideration here is I'm not sure if there is a situation when the isDirty=true but navigation is actually allowed.

Other than the submit button and switch to react, the public form page should not have any links that redirects the page using react-router (might be wrong here, need more context), so the only way of exiting the page is using browser functions. Thus instead of using the underlying react-router-dom functions, native web apis can be used instead for simplicity. The functionality remains the same in my testing.

The user flow for switch to react is also affected as the prompt will be shown after the feedback modal. Not sure if this is desirable and whats the best way to go about this, perhaps there is an easy way to transfer data when switching to angular? 

Screen Shot 2022-09-07 at 2 11 10 PM

Another solution could be to simply save responses localstorage and do away with prompts which might not be the best UX.

### DIFF
--- a/frontend/src/features/env/PublicFeedbackModal.tsx
+++ b/frontend/src/features/env/PublicFeedbackModal.tsx
@@ -34,6 +34,7 @@ import { ModalCloseButton } from '~components/Modal'
 import Textarea from '~components/Textarea'
 
 import { useEnvMutations, useFeedbackMutation } from '~features/env/mutations'
+import { usePublicFormContext } from '~features/public-form/PublicFormContext'
 
 import { usePublicFeedbackFormView } from './queries'
 import { isUsableFeedback } from './utils'
@@ -50,6 +51,7 @@ export const PublicFeedbackModal = ({
   responseMode?: FormResponseMode
   authType?: FormAuthType
 }): JSX.Element => {
+  const { setBlockNavigation } = usePublicFormContext()
   const modalSize = useBreakpointValue({
     base: 'mobile',
     xs: 'mobile',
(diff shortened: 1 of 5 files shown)


darrelhong:
@wanlingt will close this PR if its not needed anymore

wanlingt:
> @wanlingt will close this PR if its not needed anymore

Hi @darrelhong , thanks for your work on this and apologies for taking a while to get to your PR. Displaying the prompt when clicking switch back to angular might cause some confusion to users, additionally there are other links in the form footer that navigate away from the page. We could limit this to the back and forward button events for now, let me know what you think!

darrelhong:
@wanlingt luckily it looks like all the footer links open in a new tab, so navigation is not affected. For the case of switching to angular, I have updated to not block navigation.

darrelhong:
@wanlingt any update on this? would like to get it merged if possible

foochifa:
Heya @darrelhong we will be removing angular and the feedback modal #6190! Thanks for your contribution, but will be closing this!

PR #4814, opened 2022-09-07
feat(v2): update landing page copy

### LINKED ISSUE #4709: [0.5] Landing page updates - copy, modify stats endpoint
1. Landing page “Build secure government forms in minutes.” should not have period
2. Landing page image “Staff Travel Declarati\nons 2021” i thought we are changing the image
3. ~OGP Products modal~ (post rollout)

Passing on to PM/Design

wanlingt:
## Problem

Landing page needs updating

Closes #4709

## Solution

- Replace images with animations
- Return agency count from `getLandingPageStatistics`

**Breaking Changes** 

**Improvements**:

- Moved `LottieAnimation` out of `NewFeatureContent` into a separate file


wanlingt:
Converting to draft to work on design review comments

frontend/src/pages/Landing/LandingPage.tsx
```diff
@@ -45,10 +45,12 @@ import { useMdComponents } from '~hooks/useMdComponents'
 import Button from '~components/Button'
 import Link from '~components/Link'
 import { Tab } from '~components/Tabs'
+import { LottieAnimation } from '~templates/LottieAnimation/LottieAnimation'
```
karrui:
```suggestion
import { LottieAnimation } from '~templates/LottieAnimation'
```

frontend/src/pages/Landing/LandingPage.tsx
```diff
@@ -87,13 +89,18 @@ export const LandingPage = (): JSX.Element => {
...
           direction={{ base: 'column', lg: 'row' }}
           align="center"
           spacing={{ base: '1.5rem', md: '3.125rem', lg: '7.5rem' }}
+          pl={{ base: '1.5rem', md: '5.5rem', lg: '9.25rem' }}
         >
-          <Flex flexDir="column" flex={1}>
+          <Flex
+            flexDir="column"
+            flex={1}
+            pr={{ base: '1.5rem', md: '5.5rem', lg: '0' }}
```
karrui, line 102: any reason for the right padding? if not, I think we can just adjust the `Stack` parent's `spacing` prop, which adds spacing between its children.

karrui (approved):

frontend/src/pages/Landing/LandingPage.tsx
wanlingt: the right padding is used when the layout is a vertical stack (so when the layout is < `lg`). `spacing` won't work in this case since it'll be vertical spacing

didn't apply the right padding to the `Stack` parent because both its children have different right padding values - the text portion has right padding, but `LottieAnimation` is flush to the right viewport so has 0 right padding

wanlingt:
landing page stats are not being captured in chromatic, we can't figure out why (experimented with adjusting both msw and chromatic delay settings). @karrui suspects it's due to `toLocaleString`. ignoring for now.

PR #4815, opened 2022-09-07
feat(v2): move settings tabs for mobile to left side

### LINKED ISSUE #4744: [3] Settings tabs in mobile should be at the side instead of bottom
@pearlyong can link the design?

justynoh:
## Problem
Settings page tabs currently exist at the bottom of the page. We want them to exist on the left side instead, to be consistent with desktop.

Closes #4744 

## Solution
Remove the custom CSS for creating the sidebar at the bottom of the page.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/188815476-56bdba19-766b-46f0-bfc8-0dea539f9467.mov

### DIFF
--- a/frontend/src/features/admin-form/settings/SettingsPage.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsPage.tsx
@@ -45,8 +45,8 @@ export const SettingsPage = (): JSX.Element => {
         isManual
         orientation="vertical"
         variant="line"
-        py={{ base: '3.5rem', lg: '4rem' }}
-        px={{ base: '1.75rem', lg: '2rem' }}
+        py={{ base: '2.5rem', lg: '3.125rem' }}
+        px={{ base: '1.5rem', md: '1.75rem', lg: '2rem' }}
       >
         <Flex
           h="max-content"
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -38,47 +36,31 @@ export const SettingsPage = (): JSX.Element => {
...
       <Tabs
         isLazy
         isManual
-        orientation={tabOrientation}
+        orientation="vertical"
         variant="line"
-        py={{ base: '2.5rem', md: '3.5rem', lg: '4rem' }}
-        px={{ base: '1.5rem', md: '1.75rem', lg: '2rem' }}
+        py={{ base: '3.5rem', lg: '4rem' }}
+        px={{ base: '1.75rem', lg: '2rem' }}
```
karrui: may still need to update the padding here, since mobile padding is slightly less looking at the figma

justynoh: which part of the figma? :o

justynoh: ok, updated in [f084a00](https://github.com/opengovsg/FormSG/pull/4815/commits/f084a0043f9f3871c94b345adbcdfe7b4254bf9b). Recomputed the x and y paddings so it should look correct now

timotheeg (approved):
LGTM

When the view is really narrow, it looks like the left-margin could be reduced even further 🤔 , but we can fine-tune later.

PR #4816, opened 2022-09-07
chore(v2): remove partytown

wanlingt:
## Problem
We no longer use `partytown` to load Google Analytics 4 (related to #4665)

## Solution

Remove `partytown`.

We can explore re-enabling `partytown` with GA4 when we look into optimising our deployment pipeline.


timotheeg (approved):
lgtm - verified partytown is removed:

```bash
$ git grep -i partytown | wc -l
       0
```

PR #4821, opened 2022-09-08
feat(v2): enable refetch on mount for storage responses

### LINKED ISSUE #4737: [1] Storage mode responses do not update automatically and need to refresh in React
In angular, admin user could refresh storage mode responses by returning to admin dashboard and clicking back into the form. In react, admin user need to refresh browser in order to refresh storage mode responses

justynoh:
## Problem
Storage responses was not refetching until the admin refreshed the page. But once they refresh the page, they need to upload their secret key again, creating higher friction for admins to see their updated form responses.

Closes #4737 

## Solution
Remove custom refetch settings.

Also made some changes to the storage responses page for mobile responsiveness.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/189083049-fed255c0-cb18-4606-81e9-c2c02ec78bdf.mov

### DIFF
--- a/frontend/src/features/admin-form/responses/queries.ts
+++ b/frontend/src/features/admin-form/responses/queries.ts
@@ -90,6 +90,7 @@ export const useFormResponses = ({
     adminFormResponsesKeys.metadata(formId, params),
     () => getFormSubmissionsMetadata(formId, params),
     {
+      staleTime: 0,
       keepPreviousData: !submissionId,
       enabled: !!secretKey && (page > 0 || !!submissionId),
     },
--- a/frontend/src/features/admin-form/responses/queries.ts
+++ b/frontend/src/features/admin-form/responses/queries.ts
@@ -90,11 +90,6 @@ export const useFormResponses = ({
     adminFormResponsesKeys.metadata(formId, params),
     () => getFormSubmissionsMetadata(formId, params),
     {
-      refetchOnMount: true,
-      // Consider data stale after 10s, prevent admins from initiating too many
-      // fetches by continuously clicking out and in of the results tab to
-      // refetch data
-      staleTime: 10000,
       keepPreviousData: !submissionId,
       enabled: !!secretKey && (page > 0 || !!submissionId),
     },
--- a/frontend/src/features/admin-form/responses/queries.ts
+++ b/frontend/src/features/admin-form/responses/queries.ts
@@ -91,9 +91,10 @@ export const useFormResponses = ({
     () => getFormSubmissionsMetadata(formId, params),
     {
       refetchOnMount: true,
-      refetchOnWindowFocus: false,
-      // Data will never change.
-      staleTime: Infinity,
+      // Consider data stale after 10s, prevent admins from initiating too many
+      // fetches by continuously clicking out and in of the results tab to
+      // refetch data
+      staleTime: 10000,
       keepPreviousData: !submissionId,
       enabled: !!secretKey && (page > 0 || !!submissionId),
     },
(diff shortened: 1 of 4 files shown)


justynoh:
`staleTime`
Frankly, I picked 10s just as an order of estimation, but maybe setting staleTime isn't even the right approach at all - pls give me comments if so! 🙏  I took reference from the (soon-to-be-removed-removed-removed-REMOVED) "your submission may fail" query which doesn't even set the staleTime, so it always refetches haha. I didn't think refetching every time was necessary for this use but we can always just refetch always.

As for why 10s, we can look at some logs for reference. Not sure if this is super useful but also i figured i can just look at it to brush up on my cloudwatch syntax anyway ¯\_(ツ)_/¯ 

Ran these queries at 11:07am on a Thursday over 3 hours, thus capturing logs from 8-11am on a weekday (our busiest traffic period).

image

The highest form got 980 submissions over 3h, which amounts to about 980/(3*60) = 5.44 submissions per minute, on average, so 10s is more than sufficient to update fast enough.

Looking from the admin perspective, we can find how many admins are interested in their storage submission metadata. In the same time period we can look at what's the first and last request for each form (duration is the time between the first and last request), and how many requests they made within that time period. 

image

The maximum number of requests made was 293 in 3 hours, but that was within the span of about an hour. The remainder all made less than 100 requests. The duration was varying, though. This is very coarse-grained data, if we really want to know, we can look deeper into those who seemed to be "online" for a long time (e.g. for almost the whole 3h period) to determine what their usage patterns are.

frontend/src/features/admin-form/responses/queries.ts
```diff
@@ -91,9 +91,10 @@ export const useFormResponses = ({
     () => getFormSubmissionsMetadata(formId, params),
     {
       refetchOnMount: true,
-      refetchOnWindowFocus: false,
-      // Data will never change.
-      staleTime: Infinity,
+      // Consider data stale after 10s, prevent admins from initiating too many
+      // fetches by continuously clicking out and in of the results tab to
+      // refetch data
+      staleTime: 10000,
```
karrui: setting staletime to `0` should get the behaviour we want: which is to refetch every time the component mounts.

further reading: https://react-query-v3.tanstack.com/guides/initial-query-data#staletime-and-initialdataupdatedat

karrui (changes requested):

frontend/src/features/admin-form/responses/queries.ts
karrui: refetching on mount is good because this means users only get the updated information when they renavigate (or refocus) to the page, and not every 10 seconds

justynoh: [71edda3](https://github.com/opengovsg/FormSG/pull/4821/commits/71edda30fc6bcecda0ea2ed430b7fc217a8de05f)

frontend/src/features/admin-form/responses/queries.ts
```diff
@@ -90,10 +90,6 @@ export const useFormResponses = ({
     adminFormResponsesKeys.metadata(formId, params),
     () => getFormSubmissionsMetadata(formId, params),
     {
-      refetchOnMount: true,
-      refetchOnWindowFocus: false,
-      // Data will never change.
-      staleTime: Infinity,
       keepPreviousData: !submissionId,
```
karrui, line 94: ensure the correct behaviour happens yea

karrui (approved):

karrui:
should we also update the PR description?

frontend/src/features/admin-form/responses/queries.ts
justynoh: yes sir 🫡

PR #4827, opened 2022-09-08
fix(v2): filter email recipients in settings page before mutation

### LINKED ISSUE #4731: [1] Email settings: new email recipients disappear when admin clicks away
**To Reproduce**
In Settings: Emails where responses are sent (email mode) -
1. Add a valid email, and immediate add an invalid format email after that (e.g. abc@sss). Click away.
2. Both the newly added valid email and invalid format email disappear. 

**Expected behavior**
Newly added valid email should remain, and only the invalid format email should disappear

justynoh:
## Problem
Currently, email recipient settings has some unintuitive behavior. This PR fixes that.

Closes #4731 

In addition to the above, I also discovered a bug where an onBlur with invalid input reset the field to its original state (all the way at the beginning of the first render), caused by `reset` being called with no input. This is because `useForm` defaultValues gets cached in the first render, so it doesn't update on future renders. 

image

## Solution
For consistency with server, call reset in useEffect so that defaultValues get updated on every render. 

Also, deal with bad input onBlur by filtering the input before submitting. Updated error message for invalid emails to make this behavior known to users

**Breaking Changes** 

https://user-images.githubusercontent.com/25571626/189048871-3ad0e78e-2e28-4604-9b4f-4de5cbf3c58e.mov

### DIFF
--- a/frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
@@ -32,14 +32,20 @@ export const EmailFormSection = ({
 
   const {
     formState: { errors },
+    reset,
   } = formMethods
 
   const { mutateFormEmails } = useMutateFormSettings()
 
-  const handleSubmitEmails = ({ emails }: { emails: string[] }) => {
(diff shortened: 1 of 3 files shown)


karrui (approved):
lgtm

Issue #4757, opened 2022-08-29
Copy change for collaborators error message


mantariksh:
Change to "You don't have permission to modify this form."
<img width="623" alt="Screenshot 2022-08-29 at 6 23 17 PM" src="https://user-images.githubusercontent.com/29480346/187180491-5b2979f8-7b1c-4321-9526-d5f635162a13.png">

Issue #4746, opened 2022-08-29
Webhook retries toggle success/failure copy update


mantariksh:
Webhook retries have been enabled.
Webhook retries have been disabled.

Issue #4736, opened 2022-08-29
[0.5] Remove warning toast for respondent when form is out of date


mantariksh:
Remove this feature, we will address the problem post-rollout

mantariksh:
proposal:
1. Push this until after we implement save draft
2. Only show this when we are sure submission will fail, i.e. there are new form fields (I think this is the only case?)

mantariksh:
remove this feature for React rollout, will link more details on how to address this problem

mantariksh:
the issue title, description and first comment all say the same thing

PR #4829, opened 2022-09-08
fix(v2): improve copy for collaborator error and webhook retry toasts, remove public form desync

justynoh:
## Problem
1. #4736 
2. #4746 
3. #4757 

## Solution
Change the toast contents as appropriate. 

**Breaking Changes** 

## Screenshots

image

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/utils/getMutationErrorMessage.ts
+++ b/frontend/src/features/admin-form/create/builder-and-design/utils/getMutationErrorMessage.ts
@@ -3,11 +3,13 @@ import { StatusCodes } from 'http-status-codes'
 import { HttpError } from '~services/ApiService'
 
 export const getMutationErrorMessage = (error: Error): string => {
-  if (!(error instanceof HttpError)) return error.message
-  switch (error.code) {
-    case StatusCodes.FORBIDDEN:
-      return 'You do not have permission to modify this form.'
-    default:
-      return error.message
+  if (error instanceof HttpError) {
+    switch (error.code) {
+      case StatusCodes.FORBIDDEN:
+        return 'You do not have permission to modify this form.'
+      default:
+        break
+    }
   }
+  return error.message
 }
--- a/frontend/src/features/admin-form/create/builder-and-design/utils/getMutationErrorMessage.ts
+++ b/frontend/src/features/admin-form/create/builder-and-design/utils/getMutationErrorMessage.ts
@@ -1,0 +1,13 @@
+import { StatusCodes } from 'http-status-codes'
+
+import { HttpError } from '~services/ApiService'
+
+export const getMutationErrorMessage = (error: Error): string => {
+  if (!(error instanceof HttpError)) return error.message
+  switch (error.code) {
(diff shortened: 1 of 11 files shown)


frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -147,27 +145,9 @@ export const PublicFormProvider = ({
...
-              <Link href={window.location.href}>Refresh</Link> for the latest
-              version of the form.
-            </Text>
-          ),
-          duration: null,
-        })
-      }
+    if (data && !cachedDto) {
+      trackVisitPublicForm(data.form)
+      setCachedDto(data)
```
karrui: if we do not want to track changes, we might not even need to cache anything and just need to set staleTime=Infinity in the query.

frontend/src/features/admin-form/create/builder-and-design/utils/getMutationErrorMessage.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { HttpError } from '~services/ApiService'
+
+export const getMutationErrorMessage = (error: Error): string => {
+  if (!(error instanceof HttpError)) return error.message
+  switch (error.code) {
+    case StatusCodes.FORBIDDEN:
+      return 'You do not have permission to modify this form.'
+    default:
+      return error.message
```
wanlingt: nit: feel like it'd be easier to understand without the negation, like in [admin-form/common/mutation](https://github.com/opengovsg/FormSG/blob/form-v2/develop/frontend/src/features/admin-form/common/mutations.ts#L134-L153)
```suggestion
  if (error instanceof HttpError) {
    switch (error.code) {
      case StatusCodes.FORBIDDEN:
        return 'You do not have permission to modify this form.'
      default:
        return error.message
 } 
 return error.message
```

frontend/src/features/public-form/PublicFormProvider.tsx
justynoh: oh yes, that's true. Did we decide to dynamically update preview form also? or just static as well? should we remove the cachedDto for preview forms too?

karrui: yup, lets remove it for preview too until we can revisit it.

frontend/src/features/admin-form/create/builder-and-design/utils/getMutationErrorMessage.ts
justynoh: [90904e8](https://github.com/opengovsg/FormSG/pull/4829/commits/90904e822d5db8485d7585a35227795b309648d4)

frontend/src/features/public-form/PublicFormProvider.tsx
justynoh: [90904e8](https://github.com/opengovsg/FormSG/pull/4829/commits/90904e822d5db8485d7585a35227795b309648d4)

karrui (approved):
lgtm

PR #4832, opened 2022-09-08
fix: non-existent dates and open ended date range

### LINKED ISSUE #4706: [2] bug: Custom date validation does not allow open ended range, better error message
### 31st dates are allowed for months that do not have it. 
If you try to create those, it will say “Max date cannot be less than min date”. Error message should be changed.

### Open ended date ranges are not currently allowed

tshuli:
## Problem
- Closes #4706 

- Problem 1: 31st dates are allowed for months that do not have it (or 29/30/31 allowed for non-leap year february too). This occurs only if you manually key in the date to the input.

  Solution: Currently, date input field already disallows invalid dates such as '31/13/2022' from being keyed in. Extends the existing functionality to disallow non-existent dates (e.g. 31/09/2022) from being keyed in as well. 

  Note: The ticket suggested that the error message displayed was wrong and we should change it. What actually happens is that when a non-existent date is keyed in to the input field, the value stored is `null`, and hence the error message corresponds to the case when no date is entered. 

- Problem 2: Open ended date ranges are not currently allowed

  Solution: Fixed code logic to allow open ended date range.

### DIFF
--- a/frontend/src/components/DatePicker/DateInput.tsx
+++ b/frontend/src/components/DatePicker/DateInput.tsx
@@ -1,4 +1,10 @@
-import { KeyboardEventHandler, useCallback, useMemo, useRef } from 'react'
+import {
+  ChangeEvent,
+  KeyboardEventHandler,
+  useCallback,
+  useMemo,
+  useRef,
+} from 'react'
 import FocusLock from 'react-focus-lock'
 import {
   Flex,
@@ -13,7 +19,7 @@ import {
   useFormControlProps,
 } from '@chakra-ui/react'
 import { ComponentWithAs, forwardRef } from '@chakra-ui/system'
-import { format } from 'date-fns'
+import { format, isValid, parse } from 'date-fns'
 
 import { BxCalendar } from '~assets/icons'
 import IconButton from '~components/IconButton'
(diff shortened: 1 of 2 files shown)


frontend/src/components/DatePicker/DateInput.tsx
```diff
@@ -65,6 +71,14 @@ export const DateInput = forwardRef<DateInputProps, 'input'>(
       [onChange],
     )
 
+    /**
+     * Disallow manual input of non-existant dates (e.g. 31 Sep 2022)
+     */
+    const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+      if (isValid(parse(e.target.value, 'yyyy-MM-dd', new Date())))
```
karrui, line 78: i don't think this will work; this may prevent users from changing the dates entirely, or typing a date manually i think. Since the start of a date will always be invalid

karrui (changes requested):

frontend/src/components/DatePicker/DateInput.tsx
karrui: as discussed, date inputs behave a little differently... will probably need some changing later when the date input changes to a text one, but for now it's ok

karrui (approved):
lgtm,

PR #4833, opened 2022-09-08
feat(form-v2): copy paste emails into tags

### LINKED ISSUE #4828: Support copy and pasting emails into email recipients tag field
Currently, pasting a list of comma-separated emails in the email recipients settings and pressing "Enter" will result in the entire list being amalgamated into a single (invalid) tag. We don't support this feature now, but I could imagine situations where admins may want to copy and paste a list of people who they want to add as recipients for multiple forms.

https://user-images.githubusercontent.com/25571626/189051832-8ad98afb-263b-4e7b-871d-fa018f77f9ce.mov

tohhengxing:
## Problem

hi justyn! here's a PR for your request

Closes #4828 

## Solution

rules for email string is no comma, so i split the input value by comma and pass it into the addTag handler function.
case 1: single email -> no issue because split will default to array
case 2: string with commas -> becomes array of emails

**Breaking Changes** 

    - Details ...

## Before & After Screenshots

**AFTER**:

<details>

https://user-images.githubusercontent.com/102740235/189140916-3000a347-a4ef-49ce-be5d-4f856db29c3f.mp4

</details>


justynoh:
Thanks for this @tohhengxing! We'll take a look at it, in the meantime I've opened the workflows

karrui (approved):
lgtm. as a stretch goal (can be in new PR), could we add warning/error state to tags if they are a duplicate? so as to fit the error (and what happens during invalid emails)

frontend/src/components/TagInput/TagInput.tsx
```diff
@@ -91,8 +91,7 @@ export const TagInput = forwardRef<TagInputProps, 'input'>(
       (event: SyntheticEvent, tag: string) => {
         if (event.isDefaultPrevented()) return
         if (preventDuplicates && value.includes(tag)) return
```
karrui: actually, might need to do some error handling here so tags are not duplicated when they are split

karrui (changes requested):

tohhengxing:
hi @karrui, i'll work on the changes requested and the stretch goal this weekend! Apologies can't work on it during weekday due to academic commitments. Hope this isn't an issue 🙏.

frontend/src/components/TagInput/TagInput.tsx
tohhengxing: > so tags are not duplicated when they are split

there is a preventDuplicates prop.

if preventDuplicates is true, then which behaviour do you mean?

1) if the new tags contain duplicates of the existing tags -> don't add any tag at all? (this follows existing behaviour)
or
2) if the new tags contain duplicates of the existing tags -> remove duplicate tags from new tags and add?

karrui: option 2 looks better

tohhengxing: https://user-images.githubusercontent.com/102740235/191047376-47c969d9-c9da-49e8-8a4e-e1b19570ff64.mp4

frontend/src/components/TagInput/TagInput.tsx
```diff
@@ -90,8 +90,17 @@ export const TagInput = forwardRef<TagInputProps, 'input'>(
...
         if (event.isDefaultPrevented()) return
-        if (preventDuplicates && value.includes(tag)) return
-        onChange(value.concat(tag.split(',')))
+        if (preventDuplicates) {
+          if (value.includes(tag)) return
+          const existingTags = new Set(value)
+          const newTags = new Set(tag.split(','))
+          const uniqueTags = [...newTags].filter(
+            (tag) => !existingTags.has(tag),
+          )
```
karrui: nit: might not need to filter, we can just add new values to the set and create a list at the end instead of filtering? might be cleaner. ordering will not be lost as sets are ordered in javascript.

```ts
onChange(Array.from(new Set([...value, ...tag.split(',')]))
```

karrui (approved):
lgtm

frontend/src/components/TagInput/TagInput.tsx
tohhengxing: >ordering will not be lost as sets are ordered in javascript

Wow i didn't know that! Thank you!

Done, as requested

karrui: Note that JavaScript sets are ordered in insertion order, not lexical, fyi, hehe

karrui (approved):

frontend/src/components/TagInput/TagInput.tsx
tohhengxing: swee, thanks!

PR #4835, opened 2022-09-09
fix(v2): misc ui fixes when drawer is narrow

darrelhong:
## Problem
UI bugs that occur when both drawer and content are open and drawer are open and widths are narrow

## Solution
See commits, mostly using responsive breakpoints

## Before & After Screenshots

**BEFORE**:
Screen Shot 2022-09-08 at 11 11 44 PM

**AFTER**:
Screen Shot 2022-09-08 at 11 11 25 PM

**BEFORE**:
Screen Shot 2022-09-09 at 10 48 47 AM

**AFTER**:
Screen Shot 2022-09-09 at 10 48 39 AM

**BEFORE**:
Screen Shot 2022-09-09 at 1 28 20 PM

**AFTER**:
Screen Shot 2022-09-09 at 1 28 38 PM

**BEFORE**:
Screen Shot 2022-09-09 at 11 39 39 AM

**AFTER**:
Screen Shot 2022-09-09 at 11 38 27 AM

**BEFORE**:
Screen Shot 2022-09-09 at 1 08 08 PM

**AFTER**:
Screen Shot 2022-09-09 at 1 07 58 PM

PR #4836, opened 2022-09-09
fix(v2): responsive layouts based on width of drawer

darrelhong:
## Problem
breakpoints might not be flexible enough as the current layout allows a split drawer and content up to minimum width of 768px which means each column can get very narrow. Since breakpoints are based on the width of the entire page, it can't account for the split panels. This mainly affects the custom date range input in form builder.

## Solution
Most scenarios can be satisfied by simply setting breakpoints at `lg` instead of `md`, as per #4835. This PR adds a hook that watches the width of the drawer container then layouts can be changed at any arbitrary width.

Pros:
More flexible

Cons:
Not the cleanest code

Room for improvement:
hook could probably be optimised to give it more structure as opposed to returning the raw width?

## Before & After Screenshots

**BEFORE**
Screen Shot 2022-09-09 at 1 47 06 PM

**WITH BREAKPOINTS**
Screen Shot 2022-09-09 at 1 46 23 PM

**WITH ARBITRARY WIDTH**
Screen Shot 2022-09-09 at 1 47 49 PM

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/BuilderAndDesignDrawer.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/BuilderAndDesignDrawer.tsx
@@ -22,7 +22,7 @@ import { FieldListDrawer } from './FieldListDrawer'
 
 export const BuilderAndDesignDrawer = (): JSX.Element | null => {
   const isMobile = useIsMobile()
-  const { activeTab, isDrawerOpen } = useCreatePageSidebar()
+  const { activeTab, isDrawerOpen, drawerRef } = useCreatePageSidebar()
   const createOrEditData = useFieldBuilderStore(stateDataSelector)
 
   const drawerMotionProps = useMemo(() => {
@@ -78,7 +78,7 @@ export const BuilderAndDesignDrawer = (): JSX.Element | null => {
           overflow="hidden"
           {...drawerMotionProps}
         >
-          <Flex w="100%" h="100%" flexDir="column">
+          <Flex w="100%" h="100%" flexDir="column" ref={drawerRef}>
             {renderDrawerContent}
           </Flex>
         </MotionBox>
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
@@ -23,6 +23,8 @@ import Input from '~components/Input'
 import Textarea from '~components/Textarea'
 import Toggle from '~components/Toggle'
 
+import { useCreatePageSidebar } from '~features/admin-form/create/common'
+
 import { DrawerContentContainer } from '../common/DrawerContentContainer'
 import { FormFieldDrawerActions } from '../common/FormFieldDrawerActions'
 import { EditFieldProps } from '../common/types'
@@ -31,6 +33,8 @@ import { useEditFieldForm } from '../common/useEditFieldForm'
 type EditDateProps = EditFieldProps<DateFieldBase>
 
 const EDIT_DATE_FIELD_KEYS = ['title', 'description', 'required'] as const
+
+const MIN_WIDTH_FOR_2_COL = 405
 
 type EditDateInputs = Pick<
   DateFieldBase,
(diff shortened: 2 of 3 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
```diff
@@ -194,20 +200,24 @@ export const EditDate = ({ field }: EditDateProps): JSX.Element => {
...
-              />
-              <Controller
-                control={control}
-                rules={{
-                  deps: ['dateValidation.customMinDate'],
-                }}
-                name="dateValidation.customMaxDate"
-                render={({ field }) => <DateInput {...field} />}
-              />
+              <Box gridColumn={drawerWidth < 405 ? '1/3' : undefined}>
```
karrui: should we extract the magic number (`405`) into a constant in this file so future readers will know what it represents? Could we also add comments to the necessity of controlling the column by drawer widths?

frontend/src/hooks/useDrawerWidth.ts
```diff
@@ -0,0 +1,20 @@
+import { RefObject, useLayoutEffect, useState } from 'react'
+
+export const useDrawerWidth = (drawerRef: RefObject<HTMLDivElement>) => {
```
karrui: should we add documentation for the purpose of this hook?

karrui (approved):
hook is fine, approach makes sense. thanks, lgtm.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
darrelhong: updated

frontend/src/hooks/useDrawerWidth.ts
darrelhong: decided not to write own hook and use useMeasure form react-use instead, it should be much clearer now

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/BuilderAndDesignDrawer.tsx
```diff
@@ -25,6 +25,9 @@ export const BuilderAndDesignDrawer = (): JSX.Element | null => {
   const { activeTab, isDrawerOpen } = useCreatePageSidebar()
   const createOrEditData = useFieldBuilderStore(stateDataSelector)
 
+  const { useMeasureDrawer } = useCreatePageSidebar()
+  const [drawerRef] = useMeasureDrawer
+
```
timotheeg: This can probably be merged into line 25 instead:

```typescript
const { activeTab, isDrawerOpen, useMeasureDrawer: [drawerRef] } = useCreatePageSidebar()
```

timotheeg (commented):
Looking good, one minor comment to save on an unnecessary function call.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/BuilderAndDesignDrawer.tsx
darrelhong: tunnel vision got me, fixed. kept it as 2 lines as its a tad more readable imo

frontend/src/features/admin-form/create/common/CreatePageSidebarContext/CreatePageSidebarContext.tsx
```diff
@@ -151,6 +154,8 @@ export const useCreatePageSidebarContext =
       setPendingTab(undefined)
     }, [isMobile, pendingTab, setFieldsToInactive])
 
+    const useMeasureDrawer = useMeasure()
```
karrui: nit: we should not prefix non-hooks with `use` keyword, causes unnecessary confusion. should we call this `drawerDimensions` instead?

karrui: seems like we only need the drawer width too; suggest just returning `drawerWidth` from this hook instead of the entire dimensions.

darrelhong: agree on both points

karrui (approved):
lgtm

PR #4837, opened 2022-09-11
fix(v2): reset emails input using state to prevent erratic behavior

### LINKED ISSUE #4765: Email settings: Erratic Behavior on Email Deletion
**To Reproduce**

Create email mode form. Add additional email in email settings (i.e. total of 2 emails). Exit to admin dashboard. Enter form settings again. Attempt to delete both emails. After clicking delete on second email, both emails should appear again. Subsequent attempts to delete behaves erratically (see gif).
_(note: this bug does not occur if you do not exit to admin dashboard after adding the second email)_

Recording 2022-08-30 at 09 51 48

**Expected behavior**

When clicking delete on the last remaining email, nothing should happen. Second email should not appear again.

darrelhong:
## Problem
Emails input resets to defaultState when the one remaining email is removed. DefaultState is taken when useForm is initiated so it is whatever the original list of emails was.

Closes #4765

## Solution
Reset email input using values from store instead. Also added a error toast as previously no indication of any error if 'X' button is clicked last remaining email. 

**AFTER**:
sr

## For consideration
Currently the api calls only happen onBlur however, it might be more natural to update after pressing 'Enter' key.

There is different behaviour when clicking 'X' vs pressing backspace so that might want to be improved upon. Clicking 'X' will not allow last email to be removed, whereas backspace will allow last email to be removed, but will be reset to original state onBlur. (see above video)

### DIFF
--- a/frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
@@ -5,7 +5,7 @@ import {
   useForm,
   useFormContext,
 } from 'react-hook-form'
-import { FormControl } from '@chakra-ui/react'
+import { FormControl, useToast } from '@chakra-ui/react'
 import { get, isEmpty, isEqual } from 'lodash'
 import isEmail from 'validator/lib/isEmail'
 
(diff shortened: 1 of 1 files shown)


karrui:
Apologies, the issue may be a duplicate. May be fixed by https://github.com/opengovsg/FormSG/pull/4827. I will review this if it turns out not to be the same issue

darrelhong:
woops, looks like it

karrui:
duplicate of #4827. thanks for contributing nonetheless!

PR #4838, opened 2022-09-11
fix(v2): page scrolls completely to bottom on mobile

### LINKED ISSUE #4767: Can't scroll to the end in Admin builder preview
Thank you button gets cut off

https://user-images.githubusercontent.com/28617936/187331505-0b21836b-b3f1-4e15-b7eb-79ab67ab847b.mp4

darrelhong:
## Problem
page does not scroll all the way to the bottom on mobile browsers on first scroll. Issue is better described in these articles.
[https://dev.to/nirazanbasnet/dont-use-100vh-for-mobile-responsive-3o97](https://dev.to/nirazanbasnet/dont-use-100vh-for-mobile-responsive-3o97)
[https://chanind.github.io/javascript/2019/09/28/avoid-100vh-on-mobile-web.html](https://chanind.github.io/javascript/2019/09/28/avoid-100vh-on-mobile-web.html)
[https://stackoverflow.com/questions/52848856/100vh-height-when-address-bar-is-shown-chrome-mobile](https://stackoverflow.com/questions/52848856/100vh-height-when-address-bar-is-shown-chrome-mobile)

Closes #4767

## Solution
simplest solution seems to be replacing 100vh with 100% instead.

**AFTER**
iosgif
#
android

## Considerations
One side effect is now the address bar on mobile browsers will not scroll away, limiting screen height for content. Removing height property completely will fix this issue but position: sticky for headers will stop working. To get both sticky headers, and address bar to scroll away, an option is to set fixed header height and equal padding: top to the content. However, as the headers are currently not fixed in height, this is not so straightforward as well.

### DIFF
--- a/frontend/src/features/admin-form/common/AdminFormLayout.tsx
+++ b/frontend/src/features/admin-form/common/AdminFormLayout.tsx
@@ -47,7 +47,7 @@ export const AdminFormLayout = (): JSX.Element => {
   }
 
   return (
-    <Flex flexDir="column" height="100vh" overflow="hidden" pos="relative">
+    <Flex flexDir="column" height="100%" overflow="hidden" pos="relative">
       {bannerProps ? (
         <Banner variant={bannerProps.variant}>{bannerProps.msg}</Banner>
       ) : null}
--- a/frontend/src/features/user/billing/BillingPage.tsx
+++ b/frontend/src/features/user/billing/BillingPage.tsx
@@ -22,7 +22,7 @@ export const BillingPage = (): JSX.Element => {
     setEsrvcId(esrvcId?.trim())
 
   return (
-    <Flex direction="column" h="100vh">
+    <Flex direction="column" h="100%">
       <AdminNavBar />
       <Container
         overflowY="auto"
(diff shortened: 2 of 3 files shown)


frontend/src/features/admin-form/common/AdminFormLayout.tsx
```diff
@@ -47,7 +47,7 @@ export const AdminFormLayout = (): JSX.Element => {
   }
 
   return (
-    <Flex flexDir="column" height="100vh" overflow="hidden" pos="relative">
+    <Flex flexDir="column" height="100%" overflow="hidden" pos="relative">
```
karrui, line 50: Should we set `flex=1` (and make update the parent so `flex` works)? This is so this layout component still fills the screen instead of showing white bottom padding when the content is below full viewport height. That was the original intent of setting 100vh.

darrelhong: do you have a screenshot of when this happens, the layout component should still fill the screen even when content is short. if not a maybe min-height: 100vh might do the trick

karrui:
Thanks for attempting this issue! I've opened a PR to (hopefully) fix this issue with a more robust fix for all `vh` usages.

https://github.com/opengovsg/FormSG/pull/4860

PR #4840, opened 2022-09-12
fix(v2): set default behavior for NumberInput to not focus on clicking steppers

### LINKED ISSUE #4753: [2] Table field requires double clicks to add more rows
To replicate:
1. Enable "Add more rows"
2. Reduce minimum number of rows
3. Now "Add more rows" toggle needs two clicks to toggle

justynoh:
## Problem
After clicking the `- | +` steppers for minimum rows in the EditTable drawer, the number input gets highlighted, causing the associated Toggle for allowing respondents to add more rows to not work. There is currently a bug with Chakra which causes toggles to not work when something on screen is highlighted, as pointed out in the issue.

Closes #4753 

## Solution
Avoid the issue entirely by setting `focusInputOnChange` to `false`, which controls if the input is focused when the stepper is clicked. This is set as a default by changing the `NumberInput` component directly, so that the behavior of not focusing on click is consistent throughout the app. 

image

**Breaking Changes** 

### DIFF
--- a/frontend/src/components/NumberInput/NumberInput.tsx
+++ b/frontend/src/components/NumberInput/NumberInput.tsx
@@ -29,7 +29,6 @@ export interface NumberInputProps extends ChakraNumberInputProps {
    * Whether to show the increment and decrement steppers. Defaults to true.
    */
   showSteppers?: boolean
-
   /**
    * Color scheme of number input.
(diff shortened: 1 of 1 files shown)


karrui (approved):
lgtm

PR #4849, opened 2022-09-13
fix(form-v2): change toast copy for storage mode form activation

### LINKED ISSUE #4846: Change toast copy for storage mode form activation
When activating storage mode form, toast says `Your form is now open. If you expect a large number of responses, AutoArchive your mailbox to avoid losing any of them.`

As storage mode has no email, change to `Your form is now open. Store your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`

tohhengxing:
## Problem

change toast message to `Your form is now open. Store your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`

Closes #4846 

## Before & After Screenshots

**AFTER**:

image

### DIFF
--- a/frontend/src/features/admin-form/settings/mutations.ts
+++ b/frontend/src/features/admin-form/settings/mutations.ts
@@ -95,11 +95,14 @@ export const useMutateFormSettings = () => {
       onSuccess: (newData) => {
         // Show toast on success.
         const isNowPublic = newData.status === FormStatus.Public
-        const toastStatusMessage = isNowPublic
-          ? newData.responseMode === FormResponseMode.Encrypt
+        const toastStatusPublicMessage =
+          newData.responseMode === FormResponseMode.Encrypt
             ? `Your form is now open.\n\nStore your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`
             : `Your form is now open.\n\nIf you expect a large number of responses,  [AutoArchive your mailbox](https://go.gov.sg/form-prevent-bounce) to avoid losing any of them.`
-          : 'Your form is closed to new responses.'
+        const toastStatusClosedMessage = 'Your form is closed to new responses.'
+        const toastStatusMessage = isNowPublic
+          ? toastStatusPublicMessage
+          : toastStatusClosedMessage
 
         handleSuccess({ newData, toastDescription: toastStatusMessage })
       },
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/settings/mutations.ts
```diff
@@ -95,7 +95,7 @@ export const useMutateFormSettings = () => {
         // Show toast on success.
         const isNowPublic = newData.status === FormStatus.Public
         const toastStatusMessage = isNowPublic
-          ? `Your form is now open.\n\nIf you expect a large number of responses,  [AutoArchive your mailbox](https://go.gov.sg/form-prevent-bounce) to avoid losing any of them.`
+          ? `Your form is now open.\n\nStore your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`
```
justynoh: Apologies if the issue was not written clearly! What we want is actually for the toast to be the original (autoarchive mailbox) message if the form is in email mode, and the new secret key message if the form is in storage mode. 

You'll need to add a differentiation between the two cases. This change would replace the toast intended for email mode forms.

justynoh (changes requested):

frontend/src/features/admin-form/settings/mutations.ts
tohhengxing: ok no problem, i will resolve now

frontend/src/features/admin-form/settings/mutations.ts
```diff
@@ -95,7 +96,9 @@ export const useMutateFormSettings = () => {
         // Show toast on success.
         const isNowPublic = newData.status === FormStatus.Public
         const toastStatusMessage = isNowPublic
-          ? `Your form is now open.\n\nIf you expect a large number of responses,  [AutoArchive your mailbox](https://go.gov.sg/form-prevent-bounce) to avoid losing any of them.`
+          ? newData.responseMode === FormResponseMode.Encrypt
+            ? `Your form is now open.\n\nStore your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`
+            : `Your form is now open.\n\nIf you expect a large number of responses,  [AutoArchive your mailbox](https://go.gov.sg/form-prevent-bounce) to avoid losing any of them.`
```
justynoh: nit: should we separate out this computation for readability?

justynoh (approved):
lgtm, thanks for the contribution!

frontend/src/features/admin-form/settings/mutations.ts
tohhengxing: ok done

Issue #4177, opened 2022-07-14
[2] a11y: required fields are reporting as "required invalid data" even when there are no errors yet


karrui:
Seems to be a quirk of screen readers (maybe don't flame me if i'm wrong), known issues from quick Googling.

One fix could be to set `aria-invalid=false` when there are no errors explicitly: https://stackoverflow.com/a/72463036

We should experiment with a bunch of aria props so we can check which approach is best.

Further reading: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid

justynoh:
This may not be the universally correct approach - refer to this set of reference forms to see why

https://russmaxdesign.github.io/accessible-forms/index.html#required

 It seems that setting aria-invalid doesn't really cover everything. Instead, the approach suggested here is to ignore the `required` attribute completely, and use `aria-required` instead along with backend validation before form submission (which we already have with rhf, anyway, so we don't really need the `required` attribute, right? - correct me if i'm wrong)

The downside to this approach is that we provide `isRequired` to chakra `use*(props)`, which return css props that have the `required` property, so we will have to override that with `required = false` universally somehow...

karrui:
closed by #4916

PR #4851, opened 2022-09-13
fix(a11y): yes/no fields, formlabels

### LINKED ISSUE #4176: a11y: yes/no fields are too verbose
should just say radio group, yes option, no option

justynoh:
## Problem
Yes/No fields currently are too verbose, as they used to say "No, selected, required invalid data, radio button, 1 of 2". Then, tabbing to the next option actually goes to "No, selected" before the yes option is read.

Also, some misc a11y fixes for the start page of the form (everything above form fields).

Closes #4176, proposes solution for #4177

## Solution
`FormLabel` - add "(required)" visually hidden component for a11y

Yes/No field - refactor component to better suit its purpose (expose only `leftIcon` and `label` props rather than generic react `children`; this is also better for a11y to easily strip away the additional cosmetic icon)

**Breaking Changes** 

## Screenshots

**Notice, importantly, that forcing `required={false} aria-required={false}` does not change the behavior of the form.**

Before:

https://user-images.githubusercontent.com/25571626/189828059-410433d4-bd73-456d-b10a-0aa9097210f2.mov

After:

https://user-images.githubusercontent.com/25571626/189827455-c7fb3840-3511-4f0f-8899-30e94ac7d45c.mov

### DIFF
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -3,14 +3,11 @@ import { BiCheck, BiX } from 'react-icons/bi'
 import {
   forwardRef,
   HStack,
-  Icon,
   useFormControlProps,
-  useMultiStyleConfig,
   useRadioGroup,
 } from '@chakra-ui/react'
 import pick from 'lodash/pick'
 
-import { YESNO_THEME_KEY } from '~theme/components/Field/YesNo'
 import { FieldColorScheme } from '~theme/foundations/colours'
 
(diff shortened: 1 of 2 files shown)


justynoh:
After discussion with pearly, reflected that it may be a better option for radio fields to aria-required on the radio group, and remove required on the individual radio inputs. can revisit this later.

justynoh:
closed to overwrite v2/fix/a11y-yesno branch. will reopen another PR for it, and split up with the "required invalid data" fixes

PR #4857, opened 2022-09-14
fix(a11y): improve section and form instructions focus a11y

justynoh:
## Problem
Found some small AFIs in sections and form instructions a11y.

1. Section fields navigation was inconsistent when scrolling through the form and navigating to sections. In particular when users scroll through the form, they get to the section and it says "heading level 2, 2 items, <section name> <section description>". However when navigating, it only says "Navigated to <section name> section". We should update so that the description is also read out.
2. Form instructions did not have a11y parity with the sections (it didn't say "navigated to instructions section").

## Solution
1. Change `navigatedSectionTitle` to `navigatedSectionId` to get the title and description so it can be read out in the `VisuallyHidden` on click.
2. Refactor out instructions section id into a constant, and add support for instructions to be read out after navigation click.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
+++ b/frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
@@ -18,13 +18,15 @@ import { useIsMobile } from '~hooks/useIsMobile'
 import { usePublicFormContext } from '~features/public-form/PublicFormContext'
 
 import { useFormSections } from '../FormFields/FormSectionsContext'
+import { PUBLICFORM_INSTRUCTIONS_SECTIONID } from '../FormInstructions/FormInstructionsContainer'
 
 import { SidebarLink } from './SidebarLink'
 
 export const SectionSidebar = (): JSX.Element => {
-  const { sectionScrollData, activeSectionId, navigatedSectionTitle } =
+  const { sectionScrollData, activeSectionId, navigatedSectionId } =
     useFormSections()
   const {
+    form,
     miniHeaderRef,
     submissionData,
     isMobileDrawerOpen,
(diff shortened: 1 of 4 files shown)


frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
```diff
@@ -42,6 +44,14 @@ export const SectionSidebar = (): JSX.Element => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [miniHeaderRef?.current?.clientHeight])
 
+  const navigatedSection = useMemo(() => {
+    if (!navigatedSectionId || !form) return
+    if (navigatedSectionId === PUBLICFORM_INSTRUCTIONS_SECTIONID) {
+      return { title: 'Instructions', description: form.startPage.paragraph }
```
timotheeg, line 50: Just checking: does this need an additional property `_id: PUBLICFORM_INSTRUCTIONS_SECTIONID` for compatibility?

timotheeg (commented):
LGTM. I just have a question for clarification

karrui (approved):

frontend/src/features/public-form/components/SectionSidebar/SectionSidebar.tsx
justynoh: its not exactly needed because `navigatedSection` doesn't have the type of form fields. we only ever use the title and description fields from `navigatedSection`. Maybe it's better to say this instead, for clarity?

```typescript
const {title: navigatedSectionTitle, description: navigatedSectionDescription } = useMemo(() => {...

justynoh: it didn't seem that it was needed, i guess ts is smart enough to know that only the title and description fields were referenced, haha

Issue #4183, opened 2022-07-14
a11y: all icons used in forms are being labeled as "image"


karrui:
We should either hide the icons from screen readers using aria-hidden or add clearer labels to the icons used.

e.g.
yes/no field checkmark/cross icons
rating field star icons

justynoh:
actually, it seems that being labelled as "image" is correct: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA24

> The key is for the author to semantically markup icon fonts with `role="img"`.

there is also no role for icons: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

I think it's sufficient to add aria-labels that suffix in "icon", so that it reads "Rating star icon, image" or something like that

PR #4894, opened 2022-09-19
feat(a11y): update aria-labels for icons used in public forms

justynoh:
## Problem
Quick fix to places where icons are used in public forms to add aria-labels with the "icon" keyword, as suggested in #4183 . This definitely does not make all the a11y for associated components perfect but it should help to make sure we have the icons covered. 

## Solution

**Breaking Changes** 

### DIFF
--- a/frontend/src/components/Tag/Tag.tsx
+++ b/frontend/src/components/Tag/Tag.tsx
@@ -39,7 +39,9 @@ export const TagRightIcon = forwardRef<TagIconProps, 'svg'>((props, ref) => {
   )
 })
 
-const TagCloseIcon = () => <Icon as={BiX} fontSize="1.5rem" />
+const TagCloseIcon = () => (
+  <Icon as={BiX} fontSize="1.5rem" aria-label="Remove option icon" />
+)
 
 export type TagCloseButtonProps = ChakraTagCloseButtonProps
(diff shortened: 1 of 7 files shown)


frontend/src/components/Tag/Tag.tsx
```diff
@@ -39,7 +39,9 @@ export const TagRightIcon = forwardRef<TagIconProps, 'svg'>((props, ref) => {
   )
 })
 
-const TagCloseIcon = () => <Icon as={BiX} fontSize="1.5rem" />
+const TagCloseIcon = () => (
+  <Icon as={BiX} fontSize="1.5rem" aria-label="Remove option icon" />
```
karrui, line 43: misnomer imho. should we use  'tag' instead, which is more generic. maybe we can allow prop overrides too

justynoh: The reason for using "option" is that in the close button component (right below this), `aria-label="Remove selected option"` is used hahah

justynoh: wondering also if "tag" is common terminology for this type of usecase, maybe it should actually read "Remove <tag value> icon"

karrui: actually ya hor, tag probably less obvious than option. option it is!

karrui (approved):

PR #4908, opened 2022-09-20
fix(a11y): improvements to radio, checkbox and form labels, refactor rating to reduce screen reader garbage

### LINKED ISSUE #4184: [2] a11y: rating field wonky aria props
rating fields have extra labels that only happen when you enter the group with a screen reader.

legend: 
* `->`: ctrl+opt+shift+right_arrow

e.g.
focus on rating field -> "1 selected", -> on image (unnamed) -> 2 of x, 3 of 5 (no more selection, no more on image)

<- will now show images, and extra text again.

actually gif will probably be faster to explain

ezgif-2-2b485a255c

justynoh:
## Problem
Rating field was saying "1 unselected, required invalid data, radio button, 1 of 5, 1 of 5" for unselected ratings, and selected ratings will say "3 selected, required selected invalid data, radio button, 3 of 5, 3 of 5", along with other stuff when scrolling through focus referring to the text and the star/heart icon. 

Key reason for this is that the Wrap component generates a list element (accounting for one of the "3 of 5"), and then radio button accounts for the other "3 of 5". Also, HTML input natively announces selected-ness, so there's no need for us to add our own check at the start. The fact that it's selected is kinda hidden by the fact that "invalid data" is read out everywhere but once that's fixed later on this should be good to go.

Additionally, radio and checkbox fields were slightly inconsistent, and form labels sounded weird. (Form label screen reader says "Question number (long pause) 4. Rating". Better to just get rid of the "question number" prefix, google forms doesn't even show question numbers anyway. Given that it's a form page i would think it's quite obvious that the prefix number is the question number.

Closes #4184 

## Solution
Refactor rating to make use of stack and hstacks instead.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/191166855-4c9962ea-4faa-4596-a6f7-d09de8a5fac0.mov

### DIFF
--- a/frontend/src/components/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/Checkbox/Checkbox.tsx
@@ -48,6 +48,7 @@ export const Checkbox = forwardRef<CheckboxProps, 'input'>(
         colorScheme={colorScheme}
         ref={ref}
         {...props}
+        aria-label={children?.toString()}
       >
         {children}
       </ChakraCheckbox>
--- a/frontend/src/components/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/Checkbox/Checkbox.tsx
@@ -48,7 +48,6 @@ export const Checkbox = forwardRef<CheckboxProps, 'input'>(
         colorScheme={colorScheme}
         ref={ref}
         {...props}
-        aria-label={children?.toString()}
       >
         {children}
       </ChakraCheckbox>
@@ -107,6 +106,7 @@ const OthersCheckbox = forwardRef<CheckboxProps, 'input'>((props, ref) => {
     <Checkbox
       ref={mergedCheckboxRef}
       __css={styles.othersCheckbox}
+      aria-label="Other"
       {...props}
       onChange={handleCheckboxChange}
     >
--- a/frontend/src/components/Field/Rating/Rating.tsx
+++ b/frontend/src/components/Field/Rating/Rating.tsx
@@ -149,7 +149,7 @@ export const Rating = forwardRef<RatingProps, 'input'>(
         >
           <Stack spacing={ratingLayout.spacing}>
             {options.map((row, i) => (
-              <HStack spacing={ratingLayout.spacing}>
+              <HStack spacing={ratingLayout.spacing} key={i}>
                 {row.map((value) => (
                   <RatingOption
                     name={name}
--- a/frontend/src/components/Field/Rating/Rating.tsx
+++ b/frontend/src/components/Field/Rating/Rating.tsx
@@ -135,15 +135,15 @@ export const Rating = forwardRef<RatingProps, 'input'>(
         rowGap="0.5rem"
         templateAreas={{
           base: "'caption' 'rating'",
-          lg: "'rating' 'caption'",
+          md: "'rating' 'caption'",
         }}
       >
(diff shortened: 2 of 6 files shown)


frontend/src/components/Checkbox/Checkbox.tsx
```diff
@@ -48,6 +48,7 @@ export const Checkbox = forwardRef<CheckboxProps, 'input'>(
         colorScheme={colorScheme}
         ref={ref}
         {...props}
+        aria-label={children?.toString()}
```
karrui: children is not necessarily a string, it could be a react node. suspect this will not work for the general case

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -140,38 +147,27 @@ export const Rating = forwardRef<RatingProps, 'input'>(
...
-                        </WrapItem>
-                      )
-                  }
-                </Fragment>
-              )
-            })}
-          </Wrap>
+          <Stack spacing={ratingLayout.spacing}>
+            {options.map((row, i) => (
+              <HStack spacing={ratingLayout.spacing} key={i}>
```
karrui: slight regression here as seen in chromatic

frontend/src/components/Checkbox/Checkbox.tsx
justynoh: oops that was my quick and dirty cheat to see if it would work and I forgot to clean it up HAHAHA let me update

justynoh: [13695af](https://github.com/opengovsg/FormSG/pull/4908/commits/13695aff6b12a2b7385011fd995876c7365c3fc1)

frontend/src/components/Field/Rating/Rating.tsx
justynoh: nooooo ok i look tmr

justynoh: [8f1e49e](https://github.com/opengovsg/FormSG/pull/4908/commits/8f1e49ec278a06d8375289a2d745646e1ef5d10d)

frontend/src/components/Field/Rating/Rating.tsx
```diff
@@ -108,70 +106,66 @@ export const Rating = forwardRef<RatingProps, 'input'>(
...
-                        <WrapItem display="contents" aria-hidden>
-                          <Box flexBasis="100%" h={ratingLayout.rowHeight} />
-                        </WrapItem>
-                      )
-                  }
-                </Fragment>
-              )
-            })}
-          </Wrap>
+          <Stack spacing={ratingLayout.spacingY}>
```
karrui: comment: quite confusing to see spacingY assigned to spacing... should we rename them to something less confusing (left as an exercise to you) since we are not spreading the layout props?

karrui (approved):
lgtm

frontend/src/components/Field/Rating/Rating.tsx
justynoh: [e621225](https://github.com/opengovsg/FormSG/pull/4908/commits/e62122558f4582b9fdf0695a88a417a3224671ab) renamed!

PR #4916, opened 2022-09-21
feat: add patch-package to remove `required` prop in useFormControl

### LINKED ISSUE #4177: [2] a11y: required fields are reporting as "required invalid data" even when there are no errors yet

karrui:
## Problem

Known issue of various screen readers to read out input as "invalid" even when there has not been any action taken by the user. This is tEcHniCaLLy cOrReCt since an empty required field is technically invalid, though it is very unintuitive to a screen-reader user.

The most correct (and convenient) way is to remove the `required` prop from the input and use `aria-required` instead, but  ChakraUI injects the `required` prop without allowances for overrides.

Thus, require monkey-patching the underlying package to remove the assignment of `required` prop.

Closes #4177

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -15907,6 +15907,12 @@
       "integrity": "sha512-NuHqBY1PB/D8xU6s/thBgOAiAP7HOYDQ32+BFZILJ8ivkUkAHQnWfn6WhL79Owj1qmUnoN/YPhktdIoucipkAQ==",
       "dev": true
     },
+    "@yarnpkg/lockfile": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz",
+      "integrity": "sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==",
+      "dev": true
+    },
     "abab": {
       "version": "2.0.5",
       "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.5.tgz",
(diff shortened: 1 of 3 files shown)


karrui:
If we have a major disagreement with using `patch-package`, we can always remove the use of `isRequired` in form control and default to a visually hidden "required" text beside the form label text.

justynoh:
does it work even if we don't override `aria-required`? I remember testing by just overriding `isRequired` and it also didn't quite work as intended, needed to override both at the same time

karrui:
> does it work even if we don't override `aria-required`? I remember testing by just overriding `isRequired` and it also didn't quite work as intended, needed to override both at the same time

Yup, screen reader will only say "required" and not "required invalid data"

wanlingt (approved):
cool learnt something new! so the next step would be to replace `isRequired` with `aria-required` right

karrui:
> cool learnt something new! so the next step would be to replace `isRequired` with `aria-required` right

Don't need, auto added by chakra, now I'm just removing required!

timotheeg:
> Thus, require monkey-patching the underlying package to remove the assignment of required prop.

😰

No objection, but I'm not a fan of monkey patching. It introduces behaviour hidden from application code, and we have to be careful with updates, even patches. 

Can we:
    1. Add some comment in the component code where we rely on the monkey patching to indicate things work thanks to this behaviour we have introduced
    2. Pin the version of chakra-ui we use to a specific version in package.json (disallow automatic patch upgrade!) . This _should_ help with stability because chakra-react itself pins `@chakra-ui/form-control` to a specific version (`1.6.0`)

karrui:
> Can we: 1. Add some comment in the component code where we rely on the monkey patching to indicate things work thanks to this behaviour we have introduced 2. Pin the version of chakra-ui we use to a specific version in package.json (disallow automatic patch upgrade!) . This _should_ help with stability because chakra-react itself pins `@chakra-ui/form-control` to a specific version (`1.6.0`)

Sure. `patch-package` also skips the patch if the package is not 1.6.0, so it's not very risky. The only behaviour is that `required` prop is set again, which is not really a big deal

timotheeg:
Makes sense! Thanks KarRui!

karrui:
> Can we: 1. Add some comment in the component code where we rely on the monkey patching to indicate things work thanks to this behaviour we have introduced

I looked around in the codebase, the usage is hidden away in Chakra's own components... not sure where I can add the comment. 😢

karrui:
just merging in first!

Issue #4189, opened 2022-07-14
[1] a11y: add affordance for screen readers when form has been submitted


karrui:
Currently, the screen changes to the submission success page without any announcements for screen readers.
Screen reader users may feel confused with the lack of success feedback.

Suggest focusing on success text when loaded so screen readers will read it out. Or add aria declaration "Your response has been submitted"

karrui:
This was actually closed by #4917.

PR #4917, opened 2022-09-21
feat(v2): read out submission confirmation for screen readers on success

### LINKED ISSUE #4179: [0.5] a11y: add extra aria text to denote the end of the form
users using screen readers could get confused when suddenly met with a submit button. we could add more accessible text to the submit button to denote that it is the end of the form.

karrui:
## Problem

Currently, screen reader users do not know when their form has been submitted. This PR adds an aria-live region where success state is being read out on submission success.

Closes ~~#4179~~ #4189

## Solution

**Breaking Changes** 


justynoh (approved):
lgtm

PR #4918, opened 2022-09-21
fix(v2): change verification copy for email and mobile

### LINKED ISSUE #4887: Clearer copy for field verification error
image

The long-term fix for this is #4618, but in the meantime, the copy can be improved to make the required action clearer to the user.
"Verify your mobile number below."
"Verify your email address below."

wanlingt:
## Problem

Field verification error is confusing. This is an **interim** fix that only changes the copy.

Closes #4887 

## Solution
Change the verification message depending on the `fieldType`

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:
image
(can also view this in storybook)

### DIFF
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -98,10 +98,10 @@ const createBaseVfnFieldValidationRules: ValidationRuleFnEmailAndMobile<
           return true
         }
         if (schema.fieldType === BasicField.Mobile) {
-          return 'Verify your mobile number below.'
+          return 'Please verify your mobile number'
         }
         if (schema.fieldType === BasicField.Email) {
-          return 'Verify your email address below.'
+          return 'Please verify your email address'
         }
       },
     },
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -9,6 +9,7 @@ import validator from 'validator'
 
 import {
   AttachmentFieldBase,
+  BasicField,
   CheckboxFieldBase,
   DateFieldBase,
   DateSelectedValidation,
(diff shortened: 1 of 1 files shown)


karrui (approved):

frontend/src/utils/fieldValidation.ts
```diff
@@ -53,10 +54,20 @@ type OmitUnusedProps<T extends FieldBase> = Omit<
   'fieldType' | 'description' | 'disabled'
 >
 
+// fieldType is only used in email and mobile field verification
+type OmitUnusedPropsEmailAndMobile<T extends FieldBase> = Omit<
+  T,
+  'description' | 'disabled'
+>
+
```
timotheeg: Why was it necessary to do this?

wanlingt: We don't want to omit `fieldType` for email and mobile as we use it to display the appropriate verification message ("Please verify your email address / mobile number"). 

I didn't modify the original `OmitUnusedProps` (which omits `fieldType`) as that applies to other fields (e.g. dropdown, etc). Changing this would require changing other functions which depend on `OmitUnusedProps`, which wouldn't make much sense since they don't need `fieldType`.

timotheeg: Ah, I see from the code above, I should have expanded (sorry!).

Omitting prop is an interesting thing to do. I expect no changes would be needed in other functions if `OmitUnusedProps` was modified, because having additional properties which are not used are inconsequential. It would matter in cases where we create objects with a minimal set of properties to match the definition, because those objects might now fail to meet the field requirements (depending on how we construct them).

By the look of the PR, we are not doing that for emails and mobile, but are we doing it for other field types?

BTW looking at the old code, `OmitUnusedProp*` is quite a terrible name for a type. Types should indicate what they represent with positive names, in this case it might be `FieldMinimalValidation' or something like that 🤔

wanlingt: > It would matter in cases where we create objects with a minimal set of properties to match the definition, because those objects might now fail to meet the field requirements (depending on how we construct them).

Yes this is right! For example, `createDropdownValidationRules` is used in [`DropdownColumnCell`](https://github.com/opengovsg/FormSG/blob/form-v2/fix-issue-4887/frontend/src/templates/Field/Table/ColumnCell.tsx#L73), where `schema` takes the form of [`ColumnDto`](https://github.com/opengovsg/FormSG/blob/form-v2/fix-issue-4887/frontend/src/templates/Field/Table/ColumnCell.tsx#L36) (which doesn't have the property `fieldType`)

> BTW looking at the old code, OmitUnusedProp* is quite a terrible name for a type. Types should indicate what they represent with positive names, in this case it might be `FieldMinimalValidation' or something like that

hahaha ok i will rename this

wanlingt:
if no more comments by tmr (28/9), will merge this!

timotheeg (approved):

PR #4922, opened 2022-09-22
fix(v2): add better a11y to dropdown fields

### LINKED ISSUE #4185: [2] a11y: dropdown field has inconsistent behavior
**Describe the bug**
some screen readers do not trigger dropdown on entering with a screen reader (maybe it doesn't count as a focus)

when leaving the field, dropdown field could stay in an opened state due to lack of onBlur, leaving them in a trapped state when the field is entered again.

Dropdown needs more descriptor, need to announce as a combobox?
“Clear dropdown input button” => rename to “clear selection button”?

karrui:
## Problem

This PR adds a bunch of a11y tweaks to hopefully make dropdown fields a better experience.

Closes #4185 

## Solution

**Breaking Changes** 

**Features**:

- feat: prevent Enter keypress from form submission (public form)
- feat: allow dropdown to be opened with screen readers by allowing focus on dropdown caret as a button
- feat(SelectCombobox): announce input cleared and refocus on input

## Before & After Screenshots
Should have no changes

### DIFF
--- a/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
+++ b/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
@@ -67,7 +67,7 @@ export const MultiSelectProvider = ({
   filter = defaultFilter,
   nothingFoundLabel = 'No matching results',
   placeholder: placeholderProp,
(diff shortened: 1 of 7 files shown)


timotheeg (approved):
lgtm 👍

PR #4925, opened 2022-09-23
fix(v2): fix radio aria-required inconsistency

### LINKED ISSUE #4190: [3] a11y: radio field aria inconsistency
Unselected options still report invalid entry (even though some other option has been selected)

Expected behaviour should be one radio field has an option selected, unchecked radio fields should no longer prompt for invalidness.

suspect due to `aria-required` prop, will require investigation.

wenhaogoh:
## Problem

Closes #4190 

## Solution

Although the automatic injection of `required` props by Chakra's `useFormControl` has been disabled by this [PR](https://github.com/opengovsg/FormSG/pull/4916), they are still injected by radio-related Chakra hooks eg. `useRadio`. Hence, the `required` prop still has to be disabled manually to produce the desired behaviour.

After some discussion with @justynoh, the desired behaviour of radio field's a11y is as such:

- Radio Group
  - aria-invalid
    - `true` when error message is displayed (either for radio group or other input)
  - aria-required
    - `true` when required option is checked

- Radio + Other Radio
  - aria-invalid
    - never `true`
  - aria-required
    - never `true`

- Other Input
  - aria-invalid
    - `true` when error message is displayed (for other input)
  - aria-required
    - `true` when other option is checked

### DIFF
--- a/frontend/src/components/Radio/Radio.tsx
+++ b/frontend/src/components/Radio/Radio.tsx
@@ -266,7 +266,12 @@ const OthersRadio = forwardRef<RadioProps, 'input'>((props, ref) => {
   }, [isChecked, othersInputRef])
 
   return (
-    <Radio ref={mergedRadioRef} {...props} __css={styles.othersRadio}>
+    <Radio
+      ref={mergedRadioRef}
+      {...props}
+      __css={styles.othersRadio}
+      isRequired={false}
+    >
       Other
     </Radio>
   )
@@ -299,6 +304,7 @@ export const OthersInput = forwardRef<InputProps, 'input'>(
         ref={mergedInputRef}
         {...props}
         onChange={handleInputChange}
+        isRequired={othersRadioRef.current?.checked}
       />
     )
   },
--- a/frontend/src/components/Radio/Radio.tsx
+++ b/frontend/src/components/Radio/Radio.tsx
@@ -215,7 +215,11 @@ export const Radio = forwardRef<RadioProps, 'input'>(
         {...labelProps}
         __css={rootStyles}
       >
-        <input className="chakra-radio__input" {...inputProps} />
+        <input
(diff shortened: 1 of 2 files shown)


frontend/src/templates/Field/Radio/RadioField.tsx
```diff
@@ -87,9 +87,18 @@ export const RadioField = ({
...
+            }
+            aria-required={schema.required}
           >
             {schema.fieldOptions.map((option, idx) => (
-              <Radio key={idx} value={option} {...(idx === 0 ? { ref } : {})}>
+              <Radio
+                key={idx}
+                value={option}
+                {...(idx === 0 ? { ref } : {})}
+                isRequired={false}
```
justynoh, line 101: Can we add a comment to explain why we set `isRequired` in places that we do this override so that we don't delete this by accident in future hehe 🙏

justynoh (approved):
lgtm

frontend/src/templates/Field/Radio/RadioField.tsx
wenhaogoh: Added!

frontend/src/components/Radio/Radio.tsx
```diff
@@ -295,6 +305,8 @@ export const OthersInput = forwardRef<InputProps, 'input'>(
         ref={mergedInputRef}
         {...props}
         onChange={handleInputChange}
+        // This is to enhance radio a11y: https://github.com/opengovsg/FormSG/pull/4925
```
karrui: we should not link to any PRs, we should add a specific comment. URLs may die

wenhaogoh: amended!

karrui (approved):
lgtm

PR #4926, opened 2022-09-25
perf(v2): narrow slice selection and remove memo

darrelhong:
## Problem
This PR should further alleviate performance issues as mentioned PR #4885 and issues #4766 #4764.

## Solution
The first commit seems to yield significant gains (>40%) by only selecting the necessary slice from `useFieldBuilderStore` `stateData`. `stateData.field` will change when editing field, causing components to update frequently. Following the suggested pattern in [zustand recipes](https://docs.pmnd.rs/zustand/recipes/recipes) to select multiple slices.

Need some help to verify if this translates into production by testing on a form with many table fields as it seems table fields are pretty render intensive. There should be a reduction in task duration when editing a field.

Currently, all fields will still rerender on input change but it seems either react/zustand is handling it much better perhaps because its doing less comparisons(?)

Second commit does not really do seem to have much impact but the additional `memo` is definitely overhead. The BuilderFields component will always rerender whenever `stateData` changes as the `fields` prop it receives will always change on input.

To verify, add a console.log to the comparison function of `memo` and it should not get printed out.

**BEFORE**
Screen Shot 2022-09-25 at 3 47 03 PM

**AFTER**
Screen Shot 2022-09-25 at 3 46 20 PM

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
@@ -1,5 +1,3 @@
-import { memo, useMemo } from 'react'
-
 import { AdminFormDto } from '~shared/types/form'
 
 import { augmentWithQuestionNo } from '~features/form/utils'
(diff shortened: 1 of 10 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
```diff
@@ -13,27 +11,24 @@ interface BuilderFieldsProps {
...
-  },
-  (prev, next) =>
-    prev.fields === next.fields && prev.isDraggingOver === next.isDraggingOver,
-)
+export const BuilderFields = ({
+  fields,
+  visibleFieldIds,
+  isDraggingOver,
+}: BuilderFieldsProps) => {
+  console.log('rerender')
```
karrui: Extraneous console log

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -409,7 +409,7 @@ type MemoFieldRowProps = {
   showMyInfoBadge?: boolean
 }
 
-const MemoFieldRow = memo(({ field, ...rest }: MemoFieldRowProps) => {
+const MemoFieldRow = ({ field, ...rest }: MemoFieldRowProps) => {
```
karrui: should we rename this if we are not memoizing this code?

karrui (approved):
lgtm with comment

karrui (changes requested):
oops should be request changes

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
darrelhong: good catch

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
darrelhong: done

karrui (approved):
lgtm!

PR #4950, opened 2022-09-27
feat(v2): add single quote to CSV output

wanlingt:
## Problem
Works towards [#133](https://github.com/datagovsg/formsg-private/issues/133)

## Solution

Use regex expressions to determine if the field value starts with a formula character. If so, prefix it with a single quote. The only exceptions to this are field values that only contain numbers, + or - e.g. +65, -1.

[Design doc here](https://docs.google.com/document/d/1M5kc5IckkhLjA9bZ_WI9gHq3WIuSUFCo4JFzuPEsB50/edit)

**Breaking Changes** 

    - Downloaded CSVs might have cells that start with an additional single quote, if any of their field values start with a formula character

## Tests

- [x] Enter a value starting with a formula character into a **storage form**'s short text, long text and table fields. Download the responses. These values should be prefixed with a single quote.
- [x]  Repeat the same for a **form's feedback**.

### DIFF
--- a/frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
+++ b/frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
@@ -41,7 +41,9 @@ export class FeedbackCsvGenerator extends CsvGenerator {
       // prefix it with a single quote to prevent formula injection
       if (!isPureNumber && hasFormulaChars) {
         feedbackComment = `'${feedback.comment}`
-      } else feedbackComment = feedback.comment
+      } else {
+        feedbackComment = feedback.comment
+      }
     }
 
     this.addLine([createdAt, feedbackComment, feedback.rating])
--- a/frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
+++ b/frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
@@ -4,10 +4,7 @@ import { formatInTimeZone } from 'date-fns-tz'
 import { FormFeedbackDto } from '~shared/types'
 
 import { CsvGenerator } from '../../common/utils/CsvGenerator'
-import {
-  FORMULA_INJECTION_REGEXP,
-  PURE_NUMBER_REGEXP,
-} from '../../ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator'
+import { processFormulaInjectionText } from '../../ResponsesPage/storage/utils/processFormulaInjection'
 /**
  * Class to encapsulate the FeedbackCsv and its attributes
  */
@@ -31,20 +28,9 @@ export class FeedbackCsvGenerator extends CsvGenerator {
       'dd MMM yyyy hh:mm:ss a',
     ) // Format in SG timezone
 
-    let feedbackComment = ''
-    if (feedback.comment) {
-      // Check if fieldRecord is a pure number
-      const isPureNumber = PURE_NUMBER_REGEXP.test(feedback.comment)
-      // Check if fieldRecord starts with formula characters
-      const hasFormulaChars = FORMULA_INJECTION_REGEXP.test(feedback.comment)
-      // if fieldRecord is not a pure number, and starts with formula characters,
-      // prefix it with a single quote to prevent formula injection
-      if (!isPureNumber && hasFormulaChars) {
-        feedbackComment = `'${feedback.comment}`
-      } else {
-        feedbackComment = feedback.comment
-      }
-    }
+    const feedbackComment = feedback.comment
+      ? processFormulaInjectionText(feedback.comment)
+      : ''
 
     this.addLine([createdAt, feedbackComment, feedback.rating])
   }
--- a/frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
+++ b/frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
@@ -4,7 +4,10 @@ import { formatInTimeZone } from 'date-fns-tz'
 import { FormFeedbackDto } from '~shared/types'
 
 import { CsvGenerator } from '../../common/utils/CsvGenerator'
-import { FORMULA_INJECTION_REGEXP } from '../../ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator'
+import {
+  FORMULA_INJECTION_REGEXP,
+  PURE_NUMBER_REGEXP,
+} from '../../ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator'
 /**
  * Class to encapsulate the FeedbackCsv and its attributes
  */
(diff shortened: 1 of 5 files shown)


karrui (approved):
lgtm, but wait for one more set of eyes pls :P. 

@amitogp @mantariksh would we need to communicate this change to users too? not sure what we decided on.

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
```diff
@@ -14,6 +14,8 @@ type UnprocessedRecord = Merge<
   { record: Dictionary<Response> }
 >
 
+export const PURE_NUMBER_REGEXP = new RegExp('^[+-][0-9]+')
+export const FORMULA_INJECTION_REGEXP = new RegExp('^[+=@-].*')
```
timotheeg: Minor comments:
* No need to use the `RegExp` constructor, unless we are constructing a pattern from parts
* Pure numbers regexp should have an end anchor as well. Otherwise in its current form, it still allows formulas like this: `+3*SUM(A1)` (in fact do add a test for an input like that please)
* Checking for `.*` at the end of a regexp can be discarded

```suggestion
export const PURE_NUMBER_REGEXP = /^[+-]\d+(\.\d+)?$/
export const FORMULA_INJECTION_REGEXP = /^[+=@-]/
```

Maybe we can simplify `PURE_NUMBER_REGEXP` to `/^[+-][\d.]+$/` ? That's not exact number regexp, but it's safe from an injection perspective.

frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
```diff
@@ -22,13 +25,27 @@ export class FeedbackCsvGenerator extends CsvGenerator {
...
+      const isPureNumber = PURE_NUMBER_REGEXP.test(feedback.comment)
+      // Check if fieldRecord starts with formula characters
+      const hasFormulaChars = FORMULA_INJECTION_REGEXP.test(feedback.comment)
+      // if fieldRecord is not a pure number, and starts with formula characters,
+      // prefix it with a single quote to prevent formula injection
+      if (!isPureNumber && hasFormulaChars) {
+        feedbackComment = `'${feedback.comment}`
+      } else {
+        feedbackComment = feedback.comment
+      }
```
justynoh: comments still say `fieldRecord` :P 

this logic is identical, maybe can extract to a util `shouldAddQuotes` (or whatever name you desire) and then just call `shouldAddQuotes(feedback.comment)` and `shouldAddQuotes(fieldRecord.getAnswer(colIndex))`

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
justynoh: thinking about whether we should check just `/^[+-.\d]*$/` for expressions like `-4+2-0` or something like that, since i guess those are formulas but i don't think they actually pose any risk (correct me if i'm wrong). It might be simpler to describe to admins too, to explain the behavior of what we add quotes to and what we don't

justynoh: thinking about whether we should check just `/^[+-.\d]*$/` for expressions like "-4+2-0" or something like that, since i guess those are formulas but i don’t think they actually pose any risk (correct me if i’m wrong). It might be simpler to describe to admins too, to explain the behavior of what we add quotes to and what we don’t

frontend/src/features/admin-form/responses/FeedbackPage/utils/FeedbackCsvGenerator.ts
wanlingt: good catch + suggestion! made the changes [8ceacfd](https://github.com/opengovsg/FormSG/pull/4950/commits/8ceacfd3e148215efa7a8d4bed55151220c33903)

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
wanlingt: @justynoh yeah I don't think they pose any risks if they are just digits and + or -. Didn't come across that being mentioned in any of the sites which talked about formula injection - the malicious commands all contained at least 1 alphabet. thanks for spotting that case!
Made the change in [858f4dc](https://github.com/opengovsg/FormSG/pull/4950/commits/858f4dc713acccb1b2d39af79f4b1db366b39d41), decided to go with `/^[+-.\d]+$/`

wanlingt: @timotheeg thanks tim! added the additional test here [c0b1511](https://github.com/opengovsg/FormSG/pull/4950/commits/c0b15116ee5752bcee930794fbe940261b52ff71)

amitogp:
@karrui @wanlingt the plan is to release it with react and announce it as a security improvement along with other react enhancements. Something like: "Enhanced security to stop malicious inputs to form responses." @pearlyong please help in adding it in our react enhancements. @kennethchangOPENGOV will need your help in adding details of this to our guide.

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
wanlingt: @timotheeg 's offline comment: `-` is a range separator, so it should only be added at the end of the regexp. otherwise it captures ASCII characters between `+` and `.`
[e0da940](https://github.com/opengovsg/FormSG/pull/4950/commits/e0da9401505615dd0dc45fb5ba717a6a6dfd88dd)

timotheeg (approved):

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
justynoh: out of curiosity, then, if they all contain at least 1 alphabet, why not just use a single regex `/^[-+@=].*[a-zA-Z]/` haha. Maybe there is a case I'm not seeing

wanlingt: @justynoh that regex wouldn't capture other symbols (e.g. `|`) and formulas like `=SUM(A1)`. also I think it's better to play it safe and not have Excel react to any formulas at all (other than ones with numbers and + | -, which we know are harmless).

wanlingt:
> 



> @karrui @wanlingt the plan is to release it with react and announce it as a security improvement along with other react enhancements. Something like: "Enhanced security to stop malicious inputs to form responses." @pearlyong please help in adding it in our react enhancements. @kennethchangOPENGOV will need your help in adding details of this to our guide.

@amitogp I've created new issues for these followup actions
@pearlyong - #4970 
@kennethchangOPENGOV - #4971

src/public/modules/forms/helpers/__tests__/FeedbackCsvGenerator.spec.ts
```diff
@@ -65,5 +65,33 @@ describe('FeedbackCsvGenerator', () => {
...
+        created: new Date('2019-11-05T13:12:14').toISOString() as DateString,
+        lastModified: new Date(
+          '2019-11-05T13:12:14',
+        ).toISOString() as DateString,
+      }
+      const insertedCreatedDate = moment(feedback.created)
+        .tz('Asia/Singapore')
+        .format('DD MMM YYYY hh:mm:ss A')
+
+      const insertedLine = `${insertedCreatedDate},'${feedback.comment},${feedback.rating}`
```
karrui, line 85: @wanlingt there's a bug here right? there should be a single quote in the expected line

karrui: or actually should not have anything here since there was no change to AngularJS's feedback csv generator.

```suggestion
      const insertedLine = `${insertedCreatedDate},${feedback.comment},${feedback.rating}`
```

wanlingt: my bad, you are right! this tests for feedback csv generator in AngularJS not React. Will have to add a separate test for React

wanlingt: fixed in your PR [8b33acc](https://github.com/opengovsg/FormSG/pull/5010/commits/8b33acc7ab4d3896b83b18548eb0982660a2e0a3)

PR #4958, opened 2022-09-27
fix(v2): reset twilio settings form with empty string on deletion of credentials

### LINKED ISSUE #4933: [2]Removing twillio credentials will not clear twillio settings fields
To reproduce: 
1. enter twillio credentials, save
2. Then, refresh the page, then delete credentials. The masked strings won’t be removed:

image

Trying to click “save” again will not work though, it will show errors for all fields because the contents of the fields is “***”.

justynoh:
## Problem
Deleting Twilio credentials does not currently clear the fields, but leaves the string `'******'` in the fields.

Closes #4933 

## Solution
This was caused by resetting the field on deletion to its default value, which was `'****'`. Instead, reset to `''` on deletion.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/192466244-f1506be8-b608-4168-94c5-a2b04e7330c4.mov


karrui (approved):
lgtm

PR #4964, opened 2022-09-28
fix(v2): extend empty workspace flexbox to fill screen

justynoh:
There was a weird break in the background color for empty workspace because the height wasn't 100vh. This PR fixes that.

Before
image

After
image

### DIFF
--- a/frontend/src/features/workspace/components/EmptyWorkspace/EmptyWorkspace.tsx
+++ b/frontend/src/features/workspace/components/EmptyWorkspace/EmptyWorkspace.tsx
@@ -22,9 +22,10 @@ export const EmptyWorkspace = ({
 
   return (
     <Flex
-      justify="center"
+      justify="flex-start"
       flexDir="column"
       align="center"
+      h="100vh"
       px="2rem"
       py="1rem"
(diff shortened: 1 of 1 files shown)


frontend/src/features/workspace/components/EmptyWorkspace/EmptyWorkspace.tsx
```diff
@@ -22,9 +22,10 @@ export const EmptyWorkspace = ({
 
   return (
     <Flex
-      justify="center"
+      justify="flex-start"
       flexDir="column"
       align="center"
+      h="100vh"
```
karrui: should we use the `css` prop instead? `100vh` SUCKS on iOS safari.

css utils are in `frontend/src/utils/fillHeightCss.ts`

justynoh: okehh [3b2368f](https://github.com/opengovsg/FormSG/pull/4964/commits/3b2368f41f136d270e6baa6b02d9a711acb16288)

karrui (approved):
lgtm

PR #4980, opened 2022-09-29
feat: bunch of table field (a11y) improvements, misc public form fixes

### LINKED ISSUE #4192: [3] a11y: add more affordances to table field
we should add aria descriptions of current row/column of focused field in the table.

e.g.  column 1, row 1

add more rows button also confusing to pure screen reader users.
Should add additional description at the start of the field after title and description:

"This is a table field. You can add more rows if you’d like, up to X rows"

karrui:
## Problem

This PR adds a bunch of table field a11y improvements. 
Also some public form submission fixes. See changelog below.

Hopefully closes #4192, but probably need more user tests.

## Solution

**Breaking Changes** 

**Features**:

- feat: add accessible table field description and label

**Bug Fixes**:

- fix(Select): use Portal for dropdown menu so Table field container does not constrain dropdown dimensions.
- fix: move onSubmit to submit button click, remove incorrect handling of "enter" keypresses on the form level
  - previous implementation of preventing enter keypresses was affecting the whole form, where enter can no longer be used to trigger buttons.
  - however, we still do want to prevent erroneous form submissions with the enter keypress, and thus should only trigger form submission on button click.
- feat: add slightly more context to table aria descriptions and when table row changes have occurred

## Before & After Screenshots
No change, all purely on the aria level.

### DIFF
--- a/frontend/src/components/Dropdown/SingleSelect/SingleSelect.stories.tsx
+++ b/frontend/src/components/Dropdown/SingleSelect/SingleSelect.stories.tsx
@@ -5,6 +5,7 @@ import { FormControl } from '@chakra-ui/react'
 import { useArgs } from '@storybook/client-api'
 import { Meta, Story } from '@storybook/react'
 
+import { fixedHeightDecorator } from '~utils/storybook'
 import Button from '~components/Button'
 import FormErrorMessage from '~components/FormControl/FormErrorMessage'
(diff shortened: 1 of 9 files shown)


wanlingt (approved):
lgtm!

justynoh (approved):
lgtm

PR #4987, opened 2022-09-30
fix: upgrade chromium to 102.0.5005.182-r0

justynoh:
## Problem
Chromium is failing again!!! 

## Solution

Upgrade to 102.0.5005.182-r0 which is the version supported by alpine v3.16 now

image

**Breaking Changes** 


karrui (approved):
lgtm

timotheeg (approved):

Issue #4752, opened 2022-08-29
Table field title and delete icon should be centre aligned (vertically)


mantariksh:
<img width="419" alt="Screenshot 2022-08-29 at 6 15 56 PM" src="https://user-images.githubusercontent.com/29480346/187179207-7a4567c5-a7a1-4d51-89b4-aeeeab593d14.png">

Issue #4658, opened 2022-08-22
Remove Twilio credentials button looks like an error message


karrui:
![image44](https://user-images.githubusercontent.com/22133008/185896015-1b24d803-fafc-4ad1-90b8-52363692692c.png)

pearlyong:
supposed to be danger button, [figma](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=5177%3A102162)

Issue #4749, opened 2022-08-29
Image field design update


mantariksh:
Image box layout is updated, see [correct padding](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=2580%3A1447)

justynoh:
seems correct currently. can check what was this about? maybe it's already fixed

karrui:
If padding seems correct then it should already be fixed. 

I cleaned up the padding in #4525 I think

PR #4990, opened 2022-09-30
fix(v2): align delete column button with label in table edit drawer, make Twilio credential delete a button

justynoh:
## Problem

1. #4752 
2. #4658

## Solution
1. Need to add a negative margin for the delete icon. However, there is an issue of the margins not working due to chakra enforcing margin-top in the `Stack` component (see this [issue](https://github.com/chakra-ui/chakra-ui/issues/2578)). Therefore, switch to grid so that our negative margins work.
2. Remove `variant='link'` from button component.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditTable/EditTableColumns.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditTable/EditTableColumns.tsx
@@ -8,6 +8,7 @@ import {
 import { BiPlus, BiTrash } from 'react-icons/bi'
 import {
   FormControl,
+  Grid,
   Stack,
   StackDivider,
   VisuallyHidden,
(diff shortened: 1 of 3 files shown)


wanlingt (commented):
lgtm, but I'm not sure if it was intentional to make the description required for a11y purposes - it's [used in the alt text](https://github.com/opengovsg/FormSG/blob/form-v2/develop/frontend/src/templates/Field/Image/ImageField.tsx#L59) 
@karrui can confirm?

karrui:
> lgtm, but I'm not sure if it was intentional to make the description required for a11y purposes

Yes, it was intentional! Description will be required (only on the client so it's backwards compatible) in react.

justynoh:
@wanlingt @karrui i'll revert that commit, do you know what the issue #4749 was about? seems like currently the same as figma leh

wanlingt (approved):

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditTable/EditTableColumns.tsx
```diff
@@ -81,10 +82,11 @@ export const EditTableColumns = ({
             isReadOnly={isLoading}
             isInvalid={!!errors?.columns?.[index]?.title}
           >
-            <Stack justify="space-between" align="center" direction="row">
+            <Grid templateColumns="1fr auto">
```
karrui, line 85: not sure if the padding will be off like this, since stack used to give everything inside 0.5rem of spacing

justynoh: image

Looks right, the changes to UI are because the bottomline for the delete button was lower than that for the "Column 1" form labels in the past when it wasn't supposed to be. Now it's aligned, so everything is moved up

PR #5019, opened 2022-10-05
perf: instrument email submission controller with datadog traces

### LINKED ISSUE #4820: [2] Instrument email submission to investigate latency
Email submissions are slow, we should investigate why. ~~May subsume #3754 if investigation shows that we can improve this with config.~~
This issue only covers the instrumentation. Based on the investigation, we will open subsequent stories to fix the latency.

justynoh:
## Problem
We are having an issue with email submissions taking a long time. This PR adds instrumentation to allow us to investigate the issue.

Closes #4820 

## Solution

From existing traces we can see that the long tail which takes a long time occurs after saving the submission metadata to the database. Add datadog traces to the remainder of the function to evaluate how long each step takes.

**Breaking Changes** 

## Screenshots

On staging, this is how the traces look like now:

image


timotheeg (approved):
👍

timotheeg (approved):

PR #5020, opened 2022-10-05
feat: add rate limit to verification endpoints too

karrui:
previously only had rate limits on auth.

Verification endpoints won't invoke the super expensive hash comparisons since the verification document will not exist in the first place, but verification endpoints also invoke some database queries, might as well rate limit those too.

Fixing some CodeQL flagged issues


justynoh (approved):
lgtm

PR #5069, opened 2022-10-06
chore(ci): remove redundant rules in mergify config

karrui:
Round 2. Some dependabot updates do not trigger success criteria for some workflows. 

This PR removes redundant and sometimes harmful rules that prevent mergify from working correctly. The removed rules are already covered by branch protection rules, which mergify respects


karrui:
Relooking at the rules, may not actually need those criteria. will remove.

wanlingt (approved):

PR #5071, opened 2022-10-06
fix: run `npm audit fix` on frontend directories

karrui:
This PR fixes the auto-fixable npm packages afforded by `npm audit fix`.


karrui:
i guess... things break due to super old stuff on root. Will only audit fix frontend directory then.

EDIT: oops it was frontend after all. Maybe we'll do this next time :P

PR #5079, opened 2022-10-09
perf(v2): use modified version of useMeasure to prevent choppy animation

darrelhong:
## Problem
realised my previous PR #4836 causes drawer animation to be choppy as constantly changing width triggers rerenders.

## Solution
modified useMeasure to debounce setState calls so it doesnt get called so often. Animations are now smooth but there still seems to be a very slight delay when opening drawer as compared to without the useMeasure hook. maybe reverting to simple breakpoints can be considered.

### DIFF
--- a/frontend/src/hooks/useMeasure.ts
+++ b/frontend/src/hooks/useMeasure.ts
@@ -1,0 +1,54 @@
+// from react-use https://github.com/streamich/react-use/blob/master/src/useMeasure.ts
+// modified to debounce setRect call prevent rerenders during animation
+import { useLayoutEffect, useMemo, useState } from 'react'
+import { debounce } from 'lodash'
+
+export type UseMeasureRect = Pick<
+  DOMRectReadOnly,
+  'x' | 'y' | 'top' | 'left' | 'right' | 'bottom' | 'height' | 'width'
+>
+export type UseMeasureRef<E extends Element = Element> = (element: E) => void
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useMeasure.ts
```diff
@@ -0,0 +1,54 @@
...
+  left: 0,
+  bottom: 0,
+  right: 0,
+}
+
+export function useMeasure<E extends Element = Element>(): UseMeasureResult<E> {
+  const [element, ref] = useState<E | null>(null)
+  const [rect, setRect] = useState<UseMeasureRect>(defaultState)
+
+  const deboucedSetRect = useMemo(() => debounce(setRect, 100), [])
```
karrui, line 31: nit: since we are adding a new hook, should we attempt to make it generalisable, i.e allow passing of debounce timeout (with a default)?

karrui (approved):
lgtm. Not a big deal anyways, ppl don't really dynamically resize their browsers

frontend/src/hooks/useMeasure.ts
darrelhong: definitely but since its use is currently limited to this one use case, i think its ok to leave it as it is. will be strightforward to add it in the future

karrui:
Reopening to trigger snapshots maybe

PR #5087, opened 2022-10-10
fix: harden ssh server

timotheeg:
## Problem
FormSG runs in Elastic BeanStalk and SSH is not opened by default. Occasionally we do ssh in for some inspection (run traceroutes, inspect EFS and shared files), this is done with access control limited to VPN.

Even though the SSH access scheme is locked down pretty well, the ssh server itself should disallow weak ciphers and weak key exchange algorithms, in particular CBC ciphers, and weak diffie_helman variant of key exchange algorithms.

## Solution
Rather than building a custom image that we have to maintain, we let the EB deployment harden the SSH server on deployment through the use of an ebextension file.

We cannot just removed ciphers and algos we don't want, but instead we must state the ciphers and algos we specifically allow. 

- [x] Tested on staging and verified by pairing with @justynoh

### DIFF
--- a/.ebextensions/harden_sshd.config
+++ b/.ebextensions/harden_sshd.config
@@ -1,9 +1,9 @@
-container_commands:
+commands:
     harden_sshd_remove_weak_ciphers_and_key_exchange_algos:
         command:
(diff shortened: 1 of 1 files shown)


justynoh (approved):
lgtm

PR #5088, opened 2022-10-10
chore: remove the double build-step for EB deployment

timotheeg:
## Problem
Our deployments are slow because we build twice:
- once as a github action step
- once in the docker image creation

That should be unnecessary. If the docker image building fails, the overall github action should also fails.

## Solution
Drop build step from github action

TODO:
- [x] Test on staging that a deployment is successful with a single build step
- [x] Test on staging that, say, a compilation error, causes the overall github workflow to fail when the docker build step fails

### DIFF
--- a/.github/workflows/deploy-eb.yml
+++ b/.github/workflows/deploy-eb.yml
@@ -32,13 +32,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


justynoh (approved):
lgtm

timotheeg:
Tested fail run here: https://github.com/opengovsg/FormSG/actions/runs/3217685588

Issue #5061, opened 2022-10-06
Paragraph field does not display the content as typed


justynoh:
<img width="357" alt="image" src="https://user-images.githubusercontent.com/25571626/194281816-2b510b68-02ad-4308-8a25-ffb5b06ab592.png">

This is an issue both with the builder and public form. As compared to angular, which shows the paragraph field contents exactly (even with multiple newlines exactly the same)

karrui:
grrr i thought this was fixed, regression somewhere :(. or is it edge/gsib-only?

justynoh:
the screenshots were on edge and gsib, seems like the same on mac tho



![Image](https://user-images.githubusercontent.com/25571626/194286607-198f6bf0-95e8-47cb-90f9-43bd51335e8e.png)

Issue #5058, opened 2022-10-06
Tooltip for description of "Include PDF response" for email confirmations in storage mode.


justynoh:
Email field in storage mode allows for email confirmations, but disables Include PDF responses. However there is no information provided as to why it is disabled. 

<img width="431" alt="image" src="https://user-images.githubusercontent.com/25571626/194280655-0c3868c9-5a3a-4634-8003-38d5fed346ce.png">

Angular has a tooltip for this, which we can replicate as a description under the toggle.

<img width="491" alt="image" src="https://user-images.githubusercontent.com/25571626/194280487-802157a1-cb64-4599-85bc-8158455d8b7e.png">

Issue #5055, opened 2022-10-06
Update toasts for paragraph field


justynoh:
Currently, create/update/dupe/delete (CUDD?) operations for paragraph field still say "Statement" in the toast. We should change this to "Paragraph".

![image](https://user-images.githubusercontent.com/25571626/194280921-fa0ce04c-c7b3-4d6c-9b6e-6c7e8ddbfe4a.png)

Issue #5050, opened 2022-10-06
Long question titles are not wrapped in various places


justynoh:
Long question titles are not wrapped or cut off in places resulting in strange UI.

1. in logic:
![PXL_20221006_053956074](https://user-images.githubusercontent.com/25571626/194276966-97e4b290-c762-4076-b015-72933e8ade68.jpg)

2. In question titles and descriptions (except tables, for some reason)
<img width="491" alt="image" src="https://user-images.githubusercontent.com/25571626/194277168-02a22d40-0456-43c5-b99d-d38e556f6e0c.png">

PR #5105, opened 2022-10-11
fix: misc builder fixes for builder tool and public forms, add MarkdownText component

justynoh:
## Problem
1. #5058 
2. #5050 
3. ~~#5055~~
4. #5061 

## Solution

1. Add descriptions for toggles that mimic tooltips that exist in angular in `EditTable`.
2. Add overflow-wrap in `FormLabel`, add overflow hidden in `ThenShowBlock`, add wrap anywhere in checkbox and radio labels.
3. ~~Toasts for paragraph field are determined by the title of the field determined by the question title for Statement/Paragraph fields. Change this to paragraph in `EditParagraph`. I THINK this is backward compatible, since the name of the questions isn't enforced anywhere, and we filter out entire statements/paragraphs during submission.~~ It is backwards compatible but makes for strange UI (statements created in angular are called "Statement" while those created in react are called "Paragraph", so the toasts might show differently depending on actions). I'll open another PR to fix it with a more sweeping change.
4. This one was a bit complicated. Looking at the angular app, we realize that links are automatically created from URLs in many more places in the app than we accounted for:
  - Form instructions
  - Header descriptions
  - Paragraphs
  - Image captions
  - Form descriptions
  - Thank you page description
 
  Change all of these to make use of ReactMarkdown. But react markdown does a weird thing where it sucks up consecutive newlines... (see [this issue](https://github.com/remarkjs/react-markdown/issues/278)) To get newlines to show up, wrap them in a box with whitespace = pre-wrap. This means that **3 or more `\n`'s always looks like only 2 `\n`'s, because it gets rendered as two consecutive `<p>` tags by react markdown**. Thus add new processing to the `children` text to inject new-line nodes into the markdown AST.

**Breaking Changes** 


frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -145,7 +148,12 @@ const FormLabelDescription = ({
   })
 
   return useMarkdown ? (
-    <ReactMarkdown components={mdComponents}>{children}</ReactMarkdown>
+    // Wrap markdown with a <div white-space='pre-wrap'> to get consecutive newlines to show up
+    <Box whiteSpace="pre-wrap">
+      <ReactMarkdown components={mdComponents} remarkPlugins={[gfm]}>
```
karrui: as discussed, should we create a separate `TextMarkdown` component that uses all the plugins we want to have a consistent markdown usage across the app?

also use something like [remark-breaks](https://github.com/remarkjs/remark-breaks) so breaks are (in)correctly rendered. (breaks markdown spec but for our use case its fine lol)

wanlingt (approved):
lgtm! (other than kr's comment)

nit: btw the react markdown link leads to the homepage and not the issue

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
justynoh: looking at https://github.com/vmware-tanzu/kubeapps/pull/4765 but it didn't work - some error with gfm, breaks and react-markdown not playing nice together.

Tried following this "little plugin" but it's still not working (now it shows up worse than it would have with pre-wrap): https://github.com/remarkjs/react-markdown/issues/278#issuecomment-971895356

image

karrui:
ok hope stories catches all regressions lol

frontend/src/components/MarkdownText/MarkdownText.tsx
```diff
@@ -6,18 +6,27 @@ import gfm from 'remark-gfm'
 
 interface MarkdownTextProps {
   components?: Components
+  /**
+   * Whether to allow sequetial new lines to generate sequential line breaks.
```
karrui: lazy to push again to fix this typo...

karrui: [9a69184](https://github.com/opengovsg/FormSG/pull/5105/commits/9a69184d9a06891a4341ac48b474fadffb0cd5a4)

frontend/src/templates/Field/Paragraph/ParagraphField.tsx
```diff
@@ -24,9 +25,12 @@ export const ParagraphField = ({
       },
     },
   })
+
   return (
-    <ReactMarkdown components={mdComponents}>
-      {schema.description}
-    </ReactMarkdown>
+    <Box mt="1rem">
```
karrui: any reason why you added a margin top to this? @justynoh

justynoh: shoot, no, no reason. should be deleted, i must have missed deleting it after copy paste oops

justynoh: [7cbbfff](https://github.com/opengovsg/FormSG/pull/5105/commits/7cbbfff14b8b887817d5f4d8ae55d8b05023d9b3)

PR #5110, opened 2022-10-11
fix: add transformAllIsoStringsToDate step on form preview dto

karrui:
## Problem

Preview form was crashing due to type mismatch in the preview form DTO.
This PR fixes that :(.

No prod impact since it is not released yet.

## Solution

**Breaking Changes** 

- [x] No - this PR is backwards compatible, but we should really get it merged in.

**Bug Fixes**:

- fix: add `transformAllIsoStringsToDate` step on form preview dto

## Tests
- [ ] Open date picker with custom date range in preview. Should work properly and dates correctly disabled if out of range. Should have no runtime errors


justynoh (approved):
+11111

wanlingt (approved):

PR #5115, opened 2022-10-11
fix: update toasts for non-input fields, design and logic, remove steppers from form limit toggle

### LINKED ISSUE #5054: Remove steppers for maximum number of responses in Settings - General
Without focusing the input, clicking out never triggers onBlur. Clicking the steppers will increment the number but not focus is. Therefore we never save the number, which is confusing for users. 

We should just get rid of the steppers, since anyway the number of admins who are going to click the steppers is likely very little.

### LINKED ISSUE #5055: Update toasts for paragraph field

### LINKED ISSUE #5056: Update toasts for non-input fields
Related to #5055, but separate issue since this can be done post-rollout.

Actually, all non-input fields should have a different toast description, since CUDD operations for fields currently say 

The field "Image" has been updated.

But actually since there is no "canonical name" for image and paragraph fields, we might want to say "The image has been updated" and "The paragraph has been updated" rather than isolating "Image" and "Paragraph" as the name of the field, which doesn't seem intuitive. The header should also say "The header "XXX" has been updated." right? Just a suggestion.

justynoh:
## Problem
Toasts for non-input fields say things like 

The field "statement" has been updated.
The field "image" has been deleted.

This PR updates the toasts by replacing `field "${field.title}"` with conditionally determining what to put in that section of the string.

Also i realized design and logic language was not standardized with the field language, so I updated it as well.

Also, remove steppers for form limit toggle since no one is using that anyway lolol

Closes #5054, #5055 and #5056

## Solution

Added `getMutationToastDescriptionFieldName` which performs the computation. This function is used by all the mutations for create, update, delete, dupe.

**Breaking Changes** 

## Screenshots

image

image


timotheeg (approved):
lgtm. 

Just checking: did a designer or PM vet all textual and casing changes?

wanlingt (approved):
lgtm! 

+1 for design and pm review on copy, noticed the changes are mainly from active to passive voice

frontend/src/features/admin-form/settings/components/FormLimitToggle.tsx
```diff
@@ -90,7 +90,7 @@ const FormLimitBlock = ({
         ref={inputRef}
         // min={currentResponseCount + 1}
```
karrui: remove unused prop (sorry i know not you LOL)

karrui (approved):
lgtm

frontend/src/features/admin-form/settings/components/FormLimitToggle.tsx
justynoh: image

[3e0867d](https://github.com/opengovsg/FormSG/pull/5115/commits/3e0867d55278e179136701f7e5404b60d64f692d)

PR #5128, opened 2022-10-12
fix: ui admin cookie gets samesite=lax

timotheeg:
## Problem
A chrome bug prevents us to set admin ui cookie with 302 redirects. See [this post](https://bugs.chromium.org/p/chromium/issues/detail?id=696204#c106)

## Solution
Set the cookie with `samesite=lax`

cc @justynoh


justynoh (approved):

timotheeg:
Merged hotfix before all checks have run for speed. Only one change to a string, no business logic code change.

Paired with @justynoh @wanlingt

PR #5135, opened 2022-10-12
test: add login e2e tests with Playwright for React client, update build script

### LINKED ISSUE #5112: investigate e2e tests for react
Current e2e tests will soon become useless due to being tailored for AngularJS client. 
To get some semblance of safety during automated dependency updates (and releases), we should set up e2e for the react client.

Should be a heavy undertaking, this issue marks the set up of the e2e framework to be used for react.

suggest using playwright framework (actually already start liow so NO CHOICE I CHOOSE IT)

karrui:
## Problem

For added peace of mind when releasing, this PR adds e2e tests to test React app. Started with login tests, was the easiest to setup tests for. 

Misc improvement to build script so they are ran in parallel instead of in series (thanks @qin-guan for raising possibility), leading to a slight improvement in build times.

Closes #5112 

## Solution

**Breaking Changes** 

**Features**:

- test: add playwright setup for e2e testing

**Improvements**:

- chore: use concurrently to run npm build script 

## Tests

New e2e tests, and CI/CD action.


karrui:
~~setting as draft whilst i check why concurrent fails~~
Note to self: `-k` flag does not mean kill if non-zero, it means kill on any zero lol

justynoh (approved):

PR #5137, opened 2022-10-12
fix(client): fix logic multiselection dropdown from not showing

karrui:
## Problem

Users were unable to select dropdown options in logic builder if the dropdown option was a multi-select. Bug was introduced in #4778 due to library and code changes in rendering of dropdown.

This PR also fixes UI overflow on dropdown menu on long dropdown options. 

## Solution

**Breaking Changes** 

**Improvements**:
- feat: constrain dropdown item to a single line

**Bug Fixes**:
- fix: add floating portal to MultiSelectMenu
- fix: use minW instead of overflow=hidden to constrain width of logic dropdown


justynoh (approved):
thanks

PR #5138, opened 2022-10-12
fix(react): use string comparison for non-number equality checks

karrui:
## Problem

This PR fixes the bug where form logic will not correctly trigger the showing or hiding of certain fields due to strict equality checks returning false, such as where numbers and strings are compared to each other.

thought about using == instead of strict string equality, but string comparison is probably more explicit

Validated that the missed edge cases now work.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- fix: use string comparison for all logic equality checks

### DIFF
--- a/frontend/src/features/logic/utils/isConditionFulfilled.ts
+++ b/frontend/src/features/logic/utils/isConditionFulfilled.ts
@@ -49,8 +49,8 @@ export const isConditionFulfilled = (
     case LogicConditionState.Either: {
       // currentValue must be in a value in condition.value
       const condValuesArray = Array.isArray(condition.value)
-        ? condition.value
-        : [condition.value]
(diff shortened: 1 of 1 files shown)


justynoh:
Is this correct for decimals? Would it be better to fix the saved type for the condition as a string

karrui:
> Is this correct for decimals? Would it be better to fix the saved type for the condition as a string

LTE, GTE, Equals checks for decimals all cast to numbers, so it should be fine.

suspect the types are being casted by something else, could be react-hook-forms. :(. Fixing it here still gives a source of truth to the logic comparisons, i think ok la

edit: updated PR title for clarity

justynoh (approved):
ok, trust 👍

PR #5139, opened 2022-10-12
build: release v6.4.0

karrui:
## New

- fix(client): fix logic multiselection dropdown from not showing [`#5137`](https://github.com/opengovsg/FormSG/pull/5137)
- fix(react): use string comparison for non-number equality checks [`#5138`](https://github.com/opengovsg/FormSG/pull/5138)


justynoh (approved):

karrui:
i'll test on staging first before pushing release

PR #5146, opened 2022-10-13
fix(client): correctly display transfer ownership error from the server

### LINKED ISSUE #5147: Transfer ownership errors should be more descriptive
**Describe the bug**
Client is giving a generic error message on transfer ownership errors when the owner does not exist yet.
Should return the error message from server, which already has a good error message.

karrui:
## Problem

When 400 status code errors occur when transferring ownership of a form, a vague error message is displayed to the user. This PR updates the error message displayed to the one returned from the server, reducing confusion.

Closes #5147

## Solution

**Breaking Changes** 

**Improvements**:
- fix: correctly display transfer ownership error from the server

## Before & After Screenshots

**AFTER**:

Screenshot 2022-10-13 at 11 00 22 AM

### DIFF
--- a/frontend/src/features/admin-form/common/mutations.ts
+++ b/frontend/src/features/admin-form/common/mutations.ts
@@ -87,11 +87,15 @@ export const useMutateCollaborators = () => {
 
   const getMappedBadRequestErrorMessage = (
     formCollaboratorAction: FormCollaboratorAction,
+    originalErrorMessage: string,
   ): string => {
     let badRequestErrorMessage
     switch (formCollaboratorAction) {
       case FormCollaboratorAction.ADD:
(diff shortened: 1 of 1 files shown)


justynoh (approved):
actually should we default to just showing the error message returned by the server? like

`Sorry, the following error occurred. Please refresh the page and try again later. ${originalErrorMessage}`

karrui:
> actually should we default to just showing the error message returned by the server? like
> 
> `Sorry, the following error occurred. Please refresh the page and try again later. ${originalErrorMessage}`

Had a few discussions like this before, i think the conclusion is might be too verbose (and we should not actually be displaying original error messages actually and craft the client errors on the client)

Issue #5153, opened 2022-10-13
Switch modal should show submit intead of next.


amitogp:

amitogp:
For the feedback button on the bottom right, users may not know that pressing "next" automatically sends a feedback

PR #5154, opened 2022-10-13
fix: feedback modal submit button copy

justynoh:
## Problem

#5153 

## Solution

**Breaking Changes** 


justynoh:
not planned loool

PR #5166, opened 2022-10-13
fix: add correct parsing for myinfo dates

justynoh:
## Problem

Myinfo dates were not getting parsed correctly. This PR adds it as a hotfix.

## Solution
Parse myinfo dates with the format yyyy-MM-dd prior to passing it to the date field.

**Breaking Changes** 


karrui (approved):
lgtm 🙏

PR #5170, opened 2022-10-14
fix: add myinfo attribute to parsed responses if MyInfo field

karrui:
## Problem

Instead of relying on response object, MyInfo state should ultimately come from server/db. This PR ensures that if a form field is a MyInfo field, the attribute is added to the response.

Closes https://github.com/opengovsg/formsg-private/issues/111

## Solution

**Breaking Changes** 

- fix: add myinfo attribute to parsed responses if MyInfo field

### Tests
- [x] Submit myinfo **email** mode form on _React_ client. MyInfo responses should be correctly prefixed (and thus validated)
- [x] Submit myinfo **email** mode form on _AngularJS_ client. MyInfo responses should be correctly prefixed (and thus validated)

### DIFF
--- a/src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
+++ b/src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
@@ -151,6 +151,11 @@ export default class ParsedResponsesObject {
         processingResponse.isUserVerified = formField.isVerifiable
       }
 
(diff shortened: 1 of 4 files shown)


karrui:
Tested on stagingzxc

justynoh (approved):

wanlingt (approved):
thanksss

PR #5172, opened 2022-10-14
feat: separate admin and public environment switch

karrui:
## Problem

This PR fixes the edge case where switching in public form from React to AngularJS if you are logged in fails, since the admin form switch URL is used over the public form.


wanlingt (approved):
thanks!

wanlingt:
Testing on staging - does switching from react to angular work? does feedback report get sent?

- [x] email mode - admin logged in
- [x] email mode - admin not logged in
- [x] storage mode - admin logged in
- [x] storage mode - admin not logged in
- [x] admin dashboard switch react to angular

PR #5175, opened 2022-10-16
fix: prevent event default when enter

### LINKED ISSUE #5157: bug: forms with 1 text field refreshes on enter
**Describe the bug**
Forms with only 1 text field refreshes when the user presses the enter button after focusing on the field.

**To Reproduce**
1. Create new form
2. Add short field
3. Click preview/Open form link
4. Select field
5. Enter something (Optional)
6. Press enter
7. The page will refresh and any input will be cleared

**Expected behavior**
The page should not refresh on enter.

tohhengxing:
## Problem
This bug is weird. I wasn't able to trace the root cause, so I just handled the default event.
Here are some findings which might help
1. The refresh is triggered by the client
2. There is a query string appended after the refresh for some reason

Closes #5157 

## Solution
handled the default event.

## Before & After Screenshots

**BEFORE**:
<details>

https://user-images.githubusercontent.com/102740235/196020433-c35365af-0a19-4bfa-a8c0-f9de01e14e61.mp4

</details>

**AFTER**:
<details>

https://user-images.githubusercontent.com/102740235/196020437-e5a35da4-c29d-4e42-826d-ff0960f41edf.mp4

</details>


justynoh:
Thanks for the contribution @tohhengxing!

1. Should we add a comment above that line to explain what it fixes so that we know? A link to the issue number will also help.
2. Did you manage to investigate if this was only an issue with short text fields or do other fields trigger this behavior too? (e.g. number, long text, decimal)

tohhengxing:
> Thanks for the contribution @tohhengxing!
> 
> 1. Should we add a comment above that line to explain what it fixes so that we know? A link to the issue number will also help.
> 2. Did you manage to investigate if this was only an issue with short text fields or do other fields trigger this behavior too? (e.g. number, long text, decimal)

hey justyn. unfortunately the issue is present for almost every field type as long as there's only one field in the form. Do you need me to list out all the fields?

I looked through all code that had a keydown with enter, none of them were activated so I am not sure what's the issue either.

Might need more investigation into the reason. I can add the handler to all the fields but that would be a dirty fix.

justynoh:
Probably shouldn't always take advice from reddit, but this seems like a HTML thing, based on [this thread](https://www.reddit.com/r/reactjs/comments/czbuet/comment/eyxc2nt/?context=8&depth=9) and a bunch of other threads (just google "pressing enter refreshes page with input")

can we put the preventDefault in the submission handler instead? does that work

karrui:
> Probably shouldn't always take advice from reddit, but this seems like a HTML thing, based on [this thread](https://www.reddit.com/r/reactjs/comments/czbuet/comment/eyxc2nt/?context=8&depth=9) and a bunch of other threads (just google "pressing enter refreshes page with input")
> 
> can we put the preventDefault in the submission handler instead? does that work

don't think that works, that just disables the enter key to activate all buttons, which is bad for a11y

justynoh:
We should probably just push the fix down into the `Input` component itself, then. I don't think we ever want the behavior of enter to submit anywhere in the app anyway. Maybe we should add it to `NumberInput` as well

tohhengxing:
hi. I have added the changes requested. Unfortunately, I still don't know why the problem exists in the first place...oh well i guess some problems aren't meant to be known :P

I have tested for all of the following components.

Input

Basic Fields
1. Short Answer
2. NRIC
3. UEN

MyInfo
1. Name
2. Passport Number
3. Vehicle Number
4. Registered Address
5. Name of Employer
6. Marriage Certificate Number

Number Input
Basic
1. Mobile Number
2. Home Number
3. Number
4. Decimal

MyInfo
1. Mobile Number

justynoh:
closing and opening to retrigger ci D:

justynoh (approved):
anyway lgtm, thanks @tohhengxing !

PR #5176, opened 2022-10-16
fix: remove "This field is required" from Home Number field preview

### LINKED ISSUE #5059: Home number preview always shows "This field is required"
It shouldn't.

image

wanlingt:
## Problem

Home number field (required) shows the message "This field is required" in the form builder. This message should only appear when the user tries to submit without filling in the form.

Closes #5059

## Solution

Remove `defaultValue` - this makes it consistent with the Mobile field

## Before & After Screenshots

**BEFORE**:
Refer to issue

**AFTER**:

https://user-images.githubusercontent.com/56983748/196028051-1fd1134c-1612-4da1-8c98-aae020d17553.mov


karrui (approved):
lgtm

PR #5187, opened 2022-10-18
fix: inject RUM session IDs into feedback form

### LINKED ISSUE #5186: Report RUM session ID automatically with user feedback

justynoh:
## Problem

We want to know the RUM session ID for recorded sessions to make it easier to find in datadog.

Releasing as hotfix `v6.6.1`.

Closes #5186 

## Solution

Add a new field to the bug report form for RUM session ID. Inject the session ID into the programmatic form submission. This has been successfully tested on staging.

**Breaking Changes** 

## Screenshots

image

image


karrui (approved):
the target form already has the datadog field added yea

wanlingt (approved):

justynoh:
@karrui not yet, since having the target field without it being represented in code causes submissions to fail for current users. but will add once it deploys

PR #5188, opened 2022-10-18
feat: prevent benign frontend errors from being sent to datadog RUM

karrui:
## Problem

This PR prevents benign frontend errors from being sent to datadog RUM, so the errors we see on RUM can be less noisy and more useful. Do not want a repeat of Sentry in AngularJS :P.


justynoh (approved):
+1 
tagging @staceytan1998 - this would be the place to determine if we want to throw away sessions lasting < 1 min. Or, will having the datadog rum session ID in the feedback form be good enough from #5187?

PR #5190, opened 2022-10-18
feat: add breadcrumbs to admin form navbar

### LINKED ISSUE #5165: Replace top navbar Back button with breadcrumb
- Users are confusing the Back button with going back to the Builder, when it actually goes back to the Workspace
- Therefore, we're replacing the Back button with a breadcrumb, which will show All forms / Form name so it's clear that clicking on All forms goes back to Workspaces
- In this case, we won't be able to accommodate the Saved at... under breadcrumb, so we are moving it to the right

[figma here](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=11333%3A134780)

justynoh:
## Problem
We want to remove the back button in the admin form navbar, and replace it with a breadcrumb instead. This PR does this.

Closes #5165

## Solution
~~Use chakra's breadcrumb component to help with this. Considered using `use-react-router-breadcrumbs` but that seems like overkill for this use, since we only use breadcrumbs in this single place in our app, and we need to configure it to recognize the router paths anyway.~~ NO WE DON'T because i can't style the inner `<ol>` and i'm done with battling chakra. Haiz. nvm ba, since breadcrumb only appears in one place. Just use flex lor. Should announce as breadcrumb tho for a11y - should be a quick one. i'll fix it

**Breaking Changes** 

## Screenshots

desktop
image

tablet
image

mobile
image


karrui (approved):
lgtm

frontend/src/features/admin-form/common/components/AdminFormNavbar/AdminFormNavbarBreadcrumbs.tsx
```diff
@@ -27,19 +27,20 @@ export const AdminFormNavbarBreadcrumbs = ({
...
+      <BreadcrumbItem whiteSpace="nowrap">
         <BreadcrumbLink color="primary.500" onClick={handleBackButtonClick}>
           {isMobile ? <Icon as={BiHomeAlt} mb="-2px" /> : 'All forms'}
         </BreadcrumbLink>
       </BreadcrumbItem>
 
-      <BreadcrumbItem>
-        <Skeleton isLoaded={!!formInfo}>
-          <Text>{formInfo ? formInfo.title : 'Loading...'}</Text>
+      <BreadcrumbItem overflow="hidden">
```
karrui: i see you managed to work around it lulz

karrui (approved):
lgtm

frontend/src/features/admin-form/common/components/AdminFormNavbar/AdminFormNavbarBreadcrumbs.tsx
justynoh: no i didn't this was still a buggy implementation see the fix [48dd739](https://github.com/opengovsg/FormSG/pull/5190/commits/48dd7392ed8f5d9d04c2fa90b953f449f8a2ca29)

karrui (approved):

PR #5191, opened 2022-10-18
feat: add frontend app versioning to datadog RUM

karrui:
## Problem

This PR adds versioning to RUM so we can have version breakdowns of errors, etc, for easier spotting of whether version change caused regressions/fixed issues.


karrui:
Tested on DD, now there's versioning woohoo. Staging's stats will probably not be accurate though since we redeploy multiple times with the same versions

timotheeg (approved):
Nice!

PR #5192, opened 2022-10-18
feat: get better sort (and highlighting) in dropdown searches

karrui:
## Problem

Ok title was click bait. We do get better sorts and highlighting by replacing `match-sorter` + `react-highlight-words` combo though, but the main aim of this PR is to fix a `SyntaxError` that may occur (still can't repro, shrug). 

Almost certain this is due to https://github.com/opengovsg/FormSG/blob/7ef9af45f7c158756fbc938c3a4d063ebd66b030/frontend/src/components/Dropdown/components/DropdownItem/DropdownItemTextHighlighter.tsx#L40

Anyways, tried a new library that seems much nicer and has better highlighting so.... win-win?

Example error:
```js
SyntaxError: Invalid regular expression: /[Installation & Commissioning Air-conditioning Systems \(R32 Split Unit\) 25 Nov \(Fri\) 9am - 5pm \[CE \] ]/: Range out of order in character class
```


frontend/src/components/Dropdown/utils/defaultFilter.ts
```diff
@@ -23,7 +22,23 @@ export const defaultFilter = <Item extends ComboboxItem>(
...
     }
     sortKeys.push('description')
-    matchSorterOptions = { keys: sortKeys }
+
+    return fuzzysort
+      .go(value, items, {
+        all: true,
+        keys: sortKeys,
+        // Create a custom combined score to sort by. -500 to the desc score makes it a worse match
+        scoreFn: (a) => {
```
justynoh, line 31: i know the docs use `a` but should we rename to something like `sortKeysMatches` or something more descriptive

frontend/src/components/Dropdown/components/MultiDropdownItem/MultiDropdownItem.tsx
```diff
@@ -57,18 +57,32 @@ export const MultiDropdownItem = ({
...
-              showHoverBg={isHighlighted}
-              textToHighlight={label}
-            />
+            <Text
+              textStyle="body-1"
+              minWidth={0}
+              textOverflow="ellipsis"
+              whiteSpace="nowrap"
+              overflowX="hidden"
+            >
```
justynoh, line 66: just to clarify - this is not for sort and highlighting right? this is just to restrict the content to 1 line

justynoh (approved):
lgtm

PR #5193, opened 2022-10-18
fix: set inline display on all markdown links

karrui:
## Problem

Links were rendering as inline-flex in react-markdown due to the `Link` stylings in the default theme, causing unwarranted layout shifts. Set to default to inline display so text flows.

Added a new story to ensure no regression.


justynoh (approved):
i watched

PR #5215, opened 2022-10-20
feat: record feedback ratings for instrumentation

timotheeg:
## Problem
We don't have a live view of feedback ratings in datadog. Feedback ratings are tracked in our product charts but the view is limited to the past 30 days.

image

## Solution
Record feedback events to datadog. Those will allow very granular view in to the feedback ratings for any time period.

The metric type [distribution](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types) is used, such that we can track counts, and average.

Note 1: a counter metric would be suitable for a distribution graph (if the ratings are tracked as tags to the metric), but average wouldn't be available. a metric type `distribution` gives us what we need (in fact, more than what we need, sadly).

Note 2: We are not tagging the feedback by form id because tag cardinality too high.

View in DataDog:
image

### DIFF
--- a/src/app/modules/feedback/feedback.controller.ts
+++ b/src/app/modules/feedback/feedback.controller.ts
@@ -2,6 +2,7 @@ import { celebrate, Joi, Segments } from 'celebrate'
 import { StatusCodes } from 'http-status-codes'
 import { ErrorDto, PrivateFormErrorDto } from 'shared/types'
 
+import config from '../../config/config'
 import { statsdClient } from '../../config/datadog-statsd-client'
 import { createLoggerWithLabel } from '../../config/logger'
 import { createReqMeta } from '../../utils/request'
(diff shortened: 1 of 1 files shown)


src/app/modules/feedback/feedback.controller.ts
```diff
@@ -55,6 +57,14 @@ const submitFormFeedback: ControllerHandler<
     .andThen(() => FormService.retrieveFullFormById(formId))
     .andThen((form) => FormService.isFormPublic(form).map(() => form))
     .andThen((form) => {
+      statsdClient.distribution('formsg.feedback.rating', rating, 1, {
+        rating: `${rating}`,
+        ui:
+          req.cookies?.[config.reactMigration.respondentCookieName] === 'react'
+            ? 'react'
+            : 'angular',
+      })
```
timotheeg, line 66: ui tracking is temporary, so I'm not trying to make this clean.

karrui (approved):
lgtm

PR #5218, opened 2022-10-20
feat: use MyInfo constants for title on MyInfo field creation

karrui:
## Problem

This PR updates MyInfo field creations to always use the shared constant values for the title for consistency.

@timotheeg this approach has a con in which IF the shared constants are incorrect, it will require a redeploy to correct. Assume that's a fair tradeoff?

## Solution

**Breaking Changes** 

**Features**:

- feat: assign myinfo title to stored constant on field creation 

## Tests
- [x] Use fetch/curl/something to create a MyInfo field with a random, non-myinfo title. The resulting title should still be what is stored in the shared constants.


karrui:
~~Testing on staging first~~
Tested on staging, works fine.

```
fetch("***********/api/v3/admin/forms/**********/fields?to=7", {
  ...,
  "body": "{\"disabled\":false,\"required\":true,\"title\":\"Anyhowlysettitlehaha\",\"description\":\"\",\"fieldType\":\"dropdown\",\"myInfo\":{\"attr\":\"race\"},\"fieldOptions\":[]}",
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});
```

justynoh (approved):

PR #5225, opened 2022-10-20
build: use new CD pipeline to allow for datadog source maps, add cache control

karrui:
## Problem

Updated pipeline to build and upload React source maps to Datadog in a secure manner. The previous build pipeline had discrepancies with what Docker build output (probably due to `chakra-ui/cli` not running (since the source files were not copied into the docker build yet, and `patch-package` not running properly).

Also update express to apply no-cache headers to our served React's `index.html` so we can hopefully lessen chunk 404 errors.

Should hopefully work now. Source maps are now uploading properly inside the docker build.

Ignore the bunch of commits where I flailed around like a headless monkey. But we should really clean up our docker/CD pipeline lol it seems like bandaids on top of bandaids.

Closes https://github.com/opengovsg/formsg-private/issues/112 (again)

## Solution

**Breaking Changes** 

## Deploy Notes

- Update CD pipeline
- Correctly copy frontend/patches in Dockerfile.production so patches actually get applied

### DIFF
--- a/.github/workflows/deploy-eb.yml
+++ b/.github/workflows/deploy-eb.yml
@@ -34,9 +34,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Setup secrets for datadog sourcemap deployment
         run: |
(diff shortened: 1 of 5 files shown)


timotheeg (approved):
Thanks KarRui! Looks good to me.

PR #5235, opened 2022-10-23
fix: endpage state is dirty when paragraph or buttonLink fields are undefined

darrelhong:
https://user-images.githubusercontent.com/39296145/197379039-536faf21-c815-4df1-af2c-df3f74f214a7.mov

Fixes a super small bug where unsaved changes prompt will appear when no edits are made to end page input. react-hook-forms thinks end page inputs is dirty when fields are missing in `defaultValues`, reason being it's assigns default empty strings to input and deems empty string not equal to undefined.

Fix
check if length of `dirtyFields` is 0 instead of using isDirty. [dirtyFields](https://github.com/react-hook-form/react-hook-form/issues/4740#issuecomment-818598647) will be empty when keys are not supplied in `defaultValues`

### DIFF
--- a/frontend/src/features/admin-form/create/end-page/EndPageDrawer.tsx
+++ b/frontend/src/features/admin-form/create/end-page/EndPageDrawer.tsx
@@ -74,7 +74,7 @@ export const EndPageInput = (): JSX.Element => {
 
   const {
     register,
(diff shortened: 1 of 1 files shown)


justynoh (approved):
lgtm, thanks @darrelhong !

Issue #5239, opened 2022-10-25
React app is not setting collaborators in lowercase


karrui:
History lesson: When FormSG was first built, MongoDB did not support case insensitive search very well (had to use regexes), and thus the AngularJS transformed all emails to lowercase before sending them to the server. 

However, this was a mistake-ish, and the server should be the one lowercasing emails (for uniqueness) before updating the document.

This should be fixed, and there should also be backwards compatibility for old, non-lowercased emails. Moving forward, the setter should automatically set emails to the correct case in the database

karrui:
TODO: add collation index for case-insensitive search in the future, needs to be released concurrently with code that references that index, or current app will be sluggish (since new index is not compatible with old index)

Can probably do in the future...

PR #5240, opened 2022-10-25
feat: use lowercased string operations when setting/getting emails from db

karrui:
## Problem

This PR uses lowercased email equality to determine ownership of forms. Moves email cased-ness responsibility to the server instead of the client.

Related to #5239

## Solution

**Breaking Changes** 

- [x] No - this PR is backwards compatible, even though there are schema changes. Moved email cased-ness responsibility to the server instead of the client.

**Features**:
- feat: update form and user models to set to lowercased automatically

**Improvements**:

- feat: add email case insensitivity when checking form permissions
- feat: add case insensitive email check when transferring ownership
- feat: search permission list with lowercased emails

## Before & After Screenshots
Should have no changes.

## Tests

- [ ] Add collaborator with mixed case email. Return response should correctly be lowercased.
- [ ] In the staging DB (pls don't do on prod), edit a form in which you are a collaborator to your email, but in mixed case. You should still have access to the form.
- [ ] Transfer ownership of your form with a mixed case email. Should correctly transfer (if the lowercased version of the email exists) (can just add <your-email>+1@open.gov.sg (log in once to create the user) for ease of testing.

### DIFF
--- a/frontend/src/features/admin-form/common/queries.ts
+++ b/frontend/src/features/admin-form/common/queries.ts
@@ -105,7 +105,8 @@ export const useAdminFormCollaborators = (formId: string) => {
     }
     // Else use permissionList first
     return form.permissionList.some(
-      (perms) => perms.write && perms.email === user.email,
+      (perms) =>
+        perms.write && perms.email.toLowerCase() === user.email.toLowerCase(),
     )
   }, [collaborators, form, isFormAdmin, user])
 
--- a/frontend/src/features/admin-form/common/queries.ts
+++ b/frontend/src/features/admin-form/common/queries.ts
@@ -99,7 +99,8 @@ export const useAdminFormCollaborators = (formId: string) => {
     // Collaborators is source of truth if it has already loaded.
     if (collaborators) {
       return collaborators.some(
-        (perms) => perms.write && perms.email === user.email,
+        (perms) =>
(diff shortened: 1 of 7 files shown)


frontend/src/features/admin-form/common/queries.ts
```diff
@@ -99,7 +99,8 @@ export const useAdminFormCollaborators = (formId: string) => {
     // Collaborators is source of truth if it has already loaded.
     if (collaborators) {
       return collaborators.some(
-        (perms) => perms.write && perms.email === user.email,
+        (perms) =>
+          perms.write && perms.email.toLowerCase() === user.email.toLowerCase(),
       )
     }
     // Else use permissionList first
```
wanlingt, line 106: should we add it to the lines below too?
```suggestion
    // Else use permissionList first
    return form.permissionList.some(
      (perms) => perms.write && perms.email.toLowerCase() === user.email.toLowerCase(),
    )
```

wanlingt (approved):
lgtm!

frontend/src/features/admin-form/common/queries.ts
karrui: good catch! 4e01e48

timotheeg (approved):

PR #5247, opened 2022-10-25
feat: hopefully add robustness to mobile viewport rendering

### LINKED ISSUE #5244: useIsMobile hook does not trigger desktop breakpoint on certain browsers
**Describe the bug**

Our `useIsMobile` hooks are not re-firing on certain browsers (and on certain elements), leading to an interesting state where some elements are being displayed and some are not. 

Not able to replicate, but we may want to look into updating our `useIsMobile` hook to use media queries instead of breakpoints.

See https://chakra-ui.com/docs/hooks/use-media-query

Possibly related to #5243

Screenshots submitted by form admin:

image
image001

karrui:
## Problem

Not sure if this will solve it, but maybe? The later versions of `@chakra-ui/media-query` allow for base default breakpoints??? we can try using media queries if this does not help still. 

Closes #5244 hopefully

## Solution

**Breaking Changes** 

**Improvements**:

- refactor: use new useBreakpointValue import,
- refactor: add useModalSize hook to reduce repeated code
- build: update browserlist to be same in dev and prod  (loader seems to break on esm modules if using development profile)

## Before & After Screenshots
Should have no changes

## Deploy Notes

**New dependencies**:

- `@chakra-ui/media-query` : separate, newer version than the one bundled with our current `@chakra-ui/react` version (that we can't update yet since we are on a lower version of react)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1489,6 +1489,11 @@
         "@chakra-ui/utils": "1.10.4"
       }
     },
+    "@chakra-ui/breakpoint-utils": {
+      "version": "2.0.4",
+      "resolved": "https://registry.npmjs.org/@chakra-ui/breakpoint-utils/-/breakpoint-utils-2.0.4.tgz",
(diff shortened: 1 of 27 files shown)


karrui:
zzzz storybook using ancient webpack, cannot handle esm yet!!!

karrui:
maybe this can wait, i'll try another approach

karrui:
superceded by #5250

PR #5250, opened 2022-10-26
feat: use media queries for `useIsMobile` hook

### LINKED ISSUE #5244: useIsMobile hook does not trigger desktop breakpoint on certain browsers

karrui:
## Problem

Use media queries instead of the result of `useBreakpointValue` to trigger layout changes.  May break on virtual browsers, we should test before releasing if we can our hands on such devices.

Supercedes #5247
Closes #5244 hopefully

## Solution

Documentation for [chakra's useMediaQuery()](https://chakra-ui.com/docs/hooks/use-media-query)

**Breaking Changes** 

### DIFF
--- a/frontend/src/hooks/useIsMobile.ts
+++ b/frontend/src/hooks/useIsMobile.ts
@@ -1,11 +1,9 @@
-import { useBreakpointValue } from '@chakra-ui/react'
+import { useMediaQuery } from '@chakra-ui/react'
+
+import { BREAKPOINT_VALS } from '~theme/foundations/breakpoints'
 
 export const useIsMobile = (): boolean => {
-  const isMobile = useBreakpointValue({
-    base: true,
-    xs: true,
-    md: false,
-  })
+  const [isLargerThanMd] = useMediaQuery(`(min-width: ${BREAKPOINT_VALS.md})`)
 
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useIsMobile.ts
```diff
@@ -1,11 +1,9 @@
...
+
+import { BREAKPOINT_VALS } from '~theme/foundations/breakpoints'
 
 export const useIsMobile = (): boolean => {
-  const isMobile = useBreakpointValue({
-    base: true,
-    xs: true,
-    md: false,
-  })
+  const [isLargerThanMd] = useMediaQuery(`(min-width: ${BREAKPOINT_VALS.md})`)
```
timotheeg: This is pretty neat!

timotheeg (approved):

PR #5251, opened 2022-10-26
feat: correctly(?) upload sourcemaps for Datadog, again

karrui:
## Problem

Fix source map URL to the exact environment, also add repository link so any errors can direct us to the offending file immediately.

Part of improving observability.

## Solution

**Breaking Changes** 

**Improvements**:

- Hopefully upload source maps correctly.

## Before & After Screenshots
No changes

## Deploy Notes

- Add new Github environment secrets `APP_URL` (not really secret, but just adding it for per-env setup)
- add new font-wqy-zenhei package package in our `Dockerfile.prod` since the one we were relying on seems to have been pulled/renamed.

### DIFF
--- a/Dockerfile.production
+++ b/Dockerfile.production
@@ -39,7 +39,7 @@ RUN --mount=type=secret,id=dd_api_key \
     npx @datadog/datadog-ci sourcemaps upload ./dist/frontend \
     --service=formsg-react \
     --release-version=$APP_VERSION \
-    --minified-path-prefix=$APP_URL/static/js \
+    --minified-path-prefix=$APP_URL \
     --repository-url=$REPO_URL
 
 RUN npm prune --production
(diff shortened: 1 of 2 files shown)


Dockerfile.production
```diff
@@ -77,7 +80,7 @@ ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
 ENV PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium-browser
 
 # This package is needed to render Chinese characters in autoreply PDFs
-RUN echo @edge http://nl.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories && apk add wqy-zenhei@edge
+RUN echo @edge http://nl.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories && apk add font-wqy-zenhei@edge
```
karrui, line 83: @wanlingt you can also add this in the interim so your staging-alt deployment passes before this PR is merged in

timotheeg (approved):

PR #5270, opened 2022-10-27
feat: add retry/refresh when chunk loading fails

karrui:
## Problem

Seeing a small amount of chunk loading errors due to older clients.

Forcing a refresh on lazy loading errors (on first refresh) should get latest the `index.html` from the server (which references new chunks), reducing chunk errors

## Solution

**Breaking Changes** 

**Improvements**:

- feat: add retry/refresh when lazy loading fails

## Before & After Screenshots
Should have no changes

### DIFF
--- a/frontend/src/utils/lazyRetry.ts
+++ b/frontend/src/utils/lazyRetry.ts
@@ -1,0 +1,36 @@
+import { ComponentType } from 'react'
+
+/**
+ * Retry React.lazy calls on error, or refresh the page if it was a chunk load
+ * error to hopefully retrieve the correct version of the chunk to due
+ * no-cache header.
+ *
+ * Retrieved from https://www.codemzy.com/blog/fix-chunkloaderror-react.
+ *
+ * @note this only works for route based code splitting https://reactjs.org/docs/code-splitting.html#route-based-code-splitting
+ * Will cause infinite loop if used for normal code splitting since multiple lazy calls may be triggered in a single component.
+ */
+export const lazyRetry = <T extends ComponentType>(
+  componentImport: () => Promise<{ default: T }>,
+) => {
+  return new Promise<{ default: T }>((resolve, reject) => {
+    // check if the window has already been refreshed
+    const hasRefreshed = JSON.parse(
+      window.sessionStorage.getItem('retry-lazy-refreshed') || 'false',
+    )
+    // try to import the component
+    componentImport()
+      .then((component) => {
+        window.sessionStorage.setItem('retry-lazy-refreshed', 'false') // success so reset the refresh
+        resolve(component)
+      })
+      .catch((error) => {
+        if (!hasRefreshed) {
+          // not been refreshed yet
+          window.sessionStorage.setItem('retry-lazy-refreshed', 'true') // we are now going to refresh
+          return window.location.reload() // refresh the page
+        }
+        reject(error) // Default error behaviour as already tried refresh
+      })
+  })
+}
--- a/frontend/src/utils/lazyRetry.ts
+++ b/frontend/src/utils/lazyRetry.ts
@@ -13,21 +13,23 @@ import { ComponentType } from 'react'
 export const lazyRetry = <T extends ComponentType>(
   componentImport: () => Promise<{ default: T }>,
 ) => {
+  const sessionStorageKey = `retry-lazy-${process.env.REACT_APP_VERSION}-refreshed`
+
   return new Promise<{ default: T }>((resolve, reject) => {
     // check if the window has already been refreshed
     const hasRefreshed = JSON.parse(
-      window.sessionStorage.getItem('retry-lazy-refreshed') || 'false',
+      window.sessionStorage.getItem(sessionStorageKey) || 'false',
     )
     // try to import the component
     componentImport()
       .then((component) => {
-        window.sessionStorage.setItem('retry-lazy-refreshed', 'false') // success so reset the refresh
+        window.sessionStorage.setItem(sessionStorageKey, 'false') // success so reset the refresh
         resolve(component)
       })
       .catch((error) => {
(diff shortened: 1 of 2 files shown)


frontend/src/utils/lazyRetry.ts
```diff
@@ -0,0 +1,36 @@
...
+    componentImport()
+      .then((component) => {
+        window.sessionStorage.setItem('retry-lazy-refreshed', 'false') // success so reset the refresh
+        resolve(component)
+      })
+      .catch((error) => {
+        if (!hasRefreshed) {
+          // not been refreshed yet
+          window.sessionStorage.setItem('retry-lazy-refreshed', 'true') // we are now going to refresh
+          return window.location.reload() // refresh the page
```
timotheeg, line 33: If there is a client-side cache issue, this soft refresh might not be enough to get the latest version 🤔

We could try a more aggressive refresh with a dummy query string arg, like `?r=${Math.random()`

timotheeg, line 33: Also, just checking, does a call to `window.location.reload()` stop JS execution immediately in the current page?

I'm asking because there are multiple `componentImport()` in flight in parallel. If they have differing success/failure status, they could overwrite the value of `retry-lazy-refreshed` (That wouldn't stop the refresh itself though)

frontend/src/utils/lazyRetry.ts
```diff
@@ -0,0 +1,36 @@
...
+    // try to import the component
+    componentImport()
+      .then((component) => {
+        window.sessionStorage.setItem('retry-lazy-refreshed', 'false') // success so reset the refresh
+        resolve(component)
+      })
+      .catch((error) => {
+        if (!hasRefreshed) {
+          // not been refreshed yet
+          window.sessionStorage.setItem('retry-lazy-refreshed', 'true') // we are now going to refresh
```
timotheeg: Why do we need to store this in `sessionStorage`? `sessionStorage` survives a page refresh, but the `hasRefreshed` flag is meant to apply only for the current instance of the app, correct?

frontend/src/utils/lazyRetry.ts
karrui: it should, since our `index.html` has no caching, so a refresh should refetch a fresh `index.html`... in theory.

karrui: yes, that's why there was a big note that this current component only works if we do route-level lazy loading (meaning only a single component load in flight). do you want me to change it to allow parallel lazy loads even with that in mind?

frontend/src/utils/lazyRetry.ts
karrui: makes more sense to store it here than in localStorage, no? but you are right.. only happens for a single deploy...

wonder if......... we should just refresh if the version of the app has changed......

timotheeg: Well, I reckon it shouldn't be in sessionstorage or localstorage; it should just be in app memory. A singleton flag would do.

karrui: if its in memory won't it be cleared after the first refresh? then it'll just lead to an infinite refresh loop if the chunk just 404s forever

frontend/src/utils/lazyRetry.ts
timotheeg: The way I was looking at this is like that: if the links are broken now, we're assuming a bad cache. Something was cached incorrectly for some reason (??). If that incorrect cache is expiry based, the browser might still use whatever copy it holds on a refresh, without doing a full reload.

This is a bit of stretch, with a no-cache policy header, there should not be random expiry based cache happening, so I'm good that a reload should fetch the right thinngs.

That being said... are there ways other than cache we can get 404s on chunks? And I'm thinking maybe yes 😢 . Splitting the app in chunks means the client makes more requests to fetch assets, and at release time we still have both old and new version servers running in parallel, so some requests for new chunks could hit old servers (or vice versa) and fail.

During deployments, there could even be consecutive failures depending on which servers the requests hit. Urgh 😢 

And if we eventually do extended blue/green deployments, that problem would only get worse. We'll have to address it at some point.

frontend/src/utils/lazyRetry.ts
timotheeg: Oh you meant to do a single refresh overall? And if still broken, user would refresh manually thereafter?

I thought we would try to refresh till it works (under the assumption is eventually will work). In the cases of deployments (with vN and vN+1 existing in parallel), we could have multiple consecutive failures, but they would eventually resolve.

But you're right! If we have a broken deployment, and all clients go into refresh frenzy, they might DDOS us 😰. For continuous refresh, we should have some backoff mechanism.

karrui: single refresh should account for the bulk of the errors... but this would be a single refresh for the lifetime of the tab... which doesn't mean much. Maybe we should update once per version, which would already exist with `process.env.REACT_APP_VERSION`!

frontend/src/utils/lazyRetry.ts
timotheeg: 😅😱 Busted! I didn't read the big comment 🤦😅

Thanks for the clarification!

frontend/src/utils/lazyRetry.ts
karrui: done in [bbb0eb1](https://github.com/opengovsg/FormSG/pull/5270/commits/bbb0eb1d298d543ba4ca95bb4ae6a5ea54b7874f)

timotheeg: Aye, that would be nice 👍

timotheeg (approved):
LGTM 👍 !

And we can iterate this as we go (if needed)!

PR #5272, opened 2022-10-27
feat: add meta tags to react app

### LINKED ISSUE #5302: add meta tags for social media preview

wanlingt:
## Problem

To maintain feature parity with AngularJS, we want to add meta tags for public form social media previews 

Closes #5302 

## Solution

- Inject `title`, `description` and `image` tags
- `image` tag is not dynamically generated; can explore other options like [prerender.io](https://prerender.io/) if we wish to do image dynamic generation
- `serveFormReact` takes in a new parameter, `isPublicForm`, which determines if it should return `servePublicFormReact`. Meta tags are generated in `servePublicFormReact`
- As we now access `index.html` at bootstrap time, `npm run build` is required for both e2e and backend tests. The `ci` script has been modified to combine build, e2e and backend tests into the same step

## Before & After Screenshots

**AFTER**:
**Non-public form pages**
image

**Public form page**
image

## Tests

- [Preview and generate](https://www.opengraph.xyz/) Open Graph meta tags, check that `title`, `description` and `image` are populated

## Deploy Notes
**New dependencies**:
- `html-escaper` : used to escape the strings we inject into `index.html`


src/app/modules/react-migration/react-migration.routes.ts
```diff
@@ -7,7 +7,7 @@
 
 ReactMigrationRouter.get(
   '/:formId([a-fA-F0-9]{24})',
-  ReactMigrationController.serveForm,
+  ReactMigrationController.servePublicForm,
```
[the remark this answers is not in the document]
karrui: should we rate limit? @timotheeg? or ignore. i think we can rate limit to use our submissions rate limit of 80/second, even though we have cloudflare caching.

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -56,6 +56,14 @@ const serveFormReact =
...
       },
     })
+    let reactHtml = readFileSync(path.join(reactFrontendPath, 'index.html'), {
+      encoding: 'utf8',
+    })
+    reactHtml = reactHtml
+      // there are multiple __OG_TITLE__ and we are unable to use string.replaceAll so we use regexp
+      .replace(new RegExp('__OG_TITLE__', 'g'), 'FormSG')
+      .replace('__OG_DESCRIPTION__', 'Form Manager for Government')
+
```
karrui: should we move this block down to after the public form conditional so we do not do double work?

wanlingt: yes

wanlingt: [9b407b1](https://github.com/opengovsg/FormSG/pull/5272/commits/9b407b152ac4c2f3da77c6257c3cb4f211d0c3a6)

wanlingt:
@pearlyong @staceytan1998 took the old photo from the angular code for social media previews - let me know if the image should be updated

staceytan1998:
Social images here, 

Public form links: https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=13764%3A134946 

Non-public form links: https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=13821%3A135781 

Let me know if you need these in any other format!

src/app/modules/react-migration/react-migration.routes.ts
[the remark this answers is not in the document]
wanlingt: [cfe98ed](https://github.com/opengovsg/FormSG/pull/5272/commits/cfe98ed858d0072e19c662ed8fa26aa71b102089) - used the submissions rate limit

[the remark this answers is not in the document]
timotheeg: Apologies for the delay in commenting here 😱 , this is a GET on the form's html page, I reckon there should be no rate limiting at all 🤔 

The issue is the file system access and we can easily clear that. I'd recommend we load the file from filesystem at bootstrap and keep it in memory. We can then run the replacements on requests (which is light).

Just a note on that: loading a filesystem asset into memory is OK to do in an immutable infrastructure. If we were releasing by pushing files on deployment on long-running servers, it wouldn't be the right thing to do, or we'd need additional processes, of timeout, or filesystem hooks, to be notified of a change to the file and reload it.

With immutable infra, the file will never change for the lifetime of the nodejs app. Releases spin a new server, and again the index file don't change for the lifetime of the new app, so it's OK to cache it in memory.

One more note on that, reading from filesystem is by itself not terrible either. Linux has an excellent filesystem cache, and so for `frontend/index.html`, which has very frequent access, it's very likely the filesystem read call actually gets the content from OS FS cache in memory, rather than actually accessing the disk. Still, since we know the read characteristics, we might as well handle the in-memory cache explicitly in app ourselves.

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -49,24 +114,47 @@ const serveFormReact: ControllerHandler = (_req, res) => {
...
+      .replace(new RegExp('__OG_TITLE__', 'g'), 'FormSG')
+      .replace('__OG_DESCRIPTION__', '')
+  } else {
+    const { title, description } = createMetatagsResult.value
+    reactHtml = reactHtml
+      // there are multiple __OG_TITLE__ and we are unable to use string.replaceAll so we use regexp
+      .replace(new RegExp('__OG_TITLE__', 'g'), title)
+      .replace('__OG_DESCRIPTION__', description ?? '')
+      .replace('<title>FormSG</title>', `<title>${title}</title>`)
+      .replace('__OG_IMAGE__', 'og-img-metatag-publicform.png')
```
timotheeg: It feels like the code to set the OG tags are duplicated, it might be nice to have a refactor with a pipelining approach instead. A given GET flow basically gets the og data, and eventually returns the file with replacement. For all get flows, the last step of the processing pipeline should be the same code 🤔

timotheeg (changes requested):

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -38,7 +43,67 @@ export const ADMIN_COOKIE_OPTIONS_WITH_EXPIRY = {
...
+          unknown,
+          unknown,
+          Record<string, string>
+        >,
+        res,
+        next,
+      )
+    }
+
+    let reactHtml = readFileSync(path.join(reactFrontendPath, 'index.html'), {
```
timotheeg: ⚠️ We should not do `readFileSync()` in code flows triggered from user inputs!! ⚠️ 

Use the async `fs.readFile()` instead and if you want the promise version of it, transform it with `Util.promisy()`

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -38,8 +43,86 @@ export const ADMIN_COOKIE_OPTIONS_WITH_EXPIRY = {
...
+  description,
+  image,
+}: MetaTags): string => {
+  return (
+    reactHtml
+      // there are multiple __OG_TITLE__ and we are unable to use string.replaceAll so we use regexp
+      .replace(new RegExp('__OG_TITLE__', 'g'), title)
+      .replace('<title>FormSG</title>', `<title>${title}</title>`)
+      .replace('__OG_DESCRIPTION__', description)
+      .replace('__OG_IMAGE__', image)
```
timotheeg: Some comments:
* no need for the comment 😅 
* don't use the `RegExp()` constructor unless you are constructing the regex pattern from variables, for static patterns, use the slash-slash syntax instead
* ⚠️ there are also multiple `__OG_DESCRIPTION__` and `__OG_IMAGE__` in the source file, all regexps should use the `global` flag on for all replacements
* Since we are using the same title value for the page main `<title>` tag, might as well replace the source `<title>FormSG</title>` into `<title>__OG_TITLE__</title>`, and remove the title-specific replacement statement

```suggestion
      .replace(/__OG_TITLE__/g, title)
      .replace(/__OG_DESCRIPTION__/g, description)
      .replace(/__OG_IMAGE__/g, image)
```

Also just noting: this is doing html injection in text nodes and attributes. We know from the values below that the entries are static and html-safe (thankfully! - meaning no markup of xss injection possible), but we should consider doing html-encoding for safety anyway in this function.

Ref: [OWASP recommendations on html and attribute encoding](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#RULE_.231_-_HTML_Escape_Before_Inserting_Untrusted_Data_into_HTML_Element_Content)

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -49,24 +132,45 @@ const serveFormReact: ControllerHandler = (_req, res) => {
...
+      image: 'og-img-metatag-publicform.png',
+    })
+  }
+
   return (
     res
       // Prevent index.html from being cached by browsers.
       .setHeader('Cache-Control', 'no-cache')
-      .sendFile(path.join(reactFrontendPath, 'index.html'))
+      .send(reactHtmlWithMetaTags)
```
timotheeg: This file still contains 3 calls to `replaceWithMetaTags()`, and 2 calls `.setHeader('Cache-Control', 'no-cache')` 🤔 . I reckon the flow can be improved with a processing pipeline where the different conditional flows only collect the meta tag data, but there's a single shared final action to set the header and send the body.

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -38,8 +43,86 @@ export const ADMIN_COOKIE_OPTIONS_WITH_EXPIRY = {
...
+      message: 'serveFormReact',
+      meta: {
+        action: 'routeReact.serveFormReact',
+        __dirname,
+        cwd: process.cwd(),
+        reactFrontendPath,
+      },
+    })
+
+    if (isPublicForm && !!get(req.params, 'formId')) {
```
timotheeg: /nit Unless the linter is complaining, the `!!` is not necessary in a pure evaluation (i.e. the evaluation is not being assigned to a variable of type boolean)

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -38,25 +42,65 @@ export const ADMIN_COOKIE_OPTIONS_WITH_EXPIRY = {
...
+const replaceWithMetaTags = ({
+  title,
+  description,
+  image,
+}: MetaTags): string => {
+  return reactHtml
+    .replace(/(__OG_TITLE__)/g, DOMPurify.sanitize(title))
+    .replace(
+      /(<title>__OG_TITLE__<\/title>)/g,
+      DOMPurify.sanitize(`<title>${title}</title>`),
```
timotheeg: ⚠️ That looks incorrect: the `<title> tag **is** markup on purpose, so it should not be sanitized. Only the title itself shoud be sanitized.

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -38,25 +42,65 @@ export const ADMIN_COOKIE_OPTIONS_WITH_EXPIRY = {
...
+    )
```
timotheeg: Suggestion: in index.html, replace `<title>FormSG</title>` by `<title>__OG_TITLE__</title>`, and reduce the replacement code to this:

```suggestion
    .replace(/(__OG_TITLE__)/g, DOMPurify.sanitize(title))
```

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -38,25 +42,61 @@ export const ADMIN_COOKIE_OPTIONS_WITH_EXPIRY = {
...
+}
+const replaceWithMetaTags = ({
+  title,
+  description,
+  image,
+}: MetaTags): string => {
+  return reactHtml
+    .replace(/(__OG_TITLE__)/g, DOMPurify.sanitize(title))
+    .replace(/(__OG_DESCRIPTION__)/g, DOMPurify.sanitize(description))
+    .replace(/(__OG_IMAGE__)/g, DOMPurify.sanitize(image))
```
timotheeg: Few extra minor nits and questions (sorry! 😅 )
* brackets are not needed in the regex patterns
* Does `DOMPurify.sanitize()` have different mechanisms to encode raw markup on one hand, and attributes on the other hand? some of the markup injection we are doing here is markup (`<title>` tag) while the rest are attributes.
* Sanitization might be over kill for this (not sure 🤔), and [html-escaping](https://www.npmjs.com/package/html-escaper) might have been enough

BTW I realized my previous comment was wrong: considering the form title is user-supplied, we don't know what content it will have exactly, so escaping or sanitising was indeed the right thing to do 👍

timotheeg: Just adding a comment from our earlier discussion. In this use case, doing escaping is better than sanitizing, because the values are not meant to contain HTML at all.

So last TODO for this PR is to replace `DOMPurify.sanitize()` with `html-escaper.escape()`

timotheeg (approved):
LGTM

timotheeg:
Looks like the PR needs a rebase, or merge from `develop`... There's an install conflict on `@types/http-error`

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -38,25 +42,61 @@ export const ADMIN_COOKIE_OPTIONS_WITH_EXPIRY = {
...
+      tags = await getPublicFormMetaTags(get(req.params, 'formId') ?? '')
+    }
+
+    const reactHtmlWithMetaTags = replaceWithMetaTags(tags)
+
+    return (
+      res
+        // Prevent index.html from being cached by browsers.
+        .setHeader('Cache-Control', 'no-cache')
+        .sendFile(reactHtmlWithMetaTags)
```
timotheeg: This looks wrong now 🤔 

`reactHtmlWithMetaTags` is not a path to a file.

timotheeg:
One more late comment (that we can fix later, soon-ish):, we shouldn't put the static assets at the root of `/frontend/public`. Instead, let's put assets under the subfolder `/frontend/public/static`, and we can have further subfolders if we need.

The main reason I ask this is so, if we wanted, we could eventually have path-based routing very early (say on CF itself) on `/static/`. When assets are directly under `/`, it's harder to split traffic early.

wanlingt:
> One more late comment (that we can fix later, soon-ish):, we shouldn't put the static assets at the root of `/frontend/public`. Instead, let's put assets under the subfolder `/frontend/public/static`, and we can have further subfolders if we need.
> 
> The main reason I ask this is so, if we wanted, we could eventually have path-based routing very early (say on CF itself) on `/static/`. When assets are directly under `/`, it's harder to split traffic early.

ok! i'll open a new PR for this - #5338

PR #5330, opened 2022-11-03
fix: decimal field validation allows 0 to be the min or max

wanlingt:
## Problem

Frontend validation checks for the decimal field were failing when 0 was set as the min or max value.

## Solution

As identified by @justynoh , when 0 was set as the min/max value, it was not being registered by the checks since this was based on a boolean. To solve this, set `customMax` / `customMin` `!== null` instead of relying on just `customMax` / `customMin`

## Screenshots
*Before*
image

## Tests

- [ ] Create a decimal field. Set 0 as the min value. In the public form, enter -5. A validation message should pop up and you should not be able to submit the form. 
- [ ] Create a decimal field. Set 0 as the max value. In the public form, enter 5. A validation message should pop up and you should not be able to submit the form.

### DIFF
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -250,22 +250,21 @@ export const createDecimalValidationRules: ValidationRuleFn<
       if (!validateByValue) return true
 
       if (
-        customMin &&
-        customMax &&
+        customMin !== null &&
+        customMax !== null &&
         (numVal < customMin || numVal > customMax)
       ) {
         return `Please enter a decimal between ${formatNumberToLocaleString(
           customMin,
         )} and ${formatNumberToLocaleString(customMax)} (inclusive)`
       }
-
-      if (customMin && numVal < customMin) {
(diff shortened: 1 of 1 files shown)


frontend/src/utils/fieldValidation.ts
```diff
@@ -250,22 +250,21 @@ export const createDecimalValidationRules: ValidationRuleFn<
...
+        customMax !== null &&
         (numVal < customMin || numVal > customMax)
       ) {
         return `Please enter a decimal between ${formatNumberToLocaleString(
           customMin,
         )} and ${formatNumberToLocaleString(customMax)} (inclusive)`
       }
-
-      if (customMin && numVal < customMin) {
+      if (customMin !== null && numVal < customMin) {
```
justynoh, line 261: supernit: what happened to the whitespace HAHAHA

justynoh (approved):

frontend/src/utils/fieldValidation.ts
wanlingt: what whitespace

PR #5351, opened 2022-11-08
fix: add frontend and backend validation for empty field titles

### LINKED ISSUE #5345: Ensure that form field names cannot be empty
React allows an admin to type a space for a field title, which is allowed to be saved in a form definition, but is incorrect. 

Angular did not allow this behaviour.

Either way service-side validation is missing. 

Forms with empty field titles should not exist in DB. Whichever are there need to be cleared (maybe with admins notified?)

cc @justynoh

justynoh:
## Problem

We did not perform empty string validation for field titles correctly. In particular, field titles were not trimmed before being validated.

Closes #5345 

## Solution
Backend: Add `.trim()` to validator for field creation and updation middlewares.

Frontend (react): Add custom validator for required fields to trim input before checking non-emptiness. This required a bit of refactoring for the entire field validation util file, so we will need to retest all the validators again, especially for public forms.

**Breaking Changes** 

## Tests

Backend
- [x] PUT a field with `"title":"field title"`. This should succeed, and the field should appear in the form with title `field title`.
- [x] PUT a field with `"title":"  field title "`. This should succeed, and the field should appear in the form with title `field title`.
- [x] PUT a field with `"title":""`. This should fail with status code 400. The field should not appear in the form.
- [x] PUT a field with `"title":"  "`. This should fail with status code 400. The field should not appear in the form.

Sample failure and success cases:
image

Frontend
- [x] Attempt to create a field with title just containing a single space. The frontend should display an error that the title field is required.
- [x] Attempt to create a field with title ` field title  `. This should succeed, and the field title should be `field title`.
- [x] Test that the validators for all fields work correctly in public forms
   - yes/no: required
   - home no, mobile no, email: required, valid home no / mobile no / email entered*
   - number, decimal: required, custom min/max value
   - short text, long text: required, custom min/max/exact length*
   - uen, nric: required, valid uen/nric entered
   - checkbox: required, custom min/max number of options
   - date: required, custom date validations
   - radio: required
   - tables: columns required (both short text and dropdowns)

* It's important in this case to check that the behavior re: leading/trailing whitespace should continue to be the same as current behavior. (e.g. does `a b c ` count as 5 or 6 characters, for short/long text input length custom validation)

### DIFF
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -75,12 +75,19 @@ type ValidationRuleFnEmailAndMobile<T extends FieldBase = FieldBase> = (
   schema: MinimumFieldValidationPropsEmailAndMobile<T>,
 ) => RegisterOptions
 
-const createRequiredValidationRules = (
+const requiredValidationFn =
+  (schema: Pick<FieldBase, 'required'>) => (value: unknown) => {
+    if (!schema.required) return true
+    // Trim strings before checking for emptiness
+    const trimmedValue = typeof value === 'string' ? value.trim() : value
+    return !!trimmedValue || REQUIRED_ERROR
+  }
+
+const createRequiredInValidationRules = (
   schema: Pick<FieldBase, 'required'>,
-): RegisterOptions['required'] => {
+): RegisterOptions['validate'] => {
   return {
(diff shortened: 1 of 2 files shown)


wanlingt (approved):
lgtm!

PR #5360, opened 2022-11-09
fix: revive pressing enter to request and submit OTP

### LINKED ISSUE #5340: pressing "enter" on the login page does not submit email/otp
Go to the login page and enter your email. Press "enter". It should submit the email and request for the otp, but it doesn't anymore. Users now have to actually click the "Sign in" button to request for the otp/submit the otp.

Regression probably introduced in #5175, we should prevent default on the public form submit button instead. (The thread in the PR says we should not due to a11y, but I discussed with KR in person and we're not sure if it actually does affect a11y, should test whether it indeed affects a11y.)

justynoh:
## Problem
Because we prevented default on all inputs so that public forms do not get reloaded upon submission, this also removed defaults for forms such as the admin logins.

Closes #5340 

## Solution
Add `preventDefaultOnEnter` prop to `Input` and `NumberInput` components, and use it where necessary to prevent form reload on enter for public forms. User input that is not used in public forms (e.g. admin login flow, emergency contact modal, mobile and email otp verification on public forms) are restored to previous functionality (where pressing "Enter" does the expected thing).

**Breaking Changes** 


wanlingt (approved):
lgtm!

PR #5364, opened 2022-11-09
fix: augment admin logic form fields data with myinfo values

### LINKED ISSUE #5363: Logic unable to be created for dropdown MyInfo fields
Create a MyInfo field based on dropdown (e.g. Marital status)
Go to Logic
Try to create a logic based on it
It will work in angular but not react

😢

justynoh:
## Problem
Logic on dropdown myinfo fields is not working :(

Closes #5363 

## Solution
We need to augment the fields with MyInfo data before it can be used by the logic builder.

**Breaking Changes** 


wanlingt (approved):
lgtm!

PR #130, opened 2020-08-13
feat: log IP, submissionId and formId together

mantariksh:
## Problem

If we have security incidents involving potentially malicious IP addresses, it is important to be able to trace the IP back to specific submission IDs. This is currently not possible as we do not log the IP together with the form and submission IDs.

Closes #129 

## Solution

Add the form ID and IP address to our submission log string.

### Before
image
### After
image

PR #5369, opened 2022-11-10
feat: use 2 nodemailer clients

wanlingt:
## Problem

Closes [#126](https://github.com/opengovsg/formsg-private/issues/126)

## Solution

- Create 2 nodemailer clients - 1 using SG env vars, another using US env vars
- Conditionally send mail from SG client depending on a rollout threshold `NODEMAILER_THRESHOLD_SG`

**Breaking Changes** 

## Tests

- [x] Before sending emails: check the SES dashboard (US and SG) to see the count of emails sent. Trigger the sending of multiple emails (e.g. OTPs). These emails should be sent successfully. Check the cloudwatch logs to determine which nodemailer client is being used. Check the SES dashboard (US and SG) to see if the new count of emails tallies with the cloudwatch logs.

```
fields @timestamp, @message
| filter label like 'mail'
| sort @timestamp desc
```

## Deploy Notes

- `SES_PORT_US `
- `SES_HOST_US`
- `SES_USER_US`
- `SES_PASS_US`
- `SES_PORT_SG`
- `SES_HOST_SG`
- `SES_USER_SG`
- `SES_PASS_SG`
- `NODEMAILER_SG_WARMUP_START_DATE`: Date to start rollout of emails from SG nodemailer client. Should be in the format `YYYY-MM-DDTHH:MM:SS+08:00`

The permissions for the SES sending policy should also be modified accordingly to allow both US and SG nodemailer clients to send emails


__tests__/e2e/setup/.test-env
```diff
@@ -67,6 +67,12 @@ SESSION_SECRET=sandcrawler-138577
 SES_PORT=1025 
 SES_HOST=0.0.0.0
 
+# Remove these when SES migration is over (opengovsg/formsg-private#130)
+SES_PORT_US=1025
+SES_HOST_US=0.0.0.0
+SES_PORT_SG=1025
+SES_HOST_SG = 0.0.0
```
timotheeg: Inconsistent (invalid?) syntax 
```suggestion
SES_HOST_SG=0.0.0.0
```

src/app/services/mail/mail.service.ts
```diff
@@ -157,13 +164,24 @@ export class MailService {
       })
 
       try {
+        let sendMailFromSG = false
+        const rand = Math.random() * 100
+        sendMailFromSG = rand <= config.nodemailer_client_threshold_sg
```
timotheeg: /nit: there's no need to initialize `sendMailFromSG` as a non-const with a default. The assignment is systematic and guaranteed to yield a boolean:
```suggestion
        const rand = Math.random() * 100
        const sendMailFromSG = rand <= config.nodemailer_client_threshold_sg
```

timotheeg: Also, since we have APM traces, could you try adding a tag of the US/SG decision into the trace span? That might allow easier searching and graphing later. Maybe something like this?

```typescript
const span = tracer.scope().active()
if (span) span.setTag('ses.region', sendMailFromSG ? 'sg', 'us')
```

I'm not sure if that should be on the outer span, or in the `nodemailer/sendMail` span btw 🤔 , Try with the outer span first, it might make searches easier (am not sure).

Documentation [here](https://docs.datadoghq.com/tracing/guide/add_span_md_and_graph_it)

timotheeg: ⚠️ This line should use `<` instead of `<=`. The reason for that is `Math.random()` returns values in the range `[0,1[` (greater than or equal to 0 and less than 1 - [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random))

A strict less-than operator will ensure a threshold of zero will indeed always yield evaluations as `false`, and a threshold of `100` will always yield evaluations as `true`.

I don't remember if we verified that for our react migration threshold checks btw 😰... 

In practice it probably doesn't matter all that much because getting `0` exactly from `Math.random()` practically never happens. But still, for code correctness, the check operation should be `<` instead of `<=`.

__tests__/e2e/setup/.test-env
justynoh: OMG. this is why the tests are failing. thanks for catching this!!

src/app/services/mail/mail.service.ts
justynoh: fixed!

src/app/config/schema.ts
```diff
@@ -436,6 +437,39 @@ export const prodOnlyVarsSchema: Schema<IProdOnlyVarsSchema> = {
...
+    env: 'SES_PASS_SG',
+    sensitive: true,
+  },
+  // Remove this when SES migration is over (opengovsg/formsg-private#130)
+  nodemailer_client_threshold_sg: {
+    doc: 'Percentage threshold to send mail from SG nodemailer client',
+    format: 'int',
+    default: 0,
+    env: 'NODEMAILER_THRESHOLD_SG',
+  },
```
timotheeg: Apologies, I'm making a change to the requirement 😅😰.

This threshold is like the thresholds we had for the react migration. In react, setting the threshold value manually made sense, because we wanted high-control deliberate steps, and no change over the weekends.

For SES, the warm up would be better served with a gradual and continuous increase over the fixed warm-up period of 6 weeks (see doc [here](https://docs.aws.amazon.com/ses/latest/dg/dedicated-ip-warming.html)). 

Therefore, I'd recommend new requirement/behaviour as follows:
1. env variable represents whether the split is active at all (so we can turn it off if needed)
2. env var value is a date of start of warm up as `YYYY-MM-DDTHH:MM:SS+08:00`
3. hardcode 6 WEEKS as a constant in code
4. derive threshold value from a function (which basically embeds the rollout strategy)

Example linear growth (pseudo code);
```javascript
const START_TS = (new Date(env.SES_US_SG_START_DATE)).getTime()
const WARM_UP_DURATION = 6 * 7 * 24 * 60 * 60 * 1000; // 6 weeks

// return threshold on [0, 1]
function getThreshold() {
    const now = Date.now()
    const elapsed = Math.max(0, Math.min(now - START_TS, WARM_UP_DURATION))
    const linear = elapsed / WARM_UP_DURATION

    return linear
}
```

Or quadratic growth (slower start, faster end):
```javascript
const START_TS = (new Date(env.SES_US_SG_START_DATE)).getTime()
const WARM_UP_DURATION = 6 * 7 * 24 * 60 * 60 * 1000; // 6 weeks

// return threshold on [0, 1]
function getThreshold() {
    const now = Date.now()
    const elapsed = Math.max(0, Math.min(now - START_TS, WARM_UP_DURATION))
    const linear = elapsed / WARM_UP_DURATION

    return linear * linear
}
```

timotheeg: The system above can then be plugged in to the check that's already done in the PR:
```javascript
if (Math.random() < getThreshold()) {
    // SES SG
}
else {
    // SES US
}
```

wanlingt: ok! this approach does make more sense, will make the change

src/app/services/mail/mail.service.ts
```diff
@@ -157,13 +167,36 @@ export class MailService {
...
-        )
+        const getThreshold = () => {
+          const now = Date.now()
+          const elapsed = Math.max(
+            0,
+            Math.min(now - START_TS, WARM_UP_DURATION),
+          )
+          const linear = elapsed / WARM_UP_DURATION
+          return linear * linear
+        }
```
timotheeg: /nit this function is not dependent on the call context (it only relies on `START_TS` and `WARM_UP_DURATION`, which are module constants), so the function itself can be declared just once at module level.

And then maybe just for extra clarity the function can be renamed `getWarmUpThreshold()` 😅

src/app/config/schema.ts
```diff
@@ -436,6 +437,39 @@ export const prodOnlyVarsSchema: Schema<IProdOnlyVarsSchema> = {
...
+    format: String,
+    default: null,
+    env: 'SES_PASS_SG',
+    sensitive: true,
+  },
+  // Remove this when SES migration is over (opengovsg/formsg-private#130)
+  nodemailer_sg_warmup_start_date: {
+    doc: 'Date where SG nodemailer client will start sending emails',
+    format: String,
+    default: '',
```
timotheeg: There should either not be a default, or the default should be some date ridiculously in the future so the default behaviour is to be at rollout threshold value of 0 (e.g. `2099-01-01T23:59:59+08:00`)

timotheeg: Well... Turns out that in practice an invalid date does have the behaviour of threshold `0`, but I reckon it's by accident rather than a conscious choice, correct? 😅 

As in: if the env var is not representing a valid date, `START_TS` will be an `Invalid Date` with a `getTime()` value of `NaN`. 

The calculations based on that in `getThreshold()` will eventually return `NaN` as well, and an interesting fact of `NaN` is that comparisons to it are **always false** (i.e. `1 < NaN` is false, but `NaN < 1` is also false). Therefore `rand < NaN` will always be false, and in effect that represents a rollout threshold of `0`. 

So again, it works, but it's dirty. It would be better to make the behaviour explicit, and have an explicit handling for `if (isNaN(START_TS)) return 0` as first line of the function `getThreshold()`.

src/app/config/config.ts
```diff
@@ -155,8 +156,56 @@ const mailConfig: MailConfig = (function () {
...
+
+  return {
+    mailFrom,
+    official,
+    mailer,
+    transporter,
+  }
+})()
+
+// Use this mail config when SES migration is over, after renaming env vars back to without _sg suffix (opengovsg/formsg-private#130)
```
tshuli: Add TODO(#130) prefix, so that it is easier to find all the TODOs in the codebase
```suggestion
// TODO(#130): Use this mail config when SES migration is over, after renaming env vars back to without _sg suffix (opengovsg/formsg-private#130)
```

src/app/config/schema.ts
```diff
@@ -382,29 +382,30 @@ export const optionalVarsSchema: Schema<IOptionalVarsSchema> = {
 }
 
 export const prodOnlyVarsSchema: Schema<IProdOnlyVarsSchema> = {
-  port: {
+  // Delete these env vars when SES migration is over (opengovsg/formsg-private#130)
```
tshuli: Same here, add TODO (and rest of the PR)

tshuli (approved):
lgtm, other than Tim's comments and a minor nit

src/app/services/mail/mail.service.ts
wanlingt: [c77a6cd](https://github.com/opengovsg/FormSG/pull/5369/commits/c77a6cd2f529c179152ed3f9be395ab139d57cdf)

src/app/config/schema.ts
wanlingt: [c77a6cd](https://github.com/opengovsg/FormSG/pull/5369/commits/c77a6cd2f529c179152ed3f9be395ab139d57cdf)

src/app/config/config.ts
wanlingt: no wonder, I was thinking it looked a bit odd but couldn't figure out why. thanks! [d167864](https://github.com/opengovsg/FormSG/pull/5369/commits/d167864734f46ad3fa204bc4db5d370b8bcc6ce8)

src/app/services/mail/mail.service.ts
timotheeg: Hey? I'm not seeing that change in the commit? did you forget to add the file? 😅

timotheeg: nvm, I see them in the second "TODO" commit, oops!

src/app/services/mail/mail.service.ts
```diff
@@ -64,6 +64,21 @@ const DEFAULT_RETRY_PARAMS: MailServiceParams['retryParams'] = {
...
 
+// TODO #130 Delete references to US SES when SES migration is over (opengovsg/formsg-private#130)
+const START_TS = new Date(config.nodemailer_sg_warmup_start_date).getTime()
+const WARM_UP_DURATION = 6 * 7 * 24 * 60 * 60 * 1000 // 6 weeks
+
+const getWarmUpThreshold = () => {
+  // if START_TS is an empty string, set threshold to 0
+  if (isNaN(START_TS)) {
+    return 0
+  }
```
timotheeg: Just for reference: this safety check is not just to catch empty strings, but to catch any input error in our config. Any config input (accidentally) not parsable as an ISO Date would be caught here. 

This should never happen obviously (with config update pairing and all that), but it's just simple defensive programming that can catch "simple" mistakes:

For example:
```javascript
> (new Date(' 2022-11-15T02:47:30')).getTime() // accidental leading space
NaN

> (new Date('2022-11-15T02:47:30 ')).getTime() // accidental trailing space
NaN

> (new Date('2022-11-15T02:47:30+08')).getTime() // invalid timezone suffix definition
NaN
```

timotheeg (approved):
LGTM 👍

PR #5373, opened 2022-11-10
fix: make custom field validations consistent between angular and react

### LINKED ISSUE #5372: Make field validation behavior consistent between angular and react
Some field validation behavior is inconsistent between angular and react. In particular, for short text, the length validation performed in angular occurs after trimming the input, whereas in react it is before. We should check if the validations are consistent.

justynoh:
## Problem
Some field validation are inconsistent between angular and react. 

Closes #5372 

## Solution
In angular, we always trim text fields before validation and submission, so this PR implements this for react.

Additionally, I snuck in a quick change to the `Input` component which had a small bug before. Sorry!

**Breaking Changes** 

## Tests

- [x] Create a short/long text field with some custom length validation. Then, fill the form field with some leading/trailing spaces. The character count should not count the leading/trailing spaces.
- [x] Submit the form with the text field containing leading/trailing spaces, and look at the sent payload from the Network panel. The sent answer should be trimmed, and contain the correct number of characters that the custom validation requested for.

### DIFF
--- a/frontend/src/components/Input/Input.tsx
+++ b/frontend/src/components/Input/Input.tsx
@@ -38,33 +38,43 @@ export const Input = forwardRef<InputProps, 'input'>((props, ref) => {
     'preventDefaultOnEnter',
   ])
 
-  const chakraInputProps: ChakraInputProps = useMemo(
-    () => ({
-      ref,
-      ...inputProps,
-      sx: props.sx ?? inputStyles.field,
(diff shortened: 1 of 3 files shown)


justynoh:
Separately, should we implement backend validation for the custom validations set by admins? I don't see it anywhere, unless I'm blind...

timotheeg (approved):
lgtm

PR #5376, opened 2022-11-13
fix: local dev

darrelhong:
Changes:

- copy 'shared' project package.json files to container so it doesn't break install
- update frontend start command with  --openssl-legacy-provider flag
- don't install frontend dependencies by default in postinstall as it is not required in local dev. Install frontend dependencies production dockerfile


tshuli:
Thanks @darrelhong for your contributions 🙏! Made minor edits in #5379 to install the frontend deps, which are required for our CI

PR #5377, opened 2022-11-13
feat: use-template for forms

### LINKED ISSUE #5161: [Review] Use-template feature for forms on react

wanlingt:
## Problem

Closes #5161
## Solution

**Breaking Changes** 

**Features**:
- Public forms can be used as a template

**Details**:
- Added the use-template link to `ShareFormModal`. This input field is disabled if the form is closed.
- Added a new component `TemplateFormPage`, which is similar in structure to `PreviewFormPage`. This also uses a new `TemplateFormProvider`, which only differs from `PreviewFormProvider` in the query used (`useFormTemplate`)
- Use `DuplicateFormModal` to duplicate the template form. The `formId` is now injected into `DupeFormWizardProvider` at a higher level, so that `DuplicateFormModal` can be used across workspace rows and the template form page.
- Use `z-index` in form page's `MiniHeader` instead of `Portal` to enable stacking of the sticky headers

**Improvements**:

- Updated the design of `PreviewFormPage`

## Before & After Screenshots

**AFTER**:

https://user-images.githubusercontent.com/56983748/202135080-868110c0-1804-45b3-847c-894eafc6803c.mov

## Tests

New stories have been added.

Use-template:
- [x] As a logged in admin, visit the use-template endpoint of a public form `admin/form/:formId/use-template`. Click on the button 'Use this template'. The form title should be prefixed with '[Template]' and you should be able to duplicate the form successfully.
- [x] As a logged in admin, visit the use-template endpoint of a **public** form that you're not the owner of. You should be able to create a duplicate of the form template successfully.
- [x] As a logged in admin, visit the use-template endpoint of a **private** form that you're not the owner of. You should see the admin forbidden error page.
- [x] As a public user, visit the use-template endpoint of a public form. You should be redirected to the login page.

### DIFF
--- a/frontend/src/features/admin-form/common/AdminViewFormService.ts
+++ b/frontend/src/features/admin-form/common/AdminViewFormService.ts
@@ -9,7 +9,7 @@ import {
   SmsCountsDto,
 } from '~shared/types/form/form'
 
-import { PUBLICFORM_USETEMPLATE_ROUTE } from '~constants/routes'
+import { ADMINFORM_USETEMPLATE_ROUTE } from '~constants/routes'
 import { transformAllIsoStringsToDate } from '~utils/date'
 import { ApiService } from '~services/ApiService'
 
@@ -80,7 +80,7 @@ export const viewFormTemplate = async (
   formId: string,
 ): Promise<PreviewFormViewDto> => {
   return ApiService.get<PreviewFormViewDto>(
-    `${ADMIN_FORM_ENDPOINT}/${formId}/${PUBLICFORM_USETEMPLATE_ROUTE}`,
+    `${ADMIN_FORM_ENDPOINT}/${formId}/${ADMINFORM_USETEMPLATE_ROUTE}`,
   )
     .then(({ data }) => {
       // Add default mock authenticated state if previewing an authenticatable form
--- a/frontend/src/features/admin-form/common/AdminViewFormService.ts
+++ b/frontend/src/features/admin-form/common/AdminViewFormService.ts
@@ -78,7 +78,9 @@ export const previewForm = async (
 export const viewFormTemplate = async (
   formId: string,
 ): Promise<PreviewFormViewDto> => {
-  return ApiService.get<PreviewFormViewDto>(`${formId}/use-template`)
+  return ApiService.get<PreviewFormViewDto>(
+    `${ADMIN_FORM_ENDPOINT}/${formId}/use-template`,
+  )
     .then(({ data }) => {
       // Add default mock authenticated state if previewing an authenticatable form
       // and if server has not already sent back a mock authenticated state.
--- a/frontend/src/features/admin-form/common/AdminViewFormService.ts
+++ b/frontend/src/features/admin-form/common/AdminViewFormService.ts
@@ -69,6 +69,33 @@ export const previewForm = async (
     .then(transformAllIsoStringsToDate)
 }
 
+/**
+ * Gets the public view of a form. Used for viewing the form from the form template page.
+ * Must be an admin.
+ * @param formId formId of form in question
+ * @returns Public view of a form
+ */
+export const viewFormTemplate = async (
+  formId: string,
+): Promise<PreviewFormViewDto> => {
+  return ApiService.get<PreviewFormViewDto>(`${formId}/use-template`)
+    .then(({ data }) => {
+      // Add default mock authenticated state if previewing an authenticatable form
+      // and if server has not already sent back a mock authenticated state.
+      if (data.form.authType !== FormAuthType.NIL && !data.spcpSession) {
+        data.spcpSession = { userName: PREVIEW_MOCK_UINFIN }
+      }
+
+      // Inject MyInfo preview values into form fields (if they are MyInfo fields).
+      data.form.form_fields = data.form.form_fields.map(
+        augmentWithMyInfoDisplayValue,
+      )
+
+      return data
+    })
+    .then(transformAllIsoStringsToDate)
+}
+
 export const getFreeSmsQuota = async (formId: string) => {
   return ApiService.get<SmsCountsDto>(
     `${ADMIN_FORM_ENDPOINT}/${formId}/verified-sms/count/free`,
--- a/frontend/src/features/admin-form/common/AdminViewFormService.ts
+++ b/frontend/src/features/admin-form/common/AdminViewFormService.ts
@@ -9,6 +9,7 @@ import {
   SmsCountsDto,
 } from '~shared/types/form/form'
 
+import { PUBLICFORM_USETEMPLATE_ROUTE } from '~constants/routes'
 import { transformAllIsoStringsToDate } from '~utils/date'
 import { ApiService } from '~services/ApiService'
 
@@ -79,7 +80,7 @@ export const viewFormTemplate = async (
   formId: string,
 ): Promise<PreviewFormViewDto> => {
   return ApiService.get<PreviewFormViewDto>(
-    `${ADMIN_FORM_ENDPOINT}/${formId}/use-template`,
+    `${ADMIN_FORM_ENDPOINT}/${formId}/${PUBLICFORM_USETEMPLATE_ROUTE}`,
   )
     .then(({ data }) => {
       // Add default mock authenticated state if previewing an authenticatable form
--- a/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
+++ b/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
@@ -1,29 +1,59 @@
 import { BiShow } from 'react-icons/bi'
 import { Link as ReactLink } from 'react-router-dom'
-import { Flex, Icon, Text } from '@chakra-ui/react'
+import { Flex, Icon, Stack, Text } from '@chakra-ui/react'
 
-import { ADMINFORM_ROUTE } from '~constants/routes'
+import { ADMINFORM_ROUTE, DASHBOARD_ROUTE } from '~constants/routes'
 import Button from '~components/Button'
+import Link from '~components/Link'
 
 import { usePublicFormContext } from '~features/public-form/PublicFormContext'
 
-export const PreviewFormBanner = (): JSX.Element => {
+interface PreviewFormBannerProps {
+  isTemplate?: boolean
+}
+
+export const PreviewFormBanner = ({
+  isTemplate,
+}: PreviewFormBannerProps): JSX.Element => {
   const { formId } = usePublicFormContext()
 
   return (
-    <Flex bg="primary.100" py="1rem" px="2rem" display="flex" width="100%">
+    <Flex
+      bg="primary.100"
+      py="1rem"
+      px="2rem"
+      display="flex"
+      width="100%"
+      // position="fixed"
+    >
       <Flex align="center" flex={1} justify="space-between" flexDir="row">
         <Flex align="center">
           <Icon aria-hidden as={BiShow} fontSize="1rem" mr="0.75rem" />
           <Text textStyle="subhead-3">Form Preview</Text>
         </Flex>
-        <Button
-          aria-label="Click to edit the form"
-          as={ReactLink}
-          to={`${ADMINFORM_ROUTE}/${formId}`}
-        >
-          Edit form
-        </Button>
+        {isTemplate ? (
+          <Stack spacing="1rem" direction="row">
+            <Link
+              variant="standalone"
+              aria-label="Click to return to the admin dashboard"
+              as={ReactLink}
+              to={DASHBOARD_ROUTE}
+            >
+              Back to FormSG
+            </Link>
+            <Button aria-label="Click to use this template">
+              Use this template
+            </Button>
+          </Stack>
+        ) : (
+          <Button
+            aria-label="Click to edit the form"
+            as={ReactLink}
+            to={`${ADMINFORM_ROUTE}/${formId}`}
+          >
+            Edit form
+          </Button>
+        )}
       </Flex>
     </Flex>
   )
--- a/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
+++ b/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
@@ -1,6 +1,16 @@
+import { useCallback } from 'react'
 import { BiShow } from 'react-icons/bi'
 import { Link as ReactLink } from 'react-router-dom'
-import { Flex, Icon, Stack, Text, useDisclosure } from '@chakra-ui/react'
+import { Waypoint } from 'react-waypoint'
+import {
+  Flex,
+  Icon,
+  Portal,
+  Slide,
+  Stack,
+  Text,
+  useDisclosure,
+} from '@chakra-ui/react'
 
 import { ADMINFORM_ROUTE, DASHBOARD_ROUTE } from '~constants/routes'
 import Button from '~components/Button'
@@ -9,6 +19,17 @@ import Link from '~components/Link'
 import { usePublicFormContext } from '~features/public-form/PublicFormContext'
 import { DuplicateFormModal } from '~features/workspace/components/DuplicateFormModal'
 
+export const StickyPreviewHeader = ({
+  isOpen,
+}: {
+  isOpen: boolean
+}): JSX.Element => (
+  <Portal>
+    <Slide aria-hidden direction="top" in={isOpen}>
+      <PreviewFormBanner isTemplate />
+    </Slide>
+  </Portal>
+)
 interface PreviewFormBannerProps {
   isTemplate?: boolean
 }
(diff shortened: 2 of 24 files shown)


frontend/src/features/admin-form/common/AdminViewFormService.ts
```diff
@@ -69,6 +69,35 @@ export const previewForm = async (
...
+ * Gets the public view of a form. Used for viewing the form from the form template page.
+ * Must be an admin.
+ * @param formId formId of form in question
+ * @returns Public view of a form
+ */
+export const viewFormTemplate = async (
+  formId: string,
+): Promise<PreviewFormViewDto> => {
+  return ApiService.get<PreviewFormViewDto>(
+    `${ADMIN_FORM_ENDPOINT}/${formId}/use-template`,
```
justynoh: Should we reference the `PUBLICFORM_USETEMPLATE_ROUTE` constant in `../../../constants/routes` instead?

frontend/src/features/admin-form/share/ShareFormModal.tsx
```diff
@@ -161,6 +161,32 @@ export const ShareFormModal = ({
                 </Stack>
               </Skeleton>
             </FormControl>
+            <FormControl isReadOnly>
+              <FormLabel isRequired useMarkdownForDescription>
```
justynoh: supernit: since we don't have a description, we probably don't need `useMarkdownForDescription` here haha

separately, actually, i wonder if we _should_ have a description that reminds admins that if their form is closed, other admins will not be able to use-template

frontend/src/features/admin-form/share/ShareFormModal.tsx
```diff
@@ -161,6 +161,32 @@ export const ShareFormModal = ({
...
+                    isReadOnly
+                    isDisabled={isFormPrivate}
+                    value={`${shareLink}/use-template`}
+                  />
+                  {formId ? (
+                    <InputRightElement>
+                      <CopyButton
+                        colorScheme="secondary"
+                        stringToCopy={`${shareLink}/use-template`}
+                        aria-label="Copy use-template form link"
```
justynoh: nit: For a11y, should we rephrase "use-template" in `aria-label`?

frontend/src/features/admin-form/share/ShareFormModal.tsx
```diff
@@ -161,6 +161,32 @@ export const ShareFormModal = ({
...
+                    data-chromatic="ignore"
+                    isReadOnly
+                    isDisabled={isFormPrivate}
+                    value={`${shareLink}/use-template`}
+                  />
+                  {formId ? (
+                    <InputRightElement>
+                      <CopyButton
+                        colorScheme="secondary"
+                        stringToCopy={`${shareLink}/use-template`}
```
justynoh: nit: should we factor out the link, and also reference the route constant again?

frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
```diff
@@ -1,39 +1,183 @@
...
-          to={`${ADMINFORM_ROUTE}/${formId}`}
-        >
+    <>
+      <Flex bg="primary.100" py="1rem" px="2rem" display="flex" width="100%">
+        <Flex align="center" flex={1} justify="space-between" flexDir="row">
           <Flex align="center">
-            <Icon aria-hidden as={BxsEditAlt} fontSize="1rem" mr="0.75rem" />
-            <Text textStyle="subhead-2">Edit form</Text>
+            <Icon aria-hidden as={BiShow} fontSize="1rem" mr="0.75rem" />
+            <Text textStyle="subhead-3">Template Preview</Text>
```
justynoh: Based on figma, templates should say "Template preview" but form preview should say "Form preview", hahah

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
```diff
@@ -53,7 +54,12 @@ export const useDupeFormWizardContext = (): CreateFormWizardContextReturn => {
       responseMode: containsMyInfoFields
         ? FormResponseMode.Email
         : FormResponseMode.Encrypt,
-      title: makeDuplicateFormTitle(previewFormData.form.title, dashboardForms),
+      title: makeDuplicateFormTitle(
+        isTemplate
+          ? `Template_${previewFormData.form.title}`
```
justynoh: not about your PR, but omggg why are we still prefixing with "Template_" so gross :(

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
```diff
@@ -14,17 +14,18 @@ import {
...
   const { data: dashboardForms, isLoading: isWorkspaceLoading } = useWorkspace()
-  const { activeFormMeta } = useWorkspaceRowsContext()
 
   const { data: previewFormData, isLoading: isPreviewFormLoading } =
     usePreviewForm(
-      activeFormMeta?._id ?? '',
+      formId,
       // Stop querying once submissionData is present.
-      /* enabled= */ !!activeFormMeta,
+      /* enabled= */ !!formId,
```
justynoh: because of this check, i wonder if actually we actually need `formId` to be provided. could we possibly make `formId` an optional argument, then in `WorkspaceRowsContext.tsx` we don't have to handle the undefined case?

```javascript
        formId={activeFormMeta?._id}

File: frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx [hunk 8]
```diff
@@ -14,17 +14,18 @@ import {
...
-export const useDupeFormWizardContext = (): CreateFormWizardContextReturn => {
+export const useDupeFormWizardContext = (
+  formId: string,
+  isTemplate: boolean,
+): CreateFormWizardContextReturn => {
   const { data: dashboardForms, isLoading: isWorkspaceLoading } = useWorkspace()
-  const { activeFormMeta } = useWorkspaceRowsContext()
 
   const { data: previewFormData, isLoading: isPreviewFormLoading } =
     usePreviewForm(
```
justynoh, line 28: are we sure that this works as intended if we only call `usePreviewForm` for the template feature? In the backend, `handlePreviewAdminForm` checks for the current user, while `handleGetTemplateForm` does not. So when a different admin tries to dupe the form, this might not work.

justynoh (changes requested):

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
wanlingt: @staceytan1998 any comments HAHA

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
wanlingt: i cri u are right

frontend/src/features/admin-form/share/ShareFormModal.tsx
wanlingt: [d147cf9](https://github.com/opengovsg/FormSG/pull/5377/commits/d147cf94f10e80bfad26efb58c95c08305369a6a)

frontend/src/features/admin-form/share/ShareFormModal.tsx
wanlingt: yes good catch! [d147cf9](https://github.com/opengovsg/FormSG/pull/5377/commits/d147cf94f10e80bfad26efb58c95c08305369a6a)

frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
wanlingt: [d147cf9](https://github.com/opengovsg/FormSG/pull/5377/commits/d147cf94f10e80bfad26efb58c95c08305369a6a)

frontend/src/features/admin-form/share/ShareFormModal.tsx
wanlingt: yeah good idea! i've removed `useMarkdownForDescription` for now in this commit [d147cf9](https://github.com/opengovsg/FormSG/pull/5377/commits/d147cf94f10e80bfad26efb58c95c08305369a6a)

@staceytan1998 shall we add the reminder description for admins? maybe something like 'If your form is closed, other admins will not be able to view a template of your form"

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
wanlingt: we'll change this to [Template] after we allow special characters in the form title

frontend/src/features/admin-form/share/ShareFormModal.tsx
wanlingt: added the reminder to the share form modal [ec6c92e](https://github.com/opengovsg/FormSG/pull/5377/commits/ec6c92e187fa1c976db8d50eb36868208e6c550b)

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
wanlingt: made the queries somewhat conditional here [db5601b](https://github.com/opengovsg/FormSG/pull/5377/commits/db5601b5a604971618758486ec619167b987d3b7), but not sure if it's the best way to do so, lmk what you think!

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
wanlingt: i left it as is as it's technically needed for in the queries `usePreviewForm(formId)` / `useTemplateForm(formId)`

frontend/src/features/admin-form/common/AdminViewFormService.ts
wanlingt: [d147cf9](https://github.com/opengovsg/FormSG/pull/5377/commits/d147cf94f10e80bfad26efb58c95c08305369a6a)

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
```diff
@@ -24,13 +27,23 @@ export const useDupeFormWizardContext = (
...
 
+  const { data: templateFormData, isLoading: isTemplateFormLoading } =
+    useFormTemplate(
+      formId,
+      // Stop querying if formId does not exist or if it's not in preview mode
+      /* enabled= */ !!formId || isTemplate,
+    )
+
+  const data = templateFormData || previewFormData
+  const isLoading = isTemplateFormLoading || isPreviewFormLoading
```
justynoh: Suggestion:

```javascript

const { data, isLoading } = isTemplate ? useFormTemplate(formId, !!formId) : usePreviewForm(formId, !!formId)

File: frontend/src/features/admin-form/template/TemplateFormProvider.tsx [hunk 10]
```diff
@@ -65,7 +65,8 @@ export const TemplateFormProvider = ({
 
   const isFormNotFound = useMemo(() => {
     return (
-      error instanceof HttpError && (error.code === 404 || error.code === 410)
+      error instanceof HttpError &&
```
justynoh, line 68: not about this line, but there's a linter complaint in this file:

```
Check warning on line 21 in frontend/src/features/admin-form/template/TemplateFormProvider.tsx

GitHub Actions / frontend_lint
'AdminForbiddenErrorPage' is defined but never used
```

justynoh (approved):
lgtm

frontend/src/features/admin-form/template/TemplateFormProvider.tsx
```diff
@@ -0,0 +1,188 @@
...
+  PublicFormContext,
+  SubmissionData,
+} from '~/features/public-form/PublicFormContext'
+import { useCommonFormProvider } from '~/features/public-form/PublicFormProvider'
+
+import { useTimeout } from '~hooks/useTimeout'
+import { HttpError } from '~services/ApiService'
+import { FormFieldValues } from '~templates/Field'
+
+import AdminForbiddenErrorPage from '~pages/AdminForbiddenError'
```
justynoh: there's a linter complaint here:

```
Check warning on line 21 in frontend/src/features/admin-form/template/TemplateFormProvider.tsx

GitHub Actions / frontend_lint
'AdminForbiddenErrorPage' is defined but never used
```

frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx
justynoh: haha as in, we can leave it undefined, then in the dupe form wizard provider do 

(but either way is ok la. is up to preference haha)

```javascript
const useDupeFormWizardContext = (formId = '', isTemplate) => ...

File: frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx (hunk 9 above)
Comment (wanlingt) [reply frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx]: cant do that :( `React Hook "useFormTemplate" is called conditionally. React Hooks must be called in the exact same order in every component render`

File: frontend/src/features/admin-form/template/TemplateFormProvider.tsx (hunk 11 above)
Comment (wanlingt) [reply frontend/src/features/admin-form/template/TemplateFormProvider.tsx]: ooh yes tq [bccfb1c](https://github.com/opengovsg/FormSG/pull/5377/commits/bccfb1c8c07b6fa6d7a8500c5beed4760959c675)

File: frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx (hunk 9 above)
Comment (justynoh) [reply frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx]: oh ya LOL lame, sure then hahaha that's fine

File: frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx (hunk 6 above)
Comment (wanlingt) [reply frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.tsx]: [3724a12](https://github.com/opengovsg/FormSG/pull/5377/commits/3724a12d3e816e7ee2988d401dfb0cf0915f1d72) updated yay

**Comment by wanlingt** [comment]
@justynoh rerequesting review for the routing and new form title validation portions!

File: frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx [hunk 12]
```diff
@@ -1,39 +1,197 @@
...
+  )
+
+  return (
+    <>
+      {isTemplate ? <StickyPreviewHeader isOpen={isOpen} /> : null}
+      <PreviewFormBanner isTemplate={isTemplate} />
+      {
+        /* Sentinel to know when sticky navbar is starting */
+
+        <Waypoint topOffset="64px" onPositionChange={handlePositionChange} />
```
timotheeg, line 193: I see we already have a Sentinel waypoint in the mini header. Do that and this preview banner ever show together? If yes do the waypoints work well together?

src/app/models/form.server.model.ts
```diff
@@ -163,10 +167,13 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
         type: String,
-        validate: [
-          /^[a-zA-Z0-9_\-./() &`;'"]*$/,
-          'Form name cannot contain special characters',
-        ],
+        validate: {
+          // validation fails if regex pattern exists in the title
+          validator: function (v: string) {
+            return !UNICODE_ESCAPED_REGEX.test(v)
+          },
```
timotheeg: /super nit: use shorter syntax for a simple function like this
```suggestion
          validator: v => !UNICODE_ESCAPED_REGEX.test(v),
```

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -136,7 +136,24 @@ const createFormValidator = celebrate({
       })
       .required()
       // Allow other form schema keys to be passed for form creation.
-      .unknown(true),
+      .unknown(true)
+      .custom((value, helpers) => {
+        // If there are unicode-escaped characters are not valid utf-8 encoded,
```
timotheeg: /nit grammar in this comment reads funny?

justynoh:
@wanlingt can I suggest that we merge the last two commits (allowing special characters in form titles and updating template prefix) in a separate PR instead? So that we get better separation of features if we have to rollback either one

src/app/models/form.server.model.ts
```diff
@@ -163,10 +167,13 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     {
       title: {
         type: String,
-        validate: [
-          /^[a-zA-Z0-9_\-./() &`;'"]*$/,
```
timotheeg: /justMusing: Ooohh, so we did allow some special characters even previously, we could have set the prefix as `(Template) `  and reduce the changeset 🤔

src/app/models/form.server.model.ts
timotheeg: More importantly: are you sure this test is correct here? `UNICODE_ESCAPED_REGEX` detects the incorrect utf8 characters by inspecting the json-encoded version of the value, but `v` here is the raw string value, no?

src/app/modules/form/admin-form/admin-form.middlewares.ts
```diff
@@ -26,5 +27,23 @@ export const updateSettingsValidator = celebrate({
...
+      // If this is saved into the db, an error is thrown when the driver attempts to read the db document as the driver interprets this as an escaped unicode sequence
+      // Since valid unicode-escaped characters will be processed correctly (e.g. \u00ae is processed as ®), they will not trigger an error
+      // Also note that if the user intends to input a 6-character string of the same form e.g. \udbbb, the backslash will be escaped (i.e. double backslash) and hence this will also not trigger an error
+
+      const valueStr = JSON.stringify(value)
+
+      if (UNICODE_ESCAPED_REGEX.test(valueStr)) {
+        return helpers.message({
+          custom:
+            'Please check that there are no improperly encoded characters in your input',
```
timotheeg: I wonder if folks would understand this error message 🤔 . Since we **know** there are invalid UTF8 characters in the input, we should probably say it outright.
```suggestion
            'There are invalid characters in your input',
```

timotheeg:
> @wanlingt can I suggest that we merge the last two commits (allowing special characters in form titles and updating template prefix) in a separate PR instead? So that we get better separation of features if we have to rollback either one

Good idea! This PR could do a simpler change of allowing 2 new characters in the title `[]` (by expanding the regex just a bit), and a secondary PR to allow unrestricted unicode content could follow. This whole process might take a bit more than  the 1 point we planned though (or was it zero even? 😅 )

wanlingt:
> @wanlingt can I suggest that we merge the last two commits (allowing special characters in form titles and updating template prefix) in a separate PR instead? So that we get better separation of features if we have to rollback either one

yes! good suggestion. I've reverted the last two commits in this PR, will open a separate PR for them :) so now the only change i've added is to allow square brackets in the title [f7f7c2c](https://github.com/opengovsg/FormSG/pull/5377/commits/f7f7c2caa24dd1f4fa7d914891c5e185831f8aec)

@timotheeg fyi too!

frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
wanlingt: yup, the sticky use-template preview banner and the mini header show up in use-template mode. they seem to work well together! the waypoints are specific to each element

frontend/src/features/admin-form/template/TemplateFormProvider.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const isAuthRequired = useMemo(
+    () =>
+      !!data?.form &&
+      data.form.authType !== FormAuthType.NIL &&
+      !data.spcpSession,
+    [data?.form, data?.spcpSession],
+  )
+
+  const { submitEmailModeFormMutation, submitStorageModeFormMutation } =
+    usePreviewFormMutations(formId)
```
karrui, line 105: We should not allow submission on templates - using the preview endpoint may fail if the form does not belong to the user. Just disable altogether imho

wanlingt: good point! opened a new pr [here](https://github.com/opengovsg/FormSG/pull/5443) to disable submission

PR #5379, opened 2022-11-14
fix: local dev

tshuli:
- Builds on #5376


justynoh (approved):
lgtm!

tshuli:
Thanks for @darrelhong for contributing in #5376

darrelhong:
Reason for not installing frontend deps by default is just to save some time when rebuilding local container

PR #5411, opened 2022-11-16
fix: update feedback modal and switch env functionality

### LINKED ISSUE #5271: Change modal for floating feedback button in React admin
Feedback about React admin currently can be provided through the infobox or the floating feedback button, and accessing the Switchback modal. Users only see the Switchback modal and give feedback when they intend to switch, so all the feedback we’ve been getting skews toward negative even though acceptance rate of React is high.

This design will provide users an avenue to provide general feedback that might not be related to switching back.

- Keep infobox and link to modal
- Change modal for floating feedback button
- Change copy for tooltip of that button

[Design here](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=13544%3A133966)

### LINKED ISSUE #5385: remove feedback form
When users wish to switch back to Angular, clicking the switch back option should bring them directly to Angular, without displaying the feedback form modal.
The switch env feedback form floating icon should be updated as mentioned in #5271

justynoh:
## Problem
We want to 
- Let admins switch back without filling the form, and
- Let admins give feedback using the switch env icon without switching.

Closes #5271 
Closes #5385 

## Solution
Small refactor to split up the hooks for the switch mutation and the feedback submission. Split up the admin and public feedback forms into separate modals, since they now are basically completely different (and we want to keep the admin modal around for the foreseeable future, unlike the public modal).

To accommodate this, communicated with kenneth on using a different form to collate admin feedback. He agreed to this, which means we need to inject the admin feedback id as well. Decided to use query params for the api endpoints to get the formid and submit the form for backward compatibility (see below).

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/202183189-741b6485-7e7d-4ff5-a4a9-9c0bfc79125a.mov

## Tests

Admins
- [x] Click the "switch back" banner, this should instantly switch users back to angular.
- [x] Click the feedback icon, it should open a modal with rating and feedback fields. Fill the form. It should submit successfully to the admin feedback form.

Public
- [x] Click the "switch back" banner, this should open a modal with long text field (and email field if not logged in as admin). Fill the form. It should submit successfully to the public feedback form.

## Deploy notes

**Environment variables**

Deprecated env vars
- `REACT_TO_ANGULAR_FEEDBACK_FORM_ID`: the single feedback form used by both public respondents and admins

New env vars
- `REACT_SWITCH_ENV_FEEDBACK_FORM_ID_ADMIN`: the admin feedback form
- `REACT_SWITCH_ENV_FEEDBACK_FORM_ID_RESPONDENT`: the respondent feedback form

**Endpoints**
- Updated `/forms/submissions/email/switchenvfeedback` and `/forms/switchenvfeedback` to take query param `view` which can be either `admin` or undefined (for backward compatibility). If the view is admin, we submit to the admin feedback form, otherwise we submit to the public feedback form. Reason is because we are reusing the old feedback form for the public feedback, but collecting admin feedback in a new form. This preserves backward compatibility for those who may submit their feedback on an old client which does not inject a query param.


frontend/src/features/workspace/components/AdminSwitchEnvMessage.tsx
```diff
@@ -63,7 +52,7 @@ export const AdminSwitchEnvMessage = (): JSX.Element => {
             as="u"
             variant="link"
             display="inline"
-            onClick={onOpen}
+            onClick={handleClick}
```
timotheeg, line 55: Just checking: is `handleClick` as a callback necessary? Could we just:
```suggestion
            onClick={adminSwitchEnvMutation.mutate}
```

timotheeg (approved):

wanlingt (commented):
(as discussed, leaving a comment for documentation) will the rating and feedback for admins be a long-term functionality? if yes it might be cleaner to create a new form + separate modal for this, instead of reusing the react feedback form

also, the modal header for admins should be 'Share your feedback'

frontend/src/features/workspace/components/AdminSwitchEnvMessage.tsx
justynoh: not necessary, but just wrapped it to match the naming conventions we usually use!

shared/types/env.ts
```diff
@@ -11,4 +11,5 @@ export interface SwitchEnvFeedbackFormBodyDto {
   email: string
   rumSessionId: string
   switchReason: string
+  rating: string
```
tshuli, line 21: as above comment

frontend/src/features/env/SwitchEnvFeedbackModal.tsx
```diff
@@ -146,85 +148,33 @@ export const SwitchEnvFeedbackModal = ({
...
+                      rules={{ required: 'This field is required' }}
+                      control={control}
+                      name={'rating'}
+                      render={({ field: { value, onChange, ...rest } }) => (
+                        <Rating
+                          numberOfRatings={5}
+                          variant={'star'}
+                          helperText={'1: Poor, 5: Excellent'}
+                          value={Number(value)}
+                          onChange={(val) => onChange(val?.toString())}
```
tshuli: qn: why are we casting this to string?

tshuli (approved):
lgtm with one clarification

frontend/src/features/env/SwitchEnvFeedbackModal.tsx
justynoh: the type of the `Rating` component and the `onChange` provided by the controller is different in that one allows undefined while the other does not, so we need it to get the types to work out 😓

justynoh:
going to re-request for reviews, since quite a lot has changed since last review

frontend/src/features/env/SwitchEnvFeedbackModal.tsx
justynoh: also, public form `RatingField` does the same thing oop hahaha

frontend/src/features/env/AdminFeedbackModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+      <ModalOverlay />
+      <ModalContent>
+        <ModalCloseButton />
+        <chakra.form noValidate onSubmit={handleSubmitForm}>
+          <ModalHeader pr="48px">Share your feedback</ModalHeader>
+          <ModalBody mt="0" pt="0">
+            <Stack spacing="1rem">
+              <FormControl>
+                <Input type="hidden" {...register('url')} value={url} />
+              </FormControl>
```
karrui: nit: probably don't need form control wrapping this since the input is hidden

frontend/src/features/env/AdminFeedbackModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+      isOpen={isOpen}
+      onClose={handleClose}
+      size={modalSize}
+      initialFocusRef={initialRef}
+    >
+      <ModalOverlay />
+      <ModalContent>
+        <ModalCloseButton />
+        <chakra.form noValidate onSubmit={handleSubmitForm}>
+          <ModalHeader pr="48px">Share your feedback</ModalHeader>
```
karrui: we should use `rem` measurements over `px`

```suggestion
          <ModalHeader pr="3rem">Share your feedback</ModalHeader>
```

frontend/src/features/env/AdminFeedbackModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+          <ModalBody mt="0" pt="0">
+            <Stack spacing="1rem">
+              <FormControl>
+                <Input type="hidden" {...register('url')} value={url} />
+              </FormControl>
+
+              <FormControl isRequired isInvalid={!isEmpty(errors)}>
+                <FormLabel>How was your experience using FormSG?</FormLabel>
+                <Controller
+                  rules={{ required: 'This field is required' }}
```
karrui: nit: should we use the `REQUIRED_ERROR` validation constant that is already declared for consistency?

frontend/src/features/env/AdminFeedbackModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  control={control}
+                  name={'rating'}
```
karrui: supernit: 
```suggestion
                  name="rating"
```

edit: and all the rest haha don't want to spam PR review with supernit comments

frontend/src/features/env/AdminFeedbackModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  render={({ field: { value, onChange, ...rest } }) => (
+                    <Rating
+                      numberOfRatings={5}
+                      variant={'star'}
```
karrui: supernit: 
```suggestion
                      variant="star"
```

frontend/src/features/env/EnvService.ts
```diff
@@ -82,28 +59,42 @@ export const submitSwitchEnvFormFeedback = async ({
...
+
   const formData = createSwitchFeedbackSubmissionFormData(
+    formFields,
     formInputs,
     feedbackForm,
   )
+
   return ApiService.post<SuccessMessageDto>(
-    `${PUBLIC_FORMS_ENDPOINT}/submissions/email/switchenvfeedback?captchaResponse=null`,
+    `${PUBLIC_FORMS_ENDPOINT}/submissions/email/switchenvfeedback?captchaResponse=null&view=${
```
karrui: just noticed this API call haha, should we do it in a more idiomatic `axios` way?

```ts
return ApiService.post(
  `${PUBLIC_FORMS_ENDPOINT}/submissions/email/switchenvfeedback`,
  null, // no query body
  { 
    params: {
      captchaResponse: null, // do we even need this? I think it defaults to null on the server validator.
      view: isAdmin ? 'admin' : 'public'
    }
  }
}
  
```

frontend/src/features/env/EnvService.ts
```diff
@@ -82,28 +59,42 @@ export const submitSwitchEnvFormFeedback = async ({
...
  * identify information obtained from Singpass/Corppass/MyInfo.
  * @returns Public view of form, with additional identify information
  */
-export const getSwitchEnvFormView = async (): Promise<PublicFormViewDto> => {
+export const getFeedbackFormView = async (
+  admin?: boolean,
+): Promise<PublicFormViewDto> => {
   return ApiService.get<PublicFormViewDto>(
-    `${PUBLIC_FORMS_ENDPOINT}/switchenvfeedback`,
+    `${PUBLIC_FORMS_ENDPOINT}/switchenvfeedback?view=${
```
karrui: same here

frontend/src/features/env/SwitchEnvFeedbackModal.stories.tsx
```diff
@@ -28,27 +29,11 @@ const onClose = () => {
...
-  return (
-    <SwitchEnvFeedbackModal
-      onChangeEnv={() => console.log('change env')}
-      onSubmitFeedback={async () => console.log('submit feedback')}
-      onClose={onClose}
-      isOpen={true}
-      isAdminView={true}
-    />
-  )
+  return <AdminFeedbackModal onClose={onClose} isOpen={true} />
```
karrui: nit: props are always truthy in react don't need `true` assignment

```suggestion
  return <AdminFeedbackModal onClose={onClose} isOpen />
```

frontend/src/features/env/mutations.ts
```diff
@@ -42,18 +41,28 @@ export const useEnvMutations = (
...
-    submitSwitchEnvFormFeedbackMutation,
   }
 }
+
+export const useAdminFeedbackMutation = () => {
+  const { data: adminFeedbackForm } = useAdminFeedbackFormView()
+  return useMutation((args: AdminFeedbackFormDto) => {
+    return submitSwitchEnvFormFeedback({
+      formInputs: args,
+      feedbackForm: adminFeedbackForm,
```
karrui: seems like a dangerous regression here, what if the feedback form query has not resolved yet? may need some error handling.

would if be safer to pass the admin feedback form as an arg to this mutation instead? And the parent component should be rendering loading state if feedbackform has not been fetched.

frontend/src/features/env/mutations.ts
```diff
@@ -42,18 +41,28 @@ export const useEnvMutations = (
...
+    })
+  })
+}
+
+export const usePublicFeedbackMutation = () => {
+  const { data: publicFeedbackForm } = usePublicFeedbackFormView()
```
karrui: samez

frontend/src/features/env/queries.ts
```diff
@@ -5,11 +5,12 @@ import { ClientEnvVars } from '~shared/types/core'
...
 import { ApiError } from '~typings/core'
 
-import { getClientEnvVars, getSwitchEnvFormView } from './EnvService'
+import { getClientEnvVars, getFeedbackFormView } from './EnvService'
 
 const envKeys = {
   base: ['env'],
-  viewSwitchEnvForm: ['viewSwitchEnvForm'] as const,
+  adminFeedbackForm: ['adminFeedbackForm'] as const,
+  publicFeedbackForm: ['publicFeedbackForm'] as const,
```
karrui, line 13: comment: feels like this whole feedback thing should not be in the `env` feature module and instead should be in it's own module. used to close one eye because was supposed to be temporary. is that still the case? if going to be stuck here forever we should move it to its own module.

out of scope of this PR though, we could check whether it's here to stay or not and create an issue for relocation otherwise.

src/app/modules/form/public-form/public-form.middlewares.ts
```diff
@@ -1,9 +1,14 @@
...
+
 import { reactMigration } from '../../../config/config'
 import { ControllerHandler } from '../../core/core.types'
 
 export const injectFeedbackFormUrl: ControllerHandler = (req, res, next) => {
-  const formId = reactMigration.reactToAngularFeedbackFormId
+  const formId =
+    get(req.query, 'view') === 'admin'
+      ? reactMigration.adminSwitchEnvFeedbackFormId
+      : reactMigration.publicSwitchEnvFeedbackFormId
```
karrui: nit: should we add a joi validation step now that we look at the query? not a big deal since we ignore all values except `admin`.. but still good to have? if we also ignore all values except `admin`, why are we assigning the value of `public` from the frontend? can just be undefined in that case haha

frontend/src/features/env/EnvService.ts
justynoh: captchaResponse is required by `validateCaptchaParams` haha

src/app/modules/form/public-form/public-form.middlewares.ts
justynoh: I think it's okay, since soon we'll be getting rid of the public form anyway and always default to the single admin form. Then we can also get rid of the query param and just serve the admin form id. What do you think?

frontend/src/features/env/queries.ts
justynoh: I agree, but for now kept it here since the future is a bit unclear. We also don't yet have clarity on whether we're reusing the feedback icon for anything else (e.g. enable feedback to admins in public forms, like you said).

Based on my convo with pearly, we are going to be deprecating the entire public feedback form soon, and only keep the admin feedback form, once we close off switchbacks to angular. Also currently, the feedback and switch env flows are separated for admins, but they are still together for respondents, so I decided not to separate them for now. We can rename and refactor a bit when we delete the respondent modal.

frontend/src/features/env/mutations.ts
justynoh: point taken, thanks! will fix

frontend/src/features/env/AdminFeedbackModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+              justify="right"
+              direction={{ base: 'column-reverse', md: 'row' }}
+            >
+              <Button isFullWidth={isMobile} variant="clear" onClick={onClose}>
+                Cancel
+              </Button>
+              <Button
+                isFullWidth={isMobile}
+                type="submit"
+                isDisabled={isLoading}
```
karrui, line 166: Nit: should we just use the isLoading prop? Adds a spinner when loading too.

frontend/src/features/env/AdminFeedbackModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+            </Skeleton>
+          </ModalBody>
+          <ModalFooter mt={{ base: '2.5rem', md: '0' }}>
+            <Stack
+              w="100%"
+              spacing="1rem"
+              justify="right"
+              direction={{ base: 'column-reverse', md: 'row' }}
+            >
+              <Button isFullWidth={isMobile} variant="clear" onClick={onClose}>
```
karrui: This should be disabled when feedback is being submitted

frontend/src/features/env/PublicFeedbackModal.tsx
```diff
@@ -0,0 +1,212 @@
...
+              <Stack
+                w="100%"
+                spacing="1rem"
+                justify="right"
+                direction={{ base: 'column-reverse', md: 'row' }}
+              >
+                <Button
+                  isFullWidth={isMobile}
+                  variant="clear"
+                  onClick={onClose}
```
karrui, line 194: Same

karrui (approved):
Lgtm

frontend/src/features/env/AdminFeedbackModal.tsx
justynoh: I thought about it, but i decided not to use it since the submit button in a spinner state gives the impression that a submission is being made. I thought that disabling the button made more sense if the form hasn't loaded yet haha

karrui: Ok it's not a post request? then should we disable on fetch and set loading on post then?

justynoh: yup, done in [086308e](https://github.com/opengovsg/FormSG/pull/5411/commits/086308e40d434b870badf2f25189796b2e73743e)!

frontend/src/features/env/AdminFeedbackModal.tsx
justynoh: [086308e](https://github.com/opengovsg/FormSG/pull/5411/commits/086308e40d434b870badf2f25189796b2e73743e)

frontend/src/features/env/PublicFeedbackModal.tsx
justynoh: [086308e](https://github.com/opengovsg/FormSG/pull/5411/commits/086308e40d434b870badf2f25189796b2e73743e)

PR #5433, opened 2022-11-21
fix: v.6.18.2: remove whitespace from dropdown validation and logic checks

wanlingt:
## Problem

When dropdown options are created in Angular, the leading and trailing whitespace are not trimmed. Respondents are unable to submit their forms in React, as React compares the trimmed version of the selected dropdown option with the (untrimmed) dropdown options and invalidates the submission

## Solution

- Trim dropdown options
- Trim condition values in logic


justynoh (approved):
lgtm as a hotfix

however, I'm wondering that if want to enforce trimming for dropdown (or even radio, checkbox, and table dropdowns options), this should be done when the field is saved, on the backend. then when the field definition is retrieved, we can always refer directly to the source of truth rather than tracking all the places where we need to trim the options. maybe we can just open a ticket for this to clean up

wanlingt:
> lgtm as a hotfix
> 
> however, I'm wondering that if want to enforce trimming for dropdown (or even radio, checkbox, and table dropdowns options), this should be done when the field is saved, on the backend. then when the field definition is retrieved, we can always refer directly to the source of truth rather than tracking all the places where we need to trim the options. maybe we can just open a ticket for this to clean up

yeah agree!! ok ill add a todo to remove this after we remove angular...

wanlingt:
> lgtm as a hotfix
> 
> however, I'm wondering that if want to enforce trimming for dropdown (or even radio, checkbox, and table dropdowns options), this should be done when the field is saved, on the backend. then when the field definition is retrieved, we can always refer directly to the source of truth rather than tracking all the places where we need to trim the options. maybe we can just open a ticket for this to clean up

ALTHOUGH on second thought, this would apply to all old forms created in angular. so we'd have to decide if we wanna modify the forms or keep this trimmed in the backend. will tag it to the issue regardless

tshuli (approved):
lgtm!

For discussion, since we did allow dropdown options with trailing / leading whitespaces previously, it may be possible (although unlikely) that admins have used that to differentiate dropdown options. So this fix would invalidate their existing form logic.

PR #5435, opened 2022-11-21
chore: merge release v6.18.2 into develop 

wanlingt:
This merge commit contains `v6.18.1` and `v6.18.2` hotfixes


timotheeg (approved):
@wanlingt remember to push the tag `v6.18.2`, also I think you didn't pull `v6.18.1` when you made the hotfix so the deploy changeset into the `changelog.md` is incorrect (not important, but just to keep in mind for next time)

wanlingt:
> @wanlingt remember to push the tag `v6.18.2`, also I think you didn't pull `v6.18.1` when you made the hotfix so the deploy changeset into the `changelog.md` is incorrect (not important, but just to keep in mind for next time)

oh no! okay, pushed the tag `v.6.18.2`. didn't realise that the tag was separate from pulling the updated code - will remember to fetch the tag in the future

PR #5436, opened 2022-11-21
feat: allow special chars in form title

### LINKED ISSUE #5406: Allow special characters in form titles

wanlingt:
## Problem

We want to allow special characters in the form title

Closes #5406

## Solution

- Allow all special characters in form titles, apart from invalid utf8 encoded characters (refer to #4173 for precedent for form field titles). 
- A separate function `verifyValidUnicodeString` is used to represent the custom Joi validation for invalid utf-8 encoded characters. This function is added as a new custom validation to `createFormValidator` , `updateSettingsValidator`, and replaces the existing implementations in `handleUpdateFormField`, `handleCreateFormField`

## Tests

Form title:
- [x] Create a new form with a previously-disallowed special character in the title (e.g. `狗`). You should be able to create the form successfully.
- [x] Copy the network request of the step above as cURL (POST). Now modify the network request so that the title includes `\udbbb` e.g. `title\udbbb`. Send the network request. You should receive 400.
- [x] Rename an existing form title to now include a previously-disallowed special character in the title (e.g. `狗`). You should be able to update the form successfully.
- [x] Copy the network request of the step above as cURL (PATCH). Now modify the network request so that the title includes `\udbbb` e.g. `title\udbbb`. Send the network request. You should receive 400.


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -136,7 +136,23 @@ const createFormValidator = celebrate({
       })
       .required()
       // Allow other form schema keys to be passed for form creation.
-      .unknown(true),
+      .unknown(true)
+      .custom((value, helpers) => {
+        // If there are invalid utf-8 encoded unicode-escaped characters,
+        // node 14 treats the sequence of characters as a string e.g. \udbbb is treated as a 6-character string instead of an escaped unicode sequence
```
karrui: We are now on node 18, does this still hold?

timotheeg: Yup it still holds:
```javascript
➜  ~ node --version
v18.12.0

➜  ~ node
Welcome to Node.js v18.12.0.
Type ".help" for more information.

> JSON.stringify('\u4E24') // valid unicode sequence
'"两"'

> JSON.stringify('\udbbb') // invalid unicode sequence
'"\\udbbb"'
```

timotheeg: Maybe we can replace `node 14` all the comments by simply `nodejs`.

timotheeg (commented):
So the same code is duplicated 4 times now, can it be extracted into a shared function? Maybe something like this:

```typescript
verifyValidUnicodeString(value, helpers, errorKey='custom') {
  // explanation comment + validation check + error object returned as
  return helpers.message({
    [errorKey]: 'There are invalid characters in your input',
  })
}

// And then invocations as:
.custom(verifyValidUnicode)
```

src/app/modules/form/admin-form/admin-form.controller.ts
wanlingt: Thanks for helping to check! @timotheeg 

Updated the comments here [8e6b94b](https://github.com/opengovsg/FormSG/pull/5436/commits/8e6b94b9bee14935c52c05d8974ec8cac48a1f80)

wanlingt:
> So the same code is duplicated 4 times now, can it be extracted into a shared function? Maybe something like this:
> 
> ```ts
> verifyValidUnicodeString(value, helpers, errorKey='custom') {
>   // explanation comment + validation check + error object returned as
>   return helpers.message({
>     [errorKey]: 'There are invalid characters in your input',
>   })
> }
> 
> // And then invocations as:
> .custom(verifyValidUnicode)
> ```

Extracted it into a separate function - [e2ad263](https://github.com/opengovsg/FormSG/pull/5436/commits/e2ad2634dc22fd89a8410a9086f234d7279eaed7)

timotheeg (approved):
lgtm

PR #5442, opened 2022-11-22
fix(docker-compose): include US and SG SES env vars

mantariksh:
## Problem

Dev environment doesn't work out of the box because `SES_PORT` and `SES_HOST` env vars for both US and SG are not included by default.

## Solution

Add them to `docker-compose.yml` so each dev doesn't have to add them to an env file individually.


wanlingt (approved):

karrui (approved):
Welcome back

PR #5456, opened 2022-11-24
fix: update frontend error messaging for traceability

justynoh:
## Problem
We are getting some error messages which are not very traceable. This PR adds some instrumentation with error labelling for traceability.

## Solution
**Breaking Changes** 

### DIFF
--- a/frontend/src/services/ApiService.ts
+++ b/frontend/src/services/ApiService.ts
@@ -24,27 +24,27 @@ export const transformAxiosError = (e: Error): ApiError => {
     if (e.response) {
       const statusCode = e.response.status
       if (statusCode === StatusCodes.TOO_MANY_REQUESTS) {
-        return new HttpError('Error [429]: Please try again later.', statusCode)
+        return new HttpError('Error [001]: Please try again later.', statusCode)
       }
       if (typeof e.response.data === 'string') {
-        return new HttpError(`Error [001]: ${e.response.data}`, statusCode)
+        return new HttpError(`Error [002]: ${e.response.data}`, statusCode)
       }
       if (e.response.data?.message) {
         return new HttpError(
-          `Error [002]: ${e.response.data.message}`,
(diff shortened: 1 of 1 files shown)


frontend/src/services/ApiService.ts
```diff
@@ -24,21 +24,27 @@ export const transformAxiosError = (e: Error): ApiError => {
...
       }
       if (e.response.statusText) {
-        return new HttpError(e.response.statusText, statusCode)
+        return new HttpError(
+          `Error [003]: ${e.response.statusText}`,
+          statusCode,
+        )
       }
-      return new HttpError(`Http ${statusCode} error`, statusCode)
+      return new HttpError(`Error [${statusCode}]`, statusCode)
```
tshuli: To be consistent with the other labels, how about Error `[004]: ${statusCode} error`?

frontend/src/services/ApiService.ts
```diff
@@ -24,21 +24,27 @@ export const transformAxiosError = (e: Error): ApiError => {
     if (e.response) {
       const statusCode = e.response.status
       if (statusCode === StatusCodes.TOO_MANY_REQUESTS) {
-        return new HttpError('Please try again later.', statusCode)
+        return new HttpError('Error [429]: Please try again later.', statusCode)
```
tshuli: How about `Error [001]: Please try again later` (and downshifting the rest) so that the numbering is consistent for tracing?

tshuli (approved):
just some questions on the numbering system, else lgtm!

frontend/src/services/ApiService.ts
justynoh: okie shifted everything down!

frontend/src/services/ApiService.ts
justynoh: same as above

PR #5458, opened 2022-11-25
feat: search and filter for forms on the dashboard

### LINKED ISSUE #5397: Implement search and filter for form organization

justynoh:
## Problem
We want to implement search and filter for the dashboard. This PR does this.

Closes #5397

## Solution

- ~~Major updates to searchbar component to add filter element on the right, as well as mobile support for it (although not used in this PR). Updated stories to include those with filters too.~~ Searchbar not touched, decided to create a new WorkspaceSearchbar instead since it was different from all the other searchbars.
- Add searchbar to the workspace header in all three modes, desktop tablet and mobile. Some special branching had to be done so that each specific case could be accomodated. Figma [here](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=14174%3A145335&t=TrlRawLlHmQDsYSP-1)
- Update workspace context to accommodate search options, and use fuzzysort to help with dynamic fuzzy search.

**Breaking Changes** 

## Screenshots

Desktop

https://user-images.githubusercontent.com/25571626/203884749-ab8efb3f-ccf2-49b2-8c00-a4911e75de87.mov

Tablet

https://user-images.githubusercontent.com/25571626/203884783-16bd0c92-e2dc-445b-9b31-fc1df88c5d03.mov

Mobile

https://user-images.githubusercontent.com/25571626/203884798-fc5bcb71-1be1-447a-991a-b8afe3dea379.mov

## Tests

In all three modes (desktop, tablet, mobile):
- [ ] Searching and filtering works as intended. Dynamic search expected (i.e. the search should occur onChange of the input).
- [ ] For mobile and tablet modes, closing the searchbar also clears the current search (but the current filter is left as is).


frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -1,32 +1,48 @@
...
+
 import IconButton from '../IconButton'
 
 export interface SearchbarProps extends Omit<InputProps, 'onChange'> {
   /**
-   * Function to be invoked when user presses enter (to search) or clicks the
-   * 'search' icon.
+   * Optional. Function to be invoked when user presses enter (to search) or
+   * clicks the 'search' icon. Note: at least one of onSearch and onChange must
+   * be provided for the searchbar to work properly!
```
mantariksh: you can consider using [`RequireAtLeastOne`](https://github.com/sindresorhus/type-fest/blob/main/source/require-at-least-one.d.ts) to enforce this. for example,

```typescript
type SearchbarHandlers = {
  onSearch: (searchValue: string) => void
  onChange: (newValue: string) => void
}

export interface SearchbarProps extends Omit<InputProps, 'onChange'> {
  handlers: RequireAtLeastOne<SearchbarHandlers, 'onSearch' | 'onChange'>
  // other props
}
```

mantariksh: or you could even not nest it under a `handlers` key and just use `RequireAtLeastOne` on the whole of `SearchbarProps`

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -58,51 +73,113 @@ export interface SearchbarProps extends Omit<InputProps, 'onChange'> {
   value?: string
 
   /**
-   * Function to be invoked when the value in the searchbar input changes (but
-   * the search button has not been clicked).
+   * Optional. Function to be invoked when the value in the searchbar input
```
mantariksh: tiny nit: not sure we need the word "Optional" everywhere, as it's already shown by the type. plus if we start putting "Optional" for every optional prop, we'd need to do it consistently, e.g. `isExpandable` doesn't have it rn.

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: omg, THANKS!!! 

❤️ static guarantees

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -58,51 +73,113 @@ export interface SearchbarProps extends Omit<InputProps, 'onChange'> {
...
+        isFullWidth: true,
+        justifyContent: 'flex-start',
+        variant: 'clear',
+        colorScheme: 'secondary',
+        textStyle: 'body-1',
+      }),
+      [],
+    )
+
+    const filterActive = useMemo(
```
mantariksh: nit: in general we name booleans or functions which return booleans starting with `is` or `has`, e.g. `isFilterActive`

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -58,51 +73,113 @@ export interface SearchbarProps extends Omit<InputProps, 'onChange'> {
...
+   * Optional. If provided, adds a filter button to the searchbar. The initial
+   * (default) value of the filter, which is most likely the option for "No filter".
+   */
+  filterValue?: string
+
+  /**
+   * Optional. The remainder of the filter options, not including the default
+   * filterValue. Will be ignored unless filterValue has been provided.
+   */
+  filterOptions?: string[]
```
mantariksh: hm this seems to break the convention of providing one prop specifying the current value and another prop specifying the full array of options (including the current value). see e.g. [`react-dropdown`](https://www.npmjs.com/package/react-dropdown). any particular reason why `filterValue` must not be an element of `filterOptions`?

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -58,51 +73,113 @@ export interface SearchbarProps extends Omit<InputProps, 'onChange'> {
...
+
+  /**
+   * Optional. Function to be invoked when a filter option has been selected.
+   * Will be ignored unless filterValue has been provided.
```
mantariksh: I'm wondering if we can design a more intuitive API for indicating "No filter". intuitively the behaviour I'd expect from this component would be something like:
- as long as I provide `filterOptions` and `onFilter`, the filter button gets added
- if I don't provide `filterValue`, the component defaults to "no filter"

wdyt?

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: I guess i personally prefer to be more explicit, but I agree that the type already shows it. I guess more importantly is to explain how the behavior changes if the prop is omitted, haha

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: hmm, yeah. This was a decision that I did think about, and I went for this for two reasons. 

1. In this use case, default filter value will generally always be "no filter", so it made sense to view it not really as a filter option, but rather as the _name associated with_ "No filter". (for e.g. in the workspace context, "No filter" is called "All forms"). Then, filter options should be the list of options that represent the actual ways to filter whatever content you need. Perhaps in this situation the naming of `filterValue` is not great (maybe `noFilterName` or something like that?).
2. Asking users for an array along with a specified element of that array just seems funny to me. It's possible to also get rid of this by getting rid of `filterValue` and promising that `filterOptions[0]` will be the initial filter value, but then you lose the guarantee that `filterOptions` will be non-empty. Of course we can throw an error, but that also feels a bit inelegant i guess.

All that being said, I do see your point that the convention is just to provide the option array + an option from that array, and I don't think there's any issue with that (especially the example with react-dropdown). It would also make it more flexible, in the case that we want to initialize the filter with an option that isn't going to be shown as the first option. wdyt?

mantariksh: personally I'd rather keep it unopinionated, so I wouldn't assume that you always want a "no filter" state (e.g. you might have a case where some filter is always active). my take is that we should stick with convention and just specify an element of the `filterOptions` array.

agree that it seems funny but specifying a `filterOptions` prop that doesn't actually include all the `filterOptions` is also funny. in this case I think it's best to stick to what devs are generally used to.

but will leave final decision up to you! it's our own component so we can always update it

frontend/src/components/Searchbar/Searchbar.tsx
mantariksh: no strong opinions about this, am ok to not change

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: nah, i agree with you. I'll update it!

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: To me ideally, since `onSearch` and `onChange` control the 'searchableness' of the searchbar, `onFilter` should control the 'filterableness' of the filter section. The issue with that is then that, if there is no filterOption, we run into the issue of having no filter value to select. So the "correct" way to do this is to require /both/ filterOptions and onFilter, for the filter button to appear. But in that case then we should bundle them together and require an object `{filterOptions, onFilter}`, but that doesn't seem very conventional either. 

Given our other conversation (see above), here's what I would do:

- a non-empty filterOptions and onFilter must be provided in order for the filter button to be added
- if filterValue is provided AND filterValue \in filterOptions, we use that as the initial filter value. Otherwise, we default to filterOptions[0] as the initial filter value. 

Do you have any thoughts about this API?

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: [72a9d92](https://github.com/opengovsg/FormSG/pull/5458/commits/72a9d92bca2923427e1c5f8dbf187bf4b5bdb538)

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: [72a9d92](https://github.com/opengovsg/FormSG/pull/5458/commits/72a9d92bca2923427e1c5f8dbf187bf4b5bdb538)

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: [72a9d92](https://github.com/opengovsg/FormSG/pull/5458/commits/72a9d92bca2923427e1c5f8dbf187bf4b5bdb538)

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: implemented this for now in [72a9d92](https://github.com/opengovsg/FormSG/pull/5458/commits/72a9d92bca2923427e1c5f8dbf187bf4b5bdb538), feel free to give more comments!

mantariksh: yup sgtm, and the docs can clarify that `filterValue` is ignored if it is not an element of `filterOptions`

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -58,51 +74,130 @@ export interface SearchbarProps extends Omit<InputProps, 'onChange'> {
...
+      [filterOptions, filterValueProp],
+    )
 
-    const [innerIsExpanded, setInnerIsExpanded] = useState<boolean>(
-      isExpandable ? isExpanded : true,
+    const [value, setValue] = useState<string>(valueProp ?? '')
+    const [isExpanded, setIsExpanded] = useState(
+      !isExpandable || isExpandedProp,
     )
+    const [filter, setFilter] = useState<string>(filterValue)
```
mantariksh: this gets a bit confusing because there's now both `filter` and `filterValue`. shall we make the naming a bit more explicit? e.g. maybe `filterValue` can be renamed to `initialFilterValue` and/or `filter` can be `currentFilter`

frontend/src/components/Searchbar/Searchbar.tsx
```diff
@@ -111,20 +206,51 @@ export const Searchbar = forwardRef<SearchbarProps, 'input'>(
...
+    const onFilterSelect = useCallback(
+      (option: string) => {
+        if (onFilter) onFilter(option)
+        setFilter(option)
+      },
+      [onFilter],
+    )
+
+    if (!isExpanded) {
+      return (
```
mantariksh: nice, I like this much better

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -73,94 +134,82 @@ export const WorkspaceHeader = ({
...
-                </MenuButton>
+                {renderFilterButton(isOpen)}
                 <Menu.List>
-                  <Menu.Item onClick={() => setActiveFilter(null)}>
-                    {renderFilterButton(null)}
-                  </Menu.Item>
-                  {Object.values(FilterOption).map((filterOption) => (
-                    <Menu.Item onClick={() => setActiveFilter(filterOption)}>
-                      {renderFilterButton(filterOption)}
+                  {[null, ...Object.values(FilterOption)].map((option, i) => (
```
mantariksh: can we leave a comment here about why there's a `null`?

mantariksh (commented):
I see the background colour of the search icon is different from the search bar. is that in line with the design?

karrui:
> I see the background colour of the search icon is different from the search bar. is that in line with the design?

Will fix, saw some layout shifts too. Will fix those too. 
Will also add highlight to form titles ala dropdown field search

frontend/src/features/workspace/WorkspaceContext.tsx
```diff
@@ -9,8 +9,10 @@ export interface WorkspaceContextProps {
   totalFormsCount?: number
   displayedForms: AdminDashboardFormMetaDto[]
   displayedFormsCount: number
-  activeFilter: FilterOption | null
-  setActiveFilter: (filterOption: FilterOption | null) => void
+  activeSearch: string
```
mantariksh, line 12: nice, I think empty string rather than `null` makes more sense

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -1,30 +1,25 @@
...
+  Flex,
   Grid,
-  Icon,
-  MenuButton,
   Skeleton,
-  Stack,
   Text,
   useDisclosure,
 } from '@chakra-ui/react'
+import { useMediaMatch } from 'rooks'
```
mantariksh: rooks

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -36,139 +31,89 @@ export const WorkspaceHeader = ({
   handleOpenCreateFormModal,
 }: WorkspaceHeaderProps): JSX.Element => {
   const isMobile = useIsMobile()
+  const isDesktop = useMediaMatch(`(min-width: ${BREAKPOINT_VALS.lg})`)
```
mantariksh: maybe abstract into `useIsDesktop` just like `useIsMobile`?

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -36,139 +31,89 @@ export const WorkspaceHeader = ({
...
       gridTemplateAreas={{
-        base: "'header filter' 'create create'",
-        md: "'header filter create'",
+        base: isSearchExpanded
+          ? "'header searchicon filter' 'search search search' 'create create create'"
+          : "'header searchicon filter' 'create create create'",
+        md: isSearchExpanded
+          ? "'header searchicon filter create' 'search search search search'"
+          : "'header searchicon filter create'",
+        lg: "'header searchfilter create'",
```
mantariksh: nit: use camelcase for grid areas for readability? or is all lowercase a convention

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -36,139 +31,89 @@ export const WorkspaceHeader = ({
...
+      }}
+      gridTemplateColumns={{
+        base: isSearchExpanded ? '1fr auto' : '1fr auto auto',
```
mantariksh, line 69: could you explain this - isn't it 3 columns regardless of `isSearchExpanded`? so why have a `'1fr auto'` here?

frontend/src/features/workspace/components/WorkspaceSearchbar/MobileWorkspaceSearchbar.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const handleToggle = useCallback(() => {
+      if (isExpanded) {
+        setInternalValue('')
+      }
+      onToggleExpansion()
+    }, [isExpanded, onToggleExpansion, setInternalValue])
+
+    return (
+      <>
+        <Box gridArea="searchicon">
```
mantariksh: tiny nit and might not be worth it, but I wonder if the `gridArea` names should be in the props somehow. currently the styling of this component and its parent are coupled through the grid area names, might be hard to debug in future. but also might make the props messy so up to you

frontend/src/theme/components/Menu.ts
```diff
@@ -64,7 +64,6 @@ const variantClear: PartsStyleFunction<typeof parts> = (_props) => {
   return {
     button: {
       minH: 'auto',
-      p: '0.25rem',
```
mantariksh, line 67: hope this doesn't break anything else

mantariksh (approved):
some nits

frontend/src/components/Searchbar/Searchbar.tsx
justynoh: renaming to `initialFilterValue` and `filterValue`, for consistency with the search state being called `value`

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
justynoh: okie

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
justynoh: I think you're right here! will fix

frontend/src/features/workspace/components/WorkspaceHeader/WorkspaceHeader.tsx
justynoh: will update, though it seems that everywhere else we use lowercase haha. can update elsewhere whenever else

frontend/src/features/workspace/components/WorkspaceSearchbar/MobileWorkspaceSearchbar.tsx
justynoh: Personally I think it'll make the props a bit messy but i agree we should do something to make the coupling more explicit. Will add comments to address this

justynoh:
Note: @karrui and I decided to remove the changes to searchbar component in this PR, if we need a search and filter component, we can add it later as a different component.

Issue #5452, opened 2022-11-23
Deprecate validity of old MyInfo cookies


mantariksh:
Once expiry time of MyInfo forms is extended beyond 30min through usage of a new cookie format, remove code which accepts the old cookie format. This can be done 30min after the new cookies start being used.

PR #5464, opened 2022-11-28
fix: extend MyInfo login expiry to match env var

### LINKED ISSUE #3459: MyInfo cookie expires after 30 minutes without warning
The JWT returned from MyInfo has a 30min expiry, so when users attempt to submit their MyInfo form after 30min, we reject the submission as the JWT is invalid. However, no warning is given to the user about this, so they have to refresh and fill in the form from scratch.

mantariksh:
## Problem

MyInfo logins currently expire after only 30 minutes. If user attempts to submit the form after 30min, they see an error and have to refresh, re-login and fill out the whole form again. Many users are complaining that they don’t have enough time to fill out their forms, especially because there have been a lot of long MyInfo forms (180 - 190 fields) created recently for schools registration.

This 30-minute expiry is unintentional behaviour. It used to be the case that MyInfo logins were valid for 3 hours, but then MyInfo reduced the validity period of their tokens, and our logins were cut short because we coupled the validity of our MyInfo login to the expiry of their tokens.

Closes #3459 

## Solution

1. When user hits the `/myinfo/login` endpoint with the auth code, the auth code is stored in a HTTP-only cookie.
2. When the user is redirected to the public form endpoint, the auth code is exchanged for the access token, and the access token is then presented to retrieve the MyInfo data. This way, the client never sees the access token.
3. Once the data is retrieved, the access token is thrown away. FormSG creates its own MyInfo login JWT containing the UIN/FIN of the user. This JWT has an expiry that we specify via environment variable.
4. Upon form submission, we verify the signature on the login JWT.

**Breaking Changes** 

  - The old cookie format is still accepted, for users who are currently logged in to MyInfo at the point that this change is deployed. It will be deprecated when #5452 is addressed.

## Tests

- [x] Log in to a MyInfo form with the previous version of the code so you are logged in with the old cookie. Deploy the new server. Submit the form and ensure it can be submitted successfully.
- [x] Log in to a MyInfo form with the new server. Check that the expiry time of the cookie called MyInfoCookie is in 3 hours. Ensure you can submit successfully
- [x] Log in to a MyInfo form. Manually delete the cookie called MyInfoCookie in your console and submit the form. Submission should be rejected.
- [x] Log in to a MyInfo form. Manually edit the cookie called MyInfoCookie in your console and replace the JWT with garbage. Submit the form. Submission should be rejected.
- [x] Log in to a MyInfo form. Log out and make sure that the cookie called MyInfoCookie is deleted.
- [x] Log in to a MyInfo form but do not consent to providing your MyInfo data. Check that the MyInfo error message is shown when you return to the form.
- [x] Production NDI env var config contains a new env var called MYINFO_JWT_SECRET generated using the following: require('crypto').randomBytes(32).toString('base64')

## Deploy Notes

- `MYINFO_JWT_SECRET`: Secret for signing MyInfo JWT.


mantariksh:
@timotheeg @tshuli  For the record, CodeQL raised the following alert:
Screenshot 2022-11-28 at 10 10 56 AM

I dismissed this alert based on the fact that:
1. We have rate-limiting via Cloudflare
2. The MyInfo circuit breaker allows the MyInfo server to recover from a DOS attack by short-circuiting when their server is unresponsive

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -219,70 +218,75 @@ export const handleGetPublicForm: ControllerHandler<
           return res.json({ form: publicForm, isIntranetUser })
         })
     case FormAuthType.MyInfo: {
+      // We always want to clear existing login cookies because we no longer
+      // have the prefilled data
+      res.clearCookie(MYINFO_LOGIN_COOKIE_NAME)
+      const authCodeCookie: unknown = req.cookies[MYINFO_AUTH_CODE_COOKIE_NAME]
+      // No auth code cookie because user is accessing the form before logging
+      // in
+      if (authCodeCookie === null || authCodeCookie === undefined) {
```
karrui: nit: should we use a falsey check?

```suggestion
      if (!authCodeCookie) {
```

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -219,70 +218,75 @@ export const handleGetPublicForm: ControllerHandler<
...
+          logger.error({
+            message: 'MyInfo login error',
+            meta: logMeta,
+            error,
           })
-      )
+          // No need for cookie if data could not be retrieved
+          // NOTE: If the user does not have any cookie, clearing the cookie still has the same result
+          return res
+            .clearCookie(MYINFO_LOGIN_COOKIE_NAME, MYINFO_LOGIN_COOKIE_OPTIONS)
```
karrui: this has already been cleared above in line 223, should we remove it here?

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -306,16 +312,36 @@ const isMyInfoFormWithEsrvcId = <F extends IFormSchema>(
...
+  if (
+    cookie &&
+    typeof cookie === 'object' &&
+    hasProp(cookie, 'uinFin') &&
+    typeof cookie.uinFin === 'string'
+  ) {
+    return true
+  }
+  return false
+}
```
karrui: nit: simplify to returning conditional immediately

```suggestion
return (
  !!cookie &&
  typeof cookie === 'object' &&
  hasProp(cookie, 'uinFin') &&
  typeof cookie.uinFin === 'string'
  ) 
}
```

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -333,80 +359,143 @@ export const isMyInfoCookie = (
...
 
+/**
+ * Extracts a MyInfo login cookie from a request's cookies
+ * @param cookies Cookies in a request
+ */
+export const extractMyInfoLoginJwt = (
+  cookies: Record<string, unknown>,
+): Result<string, MyInfoMissingLoginCookieError> => {
+  const jwt = cookies[MYINFO_LOGIN_COOKIE_NAME]
+  if (typeof jwt === 'string') {
```
karrui: should we also check whether the string is empty? 
```suggestion
  if (typeof jwt === 'string' && !!jwt) {
```

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -372,7 +367,7 @@ const submitEmailModeForm: ControllerHandler<
           )
           // MyInfo access token is single-use, so clear it
           return res
-            .clearCookie(MYINFO_COOKIE_NAME, MYINFO_COOKIE_OPTIONS)
+            .clearCookie(MYINFO_LOGIN_COOKIE_NAME, MYINFO_LOGIN_COOKIE_OPTIONS)
```
karrui, line 370: comment: Feel like we clear many times, but i guess it doesn't hurt.

karrui (approved):
awesome. lgtm with some nits

src/app/modules/form/public-form/public-form.controller.ts
mantariksh: can also!

src/app/modules/form/public-form/public-form.controller.ts
mantariksh: good catch, removed

src/app/modules/myinfo/myinfo.util.ts
mantariksh: yup sure, catches the error earlier in the chain

src/app/modules/submission/email-submission/email-submission.controller.ts
mantariksh: hmm, maybe it seems that way because now there are two cookies, so we call `clearCookie` twice at the public form endpoint. actually this is the only call to `clearCookie` at this (submissions) endpoint, so I don't think this one is avoidable. for the other one, I've already reduced it to one `clearCookie` for each cookie type, which should be the minimum.

src/app/modules/myinfo/myinfo.constants.ts
```diff
@@ -31,6 +44,16 @@ export const MYINFO_COOKIE_OPTIONS = {
   secure: !config.isDev,
 }
 
+/**
+ * Settings for the MyInfo auth code cookie
+ */
+export const MYINFO_AUTH_CODE_COOKIE_OPTIONS = {
+  // Important for security - access token cannot be read by client-side JS
```
tshuli: nit: should be auth code, not access token

mantariksh: oops tq

src/app/modules/form/public-form/public-form.controller.ts
mantariksh: 676072bc

src/app/modules/form/public-form/public-form.controller.ts
mantariksh: 676072bc

src/app/modules/myinfo/myinfo.util.ts
mantariksh: 676072bc

src/app/modules/myinfo/myinfo.util.ts
mantariksh: 676072bc

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -333,80 +356,143 @@ export const isMyInfoCookie = (
...
+  string,
+  MyInfoInvalidAuthCodeCookieError | MyInfoCookieStateError
+> => {
+  if (!isMyInfoAuthCodeCookie(cookie)) {
+    return err(new MyInfoInvalidAuthCodeCookieError(cookie))
+  }
+  if (cookie.state !== MyInfoAuthCodeCookieState.Success) {
+    return err(new MyInfoCookieStateError())
+  }
+  return ok(cookie.authCode)
```
tshuli, line 479: should we add in a check that authCode is non-empty string? there is no guarantee from the type, and we use this directly to retrieve the access token in L213 of `myinfo.service`

mantariksh: sure sgtm

src/app/modules/myinfo/myinfo.constants.ts
```diff
@@ -15,6 +15,19 @@ export const MYINFO_ROUTER_PREFIX = '/myinfo'
...
+ * from the /myinfo/login endpoint to the public form endpoint.
+ */
+export const MYINFO_AUTH_CODE_COOKIE_NAME = 'MyInfoAuthCode'
+
+/**
+ * Lifetime in ms of cookie containing auth code, which is used
+ * to pass the auth code from the /myinfo/login endpoint to the
+ * public form endpoint.
+ */
+export const MYINFO_AUTH_CODE_COOKIE_AGE_MS = 3 * 60 * 1000
```
tshuli: I wonder if this should be something shorter like 30 seconds instead of 3 minutes, since it should be used immediately. Longer expiry means there is a chance that if the user doesn't follow through on the redirect, someone else could use the auth code later on. But not a big deal tbh

src/app/modules/myinfo/myinfo.service.ts
```diff
@@ -457,13 +459,13 @@ export class MyInfoServiceClass {
...
   extractUinFin(
-    accessToken: string,
+    loginJwt: string,
   ): Result<string, MyInfoInvalidAccessTokenError> {
     return Result.fromThrowable(
-      () => this.#myInfoGovClient.extractUinFin(accessToken),
+      () => jwt.verify(loginJwt, spcpMyInfoConfig.myInfoJwtSecret),
       (error) => {
         logger.error({
           message: 'Error while extracting uinFin from MyInfo access token',
```
tshuli: suggest - update log msg to jwt has expired?

tshuli (approved):
lgtm

src/app/modules/myinfo/myinfo.constants.ts
mantariksh: hmm yeah, I suppose it's just down to how slow we expect clients to be. I've seen really slow 3G connections before. I'll cut it to 2 minutes as a compromise but ultimately we're really just going for "as short as possible without disrupting slow users" and there's no correct answer I think haha

src/app/modules/myinfo/myinfo.service.ts
mantariksh: hmm, it's not necessarily expiry here, e.g. if the client submits arbitrary garbage for JWT, it just can't be decoded. but I've fixed the log message above on line 418 (from "access token" to "login JWT")

src/app/modules/myinfo/myinfo.constants.ts
tshuli: 2 minutes sounds good!

src/app/modules/myinfo/myinfo.constants.ts
mantariksh: 357adda2

src/app/modules/myinfo/myinfo.util.ts
mantariksh: c51facf5

src/app/modules/myinfo/myinfo.constants.ts
mantariksh: eee75517

PR #5468, opened 2022-11-28
fix: update field validators with more specific type guarantees

justynoh:
## Problem
There is an issue with required checkboxes being able to be submitted even with invalid responses. This PR fixes that

https://user-images.githubusercontent.com/25571626/204191918-61ca4b66-01da-4503-92b1-0aec6dc80e33.mov

https://user-images.githubusercontent.com/25571626/204191948-b4b3b70a-d143-45fe-927e-9059d5c0301d.mov

## Solution
Refactor validators to use `FormFieldValues` as a minimal type guarantee.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/204191996-6f1737af-fb69-4e0e-a234-add7d36bdd97.mov

### DIFF
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -4,6 +4,7 @@
  */
 import { RegisterOptions } from 'react-hook-form'
 import { isValid, parse } from 'date-fns'
+import { identity } from 'lodash'
 import simplur from 'simplur'
 import validator from 'validator'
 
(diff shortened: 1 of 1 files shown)


wanlingt (approved):
@karrui and i: lgtu

wanlingt (changes requested):
found bug in chromatic

karrui (approved):

PR #5474, opened 2022-11-28
build: add .gitpod.yml

mantariksh:
Dev environment takes a very, very long time to build (10-15min) and requires very specific config options. I thought we could take advantage of [Gitpod prebuilds](https://www.gitpod.io/docs/configure/projects/prebuilds) to maintain a working cloud-based IDE.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,7 @@
+FROM gitpod/workspace-full:latest
+
+RUN bash -c 'VERSION="18" \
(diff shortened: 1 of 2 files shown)


mantariksh:
I ran into [this issue](https://github.com/gitpod-io/gitpod/issues/11183):
Screenshot 2022-11-29 at 9 20 58 AM

Which means prebuilds don't work, which means gitpod has little value-add.

Shelving this until Gitpod fixes the issue, or we find a good alternative.

PR #5483, opened 2022-11-29
ci: separate e2e and backend ci flows

wanlingt:
## Problem

The e2e and backend tests were combined in #5272 so that they can both use `npm run build`, however they take a long time to run.

## Solution

Save the build files as an artifact so that they can be used later in the separated e2e and backend tests


.github/workflows/ci.yml
```diff
@@ -135,3 +135,22 @@ jobs:
...
+      steps:
+        - uses: actions/checkout@v2
+        - name: Use Node.js
+          uses: actions/setup-node@v3
+          with:
+            node-version: 14
+            cache: 'npm'
+            cache-dependency-path: '**/package-lock.json'
+        - run: npm ci
+        - run: npm run build
```
justynoh: we should actually shard out the build into it's own step right? since they're common between both backend and e2e, then they can share on the build

karrui: yarh

wanlingt: yeah context is this is what we had initially, but the files from the build step weren't being used in e2e and backend tests so we combined the build, e2e and backend steps tgt. discussed w @karrui - can use a build artifact

wanlingt:
note to self: update node version to 18

wanlingt:
@justynoh @karrui successful ci [here](https://github.com/opengovsg/FormSG/actions/runs/3598169928)

will remove the `build_backend_and_e2e_test` requirement for merging PRs when this is approved, and add the build, backend, e2e requirement for merging PRs separately in github settings

i've also opened a separate [issue](https://github.com/opengovsg/FormSG/issues/5512) for upgrading the workflows to use node version 18 because i'm having some trouble with NODE_OPTIONS and want to merge this in first haha

karrui:
Wondering whether we can rerun tests with this flow if the build artefacts disappear too fast. May need some more investigation.

wanlingt:
> Wondering whether we can rerun tests with this flow if the build artefacts disappear too fast. May need some more investigation.

good point! I'll just remove the retention policy https://github.com/opengovsg/FormSG/pull/5483/commits/68512e4e6ad60854850ea88b98573c172e7ae348
from [docs](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts): GitHub stores artifacts for 90 days, and is free for public repos

(will reinstate the conditionals https://github.com/opengovsg/FormSG/pull/5483/commits/eedc5f407c31fd14dcdee96ba2b783d248b37645 after this PR is approved)

.github/workflows/ci.yml
```diff
@@ -109,29 +156,21 @@ jobs:
...
-        run: npm run test-e2e
-      - name: backend_test
-        if: ${{ needs.changes.outputs.backend == 'true' }}
-        run: npm run test:backend
-      - name: Coveralls
-        if: ${{ needs.changes.outputs.backend == 'true' }}
-        uses: coverallsapp/github-action@master
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - run: npm run test-e2e
```
karrui: supernit: new line

.github/workflows/ci.yml
```diff
@@ -79,6 +103,29 @@ jobs:
       - run: npm ci
       - run: npm run lint:frontend
 
+  backend_test:
+    needs: [changes, install, build]
+    # if: ${{ needs.changes.outputs.backend == 'true' }}
```
karrui: should we reenable this check too? it should only run if there are backend changes imo

.github/workflows/ci.yml
```diff
@@ -109,29 +156,21 @@ jobs:
       - run: npm ci
       - run: npm run lint-ci
 
-  build_backend_and_e2e_test:
-    needs: [changes, install]
+  src_e2e:
+    needs: [changes, build]
+    # if: ${{ needs.changes.outputs.backend == 'true' || needs.changes.outputs.angularjs == 'true' }}
```
karrui: same

.github/workflows/ci.yml
```diff
@@ -49,6 +48,31 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      NODE_OPTIONS: '--max-old-space-size=4096'
+    steps:
+      - uses: actions/checkout@v2
+      - run: mkdir -p build-output/
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
karrui, line 62: we can shunbian upgrade this node version to 18 liow

.github/workflows/ci.yml
```diff
@@ -79,6 +103,29 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@v3
+        if: always()
+        with:
+          name: build-output
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
karrui, line 119: Same (also for the rest. Actually may be in a different PR hor)

karrui (approved):
lgtm if the stuff commented are done in another PR

.github/workflows/ci.yml
wanlingt: yup will reenable after it's approved! wanted to show that the separate flows run successfully

.github/workflows/ci.yml
wanlingt: ya correct separate pr!

.github/workflows/ci.yml
wanlingt: [af93bca](https://github.com/opengovsg/FormSG/pull/5483/commits/af93bcac96b49e9550eea3074776d4f226fdfdbf)

.github/workflows/ci.yml
wanlingt: [af93bca](https://github.com/opengovsg/FormSG/pull/5483/commits/af93bcac96b49e9550eea3074776d4f226fdfdbf)

justynoh (approved):
lgtm

PR #5486, opened 2022-11-29
feat: add public form `/use-template` redirection to admin form template page

karrui:
## Problem

This PR adds backwards compatibility with the URL form admins are used to. 
Redirects `/:formId/use-template` to `/admin/form/:formId/use-template`.

Also fixes bad UX when users call `/use-template` and gets redirected to broken AngularJS + ReactJS iframe page.

## Solution

**Breaking Changes** 

**Features**:
- feat: add public form use-template redirection

## Tests

- [ ] Append `/use-template` in a public form URL **when logged out**. Should be redirected to login page, and then to the correct template page once logged in.
- [ ] Append `/use-template` when logged in. Should redirect to the template page. And creating template should work.


justynoh (approved):
lgtm

Issue #5389, opened 2022-11-15
Take SGID out of beta


amitogp:

amitogp:
Keeping 1 day capacity for PR review and due diligence.

PR #5531, opened 2022-12-05
feat: enable sgID for all users

foochifa:
## Problem

Take sgID out of beta

Closes [#5389]

## Solution

- remove sgID betaflags 
- remove sgID from BetaService
- enable sgID radio buttons for all users

**Breaking Changes** 

-  No - this PR is backwards compatible  

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

For this test, do the following:
 

>  Access a form with an admin account without sgid betaflag (check that betaFlags.sgid does not exist for the admin in the database). Go to settings/Singpass. Select 'Singpass App-only login (Free)'. Save settings and open/activate the form for submission. 

On the framework/submission methods below:

- [ ] On Angular version, with Storage Mode
- [ ] On Angular version, with Email Mode 
- [ ] On React version, with Storage Mode
- [ ] On React version, with Email Mode

As a user, do the following:

> Go to an existing form link (from the above test) with sgid enabled and ensure you can sign in using Singpass QR code and submit the form. 

On the framework/submission methods below:

- [ ] On Angular version, with Storage Mode
- [ ] On Angular version, with Email Mode 
- [ ] On React version, with Storage Mode
- [ ] On React version, with Email Mode

## Deploy Notes

**New scripts**:

- `remove-sgid-flag.js` : deletes sgid betaflag from all users

### DIFF
--- a/frontend/src/constants/links.ts
+++ b/frontend/src/constants/links.ts
@@ -74,5 +74,6 @@ export const LANDING_PAGE_EXAMPLE_FORMS = [
 export const OGP_ALL_PRODUCTS = 'https://www.open.gov.sg/products/overview'
 export const OGP_POSTMAN = 'https://postman.gov.sg'
 export const OGP_FORMSG_COLLATE = 'https://collate.form.gov.sg'
+export const OGP_SGID = 'https://go.gov.sg/sgid-formsg'
 
 export const OGP_FORMSG_REPO = 'https://github.com/opengovsg/formsg'
--- a/frontend/src/constants/links.ts
+++ b/frontend/src/constants/links.ts
@@ -74,6 +74,5 @@ export const LANDING_PAGE_EXAMPLE_FORMS = [
 export const OGP_ALL_PRODUCTS = 'https://www.open.gov.sg/products/overview'
 export const OGP_POSTMAN = 'https://postman.gov.sg'
 export const OGP_FORMSG_COLLATE = 'https://collate.form.gov.sg'
-export const OGP_SGID = 'https://go.gov.sg/sgid-formsg'
 
 export const OGP_FORMSG_REPO = 'https://github.com/opengovsg/formsg'
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
@@ -12,13 +12,11 @@ import { FormAuthType, FormSettings, FormStatus } from '~shared/types/form'
 
 import { BxsHelpCircle } from '~assets/icons/BxsHelpCircle'
 import InlineMessage from '~components/InlineMessage'
-import Link from '~components/Link'
 import Radio from '~components/Radio'
 import Tooltip from '~components/Tooltip'
 
 import { useAdminForm } from '~features/admin-form/common/queries'
 import { isMyInfo } from '~features/myinfo/utils'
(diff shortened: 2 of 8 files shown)


frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx
```diff
@@ -11,7 +11,6 @@ import { Box, Flex, Icon, Skeleton } from '@chakra-ui/react'
 import { FormAuthType, FormSettings, FormStatus } from '~shared/types/form'
 
 import { BxsHelpCircle } from '~assets/icons/BxsHelpCircle'
-import { OGP_SGID } from '~constants/links'
 import InlineMessage from '~components/InlineMessage'
 import Link from '~components/Link'
```
mantariksh: let's set up linting tools on your vscode so you don't encounter these linting issues

shared/types/user.ts
```diff
@@ -13,7 +13,6 @@ export const UserBase = z.object({
   agency: AgencyBase.shape._id,
   betaFlags: z
     .object({
-      sgid: z.boolean().optional(),
     })
```
mantariksh: is there a difference between `z.object({})` and `z.object()`?

src/public/modules/forms/admin/directives/settings-form.client.directive.js
```diff
@@ -144,11 +143,9 @@ function settingsFormDirective(
           )
         }
 
-        $scope.isDisableAuthType = (authType) => {
+        $scope.isDisable = () => {
```
mantariksh: why rename this from `isDisableAuthType` to `isDisable`? actually this function is still specific to disabling the auth type right

shared/types/user.ts
foochifa: I receive an error if i did use `z.object()` instead. 
image

The other option is to remove betaflags entirely, however, I don't think that will be a good option? As we will likely have future betas.

src/public/modules/forms/admin/directives/settings-form.client.directive.js
foochifa: ~The reason I refactored this is because it originally took in an authType argument that is only used when checking for authType.val == sgid.~
~Currently it no longer checks for authType but instead mainly checks if the form is public or if there is an myInfoSP warning.~

~Also I tried to align with the React version which is in AuthSettingsSections.tsx and uses the method name `isDisabled`~

~P.S. I realised I should have named the method as `isDisabled` instead hahaha.~

Have understood that `isDisableAuthType` is to correlate the disable function specifically to the auth/singpass component. Will update the PR! Thanks Anty

shared/types/user.ts
mantariksh: sgtm, let's leave it as empty object

frontend/src/constants/links.ts
```diff
@@ -74,6 +74,5 @@ export const LANDING_PAGE_EXAMPLE_FORMS = [
 export const OGP_ALL_PRODUCTS = 'https://www.open.gov.sg/products/overview'
 export const OGP_POSTMAN = 'https://postman.gov.sg'
 export const OGP_FORMSG_COLLATE = 'https://collate.form.gov.sg'
-export const OGP_SGID = 'https://go.gov.sg/sgid-formsg'
```
mantariksh: thanks for the attention to detail, good that we cleaned this up

mantariksh (approved):

PR #5536, opened 2022-12-06
test: remove example playwright tests

mantariksh:
Annoying when PRs get blocked because the [example tests fail](https://github.com/opengovsg/FormSG/actions/runs/3628053436/jobs/6118629994). No point keeping the examples in the codebase as we can always refer to the online docs and the API might change.


karrui (commented):
Oops

karrui (approved):

PR #5537, opened 2022-12-06
fix: filter out unusable React feedback

mantariksh:
## Problem

Since feedback is compulsory when switching away from React, users often submit whitespace, a few nonsense characters or NIL/NA/Nothing/None for feedback. Filtering through these responses is unnecessary ops work.

## Solution

Filter out responses with fewer than 5 characters or NIL/NA/Nothing/None. In addition, use `mutate` instead of `mutateAsync` so that failed feedback submissions do not cause a frontend error, since this is a non-critical operation.


amitogp (approved):
LGTM

Issue #4889, opened 2022-09-19
upgrade react app dependencies


karrui:
Dependencies have not been upgraded (or only upgraded on adhoc instances) since we were all working on gaining parity with AngularJS. Now that React is *almost* complete, we should take a look at upgrading our dependencies.

This is partially triggered by https://github.com/opengovsg/FormSG/pull/4885.

Big ones to upgrade as of my quick glance at the repository:
- [ ] React 18 (blocker for some dependencies due to the new concurrent/double rendering bugs in strict mode)
- [ ] Chakra v2 (blocked by React 18)
- [ ] react-beautiful-dnd (does not work with 18, but there is a drop-in replacement/fork now in https://github.com/hello-pangea/dnd)

The other dependencies should be able to be updated without trouble.

darrelhong:
might want to look into react-scripts as well #4567

PR #5547, opened 2022-12-08
fix: fix hot reload not working due to missing `process` variable on React

karrui:
## Problem

Our app sometimes fails to hot reload and requires a manual refresh due to a cryptic error where `process` is not defined. The full explanation can be found in SO, see https://stackoverflow.com/questions/70368760/react-uncaught-referenceerror-process-is-not-defined.

This PR pins the `react-error-overlay` package to v6.0.9, at least until we upgrade to CRA5, tracked in #4889 

## Solution

**Breaking Changes** 

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -40970,6 +40970,12 @@
         "node": ">=8"
       }
     },
+    "node_modules/react-dev-utils/node_modules/react-error-overlay": {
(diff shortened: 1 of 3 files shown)


mantariksh (approved):
eh, lgtm

timotheeg (approved):

PR #5558, opened 2022-12-13
feat: feature update for search and filter

justynoh:
## Problem
We need to update What's new for search and filter. This PR adds this.

## Solution
Added animation file for new What's New. Added text for the what's new panel as per design.

We intend to release this tomorrow (Wed 14 Dec), so the date is set as such.

**Breaking Changes** 

## Screenshots

image

## Tests
- [ ] As a user who has not seen any feature updates (`lastSeenFeatureUpdate` flag does not exist), view the dashboard. The red circle should be seen on the What's new link. After clicking the What's new, the `lastSeenFeatureUpdate` should be set to 2 and the red circle should disappear.
- [ ] As a user who has seen the previous feature update (`lastSeenFeatureUpdate` flag is set to `1`), view the dashboard. The red circle should be seen in the What's new link. After clicking What's new, the `lastSeenFeatureUpdate` should be set to 2 and the red circle should disappear.


wanlingt (approved):
lgtm!

while reviewing this i realised we don't have a deadline for removing the what's new content, could be something to consider, maybe something like 3 months? @staceytan1998

PR #5561, opened 2022-12-13
fix: move respondent switch env to the end of the form

justynoh:
## Problem
We want to move the switch env message to the end of the form.

## Solution
Move the switch env message into the form fields component (it could not simply be be moved down within the form wrapper, since children includes the form footer).

## Screenshots
Public without auth
image

Public with auth
image

Preview mode (same as before - should not show the switch env message)
image

### DIFF
--- a/frontend/src/features/public-form/components/FormFields/FormFieldsContainer.tsx
+++ b/frontend/src/features/public-form/components/FormFields/FormFieldsContainer.tsx
@@ -1,10 +1,8 @@
 import { useMemo } from 'react'
 import { Box } from '@chakra-ui/react'
 
-import { FormResponseMode } from '~shared/types'
 import { FormAuthType } from '~shared/types/form/form'
 
-import { useEnv } from '~features/env/queries'
 import { usePublicFormContext } from '~features/public-form/PublicFormContext'
 
 import { FormAuth } from '../FormAuth'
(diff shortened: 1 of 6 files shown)


wanlingt (approved):
lgtm!

justynoh:
closing and reopening to retrigger tests

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
```diff
@@ -42,7 +42,7 @@ export const PublicFormSubmitButton = ({
   }, [formInputs, formFields, formLogics])
 
   return (
-    <Stack px={{ base: '1rem', md: 0 }} pt="2.5rem" pb="4rem">
+    <Stack px={{ base: '1rem', md: 0 }} pt="2.5rem" pb="1.5rem">
```
karrui: should we use a margin here instead so we can use margin collapse on the switcher so we do not need to remember to change this when we remove the switcher?

```tsx
<Stack px={{ base: '1rem', md: 0 }} mt="2.5rem" mb="4rem">
...
</Stack>

// Then in the switch env message component
// -2.5rem so the total spacing between the components is still 1.5rem
<Box mt="-2.5rem">...</Box>
```

frontend/src/features/public-form/components/FormFields/FormFieldsContainer.tsx
```diff
@@ -40,11 +50,41 @@ export const FormFieldsContainer = (): JSX.Element | null => {
     )
   }, [form, handleSubmitForm, isAuthRequired, isLoading])
 
+  // TODO #4279: Computes whether to show the switch env message, remove after React rollout is complete
```
karrui: should we move this into publicswitchenvmessage component? This is sticking around longer than we thought lol.

the component itself can control whether or not it is shown. can pass in form response type as a prop.

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
justynoh: hmm. i didn't cuz need to condition on whether form is authed or not, the authed form bottom padding is different. but yeah, i think since it'll be removed eventually anyway then it's fine haha

karrui: I scared we forget then end up regression

frontend/src/features/public-form/components/FormFields/FormFieldsContainer.tsx
justynoh: [9c49ab2](https://github.com/opengovsg/FormSG/pull/5561/commits/9c49ab2fe57d9b14eb9ec7ae9d03c69116e28023)

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
justynoh: [9c49ab2](https://github.com/opengovsg/FormSG/pull/5561/commits/9c49ab2fe57d9b14eb9ec7ae9d03c69116e28023)

PR #5565, opened 2022-12-14
fix: remove background for desktop mode workspace filter button when filter is active

justynoh:
## Problem
Fixes a design error made in the workspace search and filter component for desktop mode. We do not want a grey background when the filter is active.

## Solution
Remove the check the updates the background color if there is an active filter. 

**Breaking Changes** 


foochifa (approved):
lgtm!

PR #5568, opened 2022-12-14
feat: upgrade axios to 1.2.1

mantariksh:
## Problem

`axios` changed from CJS to ESM exports in `v1.0.0`, resulting in [breaking changes](https://github.com/opengovsg/FormSG/pull/5177) and some teething issues. These issues should be addressed in the latest version. Moreover, we are hoping that a version upgrade will magically fix an issue which some users are facing where submitting forms results in `axios` returning a `Network Error` despite their internet connection being fine.

## Solution

Started by going through the `axios` [changelog for 1.0.0](https://github.com/axios/axios/blob/v1.x/CHANGELOG.md#100---2022-10-04) to verify that we weren't affected by the breaking changes.

Then, see: [yak shaving](https://en.wiktionary.org/wiki/yak_shaving).
1. Upgraded `axios` to latest version. This resulted in [tests failing](https://github.com/opengovsg/FormSG/actions/runs/3691867360/jobs/6250272335). I'm still not really sure why - [this issue](https://github.com/axios/axios/issues/5101) says it's because `jest` is run in NodeJS (which is only CJS-compatible by default), but we have `ts-jest` which [should transform the module correctly](https://kulshekhar.github.io/ts-jest/docs/processing#ts-jest-process).
2. Tried upgrading `jest` because of [this lovely comment](https://github.com/axios/axios/issues/5101#issuecomment-1291347644). Resulted in a missing peer dependency `jest-environment-jsdom`, which stopped being shipped together with `jest` in [28.0.0](https://github.com/facebook/jest/blob/main/CHANGELOG.md#2800).
3. Had to upgrade `ts-jest` as well since its [major versions follow `jest`'s](https://github.com/kulshekhar/ts-jest#versioning).
4. Found out that `ts-jest` stopped exporting `mocked` and `MockedObjectDeep` in [28.0.0](https://github.com/kulshekhar/ts-jest#versioning) because `jest` started providing it [natively](https://jestjs.io/docs/jest-object#jestmockedsource-options). Had to upgrades `@types/jest` and change all imports of `mocked` and `MockedObjectDeep` to `jest.mocked` and `jest.Mocked` respectively.
5. The one frontend test which uses `jest.mocked` was still failing, saying `jest.mocked is not a function`. Dug around the `craco` config to try to figure out how to override the `jest` version, couldn't find an easy way. Finally resorted to workaround of importing `mocked` from `jest-mock` instead, which is fine because it's fully supported.

## Tests
We use `axios` for:
- Every single API call from frontend to backend
- Verifying reCAPTCHA responses through Google reCAPTCHA API
- Sending webhooks

I wrote these tests to try to test all the ways in which we might be using `axios`.

**GET**
- [x] Load admin dashboard

**PUT**
- [x] Create form logic. Make an update to the form logic you just created and check that it goes through correctly.

**POST (application/json)**
- [x] Create a form field

**POST (multipart/form-data)**
- [x] Make a form submission and check that it goes through 

**POST (with query params)**
- [x] Reorder form fields

**PATCH**
- [x] Make an update to form settings and check that settings are updated correctly.

**DELETE**
- [x] Archive a form

**Interceptors**
This test relies on buggy behaviour because I couldn't think of any other easy way to test this.
- [x] Log in to admin dashboard. In a separate tab, log in to a MyInfo form. Remove the `MyInfoCookie` cookie using developer tools. Submit the form and get a 401. Switch back to admin dashboard tab and check that you are logged out.

**reCAPTCHA**
- [x] Check that you can submit a form with recaptcha, assuming you are not a robot.

**Webhooks**
- [x] Send a webhook successfully, check that it is logged correctly in FormSG logs ("Webhook POST succeeded")
- [x] Send a webhook which fails ("Webhook POST failed")

### DIFF
--- a/src/app/loaders/express/__tests__/helmet.spec.ts
+++ b/src/app/loaders/express/__tests__/helmet.spec.ts
@@ -1,5 +1,4 @@
 import helmet from 'helmet'
-import { mocked } from 'ts-jest/utils'
 
 import config from 'src/app/config/config'
 import { sentryConfig } from 'src/app/config/features/sentry.config'
@@ -10,11 +9,11 @@ import helmetMiddlewares from '../helmet'
 
 describe('helmetMiddlewares', () => {
   jest.mock('helmet')
-  const mockHelmet = mocked(helmet, true)
+  const mockHelmet = jest.mocked(helmet)
   jest.mock('src/app/config/config')
-  const mockConfig = mocked(config, true)
+  const mockConfig = jest.mocked(config)
   jest.mock('src/app/config/features/sentry.config')
-  const mockSentryConfig = mocked(sentryConfig, true)
+  const mockSentryConfig = jest.mocked(sentryConfig)
 
   const cspCoreDirectives = {
     imgSrc: [
--- a/src/app/utils/__tests__/limit-rate.spec.ts
+++ b/src/app/utils/__tests__/limit-rate.spec.ts
@@ -1,4 +1,7 @@
-import RateLimit from 'express-rate-limit'
+import RateLimit, {
+  Options as RateLimitOptions,
+  Store,
+} from 'express-rate-limit'
 
 import expressHandler from 'tests/unit/backend/helpers/jest-express'
 
@@ -8,8 +11,29 @@ const MockRateLimit = jest.mocked(RateLimit)
 // eslint-disable-next-line import/first
 import { limitRate } from 'src/app/utils/limit-rate'
 
-const MOCK_MAX = 5
-const MOCK_WINDOW = 10
+/**
+ * express-rate-limit passes this set of options to the
+ * custom handler we specify. When we test our custom
+ * handler, we should make sure that it behaves correctly
+ * when these options are passed in.
+ */
+const MOCK_OPTIONS: RateLimitOptions = {
+  windowMs: 60000,
+  max: 5,
+  message: 'Too many requests, please try again later.',
+  statusCode: 429,
+  legacyHeaders: true,
+  standardHeaders: false,
+  requestPropertyName: 'rateLimit',
+  skipFailedRequests: false,
+  skipSuccessfulRequests: false,
+  keyGenerator: () => '0.0.0.0',
+  handler: (_req, res) => res.status(429).send('Too many requests'),
+  onLimitReached: (_req, res) => res.sendStatus(200),
+  skip: () => false,
+  requestWasSuccessful: () => true,
+  store: {} as Store,
+}
 
 describe('limitRate', () => {
   beforeEach(() => {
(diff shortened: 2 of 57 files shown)


src/app/loaders/express/__tests__/helmet.spec.ts
```diff
@@ -10,11 +9,11 @@ import helmetMiddlewares from '../helmet'
 
 describe('helmetMiddlewares', () => {
   jest.mock('helmet')
-  const mockHelmet = mocked(helmet, true)
+  const mockHelmet = jest.mocked(helmet)
```
mantariksh, line 12: note to reviewer:
the old `mocked` from `ts-jest/utils` has a second arg, `deep`, which specifies whether to mock the object deeply and defaults to `false`.

the new `jest.mocked` has a second arg, `options?: { shallow: boolean }` which defaults to `false`, i.e. deep mocking is enabled by default.

src/app/utils/__tests__/limit-rate.spec.ts
```diff
@@ -35,7 +34,7 @@ describe('limitRate', () => {
     const handler = MockRateLimit.mock.calls[0][0]!.handler!
     const mockNext = jest.fn()
     const mockRes = expressHandler.mockResponse()
-    handler(expressHandler.mockRequest(), mockRes, mockNext)
+    handler(expressHandler.mockRequest(), mockRes, mockNext, {})
```
timotheeg: Why is this needed?

timotheeg (approved):
LGTM 👍 

So all the mocks are deep mocks now, no issue with that, correct?

src/app/utils/__tests__/limit-rate.spec.ts
mantariksh: Haha, initially I just put that to stop the types from complaining, but since you commented...

`RateLimit` is a function which returns an `express` rate-limiting middleware. `RateLimit` accepts many options to configure this middleware, one of which is a custom `handler` to call when the rate limit is exceeded.

This particular test in `limit-rate.spec.ts` checks that our custom handler behaves correctly.

`express-rate-limit` passes 4 args to the custom `handler`: `req`, `res`, `next` as well as the object of `options` which was used to set up the middleware.

Previously, this test didn't pass the `options` to the handler when testing its behaviour. However, ideally the test should match the behaviour of `express-rate-limit` which is to pass those options. So I've updated the test to do this in [652077e](https://github.com/opengovsg/FormSG/pull/5568/commits/652077ecbf625d4e3ff729c6fe2a711dcd3f8ea8).

mantariksh:
@timotheeg 

> So all the mocks are deep mocks now, no issue with that, correct?

That's right. We generally don't export nested functions anyway, so deep/shallow shouldn't make a difference.

timotheeg (approved):
Looking good! Thanks for the explanations @mantariksh !

PR #5579, opened 2022-12-16
fix: return 404 for unmatched static assets

### LINKED ISSUE #5577: Catch-all router should still report 404s
The catch-all router serves the main react app (html content). This is too aggressive. In particular, requests for assets under `/static` that reach the catch-all handler should get 404 responses.

timotheeg:
## Problem
The catch-all handler we have for react (which is mostly used to catch all the admin routes) is hiding 404 responses because it always returns 200 with html content.

We know some users sometimes see blank pages and it's likely during release when there's a mismatch between static assets (old assets requested on new server, or new assets requested on old server).

fixes #5577 

Additional info: the reason we have a catch-all in react is to capture all the admins routes without having to list them explicitly in express. Listing all app routes (admins and respondent) rather than `*` would be an alternative solution, since then the final catcher would be the error handler, which would return 404 for all non-matched routes.

## Solution
This PR doesn't fix the problem (this will come in a separate process), but it restored our visibility into 404s for static assets.

Tested locally:
```bash
➜ curl -v http://localhost:5001/617acb2cf1180300148af690 2>&1 | grep '< HTTP'
< HTTP/1.1 200 OK

➜ curl -v http://localhost:5001/static/ 2>&1 | grep '< HTTP'
< HTTP/1.1 404 Not Found

➜ curl -v http://localhost:5001/static/foo 2>&1 | grep '< HTTP'
< HTTP/1.1 404 Not Found

➜ curl -v http://localhost:5001/foo.bar 2>&1 | grep '< HTTP'
< HTTP/1.1 404 Not Found

➜ curl -v http://localhost:5001/robots.txt 2>&1 | grep '< HTTP'
< HTTP/1.1 200 OK

➜ curl -v http://localhost:5001/static/js/1.e7cd9c85.chunk.js 2>&1 | grep '< HTTP'
< HTTP/1.1 200 OK

➜ curl -v http://localhost:5001/static/js/1.aaaaaaaa.chunk.js 2>&1 | grep '< HTTP'
< HTTP/1.1 404 Not Found
```

### DIFF
--- a/src/app/loaders/express/error-handler.ts
+++ b/src/app/loaders/express/error-handler.ts
@@ -23,92 +23,70 @@ const isHTTPLikeError = (
 }
 
 // 404 since no middleware responded
-export const catchNonExistentRoutesMiddleware: RequestHandler = function (
-  _req,
-  res,
-) {
+export const catchNonExistentRoutesMiddleware: RequestHandler = (_req, res) => {
   res.sendStatus(StatusCodes.NOT_FOUND)
 }
 
-const errorHandlerMiddlewares = (): (
-  | ErrorRequestHandler
-  | RequestHandler
-)[] => {
-  // Assume 'not found' in the error msgs is a 404. this is somewhat silly, but valid, you can do whatever you like, set properties, use instanceof etc.
-  const genericErrorHandlerMiddleware: ErrorRequestHandler = function (
-    err,
-    req,
-    res,
-    next,
-  ) {
-    // If headers have already been sent, don't send again
-    if (res.headersSent) {
-      return
+export const genericErrorHandlerMiddleware: ErrorRequestHandler = (
+  err,
+  req,
+  res,
+  next,
+) => {
+  // If headers have already been sent, don't send again
+  if (res.headersSent) {
+    return
+  }
+
+  // If the error object doesn't exists
+  if (!err) {
+    return next()
+  } else {
+    const genericErrorMessage =
+      'Apologies, something odd happened. Please try again later!'
+
+    const log_meta = {
+      action: 'genericErrorHandlerMiddleware',
+      // formId is only present for Joi validated routes that require it
+      formId: get(req, 'form._id', null),
     }
 
-    // If the error object doesn't exists
-    if (!err) {
-      return next()
-    } else {
-      const genericErrorMessage =
-        'Apologies, something odd happened. Please try again later!'
-
-      const log_meta = {
-        action: 'genericErrorHandlerMiddleware',
-        // formId is only present for Joi validated routes that require it
-        formId: get(req, 'form._id', null),
-      }
-
-      // Error page
-      if (isCelebrateError(err)) {
-        logger.error({
-          message: 'Joi validation error',
-          meta: {
-            ...log_meta,
-            details: Object.fromEntries(err.details),
-          },
-          error: err,
-        })
-        return celebrateErrorHandler(err, req, res, next)
-      }
-
-      if (isBodyParserError(err)) {
-        logger.error({
-          message: 'Body parser error',
-          meta: {
-            ...log_meta,
-            details: {
-              type: err.type,
-              message: err.message,
-            },
-          },
-          error: err,
-        })
-        if (err.expose) {
-          return res
-            .status(err.status)
-            .json({ message: err.message ?? genericErrorMessage })
-        }
-        return res.status(err.status).json({ message: genericErrorMessage })
-      }
-
-      if (isHTTPLikeError(err)) {
-        logger.error({
-          message: `HTTP Error ${err.statusCode}`,
-          meta: {
-            ...log_meta,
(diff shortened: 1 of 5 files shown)


src/app/loaders/express/error-handler.ts
```diff
@@ -22,6 +22,14 @@ const isHTTPLikeError = (
   return types.isNativeError(err) && 'statusCode' in err
 }
 
+// 404 since no middleware responded
+export const catchNonExistentRoutesMiddleware: RequestHandler = function (
```
mantariksh: ha, we probably missed this file while converting all functions to arrow functions. can fix another time

src/app/loaders/express/error-handler.ts
```diff
@@ -115,14 +123,6 @@ const errorHandlerMiddlewares = (): (
...
-  // Assume 404 since no middleware responded
-  const catchNonExistentRoutesMiddleware: RequestHandler = function (
-    _req,
-    res,
-  ) {
-    res.sendStatus(StatusCodes.NOT_FOUND)
-  }
-
   return [genericErrorHandlerMiddleware, catchNonExistentRoutesMiddleware]
 }
```
mantariksh, line 117: tiny nit: in an ideal world I would remove line 129 and change `errorHandlerMiddlewares` to a named export just so it's clear what the difference is between the multiple exports from this file. but not a big deal

mantariksh (approved):
just to confirm, the browser behaviour will be the same right? users will still see a blank page, just that this time it's because we're serving a 404 instead of html

src/app/loaders/express/index.ts
```diff
@@ -131,10 +133,19 @@ const loadExpressApp = async (connection: Connection) => {
   // New routes in preparation for API refactor.
   app.use('/api', ApiRouter)
 
+  // serve static assets
```
tshuli:
```suggestion
  // serve static assets. `dist/frontend` contains the root files as well as a `/static` folder
  // express.static calls next() if the file is not found
```

tshuli: suggested documentation for future devs

src/app/loaders/express/index.ts
```diff
@@ -131,10 +133,19 @@ const loadExpressApp = async (connection: Connection) => {
...
   app.use('/public', express.static(path.resolve('dist/angularjs')))
-  app.get('/old/', HomeController.home)
 
+  // Requests for known static asset patterns which were not served by
+  // the static handlers above should return 404s
+  app.get(/^\/(public|static)\//, catchNonExistentRoutesMiddleware)
+
+  // Requests for root files (e.g. /robots.txt or /favicon.ico) that were
+  // not served statically above will also return 404
+  app.get(/^\/[^/]+\.[a-z]{2,}$/, catchNonExistentRoutesMiddleware)
```
tshuli: No objections per se but just wondering why we need to be so specific with the filename format - are there no legitimate cases in future where we could have:
- filename with no extension
- filename with single-character extension

tshuli (approved):
lgtm, with just some minor comments!

timotheeg:
> just to confirm, the browser behaviour will be the same right? users will still see a blank page, just that this time it's because we're serving a 404 instead of html

Correct! This PR is not a functional fix from the user perspective, it's an observability fix for us to have visibility on errors.

src/app/loaders/express/index.ts
timotheeg: Aye, I wanted to capture files with dots specifically `{name}.{extension}` so I wouldn't risk matching forms with urls `/{formid}` (those are not meant to be served statically, but also should not be 404s).

I can drop the extension specificity though, this is indeed not useful!

src/app/loaders/express/error-handler.ts
timotheeg: I'll do it!

timotheeg: I wonder why the 2 handlers in `errorHandlerMiddlewares` were defined as local functions btw I'll convert them both to module functions for consistency.

timotheeg: Done in https://github.com/opengovsg/FormSG/pull/5579/commits/5cc48bfe3cd4f037fab3ff12a7893e07656805d4

src/app/loaders/express/index.ts
timotheeg: Done in https://github.com/opengovsg/FormSG/pull/5579/commits/bbd5ab5d0c7c76bc79d6889f0df430988c88ae51

src/app/loaders/express/error-handler.ts
timotheeg: Fixed in https://github.com/opengovsg/FormSG/pull/5579/commits/5cc48bfe3cd4f037fab3ff12a7893e07656805d4 too

PR #5581, opened 2022-12-16
build: hotfix v6.23.1 for whitespace in email field and form title

mantariksh:
## Changes
- fix(deps): bump formsg-javascript-sdk from 0.9.0 to 0.10.0 [`#5578`](https://github.com/opengovsg/FormSG/pull/5578)
- build: release v6.23.0 [`#5572`](https://github.com/opengovsg/FormSG/pull/5572)
- fix: trim email input in frontend and backend [`9662591`](https://github.com/opengovsg/FormSG/commit/96625918646779c5575aad563ffd521fd44776bc)

## Problems

1. Email field validation allows trailing/leading whitespace in the frontend, but not on the backend. This results in many errors where users submit whitespace with no frontend error, but their submission is rejected by the backend.
2. Submission decryption in storage mode fails when field title contains only whitespace.

## Solutions

This hotfix fixes both issues by respectively
1. Trimming email inputs on the frontend before submission, and on the backend before validation
2. Bumping to the latest version of `formsg-javascript-sdk` which allows white-space only field titles

## Tests
- [x] Create an email field in an email mode form with OTP verification. Fill in an email with Attempt to verify OTP. Check that in the network payload to the `/generate` endpoint, the email is trimmed.
- [x] Continue to verify OTP and submit the email with whitespace. Check that in network payload to submissions endpoint, email is trimmed. Submission should succeed.
- [x] Repeat above for storage mode.
- [x] Go back to email mode field and make a manual network request with an email field answer where there is a trailing whitespace. Submission should succeed.
- [x] Edit a storage mode form in the staging database to have a field where the title is whitespace. Submit the form and decrypt the response. Response should be decryptable.


src/app/utils/field-validation/validators/emailValidator.ts
```diff
@@ -38,7 +38,9 @@ const makeEmailDomainValidator: EmailValidatorConstructor =
     const emailAddress = String(answer)
     if (!(isVerifiable && hasAllowedEmailDomains && allowedEmailDomains.length))
       return right(response)
-    const emailDomain = ('@' + emailAddress.split('@').pop()).toLowerCase()
+    const emailDomain = ('@' + emailAddress.split('@').pop())
+      .toLowerCase()
+      .trim()
```
timotheeg: This is not the error we were seeing though is it? Don't we need to trim before the `isEmail()` check line 24 too?

timotheeg: Just for extra info, the domain error is:
`EmailValidator:\t answer is not a valid email domain`
while the plain email error is
`EmailValidator:\t answer is not a valid email`

image

mantariksh: dang, you are right sorry my bad

mantariksh: rewrote history, fixed in [9662591](https://github.com/opengovsg/FormSG/pull/5581/commits/96625918646779c5575aad563ffd521fd44776bc)

src/app/utils/field-validation/validators/emailValidator.ts
```diff
@@ -21,7 +21,7 @@ type EmailValidatorConstructor = (
  */
 const emailFormatValidator: EmailValidator = (response) => {
   const { answer } = response
-  return isEmail(answer)
+  return isEmail(answer.trim())
```
timotheeg, line 24: Hmm, might need an additional casting to string to prevent possible run-time errors? 🤔 
```suggestion
  return isEmail(String(answer).trim())
```

timotheeg (approved):

src/app/utils/field-validation/validators/emailValidator.ts
mantariksh: this is already checked before this function is called, in `isProcessedSingleAnswerResponse`

PR #5582, opened 2022-12-16
feat: add form type, user agent and attachment types to public switchback form 

### LINKED ISSUE #5573: add more details to the switchback feedback form
Add the browser + device types and version ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgent))
Add the attachment file types

wanlingt:
## Problem

More details about the problem each user is facing will help us to troubleshoot errors more effectively

Closes #5573

## Solution

Add the following fields to the feedback form:
- form response mode
- form auth type
- browser and device information using [userAgent](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgent)
- attachment file types

**Note**
In the [userAgent MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgent), it states that the `userAgent` data is unreliable as the user agent string is user configurable. We'll still go with this approach given that the feedback form is temporary. We could also refine this further (e.g. [using duck-typing methods](https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browsers)) in the future if we find that the feedback form data is too unreliable.

Some other resources:
- [list of user agent strings](https://deviceatlas.com/blog/list-of-user-agent-strings)
- [user agent string syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#syntax)

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:
image

image

## Tests

1. Fill in the feedback modal, check that the submission is successful. Variations to test are
- [x] No checkbox selected
- [x] Only 1 non-others checkbox option selected
- [x] >1 non-others checkbox options selected
- [x] 1 non-others + 1 Others checkbox options selected
- [x] Only others checkbox option selected, without text input (validation error should appear)
- [x] Only others checkbox option selected, with text input

2.
- [x] Fill in the feedback modal again on another device and browser. This info should be reflected differently in the userAgent field

## Deploy Notes

Add the new optional short text fields to the feedback form - `attachmentType`, `userAgent`, `responseMode`, `authType`


frontend/src/features/public-form/components/FormFields/FormFieldsContainer.tsx
```diff
@@ -58,6 +58,7 @@ export const FormFieldsContainer = ({
         <PublicSwitchEnvMessage
           responseMode={form?.responseMode}
           isAuthRequired={isAuthRequired}
+          authType={form?.authType}
```
mantariksh, line 61: `form` should be defined here right, since there's a check on line 31?

mantariksh (commented):
code lgtm with one small nit. did we consider making the attachment types field a checkbox with "others" instead? anecdotally the vast majority of attachments are PDFs, zips and screenshots, so I wonder if we might get better data if we reduce the friction of indicating the attachment type

frontend/src/features/public-form/components/FormFields/FormFieldsContainer.tsx
timotheeg: Looks like the form check line 31 only affects the `renderFields` memo, so it's not guaranteed to be defined here (curse indentation! 😜)

timotheeg (approved):
I think this looks good for now.

I agree with Anty's comment on reducing friction though, and we could refine the attachment type later 🤔 

In fact, are we able to get the file names (and loosely extract the type from them) from the visible form fields? I reckon many users may not know the file types of whatever they are attaching, so if we can get the data transparently, it'd be ideal.

wanlingt:
> I think this looks good for now.
> 
> I agree with Anty's comment on reducing friction though, and we could refine the attachment type later 🤔
> 
> In fact, are we able to get the file names (and loosely extract the type from them) from the visible form fields? I reckon many users may not know the file types of whatever they are attaching, so if we can get the data transparently, it'd be ideal.

yes good point @mantariksh on reducing friction, i've replaced the input with checkboxes for the attachment type field. @timotheeg will look into extracting the file names in a separate PR!

frontend/src/features/env/PublicFeedbackModal.tsx
```diff
@@ -80,6 +91,28 @@ export const PublicFeedbackModal = ({
...
+  const othersValidationRules = useMemo(
+    () => ({
+      validate: (value?: string | string[]) => {
+        const currCheckedVals = getValues(checkboxInputName)
+        return (
+          !(
+            Array.isArray(currCheckedVals) &&
+            currCheckedVals.includes(CHECKBOX_OTHERS_INPUT_VALUE)
+          ) ||
+          !!value ||
```
timotheeg: If `value` is of type `string[]`, this test will always be true. 

Also, do we need to `trim()` before validation?

frontend/src/features/env/PublicFeedbackModal.tsx
```diff
@@ -80,6 +91,28 @@ export const PublicFeedbackModal = ({
...
 
+  const checkboxInputName = 'attachmentType'
+
+  const CHECKBOX_OTHERS_INPUT_VALUE =
+    '!!FORMSG_INTERNAL_CHECKBOX_OTHERS_VALUE!!'
+  const ATTACHMENT_TYPE_OPTIONS = ['PDF', 'JPEG', 'PNG', 'ZIP']
+
+  const othersValidationRules = useMemo(
+    () => ({
+      validate: (value?: string | string[]) => {
```
timotheeg: When would value be `string[]`?

frontend/src/features/env/EnvService.ts
```diff
@@ -23,7 +26,25 @@ const createFeedbackResponsesArray = (
...
   return feedbackFormFieldsStructure.map((question, i) => {
     const { _id, fieldType } = feedbackForm.form.form_fields[i]
-    const answer: string = formInputs[question] ?? ''
+    if (fieldType === BasicField.Checkbox) {
+      const answerArray: string | string[] = formInputs[question] ?? []
+      if (formInputs[othersInputName] && Array.isArray(answerArray)) {
+        if (formInputs[question] === '[""]') {
+          // Remove case where non-Others checkbox is selected
+          answerArray.pop()
+        }
```
timotheeg: Apologies, this code confuses me 😢 

`answerArray` is defined from `formInputs[question]`, so they have the same value; unless `formInputs[question]` is `undefined` (in which case `answerArray` would be an empty array).

So, when `answerArray` is an array (per the conditional being checked), then either `formInputs[question]` was an array, or `formInputs[question]` was `undefined`. In both cases `formInputs[question]` can never be equal to the string `[""]`, and the nested conditional block never runs, correct?

Could you help me understand how this works / is supposed to work? 😅

frontend/src/features/env/PublicFeedbackModal.tsx
wanlingt: changed the type to `string` and added a `trim()` [here](https://github.com/opengovsg/FormSG/pull/5582/commits/ce1650222cf66faacd5dcbdab9d5cbb5e074d27d#diff-b2271c6e3f57c5e2f1759290092b864d15724f49237d1c5348bfc51ee5eac5a1R102-R109)!

frontend/src/features/env/PublicFeedbackModal.tsx
wanlingt: `string` only oops, modified [here](https://github.com/opengovsg/FormSG/pull/5582/commits/ce1650222cf66faacd5dcbdab9d5cbb5e074d27d#diff-b2271c6e3f57c5e2f1759290092b864d15724f49237d1c5348bfc51ee5eac5a1R102-R109)

frontend/src/features/env/EnvService.ts
wanlingt: sorry this was so confusing! I've made changes to the conditional and added more comments here [ce16502](https://github.com/opengovsg/FormSG/pull/5582/commits/ce1650222cf66faacd5dcbdab9d5cbb5e074d27d), hope it makes sense now!!

frontend/src/features/env/EnvService.ts
```diff
@@ -23,6 +26,25 @@ const createFeedbackResponsesArray = (
...
     const { _id, fieldType } = feedbackForm.form.form_fields[i]
+    // Checkbox answers have to be in an answerArray
+    if (fieldType === BasicField.Checkbox) {
+      // if no checkbox is selected, return an empty array
+      const answerArray: string[] = formInputs[question] || []
+
+      // if Others is selected, remove '!!FORMSG_INTERNAL_CHECKBOX_OTHERS_VALUE!!'
+      // which is the last value in the array when 'Others' is selected,
+      // and replace it with the value entered in the Others Input field
+      if (formInputs[othersInputName]) {
```
mantariksh, line 39: this will break if we ever have multiple checkboxes with "Others" in this modal right? unlikely but maybe we should just leave a comment here in case we forget

wanlingt: yes correct! okie will add a comment

shared/types/env.ts
```diff
@@ -1,15 +1,22 @@
...
+
 export enum UiCookieValues {
   React = 'react',
   Angular = 'angular',
 }
 
 // TODO #4279: Remove after React rollout is complete
 export interface PublicFeedbackFormDto {
-  [key: string]: string
+  [key: string]: any
```
mantariksh, line 10: nice

mantariksh (approved):
lgtm, would be good to add some manual tests for different combinations of checkbox options (e.g. regular options without Others, regular options with Others, Others only, Others without text input...) just to make sure we're handling all the weird "Others" stuff correctly

frontend/src/features/env/EnvService.ts
wanlingt: [486ada2](https://github.com/opengovsg/FormSG/pull/5582/commits/486ada2f79e65c1e58bba9bbdaf4b8c0a0b262ba)

PR #5595, opened 2022-12-20
feat: push static assets to s3

tshuli:
## Problem
With each new deployment, new JS bundles are generated for the frontend. These js bundles have a unique filename (hash digest). During a rolling deployment where both old and new server instances are concurrently accepting traffic, users have a high chance of not being able to retrieve the js bundle, leading to a blank screen.

## Solution
- At each build, push the generated static assets to s3 bucket. 
- When client requests a static asset, server will attempt to serve from locally. If that fails, server will attempt to retrieve from s3 bucket

In Github Actions
- `S3_STATIC_ASSETS_BUCKET_NAME`

In Parameter Store
- `STATIC_ASSETS_S3_BUCKET `

It is necessary to duplicate the env vars as they are required at build time and on the backend.
1) `S3_STATIC_ASSETS_BUCKET_NAME` is used to push the static files to S3 and configure craco at build time. 
2) `STATIC_ASSETS_S3_BUCKET` is used on the back-end to set the S3 bucket url for server to try and retrieve the assets if they are not present locally.


.github/workflows/deploy-eb.yml
```diff
@@ -97,12 +99,22 @@ jobs:
...
 
+      - name: Copy Static Assets from Docker Image and Push to S3
+        env:
+          S3_STATIC_ASSETS_BUCKET_NAME: ${{ secrets.S3_STATIC_ASSETS_BUCKET_NAME }}
+          ECR_REPOSITORY: ${{ secrets.ECR_REPO }}
+        run: |
+          mkdir -p /tmp/s3static
+          chmod 777 /tmp/s3static
+          docker run --rm -v /tmp/s3static:/tmp/s3static $ECR_REPOSITORY:$IMAGE_TAG cp -r ./dist/frontend/. /tmp/s3static
+          aws s3 sync /tmp/s3static s3://$S3_STATIC_ASSETS_BUCKET_NAME
```
timotheeg, line 114: Since the assets have hash in their names and are therefore immutable, we should give them a long expiry (10 years?) with a cli arg `--cache-control max-age=315360000`

.github/workflows/deploy-eb.yml
```diff
@@ -97,12 +99,22 @@ jobs:
...
           aws s3 cp $IMAGE_TAG.zip s3://$BUCKET_NAME/$IMAGE_TAG.zip
 
+      - name: Copy Static Assets from Docker Image and Push to S3
+        env:
+          S3_STATIC_ASSETS_BUCKET_NAME: ${{ secrets.S3_STATIC_ASSETS_BUCKET_NAME }}
+          ECR_REPOSITORY: ${{ secrets.ECR_REPO }}
+        run: |
+          mkdir -p /tmp/s3static
+          chmod 777 /tmp/s3static
+          docker run --rm -v /tmp/s3static:/tmp/s3static $ECR_REPOSITORY:$IMAGE_TAG cp -r ./dist/frontend/. /tmp/s3static
```
timotheeg, line 113: Just checking: this is using the local image, not redownloading it from AWS, correct?

timotheeg (approved):
LGTM

.github/workflows/deploy-eb.yml
tshuli: yes, local image

.github/workflows/deploy-eb.yml
tshuli: KIV for now, since there are non-immutable assets which don't have hash (e.g. static images)

timotheeg (approved):
+1 again

src/app/loaders/express/error-handler.ts
```diff
@@ -22,9 +24,41 @@ const isHTTPLikeError = (
...
+      message: 'Serving static asset from S3',
+      meta: {
+        action: 'catchNonExistentStaticRoutesMiddleware',
+        url: req.originalUrl,
+      },
+    })
+    data.pipe(res)
+  } catch (err) {
+    // Else return 404
+    logger.error({
```
timotheeg, line 52: The error here is not necessarily a 404, correct? which is why we are logging?

404s will be logged by express anyways and we'll know it comes from here.

tshuli: Yup - in fact it will be a 403 from s3 if the item doesn't exist in s3

tshuli: Added doc in [012ce00](https://github.com/opengovsg/FormSG/pull/5595/commits/012ce00898f4398cccccf9abca2752572015e0cb)

PR #5596, opened 2022-12-20
fix: view full public form page when printed 

### LINKED ISSUE #5511: Print and "save as PDF" does not render properly for React
Reference Ticket: https://formsg.zendesk.com/agent/tickets/16447 

When the table field is the last field in a form and printed, the last row and the form footer gets truncated

https://go.gov.sg/form-github-issue-pdfprint-2022

wanlingt:
## Problem

A public form page in print mode is sometimes truncated at the end if tables with multiple columns are present in the form.

Also, there is inconsistency in the column headers printed - Safari and Firefox print tables without duplicated column headers, while Chrome does.

Closes #5511

## Solution

It's likely that the table is truncated because the browser renders its height inaccurately in print mode. Figured this out after multiple attempts at setting various css media print queries in the elements on the public form page, e.g. `display: 'block'`, `overflow: 'visible'`, `break-inside` / `break-after` / `break-before` and `height`

To solve this, I
1. set the height of the table field explicitly in the print media query (`printTableHeight`) - this height varies according to the number of rows and columns. 
2. added `breakInside: 'avoid'` to `Td` to prevent each table cell from getting truncated
3. added new `useIsPrint` hook to check when the page is in print mode. If yes, print the column headings for each table cell.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests

- [x] Create a form that has a table as its last field. On a separate branch, check that this table is truncated when printed. Returning back to this branch, this table should no longer be truncated; the full form should appear - including the submit button and the form footer


frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -135,6 +135,11 @@ export const TableField = ({
     return description
   }, [fields.length, schema.addMoreRows, schema.maximumRows])
 
+  // If a table is the last field, Chrome and MS Edge sometimes truncate the table in print mode.
+  // We set the height of the table explicitly (depending on the number of cols and rows)
+  // for use in the media query so that the browser is able to render the full table in print mode
+  const printTableHeight = schema.columns.length * rows.length * 7.25
```
karrui: Should we not hardcode the height? Will it work if we use the exact height of the table itself? We could set a ref on the table and get its current height. Or is 7.25rem a product of experimentation?

wanlingt: yeah the thing is it's not the table height as it appears on the screen, cos when we print it becomes one column
e.g. 
image
becomes
image

7.25rem was partially a product of experimentation - i added up the height of each cell + padding + margins, and gave a bit more buffer

karrui (approved):
Lgtm

timotheeg:
Just adding a comment that we're discussing if we can identify what might cause the table height to be computed incorrectly rather and tweak/trick that with custom rules.

The issue with computing the table height is that the header behaviour is not consistent. Chrome will show the field headers multiple times while safari does not, and so it's hard to consistently account for a row/cell height.

wanlingt:
> The issue with computing the table height is that the header behaviour is not consistent. Chrome will show the field headers multiple times while safari does not, and so it's hard to consistently account for a row/cell height.

fixed this in [34aa54c](https://github.com/opengovsg/FormSG/pull/5596/commits/34aa54c4c51df814d7f2b3816e1677e3d3030a11), now headers show up if the page is in print mode

timotheeg:
One more note from our tests earlier. 

So it looks like chrome:
1. Computes the number of page to print without realizing that the table display will change from columns to row and thus is under counting (by a lot!)
2. Proceed to render **correctly** based on the size available on the print view. Sadly, whatever should be rendered at the end (fields, footer, etc.) may be clipped away, simply because there is no page to render into 😑 

In all our tests, we manage to have a last page truncation happen, but if the theory above holds, a form with several tables (or a tables with **many** rows) could have its overall height so wrong that multiple pages are missing at the end.

frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -135,6 +135,13 @@ export const TableField = ({
     return description
   }, [fields.length, schema.addMoreRows, schema.maximumRows])
 
+  // If a table is the last field, Chrome and MS Edge sometimes truncate the table in print mode.
```
timotheeg: From our experiments, looks like height for all tables is not computed correctly, so the comment should be updated to remove `is the last field`

timotheeg (commented):
Looking good. Just 2 minor comments from me

frontend/src/templates/Field/Table/TableField.tsx
wanlingt: [023dee2](https://github.com/opengovsg/FormSG/pull/5596/commits/023dee2d290577a3d91eeb0cedc91ef2939b9970)

frontend/src/templates/Field/Table/ColumnCell.tsx
```diff
@@ -138,7 +140,8 @@ export const ColumnCell = ({
       isInvalid={!!get(errors, inputName)}
     >
       <FormLabel
-        as={isMobile ? undefined : VisuallyHidden}
+        // display column header in print and mobile modes
+        as={isPrint ? undefined : isMobile ? undefined : VisuallyHidden}
```
timotheeg: Heeeeeyyy?? github dropped my second comment 😭 Am not sure what happened but I'm sorry about it. I had this extra nit:
```suggestion
        as={isPrint || isMobile ? undefined : VisuallyHidden}
```

timotheeg: I pushed the change to save a bit of time 😅

timotheeg (approved):

frontend/src/templates/Field/Table/ColumnCell.tsx
wanlingt: sorry!!! was afk

PR #5597, opened 2022-12-20
test: fix selectors for tests

justynoh:
## Problem
Playwright is failing, possibly because we upgraded @playwright/test.

## Solution
Some selectors were breaking, so this PR fixes them.

### DIFF
--- a/__tests__/e2e/constants/field.ts
+++ b/__tests__/e2e/constants/field.ts
@@ -285,15 +285,15 @@ export const ALL_FIELDS: E2eFieldMetadata[] = [
         },
       },
       {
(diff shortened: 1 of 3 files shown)


timotheeg (approved):

timotheeg:
I didn't see you change the test data too. What's up with that? 😅

justynoh:
playwright was being annoying with case-sensitivity. In the old test cases it was "Female" and "Male", so looking for the string "Male" would match both locations for the buttons "Fe**male**" and "**Male**"

So i changed the test data to get rid of the issue hahaha

PR #5605, opened 2022-12-21
fix: avoid using merge to set query after reordering fields

### LINKED ISSUE #5603: Moving fields causes rendering issue for fields with 'options'
**Describe the bug**
When moving fields in a form that has fields with options (radio/checkbox/dropdown etc) followed by another field with more options. 

The fields that are passed by the shifting field may render the options of the larger field into the radio/checkbox/dropdown field as additional options.

**To Reproduce**
1. Go to a form with radio fields followed by checkbox fields (with more options than the preceding radio field). ([form](https://form.gov.sg/639299ffa78b2700113e78d2) where bug was found)
2. Either create a new field or select the fields at the bottom
3. Move the field up, passing by the radio field in pt 1.
4. After shifting the fields, go back to check radio field, some of the options items from the following checkbox field may have become additional radio buttons.

(Can substitute radio/checkbox freely and bug should still be reproducible) 

**Expected behavior**
Fields should remain unchanged when moving other fields around.

**Screenshots**

BEFORE
image

AFTER
image

**Additional context**
The issue only happens on the frontend, upon refreshing the page, the fields will return back to normal. However, form admins get confused by the bug and will go in and try to change the radio fields.

Also, the form will return to normal after a while.

justynoh:
## Problem
Reordering fields with options with different lengths causes the excess options in the field with longer options to be replicated in the field with shorter options.

https://user-images.githubusercontent.com/25571626/208871283-c46de3ca-a5d3-45cb-81a4-af13271d8588.mov

Closes #5603 

## Solution
The reason for this is that we use lodash's `merge` when we set the query. This modifies the old array values in-place, and does not replace the array. Therefore, avoid using `merge`.  

**Breaking Changes** 

https://user-images.githubusercontent.com/25571626/208871811-932bd2cf-479c-490a-bd9b-fd47ba1ce691.mov

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/mutations/useReorderFormField.ts
+++ b/frontend/src/features/admin-form/create/builder-and-design/mutations/useReorderFormField.ts
@@ -1,7 +1,6 @@
 import { useCallback } from 'react'
 import { useMutation, useQueryClient } from 'react-query'
 import { useParams } from 'react-router-dom'
-import { merge } from 'lodash'
 
 import { FormFieldDto } from '~shared/types/field'
 import { AdminFormDto } from '~shared/types/form'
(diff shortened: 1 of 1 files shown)


mantariksh (approved):
image
TM

justynoh:
big LG

PR #5612, opened 2022-12-22
build: Release 6.23.3 hotfix

tshuli:
## Problem
- In 6.23.2, we introduced reverse proxy functionality to stream files from s3 
- However, we did not pass on the headers from s3, particularly `content-type`
- This led to MIME type errors 
- This hotfix includes the headers from s3 in the response to the client

As there are no changes to the frontend code, the js bundles have not changed


timotheeg (approved):
Code looks good to me. Update to changelog is a bit weird, but we can probably ignore (?)

tshuli:
fixed the changelog - was because 6.24.0 tag was in the repo

timotheeg (approved):

PR #5633, opened 2022-12-23
fix: remove deprecated GA command set-output

timotheeg:
## Problem
Our actions and some of the actions we import use the deprecated command `set-output`. Github is warning us it will be removed soon and we should use the new setup, per the instructions [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

image

image

## Solution
* Upgrade `set-output` commands
* Upgrade imported actions that use `set-output`


timotheeg:
The warning on the `configure-aws-ctredentials` is still there, I need to dig deeper if it has been fixed upstream and if there's a way for us to clear the warning.

justynoh (approved):
lgtm

PR #5659, opened 2022-12-27
feat: add console.error to instances of arrayBuffer in the frontend

### LINKED ISSUE #5656: Add try catch handling around arraybuffer references in our code
Will hopefully help us identify the source of error (react app code or compression libraries)

related to #5552

wanlingt:
## Problem

We want to investigate the arrayBuffer issue by documenting where the error occurs

Closes #5656

## Solution

Add `console.error` to the 2 instances where arrayBuffer appears in the frontend React code

### DIFF
--- a/frontend/src/features/public-form/utils/createSubmission.ts
+++ b/frontend/src/features/public-form/utils/createSubmission.ts
@@ -180,7 +180,11 @@ const encryptAttachment = async (
     }
     return { id, encryptedFile: encodedEncryptedAttachment }
   } catch (error) {
+    // TODO: remove error logging when error about arrayBuffer not being a function is resolved
     console.error(error)
+    console.error(`Information about attachment ${id}`)
+    console.error(typeof attachment)
+    console.error(attachment)
     return { id, encryptedFile: undefined }
   }
 }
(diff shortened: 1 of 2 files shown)


frontend/src/features/public-form/utils/createSubmission.ts
```diff
@@ -166,17 +166,21 @@ const encryptAttachment = async (
...
+      publicKey,
+    )
+    const encodedEncryptedAttachment = {
+      ...encryptedAttachment,
+      binary: encodeBase64(encryptedAttachment.binary),
+    }
+    return { id, encryptedFile: encodedEncryptedAttachment }
+  } catch (error) {
+    console.error(error)
+    return { id, encryptedFile: undefined }
```
timotheeg: Hmm, do we really want to do that? Does that create an artificial success for submission, while missing the attachment?

If we catch the error, it might not be reported to DD any more? I thought we would only rethrow 🤔 .

Also if there is an exception, and we assume it's `arrayBuffer is not a function`, we will want to print more information about the context of the error, not just the error (which we already have in DD). If `attachment` is a `File` (per TS definition above), then it *should* have a `.arrayBuffer()` method. If there's none, what else will help us figure out what's wrong? I'd probably add `console.error(attachment)`, and maybe even print  `typeof attachment` for good measure. And anything else you could think of?

frontend/src/services/FileHandlerService.ts
```diff
@@ -48,6 +48,7 @@ const computeChecksumMd5 = (file: File): Promise<string> => {
     fileReader.onload = (e) => {
       const arrayBuffer = e.target?.result
       if (!(arrayBuffer instanceof ArrayBuffer)) {
+        console.error('Empty array buffer - error reading the file')
```
timotheeg: Has this been an issue we have seen as well?

timotheeg (approved):

PR #5665, opened 2022-12-28
chore: swap cookie names for admin ui from `v2-admin-ui` to `v2-admin-ui-Jan-2023`

### LINKED ISSUE #5662: Forcefully push admins to react and create a secret passage to angular
This issue is a follow-up to https://github.com/opengovsg/FormSG/issues/5388

## TODO
1. Change cookie name of admin choice to something new (maybe add some random suffix to the name, like `_jan23`)
2. Add some code in serve default to delete the old cookie name

## Description
By using a new cookie name, the old cookie value, which represents the previous choice of the admin, will be ignored, and since the default serving behaviour is to be on react 100%, all admins will now be served React. Since the release threshold is now 100%, admins don't have a in-app link to switch back to angular any more.

## Note:
We know some admins chose to be on angular because they just don't like new things (human nature FTL), but it's possible some admins chose angular because they did indeed hit issues that we don't know about yet. These represent a little bit of unknown unknowns. 

Some of the unknown unknows could in fact be blockers that need to either be addressed with hotfix, or at least allow the admin to go back to angular after all to be unblocked.

## Fail Safe switch back
While the app does NOT show a link to switch to angular any more, we can send the switch link (https://form.gov.sg/environment/angular) to the admin manually (through an interaction in a Zendesk ticket).

By accessing that link, the new cookie will be set with a value of angular and thus represent a fresh choice to be on the old UI. 

We need to have a hard stance on giving that link to users. Only legitimate bugs with no workarounds should justify giving the admins a switch-back link to angular. In al other cases, we expect admins to get used to the react UI, like everybody else.

Note that we should track if/when we give the switch back link manually, because it will mean some admins will still be on angular. When we fix whatever issue was reported, we can either:
1. repeat the process described in this ticket to change the cookie one more time
2. remove angular entirely, removing switch back capability for good.

justynoh:
## Problem
We want to force admins who were on angular onto react now.

Closes #5662 

## Solution

Add a new cookie, `v2-admin-ui-Jan-2023` that is the new cookie that will be used. The deprecated cookie is cleared.

**Breaking Changes** 

## Deploy Notes

- `REACT_MIGRATION_ADMIN_COOKIE_NAME_OLD` : The name of the old admin cookie to be deleted. Defaults to `v2-admin-ui`.

## Tests

Assuming admin rollout is at 100%,
- [x] As an admin who has no cookies, log in. Should be served react, with `v2-admin-ui-Jan-2023` = `react`.
- [x] As an admin who has `v2-admin-ui` = `react`, log in. Should be served react, with `v2-admin-ui` cleared and `v2-admin-ui-Jan-2023` = `react`.
- [x] As an admin who has `v2-admin-ui` = `angular`, log in. Should be served react, with `v2-admin-ui` cleared and `v2-admin-ui-Jan-2023` = `react`.
- [x] As an admin who has no cookies, go to `/environment/angular`. Then, log in. Should be served angular, with `v2-admin-ui-Jan-2023` = `angular`.
- [x] As an admin who has `v2-admin-ui` = `react`, go to `/environment/angular`. Then, log in. Should be served angular, with `v2-admin-ui` cleared and `v2-admin-ui-Jan-2023` = `angular`.
- [x] As an admin who has `v2-admin-ui` = `angular`, go to `/environment/react`. Then, log in. Should be served react, with `v2-admin-ui` cleared and `v2-admin-ui-Jan-2023` = `react`.

### DIFF
--- a/src/app/modules/react-migration/react-migration.controller.ts
+++ b/src/app/modules/react-migration/react-migration.controller.ts
@@ -226,8 +226,8 @@ export const servePublicForm: ControllerHandler<
 
 export const serveDefault: ControllerHandler = (req, res, next) => {
   // Delete the deprecated cookie.
-  if (req.cookies) {
-    res.clearCookie(config.reactMigration.adminCookieName)
+  if (req.cookies[config.reactMigration.adminCookieNameOld]) {
+    res.clearCookie(config.reactMigration.adminCookieNameOld)
   }
 
   // Admins assigned react, or who choose react will stay on react until they opt out
@@ -245,13 +245,13 @@ export const serveDefault: ControllerHandler = (req, res, next) => {
     showReact = false
     // Delete existing cookie to prevent infinite redirection
     if (req.cookies) {
-      res.clearCookie(config.reactMigration.adminCookieNameJan2023)
+      res.clearCookie(config.reactMigration.adminCookieName)
     }
   } else if (req.cookies) {
-    if (config.reactMigration.adminCookieNameJan2023 in req.cookies) {
+    if (config.reactMigration.adminCookieName in req.cookies) {
       // Check if admin had already chosen react previously
       showReact =
-        req.cookies[config.reactMigration.adminCookieNameJan2023] ===
+        req.cookies[config.reactMigration.adminCookieName] ===
         UiCookieValues.React
     }
   }
@@ -271,7 +271,7 @@ export const serveDefault: ControllerHandler = (req, res, next) => {
     })
 
     res.cookie(
-      config.reactMigration.adminCookieNameJan2023,
+      config.reactMigration.adminCookieName,
       showReact ? UiCookieValues.React : UiCookieValues.Angular,
       // If assigned react, admins will stay on it unless they opt out.
       // If assigned angular, the admin cookie is for a single session
@@ -310,7 +310,7 @@ export const adminChooseEnvironment: ControllerHandler<
       ? UiCookieValues.React
       : UiCookieValues.Angular
   res.cookie(
-    config.reactMigration.adminCookieNameJan2023,
+    config.reactMigration.adminCookieName,
     ui,
     // When admin chooses to switch environments, we want them to stay on their
     // chosen environment until the alternative is stable.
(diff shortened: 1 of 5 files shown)


src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -240,13 +245,13 @@ export const serveDefault: ControllerHandler = (req, res, next) => {
     showReact = false
     // Delete existing cookie to prevent infinite redirection
     if (req.cookies) {
-      res.clearCookie(config.reactMigration.adminCookieName)
+      res.clearCookie(config.reactMigration.adminCookieNameJan2023)
```
timotheeg: Oh... I didn't think you'd change that code.

From a functional perspective, we still have an `adminCookieName` which controls overrides so that should stay the same, just it will get a new value (with the _jan23 suffix) Then we now also have a `oldCookieName` variable that we want to track to delete it.

timotheeg (commented):
Well, the PR works, but it's a bit differrent from what I expected  you would do 😅

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -225,6 +225,11 @@ export const servePublicForm: ControllerHandler<
 }
 
 export const serveDefault: ControllerHandler = (req, res, next) => {
+  // Delete the deprecated cookie.
+  if (req.cookies) {
```
timotheeg: We could check if there's the old cookie at all before sending the delete.

justynoh: [7ebfd3b](https://github.com/opengovsg/FormSG/pull/5665/commits/7ebfd3b5fead285a25667d4bdce15d019e1253a7)

src/app/modules/react-migration/react-migration.controller.ts
justynoh: [7ebfd3b](https://github.com/opengovsg/FormSG/pull/5665/commits/7ebfd3b5fead285a25667d4bdce15d019e1253a7)

timotheeg (approved):
LGTM

PR #5670, opened 2022-12-29
fix(dev): form auth for dev env

### LINKED ISSUE #5384: Improve mockpass for SP/CP OIDC
**Problem**

- @opengovsg/mockpass currently does not fully mimic the Singpass / Corppass OIDC integration. For eg.
  - mockpass still uses pem keys, and not the jwks format now used by Singpass / Corppass
  - mockpass does not expose discovery and jwks endpoints
  - mockpass does not read from the application's jwks endpoint, and hence does not encrypt the payload using the application's public key / check the assertion jwt signed by the application when the token endpoint is called

**Goal**
- Improve mockpass to better mimic Singpass / Corppass OIDC. This will improve the local dev experience (instead of testing against SP/CP staging servers).

justynoh:
## Problem
Form authentication is broken on local dev environments because mockpass does not support NDI OIDC as an authentication service. Mockpass [was recently updated](https://github.com/opengovsg/mockpass/pull/473) to support NDI OIDC, so this PR upgrades mockpass along with the necessary env vars. This fixes Singpass, Corppass and MyInfo.

Closes #5384 

## Solution

- Upgrade mockpass to the latest version.
- Update our docker-compose to add the latest mockpass endpoints.

I also took the chance to fix SGID. The reason why SGID was failing in dev env was because we always [use the default `hostname` provided by @opengovsg/sgid-client](https://github.com/opengovsg/sgid-client/blob/develop/src/index.ts), which is, of course, the actual production sgid endpoint. There was an `SGID_ENDPOINT` env var included in our docker-compose but it was never referenced in code.

- Add `SGID_HOSTNAME` as an environment variable. If defined, it uses this SGID host. Otherwise, it defaults to the host provided by the SGID client. 

**Breaking Changes** 

## Screenshots

Singpass, Corppass, SGID

https://user-images.githubusercontent.com/25571626/209909839-fe4ad057-55e3-4213-bc26-2de962dfa8d9.mov

(did I find a bug at 1:23? 🤔 )

MyInfo

https://user-images.githubusercontent.com/25571626/209909851-93c7a4c4-936c-433e-a81f-0c2b8012b306.mov

## Tests
- [ ] All form auth should work on dev env by redirection to mockpass.
- [ ] All form auth should work on staging by redirection to respective staging endpoints (Singpass, Corppass, SGID).

## Deploy Notes

- `SGID_HOSTNAME` : The SGID authentication API hostname. Optional. Omit to use the default SGID API hostname.

Prior to deployment, we should clean up after ourselves and delete the `SGID_ENDPOINT` environment variable from parameter store. This has been tested on staging.

### DIFF
--- a/src/app/modules/sgid/sgid.service.ts
+++ b/src/app/modules/sgid/sgid.service.ts
@@ -40,10 +40,13 @@ export class SgidServiceClass {
     cookieMaxAgePreserved,
     privateKeyPath,
     publicKeyPath,
+    hostname,
     ...sgidOptions
   }: ISgidVarsSchema) {
     this.privateKey = fs.readFileSync(privateKeyPath, { encoding: 'utf8' })
     this.client = new SgidClient({
+      // If hostname is empty, use the default provided by sgid-client.
+      hostname: hostname || undefined,
       ...sgidOptions,
       privateKey: this.privateKey,
     })
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -84,7 +84,7 @@ services:
       - MYINFO_CLIENT_ID=mockClientId
       - MYINFO_CLIENT_SECRET=mockClientSecret
       - MYINFO_JWT_SECRET=mockJwtSecret
-      - SGID_ENDPOINT=http://localhost:5156/sgid
+      - SGID_HOSTNAME=http://localhost:5156/sgid
       - SGID_CLIENT_ID=sgidclientid
       - SGID_CLIENT_SECRET=sgidclientsecret
       - SGID_REDIRECT_URI=http://localhost:5001/sgid/login
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,16 +61,16 @@ services:
       - TWILIO_API_KEY=mockTwilioApiKey
       - TWILIO_API_SECRET=mockTwilioApiSecret
       - TWILIO_MESSAGING_SERVICE_SID=mockTwilioMsgSrvcSid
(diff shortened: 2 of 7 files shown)


src/app/modules/sgid/sgid.service.ts
```diff
@@ -40,10 +40,13 @@ export class SgidServiceClass {
     cookieMaxAgePreserved,
     privateKeyPath,
     publicKeyPath,
+    hostname,
     ...sgidOptions
   }: ISgidVarsSchema) {
     this.privateKey = fs.readFileSync(privateKeyPath, { encoding: 'utf8' })
     this.client = new SgidClient({
+      // If hostname is empty, use the default provided by sgid-client.
+      hostname: hostname || undefined,
```
timotheeg, line 49: /justMusingNoNeedToDoAntyhing: `|| undefined` feels a bit "funny". It's there to account for when`sgid.config` itself defaulted to empty string, correct? Could we not just removed the default as empty string in config, or default it to `undefined` there directly?

timotheeg (approved):
This is awesome! Thanks Justyn!

src/app/modules/sgid/sgid.service.ts
justynoh: i wanted to default it to undefined in the config, but convict wouldn't let me do it :( actually let me play around with this to see if it'll let me do it. I completely agree that this is weird haha

justynoh: image

seems like convict strictly doesn't allow `default` to be undefined, so we have to specify something. Most people seem to be using `nullable` property, but that just sets the default value to `null`, not `undefined`, so anyway we would have to do this check haha

PR #5672, opened 2022-12-29
fix: add submission time to end page, returned from server

### LINKED ISSUE #4880: [2] Display submission time from server
### Context

FormSG displays the submission time in the thank you page. The time is displayed from checking the client clock. Client clock are notoriously unreliable, as they can be off by few seconds or more. 

We have received a question from a form admin who had their form start time at 9am, but saw a form submission at 08:59am, and was wondering whether an error happened.

There was no error, but the client clock was behind by a bit (for whatever reason, accidental or malicious)

### TODO / Suggestion

(from @karrui 💪) We should return the submission time (in UTC) from our server and use that to display on the client success page instead of using client time.

The submission endpoint should return the submission timestamp in ISO format (**with timezone**!), like this:
```javascript
{
    "datetime": "2022-09-15T01:00:00.000+00:00"
}
```
Note that the UTC timezone can also be specified like this:
```javascript
{
    "datetime": "2022-09-15T01:00:00.000Z"
}
```

On the client, a `Date` object can be instantiated from that ISO string representation:
```javascript
const submissionDatetime = new Date("2022-09-15T01:00:00.000+00:00")
```
And the final date+time can be displayed by extracting the required various "bits" of the date and time in local timezone:
```javascript
// if client timezone is Singapore as GMT+8
submissionDatetime.toString() // "Thu Sep 15 2022 09:00:00 GMT+0800 (Singapore Standard Time)"

submissionDatetime.getDate() // 15
submissionDatetime.getHours() // 9
submissionDatetime.getMinutes() // 0
```
It's probably a good idea to use a date-to-string functions to format the final value, rather than doing it by hand. The existing code for the thank you page likely already has that code btw, and hopefully the only change needed will be to supply the Date object coming from the server to it, rather than instantiating a Date object on the client.

### LINKED ISSUE #5444: Edit Thank You page to add timestamp and change image
1. We previously removed the timestamps on the Thank You page because it was too cluttered. But users have written in asking for it, main reason is for proof of submission to admins. Therefore we're bringing it back and reducing the font size to caption-2 to minimise clutter.

2. Editing the image to a more neutral one

[figma link](https://www.figma.com/file/9oE4ECjdeZQ1RBpwZVtnB3/Form-Design-Master?node-id=6942%3A102593)

justynoh:
## Problem

We want to add the submission time to the form end page, and also return the submission time from the server rather than using the browser time.

Closes #5444 
Closes #4880

## Solution

Both submission endpoints now return a new value, `timestamp`. This is a date represented as ms from epoch. The public form handlers accept the date and format them as required for display on the frontend.

Additionally, the thank you page svg has been updated.

**Breaking Changes** 

## Screenshots

image

## Tests

- [ ] Make a form submission in react with a different browser time than the actual time (change your local computer's time - you can check your browser time by doing `new Date()` in the console). The displayed timestamp should be of the actual time.
- [ ] Make a form submission in angular with a different browser time than the actual time. The displayed timestamp should be of the actual time.


frontend/src/features/public-form/components/FormEndPage/components/EndPageBlock.tsx
```diff
@@ -40,6 +41,11 @@ export const EndPageBlock = ({
     },
   })
 
+  const submissionTimestamp = useMemo(
+    () => format(new Date(submissionData.timestamp), 'dd MMM yyyy, HH:mm:ss z'),
```
timotheeg, line 45: Since the type of `submissionData.timestamp` is already a `Date`, the `Date()` constructor wrapper here is unnecessary, correct?

Btw, I'm thinking the oversight is probably due to the change of types between Dto and app data type. In the PR so far, `SubmissionResponseDto.timestamp` is of type number, but `submissionData.timestamp` is of type `Date`.

To me, the term `timestamp` connotates a number (ms from epoch), so I'm thinking  maybe we could avoid some confusion with some slight refactoring. Like we could either:
* maintain `timestamp` as a number everywhere, and only convert to `Date` in the components that need it, e.g. for date computation or rendering (like is done here)
* change the field name in `submissionData` (maybe use `date`, or `datetime`) to indicate the field is a `Date`

I know TypeScript helps us with type, but having the same field name be of different types in different context is confusing for humans, not compilers 😅

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -374,6 +374,7 @@ const submitEmailModeForm: ControllerHandler<
               // Return the reply early to the submitter
               message: 'Form submission successful.',
               submissionId: submission.id,
+              timestamp: (submission.created || new Date()).getTime(),
```
timotheeg, line 377: I wonder if `|| new Date()` is overly defensive? Is there really a case where we have a `submission` without a `created` field on it?

(Same question/comment for the encrypt controller below)

timotheeg (commented):
Thanks for taking this @justynoh

src/app/modules/submission/email-submission/email-submission.controller.ts
justynoh: haha `submission.created` is a `number | undefined` so typescript won't let me set `timestamp` without a default 😅

frontend/src/features/public-form/components/FormEndPage/components/EndPageBlock.tsx
justynoh: whoops. you are completely correct - i played around with the types initially, thinking what would be the best type for the dto and then where to perform the conversion. I agree we should keep to ms from epoch as far as possible! this is fixed in [7b0417a](https://github.com/opengovsg/FormSG/pull/5672/commits/7b0417a219cd4c5680252eb74b20e28705deca47)

frontend/src/features/public-form/components/FormEndPage/FormEndPage.stories.tsx
```diff
@@ -29,7 +29,7 @@ Default.args = {
   },
   submissionData: {
     id: 'mockSubmissionId',
-    timeInEpochMs: 1648545566989,
+    timestamp: new Date(1648545566989),
```
timotheeg: Change this back to number as well?

src/app/modules/submission/email-submission/email-submission.controller.ts
timotheeg: Ah, thanks for the clarification.... Oh well 🙄 🤷

frontend/src/features/public-form/components/FormEndPage/FormEndPage.stories.tsx
justynoh: oops! thanks, [dcf2dab](https://github.com/opengovsg/FormSG/pull/5672/commits/dcf2dab6e79770e81e3c54afa9194b445862d347)

timotheeg (approved):
LGTM 👍

karrui:
Haven't looked closely, but if the server returns the time of submission in Date, remember to use the client side middle ware to reserialize the stringified date back to a Date.

justynoh:
the dto is a number in ms from epoch, not Date hehe

PR #5700, opened 2023-01-10
build: hotfix v6.28.2 for datadog error logs

wanlingt:
## Problem

Remove formInputs from error logs, use attachment extension instead of attachment name


timotheeg (approved):

frontend/src/features/public-form/utils/createSubmission.ts
```diff
@@ -188,7 +188,7 @@ const encryptAttachment = async (
         error: error,
         attachmentId: id,
         attachmentType: typeof attachment,
-        attachmentName: attachment.name,
+        attachmentExtension: attachment.name.split('.').pop(),
```
timotheeg: I am just a little bit worried for run-time errors here, for the unlikely cases where `attachment.name` is not a string 🤔 , maybe we can add some defensive check:

```javascript
attachmentExtension: attachment.name?.split('.').pop(),
```

wanlingt: [1f559c5](https://github.com/opengovsg/FormSG/pull/5700/commits/1f559c53d2984665020eafe2c1f59d4dfd2db8c8)

timotheeg (approved):
👍 again

PR #5705, opened 2023-01-11
feat: improved FE logging for unexpected errors

timotheeg:
## Context

We are receiving frontend errors that we haven't managed to reproduce. These errors are not tracked automatically by RUM because they are caught and displayed in a user toast. 

That is enough information to lead us in the right direction but not enough to toubleshoot definitely.

We have previously added DD browser logs but they are still not sufficient.

## Approach

Several incremental improvements are done to make the log more useful to us:
* Add more context fields into error logs
* Split code known to give error into recognizable steps in logs

Additional bonus refactoring
* in APIService's `transformAxiosError()` method, get rid of single letter variable
* minor code reduction in `createSubmission.ts`


timotheeg:
@wanlingt , please take a look when you have the chance (it's a short PR)

wanlingt (approved):
lgtm!

Issue #5510, opened 2022-12-01
Timezones not considered on datepicker


nicholasluimy:
**Describe the bug**
It is unclear when dealing with datepickers if we are dealing with local or Singapore timezones

Selecting a date would result in a different date displayed depending on local context -> E.g selecting a date in +8 GMT (E.g 10th Jan) could result in a different date displayed if the user is locally in a different timezone  (E.g 9th Jan, if in a +6 GMT timezone) and we're barely into the day in +8 GMT (E.g 0100hrs)

**To Reproduce**
1. Set location/time on device to different timezone that is currently on different day than Singapore time
2. Selecting a date results in a different date input into the form based off local timezone
3. Switch location/time on device to Singapore time
4. Inputs now work as expected

**Expected behavior**
A clear highlight of the timezone we are dealing with, or specify that all date/time inputs are in Singapore's frame of reference

**Desktop (please complete the following information):**
- Chrome browser Version 107.0.5304.121 (Official Build) (x86_64)

nicholasluimy:
For a sample of how this could be implemented

![image](https://user-images.githubusercontent.com/8996422/205141528-8b778758-8d15-4e9d-9e5b-bd45f33537da.png)


This issue was additionally confirmed with receipt of form submission - the erroneous dates were not cleaned/validated on the backend, and the dates in the non singapore timezone was used instead

LinHuiqing:
For now, our date pickers are used to indicate specific dates alone with no concept of time, both from the admin and respondent perspective. As such, no awareness of timezone should be built into date pickers. If admins are specific date information, it is up to the admins to specify what timezone they are referring to in their questions and descriptions.

PR #5731, opened 2023-02-05
fix: date picker calendar select correct date in negative UTC offset timezones

darrelhong:
This PR fixes an issue brought up in #5510. It seems to be caused modifications to date in `handleDateChange`. 
demo:
https://user-images.githubusercontent.com/39296145/216810183-b8ca0036-416b-45c1-be25-530eb6c90729.mov

1. Date picker selection returns date at 00:00:00 in local timezone
2. Default 'UTC' string is passed to `zonedTimeToUtc`
3. `zonedTimeToUtc` returns equivalent datetime in UTC after comparing local timezone to 'UTC' time
4. Any local timezone that is negative offset from UTC will result in the date being change to the day before.

For better visualisation:
*BEFORE*
Screen Shot 2023-02-05 at 2 08 35 AM

*AFTER*
Screen Shot 2023-02-05 at 2 07 19 AM

## Solution
Instead of specifying 'UTC' to `zoneTimeToUtc`, provide the local timezone instead. This fixes the issue but I'm not actually sure whether its correct as it feels like its just reversing the use of `zoneTimeToUtc`. More info at [marnusw/date-fns-tz.js#174](https://github.com/marnusw/date-fns-tz/issues/174)

I believe this can be discussed further @karrui

**Breaking Changes** 

### DIFF
--- a/frontend/src/components/DatePicker/DatePickerContext.tsx
+++ b/frontend/src/components/DatePicker/DatePickerContext.tsx
@@ -84,7 +84,7 @@ const useProvideDatePicker = ({
   isReadOnly: isReadOnlyProp,
   isRequired: isRequiredProp,
   isInvalid: isInvalidProp,
-  timeZone = 'UTC',
+  timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone,
   locale,
   isDateUnavailable,
   allowManualInput = true,


darrelhong:
@karrui @justynoh @mantariksh any update on this, maybe could use this a temp fix before a more comprehensive fix

LinHuiqing:
Hi @darrelhong, thanks for your contributions! Sorry we took so long to get around to this. We've decided to solve the datepicker issue by removing the concept of timezone altogether in this fix (https://github.com/opengovsg/FormSG/pull/6025). As you mentioned, your fix will be essentially undo-ing the effects of `zoneTimeToUtc`, which is removing the concept of timezones, so in effect, what your PR does is the same as what #6025 does.
As such, we'll be closing this PR. Do let us know if you have any comments on the chosen fix though. Thank you!

PR #5743, opened 2023-02-10
fix: save secret key per form with useQuery to prevent need for entering secret key

### LINKED ISSUE #5514: Users need to key in secret key each time they exit the form
On Angular, users would be able to key in multiple form's secret keys and it will be accessible for a period of time.

However, on React, users would be prompted to enter the secret key each time they enter a different form.

Ticket (non-exhaustive, just started tracking but have seen old requests for this):
https://formsg.zendesk.com/agent/tickets/17401

darrelhong:
## Problem
Users need to key in secret key each time they exit the form. PR also serves to achieve feature parity with angular

Closes #5514

## Solution
use useQuery to store secretKey per form. Maintained signature of secretKey and setSecretKey in `useSecretKeys.ts` for minimal code change. queryFn resolves to string for type inference. Set stale time to infinity so that queryFn is never called again, resetting secretKey. 

cc @amitogp @kennethchangOPENGOV

**Breaking Changes** 

    - Details ...

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/useSecretKey.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/useSecretKey.ts
@@ -1,7 +1,10 @@
-import { useEffect, useRef, useState } from 'react'
+import { useCallback, useEffect, useRef } from 'react'
+import { useQuery, useQueryClient } from 'react-query'
 import { BroadcastChannel } from 'broadcast-channel'
 
 import { useHasChanged } from '~hooks/useHasChanged'
+
+import { adminFormResponsesKeys } from '../../queries'
 
 const SECRETKEY_BROADCAST_KEY = 'formsg_private_key_sharing'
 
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/responses/ResponsesPage/storage/useSecretKey.ts
```diff
@@ -17,7 +18,16 @@ type SecretKeyBroadcastMessage =
     }
 
 export const useSecretKey = (formId: string) => {
-  const [secretKey, setSecretKey] = useState<string>()
+  const { data: secretKey } = useQuery({
+    queryKey: ['secretKey', formId],
```
justynoh: should we define the queryKey in `frontend/src/features/admin-form/responses/queries.ts` under `adminFormResponsesKeys` instead (since the secret key is entered under that module)?

justynoh (commented):
Thanks for the contribution @darrelhong ! Just one small comment. Additionally, there's a linter warning - would you be able to fix it? thanks!

frontend/src/features/admin-form/responses/ResponsesPage/storage/useSecretKey.ts
darrelhong: @justynoh moved the query key to the the file, but do you prefer if whole query was extracted instead?

justynoh: I think keeping the query in this file is fine, we don't need to extract it. 

Actually, I was wondering if it's possible to avoid having an awkward `queryFn: () => Promise.resolve('')` at all. Instead of using `useQuery` with a dummy `queryFn`, can we just use `useQueryClient` with `QueryClient.getQueryData` and `QueryClient.setQueryData`?

darrelhong: I havent tried it but I think in that case secretKey will not be reactive, and not respond to changes when setQueryData is called

darrelhong: @justynoh updated with ìnitialData` option which i think you would prefer as its more declarative and no dummy query and supports type inference

justynoh (approved):
Great, lgtm - thanks for the contribution @darrelhong!

justynoh:
Closing and reopening to try and get tests to run

PR #5750, opened 2023-02-13
chore: remove numbering for API service errors

justynoh:
## Problem

Since we've already identified the genre of the errors that has been most common, we can remove the numbering which can be confusing for users who encounter the error.

## Solution

**Breaking Changes** 


tshuli (approved):
lgtm

PR #5752, opened 2023-02-13
chore: add req meta to static asset endpoint log

tshuli:
## Problem
Adds req meta to logs for static asset retrieval


justynoh (approved):
lgtm

PR #5763, opened 2023-02-14
fix: inefficient form metadata query leading to excessive memory usage

### LINKED ISSUE #3394: [8] Database read time spikes due to response downloads
When admin of large form downloads responses (thousands or tens of thousands), the current implementation issues one very large and long running query to mongoDB.

This consumes a long running operation on node and on mongo. While both should be broken down and driven in multiple requests from the client itself, at minimum we should split the mongo query into smaller paginated queries to ensure no query may lock the DB.

tshuli:
## Problem

- We have been seeing very slow `formsg.submissions` operations which have very high db disk util
- Initial suspicion was that this was due to download of large forms

- Closes #3394

## Solution
- For load test example, see forms `63e9b7e27e26fb0012795ca8` (200k small responses) and `63e9d46e6d6d6f0012d14402` (20k large responses of >700KB each) on staging (log in using form team email, key in 1pw).
- It turns out that response download is already efficient. The current implementation returns a cursor which iterates over the matching documents. It appears that mongoDB has implemented this in a non-blocking fashion (from testing, multiple concurrent downloads of large forms can proceed concurrently without spike in disk util).
- Instead, the reason for the high disk utilisation was inefficient query implementation in the `/metadata` endpoint, which returns response metadata for storage mode submissions. This is triggered once the admin navigates to response tab.
  - In `submission.server.model.ts`, the following code retrieves all submission documents corresponding to the form, stores it all in memory (or disk, because `allowDiskUse(true)`), in order to count the number of responses. This led to excessive disk utilisation and slow query (see `NOTE` below).
  - To solve this, this has been replaced with a simpler `countDocuments` query.

**Existing code**
```
.facet({
  pageResults: [
    { $skip: numToSkip },
    { $limit: pageSize },
    { $project: { _id: 1, created: 1 } },
  ],
  allResults: [
    { $group: { _id: null, count: { $sum: 1 } } },
    { $project: { _id: 0 } }, // NOTE: Means project ALL fields except _id
  ],
})
// prevents out-of-memory for large search results (max 100MB).
.allowDiskUse(true) 
// NOTE: As the above query required excessive memory, 
// it was written to disk instead and this led to high disk utilisation 
.then((result: MetadataAggregateResult[]) => {
  const [{ pageResults, allResults }] = result
  const [numResults] = allResults
  const count = numResults?.count ?? 0
```

**Improvements**:

## Before & After Screenshots

**BEFORE**:
- Timeout (504) for metadata endpoint for forms with many submissions
timeout

- 100% disk utilisation on DB
diskutil

**AFTER**:

- metadata endpoint resolves in 200+ms
fasteresolve

- No impact to db disk util
lowdiskutil

## Tests
- [ ] Create storage mode form. Submit 20 responses. Check that metadata and pagination shows correctly on results tab.

### DIFF
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -1,6 +1,5 @@
 import moment from 'moment-timezone'
 import mongoose, { Mongoose, QueryCursor, Schema } from 'mongoose'
-import type { FixedLengthArray } from 'type-fest'
 
 import {
   FormAuthType,
@@ -274,10 +273,7 @@ EncryptSubmissionSchema.statics.findSingleMetadata = function (
  * Unexported as the type is only used in {@see findAllMetadataByFormId} for
  * now.
  */
-type MetadataAggregateResult = {
-  pageResults: Pick<ISubmissionSchema, '_id' | 'created'>[]
-  allResults: FixedLengthArray<{ count: number }, 1> | []
-}
+type MetadataAggregateResult = Pick<ISubmissionSchema, '_id' | 'created'>
 
 EncryptSubmissionSchema.statics.findAllMetadataByFormId = function (
   formId: string,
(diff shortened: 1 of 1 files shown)


timotheeg (approved):
👍 Great work!

PR #5766, opened 2023-02-14
chore: remove instrumentation for email performance monitoring and SES migration

justynoh:
## Problem

We created additional instrumentation over our mailing system in #5019 and #5369 and #5442. This PR tears them down, since they are not needed anymore.

Closes https://github.com/opengovsg/formsg-private/issues/130 and closes https://github.com/opengovsg/formsg-private/issues/132.

Cleaning up old SES config will also close https://github.com/opengovsg/formsg-private/issues/91.

## Solution

Delete code that references datadog traces and SES in the US.

**Breaking Changes** 

## Tests

- [x] OTP emails can be sent to admins to log in.
- [x] Email responses can be sent to admins after submitting an email mode form.
- [x] OTP emails can be sent to respondents when verifying an email in a public form.
- [x] Confirmation emails can be sent to respondents when submitting and email mode form.

## Deploy Notes

for `x` = `HOST`, `PORT`, `USER`, `PASS`:

- `SES_{x}` : Should be the value of the old variable `SES_{x}_SG`
- `SES_{x}_US` and `SES_{x}_SG`: Should be deleted

### DIFF
--- a/src/app/services/mail/mail.service.ts
+++ b/src/app/services/mail/mail.service.ts
@@ -1,4 +1,3 @@
-import { tracer } from 'dd-trace'
 import { get, inRange, isEmpty } from 'lodash'
 import moment from 'moment-timezone'
 import { err, errAsync, okAsync, Result, ResultAsync } from 'neverthrow'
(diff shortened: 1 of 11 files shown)


justynoh:
(This PR must be released before we tear down the infra for SES US)

timotheeg (commented):
Looking good!

Unless the tracing of the email sending is costing a lot (I assume no right now), let's keep them in for now (traces for email sending + pdf generation), it gives us visibility in processes we are otherwise blind to.

src/app/services/mail/mail.service.ts
```diff
@@ -1,4 +1,4 @@
-import { tracer } from 'dd-trace'
+import tracer from 'dd-trace'
```
timotheeg, line 1: is this change intended?

justynoh: oops no, but it shouldn't make a difference! it was probably introduced when I reinstated the tracer spans. But it still works, tested on staging + the default import was already used in `mail.utils.ts` before

timotheeg (approved):
LGTM

PR #5767, opened 2023-02-14
fix(a11y): move aria required and invalid from individual rating options to rating group

foochifa:
## Problem

Currently every individual rating option are tagged as aria-required and aria-invalid, which results in Mac Voiceover reading it as 'required selected invalid data'. 

Additionally, the labels of the rating option are not informative and aria labels should be implemented to describe how the rating options are visually represented.

Closes [#5503]

## Solution

Added aria labels to describe the rating options
Remove aria-required and aria-invalid labels by setting them as false (similar to our approach for radio buttons)
Updated Rating field container and Rating group to adopt aria-required and aria-invalid labels

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

## Before & After Screenshots

**BEFORE**:

Recording 2023-02-14 at 15 53 57

**AFTER**:

Recording 2023-02-14 at 15 57 52

## Tests

On Mac, ensure voiceover is turned on:
- [ ] Go to a form with ratings set as required. Go through the ratings field and ensure that the required label is in the rating group (not each option).
- [ ] Try and submit form without indicating a rating (submission will fail). Voiceover will redirect to top of form. Go to ratings field and make sure 'invalid' label is together with the required label in the rating group (not each option). Select an option, submit the form.
- [ ] Go to a form with ratings, but not set as required. Go through the ratings field. Ensure that no required label is tagged on to the rating field (group or individual options). Submit the form.

On iOS, ensure voiceover is turned on:
- [ ] Go to a form with ratings set as required. Go through the ratings field and ensure that the required label is in the rating group. Ensure that each individual option do not read 'required' and do not read 'unticked' when it is selected.

PR #5768, opened 2023-02-14
fix(a11y/windows): read out question text when using Tab to navigate

### LINKED ISSUE #5588: a11y: question text is not read out when using Tab to navigate (Windows)
**Issue**
When the user uses the Tab key to navigate from one form element to another, the question text does not get read out for some questions, and only the question options gets read out. This issue does not appear when using the arrow keys to navigate instead.

This issue affects questions 8, 10, and 12.

i.e. The affected form questions are the following:

- 1 to 5 star rating
- Yes/No buttons
- File upload

**Suggestion by Etch**
The issue seems to be caused by the ARIA labels in the options for the affected questions taking precedence over the ARIA label in the question text.

For form questions of this nature, one potential fix for this issue is to place the radiogroup inside a <fieldset> tag.

Then, place the question text inside a <legend> tag above the radiogroup.

This would fix the ARIA labels to read the question text and question options properly using the Tab key for navigation.

Additionally, for the Yes/No button question type (i.e. Question 10), the role="button" in the <label> tag needs to be removed for this implementation to work.

This has the added benefit of making the input type detected as radio buttons to screen readers, as it currently only is detected as a selectable button, which means keyboard shortcuts to jump to radio buttons using screen readers do not work for Question 10.

**Screenshot**
https://drive.google.com/file/d/1iqMyET-YQg9jNFZzUkXYqVSGJ4IjME0n/view?usp=share_link

LinHuiqing:
## Problem

On Windows devices, when the Tab key is used for navigation, the question texts for the following question types are not read out:

- Yes/no
- Attachment
- Rating (addressed in PR #5767)

Refer to issue #5588 for more details.

Closes #5588

## Solution

As Tab key reads out information for options directly, pass the question text into the options to be read out as aria-labels.

**Breaking Changes** 

**Bug Fixes**:

- Questions will now be read out for Yes/no and Attachment fields

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/37061143/218991896-f1b8c510-7191-4de3-90f1-ad39a631ab5c.mp4

**AFTER**:

https://user-images.githubusercontent.com/37061143/218986659-6f9dbc3e-8bcf-4ee9-aa3e-c229e2d3501b.mp4

## Tests

- [ ] Using Windows' NVDA software, use Tab navigation to get around in a form with Yes/no and Attachment fields. 
- [ ] You should be able to hear the question text for Yes/no fields now.
- [ ] You should be able to hear the question text for Attachment fields now.

### DIFF
--- a/frontend/src/components/Field/Attachment/AttachmentDropzone.tsx
+++ b/frontend/src/components/Field/Attachment/AttachmentDropzone.tsx
@@ -8,19 +8,21 @@ interface AttachmentDropzoneProps {
   inputProps: DropzoneInputProps
   isDragActive: DropzoneState['isDragActive']
   readableMaxSize?: string
+  question?: string
 }
 
 export const AttachmentDropzone = ({
   inputProps,
   isDragActive,
   readableMaxSize,
+  question,
 }: AttachmentDropzoneProps): JSX.Element => {
   const styles = useStyles()
 
   return (
     <>
       <VisuallyHidden>
-        Click to upload file, maximum file size of {readableMaxSize}
+        {question} Click to upload file, maximum file size of {readableMaxSize}
       </VisuallyHidden>
       <chakra.input {...inputProps} data-testid={inputProps.name} />
       <Icon aria-hidden as={BxsCloudUpload} __css={styles.icon} />
--- a/frontend/src/components/Field/Attachment/Attachment.tsx
+++ b/frontend/src/components/Field/Attachment/Attachment.tsx
@@ -80,6 +80,7 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
       value,
       name,
       colorScheme,
+      title,
       ...props
     },
     ref,
@@ -250,6 +251,7 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
                 isDragActive={isDragActive}
                 inputProps={processedInputProps}
                 readableMaxSize={readableMaxSize}
+                question={title}
               />
             )}
           </Box>
--- a/frontend/src/components/Field/YesNo/YesNo.tsx
+++ b/frontend/src/components/Field/YesNo/YesNo.tsx
@@ -89,6 +89,7 @@ export const YesNo = forwardRef<YesNoProps, 'input'>(
           // Ref is set here for tracking current value, and also so any errors
           // can focus this input.
           ref={ref}
+          title={props.title}
         />
         <YesNoOption
           side="right"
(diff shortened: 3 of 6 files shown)


frontend/src/components/Field/Attachment/AttachmentDropzone.tsx
```diff
@@ -8,19 +8,21 @@ interface AttachmentDropzoneProps {
...
   readableMaxSize,
+  question,
 }: AttachmentDropzoneProps): JSX.Element => {
   const styles = useStyles()
 
   return (
     <>
       <VisuallyHidden>
-        Click to upload file, maximum file size of {readableMaxSize}
+        {question} Click to upload file, maximum file size of {readableMaxSize}
```
foochifa, line 25: Is there a reason why we put the question here over including it as an aria-label in the chakra.input below?

LinHuiqing: Oh, because:
1. The question should come before what the user should do.
2. I kind of assumed this was a better place to put it as VisuallyHidden is intended for screen readers and previous engineers chose to. 

Anyways, I tried it out with the following snippet:
```ts
<chakra.input
  aria-label={`${question} Click to upload file, maximum file size of ${readableMaxSize}`}
  {...inputProps}
  data-testid={inputProps.name}
/>
```

On windows, this will just read "button required" as seen in the video below:

https://user-images.githubusercontent.com/37061143/219065430-3e9b504d-0077-4c6c-bea4-aed94ed96e97.mp4

So this probably means NVDA does not pick up the `aria-label` of `chakra.input`. Thus, I'm guessing `VisuallyHidden` was used (by @justynoh I think) for screen readers to pick up and give more informative descriptions to our visually impaired users.

Relatedly, I just read [this article](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/) that points out the flaws of `aria-label`s and explains when to use `VisuallyHidden` elements.

foochifa: Oh nice! Thanks I realised that aria-labels also don't work as expected all the time hahaha, maybe it has to do with assigning a role to the input or perhaps another option would be to envelop the input+icon field with a label container like in radio & Yes/No.

But seems like visually hidden works well here!

foochifa:
lgtm!

foochifa (approved):

foochifa:
Yikes wait sorry, could you edit the PR and change the test to be checkboxes, also separate the test for both Yes/No and attachment fields thanks!

LinHuiqing:
Okays done!

PR #5786, opened 2023-02-16
fix(a11y): change heading label from instruction and section field container to h2 element

### LINKED ISSUE #5501: a11y: Paragraph shouldn't be labelled h2
**Checked:**
NVDA: Paragraph is reading as h2
Mac Voiceover: the entire heading and paragraph reads together (heading level 2, 2 items)
iPhone Voiceover: Paragraph reads as h2
Talkback: reads out heading and paragraph together. but does not announce h2 

**Proposal:**
Paragraph is only reading out as h2 on NVDA and iPhone Voiceover, could we try and check why?

foochifa:
## Problem

Currently, we include the heading role in the parent element of both the instruction/heading of the section and the descriptions.

This causes the description/paragraph to also be labelled as "heading level 2" and will be read out as such on iOS Voiceover and window's NVDA screen reader. 

Closes #5501

## Solution

Removed `role=heading` in the parent element of section and instruction fields. Using `role=heading` is actually [dissuaded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role), and we should use `<h2>` element instead. 

Change the element of instruction and heading from `<Text>` to `<Box as="h2">` using the advice from this [issue](https://github.com/chakra-ui/chakra-ui/issues/3658) to retain our text styles.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

Heading level 2 after both the headers and the paragraphs

**AFTER**:

Heading level 2 after the header and before the paragraph 

## Tests

On Windows:
- [x] Open a form with instructions. Use NVDA screen reader, navigate to the instruction section using the navbar. The screen reader should read 'Instructions Heading Level 2 (your instruction here)'.
- [x] Open a form with a heading field, including description of the heading. Use NVDA screen reader, navigate to the section using the navbar. The screen reader should read '(your heading here) Heading Level 2 (your description here)'.

On iOS:
- [x] Open a form with instructions. Use iOS voiceover, navigate to the instruction section. The screen reader should read 'Instructions Heading Level 2'. And upon next swipe it should read your instruction without any 'Heading Level 2' attached behind it.
- [x] Open a form with a heading field, including description of the heading. Use iOS voiceover, navigate to your section. The screen reader should read '(your heading here) Heading Level 2'. And upon next swipe it should read your description without any 'Heading Level 2' attached behind it.

### DIFF
--- a/frontend/src/features/public-form/components/FormInstructions/FormInstructions.tsx
+++ b/frontend/src/features/public-form/components/FormInstructions/FormInstructions.tsx
@@ -1,4 +1,4 @@
-import { Box, Text } from '@chakra-ui/react'
+import { Box } from '@chakra-ui/react'
 
 import { FormColorTheme } from '~shared/types'
 
@@ -27,9 +27,9 @@ export const FormInstructions = ({
 
   return (
     <>
-      <Text textStyle="h2" color={sectionColor}>
+      <Box as="h2" textStyle="h2" color={sectionColor}>
         Instructions
-      </Text>
+      </Box>
       <Box mt="1rem">
         <MarkdownText multilineBreaks components={mdComponents}>
           {content}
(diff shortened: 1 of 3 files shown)


LinHuiqing:
Checked for Windows and it seems good to go! Can't check for iOS tho... Maybe @justynoh and @wanlingt can help try out for the behaviour on iOS?

justynoh (approved):
lgtm

justynoh:
sorry i don't have iOS :)

foochifa:
Its still on staging-alt if anyone wants to try it on iOS @staceytan1998 @wanlingt

PR #5796, opened 2023-02-20
fix: strip whitespace for logic conditions on frontend

tshuli:
## Problem

- Follow up to #5433, which trimmed whitespace for logic checks on the backend. 
- In some cases, existing logic contains trailing whitespace. This PR ensures that existing logic with whitespace is also trimmed on the frontend so that form logic works properly.

## Tests
- [ ] Create form with dropdown option as logic condition for prevent submission. In the DB, add trailing whitespace to the logic condition for that dropdown option. Check that submission is prevented when the option is chosen.
- [ ] Repeat for show field logic.


wanlingt (approved):
lgtm!

PR #5799, opened 2023-02-21
fix: meta tags for public form pages

### LINKED ISSUE #5589: meta tags are not showing up correctly for public form pages
**Describe the bug**
The title, description and image for the meta tags are defaulting to the form.gov.sg ones, although they should be different

**Expected behavior**
Related to #5114 

Non-public form pages
image

Public form pages (Open forms)
image

Also, try to set up the og_image to be a fully qualified url too (dynamically - probably need to read the host name from the request headers)

wanlingt:
## Problem

Social media previews were not being generated correctly for public forms.

`formResult.value.status` was `undefined` as `status` was not retrieved in the form result. Hence the check for public forms was always returning `false`, and the default meta tags would be used for public forms, instead of the public form-specific meta tags.

Closes #5589

## Solution

To fix this, I specified `status` when retrieving the form using `FormService.retrieveFormKeysById`.

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:

Public form
image

Private form
image

## Tests

[Preview and generate](https://www.opengraph.xyz/) Open Graph meta tags, check that `title`, `description` and `image` are populated for
- [x] public form
- [x] private form

### DIFF
--- a/src/app/modules/react-migration/react-migration.controller.ts
+++ b/src/app/modules/react-migration/react-migration.controller.ts
@@ -217,10 +217,10 @@ export const servePublicForm: ControllerHandler<
   })
 
   if (showReact) {
-    const isPublicForm = Boolean(
-      !formResult.isErr() && formResult.value.status === FormStatus.Public,
-    )
-    return serveFormReact(/* isPublic= */ isPublicForm)(req, res, next)
+    const isPublicForm =
+      !formResult.isErr() && formResult.value.status === FormStatus.Public
+
+    return serveFormReact(isPublicForm)(req, res, next)
   } else {
     return serveFormAngular(req, res, next)
   }
--- a/src/app/modules/react-migration/react-migration.controller.ts
+++ b/src/app/modules/react-migration/react-migration.controller.ts
@@ -220,23 +220,7 @@ export const servePublicForm: ControllerHandler<
     const isPublicForm = Boolean(
       !formResult.isErr() && formResult.value.status === FormStatus.Public,
     )
-    const formStatus = !formResult.isErr()
-      ? `${formResult.value.status}`
-      : 'formresulterr'
-    logger.info({
-      message: 'metatags test',
-      meta: {
-        action: 'metatags test',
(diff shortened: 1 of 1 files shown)


src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -216,9 +217,10 @@ export const servePublicForm: ControllerHandler<
   })
 
   if (showReact) {
-    return serveFormReact(
+    const isPublicForm = Boolean(
       !formResult.isErr() && formResult.value.status === FormStatus.Public,
-    )(req, res, next)
+    )
+    return serveFormReact(/* isPublic= */ isPublicForm)(req, res, next)
```
tshuli: dangling comment?

tshuli (approved):
minor nit but lgtm!

src/app/modules/react-migration/react-migration.controller.ts
```diff
@@ -216,9 +217,10 @@ export const servePublicForm: ControllerHandler<
   })
 
   if (showReact) {
-    return serveFormReact(
+    const isPublicForm = Boolean(
```
timotheeg: /nit the `Boolean()` wrapper should not be necessary. Both operands of the logical AND operator produce booleans, so the type should already be correct.

src/app/modules/react-migration/react-migration.controller.ts
wanlingt: removed in [00b75a7](https://github.com/opengovsg/FormSG/pull/5799/commits/00b75a7cc63e73d54ee8d1b501676332a0572be1)

src/app/modules/react-migration/react-migration.controller.ts
wanlingt: removed in [00b75a7](https://github.com/opengovsg/FormSG/pull/5799/commits/00b75a7cc63e73d54ee8d1b501676332a0572be1)

PR #5800, opened 2023-02-21
feat: retain metadata after image compression

### LINKED ISSUE #5754: Image metadata stripped after compression
**Describe the bug**
Our image compression method in react, [browser-image-compression](https://github.com/Donaldcwl/browser-image-compression), strips the metadata from the image. This is likely caused by the usage of canvas for the [compression](https://stackoverflow.com/questions/14672746/how-to-compress-an-image-via-javascript-in-the-browser/44849182#44849182).

**To Reproduce**
1. Go to any form with an attachment field
2. Drag and drop an image (with existing metadata) that has a larger size than the max attachment file limit
3. Submit the form
4. Check form response for the image attachment
5. Compare the metadata of the image in the response to the original

**Expected behavior**
Metadata should be retained after compression.

**Additional context**
This is a regression from angular, as it seems that angular was able to retain the metadata as seen from this [ticket](https://formsg.zendesk.com/agent/tickets/30210).

foochifa:
## Problem
In React, the image compression library that we currently use, [browser-image-compression](https://www.npmjs.com/package/browser-image-compression), does not retain EXIF metadata of images, as it strips the data before compressing it with html canvas. 

As some of our users rely on image metadata, this feature will help retain their workflow even if submission size of images is large. Our previous version in angularjs retains this metadata.

Closes #5754

## Solution
Replace the image compression library with [compressorjs](https://github.com/fengyuanchen/compressorjs)

**Breaking Changes** 

    - Details ...

**Features**:
EXIF data of JPEG images will now be preserved even after compression.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- [ ] Go to a form with an attachment field. Attempt to upload a jpeg, with metadata, that has a file size larger than the attachment field. The image should be successfully compressed. Submit the form.
- [ ] Check that the metadata of the original image is preserved in the compressed image, which was submitted to the form.
- [ ] Go to a form with an attachment field. Attempt to upload a png that has a file size larger than the attachment field. The image should be successfully compressed. Submit the form.

**New dependencies**:

- `compressorjs` : javascript lossy and async image compressor using the browser's canvas.toBlob API


timotheeg:
Any news from upstream on this one @foochifa ?

foochifa:
Oh yes @timotheeg ! The package was just released earlier this week, sorry have been busy with the payment issues. Though there is one issue that may arise from this:

Compressorjs does not have a max_size paramter. Hence, even after compression (only for Png files, Jpeg will still work) it may still be quite large. Was testing a png file that compresses from just 2.6MB to 1.1MB even after dropping quality and width/height to 1440 (this is also quite rare, after testing about 8 png files (of size 1-13.1), it only happened for one specific png). 

Thats why I have added error checking, to check if:
- compression fails cause browser doesn't support
- even after compression, size exceeds the limit

Also the metadata only works for jpeg files currently too haha there isn't any good package to handle metadata for png files (even though exif should have been integrated in 2017)

I have 4 options for this:
1. When image is still too large after compressing -> compress again at 0.5/0.75 of the previous width (may make compress for some Pngs slower, but it seems like compressorjs is already a bit faster than browser-image-compression)
2. Keep it as it is, on rare cases (very small, high quality, png files) it might force the user to compress the files themselves
3. Set a threshold in compressorjs to convert png -> jpeg -> compress -> png again (this will lose features like invisible background, though its unlikely for images with invisible background to be too large after compressing)
4. Keep browser-image-compression, but extract and append exif data using another package like [exifreader](https://www.npmjs.com/package/exifreader)

Current implementation is **option 3**

foochifa:
Some cool things I have found too:

1. Theoretically a jpeg exif data can be [infinite](https://stackoverflow.com/questions/3248946/what-is-the-maximum-size-of-jpeg-metadata/17415204#17415204), so if the exif data is already > maxSize, the compressed image with metadata will always exceed the maxSize. 
2. The [largest possible png](https://stackoverflow.com/questions/20415752/largest-potential-png-size-given-certain-dimensions) is width x height x 4B, so around 8MB if we get a square image and resize it to 1440x1440. 
3. Whilst for [jpeg](https://stackoverflow.com/questions/9806091/maximum-file-size-of-jpeg-image-with-known-dimensions) is ~width x height x 1B, 2MB if we get a square image.

foochifa:
CLOSING THIS our current `browser-image-compression` library, which I think is better as it provides more functionality (having a max_size to iterative compress the file down) [just got an update](https://github.com/Donaldcwl/browser-image-compression/pull/183) that helps to preserve exif metadata for JPEGs.

Will open a separate PR to bump browser-image-compression and enable the preserve exif flag

timotheeg:
Awesome! Thanks @foochifa !

PR #5828, opened 2023-02-22
feat: add prefix to OTPs

### LINKED ISSUE #5710: Improve OTP identification
## Context

Email and SMS OTPs can sometimes be slow to deliver, especially email OTP over government mailboxes, where the email filtering and validation can take time.

Because of that, many user request new OTPs, and when they receive the first email (delayed), they are dismayed to get an error that the OTP did not match. That's because there can only be a single active OTP for a user at a time, and the request for a second one made the first invalid. 

Unfortunately, users have no way of knowing which OTP is active, in relation to the emails and sms-es they are receiving. 

## Proposal: An OTP identification prefix

An identification prefix can be generated when an OTP is requested.  The prefix is sent as part of the email/sms, and will be shown in the submit form as static text, but it is not a user input. 

As in, the OTP will still be a 6-digits code, like `123456`, but the SMS/email contain the code `ABC-123456`, and the text field to enter the OTP shows `ABC-[...]` . That way people will know what OTP they are looking at.

Reference: DBS uses that prefix system in their OTP, see SMS content below:

> Fr DBS: Do not reveal OTP to anyone for your online txn of SGDXX.XX on VISA ending 9999 at Shopee. To approve, use **HTW-191206** within 3 min.

LinHuiqing:
## Problem

When multiple OTP requests are triggered, but messages come in after a long time, admin users have no way of checking which is the right OTP to key in other than trying out all of them. Refer to #5710 for more context.

Closes #5710

## Solution

Add a 3 letter prefix to OTPs to the OTP emails and the OTP input field.

**Breaking Changes** 
    - When the old UIs hit the new servers, since the response was never used, there would be no issue.
    - When the new UIs hit the old servers, `const { otpPrefix } = response` (in `frontend/src/features/login/LoginPage.tsx` line 119 to 122) will yield `undefined`. In this case, anything related to the OTP prefix should not be shown to users.

**Features**:

- Randomly generate 3 letters to be OTP prefix
- Add OTP prefix to OTP in email
- Add OTP prefix on the left of OTP input field
- Fix tests affected by adding of OTP prefix

## Before & After Screenshots

**BEFORE**:

OTP input form:
Screenshot 2023-02-22 at 3 06 59 PM

Email:
Screenshot 2023-02-22 at 3 09 52 PM

**AFTER**:

OTP input form:
Screenshot 2023-02-22 at 3 08 12 PM

OTP prefix on input form is updated when OTP is resent:

https://user-images.githubusercontent.com/37061143/220548740-52cacc08-570e-4713-8b3a-c8bdaa0741f7.mov

Email:
Screenshot 2023-02-22 at 3 10 51 PM

## Tests

At the login page, enter your .gov.sg email. You should be at the OTP input form.
- [ ] The OTP input should be prefixed by 3 letters followed by a dash (e.g. `ABC-`).
- [ ] The OTP email should have the OTP prefix before the OTP (e.g. `ABC-12345`).
- [ ] Wait for a OTP resend to be allowed, then resend the OTP. 
  - [ ] The OTP prefix on the email should change when the OTP sent is different.
  - [ ] The OTP prefix on the OTP input field should be updated to the right OTP prefix.
  - [ ] Entering the old OTP (with the old prefix) should not allow the user to login.
- [ ] Entering the OTP with the same OTP prefix as the input form should allow admin user to login.

### DIFF
--- a/frontend/src/components/Input/Input.tsx
+++ b/frontend/src/components/Input/Input.tsx
@@ -4,6 +4,7 @@ import {
   Icon,
   Input as ChakraInput,
   InputGroup,
+  InputLeftAddon,
   InputProps as ChakraInputProps,
   InputRightElement,
   useMultiStyleConfig,
@@ -55,14 +56,28 @@ export const Input = forwardRef<InputProps, 'input'>((props, ref) => {
 
   // Return normal input component if not success state.
   if (!props.isSuccess) {
-    return (
-      <ChakraInput
-        ref={ref}
-        {...preventDefault}
-        {...inputProps}
-        sx={props.sx ?? inputStyles.field}
-      />
-    )
+    if (props.prefix) {
+      return (
+        <InputGroup>
+          <InputLeftAddon pointerEvents="none" children={props.prefix} />
+          <ChakraInput
+            ref={ref}
+            {...preventDefault}
+            {...inputProps}
+            sx={props.sx ?? inputStyles.field}
+          />
+        </InputGroup>
+      )
+    } else {
+      return (
+        <ChakraInput
+          ref={ref}
+          {...preventDefault}
+          {...inputProps}
+          sx={props.sx ?? inputStyles.field}
+        />
+      )
+    }
   }
 
   return (
--- a/frontend/src/features/login/LoginPage.tsx
+++ b/frontend/src/features/login/LoginPage.tsx
@@ -99,6 +99,7 @@ export const LoginPage = (): JSX.Element => {
   const { data: { siteBannerContentReact, isLoginBannerReact } = {} } = useEnv()
   const [, setIsAuthenticated] = useLocalStorage<boolean>(LOGGED_IN_KEY)
   const [email, setEmail] = useState<string>()
+  const [otpPrefix, setOtpPrefix] = useState<string>('')
   const { t } = useTranslation()
 
   // TODO (#4279): Revert back to non-react banners post-migration.
@@ -115,8 +116,10 @@ export const LoginPage = (): JSX.Element => {
 
   const handleSendOtp = async ({ email }: LoginFormInputs) => {
     const trimmedEmail = email.trim()
-    await sendLoginOtp(trimmedEmail)
-    return setEmail(trimmedEmail)
+    await sendLoginOtp(trimmedEmail).then(({ otpPrefix }) => {
+      setOtpPrefix(otpPrefix)
+      setEmail(trimmedEmail)
+    })
   }
 
   const handleVerifyOtp = async ({ otp }: OtpFormInputs) => {
@@ -143,7 +146,7 @@ export const LoginPage = (): JSX.Element => {
     if (!email) {
       throw new Error('Something went wrong')
     }
-    await sendLoginOtp(email)
+    await sendLoginOtp(email).then(({ otpPrefix }) => setOtpPrefix(otpPrefix))
   }
 
   return (
@@ -186,6 +189,7 @@ export const LoginPage = (): JSX.Element => {
             ) : (
               <OtpForm
                 email={email}
+                otpPrefix={otpPrefix}
                 onSubmit={handleVerifyOtp}
                 onResendOtp={handleResendOtp}
               />
--- a/src/app/modules/auth/auth.controller.ts
+++ b/src/app/modules/auth/auth.controller.ts
@@ -111,7 +111,7 @@ export const _handleLoginSendOtp: ControllerHandler<
         })
     )
   } else {
-    // Step 4b: Error occurred whilst creating otp.
+    // Step 3b: Error occurred whilst creating otp.
     const error = loginOtpResult.error
     logger.error({
       message: 'Error creating login OTP',
--- a/src/app/modules/auth/auth.controller.ts
+++ b/src/app/modules/auth/auth.controller.ts
@@ -1,5 +1,6 @@
 import { StatusCodes } from 'http-status-codes'
 import { isEmpty } from 'lodash'
+import { SendOtpResponseDto } from 'shared/types/user'
 
 import { SUPPORT_FORM_LINK } from '../../../../shared/constants/links'
 import { createLoggerWithLabel } from '../../config/logger'
@@ -57,7 +58,7 @@ export const handleCheckUser = [
 
 export const _handleLoginSendOtp: ControllerHandler<
   unknown,
-  { message: string } | string,
+  { message: string } | SendOtpResponseDto,
   { email: string }
 > = async (req, res) => {
   // Joi validation ensures existence.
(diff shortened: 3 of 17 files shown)


frontend/src/components/Input/Input.tsx
```diff
@@ -4,6 +4,7 @@ import {
   Icon,
   Input as ChakraInput,
   InputGroup,
+  InputLeftAddon,
```
timotheeg, line 7: Cool! I'm impressed this is a component provided by default by chakra! 😄

src/app/modules/auth/auth.controller.ts
```diff
@@ -69,42 +70,60 @@ export const _handleLoginSendOtp: ControllerHandler<
...
+          })
+          const { errorMessage, statusCode } = mapRouteError(
+            error,
+            /* coreErrorMessage=*/ 'Failed to send login OTP. Please try again later and if the problem persists, contact us.',
+          )
+          return res.status(statusCode).json({ message: errorMessage })
+        })
+    )
+  } else {
+    // Step 4b: Error occurred whilst creating otp.
```
timotheeg: /nit incorrect step counter:
```suggestion
    // Step 3b: Error occurred whilst creating otp.
```

src/app/modules/auth/auth.controller.ts
```diff
@@ -69,42 +70,60 @@ export const _handleLoginSendOtp: ControllerHandler<
...
-          error,
+        // Step 5a: Successfully sent login otp.
+        .map(() => {
+          logger.info({
+            message: 'Login OTP sent successfully',
+            meta: logMeta,
+          })
+          return res
+            .status(StatusCodes.OK)
+            .json({ message: `OTP sent to ${email}`, otpPrefix })
```
timotheeg, line 97: This constitute a change of response type (not just field within an envelop), so we must account for backward compatibility at deployment time, when new UI may hit old servers and vice-versa.

So just to confirm: it looks like the previous response was never used, so old UI hitting new server should be OK, correct? And for the case of new UI hitting old server, code would get `const { otpPrefix } = response` yielding `undefined` (but at least JS should not break with a runtime error, since it can still destructure a string), yes?

Site note: sad that the previous step 4a and 4b did not have the same response type (`string` and `{message:string}` respectively) 😢 , we should favour always having envelops, rather than returning primitives.

frontend/src/features/login/LoginPage.tsx
```diff
@@ -115,8 +116,10 @@ export const LoginPage = (): JSX.Element => {
 
   const handleSendOtp = async ({ email }: LoginFormInputs) => {
     const trimmedEmail = email.trim()
-    await sendLoginOtp(trimmedEmail)
-    return setEmail(trimmedEmail)
+    await sendLoginOtp(trimmedEmail).then(({ otpPrefix }) => {
+      setOtpPrefix(otpPrefix)
```
timotheeg, line 120: Just noting: during deployment, if/when receiving a response from an old server, this would be `undefined` rather than a string. It looks harmless in the code and its uses, and would always be a string as soon as deployment is over (few minutes tops).

src/app/utils/otp.ts
```diff
@@ -17,6 +17,18 @@ export const generateOtp = (): string => {
     .join('')
 }
 
+/**
+ * Randomly generates and returns a 3 letter OTP prefix.
+ * @returns 3 letter OTP prefix string
+ */
+export const generateOtpPrefix = (): string => {
+  // Generates cryptographically strong pseudo-random data. 65 is the starting ASCII character code for upper case letters.
```
timotheeg, line 25: Nice code! I smiled at using cryptographically strong randomness here, which seemed unnecessary 🙂. The cost is low though, so let's keep it!

frontend/src/components/Input/Input.tsx
```diff
@@ -55,14 +56,28 @@ export const Input = forwardRef<InputProps, 'input'>((props, ref) => {
...
+      return (
+        <InputGroup>
+          <InputLeftAddon pointerEvents="none" children={props.prefix} />
+          <ChakraInput
+            ref={ref}
+            {...preventDefault}
+            {...inputProps}
+            sx={props.sx ?? inputStyles.field}
+          />
+        </InputGroup>
```
timotheeg, line 69: Do we need a guarantee that overall width is respected if it is provided for the Input props?

timotheeg (commented):
Nice PR @LinHuiqing !

I was having the thought that for backward compatibility of the release time, it could be confusing to an admin on the old UI (that he loaded just before deployment) that he gets the OTP email with prefix from the new servers, but the the old UI has no indication of what the prefix is for. I reckon the problem is minimal and can be be ignored however, because the OTP field has max length 6, and an error message indicating `Only numbers are allowed.`, which should make clear what the admin needs to do. Same reasoning applies when we add prefixes to email/sms verification OTP in forms as well.

src/app/modules/auth/auth.controller.ts
LinHuiqing: Yep seems like what you've brought up is true! Will add the 2 cases into the PR 👍 
We can slowly improve on the the response types :')

frontend/src/features/login/LoginPage.tsx
LinHuiqing: Just to note, the OTP field should look like the following when the UI is receiving a response from an old server (simulated with React DevTools):

Screenshot 2023-02-24 at 9 21 58 AM

LinHuiqing: Sorry, as discussed, the screenshot is wrong :') there will be a ' -' string in the grey prefix part.

timotheeg: I wonder if it was not doubly-wrong even? 😅🤔 

As-in, the string interpolation of `undefined` and the dash would probably have shown the string `undefined -`, no?

```
> let foo;
undefined

> `${foo} -`
'undefined -'
```

LinHuiqing: oh yea 🤦‍♀️ but its okay I think it should be fine now with the latest commit [6480f5f](https://github.com/opengovsg/FormSG/pull/5828/commits/6480f5fa3ccba88ef24a9988cecb16813a01ec8d)! let me test it out

LinHuiqing: Seems to work on my end!

frontend/src/components/Input/Input.tsx
LinHuiqing: What do you mean by this actually? 😅 Do you mean the 6 character width of the OTPs?

LinHuiqing:
Wait the OTP email still has a dash because of the template, let me see what I can do abt that

frontend/src/components/Input/Input.tsx
timotheeg: No, I mean on the layout. As in, if someone uses an Input component at whatever width, and we pass the prop as-is to the ChakraInput within the InputGroup, does having a prefix mean the overall width of the InputGroup be longer than the consumer intends?

timotheeg:
> Wait the OTP email still has a dash because of the template, let me see what I can do abt that

I think you can ignore this case. If the email includes the prefix (new backend), but the frontend doesn't reflect it (old ui), I think it'll be OK. I wrote about it in [my first note](https://github.com/opengovsg/FormSG/pull/5828#pullrequestreview-1310494593).

LinHuiqing:
> > Wait the OTP email still has a dash because of the template, let me see what I can do abt that
> 
> I think you can ignore this case. If the email includes the prefix (new backend), but the frontend doesn't reflect it (old ui), I think it'll be OK. I wrote about it in [my first note](https://github.com/opengovsg/FormSG/pull/5828#pullrequestreview-1310494593).

Sounds good then! Just to document, if the `otpPrefix` is `undefined`, the email will look like this:

Screenshot 2023-02-24 at 12 01 31 PM

timotheeg:
How could this situation happen? The email and prefix are both server code, so there's no old/new issue here, right?

frontend/src/components/Input/Input.tsx
LinHuiqing: Hm, do we want the generalised `Input` component to have that level of awareness though? We can have a check on the `OtpForm` component though, if we want to be sure of the prefix width.

LinHuiqing:
Ooh, that is true, it won't happen then! My bad, I was passing in `undefined`.

timotheeg (approved):

frontend/src/components/Input/Input.tsx
LinHuiqing: As discussed, this is not really a concern as `inputProps` is defined to include the props `isSuccess`, `isPrefilled` and `preventDefaultOnEnter`, which does not include width properties which might be affected by the adding of `InputLeftAddon`.

PR #5830, opened 2023-02-22
fix: squished logic block equals to/either option

### LINKED ISSUE #5508: logic block's equals to/either option is squished when the selected option has a long name
image

dexxtan:
Add minimum width 0 for flex child items that do not respect flex basis settings

## Problem

Closes or Fixes #5508 

## Solution

Flex basis which should be used to size the width of the rendered multi select input is not respected because there is a max width set on the multi select combo box input beside it. This results in the squished component.

The solution is to:
- convert max width into flex basis on the multi select combo box container
- rely on current set flex grow and flex shrink settings on the multi select combo box container to size that input
- not let the dropdown icon overflow to the next line due to insufficient width available
- and also apply min width 0 on both inputs for desktop screen sizes to let the implicit flex basis auto of the multi select input on the left size the input correctly
- also add a useEffect hook to set min width for the multi select input in case of other types of inputs needing this. It can be simplified from a switch to a simple if else since there is only one condition now

**Breaking Changes** 

    - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

From issue #5508:
issue

**AFTER**:

See issue fixed:
Screenshot 2023-02-23 at 12 54 41 AM

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/frontend/src/components/Dropdown/components/MultiSelectCombobox/MultiSelectCombobox.tsx
+++ b/frontend/src/components/Dropdown/components/MultiSelectCombobox/MultiSelectCombobox.tsx
@@ -25,7 +25,8 @@ const MultiItemsContainer: FC = ({ children }) => {
       // Margin difference for selected items.
       my="-3px"
       // Padding for dropdown toggle.
-      maxW="calc(100% - 2.5rem)"
+      flexBasis="calc(100% - 2.5rem)"
+      minW="0px"
     >
       {children}
     </Box>
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/EditConditionBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/EditConditionBlock.tsx
@@ -180,6 +180,16 @@ export const EditConditionBlock = ({
     }
   }, [currentSelectedField])
 
+  const logicValueWrapperWidth = useMemo(() => {
+    if (!ifValueTypeValue) return 'auto'
+    switch (ifValueTypeValue) {
+      case LogicIfValue.MultiSelect:
+        return '0px'
+      default:
+        return 'auto'
+    }
+  }, [ifValueTypeValue])
+
   const validateValueInputComponent = useCallback(
     (val) => {
       switch (ifValueTypeValue) {
@@ -367,6 +377,7 @@ export const EditConditionBlock = ({
                 isRequired
                 isReadOnly={isLoading}
                 isInvalid={!!get(errors, `${name}.value`)}
+                minW={{ md: logicValueWrapperWidth }}
               >
                 <VisuallyHidden
                   as="label"


dexxtan:
Hi, just leaving a comment here to please bear with me while I try to figure out the necessary things for contributing and PR reviews

justynoh (commented):
Hi @dexxtan, thanks for the contribution!

frontend/src/components/Dropdown/components/MultiSelectCombobox/MultiSelectCombobox.tsx
```diff
@@ -25,7 +25,8 @@ const MultiItemsContainer: FC = ({ children }) => {
       // Margin difference for selected items.
       my="-3px"
       // Padding for dropdown toggle.
-      maxW="calc(100% - 2.5rem)"
+      flexBasis="calc(100% - 2.5rem)"
+      minW="0px"
```
justynoh, line 29: We want to try and avoid modifying the components. Is there a solution where we just modify the layout higher up in the tree? Perhaps we can pass in some `flex` props in the `EditConditionBlock` to make it work.

justynoh:
Hi @dexxtan , just wanted to check in if you're still working on this?

frontend/src/components/Dropdown/components/MultiSelectCombobox/MultiSelectCombobox.tsx
timotheeg: This addition is a static addition that is harmless in normal use (correct?), I'm thinking it's acceptable 🤔

frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditCondition/EditConditionBlock.tsx
```diff
@@ -180,6 +180,16 @@ export const EditConditionBlock = ({
     }
   }, [currentSelectedField])
 
+  const logicValueWrapperWidth = useMemo(() => {
+    if (!ifValueTypeValue) return 'auto'
```
timotheeg, line 184: Is this line needed because the typescript linter is complaining? If not, the `switch`'s `default` block below already captures the falsy case, so the line here is redundant.

In fact, since the change is for a specific check, I suppose the whole memo could be simplified to:

```javascript
return ifValueTypeValue === LogicIfValue.MultiSelect ? '0px' : 'auto'
```

LinHuiqing: Yep, you're right!

frontend/src/components/Dropdown/components/MultiSelectCombobox/MultiSelectCombobox.tsx
LinHuiqing: Do correct me if I'm wrong, but I think this line change is unnecessary based on my observations on the UI behaviour.

LinHuiqing:
Opening new PR for this so we can run checks and merge it in! fyi: @dexxtan, not sure if you're still looking at this

LinHuiqing:
Closing in favour of #6305 
Let's move discussions, if any, to new PR here: https://github.com/opengovsg/FormSG/pull/6305

PR #5837, opened 2023-02-24
feat: enable receipt download

### LINKED ISSUE #5808: Enable download of payment receipt

tshuli:
## Problem
- Previously, after payment is completed, we provide a link for user to view the payment on Stripe website
- This PR introduces a new payment receipt download endpoint which returns the receipt pdf. The receipt is then downloaded onto the client machine.
- Closes #5808

**Details**
We make use of the [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header to download the file directly using the browser. 

**Screenshot**
- Receipt is saved as pdf
chrome-capture-2023-2-13

## Tests
- [ ] Create a form with payment. Complete payment. You should see receipt download button. Test that receipt download works successfully.


frontend/src/features/public-form/PublicFormService.ts
```diff
@@ -152,16 +154,37 @@ export const submitFormFeedback = async (
...
+  return ApiService.get(
+    `payments/receipt/${formId}/${submissionId}/download`,
+  ).then(({ data }) => {
+    const buffer = Buffer.from(data.pdfBuffer.data)
+
+    const blob = new Blob([buffer], { type: 'application/pdf' })
+    FileSaver.saveAs(blob, `${submissionId}-receipt.pdf`)
+    return
+  })
+}
```
timotheeg: I don't think we need to handle this ourselves. We can leverage the browser's native capabilities to download a file by using the [Content-Disposition header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) and not have to hold anything in memory. The `Download` button should just act as link to make it happen.

There is additional an issue with separation of concerns here 🤔. The file is generated as pdf on the server, so the server should set a content-type header and the appropriate file name+extension in the response.

src/app/utils/convert-html-to-pdf.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Utility function to generate a PDF from HTML.
+ * Used to send autoreply emails, and to generate payment receipts
+ * @param summaryHtml HTML to generate PDF from
+ * @returns PDF buffer
+ */
+export const generatePdfFromHtml = async (
+  summaryHtml: string,
+): Promise<Buffer> => {
+  const browser = await puppeteer.launch({
```
timotheeg: Let's wrap this function with a DD span, so we can monitor how long the pdf generation takes.

src/app/modules/payments/stripe.controller.ts
```diff
@@ -191,17 +193,70 @@ export const getPaymentReceipt: ControllerHandler<{
...
+        axios
+          .get<string>(receiptUrl)
+          // convert to pdf and return
+          .then((receiptUrlResponse) => {
+            const html = receiptUrlResponse.data
+            const pdfBufferPromise = generatePdfFromHtml(html)
+            return pdfBufferPromise
+          })
+          .then((pdfBuffer) => {
+            return res.status(StatusCodes.OK).json({ pdfBuffer })
```
timotheeg: I don't think json is the right medium to send a full file content. Let's have this endpoint return the file as-is, And supply relevant headers:
* [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)
* [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition)

Also [Content-Length](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) if express doesn't include that by default.

timotheeg (commented):
Good work!

I'm not really a fan of us taking over pdf rendering when the browser has this capability built-in 🤔 . This case is unlike the pdf attachment to emails, where the whole operation is server-side. 

And even if we do take over pdf generation, we should leverage browser capabilities wherever we can. Attachment download in binary is well supported, so there's no need for us to transmit a binary file in a json envelop, encoded to fit in a string and taking more space than needed in transmission.

src/app/utils/convert-html-to-pdf.ts
wanlingt: [a1e656d](https://github.com/opengovsg/FormSG/pull/5837/commits/a1e656d1de54860d037c5e4591a7c0470c22fff9)

src/app/modules/payments/stripe.controller.ts
wanlingt: [9a2c271](https://github.com/opengovsg/FormSG/pull/5837/commits/9a2c2716692a8899b0f75ce1f9bd3f0fca1cc002)

Content-Length is automatically assigned by express-js ([source](http://expressjs.com/en/4x/api.html))

frontend/src/setupProxy.js
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable no-undef */
+const { createProxyMiddleware } = require('http-proxy-middleware')
+
+// Added a proxy middleware as api routes with text/html in its Accept header
+// were not getting redirected to the proxy
```
timotheeg: Clarification on the behaviour (+ GammarNazi 😜)

```suggestion
// Added a proxy middleware, as requests to api routes with text/html in their Accept header
// were not getting proxied to the backend
```

src/app/modules/payments/stripe.controller.ts
```diff
@@ -191,17 +193,76 @@ export const getPaymentReceipt: ControllerHandler<{
...
+      return (
+        axios
+          .get<string>(receiptUrl)
+          // convert to pdf and return
+          .then((receiptUrlResponse) => {
+            const html = receiptUrlResponse.data
+            return tracer.trace('generatePdfFromHtml', () => {
+              const pdfBufferPromise = generatePdfFromHtml(html)
+              return pdfBufferPromise
+            })
```
timotheeg: It's probably better to create the DD span in the function `generatePdfFromHtml()` itself, so it will create spans regardless of who the caller is.

src/app/routes/api/v3/payments/payments.routes.ts
```diff
@@ -7,14 +7,25 @@ export const PaymentsRouter = Router()
...
+/**
+ * Downloads the receipt pdf
+ * @route GET /receipt/:formId/:submissionId/download
  *
  * @returns 200 with receipt URL exists
  */
 PaymentsRouter.route(
-  '/receipt/:formId([a-fA-F0-9]{24})/:submissionId([a-fA-F0-9]{24})',
-).get(StripeController.getPaymentReceipt)
+  '/receipt/:formId([a-fA-F0-9]{24})/:submissionId([a-fA-F0-9]{24})/download',
```
timotheeg, line 28: Let's add a TODO comment here that we should consider rate limiting this endpoint 🤔 . Please create an issue for that as well.

frontend/src/setupProxy.js
wanlingt: hahaha thanks tim, [aa46d3c](https://github.com/opengovsg/FormSG/pull/5837/commits/aa46d3c02b7afecbf6c75c33cca9b9bfb545f0b1)

src/app/modules/payments/stripe.controller.ts
wanlingt: [e8982e6](https://github.com/opengovsg/FormSG/pull/5837/commits/e8982e6dabba323362c014aae8d95ab16fa3dd61)

src/app/routes/api/v3/payments/payments.routes.ts
wanlingt: [c8ea1be](https://github.com/opengovsg/FormSG/pull/5837/commits/c8ea1bead9d14ab779ecdda23d00f62ca154993c)

issue #5924

src/app/services/mail/mail.utils.ts
```diff
@@ -174,7 +149,7 @@ export const generateAutoreplyPdf = (
   return tracer.trace('generateAutoreplyPdf', () =>
```
timotheeg: Now that `generatePdfFromHtml()` does its own span, we can probably remove this `span` 🤔 (because it doesn't wrap any other action than the pdf generation)

src/app/routes/api/v3/payments/payments.routes.ts
timotheeg: Thanks @wanlingt , I added some rationale in the ticket.

timotheeg (approved):
LGTM 👍 

I added one small comment on the DD span, but other than that, I think it's ready to go! 🎉

tshuli:
lgtm, thanks @wanlingt for taking this up!

only one minor thing which I considered during the review - currently we return the pdf file directly in the api response, and trigger the download using `Content-Disposition: attachment` header, and therefore we can't detect when the download is started or completed. This is because the browser handles the download process directly and doesn't provide any hooks or events to notify us when the download is complete.

As a result, we can't display the loading states in the save button during the 2-3s it takes for us to generate the pdf, or handle errors if the download fails (e.g. our server fails to fetch receipt from stripe, we still display the toast with  download started).
Recording 2023-03-14 at 14 00 33

Perhaps a more complete implementation might indeed be to store the generated pdf in s3, and return the pre-signed url in the api call for the browser to navigate to. This would allow us to properly handle all the states leading up to the start of the file download. But we discussed that we would not do that for this PR, so as to keep things simple.

This note is just for future reference, if this issue ever comes up again.

src/app/services/mail/mail.utils.ts
wanlingt: ooh yes thanks for spotting that tim. removed! [d662ee0](https://github.com/opengovsg/FormSG/pull/5837/commits/d662ee0b245d096eb0dd00d1b8522df39a47987d)

Issue #5803, opened 2023-02-22
Show payment details in individual response page


wanlingt:

LinHuiqing:
Implemented for response payment details in #5839:
- Display payment amount in dollars
- Add payment date

PR #5839, opened 2023-02-26
feat: add some payment details to responses page

wanlingt:
## Problem

Some payment details in the responses page have been hard-coded

Makes progress on #5803

## Solution

- Display payment amount in dollars
- Add payment date

Transaction fee is the remaining field that needs to be shown in the responses page. This will be done in a separate PR as we need to create a new transaction_fee property in the payments object, and also it requires refactoring of the `paymentIntents` Stripe API call (which is currently used in `getReceipt`).

**Breaking Changes** 

## Screenshots
**After**
image

## Tests

- [ ] Check that payment amount shows up in dollars and cents, instead of purely cents
- [ ] Check that the payment date displays the same date, time and timezone that the payment was made

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -194,7 +194,7 @@ export const IndividualResponsePage = (): JSX.Element => {
                     <Text textStyle="subhead-1">Payment amount:</Text>
                     <Text>
                       S$
-                      {(paymentData.amount / 100).toLocaleString('en-IN', {
+                      {(paymentData.amount / 100).toLocaleString('en-GB', {
                         minimumFractionDigits: 2,
                         maximumFractionDigits: 2,
                       })}
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -201,7 +201,7 @@ export const IndividualResponsePage = (): JSX.Element => {
                   </Stack>
                   <Stack direction={{ base: 'column', md: 'row' }}>
                     <Text textStyle="subhead-1">Payment status:</Text>
-                    <Text>{paymentData.status}</Text>
+                    <Text>{paymentData.status.toUpperCase()}</Text>
                   </Stack>
                   <Stack direction={{ base: 'column', md: 'row' }}>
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -186,34 +187,45 @@ export const IndividualResponsePage = (): JSX.Element => {
...
-            ) : (
+            ) : paymentData ? (
               <>
                 <Stack>
                   <Stack direction={{ base: 'column', md: 'row' }}>
                     <Text textStyle="subhead-1">Payment amount:</Text>
-                    <Text>{paymentData?.amount}</Text>
+                    <Text>
+                      S$
+                      {(paymentData.amount / 100).toLocaleString('en-IN', {
```
timotheeg: Why `en-IN`? We should specify British English norms with `en-GB` ([link here](https://en.wikipedia.org/wiki/Language_localisation#Language_tags_and_codes)), or maybe `en-SG` (if that exists 😅)

```suggestion
                      {(paymentData.amount / 100).toLocaleString('en-GB', {
```

LinHuiqing:
Seems like the payment status isn't updating (even after refreshing the page): 

Screenshot 2023-02-27 at 10 30 22 AM

wanlingt:
> Seems like the payment status isn't updating (even after refreshing the page):
> 

Yup, that's a separate issue though. When I tested it, the status of the payment object isn't being correctly updated in the DB through the webhooks or after the API call to Stripe. Was it the same for you in this case? I'll create a separate issue for this! This PR just focuses on displaying that property from the payment object in the FE.

timotheeg (approved):

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
wanlingt: changed here [759236d](https://github.com/opengovsg/FormSG/pull/5839/commits/759236d4a79610214035bdf0175e348cd4e11e3d)! (used `en-GB`)

LinHuiqing:
> > Seems like the payment status isn't updating (even after refreshing the page):
> 
> Yup, that's a separate issue though. When I tested it, the status of the payment object isn't being correctly updated in the DB through the webhooks or after the API call to Stripe. Was it the same for you in this case? I'll create a separate issue for this!

Ohh okay makes sense, my bad. Yep, the status of the payment object isn't updated on the DB on my end too.

Issue #5822, opened 2023-02-22
Protect form update API endpoint to NOT allow overwriting payment.account_id


LinHuiqing:
The target account id can only be set with a successful oAuth flow, a simple form edit should never be able to edit/delete the payment.account_id

PR #5843, opened 2023-02-27
fix: remove payment account id from payment updates

### LINKED ISSUE #5812: Remove target account id from "edit payment" page

foochifa:
Closed in favour of #5929 and #5903 

## Problem

Currently we show and allow changes to payment account id by the admin through the payments drawer tab.

However, this behaviour should not be allowed, and payment account id should only be modifiable throught the Stripe OAuth API

Closes #5812, #5822

## Solution

Removed the target account id field in payments drawer tab
Decouple account id and publishable key from payments update

**Breaking Changes** 

- No - this PR is backwards compatible  .

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- [ ] go to a form with payments enabled. Go to the payment drawer tab, ensure that there isn't an account id field present.
- [ ] go to a form with payments enabled. Attempt to change the account id through the `/payment` endpoint. Your request should be rejected

### DIFF
--- a/shared/types/form/form.ts
+++ b/shared/types/form/form.ts
@@ -240,4 +240,4 @@ export type StartPageUpdateDto = FormStartPage
 export type EndPageUpdateDto = FormEndPage
 export type FormPermissionsDto = FormPermission[]
 export type PermissionsUpdateDto = FormPermission[]
-export type PaymentsUpdateDto = FormPayments
+export type PaymentsUpdateDto = Omit<FormPayments, 'target_account_id'>
--- a/shared/types/form/form.ts
+++ b/shared/types/form/form.ts
@@ -68,13 +68,18 @@ export enum FormResponseMode {
   Email = 'email',
 }
 
-export type FormPayments = {
+type FormPaymentsAccount = {
+  target_account_id: string
+  publishable_key: string
+}
+
+type FormPaymentsField = {
   enabled: boolean
-  target_account_id?: string
-  publishable_key?: string
   amount_cents?: number
   description?: string
 }
+
+export type FormPayments = Partial<FormPaymentsAccount> & FormPaymentsField
 
 export interface FormBase {
   title: string
@@ -240,5 +245,4 @@ export type StartPageUpdateDto = FormStartPage
 export type EndPageUpdateDto = FormEndPage
 export type FormPermissionsDto = FormPermission[]
 export type PermissionsUpdateDto = FormPermission[]
-// target account id should only be changed through Stripe's OAuth Callback
-export type PaymentsUpdateDto = Omit<FormPayments, 'target_account_id'>
+export type PaymentsUpdateDto = FormPaymentsField
--- a/shared/types/form/form.ts
+++ b/shared/types/form/form.ts
@@ -240,4 +240,5 @@ export type StartPageUpdateDto = FormStartPage
 export type EndPageUpdateDto = FormEndPage
 export type FormPermissionsDto = FormPermission[]
 export type PermissionsUpdateDto = FormPermission[]
+// target account id should only be changed through Stripe's OAuth Callback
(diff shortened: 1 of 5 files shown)


shared/types/form/form.ts
```diff
@@ -240,4 +240,5 @@ export type StartPageUpdateDto = FormStartPage
 export type EndPageUpdateDto = FormEndPage
 export type FormPermissionsDto = FormPermission[]
 export type PermissionsUpdateDto = FormPermission[]
-export type PaymentsUpdateDto = FormPayments
+// target account id should only be changed through Stripe's OAuth Callback
+export type PaymentsUpdateDto = Omit<FormPayments, 'target_account_id'>
```
justynoh: I think we can be a bit more explicit here rather than omitting specific keys from the `FormPayments` object. Since `FormPayments` contains five keys,

```typescript
export type FormPayments = {
  enabled: boolean
  target_account_id?: string
  publishable_key?: string
  amount_cents?: number
  description?: string
}
```

and two of them (`target_account_id` and `publishable_key`) can only be written by OAuth callback and the other three can only be written by updating the form field, this creates a nice separation of concern that we should reflect in the code. My suggestion would be to do something like this:

```typescript
export type FormPaymentsAccount = {
    target_account_id: string
    publishable_key: string
}

export type FormPaymentsField = {
    enabled: boolean
    amount_cents?: number
    description?: string
}

export type FormPayments = Partial<FormPaymentsAccount> & FormPaymentsField
``` 

You can even go one step further and require that enabled is true iff `amount_cents` and `description` exists:

```typescript
type FormPaymentsFieldDisabled = {
   enabled: false
}

type FormPaymentsFieldEnabled = {
   enabled: true
   amount_cents: number
   description: string
}

type FormPaymentsField = FormPaymentsFieldDisabled | FormPaymentsFieldEnabled
```

You would also need to update the type of the endpoint handlers for both oauth and payment field update. What do you think?

foochifa: Oh thanks @justynoh! Yeah it makes more sense this way method should be cleaner, I just realised that publishable key isn't editable on the payments tab too 😅. Will make the amendments thank you!

foochifa: @justynoh 
:sob: FieldErrors can't handle union types

https://stackoverflow.com/questions/74219465/typescript-react-hook-form-error-handling-with-zod-union-schema

https://github.com/react-hook-form/react-hook-form/issues/9287

I could use the workaround to explicit cast the error 
`const formErrors: FieldErrors<FormPaymentsFieldEnabled> = errors` 
Before accessing `formErrors.amounts_cents` But its kinda ugly and we have to export `FormPaymentsFieldEnabled`

Else could either drop the discriminated union or we could potentially set the other parameters in `FormPaymentsFieldDisabled`  as optional undefined like:
```typescript
type FormPaymentsFieldDisabled = {
  enabled: false
  amount_cents?: undefined
  description?: undefined
}
```

foochifa: Actually thinking about it, do we need to prevent updates of amount and description if payment is disabled? 😅 Not sure how it will impact direct api calls though

justynoh: up to you haha. the problem here is with frontend, can always redefine the frontend form types:

```typescript
// shared/types
type FormPaymentsField = ...

// frontend/src/.../PaymentFIeld.tsx (or whatever)
type FormPaymentsFieldInputs = {
  enabled: boolean,
  amount_cents: number,
  description: string
}
```

Of course this duplicates the definition of the shared type, so maybe not so clean. Just go with whichever one you feel is cleanest!

Just make sure we have backend checks, then it'll be ok haha

foochifa:
Will update and streamline this PR after updating payment schema #5872

PR #5852, opened 2023-02-28
feat: input and display payment amount in [dollars].[cents]

### LINKED ISSUE #5804: Show amounts in dollars, even if they are stored in cents internally

LinHuiqing:
## Problem

Payment amount was displayed in cents as handling money in integers avoids issues that might be caused by floating points. However, this is not what users are used to seeing - we are used to seeing currency displayed as dollars and cents, separated by decimal points.

Closes #5804 

## Solution

- Before saving, convert the 2dp payment amount to the appropriate integer amount in cents.
- Before displaying the previously saved values (which are saved in cents), convert the amount in cents back to dollars and cents.

**Breaking Changes** 

- [x] No - this PR does not contain breaking changes

**Features**:

- Update look of field where admin enters the amount they want to collect for payment.
- `MoneyInput` component for where admin should key in the payment amount for the form.
- Separate typing for internal state of form and display state of forms.

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/37061143/221749008-917e80d1-adbe-4584-889d-8013a72fb647.mov

**AFTER**:

https://user-images.githubusercontent.com/37061143/223621054-79a6c9ec-6542-4104-9718-f8d304afaf07.mov

## Tests

After connecting the form to stripe, go to create > add payments. Check for the following:
- [ ] Input box should be labelled "Payment Amount"
- [ ] The sample payment block on the right should reflect the right payment amount for any valid amount entered:
  - [ ] Any integer amount (representing dollars)
  - [ ] Any float with 2 dp
  - [ ] Any float with 1 dp
  - [ ] Any valid amount starting with '.'
  - [ ] Any valid amount ending with '.'
- [ ] If multiple '.' are entered in the box, payment amount input box should be invalid and ask admin to key in a valid decimal.
- [ ] If value keyed in by admin has more than 2 decimal places, payment amount input box should be invalid and ask admin to keep the value to 2dp.
- [ ] If value keyed in by admin is more than $1000, payment amount input box should be invalid and ask admin to key in a value below $1000.
- [ ] If value keyed in by admin is a negative number, payment amount input box should be invalid and ask admin to key in a positive value.
- [ ] If value keyed in by admin is any other invalid monetary amount (e.g. "23+12"), payment amount input box should be invalid and ask admin to key in a valid amount.
- [ ] The correct payment amount should be saved (and retrieved thereafter) for any valid amount entered:
  - [ ] Any integer amount (representing dollars)
  - [ ] Any float with 2 dp
  - [ ] Any float with 1 dp
  - [ ] Any valid amount starting with '.'
  - [ ] Any valid amount ending with '.'
- [ ] The correct payment amount should be seen on the respondents' payment page for any valid amount entered: 
  - [ ] Any integer amount (representing dollars)
  - [ ] Any float with 2 dp
  - [ ] Any float with 1 dp
  - [ ] Any valid amount starting with '.'
  - [ ] Any valid amount ending with '.'


frontend/src/features/admin-form/create/payment/PaymentContent.tsx
```diff
@@ -67,7 +67,11 @@ export const PaymentContent = (): JSX.Element => {
                       <Text textStyle="body-1" textColor="secondary.700">
                         Your credit card will be charged:{' '}
                         <Text as="span" fontWeight="bold">
-                          S${((payments?.amount_cents ?? 0) / 100).toFixed(2)}
+                          S$
+                          {(payments?.payment_amount === undefined
+                            ? 0
+                            : Number(payments.payment_amount)
+                          ).toFixed(2)}
```
timotheeg: /nit falsy check is probably good enough

```suggestion
                          {(payments?.payment_amount
                            ? 0
                            : Number(payments.payment_amount)
                          ).toFixed(2)}
```

frontend/src/components/NumberInput/NumberInput.tsx
```diff
@@ -88,31 +94,65 @@ export const NumberInput = forwardRef<NumberInputProps, 'input'>(
...
+                  __css={styles.icon}
+                />
+                <Text>SGD</Text>
+              </HStack>
+            </InputLeftAddon>
+            {inputBox(true)}
+          </InputGroup>
+        ) : (
+          inputBox(false)
+        )}
```
justynoh: If we want a money input, would it be more appropriate to create a separate component for `MoneyInput` instead of trying to bake the functionality into `NumberInput`? Adding a specific `InputLeftAddon` feels somewhat outside the scope of what `NumberInput` is supposed to be for.

src/app/models/form.server.model.ts
```diff
@@ -446,14 +446,14 @@ const compileFormModel = (db: Mongoose): IFormModel => {
           type: String,
           default: '',
         },
-        amount_cents: {
+        payment_amount: {
```
justynoh: Is there any specific reason to change the variable name here?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -69,7 +69,13 @@ export const PaymentInput = (): JSX.Element => {
...
   } = useForm<FormPayments>({
     mode: 'onBlur',
-    defaultValues: paymentsData,
+    defaultValues: {
+      ...paymentsData,
+      // Change payment_amount value to 2 decimal places temporarily for the form
+      payment_amount: paymentsData?.payment_amount
+        ? paymentsData?.payment_amount / 100
+        : 0,
+    },
```
justynoh: Rather than dealing with floating point numbers throughout the frontend, I wonder if it would be more appropriate to always keep an integer internal state, and then only convert the integer to it's currency representation at the point of rendering to the user. Not fully investigated, just the skeleton of the idea:

```typescript
// utils
const monetizeInteger = (val: number): string => ... // convert to "$xxx.xx" format

// input
const onChange = (val: string) => {
  // Strip out all non-digit characters
  const amount_cents = Number(val.replace(/\D/g,''))
  // Recompute correct value
  ... monetizeInteger(amount_cents)
}

File: frontend/src/components/NumberInput/NumberInput.tsx (hunk 2 above)
Comment (LinHuiqing) [reply frontend/src/components/NumberInput/NumberInput.tsx]: Makes sense! Will make a separate component for money 👍

File: src/app/models/form.server.model.ts (hunk 3 above)
Comment (LinHuiqing) [reply src/app/models/form.server.model.ts]: I made the change to the variable name because I was using the same type for when the payment amount is in cents and when its in dollars and cents. I was a bit conflicted about this too when I was implementing it. 
As discussed w @timotheeg, I'll change it back! It's better for backward compatibility and understanding too.

File: frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx (hunk 4 above)
Comment (LinHuiqing) [reply frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx]: I think for form control, the value in `payments` object is directly tied to the value shown in the input box(?) That's why I did the conversion before loading and before saving the values. Lemme see what I can do to properly type it.

File: frontend/src/components/NumberInput/NumberInput.tsx (hunk 2 above)
Comment (LinHuiqing) [reply frontend/src/components/NumberInput/NumberInput.tsx]: Note: I've separated the component used into MoneyInput, but the currency flag and prefix is currently hard-coded in the component itself to be the Singapore flag and "SGD" respectively. ([b9cff23](https://github.com/opengovsg/FormSG/pull/5852/commits/b9cff23c97ebc6ae9f4b8f7dd306075e0f2e3e2c))

File: src/app/models/form.server.model.ts (hunk 3 above)
Comment (LinHuiqing) [reply src/app/models/form.server.model.ts]: I've introduced `FormPaymentsDisplay` to separate typing for input values on the form and internal states (typed with `FormPayments`). So internal states will use `amount_cents` while floats can still be displayed on the form in `PaymentDrawer`.

File: frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx (hunk 4 above)
Comment (LinHuiqing) [reply frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx]: Oh for this, I can't seem to find a way to do this easily... So I simply kept the typing used for the form values displayed (`FormPaymentsDisplay`) and form loading / submission (`FormPayments`) separate. `PaymentStore` will always be storing the payment amount in cents (i.e. integers), and the form itself has a set of states where payment amount is reflective of the dollar value (i.e. 2dp floats).

**Comment by LinHuiqing** [comment]
I think the `PaymentDrawer` needs frontend tests, but I'll create a separate issue for now?

File: frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx [hunk 5]
```diff
@@ -108,32 +120,72 @@ export const PaymentInput = (): JSX.Element => {
...
-  const amountValidation: RegisterOptions<FormPayments, 'amount_cents'> =
-    useMemo(
-      () => ({
-        validate: {
-          validNumber: (val) => {
-            // Check whether input is a valid number, avoid e
-            return !isNaN(Number(val)) || 'Please enter a valid number'
-          },
+  const maxPaymentAmount = 1000 // due to IRAS requirements and agency financial institutions are expected to be in SG
+  const currencyFormatter = new Intl.NumberFormat('en-IN', {
```
timotheeg: Why `en-IN`?

timotheeg (approved):
LGTM

PR #5855, opened 2023-02-28
fix: add auto-trimming for NRIC field

### LINKED ISSUE #5854: NRIC field does not auto correct leading/trailing whitespaces
**Describe the bug**
Another regression in the React App (works in Angular) - if the form filler inputs an NRIC with whitespaces into the NRIC field, they are unable to submit their form

**Expected behavior**
NRIC field should auto trim leading/trailing whitespaces

**Additional Context**
Also mentioned in https://github.com/opengovsg/FormSG/issues/5655#issuecomment-1367152964

KenLSM:
## Problem

Spaces could be unintentionally included by the user when they are entering their NRIC. This can be incredibly frustrating for the user to identify why their NRIC had failed the validation. This is a regression issue arising from the react-angular migration.

Closes #5854

## Solution

We can remove one potential source of frustration through the inclusion of trimming the NRIC as there are no need for trailing/leading spaces.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-02-28 at 7 17 20 PM

**AFTER**:
Screenshot 2023-02-28 at 6 57 29 PM

## Tests

- [ ] go to a form with an NRIC input. Attempt to enter a valid NRIC. It should pass validation without any error message appearing.
- [ ] go to a form with an NRIC input. Attempt to enter a valid NRIC with trailing or leading spaces. The input should not be updating with spaces present. It should pass validation without any error message appearing.

### DIFF
--- a/frontend/src/templates/Field/Nric/NricField.test.tsx
+++ b/frontend/src/templates/Field/Nric/NricField.test.tsx
@@ -8,6 +8,9 @@ import * as stories from './NricField.stories'
 
 const { ValidationRequired, ValidationOptional } = composeStories(stories)
 
+const VALID_NRIC = 'S0000002G'
+const INVALID_NRIC = 'S0000001B'
+
 describe('validation required', () => {
   it('renders error when field is not filled before submitting', async () => {
     // Arrange
@@ -38,12 +41,12 @@ describe('validation required', () => {
 
     // Act
     // Valid NRIC
-    await user.type(input, 'S0000002G')
+    await user.type(input, VALID_NRIC)
     await user.click(submitButton)
 
     // Assert
     // Should show success message.
-    const success = screen.getByText('You have submitted: S0000002G')
+    const success = screen.getByText(`You have submitted: ${VALID_NRIC}`)
     expect(success).not.toBeNull()
     const error = screen.queryByText('Please fill in required field')
     expect(error).toBeNull()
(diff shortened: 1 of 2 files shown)


frontend/src/templates/Field/Nric/NricField.tsx
```diff
@@ -28,7 +29,12 @@ export const NricField = ({ schema }: NricFieldProps): JSX.Element => {
         aria-label={`${schema.questionNumber}. ${schema.title}`}
         defaultValue=""
```
timotheeg, line 29: Is `defaultValue=""` still needed now that we pass a value in explicitly?

frontend/src/templates/Field/Nric/NricField.tsx
```diff
@@ -28,7 +29,12 @@ export const NricField = ({ schema }: NricFieldProps): JSX.Element => {
...
         preventDefaultOnEnter
-        {...register(schema._id, validationRules)}
+        value={value}
+        {...register(schema._id, {
+          ...validationRules,
+          onChange: (event: React.ChangeEvent<HTMLInputElement>) =>
+            setValue(event.target.value.trim()),
+        })}
```
timotheeg: So just to confirm the render flow: if a user inputs a leading/trailing space, [validation](https://github.com/opengovsg/FormSG/blob/0f211f486f06c57a3d898dcb136eeb38cd38df7b/frontend/src/utils/fieldValidation.ts#L321-L339) will fail, but `onChange()` will fire and correct the input, validation will run again, and it will pass, and the user will not see any error message flash, correct?

frontend/src/templates/Field/Nric/NricField.test.tsx
```diff
@@ -8,6 +8,9 @@ import * as stories from './NricField.stories'
 
 const { ValidationRequired, ValidationOptional } = composeStories(stories)
 
+const VALID_NRIC = 'S0000002G'
+const INVALID_NRIC = 'S0000001B'
```
timotheeg, line 12: 👍 👏

timotheeg (approved):

frontend/src/templates/Field/Nric/NricField.tsx
KenLSM: Yes to `user will not see any error message flash`.  The flow works as expected for users that is correcting their input, but will fail when user had explicitly added a space. But to be pedantic, it will be called with the data that the user had input as opposed to the output of `setState`.

The reason for why this happens is due to the fact that `react-hooks-form` operates on uncontrolled component, and my initial implementation is turning the component into a controlled component, which breaks the assumption that `react-hooks-form` had.

The correct way to manipulate the form data is to use `setValue` hook provided by the form's context.

frontend/src/templates/Field/Nric/NricField.tsx
KenLSM: Have added a switch to the implementation which now requires the use of the `defaultValue` field.

frontend/src/templates/Field/Nric/NricField.tsx
timotheeg: Thanks for the explanation!

timotheeg (approved):
LGTM

PR #5861, opened 2023-03-01
fix: decouple payment response from individual response page

### LINKED ISSUE #5857: Individual response page throws an error when the form does not contain payment object
This is because `submission/${submissionId}/payment` endpoint is called regardless of `payments.enabled` stauts

foochifa:
## Problem

Currently, irregardless of whether the form has payment enabled, we will call the `/submission/{submissionId}/payment` endpoint. This returns a 500 error for non-payment enabled forms, as the payment object will not exist.

Closes #5857

## Solution

Decouple the payment response from the overall individual response page. We will only call the endpoint if payment has been enabled on the form.

**Breaking Changes** 

## Before and After Screenshots

**Before**
image
image

## Tests

- [ ] go to a storage form without payments enabled. Try to access the individual response in the 'results' tab. There should not be any error.
- [ ] go to a storage form with payments enabled. Try to access the individual response in the 'results'' tab. There should not be be any error message and the payment details field should be visible

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
@@ -1,4 +1,4 @@
-import { Stack, Text } from '@chakra-ui/react'
+import { Skeleton, Stack, Text } from '@chakra-ui/react'
 
 import { useIndividualPaymentSubmission } from './queries'
 
@@ -7,8 +7,12 @@ export const IndividualPaymentResponse = (submissionId: {
 }): JSX.Element => {
   if (!submissionId) throw new Error('Missing submissionId')
 
-  // TODO: get paymentData from IndividualResponsePage through submission endpoint
-  const { data: paymentData } = useIndividualPaymentSubmission()
+  // TODO: get paymentData from IndividualResponsePage through submission endpoint when we refactor the endpoints
+  const {
+    data: paymentData,
+    isError,
+    isLoading,
+  } = useIndividualPaymentSubmission()
 
   return (
     <Stack>
@@ -21,7 +25,10 @@ export const IndividualPaymentResponse = (submissionId: {
       >
         Payment
       </Text>
-      {paymentData ? (
+      {isError || isLoading ? (
+        //TODO: check with design for skeleton layout
+        <Skeleton height="1.5rem" />
+      ) : paymentData ? (
         <>
           <Stack>
             <Stack direction={{ base: 'column', md: 'row' }}>
@@ -64,6 +71,7 @@ export const IndividualPaymentResponse = (submissionId: {
           </Stack>
         </>
       ) : (
+        //TODO: check with design for exact copy
         <Text>Payment data not found</Text>
       )}
     </Stack>
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
@@ -1,0 +1,76 @@
+import { Stack, Text } from '@chakra-ui/react'
+
+import { useIndividualPaymentSubmission } from './queries'
+
+export const IndividualPaymentResponse = (submissionId: {
+  submissionId: string
+}): JSX.Element => {
+  if (!submissionId) throw new Error('Missing submissionId')
+
+  const {
+    data: paymentData,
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Text
+        textStyle="h2"
+        as="h2"
+        color="primary.500"
+        mb="0.5rem"
+        _notFirst={{ mt: '2.5rem' }}
+      >
+        Payment
+      </Text>
+      {isPaymentLoading || isPaymentError ? (
```
foochifa: from @wanlingt, looking forward the loading & error portion will likely be covered under useIndividualSubmission() fin IndividualResponsePage.

we can potentially leverage on that, and use the loading decryption skeleton from Individual Response page instead of returning `<Text>Payment was not enabled when this form was submitted</Text>` (Which isn't being called during error anyway).

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
```diff
@@ -0,0 +1,71 @@
+import { Stack, Text } from '@chakra-ui/react'
+
+import { useIndividualPaymentSubmission } from './queries'
+
+export const IndividualPaymentResponse = (submissionId: {
+  submissionId: string
+}): JSX.Element => {
+  if (!submissionId) throw new Error('Missing submissionId')
+
+  // TODO: get paymentData from IndividualResponsePage through submission endpoint
```
wanlingt: sorry I don't understand / have forgotten what this TODO is about! can provide more context?

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { data: paymentData } = useIndividualPaymentSubmission()
```
wanlingt: now I'm wondering if it makes sense to have the error and isLoading states (and the LoadingDecryption or just a normal loading skeleton) here too.. :sweat_smile: haha what do you think?

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
foochifa: Ah, cause currently we are getting payment data from `submissions/:submissionId/payment` endpoint. 

However, ideally, the payment data should have been retrieved together with all the other data from `submissions/:submissionId`. So eventually we will be passing the entire paymentdata as a parameter to IndividualPaymentResponse.

wanlingt: ah gotcha, so after the future PR where we fix the endpoints

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
foochifa: Oh but then we will need to make a separate loadingdecrpytion skeleton for the payment response too right? Else it will look a lil awkward hahaha

Though I think it might be worth doing it, at least with a payment header or something, so the admin who sees it knows that he has payment enabled. Thanks @wanlingt will add it in!

wanlingt: yup! or maybe can just use the standard Skeleton component from chakra not sure will have to consult @staceytan1998 !

thanks chifa :)

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
foochifa: yup! This issue #5864

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualPaymentResponse.tsx
wanlingt: [00473ba](https://github.com/opengovsg/FormSG/pull/5861/commits/00473bac19938dd74884715d2354f1b16e302bc3)

wanlingt (approved):
lgtm ;)

Issue #5866, opened 2023-03-01
Improve OTP identification for respondent mobile number


LinHuiqing:
This issue to is track implementation for respondent OTP verification for mobile number. Go to #5865 for email.

## Context

When respondents try to verify their email / mobile number, they might get impatient and trigger multiple otps before the OTPs come into their inbox / messages. This causes frustration when users key in outdated OTPs and are unable to verify their emails / mobile numbers.

Branched off #5710

## Proposed Solution

An identification prefix can be generated when an OTP is requested.  The prefix is sent as part of the email/sms, and will be shown in the submit form as static text, but it is not a user input. 

As in, the OTP will still be a 6-digits code, like `123456`, but the SMS/email contain the code `ABC-123456`, and the text field to enter the OTP shows `ABC-[...]` . That way people will know which OTP they need to look for.

Reference: DBS uses that prefix system in their OTP, see SMS content below:

> Fr DBS: Do not reveal OTP to anyone for your online txn of SGDXX.XX on VISA ending 9999 at Shopee. To approve, use **HTW-191206** within 3 min.

Refer to #5828 for OTP prefix implementation for admin logins.

PR #5874, opened 2023-03-02
feat: otp prefix for respondents email and mobile fields

### LINKED ISSUE #5865: Improve OTP identification for respondent emails
This issue to is track implementation for respondent OTP verification for emails. Go to #5866 for mobile number.

## Context

When respondents try to verify their email / mobile number, they might get impatient and trigger multiple emails before the emails come into the inbox. This causes frustration when users key in outdated OTPs and are unable to verify their emails / mobile numbers.

Branched off #5710

## Proposed Solution

An identification prefix can be generated when an OTP is requested.  The prefix is sent as part of the email/sms, and will be shown in the submit form as static text, but it is not a user input. 

As in, the OTP will still be a 6-digits code, like `123456`, but the SMS/email contain the code `ABC-123456`, and the text field to enter the OTP shows `ABC-[...]` . That way people will know what OTP they are looking at.

Reference: DBS uses that prefix system in their OTP, see SMS content below:

> Fr DBS: Do not reveal OTP to anyone for your online txn of SGDXX.XX on VISA ending 9999 at Shopee. To approve, use **HTW-191206** within 3 min.

Refer to #5828 for OTP prefix implementation for admin logins.

KenLSM:
## Problem
When multiple OTP requests are triggered, but messages come in after a long time, respondents have no way of checking which is the right OTP to key in other than trying out all of them. Refer to https://github.com/opengovsg/FormSG/issues/5710 for the original context.

Closes #5865, #5866 

## Solution
Added random prefix for respondent to uniquely identify which OTP they should be expected to input.

**Breaking Changes** 
- [X] Yes - If the responder hits an old BE server while on new FE, the otp verification flow will be broken for the user until our old BE servers are updated.

**Features**:

- Randomly generate 3 letters to be OTP prefix
- Add OTP prefix to OTP in email
- Add OTP prefix on the left of OTP input field
- Fix tests affected by adding of OTP prefix

## Before & After Screenshots
**BEFORE**:
Responder Mail OTP
Screenshot 2023-03-03 at 10 08 14 AM

Responder Mobile Number OTP
Screenshot 2023-03-03 at 10 07 38 AM

Email Response Copy

Screenshot 2023-03-03 at 10 10 52 AM

**AFTER**:
Responder Mail OTP
Screenshot 2023-03-02 at 9 48 57 PM

Responder Mobile Number OTP
Screenshot 2023-03-02 at 9 48 46 PM

Email Response Copy
Screenshot 2023-03-03 at 10 09 43 AM

## Tests
- [ ] OTP prefix should exist on FE on OTP enabled

  - [ ] email verification field
  - [ ] mobile number verification field
- [ ] OTP prefix should sent to user's
  - [ ] mobile device
  - [ ] email


frontend/src/features/verifiable-fields/components/VerificationBox/VerificationBox.tsx
```diff
@@ -20,6 +26,7 @@ type VfnFieldValues = {
 
 export interface VerificationBoxProps {
   fieldType: VerifiableFieldType
+  otpPrefix: string | null
```
LinHuiqing: Is there a reason why we are using `null` instead of `undefined` (or `?`) here? Just trying to understand what are the general pros and cons cos I would think both have the same effect here but `?` might be slightly cleaner?

frontend/src/features/verifiable-fields/VerifiableFieldService.ts
```diff
@@ -54,13 +56,13 @@ export const triggerSendOtp = async ({
   fieldId: string
   /** The value of the verification field to verify. Usually an email or phone number */
   answer: string
-}): Promise<void> => {
+}): Promise<SendFormOtpResponseDto> => {
```
LinHuiqing, line 59: Will this mean that if the new UI sends a request to the old server, the frontend will break? Since it the new UI will expect `{ otpPrefix: string }` in the payload but that won't be served by the old servers.

KenLSM: Yeah. I assume that aside from the initial release window, our BE almost be on the latest version as opposed to the FE where user doesn't refresh.

frontend/src/features/verifiable-fields/components/VerificationBox/VerificationBox.tsx
KenLSM: `otpPrefix?: string` would mean that the component allows for omission of `otpPrefix`

```jsx
// valid
<VerificationBox fieldType={type} otpPrefix={""} />

// also valid
<VerificationBox fieldType={type} />
```

But if we do `otpPrefix: string | null`, the component is required to handle and provide the `otpPrefix` props


```jsx
// valid
<VerificationBox fieldType={type} otpPrefix={""} />

// invalid
<VerificationBox fieldType={type} />

// now it is valid
<VerificationBox fieldType={type} otpPrefix={null} />
```

From the design standpoint, I was making the assumption that we want the `otpPrefix` to available rather than omitted.

LinHuiqing: Ah, good point. In this case, is there a reason why we want to keep the null option instead of it being strictly a `string` then?

frontend/src/features/verifiable-fields/VerifiableFieldService.ts
LinHuiqing: Maybe we'd want to include this as a note in the PR? Though the impact of this is probably low (unless we cut the release at the same time as a huge event involving OTPs).

frontend/src/features/verifiable-fields/components/VerificationBox/VerificationBox.tsx
KenLSM: Right. `null` is rather redundant here since the parent component would have been required to handle this case. Will remove.

frontend/src/features/verifiable-fields/VerifiableFieldService.ts
KenLSM: Ah, makes sense. Forgot that this is still released together.

src/app/modules/verification/verification.controller.ts
```diff
@@ -283,23 +288,27 @@ export const _handleGenerateOtp: ControllerHandler<
         }
       })
       .andThen((form) =>
-        generateOtpWithHash(logMeta, SALT_ROUNDS)
-          .andThen(({ otp, hashedOtp }) =>
-            // Step 3: Send otp
+        generateOtpWithHash(logMeta, SALT_ROUNDS).andThen(
+          ({ otp, hashedOtp, otpPrefix }) =>
+            // Step 3: Send Otp
```
LinHuiqing, line 293: side note: why did you capitalise the O :joy:

LinHuiqing (approved):
lgtm!

src/app/modules/verification/verification.controller.ts
KenLSM: Ah you discovered my secret ploy to get more LOC changes 💀   /s. 

Not sure why I did that 😆

LinHuiqing: Wah okay I should start changing up all the comments slowly in my PRs too 👀 (jkjk)

Issue #5362, opened 2022-11-09
Remove `--legacy-peer-deps` from install step


justynoh:
As a result of upgrading to npm 8 in #5344, we had to instruct npm to use the old peer dependency resolution strategy. We should look into how we can resolve this and make our packages consistent with each other (if that's even at all possible).

PR #5876, opened 2023-03-05
test(e2e/email): add form auth tests

justynoh:
## Problem

When we initially migrated e2e tests to Playwright, mockpass was not compatible with our OIDC version. Thus, we omitted e2e tests for forms with authentication. Since mockpass has been updated, form authentication has been restored locally, and therefore we should restore these tests.

## Solution

Added tests for SP, CP, SGID and MyInfo. Small adjustment to be made for the myinfoable types in `field.ts`.
Had to upgrade node version for our CI to 18 to be compatible with mockpass.
Sorry about the lines changed metric, a large proportion of the changes made in `createForm` is just a giant refactor (to account for auth-ing the form with mockpass).

Separately, I got annoyed with having to type legacy-peer-deps every time so i added it to .npmrc and updated the README.


.npmrc
```diff
@@ -1 +1,2 @@
 unsafe-perm=true
+legacy-peer-deps=true
```
KenLSM, line 2: I was discussing this with @timotheeg on this, I think we can manual key in this flag as it is not supposed to be a permanent fixture in our development process.

The assumption is that `--legacy-peer-deps` is not required after angular is exterminated.

justynoh: That's fair, I think it's not necessary but mostly just a temporary DX thing. I'm ok to get rid of it, nbd haha.

For context though: I'm not confident that we can get rid of it along with angular, we may end up needing to do some dependency cleanup in a separate PR. The flag was introduced because a lot of angular packages violated the new resolution strategy, but other packages in backend/react also did. I don't know the extent of how easy it will be to upgrade everything so we can get rid of the flag, which is why I thought we might as well add it first, and then remove it later along with this ticket https://github.com/opengovsg/FormSG/issues/5362.

If you still feel we can leave it out from this PR I have no issues removing it!

KenLSM: I see, I think tracking this from another issue works too. I, too, am not sure if other packages are affected as well. Then it makes sense to keep this change as it is due to `npm 8` upgrade.

__tests__/e2e/utils/mail.ts
```diff
@@ -52,31 +52,22 @@ const MAIL_CLIENT = {
...
-    .filter((e) => e.to[0].address === toEmail)
-    .sort((a, b) => (a.time > b.time ? -1 : 1))
-}
-
-const getEmailsBySubject = (inbox: MailData[], subject: string): MailData[] => {
-  return inbox
-    .filter((e) => e.subject === subject)
-    .sort((a, b) => (a.time > b.time ? -1 : 1))
+const getEmailsBy = async (
+  filter: (email: MailData) => boolean,
```
KenLSM: Nit: `filter` is accepting a function instead of an object key or any other sensible types. Perhaps calling it `filterFn` or `filterCallbackFn` would help imply that.

__tests__/e2e/utils/mail.ts
```diff
@@ -132,3 +123,31 @@ const getSubmissionAttachments = async (
...
+  )
+
+  const lastEmail = emails.pop()
+  if (!lastEmail) {
+    throw Error(
+      `mailbox does not contain autoreply email for response ID ${responseId}`,
+    )
+  }
+
+  await MAIL_CLIENT.deleteById(lastEmail.id)
```
KenLSM, line 150: Just curious, why do we need to delete the email after retrieval?

__tests__/e2e/helpers/verifySubmission.ts
```diff
@@ -91,33 +161,33 @@ const TABLE_HANDLER = {
...
-    }
-    return title
-  } else {
-    if (isInJson) {
-      if (field.title.startsWith('[verified] ')) {
-        return field.title.substring(11)
-      } else if (field.title.startsWith('[MyInfo] ')) {
-        return field.title.substring(9)
-      }
+  } else if (field.fieldType === BasicField.Attachment) {
```
KenLSM: We could change this `else if` clause into `if` since the previous `if` returns.

__tests__/e2e/helpers/verifySubmission.ts
```diff
@@ -91,33 +161,33 @@ const TABLE_HANDLER = {
...
+    switch (formMode) {
+      case FormResponseMode.Email:
+        return `[attachment] ${field.title}`
+      case FormResponseMode.Encrypt:
+        return field.title
     }
-    return field.title
+  } else if (isVerifiableFieldType(field)) {
```
KenLSM: Same as the `else if` above.

__tests__/e2e/helpers/createForm.ts
```diff
@@ -614,3 +294,368 @@ const addCollaborators = async (
...
+
+  for (const logic of formLogics) {
+    // The 0th button called 'Add logic' is the sidebar tab nav button
+    await page.getByRole('button', { name: 'Add logic' }).nth(1).click()
+
+    // Add logic conditions
+    for (let i = 0; i < logic.conditions.length; i++) {
+      const { field, state, value } = logic.conditions[i]
+
+      if (i) await page.getByRole('button', { name: 'Add condition' }).click()
```
KenLSM: Nit: is `if (i)` trying to skip the first logic? Would prefer if this skip is more explicit through `if (i > 0)` or even `if (i !== 0)`.

__tests__/e2e/helpers/createForm.ts
```diff
@@ -553,30 +233,30 @@ const addAuthSettings = async (
...
       : formSettings.authType === FormAuthType.MyInfo
       ? ' with MyInfo'
       : ' (Corporate)'
   }`
 
   await page
     .locator('label', { has: page.getByRole('radio', { name }) })
-    .click()
+    .first() // Since 'Singpass' will match all radio options, pick the first matching one.
+    .click({ position: { x: 1, y: 1 } }) // Clicking the center of the sgid button launches the sgid contact form, put this here until we get rid of the link
```
KenLSM, line 245: 😆 how did you discover this worked?

KenLSM (approved):

__tests__/e2e/helpers/createForm.ts
justynoh: Haha, used debug mode which showed that it was clicking on the link in the middle of the button :')

__tests__/e2e/utils/mail.ts
justynoh: Haha this bug plagued me for a solid half a day :') 

There may be interference between OTP field verification emails and autoreply emails, since they're both sent to the same email address (i.e. the one that was entered in the email field). The way the tests are written, the OTP email getter simply pops the latest email from the inbox which was received by the recipient. If multiple tests use the same email field value, then an autoreply email may be picked up rather than an OTP email. 

It might have been possible to put a better filter in the extractOtp function to get emails with a particular subject, but it would also have been hacky because the email subject for admin logins and OTP verifications isn't the same. We also can't filter by regex because response IDs are in the body of autoreply emails, and may have 6 consecutive digits within their body. Anyway this seemed to work fine for the couple test runs I did, so I figured we can leave it as is for now. If it becomes flaky later on, we can look at improving it!

justynoh:
Also, if you're reading this, please know that all the force pushes are there because I was trying to debug tests that were working fine locally but not in CI. It turns out that mockpass used `fetch` which wasn't available out of the box in node 14 which our CI was running on previously 😠

wanlingt:
do we want to add this to branch protection rules? doing release now and I realise I don't have to wait for these tests to pass to merge into `release-al2`

PR #5881, opened 2023-03-06
feat: add local dev support for twilio

KenLSM:
## Problem

Testing SMS preview on `local` dev is only available if we push our code to `staging` and trigger an SMS. 
This can be frustrating as developers have to "book" the staging environment for the duration of the test.

Closes opengovsg/formsg-private#140

## Solution

Re-routing to [`stoplight/prism`](https://stoplight.io/open-source/prism) that reflects the APIs based on Twilio's [OpenAPI specification](https://github.com/twilio/twilio-oai/blob/main/spec/json/twilio_api_v2010.json).

**Breaking Changes** 

**Features**:

- Sending SMS on local dev will now return a proper response instead of BE crashing
- Sending SMS on local dev will send a copy of the message body over to `maildev`

## Before & After Screenshots

**BEFORE**:

**AFTER**:

post

## Tests

- [ ] SMS should not be intercepted if it is not in dev environment (dev = local dev)

## Deploy Notes

- `LOCAL_DEV`: Used by BE to determine if it is currently running on local development environment
- `TWILIO_ACCOUNT_SID` : Updated to follow specification format
- `TWILIO_MESSAGING_SERVICE_SID` : Updated to follow specification format

**New scripts**:

- `mocktwilio` : added new docker container that will provide proper API response

**New dependencies**:

- `"@stoplight/prism-cli": "^4.10.6"` : added to provide mocking support

**New dev dependencies**:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -172,7 +172,7 @@ services:
     depends_on:
       - backend
     network_mode: 'service:backend' # reuse backend service's network stack so that it can resolve localhost:4010 to prismtwilio:4010
-    command: mock -d https://raw.githubusercontent.com/twilio/twilio-oai/main/spec/json/twilio_api_v2010.json
+    command: mock https://raw.githubusercontent.com/twilio/twilio-oai/main/spec/json/twilio_api_v2010.json
 
 volumes:
   mongodb_data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -54,10 +54,10 @@ services:
       # https://github.com/opengovsg/formsg-javascript-sdk/blob/develop/src/resource/signing-keys.ts
       - SIGNING_SECRET_KEY=HDBXpu+2/gu10bLHpy8HjpN89xbA6boH9GwibPGJA8BOXmB+zOUpxCP33/S5p8vBWlPokC7gLR0ca8urVwfMUQ==
       # Mock Twilio credentials. SMSes do not work in dev environment.
-      - TWILIO_ACCOUNT_SID=ACmockTwilioAccountSid
+      - TWILIO_ACCOUNT_SID=AC00000000000000000000000000000000
       - TWILIO_API_KEY=mockTwilioApiKey
       - TWILIO_API_SECRET=mockTwilioApiSecret
-      - TWILIO_MESSAGING_SERVICE_SID=mockTwilioMsgSrvcSid
+      - TWILIO_MESSAGING_SERVICE_SID=MG00000000000000000000000000000000
       - SP_OIDC_NDI_DISCOVERY_ENDPOINT=http://localhost:5156/singpass/v2/.well-known/openid-configuration
       - SP_OIDC_NDI_JWKS_ENDPOINT=http://localhost:5156/singpass/v2/.well-known/keys
       - SP_OIDC_RP_CLIENT_ID=rpClientId
@@ -166,6 +166,14 @@ services:
      - STRIPE_API_KEY=secretKey
      - STRIPE_DEVICE_NAME=StripeWebhookListener
      
+  mocktwilio:
+    image: stoplight/prism:4
+    container_name: formsg-mocktwilio
+    depends_on:
+      - backend
+    network_mode: 'service:backend' # reuse backend service's network stack so that it can resolve localhost:4010 to prismtwilio:4010
+    command: mock -d https://raw.githubusercontent.com/twilio/twilio-oai/main/spec/json/twilio_api_v2010.json
+
 volumes:
   mongodb_data:
     driver: local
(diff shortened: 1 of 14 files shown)


docker-compose.yml
```diff
@@ -54,10 +54,10 @@ services:
       # https://github.com/opengovsg/formsg-javascript-sdk/blob/develop/src/resource/signing-keys.ts
       - SIGNING_SECRET_KEY=HDBXpu+2/gu10bLHpy8HjpN89xbA6boH9GwibPGJA8BOXmB+zOUpxCP33/S5p8vBWlPokC7gLR0ca8urVwfMUQ==
       # Mock Twilio credentials. SMSes do not work in dev environment.
-      - TWILIO_ACCOUNT_SID=ACmockTwilioAccountSid
+      - TWILIO_ACCOUNT_SID=AC00000000000000000000000000000000
```
KenLSM, line 57: Needed to update this as the validation server expects `"pattern": "^AC[0-9a-fA-F]{32}$"`.

docker-compose.yml
```diff
@@ -54,10 +54,10 @@ services:
...
       - TWILIO_API_KEY=mockTwilioApiKey
       - TWILIO_API_SECRET=mockTwilioApiSecret
-      - TWILIO_MESSAGING_SERVICE_SID=mockTwilioMsgSrvcSid
+      - TWILIO_MESSAGING_SERVICE_SID=MG00000000000000000000000000000000
```
KenLSM, line 60: Needed to update this as the validation server expects `"pattern": "^MG[0-9a-fA-F]{32}$"`.

src/app/config/config.ts
```diff
@@ -40,6 +40,7 @@ const isDev =
   basicVars.core.nodeEnv === Environment.Dev ||
   basicVars.core.nodeEnv === Environment.Test
 const nodeEnv = isDev ? basicVars.core.nodeEnv : Environment.Prod
+const localDev = basicVars.core.localDev
```
KenLSM: Need another eyes on checking if this is the best way to determine if executed environment is on local development server.

src/app/services/mail/mail.service.ts
```diff
@@ -879,6 +879,22 @@ export class MailService {
...
+      from: this.#senderFromString,
+      html: smsMessage,
+      subject: '[mocktwilio] SMS OTP Verification',
+      replyTo: this.#officialMail,
+      bcc: this.#senderMail,
+    }
+    return this.#sendNodeMail(mailOptions, {
+      mailId: 'sendWarningMailForAdmin',
+    })
+  }
```
timotheeg: Hmm, this function does something more specific than its name suggests. `subject` should probably be an input arg instead of hardcoded.

docker-compose.yml
```diff
@@ -148,6 +148,14 @@ services:
     ports:
       - '1080:1080'
 
+  mocktwilio:
```
timotheeg, line 169: Just checking: why does this definition not expose the port `4010` explicitly?

src/app/config/config.ts
```diff
@@ -40,6 +40,7 @@ const isDev =
   basicVars.core.nodeEnv === Environment.Dev ||
   basicVars.core.nodeEnv === Environment.Test
 const nodeEnv = isDev ? basicVars.core.nodeEnv : Environment.Prod
+const isLocalDev = basicVars.core.localDev
```
timotheeg: Could we have leveraged `basicVars.core.nodeEnv === Environment.Dev` instead of introducing a new `isLocalDev` variable? 

Hmm, actually not sure 🤔 . You could have a dev environment testing real twilio crendentials too, so they're different concerns. But then again, the same applies to `isLocalDev`, so maybe we need something very specific like `useMockTwilio` 🤔

timotheeg (commented):
Good work Ken! 

Just some minor questions from me.

docker-compose.yml
KenLSM: I assumed that we didn't need to call/interact with `mocktwilio` container. Only `backend` needs to talk to `mocktwilio`.

I suppose we can open the port `4010`  and allow sending API calls to it, but I couldn't think of a use case currently.

src/app/config/config.ts
KenLSM: Yeah I think `useMockTwilio` makes sense too. I wasn't exactly sure what other common use cases we have (i.e., all test env SMS OTPs should go to email), I'll ask tomorrow in sync.

KenLSM (commented):
I've forgotten to hit the review change button on GH apologies if the comments looked out of order. They were added before the review.

src/app/services/mail/mail.service.ts
KenLSM: Agreed. Added subject as required params.

timotheeg (approved):
LGTM!

PR #5882, opened 2023-03-06
chore: update package.json to loosen up engine version checks

KenLSM:
## Problem

Resolving warnings that is clouding the output logs. 

We should keep our logs as clean as possible if there are warnings/errors and take actions to resolve it if we see them. It should not be treated as a norm and get desensitized.

## Solution

Reducing the specification of node and npm. We don't actually require `node` and `npm` to be strictly  `"18.12.x"` and `"8.19.x"`

The remaining warning from openid-client can be addressed by getting [opengovsg/sgid-client](https://github.com/opengovsg/sgid-client) to update their `openid-client` deps to [`5.1.10`](https://github.com/panva/node-openid-client/commit/9aefba30dcf0e312051e6844b35b06bc457488d5) in another PR.

```
$ npm why openid-client
openid-client@5.1.5
node_modules/@opengovsg/sgid-client/node_modules/openid-client
  openid-client@"5.1.5" from @opengovsg/sgid-client@1.0.4
  node_modules/@opengovsg/sgid-client
    @opengovsg/sgid-client@"1.0.4" from the root project

openid-client@5.3.1
node_modules/openid-client
  openid-client@"^5.3.1" from the root project
```

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-03-06 at 4 50 02 PM

**AFTER**:

Screenshot 2023-03-06 at 5 24 21 PM

## Tests

- [ ] installing packages should not contain `npm` + `node` version warning.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -236,8 +236,8 @@
         "worker-loader": "^2.0.0"
       },
       "engines": {
-        "node": "~18.12.0",
-        "npm": "~8.19.2"
+        "node": "18",
+        "npm": "8"
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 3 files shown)


justynoh (approved):
thanks! lgtm

PR #5909, opened 2023-03-09
fix: bump downshift from 6.1.12 to 7.3.2, a11y fixes for single select and multiselect

### LINKED ISSUE #5502: a11y: Auto-set focus to be on the first option of a dropdown menu (iOS, Android, Mac)
**Problem**
When users encounter the combobox, it reads out as an edit box which is correct but is potentially confusing as a combobox is a cross between editbox and dropdown. When users Enter in editbox, the focus either jumps to the bottom of the page or does nothing, depending on the device.

**Related issue**
- On iOS Voiceover, when the dropdown opens if a user swipes it will lead them to the X button instead of going down the list

**Proposal**
- If users press Enter or Space, open the dropdown list and focus on the first item in the group.

### LINKED ISSUE #5504: a11y: Table field navigation
- On iOS Voiceover, when in the first column, it reads out the entire question again
- on MacOS Voiceover, when entering the table field, focus jumps to the second column
- on NVDA, when attempting to open the combobox, focus jumps to bottom of page
- Can we change Comboboxes to normal dropdown fields to avoid problems mentioned in #5502 and above

justynoh:
## Problem
There are various bugs in dropdown a11y. This PR rectifies them and makes the dropdown screen reader experience better.

Closes #5502
Closes #5504 

## Solution

[Replicated a PR from design system](https://github.com/opengovsg/design-system/pull/252) that fixes dropdown a11y, this should fix dropdown fields and dropdowns in tables.

**Breaking Changes** 

## Tests

This PR has been tested by design system, but tagging @staceytan1998 for assistance on further testing! Can be done on chromatic or pushed to staging to test within a form field setting.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -29,7 +29,7 @@
         "dayjs": "^1.10.7",
         "dayzed": "^3.2.2",
         "dedent": "^0.7.0",
-        "downshift": "^6.1.7",
+        "downshift": "^7.3.2",
         "file-saver": "^2.0.5",
         "focus-visible": "^5.2.0",
         "framer-motion": "^5.4.5",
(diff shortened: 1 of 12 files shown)


justynoh:
Closing this PR temporarily as stacey is away

justynoh:
reopening for review!

timotheeg (approved):
LGTM

PR #5911, opened 2023-03-09
feat: add unlisted agency issued UENs list

KenLSM:
## Problem

We discovered that are instance of agencies publishing their own Entity Type Indicators which are not listed under [ARCA's UEN list](https://www.uen.gov.sg/ueninternet/faces/pages/admin/aboutUEN.jspx)

## Solution

Add our own list of known Entity Type Indicators, which currently contains `["EC"]`.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-03-09 at 11 06 03 AM

**AFTER**:

Screenshot 2023-03-09 at 10 58 34 AM

## Tests

- [ ] In the UEN Field, inputing `S68EC0001C` should be valid
- [ ] In the UEN Field, inputing `S68ED0001C` should not be valid

### DIFF
--- a/shared/utils/uen-validation.ts
+++ b/shared/utils/uen-validation.ts
@@ -2,7 +2,7 @@
  * Validate entity-type indicators, as per
  * https://www.uen.gov.sg/ueninternet/faces/pages/admin/aboutUEN.jspx
  */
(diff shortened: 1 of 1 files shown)


timotheeg (approved):
Very nice implementation!

LGTM

PR #5884, opened 2023-03-07
chore(deps): upgrade react client dependencies

justynoh:
Closed as an attempt, but here are the basic ideas as a reminder later on.

1. `react-scripts` 4.0.3 -> 5.0.1, `craco` 6.4.3 -> 7.0.0, `eslint` 7.32.0 -> 8.35.0 to be upgraded together
2. `html-webpack-plugin` 4.5.2 -> 5.5.0, add `webpack@5.75.0` as a dev-dep
    a. Polyfill required for csv-string dependency, add it to `craco.config.js` and install `stream-browserify` as a dep
    b. install `postcss8` as a dev-dep
    c. will need to make typing changes to various files (mostly convert from nullish coalescing to optional chaining) 
6. Can delete frontend/__tests__ folder

Storybook:
After all of the above
1. Run `npx sb upgrade` 
2. Update storybook config with `core: { builder: 'webpack5' }` and new webpack configs (including injecting polyfill)

Issue #5721, opened 2023-01-27
Use `@opengovsg/design-system-react` dependency


karrui:
The time has come... Should we use the design system package and remove all current duplicate design system components in FormSG?

darrelhong:
design system is dependent on chakra 2 and react 18, does form need to update before being compatible?

justynoh:
attempted but not worth to focus on now. see #5884

PR #5424, opened 2022-11-19
chore: cra5

darrelhong:
react-scripts 4.0.3 -> 5.0.1
craco 6.4.3 -> 7.0.0 with react-scripts 5 support
eslint 7.32.0 -> 8.28.0 to match react-scripts
manually add postcss 8 to fix prod build errors
update storybook with  `npx sb upgrade`
update storybook config manually as craco preset doesn't seem to work anymore

PR #5918, opened 2023-03-12
feat: react 18, cra 5, chakra 2

darrelhong:
## Problem
Since there are plans to use @opengovsg/design-system-react (#5721) in the future, current react app needs to be updated to as Chakra UI 2 requires React 18 as well. It was also necessary to update other dependencies as well

Related issues #5884 #5424  #5450

## Solution
Will go through the steps and reasons for each upgrade by commit
- replaced react-beautiful-dnd with drop-in @hello-pangea/dnd. Original library does not support React 18 and is not maintained anymore, new library seems to have active community around it
- Updated to react 18 and chakra 2
- prettier problems appeared after update, used pretter --write to fix
- chakra 2 did not seem to be compatible with CRA 4, therefore, worked on top of my previous MR in upgrading to CRA5, details as follows
- react-scripts 4.0.3 -> 5.0.1
- craco 6.4.3 -> 7.0.0 with react-scripts 5 support
- eslint 7.32.0 -> 8.28.0 to match react-scripts
- manually add postcss 8 to fix prod build errors
- update storybook with npx sb upgrade
- update storybook config manually as craco preset doesn't seem to work anymore
- remove all references to React.FC and it does not support children by default, replaced with PropsWithChildren helper type
- remove isFullWidth as is not supported by chakra anymore, replaced with width: 'full'.
- update to tanstack query v4 as it natively supports react 18, v3 also had typescript issues due to React.FC children prop
- used codemods from migration guide to update imports, followed by prettier --write and eslint --fix
- With regards to other breaking changes, I believe it mostly does not affect this app as mainly the basic features of react-query are used. the changes that may affect functionality might be relating to [disabled queries](https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#disabled-queries) and [setQueryData](https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#onsuccess-is-no-longer-called-from-setquerydata)
- update to framer-motion 10 as it also had issues with React.FC children prop
- fix various typescript error that occurs due to changes in Chakra hooks etc, this mainly involved typecasting to the correct type.
- there is one outstanding issue with react-table and useTable columns field which I'm not sure what is the correct fix. it appears to be related to react 18 types as well, as downgrading react type definitions to 17 fixed. Currently disabled typescript checks on it.
- update react-joyride as previous did not support react 18
- fix storybook to no longer use craco plugin but used cra plugin instead
- dont override colors.white as it use by chakra components.
- fix menu button width issues
- update html-webpack-plugin to 5.5.0 to match webpack 5, added webpack-cli 5 to fix build step

**Improvements**:
All the benefits of webpack 5 and React 18

## Deploy Notes
npm build, start and storybook commands working

Most of the key features seem to be working and have already fixed some UI bugs, however might need to have a conduct a detailed regression as multiple key dependencies have been updated.

**New dependencies**:

`@chakra-ui/react` : 1.8.6 -> 2.5.1
`@chakra-ui/cli`: 1.9.0 -> 2.3.0
`@emotion/react`: 11.7.0 -> 11.10.6
`@emotion/styled`: 11.6.0 -> 11.10.6
`react-beautiful-dnd`: removed
`@hello-pangea/dnd`: 16.2.0
`@tanstack/react-query`: 4.26.1
`@tanstack/react-query-devtools`: 4.26.1
`react`: 17.0.2 -> 18.2.0
`react-dom`: 17.0.2 -> 18.2.0
`@types/react`: 18.0.28
`@types/react-dom`: 18.0.11
`framer-motion`: 5.4.5 -> 10.2.4
`react-joyride`: 2.4.0 -> 2.5.3
`stream-browserify`: 3.0.0
`@craco/craco`: 6.4.3 -> 7.1.0
storybook and plugins: 6.5.5 -> 6.5.13
`eslint`: 7.32.0 -> 8.36.0
`postcss`: 8.4.21
`webpack`: 5.76.1
`html-webpack-plugin`: 4.5.2 -> 5.5.0

cc @justynoh @karrui @KenLSM @timotheeg @mantariksh


timotheeg:
Hey @darrelhong , thanks for another contribution! 

Apologies for leaving several of your PRs dangling. We have been rather busy with commitments and deadlines, but we should be resuming our feature delivery soon, and we'll be relooking at open PRs in the next few days/weeks.

On this PR specifically, first of all, it's huge (!), so big thank you for taking the time and effort to work on it!

That being said, I need to give you some feedback here. PRs are not free. Just as it takes you lots of time to write the PR, it takes us a lot of time and effort to go through PRs reviews, assess quality and completeness, and test. Large unsolicited PRs are typically problematic because 1) the team may have plans to tackle an issue in some specific way; 2) some folks may have already started on implementation, potentially causing duplication of work; and 3) It's hard for us to test exhaustively all the changes (so if we have other priorities, we may not be able to attend to the PR at all)

In this case, we had a plan to remove the Angular UI, and all the dev dependencies it relies on first, and then upgrade whatever is left for React and Chakra. Unfortunately, we are not quite ready to remove angular from the code base entirely atm because of one lingering issue with Axios, and so we're likely to ignore this PR for now.

Secondly, we typically mark issues we hope to receive contributions on with a label `contribute`. It would be much better to pick those rather than jumping on random issues. And for issues that looks large, unclear, or complicated, it would be best to contact the team and have a conversation before starting work, to check if there are more details or ideas on a potential approach that have been discussed by the team, but that have not been documented in the issue.

Lastly, please do not tag folks explicitly in contribution PRs. For one thing notifications are disruptive, and based on our schedules and priorities, we can make no commitments on response time. We will strive to reply to contribution PRs in a timely manner, but we can make no promises. For a second thing, the team is fluid, with people leaving and joining on somewhat regular basis. Antariksh and KarRui have recently left FormSG, so they should not be disturbed for FormSG matters.

I don't mean to sound too negative. Doing the upgrades you did here are something we must do in the not too distant future, and so your effort are still very much appreciated! But again, we need to establish some line of communication so we can collaborate effectively. We'll be discussing this internally, and hopefully we'll have some guidelines to share soon!

darrelhong:
understand the context much better now, please excuse me for the tagging as I'm mostly left hanging. It would be good to mention if MRs are being looked into or not. From my side I did not and will not know the situation of the team. Previously, I had MRs that were accepted so I continued to try new things. 

I submitted this MRs as I saw there was a few attempts to do the same before, so it is just meant as a experiment to show how it could be achieved, unfortunately I wasn't aware of where it was on the priority. I believe most of my other PRs are also tagged to existing issues. Thanks for the reply, and once again sorry for the disruption.

PR #5921, opened 2023-03-13
refactor: simplify payment amount regex

LinHuiqing:
## Problem

Payment amount regex was more complicated as it mandated a non-zero positive amount to be entered. However, we have since decided to use `useForm`'s validation rules to mandate the minimum and maximum amounts for cleaner implementation.

## Solution

Simplified the regex to limit input to 2 decimal places.

**Breaking Changes** 

**Improvements**:

- Disabled steppers for payment amount input field in payment drawer

## Tests

- [ ] Any of the following valid payment amounts (`x` where `0.50 <= x <= 100` should be accepted by the payment amount field and saved successfully thereafter:
  - [ ] Whole numbers.
  - [ ] Whole numbers followed by `.`.
  - [ ] Numbers with 1 decimal place.
  - [ ] Numbers with 2 decimal places.
  - [ ] Any input other than what is defined above (i.e. valid payment amount) should be rejected.
- [ ] When in the payment amount input field, pressing the arrow keys should not change the value in the field.

### DIFF
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -201,6 +201,7 @@ export const PaymentInput = (): JSX.Element => {
                 render={({ field }) => (
                   <MoneyInput
                     flex={1}
+                    step={0}
                     inputMode="decimal"
                     placeholder="0.00"
                     {...field}
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -142,11 +142,9 @@ export const PaymentInput = (): JSX.Element => {
             /* Regex allows: 
                - leading and trailing spaces
                - max 2dp
-               - strictly positive (>0)
             */
-            /^\s*0*([1-9]\d*(\.\d{0,2})?|0\.(0[1-9]|[1-9]\d?))\s*$/.test(
-              val ?? '',
-            ) || 'Please enter a valid payment amount'
+            /^\s*\d*\.?\d{0,2}\s*$/.test(val ?? '') ||
+            'Please enter a valid payment amount'
           )
         },
         validateMin: (val) => {
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -143,7 +143,7 @@ export const PaymentInput = (): JSX.Element => {
                - leading and trailing spaces
                - max 2dp
             */
-            /^\s*\d*\.?\d{0,2}\s*$/.test(val ?? '') ||
+            /^\s*(\d+)(\.\d{0,2})?\s*$/.test(val ?? '') ||
             'Please enter a valid payment amount'
           )
         },


frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -142,11 +142,9 @@ export const PaymentInput = (): JSX.Element => {
             /* Regex allows: 
                - leading and trailing spaces
                - max 2dp
-               - strictly positive (>0)
             */
-            /^\s*0*([1-9]\d*(\.\d{0,2})?|0\.(0[1-9]|[1-9]\d?))\s*$/.test(
-              val ?? '',
-            ) || 'Please enter a valid payment amount'
+            /^\s*\d*\.?\d{0,2}\s*$/.test(val ?? '') ||
```
timotheeg: Hmm, simplifying yes, but don't compromise on readability and correctness 😅 . Consecutive optional tokens in a regexp are "dangerous" because they could all not match, and the validation would still pass. We need to check strictly for what a valid input is.

In this case, the regexp so far allows an empty input, or just whitespace input, or just a dot 😅 .

Think of it this way, and write the regexp to match:
* Any sequence of digits (including zero since the min/max validator will protect against that)
* Optionally decimal places up to 2 digits (including zeros, since again the min/max validator protect us against zero). 

So I'd recommend this:

```suggestion
            /^\s*\d+(\.\d{0,2})?\s*$/.test(val ?? '') ||
```

timotheeg: Hm, maybe I can clarify further here:

I could have made less changes to your regexp like so:

```javascript
/^\s*\d+\.?\d{0,2}\s*$/
```
That guarantees we would no longer allow empty input, just a dot, etc. That works, but it can confusing to see why. Basically it works because when no dot is injected in the input, `\d+\d{0,2}` is the same as `\d+`.

For readability, it's typically better to match groups as humans expect them, so it's easier to reason about. In this case:
* 1 compulsory unit part
* 1 optional decimal part (with max 2-digits constraints)

By the way, in this case, the fact that we allow trailing dots with no digits is debatable, it could be even clearer to write this:

```javascript
/^\s*(\d+)(\.\d{1,2})?\s*$/
```
That way the decimal part must be at least a dot and a digit (up to 2 digits), and the decimal part itself is optional.

(And sorry if all this is too obvious btw 😅)

LinHuiqing: Ah, thanks for the correction and explanations! +xp on regex :) 
As for allowing the trailing dot, I think my original intention for allowing it is that the error message will pop up as the admins type in the amount, which might be quite jarring, as seen in the video:

https://user-images.githubusercontent.com/37061143/224887197-a6804c56-e7bd-42e7-923b-1a93c1a1b829.mov

And we will be able to parse the number if it has a trailing dot anyway, so I thought we should allow it for better user experience.

timotheeg: Understood 👌 

I'm thinking that another way to manage this kind of situations could be to have the error message shown with  some debounce delay, so it allows a little more time for the admin to enter the decimal part. And if they do not, and the error eventually pops up, it's kind of obvious why the input is considered incorrect.

But anyway, that's not a priority! It's fine to keep it as `{0,2}` for now. Do update the regexp with the `\d+`, and the group for the decimal part, and we can merge the PR.

LinHuiqing: Ooh for debounce do you mean by time? I.e. only when the trailing dot is present in the field after a certain amount of time, then the error should pop up.
Also, done!

timotheeg: Thanks! I meant debounce by time yes. There'd be no need to check for the presence of dot specifically. Just adding some debounce delay on keyup/change events, to allow the user a little more time to complete his/her input before showing whatever validation error message they hit.

timotheeg (approved):

PR #5926, opened 2023-03-14
test(e2e): add storage mode e2e tests

### LINKED ISSUE #5113: write initial e2e tests for react
As a first cut, should be more or less parity with current e2e tests in what they cover. may not need to be that detailed.

- [x] login e2e tests
- [x] form creation
- [x] form duplication
- [x] form field updating
- [x] email mode form filling (and submission)
- [ ] storage mode form filling (and submission)
- [x] auth'd form filling (and submission)
- [ ] storage mode responses downloading

justynoh:
## Problem
We don't have e2e tests for storage mode yet! This PR adds them and achieves parity with Angular.

Closes #5113

## Solution
Shard out some of the test case data that is shared between the two modes.
Add support for storage mode in all relevant handlers. Specifically, changes needed to be made to createForm and verifySubmission to accommodate storage mode individual response page and CSV modes which each displayed data differently (help...). Some utils had to be modified as well in the process.


justynoh:
Closing and reopening to get CI to run

justynoh:
snyk pls 🥺

__tests__/e2e/constants/field.ts
```diff
@@ -166,15 +166,15 @@ export const ALL_FIELDS: E2eFieldMetadata[] = [
   {
     title: 'Birthday',
     fieldType: BasicField.Date,
-    val: format(new Date(), DATE_INPUT_FORMAT),
+    val: format(new Date(952970366), DATE_INPUT_FORMAT),
```
KenLSM, line 169: Is this date value `952970366` special?

justynoh: image
:)

justynoh: ok no but on a serious note, I couldn't leave it as the current date because our Storage mode's individual response page also displays the date of submission which would also be the current date. So the `expect()` for finding the response to this question would certainly always succeed, regardless of whether the response to the question was shown.

__tests__/e2e/utils/response.ts
```diff
@@ -0,0 +1,170 @@
...
+              : `Others: ${selected}`
+          })
+          .join(isCsv ? ';' : ', '),
+      ]
+    case BasicField.Date:
+      // Need to re-parse, because input is in dd/mm/yyyy format whereas response is in dd MMM yyyy format.
+      return [
+        field.val
+          ? format(
+              parse(field.val, DATE_INPUT_FORMAT, new Date()),
```
KenLSM, line 128: Is this date supposed to be matching `952970366`?

justynoh: it doesn't have to! `field.val` will be equal to `14/03/2000` since it was formatted. This parse and re-format converts the `field.val` to the `14 Mar 2000` format. This is necessary since we display dates on the public form (in dd/mm/yyyy format) differently from the responses that admins see (in dd MMM yyyy format).

Of course, since field.val contains no information about the time, the time data from `952970366` will be lost. But it's not an issue since we never use the time information anyway

KenLSM: Got it! Thanks a lot for the explanation @justynoh!

KenLSM (approved):
LGTM!

PR #389, opened 2020-09-30
feat: soft-launch rate-limiting of API endpoints

mantariksh:
## Problem

We want to limit the rate of requests to public endpoints as a security measure against DDoS attacks. However, we need to ensure that the limits do not affect regular application usage. Hence it is necessary to ensure that the limits do not get exceeded on a regular basis, before actually denying requests which exceed the limit.

After some discussion, we decided to apply the limits to three endpoints: `/auth/sendotp` as well as both submissions endpoints, `v2/submissions/email` and `v2/submissions/encrypt`.

## Solution

Use `express-rate-limit` with a custom handler which logs a message if the rate limit is exceeded, then carries on handling the request as normal. Note that the package uses per-IP limits by default.

The log message looks like this:
Screenshot 2020-10-06 at 10 35 24 AM

### Determining the limit
It is important to apply a limit which does not affect regular application usage. To determine the limit, I counted the per-minute requests per IP address over the past week to each of the endpoints.

Between both `v2/submissions` endpoints, the maximum was 46 requests to `/v2/submissions/email`:
Screenshot 2020-10-05 at 5 27 49 PM

Multiplying this by 4 to get the maximum seemed reasonable, since programmatic attacks are likely to exceed regular usage numbers by far more. Moreover, the limit is applied per-instance rather than across instances (see next section). Hence I chose a per-minute limit of 200.

As for the `/auth/sendotp` endpoint, the maximum was 4 requests in a minute. To be safe, I set the per-minute maximum at 60.

### Per-instance vs instance-wide limits
`express-rate-limit` uses an in-memory store by default. This means that the rate limits are applied per-instance. Applying them instance-wide would require setting up a store in our database using [MongoDB plugin](https://www.npmjs.com/package/rate-limit-mongo). However, the plugin does not seem to support reusing existing connections, which means we would have to create a new connection to the database. I decided against this to avoid concurrency issues with multiple instances accessing the same documents, so the limits are applied per-instance.

## Tests
- [ ] Production environment is updated with the environment variables `SUBMISSIONS_RATE_LIMIT=200` and `SEND_AUTH_OTP_RATE_LIMIT=60`.
- [ ] In production, an alarm has been set for any occurrence of the "Rate limit exceeded" log message.
- In the staging environment, set `SUBMISSIONS_RATE_LIMIT=2` and `SEND_AUTH_OTP_RATE_LIMIT=2`.
[...]

PR #5927, opened 2023-03-14
feat: add rate limiting for receipt download

### LINKED ISSUE #5924: rate limit the downloadPaymentReceipt endpoint
We should add rate limits to the route GET `/receipt/:formId/:submissionId/download` in `payments.routes.ts`

Rationale for rate limit:
* endpoint does external calls to stripe, so we want to make sure we can't be used as a source of abuse
* pdf generation is heavy, so we should protect our capacity

wanlingt:
## Context

This PR adds rate limits to the payment receipt download endpoint, because:

- endpoint does external calls to stripe, so we want to make sure we can't be used as a source of abuse
- pdf generation is heavy, so we should protect our capacity

Closes #5924

## Solution

Add a new config `rateLimit.downloadPaymentReceipt` - the default is set at 10

## Tests

- [ ] Production environment is updated with the environment variables `DOWNLOAD_PAYMENT_RECEIPT_RATE_LIMIT=10` (any other thoughts on a recommended number?)
- (FYI from #389) In production, an alarm has been set for any occurrence of the "Rate limit exceeded" log message.
- In the staging environment, set `DOWNLOAD_PAYMENT_RECEIPT_RATE_LIMIT=2`
    - [x] On the payment thank you page, try to download the payment receipt as many times as you can within 60s. Ensure that you are able to download the receipt only twice, and that a "Rate limit exceeded" message gets logged with metadata about the rate limit (IP address, endpoint, current number of requests).
    - [x] Set `DOWNLOAD_PAYMENT_RECEIPT_RATE_LIMIT=10`

## Deploy Notes

- `DOWNLOAD_PAYMENT_RECEIPT_RATE_LIMIT` : Per-minute, per-IP request limit to download the payment receipt from Stripe

### DIFF
--- a/src/app/routes/api/v3/payments/payments.routes.ts
+++ b/src/app/routes/api/v3/payments/payments.routes.ts
@@ -1,6 +1,8 @@
 import { Router } from 'express'
 
+import { rateLimitConfig } from '../../../../config/config'
 import * as StripeController from '../../../../modules/payments/stripe.controller'
+import { limitRate } from '../../../../utils/limit-rate'
 
 export const PaymentsRouter = Router()
 
@@ -26,7 +28,10 @@ PaymentsRouter.route(
 // TODO: consider rate limiting this endpoint #5924
 PaymentsRouter.route(
   '/receipt/:formId([a-fA-F0-9]{24})/:submissionId([a-fA-F0-9]{24})/download',
-).get(StripeController.downloadPaymentReceipt)
+).get(
+  limitRate({ max: rateLimitConfig.downloadPaymentReceipt }),
+  StripeController.downloadPaymentReceipt,
+)
 
 PaymentsRouter.route('/stripe/callback').get(
   StripeController.handleConnectOauthCallback,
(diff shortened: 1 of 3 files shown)


src/app/routes/api/v3/payments/payments.routes.ts
```diff
@@ -26,7 +28,10 @@ PaymentsRouter.route(
 // TODO: consider rate limiting this endpoint #5924
 PaymentsRouter.route(
   '/receipt/:formId([a-fA-F0-9]{24})/:submissionId([a-fA-F0-9]{24})/download',
-).get(StripeController.downloadPaymentReceipt)
+).get(
+  limitRate({ max: rateLimitConfig.downloadPaymentReceipt }),
+  StripeController.downloadPaymentReceipt,
```
KenLSM, line 33: Is this rate limit differentiated by different `submissionId`, `formId`?

I.e., 
```
receipt/someformid/somesubmissionid/download
receipt/someformid/somesubmissionid2/download
receipt/someformid/somesubmissionid2/download // 1. Blocked here as overall req to this route has exceeded limit
receipt/someformid/somesubmissionid2/download // 2. Or blocked here as uniq req to this route has exceeded limit
```

wanlingt: yup it's differentiated by `submissionId`, as `submissionId` is in the route `'/receipt/:formId([a-fA-F0-9]{24})/:submissionId([a-fA-F0-9]{24})/download'`

so that would be case 2. in your example, if the limit is 2

KenLSM (approved):

PR #5935, opened 2023-03-15
feat(db): add pending submissions collection

justynoh:
## Problem
We want a pending submissions collection. We don't have a pending submissions collection.

## Solution
We make a pending submissions collection.

This is done via sharing schemas with submissions which gives us all the benefits of the statics + type guarantees of the submissions model a la [this suggestion](https://stackoverflow.com/questions/13337155/mongoose-same-schema-for-different-collections-in-mongodb).

`paymentId` is added to the storage mode submissions only.

The old implementation of pending submissions (with the pending flag in the submissions collection) is removed.

**Breaking Changes** 

### DIFF
--- a/src/app/models/pending_submission.server.model.ts
+++ b/src/app/models/pending_submission.server.model.ts
@@ -16,8 +16,8 @@ import {
 } from './submission.server.model'
 
 export const PENDING_SUBMISSION_SCHEMA_ID = 'PendingSubmission'
-export const EMAIL_PENDING_SUBMISSION_SCHEMA_ID = 'emailPendingSubmission'
-export const ENCRYPT_PENDING_SUBMISSION_SCHEMA_ID = 'encryptPendingSubmission'
+const EMAIL_PENDING_SUBMISSION_SCHEMA_ID = 'emailPendingSubmission'
+const ENCRYPT_PENDING_SUBMISSION_SCHEMA_ID = 'encryptPendingSubmission'
 
 const compilePendingSubmissionModel = (db: Mongoose): ISubmissionModel => {
   const PendingSubmission = db.model<ISubmissionSchema, ISubmissionModel>(
--- a/src/app/models/pending_submission.server.model.ts
+++ b/src/app/models/pending_submission.server.model.ts
@@ -16,8 +16,8 @@ import {
 } from './submission.server.model'
 
 export const PENDING_SUBMISSION_SCHEMA_ID = 'PendingSubmission'
-const EMAIL_PENDING_SUBMISSION_SCHEMA_ID = 'emailPendingSubmission'
-const ENCRYPT_PENDING_SUBMISSION_SCHEMA_ID = 'encryptPendingSubmission'
+const EMAIL_PENDING_SUBMISSION_SCHEMA_ID = 'EmailPendingSubmission'
+const ENCRYPT_PENDING_SUBMISSION_SCHEMA_ID = 'EncryptPendingSubmission'
 
 const compilePendingSubmissionModel = (db: Mongoose): ISubmissionModel => {
   const PendingSubmission = db.model<ISubmissionSchema, ISubmissionModel>(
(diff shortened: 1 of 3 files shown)


justynoh:
## Design notes

Some conscious choices had to be made about what level of identicality we should be enforcing here. 

(Option 1) In particular I decided not to have an additional `PendingSubmission` shared type, and decided only to create a pending submission collection using the same `Submission` types. In fact, I made it such that the pending submissions collection is completely identical to the submissions collection, so we get all the "extras" (indexes, statics, methods) already defined on the submissions collection. 

What are the alternatives to this approach? 

(Option 2) One way is to only enforce the identicality at the base type level, but create two separate schemas for it. But this was clearly a somewhat strange way to do it, since all the fields are shared anyway. In that case, there would be no point repeating the code to define the schema. You lose all the "extras" for more work, so that's no good.

(Option 3) Another way is to completely redefine them from the types up. So something like a shared type `type PendingSubmission = Submission`, and then redefine the schemas and models (schemas could potentially be shared, since the types are the same). 

I didn't really have a good answer as to why Option 1 or the Option 3 would be better. My thought is that a type should really represent a "type of object", so we should not create a new type for pending submissions, since they're... always supposed to be the same thing. Enforcing that the shape of pending submissions and submissions is the same at the type level feels the safest, since we always want this guarantee that pending submissions should be movable in to submissions anytime. If there are any opinions on the design choices of this PR, please feel free to lmk!

src/app/models/submission.server.model.ts
```diff
@@ -182,6 +178,10 @@ const EncryptSubmissionSchema = new Schema<
     type: Number,
     required: true,
   },
+  paymentId: {
+    type: Schema.Types.ObjectId,
+    ref: 'Payment',
```
wanlingt: use `PAYMENT_SCHEMA_ID` instead?

src/app/models/pending_submission.server.model.ts
```diff
@@ -0,0 +1,40 @@
...
+  )
+  PendingSubmission.discriminator(SubmissionType.Email, EmailSubmissionSchema)
+  PendingSubmission.discriminator(
+    SubmissionType.Encrypt,
+    EncryptSubmissionSchema,
+  )
+  return db.model<ISubmissionSchema, ISubmissionModel>(
+    PENDING_SUBMISSION_SCHEMA_ID,
+    SubmissionSchema,
+  )
```
timotheeg: I know this copies the code from `compileSubmissionModel()`, but why are we returning this call instead of `PendingSubmission`?

src/app/models/submission.server.model.ts
```diff
@@ -389,7 +389,7 @@ EncryptSubmissionSchema.statics.findEncryptedSubmissionById = function (
 
 const compileSubmissionModel = (db: Mongoose): ISubmissionModel => {
   const Submission = db.model<ISubmissionSchema, ISubmissionModel>(
-    'Submission',
+    SUBMISSION_SCHEMA_ID,
```
timotheeg, line 393: 👏 👍

src/app/models/submission.server.model.ts
justynoh: omg ur right, thanks!

timotheeg (approved):

timotheeg:
On your choice for option 1, I think you did the right thing. Pending submissions are exactly submissions, just they are not complete yet. Keeping the exact same schema and not diverging will help. 

There is still a sliver of a chance we go back to a pending flag after all (we need to have another engineering discussion on the submission ids), so not having divergence on the submissions and pending submissions schema keeps us as compatible as we can be.

wanlingt (approved):
lgtm! agree with the choice of Option 1.

I suppose Option 3 might be better if we were to add other properties to `PendingSubmission` in the future, such that it's an extension of `Submission`.. but since we're copying it over wholesale this is probably highly unlikely

src/app/models/pending_submission.server.model.ts
justynoh: Yeah, I think you might be right too. Let me update it and we can give it a shot!

Issue #5820, opened 2023-02-22
Handle Payouts


foochifa:
We receive payout notification (webhooks) which are aggregate of many payments. We need to process these payouts notifications and update the payments independently.

Note: 
Need to care for iteration cost and replay idempotency.
Likely we need to split the webhooks to jobs and have a way to process them with some rate and concurrency control ([SQS](https://aws.amazon.com/sqs/)? [Bull Queues](https://github.com/OptimalBits/bull)?). The webhook handler should not be a long blocking job that risksbreaking.

timotheeg:
We receive payout notification (webhooks) which are aggregate of many payments. We need to process these payouts notifications and update the payments independently.

justynoh:
Solved in #5937 , but true that as payment forms get more popular we might need to revisit our latency as we modify many documents at once. Can relook in the future imo

PR #5937, opened 2023-03-16
fix: webhook handler for Stripe events

### LINKED ISSUE #5805: Store all payment activities from stripe webhook
Store in append-only mode every interaction we receive on a given payment

justynoh:
## Problem
In the previous iteration of webhook handling, Stripe events were not handled completely correctly. A lot of functionality was missing and a number of specs have changed since then.

Closes #5805 and #5820

## Solution

Don't worry, about half the lines are from a test file :)

This PR basically overhauls the entire old webhook handling functionality and rewrites it in a (what I would think) better and more reliable way, making use of utility functions that were added in previous PRs.

Changelog:
- Added `stripe-event-types` to be a dev-dep for better type discrimination.
- Added new payment statuses `PartiallyRefunded`, `FullyRefunded` and `Disputed`.
- Some updates for consistency between payment model and payment type (change model type of `webhookLog`, change type of `payoutMeta`)
- Some additional documentation updates here and there
- `stripe.controller.ts`
   - rewrite of most of the webhook handler to process events correctly
   - remove references to `getPaymentFromLatestSuccessfulCharge` to find by submission Id instead.
- `stripe.service.ts`
   - delete `getPaymentFromLatestSuccessfulCharge` and port the code over to `confirmStripePaymentPendingSubmission` in addition to using `PaymentsService.confirmPaymentPendingSubmission`. 
   - write `processStripeEvent` which is a generic event handler for all events, rather than letting the controller decide on the update to perform.
- `stripe.utils.ts`
   - move `getRedirectUrl` from `payments.utils.ts` since it is a stripe-specific redirect url 
   - add helper functions to extract payment id `getMetadataPaymentId`, recompute the payment state using the received charge events `computePaymentState` (used in the event handler), and compute the payout state `getPayoutState`

State diagram for `computePaymentState`
image

State diagram for `getPayoutState`
image

## Additional notes
There are several places where we might need to put alarms. In particular,
- `computePaymentState`: If we run into step 2 fallback (indicates that an unknown transition occurred and we should account for it in our state machine)
- `_handleStripeEventUpdates`: If we have thrown errors in the webhook handler in Step 4 (can filter for which specific error we would like to catch, e.g. `PaymentNotFoundError` is probably not as bad as `PendingSubmissionNotFoundError`).

TODOs
- Add a `mapRouteError` to the payments functions to return appropriate status codes.
- Extract `paymentId` from metadata into a constant (in general, extract metadata keys into constants)
- Figure out how to deal with race condition (see comment below)
- Add tests for state machine reducers in `stripe.utils.ts`
- Determine what alarms need to be set
  - when `paymentStateReducer` fails to return a known state
  - when the handler itself returns non-OK error codes or times out (see comment below)


justynoh:
There is a small bug in the code, although it does not impact correctness. There is a race condition when two webhooks get sent simultaneously and are received by two different servers:

image

In `processStripeEvent`, the general structure looks like this:
```typescript
const processStripeEvent = () => {
  // ...
  PaymentModel.findById() // (1)
  // ... some computation ... (2)
  payment.save() // (3)
}
```

What could potentially happen is that when webhook A arrives, it is taken by server A, and when webhook B arrives, it is taken by server B.
- server A executes (1)
- server B executes (1)
- server A and B both execute (2) 
- server A executes (3)
- server B executes (3)

In this case, because server A saved the document first, server B's document is out of sync, causing a version error. This will happen anytime the operations are overlapping (i.e. two servers are working on the same payment document, at the same time). 

The reason why this does not affect correctness is that even if server A succeeds and server B fails, Stripe will retry the webhook again later on, so effectively the two successful operations do run sequentially. However, this is dependent on Stripe's retry timing (it really is just a few seconds later which is good enough for our purposes), so feels a bit dirty. If we can make sure all webhooks succeed on their first send, we should do that.

So far here's what I've considered as to remedying this bug.
- The straightforward thing to do is to wrap it in a transaction... but given that `confirmPaymentPendingSubmission` itself is already in a transaction, I want to avoid doing that. 
- Another option I've considered is to use `PaymentModel.findByIdAndUpdate` which is supposed to be atomic. However I'm not entirely sure if this is possible because the update operation is not that simple. I _think_ possible to use aggregation pipelines to update but the specifics might prove otherwise.

I'm open to other suggestions as well!

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -0,0 +1,563 @@
...
+      const result = await StripeService.processStripeEvent(
+        String(payment._id),
+        eventChargeFailed,
+      )
+
+      // Assert
+      expect(result.isOk()).toEqual(true)
+
+      const updatedPayment = await Payment.findById(payment.id)
+      if (!updatedPayment) throw new Error('Payment not found!')
```
KenLSM: Is throwing errors from tests expected (pun intended)? If we are asserting that `updatedPayment` exist, it might work if we use `expect(updatedPayment).toBeTruthy()`

justynoh: the intention of throwing the error was to distinguish between a functional failure and a test case error. In this case, it should be impossible for the payment to not exist, and not really something we are expecting from the test case... UNLESS the tests should also verify that the payment document was not deleted by the function, but that seemed a bit extreme haha. tbh i'm not entirely sure what the right thing to do here is, what do you think?

KenLSM: > the tests should also verify that the payment document was not deleted by the function,

I feel that it is reasonable to verify it, if we want to ensure that `await Payment.findById(payment.id)` will always return something.

KenLSM:
Yeah the `processStripeEvent` is a very interesting problem.

I feel that its a problem of our monolithic repo. If we could route by sharding specific payment-id over to the same server + queue, then there would be no issue of two event from the same id be consumed by two different server.

src/app/modules/payments/stripe.controller.ts
```diff
@@ -191,10 +241,7 @@ export const checkPaymentReceiptStatus: ControllerHandler<{
     },
   })
 
-  return StripeService.getPaymentFromLatestSuccessfulCharge(
```
KenLSM, line 194: This change has conflict with [dedicated payment page](https://github.com/opengovsg/FormSG/pull/6001/files#diff-a8cc140505581213ebf38ff6c87b84cba6e125176f67467358811b15a684e747L180). 


```
return PaymentService.findPaymentBySubmissionId(submissionId)

vs

return PaymentService.findPaymentByPaymentIntentId(paymentId)
```

src/app/modules/payments/stripe.utils.ts
```diff
@@ -0,0 +1,256 @@
...
+ * @returns err(StripeMetadataValidPaymentIdNotFoundError) if the payment id was not found or is an invalid BSON object id
+ */
+export const getMetadataPaymentId = (
+  metadata: Stripe.Metadata,
+): Result<string, StripeMetadataValidPaymentIdNotFoundError> => {
+  const logMeta = {
+    action: 'getMetadataPaymentId',
+    metadata,
+  }
+  const paymentId = get(metadata, 'paymentId') // TODO: Extract this value to a constant?
```
KenLSM, line 49: Yeah I agree that we should have it extracted as a constant, if not it might be difficult to source where this metadata originates from.
```ts
export interface StripePaymentMetaDto extends Stripe.Metadata {
  paymentId: string
  formId: string
}
```

```ts
export const getMetadataPaymentId = (
  metadata: StripePaymentMetaDto,
): Result<string, StripeMetadataValidPaymentIdNotFoundError> => {
  ...
  const paymentId = get(metadata, 'paymentId')
```

And going further, we can specify this in `src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts`. We then can do a global text search to find the association.
```ts
const paymentMetaData: PaymentMetaDto = {
  formId,
  paymentId,
}
// Step 3: Create the payment intent via API call to stripe.
// Stripe requires the amount to be an integer in the smallest currency unit (i.e. cents)
const createPaymentIntentParams: Stripe.PaymentIntentCreateParams = {
  amount,
  currency: paymentConfig.defaultCurrency,
  payment_method_types: [
    'card',
    /* 'grabpay', 'paynow'*/
  ],
  description: form.payments_field.description,
  receipt_email: req.body.paymentReceiptEmail,
  metadata: paymentMetaData,
}
```

src/app/modules/payments/stripe.controller.ts
```diff
@@ -91,84 +68,157 @@ export const _handleStripeEventUpdates: ControllerHandler<
...
+          ? event.data.object.charge
+          : event.data.object.charge.id
+
+      const payment = await PaymentModel.findOne({ chargeIdLatest })
+      if (!payment) {
+        logger.warn({
+          message: 'Received dispute event with unknown latest charge id',
+          meta: { ...logMeta, chargeIdLatest },
+        })
+        return res.sendStatus(StatusCodes.BAD_REQUEST)
```
KenLSM, line 139: Hm, this will result in stripe continuing to send us such request 🤔. 

Does it make sense for us to drop it? I assume that subsequent webhook from stripe will still send us nothing, and hog our servers.

KenLSM: In the situation where we have not responded with a non-`OK`, over three days, stripe might disable our webhook.

So I think it might be risky if we continue to reject the invalid webhook.

src/app/modules/payments/stripe.utils.ts
```diff
@@ -0,0 +1,256 @@
...
+    (
+      event: Stripe.DiscriminatedEvent,
+    ): event is
+      | Stripe.DiscriminatedEvent.ChargeEvent
+      | Stripe.DiscriminatedEvent.ChargeDisputeEvent =>
+      event.type === 'charge.failed' ||
+      event.type === 'charge.pending' ||
+      event.type === 'charge.refunded' ||
+      event.type === 'charge.succeeded' ||
+      event.type.startsWith('charge.dispute.'),
```
KenLSM: I think this might be terser,
```suggestion
      [
        'charge.failed',
        'charge.pending',
        'charge.refunded',
        'charge.succeeded',
      ].includes(event.type) || event.type.startsWith('charge.dispute.'),
```

src/app/modules/payments/stripe.controller.ts
```diff
@@ -91,84 +68,157 @@ export const _handleStripeEventUpdates: ControllerHandler<
...
-    case 'payment_intent.partially_funded':
-    case 'payment_intent.payment_failed':
-    case 'payment_intent.processing':
-    case 'payment_intent.requires_action':
-    case 'payment_intent.succeeded':
-      await findPaymentAndUpdate(event.data.object.metadata, {}, event)
+    }
+    case 'charge.refund.updated': {
+      const chargeIdLatest =
+        typeof event.data.object.charge === 'string'
```
KenLSM: This extraction of chargeId from `object.charge` is scattered in quite a few places, I think it makes sense to make it into a utils func

KenLSM:
I think we also need to setup monitoring for the webhook p99/p95 latency to detect if our webhook is taking too long to return before Stripe starts to disable our webhook. And perhaps the response code as well.

Comment here as I'm not sure if the changes is relevant to this scope.

src/app/modules/payments/stripe.utils.ts
```diff
@@ -0,0 +1,256 @@
...
+  )
+
+  // Step 1: State machine for computing the current value of the payment state.
+  // A "null" status represents an unknown state, and should be recovered from later.
+  let status: Payment['status'] | null = PaymentStatus.Pending
+  let chargeIdLatest: Payment['chargeIdLatest']
+
+  for (const event of chargeEvents) {
```
KenLSM: Im thinking that we can extract this out as a reducer function, this will allow us to test it. 

```ts
  const chargeTransitionReducer = (
    state: ChargeStateType,
    curEvent:
      | Stripe.DiscriminatedEvent
      | Stripe.DiscriminatedEvent.ChargeDisputeEvent,
  ) => {
    switch (state.status) {
      case PaymentStatus.Pending:
      case PaymentStatus.Failed:
        if (curEvent.type === 'charge.failed') {
          return {
            status: PaymentStatus.Failed,
            chargeIdLatest: curEvent.data.object.id,
          }
        }
        if (curEvent.type === 'charge.succeeded') {
          return {
            status: PaymentStatus.Succeeded,
            chargeIdLatest: curEvent.data.object.id,
          }
        }
        return {
          status: null,
          chargeIdLatest: curEvent.data.object.id,
        }
       ....
  }


  const initialChargeState = { status: null, chargeIdLatest: null }
  type ChargeStateType = {
    status: Payment['status'] | null
    chargeIdLatest: Payment['chargeIdLatest'] | null
  }

  const chargeState: ChargeStateType = chargeEvents.reduce(
    applyChargeTransition,
    initialChargeState,
  )
  if (chargeState.status) return ok(chargeState)

```

src/app/modules/payments/stripe.utils.ts
```diff
@@ -0,0 +1,256 @@
...
+          getChargeId(event.data.object.charge) === chargeIdLatest
+        ) {
+          status = PaymentStatus.Disputed
+        } else {
+          status = null
+        }
+        break
+      case PaymentStatus.Disputed:
+        if (
+          (event.type === 'charge.dispute.funds_withdrawn' ||
```
KenLSM: Same thing as above, we might be able to make it terser if we use `Array.prototype.includes`.

src/app/modules/payments/stripe.utils.ts
```diff
@@ -0,0 +1,256 @@
...
+): Ok<Payment['payout'], never> => {
+  const payoutEvents = events.filter(
+    (event): event is Stripe.DiscriminatedEvent.PayoutEvent =>
+      event.type.startsWith('payout.'),
+  )
+
+  let payout
+
+  for (const event of payoutEvents) {
+    switch (event.type) {
```
KenLSM: Also reduce-able!

src/app/modules/payments/stripe.controller.ts
justynoh: oh no HAHA ok, will update to be consistent with yours

EDIT: On second thought, there are two better options. 1) can we just merge and then fix the conflicts later? otherwise, 2) I wanted to remove `getPaymentFromLatestSuccessfulCharge` in this PR because it's been refactored into the webhook handler, so indicating that we've used it. But i can just restore it and we can delete the function later. What do you think?

src/app/modules/payments/stripe.utils.ts
justynoh: I agree! But should we do it in this PR? It's already quite big so maybe it's worth doing separately, haha

src/app/modules/payments/stripe.controller.ts
KenLSM: Yeah, I think its better for yours to merge first (as option 1), and then we re-evaluate it on the other PR.

KenLSM (approved):
No blocking issues, LGTM!

src/app/modules/payments/stripe.controller.ts
justynoh: yeah, you're right, I didn't really know what we wanted from this scenario. In theory, this should never happen, but if it does, i figured we might be able to set an alarm, do some kind of recovery and when the webhook gets retried, it might succeed again (looking at the case here where some unknown bug/race condition in another part of the code caused the charge id to be desynced). 

If we dropped the webhook and returned `OK`, even if we managed to forward-recover, we would have to manually re-trigger the webhook from the Stripe dashboard. I'm not sure if that's super easy to do (we'll have the event ID, but it could still be hard to find and re-trigger)

I'm not entirely sure the correct thing to do is here, what do you think!

src/app/modules/payments/stripe.utils.ts
justynoh: this one will have a tradeoff since it's not a custom typeguard. We rely on the equality check to ensure that `event.data.object` is a `Stripe.Dispute` which makes the `.charge` attribute available later on:

```typescript
getChargeIdFromNestedCharge(event.data.object.charge) === chargeIdLatest
```

However it's possible to use `.includes` and then do something like
```typescript
getChargeIdFromNestedCharge((event.data.object as Stripe.Dispute).charge) === chargeIdLatest
```
what do you think is better?

src/app/modules/payments/stripe.utils.ts
justynoh: [3a5c6c8](https://github.com/opengovsg/FormSG/pull/5937/commits/3a5c6c8ee0400042dc3721f7266fd4b977135c80)

src/app/modules/payments/stripe.controller.ts
justynoh: [f653ab5](https://github.com/opengovsg/FormSG/pull/5937/commits/f653ab513d15b3163ef302d83046d80bbc6911cb)

src/app/modules/payments/stripe.utils.ts
justynoh: [4ddaa44](https://github.com/opengovsg/FormSG/pull/5937/commits/4ddaa446d617dce8617b3f7768cf9f2e14938493)

src/app/modules/payments/stripe.utils.ts
justynoh: [4ddaa44](https://github.com/opengovsg/FormSG/pull/5937/commits/4ddaa446d617dce8617b3f7768cf9f2e14938493)

src/app/modules/payments/__tests__/stripe.service.spec.ts
justynoh: [1ffc37a](https://github.com/opengovsg/FormSG/pull/5937/commits/1ffc37ae1dc884e4fc94dd737e0098c2a097fc46)

src/app/modules/payments/stripe.utils.ts
KenLSM: TIL, type narrowing isn't supported on [`Array.prototype.includes`](https://github.com/microsoft/TypeScript/issues/15048). 

I think for this specific case, we do make use of type narrowing, and thus it actually changes the intended typecheck behaviour at the cost of being terse. That isn't what we want.

With that, I think we should keep the original implementation. 🙈

src/app/modules/payments/stripe.controller.ts
KenLSM: Understand. I think we don't have enough data to justify either way -- on whether it is risky to keep rejecting or whether there's an unknown bug/race condition.

Agree to keep status quo.

PR #5940, opened 2023-03-17
build: release 6.36.1 hotfix

tshuli:
## Problem
- In release [6.32.0](https://github.com/opengovsg/FormSG/pull/5832), we introduced a [bug fix ](https://github.com/opengovsg/FormSG/pull/5796)  to fix another issue where logic with trailing spaces was preventing submission. 
- This inadvertently led to  another bug where logic conditions are not triggering if the dropdown condition contains whitespace
- To resolve the bug, this PR trims the whitespace on dropdown options before running logic check. 

## Tests
- [x] Create form dropdown as show logic. In DB, add whitespace to both the dropdown option and logic option.
- [x] Show logic should still work when user selects the dropdown option with whitespace.
- [x] Test again if dropdown has no whitespace but logic has whitespace
- [x] Test again if dropdown has whitespace but logic has no whitespace


timotheeg (approved):
👍

justynoh (approved):
lgtm

PR #5942, opened 2023-03-19
feat: show transaction fee in individual response page

### LINKED ISSUE #5803: Show payment details in individual response page

wanlingt:
## Problem

We want to display the transaction fee in the individual response page.

Closes #5803

## Solution

1. Add `stripeTransactionFee` and `receiptUrl` to payments model
2. Retrieve transaction fee from Stripe using the latest successful charge's [balance transaction object](https://stripe.com/docs/api/balance_transactions/object?lang=node). This is done at the same time as receipt retrieval, as the receipt and Stripe's transaction fee amount only exist when there is a successful charge. The payment object gets updated with `stripeTransactionFee` and `receiptUrl`.
3. Refactor `getReceiptUrl` into `getPaymentFromLatestSuccessfulCharge` to return the payment object instead of the receipt URL

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:
image

## Tests

- [x] Make a payment on a public form. As the admin, go to the individual response page. Check that the transaction fee corresponds to the `stripeTransactionFee` value in the DB.


shared/types/payment.ts
```diff
@@ -17,6 +21,8 @@ export type Payment = {
   payoutId: string
   payoutDate: Date
   created: DateString
+  stripeTransactionFee: number
```
justynoh: should we just call this `transactionFee`? this would avoid explicitly referring to stripe as much as possible in code

src/app/models/payment.server.model.ts
```diff
@@ -42,6 +42,10 @@ const compilePaymentModel = (db: Mongoose): IPaymentModel => {
       payoutDate: {
         type: Date,
       },
+      stripeTransactionFee: {
+        type: Number,
+      },
+      receiptUrl: { type: String },
```
justynoh: supernit: 
```typescript
receiptUrl: {
  type: String,
}
```

src/app/modules/payments/payments.service.ts
```diff
@@ -80,3 +80,43 @@ export const findPaymentBySubmissionId = (
...
+  // Retrieve payment object from database and
+  // Update payment's receipt url
+  return ResultAsync.fromPromise(
+    PaymentModel.findByIdAndUpdate(
+      paymentId,
+      {
+        $set: {
+          receiptUrl,
+          stripeTransactionFee,
+          updatedAt: new Date(),
```
justynoh: do we need to explicitly update this field? I expected that it would be updated automatically, as defined by the `timestamps` in the model

justynoh (approved):
lgtm

shared/types/payment.ts
wanlingt: yup good point! I put it as `stripeTransactionFee` initially as there were many other types of fees e.g. 
image ([source](https://stripe.com/docs/api/balance_transactions/object)) and I wanted to be clear that this was the fee that Stripe charges. but on on further thought you're right, we can leave it as `transactionFee` since we know what we're referring to

src/app/modules/payments/payments.service.ts
wanlingt: [073bdd5](https://github.com/opengovsg/FormSG/pull/5942/commits/073bdd55fcd76bf9fd8d4008cec21e733fee47d4)

src/app/models/payment.server.model.ts
wanlingt: [073bdd5](https://github.com/opengovsg/FormSG/pull/5942/commits/073bdd55fcd76bf9fd8d4008cec21e733fee47d4)

shared/types/payment.ts
wanlingt: [073bdd5](https://github.com/opengovsg/FormSG/pull/5942/commits/073bdd55fcd76bf9fd8d4008cec21e733fee47d4)

PR #5944, opened 2023-03-20
fix: add discriminator to pending submission collection, submission confirmation service

justynoh:
## Problem
This PR fixes a bug in the pending submission collection where the incorrect discriminator value is used.

## Solution
If a value (third argument) is not provided, the discriminator value defaults to the discriminated collection's schema ID. But this is not what we want, since we need the pending submissions schema IDs to be unique from that of the corresponding schema IDs from the submissions collection, but the values must be the same. So we add the third argument, which is the value of the discriminator key.

## Additional notes

Additionally, I took the opportunity to write the transaction as promised here.

image

Due to a bunch of changes to the separation of concerns, I repurposed one of the functions and marked several other functions as being refactorable into functions that can be used later for our implementation. This will help us avoid duplicating work later.

AND THEN

Once I started writing I realized that the transaction is meant to guarantee that a bunch of stuff appears at the same time (i.e. the transaction fee, submission id etc) so I also refactored the payment model to group those data together. Hopefully this makes it more wieldy for us in the future to check if payments are completed and access the attributes without having to add `!` guards everywhere.

OK SO 

This pr has already had a ton of scope creep and I kind of want to stop it here. Hopefully the comments and TODOs are good enough markers so that future us remembers what else needs to be done, haha.

lesson learnt: just a bit of scope creep can become a lot of scope creep, so just don't creep at all

### DIFF
--- a/src/app/modules/payments/payments.service.ts
+++ b/src/app/modules/payments/payments.service.ts
@@ -10,7 +10,10 @@ import { DatabaseError } from '../core/core.errors'
 import { PendingSubmissionNotFoundError } from '../submission/submission.errors'
 import * as SubmissionService from '../submission/submission.service'
 
-import { PaymentNotFoundError } from './payments.errors'
+import {
+  PaymentAlreadyConfirmedError,
+  PaymentNotFoundError,
+} from './payments.errors'
 
 const logger = createLoggerWithLabel(module)
 const PaymentModel = getPaymentModel(mongoose)
@@ -129,7 +132,10 @@ export const confirmPaymentPendingSubmission = (
   transactionFee: number,
 ): ResultAsync<
   IPaymentSchema,
-  PaymentNotFoundError | PendingSubmissionNotFoundError | DatabaseError
+  | PaymentNotFoundError
+  | PendingSubmissionNotFoundError
+  | PaymentAlreadyConfirmedError
+  | DatabaseError
 > => {
   const logMeta = {
     action: 'confirmPaymentPendingSubmission',
@@ -152,7 +158,14 @@ export const confirmPaymentPendingSubmission = (
     })
 
     return (
+      // Step 1: Retrieve the payment by payment id and check that the payment
+      // has not already been confirmed.
       findPaymentById(paymentId)
+        .andThen((payment) =>
+          payment.completedPayment
+            ? errAsync(new PaymentAlreadyConfirmedError())
+            : okAsync(payment),
+        )
         .andThen((payment) =>
           // Step 2: Copy the pending submission to the submissions collection
           SubmissionService.copyPendingSubmissionToSubmissions(
--- a/src/app/modules/payments/payments.service.ts
+++ b/src/app/modules/payments/payments.service.ts
@@ -160,7 +160,7 @@ export const confirmPaymentPendingSubmission = (
     return (
       // Step 1: Retrieve the payment by payment id and check that the payment
       // has not already been confirmed.
-      findPaymentById(paymentId)
+      findPaymentById(paymentId, session)
         .andThen((payment) =>
           payment.completedPayment
             ? errAsync(new PaymentAlreadyConfirmedError())
--- a/src/app/modules/payments/payments.service.ts
+++ b/src/app/modules/payments/payments.service.ts
@@ -6,15 +6,43 @@ import { IPaymentSchema } from '../../../types'
 import { createLoggerWithLabel } from '../../config/logger'
 import getPaymentModel from '../../models/payment.server.model'
 import { getMongoErrorMessage } from '../../utils/handle-mongo-error'
-import { ApplicationError, DatabaseError } from '../core/core.errors'
+import { DatabaseError } from '../core/core.errors'
+import { PendingSubmissionNotFoundError } from '../submission/submission.errors'
+import * as SubmissionService from '../submission/submission.service'
+
+import { PaymentNotFoundError } from './payments.errors'
 
 const logger = createLoggerWithLabel(module)
 const PaymentModel = getPaymentModel(mongoose)
 
-export class PaymentNotFoundError extends ApplicationError {
-  constructor(message = 'Payment not found') {
-    super(message)
-  }
+/**
+ * Retrieves payment by Id.
+ * @param paymentId the payment id of the payment to be retrieved
+ * @returns ok(payment) if payment exists
+ * @returns err(PaymentNotFoundError) if the payment does not exist
+ * @returns err(DatabaseError) if error occurs whilst querying the database
+ */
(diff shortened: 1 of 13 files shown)


src/app/modules/submission/submission.service.ts
```diff
@@ -203,3 +207,58 @@ export const findSubmissionById = (
     return okAsync(submission)
   })
 }
+
+/**
+ * @param pendingSubmissionId the id of the pending submission to confirm
+ * @param session (optional) the mongoose transaction session to be used, if any
```
wanlingt: is there a case where this will be done outside of a transaction session?

justynoh: not sure tbh. you're absolutrly right that it is a thing that we might want to enforce. 

It might be a case where enforcing it now makes sense, and if we ever need to relax it later we can do that more easily. However, I didn't want to pre-emptively block feature dev if we ever need this functionality without sessions (since we may still have unknown unknowns).

That being said, I'm ok with making it a required parameter, and then if we need to relax it later we can always do that!

KenLSM: I think we should enforce it. It can be a costly mistake, and an evasive bug.

src/app/modules/submission/submission.service.ts
```diff
@@ -203,3 +207,58 @@ export const findSubmissionById = (
...
+ *
+ * @returns ok(submission document) if a submission document is successfully created
+ * @returns err(PendingSubmissionNotFoundError) if pending submission does not exist in the database
+ * @returns err(DatabaseError) if database errors occurs while copying the document over
+ */
+export const confirmPendingSubmission = (
```
KenLSM: From what I understand `confirmPendingSubmission` is an business logic that should involve two data logic of updating 1) `payment` fields and 2) `copyPendingSubmissionToSubmission`. Both of them should happen in as transaction. There wouldn't be a case where there's a 1) without 2) fields updating if I'm not wrong.

By that argument, this action doesn't need a session, but instead, it should manage its own session.

justynoh: I think this is related to the other conversation so we can just have one thread for this. Context for writing this was the design we had before yesterday, where the responsibility of moving submissions was not coupled with updating the other payment-related data yet (e.g. receipt_url etc).

I had envisioned that this service would only be responsible for moving documents from `pendingSubmissions` to `submissions` and nothing else. I realized while writing the webhook handler that this specific part of the code was becoming pretty clunky and hard to read in the middle of all the other payments-related business logic. Therefore, I wanted to factor out this potentially more generic operation of moving documents from pending to complete, which could be used in places other than payments later on. 

Things have changed since then, obviously! Given the new design, I agree that we should have a single function that updates the payment document with the critical business logic in its own session. Now in terms of code design, how should we think about it? (Maybe it's a false dichotomy and you can point out if it is but) It feels like there is a balance between safety and maintainability here. Should the entire logic be encapsulated into one function, so that it's impossible for someone to use a subsection of the logic in an unintended way (safety)? Or, are we ok with still writing helpers (but maybe not exporting them, for readability)? As an illustration: the first option would look like 

```typescript
export const confirmPayment = (...args) => {
  await PendingSubmission.findById()
      .then((pendingSubmission) => { Submission.save(pendingSubmission) }) // of course, not as simple as this, and will probably look very hard to read
      .then((submission) => { Payment.update(paymentId, {submission.id}) } )
      // ...etc
```

whereas the second option _could_ look like
```typescript
const movePendingSubmissionToSubmission = (pendingSubmissionId) => {
  // identical to the confirmPendingSubmission function written in this PR
  await PendingSubmission.findById()
      .then((pendingSubmission) => { Submission.save(pendingSubmission) })
}

const updatePaymentWithSuccessMetadata = (...args) => {
  await Payment.update(...)
}

export const confirmPayment = (...args) => {
  // Much easier to read and more maintainable
  await movePendingSubmission(pendingSubmissionId)
    .then(updatePaymentWithSuccessMetadata)
```

Would be interested to hear what you/people think!

KenLSM: Im buying option 2. Looks cleaner without any sacrifice to functionality or additional risks. Testing private functions can be done through rewire too.

> Things have changed since then, obviously! 

Apologies for time travelling xD

justynoh: haha np! sounds good, will rewrite to accomodate this. In this case, the tradeoff here is that it might look a bit strange - the `movePendingSubmissionToSubmission` function will be in the payments service, despite not really touching the payments model at all. I think that's fine though, and we'll manage if we never need to generalize the function. thanks for the input!

EDIT: actually on second thought, I'm gonna elect to leave it within the submission service, but enforce a session to be provided as discussed earlier

wanlingt: thanks for the elaborations! agree that the second option looks easier to understand

justynoh:
plz see the edits above 😭

src/app/modules/payments/payments.service.ts
```diff
@@ -81,41 +109,111 @@ export const findPaymentBySubmissionId = (
...
+    return new DatabaseError(getMongoErrorMessage(error))
+  }).andThen((session) => {
+    session.startTransaction({
+      readPreference: 'primary',
+      readConcern: { level: 'snapshot' },
+      writeConcern: { w: 'majority' },
+    })
+
+    return (
+      findPaymentById(paymentId)
```
KenLSM: Do we need to check if the pendingSubmission has been copied to submission?

justynoh: OMG. LOL. You are right let me make this change real quick

justynoh: ok, done!

KenLSM: Also another (possibly dumb) question, why don't we need to pass session into `findPaymentById`?

Does `payment.save` already ensure that the initial payment entry isn't already altered before saving?

i.e.,
```
Sequence
1. const payment = await findPaymentById(314)
> { __id: 314,  __updated_time: 100, status: 'catey' }

# halfway through from another server
2. payment2 <- anotherFuncThatAltersPaymentId(314)
> { __id: 314,  __updated_time: 101, status: 'dogey' }

3. payment.status = 'ottey'
> { __id: 314,  __updated_time: 100, status: 'ottey' }

4. payment.save({ session }) // goes through or throws
> { __id: 314,  __updated_time: 102, status: ?? }
```

justynoh: hmmmm, yeah, i think you're right, will add it in 👍

KenLSM (approved):
LGTM Justyn. Transactions are tough!

PR #5945, opened 2023-03-20
docs: Script to find invalid whitespaces

tshuli:
- Adds a reference script which was used for [trailing whitespaces incident](https://docs.google.com/document/d/13bqWVNIOKQtHy0a8nW41itKWQQArNAvj49jg0zxncdQ/edit)
- Reference hotfix: #5940

### DIFF
--- a/scripts/20230317_whitespace_issue/index.js
+++ b/scripts/20230317_whitespace_issue/index.js
@@ -8,9 +8,9 @@ async function getStats(db) {
 
   const cursor = await collection
     .find({
-      responseMode: "encrypt",
-      status: "PUBLIC",
-      "form_logics.1": {$exists: true}
+      responseMode: 'encrypt',
+      status: 'PUBLIC',
+      'form_logics.1': { $exists: true },
     })
     .project({
       form_fields: 1,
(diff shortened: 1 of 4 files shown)


scripts/20230317_whitespace_issue/index.js
```diff
@@ -0,0 +1,203 @@
...
+  })
+
+  console.log('')
+  console.log('Number of forms with field with invalid whitespace')
+  console.log(Object.keys(formsWithInvalidWhiteSpace).length)
+  console.log('Number of forms where whitespace affects logic:')
+  console.log(Object.keys(formsWhereWhiteSpaceAffectsLogic).length)
+
+  // TODO query submission to chedk how many forms of those forms had submissions in the past 3 weeks.
+
```
timotheeg, line 114: Can remove this TODO since it was done :) 

```suggestion
```

timotheeg (approved):

timotheeg (approved):

scripts/20230317_whitespace_issue/index.js
```diff
@@ -0,0 +1,254 @@
...
+    .project({ _id: 1, form: 1 })
+    .toArray()
+
+  console.log(allSubmissions.length)
+
+  console.log('-----')
+  console.log(
+    'Total number of forms which received submissions during that timeframe:',
+  )
+  const allForms = new Set(allSubmissions.map((s) => s.form.toString()))
```
timotheeg, line 174: This works fine, but just some feedback: the process as coded here gathers a huge dataset of repeated entries,  that needs to be stored entirely in memory (~1.3M entries in this case). 

When computing stats, leverage the DB as much as possible to reduce the size of datasets to just what you need. In this case, it would have probably been better to query for an aggregate of submission count by formid, and then do a summation of the submission counts in code to derive the total number of submissions received overall.

PR #5946, opened 2023-03-20
fix: move permission list validation from model level to service layer

justynoh:
## Problem
Whenever agencies change email domains, we remove their domain from the Agency collection. However, this means that if they had any collaborators in their form collaborator list whose emails were from the old domain, their form is forever locked. They cannot make changes to their form anymore, since the form document cannot be saved with invalid collaborators. 

image

The reason for this occurring is validation that is applied at the model level. The validation is thus run on every document save, and the database rejects the invalid save on change.

## Solution
Even though we would _like_ for there to be model-level validation, this is not practical or informative. The model validations should specify invariants of the model over all operations throughout the system. However, the `agency` collection is always subject to change, and the validator will certainly not be invariant over time. Therefore, we should not in principle validate based on the `agency` collection at the model level. All we can guarantee is that the list of collaborators are emails. What we would _really_ like is that all _new_ collaborators are whitelisted.

Therefore, this PR removes the model-level validation based on the `agency` collection. In replacement, the validation occurs at the service layer. Any new collaborators added are extracted and checked for validity, while old collaborators are left untouched.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/226308884-e7524156-844b-4fa1-ba6f-12a146c9c0b2.mov

https://user-images.githubusercontent.com/25571626/226308964-6cdee75e-ff02-4da4-b145-6f97b80946d7.mov

## Tests

- [ ] Ensure that the email domain `data.gov.sg` is not in the `govtech` agency. Try to add `leonard@data.gov.sg` email as a collaborator. This should fail. 
- [ ] Add the `data.gov.sg` email domain to the `govtech` agency. Now add `leonard@data.gov.sg` and `formsg@data.gov.sg` email as collaborators. Then, remove the `data.gov.sg` email domain from the `govtech` agency. Perform the following actions in order. 
    - Try modifying some form fields. This should work.
    - Remove `formsg@data.gov.sg` as a collaborator. This should work.
    - Modify the read/write permission of `leonard@data.gov.sg`. This should fail.
    - Add `leonard@open.gov.sg` as a collaborator. This should work.
    - Remove `leonard@data.gov.sg` as a collaborator. This should work.

## Notes

Separately, I found an existing bug with transfer ownership. We currently allow ownership to be transferred to users whose emails are no longer valid domains belonging to any agency. You can see this in the second video clip, right at the end. I'll open a ticket for it after looking through the existing issues to see if it's duplicated.

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.service.ts
@@ -43,6 +43,7 @@ import {
 import { EditFormFieldParams, FormUpdateParams } from '../../../../types/api'
 import config, { aws as AwsConfig } from '../../../config/config'
 import { createLoggerWithLabel } from '../../../config/logger'
+import getAgencyModel from '../../../models/agency.server.model'
 import getFormModel from '../../../models/form.server.model'
 import * as SmsService from '../../../services/sms/sms.service'
 import { twilioClientCache } from '../../../services/sms/sms.service'
@@ -83,6 +84,7 @@ import {
   CreatePresignedUrlError,
   EditFieldError,
   FieldNotFoundError,
+  InvalidCollaboratorError,
   InvalidFileTypeError,
 } from './admin-form.errors'
 import {
@@ -94,6 +96,7 @@ import {
 
 const logger = createLoggerWithLabel(module)
 const FormModel = getFormModel(mongoose)
+const AgencyModel = getAgencyModel(mongoose)
 
 export const secretsManager = new SecretsManager({
   region: config.aws.region,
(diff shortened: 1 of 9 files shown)


LinHuiqing (approved):
lgtm!

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -751,26 +754,71 @@ export const updateForm = (
...
+    formId: form._id,
+  }
+
+  // Get the updated (added or modified) collaborator emails (i.e. they are not
+  // in the original collaborator list).
+  const updatedCollaboratorEmails = updatedCollaborators
+    .filter(
+      (c1) =>
+        !form.permissionList.some(
+          (c2) => c1.email === c2.email && c1.write === c2.write,
```
timotheeg, line 777: Just checking: is normalization of the emails (trim + lowercasing) done before this filtering occurs? (question is also relevant to the agency search below)

justynoh: these are both done in the frontend. we have validation middleware requiring that the emails be emails (which means they must already be trimmed), but i don't think this enforces a lowercase restriction. we should do it here too. Thanks for catching! 

I'll open a PR to force lowercase validation in the middleware

PR #5928, opened 2023-03-14
WIP: Exp/api fetch

KenLSM:
Not intended for merge.

PR #5948, opened 2023-03-21
feat: implement fetch API fallback for network error

### LINKED ISSUE #5826: Experiment with fallback APIs when network errors are encountered
Axios is giving us Network errors that we know are not "real" network errors. 

It's possible we have issues with what we send, but also, we could be encountering a bug with Axios. 

We want to:
- Implement a fallback to Axios when network errors are detected (say with [got](https://www.npmjs.com/package/got) or [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API))
- Generously log so we can tell what is happening
- Make a decision to have a single method that is reliable to submit form data, and remove code that is not reliable

tshuli:
## Problem
- Builds off #5928 
- To deal with network error, implements the following
  - retry submission with fetch whenever axios submission fails with Network Error
  - when axios submission fails, attempts to poll client env vars using both axios and fetch
  - implements new `useFetchForSubmissions` flag which allows us to toggle between using fetch instead of axios for submissions. This is used for testing, as well as to provide the option to switch to fetch if logs reveals that fetch works as a fallback for axios

Closes #5826 

### Test normal axios mode
- [ ] Email and storage mode submissions work without captcha
  - [ ]   Datadog logs should show `handleSubmitForm: submitting via axios`
- [ ] Email and storage mode submissions work with captcha
  - [ ]   Datadog logs should show `handleSubmitForm: submitting via axios`
- [ ] Email and storage mode preview submissions work
  - [ ]   Datadog logs should show `handleSubmitForm: submitting via axios`

### Test new fetch mode (on staging)
- [ ] set `REACT_MIGRATION_USE_FETCH_FOR_SUBMISSIONS` to true in the EB environment
- [ ] Email and storage mode submissions work without captcha
  - [ ]   Datadog logs should show `handleSubmitForm: submitting via fetch`
- [ ] Email and storage mode submissions work with captcha
  - [ ]   Datadog logs should show `handleSubmitForm: submitting via fetch`
- [ ] Email and storage mode preview submissions work
  - [ ]   Datadog logs should show `handleSubmitForm: submitting via fetch`

`REACT_MIGRATION_USE_FETCH_FOR_SUBMISSIONS`: flag to use fetch for submissions instead of axios


src/app/loaders/express/locals.ts
```diff
@@ -64,6 +66,7 @@ const environment = ejs.render(
     var reactMigrationRespondentRolloutStorage = "<%= reactMigrationRespondentRolloutStorage%>"
     var reactMigrationAdminRollout = "<%= reactMigrationAdminRollout%>"
     var reactMigrationAngularPhaseOutDate = "<%= reactMigrationAngularPhaseOutDate%>"
+    var reactMigrationUseFetchForSubmissions = "<%= reactMigrationUseFetchForSubmissions%>"
```
timotheeg: For this too, I would try to not convert to string to remove typing risks (`"false"` is truthy 😰 ) ... (although it looks like this js chunk is actually not used anywhere anyway 🤷)
```suggestion
    var reactMigrationUseFetchForSubmissions = <%= reactMigrationUseFetchForSubmissions%>
```

frontend/src/features/admin-form/common/AdminViewFormService.ts
```diff
@@ -196,3 +196,65 @@ export const submitStorageModeFormPreview = async ({
...
+    publicKey,
+  )
+
+  const response = await fetch(
+    `${API_BASE_URL}/${ADMIN_FORM_ENDPOINT}/${formId}/preview/submissions/encrypt`,
+    {
+      method: 'POST',
+      body: JSON.stringify(submissionContent),
+      headers: {
+        'Content-Type': 'application/json',
```
wanlingt, line 255: is there a reason why we use `JSON.stringify` and specify the Content-Type headers here (vs `submitEmailModeFormPreviewWithFetch` and the `submitStorageModeFormPreview` axios call)?

frontend/src/features/admin-form/common/AdminViewFormService.ts
```diff
@@ -196,3 +196,65 @@ export const submitStorageModeFormPreview = async ({
     submissionContent,
   ).then(({ data }) => data)
 }
+
+/**
+ * Submit an email mode form in preview mode using fetch
+ * TODO(#5826): Fallback using Fetch. Remove once network error is resolved
```
wanlingt, line 202: Should we rename this or make the description for #5826 clearer? I feel it's unclear about whether this fetch method tackles #5826 (this is what it's doing right?), or whether we still have a TODO left in the form of #5826. Also, are we removing this when the network error is resolved, since resolved might mean keeping fetch instead of axios?

wanlingt (commented):
thanks shuli! lgtm for the parts where we replace axios with fetch.

could we make the tests more comprehensive? 
- for example, specify that we should "set `REACT_MIGRATION_USE_FETCH_FOR_SUBMISSIONS` flag to `true`" before carrying out the tests. 
- add a test to check that the DD logs show up for staging
- add a test for `axiosDebugFlow`

frontend/src/features/admin-form/common/AdminViewFormService.ts
timotheeg: For extra info here:
* For email mode, we generate a [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object, which is understood by both Axios and `fetch()`, so we don't need to set custom headers.
* For Storage mode, we supply a POJO object. Axios automatically applies JSON-encoding to it, and sets the header accordingly, while when we use `fetch()`, we have to do it ourselves (see [docs here](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#uploading_json_data))

timotheeg (commented):
Functionally, this looks OK. The code in `PreviewFormProvider.tsx` feels like it has a lot of repetition in term of gathering payload and preparing log data. Some refactoring would be welcome (not over the top though, since we know some of the code will eventually be dropped). 

Perhaps, the fetch() calls in fetch mode and in error catching mode could use local functions.

timotheeg:
@tshuli, I pushed [a commit](https://github.com/opengovsg/FormSG/pull/5948/commits/c1899ad03173983c8fe4f7dd3afcd2fb61eea948) to show how `PreviewFormProvider.tsx` can be refactor a bit, there could be more, but I don't want to go over the top. I'll do the same for `PublicFormProvider.tsx` shortly, please help me take a look at the code to be sure I didn't mess up anything. Thanks!

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -227,43 +243,133 @@ export const PublicFormProvider = ({
...
+            datadogLogs.logger.info(`handleSubmitForm: submitting via fetch`, {
+              meta: {
+                action: 'handleSubmitForm',
+                responseMode: 'storage',
+                method: 'fetch',
+                useFetchForSubmissions,
+              },
+            })
+
+            return submitEmailModeFormFetchMutation
```
timotheeg: This should be using `submitStorageModeFormFetchMutation()`

timotheeg:
@tshuli , I added one more commit to refactor `PublicFormProvider.tsx` (and I noticed one copy/paste issue in the file while doing it 😅). I'll do one more round of review on the whole PR later today, and meanwhile, please help me check the last 2 commits, thanks!.

frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
```diff
@@ -100,65 +105,200 @@ export const PreviewFormProvider = ({
...
+                })
+                showErrorToast(error)
+              })
+          }
+
+          // Using mutateAsync so react-hook-form goes into loading state.
+
+          // TODO (#5826): Toggle to use fetch for submissions instead of axios. If enabled, this is used for testing and to use fetch instead of axios by default if testing shows fetch is more  stable. Remove once network error is resolved
+          if (useFetchForSubmissions) {
+            submitStorageFormWithFetch()
```
timotheeg: missing return
```suggestion
            return submitStorageFormWithFetch()
```

frontend/src/features/admin-form/common/AdminViewFormService.ts
timotheeg: I've updated #5826 with some details. Depending on the result of the experiment, ideally we keep a single mechanism to do our requests, and indeed all the comments from this PR should be removed after the final decision is taken.

timotheeg:
@wanlingt I've added some tests to verify the fetch flow works, but on staging/prod, it's not going to be possible to make the axios call specifically fail. We can try to have unit tests for that though.

frontend/src/features/admin-form/common/AdminViewFormService.ts
tshuli: yup! thanks Tim for clarifying 💯

timotheeg:
Just an extra note on the small refactor: one benefit of having the `Submit With Fetch` done with local functions is that when we test fetch mode, it's the exact same code that get executed in axios fallback, so we can have higher confidence that the code works as intended (though still not as good as testing an actual axios failure of course)

frontend/src/features/public-form/PublicFormProvider.tsx
tshuli: this was my bad, thanks for catching!

tshuli:
> thanks shuli! lgtm for the parts where we replace axios with fetch.
> 
> could we make the tests more comprehensive?
> 
> * for example, specify that we should "set `REACT_MIGRATION_USE_FETCH_FOR_SUBMISSIONS` flag to `true`" before carrying out the tests.
> * add a test to check that the DD logs show up for staging
> * add a test for `axiosDebugFlow`

Thanks for the helpful comments @wanlingt ! I beefed up the logging and added in specific dd logs to look out for in the tests

tshuli:
> @tshuli , I added one more commit to refactor `PublicFormProvider.tsx` (and I noticed one copy/paste issue in the file while doing it 😅). I'll do one more round of review on the whole PR later today, and meanwhile, please help me check the last 2 commits, thanks!.

Thanks @timotheeg ! The refactors look good, only a missing `responseMode: 'storage'` in the log meta which I've added back in

frontend/src/services/ApiService.ts
```diff
@@ -43,8 +43,10 @@ export const transformAxiosError = (error: Error): ApiError => {
         meta: {
           action: 'transformAxiosError',
           error: {
+            code: error?.code,
             message: error?.message,
             stack: error?.stack,
+            dump: error?.toJSON(),
```
timotheeg: Hmm, I don't think errors are guaranteed to have a `.toJSON()` method associated with them, so this could cause a runtime error. Could you double check?

timotheeg: Quick test in chrome console:
image

I've pushed a commit to add a safety check.

tshuli: ~would `dump: JSON.stringify(error)` be better for the error dump? instead of the safety check~ nope, `JSON.stringify` returns `{}` on an `Error` object because it doesn't have enumerable properties

frontend/src/services/ApiService.ts
```diff
@@ -43,8 +43,10 @@ export const transformAxiosError = (error: Error): ApiError => {
         meta: {
           action: 'transformAxiosError',
           error: {
+            code: error?.code,
             message: error?.message,
             stack: error?.stack,
+            dump: error?.toJSON ? error.toJSON() : undefined,
```
tshuli, line 49:
```suggestion
            dump: JSON.stringify(error),
```

frontend/src/services/ApiService.ts
timotheeg: Yes indeed! That is the reason why we read properties explicitly (`message`, `stack`) to add them to the logs. It's a bit sad, but manageable when we know what to do.

timotheeg (approved):
I think this is good to go! 👍

PR #5950, opened 2023-03-21
chore: dont update index.html after dd chunk build

darrelhong:
## Problem
Not a big issue, just some ergonomic updates to build command. When building the app locally, the file frontend/public/index.html is modified when webpack injects markup to load the datadog chunk. Git reflects that as a local change, which is confusing for developers, and there is a risk to commit the change accidentally.

## Solution
2 options proposed
1. reset index.html after dd build (may need to test in CI env where git might not exist?)

2. write output index.html to temp directory and modify webpack config to reference temp output in final cra build

**Breaking Changes** 

    - Details ...

**Improvements**:
- will not accidentally commit unwanted changes to index.html, and prevent confusion as to whether changes should be committed

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -30,3 +30,6 @@ build-storybook.log
 
 #Datadog chunk files
 datadog-chunk.*.js
+
+# temp directory for modified index.htm;
+/temp
--- a/frontend/craco.config.js
+++ b/frontend/craco.config.js
@@ -1,5 +1,6 @@
 /* eslint-env node */
 const CracoAlias = require('craco-alias')
+const HtmlWebpackPlugin = require('html-webpack-plugin')
 const merge = require('lodash/merge')
 const path = require('path')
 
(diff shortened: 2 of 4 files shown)


timotheeg:
Thanks @darrelhong. 

Not having unnecessary local changes after a build is indeed a good goal! Give us some time to get back to you on the approaches.

One piece of feedback for the PR description: the first section `Problem` should indicate why this PR exists in the first place, what problem it is solving. You can think of the 2 section `Problem/Solution`, as `Context/Approach` also. 

In this case, the problem should describe the line you had on the Improvement section you added, maybe something like this:

> When building the app locally, the file `frontend/public/index.html` is modified when webpack injects markup to load the datadog chunk. Git reflects that as a local change, which is confusing for developers, and there is a risk to commit the change accidentally.

foochifa:
Hey @darrelhong we will be closing this PR currently! We definitely need to relook into this (welcome you to create an issue for this), however, as this is non-urgent we will focus on removing angular first. Once all the different routings are cleaned up it will make resolving this issue easier!

Thanks for the contribution 🙏

PR #5972, opened 2023-03-22
feat: payment max and min limit

### LINKED ISSUE #5869: encode in max payment limit per form in a variable

tshuli:
## Problem

- Adds env var to set max payment limit for forms (default is $1000, to align with IRAS max limit for invoices)
- Adds env var to set min payment limit for forms (default is $0.50, set by stripe)
Closes #5869 

## Tests
- [ ] Create payment form. Attempt to set payment amount above $1000. Should fail on frontend.
- [ ] Attempt to set payment amount above $1000 via api call. Should fail.
- [ ] Setting payment amount of <=$1000 works.
- [ ] Repeat above for lower payment threshold of 50 cents.

- `PAYMENT_MAX_PAYMENT_AMOUNT_CENTS`: Max amount that can be collected in each payment, in cents
- `PAYMENT_MIN_PAYMENT_AMOUNT_CENTS`: Min amount that can be collected in each payment, in cents

### DIFF
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -131,7 +131,7 @@ export const PaymentInput = (): JSX.Element => {
 
   const handleCloseDrawer = useCallback(() => handleClose(false), [handleClose])
 
-  const minPaymentAmount = 0.5 // stipulated by Stripe
+  const MIN_PAYMENT_AMOUNT = 0.5 // stipulated by Stripe
 
   const amountValidation: RegisterOptions<
     FormPaymentsDisplay,
@@ -151,9 +151,9 @@ export const PaymentInput = (): JSX.Element => {
         },
         validateMin: (val) => {
           return (
-            Number(val?.trim()) >= minPaymentAmount ||
+            Number(val?.trim()) >= MIN_PAYMENT_AMOUNT ||
             `Please enter a payment amount above ${formatCurrency(
-              minPaymentAmount,
+              MIN_PAYMENT_AMOUNT,
             )}`
           )
         },
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -56,7 +56,8 @@ export const PaymentInput = (): JSX.Element => {
   const isMobile = useIsMobile()
   const { paymentsMutation } = useMutateFormPage()
 
-  const { data: { maxPaymentAmountCents } = {} } = useEnv()
+  const { data: { maxPaymentAmountCents, minPaymentAmountCents } = {} } =
+    useEnv()
 
   const setIsDirty = useDirtyFieldStore(setIsDirtySelector)
 
@@ -131,8 +132,6 @@ export const PaymentInput = (): JSX.Element => {
 
   const handleCloseDrawer = useCallback(() => handleClose(false), [handleClose])
 
-  const MIN_PAYMENT_AMOUNT = 0.5 // stipulated by Stripe
-
   const amountValidation: RegisterOptions<
     FormPaymentsDisplay,
     'display_amount'
@@ -150,10 +149,12 @@ export const PaymentInput = (): JSX.Element => {
           )
         },
         validateMin: (val) => {
+          if (minPaymentAmountCents === undefined) return true
           return (
-            Number(val?.trim()) >= MIN_PAYMENT_AMOUNT ||
+            // val is in dollars
+            Number(val?.trim()) >= minPaymentAmountCents / 100 ||
             `Please enter a payment amount above ${formatCurrency(
-              MIN_PAYMENT_AMOUNT,
+              minPaymentAmountCents / 100,
             )}`
           )
         },
@@ -169,7 +170,7 @@ export const PaymentInput = (): JSX.Element => {
         },
       },
     }),
-    [maxPaymentAmountCents],
+    [maxPaymentAmountCents, minPaymentAmountCents],
   )
 
   const handleUpdatePayments = handleSubmit(() => {
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -152,9 +152,10 @@ export const PaymentInput = (): JSX.Element => {
           if (minPaymentAmountCents === undefined) return true
           return (
             // val is in dollars
-            Number(val?.trim()) >= minPaymentAmountCents / 100 ||
+            Number(val?.trim()) >=
+              Number(centsToDollars(minPaymentAmountCents)) ||
             `Please enter a payment amount above ${formatCurrency(
-              minPaymentAmountCents / 100,
+              Number(centsToDollars(minPaymentAmountCents)),
             )}`
           )
         },
@@ -162,9 +163,10 @@ export const PaymentInput = (): JSX.Element => {
           if (maxPaymentAmountCents === undefined) return true
           return (
             // val is in dollars
-            Number(val?.trim()) <= maxPaymentAmountCents / 100 ||
+            Number(val?.trim()) <=
+              Number(centsToDollars(maxPaymentAmountCents)) ||
             `Please keep payment amount under ${formatCurrency(
-              maxPaymentAmountCents / 100,
+              Number(centsToDollars(maxPaymentAmountCents)),
             )}`
           )
         },
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -56,7 +56,7 @@ export const PaymentInput = (): JSX.Element => {
   const isMobile = useIsMobile()
   const { paymentsMutation } = useMutateFormPage()
 
-  const { data: { paymentMaxLimit } = {} } = useEnv()
+  const { data: { maxPaymentAmount } = {} } = useEnv()
 
   const setIsDirty = useDirtyFieldStore(setIsDirtySelector)
 
@@ -132,7 +132,6 @@ export const PaymentInput = (): JSX.Element => {
   const handleCloseDrawer = useCallback(() => handleClose(false), [handleClose])
 
   const minPaymentAmount = 0.5 // stipulated by Stripe
-  const maxPaymentAmount = 1000 // due to IRAS requirements and agency financial institutions are expected to be in SG
 
   const amountValidation: RegisterOptions<
     FormPaymentsDisplay,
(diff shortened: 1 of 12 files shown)


frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -148,24 +149,28 @@ export const PaymentInput = (): JSX.Element => {
...
         },
         validateMin: (val) => {
+          if (minPaymentAmountCents === undefined) return true
           return (
-            Number(val?.trim()) >= minPaymentAmount ||
+            // val is in dollars
+            Number(val?.trim()) >= minPaymentAmountCents / 100 ||
             `Please enter a payment amount above ${formatCurrency(
-              minPaymentAmount,
+              minPaymentAmountCents / 100,
```
LinHuiqing: hm, maybe it will be better if we use the centToDollars function instead of dividing by 100 to avoid division errors?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -148,24 +149,28 @@ export const PaymentInput = (): JSX.Element => {
...
         },
         validateMax: (val) => {
+          if (maxPaymentAmountCents === undefined) return true
           return (
-            Number(val?.trim()) <= maxPaymentAmount ||
+            // val is in dollars
+            Number(val?.trim()) <= maxPaymentAmountCents / 100 ||
             `Please keep payment amount under ${formatCurrency(
-              maxPaymentAmount,
+              maxPaymentAmountCents / 100,
```
LinHuiqing: Same as the above, use `centsToDollars`?

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -2628,4 +2630,124 @@ describe('admin-form.service', () => {
...
+    it('should return InvalidPaymentAmountError if payment amount exceeds maxPaymentAmountCents', async () => {
+      // Arrange
+
+      const updatedPaymentSettingsExceeded = {
+        enabled: true,
+        target_account_id: 'someId',
+        publishable_key: 'somekey',
+        amount_cents: 500000,
+        description: 'some description',
+      } as PaymentsUpdateDto
```
LinHuiqing: Should this definition make use of the spread operator to ensure that other params are the same as the ones in `updatedPaymentSettings`? So there'll be less tech debt in the future if we change our API payloads. So like:
```suggestion
      const updatedPaymentSettingsExceeded = {
        ...updatedPaymentSettings,
        amount_cents: 500000,
      } as PaymentsUpdateDto
```

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
```diff
@@ -2628,4 +2630,124 @@ describe('admin-form.service', () => {
...
+    it('should return InvalidPaymentAmountError if payment amount is below minPaymentAmountCents', async () => {
+      // Arrange
+
+      const updatedPaymentSettingsBelow = {
+        enabled: true,
+        target_account_id: 'someId',
+        publishable_key: 'somekey',
+        amount_cents: 49,
+        description: 'some description',
+      } as PaymentsUpdateDto
```
LinHuiqing: Same comment as above, making use of spread (`...`) might be better:
```suggestion
      const updatedPaymentSettingsBelow = {
        ...updatedPaymentSettings,
        amount_cents: 49,
      } as PaymentsUpdateDto
```

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1514,14 +1516,27 @@ const deleteTwilioTransaction = async (
...
+
+  // Check if payment amount exceeds maxPaymentAmountCents
+  if (amount_cents && amount_cents > paymentConfig.maxPaymentAmountCents) {
+    return errAsync(new InvalidPaymentAmountError())
+  }
+
+  // Check if payment amount is below minPaymentAmountCents
+  if (amount_cents && amount_cents < paymentConfig.minPaymentAmountCents) {
+    return errAsync(new InvalidPaymentAmountError())
+  }
```
LinHuiqing: Should this be combined instead since the error thrown is the same? So like if `< min` or `> max`, then `return errAsync`

LinHuiqing:
Also nit: can we remove the unused imports flagged out by GitHub Action's frontend lint haha

src/app/modules/form/admin-form/__tests__/admin-form.service.spec.ts
tshuli: Sounds good! made the change in [fa5308e](https://github.com/opengovsg/FormSG/pull/5972/commits/fa5308ee0cf1836d0c9ac2ea43ecfb259f8453a1)

src/app/modules/form/admin-form/admin-form.service.ts
tshuli: thanks! in [0bd5f0b](https://github.com/opengovsg/FormSG/pull/5972/commits/0bd5f0b54ff92a79bd2e7c9f686d9e8900de8416)

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
tshuli: for this, i'm thinking to keep the current implementation, because:

1) (I'm assuming we would do the same for L155 too) `centsToDollars` takes a number and returns a string, so we need to cast to Number again for the comparison, which is not ideal
2) `formatCurrency` already handles the formatting for us
3) there is no division error because the denominator is > 0

thoughts?

LinHuiqing: I'm inclined to say that (1) isn't that big of a point since we're doing the same for the input anyway, and it will not introduce any errors.
For 3, I recall there were specific numbers that became long floats then divided by 100 🤔 Let me try to find these numbers again. But yes, you're right that it will be handled by `formatCurrency` (2). 
I was discussing this with @timotheeg previously and we went with the `centsToDollars` and `dollarsToCents` implementation when converting between the values where we use string parsing for conversion as it is cleaner and avoids all the numerical errors that come with division/ multiplication. Referring to [this PR](https://github.com/opengovsg/FormSG/pull/5852), but we had the discussion face-to-face so sadly it's not documented.

Oh also to clarify, my main issue is the `/ 100` part, not the use of `formatCurrency` to string format.

LinHuiqing: Example of division error which won't be fixed by `formatCurrency`:
Screenshot 2023-03-22 at 4 22 36 PM

LinHuiqing: This quirk might not be significant considering there's the IRAS limit, but just something to take note in the future if the amount increases? Also the correctness of the way we process the amounts, I guess.

tshuli: Thanks for this! I can go along [e5d988d](https://github.com/opengovsg/FormSG/pull/5972/commits/e5d988d64301c0d3a9e4dd863734faa3842bf4bc)

src/app/loaders/express/locals.ts
```diff
@@ -64,6 +68,9 @@ const environment = ejs.render(
     var reactMigrationRespondentRolloutStorage = "<%= reactMigrationRespondentRolloutStorage%>"
     var reactMigrationAdminRollout = "<%= reactMigrationAdminRollout%>"
     var reactMigrationAngularPhaseOutDate = "<%= reactMigrationAngularPhaseOutDate%>"
+    // Payment
+    var maxPaymentAmountCents = "<%= maxPaymentAmountCents%>"
+    var minPaymentAmountCents = "<%= minPaymentAmountCents%>"
```
timotheeg: There is no need to convert the values from number to string. These are are env vars where we control the type in the config definition

```suggestion
    var maxPaymentAmountCents = <%= maxPaymentAmountCents%>
    var minPaymentAmountCents = <%= minPaymentAmountCents%>
```

Side note not related to the PR: I'm not sure why we supply the env vars as old-style `var` defined JS, rather than using JSON (which is also typed). We could probably revisit and simplify this at some point.

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -148,24 +149,30 @@ export const PaymentInput = (): JSX.Element => {
...
         validateMin: (val) => {
+          if (minPaymentAmountCents === undefined) return true
           return (
-            Number(val?.trim()) >= minPaymentAmount ||
+            // val is in dollars
+            Number(val?.trim()) >=
+              Number(centsToDollars(minPaymentAmountCents)) ||
             `Please enter a payment amount above ${formatCurrency(
-              minPaymentAmount,
+              Number(centsToDollars(minPaymentAmountCents)),
```
timotheeg: Two comments on approach here:

1. the config value `minPaymentAmountCents` is the reference value. When we want to compare whether an input is valid against it, the input should be transformed/normalized to make a legitimate comparison, but the reference value itself should be left alone.

2. Not critical in this case, but still an improvement to the previous code would be to be consistent with how we deal with amounts. Although we are not doing arithmetic operations here, let's be consistent in the thinking that integer arithmetic is safe, while floating point arithmetic is not. And with that in mind, the safe ways to handle the operations here is to convert the input with `dollarsToCents()` and then compare to the reference value in cents, and not the other way around (of updating everything to dollars)

Same comment for line 167 below.

So Basically:
* No need to do any transformation to `minPaymentAmountCents` and `maxPaymentAmountCents`
* Transform the string input to a cents value before comparing to the min and max thresholds

tshuli: thanks for this! made changes in [663724e](https://github.com/opengovsg/FormSG/pull/5972/commits/663724e45e6f2f4d8a5649106555709ccf657d5a)

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -147,24 +148,28 @@ export const PaymentInput = (): JSX.Element => {
           )
         },
         validateMin: (val) => {
+          if (minPaymentAmountCents === undefined) return true
```
timotheeg, line 151: /justNoting (nothing to do): this defensive checks helps at release (new code hitting old server for env), but feels unnecessary at steady state thereafter 🤔  We should expect the client to be able to load valid env vars to operate, or things wouldn't work generally.

timotheeg (approved):

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -147,24 +148,28 @@ export const PaymentInput = (): JSX.Element => {
...
           return (
-            Number(val?.trim()) >= minPaymentAmount ||
+            // val is in dollars
+            (val && dollarsToCents(val) >= minPaymentAmountCents) ||
             `Please enter a payment amount above ${formatCurrency(
-              minPaymentAmount,
+              Number(centsToDollars(minPaymentAmountCents)),
```
LinHuiqing, line 156: nit: is `formatCurrency(Number(...))` still necessary? 🤔 line 165 too (if no, can remove `formatCurrency` function altogether I think

tshuli: Yes, so that the currency is correctly displayed. Otherwise we could just put `$${centsToDollars(...)}` with `$` manually but using the currency format seems more correct.

LinHuiqing: 🤔 okay! sure, yes makes sense, especially if we have other currencies in the future!

PR #5992, opened 2023-03-27
fix: lowercase collaborator endpoint emails

justynoh:
## Problem
Email inputs on the collaborator endpoints are guaranteed to be trimmed, but not lowercased as brought up in [this comment](https://github.com/opengovsg/FormSG/pull/5946#discussion_r1148771485). This PR adds middleware validation to ensure they are forced to lowercase, a la `auth.middlewares`.

## Solution

Add `.lowercase()` to the middleware validators, which forces the emails to lowercase.

**Breaking Changes** 

## Tests

**Updating collaborators**

- [X] **Updating collaborators via the frontend still works.** Add a collaborator for an existing agency via the frontend. This should still work.
- [X] **Request succeeds on trimmed, lowercased email.** Send a PUT `/:formId/collaborators` with a trimmed, lowercased email such as `[{"write":true,"email":"abc@open.gov.sg"}]`. This should return a 200 OK.
image

- [X] **Request succeeds on trimmed, mixed-case email.** Send a PUT `/:formId/collaborators` with a trimmed, mixed-case email such as `[{"write":true,"email":"dEF@opEn.gOV.sG"}]`. This should return a 200 OK, but with the email cast to lowercase.
image

- [X] **Request fails on untrimmed email.** Send a PUT `/:formId/collaborators` with an untrimmed email such as `[{"write":true,"email":" abc@open.gov.sg"}]`. This should return a 400 Validation failed.
image

**Transfer ownership**

- [x] **Transferring ownership via the frontend still works.** Transfer ownership to a collaborator via the frontend. This should still work.
- [x] **Request succeeds on trimmed, lowercased email.** Send a POST `/:formId/collaborators/transfer-owner` with a trimmed, lowercased email such as `{"email":"abc@open.gov.sg"}`. This should return a 200 OK.
- [X] **Request succeeds on trimmed, mixed-case email.** Send a POST `/:formId/collaborators` with a trimmed, mixed-case email such as `{"email":"abc@opEn.gOV.sG"}`. This should return a 200 OK, but with the email cast to lowercase.
image

- [x] **Request fails on untrimmed email.** Send a POST `/:formId/collaborators` with an untrimmed email such as `{"email":" abc@open.gov.sg "}`. This should return a 400 Validation failed.
image

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -174,12 +174,9 @@ const transferFormOwnershipValidator = celebrate({
   [Segments.BODY]: {
     email: Joi.string()
       .required()
-      .email({
-        minDomainSegments: 2, // Number of segments required for the domain
-        tlds: { allow: true }, // TLD (top level domain) validation
-        multiple: false, // Disallow multiple emails
-      })
-      .message('Please enter a valid email'),
+      .email()
+      .message('Please enter a valid email')
+      .lowercase(),
   },
 })
 
@@ -2354,7 +2351,8 @@ export const handleUpdateCollaborators = [
         email: Joi.string()
           .required()
           .email()
-          .message('Please enter a valid email'),
+          .message('Please enter a valid email')
+          .lowercase(),
         write: Joi.bool().optional(),
         _id: Joi.string().optional(),
       }),
--- a/src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.routes.spec.ts
@@ -2246,7 +2246,7 @@ describe('admin-form.routes', () => {
 
     it('should return 400 when the new owner is not in the database', async () => {
       // Arrange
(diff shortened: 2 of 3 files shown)


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -174,12 +174,9 @@ const transferFormOwnershipValidator = celebrate({
   [Segments.BODY]: {
     email: Joi.string()
       .required()
-      .email({
-        minDomainSegments: 2, // Number of segments required for the domain
-        tlds: { allow: true }, // TLD (top level domain) validation
-        multiple: false, // Disallow multiple emails
-      })
-      .message('Please enter a valid email'),
+      .email()
```
LinHuiqing, line 177: Is there a reason for looser checks for emails here? 😅

justynoh: the explicit settings made here are identical to the defaults, so I just got rid of the additional verbosity. additionally, we don't require these explicitly anywhere else in code

timotheeg (approved):

PR #5998, opened 2023-03-28
feat: add payment description significance

### LINKED ISSUE #5956: Payment description should mention its significance
Payment description should mention:
"This will appear in the payment receipt"

foochifa:
## Problem

We should let our admins know that the description in the payments drawer will be reflected in the receipt.

Closes #5956

## Solution

Add tooltip to payment drawer

And minor changes to UI based on design master figma:
1. Shift description to before pay amount
2. Add divider between description and pay amount
3. Rename description to 'Name'

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests
- [ ] go to form will payment enabled. Ensure that the payment drawer in 'create' has the description 'Name is reflected on payment receipt' below 'Name' field. (This is to ensure UI changes are done and correct)
- [ ] update the payment field of the form (This is to ensure that payment update features are not compromised)

### DIFF
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -208,30 +208,6 @@ export const PaymentInput = (): JSX.Element => {
           <>
             <FormControl
               isReadOnly={paymentsMutation.isLoading}
-              isInvalid={!!errors.display_amount}
-            >
-              <FormLabel isRequired>Payment Amount</FormLabel>
-              <Controller
-                name="display_amount"
-                control={control}
-                rules={amountValidation}
-                render={({ field }) => (
-                  <MoneyInput
-                    flex={1}
-                    step={0}
-                    inputMode="decimal"
-                    placeholder="0.00"
-                    {...field}
-                  />
-                )}
-              />
-              <FormErrorMessage>
-                {errors.display_amount?.message}
-              </FormErrorMessage>
-            </FormControl>
-
-            <FormControl
-              isReadOnly={paymentsMutation.isLoading}
               isInvalid={!!errors.description}
               isRequired
             >
@@ -252,6 +228,30 @@ export const PaymentInput = (): JSX.Element => {
                 })}
               />
               <FormErrorMessage>{errors.description?.message}</FormErrorMessage>
+            </FormControl>
+            <Divider />
+            <FormControl
+              isReadOnly={paymentsMutation.isLoading}
+              isInvalid={!!errors.display_amount}
+            >
+              <FormLabel isRequired>Payment Amount</FormLabel>
+              <Controller
+                name="display_amount"
+                control={control}
+                rules={amountValidation}
+                render={({ field }) => (
+                  <MoneyInput
+                    flex={1}
+                    step={0}
+                    inputMode="decimal"
+                    placeholder="0.00"
+                    {...field}
+                  />
+                )}
+              />
+              <FormErrorMessage>
+                {errors.display_amount?.message}
+              </FormErrorMessage>
             </FormControl>
           </>
         )}
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -62,7 +62,12 @@ const formatCurrency = new Intl.NumberFormat('en-SG', {
   maximumFractionDigits: 2,
 }).format
 
-const DESCRIPTION_TOOLTIP = 'Description will be reflected on payment receipt'
+/**
+ * Description in payment field will be rendered as 'Name' in the Frontend, but kept as description in the backend
+ * This is for design purpose as 'Name' conveys clearer information to the users,
+ * Whilst description will still be used in the backend for consistency with Stripe's API
+ */
+const NAME_TOOLTIP = 'Name will be reflected on payment receipt'
 
 export const PaymentInput = (): JSX.Element => {
   const isMobile = useIsMobile()
@@ -212,10 +217,10 @@ export const PaymentInput = (): JSX.Element => {
               isRequired
             >
               <Flex>
-                <FormLabel>Description</FormLabel>
+                <FormLabel>Name</FormLabel>
                 <Spacer />
                 <Tooltip
-                  label={DESCRIPTION_TOOLTIP}
+                  label={NAME_TOOLTIP}
                   placement="top"
                   textAlign="center"
                 >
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -193,6 +206,30 @@ export const PaymentInput = (): JSX.Element => {
...
+              <Flex>
+                <FormLabel>Description</FormLabel>
+                <Spacer />
+                <Tooltip
+                  label={DESCRIPTION_TOOLTIP}
+                  placement="top"
+                  textAlign="center"
+                >
+                  <Icon as={BxsHelpCircle} aria-hidden ml="0.5rem" />
+                </Tooltip>
```
justynoh: actually just out of curiosity, why are we using a tooltip? could we be using the description field for form labels instead a la this -> 

image

justynoh (approved):
lgtm

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: That does make more sense, @staceytan1998 what do you think?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -51,6 +62,8 @@ const formatCurrency = new Intl.NumberFormat('en-SG', {
   maximumFractionDigits: 2,
 }).format
 
+const DESCRIPTION_TOOLTIP = 'Description will be reflected on payment receipt'
```
LinHuiqing: Why not just follow the terminology on @staceytan1998's design here? Also for the label itself (commented below)
```suggestion
const DESCRIPTION_TOOLTIP = 'Name will be reflected on payment receipt'
```

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -193,6 +206,30 @@ export const PaymentInput = (): JSX.Element => {
 
         {watchedEnabled && (
           <>
+            <FormControl
+              isReadOnly={paymentsMutation.isLoading}
+              isInvalid={!!errors.description}
+              isRequired
+            >
+              <Flex>
+                <FormLabel>Description</FormLabel>
```
LinHuiqing: As above, follow @staceytan1998's terminology for better user experience:
```suggestion
                <FormLabel>Name</FormLabel>
```

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
LinHuiqing: The cognitive load of understanding that the payment "Name" understood by users is the "description" understood by Stripe's API should be borne by us if this is better for users.

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -193,6 +206,30 @@ export const PaymentInput = (): JSX.Element => {
...
+                <Spacer />
+                <Tooltip
+                  label={DESCRIPTION_TOOLTIP}
+                  placement="top"
+                  textAlign="center"
+                >
+                  <Icon as={BxsHelpCircle} aria-hidden ml="0.5rem" />
+                </Tooltip>
+              </Flex>
+              <Textarea
```
LinHuiqing: Change `Textarea` to `Input` as shown in @staceytan1998's design to convey the rough text length expected.

LinHuiqing (commented):
You might wanna remove manual tests that will fail due to other PRs (e.g. #5944) too

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: Primarily because currently in our backend we still refer to it as 'description' I think the change from description to 'name' should be an issue by itself, and should be reflected throughout our system

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: Thanks! [63d3568](https://github.com/opengovsg/FormSG/pull/5998/commits/63d3568194fa66f0c71f18a6bf55cd20f0e19515)

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
LinHuiqing: Oh, my point is that we can keep the "description" terminology internally since its consistent with what stripe API accepts, just change what the user sees on frontend so they can better understand what it means. So no need a separate issue.

foochifa: My other concern is if eventually we do end up having a 'description' or 'sub-description' (currently it was removed by Amit) field as seen in Stacey's previous design. Then we might end up with further confusion.

foochifa: Upon thinking through it, I think I agree, will make the modifications and add comments to highlight the change in definitions from backend schema to frontend display. Thanks @LinHuiqing

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: [aa9d346](https://github.com/opengovsg/FormSG/pull/5998/commits/aa9d3465942aec055d626177146e41c6b4aed2d2)

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: [aa9d346](https://github.com/opengovsg/FormSG/pull/5998/commits/aa9d3465942aec055d626177146e41c6b4aed2d2)

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
staceytan1998: Ok, will change the design! Thanks for bringing it up @justynoh !

foochifa: Thanks for the suggestion @justynoh, [5ec9794](https://github.com/opengovsg/FormSG/pull/5998/commits/5ec9794af2dde44085a7d7444327e025c11cf38a)

LinHuiqing (approved):
Do update the manual tests section to:
- remove mention of tooltip
- clearly state their purpose (i.e. 1. ensure that relevant UI changes have been made; 2. existing functions of payment drawer have not been compromised)

otherwise, lgtm + tested :)

PR #5999, opened 2023-03-28
feat: update encrypt submission controller to use pending submission collection

### LINKED ISSUE #5870: Clone pending submission to submission collection when charge succeeds
* Clone pending submission to submission collection
* Update payment document with:
    * submission id
    * status to charged

steps B6B, B7, B8

justynoh:
## Problem
The original version of our code did not use pending submissions collection and handle payments / pending submissions / payment intents in the correct order. This PR does this.

 Closes #5870

## Solution

The order of actions is payment created --(A)-> pending submission created --(B)-> pending submission saved --(C)-> payment intent created --(D)-> payment saved. The reason for this is primarily that to save a payment, we need the pending submission id and payment intent ids. However we also get a strong guarantee from this that whenever we check the payment document (i.e. our source of truth), we are assured that pending submission and payment intent exists.

What happens if things fail between each action? 

(A) and (B) No damage is done - nothing has been saved to DB. 
(C) We might have floating pending submissions in DB pointing to a bogus payment id which does not exist in DB. But this is not an issue since we can always clean it up. In any case, no payment intent is created, so no money can change hands. 
(D) This is where we might need to think about the right action. We have an open payment intent but no payment document to back it up. Since payment document creation failed, we don't have visibility on this payment intent which may be dangerous (could be scraped and used for attacks). Therefore, we have to proactively roll back the creation by cancelling the payment intent. We should also set an alarm on this branch of the code.

## Deploy Notes

Required downstreaam features before deployment
- updates to individual response page 
- dedicated payment page 

## Other notes

We will need to ensure that the post-submission actions get factored out and done when pending submissions are converted to real submissions:
- Webhooks get sent
- email confirmations get sent

Will create a new ticket for this. EDIT: See #6000 (wow 6000 liao! 🎉)

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -352,7 +352,11 @@ const submitEncryptModeForm: ControllerHandler<
   ) {
     const amount = form.payments_field.amount_cents
     // Step 1: Create payment without payment intent id and pending submission id.
-    if (!amount || amount <= 0) {
+    if (
+      !amount ||
+      amount < paymentConfig.minPaymentAmountCents ||
+      amount > paymentConfig.maxPaymentAmountCents
+    ) {
       logger.error({
         message:
           'Error when creating payment: amount is not a positive integer',
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -13,7 +13,6 @@ import {
   FormSubmissionMetadataQueryDto,
   Payment,
   PaymentChannel,
-  PaymentStatus,
   StorageModeSubmissionDto,
   StorageModeSubmissionMetadataList,
   SubmissionErrorDto,
@@ -24,6 +23,7 @@ import { paymentConfig } from '../../../config/features/payment.config'
 import { createLoggerWithLabel } from '../../../config/logger'
 import { stripe } from '../../../loaders/stripe'
 import getPaymentModel from '../../../models/payment.server.model'
+import { getEncryptPendingSubmissionModel } from '../../../models/pending_submission.server.model'
 import { getEncryptSubmissionModel } from '../../../models/submission.server.model'
 import * as CaptchaMiddleware from '../../../services/captcha/captcha.middleware'
 import * as CaptchaService from '../../../services/captcha/captcha.service'
@@ -62,6 +62,7 @@ import IncomingEncryptSubmission from './IncomingEncryptSubmission.class'
 
 const logger = createLoggerWithLabel(module)
 const EncryptSubmission = getEncryptSubmissionModel(mongoose)
+const EncryptPendingSubmission = getEncryptPendingSubmissionModel(mongoose)
 const Payment = getPaymentModel(mongoose)
 
 // NOTE: Refer to this for documentation: https://github.com/sideway/joi-date/blob/master/API.md
@@ -334,7 +335,7 @@ const submitEncryptModeForm: ControllerHandler<
     }
   }
 
-  const submission = new EncryptSubmission({
+  const submissionContent = {
     form: form._id,
     authType: form.authType,
     myInfoFields: form.getUniqueMyInfoAttrs(),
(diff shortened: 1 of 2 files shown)


LinHuiqing:
Should we be okay with the series of broken states these PRs are going to introduce? :') Or should we loosen the validation checks and add them as the different parts are implemented?

Screenshot 2023-03-28 at 11 24 54 AM

justynoh:
imo it's okay for now, we're developing a bunch of things in parallel anyway and trying to fix things up as we go along

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -334,102 +335,92 @@ const submitEncryptModeForm: ControllerHandler<
...
+    // Step 1: Create payment without payment intent id and pending submission id.
+    if (!amount || amount <= 0) {
       logger.error({
         message:
-          'Error when creating payment intent, amount is not a positive integer',
-        meta: {
-          submissionId,
-          ...logMeta,
-        },
+          'Error when creating payment: amount is not a positive integer',
```
LinHuiqing: The limits of amount should be bound by the env vars `PAYMENT_MAX_PAYMENT_AMOUNT_CENTS` and `PAYMENT_MIN_PAYMENT_AMOUNT_CENTS` (implemented in #5972). These could be used to provide a better check than a non-negative check?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -334,102 +335,92 @@ const submitEncryptModeForm: ControllerHandler<
...
+    })
+    const paymentId = payment.id
+
+    // Step 2: Create and save pending submission.
+    const pendingSubmission = new EncryptPendingSubmission({
+      ...submissionContent,
+      paymentId,
+    })
+
+    try {
```
LinHuiqing, line 382: Using `neverthrow` might be better than all these try/catch statements?

LinHuiqing:
> imo it's okay for now, we're developing a bunch of things in parallel anyway and trying to fix things up as we go along

Hmm I'm just thinking of this from service recovery point of view, like if any of these cause the app to break when we deploy, it will mean that these changes need to rolled back altogether to get to the previous usable state. But if we're okay with this, better add a note on the expected app behaviour at this point of time and link this up to related PRs upstream (e.g. #5944) and maybe downstream.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Maybe this can be done in future PRs too, but it will be tech debt :') Just a comment since we're not just in "hackathon" mode anymore, right?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
justynoh: Oh you're right! Thanks for that. Will update

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
justynoh: Agreed. The rest of the controller uses the try/catch style though, and I didn't want to introduce neverthrow here out of the blue. The entire controller is too long and needs refactoring anyway so I figured we can refactor the entire thing once we merge into develop in beta. It's also a comment I added in #6000, feel free to tag on to that ticket!

justynoh:
Agreed! No PR open for it yet but will note it in deploy notes.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: man sounds messy already :') but yup okay!

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
justynoh: done! [bc9e3a2](https://github.com/opengovsg/FormSG/pull/5999/commits/bc9e3a2e7975467e5d2bf5277a617d4c280b5bf4)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -334,102 +335,96 @@ const submitEncryptModeForm: ControllerHandler<
...
+      amount > paymentConfig.maxPaymentAmountCents
+    ) {
       logger.error({
         message:
-          'Error when creating payment intent, amount is not a positive integer',
-        meta: {
-          submissionId,
-          ...logMeta,
-        },
+          'Error when creating payment: amount is not a positive integer',
```
LinHuiqing: You forgot to change the error message accordingly 😅

justynoh: [91621ad](https://github.com/opengovsg/FormSG/pull/5999/commits/91621ad008f3cab0db82380699c8dea88d77e9b5)

KenLSM (approved):

PR #6001, opened 2023-03-28
feat: dedicated payment page

### LINKED ISSUE #5932: Create dedicate payment page with unique url
Can be used for #5806 

Dependency
- #5872
- #5870

KenLSM:
## Problem

Closes #5932

## Solution

**Breaking Changes** 

    - Requires Backend to be deployed due to calls to new api endpoints (`/payment/*`)

**Features**:
- Introduced of new dedicated URL for payment page under `/:formid/payment/:paymentId`.
- Decoupled `submissionId` as the unique identifier from FE. It is now referred to as `paymentId` on FE.
  - BE utilizes `paymentIntentId` issued from `stripe` under the hood
-  Added new API,  `/payment/:paymentId/getinfo`, to query for payment info that's necessary to show payment element.
-  Support api `/payment/:paymentId/getreceipt`
-  Added polling for stripe payment status
-  Added loading state when payment from stripe has been confirmed, but BE has not received receipt
-  Updated redirection from `proceed to pay` to new payment url

**Improvements**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

Payment
- [ ] Able to handle idempotency in payment
  - [ ] Attempt to make payment on two different tabs of the same payment url, it should reject the second successful payment. It should show the receipt block.

---

## Additional Notes

**Folder Changes**
This PR also made some folder changes to group payment related FE Components into `features/payment` folder. Stripe-specific components are placed as a child to the payment folder.

**Reference Id Changes**
As we are migrating to having a dedicated page, the `client_secret` shouldn't be injected through `queryParams`, thus we will need to use a different `identifier` for FE and BE to communicate which payment it is. FE should agnostic to which `identifier` it is, so long BE is consistent.

As such, `payment_intent_id` that we obtained from `stripe.createPaymentIntent` is chosen. It is a unique key that is idempotent, and we can extend it to use `UUID` (or other forms of identifiers) post-mvp if we decide to integrate with other `payment_channels`.

**Receipt URL fetched from DB**
Receipt URL has been changed from being fetch from stripe at the moment of request to be synced with the DB. i.e., we will only return the receipt download URL `iif` DB has it recorded, even though it is clearly retrievable from BE through `stripe.charge.retrieve`. This is due to the implementation of `confirmSubmission` flow where we will treat the update of `payment` + `copyToSubmission` as a single atomic transaction. If the copy fails, the responder will not be able to download the receipt as well, both are regarded as incomplete submission attempt.


justynoh:
Should we have the `frontend/src/features/payment` folder actually be in `frontend/src/features/public-form`? All those components look like they only belong on the public-facing side of the app. Additionally, the payment page is wrapped in a public form provider anyway, so we won't be able to use it anywhere else. Maybe we can have a file `PublicFormPaymentPage.tsx` in the main folder, and then stuff the other files in `components/FormPayment` or something like that.

Just seemed like if we're importing a bunch of stuff from `~/features/public-form` into this file, it's basically asking to go into the `~features/public-form` folder haha

frontend/src/constants/routes.ts
```diff
@@ -43,3 +43,4 @@ export const ACTIVE_ADMINFORM_RESULTS_ROUTE_REGEX = new RegExp(
   `${ADMINFORM_ROUTE}/([a-fA-F0-9]{24})/${ADMINFORM_RESULTS_SUBROUTE}(/${RESULTS_FEEDBACK_SUBROUTE})?/?`,
   'i',
 )
+export const PAYMENT_PAGE_SUBROUTE = 'payment/:paymentPageId'
```
justynoh: nit: should we leave a comment here about what the `paymentPageId` represents? so that we won't be confused later on about what id is being passed here, and in case we want to update it to something else

frontend/src/features/payment/stripe/StripeElementWrapper.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const { data } = useGetPaymentStatusFromStripe({
+    clientSecret: paymentInfoData.client_secret,
+    stripe,
+    refetchKey,
+  })
+
+  const viewType = getPaymentViewType(data?.paymentIntent?.status)
+
+  switch (viewType) {
+    case 'invalid':
```
justynoh: For the view names, should we name them more similarly to states states instead? for instance, `incomplete`, `cancelled`, `succeeded`, etc. Or, at least, should we make it more internally consistent (e.g. `invalid` is an adjective but `receipt` is a noun)

src/app/modules/payments/__tests__/payment.service.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import getPaymentModel from 'src/app/models/payment.server.model'
+
+import dbHandler from 'tests/unit/backend/helpers/jest-db'
+
+import { findPaymentByPaymentIntentId } from '../payments.service'
+
+const Payment = getPaymentModel(mongoose)
+describe('payments.service', () => {
+  beforeAll(async () => await dbHandler.connect())
```
justynoh: should this be `beforeEach`, if we have a test that closes the connection to the database?

src/app/modules/payments/__tests__/payment.service.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    beforeEach(async () => {
+      await dbHandler.clearCollection(Payment.collection.name)
+    })
+    afterEach(() => jest.clearAllMocks())
+    it('should return with error if payment intent is not found', async () => {
+      const result = await findPaymentByPaymentIntentId(
+        new ObjectId().toHexString(),
+      )
+      expect(result.isErr()).toBeTrue()
+    })
```
justynoh: should we add some mock data and test success cases too?

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -561,4 +579,53 @@ describe('stripe.service', () => {
...
+      // Act
+      const result = await StripeService.linkStripeAccountToForm(mockForm, {
+        accountId: expectedAccountId,
+        publishableKey: 'publishableKey',
+      })
+
+      // Assert
+      expect(mockForm.addPaymentAccountId).toHaveBeenCalledTimes(0)
+      expect(result._unsafeUnwrap()).toBe(expectedAccountId)
+    })
```
justynoh: should we rename this test? perhaps something like `"should return existing account information when called with new account to be linked"`

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -561,4 +579,53 @@ describe('stripe.service', () => {
       expect(result._unsafeUnwrapErr()).toBeInstanceOf(DatabaseError)
     })
   })
+  describe('linkStripeAccountToForm', () => {
+    it('should call func to attach payment account information', async () => {
+      // Arrange
+      const EncryptedForm = getEncryptedFormModel(mongoose)
```
justynoh: nit: should we put this in the global file scope, so that future tests needing the encrypted form model won't have to get the model again?

src/app/modules/payments/payments.service.ts
```diff
@@ -29,7 +29,12 @@ export const findPaymentById = (
   session?: mongoose.ClientSession,
 ): ResultAsync<IPaymentSchema, PaymentNotFoundError | DatabaseError> => {
   return ResultAsync.fromPromise(
-    PaymentModel.findById(paymentId).session(session ?? null),
+    PaymentModel.findById(
+      paymentId,
+      null,
+      // readPreference from transaction isn't respected, thus we are setting it on operation
+      session ? { readPreference: 'primary' } : null,
+    ).session(session ? session : null),
```
justynoh, line 37: 😢

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -522,19 +522,19 @@ const submitEncryptModeForm: ControllerHandler<
       })
     }
 
+    // Attach required payment configs if client secret is present. Otherwise,
+    // client will display error message. We still return 200 OK because the
+    // state is recoverable.
+    const paymentData = paymentClientSecret
+      ? { paymentIntentId: paymentIntent?.id }
+      : null
+
```
justynoh: Is it necessary still to check on whether the client secret is present? I feel like we could just return the payment intent id if it exists.

src/app/routes/api/v3/payments/payments.routes.ts
```diff
@@ -15,19 +15,19 @@ PaymentsRouter.get('/stripe')
...
- * @route GET /receipt/:formId/:submissionId/download
+ * @route GET /payments/receipt/:formId/:submissionId/download
  *
  * @returns 200 with receipt URL exists
  */
-// TODO: consider rate limiting this endpoint #5924
 PaymentsRouter.route(
-  '/receipt/:formId([a-fA-F0-9]{24})/:submissionId([a-fA-F0-9]{24})/download',
+  // '/receipt/:formId([a-fA-F0-9]{24})/:submissionId([a-fA-F0-9]{24})/download',
+  '/receipt/:formId([a-fA-F0-9]{24})/:paymentId/download',
```
justynoh: nit: should we delete the extraneous comment, update the route in the documentation and update the `@returns` with the failure cases?

justynoh: question: not related to your PR, but this is not quite RESTful right? should it actually be `payments/:formId/:paymentId/download-receipt` or something like that?

src/app/routes/api/v3/payments/payments.routes.ts
```diff
@@ -36,3 +36,9 @@ PaymentsRouter.route(
 PaymentsRouter.route('/stripe/callback').get(
   StripeController.handleConnectOauthCallback,
 )
+
+/**
+ * returns clientSecret and publishableKey from paymentId
+ * @route /payments/:paymentId/getinfo
+ */
```
justynoh: nit: should we add `@returns` documentation to this route?

src/app/modules/pending-submission/pending-submission.service.ts
```diff
@@ -0,0 +1,51 @@
...
+          action: 'findSubmissionById',
+          pendingSubmissionId,
+        },
+        error,
+      })
+      return new DatabaseError(getMongoErrorMessage(error))
+    },
+  ).andThen((submission) => {
+    if (!submission) {
+      console.log({ pendingSubmissionId })
```
justynoh: should this be a `logger.error`?

src/app/modules/payments/stripe.controller.ts
```diff
@@ -388,3 +420,87 @@ export const handleConnectOauthCallback = [
...
+        stripeFullIntentObj,
+        publishableKey: form.payments_channel?.publishable_key || '',
+      }))
+    })
+    .map(({ stripeFullIntentObj, publishableKey }) => {
+      return res.status(StatusCodes.OK).json({
+        client_secret: stripeFullIntentObj.client_secret || '',
+        publishableKey,
+      })
+    })
```
justynoh:
```suggestion
      ).map((paymentIntent) => {
      return res.status(StatusCodes.OK).json({
        client_secret: paymentIntent.client_secret || '',
        publishableKey: form.payments_channel?.publishable_key ?? '',
      })
    })
    })
```

frontend/src/features/payment/stripe/components/StripePaymentBlock.tsx
```diff
@@ -77,18 +74,26 @@ const StripeCheckoutForm = ({
...
-      // Show error to your customer (for example, payment details incomplete)
       setStripeMessage(result.error.message)
     } else {
-      setStripeMessage('')
-      // Your customer will be redirected to your `return_url`. For some payment
-      // methods like iDEAL, your customer will be redirected to an intermediate
-      // site first to authorize the payment, then redirected to the `return_url`.
+      // In the event that customer is not on a payment that has a redirected flow,
+      // we will trigger a payment status refetch
+      triggerPaymentStatusRefetch()
```
justynoh: should we still keep the `setStripeMessage('')`, so that we keep the state consistent?

frontend/src/features/payment/stripe/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { PaymentIntent } from '@stripe/stripe-js'
+
+export const getPaymentViewType = (
+  status: PaymentIntent.Status | undefined,
+) => {
+  if (!status) return 'invalid'
+  if (['succeeded'].includes(status)) return 'receipt'
+  if (['canceled'].includes(status)) return 'canceled'
+  return 'payment'
+}
```
justynoh: wondering if these states should be extracted to an enum

src/app/modules/payments/stripe.controller.ts
```diff
@@ -227,66 +237,88 @@ export const handleStripeEventUpdates = [
...
+      //       })
+      //       return res
+      //         .status(StatusCodes.INTERNAL_SERVER_ERROR)
+      //         .json({ message: 'Missing receipt url' })
+      //     }
+      //     return res.status(StatusCodes.OK).json({ isReady: true })
+      //   })
+      //   .mapErr((error) => {
+      //     return res.status(StatusCodes.NOT_FOUND).json({ message: error })
+      //   })
```
justynoh, line 291: discussion point: do we need to have a manual sync flow? or should we be polling and ask the user to refresh / click a button to refetch? i feel like the latter is better haha

justynoh (commented):
actually I'm not too sure why we can't just use the paymentId directly throughout the whole flow. switching language between the model's paymentId and the paymentId from the frontend (which is actually the payment intent id to the be) is quite confusing imo. Would propose to use the mongodb payment id for the whole flow, and return the payment intent id from the backend to the frontend when the state requires use of the payment intent id (i.e. only in the payment-incomplete state)

KenLSM:
> Should we have the `frontend/src/features/payment` folder actually be in `frontend/src/features/public-form`? All those components look like they only belong on the public-facing side of the app. Additionally, the payment page is wrapped in a public form provider anyway, so we won't be able to use it anywhere else. Maybe we can have a file `PublicFormPaymentPage.tsx` in the main folder, and then stuff the other files in `components/FormPayment` or something like that.
> 
> Just seemed like if we're importing a bunch of stuff from `~/features/public-form` into this file, it's basically asking to go into the `~features/public-form` folder haha

Yeah! Good catch. I think I had tunnel-visioned during development of this feature!

Edit:
Updated in c7cd457f

src/app/modules/payments/stripe.controller.ts
KenLSM: Yeah there's a polling happening (provided by react-query).
Some other details, the BE is returning `404` which react-query will recognize as a failure and retry until it succeeds.

I'm also on the bus for polling initially, but iirc @timotheeg mentioned that we should trigger a sync.

frontend/src/features/payment/stripe/components/StripePaymentBlock.tsx
KenLSM: It's reset on [L65](https://github.com/opengovsg/FormSG/pull/6001/files/ff6bab87ece2f0d878b03c8d145fd8876e3dc14c#diff-abbd02937e4bd0ccb8beaacd056226b8e3b9bfa0faf9f18db916e04710dd3dccR65).

I was thinking that we should clear the error message on `Submit Payment` button press so that the previous error message, if any, would be cleared. If not, the responder will be still seeing the previous error message even though they've updated their payment information and clicked `Submit Payment`.

frontend/src/constants/routes.ts
KenLSM: I was thinking that this particular Id should be agnostic/decoupled on FE. I.e., for stripe payments, it could be `pi_xyz`, and for other possible integrations like paynow, it could be `pn_xyz`.

BE should be sensitive to what this value is, and should be consistent in referencing it. I.e., if we need to change the `paymentPageId` from stripe's payment intent, to `UUID` or Mongo's `objectId` under the hood, then we should only change what BE passes to FE, and how BE ~reference~understands this field. FE will treat this as a reference variable, and doesn't have to know how to manipulate or what it actually is.

Wdyt?

frontend/src/features/payment/stripe/StripeElementWrapper.tsx
KenLSM: Yeah makes sense. 

Will go with `payment -> incomplete`, and `receipt -> succeeded`.

KenLSM: 9bba127f

frontend/src/constants/routes.ts
justynoh: > we should only change what BE passes to FE, and how BE understands this field. FE will treat this as a reference variable, and doesn't have to know how to manipulate or what it actually is.

I agree with the principle, but I'm not sure I see what advantage using the payment intent id over the payment id gives us in the fe. as an example, in the implementation we pass the payment's payment intent id to the fe from the encrypt controller and receive the payment intent id in e.g. `getPaymentInfo`. But the fe wouldn't see anything different from if we had passed the mongo object id instead. Is there somewhere in the code that, if we were to only pass the objectid, it would require some casing behavior in the fe?

frontend/src/features/payment/stripe/components/StripePaymentBlock.tsx
justynoh: good point! i missed that line, thanks for the clarification 👍

frontend/src/features/payment/stripe/utils.ts
KenLSM: 9bba127

frontend/src/constants/routes.ts
KenLSM: I just realized that I was arguing about a different thing 🙈 . Right, it would make the flow simpler if BE uses `paymentId` over `paymentIntentId` until the point where FE needs to communicate with `stripe`.

src/app/modules/pending-submission/pending-submission.service.ts
KenLSM: Removed in d994a723. Not managing to find the submission could be a non-error (user fuzzing the ids).

src/app/modules/payments/stripe.controller.ts
KenLSM: c0b6b18d

src/app/routes/api/v3/payments/payments.routes.ts
KenLSM: Yeah it's not that RESTful. `/payments/:formId/:paymentId/*` clashes with `/payments/:paymentId` too.

We might need to go with these if we don't drop `formId`
- `/payments/receipt/download?formid=formId&paymentid=paymentId`
- `/payments/receipt/status?formid=formId&paymentid=paymentId`

Or we could drop `formId`
- `/payments/:paymentId/receipt/download`
- `/payments/:paymentId/receipt/status`

Wdyt?

frontend/src/constants/routes.ts
KenLSM: Switched BE reference to `paymentId` in a9b31057

src/app/modules/payments/__tests__/payment.service.spec.ts
KenLSM: Hoisted in a9b31057

src/app/modules/payments/__tests__/payment.service.spec.ts
KenLSM: Addressed in a9b31057

src/app/modules/payments/__tests__/payment.service.spec.ts
KenLSM: Added in a9b31057, also swapped the tests to be testing on `findPaymentById` since `findPaymentByPaymentIntentId` doesn't anymore

src/app/modules/payments/__tests__/stripe.service.spec.ts
KenLSM: Hoisted in a9b31057

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
KenLSM: Cleaned up in a9b31057

src/app/routes/api/v3/payments/payments.routes.ts
KenLSM: a9b31057

src/app/modules/payments/__tests__/stripe.service.spec.ts
KenLSM: 7838a176

frontend/src/constants/routes.ts
justynoh: since we're using payment ids now, should we just change this param to `:paymentId` instead and use that throughout the code?

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -273,16 +275,20 @@ export const PublicFormProvider = ({
...
+                    if (paymentData) {
+                      navigate(`/${formId}/payment/${paymentData.paymentId}`)
+                      return
+                    }
                     setSubmissionData({
                       id: submissionId,
                       timestamp,
-                      paymentClientSecret,
-                      paymentPublishableKey,
+                      paymentData,
```
justynoh: is it necessary to set payment data here? not sure why `SubmissionData` needs to be augmented with this additional field if it isn't meant for use in the payment flow. afaik, this SubmissionData is only used for display on the endpage

frontend/src/features/public-form/components/FormPaymentPage/queries.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const useGetPaymentInfo = (
+  paymentId: string,
+): UseQueryResult<GetPaymentInfoDto, ApiError> => {
+  return useQuery<GetPaymentInfoDto, ApiError>(
+    paymentId,
+    () => getPaymentInfo(paymentId),
+    { suspense: true },
+  )
+}
```
justynoh: nit: should we extract the query keys out to a query key handler for better documentation? a la 

```typescript
// frontend/src/features/admin-form/settings/queries.ts
export const adminFormSettingsKeys = {
  base: [...adminFormKeys.base, 'settings'] as const,
  id: (id: string) => [...adminFormSettingsKeys.base, id] as const,
  payment_channel: (id: string) =>
    [...adminFormSettingsKeys.id(id), 'payment_channel'] as const,
  payment_field: (id: string) =>
    [...adminFormSettingsKeys.id(id), 'payment_field'] as const,
}

File: frontend/src/features/public-form/components/FormPaymentPage/stripe/StripeElementWrapper.tsx [hunk 18]
```diff
@@ -0,0 +1,138 @@
...
+    bg="white"
+    w="100%"
+  >
+    {children}
+  </Box>
+)
+/**
+ * Handles decision to render StripePaymentModal or StripeReceiptModal
+ * @returns
+ */
```
justynoh, line 67: supernit: should we update the documentation?

frontend/src/features/public-form/components/FormPaymentPage/stripe/StripeElementWrapper.tsx
```diff
@@ -0,0 +1,138 @@
...
+              paymentPageId={paymentPageId}
+            />
+          </PaymentBox>
+        </>
+      )
+    default: {
+      // Force TS to emit an error if the cases above are not exhaustive
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const exhaustiveCheck: never = viewStates
+      throw new Error(`Undefined view type: ${viewStates}`)
```
justynoh, line 130: hahaha this is quite cool. but maybe when we add a processing state, we can default to that.

frontend/src/features/public-form/components/FormPaymentPage/stripe/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  Pending,
+}
+export const getPaymentViewStates = (
+  status: PaymentIntent.Status | undefined,
+): PaymentViewStates => {
+  if (!status) return PaymentViewStates.Invalid
+  if (['succeeded'].includes(status)) return PaymentViewStates.Succeeded
+  if (['canceled'].includes(status)) return PaymentViewStates.Canceled
+  return PaymentViewStates.Pending
+}
```
justynoh, line 16: Do we also need a `Processing` state, when the payment intent status is processing? as per 
image

src/app/modules/payments/__tests__/payment.service.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+  afterEach(async () => await dbHandler.closeDatabase())
+
+  describe('findPaymentById', () => {
+    beforeEach(async () => {
+      await dbHandler.clearCollection(Payment.collection.name)
+    })
+    afterEach(() => jest.clearAllMocks())
+
+    it('should return with without error if payment id is found', async () => {
```
justynoh: nit: `with without` 👀

src/app/modules/payments/stripe.controller.ts
```diff
@@ -388,3 +423,98 @@ export const handleConnectOauthCallback = [
...
+                formResponseMode: form.responseMode,
+              },
+            })
+            return errAsync(
+              new ResponseModeError(
+                FormResponseMode.Encrypt,
+                form.responseMode,
+              ),
+            )
+          }
```
justynoh: nit: would it be better to use `checkFormIsEncryptMode` instead, so that we can do `.andThen(checkFormIsEncryptMode)`?

src/app/modules/payments/stripe.controller.ts
```diff
@@ -388,3 +423,98 @@ export const handleConnectOauthCallback = [
...
+          break
+        case PendingSubmissionNotFoundError:
+          res.status(StatusCodes.NOT_FOUND)
+          break
+        case StripeFetchError: // fall-through
+        case PaymentAccountInformationError: // fall-through
+        default:
+          res.status(StatusCodes.INTERNAL_SERVER_ERROR)
+          break
+      }
```
justynoh: supernit: (forgot if i mentioned already but) would be nice to extract this to a `mapRouteErr` util, can be reused in other controllers too

justynoh (approved):
lgtm but mostly nits

src/app/routes/api/v3/payments/payments.routes.ts
justynoh: i would be in favor of removing `formId`, once the payment is created we should reference that directly. we can track this in a separate issue, this PR getting too big haha. we can create a new issue for api refactor!

src/app/modules/payments/stripe.controller.ts
justynoh: hmm, ok, we can track this in a separate issue?

KenLSM: Opened here opengovsg/formsg-private/issues/143

frontend/src/features/public-form/components/FormPaymentPage/stripe/utils.ts
KenLSM: I'm thinking that we don't have a processing state yet (until we integrate `paynow` or non-`card` processing types), and thus, not requiring to handle this state/design 🤔

src/app/modules/payments/stripe.controller.ts
KenLSM: Cleaned up in 57585489

src/app/modules/payments/__tests__/payment.service.spec.ts
KenLSM: Fixed typo in 57585489

frontend/src/features/public-form/PublicFormProvider.tsx
KenLSM: Removed in 57585489

frontend/src/constants/routes.ts
KenLSM: Renamed in 57585489

frontend/src/features/public-form/components/FormPaymentPage/queries.ts
KenLSM: Added key factory in 57585489

frontend/src/features/public-form/components/FormPaymentPage/stripe/utils.ts
justynoh: i feel like for safety's sake, we should just handle it (since one of the valid states for payment_intent is processing). nonetheless, doesn't have to be done now. we can open a ticket and deal with it later

KenLSM: Tracking in #6043

PR #5917, opened 2023-03-10
feat: add respondent email to receive receipts

LinHuiqing:
## Problem

Currently, users have no way of confirming that their submission and/or payment has been received, other than manually contacting the admins or if the admins follow up with their form response.

Closes #5814, closes #5897, closes #5898

## Solution

Add an email input field at the end of the payment form for respondents to input the email that will receive any receipts.

**Breaking Changes** 

- [x] Yes - this PR contains breaking changes
    - When a new UI tries to send a `:formId/submission/encrypt` POST request to an old server to submit a form with payment, the body would include the `paymentReceiptEmail` which is unexpected by the old server, causing a validation error. 
    - Fortunately, the impact of this is will be limited as this only affect forms with payments, and only a few users from JTC has the beta flag required to unlock the payment feature.

**Features**:

- Upcoming payment preview at the end of the form
- Email field in payment preview to input email for receipts
- Use email input in payment preview as email address to receive stripe receipt

**Improvements**:

- Updated copy for:
  - button at the end of the form
  - payment modal

## Before & After Screenshots

**BEFORE**:

Form with payment enabled:

Screenshot 2023-03-10 at 5 14 26 PM

Payment modal:

Screenshot 2023-03-10 at 5 14 33 PM

**AFTER**:

Form with payment enabled:

Screenshot 2023-03-10 at 5 15 21 PM

Payment modal:

Screenshot 2023-03-10 at 5 15 28 PM

Payment Flow with Email Receipt:

https://user-images.githubusercontent.com/37061143/224479954-fb88ad37-1952-40f4-aa3f-74e485470c8c.mp4

## Tests

Can be checked with stripe's test credentials:
- [ ] Check that the payment preview is the same as screenshot above.
- [ ] Check that the payment modal is the same as screenshot above.
- [ ] Check that the payment block is the same as screenshot above.

Requires stripe's live credentials to run:
- [ ] 1. Input an email address that has an inbox you have access to in the payment preview portion
       2. Click "proceed to pay" at the end of the form and then on the payment modal.
       3. On the payment block, key in card details and click "Submit payment".
       4. If payment succeeds, you should receive a receipt email in the inbox of the email address you entered.

Issue #5977, opened 2023-03-22
Move add and remove payment account methods in IFormSchema to IEncryptedFormSchema


foochifa:
Currently our add and remove payment account methods in `src/types/form.ts` is still nested in IFormSchema.

We should shift it to IEncryptedFormSchema, and update our IPopulatedEncryptedForm

Notes:
1. Previously when trying to use an interface for IEncryptedFormSchema, run into errors trying to extend IFormSchema and IEncryptedForm
2. Currently IPopulatedEncryptedForm does not extend IEncryptedFormSchema (it extends IPopulatedForm & IEncryptedForm)

Issue #1844, opened 2021-05-10
[API-post] Delete unused endpoints after checking logs.


karrui:
See [refactor design document](https://docs.google.com/document/d/1dxQciANUILj1mP8pZb-BiwHsgiX_FH5JiMem5cSkzQY/edit#heading=h.199pf2fyqba2) for list of endpoints

Note that this is partially blocked by #1845

PR #6003, opened 2023-03-28
feat: add otp verification to payment contact field

### LINKED ISSUE #5964: Make email validation mandatory for payer email field (review)
We need email validation to make sure that:
Payer can receive emails (mailbox is not full)
Payer have access to their email box
We are not under fraud attacks

LinHuiqing:
## Problem

Email fields for payment currently do not verify emails before accepting them to be tagged to payments. This may be an issue when we need to reconcile payments and there is no identifier for us to match payments to users. Verified emails provide us a way to do so.

Closes #5964

## Solution

Modify existing email verification mechanisms in form fields to use for payments.

**Breaking Changes** 

- [x] No

**Features**:

- OTP verification on payment contact email field! 
- ... and tests! 

## Before & After Screenshots

**BEFORE**:

(taken from https://github.com/opengovsg/FormSG/pull/5917, notice the lack of a "Verify" button beside the email field)

https://user-images.githubusercontent.com/37061143/229749350-82839a58-d98a-4792-9950-c8c18617b39e.mp4

**AFTER**:

https://user-images.githubusercontent.com/37061143/229748863-397bfce7-f5a7-4826-8fd1-5e24341a3b40.mov

## Tests

To test the expected behaviour on payment enabled form after feature has been implemented (on a payment form with no other verifiable fields):
1. - [ ] If payment contact field is not filled in, user should not be able to proceed with payment.
2. When “Verify” in any verifiable email field component in the form is pressed,
  - [ ] In the DB, a transaction should be created in the `verifications` table. 
    - [ ] If payment is enabled on the form, this transaction should include the `paymentField` key with the default values of `IVerificationFieldSchema` if the verifiable email field to be verified is not the payment one. 
    - [ ] If it is the payment verifiable email field, the value stored in `paymentField` key should be `IVerificationFieldSchema` but with the values populated according to the otp hash calculated.
  - [ ] On the frontend, the verification modal should open up for the user to key in the OTP.
  - [ ] An email with the OTP should be sent to the entered email.
3. Press "Resend OTP" on the payment contact email field once possible (usually after 60s).
  - [ ] OTP-prefix should be updated accordingly.
  - [ ] A new transaction should be added to the `verifications` table in the DB.
  - [ ] An email with the new OTP should be sent to the entered email.
4. Enter a wrong OTP in the payment contact field and click "Submit".
  - [ ] The verification box should let the user know that the OTP is wrong.
5. Enter the appropriate OTP in the payment contact field and click "Submit".
  - [ ] The verification box should collapse and show that the email is verified as per normal verifiable email fields.
6. - [ ] Proceed to pay in the payment element. If the app is given the live stripe credentials and the payment is successful, a stripe payment receipt should be sent to the entered email.

- [ ] Repeat tests 1-6 for a payment form with other verifiable fields.
- [ ] Repeat tests 2-6 for verifiable fields in a storage non-payment form.
- [ ] Repeat tests 2-6 for verifiable fields in an email form.
- [ ] Repeat steps 2-6 on a storage form with both payment and 2 verifiable fields in the form (1 email + 1 mobile)


LinHuiqing:
Not sure if [3bc624e](https://github.com/opengovsg/FormSG/pull/6003/commits/3bc624edcdeee0fca2b839861e519be64b476a0c) is the best way to refactor related code, will relook at it at the end!

LinHuiqing:
🤔 I should probably refactor `handleGenerateFormOtp` and `handleGeneratePaymentOtp` too, will do so later when functionality is up!

LinHuiqing:
There's a lot of overlap in code for `_handleGenerateFormOtp` and `_handleGeneratePaymentOtp` and the related tests, but I'm not quite sure how we can get around this since we will want different endpoints for them? 
Okay one solution will be to use the same endpoint `/:formId([a-fA-F0-9]{24})/fieldverifications/:transactionId([a-fA-F0-9]{24})/fields/:fieldId([a-fA-F0-9]{24})/otp/generate` but loosen the restrictions on field id in `:fieldId([a-fA-F0-9]{24})`, but not sure if this is worse / better than how it currently is? Thoughts? :')

LinHuiqing:
Side note - should tackle https://github.com/opengovsg/FormSG/issues/1844 and kill deprecated endpoints! At a glance on datadog, I don't see them being used, but will do a more in-depth check in the future.

LinHuiqing:
> There's a lot of overlap in code for `_handleGenerateFormOtp` and `_handleGeneratePaymentOtp` and the related tests, but I'm not quite sure how we can get around this since we will want different endpoints for them? Okay one solution will be to use the same endpoint `/:formId([a-fA-F0-9]{24})/fieldverifications/:transactionId([a-fA-F0-9]{24})/fields/:fieldId([a-fA-F0-9]{24})/otp/generate` but loosen the restrictions on field id in `:fieldId([a-fA-F0-9]{24})`, but not sure if this is worse / better than how it currently is? Thoughts? :')

I've gone ahead to reuse the existing endpoints and reduced code addition by 300 lines yay! We can always undo the commit if we're uncomfortable with this.

src/types/form.ts
```diff
@@ -105,6 +105,7 @@ export type PickDuplicateForm = Pick<
 export interface IFormSchema extends IForm, Document, PublicView<PublicForm> {
   form_fields?: Types.DocumentArray<FormFieldSchema> | FormFieldSchema[]
   form_logics?: Types.DocumentArray<FormLogicSchema> | FormLogicSchema[]
+  payments_field?: { enabled?: boolean }
```
justynoh: Is this necessary to add here? The payments field does not exist on non-encrypt mode forms, so I'm thinking this may not be the most appropriate. The `payments_field` prop is in `IEncryptedFormSchema`, so if `payments_field` needs to be accessed from an `IFormSchema`, could we access it after having been type guarded by a call to `checkFormIsEncryptMode`?

src/types/verification.ts
```diff
@@ -26,15 +26,19 @@ export interface IVerification {
...
   transactionId: string
-  fieldId: string
   hashedOtp: string
   signedData: string
 }
 
+export type UpdateFormFieldData = UpdatePaymentFieldData & {
+  fieldId: string
+}
+
```
justynoh: nit: for clarity, I wonder if it will be less confusing for future devs if we don't extend `UpdateFormFieldDto` from `UpdatePaymentFieldData`, as the former doesn't really sound extensible from the latter. suggestion:


```typescript
type UpdateFieldData = {
  transactionId: string
  hashedOtp: string
  signedData: string
}

export type UpdatePaymentFieldData = UpdateFieldData
export type UpdateFormFieldData = UpdateFieldData & {
  fieldId: string
}

File: src/types/verification.ts [hunk 3]
```diff
@@ -48,6 +52,10 @@ export interface IVerificationSchema
    * Extracts non-sensitive fields from a transaction
    */
   getPublicView(): PublicTransaction
+  /**
+   * Retrieves payment contact field in a transaction, or undefined if not found
+   */
+  getPaymentContactField(): IVerificationFieldSchema | undefined
```
justynoh: nit: should we rename `getField` to `getFormField`, and `getPaymentContactField` to `getPaymentField` for language consistency as well?

src/types/verification.ts
```diff
@@ -75,22 +83,39 @@ export interface IVerificationModel extends Model<IVerificationSchema> {
...
-    updateData: UpdateFieldData,
+  updateHashForFormField(
+    updateData: UpdateFormFieldData,
+  ): Promise<IVerificationSchema | null>
+  /**
+   * Updates the hash records for a payment contact field
+   * @param updateData Data with which to update field
+   */
+  updateHashForPaymentField(
+    updateData: UpdatePaymentFieldData,
```
justynoh: Looking at the model definition, I'm also wondering if it's possible instead to split the behavior via a flag rather than duplicating all the functions. For instance, we can include a `isPayment?: boolean` flag to the arguments of all of the affected functions, and in the service layer pass `fieldId === PAYMENT_CONTACT_FIELD_ID` in this argument.

src/app/modules/verification/verification.model.ts
```diff
@@ -96,17 +107,28 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
...
+    )
+      return null
+    let fields, paymentField
+    if (form_fields) {
+      fields = extractTransactionFields(form_fields)
+      if (fields.length === 0) return null
+    }
+    if (payments_field?.enabled) {
+      paymentField = { _id: PAYMENT_CONTACT_FIELD_ID, fieldType: 'email' }
+    }
```
justynoh: The previous version of the function did not require a mutable variable `fields` because it's existence was guaranteed by the inverse check `if (!form_fields) return null`. Now that we have two variables `field` and `paymentField`, and we only want to return `null` if **both** are not defined, this makes the logic a bit more complex. 

I wonder if there is a way to write this that could avoid the repeated `if` checks. In particular, we checked for the existence of `form_fields` twice. Perhaps one way is to populate both the `fields` and `paymentField` variables if necessary, and only perform an early return based on the computed values at the end. What do you think?

src/app/modules/verification/verification.model.ts
```diff
@@ -96,17 +107,28 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
...
+    const { form_fields, payments_field } = form
+    if (
+      (!form_fields || form_fields.length == 0) &&
+      (payments_field == undefined || !payments_field.enabled)
+    )
+      return null
+    let fields, paymentField
+    if (form_fields) {
+      fields = extractTransactionFields(form_fields)
+      if (fields.length === 0) return null
```
justynoh: are we certain this early return is correct? What about the case where there are non-verifiable form fields and payment enabled - would we (incorrectly) return null?

src/app/modules/verification/verification.model.ts
```diff
@@ -96,17 +107,28 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
...
+    }
+    if (payments_field?.enabled) {
+      paymentField = { _id: PAYMENT_CONTACT_FIELD_ID, fieldType: 'email' }
```
justynoh:
```suggestion
      paymentField = { _id: PAYMENT_CONTACT_FIELD_ID, fieldType: BasicField.Email }
```

src/app/modules/verification/verification.errors.ts
```diff
@@ -18,6 +18,15 @@ export class FieldNotFoundInTransactionError extends ApplicationError {
   }
 }
 
+/**
+ * Payment contact field with given ID not found in the transaction.
+ */
+export class PaymentContactFieldNotFoundInTransactionError extends ApplicationError {
```
justynoh: nit: Should we modify the name of `FieldNotFoundInTransactionError` to `FormFieldNotFoundInTransactionError` and `PaymentContactFieldNotFoundInTransactionError` to `PaymentFieldNotFoundInTransactionError`?

src/app/modules/verification/verification.util.ts
```diff
@@ -271,3 +274,56 @@ export const mapRouteError: MapRouteError = (
 export const hasAdminExceededFreeSmsLimit = (smsCount: number): boolean => {
   return smsCount > smsConfig.smsVerificationLimit
 }
+
+/**
+ *
+ * @param transaction Transaction document
+ * @returns ok(paymentField)
+ */
```
justynoh: Missing documentation? 🤔  haha

src/app/modules/verification/verification.util.ts
```diff
@@ -271,3 +274,56 @@ export const mapRouteError: MapRouteError = (
...
+export const getPaymentContactFieldFromTransaction = (
```
justynoh: Same comment about naming consistency, `Field` -> `FormField` and `PaymentContactField` -> `PaymentField`

src/app/modules/verification/verification.controller.ts
```diff
@@ -341,6 +381,15 @@ export const handleGenerateOtp = [
...
 
+/**
+ * NOTE: Exported soley for testing
+ * @param updateData
+ * @returns
+ */
+export const updateHashForPaymentFieldWrapper = (
+  updateData: UpdatePaymentFieldData,
+) => VerificationModel.updateHashForPaymentField(updateData)
+
```
justynoh: nit: would these wrapper functions be better placed in `verification.util.ts` instead?

src/app/modules/verification/verification.controller.ts
```diff
@@ -412,7 +477,24 @@ export const _handleOtpVerification: ControllerHandler<
...
+      .andThen(() => {
+        let getFieldFromTransactionFx
+        let incrementFieldRetriesFx
+        if (fieldId === PAYMENT_CONTACT_FIELD_ID) {
+          getFieldFromTransactionFx = getPaymentContactFieldFromTransaction
+          incrementFieldRetriesFx = incrementPaymentFieldRetriesWrapper
+        } else {
+          getFieldFromTransactionFx = getFieldFromTransaction
+          incrementFieldRetriesFx = incrementFormFieldRetriesWrapper
+        }
```
justynoh: I'm starting to think that maybe it's not necessary to abstract the functional conditioning away from the service layer into the controller layer. In any case, the check for whether `fieldId === PAYMENT_CONTACT_FIELD_ID` must occur at some point. In this case, we've complicated the service interfaces at the benefit of having a more uniform service layer function. 

On the contrary I feel that it would be better to have simpler controller functions, and have the service functions handle more of the detailed logic. This is also because we don't actually have separation of payment/form fields on the routes, and so we might expect that the model-level implementation detail of `FormField` vs `PaymentField` should be transparent to the controller. I'm thinking that the service layer should decide which method to call between the `FormField` and `PaymentField` variants. Better still, we don't need to modify any of the existing interfaces, since all the service functions take in `fieldId` anyway, so the check can just be shifted to service functions.  What do you think?

src/app/modules/verification/verification.model.ts
```diff
@@ -128,7 +150,27 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
...
+        $inc: {
+          'paymentField.hashRetries': 1,
+        },
+      },
+      {
+        new: true,
+        runValidators: true,
+        setDefaultsOnInsert: true,
+      },
+    ).exec()
```
justynoh: supernit: Not sure if this is a good idea, but it may be possible to abstract the arguments for all of these functions out, to avoid repeating the combinations (after all, it's just the prefix of `paymentField` or `fields.$` that changes). Nonetheless, not critical.

shared/constants/form.ts
```diff
@@ -51,3 +51,5 @@ export const ADMIN_FORM_META_FIELDS = <const>[
   '_id',
   'responseMode',
 ]
+
+export const PAYMENT_CONTACT_FIELD_ID = 'payment_contact_field'
```
justynoh, line 55: for consistency, let's stick to one term for each type of entity. I would recommend `FormFields` and `PaymentField`. You could also pick `PaymentContactField`, as long as we're consistent throughout the verification module. These will need to be made consistent in various places - the `model`, `types`, `errors` and `util` files, to name a few.

frontend/src/templates/Field/Nric/NricField.tsx
```diff
@@ -1,7 +1,7 @@
 /**
  * @precondition Must have a parent `react-hook-form#FormProvider` component.
  */
-import { useMemo, useState } from 'react'
+import { useMemo } from 'react'
```
justynoh: what's this change? haha if its removing an unused variable, maybe we can include this in a different PR into develop.

src/app/modules/verification/__tests__/verification.model.spec.ts
```diff
@@ -535,7 +539,7 @@ describe('Verification Model', () => {
       })
 
       it('should return null when the transaction ID is not found', async () => {
-        const result = await VerificationModel.updateHashForField({
+        const result = await VerificationModel.updateHashForFormField({
```
justynoh: Did we add tests for the payment versions of these functions?

src/app/modules/verification/__tests__/verification.routes.spec.ts
```diff
@@ -19,8 +19,8 @@ import { VfnRouter } from '../verification.routes'
 import {
   generateFieldParams,
   MOCK_HASHED_OTP,
+  MOCK_LOCAL_RECIPIENT,
```
justynoh, line 22: omg was this actually not being imported correctly all this while I'M DEAD

src/app/modules/verification/verification.service.ts
```diff
@@ -415,7 +401,7 @@ export const disableVerifiedFieldsIfRequired = (
 }
 
 /**
- * Compares the given otp. If correct, returns signedData, else returns an error
+ * Compares the given form otp. If correct, returns signedData, else returns an error
```
justynoh: supernit: extraneous `form `?

src/app/routes/api/v3/forms/public-forms.verification.routes.ts
```diff
@@ -41,14 +42,14 @@ PublicFormsVerificationRouter.route(
...
+  `/:formId([a-fA-F0-9]{24})/fieldverifications/:transactionId([a-fA-F0-9]{24})/fields/:fieldId([a-fA-F0-9]{24}|${PAYMENT_CONTACT_FIELD_ID})/otp/verify`,
 ).post(
   limitRate({ max: rateLimitConfig.sendAuthOtp }),
-  VerificationController.handleOtpVerification,
+  VerificationController.handleFormOtpVerification,
 )
 
 /**
- * Route for generating a new otp for a given field
+ * Route for generating a new otp for a given form field
```
justynoh: supernit: extraneous `form `?

src/app/modules/verification/verification.controller.ts
```diff
@@ -332,7 +372,7 @@ export const _handleGenerateOtp: ControllerHandler<
 /**
  * Handler for the POST /forms/:formId/fieldverifications/:transactionId/fields/:fieldId/otp/generate endpoint
  */
-export const handleGenerateOtp = [
+export const handleGenerateFormOtp = [
```
justynoh: Is there a reason for renaming this and `handleOtpVerification`? 

If the reason was to distinguish between login OTPs and Form OTPs, then we also need to avoid the use of the word "form" in `FormFields` (in distinguishing from `PaymentField`). Since in one context, we are using `Form` to refer to both the `form_fields` and `paymentField`, and in the other we are using `Form` to distinguish between the two.

frontend/src/features/verifiable-fields/Email/VerifiableEmailField.tsx
```diff
@@ -18,7 +18,7 @@ export interface VerifiableEmailFieldProps extends EmailFieldProps {
 
 /**
  * @example
- * { value: '+65 9876 5432', signature: some-signature }
+ * { value: 'form@open.gov.sg', signature: some-signature }
```
justynoh, line 21: LOL

frontend/src/features/verifiable-fields/VerifiableFieldProvider.tsx
```diff
@@ -178,7 +178,7 @@ export const VerifiableFieldProvider = ({
         },
       })
     },
-    [getValues, schema._id, setFocus, setValue, verifyOtpMutation],
+    [getValues, schema._id, setFocus, setValue, verifyFormOtpMutation],
```
justynoh: (assuming adding the word `Form` is to distinguish from login OTPs) maybe this can be done in a separate PR? we can decide on the language beforehand though, no issue.

justynoh (commented):
most nits just related to a single language confusion lol, can ignore most of them once we clarify the language of adding `Form` everywhere.

> If the reason was to distinguish between login OTPs and Form OTPs, then we also need to avoid the use of the word "form" in FormFields (in distinguishing from PaymentField). Since in one context, we are using Form to refer to both the form_fields and paymentField, and in the other we are using Form to distinguish between the two.

> (assuming adding the word Form is to distinguish from login OTPs) maybe this can be done in a separate PR? we can decide on the language beforehand though, no issue.

justynoh:
since the PR comments appear in chronological order, i feel like you can actually see my brainwaves in all the review comments lmao

frontend/src/templates/Field/Nric/NricField.tsx
timotheeg: We could establish a new norm for those 🤔. We see "somewhat often" the unused variables issues reported as warnings by github in the PRs. That's why we've been clearing them in the same stride, but you are right: that pollutes the file list on a given PR. We should instead spend 5 more mins to open trivial lint PRs just to clear those. (Well.. 5 mins to make the lint PRs, 30mins to wait for tests and merge 😢)

src/app/modules/verification/verification.model.ts
```diff
@@ -96,17 +107,28 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
...
     form: IFormSchema,
   ): Promise<IVerificationSchema | null> {
-    const { form_fields } = form
-    if (!form_fields) return null
-    const fields = extractTransactionFields(form_fields)
-    if (fields.length === 0) return null
+    const { form_fields, payments_field } = form
+    if (
+      (!form_fields || form_fields.length == 0) &&
+      (payments_field == undefined || !payments_field.enabled)
```
timotheeg: /nit

can be shortened
```suggestion
      !form_fields?.length && !payments_field?.enabled
```
or at least made consistent in style
```suggestion
      (!form_fields || form_fields.length === 0) &&
      (!payments_field || !payments_field.enabled)
```

src/app/modules/verification/verification.model.ts
LinHuiqing: Haha yes looking at the code again, I think you're right oop

src/types/verification.ts
LinHuiqing: Hm I can definitely get behind specifying `Field` as `FormField`, but as for reducing `PaymentContactField` to `PaymentField`, I'm not sure if it will be confusing in the future as there `PaymentField` more intuitively suggests that it will include contents like the dollar amount, description, etc. when it will just hold the email / future alternate contact information haha

src/app/modules/verification/verification.errors.ts
LinHuiqing: same reply as https://github.com/opengovsg/FormSG/pull/6003#discussion_r1162212118

src/app/modules/verification/verification.util.ts
LinHuiqing: hahaha 🙈

src/app/modules/verification/verification.util.ts
LinHuiqing: https://github.com/opengovsg/FormSG/pull/6003#discussion_r1162212118

src/app/modules/verification/verification.controller.ts
LinHuiqing: Hm, maybe! I initially put them here as it is only used by the controller so there's no need for it to be exported.

src/app/modules/verification/verification.controller.ts
LinHuiqing: Yea since now we're using the same endpoints for both normal form fields and payment contact field, then it makes sense for the logic to be left till the service level... 😢 okay i'll refactor and update the tests

frontend/src/templates/Field/Nric/NricField.tsx
LinHuiqing: Yea it's just to remove the unused import. I'll undo this change on this PR and start a new PR if it's still there on the `feat/payment-mvp` branch then.

src/app/modules/verification/__tests__/verification.model.spec.ts
LinHuiqing: Oh... I didn't :') thanks for spotting this! will add them

src/app/modules/verification/__tests__/verification.routes.spec.ts
LinHuiqing: Yea HAHA I don't know why jest didn't point it out / we didn't notice all this time

src/app/modules/verification/verification.controller.ts
LinHuiqing: Oh! This was from when I had both `handleGenerateFormOtp` and `handleGeneratePaymentOtp` for the 2 endpoints I set up initially. Will change it back!

frontend/src/features/verifiable-fields/VerifiableFieldProvider.tsx
LinHuiqing: Nah, https://github.com/opengovsg/FormSG/pull/6003#discussion_r1162222358. Will remove `Form`.

src/app/modules/verification/verification.controller.ts
justynoh: Okie, then can ignore all my other comments re the "form" prefix haha

frontend/src/templates/Field/Nric/NricField.tsx
justynoh: Actually is more that I feel these should be done cleanly on the most upstream branch where the change occurred. Since NRIC field wasn't touched by payments I'm p sure it's also there on develop. We should open a PR to develop to fix it haha

src/types/verification.ts
justynoh: yeah, `PaymentContactField` works fine. should we update the model and all related functions with the new terminology? then we stick to `FormFields` and `PaymentContactField` ba

src/app/modules/verification/verification.controller.ts
timotheeg: As a small extra nit to the discussion, whether they are in controller or in service, even for 2 instance of the same code, use a function:

```typescript
function getFieldFunctions(fieldId) {
  let getFieldFromTransactionFx
  let incrementFieldRetriesFx
  if (fieldId === PAYMENT_CONTACT_FIELD_ID) {
    getFieldFromTransactionFx = getPaymentContactFieldFromTransaction
    incrementFieldRetriesFx = incrementPaymentFieldRetriesWrapper
  } else {
    getFieldFromTransactionFx = getFieldFromTransaction
    incrementFieldRetriesFx = incrementFormFieldRetriesWrapper
  }
  
  return { getFieldFromTransactionFx, incrementFieldRetriesFx }
}

// [ ... ]

const { getFieldFromTransactionFx, incrementFieldRetriesFx } = getFieldFunctions(fieldId)

```

timotheeg (commented):
PR looks good to me so far! I'm not done with review, but sending 2 nits before I jump to a meeting (sorry!)

src/app/modules/verification/verification.controller.ts
LinHuiqing: The wrappers can be removed with the refactoring I've done to shift the form / payment logic to the service layer, so this should be gone!

src/types/verification.ts
LinHuiqing: With the refactor, there's no need for different data types so we can just use `UpdateFieldData` with an additional `isPayment` flag.

src/types/verification.ts
LinHuiqing: Okays 👍

src/types/verification.ts
LinHuiqing: Done in [8b1eed0](https://github.com/opengovsg/FormSG/pull/6003/commits/8b1eed09b66bddbb99cdc30b40f25646680e70ab).

src/app/modules/verification/verification.controller.ts
LinHuiqing: Done in [8b1eed0](https://github.com/opengovsg/FormSG/pull/6003/commits/8b1eed09b66bddbb99cdc30b40f25646680e70ab).

src/app/modules/verification/verification.model.ts
LinHuiqing: Done in [8b1eed0](https://github.com/opengovsg/FormSG/pull/6003/commits/8b1eed09b66bddbb99cdc30b40f25646680e70ab).

shared/constants/form.ts
LinHuiqing: With the refactor, there are no more form or payment specific methods as paths are distinguished by the `fieldId` or `isPayment` flag.

src/app/modules/verification/verification.service.ts
LinHuiqing: Done in b74c705e.

src/app/routes/api/v3/forms/public-forms.verification.routes.ts
LinHuiqing: Done in 726a0f2d.

src/app/modules/verification/verification.controller.ts
LinHuiqing: done in cb89e407

src/app/modules/verification/verification.controller.ts
LinHuiqing: Hmm regarding this, I got rid of passing around the functions completely 🤔 I simply used `isPayment` as a flag after comparing the `fieldId` to the `PAYMENT_CONTACT_FIELD_ID` value to improve code reuse. Is using separate functions and passing them around using the way you suggested above is better? Thoughts? haha

timotheeg:
Just writing down a note on the breaking change: considering we will be releasing as beta, we can expect the first release to not hit the new payment otp endpoint right away. Because of that, generally speaking, we will not need extra care for the first payment release for this "risk".

src/types/form.ts
LinHuiqing: Okay yea since it's in `IEncryptedFormSchema`, there's no need for it to be here! Removed it in fd1cb4f6.

LinHuiqing: Wait actually, then why are `addPaymentAccountId` and `removePaymentAccount` in `IFormSchema` too? They should ideally be in `IEncryptedFormSchema` right?

```ts
  /**
   * Add payment account ID to the form.
   * @param accountId the payment account ID to add
   * @returns updated form
   */
  addPaymentAccountId<T = IEncryptedFormSchema>({
    accountId,
    publishableKey,
  }: {
    accountId: FormPaymentsChannel['target_account_id']
    publishableKey: FormPaymentsChannel['publishable_key']
  }): Promise<T & DeepRequired<Pick<IEncryptedFormSchema, 'payments_channel'>>>

  /**
   * Remove payment account ID from the form.
   * @returns updated form
   */
  removePaymentAccount<T = IEncryptedFormSchema>(): Promise<T>
```

justynoh: yup, there's an issue open for this #5977, we can work on it after this

frontend/src/templates/Field/Nric/NricField.tsx
LinHuiqing: reverted in 23e38d90

src/types/form.ts
LinHuiqing: okays 👍

LinHuiqing:
@justynoh @timotheeg: Did a refactor of `VerificationSchema.statics.createTransactionFromForm` based on your comments, lemme know if it can be improved further? 👀

src/app/modules/verification/__tests__/verification.model.spec.ts
LinHuiqing: Holy, after I added the tests for the refactor (including those related to the above comment), the lines of code got bumped up to 3k (from 1.9k) :')

frontend/src/templates/Field/Nric/NricField.tsx
LinHuiqing: Okay, just checked the code on `develop`, this has been removed already!

src/app/modules/verification/verification.controller.ts
timotheeg: It's not necessarily better no; passing a boolean `isPayment` for the decision to be taken deeper is ok.

Reading the current version of the PR, `isPayment` really is a "property" of `fieldId`, so it still feels a little funny. I wonder if the conditional check can be moved even deeper to the model 🤔

src/app/modules/verification/verification.model.ts
```diff
@@ -93,31 +108,59 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
...
-    if (!form_fields) return null
-    const fields = extractTransactionFields(form_fields)
-    if (fields.length === 0) return null
-    return this.create({
-      formId: form._id,
-      fields,
-    })
+    const formFields = getVerifiableFormFields(form.form_fields)
+    if (form.responseMode === FormResponseMode.Encrypt) {
+      const { payments_field } = form as IEncryptedFormSchema
```
timotheeg, line 137: It's funny the `as IEncryptedFormSchema` is needed, I would have hoped the previous conditional check would have been enough for type narrowing 🤔

src/app/modules/verification/verification.model.ts
```diff
@@ -156,20 +201,21 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
   VerificationSchema.statics.updateHashForField = async function (
     updateData: UpdateFieldData,
   ): Promise<IVerificationSchema | null> {
+    const operationPrefix = updateData.isPayment ? 'paymentField' : 'fields.$'
```
timotheeg: 3rd repetition of this, we probably want either constants for the prefixes, or a `getVerificationPrefix()` function 🤔

src/app/modules/verification/verification.model.ts
```diff
@@ -130,19 +173,21 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
 
   VerificationSchema.statics.resetField = async function (
     transactionId: string,
+    isPayment: boolean,
     fieldId: string,
   ): Promise<IVerificationSchema | null> {
+    const operationPrefix = isPayment ? 'paymentField' : 'fields.$'
```
timotheeg: <shame on me> I could understand what the `.$` syntax was doing, but I wasn't sure how it worked, so I had to look [its doc](https://www.mongodb.com/docs/manual/reference/operator/update/positional/#---update-) up 😅

src/app/modules/verification/verification.model.ts
```diff
@@ -156,20 +201,21 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
   VerificationSchema.statics.updateHashForField = async function (
     updateData: UpdateFieldData,
   ): Promise<IVerificationSchema | null> {
+    const operationPrefix = updateData.isPayment ? 'paymentField' : 'fields.$'
     return this.findOneAndUpdate(
       {
         _id: updateData.transactionId,
-        'fields._id': updateData.fieldId,
+        ...(!updateData.isPayment && { 'fields._id': updateData.fieldId }),
```
timotheeg, line 212: /nothingToChangeJustNoting!

Whoa, this line is super awesome and gross at the same time! It manages to mix a bunch of JS gotchas into one great result 😅 👏

src/types/verification.ts
```diff
@@ -26,28 +26,32 @@ export interface IVerification {
   formId: IFormSchema['_id']
   expireAt: Date
   fields: IVerificationFieldSchema[]
+  paymentField: IVerificationFieldSchema
 }
 
 export type UpdateFieldData = {
+  isPayment: boolean
```
timotheeg, line 33: Per my other comment, `isPayment` is actually a "property" of `fieldId`, so the structure feels a little redundant... We can think about this more and change that later though (it's all internal after all), so I reckon we should merge first and have the task done for now 🤔

timotheeg (approved):
LGTM - including tests 👍 

One addition suggestion for the tests:
- [ ] repeat steps 2-6 on a storage form with both payment and 2 verifiable fields in the form (1 email + 1 mobile)

src/types/verification.ts
LinHuiqing: Yea it does feel a bit weird, but then if we just use the `fieldId`, we'll have a lot more if blocks. Alternatively, I think we can take out the boolean flag as another argument for the function (`updateHashForField`) - do you think that will make more sense?

src/app/modules/verification/verification.model.ts
LinHuiqing: hahaha yea I learnt the `...()` trick from sir stackoverflow for this 😂 pretty neat!

src/app/modules/verification/verification.model.ts
LinHuiqing: thanks! coolios!

src/app/modules/verification/verification.model.ts
LinHuiqing: done in fc647477

src/app/modules/verification/verification.model.ts
LinHuiqing: Yea, it doesn't... It's probably something to do we how we set up the schema typings, maybe something to fix in another PR soon before tech debt builds up? 🤔

LinHuiqing:
Oh yea regarding your additional test suggestion, added! 👍 
Fun fact: the verification tests currently only cover mobile OTPs :') so if we want the full suite of tests for verifications, its going to be a lot more tests (future work? :'))

LinHuiqing:
Oh yea regarding your additional test suggestion, added! 👍 @timotheeg 
Fun fact: the verification tests currently only cover mobile OTPs :') so if we want the full suite of tests for verifications, its going to be a lot more tests (future work? :'))

src/types/verification.ts
LinHuiqing: Okay I'm just gonna leave this to a separate PR if we wanna change it hahaha

src/app/modules/verification/verification.model.ts
LinHuiqing: Either way, I think this isn't something we're gonna fix in this PR so I'm just gonna merge first bah

src/app/modules/verification/verification.controller.ts
LinHuiqing: hmm actually you're right, maybe the conditional check can just be moved into the model layer with the service layer not having any knowledge of it 🤔

PR #6008, opened 2023-03-29
feat: more logging on fetch fallback

timotheeg:
## Context

Adding more comprehensive logging and minor fixes prior to releasing the axios fallback.

We have noticed that when there is a fetch error (e.g. json parse error), we do not have a handle to the response object to query and log more data from it.

Also, we spotted some minor issue with fetch URLs using double slashes like this:
```
POST /api/v3//forms/6423c090f88d780012a8c162/submissions/encrypt?captchaResponse=null 200 158ms
```

## Approach

Add minor fixes and more verbose logging

### DIFF
--- a/frontend/src/features/admin-form/common/AdminViewFormService.ts
+++ b/frontend/src/features/admin-form/common/AdminViewFormService.ts
@@ -11,7 +11,11 @@ import {
 
 import { ADMINFORM_USETEMPLATE_ROUTE } from '~constants/routes'
 import { transformAllIsoStringsToDate } from '~utils/date'
-import { API_BASE_URL, ApiService } from '~services/ApiService'
+import {
+  API_BASE_URL,
+  ApiService,
+  processFetchResponse,
+} from '~services/ApiService'
 
 import { augmentWithMyInfoDisplayValue } from '~features/myinfo/utils'
 import {
@@ -221,7 +225,7 @@ export const submitEmailModeFormPreviewWithFetch = async ({
       body: formData,
     },
   )
-  return response.json()
+  return processFetchResponse(response)
 }
 
 /**
@@ -256,5 +260,5 @@ export const submitStorageModeFormPreviewWithFetch = async ({
       },
     },
   )
-  return response.json()
+  return processFetchResponse(response)
 }
--- a/frontend/src/features/admin-form/common/AdminViewFormService.ts
+++ b/frontend/src/features/admin-form/common/AdminViewFormService.ts
@@ -27,7 +27,7 @@ import {
 import { PREVIEW_MOCK_UINFIN } from '../preview/constants'
 
 // endpoint exported for testing
-export const ADMIN_FORM_ENDPOINT = 'admin/forms'
+export const ADMIN_FORM_ENDPOINT = '/admin/forms'
 
 /**
  * Gets admin view of form.
@@ -215,7 +215,7 @@ export const submitEmailModeFormPreviewWithFetch = async ({
   const formData = createEmailSubmissionFormData(formFields, filteredInputs)
 
   const response = await fetch(
-    `${API_BASE_URL}/${ADMIN_FORM_ENDPOINT}/${formId}/preview/submissions/email`,
+    `${API_BASE_URL}${ADMIN_FORM_ENDPOINT}/${formId}/preview/submissions/email`,
     {
       method: 'POST',
       body: formData,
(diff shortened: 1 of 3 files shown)


frontend/src/features/admin-form/common/AdminViewFormService.ts
```diff
@@ -27,7 +27,7 @@ import {
 import { PREVIEW_MOCK_UINFIN } from '../preview/constants'
 
 // endpoint exported for testing
-export const ADMIN_FORM_ENDPOINT = 'admin/forms'
+export const ADMIN_FORM_ENDPOINT = '/admin/forms'
```
timotheeg, line 34: This change is to be consistent with `PUBLIC_FORMS_ENDPOINT` in `PublicFormService.ts`

frontend/src/services/ApiService.ts
```diff
@@ -78,3 +78,32 @@ ApiService.interceptors.response.use(
...
+      meta: {
+        action: 'processFetchResponse',
+        response: {
+          status: response.status,
+          statusText: response.statusText,
+          headers: [...(response.headers?.entries() || [])],
+          body: response.body,
+        },
+        error: {
+          code: error?.code,
```
tshuli: from what I can tell, there's no type guarantee that the fetch error will contain the `code` property - so omitted this.

frontend/src/services/ApiService.ts
```diff
@@ -78,3 +78,32 @@ ApiService.interceptors.response.use(
...
+          message: error?.message,
+          stack: error?.stack,
+          dump: error?.toJSON ? error.toJSON() : undefined,
```
tshuli: Revisiting this - i think it's correct to simply call `JSON.stringify(error)`. This is because if the `.toJSON()` method is defined, it will override the serialisation for `JSON.stringify()`.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior

frontend/src/services/ApiService.ts
timotheeg: It would have been fine to leave it I think. As much as we want to have type safety everywhere, if we do get errors with unexpected shape and a code field, making our extraction grab it works (in contrast to json encoding of the error object which will skip all non-enumerable properties). Plus If `error.code` is `undefined`, json stringification will remove the field anyway when it is sent to DD.

frontend/src/services/ApiService.ts
```diff
@@ -78,3 +78,34 @@ ApiService.interceptors.response.use(
...
 )
+
+export const processFetchResponse = async (response: Response) => {
+  try {
+    const data = await response.json()
+
+    // TODO: data may be present but if response is non 2XX, transform to error and throw
+    return data
+  } catch (error) {
+    if (error instanceof Error) {
```
timotheeg: This doesn't look correct? Why do we need to check the type of `error`?

frontend/src/services/ApiService.ts
```diff
@@ -78,3 +78,34 @@ ApiService.interceptors.response.use(
...
+          error: {
+            name: error.name,
+            message: error.message,
+            stack: error.stack,
+            dump: JSON.stringify(error),
+          },
+        },
+      })
+
+      throw error
```
timotheeg: This `throw` should be out of the `if`, otherwise the code is swallowing the error, and will likely cause runtime errors down the processing chain when the code assumes it was a response when it does not.

frontend/src/services/ApiService.ts
tshuli: [a52c228](https://github.com/opengovsg/FormSG/pull/6008/commits/a52c228ca54df85ff215ac97c76f91bdaed84e39)

frontend/src/services/ApiService.ts
tshuli: [541bf25](https://github.com/opengovsg/FormSG/pull/6008/commits/541bf25cefe0bd9526f7197b32409e7958124cf8)

frontend/src/services/ApiService.ts
tshuli: [541bf25](https://github.com/opengovsg/FormSG/pull/6008/commits/541bf25cefe0bd9526f7197b32409e7958124cf8)

timotheeg (commented):
LGTM! I can't approve since I opened the PR 😑😅. Please approve it @tshuli 😄

tshuli (approved):

Issue #5849, opened 2023-02-28
Disable admin from adding payment when stripe account has not been connected


LinHuiqing:
Currently, the error handling for adding payment before connecting the form to the stripe account is not informative enough for the admin to know what to do to alleviate the problem - admins can still add payment, and will only encounter a "Validation failed" error message when trying to submit:

https://user-images.githubusercontent.com/37061143/221742697-38e50e81-0b54-43ed-9c92-25e188ff1f13.mov

foochifa:
Also can implement @staceytan1998's design for when account isn't connected 
<img width="454" alt="image" src="https://user-images.githubusercontent.com/59867455/221743205-130ea9a2-1d04-4416-a25f-f0261255722c.png">

justynoh:
Related to #5848

foochifa:
Done #6037

Issue #5848, opened 2023-02-28
Delete publishable key from form model when account is unlinked


justynoh:
![Image](https://user-images.githubusercontent.com/25571626/221742096-a5aa98a7-2c5c-432a-9d44-d96bcc9c4393.png)

removePaymentAccount does not overwrite `publishableKey` :(

Dependency
- #6012

foochifa:
![image](https://user-images.githubusercontent.com/59867455/227851680-1bb2049a-4c0b-4407-83c6-6fcaf5541a46.png)

Currently, payment channel properties are set as optional as we only remove the target_account_id in removePaymentAccount. We should remove the entire payment channels instead. And make the above properties required.

KenLSM:
Note: We should first ensure that the `target_account_id` and `publishable_key` are migrated to payment records before shipping this. 

In the event that admin unlinks the account, responders will not be able to query for their payment status.

foochifa:
Done #6036

PR #6011, opened 2023-03-29
feat: add payment unsupported message in email mode

### LINKED ISSUE #5819: Disable payments in FE settings and field for email mode forms

foochifa:
## Problem

Currently payment elements in email mode displays an empty drawer and a stripe error on payment settings. We should provide information for admins to prompt them that payments is unavailable in email mode.

Closes #5819

## Solution

For PaymentDrawer
Check if form is email mode, and if it is, disables the payment drawer but still provides the default payment field in the usePaymentStore (only enabled set to false)

For PaymentSettings
Use the same svgr for webhooks, designs can be found in figma file

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image
image 
image

**AFTER**:
image
image

## Tests

In an email mode form (to confirm changes are made correctly):
- [ ] Try to access payment settings page, you should see the svgr in the screenshot
- [ ] Try to access the payment drawer in the create page. You should see a inline notification that payments is not supported in email mode
- [ ] Also ensure you cannot make modification to the payment drawer tab.

In a storage mode form (to ensure core functionality is not affected)
- [ ] Go to payment settings page, connect your stripe account, connection should succeed
- [ ] Go to payment drawer, enable and update the payment field, update should succeed


foochifa:
Whilst working on this, I also explored fixing the frontend for #5849 (as it's tangentially related, displaying an inline message for stripe connection), have decided to make a separate PR for that, after this.

foochifa:
The link to the payment guide (for the payment settings page) isn't ready yet, have made an issue to track it #6014, will update after

frontend/src/features/admin-form/create/payment/usePaymentStore.tsx
```diff
@@ -56,10 +58,10 @@ export const usePaymentStore = create<PaymentStore>()(
     },
     setToInactive: (holding) => {
       if (holding) {
-        set({ holdingState: PaymentState.EditingPayment })
+        set({ holdingState: PaymentState.Inactive })
```
KenLSM, line 58: Oh this is the bug you were mentioning where the form doesn't get set to inactive after the tab is closed?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -267,11 +277,30 @@ export const PaymentInput = (): JSX.Element => {
...
-export const PaymentDrawer = ({
-  paymentsField,
+// Will be extended for stripe unconnected messages too
+const PaymentDisabledMessage = ({
+  isEncryptMode,
 }: {
-  paymentsField: FormPaymentsField
+  isEncryptMode: boolean
 }): JSX.Element | null => {
+  return !isEncryptMode ? (
```
KenLSM: Would prefer to have final return to be the element that's intended to render if we're in the happy flow.

Example
```jsx
if (isEncryptMode) { return null }

return (<Box>...</Box>)
```

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -267,11 +277,30 @@ export const PaymentInput = (): JSX.Element => {
...
+
+type PaymentDrawerProps = {
+  isEncryptMode: boolean
+  paymentsField?: FormPaymentsField
+}
+
+export const PaymentDrawer = ({
+  isEncryptMode,
+  paymentsField,
+}: PaymentDrawerProps): JSX.Element | null => {
```
KenLSM, line 315: Curious to understand, why do we always specify the return value for Components?

I assume that variables exported from `.tsx` that are in PascalCase are always `<Components />` or `null`

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
KenLSM: On second thought, feels like this component shouldn't need to accept a `isEncryptMode` props. The parent should be the controller of whether this component should render since this component meant to show a `PaymentDisabledMessage` 🤔

KenLSM (approved):

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: I think being more verbose wouldn't hurt?

But I can see why, cause even if we remove the specification the static analyser can still assume the return types too.

@justynoh?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: The reason for this is that being in email mode isn't the only cause for a `PaymentDisabledMessage`, #5849 which is what we were experimenting ytd is the other cause. 

Technically I could do the rendering and the switch inside PaymentDrawer itself but felt it will be cleaner if I abstracted it all out!

frontend/src/features/admin-form/create/payment/usePaymentStore.tsx
foochifa: Yep HAHA

frontend/src/features/admin-form/create/payment/PaymentTab.tsx
```diff
@@ -4,18 +4,25 @@ import { useAdminForm } from '~features/admin-form/common/queries'
...
   const { data: form } = useAdminForm()
 
-  return form?.responseMode === FormResponseMode.Encrypt &&
-    form?.payments_field ? (
+  const isEncryptMode = form?.responseMode === FormResponseMode.Encrypt
+
+  const paymentsField =
+    isEncryptMode && form?.payments_field
+      ? form.payments_field
+      : defaultPaymentsField
```
justynoh: is the reason for needing `defaultPaymentsField` here because currently payments_field is undefined for all forms? If so then i would propose making the `paymentsField` input to `PaymentDrawer` and `PaymentContent` optional, and allowing the children to handle undefined input. For two reasons:

1. Since email mode forms aren't allowed to have payments fields, we should not pass defaults to the children at all, they should be undefined. (If we want to lock the toggle to be disabled in the payment drawer because of email mode, we should hardcode both the "isDisabled" and "value" attributes to guarantee this, rather than relying on the default defined elsewhere in code.) 
2. Basically no additional handling needs to happen for this work - the payment drawer and payment content elements both accept optional input as it is.

What this allows us to do is pass the `form?.payments_field` directly to the children (along with `isEncryptMode`, where necessary `responseMode` checks need to occur), and avoids passing invalid state around (i.e. an email mode form where `paymentsField === { enabled: false }`)

frontend/src/features/admin-form/create/payment/usePaymentStore.tsx
```diff
@@ -21,11 +21,13 @@ export type PaymentStore = {
   moveFromHolding: () => void
 }
 
+export const defaultPaymentsField = { enabled: false } as FormPaymentsField
+
```
justynoh: relatedly: then this doesn't have to be exported

frontend/src/features/admin-form/create/payment/PaymentTab.tsx
foochifa: In my latest commit [7be108e](https://github.com/opengovsg/FormSG/pull/6011/commits/7be108e9a5abe7f7cad0adadf127b406ab5176b5) I faced an issue when I do not set enabled as a boolean, as FormPaymentsField require enabled to exist. Hence, if I do not do `paymentCommon = {enabled:false}`, I get a bunch of Typescript errors. Let me know if you have a way to circumvent this!

However, in terms of code logic, have followed your suggestions thanks @justynoh! So we do not send a default field to the drawer, and the disabled display is already hardcoded in.

Additionally, have ensured that when payment input is disabled, there will be no mutations

foochifa: Also, I think the concept of `payments_field.enabled` is still iffy, and we should use the existence of both `payments_channel` and `payments_field` as the 'enabling factor' (after #5848 is done), and `enabled` doesn't exist in the design figma either.

foochifa:
The current payment tab/drawer/content UI is also different from the envisaged design in our figma, have created an issue #6018 to sort it out in the future

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
justynoh: in that case, is it necessary to have a separate component for this? I wonder if we can do this directly in the `PaymentDrawer` component instead:

```typescript
const PaymentDrawer = () => {
  // ...

  const isDisabledMessage = 
    isEmailMode 
    ? 'Payments are not available for email mode.' 
    : !form.payment_channel 
    ?  'Please add a payment channel before enabling payments'
    : ''

  const isDisabled = !!isDisabledMessage

  // ...

  return (
    // ...
    { isDisabledMessage && <Box> {isDisabledMessage} </Box>}
    // ...
  )
}
```

(maybe `!form.payment_channel` is not a reason for disabling, this is just an example of how it would look with multiple conditions. it would also be more easily extensible to more disabling reasons, rather than passing around props for the purpose of checking if the form should be disabled)

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
justynoh: no specific reason, and it's pretty standard throughout the codebase. Personally though I like to explicitly type all my functions :D

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: Yeah thats possible too, I was considering between these two options, but in the end I felt separating it into its own function was 'cleaner' and potentially more extendable in the future (in case we ever have more info/errors from logic etc)

frontend/src/features/admin-form/create/payment/usePaymentStore.tsx
```diff
@@ -21,11 +21,13 @@ export type PaymentStore = {
...
 }
 
+const defaultPaymentsField = { enabled: false } as FormPaymentsField
+
 export const usePaymentStore = create<PaymentStore>()(
   devtools((set, get) => ({
     state: PaymentState.Inactive,
     holdingState: null,
-    data: { enabled: false } as FormPaymentsField,
+    data: defaultPaymentsField,
```
justynoh: nit: `data: { enabled: false }`

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: Hmm but if we have additional error messages, it will end up bloating the if else chain right?

In reference to how we did myinfo's inline messages (in `FieldListDrawer`), if we have additional checks in the future it might bloat the PaymentDrawer

frontend/src/features/admin-form/create/payment/PaymentTab.tsx
justynoh: For your first comment, doing `const { paymentsData: { amount_cents: paymentAmountCents, ...paymentCommon }, ... } = usePaymentStore(...)` in `PaymentDrawer` seems to be fine for me... unless I'm misunderstanding your comment?

Also, not related to your PR but I wonder if we could pick a better variable name than `paymentCommon` for what it's supposed to represent...

For your second comment, you're right - we can revisit it in the future.

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
justynoh: I don't feel strongly either way. If there is complexity, it has to go somewhere - either we expand the number of props taken by `PaymentDisabledMessage` or the expand the logic in `PaymentDrawer`. Maybe others can weigh in if they feel strongly.

Nonetheless circling back, i agree with Ken's initial nit which is to prefer having the happy path be the final return :P

frontend/src/features/admin-form/create/payment/PaymentTab.tsx
foochifa: It will display another error here when you try to open to payment tab: 
image
As we use zustand store so Typescript will 'assume' FormPaymentsField exists when it doesn't 😅

And encrypt guarding the useEffect in paymentdrawer doesn't work too, honestly I'm having some difficulty understanding how we are using zustand's store

frontend/src/features/admin-form/create/payment/usePaymentStore.tsx
foochifa: [52331ad](https://github.com/opengovsg/FormSG/pull/6011/commits/52331ad1f1d4799e15fbcd7455eec530b0869ed3)

frontend/src/features/admin-form/create/payment/PaymentTab.tsx
justynoh: as discussed, can make data in the store optional and set the default in `PaymentDrawer`'s useEffect, a la end page store

foochifa: [1585f40](https://github.com/opengovsg/FormSG/pull/6011/commits/1585f409d8df9383d0f3943908bb1f6ce8cd8006)

Have completely removed the need for a default value, following end page store's implementation. As either way the default value seems to almost always get set as undefined by `setToInactive()`

Have make some slight improvements over the two commits as per discussed with @justynoh, thanks so much!!!

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -82,10 +82,10 @@ export const PaymentInput = ({
...
   )
 
-  // unpack payment data if it exists
-  const { amount_cents: paymentAmountCents, ...paymentCommon } = paymentsData
-    ? paymentsData
-    : ({ enabled: false } as FormPaymentsField)
+  // unpack payment data for paymentAmount if it exists
+  const paymentAmountCents = paymentsData
+    ? paymentsData.amount_cents
+    : undefined
```
justynoh: nit: `const paymentAmountCents = paymentsData?.amount_cents`

foochifa: woops 🙈 [52a6394](https://github.com/opengovsg/FormSG/pull/6011/commits/52a6394f30c92c6b19eb6bdb1ac64a8332749506)

justynoh (approved):
lgtm

PR #6028, opened 2023-04-03
feat: add strong typing for stripe metadata fields, add more metadata fields

### LINKED ISSUE #5960: Add new items to metadata
Payment metadata should contain:
Form_name
Form_id
Submission_id
Payment_id
Payer_email_id

justynoh:
## Problem
We wanted to add more fields to stripe metadata for better linkage between stripe payments and submissions. At the same time, we wanted to extract the metadata fields into a type so that we can avoid hardcoding key names.

Closes #5960 

## Solution

Extract stripe metadata fields into a type, and add a typeguard for it. Use the typeguard in places where metadata is referenced, and replace usage of lodash get with typescript key access instead.

Discussion item: We don't have access to the submission id when the metadata is injected... how?

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -7,6 +7,8 @@ import mongoose from 'mongoose'
 import Stripe from 'stripe'
 import type { SetOptional } from 'type-fest'
 
+import { StripePaymentMetadataDto } from 'src/types'
+
 import {
   ErrorDto,
   FormAuthType,
@@ -350,8 +352,8 @@ const submitEncryptModeForm: ControllerHandler<
     form.payments_field?.enabled &&
     form.payments_channel?.channel === PaymentChannel.Stripe
   ) {
+    // Step 0: Perform validation checks
     const amount = form.payments_field.amount_cents
-    // Step 1: Create payment without payment intent id and pending submission id.
     if (
       !amount ||
       amount < paymentConfig.minPaymentAmountCents ||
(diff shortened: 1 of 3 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -367,9 +369,23 @@ const submitEncryptModeForm: ControllerHandler<
...
+    const paymentReceiptEmail = req.body.paymentReceiptEmail
+    if (!paymentReceiptEmail) {
+      logger.error({
+        message:
+          'Error when creating payment: payment receipt email not provided.',
+        meta: logMeta,
+      })
+      return res.status(StatusCodes.BAD_REQUEST).json({
+        message:
+          "The form's payment settings are invalid. Please contact the admin of the form to rectify the issue.",
```
LinHuiqing, line 381: Should this error message be more specific to say the the payment receipt email cannot be found? Because form admins will still not know what is wrong given this error message by respondents.
Maybe this needs some design input?

LinHuiqing (commented):
Another comment: Do you think it will make more sense if `paymentReceiptEmail` is changed to `paymentContact` or `paymentContactEmail` instead to widen the scope of what this email can be used for in the future (e.g. submission summary, user accounts, etc)?

lgtm otherwise!

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
justynoh: haha i just reused the same error message from above, we can change if necessary later on!

justynoh:
yeah we can change it to `paymentContactEmail`, that makes sense to me! will update

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: aights! should be okay for now since we have different error codes 👍

LinHuiqing (approved):
lgtm!

PR #6029, opened 2023-04-03
chore: enforce payment channel keys to be required

justynoh:
## Problem
Payment_channel itself is already allowed to be undefined, so we should not allow it's keys to be undefined either.

## Solution
Removed `?` from `target_account_id` and `publishable_key` definition. Change conditions throughout code that used optional chaining to just checking the `payment_channel` property instead.

### DIFF
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -1,5 +1,5 @@
 import BSON, { ObjectId } from 'bson-ext'
-import { compact, merge, omit, pick, uniq } from 'lodash'
+import { compact, omit, pick, uniq } from 'lodash'
 import mongoose, {
   ClientSession,
   Mongoose,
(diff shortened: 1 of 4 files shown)


src/app/models/form.server.model.ts
```diff
@@ -184,22 +184,24 @@ EncryptedFormDocumentSchema.methods.addPaymentAccountId = async function ({
...
-    enabled: true,
-  })
+  if (!this.payments_channel) {
+    this.payments_channel = {
+      // Definitely Stripe for now, may be different later on.
+      channel: PaymentChannel.Stripe,
+      target_account_id: accountId,
+      publishable_key: publishableKey,
+    }
+  }
```
KenLSM, line 194: cmiiw, `this.payments_channel.enabled` is different from `this.payments_field.enabled`, and `enabled: true` isn't supposed to be here as it's not in the schema/not needed either?

KenLSM (approved):

src/app/models/form.server.model.ts
justynoh: yup, that's right! `enabled` doesn't exist on the `payment_channel` object type

PR #6030, opened 2023-04-03
feat: payment browser stm

### LINKED ISSUE #5970: Local storage mode to track incomplete submissions/payments
In local storage, add a marker that a form submission and payment has been submitted and is ongoing. The marker can be deleted when the payment flow is complete.

If the form page is loaded while the marker is present and recent (recency specifications TBD), we can query FormSG's server to check the status of the submission, inform the user of the details, and ask the user if:
1.  he/she is sure she wants to proceed to make a new submission
2. proceed to payment page if the submission was pending

Dependency
- #5932

KenLSM:
**Notes to Reviewers:**
Targeting `feat/dedicated-payment-page` to have an easier time reviewing as this PR dependents on that branch.

## Problem

Users may crash/unintentionally close/press back after reaching the payment page. As we do not have an account-like feature for responders to find their previous payment, we needed to add some guardrails to help them resume their payment.

Closes #5970

## Solution

1. Store their last payment record of 
2. Delete the payment record once payment is completed
3. If the user lands on the form page, we will check the STM and show the prompt if there exist a record
4. Expire the record after 1 day

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

None. This is a new feature.

**AFTER**:

**Desktop View**
Screenshot 2023-04-04 at 9 42 17 PM

**Mobile View**

Screenshot 2023-04-04 at 9 50 52 PM


frontend/package.json
```diff
@@ -149,7 +149,7 @@
     "@types/dedent": "^0.7.0",
     "@types/file-saver": "^2.0.3",
     "@types/gtag.js": "0.0.10",
-    "@types/jest": "^29.2.4",
```
KenLSM, line 152: Downgraded `jest` as it is actually a lower version on `frontend/*`.
```bash
> npm why jest
jest@26.6.0 dev
node_modules/jest
  jest@"26.6.0" from react-scripts@4.0.3
  node_modules/react-scripts
    dev react-scripts@"^4.0.3" from the root project
```

foochifa: Ahh the problem is with the type package!

image

There isn't a 26.6.x in @types/jest though 😅, but the latest version for the 26.x is 26.0.24, which also seems to be the ['more'](https://www.npmjs.com/package/@types/jest?activeTab=versions) depended on one?

frontend/src/utils/urls.ts
```diff
@@ -0,0 +1,3 @@
+export const getPaymentPageUrl = (formId: string, paymentId: string) => {
```
foochifa: Perhaps this could be in public-form/utils instead?

Also as per discussed, we probably should one day clean up our URLs hahaha

frontend/src/features/public-form/components/FormPaymentPage/FormPaymentResumeModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+  return (
+    <Stack px={{ base: '1rem', md: 0 }} pt="2.5rem" pb="4rem">
+      <Modal
+        isOpen
+        onClose={() => {
+          // do nothing, prevent dismissal through backdrop touch
+        }}
+      >
+        <ModalOverlay />
```
foochifa, line 66: Slight UI margin inconsistencies with Design:

modal header to body: 16px  -> 32px(2rem)
modal body to footer: 32px -> 40px(2.5rem)
modal footer to end of modal: 44px -> 40px(2.5rem)

frontend/src/features/public-form/components/FormPaymentPage/FormPaymentResumeModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <ModalContent>
+          <ModalHeader>Restore previous session?</ModalHeader>
+          <ModalBody whiteSpace="pre-line">
```
foochifa: Small thing but not sure if we want to be more explicit with the text styles? using frontend/src/theme/textStyles.ts?

frontend/src/hooks/payments/useBrowserStm/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  storePaymentMemory: (paymentId: string) => void,
+  clearPaymentMemory: () => void,
+] => {
+  const [paymentMemory, setPaymentMemory] = useLocalStorage(
+    PAYMENT_STM_KEY,
+    JSON.stringify({}),
+  )
+
+  const entryObj = deserialize(paymentMemory || '')
+  processEviction(entryObj)
```
foochifa: Bug here, you returned the returnObj from processEviction but did not assign it back to localstorage

Hence, it won't get evicted

KenLSM:
> There isn't a 26.6.x in @types/jest though 😅, but the latest version for the 26.x is 26.0.24, which also seems to be the ['more'](https://www.npmjs.com/package/@types/jest?activeTab=versions) depended on one?

Yeah its installed as `26.0.24` in our lockfile. But it's good to keep the `package.json` updated too.

frontend/src/features/public-form/components/FormPaymentPage/FormPaymentResumeModal.tsx
KenLSM: Removed instead, it's actually not needed.

frontend/src/hooks/payments/useBrowserStm/index.ts
KenLSM: Great catch.

After looking at it again, it's not idiomatic for the throttle function to return an object either. I've swapped the implementation in 4ea4c00a to apply a side effect instead.

frontend/src/features/public-form/components/FormPaymentPage/FormPaymentResumeModal.tsx
KenLSM: Updated here c3a260d3

Side note: I thought our semantic components would be consistent with the design 🤔

foochifa: Yeah cause we directly using the modal components from chakra UI

Not urgent but since recently we have been using a lot of Modals think we should probably wrap modal components like what we did for other components in `frontend/src/components`, currently we only have a `ModalCloseButton` 

Should probably make it consistent with our figma!

foochifa (approved):
lgtm! 

Maybe just missing tests (not sure how we gonna track them when we eventually push to prod)-> 
1. Make sure local storage gets updated after proceed to pay
2. After closing and reopening form (or back), resumeModal should appear
3. Resume modal should redirect to payment page and can make payment
4. Should be able to replicate 1-3 for multiple different pending payment forms
5. Check eviction (set eviction time to 0, and comparison to `>=`

PR #6036, opened 2023-04-04
fix: add Unconnected payment channel type to fix mongoose 'feature'

### LINKED ISSUE #5848: Delete publishable key from form model when account is unlinked

justynoh:
## Problem
This PR fixes a bug introduced in #6029 due to a mongoose 'feature' which returns empty object instead of null or undefined, when a nested schema object does not exist in the DB.

Example: Schema is `Schema({ a: { a1: String, a2: String }, b: String })`. If the document is `{ b: 'Hello world!' }`, mongoose find will return `{ a: {}, b: 'Hello world!' }`.

See also: https://github.com/Automattic/mongoose/issues/5310

Closes #5848 

## Solution

Add a new payment channel for `Unconnected` to represent unconnected payment channel. This now works the same way as the `payment_field`, which has a flag representing whether or not the other fields should be looked at. Checks for `!payment_channel` is now switched to `payment_channel !== PaymentChannel.Unconnected`.

### DIFF
--- a/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
@@ -1,6 +1,6 @@
 import { Flex, FormControl, Icon, Skeleton, Text } from '@chakra-ui/react'
 
-import { FormResponseMode } from '~shared/types'
+import { FormResponseMode, PaymentChannel } from '~shared/types'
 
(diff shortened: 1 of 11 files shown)


foochifa (approved):
LGTM thanks! Does this also solve #5848?

justynoh:
Yes, thanks for pointing that out!

foochifa:
One slight thing, this will break older forms with no target_acc, pk, and channel.Stripe
image

Reconnecting will fail at updating the payment channel.

Will have to go to the DB to amend the channel to Unconnected

PR #6037, opened 2023-04-04
feat: prevent payment changes when stripe is unconnected

### LINKED ISSUE #5849: Disable admin from adding payment when stripe account has not been connected

foochifa:
## Problem

Currently, admin can still update payment fields when their stripe account isn't connected. This can cause confusion to the admin as they are able to 'enable' payment without stripe connected, but no payment can be processed. 

Closes #5849 

## Solution

Display info message when stripe isn't connected on the payment drawer,
and also invalidate query client when linking/unlinking stripe account to initiate refetch.

**Breaking Changes** 

## Screenshots

**AFTER**:
image

## Tests

Go to a storage mode form with payment beta flag and stripe unconnected
- [ ] Check payment drawer, you should not be able to edit it
- [ ] Connect stripe account
- [ ] Check payment drawer, you should be able to edit it, edit and save
- [ ] Unconnect stripe account
- [ ] Check payment drawer, you should not be able to edit it again

### DIFF
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -322,7 +322,6 @@ export const PaymentDrawer = ({
   isStripeConnected,
   paymentsField,
 }: PaymentDrawerProps): JSX.Element | null => {
-  const { isLoading } = useAdminForm()
   const { paymentData, setData, resetData } = usePaymentStore(
     useCallback(
       (state) => ({
@@ -342,7 +341,7 @@ export const PaymentDrawer = ({
   // Allows for payment data refresh in encrypt mode
   if (!paymentData && isEncryptMode && isStripeConnected) return null
 
-  const isDisabled = !(isEncryptMode && isStripeConnected) || isLoading
+  const isDisabled = !(isEncryptMode && isStripeConnected)
 
   return (
     <CreatePageDrawerContainer>
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -289,27 +289,6 @@ export const PaymentInput = ({
   )
 }
 
-// Will be extended for stripe unconnected messages too
-const PaymentDisabledMessage = ({
-  isEmailMode,
-  isStripeUnconnected,
-}: {
-  isEmailMode: boolean
-  isStripeUnconnected: boolean
-}): JSX.Element | null => {
-  return isEmailMode || isStripeUnconnected ? (
-    <Box px="1.5rem" pt="2rem" pb="1.5rem">
-      <InlineMessage variant={'info'}>
-        {isEmailMode ? (
-          <Text>Payments are not available in email mode forms.</Text>
-        ) : (
-          <Text>Connect your Stripe account in Settings.</Text>
-        )}
-      </InlineMessage>
-    </Box>
-  ) : null
-}
-
 type PaymentDrawerProps = {
   isEncryptMode: boolean
   isStripeConnected: boolean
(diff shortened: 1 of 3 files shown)


frontend/src/features/admin-form/create/payment/PaymentTab.tsx
```diff
@@ -10,17 +10,24 @@ export const PaymentTab = (): JSX.Element => {
 
   const isEncryptMode = form?.responseMode === FormResponseMode.Encrypt
 
-  return isEncryptMode ? (
+  const isStripeConnected =
+    isEncryptMode && !!form.payments_channel?.target_account_id
```
justynoh: should we use `form.payments_channel.channel !== PaymentChannel.Unconnected` instead?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -330,11 +340,16 @@ export const PaymentDrawer = ({
   }, [paymentsField, resetData, setData])
 
   // Allows for payment data refresh in encrypt mode
-  if (!paymentData && isEncryptMode) return null
+  if (!paymentData && isEncryptMode && isStripeConnected) return null
+
+  const isDisabled = !(isEncryptMode && isStripeConnected) || isLoading
```
justynoh: should we factor the "payment enabled eligibility" check (i.e. `isEncryptMode && isStripeConnected`) out into it's own variable, since it's being reused a bunch?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -330,11 +340,16 @@ export const PaymentDrawer = ({
...
 
   return (
     <CreatePageDrawerContainer>
-      <PaymentDisabledMessage isEmailMode={!isEncryptMode} />
+      <PaymentDisabledMessage
+        isEmailMode={!isEncryptMode}
+        isStripeUnconnected={!isStripeConnected}
```
justynoh: would recommend to use the "positive" versions rather than negating, for readability.

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -292,27 +293,36 @@ export const PaymentInput = ({
...
       <InlineMessage variant={'info'}>
-        <Text>Payments are not available in email mode forms.</Text>
+        {isEmailMode ? (
+          <Text>Payments are not available in email mode forms.</Text>
+        ) : (
+          <Text>Connect your Stripe account in Settings.</Text>
+        )}
       </InlineMessage>
     </Box>
   ) : null
```
justynoh: This is kind of where I think having a sharded message component starts to gets a bit clunky. The logic for whether the `PaymentDisabledMessage` is displayed is actually identical to the `isDisabled` check, and so should be linked. Right now, there is no guarantee that if the toggle is disabled, we will definitely display a message. Instead, if we use the presence of a message to determine whether or not to disable the toggle, this would mean whether or not a message is present becomes the source of truth in whether the toggle is disabled. What do you think?

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: Aite have abstracted it out, instead of isDisbaled will just check if payment is eligible[847c211](https://github.com/opengovsg/FormSG/pull/6037/commits/847c211eb4d73ad0e523c56a0dc0f2223b95cbb1) Thanks!

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: I agree 😅 thanks Justyn [fb6ace7](https://github.com/opengovsg/FormSG/pull/6037/commits/fb6ace7ba0148a38493a86a34f43d7c608697bc7)

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -329,12 +316,28 @@ export const PaymentDrawer = ({
     return resetData
   }, [paymentsField, resetData, setData])
 
+  const isPaymentEligible = isEncryptMode && isStripeConnected
+
   // Allows for payment data refresh in encrypt mode
-  if (!paymentData && isEncryptMode) return null
+  if (!paymentData && isPaymentEligible) return null
+
+  const isPaymentDisabled = !isPaymentEligible
```
justynoh: nit: should we instead define it by presence of `paymentDisabledMessage`? This will allow us to take advantage of the single source of truth of logic.

```typescript
const paymentDisabledMessage = // ... logic ...

const isPaymentDisabled = !!paymentDisabledMessage

// Allows for payment data refresh in encrypt mode
if (!paymentData && !isPaymentDisabled) return null

return (
  <CreatePageDrawerContainer>
    {paymentDisabledMessage && (...)}
    // ...
)

justynoh (approved):

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: you the best [7df6645](https://github.com/opengovsg/FormSG/pull/6037/commits/7df6645d4a19a7bf708a9b14e70000ae84e6b921) thanks!

PR #6038, opened 2023-04-04
chore: add gdrive error message for failed fetch after network error

tshuli:
## Problem
- Some instances of unknown network error could be due to gdrive uploading
- Adds appropriate error message to prompt user to download gdrive file before uploading

### DIFF
--- a/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
+++ b/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
@@ -140,9 +140,11 @@ export const PreviewFormProvider = ({
         case FormResponseMode.Email: {
           // Using mutateAsync so react-hook-form goes into loading state.
 
-          const submitEmailFormWithFetch = function (
-            isAxiosFallback?: boolean,
-          ) {
+          const submitEmailFormWithFetch = function ({
+            isAxiosFallback,
+          }: {
+            isAxiosFallback: boolean
+          }) {
             datadogLogs.logger.info(`handleSubmitForm: submitting via fetch`, {
               meta: {
                 ...logMeta,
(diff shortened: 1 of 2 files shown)


timotheeg (commented):
Just to verify: is this error specific to email mode? What is the behaviour and error message for storage mode forms?

Also warning: the fetch functions are used by both the axios fallback and the fetch mode itself. It may be OK (maybe!) to hijack the thrown error and assume network error in axios fallback mode (since we know axios did experience network error), but it is not OK to hijack the thrown error in fetch mode.

It would probably be better to detect network errors into the result flow of the fetch operation (e.g. no status code reported), and then transform the error accordingly, like we do for axios. 

I wonder if we should also update the transform Axios error function too btw (even though those are no longer shown with the fetch fallback).

tshuli:
Thanks @timotheeg! I had assumed that we would continue using fetch as a fallback, but you are right, it should only override the error message for axios network errors. I've amended accordingly

As for filtering the error messages, I was actually thinking a generic message (instead of returning the actual fetch error) makes more sense from the user perspective - it tells them to download (if uploading), or else refresh and try again, which are the remedies the user can try to take.

timotheeg:
Update on this PR: new changes made in https://github.com/opengovsg/FormSG/pull/6042 and https://github.com/opengovsg/FormSG/pull/6069 should be capturing issues with file read errors **ahead of submission time**, which means submission failures should not be caused by gdrive issues any more, and so we shouldn't need to have custom messaging about files on submission failures.

The 2 PRs mentioned above have just been released, so we'll be monitoring for a week(-ish). If the changes have the effect we expect, we will close this PR as it will no longer be required.

PR #121, opened 2020-08-11
feat: show error upon FileReader failure

mantariksh:
## Problem

We have a high frequency of "Required but no attachment content" validation errors from attachment fields, meaning the user and the frontend think that the attachment was uploaded successfully, but the backend discovers that the attachment content is actually missing. These errors occur almost exclusively on Android phones.

Closes #23 

## Solution

The error occurs when the user attempts to upload a file directly from Google Drive. The UI behaves as if the file has been uploaded correctly, but in fact `FileReader` has failed silently because we have not set an error handler on it.

The ideal solution would be to fix file uploads from Google Drive, but this is unlikely to be straightforward and likely to introduce a lot of special cases in the code since Google Drive files clearly behave differently from files in the local file system.

Hence this solution does two things:
1. Adds an error handler which prompts the user to download their file first if they are using Google Drive.
2. Only shows that the file was uploaded successfully if the file was saved as a blob successfully. This makes sense because the upload is really only successful after we have set `vm.field.file`. It also prevents situations where the UI first behaves as if the attachment was successful, then suddenly flips to an error when `FileReader.onerror` is called.

I also took the opportunity to introduce a few small refactors to simplify the attachment component.

## Screenshots
Screenshot 2020-08-11 at 10 48 47 PM

## Tests
- [ ] Upload an attachment from Google Drive on an Android phone. The error should show as per the screenshot.
- [ ] On both desktop and mobile, upload a valid attachment. The UI should behave correctly. Make sure that this works for .zip files.
- [ ] On both desktop and mobile, upload an attachment which exceeds the maximum allowed size. You should see the correct error message, and you should subsequently be able to upload a valid attachment and submit the form.
- [ ] On both desktop and mobile, upload an attachment of an invalid file type (e.g. .js). You should see the correct error message, and you should subsequently be able to upload a valid attachment and submit the form.

PR #6042, opened 2023-04-04
fix(AttachmentField): clone attachment to (hopefully) fix NetworkError

karrui:
Old bug has come to haunt us in React, see https://github.com/opengovsg/FormSG/pull/121

This is a possible fix, still requires testing on an Android phone. 

TODO: Add tests, but first need @justynoh to test whether the bug is fixed on an Android phone first (and whether it broke anything else). Feel free to take over this PR too.

EDIT: also also may have perf implications, but i think cloning max 50MB should be ok la

### DIFF
--- a/frontend/src/utils/fileArrayBuffer.ts
+++ b/frontend/src/utils/fileArrayBuffer.ts
@@ -1,0 +1,29 @@
+// Backward-compatible version of of array buffer function.
+
+function promisifyFile(obj: FileReader): Promise<ArrayBuffer> {
+  return new Promise(function (resolve, reject) {
+    obj.onload = obj.onerror = function (evt) {
+      obj.onload = obj.onerror = null
+      console.log(obj.result)
+      evt.type === 'load' && obj.result
+        ? resolve(obj.result as ArrayBuffer)
+        : reject(new Error('Failed to read the blob/file'))
+    }
+  })
+}
+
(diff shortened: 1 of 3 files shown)


frontend/src/utils/fileArrayBuffer.ts
```diff
@@ -0,0 +1,29 @@
...
+  let buffer
+  // arrayBuffer is only compatible with Safari 14 onwards
+  // for older browsers, use readAsArrayBuffer
+  if (!file.arrayBuffer) {
+    const fr = new FileReader()
+    fr.readAsArrayBuffer(file)
+    buffer = await promisifyFile(fr)
+  } else {
+    buffer = await file.arrayBuffer()
+  }
```
timotheeg: /superNit to save on negative thinking and saving 1 byte of delivery 😅 
```suggestion
  if (file.arrayBuffer) {
    buffer = await file.arrayBuffer()
  } else {
    const fr = new FileReader()
    fr.readAsArrayBuffer(file)
    buffer = await promisifyFile(fr)
  }
```

timotheeg (approved):
Thank you! 🙏

frontend/src/utils/fileArrayBuffer.ts
justynoh: [0d85540](https://github.com/opengovsg/FormSG/pull/6042/commits/0d8554063b4fde8441e9a196d7780edd9d5b0d85)

PR #6060, opened 2023-04-05
fix: getting payment response through submissionId from completedPayment

foochifa:
## Problem

Currently, going into Individual Response Page with payment will result in DB 500 errors. This is because we are trying to find the payment document by using submissionId

However, the submissionId is now nested within a 'completedPayment' object instead

## Solution

Find the payment document by comparing completedPayment.submissionId

Update individual payment response to access transaction fee data from completedPayment object

**Breaking Changes** 

## Tests

- [ ] go to a payment form, submit the form, ensure success
- [ ] go to results page in form admin, ensure that you can see the payment details & there are no errors


KenLSM (approved):
Did pair programming. LGTM!

PR #6069, opened 2023-04-06
fix(AttachmentField): correctly clear attachment on upload error

karrui:
Fix introduced in #6042 only displays the error, but still sets the invalid file on the attachment field.
This PR correctly clears the invalid attachment when there is an upload error.


timotheeg (approved):

justynoh (approved):
lgtm

PR #6073, opened 2023-04-06
fix: allow disabling payment in payment drawer

### LINKED ISSUE #6041: Unable to disable payments from payment drawer
Currently, when you try to disable payments field directly from payment drawer, you will receive a toast saying validation failed

Image

This is because the payload still includes 'description' and 'amount_cents' which should be removed if payment is disabled.

foochifa:
## Problem

Currently we are unable to disable payment from payment drawer, this PR fixes that

And also minor UI improvements to payment drawer

Closes #6041 

## Solution

Loosen validation for update payment endpoint. Previously saved details  (description and amount) should still persist even after disabling payments

Improve UI based on new figma changes:
1. Do not remove 'Name' and 'Payment Amount' components upon disabling payments (users would still like to see whats required in a payment)
2. Add description for 'Enable' toggle, waiting for @staceytan1998 to confirm the copy with Shanty
3. Remove unneeded components (stack and button) and rely on existing `FormFieldDrawerActions` and `CreatePageDrawerContainer` components
4. Minor edit to `FormFieldDrawerActions` to allow for disabling of update button
5. Shifted `paymentDisabledMessage` to after `Edit Payment` block (thanks @justynoh)

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

| Unconnected | Connected but disabled | Connected and enabled|
| --- | --- | ---|
| image | image | image |

**AFTER**:

| Unconnected | Connected but disabled | Connected and enabled|
| --- | --- | ---|
| image | image | image |

## Tests

Ensure that payment can be disabled
- [ ] go to a payment form, connect stripe account, enable payment
- [ ] go to payment drawer and try to disable payment. It should succeed, and the previous 'Name' and 'Payment Amount' should still be captured.
- [ ] open the form and try to make a submission, there should be no payment field
- [ ] enable payment again, try to make a submission, with payment

Ensure that previous functionality is kept
- [ ] go to a payment form without stripe connected, you should see an info message in payment drawer
- [ ] you should not be able to update payment field from payment drawer
- [ ] go to an email mode form, you should not be able to update payment fields from payment drawer

### DIFF
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -204,19 +204,11 @@ export const PaymentInput = ({
         isReadOnly={paymentsMutation.isLoading}
         isDisabled={isDisabled}
       >
-        {isDisabled ? (
-          <Toggle
-            description={ENABLE_PAYMENT_INFORMATION}
-            value={1}
-            label={paymentToggleLabel}
-          />
-        ) : (
-          <Toggle
-            {...register('enabled')}
-            description={ENABLE_PAYMENT_INFORMATION}
-            label={paymentToggleLabel}
-          />
-        )}
+        <Toggle
+          {...register('enabled')}
+          description={ENABLE_PAYMENT_INFORMATION}
+          label={paymentToggleLabel}
+        />
       </FormControl>
       <FormControl
         isReadOnly={paymentsMutation.isLoading}
--- a/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
+++ b/frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
@@ -12,9 +12,7 @@ import { cloneDeep } from 'lodash'
 
 import { FormPaymentsField } from '~shared/types'
 
-import { useIsMobile } from '~hooks/useIsMobile'
 import { centsToDollars, dollarsToCents } from '~utils/payments'
-import Button from '~components/Button'
 import FormErrorMessage from '~components/FormControl/FormErrorMessage'
 import FormLabel from '~components/FormControl/FormLabel'
 import InlineMessage from '~components/InlineMessage'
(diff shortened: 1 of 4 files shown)


frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
```diff
@@ -202,89 +196,70 @@ export const PaymentInput = ({
...
+            description={ENABLE_PAYMENT_INFORMATION}
+            value={1}
+            label={paymentToggleLabel}
+          />
+        ) : (
+          <Toggle
+            {...register('enabled')}
+            description={ENABLE_PAYMENT_INFORMATION}
+            label={paymentToggleLabel}
+          />
```
timotheeg: /nit 

This looks a little funny now 🤔 . I realize this is just an extension of the previous code, but can the conditional branching be done away with? As in: can there be a single `Toggle` component that uses `isDisabled` as its property? (What does `value=1` mean btw?)

Or if not, perhaps we could still extract common properties to add them there? (assuming react allows multiple spread operators?)

```typescript
const toggleSharedProps = {
    description: ENABLE_PAYMENT_INFORMATION,
    label: paymentToggleLabel,
}

<Toggle
    {...toggleSharedProps}
    value={1}
/>


<Toggle
    {...register('enabled')}
    {...toggleSharedProps}
/>
```

timotheeg (approved):

frontend/src/features/admin-form/create/payment/PaymentDrawer.tsx
foochifa: Thanks for the catch @timotheeg! You are right and the conditional is not required anymore! (we used value=1 to hard code enabled=false in the toggle 😅, but its no longer needed after we refactored the components and UsePaymentStore previously) [29dfac3](https://github.com/opengovsg/FormSG/pull/6073/commits/29dfac3decd7eee65ffeea287f5671bed54759ec)

timotheeg (approved):
again 👍 😄

Issue #6051, opened 2023-04-05
POC fetching connected account address and GST using Stripe APIs (for invoice creation)


amitogp:
We need to display these two fields on the receipt/invoice.

Currently, we don't collect these information on the Admin side, thus, we'll have to find a way to obtain these information.

KenLSM:
GST: https://dashboard.stripe.com/settings/taxation
Address: https://dashboard.stripe.com/settings/public

KenLSM:
@amitogp we can't fetch information of connected accounts. We'll need to collect those information manually.

KenLSM:
> We will go with the approach of storing the GST/Address information on agency level (editable by Kenneth through the ProdOps admin portal).

> In the next phase, we can expose this as an editable field on forms/setting/payment page by the form admins, if it makes sense.

Tracking in #6078

PR #6074, opened 2023-04-06
feat: payment invoice

### LINKED ISSUE #6052: Generate invoice based on Stripe receipt and make it downloadable
Blocked by:
- #6051 

Following is @timotheeg's design:

Image

Following are the changes suggested by JTC:
Change to - Invoice from JTC Corporation
Invoice # present – ok
JTC GST reg– ok (Note: actual number should be MB81000306)
JTC address present – ok
Change to - Invoice Date
Amount paid - ok
Description of goods/service -  this will follow the FormSG input e.g. Tender packet or IGLS packet (for Amy to confirm)
Amount charged (includes GST) – ok
Pending Amy to provide the email address for customer contact point.

KenLSM:
## Problem

JTC needs us to issue invoices for payees. The default receipt doesn't contain enough information for payees to regard it as an invoice.

Closes #6052, #6078

## Solution

Integrating Tim's Script to preprocess the receipt before returning it out to the user to download.

This feature requires another issue to be resolved:
- #6051 

**Breaking Changes** 

**Additional Notes**
1. The receipt download functionality is merely hidden, and can be easily toggle back on. As we are not definite on whether customizing our invoices this way and whether all payees require this specific invoice format, it is best that we keep our receipt functionality, in the event that we need to switch back.

2. I've also updated the receipt urls to better match RESTful API design pattern (although it's still not looking the best with `:formId` in the middle of nowhere

**After Screenshots**
Screenshot 2023-04-06 at 9 26 24 PM

### DIFF
--- a/src/app/modules/payments/stripe.controller.ts
+++ b/src/app/modules/payments/stripe.controller.ts
@@ -296,6 +296,15 @@ export const checkPaymentReceiptStatus: ControllerHandler<{
     })
 }
 
+/**
+ * Handler for GET /api/v3/payments/:formId/:paymentId/invoice/download
+ * Receives Stripe webhooks and updates the database with transaction details.
+ *
+ * @returns 200 if webhook is successfully processed
+ * @returns 404 if the PaymentId is not found
+ * @returns 404 if the FormId is not found
+ * @returns 404 if payment.completedPayment?.receiptUrl is not found
+ */
 export const downloadPaymentInvoice: ControllerHandler<{
   formId: string
   paymentId: string
@@ -310,8 +319,11 @@ export const downloadPaymentInvoice: ControllerHandler<{
     },
   })
 
-  return PaymentService.findPaymentById(paymentId)
-    .map((payment) => {
+  return ResultAsync.combine([
+    PaymentService.findPaymentById(paymentId),
+    FormService.retrieveFullFormById(formId).andThen(checkFormIsEncryptMode),
+  ])
+    .map(([payment, populatedForm]) => {
       logger.info({
         message: 'Found paymentId in payment document',
         meta: {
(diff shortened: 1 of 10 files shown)


src/app/modules/payments/stripe.utils.ts
```diff
@@ -329,3 +331,48 @@ export const mapRouteErr = (error: ApplicationError) => {
...
+/**
+ * Converts receipt sourced from Stripe into an invoice format
+ * @param receiptHtmlSource
+ */
+export const convertToInvoiceForrmat = (receiptHtmlSource: string) => {
+  // handle special characters in addresses
+  const ADDRESS = encode('8 Jurong Town Hall Road; Singapore 609434')
+  const GST_REG_NO = encode('M91234567X')
+
+  fs.writeFileSync('./' + Date.now(), receiptHtmlSource)
```
timotheeg: I think this is not needed 😅 . It was there in the script so I could see the receipt files locally, but it's meaningless to write to disk on the servers.

KenLSM: Oops. Forgot to remove this. I'm using it to debug too 😂

KenLSM: Removed in bbb0defeed2e19db933b45a3f8ef4654667845a7

src/app/modules/payments/stripe.utils.ts
```diff
@@ -329,3 +330,50 @@ export const mapRouteErr = (error: ApplicationError) => {
...
+   * n-3  Divider Small
+   * n-2  Legal Copy: You're receiving this email because you made a purchase at XXX, which partners with Stripe...
+   * n-1  Divider Small
+   * n    Divider
+   */
+  const tables = document.querySelectorAll('table table table')
+
+  let toRemove = 5
+  while (toRemove--) {
+    tables[tables.length - toRemove - 1].remove()
```
timotheeg, line 376: Awww, more subtraction operations than my original script 😜

src/app/modules/payments/stripe.controller.ts
```diff
@@ -295,7 +296,107 @@ export const checkPaymentReceiptStatus: ControllerHandler<{
...
+                  'Some business info not available during invoice generation',
+                meta: {
+                  action: 'downloadPaymentInvoice',
+                  payment,
+                  agencyName: populatedForm.admin.agency.fullName,
+                  businessInfo: businessInfo,
+                },
+              })
+            const invoiceHtml = convertToInvoiceForrmat(html, {
+              address: businessInfo.address || '',
```
timotheeg, line 361: Empty string defaults will make the invoice look weird, but that case is a setup error to correct, so I think that's fine 👍

timotheeg (approved):
LGTM 👍

src/app/modules/payments/stripe.utils.ts
KenLSM: It's 5 extra opcode per invoice, damn. 🤣

PR #4961, opened 2022-09-27
perf(v2): improve admin form performance when editing fields

darrelhong:
Continued performance optimisation from previous PR #4926 #4885. Only the the active field should re-render now, also reduces the need for debouncing input changes.

## Solution
Remove references to `useFieldBuilderStore`'s `stateData` in child components. Update builderFields using immer. Combining immer and memo seems to work to prevent FieldRowContainer from re-rendering.

Essentially builderFields is a variable dependent on `form_fields` and `useFieldBuilderStore`'s `stateData`. There is probably a cleaner way to structure it all but can't seem to put `builderFields` inside a useState without it breaking.

Form with 10 fields (local development build)
BEFORE:
Screen Shot 2022-12-29 at 9 42 34 PM

AFTER:
Screen Shot 2022-12-29 at 9 37 50 PM

PR #6076, opened 2023-04-08
refactor: scope drawer width change only to affected components

KenLSM:
## Problem

When the drawer width changes (i.e., through resizing browser, open/closing of `<CreatePageDrawerContainer />`,) the width changes and it affects all interested party of `useCreatePageSidebar` (practically almost every component on the page). On large forms, this update is significant as it has an `O(n)` impact instead of `O(count(<EditDrawer />))`

Closes [insert issue #]

## Solution

Carving out the offending context values into a separate context.

Reduce triggers to render, which also reduce the number of huge commits from 3 to 1 every time the drawer open/closes.

In the below screenshots, they are captured through [react profiler](https://legacy.reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html#component-chart). The form tested contains 300 components (exact copy of laggy forms example)
**Before**
Screenshot 2023-04-07 at 10 56 23 PM

Notice that there are 3 tall yellow bars. First yellow bar corresponds to hooks updated after open/close drawer is triggered. The subsequent two are related to `drawerRef`-related context changing, and causing another subsequent hook to update.

**After**
Screenshot 2023-04-07 at 10 54 46 PM

Notice that we only have 1 tall yellow bars. (Note that there isn't any `<EditDrawer />` in this test form. This means that the form has a faster TTI after the drawer opens/closes(!) -- the components doesn't have to care about it either way.

---

**Additional Notes**
This is part of an investigation on fixing slow pages when admin tries to edit the form on React. The context is huge, and it is containing lots of data that updates frequently, which is not something that should be included in a context.


frontend/src/features/admin-form/create/common/CreatePageDrawer/CreatePageDrawerContainer.tsx
```diff
@@ -6,14 +6,16 @@ import { useIsMobile } from '~hooks/useIsMobile'
...
+import { useCreatePageSidebarWidth } from '../CreatePageSideBarWidthContext'
 
 export const CreatePageDrawerContainer = ({
   children,
 }: {
   children: React.ReactNode
 }): JSX.Element => {
   const isMobile = useIsMobile()
-  const { isDrawerOpen, drawerRef } = useCreatePageSidebar()
+  const { isDrawerOpen } = useCreatePageSidebar()
```
timotheeg, line 17: Should we throw this into the new context as well, so all 3 drawer layout properties are together?

timotheeg (commented):
I'm not a fan of the name `useCreatePageSidebarWidth()` hey ... 🤔 I do see the merit of splitting the drawer "layout" properties versus the drawer "content" properties (the SidePage) into 2 contexts, and so I'm wondering whether what we are after in this PR is a new context named `CreatePageDrawerContainerContext` or `CreatePageDrawerContext` rather than `CreatePageSideBarWidthContext`. I'm not entirely sure I'm making sense, here 😅 , so feedback/education on my comment is welcome!

frontend/src/features/admin-form/create/common/CreatePageDrawer/CreatePageDrawerContainer.tsx
KenLSM: `isDrawerOpen` has a lot of other dependencies. I see the point of categorizing drawer related items into `useDrawerContext` but for this case, the width (layout property) needs to be craved, as it is updated less often. 

This `isDrawerOpen` function doesn't need to be in a context either. (`Context` is a powerful tool, but I see that it's used too indiscriminately here)
```
const [activeTab, setActiveTab] = useState<DrawerTabs | null>(null)
const isDrawerOpen = useMemo(
  () => activeTab !== null && activeTab !== DrawerTabs.Logic,
  [activeTab],
)

// activeTab is used by a lot of other handlers in `CreatePageSidebarContext`
```

KenLSM: If we throw in `isDrawerOpen` back into the new `useDrawerContext`, we will still see the same problem of components updating once `activeTab` updates, due to the use of context deep in the components. 

As I notice that most of the subscribed components only care about 1-2 fields in the `useCreatePageSidebar` context., to properly solve this performance issue where most of the components are triggered to re-render (even if they don't actually need to), we will need to look at the viability of lifting the state upwards (and passing only the required data downwards, or have better use of Composition),

KenLSM: Back on the naming... I see that it's horribad as well 😆. Open for a better naming!

`CreatePageDrawerContext` or `CreatePageDrawerWidthContext` makes sense too.

justynoh: I wonder if the original context is really needed anymore. We've taken out `drawerWidth` and `drawerRef`, other than that the only layout variable left is `isDrawerOpen`. But that computation itself is cheap, and is only used once in `CreatePageDrawerContainer`, so maybe we can move that out into the container component itself. At that point, the original context only exists to track state and handle state manipulation between active/pending and builder/design/logic/endpage tabs. Could this be achieved with zustand? Would this help to improve the performance? (I don't know - looking for answers for my own learning here haha)

justynoh: also for naming... maybe `useCreatePageDrawerLayoutContext`? taking inspiration from react's `useEffect`/`useLayoutEffect` hooks

KenLSM: > I wonder if the original context is really needed anymore. We've taken out drawerWidth and drawerRef, other than that the only layout variable left is isDrawerOpen. But that computation itself is cheap, and is only used once in CreatePageDrawerContainer, so maybe we can move that out into the container component itself.

Yeah, I think that makes sense.

> At that point, the original context only exists to track state and handle state manipulation between active/pending and builder/design/logic/endpage tabs. Could this be achieved with zustand? Would this help to improve the performance? (I don't know - looking for answers for my own learning here haha)

I'm not exactly sure if `zustand` can solve the problem of the subscribed childs triggering an update (due to the subscribed `context` being updated). Perhaps it does 🤔, but that change will be pretty huge.

I do see us needing to separate the handlers/dispatchers out into their own context. My intuition here is the 300 components that are retriggering doesn't care about the state change; they only want to call the handler to update some state (which they are encapsulated from).

KenLSM: Note: the performance discussion can be pretty huge. Any improvement, in the discussion above, doesn't change the initial renders of all 300 components when the page first loads. That requires lazy loading.

justynoh:
Not totally related to this PR: To tackle the issue of form builder lagginess when the content re-renders, maybe can look at how best to `memo` the `FieldRowContainer`s to reduce lagginess on drag & click etc.

can look at a previous PR #4961 for inspiration

Preliminary suspicion is that every time the state changes in `useFieldBuilderStore`, all the `FieldRowContainer`s receive the update. Maybe we should make everything underneath `BuilderFields` dumb, so we can `memo` the `FieldRowContainer`s

justynoh (approved):
lgtm

timotheeg (approved):

LinHuiqing:
Hellos! I think this PR might be causing an issue on `develop`: 

image

This happens when I try to open the payment drawer on any payment form. Does not occur when I run code on my branch out of `release-al2`.

Okay, I'm pretty certain its this PR! Payment drawer works fine when the commit for this PR is reverted.

KenLSM:
Fixing here https://github.com/opengovsg/FormSG/pull/6160

PR #6079, opened 2023-04-10
fix: make captcha in payment scrollable

### LINKED ISSUE #6002: Recaptcha for payments, if rendered beyond the screen cannot be scrolled
Unable to scroll and you have to manually zoom out

IMG_3768

foochifa:
## Problem

Currently, for payment enabled forms, we are unable to scroll the captcha, resulting in difficulties submitting the form unless the user knows to zoom out.

This is caused by chakra modal which locks scrolling when a modal is opened.

Closes #6002 

## Solution

Two potential methods were discussed:
1. Overwrite chakra modal's props to allow scrolling ([this is not recommended by chakra](https://chakra-ui.com/docs/components/modal/usage#block-scrolling-when-modal-opens) as it might lead to problem with accessibility in the future)
2. **Close the modal before submitting** (have gotten design's okay to do so)

Have decided on the second option.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

Note the scroll at the side and lack of modal
image

## Tests

- [ ] submit a form with payments enabled, make sure captcha is shown ([can do this](https://stackoverflow.com/questions/48431361/force-google-recaptcha-challenge))
- [ ] The payment modal should close before captcha is displayed
- [ ] The captcha should be scrollable (zoom in to check if form is too short)
- [ ] Submit the payment (To check original functionality is kept)

### DIFF
--- a/frontend/src/features/public-form/components/FormPaymentModal/FormPaymentModal.tsx
+++ b/frontend/src/features/public-form/components/FormPaymentModal/FormPaymentModal.tsx
@@ -41,7 +41,12 @@ export const FormPaymentModal = ({
               <Button
                 isLoading={isSubmitting}
                 loadingText="Submitting"
-                onClick={onSubmit}
+                onClick={(event) => {
+                  onClose()
+                  if (onSubmit) {
+                    onSubmit(event)
+                  }
+                }}
               >
                 Proceed to pay
               </Button>
--- a/frontend/src/features/public-form/components/FormPaymentModal/FormPaymentModal.tsx
+++ b/frontend/src/features/public-form/components/FormPaymentModal/FormPaymentModal.tsx
@@ -22,6 +22,7 @@ export const FormPaymentModal = ({
   onClose,
   isSubmitting,
 }: FormPaymentModalProps): JSX.Element => {
+  // We need to dismiss the FormPaymentModal to release the scroll lock that affects the captcha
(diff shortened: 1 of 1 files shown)


foochifa:
Thanks @KenLSM for working with me on this

frontend/src/features/public-form/components/FormPaymentModal/FormPaymentModal.tsx
```diff
@@ -22,6 +22,12 @@ export const FormPaymentModal = ({
   onClose,
   isSubmitting,
 }: FormPaymentModalProps): JSX.Element => {
+  const closeAndSubmit = (event: MouseEvent<HTMLButtonElement>) => {
+    onClose()
```
KenLSM, line 27: Would recommend adding some comments to explain the calling of `onClose()` as it is not typical for us to explicitly close the modal before submitting.

> We need to dismiss the `FormPaymentModal` to release the scroll lock that affects the captcha

KenLSM (approved):

frontend/src/features/public-form/components/FormPaymentModal/FormPaymentModal.tsx
foochifa: Thanks Ken! [2fd6bcd](https://github.com/opengovsg/FormSG/pull/6079/commits/2fd6bcd987ae36915f6b217ba4da5fdc30909c3e)

PR #6086, opened 2023-04-11
test: payment invoice

KenLSM:
Adds test cases for:
- #6074

Separated the test case from the above PR as it had already completed its review without the tests.

**Breaking Changes** 


justynoh (approved):
lgtm

Issue #5965, opened 2023-03-22
Concatenate form name and submission id to payment description


amitogp:
This is required for easy recon on the admin side and better reference on the payer side.
Example: Description is "Tender package sale"
On receipt it should show up as: 
"Tender package sale
FormSG form: xxxxx xxxxx xxx xxxx
Submission id: xxxxxxxxxxxxxxxxxx"

justynoh:
Submission id blocked by 
- #6054

PR #6088, opened 2023-04-12
feat: stable submission id

### LINKED ISSUE #5960: Add new items to metadata

### LINKED ISSUE #6080: Implement stable submission ID
Need to catch error with copying submission ID over and report an error log. Then retry with new ID

justynoh:
## Problem
This PR implements the stable submission ID required for identification of submissions throughout the lifecycle. Closes #6080

Additionally, took the opportunity to do two items that were blocked by stable submission id. Closes #5960 and #5965

## Solution

Explicitly copy over the pending submission ID into the submission collection. Add duplicate key error handling and recovery.

## Screenshots

Receipt

image

Stripe dashboard

image

Stripe metadata

image

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -8,8 +8,6 @@ import { okAsync } from 'neverthrow'
 import Stripe from 'stripe'
 import type { SetOptional } from 'type-fest'
 
-import { StripePaymentMetadataDto } from 'src/types'
-
 import {
   ErrorDto,
   FormAuthType,
@@ -21,6 +19,7 @@ import {
   SubmissionErrorDto,
   SubmissionResponseDto,
 } from '../../../../../shared/types'
+import { StripePaymentMetadataDto } from '../../../../types'
 import { EncryptSubmissionDto } from '../../../../types/api'
 import { paymentConfig } from '../../../config/features/payment.config'
 import { createLoggerWithLabel } from '../../../config/logger'
@@ -432,9 +431,12 @@ const submitEncryptModeForm: ControllerHandler<
     const metadata: StripePaymentMetadataDto = {
       formTitle: form.title,
       formId,
+      submissionId: pendingSubmissionId,
       paymentId,
       paymentContactEmail: paymentReceiptEmail,
     }
+
+    const paymentReceiptDescription = `${form.payments_field.description}\nFormSG form: ${form.title}\nResponse ID: ${pendingSubmissionId}`
 
     const createPaymentIntentParams: Stripe.PaymentIntentCreateParams = {
       amount,
@@ -443,7 +445,7 @@ const submitEncryptModeForm: ControllerHandler<
         'card',
         /* 'grabpay', 'paynow'*/
       ],
-      description: form.payments_field.description,
+      description: paymentReceiptDescription,
       receipt_email: paymentReceiptEmail,
       metadata,
     }
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -436,7 +436,11 @@ const submitEncryptModeForm: ControllerHandler<
       paymentContactEmail: paymentReceiptEmail,
     }
 
-    const paymentReceiptDescription = `${form.payments_field.description}\nFormSG form: ${form.title}\nResponse ID: ${pendingSubmissionId}`
+    const paymentReceiptDescription = [
+      form.payments_field.description,
+      `FormSG form: ${form.title}`,
+      `Response ID: ${pendingSubmissionId}`,
+    ].join('\n')
 
     const createPaymentIntentParams: Stripe.PaymentIntentCreateParams = {
       amount,
--- a/src/app/modules/submission/submission.service.ts
+++ b/src/app/modules/submission/submission.service.ts
@@ -275,47 +275,46 @@ export const copyPendingSubmissionToSubmissions = (
           meta: logMeta,
           error,
         })
-
-        const databaseError = transformMongoError(error)
-
-        return databaseError instanceof DatabaseDuplicateKeyError
-          ? databaseError
-          : new DatabaseError(getMongoErrorMessage(error))
+        return transformMongoError(error)
       }).orElse((error) => {
-        if (error instanceof DatabaseDuplicateKeyError) {
-          // Failed to save due to duplicate keys.
-          logger.error({
-            message:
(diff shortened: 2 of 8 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -432,18 +431,21 @@ const submitEncryptModeForm: ControllerHandler<
     const metadata: StripePaymentMetadataDto = {
       formTitle: form.title,
       formId,
+      submissionId: pendingSubmissionId,
       paymentId,
       paymentContactEmail: paymentReceiptEmail,
     }
 
+    const paymentReceiptDescription = `${form.payments_field.description}\nFormSG form: ${form.title}\nResponse ID: ${pendingSubmissionId}`
```
KenLSM: /nit Would this make the description content clearer?
```suggestion
    const paymentReceiptDescription = [
      form.payments_field.description,
      `FormSG form: ${form.title}`,
      `Response ID: ${pendingSubmissionId}`
    ].join("\n")
```

src/app/modules/submission/submission.service.ts
```diff
@@ -251,17 +258,64 @@ export const copyPendingSubmissionToSubmissions = (
...
+            error,
+          })
+          // Recover by attempting to save with a different id.
+          const recoverySubmission = new SubmissionModel(submissionContent)
+          // TODO: Set alarms for both branches
+          return ResultAsync.fromPromise(
+            recoverySubmission.save({ session }),
+            (error) => {
+              logger.error({
+                message: 'Failed to recover from duplicate key error',
```
KenLSM: Not related to PR but I'm wondering if critical errors like this should have its own error code like `internalCode: P001` instead of a message.

justynoh:
Also i like how you can see the counter ticking on the last screenshot (`paymentId`: ...`678`, `submissionId`: ...`679`)

KenLSM (approved):
LGTM! Some non-blocking comments

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
justynoh: omg smartttt

EDIT: [cfc6e50](https://github.com/opengovsg/FormSG/pull/6088/commits/cfc6e5094cb663ad821e6f9fe83eb30f9bc19dbf)

src/app/modules/submission/submission.service.ts
justynoh: it could make our alarm system more fine-grained, if we can set alarms by the error code returned rather than just the status code... maybe one for the tech backlog!

src/app/modules/submission/submission.service.ts
```diff
@@ -251,17 +258,64 @@ export const copyPendingSubmissionToSubmissions = (
...
         })
-        return new DatabaseError(getMongoErrorMessage(error))
+
+        const databaseError = transformMongoError(error)
+
+        return databaseError instanceof DatabaseDuplicateKeyError
+          ? databaseError
+          : new DatabaseError(getMongoErrorMessage(error))
+      }).orElse((error) => {
+        if (error instanceof DatabaseDuplicateKeyError) {
```
LinHuiqing: seems to me that the first `instanceof DatabaseDuplicateKeyError` check is redundant?

src/app/modules/submission/submission.service.ts
```diff
@@ -251,17 +258,64 @@ export const copyPendingSubmissionToSubmissions = (
...
-      )
+      const submissionContent = omit(pendingSubmission, [
+        '_id',
+        'created',
+        'lastModified',
+      ])
+      const submission = new SubmissionModel({
+        // Explicitly copy over the pending submission's _id
+        _id: pendingSubmissionId,
+        ...submissionContent,
```
LinHuiqing:
```suggestion
        ...submissionContent,
        _id: pendingSubmissionId,
```
nit: since we want to make sure the `_id` for this part is `pendingSubmissionId`, right?

src/app/modules/submission/submission.service.ts
justynoh: `error` could be either `DatabaseDuplicateKeyError` or `DatabaseError`. Only in the former case do we want to retry, in the latter case we don't. Am i misunderstanding what you're referring to?

src/app/modules/submission/submission.service.ts
justynoh: makes sense, [431ec70](https://github.com/opengovsg/FormSG/pull/6088/commits/431ec7014bce214100600093b9f93d5e6f56801a)

src/app/modules/submission/submission.service.ts
LinHuiqing: Oh, as in because the `DatabaseError` can be wrapped in line 318 instead, right? And right after the `orElse`, the case of `if (error instanceof DatabaseDuplicateKeyError)` is handled, so the return statement here can just `return databaseError`?

src/app/modules/submission/submission.service.ts
```diff
@@ -251,17 +258,64 @@ export const copyPendingSubmissionToSubmissions = (
...
           error,
         })
-        return new DatabaseError(getMongoErrorMessage(error))
+
+        const databaseError = transformMongoError(error)
+
+        return databaseError instanceof DatabaseDuplicateKeyError
+          ? databaseError
+          : new DatabaseError(getMongoErrorMessage(error))
+      }).orElse((error) => {
```
LinHuiqing: Could you add tests for the retrying logic? Seems like its a pretty important mechanism for payments that we will not want to break with future code changes.

src/app/modules/submission/submission.service.ts
justynoh: oh, i see what you mean... hmm, either is ok i think, my view was that it helps to clarify the types of the errors going into the `orElse` more explicitly (to be sure that it's one of `DatabaseDuplicateKeyError` or `DatabaseError`). But you're right that we could be less verbose and achieve the same effect!

justynoh: [ea85744](https://github.com/opengovsg/FormSG/pull/6088/commits/ea857445b34cc99eb35b3e00e3c16f278f91ed4b)

src/app/modules/submission/submission.service.ts
justynoh: done in [44796fa](https://github.com/opengovsg/FormSG/pull/6088/commits/44796fa5173cfd6188b0a76d7e4460a39f8f403c)

LinHuiqing: aights lgtm 👍

PR #6089, opened 2023-04-12
feat: perform post-submission actions when payment is completed

### LINKED ISSUE #6000: Ensure that post submission actions get done when pending submissions are confirmed
Specifically, need to make sure
- Webhooks are fired
- Email confirmations are sent

Basically, we need to factor out everything after the payment block in encrypt submission controller and make sure those get called when submissions are confirmed too.

justynoh:
## Problem
Previously, webhooks will be fired and email confirmations will be sent once the submission is complete. This was not done in the payment flow.

Closes #6000

## Solution
Shard out the post-submission actions into a new service layer function `performEncryptPostSubmissionActions`. This is now called both in the encrypt submission controller and the payment confirmation function.

In principle, this sounds easy, but e2ee has bitten us again! We send the information required for email confirmations in the clear from FE to BE in `responses`, and then we don't store the information beyond that. Therefore, there is no persistent storage of this information that allows us to delay the sending of email confirmations.

image

With that in mind, there's only one way forward - store the responses somewhere, somehow in the DB. See below for details, but to cut a long story short, I decided to store in the payment collection for now, in the clear just because it's the least effort to do. That's why there are modifications to the payment model and types.

Feel free to give your opinions on the following two questions! Thanks 🙏  

## Discussion points

**Where to store `responses`?**
1. In the pending submissions collection. This means there will be a divergence in schema between pending submissions and submissions collection. This seems like the most natural place to put it, but will require some additional instrumentation with new services and utils for type-guarding encrypt pending submissions and email pending submissions, a la submissions currently. I don't think it's _that_ much work, but didn't want to invest effort until we decided on something.
2. In the payments collection. This means that for the first time we store actual submission data in the payments collection. This is the easiest to implement, since the payment collection is independent from everything else. And it makes _some_ sense (storing the information required only for "delayed completed submissions" in the metadata for _why_ it's delayed). [TODO: If we proceed with this, we will need to delete the data from the payment collection once the email confirmations are sent.]
3. In an entirely separate collection. This is similar to the logic of OTP verifications - we have a separate collection, but we don't expect the data to be stored there for long, and once we're done we later delete it. This isn't very high in effort, but will require us to keep track of more collections, all related to payments. Besides, payments and submissions are 1 to 1, so I don't think there's any tangible benefit of this option above option 2 in any case.

**How to store `responses`?**
1. In the clear. Of course, this "breaks" our promise of e2ee **on the form responses** (we already store the payment contact email in the clear). Are we willing to accept this?
2. Masked. This is similar to the approach for OTPs. The difference is that OTPs don't need to be "unhashed" - we just compare the hash values, so we don't store OTPs in the clear. In contrast, for responses, if we want to mask the data, the only option is an encryption mechanism, which we could do but require some effort (generate key, store in secret, etc. etc.).

### DIFF
--- a/shared/types/payment.ts
+++ b/shared/types/payment.ts
@@ -35,7 +35,6 @@ export type Payment = {
   email: string
   amount: number
   paymentIntentId: string
-  responses: any[] // Will be typed as FilteredResponse[]
 
   // Payment status tracking
   webhookLog: Stripe.Event[]
--- a/shared/types/payment.ts
+++ b/shared/types/payment.ts
@@ -35,6 +35,7 @@ export type Payment = {
   email: string
   amount: number
   paymentIntentId: string
+  responses: any[] // Will be typed as FilteredResponse[]
 
   // Payment status tracking
   webhookLog: Stripe.Event[]
--- a/src/app/models/payment.server.model.ts
+++ b/src/app/models/payment.server.model.ts
@@ -28,6 +28,7 @@ const PaymentSchema = new Schema<IPaymentSchema, IPaymentModel>(
       type: String,
       required: true,
     },
+    responses: [],
 
     webhookLog: [],
     status: {
--- a/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
@@ -6141,11 +6141,9 @@ describe('admin-form.controller', () => {
           responses: MOCK_RESPONSES,
           form: MOCK_FORM,
           encryptedContent: MOCK_ENCRYPTED_CONTENT,
-        } as IncomingEncryptSubmission),
+        } as unknown as IncomingEncryptSubmission),
       )
-      MockSubmissionUtils.extractEmailConfirmationDataFromIncomingSubmission.mockReturnValue(
-        [],
-      )
+      MockSubmissionUtils.extractEmailConfirmationData.mockReturnValue([])
       MockEncryptSubmissionService.createEncryptSubmissionWithoutSave.mockReturnValue(
         MOCK_SUBMISSION,
       )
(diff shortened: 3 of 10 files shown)


justynoh:
Update: After discussion with @amitogp and @timotheeg, we have decided that storing them in the clear in the db is ok, as long as it is transient. Therefore, once we implement the CRON job for payment cleanup, we're all good.

What's left is to decide on where they should be stored.

LinHuiqing:
Just to note: Regarding payment contact email, I think we might want to store them as hashes instead since we were concerned with storing them in the clear, and the main benefit of having them in our DB is simply for reconciliation.

justynoh:
@LinHuiqing if we need to modify our implementation of payment contact email, can we track in a separate issue? Since this PR is not related to that haha. Thanks!

LinHuiqing:
> @LinHuiqing if we need to modify our implementation of payment contact email, can we track in a separate issue? Since this PR is not related to that haha. Thanks!

Yup, definitely! Just thought to bring it up since it was brought up as a "storing stuff in the clear" precedent haha

justynoh:
## Tests

- [ ] Ensure that normal storage submission forms still work. 
  - [ ] Create a storage mode form with an associated webhook URL. The storage mode form should have at least one email confirmation field. 
  - [ ] Fill the form.
  - [ ] Ensure that email confirmations and webhooks can still be received normally.
- [ ] Ensure that the flow works for payment forms. 
  - [ ] Create a storage mode form with an associated webhook URL. The storage mode form should have at least one email confirmation field. Add a payment to it. 
  - [ ] Fill the form and click "Proceed to pay". 
  - [ ] Ensure that the payment document contains at least one item in the `responses` field. 
  - [ ] Then, complete the payment. 
  - [ ] Ensure that email confirmations and webhooks can be received normally. 
  - [ ] Finally, ensure that the payment document contains `responses: []`.

LinHuiqing (commented):
Lgtm other than that it's missing manual tests! Some tests we probably want:

- Sending of webhooks on payment forms (and non-payment forms to make sure existing stuff don't break)
- Sending of emails on payment and non-payment forms

LinHuiqing:
oh oop its out of order (how do you type so fast??)

Issue #5961, opened 2023-03-21
Stop respondent from making payment if payment intent has been canceled


amitogp:
Dependency
- #6001

amitogp:
Blocked by dedicated payment page.

amitogp:
Waiting for Design
- #5958

PR #6091, opened 2023-04-12
feat: payment processing and canceled state

### LINKED ISSUE #6043: Support Processing state in Payment Page
We don't have a `processing` state that's in Payment Page. This is a valid state that can be returned from `stripe`.

https://github.com/opengovsg/FormSG/pull/6001#discussion_r1157019528

It is a MUST for non-card payments such as paynow

KenLSM:
## Problem

There are two additional states that Stripe can return, namely payment canceled, and payment processing. The UI for these are to show the payment UI, but this is not expected as the responder has already paid.

Closes #6043, #5961 

## Solution

We will account for the two additional states and add the message blocks that indicate the correct actions/messages for them.

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-04-13 at 11 56 47 AM

**AFTER**
**Canceled**
Screenshot 2023-04-13 at 12 01 29 PM

**Processing**
Screenshot 2023-04-13 at 11 53 29 AM

**Feedback after submission has been confirmed**
Screenshot 2023-04-13 at 1 25 11 PM

**Tests**
- [ ] payment intents at canceled state should show canceled state
    - [ ] create a payment intent, go to stripe dashboard and cancel the payment intent, expect to show a canceled UI
- [ ] submission feedback form should **only** appear after payment has `succeeded`
    - [ ] create and complete a payment intent, expect that submission feedback should only appear after payment completes
    - [ ] should be able to submit and show on admin
Note: Processing is not testable from stripe. Any ideas, aside from explicitly setting it on FE?


frontend/src/features/public-form/components/FormPaymentPage/stripe/StripeElementWrapper.tsx
```diff
@@ -85,51 +67,69 @@ const StripePaymentContainer = ({
...
+              publishableKey={paymentInfoData.publishableKey}
+            />
+          </PaymentStack>
+        )
+      case PaymentViewStates.Canceled:
+        return (
+          <PaymentStack>
+            <GenericMessageBlock
+              paymentId={paymentId}
+              title={'Payment request was canceled.'}
```
LinHuiqing:
```suggestion
              title="Payment request was canceled."
```
nit: since we're just using a string directly, just use quotation marks? There are quite a few instances of this throughout.

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
```diff
@@ -161,7 +161,10 @@ export const StripePaymentBlock = ({
     return 'Please make payment.'
   }, [formTitle])
 
-  return form?.responseMode === FormResponseMode.Encrypt ? (
+  if (form?.responseMode !== FormResponseMode.Encrypt) {
+    return <></>
+  }
```
LinHuiqing: In this case, should we be redirecting to some other page / throwing an error instead? Rather than silently return an empty diff. 
Not sure if this is important this decision is important to the scope of this PR though.

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
```diff
@@ -181,7 +184,6 @@ export const StripePaymentBlock = ({
             S${centsToDollars(form?.payments_field?.amount_cents || 0)}
           </Text>
         </Text>
-
```
LinHuiqing, line 184: eeyer more of your LOC changes HAHA

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
LinHuiqing: Ah, I see that returning `<></>` is something we already do originally. Just curious, why did you extract it out? Is it for better readability or something?

LinHuiqing (approved):
code generally lgtm - add manual tests to be done?

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
KenLSM: Shhhh!

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
KenLSM: Generally we should try to leave the happy path all the way to the last. Moving this up will allow the reader to understand that this component will short-circuit if we are not in Encrypt mode.

I found that I should also add `!form` in this comparator.

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
KenLSM: Probably I tried removing `<Element />` somewhere and realized that it looked better back here 🤔

frontend/src/features/public-form/components/FormPaymentPage/stripe/StripeElementWrapper.tsx
justynoh: are you using prettier? eslint should autoformat on save

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
LinHuiqing: Oh? Do we want to use a non-null assertion here though? The component will completely break if the public form context returns undefined.

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
```diff
@@ -181,7 +184,6 @@ export const StripePaymentBlock = ({
             S${centsToDollars(form?.payments_field?.amount_cents || 0)}
```
LinHuiqing:
```suggestion
            S${centsToDollars(form?.payments_field?.amount_cents ?? 0)}
```
I know this isn't part of this PR but I just saw it - shouldn't this use `??` instead? Because we only wanna default to 0 for nullish values
@foochifa since this was part of your commit some time ago haha

frontend/src/features/public-form/components/FormPaymentPage/stripe/StripeElementWrapper.tsx
KenLSM: Hm my eslint is happy with it being wrapped with curly braces, does it complain on your side?

LinHuiqing: Actually mine is happy too hahaha

KenLSM: Addressed in 87bf06cb

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
KenLSM: Yeah. So we can remove the need to continuously check if `form` is falsey. I.e., later in the code there's a `form?.payments_field?.amount_cents`.

Addressed in 4022f596

KenLSM: It is true though that `PaymentBlock` shouldn't need to check for `form`, as it is presumed to be present. I would also prefer to change every component to be less smart...

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
KenLSM: I think for numbers (and if we trust that we will only accept numbers), it should be fine since the values that could be falsy (`undefined, null, 0`) should be treated as 0

PR #6025, opened 2023-03-31
fix: remove timezone from datepicker

LinHuiqing:
## Problem

Date picker was using the midnight of the date picked and taking into account the timezone where the user is filling in the form. However, it does not make sense to do so as users will be picking particular dates rather than timings, and the understanding of the different timezones should be established as part of the question.

Closes #6024

## Solution

Remove the portions of the code to account for timezone differences.

**Breaking Changes** 

- [x] No - this PR is backwards compatible. This the pure frontend change.

**Bug Fixes**:

- Remove timezone conversions previously included in datepicker

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/37061143/229051162-f3d111a4-d7ce-406f-b6eb-b659891296c7.mov

**AFTER**:

https://user-images.githubusercontent.com/37061143/229052471-3244b976-48d7-4c39-96e0-0c0d12a5487e.mov

## Tests

To test that the issue is fixed, [set your device's timezone](https://support.apple.com/en-gb/guide/mac-help/mchlp2996/mac) to a timezone with negative UTC offset.
- [ ] Create a form with a datepicker field.
- [ ] Open the form as a respondent and use the datepicker. The selected dates should be correctly reflected in the fields. Submit the form with the selected date on the datepicker.
- [ ] As an admin, check the responses. The latest response should have the right date indicated on the datepicker by the respondent.
- [ ] As an admin, edit the datepicker and try to set validation with a custom date range. The dates selected for the date range should be reflected correctly as well.

To test the fix did not compromise on any existing datepicker functionality, change the timezone to one with a positive UTC offset and repeat the tests.
- [ ] Open the form as a respondent and use the datepicker. The selected dates should be correctly reflected in the fields. Submit the form with the selected date on the datepicker.
- [ ] As an admin, check the responses. The latest response should have the right date indicated on the datepicker by the respondent.
- [ ] As an admin, edit the datepicker and try to set validation with a custom date range. The dates selected for the date range should be reflected correctly as well.
- [ ] For email mode form, disallow past dates / future dates validation should still work based on server time even if client timezone changes

PR #6096, opened 2023-04-13
fix: date-picker bug for negative UTC timezones

LinHuiqing:
## Problem

> (taken from #6025) 
> Date picker was using the midnight of the date picked and taking into account the timezone where the user is filling in the form. However, it does not make sense to do so as users will be picking particular dates rather than timings, and the understanding of the different timezones should be established as part of the question.

However, bug fix in PR #6025 almost caused an incident as the fix removed converting to UTC on the date picker itself, which was causing [the selection offset bug](https://formsg.zendesk.com/agent/tickets/42733). 

## Solution

Puts back the local time to UTC conversion right before submission so that dates saved in the DB are normalised to be in UTC time.

Note: local time to UTC conversion does not have to be restored for `DateField` as the responses for `DateField` are in string when encrypted anyway.

**Breaking Changes** 

**Improvements**:

- Cast time to UTC only right before submitting custom date so the conversion will be completely invisible to users
- Cast time to local timezone before being shown on the frontend so the frontend timezone will be consistent

**Bug Fixes**:

- Users with negative UTC timezones will now have their date fields correctly reflecting the dates selected on their date pickers
- `zonedTimeToUTC` was converting the timings to UTC and thus the actual date being referred to will not be preserved for respondents in negative UTC timezones. Implemented `normalizeDateToUtc` to ensure the actual date numbers are are preserved.
  - We now cast time from local timezone to UTC timezone using `Date.UTC` instead of converting time using `zonedTimeToUTC` (which we were using wrong anyway, as pointed out by @darrelhong in #5731)

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/37061143/229051162-f3d111a4-d7ce-406f-b6eb-b659891296c7.mov

**AFTER**:

Disallow past dates:

https://user-images.githubusercontent.com/37061143/231920965-0295168f-ea05-4fad-a3fe-bf6d890350cc.mov

Disallow future dates:

https://user-images.githubusercontent.com/37061143/231939426-16e43f97-7810-4429-b393-fc62d86d963c.mov

Custom date range:

https://user-images.githubusercontent.com/37061143/231940842-d14a3731-7d50-41d4-8a54-326581fd2135.mov

## Tests

To ensure that all our bases are covered, try to test the most extreme timezones. Here are the suggested timezones: 
- Most common use case: Singapore - Singapore (UTC +08)

Preparation:
- [x] Create a form with a date field and open the form.

For the tests below, timezones A and B refer to the positive and negative timezones mentioned above, in any order.

Ensure that date picker for date field with "Disallow past dates" option exhibits expected behaviour:
1. - [x] Set your timezone as A.
2. - [x] On the admin panel, refresh, select and save the "Disallow past dates" option for the date field.
3. - [x] On the respondent form, refresh and open the date picker. Only the current and future dates of the timezone should be shown.
4. - [x] Select a date. The appropriate date should be reflected in the date field.
5. - [x] Change your device's timezone to B.
6. - [x] Repeat tasks 3-4.
7. - [x] Repeat tasks 2-4.
9. - [x] Change your device's timezone to A.
10. - [x] Repeat tasks 3-4.

Ensure that date picker for date field with "Disallow future dates" option exhibits expected behaviour:
1. - [x] Set your timezone as A.
2. - [x] On the admin panel, refresh, select and save the "Disallow future dates" option for the date field.
3. - [x] On the respondent form, refresh and open the date picker. Only the current and future dates of the timezone should be shown.
4. - [x] Select a date. The appropriate date should be reflected in the date field.
5. - [x] Change your device's timezone to B.
6. - [x] Repeat tasks 3-4.
7. - [x] Repeat tasks 2-4.
9. - [x] Change your device's timezone to A.
10. - [x] Repeat tasks 3-4.

Ensure that date picker for date field with "Custom date range" option exhibits expected behaviour:
1. - [x] Set your timezone as A.
2. - [x] On the admin panel, refresh and select the "Custom date range" option for the date field.
3. - [x] Select a start date.
4. - [x] Open the end date field date picker. Only the dates on or after the selected start date should be selectable.
5. - [x] Select an end date.
6. - [x] Save the selected date range. Optionally, check that the dates saved in the DB are in UTC.
7. - [x] On the respondent form, refresh and open the date picker. Only the dates in the selected date range should be selectable.
8. - [x] Select a date. The respective date should be reflected in the date field.
9. - [x] Change your device's timezone to B.
10. - [x] Repeat tasks 7-8.
11. - [x] Repeat task 2. Ensure that the start and end dates remain the same.
12. - [x] Update the start and end dates for the date field. Save it.
13. - [x] Repeat tasks 7-8.
14. - [x] Change your device's timezone to A.
15. - [x] Repeat tasks 7-8.

Note: Here's [how to set your Mac's timezone](https://support.apple.com/en-gb/guide/mac-help/mchlp2996/mac).

Design/ product question: Should "Disallow past dates" and "Disallow future dates" be taking the past / future based on the admins' timezone, specified timezone or the users' timezone? Also, even when we have decided on the previous question, does it even make sense if the concept of time is not considered? E.g. if the recurring event I want to organise is at 10am, our date picker won't be able to know that so if we have a timezone in mind, it won't "filter" properly. Currently, it's simply implemented based on the users' timezone so its not disorienting.

### DIFF
--- a/frontend/src/utils/date.ts
+++ b/frontend/src/utils/date.ts
@@ -38,10 +38,12 @@ export const isDateAfterToday = (date: number | Date) => {
   return isAfter(date, endOfToday())
 }
 
-// Converts UTC time to the same date in local time, ignoring original timezone.
-export const fromUtcToLocalDate = (date?: Date | null) => {
-  if (!date) return date
-  return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
+export const normalizeDateToUtc = (date: Date) => {
+  return new Date(date.valueOf() - date.getTimezoneOffset() * 60 * 1000)
+}
+
+export const loadDateFromNormalizedDate = (date: Date) => {
+  return new Date(date.valueOf() + date.getTimezoneOffset() * 60 * 1000)
 }
 
 /**
--- a/frontend/src/utils/date.ts
+++ b/frontend/src/utils/date.ts
@@ -38,11 +38,13 @@ export const isDateAfterToday = (date: number | Date) => {
   return isAfter(date, endOfToday())
 }
 
-export const normalizeDateToUtc = (date: Date) => {
+export const normalizeDateToUtc = (date: Date | null) => {
+  if (!date) return date
   return new Date(date.valueOf() - date.getTimezoneOffset() * 60 * 1000)
 }
 
-export const loadDateFromNormalizedDate = (date: Date) => {
+export const loadDateFromNormalizedDate = (date: Date | null) => {
+  if (!date) return date
   return new Date(date.valueOf() + date.getTimezoneOffset() * 60 * 1000)
 }
 
--- a/frontend/src/utils/date.ts
+++ b/frontend/src/utils/date.ts
@@ -40,12 +40,12 @@ export const isDateAfterToday = (date: number | Date) => {
 
 export const normalizeDateToUtc = (date: Date | null) => {
   if (!date) return date
-  return new Date(date.valueOf() - date.getTimezoneOffset() * 60 * 1000)
+  return Date.UTC(date.getFullYear(), date.getMonth(), date.getDate())
 }
 
 export const loadDateFromNormalizedDate = (date: Date | null) => {
   if (!date) return date
-  return new Date(date.valueOf() + date.getTimezoneOffset() * 60 * 1000)
+  return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
 }
 
 /**
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -431,12 +431,8 @@ export const createDateValidationRules: ValidationRuleFn<DateFieldBase> = (
         }
 
         const { customMinDate, customMaxDate } = schema.dateValidation ?? {}
-        const customMinNoTime = customMinDate
-          ? loadDateFromNormalizedDate(customMinDate)
-          : null
-        const customMaxNoTime = customMaxDate
-          ? loadDateFromNormalizedDate(customMaxDate)
-          : null
+        const customMinNoTime = loadDateFromNormalizedDate(customMinDate)
+        const customMaxNoTime = loadDateFromNormalizedDate(customMaxDate)
         return (
           !isDateOutOfRange(parseDate(val), customMinNoTime, customMaxNoTime) ||
           'Selected date is not within the allowed date range'
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -53,10 +53,10 @@ import {
 import { VerifiableFieldBase } from '~features/verifiable-fields/types'
 
 import {
-  fromUtcToLocalDate,
   isDateAfterToday,
   isDateBeforeToday,
   isDateOutOfRange,
+  loadDateFromNormalizedDate,
 } from './date'
 import { formatNumberToLocaleString } from './stringFormat'
 
@@ -431,12 +431,15 @@ export const createDateValidationRules: ValidationRuleFn<DateFieldBase> = (
         }
 
         const { customMinDate, customMaxDate } = schema.dateValidation ?? {}
+        const customMinNoTime = customMinDate
+          ? loadDateFromNormalizedDate(customMinDate)
+          : null
+        const customMaxNoTime = customMaxDate
+          ? loadDateFromNormalizedDate(customMaxDate)
+          : null
         return (
-          !isDateOutOfRange(
-            parseDate(val),
-            fromUtcToLocalDate(customMinDate),
-            fromUtcToLocalDate(customMaxDate),
-          ) || 'Selected date is not within the allowed date range'
+          !isDateOutOfRange(parseDate(val), customMinNoTime, customMaxNoTime) ||
+          'Selected date is not within the allowed date range'
         )
       },
     },
--- a/frontend/src/utils/fieldValidation.ts
+++ b/frontend/src/utils/fieldValidation.ts
@@ -431,11 +431,12 @@ export const createDateValidationRules: ValidationRuleFn<DateFieldBase> = (
         }
 
         const { customMinDate, customMaxDate } = schema.dateValidation ?? {}
-        const customMinNoTime = loadDateFromNormalizedDate(customMinDate)
-        const customMaxNoTime = loadDateFromNormalizedDate(customMaxDate)
         return (
-          !isDateOutOfRange(parseDate(val), customMinNoTime, customMaxNoTime) ||
-          'Selected date is not within the allowed date range'
+          !isDateOutOfRange(
+            parseDate(val),
+            loadDateFromNormalizedDate(customMinDate),
+            loadDateFromNormalizedDate(customMaxDate),
+          ) || 'Selected date is not within the allowed date range'
         )
       },
     },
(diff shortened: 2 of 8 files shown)


frontend/src/utils/date.ts
```diff
@@ -38,10 +38,14 @@ export const isDateAfterToday = (date: number | Date) => {
   return isAfter(date, endOfToday())
 }
 
-// Converts UTC time to the same date in local time, ignoring original timezone.
-export const fromUtcToLocalDate = (date?: Date | null) => {
+export const normalizeDateToUtc = (date: Date | null) => {
   if (!date) return date
-  return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
+  return new Date(date.valueOf() - date.getTimezoneOffset() * 60 * 1000)
```
foochifa: 1. Can use date-fns { addMinutes }
`return addMinutes(date, date.getTimezoneOffset())`

2. This will be correct for most cases but not all. That's because `.getTimezoneOffset()` do not always return a uniform number (Daylight Savings Time). [See](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset)
What could happen is that we normalize date -> not yet DST -> offset is 300
Then loading data back -> is DST -> offset is 240
Your final result will end up being different. 
Some[ relevant discussions](https://github.com/date-fns/date-fns/issues/1401#issuecomment-621897094) and [other potential solutions](https://stackoverflow.com/questions/948532/how-to-convert-a-date-to-utc)
Though honestly not sure if this is a cause for concern for us. 

[Kinda relevant again HAHAHA](https://gist.github.com/timvisee/fcda9bbdff88d45cc9061606b4b923ca)

frontend/src/utils/fieldValidation.ts
```diff
@@ -431,12 +431,11 @@ export const createDateValidationRules: ValidationRuleFn<DateFieldBase> = (
...
         const { customMinDate, customMaxDate } = schema.dateValidation ?? {}
+        const customMinNoTime = loadDateFromNormalizedDate(customMinDate)
+        const customMaxNoTime = loadDateFromNormalizedDate(customMaxDate)
         return (
-          !isDateOutOfRange(
-            parseDate(val),
-            fromUtcToLocalDate(customMinDate),
-            fromUtcToLocalDate(customMaxDate),
-          ) || 'Selected date is not within the allowed date range'
+          !isDateOutOfRange(parseDate(val), customMinNoTime, customMaxNoTime) ||
```
foochifa: nit: realised you refactored `loadDateFromNormalizedDate`, Then can just refactor `customMinNoTime, customMaxNoTime` too as its only used once, and the naming is a bit weird too.

```
!isDateOutofRange(parseDate(val),
    loadDateFromNormalizedDate(customMinDate),
    loadDateFromNormalizedDate(customMaxDate),
    ) || 'Selected date is not within the allowed date range'
```
Same for `DateField.tsx`

foochifa (commented):
A bit premature (as its still a draft), but just wanted to highlight the date stuffs

frontend/src/utils/date.ts
LinHuiqing: Ugh. Daylight savings.

LinHuiqing: Hm I guess that's why `fromUtcToLocalDate` was done that way.

LinHuiqing: fixed in f0dbefb0

frontend/src/utils/fieldValidation.ts
LinHuiqing: done in ed2b31fd

LinHuiqing:
Will be adding the tests in a bit! Will try to be as extensive as possible so will take a while. But the videos kinda show what I think should be in the manual tests!

LinHuiqing:
Okays tests have been updated! 👍

src/app/utils/field-validation/validators/dateValidator.ts
```diff
@@ -42,10 +42,8 @@ const dateFormatValidator: DateValidator = (response) => {
  */
 const pastOnlyValidator: DateValidator = (response) => {
   // Today takes two possible values - a min (in makeFutureOnlyValidator) and max (here)
-  // Add 14 hours here to account for up to UTC + 14 timezone
-  // This allows validation to pass as long as user is on the correct date (locally)
-  // Even if they are in a different timezone
-  const todayMax = moment().utc().add(14, 'hours').startOf('day')
+  // Dates are converted to use local timezones when loaded by the DateField so no conversion required
+  const todayMax = moment()
```
tshuli: Will this affect server-side validation? E.g. singapore server is on date D, and client is on date D+1

src/app/utils/field-validation/validators/dateValidator.ts
```diff
@@ -59,10 +57,8 @@ const pastOnlyValidator: DateValidator = (response) => {
  */
 const futureOnlyValidator: DateValidator = (response) => {
   // Today takes two possible values - a min (here) and max (in makePastOnlyValidator)
-  // Subtract 12 hours here to account for up to UTC - 12 timezone
-  // This allows validation to pass as long as user is on the correct date (locally)
-  // Even if they are in a different timezone
-  const todayMin = moment().utc().subtract(12, 'hours').startOf('day')
+  // Dates are converted to use local timezones when loaded by the DateField so no conversion required
+  const todayMin = moment()
```
tshuli: Same here for server side validation

foochifa:
Might be related to Shuli's comments 
image

To replicate -> Create a date field -> set disallow past/future -> save
Go back into field -> unset disallow past/future -> validation failure

src/app/utils/field-validation/validators/dateValidator.ts
LinHuiqing: actually yea I was a bit confused as to where the validator is actually used haha lemme relook at this

LinHuiqing: ah okay yea I think there will be an issue 😓 thanks for pointing this out!

LinHuiqing:
> Might be related to Shuli's comments 
> \<image>
> To replicate -> Create a date field -> set disallow past/future -> save Go back into field -> unset disallow past/future -> validation failure

@foochifa should be fixed in [b30c58c](https://github.com/opengovsg/FormSG/pull/6096/commits/b30c58ca37dc12c545596c475055ec3de8b9e5a9)

src/app/utils/field-validation/validators/dateValidator.ts
LinHuiqing: should be fixed in [47cbc25](https://github.com/opengovsg/FormSG/pull/6096/commits/47cbc25f83d6d8f198ee104e9ca1651ae5721452)

src/app/utils/field-validation/validators/dateValidator.ts
LinHuiqing: should be fixed in [47cbc25](https://github.com/opengovsg/FormSG/pull/6096/commits/47cbc25f83d6d8f198ee104e9ca1651ae5721452)

foochifa:
image

~Yikes 😭~

~Is this caused by server side validation? My timezone is set in the US, hence 17/4/23 is a valid option.~

Sorry mb! not a prob

LinHuiqing:
> \image
> 
> Yikes 😭
> 
> Is this caused by server side validation? My timezone is set in the US, hence 17/4/23 is a valid option.

I'm assuming you selected disallow past dates, then selected 17/4/23? I can't seem to replicate it though?

foochifa (approved):
Lgtm Thanks HQ!!

PR #6097, opened 2023-04-13
feat: prevent prefill editing

### LINKED ISSUE #6093: Allow pre-fills to be lockable
**What are pre-fills?**
Refer to the guide here: https://guide.form.gov.sg/user-guides/advanced-guide/pre-fill
3 years ago, we started having prefills. This was to help respondents save time on filling the form and for admins to include certain information that they already knew about respondents.
e.g tagging clinics that respondents visited, but not wanting to create individual forms for each clinic. 

**Why is it not widely use?**
One of the current downsides of pre-fills is that there is no lock option which means that users can still edit the field. Admins have to use workarounds like putting in the description "please do not edit this field". Not all admins are convinced that this works and would be additional operational load for them to check if their respondents did or did not amend.

**Why should we introduce this??**
We had a couple of requests here and there for pre-fills to be lockable but did not have quantifiable impact. However, recently an agency wrote in sharing cost savings and open up a variety of use cases for having this simple feature (e.g. "agency verified data" without need of MyInfo like customer ID and more). They would also be propagating this workflow to other agencies as part of their research with a Ministry.

**Possible downsides?**
Previously, a consideration for not making it locked was to prevent scams but since we already highlighted the field in yellow and mentioned to respondents that some of these are pre-filled, we could give admins the choice.

tshuli:
## Problem
- Closes #6093 

## Solution

**Breaking Changes** 

    - Details ...

**Features**:

- Adds new lock feature for prefill.

## Screenshots

_Admin View_
Screenshot 2023-05-02 at 8 18 56 PM

_Public View_
Screenshot 2023-05-02 at 8 18 16 PM

## Tests
- [ ] Can create form with prefill and no prefill lock
- [ ] (via network call) attempting to create a field with no prefill but with prefill lock should return an error
- [ ] Non-prefill forms render and submit correctly
- [ ] Forms with only unlocked prefill can render and submit correctly
- [ ] Forms with only locked prefill can render and submit correctly
- [ ] Forms with combination of locked and unlocked prefill can render and submit correctly
  - [ ]  Works for singpass form with combination of locked and unlocked prefill


tshuli:
@staceytan1998 - I did a first pass, could I trouble you to review the design and propose changes if needed
- In particular, whether locked prefills should still be highlighted in yellow; this might confuse users if some yellow fields are editable and others are not
- I thought we could do it like myinfo fields, which are simply greyed-out / disabled

For merging after angular is deprecated

tshuli:
Other considerations:
- This prevents accidental modification, but user could still modify the url deliberately. Maybe instead of 'lock prefill' which implies guarantee to admins that the value won't be changed, we can call it 'prevent field editing' and in the description indicate 'Prevent users from accidentally changing the value of the field'

timotheeg:
+1 to the comments from ShuLi:
* Until we have true locking of the values (requires some form of server integration), labelling the feature "prevent field editing" sounds good to me
* Greying out like MyInfo field feels appropriate as well

LinHuiqing:
from Kenneth: let's add metrics for prefills before we push this too

staceytan1998:
Checked with Shanty on the pre-fill copy and finalised it, here is the design! Things to take note:

- [Description](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?node-id=3316-976&t=OtY9xVKP1QSaJi9m-4) for enable pre-fill toggle has been tweaked
- Addition of ['Prevent editing' setting](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?node-id=3316-1246&t=OtY9xVKP1QSaJi9m-4)
- [Copy for the infobox](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?node-id=3316-1687&t=OtY9xVKP1QSaJi9m-4) shown on the form has also been tweaked!

timotheeg:
What's the status on this prefill lock PR? @tshuli do you want to pass it to somebody else?

tshuli:
I'll see it through since it's fairly straightforward! 

Can we start merging in #afterreact PRs already? 😃

tshuli:
@staceytan1998 I've updated the designs for your review!

tshuli:
@staceytan1998 addressed your comments in chromatic, for your re-review!

frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -113,14 +123,19 @@ export const FormFields = ({
...
+                !isEmpty(fieldPrefillMap) &&
+                  Object.values(fieldPrefillMap).some(
+                    (field) => !field.lockPrefill && field.prefillValue,
+                  ) && (
+                    <InlineMessage variant="warning">
+                      Highlighted fields below have been pre-filled according to
+                      the form link you clicked. You may edit these fields if
+                      necessary, except non-editable fields with a lock icon.
+                    </InlineMessage>
+                  )
```
timotheeg: /nit: we could have more descriptive messages here to be specific in various combination of locked/not-locked pre-filled fields and their values.

timotheeg (approved):
PR looks good! Thanks @tshuli 

I am still worried about soft locks having some inherent danger.

Also additionally, let's make sure we track product metrics for this feature so monitor adoption (number of forms with locked prefills) to add in the [extractor function](https://github.com/opengovsg/formsg-private/tree/develop/functions/get_product_metrics_from_mongodb) probably

frontend/src/features/public-form/components/FormFields/FormFields.tsx
tshuli: thanks @timotheeg! having customised message is a good idea. @staceytan1998 please review the copy for the 3 cases: locked and normal prefill, normal prefill only, locked prefill only

**Normal and Locked prefill**
Screenshot 2023-05-14 at 4 08 30 PM

**Normal prefill only**
Screenshot 2023-05-14 at 4 09 04 PM

**Locked prefill only**
Screenshot 2023-05-14 at 4 09 41 PM

tshuli:
On the inherent risks with soft locks, we discussed the possibility of beta flag to better manage usage, but decided to go ahead with it. to track usage and consider a harder-lock system (e.g. requiring admins to pre-register the links).

frontend/src/features/public-form/components/FormFields/FormFields.tsx
staceytan1998: For locked prefill only, we can omit the words 'with a lock icon'! So it will be 'Highlighted fields below have been pre-filled according to the form link you clicked. These are non-editable fields.' 

Thanks Shu Li!

tshuli: edited the copy for locked prefill only

PR #6098, opened 2023-04-14
chore: prune error logs for ndi oidc key retrieval error

tshuli:
## Problem
- We use `promise-retry` to refresh ahead for our cached ndi jwks key
- If the promise fails, `promise-retry` throws the entire retryable promise, and not just the error
- This PR adjusts the logger so that only error information is logged on exception


timotheeg (approved):
LGTM 👍

PR #6099, opened 2023-04-14
fix: payment email missing in BE

KenLSM:
## Problem

BE strictly requires payment email field. However, FE wasn't able to reference the correct schema ID, resulting in the error of `Error when creating payment: payment receipt email not provided` on the BE as the field, referenced by BE, `paymentReceiptEmail` is now empty.

**Breaking Changes** 

**Bug Fixes**:

- fix to reference `const PAYMENT_CONTACT_FIELD_ID` when extracting the field in `PublicFormProvider` on FE

**Tests**
- [ ] check if FE is passing the correct field to be BE
  - [ ] able to reach payment page successfully


LinHuiqing:
Omy ps my bad on this one 😅

justynoh (approved):
lgtm

PR #6102, opened 2023-04-17
feat: payment manual sync

### LINKED ISSUE #5968: [REVIEW] Build cron system to track and move completed submissions for which we did not receive webhook or redirect enquiries
Completed payment transition pending submission to completed submissions. 

The transition is driven update to the payment intent:
- webhook
- client request when frontend redirects

For resiliency, we assume that these mechanism can both fail, and we need a catch all mechanism to perform the transition ourselves. 

Cron job:
- Find all pending payments
- for each pending payment, query stripe for payment intent details (beware of stripe rate limit)
- if payment intent is paid, update DB and transition pending submission to completed submission
- report run metrics for observability, number of payment found + payment intent status, plus check outcome
- Deferred to #6211 if payment is not paid and is older than the expiry, cancel payment, and cleanup (i.e. soft delete) 

Cron frequency:
30mins

Cron platform [TBD](https://betterprogramming.pub/cron-job-patterns-in-aws-126fbf54a276) Add Pros/Cons below and decide
- option 1: lambda with cloudwatch events
    - Pros: Easy to set up
    - max run time of 15mins (probably OK initially, as long as we have detailed metrics and we get alerted when it is not enough)

- option 2: AWS Batch

Additional tasks:
- set up monitoring dash board (runtime of cron, volume of payment processed this way, etc)
- Add alerts for unexpected outcome of these metrics
- harden infrastructure and permission for the cron job

This cron needs thoughts as it is the first app-level elements we have that will not be working from the formSG monolith app.

KenLSM:
## Problem

Form BE Payments sync with Stripe is done through webhook. Webhook can fail, although stripe feels that it is robust enough most of the time, but it is not enough for us.

We want to have additional safeguards in the situation that webhooks are delayed, or not captured by our system. We need to ensure that the payment information across Stripe and FormSG is consistent.

### Additional Requirements

Flow should support:
- future manual admin triggers

Closes: #5968, FRM-421, FRM-361

## Solution

**Reconciliation:** is a process that sweeps through the DB and checks if there are payments that have yet to reach its end-state.

1. Find all pending payments
2. For each payment account query stripe for payment intent details
3. Sends event into `StripeService.handleStripeEvent`
4. Double-checks all payment statuses with stripe, and cancels stale payments.
5. returns report run metrics. Summaries are sent to slack channel
 -  number of payment events found
 -  processed events: latest payment intent status
 -  failed events: event details

CRON job is run every 30 min.

See the comment below for further explanation on the CRON job

**Breaking Changes** 

**Features**:

- Added two new APIs on BE
  - `GET /payments/reconcile/incompletePayments`
  - `POST /payments/reconcile/account/:stripeAccount`

- Added new middleware  `withCronPaymentSecretAuthentication`
  - routes guarded by the above middleware are referred to as `ProtectedRoutes` that only those with a shared secret can call

## Tests
To test that failed events are resent:
- [ ] Fill a payment form and proceed to pay (preferably on your phone).
    - [ ] Then, revert staging to v6.41.0 (this is the release just before payments, without webhook handler). **While this deploys, keep your payment tab active!**
    - [ ] When the deployment is complete, fill the card details on the payment page and click "Submit payment". This should lead you to a blank screen.
    - [ ] Check that on the Stripe dashboard, two events have been sent out `charge.succeeded` and `payment_intent.succeeded`. Each should have been tried twice, with the next retry "in an hour".
    - [ ] Now, restore the latest version onto staging.
    - [ ] Refresh the payment page. This should show "we are waiting for your proof of payment" screen.
    - [ ] Run the lambda, and refresh the payment page. This should show the success screen.
    - [ ] Check the slack reconciliation report, that two events were successfully handled.

To test the payment cancellation flow:
- [ ] Fill a payment form and proceed to pay. DO NOT make payment.
    - [ ] Wait half an hour. Retrigger the lambda.
    - [ ] Go back to the payment page. It should show "Payment request was cancelled".

## Deploy Notes

New lambda scripts
- `form-payment-reconciliation`

Lambda environment configuration
- `SSM_ENV_SITE_NAME`: one of `prod`, `staging`, `staging-alt`, `staging-alt2` or `uat`.

New SSM configs
- `{ENV_NAME}-cron-payment parameter`
  - `CRON_PAYMENT_API_SECRET={secret-key}`
  - `SLACK_API_SECRET={slack-secret-key}`


functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,149 @@
+const { SSMClient, GetParameterCommand } = require('@aws-sdk/client-ssm')
+const { generateLinkToCloudwatch, postToSlack } = require('./src/utils')
+
+const AWS_REGION = 'ap-southeast-1'
+
+const PARAMETER_STORE_NAME = 'staging-cron-payment'
```
timotheeg: Should come from env var?

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,149 @@
...
+  const awsSsmClient = new SSMClient({ region: AWS_REGION })
+  const input = {
+    Name: PARAMETER_STORE_NAME,
+    WithDecryption: true,
+  }
+  const command = new GetParameterCommand(input)
+  return awsSsmClient
+    .send(command)
+    .then((d) => d.Parameter.Value)
+    .then((d) => {
```
timotheeg: Why 2 promise steps? Is `d.Parameter.Value` a promise?

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,149 @@
...
+      const dn = d.split('\n')
+      return dn.reduce((accum, line) => {
+        const [key, value] = line.split('=')
+        accum[key] = value
+        return accum
+      }, {})
+    })
+}
+
+const getPendingQueries = (apiSecret) => {
```
timotheeg: Is the name on purpose, or did you mean to write `getPendingPayments()` ? 😅

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,149 @@
...
+  return fetch(
+    `https://${SUB_DOMAIN}form.gov.sg/api/v3/payments/pendingPayments`,
+    {
+      headers: {
+        'x-cron-payment-secret': apiSecret,
```
timotheeg: /nit: we have a convention of making custom formsg headers' prefix be `x-formsg-` (see [mail constants](https://github.com/opengovsg/FormSG/blob/develop/src/app/services/mail/mail.constants.ts#L11-L13) and [webhook service](https://github.com/opengovsg/FormSG/blob/develop/src/app/modules/webhook/webhook.service.ts#L151) for reference)

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,149 @@
...
+  ).then((d) => d.json())
+}
+
+const reconcileAccount = (apiSecret, stripeAccountId) => {
+  return fetch(
+    `https://${SUB_DOMAIN}form.gov.sg/api/v3/payments/reconcileAccount`,
+    {
+      method: 'POST',
+      headers: {
+        'x-cron-payment-secret': apiSecret,
```
timotheeg: Same comment on header prefix

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,149 @@
...
+  return awsSsmClient
+    .send(command)
+    .then((d) => d.Parameter.Value)
+    .then((d) => {
+      const dn = d.split('\n')
+      return dn.reduce((accum, line) => {
+        const [key, value] = line.split('=')
+        accum[key] = value
+        return accum
+      }, {})
```
timotheeg: I've had plenty of issues with parsers in the past on the config, even if it was dev controlled. The top 3 offender issues are:
1. accidental empty lines
2. accidental whitespaces around the equal sign, or as prefix/suffix
3. legitimate equal sign(s) in the values of the `key=value` pairs

So I recommend having some simple "standard" protection against those, with something like this this:
```suggestion
      const keyPairRe = /^\s*([^\s=]+)\s*=\s*(.+)$/
      const dn = d.trim().split(/[\r\n]+/)
      return dn.reduce((accum, line) => {
        const matches = line.match(keyPairRe)
        if (matches) {
            const [_, key, value] = matches
            accum[key] = value.trim()
        }
        return accum
      }, {})
```

src/app/modules/payments/payments.service.ts
```diff
@@ -330,3 +337,70 @@ export const findLatestSuccessfulPaymentByEmailAndFormId = (
...
+ * @param createdHrsAgo defaults to 72
+ * @returns
+ */
+
+export const findPendingPaymentsByTime = (createdHrsAgo = 72) => {
+  const logMeta = {
+    action: 'findPaymentByTime',
+    createdHrsAgo,
+  }
+  const positiveHrsAgo = Math.max(createdHrsAgo, 1)
```
timotheeg: Didn't we say we wanted to verify with 30mins interval? the code here restrict to minimum 1 hour.

timotheeg (commented):
I finally got started on review. I'm not done yet, but sending a few minor comments first!

justynoh:
Note: here we'll use "pending" to mean "incomplete". In particular, this means that the payment status could be `Pending` or `Failed`.

**Basic structure**

Some considerations:

1. We want the CRON job to run every half an hour, reconciling payment intents as it goes along. Key requirement is that **every payment** must have its status reconciled. (We only care about the pending state transitions, i.e. pending -> failed, pending -> succeeded and failed -> succeeded.)
2. Since payment intents don't expire, cut-off times for payment creation actually don't do anything other than reduce the likelihood of race conditions. The race conditions are still not eliminated, and as we have experienced previously by Murphy's law, reducing the possibility is not enough. (Concretely, say the CRON ran at time T, and we only reconcile payments created before T-5, assuming most payments are complete within 5 min. No matter when the payment was created, either before T-5 or between T-5 and T, a user could still be trying to pay at time T, thus causing a race condition with status retrieval.) What this tells us is that **having a min-age for payments doesn't affect correctness**.
3. Since events are the primary method which allow payment statuses to be advanced, we can trigger payment movement by ensuring that FormSG receives and correctly processes all Stripe events. By re-sending all failed event deliveries, we are guaranteed* that we are consistent with Stripe.
4. Nonetheless, we still want to have an explicit correctness guarantee, so at the end of the day we still need to check that "`stripe.payment_intent.status === formsg.payment.status`".

*Assuming that our state machine is also consistent with Stripe.

Given this, the strategy is for the CRON job to simulate being a Stripe server and send events to FormSG servers, which will execute the exact same processing flow as the Stripe server. 

Here is the plan of attack. First we devise a solution to ensure that the information in the DB is correct. Assume the CRON job runs once every 30 min, at time T.

1. Retrieve all payments still in pending states. Group them by Stripe accounts. For each Stripe account, do the following:
2. Retrieve from stripe all undelivered events [1]
3. Process all the undelivered events
4. Double-check each payment status with the Stripe payment status [2]

There are several things of note here.

[1] A race condition could occur here, as pointed out earlier. If a payment is being sent at the point when the CRON job is run, part of the undelivered events may include the latest in-flight event. Or, an undelivered event may be in progress of being retried at the point when the CRON job is run. However, regardless of the race condition, this works because the event handler is idempotent. We also check for duplicate events and return 200 OK in the webhook handler, so even if Stripe retries the event later on, or our event is the later one to have arrived, both will return the response with the latest payment.

[2] This is another place that a race condition could occur. Say that during Step 3, a payment was moved to completion. However, the webhook has not arrived yet. Then at Step 4, the status on Stripe would be a success while the status on Form would still be pending. This would result in a discrepancy that could be reconciled in a future CRON job. Nonetheless, at this point, it would still be considered as incorrect and will be flagged for manual checking.

**Cancelling payments**

Retrieving all pending payments will make increasingly less sense, as this is a set that grows monotonically. We need to have a way to reduce the size of this set eventually. We want to cancel pending payments that are expired, after at least 30 min. Therefore, this flow has to be integrated with the payment cancellation flow. 

Since we should only cancel payments that have been reconciled and verified to be still in pending states even on Stripe, we can modify the process to be the following.

4. Double-check each payment status with the Stripe payment status 
    a. If the Stripe payment status is still pending, the FormSG payment status cannot be anything other than pending (otherwise our state machine is definitely incorrect!). If the payment is stale (i.e. 30 min has passed since creation), Cancel the payment. Otherwise, do nothing.
    b. If the Stripe payment status is successful and FormSG payment status is also successful, then reconciliation was successful and no further action needs to be taken
    c. If the Stripe payment status is successful and FormSG payment status is pending, then report an error. A race condition likely occurred, and will be reconciled by a future webhook or the next CRON job.

justynoh:
What needs to be done
- Add feature to transition payments to "cancelled" state on receipt of the `payment_intent.canceled` webhook
- Write feature to double check payment status
- Conditionally cancel expired payments

functions/form-payment-reconciliation/index.js
justynoh: fixed in [95ab1e8](https://github.com/opengovsg/FormSG/pull/6102/commits/95ab1e89b5031f60b29d3019a75651cfb24f589d)

functions/form-payment-reconciliation/index.js
justynoh: fixed in [95ab1e8](https://github.com/opengovsg/FormSG/pull/6102/commits/95ab1e89b5031f60b29d3019a75651cfb24f589d)

functions/form-payment-reconciliation/index.js
justynoh: [95ab1e8](https://github.com/opengovsg/FormSG/pull/6102/commits/95ab1e89b5031f60b29d3019a75651cfb24f589d)

functions/form-payment-reconciliation/index.js
justynoh: also fixed in [95ab1e8](https://github.com/opengovsg/FormSG/pull/6102/commits/95ab1e89b5031f60b29d3019a75651cfb24f589d)

src/app/modules/payments/payments.service.ts
KenLSM: This 1hr no longer valid, since cron will run every 30mins, and try to sync all incomplete payments.

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,291 @@
...
+  const command = new GetParameterCommand(input)
+  return awsSsmClient.send(command).then((res) => {
+    const parametersAsString = res.Parameter.Value
+    const parameters = parametersAsString
+      .split(/[\r\n]+/)
+      .map((line) => line.trim())
+      .filter((line) => !!line) // Filter out empty lines
+    return parameters.reduce((parameterMap, line) => {
+      const match = line.match(KEY_VALUE_PAIR_REGEX)
+      if (match && match.length === 3) parameterMap[match[1]] = match[2]
```
timotheeg, line 62: /nit: The length check for 3 is not needed. If `match` is truthy, it is guaranteed to have length 3 (thank you regexes!)

... Hmm, but is it because typescript doesn't know about it? 😅 

```suggestion
      if (match) parameterMap[match[1]] = match[2]
```

src/app/routes/api/v3/payments/payments.routes.ts
```diff
@@ -79,3 +78,38 @@ PaymentsRouter.route('/:formId([a-fA-F0-9]{24})/payments/previous').post(
...
+/**
+ * Get all payments in incomplete state (Pending or Failed) which need to be
+ * reconciled.
+ * @protected
+ * @route GET /payments/reconcile/incompletePayments
+ *
+ * @returns 200 with found payment records
+ * @returns 500 if there were unexpected errors in retrieving payment data
+ */
+ProtectedPaymentsRouter.route('/incompletePayments').get(
```
timotheeg: /nit I don't think it's actually needed to have a structure of
```
router.route(pathPattern).get(middleware, ...)
```

IIRC express accepts a simpler:
```
router.get(pathPattern, middleware, ...)
```

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,291 @@
...
+const getApi = (apiSecret) => {
+  const getIncompletePayments = async () => {
+    return fetch(`${CRON_API_PREFIX}incompletePayments`, {
+      headers: { [API_AUTH_HEADER]: apiSecret },
+    }).then(async (res) => ({ ok: res.ok, data: await res.json() }))
+  }
+
+  const reconcileAccount = async (stripeAccount, paymentIds) => {
+    return fetch(
+      `${CRON_API_PREFIX}account/${stripeAccount}?maxAgeHrs=${MAX_AGE_HRS_EVENTS}`,
```
timotheeg: /superNit: my personal practice for URI construction is to **NOT** have URI prefixes and URI namespaces include the trailing slash, so all paths computed with them for concatenation start with a `/`, and are "real" paths
e.g.
```javascript
API_PREFIX = 'https://domain.com/api/v1'
path = '/entity/property'
url = `${API_PREFIX}${path}`
```

It's a very minor thing, but applying it consistently avoids having to think about it, and occasionally making the wrong assumption, which can lead to 1) missing slashes:
```javascript
API_PREFIX = 'https://domain.com/api/v1'
path = 'entity/property'
url = `${API_PREFIX}${path}`
```

or 2) having double slashes
```javascript
API_PREFIX = 'https://domain.com/api/v1/'
path = '/entity/property'
url = `${API_PREFIX}${path}`
```
(some web servers ignore double slashes, but it's still wrong! 😅 )

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,291 @@
...
+  ENV_SITE_NAME === 'prod' ? '' : `${ENV_SITE_NAME}.`
+}form.gov.sg/api/v3/payments/reconcile/`
+
+const PARAMETER_STORE_NAME = `${ENV_SITE_NAME}-cron-payment`
+const CRON_PAYMENT_API_SECRET_KEY = 'CRON_PAYMENT_API_SECRET'
+const CRON_PAYMENT_SLACK_SECRET_KEY = 'SLACK_API_SECRET'
+
+const API_AUTH_HEADER = 'x-formsg-cron-payment-secret'
+
+const MAX_AGE_HRS_EVENTS = 7
```
timotheeg: Any rationale we could add to the comment on this 7 hours limit? Does it relate to any settings in Stripe?

functions/form-payment-reconciliation/index.js
```diff
@@ -0,0 +1,291 @@
...
+    '---',
+    '*Critical messages:*',
+  ]
+
+  console.log('CRON job starting')
+  const CRON_START_TIME = Date.now()
+
+  // Step 0(a): Load secrets from SSM
+  const secret = await getSSMSecrets()
+
```
timotheeg, line 108: Just checking: if the config doesn't match our expectations (in this case, the keys for `CRON_PAYMENT_API_SECRET_KEY` and `CRON_PAYMENT_SLACK_SECRET_KEY` should be present), does the code below that use the secrets fails "visibly", and we will know about it?

src/app/modules/auth/auth.utils.ts
```diff
@@ -63,3 +65,7 @@ export const getUserIdFromSession = (
 ): string | undefined => {
   return session?.user?._id as string | undefined
 }
+
+export const isCronPaymentAuthValid = (header: IncomingHttpHeaders) => {
+  return header['x-formsg-cron-payment-secret'] === cronPaymentConfig.apiSecret
```
timotheeg, line 70: /toDiscuss (potentiallyNit/canProbablyDefer): since the API is public, and just protected by secret auth, we might want to store the shared secret hashed in the config, and do a hash comparison here. So if a malicious user ever gets access to our config key, they won't be able exit silently, while still be able to invoke the APIs.

src/app/modules/payments/payments.service.ts
```diff
@@ -330,3 +330,24 @@ export const findLatestSuccessfulPaymentByEmailAndFormId = (
...
+/**
+ * Retrieves all payments that are in Pending or Failed statuses.
+ * @returns a list of payments that are incomplete.
+ */
+export const getIncompletePayments = (): ResultAsync<
+  IPaymentSchema[],
+  DatabaseError
+> => {
+  return ResultAsync.fromPromise(
+    PaymentModel.getByStatus(PaymentStatus.Pending, PaymentStatus.Failed),
```
timotheeg, line 343: What's the rationale for getting failed payments again? (apologies, this is probably a RTFM on me 😰 )

timotheeg (approved):
LGTM

functions/form-payment-reconciliation/index.js
justynoh: this one is purely my preference to have an explicit guarantee that the array index won't explode rather than relying on the behavior of `.match` 😅

functions/form-payment-reconciliation/index.js
justynoh: we will know about the shared secret since we will receive a whole bunch of failed requests in our reconciliation report. For missing slack secret key, there's no way to really notify us unless we set up an entirely different flow e.g. by email or something (otherwise if the lambda has no access to slack, it can't notify us via slack!). the only way is to notice that the lambda didnt run at xx:(0|3)0.

src/app/modules/payments/payments.service.ts
justynoh: failed payments can be retried!

functions/form-payment-reconciliation/index.js
justynoh: fixed!

src/app/routes/api/v3/payments/payments.routes.ts
justynoh: good point, fixed

functions/form-payment-reconciliation/index.js
justynoh: I extended it to 24 hours. 

The logic here is that if an event's re-processing failed, we should look into it. If we didn't look into it, it should continue to be retried and reported as failed in future CRON jobs. In the case that we looked into it but deemed it not to be a relevant event, we should let it exit the system at some point, so that we don't keep it around. I think a reasonable timeframe to assume we will get to looking at failed events is 24h (e.g. if this occurs while we are asleep, we need at least 8 hours buffer, assuming we all sleep at least 8 hours a night 🤡 ).

functions/form-payment-reconciliation/index.js
timotheeg: Ah I see. But hmm, you have to be careful when preference goes in the way of readability (somewhat) 🤔 : a sanity/defensive check implies something could happen, and a careful reader must expand some mental load to understand what. But in this case, the behaviour of the `.match()` is guaranteed.

BTW, if you find it clearer, you could use named groups:

```javascript
const KEY_VALUE_PAIR_REGEX = /^(?<key>[^\s=]+)\s*=\s*(?<value>\S+)$/

const match = line.match(KEY_VALUE_PAIR_REGEX)

if (match) parameterMap[match.groups.key] = match.groups.value
```

Where similarly, `match` being truthy guarantees that all defined groups have a value (yay regexes!)

PR #6106, opened 2023-04-18
chore: make hsts header follow recommendation from csa

timotheeg:
## Problem
We want to increase our domain score on CSA: https://ihp.csa.gov.sg/home

form.gov.sg is failing the HSTS check because our header `Strict-Transport-Security` has a max-age of 60 days instead of the recommendation of >= 1 year.

## Solution
Extend `max-age` of `Strict-Transport-Security` header to 400 days, to be greater than a year.

**Breaking Changes** 
    - Details ...

### DIFF
--- a/src/app/loaders/express/__tests__/helmet.spec.ts
+++ b/src/app/loaders/express/__tests__/helmet.spec.ts
@@ -122,7 +122,7 @@ describe('helmetMiddlewares', () => {
     if (hstsFn) {
       hstsFn(mockReq, mockRes, mockNext)
     }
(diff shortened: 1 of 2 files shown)


tshuli (approved):
lgtm, can I just confirm that `include subdomains` is currently also activated?

timotheeg:
> can I just confirm that include subdomains is currently also activated?

Yes it is, this was the value of the header in the CSA report:

```
Strict-Transport-Security: max-age=5184000; includeSubDomains
```

timotheeg:
I've updated HSTS in Cloudflare itself as well directly, so the scanner is already happy about that setting, but I'll still merge this to show in code that we want > 1year max-age.

PR #6118, opened 2023-04-18
chore: local dev stripe webhook listener

foochifa:
## Problem

Currently we have to manually install stripe-cli and set up a listener after authenticating, in order to forward stripe webhooks to test payment flow.

## Solution

This PR creates a docker container using [stripe-cli ](https://stripe.com/docs/cli/docker)image.

Which listens to stripe events and forwards it to our BE.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

On local env
- [ ] Update docker-compose with stripe keys and webhook secret
- [ ] Make a test payment
- [ ] Webhook should respond with success and an invoice should be created and downloadable
- [ ] Check responses, your payment details / submission should be viewable

## Deploy Notes

**New environment variables for local docker-compose**:

- `STRIPE_API_KEY` : Same key as PAYMENT_STRIPE_SECRET_KEY
- Also remember to update webhook secret in backend (seems like the webhook secret is persistent when using `--api-key` instead of `stripe login`)


foochifa:
closed in favour of #6147

PR #6129, opened 2023-04-19
fix: invoice download 502

KenLSM:
## Problem

Upon clicking invoice download, users will be brought to see a page with `bad gateway`.

This was due to the server throwing an error
As the users are directed to an API endpoint, they will see 502.

## Solution

1. On TS, I've added the an explicit type to ensure that it is mapped correctly (not sure why `IPopulatedEncryptedForm` will cause `admin` to be untyped.

2. With the TS now able to rightly determine that `agency.businessInfo` can be undefined, the rest of the references are null checked with the existence of `businessInfo`

**Breaking Changes** 

## Tests

- [] test that invoice can be downloaded (with your work email, ensure that `agency.businessInfo` is updated on db)

### DIFF
--- a/src/app/modules/payments/stripe.controller.ts
+++ b/src/app/modules/payments/stripe.controller.ts
@@ -349,7 +349,11 @@ export const downloadPaymentInvoice: ControllerHandler<{
               .business
 
             // we will still continute the invoice generation even if there's no address/gstregno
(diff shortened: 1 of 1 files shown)


justynoh (approved):
lgtm

PR #6130, opened 2023-04-19
feat: disable description and amount fields for admin when the payment field is disabled

justynoh:
## Problem
We want to disable the `description` and `amount_cents` field in the payment drawer when the payment field is disabled.

## Solution
Currently, our middlewares block empty `description`s and zero `amount_cents`. This validation has to be relaxed, so that it is only triggered when `enabled` is `true`. Also, remove model-layer validation on `amount_cents` and make it a controller validation instead.

**Breaking Changes** 


foochifa (approved):
Lgtm thanks justyn!

Tangential but since we are doing drawer do you think we should update the title of 'Name' to 'Product/service Name' which is the updated typing in figma? And also as the placeholder?

justynoh:
ok, done!

PR #6131, opened 2023-04-19
fix: update payment received subtitle copy

wanlingt:
## Solution

We want to update the copy for the message block that appears when a payment has been received 

## Before & After Screenshots

**AFTER**:
image

### DIFF
--- a/frontend/src/features/public-form/components/FormPaymentPage/stripe/StripeReceiptContainer.tsx
+++ b/frontend/src/features/public-form/components/FormPaymentPage/stripe/StripeReceiptContainer.tsx
@@ -56,7 +56,7 @@ export const StripeReceiptContainer = ({
       <PaymentStack>
         <GenericMessageBlock
           title="Your payment has been received."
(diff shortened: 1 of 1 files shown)


justynoh:
Just checking: are we using [this design](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?node-id=16028-158645&t=TlE06Oy2cBUhKwGi-4)?

wanlingt:
> Just checking: are we using [this design](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?node-id=16028-158645&t=TlE06Oy2cBUhKwGi-4)?

nope, using the updated copy provided by @amitogp in the [issues spreadsheet](https://docs.google.com/spreadsheets/d/1sjYN_Fxvc81aaS1NtH68W2HzU8Y-VG_buplCJUnwkMI/edit#gid=0). we don't want to reference Stripe. ALTHOUGH on second thought I think we should leave the second sentence in there. Will add it back in!
cc @staceytan1998

justynoh (approved):
nice, lgtm!

PR #6132, opened 2023-04-19
fix: convert from seconds to miliseconds from epoch when saving Stripe dates

justynoh:
## Problem
Dates are not appearing correctly on the individual response page because they are not being saved correctly in the DB in the first place. This was because Stripe is sending timestamps in s from epoch, and we did not convert them to ms from epoch.

## Solution
Convert Stripe dates to ms from epoch by `* 1000` before using the `Date` constructor on them.

**Breaking Changes** 

## Screenshots

image


wanlingt (approved):
lgtm!

PR #6137, opened 2023-04-19
feat(admin): add invoice url on response page

### LINKED ISSUE #6094: Replace receipt url in payment response info and CSV with invoice URL
Image

As we only provided simplified tax invoice as proof of payment, we should only invoice URLs rather than Receipt's.

KenLSM:
## Problem

On the responder's side, we are showing invoice download, but for admins, we are still showing receipt downloads.

Closes #6094

## Solution

Change receipt download into our invoice download API

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

**Individual Responses Page**
Screenshot 2023-04-19 at 9 30 33 PM

**CSV content**
Screenshot 2023-04-19 at 9 30 53 PM

**AFTER**:

**Individual Responses Page**
Screenshot 2023-04-19 at 8 57 23 PM

**CSV content**
Screenshot 2023-04-19 at 9 20 57 PM

## Tests

- [ ] new feature tests
  - [ ] payments with completed payments should see URL and can download a well-formed invoice
  - [ ] download csv should contain a well-formed invoice url  on payment forms
- [ ] regressive tests
  - [ ] download csv should still work on non-payment forms

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -47,8 +47,9 @@ const LoadingDecryption = memo(() => {
 })
 
 export const IndividualResponsePage = (): JSX.Element => {
-  const { submissionId } = useParams()
+  const { submissionId, formId } = useParams()
   if (!submissionId) throw new Error('Missing submissionId')
+  if (!formId) throw new Error('Missing formId')
 
   const { secretKey } = useStorageResponsesContext()
   const { data, isLoading, isError } = useIndividualSubmission()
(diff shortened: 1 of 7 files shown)


wanlingt (approved):
code lgtm! your Solution in the PR description is incomplete btw.

Also, I might've missed the discussion but is invoice an opt-in thing for agencies? I understand that JTC requires invoices, but what are the future plans re receipt vs invoice? I feel like we should standardise the naming and use either receipt or invoice, or both separately, as it can cause confusion to both us and users (e.g. public users see "Save payment invoice" while admins see "Receipt: Download as PDF"

KenLSM:
> Also, I might've missed the discussion but is invoice an opt-in thing for agencies? I understand that JTC requires invoices, but what are the future plans re receipt vs invoice? I feel like we should standardise the naming and use either receipt or invoice, or both separately, as it can cause confusion to both us and users (e.g. public users see "Save payment invoice" while admins see "Receipt: Download as PDF"

Yeah, I think on the product side, we're not sure if we really want invoice or not. Currently, we want to make sure that the first agency are happy to onboard. So the decisions for receipt/invoice might shift again.

The first agency has been pushing us to list it as invoice, but as Forms we want to not denote it as such. However, as not many agencies have started the onboarding process, we do not have much facts to rebut them when invoice is insisted.

PR #6157, opened 2023-04-20
fix: lift mongoose sessions for copying submissions to Stripe webhook handler level

### LINKED ISSUE #6022: Solve webhook race condition issue
See #5937 for more details [here](https://github.com/opengovsg/FormSG/pull/5937#issuecomment-1489626352)

justynoh:
## Problem
As identified in issue #6022, we wanted to see how often this happens before we fix it. Well, the time has come as we know it happens reasonably often that we need to fix it.

Closes #6022

## Solution
We issue a lock on payments throughout the entire processing of a single event, so that events are always only processed one at a time.

**Breaking Changes** 


timotheeg (approved):
LGTM

PR #6158, opened 2023-04-20
feat: add formId to payment document

foochifa:
## Problem

There will be cases where we need to lookup formId and other payment fields simultaneously.

For instance, when checking if a respondents has made a duplicate payments on a form. 

Previously we will have to find the payment document, then the submission from the submission Id, then compare the formId from submission with the req formId.

## Solution

Add formId to payment schema and update it when payment document is created upon proceed to Pay

**Breaking Changes** 

    - adds a new property to the payment schema
   
## Tests

- [ ] Fill in a payment form, select proceed to pay
- [ ] Go to DB and ensure the payment document has formId in it and that its the correct form
- [ ] Complete the payment (make sure webhook listener is on)
- [ ] Go to DB and check that payment document has been successfully updated with completedPayment. And that formId still exists

### DIFF
--- a/shared/types/payment.ts
+++ b/shared/types/payment.ts
@@ -32,6 +32,7 @@ export type PayoutMeta = {
 export type Payment = {
   // Pre-payment metadata
   pendingSubmissionId: string
+  formId: string
   email: string
   amount: number
   paymentIntentId: string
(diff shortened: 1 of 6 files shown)


LinHuiqing:
oh shucks sorry @foochifa i fake news-ed i guess you'll need to update some backend tests for this schema change 😅 I can update the tests since its my bad?

timotheeg (approved):
LGTM

PR #6163, opened 2023-04-20
fix: normalize email field values to lowercase

### LINKED ISSUE #6159: Normalise email from payment field
Currently we do not normalise the email field -> lower case domain name etc, before passing it to DB as part of payment document

This will result in inconsistencies/missed results when we try to query the payment document by the user's email

LinHuiqing:
## Problem

Email collected through payment contact email field is taken as if after verification, meaning that the emails tagged to the payments will be taken as is. The issue with this is that there might be emails with a variety of cases when they point to the same email, since modern email servers will assume case-insensitive emails.

Closes #6159

## Solution

As emails should be case-insensitive according to [MailChimp](https://mailchimp.com/resources/are-email-addresses-case-sensitive), this PR converts inputs to lowercase _as the respondent types in the email field_. This ensures that:

1. Any post-submission analysis regarding email fields will be less error-prone due to variations in casing.
2. What respondents see in the email field will be exactly what admins will receive in submission.

**Breaking Changes** 

    - Any existing emails collected through email fields will still be in a variety of cases (e.g. hUIqInG@open.goV.sg).
    - Impact of this should be minimal as payments (where the email field is used) has only used within OGP, and responses for encrypted forms are locked from us (up to admins to manipulate these).

## Tests

- [ ] On an email field, type any email with a variety in terms of upper-/lower-case. Regardless of whether the entered input is upper or lower case, the output on the field should be lower-case.
- [ ] On a payment contact field, type any email with a variety in terms of upper-/lower-case. Regardless of whether the entered input is upper or lower case, the output on the field should be lower-case.
- [ ] Submit a form with at least 1 email field and the payment contact email field. After submission, the emails should be saved correctly in the DB.

### DIFF
--- a/frontend/src/templates/Field/Email/EmailFieldInput.tsx
+++ b/frontend/src/templates/Field/Email/EmailFieldInput.tsx
@@ -45,7 +45,7 @@ export const EmailFieldInput = ({
           autoComplete="email"
           value={value?.value ?? ''}
           onChange={(event) => {
-            const value = event.target.value.trim()
+            const value = event.target.value.trim().toLowerCase()
             return handleInputChange
               ? handleInputChange(onChange)(value)
               : onChange({ value })
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -371,7 +371,7 @@ const submitEncryptModeForm: ControllerHandler<
       })
     }
 
-    const paymentReceiptEmail = req.body.paymentReceiptEmail
+    const paymentReceiptEmail = req.body.paymentReceiptEmail?.toLowerCase()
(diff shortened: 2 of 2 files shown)


timotheeg (commented):
Lowercase normalization should be performed on the backend as well for stronger guarantee.

LinHuiqing:
> Lowercase normalization should be performed on the backend as well for stronger guarantee.

@timotheeg done in [9f1a2b5](https://github.com/opengovsg/FormSG/pull/6163/commits/9f1a2b5ae47dc7276701b91f6444e153c9a95bf3). Additionally, added lowercase validation to normal email fields in [724f0d2](https://github.com/opengovsg/FormSG/pull/6163/commits/724f0d280418de0cbf7d43a3c97b5077682d7920).

timotheeg (approved):
LGTM!

PR #6170, opened 2023-04-21
feat: email check for previous payment 

### LINKED ISSUE #5973: Check if the person has just made a payment based on their email id.

foochifa:
## Problem

Users might face crashes or forgot that they have already made a payment. As we collect their email information, we can use that as an identifier for the users. Hence, if they attempt to make a payment on the same form after successfully making a payment - we will provide them with a warning and a link back to their previous payment.

Closes #5973

## Solution

Find latest successful payment (Payment Status === succeeded) based on email and formId
Endpoint :formId/payments/previous/:email will return 200 when a payment document is found and 404 when it is not

On Frontend, we will check an isDuplicate state based on whether the endpoint has returned a payment document or not. Accordingly it will choose to display PaymentModal or the new DuplicatePaymentModal with the link to the previous latest payment

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:

image

## Tests

**To check basic functionality**
Go to a payment form
- [x] make a payment successfully
- [x] attempt to make a second payment with the same email
- [x] you should see the modal directing you to your previous payment page
- [x] click the link and ensure you are the same payment page from when you made a successful payment

**To check that we get the latest successful payment**
all of these are done with the same email as above
- [x] go back and make a second payment with the same email again (take note of your payment id/payment page link)
- [x] make a payment successfully
- [x] go back to the form
- [x] make a payment intent (but do not pay)
- [x] go back to the form
- [x] fill in the form and click proceed to pay, the duplicate payment modal should show, directing you to the **second** payment that you made. Check that the payment id/payment page link is equivalent to your second payment

**To check that we do not erroneously show the duplicate payment modal**
On the same form
- [x] make a payment with a different email
- [x] you should not see the duplicate payment modal at proceed to pay

On a different form
- [x] make a payment with the original email form above
- [x] you should not see the duplicate payment modal at proceed to pay

**To ensure no regression issues on non-payment forms**
(As changes was made to PublicFormSubmitButton)
- [x] go to a non-payment storage mode form
- [x] make a submission
- [x] go to an email mode form
- [x] make a submission

### DIFF
--- a/frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
+++ b/frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
@@ -1,25 +1,28 @@
-import { MouseEventHandler, useMemo } from 'react'
+import { MouseEventHandler, useMemo, useState } from 'react'
 import { useFormState, UseFormTrigger, useWatch } from 'react-hook-form'
 import { Stack, useDisclosure, VisuallyHidden } from '@chakra-ui/react'
 
+import { PAYMENT_CONTACT_FIELD_ID } from '~shared/constants'
 import {
   FormField,
   FormResponseMode,
   LogicDto,
   MyInfoFormField,
+  PaymentStatus,
 } from '~shared/types'
 
 import { ThemeColorScheme } from '~theme/foundations/colours'
 import { useIsMobile } from '~hooks/useIsMobile'
 import Button from '~components/Button'
 import InlineMessage from '~components/InlineMessage'
-import { FormFieldValues } from '~templates/Field'
+import { FormFieldValues, VerifiableFieldValues } from '~templates/Field'
 
 import { getLogicUnitPreventingSubmit } from '~features/logic/utils'
 
 import { usePublicFormContext } from '../../PublicFormContext'
 import { DuplicatePaymentModal } from '../DuplicatePaymentModal/DuplicatePaymentModal'
 import { FormPaymentModal } from '../FormPaymentModal/FormPaymentModal'
+import { getPreviousPayment } from '../FormPaymentPage/FormPaymentService'
 
 interface PublicFormSubmitButtonProps {
   formFields: MyInfoFormField<FormField>[]
@@ -40,10 +43,17 @@ export const PublicFormSubmitButton = ({
   onSubmit,
   trigger,
 }: PublicFormSubmitButtonProps): JSX.Element => {
+  const [isDuplicate, setDuplicate] = useState(false)
+  const [paymentId, setPaymentId] = useState('')
+
   const isMobile = useIsMobile()
   const { isSubmitting } = useFormState()
   const formInputs = useWatch<FormFieldValues>({}) as FormFieldValues
   const { form, formId } = usePublicFormContext()
+
+  const paymentEmailField = formInputs[
+    PAYMENT_CONTACT_FIELD_ID
+  ] as VerifiableFieldValues
 
   const preventSubmissionLogic = useMemo(() => {
     return getLogicUnitPreventingSubmit({
@@ -58,14 +68,22 @@ export const PublicFormSubmitButton = ({
 
   const checkBeforeOpen = async () => {
     const result = await trigger()
-    if (result) onOpen()
-  }
 
+    if (result) {
+      // get previous payment
+      const payment = await getPreviousPayment(paymentEmailField.value, formId)
+      // check if duplicate
+      if (payment) {
+        setDuplicate(payment.status === PaymentStatus.Succeeded)
+        setPaymentId(payment._id)
+      }
+      onOpen()
+    }
+  }
+  console.log(paymentId)
   const isPaymentEnabled =
     form?.responseMode === FormResponseMode.Encrypt &&
     form?.payments_field?.enabled
-
-  const isDuplicate = true
 
   return (
     <Stack px={{ base: '1rem', md: 0 }} pt="2.5rem" pb="4rem">
@@ -76,7 +94,7 @@ export const PublicFormSubmitButton = ({
             onClose={onClose}
             isSubmitting={isSubmitting}
             formId={formId}
-            paymentId={'413243243'}
+            paymentId={paymentId}
           />
         ) : (
           <FormPaymentModal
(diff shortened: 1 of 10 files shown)


foochifa:
For discussion: In Frontend, currently we do not use `useQuery` and instead directly call from FormPaymentService

Consideration:
1. The latest payment should always be queried, hence, it should not be cached. This is to prevent any missed warnings.
2. useQuery will cause problems as we want to conditionally retrieve the payment document, only for payment forms (as its currently coupled in `PublicFormSubmitButton`). A potential solution will be to create a new submit button for payment.

timotheeg:
Thanks for the PR @foochifa , let's aim to release that today!

The source issue wasn't very detailed, but I reckon that this PR needs to include an expiry system. Minimally:
* Tell the respondent when the payment we found was ("2 minutes ago", "5 minutes ago", etc.)
And better yet:
* Possibly constraint the search for the previous payment to some max value (e.g. last 24h? last 48h? last 7 days?)

My rationale for asking that:
* The use case we are addressing is respondents making back-to-back payments, or not knowing if their payment went through in case of crashes
* We may not have defined what the expiry threshold should be, but the expiry system should still be present in the code and we can tune the config values based on Amit's input later.

src/app/routes/api/v3/payments/payments.routes.ts
```diff
@@ -64,3 +65,17 @@ PaymentsRouter.route('/stripe/callback').get(
...
+/**
+ * Get previous latest successful payment document if it exists
+ * @route GET /:formId/payments/previous/:email
+ *
+ * @returns 200 with payment if it exists
+ * @returns 200 without payment if it doesnt exists
+ * @returns 500 when database error occurs
+ */
+PaymentsRouter.get(
+  '/:formId([a-fA-F0-9]{24})/payments/previous/:email',
```
timotheeg: ⚠️  putting emails in URL path or query strings is not good practice because it can reveals PII even in http logs.

Can we change this route to be a post handler which takes the email from the body? 

Ooooor, we could manage this frontend hashing, store both the email and email hash in db, create indexes for both these fields, but that feels a little dirtier 🤔

foochifa:
For discussion: 

Currently `:formId/payments/previous/:email` returns 200 even if payment document is not found -> cause I think 'happy path' includes non-previous payments cases as well (which we expect to be the bulk of the payments)

We can potentially return 404s, though this will also mean an even larger number of 404s are to be expected

We also decided not to create two endpoints (i.e. one to check if duplicate + one to retrieve prev payments) instead we have one endpoint to retrieve the prev payment and set duplicate state based on whether payment document exists

Nonetheless, will not be a big change to the logic of the implementation!

src/app/modules/payments/payments.controller.ts
```diff
@@ -0,0 +1,88 @@
...
+          },
+        })
+        return res.send(payment)
+      })
+      // If payment is not found, there is no previous payment
+      // If database error, return 500
+      .mapErr((error) => {
+        // if payment isn't found, return empty response
+        if (error instanceof PaymentNotFoundError) {
+          logger.info({
```
timotheeg, line 46: This could probably be `logger.debug()`, since it's not an error, just a standard expected flow.

frontend/src/features/public-form/components/FormPaymentPage/FormPaymentService.ts
```diff
@@ -26,6 +32,23 @@ export const getPaymentReceiptStatus = async (
...
+ * @param formId the id of the form
+ * @returns payment object is the respondent has made a payment
+ * @returns undefined if the respondent has yet to make payment
+ */
+export const getPreviousPayment = async (
+  email: string,
+  formId: string,
+): Promise<PaymentDto | undefined> => {
+  return ApiService.get<PaymentDto | undefined>(
+    `${PAYMENTS_ENDPOINT}/${formId}/payments/previous/${email}`,
```
timotheeg: To change for PII protection based on my comment below.

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
```diff
@@ -57,7 +68,17 @@ export const PublicFormSubmitButton = ({
 
   const checkBeforeOpen = async () => {
     const result = await trigger()
-    if (result) onOpen()
+
+    if (result) {
+      // get previous payment
+      const payment = await getPreviousPayment(paymentEmailField.value, formId)
```
timotheeg: Just to verify: is `paymentEmailField.value` guaranteed to be normalized? (trimmed + lowercased?)

foochifa: It is guaranteed! On the FE, we will lowercase any inputs into the email field -> so the formInputs email field is guaranteed to be normalized! thanks @LinHuiqing 🙏

src/app/modules/payments/payments.controller.ts
foochifa: I realised our logger has `info` `warn` and `error`, so I kept it as `info` for now!

src/app/routes/api/v3/payments/payments.routes.ts
foochifa: Thanks Tim! [f221588](https://github.com/opengovsg/FormSG/pull/6170/commits/f2215886ca46ea8ed98f5b53c41800163b27f304)

src/app/modules/payments/payments.service.ts
```diff
@@ -254,3 +255,41 @@ export const sendPaymentConfirmationEmailByPaymentId = (
...
+    })
+      .sort({ _id: -1 })
+      .exec(),
+    (error) => {
+      logger.error({
+        message:
+          'Database error while finding latest payment by email and FormId',
+        meta: {
+          action: 'findLatestPaymentByEmailAndFormId',
+          email,
```
timotheeg, line 286: Add `formId` in the meta block (to match our DD facet)

foochifa: Thanks for catching that! [4dc5c60](https://github.com/opengovsg/FormSG/pull/6170/commits/4dc5c601d0a6879130085ad9ee8f68365c6d0ac0)

timotheeg (approved):

PR #6189, opened 2023-04-25
perf: memoize field row containers

### LINKED ISSUE #6121: Fix drag lag in form builder

foochifa:
## Problem

Currently we have a lot of unnecessary re-renders on our Build and Design Content. Many components are re-rendering despite no changes to their properties. This is caused by the large number of hooks we use in the component.

This causes extreme lag in large forms as the re-renders linearly increments the page rendering time for every single action (dragging, selecting, modifying fields in the admin-form view)

This is especially detrimental for dragging fields, as it can cause the admin view for large forms to be almost impossible to navigate.

Closes #6121

## Solution

Move out some of the hooks from `FieldRowContainer` to its parent `BuilderField` which are causing most of the unnecessary re-renders

Use react memo to memoize the `FieldRowContainer` with deep comparison of props.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

Large amount of re-render time as seen in the right.
image

https://user-images.githubusercontent.com/59867455/234180873-e3a49ef3-79a2-4253-bf83-86cb4341680a.mov

**AFTER**:

image

https://user-images.githubusercontent.com/59867455/234180987-39ea66be-7889-4b12-afa4-30c607de6e6c.mov

## Tests

**To ensure the feature works**
Use react dev tool -> profiler
- [x] go to a form in your admin view
- [x] start profiler recording
- [x] shift some of the fields around
- [x] stop recording, you should see that most fields are not re-rendered at each step, and they are memoized at the FieldRow level

Go to a large form
- [x] try to drag a field. The dragging of the field (before dropping) should feel smooth, and there should be minimal jitters.

**To ensure core features are still intact**
- [x] In any storage form, ensure the following functionality is still working
    - [x] Create a new field
    - [x] Edit the field properties of any field
    - [x] Edit the field properties, do not save, click away (close the drawer or select another field) the discard editing modal should appear
    - [x] Drag a field to change its position, field id should be updated too
    - [x] Duplicate a field
    - [x] Attempt to delete a field, the modal should open, and can be cancelled
    - [x] Delete the field, modal should open, and field can be deleted
    - [x] Preview the form. Make a submission
    - [x] Open the form to respondents. Make a submission
- [x] Do the same steps above for an email form
- [x] Do the same steps on mobile 

## Notes
There are still a few unnecessary re-renders upon drag and dropping of the fields.

Ideally, only the fields affected (field id changed) should re-render and all other fields should be unaffected. However, currently there is still some existing hooks that causes a re-render for all the fields even if they are memoized

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
@@ -1,7 +1,16 @@
 import { AdminFormDto } from '~shared/types/form'
 
+import { useCreatePageSidebar } from '~features/admin-form/create/common/CreatePageSidebarContext'
 import { augmentWithQuestionNo } from '~features/form/utils'
 import { FieldIdSet } from '~features/logic/types'
+
+import { useBuilderAndDesignContext } from '../BuilderAndDesignContext'
+import { PENDING_CREATE_FIELD_ID } from '../constants'
+import {
+  FieldBuilderState,
+  stateDataSelector,
+  useFieldBuilderStore,
+} from '../useFieldBuilderStore'
 
 import FieldRow from './FieldRow'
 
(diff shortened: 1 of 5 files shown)


timotheeg (approved):
LGTM

KenLSM (approved):

karrui:
All fields rerender after drag and drop due to recalculation of question number. May be able to reduce impact if we separate the field title from the rest of the field and memo them

foochifa:
Do we have hooks to recalculate the field ids? Probably a bit more investigation needs to be done haha

Cause seems like the hooks are causing the re-renders, as seen below the 6 fields that got their ids updated did get re-rendered correctly. But it seems like some hooks got called that resulted in all the other fields getting re-rendered even though their props didnt change 😅

image

PR #6209, opened 2023-04-27
feat: invoice no background

foochifa:
## Problem

Remove background shadow of invoice based on new flag in `payments_channel`

## Solution

Add `invoice_background` property to `payments_channel`
Remove background in `convertToInvoiceFormat` based on background flag

Currently the only method to remove the background is to go to db and manually set `invoice_background` as `false`

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

To ensure backwards compatibility
- [ ] use an old payment form
- [ ] make a payment
- [ ] the invoice should still look the same

To ensure new forms do not automatically have no background
- [ ] create a new payment form
- [ ] make a payment
- [ ] the invoice should still look the same

To test feature
- [ ] set `invoice_background` to `false` in a form from the DB
- [ ] make a payment
- [ ] the invoice should not have the shadows in the background

### DIFF
--- a/src/app/modules/payments/__tests__/stripe.controller.spec.ts
+++ b/src/app/modules/payments/__tests__/stripe.controller.spec.ts
@@ -59,7 +59,7 @@ describe('stripe.controller', () => {
         ...mockBusinessInfo,
         formTitle: mockFormTitle,
         submissionId: mockSubmissionId,
-        background: false,
+        isBackgroundOn: false,
       }
       const mockForm = {
         _id: MOCK_FORM_ID,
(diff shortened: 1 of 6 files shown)


src/app/modules/payments/__tests__/stripe.controller.spec.ts
```diff
@@ -59,6 +59,7 @@ describe('stripe.controller', () => {
         ...mockBusinessInfo,
         formTitle: mockFormTitle,
         submissionId: mockSubmissionId,
+        background: false,
```
LinHuiqing: just noting: interesting choice of putting `false` here when default is true 😂

foochifa: ikr... it fails when i put true for some reason

foochifa: I kinda rushed this out, but will look at it properly (if no choice)

LinHuiqing: D: but yea I guess, just wait for JTC to get back ig

foochifa:
No longer needed as agency has cleared the current invoice background

PR #6213, opened 2023-04-27
fix: move payment drawer to Payments tab within builder drawer

### LINKED ISSUE #6018: Shift Payment Drawer from Payment Tab to Build and Design Tab
As seen in Form Design master, there is no payment tab in the side bar, instead payment should exists as a tab in the Build and Design Tab -> Field List Drawer.

This will affect how we render the payment content as well

From: 

Image

To:

Image

justynoh:
## Problem
We want to move the payments feature from a separate drawer in the sidebar to being a tab within the builder drawer.

Closes #6018
 
## Solution
The drawer content is moved into a new tab in the builder drawer. Payment preview element from public form is repurposed so that it can be wrapped by additional functionality required by the builder tool. Such functionality includes:
- Dynamic rendering of payment preview content
- Opening payment tab in builder drawer on click
- Hover and active state for the payment preview element.

**Breaking Changes** 

## Screenshots

https://user-images.githubusercontent.com/25571626/234950713-a5d49ce8-3fe1-499d-a51a-a2f33869fddf.mov

## Tests

Regression tests
- [ ] Payment tab inputs should be visible but disabled on email mode forms.
- [ ] Payment tab should not be visible at all if user does not have beta flag.
- [ ] Payment tab inputs should be visible but disabled if user has not connected their stripe account.
- [ ] When payment is not enabled:
  - [ ] Payment preview should not appear in the builder
  - [ ] Payment can be toggled to enabled in the Payments tab
  - [ ] A name and amount can be entered and the preview will be updated in the builder
  - [ ] Payment can be saved, and drawer should return to basic field list.
  - [ ] Both builder and public form should show the correct payment field information.
- [ ] When payment is enabled:
  - [ ] Payment should appear in builder with initial values taken from form definition
  - [ ] Payment tab inputs should be prefilled with initial values taken from form definition
  - [ ] Payment values can be modified, and the preview in the builder should change 
  - [ ] Payment can be disabled, and the preview in the builder should disappear

New functionality
- [ ] When payment is enabled, opening the payment tab should scroll to the payment preview and set it to active state (if payment is disabled, opening the payment tab will just look like scrolling to the bottom of the form)
- [ ] When payment preview is clicked, the payment tab should open and the preview should be in active state
- [ ] When payment preview is in focus, followed by another field or the startpage, the preview should lose focus (and the focus should be transferred to whatever was clicked instead)
- [ ] (Vice-versa from the previous) When anything else is in focus, and the payment preview is clicked, the original thing should lose focus and payment preview should be focused instead (and the payment tab opened).
- [ ] After any modifications on the payment tab, clicking out of it (i.e. to a normal field) should cause the "unsaved changes" modal.
- [ ] (Vice-versa from the previous) After any modifications on a normal field, clicking out of it should cause the "unsaved changes" modal.


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
```diff
@@ -246,40 +248,53 @@ export const PaymentInput = ({
...
-        setData: setDataSelector(state),
-        resetData: resetDataSelector(state),
-      }),
-      [],
-    ),
-  )
+  const isEncryptMode = form?.responseMode === FormResponseMode.Encrypt
+  const isStripeConnected =
+    isEncryptMode && form.payments_channel.channel === PaymentChannel.Stripe
+  const paymentsField = isEncryptMode ? form.payments_field : undefined
```
LinHuiqing, line 264:
```suggestion
  const paymentsField = isEncryptMode && form.payments_field
```

LinHuiqing (changes requested):
There's still some odd behaviour: 

> - [ ] When payment is not enabled:
>    - [ ] ...
>    - [ ] Payment can be saved, and drawer should close.

For the above manual test case, the drawer does not close, but instead, the tab is switched back to the fields tab.

Want to add other test cases too:
- [ ] After any modifications on the payment tab, clicking out of it (i.e. to a normal field) should cause the "unsaved changes" modal.
- [ ] (Vice-versa from the previous) After any modifications on a normal field, clicking out of it should cause the "unsaved changes" modal.

For the above 2 test cases, the former passes but the latter doesn't. The latter will just cause changes to the normal field to be discarded automatically.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
justynoh: The types won't work out here - the type of `paymentsField` is `FormPaymentField | undefined`, but your suggestion changes it to `FormPaymentField | false`.

justynoh:
Oops, about that test case - it's intended to go back to the field list, the same way that saving a normal field also returns the user back to the field list. The test case was written wrongly. Thanks for pointing that out!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
LinHuiqing: Oh yea saded

LinHuiqing (approved):
lgtm! just wanted to flag out some issues/ edge cases we will wanna address in future PRs (idt they should block this):

1. When there are no form fields and payment is enabled, editing payments then clicking on the form fields will not yield any feedback, which is confusing. But I think this might be an edge case we might not have to handle?
2. Payments tab is there but empty for email forms! Not very worrying since we have a small pool of users now but that will change soon (I hope?)

justynoh:
second point fixed in [64a6556](https://github.com/opengovsg/FormSG/pull/6213/commits/64a6556b44bf47e7f2b4bc8675ab2f5f782b769b), thanks for spotting that! For the first point i agree, it's an edge case we don't have to fix for now ba

PR #6214, opened 2023-04-28
fix: add payout pending section to individual response page

### LINKED ISSUE #6143: Make payout status always visible on payments details in response
Admins should not have to wonder whether they have received the payout or not.

justynoh:
## Problem
We want to display the payout pending section even when payout details have not come in.

Closes #6143 

## Solution
Change the value of the payout details to "-" when no payout details are sent to the frontend. Display these with the appropriate tag for the payout "status", as per [figma](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?node-id=16608-163982&t=QhakbMRTzLni6fwD-4).

**Breaking Changes** 

## Before & After Screenshots

Payout pending
Screenshot 2023-04-28 at 10 53 24 AM

Payout success
Screenshot 2023-04-28 at 10 53 16 AM

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/PaymentSection.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/PaymentSection.tsx
@@ -1,4 +1,5 @@
 import { BiCheck } from 'react-icons/bi'
+import { IconType } from 'react-icons/lib'
 import { Box, Divider, Flex, Icon, Link, Text } from '@chakra-ui/react'
 import { keyBy } from 'lodash'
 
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/responses/IndividualResponsePage/PaymentSection.tsx
```diff
@@ -66,24 +36,19 @@ export const PaymentSection = ({
       ? { label: 'Disputed', colorScheme: 'warning' }
       : undefined // The remaining options should never appear.
 
+  const payoutTagProps =
+    payment.payoutId || payment.payoutDate
```
foochifa, line 40: Just curious but `payoutId` and `payoutDate` will always exists together right? i.e. 0 || 0 and 1 || 1?

foochifa (approved):
Lgtm thanks Justyn!

For tests was wondering if we could somehow trigger payouts -> was trying manual payouts but it killed my backend container 😅 
image


But firstly maybe can just manually edit the db (which is what I guess was the screenshot)
We can document the steps down too like go to payments doc -> add payout.payoutId && payout.payoutDate.

frontend/src/features/admin-form/responses/IndividualResponsePage/PaymentSection.tsx
justynoh: yup, this is a bit of a deficiency in the dto. but to avoid complication, i think this is fine. In principle, if any one of them is there, we should consider it successful (since a payout is already "known" to occur by ID or by date).

PR #6219, opened 2023-04-28
fix: update payment modal to be fullpage on mobile

### LINKED ISSUE #6216: Payment modals should be full page in mobile view (respondent)
[figma here](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?type=design&node-id=16039-155802)

KenLSM:
## Problem

Inconsistencies with design

Closes #6216

## Solution

Changes to three modals that relates to payment when on a mobile browser UA.
- `DuplicatePaymentModal.tsx`
- `FormPaymentModal.tsx`
- `FormPaymentResumeModal.tsx`

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

| Name | Mobile | PC |
|--------|--------|--------|
| `DuplicatePaymentModal` | Screenshot 2023-04-28 at 3 09 51 PM | Screenshot 2023-04-28 at 6 05 32 PM |
| `FormPaymentModal` | Screenshot 2023-04-28 at 6 06 58 PM | Screenshot 2023-04-28 at 6 05 09 PM |
| `FormPaymentResumeModal` |Screenshot 2023-04-28 at 3 52 32 PM | Screenshot 2023-04-28 at 6 04 59 PM | 

**AFTER**:

| Name | Mobile | PC |
|--------|--------|--------|
| `DuplicatePaymentModal` | Screenshot 2023-04-28 at 3 12 40 PM| Screenshot 2023-04-28 at 6 03 00 PM |
| `FormPaymentModal` | Screenshot 2023-04-28 at 1 01 01 PM | Screenshot 2023-04-28 at 6 01 26 PM |
| `FormPaymentResumeModal` |Screenshot 2023-04-28 at 6 00 13 PM |Screenshot 2023-04-28 at 6 01 18 PM |
 
## Tests

Changes on Payment forms
  - [ ] Mobile mode
    - [ ] `DuplicatePaymentModal` should show correctly in full screen with buttons in correct order
      - [ ] Cancel button should dismiss modal and remain on page
      - [ ] Proceed to pay button should lead to new payment intent created
    - [ ] `FormPaymentModal` should show correctly in full screen with buttons in correct order
      - [ ] Cancel button should dismiss modal and remain on page
      - [ ] Proceed to pay button should lead to new payment intent created
    - [ ] `FormPaymentResumeModal` should show correctly in full screen with buttons in correct order
      - [ ] Start over again button should lead to new payment intent created
      - [ ] Restore previous session button should lead to previous payment page
      - [ ] There should be no `X` button
  - [ ] PC mode
    - [ ] `DuplicatePaymentModal` should show correctly as a modal with buttons in correct order
    - [ ] `FormPaymentModal` should show correctly as a modal with buttons in correct order
    - [ ] `FormPaymentResumeModal` should show correctly as a modal with buttons in correct order

Regressions tests
- [ ] Expect submission on non-payment forms to have no issues

### DIFF
--- a/frontend/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/frontend/src/components/ButtonGroup/ButtonGroup.tsx
@@ -1,0 +1,33 @@
+import {
+  ButtonGroup as CharkaButtonGroup,
+  ButtonGroupProps as ChakraButtonGroupProps,
+} from '@chakra-ui/react'
+
+export interface ButtonGroupProps extends ChakraButtonGroupProps {
+  /**
+   * If `true`, the button group will take up the full width of its container,
+   * and its children (<Button />) will be arranged in the reverse order
+   */
+  isFullWidth: boolean
+}
+
+export const ButtonGroup = ({
+  isFullWidth,
+  children,
+  ...rest
+}: ButtonGroupProps) => {
+  const buttonGrpResponsiveLayoutProps = isFullWidth
+    ? ({
+        flexDir: 'column-reverse',
+        w: '100%',
+        spacing: 0,
+        pt: '2rem',
+        rowGap: '0.75rem',
+      } as const)
+    : ({} as const)
+  return (
+    <CharkaButtonGroup {...buttonGrpResponsiveLayoutProps} {...rest}>
+      {children}
+    </CharkaButtonGroup>
+  )
+}
--- a/frontend/src/components/ButtonGroup/index.ts
+++ b/frontend/src/components/ButtonGroup/index.ts
@@ -1,0 +1,1 @@
+export { ButtonGroup as default } from './ButtonGroup'
--- a/frontend/src/features/public-form/components/DuplicatePaymentModal/DuplicatePaymentModal.tsx
+++ b/frontend/src/features/public-form/components/DuplicatePaymentModal/DuplicatePaymentModal.tsx
@@ -2,6 +2,7 @@ import { MouseEvent, MouseEventHandler } from 'react'
 import {
   Button,
   ButtonGroup,
+  ButtonGroupProps,
   Link,
   Modal,
   ModalBody,
@@ -13,6 +14,8 @@ import {
   Stack,
   Text,
 } from '@chakra-ui/react'
+
+import { useIsMobile } from '~hooks/useIsMobile'
 
 import { getPaymentPageUrl } from '~features/public-form/utils/urls'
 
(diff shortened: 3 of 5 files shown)


foochifa:
Hmm... Should we just abstract the modal out as like `paymentModal`?

Cause it seems like for every change we have to duplicate the changes 3 times 😅, not very dry right?

And then use props instead to configure them like
```
export type paymentModalProps = {
   title: string
   body: string
   proceedMessage: string
   cancelMessage: string
   onSubmit: ...
   onCancel: ...
   url?: string
}
```

Then we can keep the modals consistent, and it might be easier in the `PublicFormSubmitButton` too when we eventually have 3 states: 1. Prev Payment, 2. Processing Payment, 3. Normal Payment Modal

So we only have to vary the props going into the modal when onOpen

KenLSM:
Yeah, I was also pondering about this, but eventually decided not to generalize it as I'm not sure if its the correct abstraction to make -- having not seen enough use case.

Reasons being that there are two different kinds of modals here 
   - `FormPaymentModal | DuplicatePaymentModal` vs `FormPaymentResumeModal`
   -  the former has different behaviour, one is close-able though backdrop/close button, while the latter modal has both button mapped to side-effects

Wdyt?

foochifa:
Hmm yeah thats true... the behaviour of the first two is different from the `FormPaymentResumeModal`

And I'm not sure if its worth it to just modify the first two (though with `ProcessingPaymentModal` coming up, in hindsight maybe I should have abstracted those two hahaha)

Would it be possible to just abstract out the props instead?

So
```
export const mobilePaymentModalResponsiveLayoutProps = { size: 'full' }
export const mobilePaymentButtonGrpResponsiveLayoutProps: ButtonGroupProps = {
        flexDir: 'column-reverse',
        w: '100%',
        spacing: 0,
        pt: '2rem',
        rowGap: '0.75rem',
      }
```
Which are being re-written for every single modal currently, then we can at least ensure that if we need to modify the modal props for all our payment modals again, we can just change this props


For the manual tests, can I also suggest to check the functionality of all 3 of the modals (Startover / resume | link in duplicatepayment etc.) under mobile mode

KenLSM:
@foochifa extended the `<ButtonGroup />` from charka instead and kept the props declaration there as I could see the usage with `<Button />` in fullwidth mode after your comments.

As for the `mobilePaymentModalResponsiveLayoutProps`, I still felt that it should be declared from the parent level instead of abstracting out.

foochifa (approved):
lgtm! Tested on mobile too.

Thanks @KenLSM for the changes, it feels cleaner now!

PR #6234, opened 2023-05-02
fix: add env parameter to Stripe metadata, check value before processing events

justynoh:
## Problem
Our Stripe webhooks are being sent everywhere! This risks our webhook handlers being disabled by Stripe for consistently rejecting webhooks that aren't meant for each environment.
 
## Solution
An `env` parameter is added to the Stripe metadata. This enables us to check which environment the webhook is meant for when it is received. If the webhook is not meant for the environment, we return a `202 Accepted`. 

Additional random stuff this PR includes:
- Also, Joi validation was breaking for previous payment endpoint. It's not necessary, so i did a small refactor to remove it in `payments.controller`.
- Need to have better response codes for Stripe webhooks, so repurposed `mapRouteErr` to be used in webhook handler as well.

Sorry for the slight mess! 🙏 

**Breaking Changes** 
- Yes - this PR contains breaking changes
  - The shape of Stripe metadata has changed, and so some webhooks may fail if not handled properly. Currently, the requirement on the metadata object having the `env` value is not enforced, to enable backward-compatibility on webhooks that did not have the env parameter previously. Once we stop receiving such webhooks, we should enable these checks in `stripe.utils.ts`.

## Deploy Notes
- `SSM_ENV_SITE_NAME` : set this to `prod` for production prior to deployment


src/app/modules/payments/stripe.utils.ts
```diff
@@ -54,25 +93,42 @@ const isStripeMetadata = (
...
+  }
+  if (!mongoose.Types.ObjectId.isValid(metadata.paymentId)) {
+    logger.warn({
+      message: 'Got Stripe metadata with invalid paymentId',
       meta: { ...logMeta, metadata },
     })
     return err(new StripeMetadataValidPaymentIdNotFoundError())
   }
+  // Explicit check for metadata.env to ensure that non-existent legacy metadata
+  // passes through on production.
```
LinHuiqing: To clarify: do you mean to ensure that an error (ie `StripeMetadataIncorrectEnvError`) is thrown for legacy metadata?

src/app/modules/payments/stripe.utils.ts
```diff
@@ -54,25 +93,42 @@ const isStripeMetadata = (
...
+ * @returns err(StripeMetadataIncorrectEnvError) if the app is incorrect
  */
 export const getMetadataPaymentId = (
   metadata: Stripe.Metadata,
-): Result<string, StripeMetadataValidPaymentIdNotFoundError> => {
+): Result<
+  string,
+  | StripeMetadataInvalidError
+  | StripeMetadataValidPaymentIdNotFoundError
+  | StripeMetadataIncorrectEnvError
```
wanlingt, line 106: hm would it make more sense to check for the env in a separate function instead of `getMetadataPaymentId`? then we can also do the check earlier and avoid processing the webhook (in `stripe.controller.ts`) if it's from the wrong env

src/app/modules/payments/stripe.utils.ts
justynoh: maybe this was a bad description. In the legacy version, `env` does not exist. probably a better comment would have been `legacy metadata with non-existent env value` 😅

src/app/modules/payments/stripe.utils.ts
justynoh: hmmm, we could but it would still be quite messy. The target of the call to `getMetadataPaymentId` is different based on different cases (see the `switch` in `handleStripeEventUpdates`), since the metadata to be checked exists in different places in each type of event. If in the future we want to extract other data from the metadata, we can expand the scope of this function to extract all relevant values. It's just that currently, we only need the payment id.

src/app/modules/payments/stripe.utils.ts
justynoh: clarified in [d33087a](https://github.com/opengovsg/FormSG/pull/6234/commits/d33087a09a2865a5c465c4bdfec2bdb9fc999061)

src/app/modules/payments/stripe.utils.ts
```diff
@@ -43,6 +81,7 @@ export const getChargeIdFromNestedCharge = (
 const isStripeMetadata = (
   obj: Stripe.Metadata,
 ): obj is StripePaymentMetadataDto =>
+  // hasProp(obj, 'env') && // TODO: Make this required later
```
wanlingt, line 84: nit: can we create an issue tagged to the TODO so we don't forget about this later on?

src/app/modules/payments/stripe.utils.ts
```diff
@@ -54,25 +93,42 @@ const isStripeMetadata = (
...
+  if (!mongoose.Types.ObjectId.isValid(metadata.paymentId)) {
+    logger.warn({
+      message: 'Got Stripe metadata with invalid paymentId',
       meta: { ...logMeta, metadata },
     })
     return err(new StripeMetadataValidPaymentIdNotFoundError())
   }
+  // Explicit check for metadata.env to ensure that legacy metadata which does
+  // not have the env value still gets processed.
+  // TODO: remove the existence check later.
```
wanlingt, line 128: nit: same for the TODO here! :)

src/app/modules/payments/stripe.utils.ts
wanlingt: ok gotcha! so it's mainly in `getMetadataPaymentId` because that's where we expect to process the metadata (?).

wanlingt (approved):
lgtm! there's a typo in your PR description btw, i think you meant `payments.controller` instead of `payments.service`. 

Adding more context in this comment for posterity (feel free to add on + add it to the PR description also):
- Webhooks are sent to all connected platform accounts' endpoints when a Stripe account is connected to multiple platform accounts
- Stripe disables endpoints if an endpoint hasn’t responded with a 2xx HTTP status code for multiple days in a row ([Stripe docs](https://stripe.com/docs/webhooks/best-practices#events-and-retries))
- test Events are sent to both live and test connect webhook endpoints ([Stripe docs](https://support.stripe.com/questions/connect-account-webhook-configurations)). Instead of using the `livemode` property, we're choosing to use `env` instead to better distinguish between our different environments.

justynoh:
You're right, updated the PR description! Also for the TODOs, will open a PR for it almost immediately. Just need to make sure we only merge it in a week after this PR goes out

PR #6236, opened 2023-05-02
feat: add business field mutation to form-level

### LINKED ISSUE #6085: [REVIEW] Business Info (GST Reg No, Address) should be extended to form-level
The `GST Reg No`, and `Business Address` are read during invoice generation on the payment page (after responder has successfully paid and we've received the receipt url from stripe).

This two fields are currently stored and read from form-level details. However, as we expect large agencies to have different `GST Reg No`, and `Business Address`, we want to add support to store and alter them at the form-level.

Storing at the form-level also supports the edge case where the form is **transferred to a different agency**, but the payment channel information should still remain the same.

The agency-level information will then be used as a prefill when the admin first.

Context:
- #6078

KenLSM:
## Problem

The GST Reg No, and Business Address are read during invoice generation on the payment page (after responder has successfully paid and we've received the receipt url from stripe).

This two fields are currently stored and read from form-level details. However, as we expect large agencies (i.e., MOH, MOE) to have different GST Reg No, and Business Address, we want to add support to store and alter them at the form-level.

Storing at the form-level also supports the edge case where the form is transferred to a different agency, but the payment channel information should still remain the same.

Closes #6085

## Solution

Introducing form-level overrides.

**Breaking Changes** 

**Features**:

- Adds options for form admins to enter their own business information at the form level
- Invoice will respect form-level overrides before reading agency-level information

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-05-03 at 12 37 17 AM

**AFTER**:

Screenshot 2023-05-03 at 12 37 30 AM

Screenshot 2023-05-03 at 12 37 04 AM

Screenshot 2023-05-03 at 10 56 42 AM

## Tests

**Form Level Overrides**
Payment form business fields can be changed
- [ ] change `gstRegNo` field to `foobar`, expect that a toast appears confirming change
- [ ] change `gstRegNo` field to ` foobar` (with leading/trailing spaces), expect that a toast appears confirming change
- [ ] field is populated with `foobar` after a refresh

Agency default is shown as placeholder if form business info is not present
- [ ] change `gstRegNo` field to `` (empty string), expect that a toast appears confirming change and showing placeholder message
  - [ ] field is populated with agency defaults (OGP's agency gstRegNo/address is used)
  - [ ] data retrieved for `form.business.gstRegNo` and `*.address` is ``

Generated Invoice should use form-level business field if present
- [ ] prepare payment form with form-level overrides on `gstRegNo` + `address` fields with `foobar`
- [ ] generated invoice should show `GST Reg No: foobar` and `Address: foobar`

Business Info Fields will trim
- [ ] change `gstRegNo` and `address` field to ` foobar` (with leading/trailing spaces), expect that a toast appears confirming change
- [ ] field is populated with `foobar` (without leading/trailing spaces) after a refresh

**Regression tests**
- [ ] payment forms without overrides
  - [ ] other fields (Form Name) should be able editable
  - [ ] generated invoice should fallback to agency's business info

### DIFF
--- a/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
@@ -1,0 +1,111 @@
+import {
+  ChangeEventHandler,
+  Dispatch,
+  FocusEventHandler,
+  KeyboardEventHandler,
+  SetStateAction,
+  useCallback,
+  useRef,
+  useState,
+} from 'react'
+import { FormControl } from '@chakra-ui/react'
+
+import { FormResponseMode, StorageFormSettings } from '~shared/types'
+
+import FormLabel from '~components/FormControl/FormLabel'
+import Input from '~components/Input'
+
+import { useMutateFormSettings } from '../../mutations'
+import { useAdminFormSettings } from '../../queries'
+
+interface BusinessFieldInputProps {
+  initialValue: string
+  handleMutation: (newAddress: string) => void
+}
+const BusinessFieldInput = ({
+  initialValue,
+  handleMutation,
+}: BusinessFieldInputProps): JSX.Element => {
+  const [value, setValue] = useState(initialValue)
+
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  const handleValueChange: ChangeEventHandler<HTMLInputElement> = useCallback(
+    (e) => {
+      setValue(e.target.value.trim())
+    },
+    [],
+  )
+
+  const handleBlur = useCallback(() => {
+    if (value === initialValue) return
+    handleMutation(value)
+  }, [handleMutation, value, initialValue])
+
+  const handleKeydown: KeyboardEventHandler<HTMLInputElement> = useCallback(
+    (e) => {
+      if (e.key === 'Enter') {
+        e.preventDefault()
+        inputRef.current?.blur()
+      }
+    },
+    [],
+  )
+
+  return (
+    <Input
+      ref={inputRef}
+      value={value}
+      onChange={handleValueChange}
+      onKeyDown={handleKeydown}
+      onBlur={handleBlur}
+      placeholder={'Using Agency Defaults'}
+    />
+  )
+}
+
+const BusinessInfoBlock = ({ settings }: { settings: StorageFormSettings }) => {
+  const { mutateFormBusiness } = useMutateFormSettings()
+  const handleAddressMutation = (newAddress: string) => {
+    mutateFormBusiness.mutate({ address: newAddress })
+  }
+  const handleGstRegNoMutation = (newGstRegNo: string) => {
+    mutateFormBusiness.mutate({ gstRegNo: newGstRegNo })
+  }
+  return (
+    <>
+      <FormControl mb="2.5rem" isReadOnly={mutateFormBusiness.isLoading}>
+        <FormLabel description="Leave empty to use your agency defaults.">
+          GST Registration Number
+        </FormLabel>
+        <BusinessFieldInput
+          initialValue={settings.business?.gstRegNo || ''}
+          handleMutation={handleGstRegNoMutation}
+        />
+      </FormControl>
+      <FormControl mb="2.5rem" isReadOnly={mutateFormBusiness.isLoading}>
+        <FormLabel description="Leave empty to use your agency defaults.">
+          Business Address
+        </FormLabel>
+        <BusinessFieldInput
+          initialValue={settings.business?.address || ''}
+          handleMutation={handleAddressMutation}
+        />
+      </FormControl>
+    </>
+  )
+}
+
+export const BusinessInfoSection = () => {
+  const { data: settings, isLoading } = useAdminFormSettings()
+
+  if (
+    isLoading ||
+    !settings ||
+    settings.responseMode !== FormResponseMode.Encrypt
+  ) {
+    return <></>
+  }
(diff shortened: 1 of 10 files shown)


frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
```diff
@@ -0,0 +1,111 @@
...
+  initialValue,
+  handleMutation,
+}: BusinessFieldInputProps): JSX.Element => {
+  const [value, setValue] = useState(initialValue)
+
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  const handleValueChange: ChangeEventHandler<HTMLInputElement> = useCallback(
+    (e) => {
+      setValue(e.target.value.trim())
```
foochifa: Was wondering if we could delay the trimming to before inputting into handle mutation

The reason for this is when we are filling in the address it becomes really unintuitive.

e.g. We will type '36 Robinson Road', however with the current trim at setState it will be '36RobinsonRoad' and we have to go back to the spaces to add them.

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
```diff
@@ -0,0 +1,111 @@
+import {
+  ChangeEventHandler,
+  Dispatch,
+  FocusEventHandler,
+  KeyboardEventHandler,
+  SetStateAction,
```
foochifa: Couple of unused imports I think, `Dispatch`, `FocusEventHandler`, `SetStateAction`

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
```diff
@@ -0,0 +1,111 @@
...
+          initialValue={settings.business?.gstRegNo || ''}
+          handleMutation={handleGstRegNoMutation}
+        />
+      </FormControl>
+      <FormControl mb="2.5rem" isReadOnly={mutateFormBusiness.isLoading}>
+        <FormLabel description="Leave empty to use your agency defaults.">
+          Business Address
+        </FormLabel>
+        <BusinessFieldInput
+          initialValue={settings.business?.address || ''}
```
foochifa, line 113: Maybe not for this PR, but we should technically add the default agency values as the initial value right?

foochifa (commented):
Just some comments on `BusinessInfoSection`!

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
KenLSM: Right 👍 , definitely should delay this.

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
KenLSM: Yeah, in another feature ticket. This is ticket is mainly to keep the scope as small as possible to ship this for form admins to change. Prefilling takes a bit more time to think about.

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
KenLSM: bcde461d

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/BusinessInfoSection.tsx
KenLSM: 160ea4cb

Delayed to after user has clicked away from the input, trim to apply just before the value was sent as a patch to BE

src/app/modules/payments/stripe.controller.ts
```diff
@@ -361,28 +361,36 @@ export const downloadPaymentInvoice: ControllerHandler<{
           // convert to pdf and return
           .then((receiptUrlResponse) => {
```
foochifa, line 362: Maybe for another ticket/PR,

but was wondering if we might want to abstract out the entire section for converting the receipt to url (line 363-400) conversion to `payment.service.ts`

this may help us to write unit tests to ensure the behaviour is correct e.g.
- will use the form address/gst before agency if available
- will still generate invoice without address/gst (which may be in the future this might be an error?)

It may also be a bit cleaner on our controller 
`.then((receiptUrlResponse) => convertToInvoicePdf(receiptUrlResponse, populatedForm))`

P.s. yikes I pressed single comment instead of review 😅

foochifa (approved):
LGTM! Thanks Ken, have tested the changes on the trimming too

Some potential additional tests:
- [ ] add trailing whitespace to address, upon blur it should be removed

Also the test case to ensure the agency level defaults are used when form business is empty can be a bit more explicit 
- [ ] enter '' for gst reg no and address, ensure the placeholder 'Using Agency Defaults' is displayed
- [ ] make a payment and generate the invoice, the address and gst reg no should be the agency defaults

And another one for when no user provided address/gst and agency defaults are available
- [ ] for a form with no form level address/gst, and no agency defaults. Ensure an invoice can be generated with no errors

src/app/modules/payments/stripe.controller.ts
KenLSM: Yeah makes sense!

PR #6243, opened 2023-05-03
fix: switch captcha to download from globally accessible domain

KenLSM:
## Problem
React is failing to render for users in CN. But angular was working fine.

Note: this is part of the effort to deprecate angular.

## Solution

Switch `www.google.com` to `www.recaptcha.net` as we are [using recaptcha globally](https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally).

**Breaking Changes** 

**Bug Fixes**:

- Also fixes a small typo in missing `www.` subdomain in `recaptcha.net`


timotheeg (approved):
lgtm!

PR #6252, opened 2023-05-03
fix: missing paymentReceiptEmail for fetch fallback submissions

tshuli:
## Problem

- Adds in missing paymentReceiptEmail when using fetch fallback to submit payment forms


KenLSM (approved):
LGTM! Thanks again for catching this

PR #6253, opened 2023-05-03
feat: reference implementation for Shifting encryption boundary

tshuli:
**No merge, reference implementation**

WIP reference implementation, for the RFC on shifting encryption boundary (see link in working doc).

This PR demonstrates how shifting of encryption boundary will allow us to put in place backend validation on storage mode responses; eventually opening up the way for more server-side actions on storage mode form data as envisioned in the RFC (e.g. virus scanning, confirmation emails, etc).

## Demo steps
- Create storage mode form. Add a few short text fields. Submit. Observe that the network call is sent without e2e encryption.
- Add field validation (e.g. minimum number of chars). Submit with valid answer. Clone the network call. Modify the answer to be invalid in the network call. Submit the invalid network call. Backend should reject with `Invalid answer`.
- Check that the valid submissions have been correctly saved in the DB with encryption and can be retrieved in responses tab.

### DIFF
--- a/frontend/src/features/public-form/PublicFormService.ts
+++ b/frontend/src/features/public-form/PublicFormService.ts
@@ -182,6 +182,7 @@ export const submitStorageModeFormWithFetch = async ({
   formId,
   publicKey,
   captchaResponse = null,
+  paymentReceiptEmail,
 }: SubmitStorageFormArgs) => {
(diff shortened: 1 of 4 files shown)


LinHuiqing:
Superseded by #6587 and #6595, thanks for the reference implementation @tshuli!

PR #6261, opened 2023-05-04
fix: date-picker bug for negative UTC timezones attempt 3

LinHuiqing:
## Problem

> (taken from #6096)
> > (taken from #6025) 
> > Date picker was using the midnight of the date picked and taking into account the timezone where the user is filling in the form. However, it does not make sense to do so as users will be picking particular dates rather than timings, and the understanding of the different timezones should be established as part of the question.
> 
> However, bug fix in PR #6025 almost caused an incident as the fix removed converting to UTC on the date picker itself, which was causing [the selection offset bug](https://formsg.zendesk.com/agent/tickets/42733). 

Unfortunately, datepicker bug fix attempt #2 (#6096) caused an actual incident due to incorrect date validation for the pastOnlyValidator ([post-mortem](https://docs.google.com/document/d/1q-Dnbzsago2chvnZZouKWrmjQG2stJNyeE2ACgAP4eg/edit?usp=sharing)).

## Solution

For bug caused by #6096 - Restore original date validators.

**Breaking Changes** 

- [x] No - this PR is backwards compatible. Custom dates were already saved in midnight UTC in the DB.

**Improvements**:

- Cast time to UTC only right before submitting custom date so the conversion will be completely invisible to users
- Cast time to local timezone before being shown on the frontend so the frontend timezone will be consistent

**Bug Fixes**:

- Users with negative UTC timezones will now have their date fields correctly reflecting the dates selected on their date pickers
- `zonedTimeToUTC` was converting the timings to UTC and thus the actual date being referred to will not be preserved for respondents in negative UTC timezones. Implemented `normalizeDateToUtc` to ensure the actual date numbers are are preserved.
  - We now cast time from local timezone to UTC timezone using `Date.UTC` instead of converting time using `zonedTimeToUTC` (which we were using wrong anyway, as pointed out by @darrelhong in #5731)

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/37061143/229051162-f3d111a4-d7ce-406f-b6eb-b659891296c7.mov

**AFTER**:

Disallow past dates:

https://user-images.githubusercontent.com/37061143/231920965-0295168f-ea05-4fad-a3fe-bf6d890350cc.mov

Disallow future dates:

https://user-images.githubusercontent.com/37061143/231939426-16e43f97-7810-4429-b393-fc62d86d963c.mov

Custom date range:

https://user-images.githubusercontent.com/37061143/231940842-d14a3731-7d50-41d4-8a54-326581fd2135.mov

## Tests

To ensure that all our bases are covered, try to test the most extreme timezones. Here are the suggested timezones: 
- Most common use case: Singapore - Singapore (UTC +08)

Preparation:
- [ ] Create a form with a date field and open the form.

For the tests below, timezones A and B refer to the positive and negative timezones mentioned above, in any order.

i) Ensure that date picker for date field with "Disallow past dates" option exhibits expected behaviour:
1. - [ ] Set your timezone as A.
2. - [ ] On the admin panel, refresh, select and save the "Disallow past dates" option for the date field.
3. - [ ] On the respondent form, refresh and open the date picker. Only the current and future dates of the timezone should be shown.
4. - [ ] Select a date. The appropriate date should be reflected in the date field.
5. - [ ] Change your device's timezone to B.
6. - [ ] Repeat tasks 3-4.
7. - [ ] Repeat tasks 2-4.
9. - [ ] Change your device's timezone to A.
10. - [ ] Repeat tasks 3-4.

ii) Ensure that date picker for date field with "Disallow future dates" option exhibits expected behaviour:
1. - [ ] Set your timezone as A.
2. - [ ] On the admin panel, refresh, select and save the "Disallow future dates" option for the date field.
3. - [ ] On the respondent form, refresh and open the date picker. Only the current and future dates of the timezone should be shown.
4. - [ ] Select a date. The appropriate date should be reflected in the date field.
5. - [ ] Change your device's timezone to B.
6. - [ ] Repeat tasks 3-4.
7. - [ ] Repeat tasks 2-4.
9. - [ ] Change your device's timezone to A.
10. - [ ] Repeat tasks 3-4.

iii) Ensure that date picker for date field with "Custom date range" option exhibits expected behaviour:
1. - [ ] Set your timezone as A.
2. - [ ] On the admin panel, refresh and select the "Custom date range" option for the date field.
3. - [ ] Select a start date.
4. - [ ] Open the end date field date picker. Only the dates on or after the selected start date should be selectable.
5. - [ ] Select an end date.
6. - [ ] Save the selected date range. Optionally, check that the dates saved in the DB are in UTC.
7. - [ ] On the respondent form, refresh and open the date picker. Only the dates in the selected date range should be selectable.
8. - [ ] Select a date. The respective date should be reflected in the date field.
9. - [ ] Change your device's timezone to B.
10. - [ ] Repeat tasks 7-8.
11. - [ ] Repeat task 2. Ensure that the start and end dates remain the same.
12. - [ ] Update the start and end dates for the date field. Save it.
13. - [ ] Repeat tasks 7-8.
14. - [ ] Change your device's timezone to A.
15. - [ ] Repeat tasks 7-8.

iv) Ensure that email mode forms with date field still work:
- [ ] Repeat test set (i), making sure to select the current date using the date picker and submitting.
- [ ] Repeat test set (ii), making sure to select the current date using the date picker and submitting.
- [ ] Repeat test set (iii), making sure to:
   - [ ] Select the start date using the date picker and submitting.
   - [ ] Select the end date using the date picker and submitting.

Note: Here's [how to set your Mac's timezone](https://support.apple.com/en-gb/guide/mac-help/mchlp2996/mac).

Design/ product question: Should "Disallow past dates" and "Disallow future dates" be taking the past / future based on the admins' timezone, specified timezone or the users' timezone? Also, even when we have decided on the previous question, does it even make sense if the concept of time is not considered? E.g. if the recurring event I want to organise is at 10am, our date picker won't be able to know that so if we have a timezone in mind, it won't "filter" properly. Currently, it's simply implemented based on the users' timezone so its not disorienting.

### DIFF
--- a/frontend/src/utils/date.test.ts
+++ b/frontend/src/utils/date.test.ts
@@ -1,0 +1,107 @@
+import * as DateUtils from './date'
+
+describe('date', () => {
+  describe('isDateBeforeToday', () => {
+    it('should return true when the input date is in the past', () => {
+      const result = DateUtils.isDateBeforeToday(new Date('2023-04-23'))
+
+      expect(result).toBe(true)
+    })
+    it('should return true when the input date is yesterday', () => {
+      const now = new Date()
+      const result = DateUtils.isDateBeforeToday(
+        new Date(now.getFullYear(), now.getMonth(), now.getDate() - 1),
+      )
+
+      expect(result).toBe(true)
+    })
+    it('should return false when the input date is within today (e.g. now)', () => {
+      const result = DateUtils.isDateBeforeToday(new Date())
+
+      expect(result).toBe(false)
+    })
+    it('should return false when the input date is today', () => {
+      const now = new Date()
+      const result = DateUtils.isDateBeforeToday(
+        new Date(now.getFullYear(), now.getMonth(), now.getDate()),
+      )
+
+      expect(result).toBe(false)
+    })
+    it('should return false when the input date is in the future', () => {
+      const now = new Date()
+      const result = DateUtils.isDateBeforeToday(
+        new Date(now.getFullYear(), now.getMonth(), now.getDate() + 1),
+      )
+
+      expect(result).toBe(false)
+    })
+  })
+
+  describe('isDateAfterToday', () => {
+    it('should return true when the input date is in the future', () => {
+      const result = DateUtils.isDateAfterToday(new Date('3023-04-23'))
+
+      expect(result).toBe(true)
+    })
+    it('should return true when the input date is tomorrow', () => {
+      const now = new Date()
+      const result = DateUtils.isDateAfterToday(
+        new Date(now.getFullYear(), now.getMonth(), now.getDate() + 1),
+      )
+
+      expect(result).toBe(true)
+    })
+    it('should return false when the input date is within today (e.g. now)', () => {
+      const result = DateUtils.isDateAfterToday(new Date())
+
+      expect(result).toBe(false)
+    })
+    it('should return false when the input date is today', () => {
+      const now = new Date()
+      const result = DateUtils.isDateAfterToday(
+        new Date(now.getFullYear(), now.getMonth(), now.getDate()),
+      )
+
+      expect(result).toBe(false)
+    })
+    it('should return false when the input date is in the past', () => {
+      const now = new Date()
+      const result = DateUtils.isDateAfterToday(
+        new Date(now.getFullYear(), now.getMonth(), now.getDate() - 1),
+      )
+
+      expect(result).toBe(false)
+    })
+  })
+
+  describe('normalizeDateToUtc', () => {
+    it('should convert local dates to UTC', () => {
+      // We can only test using this function in different system times by
+      // manually changing system time and rerunning this test
+
+      const dateString = '2023-04-23T00:00:00'
+      const localDate = new Date(Date.parse(dateString))
+      const utcDate = new Date(Date.parse(`${dateString}+00:00`))
+
+      const result = DateUtils.normalizeDateToUtc(localDate)
+
+      expect(result).toStrictEqual(utcDate)
+    })
+  })
+
+  describe('loadDateFromNormalizedDate', () => {
+    it('should convert normalised (UTC) dates to local date', () => {
+      // We can only test using this function in different system times by
+      // manually changing system time and rerunning this test
+
+      const dateString = '2023-04-23T00:00:00'
+      const utcDate = new Date(Date.parse(`${dateString}+00:00`))
+      const localDate = new Date(Date.parse(dateString))
+
+      const result = DateUtils.loadDateFromNormalizedDate(utcDate)
+
+      expect(result).toStrictEqual(localDate)
+    })
+  })
+})
--- a/frontend/src/utils/date.test.ts
+++ b/frontend/src/utils/date.test.ts
@@ -101,6 +101,11 @@ describe('date', () => {
       timezoneMock.register('Etc/GMT+12')
       const negativeUtcDate = new Date(Date.parse(dateString))
 
+      // Check that the negative UTC date created is different from UTC time and
+      // negative UTC's midnight should happen after UTC's midnight
+      expect(negativeUtcDate).not.toStrictEqual(utcDate)
+      expect(negativeUtcDate.getTime()).toBeGreaterThan(utcDate.getTime())
+
       const result = DateUtils.normalizeDateToUtc(negativeUtcDate)
 
       expect(result).toStrictEqual(utcDate)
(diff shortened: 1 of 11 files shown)


LinHuiqing:
For `pastOnlyValidator` and `futureOnlyValidator`, I wonder if the conversions done are even warranted actually, especially considering that `answer` from `response` is a `DD MMM YYYY` string with no time in it. For now, the code change in this PR does not make any changes for this, but just something for discussion for correctness!

karrui:
Looking at the main problem, I would warrant that the client should not need to know about any server implementation.
The following inputs -> outputs should be handled on the client:

no past dates: prevent past dates according to client time
no future dates: prevent future dates according to client time
date range: accept as long as client date selected in range.

All client should do is to set the date object to start of day selected before passing to the client, no need to convert to UTC. The `futureOnlyValidator` and `pastOnlyValidator` should be able to handle any day range as long as the time is a start of day.

LinHuiqing:
> Looking at the main problem, I would warrant that the client should not need to know about any server implementation. The following inputs -> outputs should be handled on the client:
> 
> no past dates: prevent past dates according to client time no future dates: prevent future dates according to client time date range: accept as long as client date selected in range.
> 
> All client should do is to set the date object to start of day selected before passing to the client, no need to convert to UTC. The `futureOnlyValidator` and `pastOnlyValidator` should be able to handle any day range as long as the time is a start of day.

So regarding this, should we stick to the implementation in this PR as of now and open a new issue for handling date 'normalisation' on backend?

FYI: 
- Did some investigation on the custom date range saving - the dates are saved in UTC midnight in production.
- Date field responses for both storage mode and email mode forms are in strings (DD MMM YYYY).

karrui (commented):
seems correct, lgtm. Can we have some unit tests on the date util functions to ensure that the two functions work as expected?

LinHuiqing:
> seems correct, lgtm. Can we have some unit tests on the date util functions to ensure that the two functions work as expected?

ah yes, we should have that in place before pushing! thanks @karrui will add it in 👍 🌮

LinHuiqing:
> > seems correct, lgtm. Can we have some unit tests on the date util functions to ensure that the two functions work as expected?
> 
> ah yes, we should have that in place before pushing! thanks @karrui will add it in 👍 🌮

Regarding unit tests for date utils, I'm finding it particularly hard to test `normalizeDateToUtc` and `loadDateFromNormalizedDate` because there is no timezone information stored in `Date` so it implicitly uses local time when methods like `getFullYear`, `getMonth` and `getDate` are used, but it is what we're passing in as input... Migrating to use strings is getting more and more tempting 😭

karrui:
> because there is no timezone information stored in `Date`

Can try mocking with [`timezone-mock`](https://www.npmjs.com/package/timezone-mock)

LinHuiqing:
> > because there is no timezone information stored in `Date`
> 
> Can try mocking with [`timezone-mock`](https://www.npmjs.com/package/timezone-mock)

Thanks for the suggestion! It works

LinHuiqing:
Rebased to develop and force pushed so that Playwright Tests can pass. New commits since last review are "[fix: normalizeDateToUtc to return Date object](https://github.com/opengovsg/FormSG/pull/6261/commits/3d14e1ac49153bfc21449ec9c2990028d241fdbd)" and after.

frontend/src/utils/date.test.ts
```diff
@@ -0,0 +1,166 @@
...
+  describe('normalizeDateToUtc', () => {
+    beforeEach(() => {
+      timezoneMock.unregister()
+    })
+    it('should convert local dates to UTC', () => {
+      const dateString = '2023-04-23T00:00:00'
+      const localDate = new Date(Date.parse(dateString))
+      const utcDate = new Date(Date.parse(`${dateString}+00:00`))
+
+      const result = DateUtils.normalizeDateToUtc(localDate)
```
karrui, line 89: Might want to assert before and after the util is called to ensure dates are different first

LinHuiqing: Hm I don't think this is especially valuable as an assertion as ideally the test should still pass when the local time is UTC. But I get it that there is no value in the test if the local time and UTC time is the same, I guess. Will it be better that this test is removed or that the local time here is set as SGT?

karrui: The rationale is to assert that the mock date library is also working, just in case :P

LinHuiqing: haha okay, I'll add the assertion for the positive/negative utc tests then! yea when i tried the library I did print the dates to make sure hahaha

LinHuiqing: checks added here: [d232fad](https://github.com/opengovsg/FormSG/pull/6261/commits/d232fad80731bd43a00af5cd640898846fc69706)

LinHuiqing:
Note: 
Playwright tests failed twice for commit [d232fad](https://github.com/opengovsg/FormSG/pull/6261/commits/d232fad80731bd43a00af5cd640898846fc69706) due to being unable to install Playwright browsers: https://github.com/actions/runner-images/issues/7048

karrui (approved):
lgtm~

PR #6263, opened 2023-05-04
fix: migrate to withTransaction for retries

justynoh:
## Problem
A payment webhook failure occurred on 2 May around 1317hrs on a live form. After investigation, the bug was deemed to have been caused by a missing retry mechanism on the payment update transaction.

At the point of `payment.save()` within the transaction, mongo threw the following error:
```
MongoError: WriteConflict error: this operation conflicted with another operation. Please retry your operation or multi-document transaction.
```
This was found to be due to a `TransientTransactionError` caused by a race condition between simultaneous incoming webhooks being processed by different servers. The presumption was that mongo issues a blocking lock at the start of the transaction, causing the entire document to not be accessible during the transaction period. This would avoid the desync that would occur when two servers read the same version and write the same (incremented) version simultaneously.

However, this was not the case. It turns out the mongo does not use this method, and instead throws an error if a write operation fails.
image 
Ref: https://medium.com/mongodb-performance-tuning/tuning-mongodb-transactions-354311ab9ed6

The recommended recovery mechanism is to retry the operation. This is encapsulated by the `withTransaction` api, as explained in the mongoose documentation.

image

## Solution
An additional type of mongo error is detected (112 WriteConflict) and if the error is thrown by mongo, it is intercepted and manually upward-propagated via throw/catch to trigger `withTransaction` to retry the operation. All other application errors are propagated upward via `neverthrow`'s `errAsync` api. At the top level, since we have to rely on `withTransaction` to abort the transaction in the event of any application error, the application errors also have to be thrown/caught (to exit and re-enter the `ResultAsync` monad).

And we can see this works. Within 10 seconds, three webhooks were sent `payment_intent.created`, `payment_intent.succeeded` and `charge.succeeded`.  We can see that as expected, the `payment_intent.succeeded` webhook retried once, and the `charge.succeeded` retried twice. 

image

At the end, the completed submission is visible in the individual response pages. 

image

It also handles application errors correctly, as being distinguished from transient transaction errors. The following screenshot is a payment not found error, triggered by an incoming webhook from another environment:

image

**Breaking Changes** 

### DIFF
--- a/src/app/constants/mongo-error.ts
+++ b/src/app/constants/mongo-error.ts
@@ -1,4 +1,5 @@
 export const MONGO_ERROR_CODE = {
+  WriteConflict: 112,
   DuplicateKey: 11000,
   InvalidBSON: 10334,
 }
--- a/src/app/modules/core/core.errors.ts
+++ b/src/app/modules/core/core.errors.ts
@@ -51,6 +51,12 @@ export class DatabaseDuplicateKeyError extends ApplicationError {
   }
 }
 
+export class DatabaseWriteConflictError extends ApplicationError {
+  constructor(message: string) {
+    super(message)
+  }
+}
+
 export class SecretsManagerError extends ApplicationError {
   constructor(message?: string) {
     super(message)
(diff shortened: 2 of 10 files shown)


justynoh:
All this being said, this has a functionality implication. We used to do post-submission actions (e.g. fire webhooks and email confirmations) immediately after creating the new submission document, even though it may not have yet been committed. Therefore, this PR also does the work to move that functionality out of the transaction, so that it is only done if the transaction has been successfully committed for a `charge.succeeded`.

Since i was modifying the post-submission actions function, I also cleaned up the types and added its errors to the `mapRouteError` function.

KenLSM (approved):
LGTM

PR #6273, opened 2023-05-05
fix: unawaited promises lint issues on session.end() on tests and stri…

KenLSM:
Most of mongodb native functions are now [returning as promises instead of callback](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_5.0.0.md#migrate-to-promise-based-api-recommended).

`session` functions are now async instead of sync with callback

**Fixes**
- `src/app/modules/payments/stripe.service.ts`: `session.endSession()` wrapping it with `ResultAsync.fromPromise` `src/app/modules/submission/__tests__/submission/submission.service.spec.ts`: `session.endSession()` here is awaited before the test concludes

### DIFF
--- a/src/app/modules/payments/stripe.service.ts
+++ b/src/app/modules/payments/stripe.service.ts
@@ -338,12 +338,18 @@ export const processStripeEvent = (
       },
     )
       .andThen(() => {
-        session.endSession()
-        return okAsync(undefined)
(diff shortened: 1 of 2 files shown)


foochifa (approved):
lgtm!

Though we still have quite a lot of work todo for linting all the server models 😅

timotheeg (approved):

Issue #6045, opened 2023-04-05
Performance optimization for large forms


timotheeg:
Large forms (>~75 fields) are very laggy in the admin interface, and somewhat laggy in the respondent view.

That makes the product unusable in these cases.

TODO:
* instrument and do performance analysis on code, to find out the bottleneck (I assume we're going to see overly heavy react render cycles for some reason)
* Experiment with solutions to improve situation, pick best fit and release

timotheeg:
@foochifa we can consider this done yeah?

foochifa:
The user is gonna test out on Wed morning!

tagging @kennethchangOPENGOV

There are still a few smaller perf issues such as lag when dropping the fields, some lag when editing the field, not sure if those are critical though

timotheeg:
Thanks! And yes I think we can be open specific issues to fix those. For now, I think we can close this on successful test Wednesday.

PR #6285, opened 2023-05-08
feat: duplicate fields to next row

foochifa:
## Problem

Currently when we duplicate fields, it is duplicated to the end of form fields. This makes it difficult for users of large forms to manoeuvre or modify their forms, and causes a lot of unneeded drag and drops.

This is related to our efforts to optimise performance for large forms in #6045

## Solution

Form fields are now duplicated to the index immediately after the original field.

During implementation we had 2 potential methods of approaching this problem.

1. Follows the current implementation, whereby Backend's `admin-form.service` will push the new duplicated field to end of form fields array. Whilst Frontend will do the same in `useDuplicateFormField`. Therefore, an explicit sync that we have to maintain between the two. So instead of pushing, both Backend and Frontend will look for the field to be duplicated and find its index + 1. This is the cleanest method of implementing this feat.
2. The alternative is to pass the insertion index from Frontend to the Backend at `/duplicate` endpoint as a `req.body`. This will allow an implicit sync between the Frontend and the Backend as the insertion index is passed between them. This can help in extensibility in the future too, if we have other features that requires duplicating to certain indexes.

Decided on option 1, as it is the cleanest approach and the Backend does not rely on client input

**Breaking Changes** 

- [x] Yes - see @LinHuiqing comment below

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/59867455/236801345-e65ba7f0-6c4f-41a1-9ab6-0f71125895ec.mov

**AFTER**:

https://user-images.githubusercontent.com/59867455/236801855-76ba2c03-76de-41be-98ae-650c1b269f39.mov

## Tests

In a storage mode form
- [ ] Create a few new fields (>2).
- [ ] Duplicate the top field, the new duplicated field should appear just below the original field. It should be auto-focused by the drawer. 
- [ ] Make some edits to the new duplicated field
- [ ] Refresh the page. The orders and details of the fields should still be the same (to ensure BE and FE queryclient are equivalent)
- [ ] Duplicate the last field, to make sure no edge issues.
- [ ] Refresh the page.
- [ ] Go to form preview, the fields in preview mode should be as per admin form view
- [ ] Open the form, the fields in public form should be as per admin form view

Email mode form
- [ ] do the same for an email mode form


LinHuiqing (commented):
Tested! But this PR is not backwards compatible! In the case where a new FE hits a old BE, the fields will be duplicated at the next index on FE but BE will save it at the end of the array of form fields. Conversely, if old FE hits new BE, FE will show that its added as the last field but BE will add it to the next one. This could be an issue with the volume of traffic we get, not sure if we have metrics as to how often admins duplicate fields?

foochifa:
Oh boy you are right... Thats the problem when FE and BE assumes each other knows, which is the problem of implementation 1

Not sure how we can circumvent this though? From speaking to Kenneth and Stacey, qualitatively, seems like duplicate is quite widely used. 

Then perhaps it might be best to only push this change at night or early mornign? 😭

LinHuiqing:
> Tested! But this PR is not backwards compatible! In the case where a new FE hits a old BE, the fields will be duplicated at the next index on FE but BE will save it at the end of the array of form fields. Conversely, if old FE hits new BE, FE will show that its added as the last field but BE will add it to the next one. This could be an issue with the volume of traffic we get, not sure if we have metrics as to how often admins duplicate fields?

Okay, checked and it could be up to ~50 in 5min. Could be substantial 🤔

foochifa:
Thanks @LinHuiqing, yeah I think we shouldn't release this during work hours... Will plan with the on-call person next week to release after hours!

LinHuiqing:
> Oh boy you are right... Thats the problem when FE and BE assumes each other knows, which is the problem of implementation 1
> 
> Not sure how we can circumvent this though? From speaking to Kenneth and Stacey, qualitatively, seems like duplicate is quite widely used.
> 
> Then perhaps it might be best to only push this change at night or early mornign? 😭

Yea actually I think there's no easy way to circumvent this that won't take too long :') If we really wanna do it safely, can break it up into multiple releases:
1. Set up endpoint on BE for new duplicate endpoint
2. Update FE to duplicated on next index and point to new BE endpoint when duplicating
3. Deprecate / remove old duplicate BE endpoint

edit: points 1 + 2 can be done in the same release, then 3 in the next release

frontend/src/features/admin-form/create/builder-and-design/mutations/useDuplicateFormField.ts
```diff
@@ -60,7 +60,9 @@ export const useDuplicateFormField = () => {
         // Should not happen, should not be able to update field if there is no
         // existing data.
         if (!oldForm) throw new Error('Query should have been set')
-        oldForm.form_fields.push(newField)
+        const insertionIndex =
+          oldForm.form_fields.findIndex((o) => o._id === fieldId) + 1
```
timotheeg, line 64: I assume `fieldId` MUST exists in `form_fields` or we'd have some big problem, but just to note: if `fieldId` is not found, then the code here doesn't do what the previous code did (it does an `unshift()` instead of a `push()`; because `findIndex()` returns -1 it not found, so the new index will be 0).

I'm not sure if we need defensive programming here of getting the reference index first and then do something like:

```typescript
const referenceIndex = findIndexAt();
if (referenceIndex > -1) {
    oldForm.form_fields.splice(referenceIndex + 1, 0, newField)
}
else {
    oldForm.form_fields.push(newField)
}
```

Basically same logic as [further below](https://github.com/opengovsg/FormSG/pull/6285/files#diff-fc907db13716e41db0eceb536856aba7ab346bc4bdb07b4ed70534d83c86d64fR542). In fact could this not use `getFormFieldIndexById()`?

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -531,15 +536,21 @@ export const duplicateFormField = (
   FormFieldSchema,
   PossibleDatabaseError | FormNotFoundError | FieldNotFoundError
 > => {
+  const fieldIndex = getFormFieldIndexById(form.form_fields, fieldId)
+  // if fieldIndex does not exist, append to end of form fields
+  const insertionIndex =
+    fieldIndex === null ? form.form_fields.length + 1 : fieldIndex + 1
```
timotheeg: ⚠️ This looks wrong since JS arrays are 0-indexed!

```suggestion
    fieldIndex === null ? form.form_fields.length : fieldIndex + 1
```

Oh! LOL, tThe code is just lucky, because `splice()` push-to-end rather than creating empty element if the insertion index is too large
```javascript
const arr = ['a']
arr.splice(arr.length + 5, 0, 'b')
console.log(arr) // ['a', 'b']
```

src/app/modules/form/form.utils.ts
```diff
@@ -105,6 +105,19 @@ export const getFormFieldById = (
...
 
+export const getFormFieldIndexById = (
+  formFields: IFormSchema['form_fields'],
+  fieldId: FormFieldSchema['_id'],
+): number | null => {
+  if (!formFields) {
+    return null
+  }
+
+  const index = formFields.findIndex((f) => fieldId === String(f._id))
```
timotheeg, line 116: Why is the `String()` wrapper necessary here when it wasn't used in [useDuplicateFormField.ts](https://github.com/opengovsg/FormSG/pull/6285/files#diff-067e20b910f5477ea8e285ad392275bcca359c3af3032d8269792d1b88c529af)?

src/app/modules/form/admin-form/admin-form.service.ts
foochifa: ~~Oh this was actually intended! Though technically field index should never be null, but in the case that it is, the behaviour will return to array.push as per before hence to insert at the end!~~ (misread HAHA)

splice will push-to-end if index is > length ahaha have tested it, originally was considering whether to set as num infinite but decided on length being the better option

src/app/modules/form/form.utils.ts
foochifa: Thats because the `f._id` here is a Mongoose `ObjectId` whilst the one in `useDuplicateFormField` is a `string` hahaha

And the reason why I had to wrap it in string instead of using `equals` comparator is because for some reason typescript or our current mongoose type does not recognise `ObjectId.equals` which is very weird 😩

frontend/src/features/admin-form/create/builder-and-design/mutations/useDuplicateFormField.ts
foochifa: Oh boy, okay let me change the behaviour of this portion.

It can't exactly use `getFormFieldIndexById()` as that method is in our BE utils, and references the field as a mongoose schema. Whilst this oldForm.form_fields is from an `AdminFormDto`

This is the problem with our current implementation where both BE and FE 'assumes' the same behaviour

foochifa: Thanks for catching this Tim [abd2b64](https://github.com/opengovsg/FormSG/pull/6285/commits/abd2b649d95ace20c05f44e756172189c1881d0b)

src/app/modules/form/admin-form/admin-form.service.ts
foochifa: Oh but yes the +1 is unneeded for the field length 😅 sorry for missing that out [855ef96](https://github.com/opengovsg/FormSG/pull/6285/commits/855ef96e38bb3e316907a88e3e3afa68fe7c4fff)

wanlingt (approved):

PR #6289, opened 2023-05-09
fix: update sms copy to remove link (policy)

timotheeg:
## Problem
The sms copy for OTPs include the appHost (likely done so we can distinguish staging/prod). 

While it is not a link, with no inclusion of a protocol or path AND sms can contain gov.sg sin Singapore, some agencies would rather not see any links at all to guarantee we comply to regulations.

## Solution
Update copy to remove `appHost` and instead use `FormSG` as a simple string.

The change is backward compatible

cc @amitogp @wanlingt


LinHuiqing (approved):
lgtm!

timotheeg:
Ignoring playwright tests, and merging now.

PR #6290, opened 2023-05-09
fix: failing test in SpcpOidBaseClientCache

kenjin-work:
## Problem
The test ``SpcpOidcBaseClientCache › retrievePublicKeysFromNdi() › should call the NDI JWKS endpoint and not retry more than 2 more times if it fails`` is flaky as it can be called a variable number of times (below 3).

Example of a failing test https://github.com/opengovsg/FormSG/actions/runs/4921370873/jobs/8791281114.

## Solution
Change the test to be limited to <= 2 tries.

**Breaking Changes** 

### DIFF
--- a/src/app/modules/spcp/__tests__/spcp.oidc.client.cache.spec.ts
+++ b/src/app/modules/spcp/__tests__/spcp.oidc.client.cache.spec.ts
@@ -433,7 +433,7 @@ describe('SpcpOidcBaseClientCache', () => {
 
       // Assert
       await expect(keyResultPromise).rejects.toThrow('Failure')
-      expect(axiosSpy).toHaveBeenCalledTimes(3)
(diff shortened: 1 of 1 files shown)


kenjin-work:
It seems that the test can indeed sometimes return the value 3...  :O.

timotheeg (approved):

timotheeg:
@kenjin-work since you are a member of formsg now, do create your PRs from branches, rather than forked repos. Otherwise some of our automation checks cannot run.

timotheeg:
Hmm, in fact, Looks like I cannot merge this PR. Do you mind closing it and opening it again from a branch?

kenjin-work:
> @kenjin-work since you are a member of formsg now, do create your PRs from branches, rather than forked repos. Otherwise some of our automation checks cannot run.

@timotheeg I don't think I've been added to the formsg repo yet :(.

PR #6296, opened 2023-05-09
chore: add script for post-release steps

wanlingt:
## Problem

The post-release steps are still done manually

## Solution

Create a script which:
1. Pushes the latest release on `release-al2` to `uat`
2. Creates a PR to merge `release-al2` to `develop`
3. Creates a new Github release from the release tag

Simply run `bash scripts/post_release_prep.sh` to execute these steps

### DIFF
--- a/scripts/post_release_prep.sh
+++ b/scripts/post_release_prep.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -x
+
+# To be executed after release has been deployed to prod
+# pre-requisites: install github CLI
+# - github documentation: https://github.com/cli/cli#installation
+
+git fetch --all --tags
+git checkout release-al2
+git reset --hard origin/release-al2
+
+# get latest release version
+release_version=$(git ls-remote --tags --sort=creatordate | grep -o 'v.*' | tail -1)
+echo ${release_version}
+
+# push latest release to UAT
+git push -f origin release-al2:uat
+
+# Create PR to merge release-al2 to develop
+gh pr create \
+  -H release-al2 \
+  -B develop \
+  -t "build: merge release ${release_version} to develop" \
+  -b ""
+
+# Create new GitHub release from release tag
+gh release create ${release_version} \
+  --generate-notes


scripts/post_release_prep.sh
```diff
@@ -0,0 +1,29 @@
...
+# - github documentation: https://github.com/cli/cli#installation
+
+git fetch --all --tags
+git checkout release-al2
+git reset --hard origin/release-al2
+
+# get latest release version
+release_version=$(git ls-remote --tags --sort=creatordate | grep -o 'v.*' | tail -1)
+echo ${release_version}
+
```
KenLSM, line 16: Im thinking if its possible to add a pause and question before continuing with release?

```
echo "Publishing $release_version, would you like to proceed? [y,n]"

read publishInput

if [[ $publishInput != 'y' ]]
then
    echo "Aborting"
    exit 1
fi

echo "done something"
```

scripts/post_release_prep.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -x
+
+# To be executed after release has been deployed to prod
+# pre-requisites: install github CLI
+# - github documentation: https://github.com/cli/cli#installation
```
KenLSM, line 7: Might also want to check if `gh` exist on the shell env before executing the script since there's a `git push` before `gh` execution making re-runs of this script to be non-idempotent (might have side-effects).

Although, this probably should be checked on the release script too.

```
which gh > /dev/null

if [[ $? -gt 0 ]]
then
    echo "github CLI not in installed. See https://github.com/cli/cli#installation"
    # or send them to https://www.google.com/search?q=github+cli
    exit 1
fi
```

timotheeg (commented):
Can we investigate if these can be done entirely automaticly with github actions? (I expect there might be some complication with protected branches, but hopefully there are solutions for automation out-there?)

Reason for my ask: the PR flow to merge from `release-al2` to `develop` is basically a waste of time for everyone involved, even with a script 😅 . Since `release-al2` is typically branched off `develop`, the risk of conflicts for backporting is very low, and it's standard process to backport from `release-al2` into `develop` anyway, so the flow of PR + run checks + approve + click-to-merge should be unnecessary.

UAT, I'm a little bit more conflicted. I have seen that we do push to UAT independently from release once in a while, so I don't know if SOP should be to always deploy to UAT after deployment to release-al2 🤔 . Are there cases, we would not want that to happen?

wanlingt:
> Can we investigate if these can be done entirely automaticly with github actions? (I expect there might be some complication with protected branches, but hopefully there are solutions for automation out-there?)

Looking into it!

> UAT, I'm a little bit more conflicted. I have seen that we do push to UAT independently from release once in a while, so I don't know if SOP should be to always deploy to UAT after deployment to release-al2 🤔 . Are there cases, we would not want that to happen?

I think Amit uses UAT for agencies to test payments fairly regularly, while previously it was only being used occasionally for specific cases. Hence thought that it might be important to keep UAT up to date! That way we have a guarantee that UAT is stable, but we can still push to UAT independently if we want to outside of the regular post-release deployments. What do you think?

LinHuiqing:
Regarding pushing to UAT, although it is something we want to keep up to date, maybe it's something we still want to do manually in the case of a release incident? Because in the case of an incident, we want to reduce the mental load of the person doing the release as much as possible (since UAT will likely have to be rolled back as well). Since we are indeed using UAT to onboard agencies for payments, I think between having a slightly old UAT and a broken UAT, the former is still better.

wanlingt:
I'm going to close this PR first. Will 
- keep the UAT process manual
- investigate if this is possible with gh actions

PR #6208, opened 2023-04-27
fix: playwright tests cancelled after timeout

LinHuiqing:
## Problem

Playwright tests are wonky and we want to fix them (to some extent) so we can at least have some e2e tests!

Closes https://github.com/opengovsg/FormSG/issues/6154

## Solution

Fixed several bugs in the playwright tests: 

- Correct order for retrieving OTP from email (fixed in [62d2dd1](https://github.com/opengovsg/FormSG/pull/6208/commits/62d2dd1e79cb13451296c52a7398d4073b4cb5bb))
- Add missing environment variable which was introduced in https://github.com/opengovsg/FormSG/pull/6234 (fixed in [0d65ebd](https://github.com/opengovsg/FormSG/pull/6208/commits/0d65ebd80596467237f394ff9d2968b614d756bb) and [3af498e](https://github.com/opengovsg/FormSG/pull/6208/commits/3af498ee26d0b8c9f0246f2052cd3368e7042075))
- Update to `getRoleBy` combobox when using dropdown for field settings (done in [87c5a40](https://github.com/opengovsg/FormSG/pull/6208/commits/87c5a4013a00e8e7f71953ca14e86ef96025e9a8)).
- `fillMultiDropdown` issues:
   - Don't close dropdown between item selections.
   - Scroll options into view if they're not in view.

Issues left to tackle regarding playwright tests:

- https://github.com/opengovsg/FormSG/issues/6306
- https://github.com/opengovsg/FormSG/issues/6307

For future reference when playwright tests timeout, here is the estimated time taken for a successful run of tests on GitHub Actions (as of this PR):

- `login.spec.ts` (9 tests): ~ 1 min
- `encrypt-submission.spec.ts` (15 tests): ~ 18 min
- `email-submission.spec.ts` (5 tests): ~ 10 min

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

PR #6300, opened 2023-05-11
ci: debug playwright tests

LinHuiqing:


LinHuiqing:
Using https://github.com/opengovsg/FormSG/pull/6208 to debug instead

PR #6303, opened 2023-05-11
perf: conditionally use hooks based on activeness of field row

### LINKED ISSUE #6045: Performance optimization for large forms

foochifa:
## Problem

Despite #6189 improving performance of dragging and updating field rows slightly, users can still face significant lags when attempting to:
1. Create new field
2. Edit fields
3. Dragging / reordering fields 

This is because of existing hooks still causing re-renders of all field rows from various user actions.

Closes #6045

## Solution

Currently, all of the field rows are too 'smart' and knows a lot of unnecessary global states/contexts, such as field builder states, entire form data (just to compute attachment size limits), and mutation methods to duplicate/delete fields. 

Firstly, we identify properties that all field rows **must** know. These properties are mostly used in the logic for `handleFieldClick` in `FieldRowContainer`.

There are specifically two properties that all fields need to keep track of:
1. `handleBuilderClick`: As the builder must be displayed upon click, thus, non-active fields must also have the correct/updated method to display the Builder
2. `isDirty`: This is to ensure if user click to another field, upon any existing unsaved changes, the dirty modal will still be registered and displayed. 

Except for the two above, all the other properties can be conditionally called based on `isActive`

**Steps done in this PR to abstract unnecessary properties out of `FieldRowContainer`**

1. Currently the `Collapse`-isble button group containing Edit (in mobile), Duplicate, and Delete, always exists for ALL field rows even if they are not displayed. This button group will only be displayed upon a field being selected (active). Hence, will conditionally render this button group based on whether the field row is active
2. As many of the hooks (such as `useDuplicateFieldMutation`) are being used for `handleDuplicateClick` and `handleDeleteClick`, we can abstract out the entire button group as its own component. Then move these hooks into that specific component. As this component will ONLY be rendered for **one** field (the active field), all re-renders from these hooks will be isolated to this `FieldRowButtonGroup` component for only one `FieldRowContainer`.
3. Additionally, as `fieldBuilderState` of the `FieldRowContainer` is primarily used when the field is active, have made it into an optional prop. This will prevent unnecessary re-renders from the updating of `fieldBuilderState`, hence making most of the fields more 'stupid'. 
4. Lastly, we are currently passing `isDraggingOver` to all field rows, however, from testing it seems that it only impacts the active field. Hence, have made it tied to activeness of the field. Allowing for faster start and drop of dragging action, as only the field being activated will be re-rendered

Other optimizations: 
1. Shifting `isDirty` hook to `BuilderFields`, for better debugging of prop changes. Now all major re-renders of `FieldRowContainers` should be seen from the change in props at the React profiler. Allowing for easier debugging of performance issues in the future
3. Shifting `colorTheme` hook to `BuilderFields`, as the underlying logic uses `useCreateTabForm` that causes re-renders upon any form definition changes (such as saving a field), despite the value remaining static without any design tab changes

Remaining re-renders that cannot be trivially solved:
1. Upon selecting any field (opening drawer)
2. First changes to a field (isDirty)
3. Saving a field (isDirty and drawer change)
4. Deleting a field (index changes, drawer change)
5. Dropping a field (index changes)

On why I did not conditionally render the drag icon:
1. Too little time saving, the chakra drag button is a lot cheaper than the button group
2. Causes issues with the formatting -> will need conditional changes to display an empty component to mimic the hidden component, as `Fade` keeps the margins of the hidden component. 

**There is likely no more trivial improvements we can make to field rows performance after this PR**

**Breaking Changes** 

## Before & After Screenshots

Video capture performing a simple process of:
1. Creating a new field
2. Reordering the field
3. Editing the field
4. Deleting the field

**BEFORE**:

takes 1min

https://github.com/opengovsg/FormSG/assets/59867455/c57a9660-381b-4e1b-bc28-a9c2ec021966

**AFTER**:

takes 35seconds

https://github.com/opengovsg/FormSG/assets/59867455/da4db689-6e3c-43b9-b015-93d99aac341c

## Tests

**To ensure no regression issues**
All the behaviours below should work as intended
- [x] create a new field, delete it
- [x] create a new field, save the field details
- [x] create a new field by dragging from builder to preview
- [x] edit a new field, the field in the preview content should update as well
- [x] edit a new field, do not save, attempt to select another field. Dirty modal should display
- [x] edit a new field, do not save, attempt to refresh, browser should prompt that it is dirty
- [x] edit a new field, save, builder should go back to the list of fields. And the preview content should reflect the edits
- [x] delete a field, the delete modal should show up, you should be able to delete the field
- [x] duplicate a field
- [x] drag a field, the field should hover over the potential droppable locations
- [x] drop the field, all affected fields should have their question numbers updated
- [x] add logic that conditionally displays the field. The fields hidden by logic should be translucent
- [x] the fields hidden by logic can be edited successfully
- [x] the fields hidden by logic can be dragged and dropped successfully
- [x] preview the form, all your changes should be reflected in the preview

On mobile
- [x] repeat the tests (except dragging from builder & selecting another field while editing)

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
@@ -50,7 +50,7 @@ export const BuilderFields = ({
           key={f._id}
           field={f}
           isHiddenByLogic={!visibleFieldIds.has(f._id)}
-          isDraggingOver={f._id === activeFieldId ? isDraggingOver : false}
+          isDraggingOver={f._id === activeFieldId && isDraggingOver}
           fieldBuilderState={
             f._id === activeFieldId ? stateData.state : undefined
           }
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
@@ -4,13 +4,14 @@ import { useCreatePageSidebar } from '~features/admin-form/create/common/CreateP
 import { augmentWithQuestionNo } from '~features/form/utils'
 import { FieldIdSet } from '~features/logic/types'
 
-import { useBuilderAndDesignContext } from '../BuilderAndDesignContext'
 import { PENDING_CREATE_FIELD_ID } from '../constants'
+import { isDirtySelector, useDirtyFieldStore } from '../useDirtyFieldStore'
 import {
   FieldBuilderState,
   stateDataSelector,
   useFieldBuilderStore,
 } from '../useFieldBuilderStore'
+import { useDesignColorTheme } from '../utils/useDesignColorTheme'
 
 import FieldRow from './FieldRow'
 
@@ -29,9 +30,17 @@ export const BuilderFields = ({
   const stateData = useFieldBuilderStore(stateDataSelector)
 
   const { handleBuilderClick } = useCreatePageSidebar()
-  const {
-    deleteFieldModalDisclosure: { onOpen: onDeleteModalOpen },
-  } = useBuilderAndDesignContext()
+
+  const activeFieldNumber =
+    stateData.state === FieldBuilderState.EditingField
+      ? stateData.field._id
+      : stateData.state === FieldBuilderState.CreatingField
+      ? PENDING_CREATE_FIELD_ID
+      : null
+
+  const colorTheme = useDesignColorTheme()
+
+  const isDirty = useDirtyFieldStore(isDirtySelector)
 
   return (
     <>
@@ -42,16 +51,12 @@ export const BuilderFields = ({
           field={f}
           isHiddenByLogic={!visibleFieldIds.has(f._id)}
           isDraggingOver={isDraggingOver}
-          isActive={
-            stateData.state === FieldBuilderState.EditingField
-              ? f._id === stateData.field._id
-              : stateData.state === FieldBuilderState.CreatingField
-              ? f._id === PENDING_CREATE_FIELD_ID
-              : false
+          fieldBuilderState={
+            f._id === activeFieldNumber ? stateData.state : undefined
           }
-          fieldBuilderState={stateData.state}
           handleBuilderClick={handleBuilderClick}
-          onDeleteModalOpen={onDeleteModalOpen}
+          isDirty={isDirty}
+          colorTheme={colorTheme}
         />
       ))}
     </>
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
@@ -31,7 +31,7 @@ export const BuilderFields = ({
 
   const { handleBuilderClick } = useCreatePageSidebar()
 
-  const activeFieldNumber =
+  const activeFieldId =
     stateData.state === FieldBuilderState.EditingField
       ? stateData.field._id
       : stateData.state === FieldBuilderState.CreatingField
@@ -50,9 +50,9 @@ export const BuilderFields = ({
           key={f._id}
           field={f}
           isHiddenByLogic={!visibleFieldIds.has(f._id)}
-          isDraggingOver={f._id === activeFieldNumber ? isDraggingOver : false}
+          isDraggingOver={f._id === activeFieldId ? isDraggingOver : false}
           fieldBuilderState={
-            f._id === activeFieldNumber ? stateData.state : undefined
+            f._id === activeFieldId ? stateData.state : undefined
           }
           handleBuilderClick={handleBuilderClick}
           isDirty={isDirty}
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
@@ -50,7 +50,7 @@ export const BuilderFields = ({
           key={f._id}
           field={f}
           isHiddenByLogic={!visibleFieldIds.has(f._id)}
-          isDraggingOver={isDraggingOver}
+          isDraggingOver={f._id === activeFieldNumber ? isDraggingOver : false}
           fieldBuilderState={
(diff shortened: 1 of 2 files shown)


foochifa:
Changes are current on staging-alt can use this form [use-template](https://staging-alt.form.gov.sg/admin/form/645cd962fedd3d00111f48c5/use-template) to test (300 fields)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
```diff
@@ -41,17 +50,13 @@ export const BuilderFields = ({
...
           field={f}
           isHiddenByLogic={!visibleFieldIds.has(f._id)}
-          isDraggingOver={isDraggingOver}
-          isActive={
-            stateData.state === FieldBuilderState.EditingField
-              ? f._id === stateData.field._id
-              : stateData.state === FieldBuilderState.CreatingField
-              ? f._id === PENDING_CREATE_FIELD_ID
-              : false
+          isDraggingOver={f._id === activeFieldNumber ? isDraggingOver : false}
```
timotheeg: This reads a little funny now. I can see we still have to pass the value for the active row, but it's funny to hardcode it as `false` otherwise, so I still wonder if we could do without passing `isDraggingOver` entirely 🤔 

You have a comment below 
> // Field only needs to know fieldBuilderState if it is active, else it is agnostic to state

And I wonder whether we can make this readably explicit in code. The props definition for `FieldRow` could define `isDraggingOver` and `fieldBuilderState` as optional, and the caller here could make it explicit:

```typescript
{fieldsWithQuestionNos.map((f, i) => (
    const activeFieldExtraProps = f._id === activeFieldId ? {
        isDraggingOver, fieldBuilderState: stateData.state
    } : {}
    
    return <FieldRow
        // all common props
        {...activeFieldExtraProps}
    />
)}
```

Or better yet, I suppose we could have an explicit type `ActiveFieldRow` which defines its props strictly, but that might make the code here more convoluted. 

In any case, this is readability concerns, which is a little subjective. I let you think about it and decide!


Other than than that, if we keep the code as is, a very small (extremely small!) nit on syntax can be:
```suggestion
          isDraggingOver={f._id === activeFieldNumber && isDraggingOver}
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -347,52 +296,14 @@ const FieldRowContainer = ({
...
-                        icon={<BiTrash fontSize="1.25rem" />}
-                        onClick={handleDeleteClick}
-                        isLoading={deleteFieldMutation.isLoading}
-                        isDisabled={isAnyMutationLoading}
-                      />
-                    </Tooltip>
-                  </ButtonGroup>
-                </Flex>
+                {isActive && (
+                  <FieldButtonGroup
```
timotheeg, line 300: Just checking: does creating a new wrapper component here by itself helps with performance? Is it because the props it depends are now explicitly set to be these 4?

I ask because while the refactor looks good, it's hard to review the code diff of the setup code and  the markup when the whole chunk is taken out and pasted below.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
```diff
@@ -29,9 +30,17 @@ export const BuilderFields = ({
   const stateData = useFieldBuilderStore(stateDataSelector)
 
   const { handleBuilderClick } = useCreatePageSidebar()
-  const {
-    deleteFieldModalDisclosure: { onOpen: onDeleteModalOpen },
-  } = useBuilderAndDesignContext()
+
+  const activeFieldNumber =
```
timotheeg: We probably want to change the variable name here to not get confused with the question number
```suggestion
  const activeFieldID =
```

timotheeg (commented):
Thanks a lot for that @foochifa !! 💪

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
foochifa: yes! because various hooks such as `useCreateTabForm` `useDuplicateFormField` `useDeleteFormField` can now be called only in the new `FieldButtonGroup` component. Thus when these hooks changes, it will only cause re-render in the `FieldButtonGroup`

image

Before:
(Hooks causing re-renders for all field rows)
image

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
foochifa: Oh yes thanks Tim! [c991e3d](https://github.com/opengovsg/FormSG/pull/6303/commits/c991e3dbf4daf3915bb7fd49df8a5c27ce41f2fd)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
timotheeg: Thanks for the explanation @foochifa 🙏 🌮

timotheeg (approved):
LGTM! The readability stuff I was mentioning are minor, we can push this first and get the perf improvements sooner rather than later.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderFields.tsx
foochifa: Yep sounds good thanks Tim! [3d859f8](https://github.com/opengovsg/FormSG/pull/6303/commits/3d859f81e5eda626938e65b6c956c61d8e340fda)

timotheeg (approved):
LGTM!

(assuming that having `isDraggingOver:undefined` and `fieldBuilderState:undefined` for the non-active fields does not cause problem down the nested elements)

PR #6305, opened 2023-05-12
fix: squished logic block either option

### LINKED ISSUE #5508: logic block's equals to/either option is squished when the selected option has a long name

LinHuiqing:
(lifted from @dexxtan's PR https://github.com/opengovsg/FormSG/pull/5830, moved here for CI checks before we can merge it in)

## Problem

Logic block's equals to/either option is squished when the selected option has a long name.

Closes #5508 

## Solution

Flex basis which should be used to size the width of the rendered multi select input is not respected because there is a max width set on the multi select combo box input beside it. This results in the squished component.

The solution is to:
- convert max width into flex basis on the multi select combo box container
- rely on current set flex grow and flex shrink settings on the multi select combo box container to size that input
- not let the dropdown icon overflow to the next line due to insufficient width available
- and also apply min width 0 on both inputs for desktop screen sizes to let the implicit flex basis auto of the multi select input on the left size the input correctly
- also add a useEffect hook to set min width for the multi select input in case of other types of inputs needing this. It can be simplified from a switch to a simple if else since there is only one condition now

**Breaking Changes** 

    - Details ...

## Before & After Screenshots

**BEFORE**:

From issue #5508:
issue

**AFTER**:

See issue fixed:
Screenshot 2023-02-23 at 12 54 41 AM

## Tests

- [ ] Create a form with a Radio field.
- [ ] In this field, create an option with a very long name.
- [ ] Go to the logic option and use this field in the condition.
- [ ] Select "either" as the option for condition and select the long option as the value. The equals to / either option dropdown should not be squished after the long option has been selected.
- [ ] Repeat the above for a Dropdown field.

### DIFF
--- a/frontend/src/components/Dropdown/components/MultiSelectCombobox/MultiSelectCombobox.tsx
+++ b/frontend/src/components/Dropdown/components/MultiSelectCombobox/MultiSelectCombobox.tsx
@@ -18,7 +18,8 @@ const MultiItemsContainer: FC = ({ children }) => {
       // Margin difference for selected items.
       my="-3px"
       // Padding for dropdown toggle.
-      maxW="calc(100% - 2.5rem)"
+      flexBasis="calc(100% - 2.5rem)"
+      minW="0px"
     >
       {children}
     </Box>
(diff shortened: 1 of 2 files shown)


foochifa:
I think we should still leave the commits authored by @dexxtan in #5830. So perhaps we could first pull his local repo into a new branch then make changes from there? Before cutting a PR from said branch. This should help reflect his work in this PR too.

LinHuiqing:
> I think we should still leave the commits authored by @dexxtan in https://github.com/opengovsg/FormSG/pull/5830. So perhaps we could first pull his local repo into a new branch then make changes from there? Before cutting a PR from said branch. This should help reflect his work in this PR too.

@foochifa done!

timotheeg (approved):
LGTM

PR #6325, opened 2023-05-15
feat: add sample-submission endpoint to retrieve sample submission data of a public form

wanlingt:
## Problem

Some integrations e.g. Zapier, Plumber, require sample form submission data to test triggers. Providing sample data is also more convenient for admins as they would not have to make a (fake) form submission just for the sake of testing the integration.

## Solution

This PR creates a new endpoint `/api/v3/forms/:formId/sample-submission` which returns a sample form submission data.

**Breaking Changes** 

## Before & After Screenshots
**AFTER**:
image

## Tests

- Unit tests have been added
- [ ] Send a `GET` request to `http://staging.form.gov.sg/api/v3/forms/:formId/sample-submission` using
  - [ ] a public form. You should get a 200 and a sample form response, based on the fields in the form
  - [ ] a form ID that does not exist. You should get a 404
  - [ ] a private form. You should get a 404
  - [ ] an archived form. You should get a 410


src/app/modules/form/form.service.ts
```diff
@@ -343,3 +344,59 @@ export const retrievePublicFormsWithSmsVerification = (
...
+  let sampleValue = null
+  let noOfOptions = 0
+  let randomSelectedOption = 0
+  switch (field.fieldType) {
+    case 'textarea':
+    case 'textfield':
+      sampleValue =
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.'
+      break
+    case 'radiobutton':
```
foochifa, line 359: Just a thought, but are we going to check for any of the validation stuffs in each field? 

for instance domain restriction in email, text min/max amount etc?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -326,6 +326,60 @@ export const handleGetPublicForm: ControllerHandler<
...
+        message: error.message,
+        // Flag to prevent default 404 subtext ("please check link") from
+        // showing.
+        isPageFound: true,
+        formTitle: error.formTitle,
+      })
+    }
+
+    return res.status(statusCode).json({ message: errorMessage })
+  }
```
timotheeg, line 359: /notRealtedToPRPerSay (no action needed; just noting)
We should probably turn `getFormIfPublic` into a middleware at some point, so this code here can be shared across all endpoints which are dependent on having a public form

src/app/modules/form/form.service.ts
```diff
@@ -343,3 +344,59 @@ export const retrievePublicFormsWithSmsVerification = (
...
+      break
+    case 'yes_no':
+      sampleValue = 'yes'
+      break
+    case 'rating':
+      sampleValue = 1
+      break
+    case 'attachment':
+      sampleValue = 'attachmentFileName'
+      break
```
timotheeg, line 388: /justNoting: This won't be enough to test actual attachments, but we can revisit that later!

src/app/modules/form/form.service.ts
```diff
@@ -343,3 +344,59 @@ export const retrievePublicFormsWithSmsVerification = (
...
+    default:
+      break
+  }
+  if (sampleValue != null) {
+    sampleData[field._id] = {
+      question: field.title,
```
timotheeg: Isn't the field id supposed to be part of the object as well in responses?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -326,6 +326,60 @@ export const handleGetPublicForm: ControllerHandler<
...
+  const publicForm = form.getPublicView() as PublicFormDto
+
+  const sampleData: Record<string, any> = {}
+  const formFields = publicForm.form_fields
+  if (!formFields) {
+    throw new Error('unable to get form fields')
+  }
+  for (const field of formFields) {
+    FormService.createSampleSubmissionData(sampleData, field)
+  }
```
timotheeg: /nit: For code readability/cleanliness, it's better to write this as a [pure function](https://en.wikipedia.org/wiki/Pure_function) that return a sample response for a given field `createSampleSubmissionAnswer(field)`, and do the mutation of the result object here (as opposed to passing the result object as first arg and mutating it with every function call).

```suggestion
  const sampleData: Record<string, any> = formFields.reduce((acc, field) => {
    const answer = FormService.createSampleSubmissionAnswer(field)
    if (!answer) return acc
    acc[field._id] = answer
    return acc
  }, {})
```

If you'd like to reduce the code in the controller, create an additional method `FormService.createSampleSubmissionResponses()`, such that the route handler can just invoke it:

```typescript
  const publicForm = form.getPublicView() as PublicFormDto
  const formFields = publicForm.form_fields

  if (!formFields) {
    throw new Error('unable to get form fields')
  }
  
  const responses = FormService.createSampleSubmissionResponses(publicForm.form_fields)
  
  return res.json({ responses })
```

And inside `FormService.createSampleSubmissionResponses`, you can call `FormService.createSampleSubmissionAnswer`.

src/app/modules/form/form.service.ts
LinHuiqing: that's true^ good point @foochifa, their processes might break if we just put in hard-coded values that are not within the values they want to set

maybe we can separate this into a separate issue and do it in a separate PR though? if not the scope of this PR might creep too much

wanlingt: yes good point @foochifa ! agree with @LinHuiqing , I'll create a separate issue and PR for it

src/app/modules/form/form.service.ts
wanlingt: (as discussed) the attachment file is added in `formsg-zapier-integration`'s code, as attachments show up as a specific format in Zapier

src/app/modules/form/form.service.ts
wanlingt: Added in [3f2dcc1](https://github.com/opengovsg/FormSG/pull/6325/commits/3f2dcc19ac6a66253f05f172c83cb3dd2d70a458)

src/app/modules/form/public-form/public-form.controller.ts
wanlingt: Thanks Tim! Created an additional method in `FormService`

[3f2dcc1](https://github.com/opengovsg/FormSG/pull/6325/commits/3f2dcc19ac6a66253f05f172c83cb3dd2d70a458)

src/app/modules/form/form.service.ts
```diff
@@ -343,3 +345,70 @@ export const retrievePublicFormsWithSmsVerification = (
...
+  let answer = {}
+  if (sampleValue != null) {
+    answer = {
+      id: field._id,
+      question: field.title,
+      answer: sampleValue,
+      fieldType: field.fieldType,
+    }
+  }
+  return answer
```
timotheeg: It feels wrong to return an empty object if the value is null (empty objects are truthy in JS) 🤔, and `createSampleSubmissionResponses()` below only skips `null`/`undefined`. 

I'd advise to either return `null`, or to return an object which contain the static fields taken from the definition (`id`, `question`, `fieldType`), and leave `answer` to be `undefined` or `null`. In the later case, code can be simplified a little:

```suggestion
  return  {
    id: field._id,
    question: field.title,
    answer: sampleValue,
    fieldType: field.fieldType,
  }
```

wanlingt: went with the suggestion! [d982f83](https://github.com/opengovsg/FormSG/pull/6325/commits/d982f831f401f1651d252f02bd99e0e792e8d65d)

src/app/modules/form/form.service.ts
wanlingt: #6335 - new issue

LinHuiqing (approved):
lgtm

PR #6334, opened 2023-05-16
fix: disable payment resume and duplicate modals in preview

foochifa:
## Problem

Currently we show the `FormPaymentModal`, `FormPaymentResumeModal` and `DuplicatePaymentModal` in preview mode if the same browser or email is used in a public form submission.

This should be disabled when the admin is in preview mode

## Solution

Check if form is in preview mode before displaying the modals.

**Breaking Changes** 

**Improvements**:

- Shifted `isPreview` from a prop in `FormFieldsContainer` and `FormEndPageContainer` to a context in `usePublicFormContext`. As `isPreview` should be static, it can be inserted as a global context without much performance consideration (from hooks updates etc)

## Before & After Screenshots

**BEFORE**:

image
image

## Tests

To ensure functionality within payments
- [ ] go to a payment form
- [ ] open the form, make a payment intent
- [ ] go to preview mode, there should be no resume modal that is shown
- [ ] go back to the public form, the resume modal should be shown
- [ ] resume session, make the payment
- [ ] go to preview mode, attempt to make a preview payment with the same email as above
- [ ] No payment modal should be shown
- [ ] go to public form, attempt to make a payment with the same email
- [ ] duplicate payment modal should be shown
- [ ] go to public form, attempt to make a payment with a new email
- [ ] proceed to pay modal should be shown

To ensure no regression problem
- [ ] go to a normal storage mode form
- [ ] go to preview mode, submit the form
- [ ] go to public form, submit the form
- [ ] do the same for an email mode form

### DIFF
--- a/frontend/src/features/admin-form/preview/PreviewFormPage.tsx
+++ b/frontend/src/features/admin-form/preview/PreviewFormPage.tsx
@@ -34,8 +34,8 @@ export const PreviewFormPage = (): JSX.Element => {
           <FormStartPage />
           <PublicFormWrapper>
             <FormInstructions />
-            <FormFields isPreview />
-            <FormEndPage isPreview />
+            <FormFields />
+            <FormEndPage />
             <FormFooter />
           </PublicFormWrapper>
         </FormSectionsProvider>
--- a/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
+++ b/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
@@ -334,6 +334,7 @@ export const PreviewFormProvider = ({
         isLoading,
         handleLogout: undefined,
         isPaymentEnabled,
+        isPreview: true,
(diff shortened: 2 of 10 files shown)


LinHuiqing (changes requested):
After the first payment is attempted on preview mode, subsequent payments attempted on preview mode will still show the modal.

foochifa:
Thanks @LinHuiqing for catching that! Have confirmed with @staceytan1998 that all modals should not be shown in preview [69a7234](https://github.com/opengovsg/FormSG/pull/6334/commits/69a723471503bb0b87da808d9f0938050d9a1367)

timotheeg (approved):
LGTM. Thanks ChiFa!

LinHuiqing (approved):
lgtm and tested! as mentioned, will open a separate issue for the ever-present white box for payment preview even when payment is not enabled

PR #6337, opened 2023-05-16
feat: MyInfo over sgID backend

### LINKED ISSUE #6323: Add backend code to support more sgID scopes
Current sgID integration only queries for the default values (NRIC, DOB, name) and uses only the NRIC for authentication.

**Describe the solution you'd like**
We should prepare the backend to allow querying for more sgID OAuth scopes.

kenjin-work:
## Problem
Retrieving MyInfo data over sgID is currently not supported.

Closes #6323 

## Solution
- Extend sgID client to convert MyInfo fields to sgID OAuth scopes.
- Reuse pre-existing MyInfo infrastructure to process and store sgID's MyInfo data.
- For now, we preserve the current behavior for sgID authentication mode (just fill up NRIC, without filling other fields.)
- This PR is mostly a refactor to split sgID into two parts: Authentication, and MyInfo.
- I will add tests in a follow up PR for the sgID MyInfo parts.

**Breaking Changes** 

**Features**:

- The full publicly available OAuth scopes supported by sgID are now supported. Enabling support of this in the frontend is only changing 10 lines of code in the backend to expose this feature.

## Before & After Screenshots

(We want no difference, since it's disabled on the frontend).

**BEFORE**:
image

**AFTER**:
image

## Tests
I've added tests to the sgID adapter to make sure it works. It should automatically be run as part of the existing test suite.


src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -403,6 +405,7 @@ export const _handleFormAuthRedirect: ControllerHandler<
             return SgidService.createRedirectUrl(
               formId,
               Boolean(isPersistentLogin),
+              [],
```
kenjin-work, line 464: Note to reviewers: This will be where the MyInfo fields are passed in when we enable this in the frontend. Blank for now as we don't want to enable this feature yet.

src/app/modules/sgid/sgid.service.ts
```diff
@@ -195,11 +201,11 @@ export class SgidServiceClass {
...
    */
   createJwt(
-    data: { 'myinfo.nric_number': string },
+    data: SGIDScopeToValue,
     rememberMe: boolean,
   ): Result<{ jwt: string; maxAge: number }, ApplicationError> {
-    const userName = data['myinfo.nric_number']
-    const payload = { userName, rememberMe }
+    const userName = data[SGID_MYINFO_NRIC_NUMBER_SCOPE]
+    const payload = { userName, data, rememberMe }
```
kenjin-work: Woops. I just realised I shouldn't even be storing this data in the JWT. Back to the drawing board!

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -321,6 +325,56 @@ export const handleGetPublicForm: ControllerHandler<
...
+    case FormAuthType.SGID_MyInfo: {
+      const accessTokenCookie = req.cookies[SGID_COOKIE_NAME]
+      if (!accessTokenCookie) {
+        return res.json({
+          form: publicForm,
+          isIntranetUser,
+        })
+      }
+      res.clearCookie(SGID_COOKIE_NAME)
+      res.clearCookie(MYINFO_LOGIN_COOKIE_NAME)
```
tshuli: If we are having separate auth type for `SGID_MyInfo`:

```
this.authType !== FormAuthType.MyInfo &&
      this.authType !== FormAuthType.SGID_MyInfo
```

why do we clear cookie for `SGID` and `MYINFO_LOGIN`; will there be a case where user has valid ongoing sessions for each of `SGID`, `MYINFO` and `SGID_MYINFO`?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -321,6 +325,56 @@ export const handleGetPublicForm: ControllerHandler<
...
+      return SgidService.extractSgidJwtMyInfoPayload(accessTokenCookie)
+        .asyncAndThen((auth) =>
+          SgidService.retrieveUserInfo({ accessToken: auth.accessToken }),
+        )
+        .andThen((userInfo) => {
+          const data = new SGIDMyInfoData(userInfo.data)
+          return MyInfoService.prefillAndSaveMyInfoFields(
```
tshuli, line 348: If we reuse the same MyInfoService to prefill fields from sgid myinfo, are we guaranteed that the types of field supported in sgid myinfo is a subset of singpass myinfo (if no, will this be an issue)?

src/app/modules/verification/verification.controller.ts
```diff
@@ -263,8 +263,12 @@ export const _handleGenerateOtp: ControllerHandler<
                 return error
               })
           }
+          // SGID_MyInfo temporarily here
+          case FormAuthType.SGID_MyInfo:
           case FormAuthType.SGID:
-            return SgidService.extractSgidJwtPayload(req.cookies.jwtSgid)
+            return SgidService.extractSgidSingpassJwtPayload(
+              req.cookies.jwtSgid,
```
tshuli: same here - should we have different cookies for sgid and sgid_myinfo?

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -403,6 +457,16 @@ export const _handleFormAuthRedirect: ControllerHandler<
...
               Boolean(isPersistentLogin),
+              [],
+              encodedQuery,
+            )
+          })
+        case FormAuthType.SGID_MyInfo:
+          return validateSgidForm(form).andThen(() => {
+            return SgidService.createRedirectUrl(
+              formId,
+              Boolean(isPersistentLogin),
```
tshuli: ref our discussion yesterday, are we still able to support isPersistentLogin if we are now storing only the access token in jwt (vs the userinfo), and exchanging for userInfo upon loading the form?

tshuli (commented):
nice work, some clarifying questions!

src/app/modules/form/public-form/public-form.controller.ts
kenjin-work: The cookie management is kind of confusing. I should probably make a third cookie type. Here's what each cookie is for:
1. `MYINFO_LOGIN_COOKIE_NAME` - contains singpass username (NRIC).
2. `SGID_COOKIE_NAME` - dual use. When using SGID Auth mode, contains the singpass username (NRIC), when using SGID MyInfo mode, is only a transient cookie holding the sgID access token.

When SGID MyInfo mode is used, we need to also set the MyInfo cookie as we need somewhere to store the singpass username, and also because we reuse MyInfo functions, which read and depend on that cookie.

I should probably split up the `SGID_COOKIE_NAME` into two SGID cookies for each SGID usage so that it's clearer, rather than reusing the same one. What do you think?

src/app/modules/form/public-form/public-form.controller.ts
kenjin-work: Good catch. Right now this depends on the frontend UI choices that we make. We're thinking of using the same form fields for MyInfo for sgID-MyInfo, so sgid myinfo will indeed be a strict subset of singpass myinfo.

src/app/modules/form/public-form/public-form.controller.ts
kenjin-work: Yup! This is because the cookie is dual-use, ref https://github.com/opengovsg/FormSG/pull/6337#discussion_r1197378671.

In sgID auth-only mode, there is no change in behavior.

src/app/modules/form/public-form/public-form.controller.ts
tshuli: Thanks for reply! I agree splitting `SGID_COOKIE_NAME` makes sense. Dual use cookie may not be ideal because `FormAuthType.SGID` sessions should be handled separately from `FormAuthType.SGID_MyInfo` (dual use implies starting one use case override the other use case)

src/app/modules/form/public-form/public-form.controller.ts
tshuli: Understand, thanks!
Off hand, I don't see any issue with one single service handling the prefill/saving of both MyInfo and sgid-Myinfo fields, but can we build it in a way that even if MyInfo and sgid-Myinfo are not subsets of another another, this still works (if not already the case)?

Then in future, if there are new fields introduced in sgid-Myinfo which are not in Myinfo, we only need to update the frontend UI to accomodate such fields

src/app/modules/form/public-form/public-form.controller.ts
tshuli: ^ as above. with dual-use cookie, will the persistent login be cleared if user then logs into a sgid-myinfo form (as the dual use cookie will be overwritten to store sgID access token)

src/app/modules/form/public-form/public-form.controller.ts
kenjin-work: Thanks. I didn't think about that. That's a great observation!

kenjin-work:
Thanks for the extremely thorough and insightful review Shu Li. I tried to address it. Now, submitting SGID-MyInfo forms don't log out an already logged in SGID-Auth form :).

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -46,6 +46,32 @@ export const handleLogin: ControllerHandler<
   }
 
   const form = formResult.value
+  // @TODO move this to separate route.
```
tshuli: If this is TODO for subsequent PR, can open an issue and then label here, e.g. // TODO(#XXXX): Move this to separate route

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -46,6 +46,32 @@ export const handleLogin: ControllerHandler<
...
+    res.cookie(SGID_MYINFO_COOKIE_NAME, jwt, {
+      maxAge,
+      httpOnly: true,
+      sameSite: 'lax', // Setting to 'strict' prevents Singpass login on Safari, Firefox
+      secure: !config.isDev,
+      ...SgidService.getCookieSettings(),
+    })
+    return res.redirect(target)
+  }
   if (form.authType !== FormAuthType.SGID) {
```
tshuli: Probably should shift this before L50(`if (form.authType === FormAuthType.SGID_MyInfo) {`), so that if form type is neither SGID or SGID_MyInfo, it is logged appropriately and `isLoginError` set

The handling for the two cases `FormAuthType.SGID_MyInfo` and `FormAuthType.SGID` then follows thereafter

src/app/modules/sgid/sgid.service.ts
```diff
@@ -212,17 +224,79 @@ export class SgidServiceClass {
   }
 
   /**
-   * Verifies a sgID JWT and extracts its payload.
+   * Create a JWT based with access token from sgID
```
tshuli: Would be helpful to add a short liner about how this is different from createSgidSingpassJwt, for future ref

E.g. something like
* Create a JWT based with access token from sgID
* This access token is then used to exchange for myinfo scopes in the public form controller
* Unlike `createSgidSingpassJwt` where the access token is exchanged for userInfo upfront and the userInfo is stored in the jwt

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -61,7 +87,12 @@ export const handleLogin: ControllerHandler<
 
   const jwtResult = await SgidService.retrieveAccessToken(code)
     .andThen((data) => SgidService.retrieveUserInfo(data))
-    .andThen(({ data }) => SgidService.createJwt(data, rememberMe))
+    .andThen(({ data }) =>
+      SgidService.createSgidSingpassJwt(
+        { 'myinfo.nric_number': data['myinfo.nric_number'] },
```
tshuli: Would be good opportunity to switch to using `SGID_MYINFO_NRIC_NUMBER_SCOPE` here instead of magic string :)

src/app/modules/sgid/sgid.service.ts
```diff
@@ -212,17 +224,79 @@ export class SgidServiceClass {
...
+   * @param accessToken - sgID access token
+   * @param rememberMe - determines how long the JWT is valid for
+   */
+  createSgidMyInfoJwt(
+    accessToken: string,
+    rememberMe: boolean,
```
tshuli: for SgidMyInfoJwt, there should not be rememberMe (?)

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -321,6 +328,56 @@ export const handleGetPublicForm: ControllerHandler<
...
+    case FormAuthType.SGID_MyInfo: {
+      const accessTokenCookie = req.cookies[SGID_MYINFO_COOKIE_NAME]
+      if (!accessTokenCookie) {
+        return res.json({
+          form: publicForm,
+          isIntranetUser,
+        })
+      }
+      res.clearCookie(SGID_MYINFO_COOKIE_NAME)
+      res.clearCookie(MYINFO_LOGIN_COOKIE_NAME)
```
tshuli: After the revised impl (which separates the sgid cookie into two - sgid and sgid_myinfo), are we still using myinfo_login_cookie for sgid_myinfo?

src/app/modules/sgid/sgid.service.ts
```diff
@@ -167,14 +177,16 @@ export class SgidServiceClass {
   }: {
     accessToken: string
   }): ResultAsync<
-    { sub: string; data: { 'myinfo.nric_number': string } },
+    { sub: string; data: SGIDScopeToValue },
```
tshuli, line 181: To update the doc for this method also since retrieveUserInfo is no longer just for NRIC

src/app/modules/form/public-form/public-form.service.ts
```diff
@@ -73,9 +73,11 @@ export const getCookieNameByAuthType = (
     | FormAuthType.SP
     | FormAuthType.CP
     | FormAuthType.MyInfo
-    | FormAuthType.SGID,
+    | FormAuthType.SGID
+    | FormAuthType.SGID_MyInfo,
 ): string => {
   switch (authType) {
+    case FormAuthType.SGID_MyInfo:
```
tshuli, line 83: should this now be `SGID_MYINFO_COOKIE_NAME`

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -321,6 +328,56 @@ export const handleGetPublicForm: ControllerHandler<
...
+        .andThen((userInfo) => {
+          const data = new SGIDMyInfoData(userInfo.data)
+          return MyInfoService.prefillAndSaveMyInfoFields(
+            form._id,
+            data,
+            form.toJSON().form_fields,
+          ).map((prefilledFields) => {
+            return res
+              .cookie(
+                MYINFO_LOGIN_COOKIE_NAME,
```
tshuli: Qn: why use the myinfo_login_cookie instead of sgid_myinfo_cookie?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -259,8 +260,8 @@ const submitEncryptModeForm: ControllerHandler<
       break
     }
     case FormAuthType.SGID: {
```
tshuli, line 264: Would be helpful to also add one more case where

`FormAuthType.SGID_MyInfo` and return
`Storage mode form is not allowed to have SGID MyInfo authorisation`

tshuli (commented):
Nice work and thanks for the quick turnaround!
Went through again in more detail and have some further comments for consideration :)

src/app/modules/form/public-form/public-form.controller.ts
kenjin-work: Yup we still do! This allows us to reuse code from MyInfo.

When SGID MyInfo mode is used, we need to also set the MyInfo cookie as we need somewhere to store the singpass username, and also because we reuse MyInfo functions, which read and depend on that cookie.

src/app/modules/form/public-form/public-form.service.ts
kenjin-work: This is the cookie that determines whether the user is logged in/out. In this case, it's correct to clear the MyInfo login cookie as that contains the login username.

src/app/modules/form/public-form/public-form.controller.ts
kenjin-work: SGID_MYINFO_COOKIE - stores access token for sgID userinfo query.
MYINFO_LOGIN_COOKIE - stores singpass username and is required for MyInfo field's pre-existing functions to work properly.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
kenjin-work: Ah completely overlooked this part. Thanks!

src/app/modules/sgid/sgid.controller.ts
kenjin-work: I think I'll remove this TODO. I don't think it should be in a separate route. They are logically a little different but most of the code is shared.

src/app/modules/form/public-form/public-form.service.ts
tshuli: Thanks! Should this now be `SGID_MYINFO_LOGIN_COOKIE_NAME`, since that is where you store the login username for SGID myinfo?

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -58,10 +66,40 @@ export const handleLogin: ControllerHandler<
...
+  }
 
   const jwtResult = await SgidService.retrieveAccessToken(code)
     .andThen((data) => SgidService.retrieveUserInfo(data))
-    .andThen(({ data }) => SgidService.createJwt(data, rememberMe))
+    .andThen(({ data }) =>
+      SgidService.createSgidSingpassJwt(
+        { 'myinfo.nric_number': data[SGID_MYINFO_NRIC_NUMBER_SCOPE] },
+        rememberMe,
+      ),
```
tshuli: Would this be neater instead of hardcoding
```
SgidService.createSgidSingpassJwt(
        { [SGID_MYINFO_NRIC_NUMBER_SCOPE]: data[SGID_MYINFO_NRIC_NUMBER_SCOPE] },
        rememberMe,
      ),
```

tshuli:
Couple of minor comments, but lgtm otherwise!

tshuli:
@timotheeg would be good if you could have another look through before this is merged in

src/app/modules/sgid/sgid.controller.ts
kenjin-work: Thanks! I completely forgot that's how you use variables as keys in JS!

src/app/modules/sgid/__tests__/sgid.service.spec.ts
```diff
@@ -68,14 +73,36 @@ describe('sgid.service', () => {
...
       expect(url._unsafeUnwrap()).toEqual(MOCK_REDIRECT_URL)
       expect(sgidClient.authorizationUrl).toHaveBeenCalledWith(
         MOCK_STATE,
-        SGID_SCOPES,
+        SGID_SINGPASS_SCOPE,
         null,
       )
     })
+
+    it('should return extract additional OAuth scopes from MyInfo', () => {
```
timotheeg: The English sounds funny? 😅 Did you mean:

```suggestion
    it('should extract additional OAuth scopes from MyInfo', () => {
```

src/app/modules/myinfo/myinfo.adapter.ts
```diff
@@ -155,7 +158,9 @@ export const internalAttrListToScopes = (
  * extract the correct data by translating internal FormSG attributes
  * to the correct keys in the data.
  */
-export class MyInfoData {
+export class MyInfoData
+  implements SGIDDataTransformer<ExternalAttr, InternalAttr>
```
timotheeg: /nit: Introducing a shared parent type between `MyInfoData` and `SGIDMyInfoData` is great, but it feels a little funny that `MyInfoData` implements `SGIDDataTransformer` 🤔 . Shouldn't the parent class be called `MyInfoDataTransformer` instead?

src/app/modules/sgid/sgid.adapter.ts
```diff
@@ -0,0 +1,134 @@
...
+      // out by other functions.
+      return ExternalAttr.NricFin
+  }
+}
+
+export const internalAttrListToScopes = (attrs: InternalAttr[]): string =>
+  // Always ask for NRIC
+  `openid ${ExternalAttr.NricFin} ${attrs
+    .map(internalAttrToScope)
+    .join(' ')}`.trim()
```
timotheeg: /nit: the `.trim()` feels superfluous. Since this is all config-time data, I'd hope there is no unexpected leading or trailing spaces 🤔 . If there are, `trim()` will only clean the ends (lead/tail), not the potentially double-spaced separators. If we are concerned for unexpected whitespace, an implementation like below could be safer to produce a valid final scope:
```typescript
  [
    'openid',
    ExternalAttr.NricFin,
    ...attrs.map(internalAttrToScope)
  ]
  .map(v => v.trim()) // is this needed?
  .join(' ')
```

Maybe more important comment: the definition of `internalAttrToScope()` suggests there could be duplicate scopes `ExternalAttr.NricFin` after the mapping of `attrs`. So to get a normalized scope, we should clean up the list to remove duplicates:
```typescript
  [
    'openid',
    ExternalAttr.NricFin,
    ...attrs.map(internalAttrToScope).filter(v => v !== ExternalAttr.NricFin)
  ]
  .map(v => v.trim()) // is this needed?
  .join(' ')
```

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -58,10 +66,42 @@ export const handleLogin: ControllerHandler<
...
+      (data) => SgidService.createSgidMyInfoJwt(data.accessToken),
+    )
+
+    if (jwtResult.isErr()) {
+      logger.error({
+        message: 'Error while handling login via sgID (MyInfo)',
+        meta,
+        error: jwtResult.error,
+      })
+      res.cookie('isLoginError', true)
```
timotheeg, line 80: /notYourPR-DontChangeAnything-JustNoting: putting an error flag in a client cookie feels a bit dirty 😅 ... Ideally, we should have sessions for respondent, and the error details should be set and retrieved from that. The client should not be able to drive error handling behaviour from cookies

timotheeg (commented):
Looking good KenJin! 🥳 

I only have some super very comments, so I think it's pretty much ready to go! 💪 

And I will try to do a second pass tomorrow morning to make sure I understand correctly all the typing magic with the generics (I'm not very good at reading generics 😅)

src/app/modules/sgid/sgid.adapter.ts
timotheeg: Oh oops, sorry that snippet I posted doesn't remove duplicates overall 😅 , it only caters to `NricFin` that we add ourselves. Using a set instead does the job overall:
```typescript
  [
    'openid',
    ...(new Set([ExternalAttr.NricFin, ...attrs.map(internalAttrToScope)])
  ]
  .map(v => v.trim()) // is this needed?
  .join(' ')

src/app/modules/sgid/sgid.controller.ts
kenjin-work: Yeah unfortunately that's how the existing code does this :(.

kenjin-work:
Thanks for the thorough review Tim!

timotheeg (approved):
LGTM

PR #6348, opened 2023-05-18
feat: direct admins to payments settings page

LinHuiqing:
## Problem

The current way the settings page is routed does not allow us to direct users to specific tabs. This will be disorienting for users, especially for Payments when they are navigated away then redirected back to us.

Closes FRM-753

## Solution

Set up routing for the different tabs on the Settings page, then update payment-related settings links to the payments tab in the settings page directly (previously we could only direct them to the generic settings page).

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

https://github.com/opengovsg/FormSG/assets/37061143/43ec4f06-e5ae-4505-a87e-4c22be5ace41

**AFTER**:

https://github.com/opengovsg/FormSG/assets/37061143/4bedff81-cde0-42eb-bb6d-8f7fb42d0ddc

## Tests

- [ ] Check that opening specific settings tabs through URL path is possible:
   - [ ] Go to `/admin/form/:formId/settings`. You should see the generic settings page with the general tab open.
   - [ ] Go to `/admin/form/:formId/settings/general`. You should see the generic settings page with the general tab open.
   - [ ] Go to `/admin/form/:formId/settings/singpass`. You should see the settings page with the singpass tab open.
   - [ ] Go to `/admin/form/:formId/settings/twilio`. You should see the settings page with the twilio tab open.
   - [ ] Go to `/admin/form/:formId/settings/webhooks`. You should see the settings page with the webhooks tab open.
   - [ ] Go to `/admin/form/:formId/settings/payments`. You should see the settings page with the payments tab open.
   - [ ] Go to `/admin/form/:formId/settings/<invalid_payment_tab_str>`. You should see the generic settings page with the general tab open.
- [ ] On a storage-mode form that is not connected to stripe, go to the payment input panel. 
   - [ ] There should be an infobox with the copy "Connect your Stripe account in Settings to add payment field.". 
   - [ ] "Settings" in the copy should be a link. 
   - [ ] Clicking on the Settings link should lead the admin to the payment settings page.
- [ ] Try to connect a stripe account to the form.
   - [ ] Admin should be redirected to payment settings page after stripe has been connected.

### DIFF
--- a/frontend/src/app/AppRouter.tsx
+++ b/frontend/src/app/AppRouter.tsx
@@ -112,9 +112,11 @@ export const AppRouter = (): JSX.Element => {
         >
           <Route index element={<CreatePage />} />
           <Route
-            path={ADMINFORM_SETTINGS_SUBROUTE}
+            path={`${ADMINFORM_SETTINGS_SUBROUTE}`}
             element={<SettingsPage />}
-          />
+          >
+            <Route path={':settingsTab'} element={<SettingsPage />} />
+          </Route>
           <Route
             path={ADMINFORM_RESULTS_SUBROUTE}
             element={<FormResultsLayout />}
--- a/frontend/src/app/AppRouter.tsx
+++ b/frontend/src/app/AppRouter.tsx
@@ -111,10 +111,7 @@ export const AppRouter = (): JSX.Element => {
           element={<PrivateElement element={<AdminFormLayout />} />}
         >
           <Route index element={<CreatePage />} />
-          <Route
-            path={`${ADMINFORM_SETTINGS_SUBROUTE}`}
-            element={<SettingsPage />}
-          >
+          <Route path={ADMINFORM_SETTINGS_SUBROUTE} element={<SettingsPage />}>
             <Route path={':settingsTab'} element={<SettingsPage />} />
           </Route>
           <Route
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
@@ -6,8 +6,9 @@ import {
   useForm,
   useWatch,
 } from 'react-hook-form'
+import { Link as ReactLink } from 'react-router-dom'
 import { useDebounce } from 'react-use'
-import { Box, FormControl, Text } from '@chakra-ui/react'
+import { Box, FormControl, Link, Text } from '@chakra-ui/react'
 import { cloneDeep } from 'lodash'
 
 import {
@@ -288,11 +289,17 @@ export const PaymentsInputPanel = (): JSX.Element | null => {
     }
   }, [paymentsField, resetData, setData, setToEditingPayment, setToInactive])
 
-  const paymentDisabledMessage = !isEncryptMode
-    ? 'Payments are not available in email mode forms.'
-    : !isStripeConnected
-    ? 'Connect your Stripe account in Settings to save this field.'
-    : ''
+  const paymentDisabledMessage = !isEncryptMode ? (
+    <Text>Payments are not available in email mode forms.</Text>
+  ) : !isStripeConnected ? (
+    <Text>
+      Connect your Stripe account in{' '}
+      <Link as={ReactLink} to={`settings/payments`}>
+        Settings
+      </Link>{' '}
+      to add payment field.
+    </Text>
+  ) : null
 
   // payment eligibility will be dependent on whether paymentDisabledMessage is non empty
   const isPaymentDisabled = !!paymentDisabledMessage
@@ -303,9 +310,7 @@ export const PaymentsInputPanel = (): JSX.Element | null => {
     <>
       {isPaymentDisabled && (
         <Box px="1.5rem" pt="2rem" pb="1.5rem">
-          <InlineMessage variant="info">
-            <Text>{paymentDisabledMessage}</Text>
-          </InlineMessage>
+          <InlineMessage variant="info">{paymentDisabledMessage}</InlineMessage>
         </Box>
       )}
       <PaymentInput isDisabled={isPaymentDisabled} />
--- a/frontend/src/features/admin-form/settings/SettingsPage.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsPage.tsx
@@ -28,16 +28,8 @@ import { SettingsPaymentsPage } from './SettingsPaymentsPage'
 import { SettingsTwilioPage } from './SettingsTwilioPage'
 import { SettingsWebhooksPage } from './SettingsWebhooksPage'
 
-const settingsTabsOrder = [
-  'general',
-  'singpass',
-  'twilio',
-  'webhooks',
-  'payments',
-]
-
 export const SettingsPage = (): JSX.Element => {
-  const { formId, settingsTab = settingsTabsOrder[0] } = useParams()
+  const { formId, settingsTab } = useParams()
   const { user } = useUser()
   const { data: flags } = useFeatureFlags()
 
@@ -58,7 +50,15 @@ export const SettingsPage = (): JSX.Element => {
   const displayPayments =
     user?.betaFlags?.payment || flags?.has(featureFlags.payment)
 
-  const tabIndex = settingsTabsOrder.indexOf(settingsTab)
+  const settingsTabsOrder = [
+    'general',
+    'singpass',
+    'twilio',
+    'webhooks',
+    ...(displayPayments ? ['payments'] : []),
+  ]
+
+  const tabIndex = settingsTabsOrder.indexOf(settingsTab ?? '')
 
   return (
     <Box overflow="auto" flex={1}>
--- a/frontend/src/features/admin-form/settings/SettingsPage.tsx
+++ b/frontend/src/features/admin-form/settings/SettingsPage.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 import { BiCodeBlock, BiCog, BiDollar, BiKey, BiMessage } from 'react-icons/bi'
 import { useNavigate, useParams } from 'react-router-dom'
 import {
@@ -28,6 +28,8 @@ import { SettingsPaymentsPage } from './SettingsPaymentsPage'
 import { SettingsTwilioPage } from './SettingsTwilioPage'
 import { SettingsWebhooksPage } from './SettingsWebhooksPage'
 
+const settingsTabsOrder = ['general', 'singpass', 'twilio', 'webhooks']
+
 export const SettingsPage = (): JSX.Element => {
   const { formId, settingsTab } = useParams()
   const { user } = useUser()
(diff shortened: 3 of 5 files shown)


frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -59,6 +69,7 @@ export const SettingsPage = (): JSX.Element => {
         variant="line"
         py={{ base: '2.5rem', lg: '3.125rem' }}
         px={{ base: '1.5rem', md: '1.75rem', lg: '2rem' }}
+        defaultIndex={tabIndex === -1 ? 0 : tabIndex}
```
wanlingt: should we also account for the case where the displayPayments is false?

foochifa: I agree with @wanlingt here! Currently if displayPayments is false we will show a blank settings content, but should be better to just redirect back to general (consistent with params of non-existent tab)

frontend/src/app/AppRouter.tsx
```diff
@@ -111,10 +111,9 @@ export const AppRouter = (): JSX.Element => {
           element={<PrivateElement element={<AdminFormLayout />} />}
         >
           <Route index element={<CreatePage />} />
-          <Route
-            path={ADMINFORM_SETTINGS_SUBROUTE}
-            element={<SettingsPage />}
-          />
+          <Route path={ADMINFORM_SETTINGS_SUBROUTE} element={<SettingsPage />}>
+            <Route path={':settingsTab'} element={<SettingsPage />} />
```
foochifa, line 115: With this it means any `settings/:settingsTab` is a new path and has the same functionality + interaction as `settings/` right?

Just a slight concern cause technically after getting redirected back, such as from payments connect, when you toggle around the settings page, the `settings/payments` will still be reflected in the url

Was wondering if there are ways to just capture the param, maybe after computing tabIndex, navigate back to `settings/`?  (Or use search params instead?)

Minor thing though, not sure if its significant

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
```diff
@@ -288,11 +289,17 @@ export const PaymentsInputPanel = (): JSX.Element | null => {
...
-    ? 'Payments are not available in email mode forms.'
-    : !isStripeConnected
-    ? 'Connect your Stripe account in Settings to save this field.'
-    : ''
+  const paymentDisabledMessage = !isEncryptMode ? (
+    <Text>Payments are not available in email mode forms.</Text>
+  ) : !isStripeConnected ? (
+    <Text>
+      Connect your Stripe account in{' '}
+      <Link as={ReactLink} to={`settings/payments`}>
```
foochifa, line 297: Tangential to your PR, but with deep linking we can do this for MyInfo fields too! @staceytan1998 

image

frontend/src/app/AppRouter.tsx
LinHuiqing: hmmm okay true... might be useful to sort this out especially if we want to do metrics 🤔

frontend/src/features/admin-form/settings/SettingsPage.tsx
LinHuiqing: thanks both, true! and will fix haha

frontend/src/app/AppRouter.tsx
LinHuiqing: Wondering if its better to somehow bind the change in tabs to the path, or use query params as you suggested

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
LinHuiqing: Yea @staceytan1998 mentioned directing users to payments input panel directly too haha future work bah

frontend/src/app/AppRouter.tsx
foochifa: Yeah so either by binding so the `/:settingsTab` reflects the current tab selected, and any erroneous path will lead back to say `settings/general` 

Else if query params we can do like `settings?tab=payments` then after computing `tabIndex` we can do a 
```
const [searchParams, setSearchParams] = useSearchParams()
searchParams.delete('tab')
setSearchParams(searchParams)
```
I think should be able to modify the url string back to `/settings`?

frontend/src/features/admin-form/settings/SettingsPage.tsx
LinHuiqing: done in [19e2b8b](https://github.com/opengovsg/FormSG/pull/6348/commits/19e2b8bea00399ede3ea3dbf9874a751ccbff0f5)!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel.tsx
foochifa: Aite okay! But for this myinfo one right its just from create drawer to `settings/singpass` which should be doable already right? 😅 just add the link to settings or something in the info message, but yeah we can update all our linkings in another PR!

LinHuiqing: Ohhh you mean from infobox to settings! Yea okays yup it'll enable that haha

frontend/src/app/AppRouter.tsx
LinHuiqing: Decided to go with the path route so it'll be easier to do metrics if we want to do user tracking in the future. Bound the tab changes to the paths, but I can't bind the default path to `/general` on navigation because it takes some time before `displayPayments` can be determined... I think that's okay though, right?

LinHuiqing: Done in [b5f9761](https://github.com/opengovsg/FormSG/pull/6348/commits/b5f9761880ea389c261c41c67e1033fe39ef7498)!

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -50,6 +52,25 @@ export const SettingsPage = (): JSX.Element => {
   const displayPayments =
     user?.betaFlags?.payment || flags?.has(featureFlags.payment)
 
+  // Note: Admins are not redirected to /general on invalid settings tabs.
+  useEffect(() => {
+    if (displayPayments) {
```
foochifa, line 62: Hmm I feel like the way we are handling `payments` in `settingsTabOrder` is a bit convoluted. Looking ahead, once we bring payments into actual prod, this will need a bigger refactor too

Why not lets just keep it within `settingsTabOrder` and then check `displayPayments` when accessing `settingsTab === 'payments'`

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -50,6 +52,25 @@ export const SettingsPage = (): JSX.Element => {
...
+      settingsTabsOrder.push('payments')
+      setTabIndex(settingsTabsOrder.indexOf(settingsTab ?? ''))
+    }
+  }, [displayPayments, settingsTab])
+
+  const [tabIndex, setTabIndex] = useState(
```
foochifa: Not sure about the usage of setState here, I think can just const tabIndex, without the useEffect above, the setState isn't really needed?

foochifa (commented):
Was wondering if there's a better way to handle the setting tab changes while still keeping track of displaypayments

```
const settingsTabsOrder = [
  'general',
  'singpass',
  'twilio',
  'webhooks',
  'payments',
]
...

  const tabIndex = settingsTabsOrder.indexOf(settingsTab ?? '')

  // if tabIndex doesn't exist or if doesn't meet displayPayments requirement
  // redirect to general tab
  if (
    tabIndex < 0 ||
    (!displayPayments && settingsTab === settingsTabsOrder[4])
  )
    navigate(`${ADMINFORM_ROUTE}/${formId}/settings/${settingsTabsOrder[0]}`)
```

Then when we will eventually move payment out of beta, we can just delete the conditional of `(!displayPayments && settingsTab === settingsTabsOrder[4])`

Rather than having to refactor the useEffect and useState

foochifa (commented):
Was wondering if there's a better way to handle the setting tab changes while still keeping track of displaypayments

```
const settingsTabsOrder = [
  'general',
  'singpass',
  'twilio',
  'webhooks',
  'payments',
]
...

  const tabIndex = settingsTabsOrder.indexOf(settingsTab ?? '')

  // if tabIndex doesn't exist or if doesn't meet displayPayments requirement
  // redirect to general tab
  if (
    tabIndex < 0 ||
    (!displayPayments && settingsTab === settingsTabsOrder[4])
  )
    navigate(`${ADMINFORM_ROUTE}/${formId}/settings/${settingsTabsOrder[0]}`)
```

Then when we will eventually move payment out of beta, we can just delete the conditional of `(!displayPayments && settingsTab === settingsTabsOrder[4])`

Rather than having to refactor the useEffect and useState

frontend/src/features/admin-form/settings/SettingsPage.tsx
LinHuiqing: Yea valid point, it did cross my mind to do this instead but I was thinking how this will reveal our beta features HAHA But okay I think that was just paranoia speaking and I don't think admins will try to enumerate our pages 😂

LinHuiqing: Hm though, we might pre-maturely redirect them to general because the displayPayments will take time :/ okay let me see what I can do

frontend/src/features/admin-form/settings/SettingsPage.tsx
LinHuiqing: The `useEffect` was needed because we need to wait for displayPayments to load first then do the navigation

foochifa: Ahh.. Okay I get what you mean now! Thanks for the explanation!

frontend/src/features/admin-form/settings/SettingsPage.tsx
LinHuiqing: okay actually another concern I had was also in the case where we have multiple tabs where we may hide from users, cos in that case, just using the array itself won't work cos its a fixed order (and won't correspond to the tab index).

LinHuiqing: Honestly, I think it isn't that bad to remove this conditional when we move to payments on prod? Just need to remove the `useEffect` hook

foochifa: Yup on second thought I think its okay to keep the hooks around, don't think we can circumvent it as we want to delay the redirect under `displayPayments` is properly loaded. 

Though I still think we should keep payments in the original `settingsTabOrder`, And just do a conditional check within the useEffect!

Also do you think we should also just check and redirect erroneous settingsTab (like `settings/whatever` to `settings/general`)? I don't think it will usually happen though so might not be a cause for concern

foochifa: Discussed to keep current implementation but add comments for documenting rationale

LinHuiqing: As discussed over huddle, we'll keep the current implementation for extensibility. Shouldn't matter since payments to prod should happen soon!! 🤞 🤞

foochifa (approved):
LGTM! Thanks HQ!

frontend/src/features/admin-form/settings/SettingsPage.tsx
LinHuiqing: comments added in [1c38c7f](https://github.com/opengovsg/FormSG/pull/6348/commits/1c38c7f887dee8ceb4fa2547f67544a8afd77fd1)

PR #6350, opened 2023-05-18
build: release 6.50.0 hotfix

LinHuiqing:
## Problem

- In release [6.50.0](https://github.com/opengovsg/FormSG/pull/6347), we introduced a bug ([f4f5998](https://github.com/opengovsg/FormSG/pull/6349/commits/f4f59984d13054f60ffeb0c770bfabf56500d4a3)) where payment-related copy meant for non-production environments shown on production environments.
- This PR fixes it by making the `SECRET_ENV` environment variable accessible on frontend through `useEnv`.

## Tests

- [x] On a storage-mode form that does not have payments enabled, go to payment settings page and check that the copy is correct and links to the guide.
- [x] When the `SECRET_ENV` is set as `production`:
   - [x] Go to a storage-mode form with payment enabled and preview the form. There should be a banner and check that the copy is expected as above.
    - [x] Go to a storage-mode form with payment enabled, open the form and try to pay as a respondent. On the payment page, check that there is **no infobox** above the payment element.
- [x] When the `SECRET_ENV` is _not_ set as `production` (e.g. `staging`):
   - [x] Go to a storage-mode form with payment enabled and preview the form. There should be a banner and check that the copy is expected as above.
   - [x] Go to a storage-mode form with payment enabled, open the form and try to pay as a respondent. On the payment page, check that there is an infobox with the expected copy on the test card.

### DIFF
--- a/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
+++ b/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
@@ -25,6 +25,7 @@ import Button, { ButtonProps } from '~components/Button'
 import Link from '~components/Link'
 
 import { UseTemplateModal } from '~features/admin-form/template/UseTemplateModal'
+import { useEnv } from '~features/env/queries'
 import { usePublicFormContext } from '~features/public-form/PublicFormContext'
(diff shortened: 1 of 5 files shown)


wanlingt (approved):
thanks!!

PR #6358, opened 2023-05-18
fix(FRM-768): add missing fields to sample-submission endpoint

wanlingt:
## Problem

The sample submission endpoint was returning `null` for some fields.

Closes FRM-768

## Solution

- Add the remaining fields - Table, Checkbox, UEN, Date, NRIC
- Also, took the chance to use [faker](https://www.npmjs.com/package/@faker-js/faker) to generate random data

**Breaking Changes** 

**Improvements**:
- Use `BasicField` types

## Tests

- [ ] Create a form with all answerable field types (so excluding Heading, Paragraph, Image). Make the form public, and send a GET request to http://form.gov.sg/api/v3/forms/:formId/sample-submission. Check that an answer is returned for each field.

## Deploy Notes

**New dependencies**:

- `@faker-js/faker` : Generate fake (but realistic) data

### DIFF
--- a/src/app/modules/form/form.service.ts
+++ b/src/app/modules/form/form.service.ts
@@ -402,6 +402,11 @@ export const createSingleSampleSubmissionAnswer = (field: FormFieldDto) => {
       }
       sampleValue = tableSampleValue
       break
+    case BasicField.Checkbox:
+      noOfOptions = field.fieldOptions.length
+      randomSelectedOption = Math.floor(Math.random() * noOfOptions)
+      sampleValue = field.fieldOptions[randomSelectedOption]
+      break
     default:
       break
   }
--- a/src/app/modules/form/form.service.ts
+++ b/src/app/modules/form/form.service.ts
@@ -1,3 +1,4 @@
+import { faker } from '@faker-js/faker'
 import mongoose from 'mongoose'
 import { err, errAsync, ok, okAsync, Result, ResultAsync } from 'neverthrow'
 
@@ -357,8 +358,7 @@ export const createSingleSampleSubmissionAnswer = (field: FormFieldDto) => {
   switch (field.fieldType) {
     case BasicField.LongText:
     case BasicField.ShortText:
-      sampleValue =
-        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.'
+      sampleValue = faker.lorem.sentence()
       break
     case BasicField.Radio:
     case BasicField.Dropdown:
@@ -367,25 +367,26 @@ export const createSingleSampleSubmissionAnswer = (field: FormFieldDto) => {
       sampleValue = field.fieldOptions[randomSelectedOption]
       break
     case BasicField.Email:
-      sampleValue = 'hello@example.com'
+      sampleValue = faker.internet.email()
       break
     case BasicField.Decimal:
-      sampleValue = 1.234
+      sampleValue = faker.number.float({ precision: 0.1 })
       break
     case BasicField.Number:
-      sampleValue = 1234
+      sampleValue = faker.number.int(100)
       break
     case BasicField.Mobile:
-      sampleValue = '+6598765432'
+      sampleValue = faker.phone.number('+659#######')
       break
     case BasicField.HomeNo:
-      sampleValue = '+6567890123'
+      sampleValue = faker.phone.number('+656#######')
       break
     case BasicField.YesNo:
       sampleValue = 'yes'
       break
     case BasicField.Rating:
-      sampleValue = 1
+      noOfOptions = field.ratingOptions.steps
+      sampleValue = faker.number.int({ min: 1, max: noOfOptions })
       break
     case BasicField.Attachment:
       sampleValue = 'attachmentFileName'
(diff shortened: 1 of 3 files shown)


src/app/modules/form/form.service.ts
```diff
@@ -348,44 +350,74 @@ export const retrievePublicFormsWithSmsVerification = (
...
+  let noOfTableRows
+  let noOfTableCols
+  const tableSampleValue = []
   switch (field.fieldType) {
-    case 'textarea':
-    case 'textfield':
-      sampleValue =
-        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.'
+    case BasicField.LongText:
+      sampleValue = faker.lorem.sentence()
```
LinHuiqing:
```suggestion
      sampleValue = faker.lorem.text()
```
Might want to use `text` instead to generate anywhere between a sentence to paragraphs! Just like how we expect users will.

src/app/modules/form/form.service.ts
```diff
@@ -348,44 +350,74 @@ export const retrievePublicFormsWithSmsVerification = (
...
       break
-    case 'radiobutton':
-    case 'dropdown':
-      noOfOptions = field.fieldOptions.length
-      randomSelectedOption = Math.floor(Math.random() * noOfOptions)
-      sampleValue = field.fieldOptions[randomSelectedOption]
+    case BasicField.ShortText:
+      sampleValue = faker.lorem.word()
```
LinHuiqing:
```suggestion
      sampleValue = faker.lorem.words()
```
I think `words` will be more reflective of what users will input in short answer fields too?

src/app/modules/form/form.service.ts
```diff
@@ -348,44 +350,74 @@ export const retrievePublicFormsWithSmsVerification = (
...
       break
-    case 'yes_no':
-      sampleValue = 'yes'
+    case BasicField.HomeNo:
+      sampleValue = faker.phone.number('+656#######')
       break
-    case 'rating':
-      sampleValue = 1
+    case BasicField.YesNo:
+      sampleValue = faker.helpers.arrayElement(['Yes', 'No'])
```
LinHuiqing, line 383: Maybe this should be separate issue: Feels somewhat odd that we can use `field.fieldOptions` for radio and dropdown options, but have to resort to hard-coding `'Yes', 'No'` for this when these options are determined on our end already. Maybe we should have like something like this:
```ts
// in a const file somewhere
export const yesNoOptions = ['Yes', 'No'] as const

// in frontend/src/templates/Field/types.ts
export type YesNoFieldValue = typeof yesNoOptions[number]
```
Then we can just import `yesNoOptions` in places like these where we want to use the options.

src/app/modules/form/form.service.ts
```diff
@@ -348,44 +350,74 @@ export const retrievePublicFormsWithSmsVerification = (
...
+      break
+    case BasicField.Checkbox:
+      sampleValue = faker.helpers.arrayElements(field.fieldOptions)
+      break
+    case BasicField.Date:
+      sampleValue = new Date(Date.now()).toLocaleDateString('en-SG', {
+        day: 'numeric',
+        month: 'long',
+        year: 'numeric',
+      })
```
LinHuiqing:
```suggestion
      sampleValue = faker.date.anytime().toLocaleDateString('en-SG', {
        day: 'numeric',
        month: 'long',
        year: 'numeric',
      })
```
Any reason to not use `faker` for this too? 👀

src/app/modules/form/form.service.ts
```diff
@@ -348,44 +350,74 @@ export const retrievePublicFormsWithSmsVerification = (
...
+      break
+    case BasicField.Nric:
+      sampleValue = faker.helpers.replaceSymbols('S9######A')
```
LinHuiqing, line 416: /justNoting
I wonder if hard-coding the NRIC with A will break workflows that will want to check for this. But this should be handled by the upcoming validation part anyway!

LinHuiqing (commented):
Generally looks good! Just some small comments. Also wow can't believe we missed the hard-coding of field types last review oop

LinHuiqing (approved):
tested! should be good to go after the stuff i commented above bah

src/app/modules/form/form.service.ts
wanlingt: [c91e6c5](https://github.com/opengovsg/FormSG/pull/6358/commits/c91e6c571c8c3b1ed13ab1eb543d84b111986842)

src/app/modules/form/form.service.ts
wanlingt: [c91e6c5](https://github.com/opengovsg/FormSG/pull/6358/commits/c91e6c571c8c3b1ed13ab1eb543d84b111986842)

src/app/modules/form/form.service.ts
wanlingt: Haha no reason! changed here :) [c91e6c5](https://github.com/opengovsg/FormSG/pull/6358/commits/c91e6c571c8c3b1ed13ab1eb543d84b111986842)

src/app/modules/form/form.service.ts
wanlingt: Yeah this is a good point! Yes and No are also being hardcoded in the options now for the YesNo field. I'll leave it as is for now, but will create a separate issue to improve this

timotheeg (approved):

src/app/modules/form/form.service.ts
wanlingt: added here FRM-892

PR #6360, opened 2023-05-19
feat: add target account id directly to payment documents

justynoh:
## Problem

In order to account for linking and re-linking of Stripe accounts to forms, we should ensure that we store the associated stripe Account with the payment documents, rather than going back to the original form. This will also be required in preparation for the async CRON job which will be implemented in #6102.

## Solution

There are several options that can be considered.
1. Don't save the target account id explicitly at all, and rely on the webhooks `account` value. 
2. Save the target account id only explicitly.
3. Save the target account id and publishable key explicitly (i.e. duplicate of `FormPaymentChannel`).

Considerations:
1. We (forms) should be responsible for saving the target account id, since we generate the initial request to stripe. We should also ensure that the webhooks all belong to the same target account id, which is the one that we initiated the payment for. Relying on the incoming webhooks is not preferred.
2. The only time the publishable key is needed is when we pass it to the frontend to call stripe's API directly. And also, in any case, holding the pub key is kind of like saying "I give you access to make API calls for me". If the form's stripe account has changed, then we should no longer be able to do things "for" this stripe account.

With these considerations, option 2 is chosen with the following benefits:
1. Validation that stripe event account contains the same account id as the associated payment document.
2. Still prevents pending payments from being completed when the stripe account on the form is changed (publishable key is not available).

**Breaking Changes** 
- **Yes - this PR contains breaking changes**

## Deploy Notes

**New scripts**:

- `20230519_add_target_account_id`: adds `targetAccountId` to payment collection, by extracting the account id from the first value in the webhook log (assumes that Stripe has been correct so far).

### DIFF
--- a/scripts/20230519_add_target_account_id/script.js
+++ b/scripts/20230519_add_target_account_id/script.js
@@ -1,0 +1,24 @@
+/* eslint-disable */
+
+// BEFORE
+// Count total number of payments with no received webhooks. This should be 0.
+db.getCollection('payments').countDocuments({ webhookLog: { $size: 0 } }) === 0
+
+// Count total number of payments.
+const total = db.getCollection('payments').countDocuments({})
+
+// Count total number of payments with no targetAccountId value. This should be equal to the total number of payments.
+db.getCollection('payments').countDocuments({ targetAccountId: { $exists: false } }) === total
+
+// UPDATE
+db.getCollection('payments').update(
+  {},
+  { $set: { targetAccountId: { $first: '$webhookLog.account' } } },
+)
+
+// AFTER
+// Count total number of payments with targetAccountId value. This should be equal to the total number of payments.
+db.getCollection('payments').countDocuments({ targetAccountId: { $exists: true } }) === total
+
+// Count total number of payments with no targetAccountId value. This should be equal to 0
+db.getCollection('payments').countDocuments({ targetAccountId: { $exists: false } }) === 0
--- a/scripts/20230519_add_target_account_id/script.js
+++ b/scripts/20230519_add_target_account_id/script.js
@@ -11,9 +11,9 @@ const total = db.getCollection('payments').countDocuments({})
 db.getCollection('payments').countDocuments({ targetAccountId: { $exists: false } }) === total
 
 // UPDATE
-db.getCollection('payments').update(
+db.getCollection('payments').updateMany(
   {},
-  { $set: { targetAccountId: { $first: '$webhookLog.account' } } },
+  [ { $set: { targetAccountId: { $first: '$webhookLog.account' } } } ],
 )
 
 // AFTER
--- a/shared/types/payment.ts
+++ b/shared/types/payment.ts
@@ -33,6 +33,7 @@ export type Payment = {
   // Pre-payment metadata
   pendingSubmissionId: string
   formId: string
+  targetAccountId: string
   email: string
   amount: number
   paymentIntentId: string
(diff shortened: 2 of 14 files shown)


src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -1562,9 +1562,9 @@ const deleteTwilioTransaction = async (
...
 
 /**
- * Update the payments of the given form
+ * Update the payments field of the given form
  * @param formId the id of the form to update the end page for
- * @param newStartPage the new start page object to replace the current one
+ * @param newPayments the new payments field to replace the current one
  * @returns ok(updated start page object) when update is successful
  * @returns err(FormNotFoundError) if form cannot be found
  * @returns err(PossibleDatabaseError) if start page update fails
```
LinHuiqing: while we're at it, update the return types documentation as well?

shared/types/payment.ts
```diff
@@ -33,6 +33,7 @@ export type Payment = {
   // Pre-payment metadata
   pendingSubmissionId: string
   formId: string
+  targetAccountId: string
```
LinHuiqing, line 36: camelCase for typescript ftw!

scripts/20230519_add_target_account_id/script.js
```diff
@@ -0,0 +1,24 @@
...
+db.getCollection('payments').countDocuments({ webhookLog: { $size: 0 } }) === 0
+
+// Count total number of payments.
+const total = db.getCollection('payments').countDocuments({})
+
+// Count total number of payments with no targetAccountId value. This should be equal to the total number of payments.
+db.getCollection('payments').countDocuments({ targetAccountId: { $exists: false } }) === total
+
+// UPDATE
+db.getCollection('payments').update(
```
LinHuiqing:
```suggestion
db.getCollection('payments').updateMany(
```
Got this error when trying to run the script on my Studio3T:
```sh
DeprecationWarning: Collection.update() is deprecated. Use updateOne, updateMany, or bulkWrite.
```

scripts/20230519_add_target_account_id/script.js
```diff
@@ -0,0 +1,24 @@
...
+  {},
+  { $set: { targetAccountId: { $first: '$webhookLog.account' } } },
```
LinHuiqing:
```sh
Uncaught MongoServerError: The dollar ($) prefixed field '$first' in 'targetAccountId.$first' is not valid for storage.
```
Not sure if it's just my end - Studio3T throws this when I try to run this line?

LinHuiqing (commented):
Generally lgtm! I'll need to test it a bit more though, sorry

src/app/modules/form/admin-form/admin-form.service.ts
justynoh: omg LOL thanks ahhaha i missed that, done in df6e178d

scripts/20230519_add_target_account_id/script.js
justynoh: yes fixed! [6f05a50](https://github.com/opengovsg/FormSG/pull/6360/commits/6f05a50b60029d851d39951de780bb02a25fba4d)

scripts/20230519_add_target_account_id/script.js
justynoh: oooo. yup, fixed in [6f05a50](https://github.com/opengovsg/FormSG/pull/6360/commits/6f05a50b60029d851d39951de780bb02a25fba4d)!

LinHuiqing (approved):
lgtm and tested!

PR #6368, opened 2023-05-22
chore: Update sgID client API version

kenjin-work:
## Problem
sgID client API needs updating from v1 to v2.

Closes FRM-899

## Solution

For dev:
- Update Mockpass.
- Update dev env vars.
- For all devs: **please delete your old docker build and container for mockpass before running the dev script**

For prod:
- Update prod env vars.

**Breaking Changes** 

    - It's not breaking for users, but might be breaking for our own development. Mockpass updates sgID to use version 2 instead of version 1. So we need to ensure that our sgID code still works (it should).

## Deploy Notes
Existing env vars for sgID hostname need updating to version 2 of the API. In both testing and prod.

Nothing new but please see deploy notes. **This PR needs updated env vars**


kenjin-work:
Please don't merge this yet. It seems to break our local development.

PR #6376, opened 2023-05-23
feat: bearer token authentication for public APIs

wanlingt:
## Problem

Closes FRM-906

## Solution

This PR allows admins with a valid API token to access public APIs. API tokens are in the format `env_version_userId_token`.

**Breaking Changes** 

**Features**:
- Created a new base route for public APIs (`/api/public/v1`)
  Public APIs will not share the same namespace as internal session-based APIs (`/api/v3`) as the external APIs will have different functions (e.g. updating multiple form fields at one go)
- The hash of the API token is stored in the `user` collection as `ApiKeyHash`
- Add public API endpoint for listing forms (GET `/api/public/v1/admin/forms`)
- Added new error types for public APIs

#### Test API
**Test 1: API works**
  - [ ] a. Generate a random string using `crypto.randomBytes(32).toString("base64")`
  - [ ] b. find your user id, and format an API key as **`test_v1_${userId}_${randomString}`**
  - [ ] c. Store a hash of that key (using 1 salt round) in the user collection, adding it using this structure: 
```
  "apiToken" :{
      "keyHash": "${hashedApiKey}",
      "createdAt": ISODate("2023-06-01T00:00:00.000Z"),
      "lastUsedAtAt": ISODate("2023-06-01T00:00:00.000Z")
      }
```
 (instructions on how to generate the api key and hash are also [here](https://www.notion.so/opengov/API-d3b711dbd53f4b8e96c52ce2d8e0a171))
  - [ ] d. Send an API request as such:
```
curl --request GET \
  --url http://form.gov.sg/api/public/v1/admin/forms \
  --header 'Authorization: Bearer ${apiKey}' \
```
  You should receive a `200 OK` response, with the full list of forms associated with your user.

**Test 2: Without authorisation header**
  - [ ] Repeat the same steps as above. 
  - [ ] In Step 2c, send a curl request without an authorisation header
```
curl --request GET \
  --url http://form.gov.sg/api/public/v1/admin/forms \
```
  - [ ] You should receive a `400 Bad Request` response, with the message "Authorisation header is missing"

**Test 3: With wrong credentials**
  - [ ] Repeat the same steps as above. 
  - [ ] In Step 2c, send a curl request with a different API token
```
curl --request GET \
  --url http://form.gov.sg/api/public/v1/admin/forms \
  --header 'Authorization: Bearer test_v1_${userId}_${WRONG_API_TOKEN}' \
```
  - [ ] You should receive a `401 Unauthorized` response

#### Regression Test
**Test 4: Regression test for `handleListDashboardForms`**
- [ ] Log in. You should see the full list of forms in your dashboard, i.e. the GET request to http://form.gov.sg/api/v3/admin/forms should return with a `200 OK`

## Deploy Notes

- `API_KEY_VERSION`: current API version
- `PUBLIC_API_RATE_LIMIT`: Per-minute, per-IP, per-instance request limit for public APIs


src/app/routes/api/external/v1/admin/forms/admin-forms.external.routes.ts
```diff
@@ -0,0 +1,26 @@
...
+   * List the forms managed by the user
+   * @security session
+   *
+   * @returns 200 with a list of forms managed by the user
+   * @returns 401 when user is not authorised
+   * @returns 422 when user of given id cannnot be found in the database
+   * @returns 500 when database errors occur
+   */
+  .get(
+    limitRate({ max: rateLimitConfig.externalApi }),
```
timotheeg: This protects the endpoint overall, but we will eventually need a rate limit per api key, so don't have a situation where one bad actors blocks the API for everyone.

In fact, in the TODOs related for that project, I would love finally getting a centralized rate limit system.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -198,9 +199,15 @@ const fileUploadValidator = celebrate({
...
 export const handleListDashboardForms: ControllerHandler<
   unknown,
-  AdminDashboardFormMetaDto[] | ErrorDto
+  AdminDashboardFormMetaDto[] | ErrorDto,
+  ApiReqBody
 > = async (req, res) => {
-  const authedUserId = (req.session as AuthedSessionData).user._id
+  let authedUserId: string
+  if (req.body.formSg?.userId) {
+    authedUserId = req.body.formSg?.userId
```
timotheeg: The namespace `req.body` looks wrong 🤔 , can we shove the `formsg` namespace straight in `req`?

timotheeg: Hmmm, in fact, instead of having to change this method to retrieve the user from 2 different objects, I'm thinking we could let the API auth put the authenticated user in the session object itself, so the old code would just work as-is.

src/app/modules/auth/auth.middlewares.ts
```diff
@@ -92,3 +95,45 @@ export const validateVerifyOtpParams = celebrate({
...
+    return res
+      .status(StatusCodes.BAD_REQUEST)
+      .json({ message: 'Invalid authorisation header format' })
+  }
+  const [apiEnv, apiVersion, userId, token] = apiKey.split(API_KEY_SEPARATOR)
+  if (!apiEnv || !apiVersion || !userId || !token) {
+    return res
+      .status(StatusCodes.BAD_REQUEST)
+      .json({ message: 'Invalid API key format' })
+  }
```
timotheeg: A regex validation scheme here could be more precise in validating the shape of the header and the key 🤔 . I can help craft that.

src/app/modules/auth/auth.middlewares.ts
```diff
@@ -92,3 +95,45 @@ export const validateVerifyOtpParams = celebrate({
...
+  return getUserByApiKey(userId, token)
+    .map((user) => {
+      if (!user) {
+        return res
+          .status(StatusCodes.UNAUTHORIZED)
+          .json({ message: 'Invalid API key' })
+      }
+      req.body.formSg = { userId: user.id }
```
timotheeg: Same comment as earlier, the namespace `req.body` feels wrong. body means something, and represents what the request actually sent (as-in: request header; request body), if we have a standard place to save the authenticated user, let's use that (in the session object), and if not, we can create a completely separate namespace to indicate that this is custom metadata.

timotheeg (commented):
Looking good WL!
 
I added some comments

src/app/routes/api/external/v1/admin/forms/admin-forms.external.routes.ts
wanlingt: FRM-948

src/app/modules/auth/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const BEARER_STRING = 'Bearer'
+export const BEARER_SEPARATOR = ' '
+export const API_KEY_SEPARATOR = '_'
+export const DEFAULT_SALT_ROUNDS = 1
```
LinHuiqing: Should this be exposed in the code or should we load this in as an env var instead? From the point of view of exposing our protocol, I think loading env vars is slightly more secure. It'll also be easier for us to tweak if we want to change it.

src/app/modules/auth/auth.service.ts
```diff
@@ -339,3 +348,50 @@ export const getFormIfPublic = (
...
+    },
+  ).map((hash) => `${hash}`)
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const generateApiKey = (user: IUserSchema): string => {
+  const key = crypto.randomBytes(32).toString('base64')
+  const apiEnv = config.publicApiConfig.apiEnv
+  const apiKeyVersion = config.publicApiConfig.apiKeyVersion
+  return `${apiEnv}_${apiKeyVersion}_${user._id}_${key}`
```
LinHuiqing, line 396: During the huddle, brought up that using `user._id` without obfuscation is kinda the same as using `email` as an identifier. But on second thought, user id is slightly better in the case where attackers somehow have a list of api keys, then the user ids provide them with an easier vector for them to figure out which agency they want to attack haha

src/app/modules/auth/constants.ts
LinHuiqing: Oh! Also I think we should set more rounds since this is for the API key which is permanent, rather than something ephemeral like our OTP!

LinHuiqing: Oh! Also I think we should set more rounds since this is for the API key which is permanent, rather than something ephemeral like our OTP!

src/app/modules/form/admin-form/admin-form.controller.ts
wanlingt: Thanks Tim, good point! I've put it in the session object itself - [32c1cc1](https://github.com/opengovsg/FormSG/pull/6376/commits/32c1cc1935b89e34e5abe9fa3803ea6520cef0e7)

src/app/modules/auth/auth.middlewares.ts
wanlingt: (as above) [32c1cc1](https://github.com/opengovsg/FormSG/pull/6376/commits/32c1cc1935b89e34e5abe9fa3803ea6520cef0e7)

src/app/modules/auth/auth.middlewares.ts
```diff
@@ -92,3 +94,51 @@ export const validateVerifyOtpParams = celebrate({
...
+    },
+  })
+  return getUserByApiKey(userId, token)
+    .map((user) => {
+      if (!user) {
+        return res
+          .status(StatusCodes.UNAUTHORIZED)
+          .json({ message: 'Invalid API key' })
+      }
+      req.session.user = { _id: user.id }
```
timotheeg: Just a note on this: this works because the session middleware is currently running for all routes served by express, and that is sort of wrong. 

We should activate the session middleware only for routes which require it (typically under the API router), and for the new "external" APIs, our auth middleware should take responsibility to setup a compatible session object. Something as small as:

```suggestion
if (!req.session) req.session = {}
req.session.user = { _id: user.id }
```

kenjin-work:
Random thought: Is implementing expiring bearer tokens (e.g. after a year) a common practice? Or is that not required here?

src/app/modules/auth/auth.middlewares.ts
KenLSM: Taking a stab...

```js
new RegExp(
  '^' + // ensure start of string
    '(?:Bearer)' + // Non-capturing group: marching the "Bearer" word
    '[ ]' + // Single space
    "(?'apiEnvToken'w+)" + // Capture group {apiEnvToken}
    '_' + // delimiter
    "(?'apiVerToken'vd+)" + // Capture group {apiVerToken}: matching starting with character v, and at least 1 digit
    '_' + // delimiter
    "(?'userIdToken'[[:alnum:]]{24})" + // Capture group {userIdToken}: matching a mongodb objectid, 24 characters of alphanumeric
    '_' + // delimiter
    "'apiToken'(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}={2}))" + // Capture group {userIdToken}: matching a mod4 base64 value
    '$', // ensure end of string,
)
```

timotheeg:
> Random thought: Is implementing expiring bearer tokens (e.g. after a year) a common practice? Or is that not required here?

Good comment @kenjin-work ! 

It's common enough, and it indeed removes risk exposure for keys, especially which are not actively in use. But it's not a must-have practice, and many small SaaS do not have expiry on their token (e.g. BetterUptime or Snyk to name just 2 of our vendors). Removing keys can be disruptive to active automations, and so it typically requires some additional mechanisms to inform users (likely with multiple reminders!) that their key will expire in, say, 60 days, 30 days, 10 days, 2 days, 1 day, BOOM 😅 

This API initiative is on POC/MVP mode right now, With the primary goal of showing what the authentication model and API namespace can be. We are not even considering building the UI to generate/revoke/regenerate the API tokens yet 😅 .

So we (cc @wanlingt as feature owner) can add requirements as we go to make the feature as robust as can. For example, just like we track lastLoginTime for UI sessions, so too we should track lastUsageTime for the token, and indeed we should add a tokenCreationTime too, so we are ready to add an expiration mechanism later on if we want.

timotheeg:
I added some minor changes, but I realized one important thing must be changed: retrieving forms atm returns the entire user object 😱. We need a filter process to make sure any non-intended fields is not returned to the client. The comment applies for both the session base API, and bearer token auth.

src/app/modules/auth/auth.middlewares.ts
timotheeg: Nice @KenLSM !

I don't think we need to go to that extent of validation though 😅. The only thing we really care about is to extract `userId`. I added in a simplified one for now.

Just to note also: bcrypt base64 encoding uses `/.` rather than `/+` (see [doc](https://en.wikipedia.org/wiki/Base64#Applications_not_compatible_with_RFC_4648_Base64))

src/app/modules/auth/constants.ts
timotheeg: > Should this be exposed in the code or should we load this in as an env var instead? From the point of view of exposing our protocol, I think loading env vars is slightly more secure. It'll also be easier for us to tweak if we want to change it.

I'm not too worried about this. The strength of the encryption is more important than hiding the number of rounds (i.e. security through obscurity on the algo). Also With bcrypt, the number of rounds is encoded in the hash, so it is visible information to anyone who gets their hand on the hash (though obviously, we should do everything we can to protect against that!).

I also think that changing the number of rounds would be a very infrequent operation, so there's no need to have it a tunable.


> I think we should set more rounds since this is for the API key which is permanent, rather than something ephemeral like our OTP!

Agreed. I increased the value to 2, which performs 4 rounds. We could do more, but we need to be mindful of processing cost. In the current POC, we compute the hash on every API call (as opposed to one time on login for session-based access).

At POC staged, I'm not overly worried about this.

timotheeg:
I added the API endpoint to retrieve submission count and submissions themselves (from our `ndjson` stream controller), and [here is a gist](https://gist.github.com/timotheeg/9e96d8902abfa11c28e7b94d18fa7c58) that shows how it can be used to load submissions from the server (all const values `formids`, `privateKey`, `apiKey` are from my local env, so obviously they need to be changed accordingly to run yourselve.

One more note: I'm not sure we need to place the public API endpoints within a namespace `/admin` 🤔 , considering we have bearer token auth, public APIs should necessarily be for admins.

timotheeg:
Also to note / possibly something for discussion.

The `apiToken` is now a structure that includes `createdAt` and `lastUsedAt` fields, and that will be useful for some instrumentation of the feature.

We might need an additional date field `regeneratedAt`, so if a key is revoked to create a new one, we don't lose stats of when admins first started to use an API key 🤔

timotheeg (approved):
LGTM. Good enough for demo even in staging / prod.

Important note though: before we start adding keys into users, we **MUST** change the query that retrieve forms, so the api key hash information NEVER leaves the DB or server. 

ATM, the auto fetch of user objects for form->admin and form->collaborators return the api token structure as a user property. **We need to return the bare minimum fields for users.**

PR #6377, opened 2023-05-23
feat: add response time to submission

foochifa:
## Problem

We want to be able to track respondent's normalised response time (response time normalised by fields available). However, currently, we do not collect nor store any information regarding response time.

Closes FRM-917

## Solution

We have to collect two metrics for every submission 
1. `responseTime` - time taken for the respondent from page load to submission
2. `numVisibleFields` - number of fields actually visible and potentially filled by the respondent at submission time

**Implementation options discussed for response time**
Option 1:
- Calculate time taken on client side, get time at start of page load in react and compare the delta when submitting
- However, client data is unreliable and can be modified

Option 2:
- Calculate the time taken on our BE, which makes it more robust
- However, we have to potentially add in a session token or some other ways to track the session of a respondent

We have decided to take Option 1 for now to expedite the collation of response time data. But will KIV option 2 for the future.

**For number of visible fields**
- `numVisibleFields` have to reside in `PublicFormProvider` as it has to be used on our `handleSubmitForm` method which is defined there
- currently we will set the visible fields whenever there is any updates in `VisibleFields.ts`, this will trigger a re-render to our `PublicFormProvider` and some of its non-memoised children. This should incur a flat performance cost of about 9-10ms every update of visible fields. The performance cost is uniform as only the contexts and containers are being updated, the variable fields are re-rendered anyway upon `VisibleFields.ts` changes.
   - Side note but if public form performance is an issue, we should look at how we handle re-renders for public form fields. There are some unnecessary mass re-renders for visible fields too, despite the memo-isation made

**Misc**

Additionally, have looked at potentially using google analytics (GA) to track engagement time instead but this poses two problems
1. GA tracks unique URLs -> each form is uniquely tracked
2. GA wouldn't be able to know the visible form fields of each respondents, which makes it very difficult to get a normalised response time

Hence, have ruled out the option

Lastly, have included the `responseMetadata` in the submissions (email & encrypt) logger. Thus, in the interim, before addressing usage of datadog metrics distributions, we could get the logger information from handle submissions, and draw our the metadata.

**Breaking Changes** 

- [ ] No - this PR is backwards compatible, `responseMetadata` is intentionally left as optional to allow for any backwards compatibility issues between the rolling update of FE and BE.

## Tests

**To ensure functionality**
- [ ] 1. Go to a public storage mode form with logic, fill in the form and submit
- [ ] 2. The submission should have been logged with the response time and number of visible fields (and the submission document in DB should reflect it too)
- [ ] 3. Repeat the steps for email mode form

Set toggle in `PublicFormProvider` for `useFetchForSubmission`
- [ ] 4. Repeat the steps 1 and 2 above for storage mode form with fetch for submission
- [ ] 5. Repeat the steps 1 and 2 above for email mode form with fetch for submission

For payment forms
- [ ] 6. Go to a payment form, fill in the form and submit
- [ ] 7. The pending submission should have been logged with the response time and number of visible fields (and the pending submission document in DB should reflect it too)
- [ ] 8. Complete the payment
- [ ] 9. The completed submission document for the payment should still contain the response metadata

**To check for regression**
- [ ] Submit a storage mode form without logic
- [ ] Submit an email mode form without logic
- [ ] Submit a preview mode, storage mode form with logic
- [ ] Submit a preview mode, email mode form with logic
- [ ] Create a use-template form with logic

Testing without `responseMetadata`
- [ ] 6. Try and call the email submission api endpoint without a `responseMetadata` in the body (can use an empty form), the submission should still be successful 
- [ ] 7. Try and call the encrypt submission api endpoint without a `responseMetadata` in the body (can use an empty form), the submission should still be successful

### DIFF
--- a/frontend/src/features/public-form/PublicFormPage.tsx
+++ b/frontend/src/features/public-form/PublicFormPage.tsx
@@ -19,8 +19,11 @@ export const PublicFormPage = (): JSX.Element => {
 
   if (!formId) throw new Error('No formId provided')
 
+  // Get date time in miliseconds when user first loads the form
+  const startTime = Date.now()
+
   return (
-    <PublicFormProvider formId={formId}>
+    <PublicFormProvider formId={formId} startTime={startTime}>
       <FormSectionsProvider>
         <Flex direction="column" css={fillMinHeightCss}>
           <FormBanner />
--- a/frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
+++ b/frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
@@ -51,6 +51,7 @@ export const VisibleFormFields = ({
     setVisibleFormFields(visibleFieldsWithQuestionNo)
 
     // set the number of visible fields in the context for public forms
+    // the setter will only exist for public forms and will be undefined for preview
     if (setNumVisibleFields)
       setNumVisibleFields(visibleFieldsWithQuestionNo.length)
   }, [
--- a/frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
+++ b/frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
@@ -8,6 +8,7 @@ import { FormFieldValues } from '~templates/Field'
 import { FormFieldWithQuestionNo } from '~features/form/types'
 import { augmentWithQuestionNo } from '~features/form/utils'
 import { getVisibleFieldIds } from '~features/logic/utils'
+import { usePublicFormContext } from '~features/public-form/PublicFormContext'
 
 import { FieldFactory } from './FieldFactory'
 import { PrefillMap } from './FormFields'
@@ -35,6 +36,7 @@ export const VisibleFormFields = ({
   const watchedValues = useWatch({ control })
   const { setVisibleFieldIdsForScrollData } = useFormSections()
   const [visibleFormFields, setVisibleFormFields] = useState(formFields)
+  const { setNumVisibleFields } = usePublicFormContext()
 
   useEffect(() => {
     const visibleFieldIds = getVisibleFieldIds(watchedValues, {
@@ -47,7 +49,16 @@ export const VisibleFormFields = ({
     )
     const visibleFieldsWithQuestionNo = augmentWithQuestionNo(visibleFields)
     setVisibleFormFields(visibleFieldsWithQuestionNo)
-  }, [formFields, formLogics, setVisibleFieldIdsForScrollData, watchedValues])
+
+    // set the number of visible fields in the context
+    setNumVisibleFields(visibleFieldsWithQuestionNo.length)
+  }, [
+    formFields,
+    formLogics,
+    setVisibleFieldIdsForScrollData,
+    watchedValues,
+    setNumVisibleFields,
+  ])
 
   return (
     <>
--- a/frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
+++ b/frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
@@ -50,8 +50,9 @@ export const VisibleFormFields = ({
     const visibleFieldsWithQuestionNo = augmentWithQuestionNo(visibleFields)
     setVisibleFormFields(visibleFieldsWithQuestionNo)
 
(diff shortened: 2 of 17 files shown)


frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
```diff
@@ -47,7 +49,17 @@ export const VisibleFormFields = ({
     )
     const visibleFieldsWithQuestionNo = augmentWithQuestionNo(visibleFields)
     setVisibleFormFields(visibleFieldsWithQuestionNo)
-  }, [formFields, formLogics, setVisibleFieldIdsForScrollData, watchedValues])
+
+    // set the number of visible fields in the context for public forms
+    if (setNumVisibleFields)
```
LinHuiqing, line 55: Will there be a case where `setNumVisibleFields` is `false`? I don't think so right? 🤔 

Side note: seems a little odd that we're using a setter as a conditional value haha

frontend/src/features/public-form/PublicFormPage.tsx
```diff
@@ -19,8 +19,11 @@ export const PublicFormPage = (): JSX.Element => {
 
   if (!formId) throw new Error('No formId provided')
 
+  // Get date time in miliseconds when user first loads the form
+  const startTime = Date.now()
```
LinHuiqing, line 23: (not related to technical implementation)
Not sure if this has been discussed for the metrics, but I wonder if it might make more sense if we take the start time based on the first field that has been touched?

LinHuiqing: Rather than doing all computations and storing all data on our side, can consider leveraging on other tools as well, another example will be DataDog's RUM (see: [Monitoring Page Performance](https://docs.datadoghq.com/real_user_monitoring/browser/monitoring_page_performance/#add-your-own-performance-timing) and [Send RUM Custom Actions](https://docs.datadoghq.com/real_user_monitoring/guide/send-rum-custom-actions/?tab=npm))

LinHuiqing: Since this is urgent, we can keep your way of implementing then change it later if we want to.

LinHuiqing (approved):
LGTM actually and tested! Other than 1 small comment you might want to address for the setter.

frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
foochifa: Yep! For Preview mode it will be `undefined`! Cause we aren't sending any `responseMetadata` on preview mode responses.

I do agree its a bit weird to use a setter as a conditional 😅 was trying to tie it to `isPreview` but couldn't really get the typing to work.

frontend/src/features/public-form/PublicFormPage.tsx
foochifa: Hmm I would think that getting start time from page load still makes sense? Cause users might take time to read the instructions/headers/questions before filling them? But they are ultimately still part of the response time right

frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
LinHuiqing: Ooh huh, okay lemme take a closer look

LinHuiqing: Okay yea I think because we need to explicitly define 2 types for preview form and public form for this to work cleanly. Okay just use the setter as a conditional I guess, can you add a comment for this?

frontend/src/features/public-form/PublicFormPage.tsx
timotheeg: Yeah, I think it makes sense from load to submission for now. And we can always revisit or add another metric later if we want 👍

frontend/src/features/public-form/components/FormFields/VisibleFormFields.tsx
foochifa: [c420ccb](https://github.com/opengovsg/FormSG/pull/6377/commits/c420ccb3e5efe22acc5b580925557220caabd873)

Thanks HQ!

PR #6387, opened 2023-05-25
feat: inject normalised response time to DD

foochifa:
## Problem

Now that #6377 injects the `responseMetadata` to DB, we should also compute and provide the normalised response time to datadog as our metrics

Closes FRM-499

## Solution

Use datadog `statsdClient.distribution` and added in a new method in `submissions.util.ts` to compute normalised response time using the formula 10 * responseTime / numVisibleFields. To note, time is currently returned in miliseconds.

Consideration for pending submissions:
- Decided to return response time at creation of pending submission instead of at webhook confirmation
- Reason being that this set of metrics is for the general form filling time (disincluding payments), and if the user have a made a pending submission - it means they have filled up the form
- Side note: we should probably have another metric to track payment time 

**Breaking Changes** 

## Tests

- [ ] Make a couple of submissions
- [ ] go to dd metrics explorer
- [ ] look for `formsg.submissions.normalisedResponseTime` 
- [ ] your submissions' response time data should be visible

### DIFF
--- a/src/app/modules/submission/email-submission/email-submission.controller.ts
+++ b/src/app/modules/submission/email-submission/email-submission.controller.ts
@@ -7,6 +7,7 @@ import {
 } from '../../../../../shared/types'
 import { IPopulatedEmailForm } from '../../../../types'
 import { ParsedEmailModeSubmissionBody } from '../../../../types/api'
+import { statsdClient } from '../../../config/datadog-statsd-client'
 import { createLoggerWithLabel } from '../../../config/logger'
 import * as CaptchaMiddleware from '../../../services/captcha/captcha.middleware'
 import * as CaptchaService from '../../../services/captcha/captcha.service'
@@ -25,7 +26,10 @@ import { SgidService } from '../../sgid/sgid.service'
 import { getOidcService } from '../../spcp/spcp.oidc.service'
 import * as EmailSubmissionMiddleware from '../email-submission/email-submission.middleware'
 import * as SubmissionService from '../submission.service'
-import { extractEmailConfirmationData } from '../submission.utils'
+import {
+  extractEmailConfirmationData,
+  getNormalisedResponseTime,
+} from '../submission.utils'
 
 import * as EmailSubmissionService from './email-submission.service'
 import { IPopulatedEmailFormWithResponsesAndHash } from './email-submission.types'
@@ -316,6 +320,20 @@ const submitEmailModeForm: ControllerHandler<
             message: 'Sending admin mail',
             meta: logMetaWithSubmission,
           })
+          if (responseMetadata)
+            statsdClient.distribution(
+              'formsg.submissions.normalisedReponseTime',
+              getNormalisedResponseTime(
+                responseMetadata.responseTimeMs,
+                responseMetadata.numVisibleFields,
+              ),
+              1,
+              {
+                mode: 'email',
+                responseTimeMs: '${responseMetadata.responseTimeMs}',
+                numOfVisibleFields: '${responseMetadata.numVisibleFields}',
+              },
+            )
 
           // Send response to admin
           // NOTE: This should short circuit in the event of an error.
--- a/src/app/modules/submission/email-submission/email-submission.controller.ts
+++ b/src/app/modules/submission/email-submission/email-submission.controller.ts
@@ -322,7 +322,7 @@ const submitEmailModeForm: ControllerHandler<
           })
           if (responseMetadata)
             statsdClient.distribution(
-              'formsg.submissions.normalisedReponseTime',
+              'formsg.submissions.normResponseTimeMetadata',
               getNormalisedResponseTime(
                 responseMetadata.responseTimeMs,
                 responseMetadata.numVisibleFields,
--- a/src/app/modules/submission/email-submission/email-submission.controller.ts
+++ b/src/app/modules/submission/email-submission/email-submission.controller.ts
@@ -320,6 +320,8 @@ const submitEmailModeForm: ControllerHandler<
             message: 'Sending admin mail',
             meta: logMetaWithSubmission,
           })
+
+          // TODO 6395 make responseMetadata mandatory
           if (responseMetadata)
             statsdClient.distribution(
               'formsg.submissions.normResponseTimeMetadata',
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -436,7 +436,7 @@ const submitEncryptModeForm: ControllerHandler<
 
     if (responseMetadata)
       statsdClient.distribution(
-        'formsg.submissions.normalisedReponseTime',
+        'formsg.submissions.normResponseTimeMetadata',
         getNormalisedResponseTime(
           responseMetadata.responseTimeMs,
           responseMetadata.numVisibleFields,
@@ -599,7 +599,7 @@ const submitEncryptModeForm: ControllerHandler<
 
   if (responseMetadata)
     statsdClient.distribution(
-      'formsg.submissions.normalisedReponseTime',
+      'formsg.submissions.normResponseTimeMetadata',
       getNormalisedResponseTime(
         responseMetadata.responseTimeMs,
         responseMetadata.numVisibleFields,
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -22,6 +22,7 @@ import {
 import { StripePaymentMetadataDto } from '../../../../types'
 import { EncryptSubmissionDto } from '../../../../types/api'
 import config from '../../../config/config'
+import { statsdClient } from '../../../config/datadog-statsd-client'
 import { paymentConfig } from '../../../config/features/payment.config'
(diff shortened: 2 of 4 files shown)


src/types/api/email_submission.ts
```diff
@@ -27,6 +27,6 @@ export type ParsedEmailModeSubmissionBody = Merge<
   EmailModeSubmissionContentDto,
   {
     responses: ParsedEmailFormFieldResponse[]
-    responseMetadata: ResponseMetadata
+    responseMetadata?: ResponseMetadata
```
LinHuiqing, line 30: Any reason why we're not keeping this as required? Can probably make `responseMetadata` required in `validateEncryptSubmissionParams` of `src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts` as well, since all submissions should have `responseMetadata` since the previous release.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -579,6 +597,22 @@ const submitEncryptModeForm: ControllerHandler<
...
+    statsdClient.distribution(
+      'formsg.submissions.normalisedReponseTime',
+      getNormalisedResponseTime(
+        responseMetadata.responseTimeMs,
+        responseMetadata.numVisibleFields,
+      ),
+      1,
+      {
+        mode: 'encrypt',
+        payment: 'false',
```
LinHuiqing: I wonder if this is where we want to fire off to datadog instead of when the submission is firmly in the submissions collection.

Also what. Why is Datadog being annoying and not accepting the boolean as a value >:(

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -579,6 +597,22 @@ const submitEncryptModeForm: ControllerHandler<
     },
   })
 
+  if (responseMetadata)
+    statsdClient.distribution(
+      'formsg.submissions.normalisedReponseTime',
+      getNormalisedResponseTime(
+        responseMetadata.responseTimeMs,
```
LinHuiqing, line 606: Should have flagged this out earlier, but maybe this should be `normResponseTimeMetadata` instead. The current name is kinda misleading haha

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Eh sorry, should have tagged this to the `payment: 'true'` one instead. Also, is it better to hard-code these values as string or to do something like `false as unknown as string`?

src/types/api/email_submission.ts
foochifa: Yeah we should probably make them all required, but I think it should be a separate PR/issue!

As currently in FE we also have our Dtos (`SubmitStorageFormArgs` `SubmitEmailFormArgs`) as optional, reason being we used the same Dto for both preview and actual submissions 😅

For now, also to expedite the metrics, I think its better to keep them optional?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
foochifa: Thank you! [820993d](https://github.com/opengovsg/FormSG/pull/6387/commits/820993df264f1d6f4b2e7ce3c8afca537c001eb2)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
foochifa: Ah so like we still 'keep it as a boolean in code' but we type it into a string? HAHA like `payment: false as unknown as string`?

foochifa: Oh and for pt1 I think I fired the distribution stats after the submission is created? We already did `submission.save()` before it and logged it too! Line 572

LinHuiqing: > Oh and for pt1 I think I fired the distribution stats after the submission is created? We already did `submission.save()` before it and logged it too! Line 572

Oh I meant to tag the comment to the `payment: 'true'` line, meaning I was referring to if that block should be here when we are creating pendingSubmissions or when we are creating submissions proper.

LinHuiqing: > Ah so like we still 'keep it as a boolean in code' but we type it into a string? HAHA like `payment: false as unknown as string`?

Yea just wondering out loud which is better. Small thing haha

src/types/api/email_submission.ts
LinHuiqing: Okays sure, can you make it a `// TODO: `?

foochifa: Okie #6395 and [069a4c0](https://github.com/opengovsg/FormSG/pull/6387/commits/069a4c0e145d2e1a7456fe6c31ced6a0b383fcdd)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
foochifa: [069a4c0](https://github.com/opengovsg/FormSG/pull/6387/commits/069a4c0e145d2e1a7456fe6c31ced6a0b383fcdd) double assertion for the boolean thing

foochifa: Oh on the pending submission part I put in my consideration in the 'solution' portion of the PR description!

TLDR is that currently responseMetadata is for filling up of form -> once they submit pending submission, they finish filling form -> doesnt matter if they pay or not their data is still useful

src/types/api/email_submission.ts
LinHuiqing: Use linear eh haha

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Okays, makes sense 👍

LinHuiqing (approved):
lgtm!

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -317,6 +321,22 @@ const submitEmailModeForm: ControllerHandler<
...
+              'formsg.submissions.normResponseTimeMetadata',
+              getNormalisedResponseTime(
+                responseMetadata.responseTimeMs,
+                responseMetadata.numVisibleFields,
+              ),
+              1,
+              {
+                mode: 'email',
+                responseTimeMs: '${responseMetadata.responseTimeMs}',
+                numOfVisibleFields: '${responseMetadata.numVisibleFields}',
```
timotheeg, line 336: ⚠️⚠️ We cannot release this! Tag cardinality on DataDog metrics should stay low (less than 1000 per metric if possible). Tracking the responseTime in ms will explode cardinality and explode our bill.

Even `numOfVisibleFields` probably shouldn't be added as a tag given we know there are broad variety of form sizes, up to thousands of fields for crazy forms.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -432,6 +434,23 @@ const submitEncryptModeForm: ControllerHandler<
...
+        getNormalisedResponseTime(
+          responseMetadata.responseTimeMs,
+          responseMetadata.numVisibleFields,
+        ),
+        1,
+        {
+          mode: 'encrypt',
+          payment: true as unknown as string,
+          responseTimeMs: '${responseMetadata.responseTimeMs}',
+          numOfVisibleFields: '${responseMetadata.numVisibleFields}',
```
timotheeg, line 450: Same comment on tag cardinality

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -579,6 +598,23 @@ const submitEncryptModeForm: ControllerHandler<
...
+      getNormalisedResponseTime(
+        responseMetadata.responseTimeMs,
+        responseMetadata.numVisibleFields,
+      ),
+      1,
+      {
+        mode: 'encrypt',
+        payment: false as unknown as string,
+        responseTimeMs: '${responseMetadata.responseTimeMs}',
+        numOfVisibleFields: '${responseMetadata.numVisibleFields}',
```
timotheeg, line 614: And again 😅

timotheeg (commented):
⚠️⚠️⚠️ This PR should not be released as-is or it will explode our DD bill. ⚠️⚠️⚠️

As much as it is sup-obtimal, just tracking responseTime by broad groups (responseMode, and maybe agency as short name if we have it) is already good enough to get a sense of average, p90, p95 people spend filling forms on FormSG as a whole, 

What we should probably do here is:
* track a distribution metric for the actual response time
* track a second distribution metric for the normalized response time by field

In both of these, the response time and number of fields should *not* be reported as tags.

foochifa:
Thanks Tim! Sorry we missed this out, will update with another PR that provides metric for:
1. Response Time
2. Normalised Response Time

And will remove the cardinal tags! Leaving the mode of submission, and whether it is a payment form.

#6406

PR #6398, opened 2023-05-29
feat: couple payment field enablement with payment channel connection

LinHuiqing:
## Problem

Payment enabled toggle is confusing for admins and is an additional step for them when they have already connected to their Stripe accounts.

Closes FRM-760

## Solution

**Breaking Changes** 

- [x] Yes - this PR contains breaking changes.
    - Existing payment fields have `enabled` field, which will be sent as part of the request when admins are trying to save their edited payment field. This is throw an error as it is an unexpected field.
    - `scripts/20230529_rm_payment_enabled/script.js` should be run in the DB as soon as possible after the new version is deployed.

**Features**:

- Replace references to `payment_field.enabled` with checks of whether `payment_channel.channel` is unconnected or not. 
- Displaying payment preview only if:
   1. Payment Channel is connected
   2. Description is not an empty string
   3. Amount in cents to be collected is not 0

### Before Stripe account is connected

Form should look like a normal form without payments.

https://github.com/opengovsg/FormSG/assets/37061143/5f624a2e-2f06-4b7e-a64c-392f01a8f107

### After Stripe account is connected, before payment details are populated

Form should look like a normal form without payments.

https://github.com/opengovsg/FormSG/assets/37061143/ea8e5ec9-aeff-43b5-a660-74d0dccdf04c

### After payment details have been populated (and Stripe is connected)

Form should have payment preview and can make payment as usual.

https://github.com/opengovsg/FormSG/assets/37061143/a493237a-d450-4a1d-b515-7a8785f2f992

### When Stripe is disconnected

Form should be look like a normal form without payments. Previously populated description and payment amount should still be visible in the payment input panel.

https://github.com/opengovsg/FormSG/assets/37061143/96d0d98d-7470-4ad3-80bf-e69f9dc56f27

## Tests

- [ ] Before Stripe account is connected, form should look like a normal form without payments.
   - [ ] Builder view:
      - [ ] Should not have the Payment Preview.
   - [ ] Preview view:
      - [ ] Should not have the Payment Preview.
      - [ ] Submit button should say "Submit now".
   - [ ] Respondent view:
      - [ ] Should not have the Payment Preview.
      - [ ] Submit button should say "Submit now".

- [ ] After Stripe account is connected, before payment details are populated, form should look like a normal form without payments.
   - [ ] Connect the form to a Stripe account.
   - [ ] Builder view:
      - [ ] Should not have the Payment Preview.
      - [ ] Payment input panel should allow admin to key in the description and payment amount. (Note: don't save this first.)
   - [ ] Preview view:
      - [ ] Should not have the Payment Preview.
      - [ ] Submit button should say "Submit now".
   - [ ] Respondent view:
      - [ ] Should not have the Payment Preview.
      - [ ] Submit button should say "Submit now".

- [ ] After payment details have been populated (and Stripe is connected), form should have payment preview and can make payment as usual.
   - [ ] Go to the payment input panel, input the description and payment amount for the payment. Save it.
   - [ ] Builder view:
      - [ ] Should have the Payment Preview.
   - [ ] Preview view:
      - [ ] Should have the Payment Preview.
      - [ ] Submit button should say "Proceed to pay".
   - [ ] Respondent view:
      - [ ] Should have the Payment Preview.
      - [ ] Submit button should say "Proceed to pay".
      - [ ] Make a payment. The flow should be the same as regular payments before.

- [ ] When Stripe is disconnected, form should be look like a normal form without payments. Previously populated description and payment amount should still be visible in the payment input panel.
   - [ ] Disconnect the form from Stripe.
   - [ ] Builder view:
      - [ ] Should not have the Payment Preview.
      - [ ] Payment input panel should have the details previously saved, but should not allow admin to change the description and payment amount.
   - [ ] Preview view:
      - [ ] Should not have the Payment Preview.
      - [ ] Submit button should say "Submit now".
   - [ ] Respondent view:
      - [ ] Should not have the Payment Preview.
      - [ ] Submit button should say "Submit now".

## Deploy Notes

**New scripts**:

- `scripts/20230529_rm_payment_enabled/script.js`: removes `payments_field.enabled` from documents, should be run ASAP after deployment.

### DIFF
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -16,6 +16,7 @@ import get from 'lodash/get'
 import simplur from 'simplur'
 
 import { PAYMENT_CONTACT_FIELD_ID } from '~shared/constants'
+import { PaymentChannel } from '~shared/types'
 import {
   FormAuthType,
   FormResponseMode,
@@ -494,7 +495,7 @@ export const PublicFormProvider = ({
 
   const isPaymentEnabled =
     data?.form.responseMode === FormResponseMode.Encrypt &&
-    data.form.payments_field.enabled
+    data.form.payments_channel.channel !== PaymentChannel.Unconnected
 
   if (isNotFormId) {
     return <NotFoundErrorPage />
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -495,7 +495,9 @@ export const PublicFormProvider = ({
 
   const isPaymentEnabled =
     data?.form.responseMode === FormResponseMode.Encrypt &&
-    data.form.payments_channel.channel !== PaymentChannel.Unconnected
+    data.form.payments_channel.channel !== PaymentChannel.Unconnected &&
+    !!data.form.payments_field.amount_cents &&
+    !!data.form.payments_field.description
 
   if (isNotFormId) {
     return <NotFoundErrorPage />
--- a/frontend/src/features/public-form/components/FormFields/FormFields.tsx
+++ b/frontend/src/features/public-form/components/FormFields/FormFields.tsx
@@ -4,6 +4,7 @@ import { useSearchParams } from 'react-router-dom'
 import { Box, Stack } from '@chakra-ui/react'
 import { isEmpty, times } from 'lodash'
 
+import { PaymentChannel } from '~shared/types'
 import { BasicField, FormFieldDto } from '~shared/types/field'
 import { FormColorTheme, FormResponseMode, LogicDto } from '~shared/types/form'
 
@@ -164,7 +165,7 @@ export const FormFields = ({
           </Box>
         )}
         {form?.responseMode === FormResponseMode.Encrypt &&
-          form?.payments_field.enabled && (
+          form?.payments_channel.channel !== PaymentChannel.Unconnected && (
             <Box
               mt="2.5rem"
               bg="white"
--- a/frontend/src/features/public-form/components/FormFields/FormFields.tsx
+++ b/frontend/src/features/public-form/components/FormFields/FormFields.tsx
@@ -165,7 +165,9 @@ export const FormFields = ({
           </Box>
         )}
         {form?.responseMode === FormResponseMode.Encrypt &&
-          form?.payments_channel.channel !== PaymentChannel.Unconnected && (
+          form?.payments_channel.channel !== PaymentChannel.Unconnected &&
+          !!form.payments_field.amount_cents &&
+          !!form.payments_field.description && (
             <Box
               mt="2.5rem"
               bg="white"
--- a/frontend/src/features/public-form/components/FormFields/FormFields.tsx
+++ b/frontend/src/features/public-form/components/FormFields/FormFields.tsx
@@ -164,8 +164,9 @@ export const FormFields = ({
             </Stack>
           </Box>
         )}
-        {form?.responseMode === FormResponseMode.Encrypt &&
-          form?.payments_channel.channel !== PaymentChannel.Unconnected &&
+        {form &&
+          form.responseMode === FormResponseMode.Encrypt &&
+          form.payments_channel.channel !== PaymentChannel.Unconnected &&
           !!form.payments_field.amount_cents &&
           !!form.payments_field.description && (
             <Box
--- a/scripts/20230529_rm_payment_enabled/script.js
+++ b/scripts/20230529_rm_payment_enabled/script.js
@@ -1,0 +1,15 @@
+/* eslint-disable */
+
+// BEFORE
+// Count total number of forms with payment_fields that have enabled value. This should be > 0.
+db.getCollection('forms').countDocuments({ "payments_field.enabled": { $exists: true } })
+
+// UPDATE - remove payments_field.enabled
+db.getCollection('forms').updateMany(
+  {},
+  [ { $unset: { "payments_field.enabled": true } } ],
+)
+
+// AFTER
+// Count total number of forms with payment_fields that have enabled value. This should be 0.
+db.getCollection('forms').countDocuments({ "payments_field.enabled": { $exists: true } }) === 0
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -153,10 +153,6 @@ const EncryptedFormSchema = new Schema<IEncryptedFormSchema>({
   },
 
   payments_field: {
-    enabled: {
-      type: Boolean,
-      default: false,
-    },
     description: {
       type: String,
       trim: true,
@@ -208,9 +204,6 @@ EncryptedFormDocumentSchema.methods.removePaymentAccount = async function () {
     channel: PaymentChannel.Unconnected,
     target_account_id: '',
     publishable_key: '',
-  }
-  if (this.payments_field) {
-    this.payments_field.enabled = false
   }
   return this.save()
 }
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -349,10 +349,7 @@ const submitEncryptModeForm: ControllerHandler<
   }
 
   // Handle submissions for payments forms
-  if (
-    form.payments_field?.enabled &&
-    form.payments_channel.channel === PaymentChannel.Stripe
-  ) {
+  if (form.payments_channel.channel === PaymentChannel.Stripe) {
     /**
      * Start of Payment Forms Submission Flow
      */
--- a/src/app/modules/verification/verification.model.ts
+++ b/src/app/modules/verification/verification.model.ts
@@ -132,15 +132,17 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
   ): Promise<IVerificationSchema | null> {
     const formFields = getVerifiableFormFields(form.form_fields)
     if (form.responseMode === FormResponseMode.Encrypt) {
-      const { payments_channel } = form as IEncryptedFormSchema
-      const paymentField = getVerifiablePaymentContactField(
-        payments_channel.channel !== PaymentChannel.Unconnected,
+      const { payments_channel, payments_field } = form as IEncryptedFormSchema
+      const paymentContactField = getVerifiablePaymentContactField(
+        payments_channel.channel !== PaymentChannel.Unconnected &&
+          !!payments_field.amount_cents &&
+          !!payments_field.description,
       )
-      if (!formFields && !paymentField) return null
+      if (!formFields && !paymentContactField) return null
       return this.create({
         formId: form._id,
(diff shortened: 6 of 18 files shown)


scripts/20230529_rm_payment_enabled/script.js
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable */
+
+// BEFORE
+// Count total number of forms with payment_fields that have enabled value. This should be > 0.
+db.getCollection('forms').countDocuments({ "payments_field.enabled": { $exists: true } })
+
+// UPDATE - remove payments_field.enabled
+db.getCollection('forms').updateMany(
+  {},
```
timotheeg, line 9: Reduce the number of documents to update (even if the updates are inconsequential):
```suggestion
  { "payments_field.enabled": { $exists: true } },
```

Also just to confirm: running this script is just cleanup. It's inconsequential to the code, since we are removing awareness of the existence of the property, correct?

frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -164,7 +165,9 @@ export const FormFields = ({
           </Box>
         )}
         {form?.responseMode === FormResponseMode.Encrypt &&
-          form?.payments_field.enabled && (
+          form?.payments_channel.channel !== PaymentChannel.Unconnected &&
+          !!form.payments_field.amount_cents &&
+          !!form.payments_field.description && (
```
timotheeg: /nit1: the conditional descent on `form?.` being done inconsistently (first 2 checks versus second 2 checks) looks a little weird. In a case like this, I think it's better to have an explicit check for `form` first:

```suggestion
        {form && form.responseMode === FormResponseMode.Encrypt &&
          form.payments_channel.channel !== PaymentChannel.Unconnected &&
          !!form.payments_field.amount_cents &&
          !!form.payments_field.description && (
```

/nit2: this is the third instance where the code contains the 4 checks to determined `isPaymentEnabled`,  it probably warrants making a utility function for it.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -349,10 +349,7 @@ const submitEncryptModeForm: ControllerHandler<
   }
 
   // Handle submissions for payments forms
-  if (
-    form.payments_field?.enabled &&
-    form.payments_channel.channel === PaymentChannel.Stripe
-  ) {
+  if (form.payments_channel.channel === PaymentChannel.Stripe) {
```
timotheeg, line 352: This concerns me somewhat. Since we no longer have a singular `enabled` boolean, the payment channel can be connected, but amount and or description not set, and thus payment is actually not enabled on the form for the incoming submission, correct?

Basically I note a distinction between making the admins able to update all the payment fields (or not) in the form definition, independently and without a `enabled` toggle gating the fields; but to display the form to respondent, and validate incoming submissions, we need to be extra careful that payment is indeed enabled.

src/app/modules/verification/verification.model.ts
```diff
@@ -134,8 +132,10 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
   ): Promise<IVerificationSchema | null> {
     const formFields = getVerifiableFormFields(form.form_fields)
     if (form.responseMode === FormResponseMode.Encrypt) {
-      const { payments_field } = form as IEncryptedFormSchema
-      const paymentField = getVerifiablePaymentContactField(payments_field)
+      const { payments_channel } = form as IEncryptedFormSchema
+      const paymentField = getVerifiablePaymentContactField(
+        payments_channel.channel !== PaymentChannel.Unconnected,
```
timotheeg: Same comment as earlier: is this test enough to guarantee that payment is enabled on the form?

src/app/models/form.server.model.ts
```diff
@@ -153,10 +153,6 @@ const EncryptedFormSchema = new Schema<IEncryptedFormSchema>({
   },
 
   payments_field: {
```
foochifa, line 155: Now that `enabled` is no longer in `payments_field` and that `payments_field` is strongly tied to the existence of the field itself. We should actually make the field items `description` & `amount_cents` required as per the interaction in FE.

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -494,7 +495,9 @@ export const PublicFormProvider = ({
 
   const isPaymentEnabled =
     data?.form.responseMode === FormResponseMode.Encrypt &&
-    data.form.payments_field.enabled
+    data.form.payments_channel.channel !== PaymentChannel.Unconnected &&
+    !!data.form.payments_field.amount_cents &&
```
foochifa, line 499: If `payments_field` children are required, then over here we can just check if `payments_field` exists

foochifa (commented):
For note too, I think maybe we should have a way for the users to 'delete' the payment field? For instance, using a similar delete button from the field row button group. That will allow them to delete the field, instead of going to settings->payments->unconnect stripe

timotheeg (commented):
Nice works HuiQing!

I have some concerns on the checking that payment is really enabled on the respondent flows. Please help clarify! 😃

src/app/models/form.server.model.ts
LinHuiqing: No eh we can't, because then what will the values be before payment account connection?

frontend/src/features/public-form/PublicFormProvider.tsx
LinHuiqing: As mentioned above, we can't cos payments_field children can't be required?

scripts/20230529_rm_payment_enabled/script.js
LinHuiqing: Nope, validation will fail if we don't do this because when we try to save the payment field through the input panel, it takes the whole `payment_field`, updates the fields based on the form values, and tries to chuck it all back to BE to deal with it. So the BE will throw an err because of the unexpected `enabled` field.

frontend/src/features/public-form/components/FormFields/FormFields.tsx
LinHuiqing: Yup we should! Currently only available in the `usePublicForm` hook but we should probably shift it somewhere more accessible for everyone.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Yep, that's why there's the check everywhere for all the 3 requirements (connection, description, amount) 😪

src/app/models/form.server.model.ts
foochifa: Oh I mean `payments_field` itself being optional but `payments_field.description` and `payments_field.amount_cents` being required. Not sure if thats possible but previously @justynoh and I thought that was the cleaner way to go. So upon stripe connection -> `payments_field` still doesn't exist

LinHuiqing: Also, we don't want to display payments before they get a chance to edit the description and amount.

src/app/modules/verification/verification.model.ts
LinHuiqing: Hm okay true, description and amount will be a stronger guarantee for this. Will do

LinHuiqing: Done in [d1ba976](https://github.com/opengovsg/FormSG/pull/6398/commits/d1ba976a4491d34b39b417d43557f36b865fe73b)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Might be much better if this is assisted by a helper function?

src/app/models/form.server.model.ts
LinHuiqing: Wait can meh HAHA

scripts/20230529_rm_payment_enabled/script.js
timotheeg: Ah dang it! Thanks for the clarification @LinHuiqing 🌮 

So it makes the timing of the release a bit stressful. As soon as the new backend code is deployed, we need to run this script, or we'll be seeing errors when admins edit their forms. Since payment is in beta, I think we know the volume will be low enough to not have problems, but if we have to do this again when payment is more broadly  adopted, we will need to have a backward compatibility plan.

How interesting 😑 , a document db gives us limitless flexibility to have extra fields that we don't care about and move fast, but our strict typing in code is in the way of our velocity.

src/app/models/form.server.model.ts
LinHuiqing: Will push this off to another PR bah

scripts/20230529_rm_payment_enabled/script.js
LinHuiqing: Yea, I think its fine because payment features are still behind payment feature flag for now! So basically, we need to run this script to clear it then we can enabled the feature flag.

frontend/src/features/public-form/components/FormFields/FormFields.tsx
LinHuiqing: nit1 done in [27dd088](https://github.com/opengovsg/FormSG/pull/6398/commits/27dd0884622a0fa4c7e96d46b74bca5881040283)

src/app/modules/verification/verification.model.ts
```diff
@@ -115,10 +115,8 @@ const compileVerificationModel = (db: Mongoose): IVerificationModel => {
     return fields
   }
 
-  const getVerifiablePaymentContactField = (
-    paymentsField?: FormPaymentsField,
-  ) => {
-    if (!paymentsField?.enabled) return null
```
mayying, line 121: to ensure a smooth change of fields, we can do a 2-step roll out, by first introducing the new field, and keep the existing field. And at the 2nd step, we can remove the existing(deprecated) field entirely. In summary:

- current, existing prod version: FE sends supports paymentEnabled field, and BE receives and checks paymentEnabled field
- upcoming release (supports both fields, but make paymentEnable optional): FE sends both supports paymentEnabled field and paymentAccConnected field, and BE checks for optional paymentEnabled field and paymentAccConnected field (ie `paymentsField.enabled OR paymentsAccConnected)
- subsequent release (remove enable field entirely): FE sends only paymentAccConnected field, and BE still checks for only paymentsAccConnected field 

Doing so, we are not pressured to run the script, and *if* there is a need of rollback, we are still backward compatible. And it's a good practice for a smooth rollout.

LinHuiqing:
@mayying thanks for the comment! For this PR actually, we might be closing it depending on the user flow design (discussed during last sync, but it might have been hard to follow if you didn't have the context). 

Anyway, yea, we can break it up into separate releases for a smooth transition haha, just that in recent history we've been somewhat lazy and usually opt for releasing in the morning before admins start using our platform 😅 But maybe for a change this risky, we should break it up as you suggested!

foochifa:
Closing this PR in favour of #6511! Thanks @LinHuiqing

PR #6402, opened 2023-05-30
revert: "feat: inject normalised response time to DD"

timotheeg:
Reverts opengovsg/FormSG#6387

The tag cardinality would explode our metrics and bill usage on DataDog

### DIFF
--- a/src/app/modules/submission/email-submission/email-submission.controller.ts
+++ b/src/app/modules/submission/email-submission/email-submission.controller.ts
@@ -7,7 +7,6 @@ import {
 } from '../../../../../shared/types'
 import { IPopulatedEmailForm } from '../../../../types'
 import { ParsedEmailModeSubmissionBody } from '../../../../types/api'
-import { statsdClient } from '../../../config/datadog-statsd-client'
 import { createLoggerWithLabel } from '../../../config/logger'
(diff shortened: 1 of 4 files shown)


KenLSM (approved):

KenLSM:
For posterity, is it related to this? 

> A custom metric is uniquely identified by a combination of a metric name and tag values (including the host tag)
> ...
> The number of custom metrics from [COUNT](https://docs.datadoghq.com/metrics/types/?tab=count#metric-types) and [RATE](https://docs.datadoghq.com/metrics/types/?tab=rate#metric-types) is calculated with the same logic.
> 
> The number of unique tag value combinations submitted for a RATE metric with this tagging scheme is four:
> 
> host:A, endpoint:X, status:200
> host:B, endpoint:X, status:200
> host:B, endpoint:X, status:400
> host:B, endpoint:Y, status:200
> This results in request.Latency reporting four custom metrics.

Ref: https://docs.datadoghq.com/account_management/billing/custom_metrics/?tab=countrate

timotheeg:
Thanks @KenLSM , and yes exactly.

And note that we already had an incident with metrics explosion in the past. [PostMortem here](https://docs.google.com/document/d/1qYfQqBPsiwcapxKVNQdr92OQ1iKz6Z1viM2yW2-llUo/edit), so we want to avoid a repeat 😰

PR #6404, opened 2023-05-30
fix: use balance transaction fees from stripe

LinHuiqing:
## Problem

One of our agencies reported that there was a discrepancy between our reported transaction fee and the fees reported by Stripe. This is due to our calculation only including stripe fees, but there could be other fees such as tax (in the case of Paynow).

Closes FRM-955

## Solution

Take all fees without looking at the type. If admins want to check out the breakdown for fee types, the Stripe dashboard can be used.

**Breaking Changes** 

## Tests

- [ ] Enable both Card and Paynow as payment methods on a Stripe account.
- [ ] Create a storage-mode form on a payment-enabled admin account. Connect the form to the Stripe account above.
- [ ] Perform a form submission and payment with the Card payment method.
   - [ ] Check that the transaction fees on FormSG matches the one on Stripe.
- [ ] Perform a form submission and payment with the Paynow payment method.
   - [ ] Check that the transaction fees on FormSG matches the one on Stripe.

### DIFF
--- a/src/app/modules/payments/stripe.service.ts
+++ b/src/app/modules/payments/stripe.service.ts
@@ -118,21 +118,13 @@ const confirmStripePaymentPendingSubmission = (
           return new StripeFetchError()
         },
       )
-        .andThen((balanceTransaction) =>
-          okAsync(
-            balanceTransaction.fee_details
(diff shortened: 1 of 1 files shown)


KenLSM (approved):
LGTM!

PR #6414, opened 2023-05-31
fix: add payment field to numVisibleFields

foochifa:
## Problem

Currently, we do not include payment field in our computation for `responseMetadata.numVisibleFields`.

## Solution

If payment is enabled on the form, `handleSubmitForm` will increment the number of visible fields by 1. 

**Breaking Changes** 

## Tests

- [ ] make a payment on a form
- [ ] check the payload, the body should contain the correct number of visible fields including payments
- [ ] check the db is updated with the correct number of visible fields
- [ ] ensure that response time metrics are still visible in DD

### DIFF
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -135,6 +135,10 @@ export const PublicFormProvider = ({
   const { isNotFormId, toast, vfnToastIdRef, expiryInMs, ...commonFormValues } =
     useCommonFormProvider(formId)
 
+  const isPaymentEnabled =
+    data?.form.responseMode === FormResponseMode.Encrypt &&
+    data.form.payments_field.enabled
+
   useEffect(() => {
     if (data?.myInfoError) {
       toast({
@@ -231,7 +235,9 @@ export const PublicFormProvider = ({
         captchaResponse,
         responseMetadata: {
           responseTimeMs: differenceInMilliseconds(Date.now(), startTime),
-          numVisibleFields,
+          numVisibleFields: isPaymentEnabled
+            ? numVisibleFields + 1
+            : numVisibleFields,
         },
       }
 
@@ -474,6 +480,7 @@ export const PublicFormProvider = ({
       useFetchForSubmissions,
       numVisibleFields,
       startTime,
+      isPaymentEnabled,
     ],
   )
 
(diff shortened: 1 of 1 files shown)


foochifa:
Opened this for discussion:

1. I think payment field currently is equivalent to an email field with verification and hence should be tracked as a field, as it could significantly skew the response time especially for very small payment forms.
2. For payments by product, not sure if we want to normalise it instead (?) or also keep track of number of visible product chocies, but I think this a convo for the future.

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -231,7 +235,9 @@ export const PublicFormProvider = ({
         captchaResponse,
         responseMetadata: {
           responseTimeMs: differenceInMilliseconds(Date.now(), startTime),
-          numVisibleFields,
+          numVisibleFields: isPaymentEnabled
+            ? numVisibleFields + 1
+            : numVisibleFields,
```
timotheeg, line 240: Jut to note: this is a case where javascript automatic type conversion would work well (boolean to int: `true->1`, `false->0`), but it's not recommended for readability of course 😅 

```javascript
          numVisibleFields: numVisibleFields + isPaymentEnabled
```

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -135,6 +135,10 @@ export const PublicFormProvider = ({
   const { isNotFormId, toast, vfnToastIdRef, expiryInMs, ...commonFormValues } =
     useCommonFormProvider(formId)
 
+  const isPaymentEnabled =
+    data?.form.responseMode === FormResponseMode.Encrypt &&
+    data.form.payments_field.enabled
```
timotheeg, line 140: Ah dang it! This conflicts with the other PR that removes `payments_field.enabled`, and I still haven't had a chat with @staceytan1998 to decide on course of action for that 😰 . I suppose we could merge this first, and update the other PR later to make a change to this line if needed)

timotheeg (approved):

frontend/src/features/public-form/PublicFormProvider.tsx
foochifa: Yep it does 😅, we will likely have another method to check the existence of payment.

I think Justyn and I both prefer checking the existence of `payments_field` so probably be like 
```
const isPaymentEnabled = 
    data?.form.responseMode === FormResponseMode.Encrypt &&
    data.form.payments_field
```

But will leave it to after discussion with hq and stacey!

frontend/src/features/public-form/PublicFormProvider.tsx
foochifa: Omg this is real big brain! 

HAHAHA though typescript will complain, and I think we need to typecast 😅

I think I will keep it to the current implementation too cause with payments by product, we might wanna revisit how we calculate payment fields in our responseTime as well (so may not be `1`?)

Thanks @timotheeg!

PR #6418, opened 2023-06-01
feat(form-workspaces-be/7): add move forms from multiple source workspaces functionality

foochifa:
## Problem

Closes FRM-770

## Solution

See #4448 

Added additional test cases for `workspace.routes.ts`, happy path test cases for `workspace.server.model.ts` `workspace.service.ts`

Will add test cases for controller after merging develop to `form-v2/form-workspaces` due to deprecated method (ts-jest/utils mocked) that is currently being used.


foochifa:
Merging this into feature branch as the core functionality has already been reviewed in #4448, the only additions are test cases

PR #6436, opened 2023-06-07
fix: payout webhooks not returning OK

KenLSM:
## Problem

Stripe Payout webhooks are not returning with `200 OK` even though they are processed correctly.

Investigation yielded an error with the promise chain to be cyclic, causing the `StripeService.handleStripeEvent` to never end. This results in the the controller `handleStripeEventFromWebhook` to think that the handler is still running and not causing `res.send()` (or its equivalent) to trigger.

TS Playground Sandbox [example](https://www.typescriptlang.org/play?ssl=19&ssc=1&pln=20&pc=1#code/DYUwLgBATiDOCuxIF4IAUoHsC2BLWIAdDLJsAG4gAUA5JgNY0CUAUCwIawCeAdgMYQAZvH5hcmHhGztcPKkwDeLCNDiIU6LHgLE4ZSrQbNlKiITAALEHPkAuCOUy4AJhGQA+CEtOm+E2JCyPCBQGDj4IG6a4Tok+tQ0QSFh2iAAjMY+KiTqUTlI5lY2TG6eSaFaEaxZqmDwUJIA9I0QAO6RfOw8PJiQMHUNEF1clrIA5hBWMEOwQ-BgmGjsY+MAoux8FqoAjvC4JEMOTq799TwmKgC+TIXWVPKlqghIrCang-lgLJcs0rLyt2Kjz8PFIoEIwEwY1ofx4aQgfmwAAdQGAQM5mCVmhBgpQoAicCjwHA2CxUU91AAmKIpCK6MEGOiMVjk8q0ggAJTgSPsww43H4QhEfDEEikMh4lPk3gpSGpqHZRDiFASRmqpkB9yY9kcLkeMp8bMqnO5NONSr0KtoRpiIEpmSy7yaLXaCK6PT64DOQx4Iws40mIUinDmCyWKx4Y3Wmx2ewO7COeqdEAuEGumoeHggNtSXNgSNe2S9HzUcu+vwlUpuljumc8ILBREh0JosOpiKJaIxTCxLQ7qJJQA)

## Solution

Decouple setting of result of its async function over to another function.

**Breaking Changes** 

## Tests

Payout event from stripe should now return a 200 OK
- [ ] Find an trigger a webhook from stripe dashboard

### DIFF
--- a/src/app/modules/payments/__tests__/stripe.service.spec.ts
+++ b/src/app/modules/payments/__tests__/stripe.service.spec.ts
@@ -712,7 +712,6 @@ describe('stripe.service', () => {
       it('should return error result when call to stripe.balanceTransactions failed', async () => {
         const processStripeEventSpy = jest
           .spyOn(StripeService, 'processStripeEvent')
-          .mockImplementationOnce(() => errAsync(new PaymentNotFoundError()))
         const balanceTransactionApiSpy = jest
           .spyOn(stripe.balanceTransactions, 'list')
           .mockImplementationOnce(
--- a/src/app/modules/payments/__tests__/stripe.service.spec.ts
+++ b/src/app/modules/payments/__tests__/stripe.service.spec.ts
@@ -3,7 +3,7 @@ import dbHandler from '__tests__/unit/backend/helpers/jest-db'
 import { ObjectId } from 'bson'
 import { keyBy } from 'lodash'
 import mongoose from 'mongoose'
-import { errAsync, ok, ResultAsync } from 'neverthrow'
+import { err, errAsync, ok, ResultAsync } from 'neverthrow'
 import { PaymentChannel, PaymentStatus, SubmissionType } from 'shared/types'
 import Stripe from 'stripe'
 
@@ -21,6 +21,7 @@ import {
 
 import { PaymentNotFoundError } from '../payments.errors'
 import * as PaymentsService from '../payments.service'
+import { StripeMetadataInvalidError } from '../stripe.errors'
 import * as StripeService from '../stripe.service'
 import * as StripeUtils from '../stripe.utils'
 
@@ -707,11 +708,10 @@ describe('stripe.service', () => {
   })
 
   describe('handleStripeEvent', () => {
-    describe('event.type of payout', () => {
-      beforeEach(() => jest.resetAllMocks())
+    describe('with event.type of payout', () => {
+      beforeEach(() => jest.restoreAllMocks())
+
       it('should return error result when call to stripe.balanceTransactions failed', async () => {
-        const processStripeEventSpy = jest
-          .spyOn(StripeService, 'processStripeEvent')
         const balanceTransactionApiSpy = jest
           .spyOn(stripe.balanceTransactions, 'list')
           .mockImplementationOnce(
@@ -720,6 +720,10 @@ describe('stripe.service', () => {
                 autoPagingEach: () => Promise.reject('boom'),
               } as unknown as Stripe.ApiListPromise<Stripe.BalanceTransaction>),
           )
+        const processStripeEventSpy = jest.spyOn(
+          StripeService,
+          'processStripeEvent',
+        )
 
         const result = await StripeService.handleStripeEvent(
           MOCK_STRIPE_EVENTS_MAP['evt_PAYOUT_CREATED'],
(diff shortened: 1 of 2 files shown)


src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -704,4 +705,79 @@ describe('stripe.service', () => {
...
+      it('should return error result when call to stripe.balanceTransactions failed', async () => {
+        const processStripeEventSpy = jest
+          .spyOn(StripeService, 'processStripeEvent')
+          .mockImplementationOnce(() => errAsync(new PaymentNotFoundError()))
+        const balanceTransactionApiSpy = jest
+          .spyOn(stripe.balanceTransactions, 'list')
+          .mockImplementationOnce(
+            () =>
+              ({
+                autoPagingEach: () => Promise.reject('boom'),
```
LinHuiqing, line 720: Wait, is the 'boom' just a placeholder string?

LinHuiqing: Okay nvm, I see the 'still gud' below 🤪

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -704,4 +705,79 @@ describe('stripe.service', () => {
...
     })
   })
+
+  describe('handleStripeEvent', () => {
+    describe('event.type of payout', () => {
+      beforeEach(() => jest.resetAllMocks())
+      it('should return error result when call to stripe.balanceTransactions failed', async () => {
+        const processStripeEventSpy = jest
+          .spyOn(StripeService, 'processStripeEvent')
+          .mockImplementationOnce(() => errAsync(new PaymentNotFoundError()))
```
LinHuiqing:
```suggestion
```
I think for this part we shouldn't be mocking an errored return since this specific test is to test that `balanceTransationApiSpy` errors out, right?

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -704,4 +705,79 @@ describe('stripe.service', () => {
...
+                autoPagingEach: (fn) => fn({ type: 'charge', source: {} }),
+              } as unknown as Stripe.ApiListPromise<Stripe.BalanceTransaction>),
+          )
+
+        const result = await StripeService.handleStripeEvent(
+          MOCK_STRIPE_EVENTS_MAP['evt_PAYOUT_CREATED'],
+        )
+
+        expect(balanceTransactionApiSpy).toHaveBeenCalledOnce()
+        expect(processStripeEventSpy).not.toHaveBeenCalledOnce()
```
LinHuiqing, line 759: I'm a bit confused by this test - if the test is to check that error result is returned when 'processStripeEvent' fails, doesn't that mean that `processStripeEventSpy` should be called?

LinHuiqing (commented):
Sorry, a bit confused for the tests so I have some questions below! 

Also, I was thinking if it will be good for us to be more informative in ous `logger.error`s? since we know that that particular logger will only be used for payout events already (same goes for the `logger.error`s before).

src/app/modules/payments/__tests__/stripe.service.spec.ts
KenLSM: Great catch. It's should have been called.

KenLSM: Added in b919420

src/app/modules/payments/stripe.service.ts
```diff
@@ -556,12 +557,21 @@ export const handleStripeEvent = (
...
             // Reducer to keep errors around
-            result = result.andThen(() => innerResult)
+            if (innerResult && innerResult.isErr()) {
+              logger.error({
+                message:
+                  'Error when calling processStripeEvent while paging through balanceTransaction',
+                meta: { ...logMeta, chargeId: charge },
+                error: innerResult,
+              })
+              payoutProcessError = errAsync(innerResult.error)
```
mayying: since autoPagingEach will be iterating the balance transaction, wouldn't there be a chance that payoutProcesError gets overwritten? In that case, should we throw an error here and let the error handler handles it, instead of initialising a separate error variable?

KenLSM: > since autoPagingEach will be iterating the balance transaction, wouldn't there be a chance that payoutProcesError gets overwritten?

Yes. It's only capturing the last error, or rather, the existence of an error. It's not meant to keep track of how many errors.

> In that case, should we throw an error here and let the error handler handles it

Throwing would short-circuit the process early, I think there is merit to letting non-erroneous transaction complete. Wdyt?

> instead of initialising a separate error variable

In case this too implicit, the intent of creating a separate error variable is to escape the everlasting promise.

mayying: > Throwing would short-circuit the process early, I think there is merit to letting non-erroneous transaction complete. Wdyt?

Yeah, that makes sense. But, we just need to bear in mind that since we are not remembering which transaction we have already processed, in the event of retry by stripe, our webhook log would have some duplicated events, unless we want to add a logic to remove duplication. Otherwise we have to always be aware of the possibility of transaction being processed more than once when we add further computation towards the event that we received. 

Also, should we use mapErr instead after asyncAndThen?

KenLSM: Refactored to `mapErr` in ef187748

> our webhook log would have some duplicated events

We have an event ID check for duplicates in `processStripeEvent`

mayying: yes, you are right, it's being checked. Thanks for pointing it out. LGTM!

mayying (approved):

PR #6442, opened 2023-06-08
chore: delete deprecated angular routes

justynoh:
## Problem
Various backend routes are no longer in use since angular is gone. This PR cleans them up.

Closes FRM-988

## Solution
Delete deprecated api routes in favor of new api routes under the `api/v3` namespace.

**Breaking Changes** 

## Tests

Regression tests
- [ ] Log in to FormSG.
- [ ] Create and submit a form with email and mobile OTP verification.
- [ ] Duplicate a form using the template function.
- [ ] Log out of FormSG.


src/app/modules/verification/verification.controller.ts
```diff
@@ -457,7 +295,7 @@ export const handleOtpVerification = [
 ] as ControllerHandler[]
 
 /**
- * Handler for resetting the verification state of a field.
+ * Handler for resetting the verification state of a field for POST /forms/:formId/fieldverifications/:id/fields/:fieldId/reset
```
LinHuiqing, line 304: I think we can get rid of the `/reset` endpoints too actually, we don't actually use these...

LinHuiqing (approved):
Lgtm and tested! I think the otp `/reset` routes can be deprecated too, but will need to check datadog just in case (though I don't see how our users can use that independently). Maybe we can do it in a separate, smaller PR anyway, but should be quite fast bah.

src/app/modules/verification/verification.controller.ts
justynoh: I think you're right. The function `resetVerifiedField` in the FE does not seem to be used anywhere. In that case we should do an e2e teardown of this endpoint. Rather than expanding the scope of this PR, I think we can do it in another!

LinHuiqing: yes sir, let's goooo! merge when ready 👍  (after conflicts have been resolved)

LinHuiqing:
Oh we should have merged this before release today 😅 @timotheeg merge when ready? or any concerns?

timotheeg:
No Concerns. Let's go!

src/app/modules/verification/verification.controller.ts
LinHuiqing: Created linear issue FRM-1049 to track this!

PR #6452, opened 2023-06-12
feat: Create issue feedback UI on public form interface

mayying:
## Problem
Currently when respondents face any difficulties in submitting a form, they have no direct way to contact the form admin for help.

Closes FRM-998.

## Solution
Create a floating action button that pops up a modal to allow respondent to submit issue they encounter while filling form.

**Breaking Changes** 

**Features**:

- Display a sticky floating message button when respondent is filling in form (and also in preview mode).
- Upon clicking the button, a pop up modal should show up allowing user to report their issue. An optional choice of supplying their contact information (email) is provided in order for admin to contact them (if admin wishes to do so).

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
- Floating message button in the bottom right corner.
image
- Pop up modal upon clicking on the button.
image
- Upon clicking send report, a toast indicating successful submission is shown.
image

## Tests

- [ ] In public form, respondent can always see the issue feedback button (message icon) in the bottom right corner of the screen to report an issue.
- [ ] Create a long form and scroll through it, respondent should still be able to see the message button.
- [ ] When hover over the button, a `Report an issue` tooltip is displayed.
- [ ] Click on the button, a pop up modal should show up, allowing user to report any issue they wish.
- [ ] Respondent cannot submit the form if the description box is empty, a validation error is shown below the text box, stating `This field is required`.
- [ ] Respondent can submit the form if the contact box is empty.
- [ ] Respondent must key in a valid email (in term of format), otherwise a validation error will be shown below the text box, stating `Please enter a valid email`
- [ ] No email OTP verification is required for the email.
- [ ] Check that it's the same flow as in the preview form, except that upon submission of the feedback, a toast showing `Thank you for submitting your feedback! Since you are in preview mode, the feedback is not stored.`
- [ ] Validate all other pages (admin creation form, post submission form) do not have the issue feedback button.
- [ ] Validate that form admin receives an email notification (at most 1 form per form per day) saying respondents are facing issues with their form
- [ ] Validate that the form issues are correctly listed on the Response tab in admin page


frontend/src/features/public-form/components/FormIssueFeedback/FormIssueFeedbackContainer.tsx
```diff
@@ -0,0 +1,11 @@
+import { Box } from '@chakra-ui/react'
+
+import { FormIssueFeedbackButton } from './FormIssueFeedbackButton'
+
+export const FormIssueFeedbackContainer = (): JSX.Element | null => {
```
LinHuiqing:
```suggestion
export const FormIssueFeedbackContainer = (): JSX.Element => {
```
`null` type not required as the return type is constant here

frontend/src/features/public-form/components/FormIssueFeedback/FormIssueFeedbackContainer.tsx
```diff
@@ -0,0 +1,11 @@
...
+  return (
+    <Box py={{ base: '1.5rem', md: '2.5rem' }} w="100%">
```
LinHuiqing: Is there a need for this box wrapper? based on what I'm seeing on my local, it seems that it just adds an empty box at the bottom of the form, which I don't think is needed?

frontend/src/features/public-form/components/FormIssueFeedback/FormIssueFeedbackButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const FormIssueFeedbackButton = (): JSX.Element => {
+  const { isOpen, onOpen, onClose } = useDisclosure()
+  const { isPreview } = usePublicFormContext()
+
+  return (
+    <Flex
+      position="fixed"
+      bottom="2.625rem"
+      right="2.75rem"
+      cursor="pointer"
```
LinHuiqing: cute LOL (though probably not needed)

frontend/src/features/public-form/components/FormIssueFeedback/FormIssueFeedbackContainer.tsx
LinHuiqing: If you take a look at the UI changeset on chromatic, you can see that this adds a space to the end of all the form things (don't think this is intended?)

frontend/src/features/public-form/PublicFormPage.tsx
```diff
@@ -9,6 +9,7 @@ import FormFields from './components/FormFields'
 import { FormSectionsProvider } from './components/FormFields/FormSectionsContext'
 import { FormFooter } from './components/FormFooter'
 import FormInstructions from './components/FormInstructions'
+import FormPreSubmissionFeedback from './components/FormIssueFeedback'
```
LinHuiqing: Is this change in name intentional? Why not keep it as `FormIssueFeedback`?

LinHuiqing (changes requested):
Thanks for the PR! Generally lgtm!

However, I think this should go out _after_ BE endpoints are out, so we don't have lost feedback from users, like they give us feedback and expect us to follow up, but we can't cos this is pure FE. Other than that, only 2 main points to fix:

1. FormIssueFeedbackContainer adds a block of white space - can probably do away with the wrapper and use the button component directly.

2. There's a bug in error handling where all boxes are highlighted when there is an error in either field (refer to screenshot below). This should not be the case - only the field with the error should be highlighted.

Screenshot 2023-06-19 at 10 27 41 AM

frontend/src/features/public-form/components/FormIssueFeedback/FormIssueFeedbackButton.tsx
LinHuiqing: Oh just noticed that we do actually use the pointer cursor for form field editing! :O cool cool let's keep this hahaha

KenLSM: Just casually popping in, iirc our `.button` has a default `cursor: "pointer"`.

LinHuiqing: Ooh okay yea that'll be cleaner 👀 (can refer to the preview / add collborator button I think @mayying)

KenLSM: As discussed offline, most of the time, we want to associate the `cursor` property to the actual html element that handles the click event, as `cursor: "pointer"` implies to the user that this "area" is interactable.

In this particular situation, the child is the ones that handles the `clickEvents`. This may cause some dissonance in terms of the touchable area for the user if the child has a `margin`, etc.

We could look at removing `as="a"` from the `<IconButton />` so that it will behave as a `button` which inherently provides `cursor: "pointer"`.

frontend/src/features/public-form/PublicFormPage.tsx
mayying: Missed this one out while i was renaming, thx for pointing it out

frontend/src/features/public-form/components/FormIssueFeedback/FormIssueFeedbackButton.tsx
mayying: thanks for sharing this. wouldnt have noticed the diff by myself. updated it!

frontend/src/features/public-form/components/FormIssueFeedback/FormIssueFeedbackContainer.tsx
mayying: thanks for sharing the changeset, did not notice this cool feature before! updated them!

KenLSM (approved):
LGTM, pending a minor lint issue with `'isEmpty' is defined but never used`

LinHuiqing:
~~@mayying anything preventing this from being pushed to develop? Other than conflicts.~~

Eh ps commented this on the wrong PR. 😓

LinHuiqing (approved):
LGTM! Realised that I need to approve too to unblock oops

PR #6454, opened 2023-06-12
feat: (BE) create and update admin feedback 

foochifa:
## Problem

We want to get feedback from admins, but currently we have no place to store nor call it.

Closes FRM-991

## Solution

The feedback flow for admin will be:
1. How's your form building experience? (Thumbs Up/Down)
2. Would you like to tell us more? 
3. Free text input

They will be 3 separate modals as seen in the [figma here](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?type=design&node-id=17430-181452&t=uPnDE24Rc90y9puj-0)

Hence, the creation of the feedback document will likely only contain the rating.

The feedback comment will then be updated after the admin selects the modal

Therefore, we will have two routes, one at `/admin/forms/feedback` to create the feedback document with a compulsory rating (will also fire off datadog statsdclient with the rating). Then an additional route at `/admin/forms/feedback/:feedbackId` to update the comment of the feedback doc.

**Breaking Changes** 

- [ ] No - this PR is backwards compatible as it only contains an addition of a collection and 2 api routes

## Tests

Happy path
- [ ] Log in, and capture the auth cookie
- [ ] Use a tool to set up api calls (like postman)
- [ ] Send a post request to `/api/v3/admin/forms/feedback` with a req.body containing {rating: 0|1}
- [ ] The request should be successful and you should receive a success message + the admin feedback dto in the response
- [ ] Send a patch request to `/api/v3/admin/forms/feedback/:feedbackId` with a req.body containing {comment: '{something}'} and feedbackId of the feedback document from above
- [ ] The request should be successful and you should receive a success message


src/app/models/admin_feedback.server.model.ts
```diff
@@ -0,0 +1,59 @@
+import { Mongoose, Schema } from 'mongoose'
+
+import { IAdminFeedbackModel, IAdminFeedbackSchema } from '../../types'
+
+import { USER_SCHEMA_ID } from './user.server.model'
+
+export const ADMIN_FEEDBACK_SCHEMA_ID = 'AdminFeedback'
+export const ADMIN_FEEDBACK_COLLECTION_NAME = 'adminfeedback'
```
KenLSM: @mayying shall we start with recommending `snake_case`-ing collection names from this point onwards?

LinHuiqing: wait but the mongo collections nomenclature is all lowercase (i googled before i think)

mayying: as discussed, let's camel case our collection name

foochifa: As discussed, will be using `camelCase`! [b8c9de6](https://github.com/opengovsg/FormSG/pull/6454/commits/b8c9de66f9af9fcc6ac66858324f7d669727eed9)

src/app/models/admin_feedback.server.model.ts
```diff
@@ -0,0 +1,59 @@
...
+  IAdminFeedbackSchema,
+  IAdminFeedbackModel
+>(
+  {
+    userId: {
+      type: Schema.Types.ObjectId,
+      ref: USER_SCHEMA_ID,
+      required: true,
+    },
+    rating: {
```
mayying, line 20: this is translated from thumb up/thumb down right? i feel that we should use boolean instead? In future, if we convert the thumb up/thumb down to 5-star basis, we can't really reuse the field (at least not directly), as it will make the old rating looks like it's 0/5 or 1/5 stars, but actually it's supposed to be at a higher rating. Maybe rename it to something else, such as usefulVotes, instead of rating?

src/app/models/__tests__/admin_feedback.server.model.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+          created: expect.any(Date),
+          lastModified: expect.any(Date),
+        }),
+      )
+    })
+
+    it('should throw validation error when rating param is missing', async () => {
+      // Act
+      const actualPromise = new FeedbackModel(
+        omit(DEFAULT_PARAMS, 'rating'),
```
KenLSM: Should this be `const paramsWithoutRating = omit(DEFAULT_PARAMS, 'rating')`, following the above convention of `Arranging` before `Act`?

src/app/models/admin_feedback.server.model.ts
```diff
@@ -0,0 +1,59 @@
+import { Mongoose, Schema } from 'mongoose'
+
+import { IAdminFeedbackModel, IAdminFeedbackSchema } from '../../types'
+
+import { USER_SCHEMA_ID } from './user.server.model'
+
+// also acts as the model Id
```
KenLSM: /nit `// alias of model Id`

foochifa: 😅 [ab89573](https://github.com/opengovsg/FormSG/pull/6454/commits/ab895734fb40ccb94be84a3c90659bde87a801b0)

src/app/models/__tests__/admin_feedback.server.model.spec.ts
foochifa: Thanks for catching that ken! (clearly I blindly copy pasted from `feedback.server.model.spec.ts` HAHAH [ab89573](https://github.com/opengovsg/FormSG/pull/6454/commits/ab895734fb40ccb94be84a3c90659bde87a801b0)

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -95,3 +99,116 @@ export const handleSubmitFormFeedback = [
   validateSubmitFormFeedbackParams,
   submitFormFeedback,
 ] as ControllerHandler[]
+
+const valdiateSubmitAdminFeedbackParams = celebrate({
+  [Segments.BODY]: Joi.object().keys({
+    rating: Joi.number().min(0).max(1).cast('string').required(),
```
mayying: just wondering: the rating here is cast to string, but in schema it's defined as a number type, will the ultimate value be stored as string or number?

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -95,3 +99,116 @@ export const handleSubmitFormFeedback = [
...
+ */
+const submitAdminFeedback: ControllerHandler<
+  unknown,
+  { message: string; feedback: IAdminFeedbackSchema } | ErrorDto,
+  { rating: number; comment?: string }
+> = async (req, res) => {
+  const sessionUserId = (req.session as AuthedSessionData).user._id
+  const { rating, comment } = req.body
+
+  return UserService.getPopulatedUserById(sessionUserId)
```
mayying: do we need the populated info? I think `findUserById` should be sufficient?

src/app/modules/feedback/feedback.service.ts
```diff
@@ -151,3 +157,103 @@ export const hasNoPreviousFeedback = (
...
+
+/**
+ * Inserts given admin feedback to the database.
+ * @param userId the userId of the admin that provided the feedback
+ * @param rating the feedback rating to insert (0 for thumbs down, 1 for thumbs up)
+ * @param comment the feedback comment to insert if available
+ * @returns ok(IAdminFeedbackSchema) if successfully inserted
+ * @returns err(DatabaseError) on database error
+ */
+export const insertAdminFeedback = ({
```
KenLSM: I'm thinking if it makes sense to separate functions that are related to `admin-feedback` as having its own service. It doesn't seem like there's much in common other than having the same term `feedback`.

src/app/modules/feedback/feedback.service.ts
```diff
@@ -151,3 +157,103 @@ export const hasNoPreviousFeedback = (
...
+          feedbackId,
+        },
+        error,
+      })
+
+      return new DatabaseError('Admin feedback could not be found')
+    },
+  ).andThen((adminFeedback) => {
+    // Ensure that feedbackId and userId are the same as params
+    if (!adminFeedback || adminFeedback.userId.toString() !== userId)
```
KenLSM: /nit would prefer to split guard statements away. `!adminFeedback` ensures that the variable access is valid, while `adminFeedback.userId.toString !== userId` ensures that the user providing feedback is the same user who gave feedback.

src/app/modules/feedback/feedback.service.ts
```diff
@@ -151,3 +157,103 @@ export const hasNoPreviousFeedback = (
...
+          action: 'updateAdminFeedback',
+          feedbackId,
+        },
+        error,
+      })
+
+      return new DatabaseError('Admin feedback could not be found')
+    },
+  ).andThen((adminFeedback) => {
+    // Ensure that feedbackId and userId are the same as params
```
KenLSM: /nit this comment might be better worded to explain why we need to ensure `userId === adminFeedback.userId`.

```suggestion
    // Only allow updating of own feedback
```

src/app/modules/feedback/feedback.controller.ts
```diff
@@ -95,3 +99,116 @@ export const handleSubmitFormFeedback = [
...
+  { feedbackId: string },
+  { message: string } | ErrorDto,
+  { rating?: number; comment?: string }
+> = async (req, res) => {
+  const { feedbackId } = req.params
+  const sessionUserId = (req.session as AuthedSessionData).user._id
+
+  const { rating, comment } = req.body
+
+  return UserService.getPopulatedUserById(sessionUserId)
```
mayying: Can we update admin feedback directly since the user session is authenticated, and in `updateAdminFeedback` we are comparing if the db's feedback user's id === req's user id

src/app/modules/feedback/feedback.service.ts
```diff
@@ -151,3 +157,103 @@ export const hasNoPreviousFeedback = (
...
+  comment,
+  rating,
+}: {
+  feedbackId: string
+  userId: string
+  comment?: string
+  rating?: number
+}) => {
+  return ResultAsync.fromPromise(
+    AdminFeedbackModel.findById(feedbackId),
```
mayying: I believe this is not needed, since findByIdAndUpdate should be able to return error when document is not found?

src/app/models/admin_feedback.server.model.ts
foochifa: Yeah, I think kinda want to make the field extendible to the future if we want to expand the number rating. Do you think it will make more sense if we just allow rating to be 1-5 and then up -> 5, down -> 1? 

So I set it as 0|1 so that when we display our metric we can show it as a fraction of 'satisfaction', it currently effectively work as a boolean 😅

src/app/modules/feedback/feedback.service.ts
foochifa: Oh I got the original `adminFeedback` document so as to use its value of `rating` or `comment` in the case that it is not provided in the call as seen here in line 242/243

```
      AdminFeedbackModel.findByIdAndUpdate(feedbackId, {
        comment: comment ? comment : adminFeedback?.comment,
        rating: rating === undefined ? adminFeedback?.rating : rating,
      }),
```

foochifa: But yeah I think you are right, I probably can remove the second call to `findByIdAndUpdate` and instead use `update` followed by `save`. Not sure if there are any iops saving here though

src/app/models/admin_feedback.server.model.ts
mayying: We can of course map it that way, but depending on which stats we get (median/average) and how long has thumb up/thumb down been around, vs the new rating system (ie if thumb up/thumb down is around for 3 months, and rating system has been around for a day, the metric will look really bad if we have more thumb downs). My point is, the data may get skewed. 

I think it's safer to not reuse the field if we want more accurate data, and later on if we want to map it to a rating system, we can always run a one-time script to do the mapping from thumb up/thumb down -> rating system. But if we reuse the field directly, it will be harder for us to seperate between the two (to justify why the rating is not as good in a negative case scenario).

src/app/modules/feedback/feedback.service.ts
mayying: I think the logic can simplified as to updating the comment and rating whenever it exists in the request. It shouldn't matter if the comment and/or rating exists in the adminFeedback. Also, shall we update the comment and rating, instead of either or? What if the initial rating is good but the comment is empty, and later on an update request is made with negative sentiment comment and there is no new updated rating (since rating is optional in the update request). The data would look really weird in the database. We may need to put in some more thoughts on this.

src/app/modules/feedback/feedback.controller.ts
foochifa: It still gets casted at `number` in the controller handler but you are right the cast here doesn't do anything 😅 [548285f](https://github.com/opengovsg/FormSG/pull/6454/commits/548285f19f26a4226195b65c07f277b4c4a9dfec)

src/app/modules/feedback/feedback.service.ts
foochifa: Refactored to its own module [eb93502](https://github.com/opengovsg/FormSG/pull/6454/commits/eb9350216c394b8ebc259da973a6ffc10ca288bc)

src/app/modules/feedback/feedback.service.ts
foochifa: ~~refactored here too [eb93502](https://github.com/opengovsg/FormSG/pull/6454/commits/eb9350216c394b8ebc259da973a6ffc10ca288bc), and made a new error that returns `FORBIDDEN` when a user tries to modify a feedback that doesn't belong to him~~

src/app/modules/feedback/feedback.service.ts
foochifa: Refactored here! [eb93502](https://github.com/opengovsg/FormSG/pull/6454/commits/eb9350216c394b8ebc259da973a6ffc10ca288bc) though slightly different wording

src/app/modules/feedback/feedback.controller.ts
foochifa: As discussed, refactored away the userId check from `user` model instead will rely on the authed data [eb93502](https://github.com/opengovsg/FormSG/pull/6454/commits/eb9350216c394b8ebc259da973a6ffc10ca288bc)!

src/app/modules/feedback/feedback.controller.ts
foochifa: Refactored as discussed above too! [eb93502](https://github.com/opengovsg/FormSG/pull/6454/commits/eb9350216c394b8ebc259da973a6ffc10ca288bc)

src/app/modules/feedback/feedback.service.ts
foochifa: This is because the flow of the feedback process is like going to be:
1. Make the feedback doc with rating only
2. Update the feedback doc with comments

As seen in the figma [here](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?type=design&node-id=17430-181452&t=n2QYeoXab9QfORnr-0)!

But for extensibility purpose I generalised the update method to take in rating if required as well, hence, kept it as a 'general' method that will patch if body is passed in

src/app/models/admin_feedback.server.model.ts
foochifa: Have discussed with Stacey! it will most likely stick to 'Up' and 'Down' as its the most actionable rating, and simple for quick engagement by the admins too!

src/app/modules/feedback/feedback.service.ts
foochifa: Belay last! We will collapse both cases into 404 and log it instead. As instead of making it a two step process, we will just be doing one call with `updateOne` as discussed with @mayying. [4b7ae59](https://github.com/opengovsg/FormSG/pull/6454/commits/4b7ae59ea2a71f87c2f38ecdde3ceff04adf56cd)

src/app/modules/feedback/feedback.service.ts
foochifa: Have simplified the logic to only one db call as discussed! [4b7ae59](https://github.com/opengovsg/FormSG/pull/6454/commits/4b7ae59ea2a71f87c2f38ecdde3ceff04adf56cd)

src/app/models/__tests__/admin_feedback.server.model.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+          ...paramsWithoutComment,
+          created: expect.any(Date),
+          lastModified: expect.any(Date),
+        }),
+      )
+    })
+
+    it('should throw validation error when rating param is missing', async () => {
+      // Arrange
+      const paramsWithoutrating = omit(DEFAULT_PARAMS, 'rating')
```
KenLSM: /s, `paramsWithoutrating` -> `paramsWithoutRating`

src/app/models/admin_feedback.server.model.ts
KenLSM: Throwing my two cents into this debate.

I think `rating: Number` represents well onto what it is currently, and what it might be in the future. Ratings are usually represented as a numerical value (as opposed to binary) in most places. Thus, giving points for extensibility. A data fix, although valid, would come with risks that could have been mitigated.

Usage of `boolean` has its merits of being smaller and safer (no need to handle for overflow etc), but it is unlikely that this will be a concern for us. Within the context of FormSG, we don't have that much traffic to deal with, and shouldn't optimize it too early.

Regarding the issue of data skew, usually it is understandable that the data would be skewed, and therefore actions to ignore certain subset of the data is done.

As the suggested change doesn't affect critical the flow of FormSG, I suggest that we lean on having a bias for action and accept this minor imperfection to ship this.

KenLSM (approved):

src/app/models/__tests__/admin_feedback.server.model.spec.ts
foochifa: 😅 [58ceeb0](https://github.com/opengovsg/FormSG/pull/6454/commits/58ceeb093aba6ad3a56712745d73b13e8b6cd5e6)

src/app/modules/admin-feedback/admin-feedback.service.ts
```diff
@@ -0,0 +1,118 @@
...
+  const updateObj = { rating, comment }
+
+  // filter out undefined properties
+  Object.keys(updateObj).forEach(
+    (key) =>
+      updateObj[key as keyof typeof updateObj] === undefined &&
+      delete updateObj[key as keyof typeof updateObj],
+  )
+
+  // if not update to be done, return ok
```
mayying: /nit if no update to be done

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+
+      // Assert
+      expect(insertSpy).toHaveBeenCalledTimes(1)
+      expect(actualResult.isErr()).toEqual(true)
+      expect(actualResult._unsafeUnwrapErr()).toBeInstanceOf(DatabaseError)
+    })
+  })
+
+  describe('updateAdminFeedback', () => {
+    const MOCK_FEEDBACK_ID = new ObjectId().toHexString()
```
mayying, line 100: should we add spy to adminFeedbackModel?

src/app/modules/admin-feedback/admin-feedback.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      })
+
+      return new DatabaseError('Admin feedback could not be created')
+    },
+  )
+}
+
+/**
+ * Updates admin feedback in the database.
+ * Will use previous value if comment or rating are not passed into method
```
mayying: /nit technically it should be `will not update the previous value if comment or rating are not passed into method`?

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+        feedbackId: MOCK_FEEDBACK_ID,
+        userId: MOCK_USER_ID,
+        rating: newRating,
+        comment: newComment,
+      })
+
+      const newFeedback = await AdminFeedbackModel.findById(MOCK_FEEDBACK_ID)
+
+      // Assert
+      expect(actualResult.isOk()).toEqual(true)
```
mayying, line 136: shall we also unwrap the result to check the updated content?

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+    })
+    afterEach(() => jest.clearAllMocks())
+
+    const MOCK_USER_ID = new ObjectId().toHexString()
+    const MOCK_RATING = 1
+    const MOCK_COMMENT = 'mock comment'
+    const MOCK_ADMIN_FEEDBACK = new AdminFeedbackModel({
+      userId: MOCK_USER_ID,
+      rating: MOCK_RATING,
+      comment: MOCK_COMMENT,
```
mayying, line 30: just for my understanding, should there also be created and lastModified?

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
```diff
@@ -2435,4 +2437,219 @@ describe('admin-form.form.routes', () => {
       )
     })
   })
+
+  describe('POST /admin/forms/feedback', () => {
```
mayying, line 2441: shall we add the test case for 422 as well?

src/app/routes/api/v3/admin/forms/admin-forms.form.routes.ts
```diff
@@ -229,3 +230,34 @@ AdminFormsFormRouter.get(
...
+  /**
+   * Update an existing admin feedback
+   * @precondition Joi validation should enforce shape of req.body before this handler is invoked.
+   * @security session
+   *
+   * @returns 200 if feedback was successfully updated
+   * @returns 400 when Joi validation fails
+   * @returns 401 when user does not exist in session
+   * @returns 404 when admin feedback cannnot be found in the database
+   * @returns 422 when user of given id cannot be found in the database
```
mayying: is this still valid?

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
foochifa: created and lastModified are auto-created by mongoose!

src/app/routes/api/v3/admin/forms/admin-forms.form.routes.ts
foochifa: omg yes, its removed, thanks for pointing it out. [8405543](https://github.com/opengovsg/FormSG/pull/6454/commits/84055434476aa416b0af672dc402825f3c6ae14d) we no longer throw 422

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
foochifa: We no longer throw 422 as we rely on the authed session for userId, and will throw 404 if not found instead. 

Have updated the docs here [8405543](https://github.com/opengovsg/FormSG/pull/6454/commits/84055434476aa416b0af672dc402825f3c6ae14d)

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
foochifa: The actual result from the update doesn't returned the updated content. Hence I use `newFeedback` to re-retrieve the document from the model, and use that for comparison in line 137-138

src/app/modules/admin-feedback/admin-feedback.service.ts
foochifa: Thank you!! [8405543](https://github.com/opengovsg/FormSG/pull/6454/commits/84055434476aa416b0af672dc402825f3c6ae14d)

src/app/modules/admin-feedback/admin-feedback.service.ts
foochifa: Thank you 😅 lots of spelling error today HAHAH [8405543](https://github.com/opengovsg/FormSG/pull/6454/commits/84055434476aa416b0af672dc402825f3c6ae14d)

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
foochifa: I think there isn't a need? As we are able to verify the interaction with the model through the results. And for most of the update calls, I will subsequently make another call to retrieve the updated results, so if the updated result has the intended behaviour it should be robust enough

The spy doesn't really do much and just check for whether it is called 😅 and I don't want to mock it so as to test mongoose's static behaviour too

mayying: what bout the case where no comment and rating, dont we need to verify that adminFeedbackModel is not called?

src/app/routes/api/v3/admin/forms/__tests__/admin-forms.form.routes.spec.ts
mayying: ok ya. i was seeing 422 below, that's why im asking this, thx for updating it.

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
mayying: got it.

src/app/modules/admin-feedback/__tests__/admin_feedback.service.spec.ts
foochifa: Oh yes thats big brain thanks @mayying [6472c7d](https://github.com/opengovsg/FormSG/pull/6454/commits/6472c7d9b9d8f4290a0a0163e0e4fe7b09b78666)

mayying (approved):
Thanks for your patience! LGTM

PR #6455, opened 2023-06-12
fix: use 400 instead of 500 for gogov err status code

LinHuiqing:
## Problem

500s thrown are causing false alarms :'(

Works towards FRM-901

## Solution

Create a general custom error for GoGov integration related errors and use 400 for that.

**Breaking Changes** 

## Tests

- [ ] Go to a form that has not been attached to a GoGov Link yet and open the share form modal.
- [ ] Open the network panel to see the network calls made.
- [ ] Try to claim a GoGov link that you know is already taken on this form.
- [ ] The `/gogov` endpoint should return a `400` status rather than a `500` status code

## Deploy Notes

Not sure where to park these:
- [ ] remove exclusion of `/gogov` endpoint from [Server Errors (Overall) >= 1](https://app.datadoghq.com/monitors/79404515) DD monitor
- [ ] remove exclusion of `/gogov` endpoint from [Server Errors (Overall) >= 10](https://app.datadoghq.com/monitors/100352488) DD monitor

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -56,7 +56,6 @@ import * as SmsService from '../../../services/sms/sms.service'
 import { createReqMeta } from '../../../utils/request'
 import * as AuthService from '../../auth/auth.service'
 import {
-  ApplicationError,
   DatabaseConflictError,
   DatabaseError,
   DatabasePayloadSizeError,
@@ -88,7 +87,7 @@ import {
   PREVIEW_CORPPASS_UINFIN,
   PREVIEW_SINGPASS_UINFIN,
 } from './admin-form.constants'
-import { EditFieldError } from './admin-form.errors'
+import { EditFieldError, GoGovError } from './admin-form.errors'
 import { updateSettingsValidator } from './admin-form.middlewares'
 import * as AdminFormService from './admin-form.service'
 import { PermissionLevel } from './admin-form.types'
@@ -2775,7 +2774,7 @@ export const handleSetGoLinkSuffix: ControllerHandler<
             },
           ),
           // TODO: fix error handling (https://linear.app/ogp/issue/FRM-901/improve-error-handling-when-calling-gogov-api)
-          () => new ApplicationError('Error occurred when claiming GoGov link'),
+          () => new GoGovError(),
         )
       })
       // Step 3: After obtaining GoGov link, save it to the form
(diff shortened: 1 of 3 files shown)


src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -2775,7 +2774,7 @@ export const handleSetGoLinkSuffix: ControllerHandler<
             },
           ),
           // TODO: fix error handling (https://linear.app/ogp/issue/FRM-901/improve-error-handling-when-calling-gogov-api)
-          () => new ApplicationError('Error occurred when claiming GoGov link'),
+          () => new GoGovError(),
```
KenLSM, line 2777: Non-issue. To note that this might be swallowing real errors such as invalid credentials/server down etc. Best to map over some expected errors (if GoGov team publishes them).

KenLSM (approved):
LGTM. Added one small non-blocking note.

src/app/modules/form/admin-form/admin-form.controller.ts
mayying, line 2777: Since this PR is closing FRM-901, shall we break down the errors further here? https://guide.go.gov.sg/developer-guide/api-documentation#errors. 
200 -> 200
400 -> 400
429 -> 429
401/402/404/5xx -> 500

LinHuiqing: Yep yep, planning to get those for the actual PR tackling [FRM-901](https://linear.app/ogp/issue/FRM-901)

LinHuiqing: as mentioned by @mayying, they do have the errors here: https://guide.go.gov.sg/developer-guide/api-documentation#errors

LinHuiqing: Nope, this PR won't be closing [FRM-901](https://linear.app/ogp/issue/FRM-901), will do that in another PR. Also, should we be lumping 401/402/404/5xx -> 500?

KenLSM: I think yes for 401, and 404. Since those errors are something that we (FormSG -> GoGov) have a contract with. But for the 400, and 429, it's something that's a contract between User and FormSG.

`401`: is an invalid API key, which is something that's not fixable by the user. It is a config issue on FormSG BE.
`404`: is also a contract between FormSG and GoGov where it is FormSG that determines the resource.

`402`: Might be interesting. What do you guys think?

justynoh: imo `402` could map to `502`, since FormSG is making API calls to Go on behalf of the user and acting as a proxy

LinHuiqing: Makes sense

LinHuiqing: will do this in a separate PR!

PR #6463, opened 2023-06-14
feat: payment variable

KenLSM:
## Problem

Supporting Variable Payments

Closes FRM-1015, Closes FRM-987, Closes FRM-971

## Solution

**Breaking Changes** 

**Features**:

**DB Model Changes**
```diff
payments_field:
	enabled: Boolean
	description: String
	amount_cents: Number; validate is integer and non-negative
+	name: String;
+	type: enum(PaymentType)


+ enum PaymentType
	+ Fixed
	+ Variable
	+ Products [reserved for payment v3]
```

## Before & After Screenshots

| Fixed Payments | Variable Payments | Disabled Payments
|--------|--------|--------|
| Screenshot 2023-06-23 at 11 13 07 PM | variable |no payment |

Payment Dropdown Component
Screenshot 2023-06-23 at 11 12 50 PM

## Tests

**Regression Flow for original payments**
Responders able to make payment, admin response still functions
- [ ] As a responder
  - [ ] using an old / existing payment form
  - [ ] able to successfully make payment on form
- [ ] As an admin
  - [ ] able to see submission response on admin response result page

Admin able to save on old forms
- [ ] As an admin
  - [ ] login using an old / existing payment form as an admin
  - [ ] able to successfully add/edit/remove non-payment fields
  - [ ] able to successfully edit/remove payment fields
- [ ] As a responder
  - [ ] able to successfully make payment on form
- [ ] As an admin
  - [ ] able to see submission response on admin response result page

**New feature flow**
Fixed Payments
- [ ] create a fixed payment form
- [ ] able to successfully make payment on form
- [ ] able to see submission response on admin response result page

Variable Payments
- [ ] As an admin
  - [ ] create a variable payment form
  - [ ] set a min and max value
- [ ] As a responder
  - [ ] able to successfully make payment on form
- [ ] As an admin
  - [ ] able to see submission response on admin response result page
  - [ ] result should reflect the dynamic amount paid

Switching Payment Types
- [ ] As an admin
  - [ ] create a variable payment form
  - [ ] set a min and max value, description, name, and save the form
  - [ ] refresh the page, and switch payment type to Fixed
  - [ ] observe that amount field is reset, and description, name are still set.
  - [ ] able to save form
  - [ ] refresh the page, and switch payment type to Variable
  - [ ] observe that min and max value fields are reset, and description, name are still set.
  - [ ] able to save form
 

## Deploy Notes

Backwards compatibility is supported, but we should proactively run the DB migration to update the entries on DB.
Script to run `scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js`


scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable */
```
mayying, line 1: Since no index is being used, a full table scan will be activated to perform the filter and update, depending on the number of forms in production, and the spec of the db, it may cause performance degradation to prod user, and also the cron processes (such as webhook handling). suggest to add filter by _id, and do a for loop batch by batch (alternatively if the number of forms since payment is launched is not too much (say lesser than 10k?), we can skip the batch, and simply add a _id >= ObjectId(`time since first payment launched`) filter to avoid full table scan. let me know if u need help with the script.

src/app/modules/form/admin-form/admin-form.payments.controller.ts
```diff
@@ -309,25 +310,62 @@ export const _handleUpdatePayments: ControllerHandler<
...
+    enabled: Joi.boolean().required(),
+    payment_type: Joi.string()
+      .allow(...Object.values(PaymentType))
+      .required(),
+    amount_cents: Joi.when('enabled', {
+      is: Joi.equal(true),
+      then: Joi.when('payment_type', {
+        is: Joi.equal(PaymentType.Fixed),
+        then: Joi.number().integer().positive().required(),
+        otherwise: Joi.number(),
```
mayying: since the variable is amount_cents, should the variable payment input type be validated as integer? or can it still be float?

src/app/modules/form/admin-form/admin-form.payments.controller.ts
```diff
@@ -309,25 +310,62 @@ export const _handleUpdatePayments: ControllerHandler<
...
+      }),
+      otherwise: Joi.number().integer(),
+    }),
+
+    min_amount: Joi.when('enabled', {
+      is: Joi.equal(true),
+      then: Joi.when('payment_type', {
+        is: Joi.equal(PaymentType.Variable),
+        then: Joi.number().integer().positive().required(),
+        otherwise: Joi.number(),
```
mayying: likewise for this

src/app/modules/form/admin-form/admin-form.payments.controller.ts
```diff
@@ -309,25 +310,62 @@ export const _handleUpdatePayments: ControllerHandler<
...
+        is: Joi.equal(PaymentType.Variable),
+        then: Joi.number().integer().min(Joi.ref('min_amount')).required(),
+        otherwise: Joi.number(),
+      }),
+      otherwise: Joi.number().integer(),
+    }),
+
+    description: Joi.when('enabled', {
+      is: Joi.equal(true),
+      then: Joi.string().required(),
```
mayying: shall we add trim?

src/app/modules/form/admin-form/admin-form.payments.controller.ts
```diff
@@ -309,25 +310,62 @@ export const _handleUpdatePayments: ControllerHandler<
...
+      otherwise: Joi.string().allow(''),
+    }),
+    name: Joi.when('enabled', {
+      is: Joi.equal(true),
+      then: Joi.string().required(),
```
mayying: same here

src/app/modules/form/admin-form/admin-form.payments.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  formId: string,
+  newPayments: PaymentsUpdateDto,
+): ResultAsync<
+  IEncryptedFormDocument['payments_field'],
+  PossibleDatabaseError | FormNotFoundError | InvalidPaymentAmountError
+> => {
+  const { enabled } = newPayments
+
+  // Check if payment amount exceeds maxPaymentAmountCents or below minPaymentAmountCents if the payment is enabled
+  if (enabled && newPayments.payment_type === PaymentType.Fixed) {
```
mayying, line 36: it may be better to put validation as a separate function, so that testing can be modularised as well. Alternatively, is it possible to use joi for these validations? for e.g. (need to be tested)

```
amount_cents: Joi.when("enabled", {
      is: Joi.equal(true),
      then: Joi.when("payment_type", {
        is: Joi.equal(PaymentType.Fixed),
        then: Joi.number().integer().positive().required().min(paymentConfig.minPaymentAmountCents).max(paymentConfig.maxPaymentAmountCents),
        otherwise: Joi.when("payment_type", {
          is: Joi.equal(PaymentType.Variable),
          then:
            Joi.number().integer().positive().required().min(Joi.ref("min_amount")).max(Joi.ref("max_amount"))
        })
      }),
      otherwise: Joi.number().integer()
    }),
```

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -237,3 +240,26 @@ export const createEncryptedSubmissionDto = (
     version: submissionData.version,
   }
 }
+
+export const getPaymentAmount = (
+  formPaymentFields: FormPaymentsField, // fields that are from document.form
+  incomingSubmissionPaymentFields?: PaymentFieldsDto, // fields that are from incoming submission
```
mayying, line 274: do we also need to check if incomingSubmissionPaymentFields exists since the middleware does not specify that the field is required.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -361,7 +362,8 @@ const submitEncryptModeForm: ControllerHandler<
      * Start of Payment Forms Submission Flow
      */
     // Step 0: Perform validation checks
-    const amount = form.payments_field.amount_cents
+    const amount = getPaymentAmount(form.payments_field, req.body.payments)
+    //  .amount_cents
```
mayying: can this be removed?

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -237,3 +240,26 @@ export const createEncryptedSubmissionDto = (
...
+  switch (payment_type) {
+    case PaymentType.Fixed:
+      return formPaymentFields.amount_cents
+    case PaymentType.Variable:
+      return incomingSubmissionPaymentFields?.amount_cents
+
+    default: {
+      // Force TS to emit an error if the cases above are not exhaustive
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const exhaustiveCheck: never = payment_type
```
mayying, line 287: why not return empty like in line 249? what is the implication of forcing an error here?

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -237,3 +240,26 @@ export const createEncryptedSubmissionDto = (
     version: submissionData.version,
   }
 }
+
+export const getPaymentAmount = (
```
mayying, line 272: it may be better to add a description here, as there are multiple forms of return here

mayying, line 272: shall we add a UT for the new method?

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
KenLSM: The number of forms with payment is not significant enough.

But however, I need this exercise. Will ping you separately on how you would write such a batching job.

src/app/modules/form/admin-form/admin-form.payments.controller.ts
KenLSM: All `amount_cents`, `min_amount`, `max_amount` should be integers. 

The reason I'm not validating it, aside from ensuring it as `typeof number`, is that the value is not read/used if it is not `PaymentType.Fixed`.

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
mayying: The number of forms with payment does not really matter that much, rather is the total number of forms created since payment feature is launched, cuz if we just add _id filter, we need to know how many rows will be affected in the index scan.

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
KenLSM: Typescript would explicitly fail to compile if we missed out a case. By returning `empty` it would mean that the code can still compile, and server would fail during runtime.

src/app/modules/form/admin-form/admin-form.payments.controller.ts
KenLSM: Ah I see what you mean. The outer `otherwise` checks for `integer()` but inner `otherwise` doesn't. This can cause conflicts when the user switches from `enable` to `disable`.

KenLSM: 7b51943d3

src/app/modules/form/admin-form/admin-form.payments.controller.ts
KenLSM: 7b51943d3

src/app/modules/form/admin-form/admin-form.payments.controller.ts
KenLSM: 7b51943d3

src/app/modules/form/admin-form/admin-form.payments.controller.ts
KenLSM: d288736f8

src/app/modules/form/admin-form/admin-form.payments.service.ts
KenLSM: > it may be better to put validation as a separate function

Agreed. With `ByProducts` this will get way too chunky.

> Alternatively, is it possible to use joi for these validations? for e.g. (need to be tested)

I think `joi` is used for `routes` -joi-middleware-> `controller`. While services can be called by other services/controller.

frontend/src/components/MoneyInput/MoneyInput.tsx
```diff
@@ -92,17 +81,7 @@ export const MoneyInput = forwardRef<MoneyInputProps, 'input'>(
...
-                // Show Flags if available. If value does not exist for any reason,
-                // a default fallback icon will be used by ChakraUI.
-                // See https://chakra-ui.com/docs/media-and-icons/icon#fallback-icon.
-                as={Flags[country]}
-                role="img"
-                __css={styles.icon}
-              />
-              <Text>{currency[country]}</Text>
-            </HStack>
+            <Text>S$</Text>
```
foochifa, line 84: Are we going to hard-set just SGD going forward?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/PaymentsInputPanel.tsx
```diff
@@ -1,73 +1,64 @@
...
-}).format
+import { FixedPaymentAmountField } from './FixedPaymentAmountField'
+import { VariablePaymentAmountField } from './VariablePaymentAmountField'
 
-type FormPaymentsInput = {
-  enabled: boolean
-  description: string
+export type FormPaymentsInput = Omit<FormPaymentsField, 'amount_cents'> & {
   display_amount: string
+  display_min_amount: string
```
foochifa, line 60: On type safety, `FormPaymentsInput` here is going to expect both `display_amount` and `display_min_amount` & `dispaly_max_amount` but that's not true right? 

it should be conditioned on whether its variable or fixed payment?

src/app/modules/form/admin-form/admin-form.payments.service.ts
mayying: Sorry for the confusion, what I meant is to move the validation to the route validation portion, instead of adding another layer of validation in this function.

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
mayying: how would server fail it though? abruptly with exit(1)? or by throwing error? I think we can do graceful exit in this case?

KenLSM: As in the behaviour ~would~could be undefined (or rather don't really want to think about it). If we could detect that such potential errors could exist, we can should try to handle them. Similar to most statically-typed languages forces you to account for all cases before allowing one to emit an executable.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/PaymentsInputPanel.tsx
```diff
@@ -196,63 +160,99 @@ export const PaymentInput = ({ isDisabled }: { isDisabled: boolean }) => {
...
-        <FormErrorMessage>{errors.display_amount?.message}</FormErrorMessage>
-        {Number(clonedWatchedInputs.display_amount) > 1000 ? (
-          <InlineMessage variant="warning" mt="2rem" useMarkdown>
-            You would need to issue your own invoice for amounts above S$1000.
-            [Learn more about
-            this](https://guide.form.gov.sg/faq/faq/payments#simplified-tax-invoices-versus-regular-tax-invoices)
-          </InlineMessage>
-        ) : null}
-      </FormControl>
+      {isUsingPayment && (
```
foochifa, line 181: Shouldn't `isUsingPayment` here encompass the entire block of payments components from `Payment type` all the way to the `VariablePaymentAmountField | FixedPaymentAmountField`? Else currently we are doing 4 `isUsingPayment` checks

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/FixedPaymentAmountField.tsx
```diff
@@ -0,0 +1,70 @@
...
+        )}
+      />
+      <FormErrorMessage>
+        {errors[DISPLAY_AMOUNT_FIELD_KEY]?.message}
+      </FormErrorMessage>
+      {Number(input[DISPLAY_AMOUNT_FIELD_KEY]) > 1000 ? (
+        <InlineMessage variant="warning" mt="2rem" useMarkdown>
+          You would need to issue your own invoice for amounts above S$1000.
+          [Learn more about
+          this](https://guide.form.gov.sg/faq/faq/payments#simplified-tax-invoices-versus-regular-tax-invoices)
```
foochifa: would it be good to use a go-gov link here? -> in the future if there are any changes we can just change the underlying link

src/app/modules/form/admin-form/admin-form.payments.service.ts
KenLSM: It is already validated at the routes [here](https://github.com/opengovsg/FormSG/pull/6463/files#diff-f46e4b758b65d2a6eed5f926a370dcbdbd3bf032308ffae0ecd0e45600f529d2R315).

This is a service-level check where the intention is to further check on config-level validators (`paymentConfig.minPaymentAmountCents`) which is out of `joi` and `db` role

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
KenLSM: It's checked in 
`incomingSubmissionPaymentFields?.amount_cents` or am I misunderstanding something?

src/app/modules/form/admin-form/admin-form.payments.service.ts
mayying: i was wondering if we could add all checks, in the joi validation, as it may cause confusion as to why there're various places for input validation checks. Also, at the very minimum, i think the min/max amount check in line 48 can be moved to joi validation?

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
mayying: got it, thanks!

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
mayying: Hmm. with the current flow, when incomingSubmissionPaymentFields is empty, we will return 500 status code, but that should not be the case as this field comes from api request, right? In that case, we may need to do a separate validation for an early return with 400, please let me know if i misunderstanding anything 😅

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
foochifa: +1 on this! 

Just went to try filtering by _id> objectId something

```
db.getCollection("forms")
.find({ 
    '_id': {$gt: ObjectId('some Id')},
    "payments_field": { $exists: true }
    }).count()

```

But doesn't seem to have a performance gain

there are a few [articles](https://studio3t.com/knowledge-base/articles/optimize-mongodb-queries/) about optimizing query performance. Maybe we have to create an index to preliminary filter by id/date before running the payment_field find query

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
KenLSM: Only for `PaymentType.Variable` then we will need `incomingSubmissionPaymentFields` to be present. Thus, we can't do an early return.

mayying: hmmm. i see, but is it correct that we will be returning 500 status in this case? is that expected?

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/PaymentBlock.tsx
```diff
@@ -162,16 +163,13 @@ export const StripePaymentBlock = ({
           <Text textStyle="h3" textColor="primary.500" mb="1rem">
             Payment
           </Text>
-          <PaymentItemDetailsBlock
+          <FixedPaymentItemDetailsBlock
```
foochifa: Was wondering if we can document this better, agreed it makes sense to re-use the fixed payment block for both variable payment and fixed payment in payment page, as by that time the payment amount is 'fixed'. 

But cause at first I got confused at this line (wondering why there wasn't a separate one for variable 😅)

frontend/src/features/public-form/components/FormPaymentPage/queries.ts
```diff
@@ -34,3 +39,61 @@ export const useGetPaymentInfo = (
     { suspense: true },
   )
 }
+
+const formatCurrency = new Intl.NumberFormat('en-SG', {
```
foochifa: Hmm am wondering if this block here including the validation should be in utils instead of `queries.ts`? as we usually use our queries files for useQuery right? which mostly implies api service calls

frontend/src/features/public-form/components/FormPaymentPage/components/types.ts
```diff
@@ -0,0 +1,16 @@
+import { FormColorTheme } from '~shared/types'
+
+import { PaymentItemNameDescriptionProps } from './PaymentItemNameDescription'
```
foochifa: Should payment item name props just be collapsed into this type file? Then this file will serve as the types/props file for our payment item components

frontend/src/components/Dropdown/components/SelectMenu.tsx
```diff
@@ -9,19 +9,63 @@ import { itemToValue } from '../utils/itemUtils'
...
+const NothingFoundItem = () => {
+  const { styles, nothingFoundLabel } = useSelectContext()
+  return (
+    <ListItem role="option" sx={styles.emptyItem}>
+      {nothingFoundLabel}
+    </ListItem>
+  )
+}
+
+const RenderIfOpen = ({
```
foochifa, line 38: is this needed 😅 cause we can just {isOpen ? ... : null} below right? or is it for readability sake?

foochifa (commented):
Thanks @mayying for reviewing too! Generally looked at the FE as mayying has already commented on the BE

Mostly nits and just clarifying some assumptions. 

have tested locally and looks good, ~~but did run into some webhook receiving issues~~ (whitespace in my keys 😩)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/PaymentsInputPanel.tsx
KenLSM: Good question. I had initially done that as well, but it was due to how `React.Fragment` works together with component's `children`.

In the parent component, `<CreatePageDrawerContentContainer />`, it checks for it's child to determine whether there should be a `<Divider />` inserted.

If we are encompass the whole payments block with a singular `isUsingPayment`, we would need to wrap the 4 child components under a singular `< />` 
```
<CreatePageDrawerContentContainer>
 {isUsingPayment && <>{...fourChildComponents}
</CreatePageDrawerContentContainer>
```

This will result in no `<Divider />` inserted between its children.

frontend/src/components/Dropdown/components/SelectMenu.tsx
KenLSM: Purely for readability sake. Having too many ternary operations in JSX is difficult to read. The current code is already having 2 nested ternary operations, without `RenderIfOpen` we would get 3-levels deep.

frontend/src/components/MoneyInput/MoneyInput.tsx
KenLSM: Yeah.

frontend/src/features/public-form/components/FormPaymentPage/queries.ts
KenLSM: Great catch. I think it is wrongly placed. Have separated it to `hooks/` and `utils/`

b16ee46ca

frontend/src/features/public-form/components/FormPaymentPage/components/types.ts
KenLSM: Agreed. 700d1fc6481c4055e65f081f4d1f2bfb041ce7ed

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/PaymentsInputPanel.tsx
foochifa: Oh yeah... Hmm okay this should be transient issue right? Because of FRM-1016

After that the components should always be there less the switching between fixed and variable

KenLSM: Yeah. Didn't want to really spend too much time writing a solution that will be eventually resolved xD.

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
KenLSM: Added! 793abffb1933385ed6cbde30a5a5e92c518922d1

KenLSM: Added some description and return type 793abffb1933385ed6cbde30a5a5e92c518922d1

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
KenLSM: 🙈 793abffb1933385ed6cbde30a5a5e92c518922d1

src/app/modules/form/admin-form/admin-form.payments.service.ts
KenLSM: Added here 6bf2024b6. This suggestion made the error response look really good 👍 

**Before**
```
{
    "message":"Invalid payment amount"
}
```

**After**
```
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "Validation failed",
    "validation": {
        "body": {
            "source": "body",
            "keys": [
                "amount_cents"
            ],
            "message": "\"amount_cents\" must be less than or equal to 100000000"
        }
    }
}
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/PaymentsInputPanel.tsx
KenLSM: I've updated the type instead, 494c2b895. BE actually still returns the information but some of the fields are not used depending on the `PaymentType`. Lmk if this makes sense.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/FixedPaymentAmountField.tsx
KenLSM: dc043e009a9a9f3a1ce3c299cf0f9f177c619081

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/PaymentsInputPanel.tsx
foochifa: Ahh okay mongoose will still return the field as empty? Okay I think this is fine 😅 thanks @KenLSM

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
mayying: > But doesn't seem to have a performance gain

Can you share your exact query and how you came to the conclusion where there is no performance gain? _id is indexed by default, you should be getting a table scan. 🤔 
Explain result with _id field
image
Explain result without _id field
image

shared/types/form/form.ts
```diff
@@ -76,12 +76,31 @@ export type FormPaymentsChannel = {
...
 }
 
-export type FormPaymentsField = {
-  enabled: boolean
+export interface PaymentTypeBase {
+  payment_type: PaymentType
+
   amount_cents?: number
-  description?: string
+  min_amount?: number
```
mayying, line 83: are amount_cents, min_amount, max_amount still needed given they are enforced in the subclass?

src/app/modules/form/admin-form/admin-form.payments.service.ts
mayying: should we remove this part then? since it's already in joi validation now

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -48,6 +48,9 @@ export const validateEncryptSubmissionParams = celebrate({
       )
       .optional(),
     paymentReceiptEmail: Joi.string(),
+    payments: Joi.object({
+      amount_cents: Joi.number().integer().positive(),
```
mayying: just curious, do we need to verify the amount limit range?

src/app/modules/form/admin-form/admin-form.payments.service.ts
KenLSM: The service should still check, as it can be called by other services/controller

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
KenLSM: Great catch. I think it makes sense to have config min+max check here. dad1b3ba6

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
KenLSM: I think it makes sense to be server-side error, since the user isn't able to fix this. Suggesting that this is highly likely a coding bug on the server side.

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
foochifa: A few experiments using `.explain('executionstats')` (find without count) in MongoDB staging cluster (229792 total forms)


`db.getCollection("forms").find({}).explain('executionStats')`

1993ms, 229792 returned, 229792 examined, 12 execution stages

`db.getCollection("forms").find({"payments_field": { $exists: true }}).explain('executionStats')` 

2136ms, 57 returned, 229792 examined, 13 execution stages

`db.getCollection("forms").find({'_id': {$gt: ObjectId('5d1fa36ca2f1698500af5fcd')},
    "payments_field": { $exists: true }}).explain("executionStats")`

2294ms, 57 returned, 198856 examined, 14 execution stages

`db.getCollection("forms").find({'_id': {$gt: ObjectId('5f1fa36ca2f1698500af5fcd')},
    "payments_field": { $exists: true }}).explain("executionStats")`

1048ms, 57 returned, 104709 examined, 14 execution stages


So @mayying is right! Just that we need to select the 'break point' properly else it will end up being a little slower due to more executions!

(previously I probably used too tiny of an objectId 😅, so the execution time was similar)

foochifa:
The FE part LGTM now! But will let @mayying approve as there are still ongoing BE discussions 🙏

mayying:
Overall LGTM, thanks for making the changes, the only part left should be the db script.

shared/types/form/form.ts
KenLSM: Yes. This is to do type narrowing once we know which subclass the branch is on.

i.e., when we ensure that `payment_type === PaymentType.Fixed`, then the type of `amount_cents` will be narrowed from `amount_cents?: number` to `amount_cents: number`. 

The reason why I wanted to keep them as optional, if the types are not yet narrowed, is that our BE actually still passes these values around.

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
```diff
@@ -4,31 +4,87 @@
 Add PaymentType.Fixed to all forms with existing payments_field object
 */
 
-// Count number of forms that contains payments_field property
-db.getCollection('forms')
-  .find({ payments_field: { $exists: true } })
-  .count()
+// Batch 1 2022
+const queriesModifiedIn2022 = {
+  payments_field: { $exists: true },
```
foochifa: Is there a need for this check when line 34 already contains the `payments_field` check?

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
```diff
@@ -4,31 +4,87 @@
...
+const queriesModifiedIn2023AprMay = {
+  lastModified: { $gte: new ISODate('2023-04-01T00:00:00.000+0000') },
+  lastModified: { $lt: new ISODate('2023-06-01T00:00:00.000+0000') },
+}
+
+const queriesModifiedIn2023JunOnwards = {
+  lastModified: { $gte: new ISODate('2023-06-01T00:00:00.000+0000') },
+}
+
+function executePaymentTypeInsertionOnQuery(dateQuery, dryRun = true) {
```
foochifa, line 33: Usually on our shell we will 'step' into each of the query right, but are we able to do that when we wrap it in a function? 

Or is that why we have the `dryRun` break 😅

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
```diff
@@ -4,31 +4,87 @@
...
 */
 
-// Count number of forms that contains payments_field property
-db.getCollection('forms')
-  .find({ payments_field: { $exists: true } })
-  .count()
+// Batch 1 2022
+const queriesModifiedIn2022 = {
+  payments_field: { $exists: true },
+  lastModified: { $gte: new ISODate('2022-01-01T00:00:00.000+0000') },
```
foochifa: /nit should we combine the two [checks](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/?_ga=2.237881088.1020962104.1688274645-959180788.1669692925#query-for-ranges)?

`lastModified: { $gte: new ISODate(), $lt: new ISODate() }`

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
```diff
@@ -4,31 +4,87 @@
...
+}
+
+function executePaymentTypeInsertionOnQuery(dateQuery, dryRun = true) {
+  console.log(dateQuery)
+  // Count number of forms that contains payments_field property
+  const formsWithPaymentsField = db
+    .getCollection('forms')
+    .find({
+      payments_field: { $exists: true },
+      ...dateQuery,
```
foochifa, line 40: `lastModified` isn't indexed right? Would this affect the query time 

pinging @mayying

foochifa (commented):
Just some nits on the db script!

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
KenLSM: Yeah, that's what I'm intending the `dryRun` to be. I was thinking that we have quite a few batched queries to run, and copy/pasting might be a point of risk / error.

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
KenLSM: Ooops, should be removed! Thanks for catching

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
KenLSM: Oh, thanks for the tip! Yes that looks terser.

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
foochifa: Oh I mean we use F6 right? at least for studio 3T

So like you copy paste the entire script then F6 all the way down 😅

image

But I think the `dryRun` method is fine too! As it separates the check and the execution which is the main focus anyway hahaha

KenLSM: I... actually didn't know of this F6 on Studio 3T 🙈. I think it makes sense to use that.

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
mayying: yeah, this query will not use index.

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
```diff
@@ -4,31 +4,87 @@
...
+  lastModified: { $gte: new ISODate('2023-06-01T00:00:00.000+0000') },
+}
+
+function executePaymentTypeInsertionOnQuery(dateQuery, dryRun = true) {
+  console.log(dateQuery)
+  // Count number of forms that contains payments_field property
+  const formsWithPaymentsField = db
+    .getCollection('forms')
+    .find({
+      payments_field: { $exists: true },
```
mayying, line 39: probably should add 'payments_field.payment_type': {$ne: 'Fixed'} too, to avoid false positive?

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
```diff
@@ -4,31 +4,87 @@
 Add PaymentType.Fixed to all forms with existing payments_field object
 */
 
-// Count number of forms that contains payments_field property
-db.getCollection('forms')
-  .find({ payments_field: { $exists: true } })
-  .count()
+// Batch 1 2022
```
mayying, line 7: Since all old forms can also be added with payments, it's best to scan the whole collection without causing performance issue. To do so, we need to utilise index which is `_id`, and we need to do a batch read/write. I write a script which you can refer to.

```
let startDate = toObjectId(new Date('2021-12-31T23:59:59.999+0800'))
const endDate = toObjectId(new Date())
const batchSize = 1000
let toUpdates = []
let totalUpdates = 0

function query(startDate, endDate) {
  return db
    .getCollection('forms')
    .find({
      payments_field: { $exists: true },
      'payments_field.payment_type': { $ne: 'Fixed' },
      _id: {
        $gt: startDate,
        $lte: endDate,
      },
    })
    .projection({ _id: 1 })
    .sort({ _id: 1 })
    .limit(batchSize)
}

function update() {
  if (toUpdates.length === 0) {
    console.log('nothing to update')
    return
  }
  const ids = toUpdates.map(function (doc) {
    return doc._id
  })
  const updateRes = db.getCollection('forms').updateMany(
    {
      _id: { $in: ids },
      payments_field: { $exists: true },
      'payments_field.payment_type': { $ne: 'Fixed' },
    },
    {
      $set: {
        'payments_field.payment_type': 'Fixed',
      },
    },
  )
  toUpdates = []
  console.log('updated', updateRes.modifiedCount)
}

while (startDate < endDate) {
  const res = query(startDate, endDate).toArray()
  if (res.length === 0) {
    break
  }
  toUpdates = toUpdates.concat(res)
  totalUpdates += res.length
  console.log(
    'Total toUpdate so far is',
    totalUpdates,
    'with startDate=',
    startDate,
    'and endDate=',
    endDate,
  )
  if (toUpdates.length >= batchSize) {
    update()
  }
  startDate = res[res.length - 1]._id
}

console.log("checking if there're anything else to update")
update()

function toObjectId(date) {
  var timestamp = Math.floor(date.getTime() / 1000)
  return ObjectId(timestamp.toString(16) + '0000000000000000')
}
```

foochifa: @mayying, was wondering if we should just ship the PR first as it's blocking other stuffs.

i think @KenLSM's current script works and it shouldn't take too much time as tested on the analytics node. Was thinking whether we push ahead and KIV this DB migration discussion for next time? 😅

(not sure if searching dates on `_id` work + @KenLSM made a good point that we are looking for `lastModified` and not `createdAt / _id`)

foochifa (approved):
Thanks @KenLSM

scripts/202306xx_populate_paymenttype_default/add_fixed_payment_type_default.js
KenLSM: Requested @foochifa to go ahead first, as this script is technically not required to be executed before we release. The code itself is backwards compatible. But it should be executed anytime before we release `By Products` to mitigate potential issues.

PR #6467, opened 2023-06-19
feat: Add getStartOfDay function

mayying:
Add function to calculate start of day. Will be used by issue feedback api.

Context: After receiving issue reported by public user, formSG will send notification email to the form admin and collaborators. The frequency of the email is kept at most once per form per day. In order to achieve that, we will have to search the `formIssue` collection within the day where the issue is reported for a particular form. Hence, this function is added here to serve as a helper function that returns the startOfDay to facilitate with the DB query.


src/app/utils/__tests__/date.spec.ts
```diff
@@ -67,4 +71,23 @@ describe('Date Util', () => {
       })
     })
   })
+
+  describe('getStartOfDay', () => {
+    it(`should return today's start of day if date is not given`, () => {
+      // Act
+      const result = getStartOfDay()
+      // Assert
+      const today = new Date(new Date().setHours(0, 0, 0, 0))
```
LinHuiqing: Tip: use `timezone-mock` package to mock timezones! Can refer to my code testing some date utils in `frontend/src/utils/date.test.ts`

src/app/utils/date.ts
```diff
@@ -31,3 +31,17 @@ export const createQueryWithDateParam = (
...
+ *
+ * @param date the date used to calculate the starting day. Defaults to today's
+ * date.
+ */
+export const getStartOfDay = (date?: Date): Date => {
+  if (!date) {
+    date = new Date()
+  }
+  return moment.tz(date, 'Asia/Singapore').startOf('day').toDate()
+}
```
LinHuiqing: Can I also clarify what we are using this function for? If we want to constantly set the timezone to Singapore, we should also name the function accordingly bah

src/app/utils/__tests__/date.spec.ts
mayying: I don't think i need to mock the timezone for this function?

src/app/utils/date.ts
mayying: the utility function itself should be self contained. refactor a little [here](https://github.com/opengovsg/FormSG/pull/6467/files/5725ee19be452e4485d49df8f98c2d4e10169d6f). let me know if it's still unclear.

src/app/utils/date.ts
```diff
@@ -31,3 +31,27 @@ export const createQueryWithDateParam = (
...
+}): Date => {
+  if (!option) {
+    option = {}
+  }
+  if (!option.date) {
+    option.date = new Date()
+  }
+  if (!option.timezone) {
+    option.timezone = 'Asia/Singapore'
+  }
```
LinHuiqing: may I propose using the magic of destructuring:
```suggestion
  option = {
    date: new Date()
    timezone: 'Asia/Singapore'
    ...option
  }
```

src/app/utils/__tests__/date.spec.ts
```diff
@@ -67,4 +71,32 @@ describe('Date Util', () => {
       })
     })
   })
+
+  describe('getStartOfDayInSingaporeTime', () => {
```
LinHuiqing: This test set could benefit from more test cases for branch coverage! So like in the cases of:
- If no options are passed in
- If only date is passed in
- If only timezone is passed in

LinHuiqing (commented):
lgtm otherwise!

src/app/utils/__tests__/date.spec.ts
mayying: ahh, i missed it after added the param, added in the new commit, thanks

src/app/utils/date.ts
mayying: Wow, TIL, thanks for this!

LinHuiqing:
In the case where the server time is in New Zealand, 1am 19 June, Singapore time will be 9pm 18 June. What do you want the function to return if you don't give any arguments? Currently, it will return 19 June 12am SGT right?

mayying:
> In the case where the server time is in New Zealand, 1am 19 June, Singapore time will be 9pm 18 June. What do you want the function to return if you don't give any arguments? Currently, it will return 19 June 12am SGT right?

It should return 18th June 12am SGT. Pls let me know if u are seeing other response. I have tested it locally as well.

src/app/utils/__tests__/date.spec.ts
```diff
@@ -67,4 +71,70 @@ describe('Date Util', () => {
       })
     })
   })
+
+  describe('getStartOfDay', () => {
+    beforeAll(() => {
+      // Mock the return value for new Date() to avoid flaky test
+      jest.useFakeTimers()
+      jest.setSystemTime(new Date(704363057000))
```
LinHuiqing: Could you leave a comment for what date + time this represents?

LinHuiqing (approved):
LGTM and tested! Small nit: can you use more varied timezones for testing? So we will almost definitely have different dates when testing.

E.g. the following timezones:

After addressing the above nit and the one below, feel free to merge!

src/app/utils/__tests__/date.spec.ts
mayying: oh my, just realised I didnt push my new commit 🤦 , should have been fixed now

src/app/utils/__tests__/date.spec.ts
```diff
@@ -67,4 +71,69 @@ describe('Date Util', () => {
...
+      const expected = new Date('2023-06-23T00:00:00+08:00')
+      expect(result).toEqual(expected)
+    })
+
+    it(`should return start of day in Singapore timezone if timezone is not given`, () => {
+      // Act
+      const date = new Date('2020-04-13T05:00:00.000+20:00')
+      const result = getStartOfDay({ date: date })
+      // Assert
+      const expected = new Date(date.setUTCHours(-8, 0, 0, 0))
```
LinHuiqing: Since we are hard-coding the input date, I think this should be hard-coded as well I think. Doesn't make sense when we do computations in the function and similarly compute stuff in the tests.

LinHuiqing: Anyway, a note for next time as well! Can just use `date.setUTCHours(-8)` and no need the other args as they are optional haha based on the documentation [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCHours).

mayying: i shouldn't hardcode this, i missed it! updated it

timotheeg:
Just a note on this PR: even when we create bitesize PRs for easy review. The PR description should be self-contained in explaining why the PR exists in the first place.

Having Context/Approach (similar to the template's Problem/Solution) should be there minimally.

mayying:
> Just a note on this PR: even when we create bitesize PRs for easy review. The PR description should be self-contained in explaining why the PR exists in the first place.
> 
> Having Context/Approach (similar to the template's Problem/Solution) should be there minimally.

Done, do let me know if it's still unclear. thanks!

PR #6469, opened 2023-06-19
feat: track webhook metrics in DataDog

wanlingt:
## Problem

Closes FRM-1009

As Tim pointed out in [this PR](https://github.com/opengovsg/formsg-private/pull/150), it makes more sense to track submissions to forms with webhooks in the app, instead of through a mongo query.

## Solution

In DataDog, record count of submissions sent to webhook URLs, with the tags `webhookType` (`zapier` | `plumber` | `generic`) and `responseCode`

**Breaking Changes** 

## Tests

- [ ] Create a storage-mode form. Set the webhook URL as a [Zapier](https://zapier.com/) URL (follow this [guide](https://guide.form.gov.sg/user-guides/advanced-guide/webhooks/zapier-integration) to set up a Zapier integration with your FormSG form to get that URL). Submit a response. A count of the response should be reflected in this [DD-graph](https://app.datadoghq.com/dashboard/f29-urp-an7/wan-lings-dashboard-test?fullscreen_end_ts=1687175566578&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1687171966578&fullscreen_widget=4704661163853978&tpl_var_env%5B0%5D=staging&from_ts=1684583088053&to_ts=1687175088053&live=false) with the tag 'zapier'
- [ ] Repeat the same test, but with a URL from [Plumber](https://plumber.gov.sg/) - you will have to create a new pipe in Plumber with a **prod** FormSG form to get this URL. The associated tag should be 'plumber'
- [ ] Repeat the same test, but with a non-zapier and non-plumber URL. The associated tag should be 'generic'

### DIFF
--- a/src/app/modules/webhook/webhook.service.ts
+++ b/src/app/modules/webhook/webhook.service.ts
@@ -27,6 +27,7 @@ import {
 } from './webhook.errors'
 import { WebhookQueueMessage } from './webhook.message'
 import { WebhookProducer } from './webhook.producer'
+import { webhookStatsdClient } from './webhook.statsd-client'
 import { formatWebhookResponse, isSuccessfulResponse } from './webhook.utils'
 import { validateWebhookUrl } from './webhook.validation'
 
@@ -221,6 +222,13 @@ export const sendWebhook = (
   })
 }
 
+export const getWebhookType = (webhookUrl: string) => {
+  const isZapier = webhookUrl.match(/^https:\/\/hooks\.zapier\.com\//)
+  const isPlumber = webhookUrl.match(/^https:\/\/plumber\.gov\.sg\/webhooks\//)
+  const webhookType = isZapier ? 'zapier' : isPlumber ? 'plumber' : 'generic'
+  return webhookType
+}
+
 /**
  * Creates a function which sends a webhook and saves the necessary records.
  * This function sends the INITIAL webhook, which occurs immediately after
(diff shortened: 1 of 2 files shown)


mayying:
The regex matches for webhook starting with `https://hooks.zapier.com`, but the test instruction gives url for zapier.com as a https://zapier.com. Likewise for plumber URL. I think the test should not pass?

wanlingt:
> The regex matches for webhook starting with `https://hooks.zapier.com`, but the test instruction gives url for zapier.com as a https://zapier.com. Likewise for plumber URL. I think the test should not pass?

Sorry it was unclear - I've updated the test descriptions accordingly!

mayying:
> 

I see, thanks for updating it! Lgtm

mayying (approved):

src/app/modules/webhook/webhook.service.ts
```diff
@@ -221,6 +222,13 @@ export const sendWebhook = (
   })
 }
 
+export const getWebhookType = (webhookUrl: string) => {
+  const isZapier = webhookUrl.match(/^https:\/\/hooks\.zapier\.com\//)
+  const isPlumber = webhookUrl.match(/^https:\/\/plumber\.gov\.sg\/webhooks\//)
```
timotheeg, line 227: Apologies I am late to make the comment, but better late than never 😅 

* Use `string.match(rexep)` when you are trying to extract part of an input with a regexp
* When all that's needed is boolean check, use `regexp.test(string)` instead.

It's not a big concern, but using match always create a data structure of the match (or `null` if no match), so it's more costly and consumes more memory. `regexp.test()` on the other hand just returns a boolean, and so is as cheap as the check can be.

BTW i you need to extract multiple entries in a regexp, use `regexp.exec(string)`

src/app/modules/webhook/webhook.service.ts
```diff
@@ -243,21 +251,29 @@ export const createInitialWebhookSender =
   > => {
     // Attempt to send webhook
     return sendWebhook(submission.getWebhookView(), webhookUrl).andThen(
-      (webhookResponse) =>
+      (webhookResponse) => {
+        webhookStatsdClient.increment('sent', 1, 1, {
+          responseCode: webhookResponse.response.status.toString(),
```
timotheeg, line 256: ⚠️ Is there a risk of runtime error here?

Can I confirm that network errors (cases where `response.status` is `undefined`!) do NOT enter this handler? And if that is the case, how do we report network errors in the metric?

A simple fix can be to remove the `.toString()` here, or add a safety check:
```typescript
responseCode: webhookResponse.response.status?.toString() || null
```

src/app/modules/webhook/webhook.service.ts
```diff
@@ -243,21 +251,29 @@ export const createInitialWebhookSender =
...
+          webhookType: getWebhookType(webhookUrl),
+        })
```
timotheeg, line 258: Additional note for tags (non-blocking)
* We should probably report a boolean tag for `is_retry: 0|1` so we can filter out retries
* Any way we can somehow know the agency?

src/app/modules/webhook/webhook.service.ts
wanlingt: Thanks Tim, changed to `regexp.test(string)` in #6475

src/app/modules/webhook/webhook.service.ts
wanlingt: I've changed this to a template literal instead! #6475

src/app/modules/webhook/webhook.service.ts
wanlingt: This DD tracking is in the `createInitialWebhookSender`, so it only counts the initial webhook that's sent, which means we don't have to filter out retries. I have added another tag `isRetryEnabled` in case we want to know whether retries have been enabled on the different categories of webhooks.

Including agency - we only have the submission ID in this webhook service, so I'd have to add a query upstream in `performEncryptPostSubmissionActions` to check for the admin and agency. I've left it for now as the agency info wasn't requested by Amit, but can definitely add it in if it's information that we'd like in the future!

PR #6474, opened 2023-06-22
chore: upgrade dd-trace to 4.2

timotheeg:
We are using dd-trace 3 which is outdated. Since we moved to nodejs 18 last year. It is high time we upgrade to version 4

image


timotheeg:
K, this PR is annoying me :( 

I'll give it one more attempt to fix, and if I cannot, I'll upgrade to latest 3.x instead of upgrading to 4 :(

KenLSM:
updated to latest dd-trace in #6728

PR #6475, opened 2023-06-22
fix: account for undefined response code when tracking webhook dd stats

wanlingt:
## Problem

@timotheeg pointed out some ways that the webhook DD metric tracking could be improved in #6469

Related to FRM-1009

## Solution

- Use `regexp.test(string)` instead of `string.match(rexep)`
- add `isRetryEnabled` as a tag
- Use template literals for tags

## Tests

(Same as those in #6469)
- [ ] Create a storage-mode form. Set the webhook URL as a [Zapier](https://zapier.com/) URL (follow this [guide](https://guide.form.gov.sg/user-guides/advanced-guide/webhooks/zapier-integration) to set up a Zapier integration with your FormSG form to get that URL). Submit a response. A count of the response should be reflected in this [DD-graph](https://app.datadoghq.com/dashboard/f29-urp-an7/wan-lings-dashboard-test?fullscreen_end_ts=1687175566578&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1687171966578&fullscreen_widget=4704661163853978&tpl_var_env%5B0%5D=staging&from_ts=1684583088053&to_ts=1687175088053&live=false) with the tag 'zapier'
- [ ] Repeat the same test, but with a URL from [Plumber](https://plumber.gov.sg/) - you will have to create a new pipe in Plumber with a **prod** FormSG form to get this URL. The associated tag should be 'plumber'
- [ ] Repeat the same test, but with a non-zapier and non-plumber URL. The associated tag should be 'generic'

### DIFF
--- a/src/app/modules/webhook/webhook.service.ts
+++ b/src/app/modules/webhook/webhook.service.ts
@@ -223,8 +223,8 @@ export const sendWebhook = (
 }
 
 export const getWebhookType = (webhookUrl: string) => {
-  const isZapier = new RegExp(/^https:\/\/hooks\.zapier\.com\//)
-  const isPlumber = new RegExp(/^https:\/\/plumber\.gov\.sg\/webhooks\//)
+  const isZapier = /^https:\/\/hooks\.zapier\.com\//
+  const isPlumber = /^https:\/\/plumber\.gov\.sg\/webhooks\//
   const webhookType = isZapier.test(webhookUrl)
     ? 'zapier'
     : isPlumber.test(webhookUrl)
--- a/src/app/modules/webhook/webhook.service.ts
+++ b/src/app/modules/webhook/webhook.service.ts
@@ -223,9 +223,13 @@ export const sendWebhook = (
 }
 
 export const getWebhookType = (webhookUrl: string) => {
-  const isZapier = webhookUrl.match(/^https:\/\/hooks\.zapier\.com\//)
-  const isPlumber = webhookUrl.match(/^https:\/\/plumber\.gov\.sg\/webhooks\//)
-  const webhookType = isZapier ? 'zapier' : isPlumber ? 'plumber' : 'generic'
+  const isZapier = new RegExp(/^https:\/\/hooks\.zapier\.com\//)
+  const isPlumber = new RegExp(/^https:\/\/plumber\.gov\.sg\/webhooks\//)
+  const webhookType = isZapier.test(webhookUrl)
+    ? 'zapier'
+    : isPlumber.test(webhookUrl)
+    ? 'plumber'
+    : 'generic'
   return webhookType
 }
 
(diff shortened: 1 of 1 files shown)


src/app/modules/webhook/webhook.service.ts
```diff
@@ -253,8 +257,9 @@ export const createInitialWebhookSender =
     return sendWebhook(submission.getWebhookView(), webhookUrl).andThen(
       (webhookResponse) => {
         webhookStatsdClient.increment('sent', 1, 1, {
-          responseCode: webhookResponse.response.status.toString(),
+          responseCode: `${webhookResponse.response.status || null}`,
           webhookType: getWebhookType(webhookUrl),
+          isRetryEnabled: `${isRetryEnabled}`,
```
timotheeg, line 262: `isRetryEnabled` is not what is important. `isRetryEnabled` is a property of the form, not the current webhook call. The property of the webhook call is whether it is the original call or a retry call.

With that, we can have stats later to know how many of our outbound webhook calls are tries.

We could try tagging the retry attempt number, 1, 2, 3, 4, ... but there's a risk that it becomes unbounded if we change retry behaviour, so that's why I was suggesting tagging a boolean `is_retry:0|1`

wanlingt: Replied to this in the first PR - repasting it here:

This DD tracking is in the `createInitialWebhookSender`, so it only counts the initial webhook that's sent, which means we don't have to filter out retries. I have added another tag isRetryEnabled in case we want to know whether retries have been enabled on the different categories of webhooks.

timotheeg: Ah understood, thanks for the clarification 👍 

I still don't think there is much value in tracking `isRetryEnabled` on the call metrics though 🤔 . As a property of the form, it would make more sense into the daily stats of the webhook feature.

But regardless, 2 additional tag values aren't much, so it's fine to leave it there for now. 👍

src/app/modules/webhook/webhook.service.ts
```diff
@@ -223,9 +223,13 @@ export const sendWebhook = (
 }
 
 export const getWebhookType = (webhookUrl: string) => {
-  const isZapier = webhookUrl.match(/^https:\/\/hooks\.zapier\.com\//)
-  const isPlumber = webhookUrl.match(/^https:\/\/plumber\.gov\.sg\/webhooks\//)
-  const webhookType = isZapier ? 'zapier' : isPlumber ? 'plumber' : 'generic'
+  const isZapier = new RegExp(/^https:\/\/hooks\.zapier\.com\//)
+  const isPlumber = new RegExp(/^https:\/\/plumber\.gov\.sg\/webhooks\//)
```
timotheeg: /nit: the `new RegExp()` constructor call is not needed. I recommend only using it if you construct regexp patterns from parts. Otherwise, stick to the litteral syntax:
```suggestion
  const isZapier = /^https:\/\/hooks\.zapier\.com\//
  const isPlumber = /^https:\/\/plumber\.gov\.sg\/webhooks\//
```

timotheeg (approved):

PR #6476, opened 2023-06-22
feat: Declare formIssue schema

mayying:
Define formIssue schema. 

image

Context: Once public user reports an issue, formSG will store the reported issue in a new collection, named `formIssue`. The queries used for this collection are:
1. `db.getCollection("formIssue").save()`
 To save new reported issue into the collection (Note that email is an optional field from the UI)
2. `db.getCollection("formIssue").countDocuments({ "formId": ObjectId(""), "_id":{$gte:ObjectId(""),$lt:ObjectId("")}}).limit(1)` 
 To search if there're other prior issues reported per form per day (to decide if an email notification is needed)
3.  `db.getCollection("formIssue").find({"formId": ObjectId("")}).sort().skip().limit()`
 To lookup issue and return to admin dashboard


shared/types/form/form_issue.ts
```diff
@@ -0,0 +1,12 @@
+import { FormDto } from './form'
+
+/**
+ * Typing for individual form issue
+ */
+export type FormIssueBase = {
+  formId: FormDto['_id']
+  issue: string
+  email?: string
+  created?: Date
```
ghost, line 10: Curious, why would created be an optional field?

sebastianwzq:
LGTM but I will let more experienced folks review/comment

shared/types/form/form_issue.ts
timotheeg: I think this is a common problem for creation, where the dataset we supply does not contain a `created` date (that gets generated by mongoose), but thereafter all reads will have a `created` field. Ideally we'd have a better way to declare the create/read distinction 🤔

src/app/models/__tests__/form_issue.server.model.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+describe('form_issue.server.model', () => {
+  beforeAll(async () => await dbHandler.connect())
+  beforeEach(async () => await dbHandler.clearDatabase())
+  afterAll(async () => await dbHandler.closeDatabase())
+
+  describe('Schema', () => {
+    const DEFAULT_PARAMS: IFormIssue = {
+      formId: new ObjectId(),
+      issue: 'I am 200 years old but your age drop down is only cap at 100',
```
timotheeg, line 20: Haha, great "problem" 😄

src/types/form_issue.ts
```diff
@@ -0,0 +1,14 @@
...
+import { FormIssueBase } from 'shared/types'
+import type { Merge } from 'type-fest'
+
+import { IFormSchema } from './form'
+
+export type IFormIssue = Merge<FormIssueBase, { formId: IFormSchema['id'] }>
+
+export interface IFormIssueSchema extends IFormIssue, Document {
+  created?: Date
+  lastModified?: Date
```
timotheeg: Is it necessary to redeclare these 2 properties? 🤔 

`IFormIssueSchema` extends `IFormIssue`, itself created from `FormIssueBase`, which already contains the 2 fields.

timotheeg (approved):
LGTM 👍 Just one nit question

src/types/form_issue.ts
mayying: Honestly, all other schemas are defined the same way, where the Base class also has a created and lastModified. But you are right, it seems to be just duplicated. I have tested it and the flow still works

PR #6473, opened 2023-06-22
build: release v6.61.0

timotheeg:
## New

- feat: Add getStartOfDay function [`#6467`](https://github.com/opengovsg/FormSG/pull/6467)
- feat: track webhook metrics in DataDog [`#6469`](https://github.com/opengovsg/FormSG/pull/6469)
- chore: Send issue reported notification email to admin [`#6468`](https://github.com/opengovsg/FormSG/pull/6468)
- feat: email onboarding and help center of payments landing page [`#6462`](https://github.com/opengovsg/FormSG/pull/6462)
- fix: payout webhooks not returning OK [`#6436`](https://github.com/opengovsg/FormSG/pull/6436)
- feat: (BE) create and update admin feedback  [`#6454`](https://github.com/opengovsg/FormSG/pull/6454)
- build: merge v6.60.0 into develop  [`#6461`](https://github.com/opengovsg/FormSG/pull/6461)
- build: release v6.60.0 [`#6460`](https://github.com/opengovsg/FormSG/pull/6460)
- test: add payout cases [`d0eb2d4`](https://github.com/opengovsg/FormSG/commit/d0eb2d4f5ec741ffa9faa2742a713c9e2e7100b1)
- test: add tests for getMetadataPaymentId, cleanup test cases [`a5a06be`](https://github.com/opengovsg/FormSG/commit/a5a06bebe35d2f9c9d819997fa172baeb0c26190)
- refactor: use mapErr to handle errors in neverthrow chain [`ef18774`](https://github.com/opengovsg/FormSG/commit/ef18774805de9e2f0cd879440d29b9a6e57ea402)

## Dependencies

- fix(deps): bump libphonenumber-js from 1.10.34 to 1.10.36 in /shared [`#6466`](https://github.com/opengovsg/FormSG/pull/6466)
- fix(deps): bump type-fest from 3.11.1 to 3.12.0 in /shared [`#6464`](https://github.com/opengovsg/FormSG/pull/6464)

## Dev-Dependencies

- chore(deps-dev): bump @typescript-eslint/parser from 5.59.11 to 5.60.0 in /shared [`#6470`](https://github.com/opengovsg/FormSG/pull/6470)
- chore(deps-dev): bump @typescript-eslint/eslint-plugin from 5.59.11 to 5.60.0 in /shared [`#6471`](https://github.com/opengovsg/FormSG/pull/6471)

### feat: track webhook metrics in DataDog [`#6469`](https://github.com/opengovsg/FormSG/pull/6469)
[...]

PR #6478, opened 2023-06-22
chore: update issue feedback mail html

mayying:
Update form issue feedback mail html content according to latest update on [figma](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?type=design&node-id=18487-191297&mode=design&t=wfhBJQcPVLhfoelV-0)

A few points happen [PR](https://github.com/opengovsg/FormSG/pull/6473). The breakdowns are listed below:
1. On inverted comma: Apparently Stacey's laptop automatically convert the inverted comma to the fat version `‘Report an issue’`. To standardise the view, we will stick to using the skinny inverted comma `'`

2. On <br/> tag: Apparently this is the incorrect html tag, and most browsers try to be immuned to invalid tag like this. Will update to use the correct tag.

### DIFF
--- a/src/app/services/mail/mail.utils.ts
+++ b/src/app/services/mail/mail.utils.ts
@@ -319,11 +319,11 @@ export const generateIssueReportedNotificationHtml = ({
   return dedent`
   <p>Dear Form admin,</p>
   <p>Respondents are facing issues on your form: ${formTitle}.<p>
-  <p>These issues are reported by respondents via the ‘Report an issue’ button 
+  <p>These issues are reported by respondents via the 'Report an issue' button 
   on public forms. We encourage you to view these reports as respondents might 
-  be facing urgent issues relating to form submission. </p>
-  <p><a href="${formResultUrl}">Login to Forms</a> and view issues reported in your 
-  ‘Results’ page.</p>
-  <p>Thank you,</br>${appName}</p>
+  be facing urgent issues relating to form submission.</p>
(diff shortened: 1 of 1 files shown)


timotheeg (approved):

src/app/services/mail/mail.utils.ts
```diff
@@ -319,11 +319,11 @@ export const generateIssueReportedNotificationHtml = ({
   return dedent`
   <p>Dear Form admin,</p>
   <p>Respondents are facing issues on your form: ${formTitle}.<p>
```
timotheeg, line 321: /notFromThisPRButShoulStillFix

⚠️ for markup safety, `formTitle` should be html-encoded (with [html entities](https://developer.mozilla.org/en-US/docs/Glossary/Entity))

We can assume `formResultUrl` and `appName` are markup-safe (so no need to change those), but since the form-title is provided by users, we need to encode it properly to ensure we generate valid markup.

timotheeg: Urgh, I see we have other markup content where this was not done 😢 (e.g. in `generatePaymentConfirmationHtml()`), We should upgrade all of those.

mayying: Ok created https://linear.app/ogp/issue/FRM-1048/ensure-user-input-variable-is-html-encoded to track, will work on it after this

timotheeg: Thanks @mayying 🙏

PR #6479, opened 2023-06-22
fix: initialize payoutProcessError with default value

KenLSM:
## Problem

1. Payout hooks from stripe are returning `502`
2. Cloudwatch was receiving `... undefined (reading match) ...`

Introduced in #6436

**Sequence of events**

1. `payoutProcessError` is not initialized to be anything.
Suppose there’s no error when running `ResultAsync`. 
2. `result` will be set with payoutProcessError := null, as it was not set to be anything.
3. Then returned back to it’s caller with `return result // null`. Caller was expecting a type ResultAsync
4. Caller freaks out

## Solution

Initialize `payoutProcessError` with `okAsync`, assuming that it will be replaced when errors occurs inside `processStripeEvent`

**Breaking Changes** 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "FormSG",
-  "version": "6.61.0",
+  "version": "6.61.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "FormSG",
-      "version": "6.61.0",
+      "version": "6.61.1",
       "hasInstallScript": true,
       "dependencies": {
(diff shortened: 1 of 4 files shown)


timotheeg (approved):
LGTM!

I think the explicit `undefined` might not even be necessary, but not an issue at all

KenLSM:
`okAsync` requires a value 😞 

```
// node_modules/neverthrow/dist/index.d.ts
declare const okAsync: <T, E = never>(value: T) => ResultAsync<T, E>;
```

timotheeg:
Oh, the PR needs a version bump to .1

timotheeg (approved):
+1

PR #6481, opened 2023-06-22
fix: potential html script injection

mayying:
## Problem
Some rich HTML emails are formatted by concatenating strings in HTML markup. Many values we inject are controlled by us (e.g. links to form) and are safe for HTML injection. Some however like the Form title, which is a user input, are not.

Closes [FRM-1048]

## Solution
Use ejs render to render html with the usage of `<%=` tag
Resource: https://ejs.co/
image

From the code, It seems that only payment confirmation email and issue notification email is susceptible to injection attack due to direct concatenation of `formTitle` (characteristic: free text input by user, unsanitised upon creation), so we will be using ejs render to protect ourself against unwanted attack.

**Breaking Changes** 

    - Details ...

**Bug Fixes**:

- Protect against potential script injection

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
html-rich text is printed as normal text
image

## Tests
- [ ] Payment confirmation email should be received successfully (Regression test)
- [ ] Create a form and set form title as `<h1>Hello, I am an email injection attack!</h1>` (This may not work in staging because cloudfront checks for sql injection, if got blocked, manually change the form title in DB)
- [ ] Enable the form for payment
- [ ] Make a payment submission
- [ ] Ensure the payment confirmation email shows title as plain text
- [ ] Ignore issue notification email for now, there is no production route to active it, will test when feature is ready for release


src/app/services/mail/mail.utils.ts
```diff
@@ -269,27 +271,21 @@ export const generateSmsVerificationWarningHtmlForCollab = (
...
+  htmlData: PaymentConfirmationData
+}): ResultAsync<string, MailGenerationError> => {
+  const pathToTemplate = `${process.cwd()}/src/app/views/templates/payment-confirmation.view.html`
+  logger.info({
+    message: 'generatePaymentConfirmationHtml',
+    meta: {
+      action: 'generatePaymentConfirmationHtml',
+      pathToTemplate,
+      __dirname,
+      cwd: process.cwd(),
```
timotheeg: /nit its probably not necessary to log those 2 fields, or at least no need to log `__dirname`, since it's not being used in the function.

I can see many log blocks above do log them, but while I cannot remember why we do it 😑, I suspect it was to verify values to move away from `__dirname`. 

I just checked a recent log which give this:
```
action: generateAutoreplyHtml
__dirname:  /opt/formsg/src/app/services/mail
cwd: /opt/formsg
```

So it looks like using `process.cwd()` is indeed safe to use, and we don't have to log the properties any more. And we can probably schedule to upgrade the other functions to consistently use `process.cwd()` in another ticket.

src/app/services/mail/mail.utils.ts
```diff
@@ -308,22 +304,19 @@ export const generatePaymentOnboardingHtml = ({
...
+  htmlData: IssueReportedNotificationData
+}): ResultAsync<string, MailGenerationError> => {
+  const pathToTemplate = `${process.cwd()}/src/app/views/templates/issue-reported-notification.view.html`
+  logger.info({
+    message: 'generateIssueReportedNotificationHtml',
+    meta: {
+      action: 'generateIssueReportedNotificationHtml',
+      pathToTemplate,
+      __dirname,
+      cwd: process.cwd(),
```
timotheeg: Same nit comment

src/app/views/templates/payment-confirmation.view.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en" xmlns="http://www.w3.org/1999/xhtml">
+  <body>
+    <p>Hello there,</p>
+    <p>
+      Your payment on <%= appName %> form: <%= formTitle %> has been received
+      successfully. Your response ID is <%= submissionId %> and your payment
+      invoice can be found <a href="<%= invoiceUrl %>">here </a>.
+    </p>
+    <p>Regards,<br />
+    <p><%= appName %> team</p>
```
timotheeg: ⚠️ Invalid markup here
```suggestion
    <p>Regards,<br />
    <%= appName %> team</p>
```

timotheeg (commented):
Thanks @mayying LGTM!

Just one minor correction needed for the markup of `generatePaymentConfirmationHtml()`.

The other nit comments are inconsequential, so you can ignore them if you want.

src/app/services/mail/mail.utils.ts
mayying: ok i will remove them then, was wondering why also, but decided to just follow.

timotheeg (approved):

timotheeg (approved):
LGTM!

PR #6483, opened 2023-06-23
feat: Implemented turnstile integration for email and stored submission

ghost:
## Problem
Users are encountering multiple slow recaptchas while using GSIB. 

Closes FRM-1052

## Solution
This merge requests contains the following

1. integrates with Cloudflare Turnstile as an alternative to Google reCAPTCHA
2. feature flags to control rollout

[Investigation](https://docs.google.com/document/d/1-VUEHnEuFdkAPL-m3emvWd8zZ6nlf7XIiuVuORSTDRY/edit) 

**Breaking Changes** 

    - Details ...

**Features**:
1. integrates with Cloudflare Turnstile as an alternative to Google reCAPTCHA
2. feature flags to control rollout

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

https://github.com/opengovsg/FormSG/assets/136435307/e0683251-13c8-490d-b212-356911ae18d6

**AFTER**:

https://github.com/opengovsg/FormSG/assets/136435307/1c6bb4a3-d5fb-4658-be0b-54179aec404f

## Deploy Notes

- `env var` : env var details

1. turnstileSiteKey
2. turnstilePrivateKey

**New scripts**:

- `script` : script details

**New dependencies**:
[Cloudflare Turnstile](https://www.cloudflare.com/en-gb/products/turnstile/)

**New dev dependencies**:


frontend/src/features/feature-flags/queries.ts
```diff
@@ -11,3 +11,16 @@ export const featureFlagsKeys = {
 export const useFeatureFlags = (): UseQueryResult<Set<string>> => {
   return useQuery(featureFlagsKeys.base, () => getEnabledFeatureFlags())
 }
+
+// Retrieves a feature flag and returns the default value
+// if an error is encountered or if the flag does not exist
+export const useFeatureFlagWithDefaults = (
```
KenLSM: With a more generic name, I think we should use this method for the rest of the code. Every flag should have an explicitly defined default.

```
// definition
useIsFeatureEnabled(flagName: string,  defaultValue: boolean): boolean

// usage
const enableTurnstile = useIsFeatureEnabled(
    featureFlags.turnstile,
    false,
  )
```

frontend/src/features/recaptcha/useRecaptcha.tsx
```diff
@@ -185,6 +199,7 @@ export const useRecaptcha = ({
     handleError,
     useEnterprise,
     grecaptcha,
+    enableTurnstileFeatureFlag,
```
KenLSM, line 202: Is this still required?

frontend/src/features/recaptcha/useRecaptcha.tsx
```diff
@@ -97,6 +101,14 @@ export const useRecaptcha = ({
 
   const captchaRef = useRef<HTMLDivElement | null>(null)
 
+  // Feature flag to control turnstile captcha rollout
+  // defaults to false
+  // todo: remove after full rollout
+  const enableTurnstileFeatureFlag = useFeatureFlagWithDefaults(
```
KenLSM: We might want to include why we want to check for `turnstile` in `turnstile`. It took me awhile to realize that both of these (grecaptcha, cf turnstile) should mutually exclusive.

frontend/src/features/turnstile/useTurnstile.tsx
```diff
@@ -0,0 +1,160 @@
...
+  getResponse: (widgetID: string) => string
+}
+
+declare global {
+  interface Window {
+    turnstile?: TurnstileCallbacks
+  }
+}
+
+// todo: check usages
```
KenLSM, line 45: I think we can remove this `todo` since we are using it.

frontend/src/features/feature-flags/queries.ts
ghost: Sure, I can add the definition above. I was also thinking of replacing the existing feature flags calls with the new function. I can create another ticket for this and do it as part of a separate diff. What do you think?

frontend/src/features/recaptcha/useRecaptcha.tsx
ghost: Sure, let me elaborate it better

frontend/src/features/turnstile/useTurnstile.tsx
ghost: Ack

frontend/src/features/feature-flags/queries.ts
KenLSM: I would prefer that to be done in a separate ticket -- just to avoid feature creep 😆

frontend/src/features/turnstile/useTurnstile.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const useTurnstile = ({
+  containerID = DEFAULT_CONTAINER_ID,
+  sitekey,
+  execution = 'execute',
+  theme = 'light',
+  appearance = 'interaction-only',
+}: UseTurnstileProps) => {
+  useScript(
+    'https://challenges.cloudflare.com/turnstile/v0/api.js?render=explicit',
```
KenLSM, line 61: Could we double check what happens on the responder side if this endpoint is not reachable? 

**Context**
We had an issue previously where our users in China couldn't access `google.com` and the page failed to load completely. There was no workaround as the captcha is required to load even if it was not enabled by the admin.

KenLSM (approved):
LGTM!

frontend/src/features/turnstile/useTurnstile.tsx
ghost: I'm afraid the page will fail to load if this endpoint is not reachable. This applies to Google reCAPTCHA too. AFAIK, captcha verification is a critical flow for forms since a submission is not possible if the captcha verification in the backend fails.

A possible approach I can think of is to guard both captchas behind a feature flag so that we can disable them completely if required but that might not be viable from a security POV.

ghost: Oh, regarding your second point, if captcha is not enabled, we will not attempt to load this script

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -217,15 +250,29 @@ export const PublicFormProvider = ({
...
+
+      if (enableTurnstileFeatureFlag) {
+        try {
+          captchaResponse = await getTurnstileResponse()
+        } catch (error) {
+          if (error instanceof RecaptchaClosedError) {
+            // Do nothing if recaptcha is closed.
+            return
+          }
+          trackReCaptchaOnError(form)
```
timotheeg, line 262: /nit the name is a little confusing here, it looks like a copy/paste mistake (?). If the error class now applies to both Recaptcha and Turnstile, perhaps we could rename the error class `captchaClosedError` instead.

frontend/src/features/turnstile/useTurnstile.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const handleChange = useCallback((response: string | null) => {
+    if (executionPromise.current.resolve) {
+      executionPromise.current.resolve(response)
+      executionPromise.current = {}
+    }
+  }, [])
+
+  const handleError = useCallback(() => {
+    if (executionPromise.current?.reject) {
```
timotheeg, line 95: Why is there a conditional chaining operator here while there's isn't one for the resolve in the handleChange check above?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -147,18 +151,53 @@ const submitEncryptModeForm: ControllerHandler<
...
-        error: captchaResult.error,
+        error: featureFlagsListResult.error,
       })
-      const { errorMessage, statusCode } = mapRouteError(captchaResult.error)
-      return res.status(statusCode).json({ message: errorMessage })
+    } else {
+      enableTurnstileFeatureFlag = featureFlagsListResult.value.includes(
+        featureFlags.turnstile,
+      )
+    }
```
timotheeg, line 169: /nit: Is this code straight duplicate from `email-submission.controller.ts`? Any way to have it in some shared utility?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -147,18 +151,53 @@ const submitEncryptModeForm: ControllerHandler<
...
+    if (enableTurnstileFeatureFlag) {
```
timotheeg, line 170: There is a bit of risk with using `enableTurnstileFeatureFlag` like that on the server: when we update the flag value, there can be a discrepancy between the frontend and the backend. If the distinction recaptcha/turnstile could be done based on the payload itself, it'd be safer. But otherwise we just have to remember to flip the feature flag at low-traffic period for safety.

src/app/services/turnstile/turnstile.constants.ts
```diff
@@ -0,0 +1,4 @@
+// Server side validation for turnstile captchas
+// https://developers.cloudflare.com/turnstile/get-started/server-side-validation/
+export const TURNSTILE_CAPTCHA_VERIFY_URL =
```
timotheeg, line 3: Shouldn't this go in a config object instead, and be driven by env var with default value to this URL, so we can mock it if needed?

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -125,13 +127,44 @@ export const PublicFormProvider = ({
...
+    sitekey: data?.form.hasCaptcha ? turnstileSiteKey : undefined,
+  })
+
+  const {
+    hasLoaded: hasRecaptchaLoaded,
+    getCaptchaResponse,
+    containerId: recaptchaContainerID,
+  } = useRecaptcha({
     sitekey: data?.form.hasCaptcha ? captchaPublicKey : undefined,
   })
```
timotheeg, line 158: justBeingSadNothingToDo that we have to initiatlize both because react cannot do conditional `use*()` 😢

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -217,15 +250,29 @@ export const PublicFormProvider = ({
...
+        try {
+          captchaResponse = await getCaptchaResponse()
+        } catch (error) {
+          if (error instanceof RecaptchaClosedError) {
+            // Do nothing if recaptcha is closed.
+            return
+          }
+          trackReCaptchaOnError(form)
+          return showErrorToast(error, form)
         }
```
timotheeg, line 275: /nit (not sure if typescript would be happy): since the catch handlers are exactly the same, the conditional on `enableTurnstileFeatureFlag` could be done if the `try{}` rather than out of it to reduce the code a little

Although again, it looks funny that the catch handler for Turnstile worries about recaptcha errors (e.g. `trackReCaptchaOnError()`)

timotheeg (commented):
Argh, I just noticed, the PR is merged 😅 I was too slow to review sorry!

frontend/src/features/turnstile/useTurnstile.tsx
ghost: According to https://developers.cloudflare.com/turnstile/get-started/client-side-rendering/ , handleError is called when there is an error (e.g. network error or the challenge failed). Added conditional chaining in the event that the error is thrown before turnstile.execute is called (i.e. before the promise is set)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
ghost: Both are written quite differently and will require significant refactoring. I can take a look at how we can restructure both controllers

frontend/src/features/public-form/PublicFormProvider.tsx
ghost: I will refactor this so that the errors are tracked separately.

frontend/src/features/turnstile/useTurnstile.tsx
timotheeg: 👍  got it, thanks for the clarification!

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
timotheeg: Oh? Per the highlight (line 154 to 169), I meant just the code to gather and compute `featureFlagsListResult`. Isn't that a straight copy?

here
```typescript
    const featureFlagsListResult = await FeatureFlagService.getEnabledFlags()
    // Feature flag to control turnstile captcha rollout
    // defaults to false
    // todo: remove after full rollout
    let enableTurnstileFeatureFlag = false
    if (featureFlagsListResult.isErr()) {
      logger.error({
        message: 'Error occurred whilst retrieving enabled feature flags',
        meta: logMeta,
        error: featureFlagsListResult.error,
      })
    } else {
      enableTurnstileFeatureFlag = featureFlagsListResult.value.includes(
        featureFlags.turnstile,
      )
    }
```

In `email-submission.controller.ts`
```typescript
  const featureFlagsListResult = await FeatureFlagService.getEnabledFlags()
  // Feature flag to control turnstile captcha rollout
  // defaults to false
  // todo: remove after full rollout
  let enableTurnstileFeatureFlag = false
  if (featureFlagsListResult.isErr()) {
    logger.error({
      message: 'Error occurred whilst retrieving enabled feature flags',
      meta: logMeta,
      error: featureFlagsListResult.error,
    })
  } else {
    enableTurnstileFeatureFlag = featureFlagsListResult.value.includes(
      featureFlags.turnstile,
    )
  }
```

timotheeg: Hopefully we could put that in a shared function such that the code in both submission handlers could be reduced to a minimum
```typescript
const enableTurnstileFeatureFlag = await FeatureFlagService.hasFeatureFlag(featureFlags.turnstile)
```

Actually, don't we already have a function like that?

timotheeg: But yeah, so just to be super clear: full refactor of the encrypt and email controller is indeed be a completely separate task; definitely out of scope for this PR, and I didn't doing mean that with my comment 😅

KenLSM: I've some plans to refactor them in d2a8816f1, c18d8bae6, and 52cf50aac (under payments by products 🙈 )

PR #6465, opened 2023-06-15
feat: admin feedback modal

foochifa:
## Problem

Implements the frontend modals for admin feedback, works on the BE endpoints from #6454

Closes FRM-992

## Solution

[Designs are here](https://www.figma.com/file/CMq0CJtawn7NGM7Zdpi8Ti/Form-Design-Master?type=design&node-id=17430-181452&t=Oa9zzJBSHsDZhbOM-0)

### Firstly, for the preconditions to show admin feedback, there are two portions:

**Eligibility based on current usage session**
     a. Create/Edit a payment field
     b. Create/Edit a myinfo field
     c. Create/Edit a form field in a form with > 5 fields

**Solution:**

Use `useSessionStorage` to keep a session variable of whether any of the above precondition have been met. Will wipe the key if the user has logged. For payment, will update session variable if payment input panel `save field` button has been selected. For myinfo, will update session variable if edit drawer's `save field` button has been selected. 

For point c, due to edit drawer submit button not being provided with any form field size context, have decided to instead set the session variable if edit field drawer is opened and the builder content has >= 5 fields (including the created field)

Have kept this as a general key to reduce complexity of `useEditFieldForm` which does not have knowledge of the current user. But will delete the key upon logout.

**Frequency of feedback shown**
    a. Show the feedback in an interval of 1 month

**Solution:**

Use `useLocalStorage` to keep the previous feedback time by the user. If the feedback time exceeds current time by one month, or if it did not exist yet, will show the modal.

Decided to use Local Storage for faster development. We should KIV and see if we want to hard-set the 1 month interval in the `user` schema in the DB. 
   
The admin feedback modal will only be showed if the above two preconditions are met for a user

### Next, the admin feedback flow
1. Thumbs Up/Down -> results will be updated in `adminFeedback` collections and DD metrics for `formsg.users.feedback.rating` will be updated
2. Call For Comment
3. Comment Box -> results will be patched onto the feedback document in step 1

**Solution:**

Enum content state, and changeState upon each step. Mutate create on step 1, keep `feedbackId` as a state in `AdminFeedbackModalContent`, mutate update on step 3 with `useForm` to capture the comment. 

### Mobile

Will disable admin feedback for mobile, and will create another ticket to implement the modal based on the updated designs by @staceytan1998 

**Breaking Changes** 

- [ ] No - this PR is backwards compatible  
    - Not breaking as #6454 has been released 

## Before & After Screenshots

**AFTER**:

On Desktop

[...]

PR #6485, opened 2023-06-26
fix: admin feedback capturing wrong local storage value

foochifa:
## Problem

working on #6465 

We are pre-emptively setting `isDisplayAdminFeedback` when `lastFeedbackTime` is `undefined`. This undefine-ness is caused by our implementation of `useLocalStorage` that may return the previous key value. In this case `ADMIN_FEEDBACK_HISTORY_PREFIX + undefined'`. 

Also updated the design of the rating modal from @staceytan1998 's correction!

## Solution

Abstracted out a new container called `AdminFeedbackContainer` that will only be called if the `useUser` return is defined. hence the above case of `undefined` admin feedback prefix access will not happen. 

**Breaking Changes** 

    - Not breaking as #6454 has been released 

## Tests

**Ensure the fix is working** 
- [ ] check that admin feedback key is in local storage with a defined time
- [ ] check that admin feedback flag is true
- [ ] refresh the dashboard
- [ ] you should not see the feedback modal

## Deploy Notes

- `ADMIN_FEEDBACK_FIELD_THRESHOLD` : (optional, default:5) threshold of the amount of fields in an active form, to enable eligibility of user for admin feedback
- `ADMIN_FEEDBACK_DISPLAY_FREQUENCY`: (optional, default:28 days in miliseconds) frequency to display the admin feedback to a user

### DIFF
--- a/frontend/src/constants/localStorage.ts
+++ b/frontend/src/constants/localStorage.ts
@@ -32,4 +32,4 @@ export const EMERGENCY_CONTACT_KEY_PREFIX = 'has-seen-emergency-contact'
 /**
  * Key to store when was the last time user has seen the admin feedback modal
  */
-export const ADMIN_FEEDBACK_HISTORY_PREFIX = 'last-seen-admin-feedback'
+export const ADMIN_FEEDBACK_HISTORY_PREFIX = 'last-seen-admin-feedback-'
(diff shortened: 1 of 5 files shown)


KenLSM (approved):
LGTM

PR #6491, opened 2023-06-27
fix: handle and wrap stripe fetches in neverthrow

KenLSM:
## Problem
When downloading stripe invoice, especially dated invoices (>3months back) the download API would crash and show a cloudflare 502(!!) to the user.

This behaviour is unexpected, and unfriendly as the user does not have any way to redress this situation.

Closes FRM-1074

## Solution

Check if API returns error, and catch the error with `mapErr`.

**Breaking Changes** 

**BEFORE**:
Screenshot 2023-06-27 at 10 20 23 PM

**AFTER**:
Screenshot 2023-06-27 at 10 19 59 PM

## Tests
**Regressions**
Ensuring that valid invoices downloads are still downloadable
As a form admin
- [ ] invoice is downloadable
As a responder
- [ ] make payment on a payment form
- [ ] invoice is downloadable

**Feature Test** (not strictly required to pass in pre-deployment check)
Ensure that expired invoices download doesn't crash (show cloudflare 502s)
As a form admin
- [ ] download of invalid invoice shows a BE error message of `StripeFetchError` to the browser instead of CF 502

### DIFF
--- a/src/app/modules/payments/__tests__/stripe.controller.spec.ts
+++ b/src/app/modules/payments/__tests__/stripe.controller.spec.ts
@@ -8,7 +8,12 @@ import { PaymentStatus, SubmissionType } from 'shared/types'
 
 import getPaymentModel from 'src/app/models/payment.server.model'
 import { getEncryptPendingSubmissionModel } from 'src/app/models/pending_submission.server.model'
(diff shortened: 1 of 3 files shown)


timotheeg (approved):
LGTM

PR #6492, opened 2023-06-28
feat: issue feedback post api service

mayying:
## Problem
Create an API to accept user issue complaints for any given form. An email notification will be sent per day per form when the API is called

Closes FRM-999

## Tests
Full E2E test can be done once https://github.com/opengovsg/FormSG/pull/6452 is merged.


src/app/modules/issue/issue.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+    'permissionList',
+    'status',
+    'title',
+  ])
+    .andThen((form) => FormService.isFormPublic(form).map(() => form))
+    .andThen((form) =>
+      insertFormIssue({
+        formId: form._id,
+        issue: issue,
+        email: email,
```
KenLSM: /nit can be terser by using Object shorthand
```js
{
        formId: form._id,
        issue,
        email,
}

```

src/app/modules/issue/issue.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+      }).map(async (formIssue) => {
+        await notifyFormAdmin({ form: form, formIssue: formIssue })
```
KenLSM: /nit can be terser by using Object shorthand as well

src/app/modules/issue/issue.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+      }),
+    )
+    .map(async () => {
```
KenLSM, line 69: `async` is probably not required here as the function is not `awaitable`

src/app/modules/issue/issue.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  return FormService.retrieveFormKeysById(formId, [
+    '_id',
+    'admin',
+    'inactiveMessage',
+    'permissionList',
+    'status',
+    'title',
+  ])
+    .andThen((form) => FormService.isFormPublic(form).map(() => form))
```
KenLSM, line 59: Completely unrelated to the changes on this PR. 

This `FormService.isFormPublic(form)` is rather idiosyncratic. It returns `ok(true)/err(false)` which is a tautology.

To check if the value is false the caller wouldn't need to access the inner value:
```js
const result = FormService.isFormPublic(form);
if(result.isErr()) {
  // don't really need to consume the actual response
}

if(result.isTrue()) {
  // don't really need to consume the actual response either 🤔 
}
```

This causes the caller to require some gymnastics to chain the function like what it is done here instead of being able to directly use the result. 

With a small change like so:
```diff
// update isFormPublic to return `form`
export const isFormPublic = (
  form: IPopulatedForm,
): Result<true, FormDeletedError | PrivateFormError | ApplicationError> => {
  if (!form.status) {
    return err(new ApplicationError())
  }
  switch (form.status) {
    case FormStatus.Public:
-      return ok(true)
+      return ok(form)
    case FormStatus.Archived:
      return err(new FormDeletedError())
    case FormStatus.Private:
      return err(new PrivateFormError(form.inactiveMessage, form.title))
  }
}
```

We can directly chain the result:
```js
FormService.retrieveFormKeysById(...args)
  .andThen(FormService.isFormPublic)
  .andThen(doSomething)
```

Wondering if you also had the same observation when working on this.

src/app/modules/issue/issue.service.ts
```diff
@@ -0,0 +1,182 @@
...
+  email?: string
+}): ResultAsync<IFormIssueSchema, FormNotFoundError | DatabaseError> => {
+  if (!mongoose.Types.ObjectId.isValid(formId)) {
+    return errAsync(new FormNotFoundError())
+  }
+  return ResultAsync.fromPromise(
+    FormIssueModel.create({
+      formId: formId,
+      issue: issue,
+      ...(email && { email: email }),
```
KenLSM: /nit can be terser by using Object shorthand as well

src/app/modules/issue/issue.service.ts
```diff
@@ -0,0 +1,182 @@
...
+          action: 'insertFormIssue',
+        },
+        error,
+      })
+      return new DatabaseError(
+        'Failed to count the document in the formIssue collection',
+      )
+    },
+  ).andThen((count) => {
+    return okAsync(count == 0)
```
KenLSM: Missed out another `=`? 

We usually still use strict equality (`===`) even though the operand are both of the same type. This is to prevent potential pitfall when the operands changes.

src/app/modules/issue/issue.service.ts
```diff
@@ -0,0 +1,182 @@
...
+ * @param formIssue the form issue object where the issue is recorded
+ */
+export const notifyFormAdmin = ({
+  form,
+  formIssue,
+}: {
+  form: IPopulatedForm
+  formIssue: IFormIssueSchema
+}): Promise<boolean> => {
+  if (form.admin && form.admin.email != '') {
```
KenLSM: Same here for strict inequality

src/app/modules/issue/issue.service.ts
```diff
@@ -0,0 +1,182 @@
...
+    const logMeta = {
+      action: 'notifyFormAdmin',
+      formIssue,
+    }
+
+    return getIsFirstIssueForFormToday({ formIssue: formIssue })
+      .andThen((isFirstIssueForFormToday) => {
+        return isFirstIssueForFormToday
+          ? MailService.sendFormIssueReportedNotificationToAdmin({
+              form: form,
```
KenLSM: /nit can use object shorthand

src/app/modules/issue/issue.service.ts
```diff
@@ -0,0 +1,182 @@
...
+ * @param form the form object where public reports an issue
+ * @param formIssue the form issue object where the issue is recorded
+ */
+export const notifyFormAdmin = ({
+  form,
+  formIssue,
+}: {
+  form: IPopulatedForm
+  formIssue: IFormIssueSchema
+}): Promise<boolean> => {
```
KenLSM: Hm. Any specific requirement to not return `ResultAsync`?

src/app/modules/issue/issue.service.ts
mayying: Ahh.. thanks for pointing it out, not used to typescript/javascript.

src/app/modules/issue/issue.service.ts
mayying: Because the error needs to be silenced, if I use ResultAsync.mapErr(), it will translate <T, E> to <T, F>, but the error will still be propagated, and i didn't want this to happen. let me know if you have an alternative.

src/app/modules/issue/issue.service.ts
mayying: done

src/app/modules/issue/issue.service.ts
mayying: done

src/app/modules/issue/issue.service.ts
mayying: done

src/app/modules/issue/issue.controller.ts
mayying: done

src/app/modules/issue/issue.controller.ts
mayying: done!

src/app/modules/issue/issue.controller.ts
mayying: done

src/app/modules/issue/issue.controller.ts
mayying: Cmiiw but currently the error is being consumed, to report why is the form not public (tho i doubt we should be doing this, for security reason). 

As for the argument of returning true/false vs the form object, I think since the function name is isXXXX, naturally, caller will expect it to function as a predicate function, which also translates to better readability. But since we are already returning an error, this is no longer appropriate for a predicate function, which means returning the form object itself will makes it easier for us to chain. Also, in term of naming, perhaps something like validatePublicForm will be better to avoid misunderstanding and assumption that the function will return a boolean type.

In summary, I agree with your point, but i will keep this here for now, and we can do some refactoring in the future. Lmk if you think otherwise.

KenLSM: 👍 Yeah definitely not suggesting refactoring here. ❌ feature creeps!

KenLSM (approved):
LGTM!

LinHuiqing:
@mayying anything preventing this from being pushed to develop? Other than conflicts.

PR #6502, opened 2023-07-03
refactor: Fix backwards compatibility issues with Turnstile integration

sebastianwzq:
## Problem
Improving backwards compatibility of turnstile integration.

Context: Currently, using the enableTurnstileFeatureFlag in frontend/backend can result in the following edge case. A user visits the site before the flag is switched on (i.e. reCAPTCHA enabled) and receives a reCAPTCHA challenge/response. The user completes the form and clicks 'Submit' after we have switched on the flag (i.e. Turnstile enabled). The challenge-response from reCAPTCHA will be forwarded to our backend and verified against Turnstile, which will fail. Users must either click submit again to receive a new turnstile challenge or refresh the page. I did some investigation and its not possible to differentiate between Turnstile/reCAPTCHA based on the challenge-response alone.

Closes FRM-1107

## Solution
This merge request solves the above problem by having the frontend explicitly tell the backend which captcha provider is being used.

## Tests

- [ ]  Verify that Turnstile captcha works (i.e. able to submit forms) after switching on the enableTurnstileFeatureFlag
- [ ] Verify that Google recaptcha works (i.e. able to submit forms) after switching off the enableTurnstileFeatureFlag

### DIFF
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -21,6 +21,7 @@ import {
   PAYMENT_VARIABLE_INPUT_AMOUNT_FIELD_ID,
 } from '~shared/constants'
 import { PaymentType } from '~shared/types'
+import { CaptchaTypes } from '~shared/types/captcha'
 import {
   FormAuthType,
   FormResponseMode,
@@ -143,12 +144,8 @@ export const PublicFormProvider = ({
   // todo: remove after full rollout
   const enableTurnstileFeatureFlag = useIsFeatureEnabled(
     featureFlags.turnstile,
-    false,
+    true,
   )
-  enum CaptchaTypes {
-    Turnstile = 'turnstile',
-    Recaptcha = 'recaptcha',
-  }
 
   let hasLoaded: boolean
   let containerID: string
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -616,6 +616,7 @@ export const PublicFormProvider = ({
       enableTurnstileFeatureFlag,
       getTurnstileResponse,
       getCaptchaResponse,
+      captchaType,
       showErrorToast,
       submitEmailModeFormMutation,
       submitStorageModeFormMutation,
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -144,7 +144,7 @@ export const PublicFormProvider = ({
   // todo: remove after full rollout
   const enableTurnstileFeatureFlag = useIsFeatureEnabled(
     featureFlags.turnstile,
-    true,
+    false,
   )
 
   let hasLoaded: boolean
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -145,8 +145,14 @@ export const PublicFormProvider = ({
     featureFlags.turnstile,
     false,
   )
+  enum CaptchaTypes {
+    Turnstile = 'turnstile',
+    Recaptcha = 'recaptcha',
+  }
+
   let hasLoaded: boolean
   let containerID: string
+  let captchaType: CaptchaTypes
 
   const {
     hasLoaded: hasTurnstileLoaded,
@@ -154,6 +160,7 @@ export const PublicFormProvider = ({
     containerID: turnstileContainerID,
   } = useTurnstile({
     sitekey: data?.form.hasCaptcha ? turnstileSiteKey : undefined,
+    enableUsage: enableTurnstileFeatureFlag,
   })
 
   const {
@@ -162,14 +169,17 @@ export const PublicFormProvider = ({
     containerId: recaptchaContainerID,
   } = useRecaptcha({
     sitekey: data?.form.hasCaptcha ? captchaPublicKey : undefined,
+    enableUsage: !enableTurnstileFeatureFlag,
   })
 
   if (enableTurnstileFeatureFlag) {
     hasLoaded = hasTurnstileLoaded
     containerID = turnstileContainerID
+    captchaType = CaptchaTypes.Turnstile
   } else {
     hasLoaded = hasRecaptchaLoaded
     containerID = recaptchaContainerID
+    captchaType = CaptchaTypes.Recaptcha
   }
 
   const { isNotFormId, toast, vfnToastIdRef, expiryInMs, ...commonFormValues } =
@@ -287,6 +297,7 @@ export const PublicFormProvider = ({
         formLogics: form.form_logics,
         formInputs,
         captchaResponse,
+        captchaType,
         responseMetadata: {
           responseTimeMs: differenceInMilliseconds(Date.now(), startTime),
           numVisibleFields: isPaymentEnabled
@@ -406,6 +417,7 @@ export const PublicFormProvider = ({
                   ...formData,
                   publicKey: form.publicKey,
                   captchaResponse,
+                  captchaType,
                   paymentReceiptEmail: paymentReceiptEmailField?.value,
                   ...(form.payments_field.payment_type === PaymentType.Variable
                     ? {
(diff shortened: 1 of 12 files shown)


frontend/src/features/recaptcha/useRecaptcha.tsx
```diff
@@ -101,15 +99,6 @@ export const useRecaptcha = ({
 
   const captchaRef = useRef<HTMLDivElement | null>(null)
 
-  // Feature flag to control turnstile captcha rollout
-  // defaults to false
-  // Cloudflare Turnstile and Google reCaptcha should be mutually exclusive
-  // todo: remove after full rollout
-  const enableTurnstileFeatureFlag = useIsFeatureEnabled(
```
ghost, line 108: dependency injection instead of retrieving feature flags here.

frontend/src/features/turnstile/useTurnstile.tsx
```diff
@@ -66,14 +64,6 @@ export const useTurnstile = ({
 
   const executionPromise = useRef<TurnstileExecutionCallback>({})
 
-  // Feature flag to control turnstile captcha rollout
-  // defaults to false
-  // Cloudflare Turnstile and Google reCaptcha should be mutually exclusive
-  // todo: remove after full rollout
-  const enableTurnstileFeatureFlag = useIsFeatureEnabled(
```
sebastianwzq, line 73: same as above, injecting dependency instead of retrieving feature flag

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -74,23 +73,6 @@ const submitEmailModeForm: ControllerHandler<
     formId,
   }
 
-  const featureFlagsListResult = await FeatureFlagService.getEnabledFlags()
```
ghost, line 77: No longer required as frontend will tell backend which captcha provider to use

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -458,6 +470,74 @@ export const PublicFormProvider = ({
           // TODO (#5826): Toggle to use fetch for submissions instead of axios. If enabled, this is used for testing and to use fetch instead of axios by default if testing shows fetch is more  stable. Remove once network error is resolved
           if (useFetchForSubmissions) {
             return submitStorageFormWithFetch()
+          } else {
```
KenLSM: @sebastianwzq we might have selected the wrong code during the rebase just now 🙈 

[this](https://github.com/opengovsg/FormSG/pull/6502/files#diff-30884334e266c2f74390ffef045635fbdda34c02ce4bc937b5365e54de57bf5bR556) might be where the `captchaType` should be added.

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -145,15 +145,22 @@ export const PublicFormProvider = ({
     featureFlags.turnstile,
     false,
   )
+  enum CaptchaTypes {
```
KenLSM: Artifact from refactoring? -- forgotten to remove this declaration and instead importing from types
Seems like this had been declared in `src/app/services/captcha/captcha.types.ts`.

frontend/src/features/public-form/PublicFormService.ts
```diff
@@ -148,6 +152,7 @@ export const submitStorageModeForm = async ({
     {
       params: {
         captchaResponse: String(captchaResponse),
+        captchaType: captchaType,
```
KenLSM: /nit 
We can use object shorthand here. 
```diff
 {
      params: {
        captchaResponse: String(captchaResponse),
-        captchaType: captchaType, 
+        captchaType,
```

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -133,34 +115,38 @@ const submitEmailModeForm: ControllerHandler<
...
-              getRequestIp(req),
-            )
-              .map(() => form)
-              .mapErr((error) => {
-                logger.error({
-                  message: 'Error while verifying captcha',
-                  meta: logMeta,
-                  error,
+            }
+            // defaults to recaptcha
```
KenLSM: /nit wondering if we should be explicit and declare `CaptchaTypes.Captcha` as fall-through to the default.

src/app/modules/submission/email-submission/email-submission.controller.ts
```diff
@@ -454,6 +440,7 @@ const submitEmailModeForm: ControllerHandler<
 }
 
 export const handleEmailSubmission = [
+  // todo: remove CaptchaMiddleware after extracting common components in Captcha and Turnstile
```
KenLSM: /nit Does your plugins support lowercase todos? Mine (vscode) doesn't capture lowercase `todo`, only uppercase `TODO` 🤔

src/app/services/turnstile/turnstile.middleware.ts
```diff
@@ -3,5 +3,6 @@ import { celebrate, Joi, Segments } from 'celebrate'
 export const validateTurnstileParams = celebrate({
   [Segments.QUERY]: Joi.object({
     captchaResponse: Joi.string().allow(null).required(),
+    captchaType: Joi.string().allow('').required(),
```
KenLSM: Might have some validation issue where users who were loading older FE (that doesn't supply `captchaType`) hitting the latest BE (that requires `captchaType`). The fix for FE is to refresh the page (which may require them to lose all working data)

src/app/modules/submission/email-submission/email-submission.controller.ts
ghost: Considered defining CaptchaTypes.Captcha explicitly as a case but I wanted to fallback to recaptcha regardless of types

src/app/modules/submission/email-submission/email-submission.controller.ts
ghost: Mine does but I will change it to uppercase

src/app/services/turnstile/turnstile.middleware.ts
sebastianwzq: Very good catch, I think I will have to leave it as an optional parameter for starters. Thoughts?

frontend/src/features/public-form/PublicFormProvider.tsx
ghost: Wanted to separate the usage of types from the frontend and backend codebase. Refactored and moved the declaration to the shared folder as discussed.

src/app/services/turnstile/turnstile.middleware.ts
ghost: Good catch, I intended to leave it as optional hence the default fallback to recaptcha but missed this here. Thank you

frontend/src/features/public-form/PublicFormService.ts
```diff
@@ -221,6 +229,7 @@ export const submitStorageModeFormWithFetch = async ({
   // Add captcha response to query string
   const queryString = new URLSearchParams({
     captchaResponse: String(captchaResponse),
+    captchaType: captchaType,
```
KenLSM: /nit can use object shorthand

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -149,55 +148,45 @@ const submitEncryptModeForm: ControllerHandler<
...
-
-      if (captchaResult.isErr()) {
-        logger.error({
-          message: 'Error while verifying captcha',
-          meta: logMeta,
-          error: captchaResult.error,
-        })
-        const { errorMessage, statusCode } = mapRouteError(captchaResult.error)
-        return res.status(statusCode).json({ message: errorMessage })
+      // defaults to recaptcha
```
KenLSM: Same here as well. For explicit fall-through for searchability

src/app/services/captcha/captcha.middleware.ts
```diff
@@ -3,5 +3,6 @@ import { celebrate, Joi, Segments } from 'celebrate'
 export const validateCaptchaParams = celebrate({
   [Segments.QUERY]: Joi.object({
     captchaResponse: Joi.string().allow(null).required(),
+    captchaType: Joi.string().allow('').required(),
```
KenLSM: This probably also need to be non-compulsory

KenLSM (approved):
LGTM (aside from highlighted changes)!

PR #6515, opened 2023-07-06
fix: set decompress to false

sebastianwzq:
## Problem
[GTA-10-009 WP1](https://docs.google.com/document/d/1G33977oPLMzGH78hrzxivRF5Wtrgfs-3sLaBlxFUFR0/edit)
Decompress defaults to true although we are not using the response from webhook requests

Closes FRM-1116, FRM-1117

## Solution
Set decompress to false and maxContentLength to 10 kb which should be sufficient for typical requests. This is in consideration that we do not require the response from webhook requests apart from storing it for archiving purposes.

## Test plan

- [ ] Successful webhook call

### DIFF
--- a/src/app/modules/webhook/webhook.service.ts
+++ b/src/app/modules/webhook/webhook.service.ts
@@ -156,6 +156,8 @@ export const sendWebhook = (
                 signature,
               }),
             },
+            decompress: false,
+            maxContentLength: 10000,
             maxRedirects: 0,
             // Timeout after 10 seconds to allow for cold starts in receiver,
             // e.g. Lambdas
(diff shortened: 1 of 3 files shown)


src/app/modules/webhook/webhook.service.ts
```diff
@@ -156,6 +156,8 @@ export const sendWebhook = (
                 signature,
               }),
             },
+            decompress: false,
+            maxContentLength: 10000,
```
KenLSM: Might want to move this `10000` to a constant under `webhook/webhook.constants.ts`

(how was this 10000 derived?)

KenLSM (changes requested):

src/app/modules/webhook/webhook.service.ts
ghost: Will move the configuration as suggested.

The webhook response is mostly expected to be empty. A quick check on the recent webhook responses in our prod analytics DB also indicates that the response is empty.

Given that the response of the webhook call is not used and does not impact our partner's processing of the call, I have revised the limit to 1MB, which should accommodate most webhook responses. What are your thoughts?

KenLSM: I see. Thanks for checking! I think this arrangement makes sense.

KenLSM (approved):

PR #6516, opened 2023-07-06
feat: create get api for listing and downloading issues

mayying:
Create GET API to list and download issues for admin to view in portal

Closes FRM-1025

### DIFF
--- a/src/app/models/form_issue.server.model.ts
+++ b/src/app/models/form_issue.server.model.ts
@@ -1,6 +1,10 @@
-import { Mongoose, Schema } from 'mongoose'
+import { Mongoose, QueryCursor, Schema } from 'mongoose'
 
-import { IFormIssueModel, IFormIssueSchema } from 'src/types'
+import {
+  FormIssueStreamData,
+  IFormIssueModel,
+  IFormIssueSchema,
+} from 'src/types'
 
 import { FORM_SCHEMA_ID } from './form.server.model'
 
@@ -36,7 +40,22 @@ const FormIssueSchema = new Schema<IFormIssueSchema, IFormIssueModel>(
 FormIssueSchema.index({ formId: 1 })
 
 /**
- * Form Issue Schema
+ * Returns a cursor for all issues for the form with formId.
+ * @param formId the form id to return the cursor for
+ * @returns a cursor to the issue retrieved
+ */
+FormIssueSchema.statics.getIssueCursorByFormId = function (
+  formId: string,
+): QueryCursor<FormIssueStreamData> {
+  return this.find({ formId }, { issue: 1, email: 1, created: 1 })
+    .batchSize(2000)
+    .read('secondaryPreferred')
+    .lean()
+    .cursor()
+}
+
+/**
+ * Form issue Schema
  * @param db Active DB Connection
  * @return Mongoose Model
  */
--- a/src/app/models/form_issue.server.model.ts
+++ b/src/app/models/form_issue.server.model.ts
@@ -42,14 +42,16 @@ FormIssueSchema.index({ formId: 1 })
 /**
  * Returns a cursor for all issues for the form with formId.
  * @param formId the form id to return the cursor for
+ * @param fields an array of field names to retrieve
  * @returns a cursor to the issue retrieved
  */
 FormIssueSchema.statics.getIssueCursorByFormId = function (
   formId: string,
+  fields: (keyof IFormIssueSchema)[],
 ): QueryCursor<FormIssueStreamData> {
-  return this.find({ formId }, { issue: 1, email: 1, created: 1 })
+  return this.find({ formId }, fields)
     .batchSize(2000)
-    .read('secondaryPreferred')
+    .read('secondary')
     .lean()
     .cursor()
 }
--- a/src/app/modules/form/admin-form/__tests__/admin-form.issue.controller.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.issue.controller.spec.ts
@@ -1,0 +1,323 @@
+import { ObjectId } from 'bson-ext'
+import { StatusCodes } from 'http-status-codes'
+import { errAsync, okAsync } from 'neverthrow'
+import { PassThrough } from 'stream'
+
+import dbHandler from '../../../../../../__tests__/unit/backend/helpers/jest-db'
+import expressHandler from '../../../../../../__tests__/unit/backend/helpers/jest-express'
+import { FormIssueMetaDto } from '../../../../../../shared/types'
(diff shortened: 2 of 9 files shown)


src/app/models/form_issue.server.model.ts
```diff
@@ -36,7 +40,22 @@ const FormIssueSchema = new Schema<IFormIssueSchema, IFormIssueModel>(
...
 /**
- * Form Issue Schema
+ * Returns a cursor for all issues for the form with formId.
+ * @param formId the form id to return the cursor for
+ * @returns a cursor to the issue retrieved
+ */
+FormIssueSchema.statics.getIssueCursorByFormId = function (
+  formId: string,
+): QueryCursor<FormIssueStreamData> {
+  return this.find({ formId }, { issue: 1, email: 1, created: 1 })
```
KenLSM: Is the selection portion required to be specified here? Wondering if we need to whitelist the fields here

src/app/models/form_issue.server.model.ts
```diff
@@ -36,7 +40,22 @@ const FormIssueSchema = new Schema<IFormIssueSchema, IFormIssueModel>(
...
+    .batchSize(2000)
+    .read('secondaryPreferred')
```
KenLSM: Not exactly sure if there's any specific impact if we use `secondary`, but from other queries, it seemed like non-critical queries are using `secondary` instead of `sp`

mayying: using `secondary` means if secondary is not avail, we fail the request. If we use `sp`, we are allowing mongo to use primary if secondary is not avail, since the query is indexed. Ref [here](https://www.mongodb.com/docs/manual/core/read-preference/#:~:text=Starting%20in%20version%204.4%2C%20secondary%20supports%20hedged%20reads%20on%20sharded%20clusters.&text=In%20most%20situations%2C%20operations%20read%20from%20secondary%20members%20but%20if,hedged%20reads%20on%20sharded%20clusters.)

src/app/models/form_issue.server.model.ts
mayying: It's not really nice and not needed to throw the entire database columns to FE, exposing the internal metadata. So I just project whatever is needed.

src/app/modules/form/admin-form/admin-form.issue.controller.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @security session
+ *
+ * @returns 200 with issue stream
+ * @returns 403 when user does not have permissions to access form
+ * @returns 404 when form cannot be found
+ * @returns 410 when form is archived
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database or stream error occurs
+ */
+export const handleStreamFormIssue: ControllerHandler<
```
KenLSM: /nit since we are returning multiple issues, does it make sense for this to be plural? (similarly for `handleGetFormIssue`)

src/app/modules/issue/issue.service.ts
KenLSM, line 1: I feel that we shouldn't be taking the namespace of `modules/issue` for this as it's closely-related to `form issues`.

src/app/models/form_issue.server.model.ts
KenLSM: Should this be done at the controller or service instead?

src/app/models/form_issue.server.model.ts
KenLSM: Yeah I know the difference, but any reason why we wanna deviate from using `secondary`? The query can be indexed, but the payload transfer will still impact it (although it is unlikely that it will be an issue here, pun not intended)

src/app/modules/form/admin-form/admin-form.issue.controller.ts
mayying: it does, but im following FormFeedback style. Should i follow the existing or add new convention? 🤔

src/app/modules/issue/issue.service.ts
mayying: do u mean i should put it under form? i was following feedback actually..

src/app/modules/form/admin-form/__tests__/admin-form.issue.controller.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+    const MOCK_REQ = expressHandler.mockRequest({
+      params: { formId: MOCK_FORM_ID },
+      session: {
+        user: {
+          _id: MOCK_USER_ID,
+        },
+      },
+    })
+
+    it('should return issues as request is processed successfully', async () => {
```
KenLSM: /nit should return issues `as -> when` request is ...

We usually use `when` which describes the arrange section

src/app/models/form_issue.server.model.ts
mayying: good point, let me change this.

src/app/models/form_issue.server.model.ts
mayying: Because I think it does not make sense to fail the query when primary is up. But if you think we should stick with `secondary` I am ok to change it

KenLSM: I think that there's merit in such queries not affecting `primary`. But not a strong opinion! I'll leave the final say to you

src/app/modules/form/admin-form/admin-form.issue.controller.ts
KenLSM: Technically, feedback has no plural 😆 (something like there's no foods, only food).

src/app/modules/issue/issue.service.ts
KenLSM: Discussion done offline.

KenLSM (approved):
LGTM aside from some nits

src/app/modules/form/admin-form/__tests__/admin-form.issue.controller.spec.ts
mayying: done https://github.com/opengovsg/FormSG/pull/6516/commits/dad6758c7ee8e83eb6e270976ea80879b5a3c8c5

src/app/modules/form/admin-form/admin-form.issue.controller.ts
mayying: done https://github.com/opengovsg/FormSG/pull/6516/commits/dad6758c7ee8e83eb6e270976ea80879b5a3c8c5

src/app/models/form_issue.server.model.ts
mayying: will revert to secondary since others is doing that too.

PR #6517, opened 2023-07-07
feat: issue feedback admin page

mayying:
## Problem
Create issue listing and download UI in admin view page under response tab. Under this PR, public users are still not able to submit issues, the feature will be enabled once https://github.com/opengovsg/FormSG/pull/6492 and https://github.com/opengovsg/FormSG/pull/6452 are pushed to develop

Closes FRM-1103

**Breaking Changes** 

**Features**:

- Admin is able to view form issues and optional email content reported under Response tab
- Admin is able to export form issues by pressing the Export button while in the Issue view
- The existing Response is now tucked under Review tab, and the default view will be Issue

## Before & After Screenshots

**BEFORE**:
image

When there are issues
**AFTER**:
image

When there isn't any issues and review
image

When there isn't any issue, but review exists
image
image

## Tests
Listing Issues
- [ ] Create a new form and view the Results > Feedback tab, observe that `You don't have any feedback yet` is shown.
- [ ] Make the form public and fill in the form as public users, and provide some review and rating after submitting the form.
- [ ] Go to the Results > Feedback tab, see that there is `0 issues to date`.
- [ ] Observe that Export button is greyed out as there're no issues to export.
- [ ] Click onto the Review button, see that the reviews and ratings entered earlier is listed on the page.
- [ ] Click onto the Export button, the reviews and ratings should be properly exported.

### DIFF
--- a/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
+++ b/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
@@ -82,11 +82,11 @@ EmptyReviewAndIssue.parameters = {
 }
 
 export const EmptyReview = Template.bind({})
-EmptyReviewAndIssue.parameters = {
+EmptyReview.parameters = {
   msw: [getEmptyAdminFormFeedback(), ...DEFAULT_MSW_ROUTES],
 }
 export const EmptyIssue = Template.bind({})
-EmptyReviewAndIssue.parameters = {
+EmptyIssue.parameters = {
   msw: [getEmptyAdminFormIssue(), ...DEFAULT_MSW_ROUTES],
 }
 
@@ -104,7 +104,7 @@ Mobile.parameters = getMobileViewParameters()
 export const LoadingDesktop = Template.bind({})
 LoadingDesktop.storyName = 'Loading/Desktop'
 LoadingDesktop.parameters = {
-  msw: [getAdminFormFeedback({ delay: 'infinite' }), ...DEFAULT_MSW_ROUTES],
+  msw: [getAdminFormIssue({ delay: 'infinite' }), ...DEFAULT_MSW_ROUTES],
 }
 export const LoadingMobile = Template.bind({})
 LoadingMobile.storyName = 'Loading/Mobile'
--- a/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
+++ b/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
@@ -6,7 +6,9 @@ import {
   createFormBuilderMocks,
   getAdminFormCollaborators,
   getAdminFormFeedback,
+  getAdminFormIssue,
   getEmptyAdminFormFeedback,
+  getEmptyAdminFormIssue,
   getStorageSubmissionMetadataResponse,
 } from '~/mocks/msw/handlers/admin-form'
 import { getUser } from '~/mocks/msw/handlers/user'
@@ -25,6 +27,7 @@ const DEFAULT_MSW_ROUTES = [
   ...createFormBuilderMocks({}, 0),
   getStorageSubmissionMetadataResponse(),
   getAdminFormFeedback(),
+  getAdminFormIssue(),
   getUser(),
   getAdminFormCollaborators(),
 ]
@@ -69,9 +72,22 @@ const Template: Story = () => {
 }
 export const Default = Template.bind({})
 
-export const EmptyFeedback = Template.bind({})
-EmptyFeedback.parameters = {
+export const EmptyReviewAndIssue = Template.bind({})
+EmptyReviewAndIssue.parameters = {
+  msw: [
+    getEmptyAdminFormIssue(),
+    getEmptyAdminFormFeedback(),
+    ...DEFAULT_MSW_ROUTES,
+  ],
+}
+
+export const EmptyReview = Template.bind({})
+EmptyReviewAndIssue.parameters = {
   msw: [getEmptyAdminFormFeedback(), ...DEFAULT_MSW_ROUTES],
+}
+export const EmptyIssue = Template.bind({})
+EmptyReviewAndIssue.parameters = {
+  msw: [getEmptyAdminFormIssue(), ...DEFAULT_MSW_ROUTES],
 }
 
 export const Tablet = Template.bind({})
--- a/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
+++ b/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
@@ -139,7 +139,7 @@ export const FeedbackPage = (): JSX.Element => {
     >
       <Grid
         mb="1rem"
-        minH="4rem"
+        minH={{ md: '4rem' }}
         alignItems="end"
         color="secondary.500"
         gridTemplateColumns={{ base: 'auto', md: 'auto 7fr 1fr' }}
@@ -167,6 +167,7 @@ export const FeedbackPage = (): JSX.Element => {
               currentFeedbackType,
               FeedbackType.Issues,
             )}
+            sx={{ borderRightWidth: '0px' }}
             onClick={() => setCurrentFeedbackType(FeedbackType.Issues)}
           >
             Issues
@@ -255,7 +256,7 @@ const getReviewInformationComponent = (
           <Text as="span" color="primary.500">
             {count}
           </Text>
-          {simplur` ${[count || 0]}feedback submission[|s] to date`}
+          {simplur` ${[count || 0]}review[|s] to date`}
         </Text>
       </Box>
     </Grid>
@@ -266,20 +267,20 @@ const getIssueInformationComponent = (
   count: number | undefined,
 ): JSX.Element => {
   return (
-    <Box>
-      <Text textStyle="h4" mb="0.5rem">
+    <Box display="flex" alignItems="center" mb="0.5rem">
+      <Text textStyle="h4">
         <Text as="span" color="primary.500">
           {count}
         </Text>
         {simplur` ${[count || 0]}issue[|s] to date`}
-        <Tooltip
-          label={`Feedback displayed here relates to form submission issues`}
-          placement="top"
-          textAlign="center"
-        >
-          <Icon as={BxsInfoCircle} aria-hidden marginX="0.5rem" />
-        </Tooltip>
       </Text>
+      <Tooltip
+        label={`Feedback displayed here relates to form submission issues`}
+        placement="top"
+        textAlign="center"
+      >
+        <Icon as={BxsInfoCircle} aria-hidden marginX="0.5rem" />
+      </Tooltip>
     </Box>
   )
 }
--- a/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
+++ b/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
@@ -142,26 +142,21 @@ export const FeedbackPage = (): JSX.Element => {
         minH={{ md: '4rem' }}
         alignItems="end"
         color="secondary.500"
-        gridTemplateColumns={{ base: 'auto', md: 'auto 7fr 1fr' }}
+        gridTemplateColumns={{ base: 'auto', md: '1fr auto auto' }}
         gridGap={{ base: '0.5rem', md: '1.5rem' }}
         gridTemplateAreas={{
-          base: "'information' 'feedbackType' 'export'",
+          base: "'information information' 'feedbackType export'",
           md: "'information feedbackType export'",
         }}
       >
-        <Container gridArea="information">
+        <Box gridArea="information" pl="0rem">
           {getInformationGridComponent(
             currentFeedbackType,
             issueProps,
             reviewProps,
           )}
-        </Container>
-        <ButtonGroup
-          gridArea="feedbackType"
-          justifySelf="flex-end"
-          isAttached
-          variant="outline"
-        >
+        </Box>
+        <ButtonGroup gridArea="feedbackType" isAttached variant="outline">
           <Button
             {...getFeedbackTypeButtonProps(
               currentFeedbackType,
@@ -236,7 +231,7 @@ const getReviewInformationComponent = (
 ): JSX.Element => {
   return (
     <Grid
-      gridTemplateColumns={{ base: 'auto', md: 'auto auto' }}
+      gridTemplateColumns={{ base: 'auto', md: 'auto 1fr' }}
       gridGap={{ base: '0.5rem', md: '1.5rem' }}
       gridTemplateAreas={{
         base: "'submissions' 'score'",
--- a/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
+++ b/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
@@ -185,6 +185,7 @@ export const FeedbackPage = (): JSX.Element => {
               reviewProps,
             )}
             handleClick={handleFeedbackDownloadClick}
+            isMobile={isMobile}
           />
         </Box>
       </Grid>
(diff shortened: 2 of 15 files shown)


frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueTable.tsx
```diff
@@ -0,0 +1,39 @@
...
+  },
+  {
+    Header: 'Date',
+    accessor: 'timestamp',
+    sortType: 'number',
+    Cell: DateCell,
+    minWidth: 80, // minWidth is only used as a limit for resizing
+    width: 80, // width is used for both the flex-basis and flex-grow
+    maxWidth: 120, // maxWidth is only used as a limit for resizing
+  },
```
wanlingt, line 22: nit: can consider extracting these since they're also used in `ReviewTable`

frontend/src/features/admin-form/responses/FeedbackPage/review/ReviewTable.tsx
```diff
@@ -0,0 +1,39 @@
...
+    maxWidth: 600,
+  },
+  {
+    Header: 'Rating',
+    accessor: 'rating',
+    sortType: 'number',
+    minWidth: 90,
+    width: 90,
+    disableResizing: true,
+  },
```
wanlingt, line 38: noticed that the widths are different from the last column (Contact) in IssueTable, is this expected? they look like they have the same width in figma!

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueService.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return ApiService.get<FormIssueDto[]>(
+    `${ADMIN_FORM_ENDPOINT}/${formId}/issues/download`,
+  ).then(({ data }) => {
+    const csvGenerator = new IssueCsvGenerator(count)
+
+    data.forEach((issue) => {
+      csvGenerator.addLineFromIssue(issue)
+    })
+
+    csvGenerator.triggerFileDownload(`${formTitle}-${formId}-issue.csv`)
```
wanlingt, line 37: (non-blocking / can be a separate PR) - wondering if it'd be useful to add logs for this!

frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
```diff
@@ -69,10 +72,23 @@ const Template: Story = () => {
...
+    getEmptyAdminFormIssue(),
+    getEmptyAdminFormFeedback(),
+    ...DEFAULT_MSW_ROUTES,
+  ],
+}
+
+export const EmptyReview = Template.bind({})
+EmptyReview.parameters = {
   msw: [getEmptyAdminFormFeedback(), ...DEFAULT_MSW_ROUTES],
 }
```
LinHuiqing, line 87: This doesn't select the review tab, so the chromatic shows a filled Issues tab instead of an empty Reviews tab. Modification required?

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueService.tsx
mayying: May I know what kind of logs do you refer to? Like info logs?

frontend/src/features/admin-form/responses/FeedbackPage/review/ReviewTable.tsx
mayying: Actually the current Rating column in our Form website is much smaller than the one on Figma. If I use the same column width that is configured, almost all emails will be wrapped. Also it makes sense to have Rating as a smaller column as it's always a single numeric character. But for email it can be wider. 
 image

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
```diff
@@ -102,3 +228,135 @@ export const FeedbackPage = (): JSX.Element => {
...
+          <Tooltip
+            label={`Feedback displayed here relates to form submission issues`}
+            placement="top"
+            textAlign="center"
+          >
+            <Icon as={BxsInfoCircle} aria-hidden marginX="0.5rem" />
+          </Tooltip>
+        </Text>
+      </Box>
+    </div>
```
LinHuiqing: Unnecessary wrapping `div`. Also, you can use `<></>` next time if a wrapping div is really needed.
```suggestion
    <Box>
      <Text textStyle="h4" mb="0.5rem">
        <Text as="span" color="primary.500">
          {count}
        </Text>
        {simplur` ${[count || 0]}issue[|s] to date`}
        <Tooltip
          label={`Feedback displayed here relates to form submission issues`}
          placement="top"
          textAlign="center"
        >
          <Icon as={BxsInfoCircle} aria-hidden marginX="0.5rem" />
        </Tooltip>
      </Text>
    </Box>
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
```diff
@@ -57,43 +139,87 @@ export const FeedbackPage = (): JSX.Element => {
...
-            </Text>
-            {prettifiedFeedbackCount}
-          </Text>
-        </Box>
+        <Container gridArea="information">
+          {getInformationGridComponent(
+            currentFeedbackType,
+            issueProps,
+            reviewProps,
+          )}
```
LinHuiqing, line 157: Can be simplified into:
```suggestion
          {currentFeedbackType === FeedbackType.Issues ? getIssueInformationComponent(issueProps.count) : getReviewInformationComponent(reviewProps.average, reviewProps.count)}
```
Then the function definition for `getInformationGridComponent` below can be removed as well.

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueTable.tsx
mayying: Hmm. Actually the columns are defined separately in DTO, and the accessor relies on the naming from the respective DTO. It can be misleading if I abstract them to another constant, a future column change to table would cause accidental change to another table. Wdyt?

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
```diff
@@ -102,3 +228,135 @@ export const FeedbackPage = (): JSX.Element => {
...
+): JSX.Element => {
+  if (currentFeedbackType === FeedbackType.Issues) {
+    return getIssueInformationComponent(issueProps.count)
+  }
+  return getReviewInformationComponent(reviewProps.average, reviewProps.count)
+}
+
+const getFeedbackTypeButtonProps = (
+  currentFeedbackType: FeedbackType,
+  requiredFeedbackType: FeedbackType,
```
LinHuiqing: (non-blocking) nit: nomenclature doesn't seem very intuitive to me - what about `selectedFeedbackType` and `feedbackType`?

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
```diff
@@ -102,3 +228,135 @@ export const FeedbackPage = (): JSX.Element => {
...
+  if (currentFeedbackType === requiredFeedbackType) {
+    return {
+      colorScheme: 'primary',
+      isActive: true,
+    }
+  }
+  return {
+    colorScheme: 'secondary',
+    isActive: false,
+  }
```
LinHuiqing: nit: can be simplified to something like this:
```suggestion
  return (currentFeedbackType === requiredFeedbackType)
    ? {
      colorScheme: 'primary',
      isActive: true,
    } : {
      colorScheme: 'secondary',
      isActive: false,
    }
```

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
```diff
@@ -102,3 +228,135 @@ export const FeedbackPage = (): JSX.Element => {
...
+}
+
+const getFeedBackDownloadButtonProps = (
+  currentFeedbackType: FeedbackType,
+  issueProps: Issue,
+  reviewProps: Review,
+): {
```
LinHuiqing, line 314: Maybe we can refactor this to pass only 2 args into it - 1) the respective isLoading state and 2) the count. The logic can be abstracted to be outside of this function itself after all.

LinHuiqing (commented):
Can be in a separate PR - the existing FeedbackTab tests should be updated to show how it looks when the reviews are populated as well.

frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
mayying: Just for clarification, do you mean the EmptyIssue? There is not route for Review tab, so the default page will always be Issue tab.

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
mayying: Actually I did this in purpose, it looks too cluttered especially when the texts are long, wdyt?

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
mayying: I dont have a strong preference, will just follow your suggestion

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
mayying: Done

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
mayying: My thought is that the caller is already very cluttered, I want to abstract the nitty gritty details out, by looking at the original caller, we do not need to know how the props are constructed, just that it is constructed. lmk wdyt

mayying:
> Can be in a separate PR - the existing FeedbackTab tests should be updated to show how it looks when the reviews are populated as well.

You mean the storybook right? I think it is updated? Please lmk if i misunderstand anything.

frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
LinHuiqing: Ah okay. In this case, maybe we can change the frontend test to test the issues tab component itself and the review tab component itself. Or set up anchoring for issue / review tab so they have routes. Can be done in a separate PR to prevent this PR from getting too big 👍

LinHuiqing:
> You mean the storybook right? I think it is updated? Please lmk if i misunderstand anything.

@mayying yep, but my concern is more of the same as the other stuff I wrote for the frontend tests. Basically, the way the tests are done now means chromatic has no visibility for the reviews tab, so it won't flag any regressions for that tab. Can be done in the separate PR bah.

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueTable.tsx
wanlingt: not fussed! can leave as it is

frontend/src/features/admin-form/responses/FeedbackPage/review/ReviewTable.tsx
wanlingt: will defer to @staceytan1998 for this

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueService.tsx
wanlingt: yup, info logs! can search `datadogLogs` in the codebase :)

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
LinHuiqing: Okay I see what you mean. Since there's a lot of repeated if/else logic for for `currentFeedbackType`, how about we use a utility function across all these? Something like this:
```ts
const getFeedbackProps = (
  currentFeedbackType: FeedbackType,
  issueProps: Issue,
  reviewProps: Review,
): {
  if (currentFeedbackType === FeedbackType.Issues) return issueProps
  else return reviewProps
}
```
Then we can use this across all the other functions that are dependent on the `currentFeedbackType` haha

mayying: we could do that, but i am not sure what is the added value it brings here?

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueService.tsx
mayying: wow, til, what is the diff between `datadogLogs` and the `logger` that I have been using?

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueTable.tsx
mayying: alright, thanks!

frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
mayying: Actually how does the frontend test works? Do you mean if the tab is not showing then the chromatic diff will not do the diff for us? Yeah, i wanted to create a new route initially but the changes will be much bigger, wanted to minimise the refactoring to aim for quicker launch.

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
LinHuiqing: To simplify the code since its repeated conditionals that we're doing. I think its cleaner this way at least, since we only do the if/else here in this 1 function instead of repeating it across all the different ones. Then the existing functions can be simplified to perform exactly what they're supposed to do without the conditional.
Just imo, wdyt?

frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
LinHuiqing: Yea if you don't see it rendered on chromatic, chromatic won't see it either, so it won't be able to compare the frontend stuff for us.
So to avoid adding routes just to test the UI, we can render the component itself and test the UI. For e.g. this is what we do for the [builder logic tab](https://www.chromatic.com/component?appId=60911f69d4403b0039b266a2&csfId=pages-adminformpage-create-logictab&buildNumber=4932&k=64abdba17e2b4d1dae169572-1200-interactive-true&h=4&b=-1).
Yea but as I mentioned, can leave this as separate issue for quicker launch! Rmb to assign Stacey as reviewer for the new tests that are added for the review tab frontend tests PR 👍

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueService.tsx
wanlingt: ya good qn!! `logger` is used in the backend, the logs go to AWS Cloudwatch. We have a lambda which pipes the Cloudwatch logs over to Datadog as well.

 `datadogLogs` are used in the frontend, the logs go directly to Datadog.

wanlingt (approved):
approving first to make it for release! also since @LinHuiqing 's comments can be addressed in a separate PR

frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
mayying: Alright, will work on this next, thanks!

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
mayying: but there is no inheritance to both props, so the caller will still need to check the type of the props to tell which property to use?

frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueService.tsx
mayying: oh, so you mean frontend code can use datadogLogs?

PR #6523, opened 2023-07-10
feat: MyInfo Children compound fields beta

kenjin-work:
## Problem
No MyInfo child fields.

Closes FRM-1019

## Solution
- Add a new `BasicField.Children` type.
- The field contains an array of `MyInfoChildAttribute/MyInfoAttribute`, which tells it what subfields to render.
- Submission data is an array of arrays. Where each subarray represents a child's data corresponding to the selected subfields (see code for example).
- These features are guarded behind a beta flag in the frontend. We don't need to guard in the backend because the beta flag is just to prevent inconsistencies during deployment and ensure backwards and forwards compatibility.

**Breaking Changes** 

With the beta flag it should be fully compatible (backwards or forwards).

**Features**:

- A new basic field type called "Children". This field is a compound field that can contain multiple subfields.
- The children field can be edited by the admin to select what subfields they want.
- Each subfield can be filled with MyInfo child birth records data and is also individually verified by MyInfo.
- New betaFlag `children`.
- The following children subfields are supported
    - name
    - dob
    - bc no/uinfin
    - vaccination status
    - gender
    - race
    - secondary race

## Before & After Screenshots

**BEFORE**:
See screenshots below with betaflag set to false.

**AFTER**:
(WITH BETAFLAG OFF)
MyInfo panel has no additional fields:
image

(WITH BETAFLAG ON)
MyInfo panel has a new field (Family (Children)):
image

Children field's edit side panel:
image

New field in Form builder:
image

Public form upon filling and selecting:
image

Email response:
image

## Tests
- [ ] Enable user.betaFlags.children
Admin form:
- [ ] Create an email mode form.
- [ ] Open admin view and enable MyInfo Singpass authentication.
- [ ] Check that a new Children field appears in the left side panel.
- [ ] Drag and drop the new field or click the new field to add it to the form.
- [ ] Try adding variable numbers of subFields.
- [ ] Toggle (enable) multiple children answers.
- [ ] Save the form. And open it to responses.

Public form:
- [ ] Open the form created in the above section.
- [ ] Log in using Singpass/mockpass
- [ ] Click the dropdown and select a child name.
- [ ] Fields should be auto populated with data. Those fields should be disabled.
- [ ] For fields that are not auto populated, they should be editable by user.
- [ ] Enter any value for those fields not yet populated.
- [ ] Submit the form.

Email response:
- [ ] Check that auto populated fields from above have the [MyInfo] label on them.
- [ ] Check that user-filled fields do not have [MyInfo] label on them.

Data collation tool:
- TODO in the other repo.

## Deploy Notes

When testing on staging, vaccination records are not supported and will error because our MyInfo test API wrongly thinks the scope is invalid. For that case, just omit testing that specific subfield.

## TODO (in order of priority):

The following will be done in some other PR because this one is already too big:
- Tests


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/EditFieldDrawer.tsx
```diff
@@ -98,7 +107,19 @@ interface MemoFieldDrawerContentProps {
 
 export const MemoFieldDrawerContent = memo<MemoFieldDrawerContentProps>(
   ({ field, ...props }) => {
+    const { user } = useUser()
     if (isMyInfo(field)) {
+      if (
+        field?.myInfo?.attr === MyInfoAttribute.ChildrenBirthRecords &&
+        user?.betaFlags?.children
```
timotheeg, line 114: if `user?.betaFlags?.children` is `false`, should that not just exit? What would line 123 show from the field?
```typescript
return <EditMyInfo {...props} field={field} />
```

frontend/src/features/admin-form/create/constants.ts
```diff
@@ -301,4 +309,47 @@ export const MYINFO_FIELD_TO_DRAWER_META: {
...
     isSubmitted: true,
   },
+  [MyInfoAttribute.ChildrenBirthRecords]: {
+    label: 'Child Records',
+    icon: BiGroup,
+    isSubmitted: true,
+  },
+
+  // The following child field's icons will never be used.
+  // So they can be any placeholder.
```
timotheeg, line 321: /nit 

> they can be any placeholder

They can, but could still be good to use at least semi-relevant icon, or named reference that indicate it's dummy. Otherwise people reading the code but missing this comment might wonder what's going on. Especially when CertNo is using a different icon than the rest (all calendars!), then it looks like there is a purpose to it.

```
const BiDummyIcon = BiCalendar // random icon that is not actually shown in app
```

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -0,0 +1,404 @@
...
+      formContext.setValue(
+        `${schema._id}.childFields`,
+        schema.childrenSubFields,
+      )
+    }
+  }, [schema.childrenSubFields, formContext, schema._id])
+
+  // Initialize with a single child section
+  useEffect(() => {
+    if (!fields || !fields.length || fields.length === 0) {
```
timotheeg: /nit `0` is falsy

```suggestion
    if (!fields || !fields.length) {
```

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+  return (
+    <FieldContainer
+      schema={schema}
+      {...fieldContainerProps}
+      errorKey={childrenInputName}
+    >
+      <VisuallyHidden id={`children-desc-${schema._id}`}>
+        {ariaChildrenDescription}
+      </VisuallyHidden>
```
timotheeg, line 120: Good stuff! which reminds me we should test all this for accessibility too. @foochifa should be able to help

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -0,0 +1,404 @@
...
+      return []
+    }
+    return myInfoChildrenBirthRecords[MyInfoAttribute.ChildName] ?? []
+  }, [myInfoChildrenBirthRecords])
+
+  // useCallback to re-compute names because for some reason watch doesn't
+  // work on this nested field in react-hook-form.
+  const allSelectedNames = useCallback((): string[] => {
+    // Really important to note that sometimes react-hook-form stores our "array"
+    // as a object with key=index and values=array entry.
```
timotheeg, line 229: > sometimes

Any pattern to note for this? Is it due to the time of instantiation or load?

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+  const childNamePath = useMemo(
+    () => `${schema._id}.child.${currChildBodyIdx}.0`,
+    [schema._id, currChildBodyIdx],
+  )
+  const { onChange: selectOnChange, ...selectRest } = register(childNamePath)
+
+  const childName = watch(childNamePath) as unknown as string
+
+  const allChilds = useMemo<string[]>(() => {
```
timotheeg: /nit funny plural? 😅 

```suggestion
  const allChildren = useMemo<string[]>(() => {
```

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+  // useCallback to re-compute names, again because of buggy allSelectedNames
+  const namesNotSelected = useCallback((): string[] => {
+    if (myInfoChildrenBirthRecords === undefined) {
+      return []
+    }
+    const temp = allSelectedNames()
+    // We want all child names that haven't already been selected.
+    // O(n^2) but n is small so it should be okay.
+    return allChilds.filter((name) => !temp.includes(name))
```
timotheeg: 👍 

just for a note, you could use a `Set` here here, but for small arrays, it's probably slower than filtering with `array.includes()` 🤷 

```suggestion
    const temp = new Set(allSelectedNames())
    // We want all child names that haven't already been selected.
    return allChilds.filter((name) => !temp.has(name))
```

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -0,0 +1,404 @@
...
+  const getChildAttr = useCallback(
+    (attr: MyInfoChildAttributes): string => {
+      if (myInfoChildrenBirthRecords === undefined) {
+        return ''
+      }
+      // First find the child
+      const indexOfChild =
+        myInfoChildrenBirthRecords[MyInfoChildAttributes.ChildName]?.indexOf(
+          childName,
+        )
```
timotheeg: So in small set sizes, this is probably very fast, but still, considering we know all the subfields are iterated on later on, we could have collated all the fields from the current child at bootstrap with a useMemo(), and then `getChildAttr()` could do a direct lookup on it rather than have to find the child index again for every subfield.

src/app/modules/myinfo/myinfo.adapter.ts
```diff
@@ -149,9 +179,55 @@ export const internalAttrToExternal = (attr: InternalAttr): ExternalAttr => {
...
+    | {
+        requirement: { code: string; desc: string }
+        fulfilled: { value: boolean }
+      }[],
+): MyInfoChildVaxxStatus => {
+  if (vaccinationRequirement === undefined || !vaccinationRequirement.length) {
+    return MyInfoChildVaxxStatus.Unknown
+  }
+  return vaccinationRequirement.some((req) => req?.requirement?.code === '1M3D')
+    ? MyInfoChildVaxxStatus.ONEM3D
```
timotheeg, line 226: LOL

timotheeg (commented):
Submitting some early comments for frontend first, while I'm still going over backend

src/app/modules/myinfo/myinfo.adapter.ts
```diff
@@ -173,6 +249,64 @@ export class MyInfoData
...
+    const records = this.#personData
+      .childrenbirthrecords as Array<MyInfoChildBirthRecordBelow21>
+    if (records === undefined) {
+      return []
+    }
+    // Note: need ?. operator because above 21 children may
+    // not have these fields.
+    switch (childAttr) {
+      case MyInfoChildAttributes.ChildName:
+        return records.map((c) => c?.name?.value ?? '')
```
timotheeg, line 270: Just to clarify: is it either of `c` or `c.name` that can be undefined, or if `c` is defined , `c.name` is guaranteed to exists?

src/app/utils/field-validation/validators/childrenValidator.ts
```diff
@@ -0,0 +1,128 @@
...
+) => ChildrenValidator
+
+/**
+ * Returns a validator to check if answerArray is empty
+ */
+const childrenAnswerValidator: ChildrenValidator = (response) => {
+  const { answerArray } = response
+
+  return answerArray.length === 0
+    ? left(`CheckboxValidator:\t Answer is empty array`)
```
timotheeg: `CheckboxValidator:` ? 😅 

(same comment on all the error blocks below)

src/app/utils/field-validation/validators/childrenValidator.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * Returns a validation function to check if the
+ * that the first answer subarray has length > 0.
+ */
+const validChildAnswerFirstArray: ChildrenValidator = (response) => {
+  const { answerArray } = response
+  const len = answerArray[0]?.length
+  return len !== undefined && len > 0
```
timotheeg: /nit

```suggestion
  const first = answerArray[0]
  return Array.isArray(first) && first.length > 0
```

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -58,16 +102,26 @@ const HASH_SALT_ROUNDS = 1
...
 ): MyInfoHashPromises => {
   const readOnlyHashPromises: MyInfoHashPromises = {}
 
   prefilledFormFields.forEach((field) => {
+    // For children fields, we need to explode their subfields.
+    if (
+      isMyInfoChildrenBirthRecords(field.myInfo?.attr) &&
+      childrenBirthRecords !== undefined
+    ) {
+      hashChildrenFieldValues(field, childrenBirthRecords, readOnlyHashPromises)
```
timotheeg, line 120: /nit I'm not a fan of passing in a object to be mutated, I would have probably kept `hashChildrenFieldValues()` to be a pure function and did the augmentation of `readOnlyHashPromises` here:
```suggestion
      Object.assign(
        readOnlyHashPromises,
        hashChildrenFieldValues(field, childrenBirthRecords)
      )
```

[Edit] I retract the comment. After seeing `compareHashedValues()` below, which is similar but harder to change, I think it's better to have the 2 similar functions be consistent in their processing.

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -83,9 +137,12 @@ const hasMyInfoAnswer = (
...
   const answer = field.answer
-  return field.fieldType === BasicField.Date
-    ? moment(new Date(answer)).format('YYYY-MM-DD')
-    : answer
+  switch (field.fieldType) {
+    case BasicField.Date:
+      return moment(new Date(answer)).format('YYYY-MM-DD')
+    default:
+      return field.answer
+  }
```
timotheeg: Why this change?

Also, default can be `return answer`

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -109,6 +166,12 @@ export const compareHashedValues = (
   const myInfoResponsesMap: MyInfoComparePromises = new Map()
   responses.forEach((field) => {
     if (hasMyInfoAnswer(field)) {
+      // Children birth records have multiple possible hash values so they
+      // need to be checked one by one.
+      if (field.myInfo.attr === MyInfoAttribute.ChildrenBirthRecords) {
+        handleMyInfoChildHashResponse(field, hashes, myInfoResponsesMap)
```
timotheeg, line 174: Same comment on mutating the Map, but I realize that in this case, there's no equivalent to `assign()` or `merge()` for Maps 😑 , so adding code here to do the merge is sort of polluting. 

I think you can ignore this comment and the previous one and leave as-is.

timotheeg (commented):
I think the PR looks very good @kenjin-work 

I only have minor nit comments and questions, and only a few changes are needed, like the `checkboxValidator` label (LOL)

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/EditFieldDrawer.tsx
kenjin-work: It shows the default MyInfo "Edit" side panel (which is the one that just shows some info like where the data originated from and doesn't let you edit anything).

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
kenjin-work: No, it just turns up when I access it certain ways (I forgot when and how, I just noticed it does 😢 )

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
kenjin-work: > Valentina Vassilyeva and her husband Feodor Vassilyev are alleged to hold the record for the most children a couple has produced. She gave birth to a total of 69 children

-Wikipedia.

Guess I will just use set for such cases. Though you're right it's probably slower for almost all cases.

src/app/utils/field-validation/validators/childrenValidator.ts
kenjin-work: Ah you caught where I copied the code from :).

src/app/modules/myinfo/myinfo.adapter.ts
kenjin-work: I have no clue (I don't think the MyInfo docs says and I don't trust that it's the case for 100% of people). So I'm just writing it like that to be sure. The code will likely never be needed.

src/app/modules/myinfo/myinfo.util.ts
kenjin-work: Woops stray change. Reverting. Thank you.

timotheeg (approved):
LGTM!

kenjin-work:
Note: although the PR does not show it, I went thru two rounds of design reviews (over Slack). So will merge.

PR #6532, opened 2023-07-12
feat: presub-change-icon-to-help

mayying:
Public user is seen using the message icon to message the admin wrt to the feedback of the form/the initiative that drives the form. However, that button is used to allow user to communicate issue they face while submitting the form. Hence, changing the icon to question mark as an effort to allow user to subconsciously know that the button should be used only when they have question.

Closes FRM-1163, Closes FRM-1166

Context: https://opengovproducts.slack.com/archives/CKHLS3W3X/p1689151781199949

Screenshot:
Before
image

After
image

Test:

- [ ] Check that the icon at the bottom right corner of public form is now a question mark instead of a message.


LinHuiqing (approved):
Lgtm, can you add a screenshot of how this looks like in the PR?

PR #6533, opened 2023-07-12
fix: presub remove additional underscore

mayying:
Remove additional underscore at white space. Credits to @LinHuiqing for noticing it.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests

- [ ] Report an issue in any public form (make sure this is the first issue reported for the day), notice the underscore is only for the word `Login to forms`


LinHuiqing (approved):
LGTM! 🤣

Issue #5919, opened 2023-03-13
Strengthen UEN verification


KenLSM:
**Is your feature request related to a problem? Please describe.**
To strengthen our UEN verification.

We are currently checking the UEN format through [ACRA](https://www.uen.gov.sg/ueninternet/faces/pages/admin/aboutUEN.jspx), however, we do not ensure if the UEN number is valid. Even though the we do not have the formula for checksum generation, we can reverse engineer it through the vast amount of dataset that's available on [data.gov.sg](https://data.gov.sg/dataset/entities-with-unique-entity-number).

**Describe the solution you'd like**
We chanced upon the open sourced [arthurdejong/python-stdnum](https://github.com/arthurdejong/python-stdnum) project that collected a whole bunch of standard formats, and included the checksum algorithm for UEN. We would want to verify the checksum against the dataset, and subsequently include the checksum algorithm into our validator.

packirisamykaran:
Hi, I would like to contribute to this project. Guide me on how I can help fix this issue

hui-jie-lim:
hi @KenLSM , i think i can pick this up if no ones working on this yet. Will need some help navigating like if there are existing checksums and where I can find them. I can have a look at the package and verify it first before working on the integration

KenLSM:
Hey @hui-jie-lim, we don't have anyone working on this in the moment. 

From our initial research, the checksum library provided by arthurdejong looked to be promising. However, as the library is not in typescript, you will most likely need to port the relevant sections of the code over to typescript, after your verification.

hui-jie-lim:
Heyy @KenLSM , I did a round of validation (only for correct UENs, not wrong ones). (See [this](https://github.com/hui-jie-lim/sg_uen_checker/tree/main)). 

Its generally sound but UENs of type UF (unregistered foreign entities) will return an error. I think its fine since these UENs cannot be used to transact with gov agencies but I am not sure if will affect the functionality of FormSG. 

Alternatively, I can surpress the error by adding 'UF' to OTHER_UEN_ENTITY_TYPES

Let me know how the team would like to proceed. Thanks!

Also, just to check - we update here? 
`FormSG/blob/develop/shared/utils/uen-validation.ts`

KenLSM:
Hey @hui-jie-lim, the lib test looks good. Would be better if we can have some test with some failure cases (perhaps 
1) switching the entity types to AA/BB which definitely doesn't exist, and 2) changing the checksum to another character).

[...]

PR #6534, opened 2023-07-12
feat: updated uen validator (ref stdnum) and tests cases 

hui-jie-lim:
## Problem

Closes [#5919]

## Solution

1. Implemented UEN validator based on [stdnum](https://github.com/arthurdejong/python-stdnum/blob/master/stdnum/sg/uen.py)
2. Added test for uen-validator, and new script for standalone sharedutils (npm run test:sharedutils), along with corresponding config files

**Breaking Changes** 

    - Details ...

**Features**:

- uen-validation

## Tests

Test comprises 2 happy and 1 unhappy test cases. Happy test cases are based on UEN codes that are known to be valid issued by ACRA and Other entities. Unhappy test case checks if uen-validator captures the checksum accurately for all types if UEN. (5XXX, 2XXX, ROC, ROB, T and S entities.) See shared/utils/__tests__/uen-validation.spec.ts

May run as a check
``` npm run test:backend shared```

### CODE REVIEW

**Comment by hui-jie-lim** [comment]
Hi @KenLSM , 

Please have a look at this PR for #5919, thank you. There seem to be a lot of new additions to package-lock.json. I think it might have been the additions generated by my local builds. If its alright, perhaps you can help to review the code first? I'll fix the package-lock file subsequently. Thanks!

Hui

File: package.json [hunk 1]
```diff
@@ -26,6 +26,7 @@
   "scripts": {
     "postinstall": "npm run postinstall:frontend && npm run postinstall:shared",
     "test": "npm run test:backend && npm run test:frontend",
+    "test:sharedutils": "npm --prefix shared test",
```
KenLSM: This might not be needed, as the `test:backend` script would be covering this section during the CIs.

If the intention was to exclusively run tests from on `shared/` locally we can use this command instead `npm run test:backend shared` or `npm run test:backend:watch shared` to run with `watch mode` on.

To run tests on a specific files, say for local dev, this can might be more suitable `npm run test:backend shared/utils/__tests__/uen-validation.spec.ts`

package.json
KenLSM, line 1: I've triggered the CI workflow, thus you might be seeing additional linting errors that were not raised up during dev. Might need to check your eslint/prettier/typescript config

shared/utils/__tests__/uen-validation.spec.ts
```diff
@@ -0,0 +1,43 @@
...
...9B','201707905H','53339181D','53339185K','53339191A','201707909M','201626636K','201615906C','53339219E','201615963E','53358965K','201626635Z','201615966W','53339220M','53358966E','201707911W','201707912H','201707913Z','201615967H','53339221K','53358969W','201615969K','53334620K','53339224A','201615971R','53358970C','53358972W','201615972G','201615973W','201707920Z','201720961D','201720969K','201616392C','201720970E','201703427W','53355848D','201703429Z','53355850J','201625295C','201720975Z','201703432R','201616387D','201616388N','201720977C','53346090E','201625297D','201625299E','53339550W','201625302H','201625311K','201616393M','201720978M','201625312C','201703434W','201720979D','53339553D','53339554B','201616400Z','201616403M','201720983K','53364379A','53344401B','201716165K','201716185R','53344402X','201716186G','201622938K','53344404K','201622940R','201622942W','53344409L','201622945K','53344412L','201716187W','53364394E','201716188H','53344413J','53364395C','201716189Z','53344414D','53364396A','53364397W','201622947M','201932505K','201932529Z','201622949N','53403911A','53333127K','53344417M','53333132C','53333133A','53403931J','T16LL0604C','201608355W','53403932D','53344420M']
```
KenLSM: `const` might be a better variable type for this as it doesn't and isn't expected to mutate through its lifetime.

KenLSM: It might make sense to sort these numbers and reduce the amount of it. It's about a 100 each, and having duplicates between `acra`, `others`, `check_sum` which can be a little on the high side.

From the implementation, since we are doing whitebox testing, we could categorize the cases by local vs business vs others or any other ways that you find it more suitable that can help to reduce the amount of cases or by picking those numbers that are generally orthogonal with each other or something even better.

shared/utils/uen-validation.ts
```diff
@@ -97,72 +140,165 @@ const isNumeric = (s: string): boolean => !!s.match(/^[0-9]+$/)
...
-
-    // check that first 8 letters are all numbers
-    const first8Chars = uen.slice(0, 8)
-    if (!isNumeric(first8Chars)) {
-      return false
-    }
-
-    // (A) Businesses registered with ACRA (SUCCESS)
-    return true
+const calc_business_check_digit = (number: string): string =>{
```
KenLSM: Generally in this repo we follow the javascript convention of using `camelCase` over `snake_case` for variables names, excluding `const variables` and functions.

shared/utils/uen-validation.ts
```diff
@@ -97,72 +140,165 @@ const isNumeric = (s: string): boolean => !!s.match(/^[0-9]+$/)
...
-    return false
+/**
+ * Helper for Business UEN. 
+ * Performs validation based on Business (ROB) numbers
+ * @param number 
+ * @returns number if Business UEN
+ */
+const validate_business = (number: string): string  =>{
+  if (!isNumeric(number.slice(0,-1))){
+    return ""
```
KenLSM: Any specific reason why we are returning a `string` instead of `boolean`? It seemed like it would be reasonable if we return a `boolean` which is immediately implies that the answer is `false` instead of the answer is an empty string.

shared/utils/uen-validation.ts
```diff
@@ -97,72 +140,165 @@ const isNumeric = (s: string): boolean => !!s.match(/^[0-9]+$/)
...
+}
+
+/**
+ * Helper for other UEN. Performs validation based on UEN in
+ * VALID_ENTITY_TYPE_INDICATORS
+ * @param number 
+ * @returns number if other UEN
+ */
+const validate_other = (number: string): string =>{
+  let rst = ['R', 'S', 'T']
```
KenLSM: We might have lost the context here if we name this `rst` without explaining the why the characters `R, S, T` is special.

shared/utils/uen-validation.ts
```diff
@@ -97,72 +140,165 @@ const isNumeric = (s: string): boolean => !!s.match(/^[0-9]+$/)
...
+
+/**
+ * Helper for other UEN. Performs validation based on UEN in
+ * VALID_ENTITY_TYPE_INDICATORS
+ * @param number 
+ * @returns number if other UEN
+ */
+const validate_other = (number: string): string =>{
+  let rst = ['R', 'S', 'T'] 
+  if (rst.indexOf(number.slice(0,1)) === -1){
```
KenLSM: The original implementation of checking the first character for `RST` is more readable as it requires less understanding of what `number.slice(0, 1) ` is doing and is more idiomatic to use `Array.prototype.includes()`. 

Might want to refer to the original implementation.

shared/utils/uen-validation.ts
```diff
@@ -97,72 +140,165 @@ const isNumeric = (s: string): boolean => !!s.match(/^[0-9]+$/)
...
+    return ""
+  }
+  if (!isNumeric(number.slice(1,3))){
```
KenLSM: Same as above where we are prioritizing readability and maintainability over LOC here.

As this section can be very tricky to implement, and read (which I'm sure you felt it as well while working on this), we would lean on being explicit rather than implicit on what we are doing to ensure that the readers are able to follow without placing too much on their cognitive workload.

shared/utils/uen-validation.ts
```diff
@@ -97,72 +140,165 @@ const isNumeric = (s: string): boolean => !!s.match(/^[0-9]+$/)
...
+ * 
+ * Business UEN based on ROB 
+ * Local Comany UEN based on ROC
+ * Other UEN.
+ *  
+ * @param number string that represemts the UEN
+ * @returns number if UEN is valid
+ */
+const validate = (number: string): string => {
+  number = standardise(number)
```
KenLSM: We shouldn't be [re-assigning the arguments](https://eslint.org/docs/latest/rules/no-param-reassign) here. Would prefer to create new variable that captures what the value now holds.

KenLSM (commented):
Hey @hui-jie-lim, firstly thanks for following through with the commitment you have. We appreciate your help here.

I've ran through the first round combing of the changes. I've also triggered the CIs for this branch. But it seems like there are a couple of linting issues lying around that our CI is not capturing as well. Meanwhile, you might want to double check on your linting config.

package.json
hui-jie-lim: Ahh - got it. It makes sense now. Will remove the change accordingly

shared/utils/__tests__/uen-validation.spec.ts
hui-jie-lim: yeahh - agree. Was just categorising them based on the source but it does seem like there's quite a bit of overlap.

shared/utils/uen-validation.ts
hui-jie-lim: got it - will make the changes accordingly

shared/utils/uen-validation.ts
hui-jie-lim: yeahh - agree this implementation was awkward. Very much prefer to use includes() but it gave me some lib compatibility issue. Hmm, let me re-implement it with includes() and update the corresponding lib then

hui-jie-lim:
Cool! Let me work on these changes. Will get back. Thanks for the review btw. I am still quite new to software development so its very helpful to have another person view my code!

shared/utils/uen-validation.ts
hui-jie-lim: I was trying to stay as close to the original code but agree a simple boolean should work. Let me work on it

shared/package.json
```diff
@@ -17,6 +17,10 @@
     "@typescript-eslint/eslint-plugin": "^5.42.1",
     "@typescript-eslint/parser": "^5.42.1",
     "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-prettier": "^4.0.0"
+    "eslint-plugin-prettier": "^4.0.0",
+    "jest": "^29.6.1"
```
KenLSM: We might not need `jest` and the `test` command under `scripts`

shared/package-lock.json
KenLSM, line 1: We might not need this (this should be automatically removed once the `jest` dependeny on `shared/package.json` is removed

shared/tsconfig.json
```diff
@@ -1,12 +1,14 @@
 {
   "compilerOptions": {
     /* Visit https://aka.ms/tsconfig.json to read more about this file */
-
     /* Basic Options */
     // "incremental": true,                         /* Enable incremental compilation */
     "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', 'ES2021', or 'ESNEXT'. */,
     "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
-    // "lib": [],                                   /* Specify library files to be included in the compilation. */
+    "lib": [
```
KenLSM: Shouldn't need to change this file. Was there any errors that was addressed this change?

shared/utils/__tests__/uen-validation.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    const business = ['53333127K','53333132C','53333133A','53333286M','53334620K','53339179B','53339181D','53339185K','53339191A','53339219E','53339220M','53339221K','53339224A','53339550W','53339553D','53339554B','53344401B','53344402X','53344404K','53344409L','53344412L','53344413J','53344414D','53344417M','53344420M','53346090E','53355848D','53355850J','53358965K','53358966E','53358969W','53358970C','53358972W','53364379A','53364394E']
+    const others = ['S13SS0001G','S36CS0039B','S37SS0006B','S37SS0013H','S37SS0014D','S39CS0041K','S39SS0014C','S46CS0044H','S46TU0001K','S46TU0002F','S46TU0003B','S46TU0004J','S47SS0014G','S62SS0014K','T09SS0032E','T09SS0034H','T09SS0035D','T09SS0036L','T09SS0042K','T09SS0048H','T09SS0050A','T09SS0053L','T09SS0056K','T09SS0058B','T09SS0060F','T09SS0062J','T09SS0063E','T09SS0064A','T16LL0604C','T16UF2321H','T16UF4146J','T16UF7020E','T16UF7023D','T16UF7025G','T16UF7437A']
+    const variants = ['53333286M','T09SS0034H','53339179B','S36CS0039B','201707903G']
```
KenLSM: This might be better named as `checksumCases`.

shared/utils/uen-validation.ts
```diff
@@ -80,8 +113,18 @@ const VALID_ENTITY_TYPE_INDICATORS = new Set([
...
+  // Foreign Entities
+  'UF' 
 ])
 
+/**
+ * Helper to standardise format - uppercase and without whitespace  
+ * @param s String
+ * @returns True if string is numeric
+ */
+const standardise = (s: string): string => s.toUpperCase().trim()
```
KenLSM: This might be better named to literally represent what it does with `upperCaseAndTrim`

shared/utils/uen-validation.ts
KenLSM, line 1: For some of the comparison statements, I feel that we could be terser by using `!statement`

i.e., 
```diff
- if (isNumeric(first8Char) === false){
+ if (!isNumeric(first8Char)){
    return false
} 
```

KenLSM:
Could you also help me to apply [this](https://gist.github.com/KenLSM/43b0bd0689970ad05198518d4440b502) git patch that I've written to fix the issue of CI not checking for linting issues on `shared/`?

Here's the [command](https://www.gitkraken.com/learn/git/git-patch#git-apply-patch) to on applying git patches.

P.S. As I was reviewing, the code, I'm spotting some prettier issues

shared/utils/uen-validation.ts
hui-jie-lim: ah, my bad - misunderstood the comments and thought that there was a preference for explicit declarations for true and false. Its clear now. Will change to 

```
if (!isNumeric(first8Char)){
```

shared/utils/uen-validation.ts
hui-jie-lim: can!

shared/tsconfig.json
hui-jie-lim: let me check. had encountered some library issues which was why it was added but unclear if its my machine or a general config requiurement

shared/package.json
hui-jie-lim: oh yes, my bad

shared/tsconfig.json
hui-jie-lim: checked, and confirm we can run without updates to this file. Revert to c3a329b

shared/package-lock.json
hui-jie-lim: hey Ken, i reverted to the file that's in c4c0d94. All's good!

package.json
hui-jie-lim: Hey Ken, I just ran prettier on uen-validation.ts and its corresponding test file. There shouldn't be any linting issues now.

package-lock.json
hui-jie-lim, line 1: hey @KenLSM , doesn't seem I was able to revert this file. Please let me know if there is a way to do it on your end, thanks!

hui-jie-lim:
Hey @KenLSM , the code should be cleaned up already - reformatted with prettier and also removed some of the unnecessary changes. Should be all good to go. Please have a look, thanks!

hui-jie-lim:
Hey @KenLSM , i saw that the previous PR failed npm ci because the packages were not in sync. I had a look at the code and notice that it was because i still had jest in shared/package.json. Now that I removed it, could you re-trigger the test again please? Thank you

KenLSM:
Don't worry about the issue from `Chromatic / chromatic-deployment (pull_request)`

hui-jie-lim:
hey @KenLSM , was wondering if there's any update on this, and if there's any follow up required on my end.

KenLSM:
Hey @hui-jie-lim apologies for the late response last week. There seems to have some tests failures on (1) linting (due to the linting that we've added for `shared/` in this PR), and also (2) test cases where it now fails due to the strengthened UEN.

(2a) 
**[Frontend](https://github.com/opengovsg/FormSG/actions/runs/5589418599/job/15138465937?pr=6546) Tests**
Extract of some of the errors
```
 const success = screen.getByText('You have submitted: S01LP1234Z')
```
Previously, `S01LP1234Z` would have passed, but with the strengthened validator, this now fails. Updating the provided UEN in the test cases to something that's valid would suffice.

(2b)
**[Playwright](https://github.com/hui-jie-lim/FormSG/blob/develop/__tests__/e2e/constants/field.ts#L317) Tests**
I believe the e2e testing is also failing due to the strengthened UEN. A similar approach of updating to a valid UEN should suffice.
[Here's probably](https://github.com/hui-jie-lim/FormSG/blob/develop/__tests__/e2e/constants/field.ts#L317) where you should update the UEN for e2e testing
```
  {
    title: 'Some UEN here please',
    fieldType: BasicField.Uen,
    val: 'T09LL0001B', // <-- this
  },
```

hui-jie-lim:
Hey @KenLSM , I have updated the UEN for the tests and ran the linting scripts. Just a quick question. I notice that I am failing the playwright test for email submissions due to timeouts on my local machine(m1 mba). It shows time out on github too. Is this expected behaviour?

KenLSM:
Yes. Expected albeit not the best way to announce test failures. The timeouts are due to the way the tests were conducted. 

The test waits until a certain element appears before concluding the tests. This also means that if there is a validation error that prevents the element from showing, the playwright tests would wait until the it timed out and fail.

KenLSM:
Hey @hui-jie-lim the feature changes looks good, and all of the tests have passed. 🎉 

However, before we can merge, we'll still have to remove the changes on `package.json` (except for the two scripts that we've added) and `package-lock.json`.

package-lock.json
hui-jie-lim, line 1: Hi @KenLSM , should I revert to [this version](https://github.com/opengovsg/FormSG/blob/b641094d40a67a888e3a8ac967a43f9d1732bd49/package-lock.json) ?

hui-jie-lim:
Done! I have removed the changes on `package-lock.json` and `package.json`

KenLSM (approved):

KenLSM:
Merged! Thanks for your contribution @hui-jie-lim 🎉 🎉

hui-jie-lim:
Awesome! Feels great to be able to contribute :)
Thanks for your help on this @KenLSM

LinHuiqing:
Note: just put the issue number directly without the square brackets to trigger automations! Will close the issue manually this time 👍 
Thanks for your contribution @hui-jie-lim!

LinHuiqing:
Manual tests: 
- [ ] Create a email mode form with a UEN field
- [ ] Submit the form with a local company UEN
- [ ] Submit the form with a business UEN
- [ ] Submit the form with a UEN from the others list
- [ ] 3 emails should the received for the 3 successful form submissions above
- [ ] Key in a UEN with the wrong checksum. The field should throw a warning.
- [ ] Perform a submission through an API call with an invalid UEN. This should fail.

PR #6535, opened 2023-07-13
fix: updating csp to fix bypass

sebastianwzq:
## Problem
Remove googletagmanager and gstatic from CSP configurations to mitigate the following VAPT issues.
1. CSP bypass via AngularJS on Google-managed hosts
2. CSP bypass via open redirect in Google Tag Manager

Closes FRM-1119

## Solution

Background info
[Google Tag manager CSP](https://developers.google.com/tag-platform/tag-manager/csp)
[reCAPTCHA CSP](https://developers.google.com/recaptcha/docs/faq)

1. Remove domains hosting AngularJS library from script-src directive
2. Remove domains deploying the open redirect endpoint from the script-src directive
3. Alternatively, load specific files as opposed to entire domains.

**Breaking Changes** 

## Tests

- [ ] Check that scripts from reCAPTCHA can load successfully in form submission page (network tab). 
- [ ] Check that reCAPTCHA can trigger successfully. Ensure that feature flags for turnstile is set to false and use VPN to trigger reCAPTCHA.
- [ ] Check that there are no script failures related to GTM/GA when you navigate the home page and log in. No errors are expected as we are not using GA4 yet.
https://analytics.google.com/analytics/web/#/report-home/a130216930w207559009p200196046

### DIFF
--- a/src/app/loaders/express/__tests__/helmet.spec.ts
+++ b/src/app/loaders/express/__tests__/helmet.spec.ts
@@ -40,8 +40,8 @@ describe('helmetMiddlewares', () => {
       'https://challenges.cloudflare.com',
       'https://js.stripe.com/v3',
       // GA4 https://developers.google.com/tag-platform/tag-manager/web/csp
-      // Commenting out due to CSP bypass issues and we are not using GA4 yet.
-      // 'https://*.googletagmanager.com/',
+      // not actively used yet, loading specific file due to CSP bypass issue
+      'https://*.googletagmanager.com/gtag/',
     ],
     connectSrc: [
       "'self'",
--- a/src/app/loaders/express/__tests__/helmet.spec.ts
+++ b/src/app/loaders/express/__tests__/helmet.spec.ts
@@ -39,6 +39,7 @@ describe('helmetMiddlewares', () => {
       'https://www.gstatic.com/recaptcha/releases/',
       'https://challenges.cloudflare.com',
       'https://js.stripe.com/v3',
+      'https://*.googletagmanager.com/gtag/',
     ],
     connectSrc: [
       "'self'",
--- a/src/app/loaders/express/__tests__/helmet.spec.ts
+++ b/src/app/loaders/express/__tests__/helmet.spec.ts
@@ -31,7 +31,6 @@ describe('helmetMiddlewares', () => {
     fontSrc: ["'self'", 'data:', 'https://fonts.gstatic.com/'],
     scriptSrc: [
       "'self'",
-      'https://www.googletagmanager.com/',
       'https://ssl.google-analytics.com/',
       'https://www.google-analytics.com/',
       'https://www.tagmanager.google.com/',
@@ -40,7 +39,7 @@ describe('helmetMiddlewares', () => {
       'https://www.gstatic.com/recaptcha/releases/',
       'https://challenges.cloudflare.com',
       'https://js.stripe.com/v3',
-      'https://*.googletagmanager.com',
+      'https://*.googletagmanager.com/gtag/',
     ],
     connectSrc: [
       "'self'",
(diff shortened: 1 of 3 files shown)


src/app/loaders/express/__tests__/helmet.spec.ts
```diff
@@ -31,17 +31,17 @@ describe('helmetMiddlewares', () => {
...
     scriptSrc: [
       "'self'",
-      'https://www.googletagmanager.com/',
       'https://ssl.google-analytics.com/',
       'https://www.google-analytics.com/',
       'https://www.tagmanager.google.com/',
       'https://www.google.com/recaptcha/',
       'https://www.recaptcha.net/recaptcha/',
-      'https://www.gstatic.com/recaptcha/',
+      'https://www.gstatic.com/recaptcha/releases/',
```
LinHuiqing, line 39: Not sure if we can be even more granular and do something like this: 

```suggestion
      'https://www.gstatic.com/recaptcha/releases/*/recaptcha__en.js',
```

LinHuiqing: The part after `releases/` seems to be some kinda ID identifying the release, and it's the same between prod and staging, so maybe we could hard-code that? Might want to see where and when it is generated before doing so though.

LinHuiqing (approved):
Lgtm in general! The `imgSrc` configurations look pretty sus to me though, not sure if we should be fixing that as well?

LinHuiqing:
> The `imgSrc` configurations look pretty sus to me though, not sure if we should be fixing that as well?

I think it could be a vulnerability? Though not flagged by cure53, but here's an example of how it could be exploited (near the start of the article): https://github.blog/2017-01-19-githubs-post-csp-journey/

LinHuiqing:
Oh also side note, @sebastianwzq I think the automations don't work if you wrap the issue string with square brackets, so can remove them from PR description for the integrations to do their magic haha
Also, do remember to indicate that this should not cause breaking changes in the PR description as well!

sebastianwzq:
> Lgtm in general! The `imgSrc` configurations look pretty sus to me though, not sure if we should be fixing that as well?

Which configuration looks suspicious? Seems to be required from [Google Tag manager CSP](https://developers.google.com/tag-platform/tag-manager/csp) and
[reCAPTCHA CSP](https://developers.google.com/recaptcha/docs/faq)

sebastianwzq:
> > The `imgSrc` configurations look pretty sus to me though, not sure if we should be fixing that as well?
> 
> I think it could be a vulnerability? Though not flagged by cure53, but here's an example of how it could be exploited (near the start of the article): https://github.blog/2017-01-19-githubs-post-csp-journey/

Thanks for sharing that! This post assumes a content injection bug as a prerequisite. We can investigate how to clean up our img src csp on a separate ticket as I suspect that will require some code changes in how we use GA.

src/app/loaders/express/__tests__/helmet.spec.ts
ghost: I actually considered hardcoding that too but am unsure of how they will name their files in new releases, hence stopping at release

LinHuiqing:
> > Lgtm in general! The `imgSrc` configurations look pretty sus to me though, not sure if we should be fixing that as well?
> 
> Which configuration looks suspicious? Seems to be required from [Google Tag manager CSP](https://developers.google.com/tag-platform/tag-manager/csp) and [reCAPTCHA CSP](https://developers.google.com/recaptcha/docs/faq)

@sebastianwzq Oh, the `'*'` haha! So actually all the other `imgSrc` configs don't matter since there's a catch-all. As you said though, we can track that as a separate issue if needed.

src/app/loaders/express/__tests__/helmet.spec.ts
LinHuiqing: Okay! Let's keep to this then, just in case.

PR #6536, opened 2023-07-14
feat: add description to form issue modal

mayying:
## Problem
Some of the public users seem to be using the form issue message to submit non-issue matter.

Closes FRM-1164

## Solution
Update description of the modal to explicitly call out that the feature is used only for issue reporting.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:

## Tests
- [ ] Open a form as public user, and click on the question mark icon at the bottom right corner.
- [ ] Ensure that the form issue modal is describing what this issue is for. Specifically, it should say `Fill this in only if you are experiencing issues and are unable to submit this form. If you would like to provide feedback, you can do so after submitting the form.


LinHuiqing (approved):
lgtm and tested!

LinHuiqing:
Should be good to go, assigned chromatic review to @staceytan1998 just in case.

LinHuiqing (approved):
lgtm! merge when ready

PR #6543, opened 2023-07-18
chore: reduce max payment limit

KenLSM:
## Problem

Form admins can set the payment limit up to 1mn which is not supported by paynow (only up to 200k per transaction limit).

Closes FRM-1173

## Solution

Switching the limits of payment config to 200k down from 1mn

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-07-18 at 3 27 27 PM
Screenshot 2023-07-18 at 3 27 23 PM

**AFTER**:

Screenshot 2023-07-18 at 3 23 19 PM
Screenshot 2023-07-18 at 3 23 15 PM

## Tests

Regression Form payments settings
- [ ] form admins should be able to update their payment amount
- [ ] form admins with previously higher form payment values should be able to update their payment amount (change on db if you don't have any existing forms with high payments)

Feature test
- [ ] form admins cannot set any value above `200000` (try `222333`)


wanlingt (approved):
lgtm

PR #6545, opened 2023-07-18
build(deps): bump winston-cloudwatch to v6.2.0

LinHuiqing:
## Problem

Updating `winston-cloudwatch` was attempted in https://github.com/opengovsg/FormSG/pull/6542, but caused release to not be able to build on our staging env due to the missing package `ip`, which was only available as a dev dependency. 

`ip` was a dependency installed as part of `proxy-agent` previously used by `winston-cloudwatch`. With the `winston-cloudwatch` update to 6.22.0, it was then removed, causing the production app to fail to build as `ip` is used in our webhook validation code.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Add `ip` dependency.

## Tests

Regression tests
- [ ] Log in to FormSG.
- [ ] Create and submit a form with email and mobile OTP verification.
- [ ] Duplicate the form using the template function, using another mode.
- [ ] Submit the new form.
- [ ] Log out of FormSG.

## Deploy Notes

**New dependencies**:

- `ip`: used for webhook validation code to check for private IP addresses

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -130,7 +130,7 @@
         "web-streams-polyfill": "^3.2.1",
         "whatwg-fetch": "^3.6.2",
         "winston": "^3.8.2",
-        "winston-cloudwatch": "^6.2.0",
+        "winston-cloudwatch": "^6.1.1",
         "zod": "^3.21.4"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


wanlingt (approved):

wanlingt:
technically this is just about adding the `ip` dependency since the `bump winston-cloudwatch version to 6.2.0` commit was un-reverted right

LinHuiqing:
> technically this is just about adding the `ip` dependency since the `bump winston-cloudwatch version to 6.2.0` commit was un-reverted right

Oops, I enabled auto-merge so this merged before I saw this. But yes! Ah I see the change set is just adding ip :') I'll update the PR description since I can't edit the merged commit anymore.

PR #6551, opened 2023-07-19
ci: adjust timeouts for playwright pipe

LinHuiqing:
## Problem

When playwright tests, we need to wait for them to timeout before we can get logs and figure out why they fail.

## Solution

Playwright tests already use default timeouts, with the exception of the thank you page. As such the timeouts for the playwright tests themselves were not changed.

**Breaking Changes** 

**Improvements**:

- Adjust playwright pipeline timeouts based on empirical values (~max time + 5 min).

## Tests

- [ ] Playwright tests should succeed when tests pass.
- [ ] Playwright tests should timeout with the timeouts set, and show logs within 40 minutes.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -6,7 +6,7 @@ on:
     branches: [develop, release-al2, master]
 jobs:
   test:
-    timeout-minutes: 60
+    timeout-minutes: 40
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/playwright.yml
```diff
@@ -6,7 +6,7 @@ on:
     branches: [develop, release-al2, master]
 jobs:
   test:
-    timeout-minutes: 60
+    timeout-minutes: 40
```
foochifa, line 9: Just for clarity sake, this first time-out is for the entire test? 

Then the subsequent timeout is for each step right? But the total sum of the subsequent timeout is 30 vs the parent 40, then do we even need this line? 😅 Or are we factoring in the `build` time?

LinHuiqing: Yes, it includes build time. And I saw a successful run with 32min, that's why it's 40min here.

LinHuiqing: this is one is for the whole job

foochifa (approved):
Thanks for the clarification! LGTM

PR #6552, opened 2023-07-19
feat: experimental login integration with SGID

justynoh:
## Problem
This PR implements an experimental admin login integration with SGID.

## Solution

A single new frontend page is created at `/ogp-login` which handles the entire login flow. To accommodate this new flow, two new backend routes are added: `/api/v3/auth/sgid/authUrl` which takes no parameters and returns the SGID authentication url, and `/api/v3/auth/sgid/login?code=XXXX&state=YYYY` which takes two query parameters, `code` and `state` and performs the access token exchange and data retrieval with SGID.

For this to work, a new SGID client class is instantiated in `AuthSgidServiceClass`. The class instantiates a new `SGIDClient` at startup, and has three functions to (1) return the auth URL, (2) perform the access token exchange, and (3) retrieve the data.

> ### Aside: Why not use the existing `SgidServiceClass`?
> This was considered. If done this way, the implication is that the same return url would have to be used, that is, `/sgid/login`, and state parsing would have to be done at the service level. But "isn't that the whole point of the state - so you can tell what part of your app the user was in before they were redirected out", you ask. In theory, yes, we could add a new case where the state could just be the string `login` and we can test for this. However, I think the tradeoff in DX is quite big here. `SgidServiceClass` handles not just the OAuth interaction with SGID, but also a whole bunch of validation for public form NRICs and so on (e.g. signing the NRIC received as a JWT, validating that the NRIC has not been tampered with when the form is submitted) and I can only imagine it will get bigger with MyInfo. Even at the first step, getting the SGID auth url, we will need to have different cases because the requested scopes will be different (NRIC for public form, OGP work email for admin). So as not to confuse the two fairly distinct use cases, I think it's just easier to keep them as separate services and maintain them separately. 

On the frontend, the single page checks if there is an existing user in session. If so, it redirects to the dashboard. Otherwise it allows the user to remain on the SGID login page, and renders the login button to trigger the login flow.

**Breaking Changes** 

## Screenshots

image

https://github.com/opengovsg/FormSG/assets/25571626/f5ab58b9-6adb-4c48-9674-1a4b7aedab2b

## Tests

**New functionality**
- [ ] Log in as a user that is from OGP. You should be succesfully redirected to your FormSG dashboard.
- [ ] Log in as a user that is not from OGP. You should simply be directed to the SGID login page.

**Regression**
- [ ] Log in with the usual OTP flow.

## Deploy Notes

- `SGID_FORM_LOGIN_REDIRECT_URI` : supersedes `SGID_REDIRECT_URI`
- `SGID_ADMIN_LOGIN_REDIRECT_URI`: `<APP_URL>/api/v3/auth/sgid/login`


src/app/modules/auth/sgid/auth-sgid.service.ts
```diff
@@ -0,0 +1,122 @@
...
+    const logMeta = {
+      action: 'createRedirectUrl',
+    }
+
+    const result = this.client.authorizationUrl(
+      SGID_LOGIN_OAUTH_STATE,
+      ['openid', SGID_OGP_WORK_EMAIL_SCOPE].join(' '),
+      null,
+    )
+    if (typeof result.url === 'string') {
```
ghost: nit: redundant check

frontend/src/features/login/SgidLoginPage.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { BiLogInCircle } from 'react-icons/bi'
+import { useMutation } from 'react-query'
+import { useSearchParams } from 'react-router-dom'
+
+import { DASHBOARD_ROUTE } from '~constants/routes'
+import { useToast } from '~hooks/useToast'
+import { getSgidAuthUrl } from '~services/AuthService'
+import Button from '~components/Button'
+import { InlineMessage } from '~components/InlineMessage/InlineMessage'
+import { ToastStatus } from '~components/Toast/Toast'
```
ghost: nit: remove unused

sebastianwzq (approved):
Took a quick look and left minor comments.

frontend/src/features/login/SgidLoginPage.tsx
justynoh: thanks! fixed

src/app/modules/auth/sgid/auth-sgid.service.ts
justynoh: ahh true. Switched to try-catch instead. Thanks!

LinHuiqing:
@justynoh merge when ready then we can cut release!

PR #6556, opened 2023-07-20
fix: paynow payout being ignored

KenLSM:
## Problem

Several payouts are found to be not confirmed.

See [investigation details](https://www.notion.so/opengov/Missing-Payout-Investigation-00dd0a4e67c24e78bf08d23d8e12029e?pvs=4)

## Solution

Payouts that were not confirmed were all from `paynow` as payout source from stripe are reported with type `payment` instead of `charge`. 

Fix to this is a single LOC which is to also include `payment` type as a valid transaction source to process.

To fully capture all the payouts, we will need to:
- collect through all `succeeded` payments that have not successfully obtained a `payout`
- retrigger the webhook payout event

**Breaking Changes** 

## Tests

**Regression**
Payout webhook should not fail
- [ ] Refire existing payout webhook from stripe dashboard
  - [ ] Observe that event returns `200 OK` on stripe
- [ ] Datadog should not report errors from BE

### DIFF
--- a/src/app/modules/payments/stripe.service.ts
+++ b/src/app/modules/payments/stripe.service.ts
@@ -559,7 +559,7 @@ export const handleStripeEvent = (
             { stripeAccount: event.account },
           )
           .autoPagingEach(async (balanceTransaction) => {
-            if (balanceTransaction.type !== 'charge') return
+            if (!['charge', 'payment'].includes(balanceTransaction.type)) return
 
             const charge = balanceTransaction.source as Stripe.Charge
             await getMetadataPaymentId(charge.metadata)
--- a/src/app/modules/payments/__tests__/stripe.service.spec.ts
+++ b/src/app/modules/payments/__tests__/stripe.service.spec.ts
@@ -240,6 +240,40 @@ const MOCK_STRIPE_EVENTS = [
     },
     type: 'payout.canceled',
   },
+  {
+    id: 'evt_PAYOUT_PAID',
+    object: 'event',
+    created: 1688972601,
+    account: 'acct_MOCK_ACCOUNT_ID',
+    data: {
+      object: {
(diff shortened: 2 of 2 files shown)


KenLSM:
@justynoh included you for sanity check as you've done a great number of work on Stripe webhooks.

sebastianwzq (approved):

wanlingt (approved):
lgtm, some (non-blocking) questions:
1. ~what is LOC~ edit: ok got it now it's line of code
2. could you include a link to the docs for the webhook shapes?

KenLSM:
> 1. what is LOC

Line of code 

> 2. could you include a link to the docs for the webhook shapes?

Super tl;dr is
Paynow type
```json
{
      "source": {
        "payment_method_details": {
          "type": "paynow"
        },
      "type": "payment"
}
```

Card type
```json
{
      "source": {
        "payment_method_details": {
          "card": {
            "brand": "visa",
            "country": "SG",
          },
          "type": "card"
        },
      },
      "type": "charge"
}
```
[More details in here](https://www.notion.so/opengov/Missing-Payout-Investigation-00dd0a4e67c24e78bf08d23d8e12029e?pvs=4#6a57b670218942be84b969b74448d3cd) -- the most verbose detail is in a block above the link on notion

src/app/modules/payments/stripe.service.ts
```diff
@@ -559,7 +559,7 @@ export const handleStripeEvent = (
             { stripeAccount: event.account },
           )
           .autoPagingEach(async (balanceTransaction) => {
-            if (balanceTransaction.type !== 'charge') return
```
ghost, line 562: nit: sudden thought. Might be useful to log scenarios when the early return happens

PR #6566, opened 2023-07-25
chore: rm redundant type assertion in url validation

### LINKED ISSUE #1788: remove redundant type checks in /shared utilities once client is fully in Typescript
this is a placeholder issue to track such occurrence in the codebase.

for example, in `src/shared/util/url-validation.ts`, even if the argument is typed, it may be called by Javascript files in the client which do not care about such types.

LinHuiqing:
## Problem

Type assertions were required in angular days as files might be called by JS files which don't care about typing. This is no longer the case as we now use a codebase completely migrated to TS.

Closes #1788

## Solution

**Breaking Changes** 

**Improvements**:

- removes redundant type assertion for url validation
- chore: remove unused isValidUrl function

## Tests

Regression test for webhook:
- [ ] Create a storage mode form
- [ ] Add a webhook to the form
- [ ] Make a submission, it should successfully post a webhook as well.

### DIFF
--- a/frontend/src/hooks/useMdComponents.tsx
+++ b/frontend/src/hooks/useMdComponents.tsx
@@ -84,7 +84,7 @@ export const useMdComponents = ({
       a: ({ node, ...props }) => {
         const { href } = props
         const isExternal =
-          typeof href === 'string' && !href.startsWith(window.location.origin)
+          (href && !href.startsWith(window.location.origin)) || false
 
         return <Link {...props} isExternal={isExternal} {...linkStyles} />
       },
--- a/frontend/src/features/logic/utils/typeguards.ts
+++ b/frontend/src/features/logic/utils/typeguards.ts
@@ -45,6 +45,6 @@ export const isLogicableField = (args: {
 export const isValueStringArray = (
   value: FormCondition['value'],
 ): value is string[] => {
-  // use .some because of limitation of typecript in calling .every() on union of array types: https://github.com/microsoft/TypeScript/issues/44373
+  // use .some because of limitation of typescript in calling .every() on union of array types: https://github.com/microsoft/TypeScript/issues/44373
   return Array.isArray(value) && !value.some((v) => typeof v === 'number')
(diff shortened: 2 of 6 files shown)


frontend/src/hooks/useMdComponents.tsx
```diff
@@ -84,7 +84,7 @@ export const useMdComponents = ({
       a: ({ node, ...props }) => {
         const { href } = props
         const isExternal =
-          typeof href === 'string' && !href.startsWith(window.location.origin)
+          (href && !href.startsWith(window.location.origin)) || false
```
kenjin-work, line 87: Just to double check: this is type safe because `href` is defined as `string` in `props` right?

Just worried that `startsWith` could be `undefined`.

kenjin-work (commented):
Just one question, otherwise LGTM!

frontend/src/hooks/useMdComponents.tsx
LinHuiqing: Yes, based on what I'm seeing on the IDE. Though, not sure where it's pulling the type though, hmm

kenjin-work: Seems it is from react-markdown Components type.  All references to `href` in their source code claims its a string https://github.com/remarkjs/react-markdown/blob/main/lib/ast-to-react.js#L34

kenjin-work (approved):

frontend/src/hooks/useMdComponents.tsx
LinHuiqing: Btw to follow up, the typing is defined here in the `react-markdown` package:
Screenshot 2023-07-25 at 5 19 39 PM
So href is `string | undefined`! Since other components don't have the `href` prop.

LinHuiqing: Oh yea, what you said! (ps just saw your comment @kenjin-work)

PR #6567, opened 2023-07-25
chore: lint and update commands in README

kenjin-work:
## Problem
Backend lint fails on dev branch. Plus README has out of date commands for e2e tests dating back to Angular days.

## Solution
Lint and update README.

**Breaking Changes** 

### DIFF
--- a/shared/types/form/form.ts
+++ b/shared/types/form/form.ts
@@ -13,7 +13,6 @@ import {
 import { DateString } from '../generic'
 import { FormLogic, LogicDto } from './form_logic'
 import { PaymentChannel, PaymentType } from '../payment'
-import { MyInfoChildrenBirthRecords } from '@opengovsg/myinfo-gov-client'
 
 export type FormId = Opaque<string, 'FormId'>
 
@@ -180,7 +179,7 @@ export type PublicStorageFormDto = Merge<
     StorageFormDto,
     // Arrays like typeof list have numeric index signatures, so their number key
     // yields the union of all numerically-indexed properties.
-    (typeof STORAGE_PUBLIC_FORM_FIELDS)[number]
+    typeof STORAGE_PUBLIC_FORM_FIELDS[number]
   >,
   PublicFormBase
 >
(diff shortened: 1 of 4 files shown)


src/types/field/childrenCompoundField.ts
```diff
@@ -1,3 +1,4 @@
+// prettier-ignore
```
kenjin-work: The formatter and prettier seems to fight over these imports. So just tell prettier to ignore it.

shared/types/form/form.ts
```diff
@@ -180,7 +179,7 @@ export type PublicStorageFormDto = Merge<
     StorageFormDto,
     // Arrays like typeof list have numeric index signatures, so their number key
     // yields the union of all numerically-indexed properties.
-    (typeof STORAGE_PUBLIC_FORM_FIELDS)[number]
+    typeof STORAGE_PUBLIC_FORM_FIELDS[number]
```
kenjin-work, line 182: I can't really tell if this change is intended. From a strict sense, `(typeof xxx)[number]` -> `string` (to be precise, a character from the typeof repr). While `typeof xxx[number]` -> `xxx[indexed]` if `xxx` is indexable.

@hui-jie-lim requesting your confirmation please :). Thanks.

LinHuiqing: did you tag the wrong person LOL but okay I updated it cos `backend-lint` was complaining: https://github.com/opengovsg/FormSG/actions/runs/5652828911/job/15313165263

kenjin-work: I think that's the author of the PR that made the change for those `typeof` parts.

LinHuiqing: Ahh yes i see it now, my bad oops. How did it get past the backend-lint 🤔

hui-jie-lim: Hello, my apologies but these changes are not intended. I think it might have been due to changes made to `FormSG/package.json` where we added shared/ to the lint scripts. 

cc: @KenLSM 
[Ref PR](https://github.com/opengovsg/FormSG/pull/6534/files)

src/types/field/childrenCompoundField.ts
LinHuiqing: It's fine for me eh actually, I don't think this is needed?

shared/types/form/form.ts
KenLSM: My bad too. Somehow i missed this during the merge 🤔

src/types/field/childrenCompoundField.ts
kenjin-work: Oh yeah I think I saw wrongly. Removed.

LinHuiqing (approved):
lgtm, thanks @kenjin-work!

Issue #2039, opened 2021-06-01
[2?] Remove customMin and customMax virtuals from number and text field schemas


tshuli:
Remove customMin and customMax virtuals from number and text field schemas, which were put in to allow for backward compatibility after customMin and customMax were removed as part of #408

justynoh:
In particular, the following need to be removed:

- Lines setting the virtuals (i.e. starting with `ValidationOptionsSchema.virtual('customMin').get(function (`) in the files `src/app/models/field/xxxField.ts` (where xxx is the relevant field type - text/number etc)
- The type generic in `src/types/field/utils/virtuals.ts`

PR #6575, opened 2023-07-27
feat: add range validation to number field

### LINKED ISSUE #378: Add range validation to number field

LeonardYam:
## Problem
Closes #378

## Solution
**Breaking Changes** 

    - Schema change in `shared/types/field/numberField.ts`

**Changes by file**
- Updated the existing `NumberValidationOptions` schema in `shared/types/field/numberField.ts` to include range validation
- Added components, validation and updated schema for range validation in the form builder drawer located in `frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx`
- Updated field creation in `frontend/src/features/admin-form/create/builder-and-design/utils/fieldCreation.ts` to reflect the updated schema
- Updated frontend validation for the `NumberField` in `frontend/src/utils/fieldValidation.ts`
- Updated mongoose schema in `src/app/models/field/numberField.ts`
- Updated backend validators in `src/app/utils/field-validation/validators/numberValidator.ts`

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image
image
image

## Tests
Manual tests for range validation:
- [ ] E2E: Range validation works as expected.
  - [ ] Select `Range of values allowed` under `Field restriction`.
    - [ ] Submit the form without entering a minimum or maximum value. This should display an error.
    - [ ] Submit the form with input `0` for minimum or maximum value. This should display an error.
    - [ ] Submit the form with an invalid range where minimum >= maximum value. This should display an error.
    - [ ] Submit the form with only one of minimum and maximum value specified. This should work and create a number field as expected.
    - [ ] Submit the form with a valid range where minimum < maximum. This should work and create a number field as expected.

Regression tests (for `EditNumber.tsx`)
- [ ] E2E: Try to create a number field with no selected validation. This should work and create a number field as expected.
- [ ] E2E: Existing number validation still works as expected. 
    - [ ] Select `Number of characters allowed` under `Field restriction`.
      - [ ] Submit the form without choosing length validation. This should display an error.
    - [ ] Select Min/Max/Exact for length validation.
      - [ ] Submit the form without an input for `Number of characters`. This should display an error.
      - [ ] Submit the form with 0 as the input. This should display an error.
      - [ ] Submit the form with a number > 0 as the input. This should work and create a number field as expected.
    
## Concerns
- [ ] Check if the naming and structuring of the new variables and schema is acceptable
- [ ] Check if the EditNumber component requires refactoring into smaller components
- [ ] Check for errors in the typing 
- [ ] Check if there needs to be design changes to the front-end components (in particular, we currently allow number fields with `Character Length` or `Number Range` selected but no custom values to be created)
- [ ] Check if the error messages are acceptable
- [ ] Updating existing tests to use new schema

## Deploy notes
**New scripts:**
- `migrate-number-field-schema.js`: This script migrates `ValidationOptions` to `ValidationOptions.LengthValidationOptions` for all existing number form fields.


justynoh:
Hi @LeonardYam! Thanks for the contribution. I'll be working with you to get this PR merged in as soon as possible 😄 

While I review this PR, could I trouble you to open another PR to clean up the old virtuals in the `numberField.ts` file? If they are retained, they will expose incorrect properties after this PR is merged in (as they are set based on the assumption of character length validation, not number range validation). To keep our codebase clean, can I ask you to take up [this (small)](https://github.com/opengovsg/formsg/issues/2039) issue as well? I've added the concrete items to be done in the issue description. Once that is ready, feel free to tag me and I'll merge it in, and then you can rebase this PR onto develop. Meanwhile, I'll take a look at this PR.

Thanks again!

LeonardYam:
Hi @justynoh, I've just rebased the pull request. Let me know if there are any issues! 😄

shared/types/field/numberField.ts
```diff
@@ -1,14 +1,30 @@
 import { BasicField, MyInfoableFieldBase } from './base'
 
 export enum NumberSelectedValidation {
+  Length = 'Character Length',
+  Range = 'Number Range',
```
justynoh: Let's stick to just the simple terms "Length" and "Range" (which you've already used as keys here) and use the same strings for the values. 

The copy we want for the frontend dropdown option should not be the concern of the shared types. We can transform the enum to the appropriate copy in the frontend component itself.

src/app/models/field/numberField.ts
```diff
@@ -10,23 +11,44 @@ import { MyInfoSchema } from './baseField'
...
+          enum: [...Object.values(NumberSelectedLengthValidation), null],
+        },
+      },
+      RangeValidationOptions: {
+        rangeMinimum: {
+          type: Number,
+        },
+        rangeMaximum: {
+          type: Number,
+        },
```
justynoh: Can we use the same variable names that are used in other parts of the codebase for this type of validation? For instance, we have checkbox validation and there we use `customMin` and `customMax`.

src/app/utils/field-validation/validators/numberValidator.ts
```diff
@@ -59,29 +62,66 @@ const maxLengthValidator: NumberValidatorConstructor =
...
+ */
+const numberRangeValidator: NumberValidatorConstructor =
+  (numberField) => (response) => {
+    const val = Number(response.answer)
+    const { rangeMinimum, rangeMaximum } =
+      numberField.ValidationOptions.RangeValidationOptions
+    const isWithinMinimum =
+      typeof rangeMinimum === 'number' && rangeMinimum <= val
+    const isWithinMaximum =
+      typeof rangeMaximum === 'number' && val <= rangeMaximum
```
justynoh: Rather than testing for the type of `rangeMinimum` and `rangeMaximum`, can we test for truthiness instead? Number field inputs can only be nonnegative, so neither of the custom values should be allowed to be 0.

src/app/utils/field-validation/validators/numberValidator.ts
```diff
@@ -59,29 +62,66 @@ const maxLengthValidator: NumberValidatorConstructor =
...
+/**
+ * Returns the appropriate number validation function
+ * based on the number validation option selected.
+ */
+const getNumberValidator: NumberValidatorConstructor = (numberField) => {
+  switch (numberField.ValidationOptions.selectedValidation) {
+    case NumberSelectedValidation.Length:
+      return getNumberLengthValidator(numberField)
+    case NumberSelectedValidation.Range:
+      return numberRangeValidator(numberField)
```
justynoh: For consistency, can we rename `numberRangeValidator` to `getNumberRangeValidator`?

shared/types/field/numberField.ts
```diff
@@ -1,14 +1,30 @@
...
+
+export enum NumberSelectedLengthValidation {
   Max = 'Maximum',
   Min = 'Minimum',
   Exact = 'Exact',
 }
 
+export type NumberLengthValidationOptions = {
+  customVal: number | null
+  selectedLengthValidation: NumberSelectedLengthValidation | '' | null
```
justynoh: Are all three types possible? Is `NumberSelectedLengthValidation | null` sufficient?

justynoh (commented):
Thanks for this! I've briefly reviewed the backend, while our designer is on leave. I'll get to the frontend when our designer is back later this week, and let you know what changes need to be made.

justynoh:
Additionally, we would need your help to write a mongo script to perform a migration of existing form documents from the old to the new schema. You can look in the `/scripts` folder for examples of how our migrations are structured. In essence, each migration exists in its own folder prefixed by the date in yyyymmdd format. What the script for this migration might look like is:

- Retrieve all forms with number field validation. Retrieve the count.
- Update all retrieved forms to convert the number field validation from the old schema to the new schema.
- Ensure that the updated count is the same as the initial count.

shared/types/field/numberField.ts
LeonardYam: I believe this was a mistake, I have removed it!

LeonardYam:
Hi @justynoh, apologies for the delay! I have gone through the previous code review and applied the suggestions accordingly.

Could I double-check on what tests I should perform to verify that the script works as expected?

justynoh:
Hi @LeonardYam, just a quick update from my end. I'll take a look as soon as I can, but this period has been quite busy for me! I'd seek your understanding to give me some time to review it when I next can, which will likely come next week. Don't worry, we haven't forgotten about your contribution! And apologies for the delay 😅

LeonardYam:
Hi @justynoh, no worries!

I've just added the mongodb script, which ended up being very verbose. However to my understanding, there aren't many alternatives since `arrayFilters` currently do not work with aggregation pipelines. I've updated the original post with the details about the script!

frontend/src/utils/fieldValidation.ts
```diff
@@ -206,34 +207,63 @@ export const createMobileValidationRules: ValidationRuleFnEmailAndMobile<
...
               currLen <= customVal ||
               simplur`Please enter at most ${customVal} digit[|s] (${currLen}/${customVal})`
             )
         }
       },
+      validNumberRange: (val: string) => {
+        if (selectedValidation !== NumberSelectedValidation.Range || !val)
+          return true
+
+        const numVal = Number(val)
```
justynoh: Since we do a cast here, should we perform a NaN check after the cast to ensure the number is valid? The behavior would probably be to return the following message: "Please enter a valid number"

frontend/src/utils/fieldValidation.ts
```diff
@@ -206,34 +207,63 @@ export const createMobileValidationRules: ValidationRuleFnEmailAndMobile<
...
+        const hasMinimum = !!customMin
+        const hasMaximum = !!customMax
```
justynoh: Are we certain that this works? We allow `customMin` to be 0, so `!!customMin` would evaluate to false, even though an explicit minimum has been prescribed. We should do an explicit `null` check rather than a boolean cast.

frontend/src/utils/fieldValidation.ts
```diff
@@ -206,34 +207,63 @@ export const createMobileValidationRules: ValidationRuleFnEmailAndMobile<
...
+          (hasMaximum && customMax < numVal)
+
+        if (!isOutsideRange) return true
+        else if (hasMinimum && hasMaximum) {
+          return `Please enter a number within the range ${customMin} to ${customMax}`
+        } else if (hasMinimum) {
+          return `Please enter a number that has a minimum value of ${customMin}`
+        } else if (hasMaximum) {
+          return `Please enter a number that has a maximum value of ${customMax}`
+        }
```
justynoh: The `else if` chain here isn't necessary since we return at every step. Could we make this less verbose and just use `if () return ...` everywhere?

Separately, for your consideration: the disadvantage of writing the code this way with `if`s is that it doesn't ensure that you have covered all cases, since you aren't statically guaranteed to return anything at all (if `hasMinimum` and `hasMaximum` are both false). If you want, my suggestion would be to write it like this:

```typescript
switch (hasMinimum) {
  case true:
    switch (hasMaximum) {
      case true: return ...
      case false: return ...
    }
  case false:
    switch (hasMaximum) {
    ...
    }
}
```
This is more verbose but gives a better static guarantee. But again, either is totally fine, it's up to you.

frontend/src/utils/fieldValidation.ts
```diff
@@ -206,34 +207,63 @@ export const createMobileValidationRules: ValidationRuleFnEmailAndMobile<
...
+          (hasMinimum && customMin > numVal) ||
+          (hasMaximum && customMax < numVal)
+
+        if (!isOutsideRange) return true
+        else if (hasMinimum && hasMaximum) {
+          return `Please enter a number within the range ${customMin} to ${customMax}`
+        } else if (hasMinimum) {
+          return `Please enter a number that has a minimum value of ${customMin}`
+        } else if (hasMaximum) {
+          return `Please enter a number that has a maximum value of ${customMax}`
```
justynoh: Copy updates:
"Please enter a number between ${customMin} and ${customMax}"
"Please enter a number that is at least ${customMin}"
"Please enter a number that is at most ${customMax}"

src/app/utils/field-validation/validators/numberValidator.ts
```diff
@@ -59,29 +62,64 @@ const maxLengthValidator: NumberValidatorConstructor =
...
+ * Returns a validation function to check if number is
+ * within the number range specified.
+ */
+const getNumberRangeValidator: NumberValidatorConstructor =
+  (numberField) => (response) => {
+    const val = Number(response.answer)
+    const { customMin, customMax } =
+      numberField.ValidationOptions.RangeValidationOptions
+    const isWithinMinimum = !!customMin && customMin <= val
+    const isWithinMaximum = !!customMax && val <= customMax
```
justynoh: Same comments here as above in the frontend field validator function, re: NaN check, Boolean cast.

src/app/utils/field-validation/validators/numberValidator.ts
```diff
@@ -59,29 +62,64 @@ const maxLengthValidator: NumberValidatorConstructor =
...
+/**
+ * Returns the appropriate number validation function
+ * based on the number validation option selected.
+ */
+const getNumberValidator: NumberValidatorConstructor = (numberField) => {
+  switch (numberField.ValidationOptions.selectedValidation) {
+    case NumberSelectedValidation.Length:
+      return getNumberLengthValidator(numberField)
+    case NumberSelectedValidation.Range:
+      return getNumberRangeValidator(numberField)
```
justynoh: Honestly, I would not be opposed to saving ourselves the verbosity of adding an extra layer of calling the `getNumberLengthValidator` and just embed that switch into this switch, then get rid of `getNumberLengthValidator` completely:

```suggestion
    case NumberSelectedValidation.Length:
      switch (numberField.ValidationOptions.LengthValidationOptions.selectedLengthValidation) {
        case NumberSelectedLengthValidation.Min:
            return minLengthValidator(numberField)
        case NumberSelectedLengthValidation.Max:
            return maxLengthValidator(numberField)
        case NumberSelectedLengthValidation.Exact:
            return exactLengthValidator(numberField)
        default:
             return right
      }
    case NumberSelectedValidation.Range:
      return rangeValidator(numberField)
```

Up to you though!

src/app/utils/field-validation/validators/numberValidator.ts
```diff
@@ -59,29 +62,64 @@ const maxLengthValidator: NumberValidatorConstructor =
...
     default:
       return right
   }
 }
 
+/**
+ * Returns a validation function to check if number is
+ * within the number range specified.
+ */
+const getNumberRangeValidator: NumberValidatorConstructor =
```
justynoh: Since this actually does the base logic of performing the validation, we should follow the same function naming convention as the other base validators a la `exactLengthValidator`. Can we change the name of this function to `rangeValidator`?

justynoh (commented):
This PR is starting to shape up nicely! A number of comments from my end, as well as some updated copy. The migration script also looks good, thanks for writing it up!

The only thing I haven't reviewed is the updated `EditNumber.tsx` frontend component. The reason for this is that we have updated screens for your reference. If you could help implement them, then on my next review I'll look at 1) the new number field creation UI and `EditNumber.tsx` and 2) the updates based on my comments. 

The screens and some other considerations are coming up in my next comment, for your reference.

justynoh:
**New UI screens**

Initial state
image

Dropdown options
image

Length validation (note that the dropdown for "Minimum/Maximum/Exact" is no longer clearable)
image

Range validation
image

**Validating the range itself**

I realized that we currently allow some strange combinations of values for the range. Some considerations for what range values we should allow during field creation (I've also given the appropriate error messages to be displayed for each error case, which should be used exactly as written):

1. We should allow the following possibilities in terms of **existence**: a) Both min and max specified, b) Only min specified, c) Only max specified. We should disallow field creation if both min and max are not specified (Error message: "Please enter range values") - in your current implementation, the field is saveable.
2. We should have the following restrictions in terms of **value**: a) If both min and max are specified, it must be that min is strictly less than max (Error message: "Minimum must be less than maximum"), b) since number fields can only contain non-negative numbers, in all cases, both the min and max values should not be 0 (Error messages: "Minimum cannot be 0", "Maximum cannot be 0").

Thanks again for contributing on this!

src/app/utils/field-validation/validators/numberValidator.ts
LeonardYam: In my opinion, the extra verbosity in this case makes the function a lot easier to understand, so I would prefer if we kept it!

frontend/src/utils/fieldValidation.ts
LeonardYam: If I'm not wrong, this should work since `numVal` is also a non-negative number and there would be no combination of values where this would cause an error. Also, after the front-end changes, we would no longer allow a value of 0 for `customMin`.

Of course, we could also go with your suggestion and be more explicit with the checks, let me know what you prefer!

frontend/src/utils/fieldValidation.ts
LeonardYam: This may not be necessary, as the frontend already ensures that `val` would be a numeric input with `val.replace(/\D/g, ''))`. In comparison, a NaN check is necessary in `createDecimalValidationRules` since `val` may not be a valid numeric input.

Of course, we could also do the check for defensive programming, let me know what you prefer!

src/app/utils/field-validation/validators/numberValidator.ts
LeonardYam: Related to above comments

frontend/src/utils/fieldValidation.ts
LeonardYam: Good point about static guarantees! I tried to change this to a switch statement but unfortunately, I think this pattern is not recognised by Typescript as a narrowing operation at the moment ([playground](https://www.typescriptlang.org/play?ts=5.2.2#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBVDtHHwDAAXlAAG9FCUSpx2IgAPJkAByUWg+X6bOk31AAB8BdBoIoAL7BBCcQQ0ZLU6SOUD8pW0hlM5msjnc3n8gCsvGlwUUcoVDIFep5EtAUtAdI1KrI5salqEPMQjutnNt0FAAEIaXTWRKANzBFQSeBkeLCBjCCGQWAKbjFBGgAAUnrZAEo0fL4O4AtB4Mws40UAAeUAARgADHmzYJ3qgNNnc4gC4zmaNBJYyOxIIU+32VAR4II5NJoHhxQHvmxXWPmRbi5BS+Ws7q-bzQLXG83VyUJh5pjKo2AcM0PGQvIhY85gewaF83GZmOwGMxIPxYNA6iJAwwJ6CIXgAEQAKKrAwiALPoEFJNiEHgA4qAAOTCFykASBBQQqOyXxkO8A78L+RyoI4abjPOTiPjowxYPoKAIIg0IJjwzGSLG0QiAYHDYgsTCsBwPBBCUQA)).

Also, wouldn't the case where both `hasMinimum` and `hasMaximum` are false be covered by the first if-statement since `!isOutsideRange` would be `true`?

LeonardYam:
Hi @justynoh, I've just went through the code review and applied the initial round of fixes.

Here's a quick look at the current UI screens: (updated copy of error messages, validation of empty and 0 values etc.)

https://github.com/opengovsg/FormSG/assets/11293086/6883e763-92e5-40ce-8db0-f44a9eb5a315

Other than the comments above, I also have some questions about the new UI screens.

I think the inclusion of a `None` option for `Field Restriction` might be slightly confusing, as clearing the selection would leave an empty input, and we would allow both an empty input and `None` to represent that no validation is required. It seems that we should either:
1. Remove `None` and let an empty input represent no validation
2. Keep `None` as an option but make the selection no longer clearable 

I'm leaning towards the first option since this would be more consistent with the current UI, as this pattern is already used in date validation for example.

On a related note, should we also keep the dropdown for length validation clearable? This way, we would be consistent with the character length validation in short answer and long answer validation.

Let me know what would be best!

frontend/src/utils/fieldValidation.ts
justynoh: Yes, we should not rely on guarantees from other upstream functions.

We want to write modular, extensible and maintainable code. It's hard to know when this function might eventually be reused later on. Relying purely on the input types and writing a function handling all cases is preferable to putting a non-static precondition on the function which relies on the user to verify that the precondition is met. Especially in this case where the unexpected case handling is basically only one line. (There might be a tradeoff if the handling is extremely complex - but it isn't in this case.)

Also, yes, I agree there should be one in `createDecimalValidationRules`... Feel free to open a ticket or PR for this if you'd like! But let's not squeeze it into this PR since it's out of scope.

frontend/src/utils/fieldValidation.ts
justynoh: Honestly this one is a bit of a supernit. But code eliminating multiple distinct cases simultaneously (e.g. in this case the `!!` eliminating null and 0) has bitten us before so I'm extra wary of it nowadays 😅

Generally, so that our code remains correct and readable, I would prefer if we base our implementations on the intent of the type representation. In particular, we know that for customMin and customMax, number represents the presence of a constraint and null represents the absence. Although we are not allowing 0 values for these anymore, maybe one day we might change our definitions and say we allow all integers to be entered into the number field. In that case, we would have to come back and change this function. 

The question is then, why should we be coding for the future when the feature does not exist yet? It's a reasonable question, and also why I think it's more of a supernit. There have been cases where I just add a comment to the effect of `Safe because customMin, customMax > 0` and just leave it, too. I think either is fine, but we should definitely document it somewhere

frontend/src/utils/fieldValidation.ts
justynoh: Yeah, tbh this one is pretty straightforward so I think the if-else chain is fine.

As an alternative suggestion, for your consideration, is to switch on the `customMin` and `customMax` directly:

```typescript
switch (customMin) {
  case null:
    switch (customMax) {
      case null:
        // No validation
      default:
        // Only customMax exists
    }
  default:
    switch (customMax) {
      case null:
        // Only customMin exists
      default:
        // Both customMin and customMax exist
    }
}
```

Again, the way you've written it is fine though, just thinking of other ways to write this.

justynoh:
Hi Leonard! Thanks for looking at the UI and doing up the new frontend. 

For your points, I agree with doing (1). We can remove the "None" option and use an empty input to represent no validation instead.

As for the dropdown for length validation, it should not be clearable. Since we already only display it when the user selects "Length validation", the user has made a conscious choice to select a validation type. Once they pick one type of length validation (Min/Max/Exact), it doesn't make sense to allow them to remove it at that level. If they want to clear the validation, they should press "X" on the validation type dropdown, not the length validation dropdown. 

Finally, as a UI nit: we should remove the form labels for "Minimum and/or maximum value" and "Number of characters allowed", above the conditionally rendered inputs. Instead, the spacing between the upper dropdown and the lower inputs should be 0.5rem.

Thanks!

frontend/src/utils/fieldValidation.ts
LeonardYam: I see. I think I would go with an explicit null check as it is more robust! (perhaps the number field would allow negative and zero values in the future?)

frontend/src/utils/fieldValidation.ts
LeonardYam: To me, the if-else chain is the most readable so I would stick to it. I've also changed `!isOutsideRange` to `isInRange` which has more clarity.

src/app/utils/field-validation/validators/numberValidator.ts
LeonardYam: I have made the changes regarding the Boolean cast!

Regarding the NaN check, I just realized that we already validate that `response.answer` is not `''` and a non-negative integer from the upstream validators. Since the composition of these validators are explicit, would the NaN check be necessary? I think we should make use of the assertions about `response.answer` from the previous validators, and add in a comment such as `// Chained validators ensure that the casting is valid`

Of course, the checks are not complex/expensive so we could always add them in. Let me know what are your thoughts on this!

LeonardYam:
Hi @justynoh, I've just went through the review again and there is only one last part I'm unsure about. Other than that, I've just realized that we actually use the `validatorjs` library in some files. Would it be better to replace the current range checks with `validator.isInt`?

Thanks for looking through the lengthy comments 😅 !

src/app/utils/field-validation/validators/numberValidator.ts
```diff
@@ -59,29 +62,64 @@ const maxLengthValidator: NumberValidatorConstructor =
...
+ * Returns a validation function to check if number is
+ * within the number range specified.
+ */
+const rangeValidator: NumberValidatorConstructor =
+  (numberField) => (response) => {
+    const val = Number(response.answer)
+    const { customMin, customMax } =
+      numberField.ValidationOptions.RangeValidationOptions
+    const isWithinMinimum = customMin !== null && customMin <= val
+    const isWithinMaximum = customMax !== null && val <= customMax
```
justynoh: Looking at this again, is the logic here correct? If say `customMin = null`, `isWithinMinimum` would short circuit to false. But there is no minimum value, so it should actually be true. Should the logic here be `customMxx === null || ...` instead?

frontend/src/utils/fieldValidation.ts
```diff
@@ -206,34 +207,68 @@ export const createMobileValidationRules: ValidationRuleFnEmailAndMobile<
...
             )
         }
       },
+      validNumberRange: (val: string) => {
+        if (selectedValidation !== NumberSelectedValidation.Range || !val)
+          return true
+
+        const numVal = parseInt(val)
+        if (Number.isNaN(numVal)) {
+          return 'Please enter a valid number!'
```
justynoh: quick copy change here: can we remove the "!"? Thanks!

frontend/src/utils/fieldValidation.ts
```diff
@@ -206,34 +207,68 @@ export const createMobileValidationRules: ValidationRuleFnEmailAndMobile<
...
+        }
+
+        const hasMinimum = customMin !== null
+        const hasMaximum = customMax !== null
+        const isInRange =
+          !(hasMinimum && numVal < customMin) &&
+          !(hasMaximum && numVal > customMax)
```
justynoh: The logic here seems correct but I think we can rewrite it in a more readable way, by de-De Morgan-ifying it. Something like this:

```typescript
const satisfiesCustomMin = !hasMinimum || customMin <= numVal
const satisfiesCustomMax = !hasMaximum || numVal <= customMax
const isInRange = satisfiesCustomMin && satisfiesCustomMax
```

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
```diff
@@ -25,44 +29,112 @@ type EditNumberProps = EditFieldProps<NumberFieldBase>
...
+    selectedValidation === NumberSelectedValidation.Length
+      ? NumberSelectedValidationInputs.Length
+      : selectedValidation === NumberSelectedValidation.Range
+      ? NumberSelectedValidationInputs.Range
+      : ('' as const)
+
+  const nextLengthValidationOptions = {
+    selectedLengthValidation:
+      LengthValidationOptions.selectedLengthValidation ||
+      NumberSelectedLengthValidation.Min,
```
justynoh: Can we default to an empty string here initially? We should then have validation on the dropdown to ensure it is filled upon save.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
```diff
@@ -98,24 +170,19 @@ export const EditNumber = ({ field }: EditNumberProps): JSX.Element => {
     'ValidationOptions.selectedValidation',
   )
 
-  const customValValidationOptions: RegisterOptions<
+  const watchedSelectedLengthValidation = watch(
+    'ValidationOptions.LengthValidationOptions.selectedLengthValidation',
+  )
+
+  const LengthCustomValValidationOptions: RegisterOptions<
```
justynoh: nit: lowercase `l` for camelCase

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
```diff
@@ -126,14 +193,67 @@ export const EditNumber = ({ field }: EditNumberProps): JSX.Element => {
...
+    () => ({
+      min: {
+        value: 1,
+        message: 'Maximum cannot be 0',
+      },
+    }),
+    [],
+  )
+
   // Effect to clear validation option errors when selection limit is toggled off.
```
justynoh: Should we update this comment to reflect that it is when the selected validation is cleared?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
```diff
@@ -126,14 +193,67 @@ export const EditNumber = ({ field }: EditNumberProps): JSX.Element => {
...
+        },
+        hasValidRange: (customMin) => {
+          const customMax = getValues(
+            'ValidationOptions.RangeValidationOptions.customMax',
+          )
+
+          return (
+            customMax === '' ||
+            customMin === '' ||
+            customMin <= customMax ||
```
justynoh: This validation should be `customMin < customMax`, a strict inequality.

src/app/utils/field-validation/validators/numberValidator.ts
justynoh: I think that's a fair argument. I'm ok with either way. Thanks for addressing the issue!

justynoh (commented):
Just a couple more nits, and we should be good to go!

As for using the validatorjs library, I think it's up to you. What you've implemented is fine, and I suspect there will be some overhead for handling `null` cases when using the library anyway, so it might not really simplify the code that much. I would prefer to just keep it as you've implemented it.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
LeonardYam: I've updated it, let me know if the comment is clear enough!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
LeonardYam: Since we intend to default to empty strings, I'm assuming we would also want to clear the input when selected validation is cleared. Let me know if there needs to be changes!

src/app/utils/field-validation/validators/numberValidator.ts
LeonardYam: Thanks for the catch!

LeonardYam:
Hi @justynoh, I've just made the relevant changes. Let me know if there are any final changes needed! 

On a side note, while I was checking the validity of the back-end validators, I actually had an issue with getting them to run as my `IncomingSubmission.responses` was always empty on form submit. Most likely, there is a misunderstanding or setup issue on my end but I wanted to check if you could reproduce this issue.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
justynoh: I think that may not be necessary. If a user 1) selects that validation, 2) enters their customization, 3) then they clear the selected validation, but 4) re-select it later, we should default to their original customization at step 2. So I don't think we need to add instrumentation to clear the input.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
justynoh: seems good enough to me. Thanks!

justynoh:
Hey Leonard,

Overall, the code on this PR looks pretty much good to go! Feel free to address the one comment I gave on your question above.

To answer your question regarding `incomingSubmission.responses` being empty - since you have encountered this, I would hazard a guess that you're testing on a storage mode form. Since storage mode responses are end-to-end encrypted, we cannot in theory (and therefore, do not in practice), perform backend validation on the form responses or encrypted response payload. This is an issue we are actively addressing but requires some work on our end on the product side before this can be done. In storage mode, the `incomingSubmission.responses` are used only to handle certain specific types of responses (in particular, email addresses entered in email fields with the "Send email confirmation" toggled on, so that our backend can send the email confirmations to respondents). Therefore, in order to test your backend validators, try using an email mode form instead.

Now, the last piece of this PR before I can approve it would be writing appropriate manual tests for the feature as well as updating the existing test suite to account for the changes. 

Concretely, you will definitely have to update the tests in `src/app/utils/field-validation/validators/__tests__/number-validation.spec.ts`. I don't see any other files off the top of my head, but you can run `npm run test:backend` to see if there are any relevant backend tests that need to be updated. 

Additionally, I would request your help to write manual tests. Generally, we try and write two types of tests - one set to cover new functionality, and another set for regression. Here is a sample of what we might expect to see in the "Tests" section of the PR:

```
## Tests

New functionality
- [ ] E2E: Range validation works correctly when both minimum and maximum is set.
    - [ ] Create a new form
    - [ ] Create a number field, and for "Field restriction" select "Range validation"
    - [ ] Click "Create" with no value entered for minimum and maximum. An error should be triggered saying a range must be entered.
    - [ ] Enter a value for minimum and maximum, where the minimum is >= maximum. An error should be triggered saying a valid range must be entered.
    - [ ] Enter a value for minimum and maximum, where the minimum is < maximum. Click "Create". The field should be successfully created.
    - [ ] Go to the public form. Enter a value outside the range, and click "Submit form". This should trigger an error on the form field.
    - [ ] Enter a value within the range instead, and click "Submit form". This should be be successfully submitted.
- [ ] BE: Range validator correctly blocks submissions where the submitted value does not pass range validation.
    - [ ] Create an email mode form, with number field with range validation.
    - [ ] On the command line, run the following cURL command: `<curl command with incorrect submission>`
    - [ ] The submission should be rejected with 400, and the error message "<error message here>".
... (add more if desired) ...

Regression tests
- [ ] E2E: Number length validation still works as expected. Repeat for Min/Max/Exact
    - [ ] Create a number field with length validation. 
    - [ ] Go to the form. Enter a number not satisfying the validation selected.
    - [ ] Click "Submit form". An error should be triggered on the number field.
... (add more if desired) ...
```

Feel free to let me know once you've updated the tests, and I'll be happy to add on if required, and then approve and merge the PR. Thanks!

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
LeonardYam: Could I confirm the desired behaviour for each input when clearing selected validation? We currently have the following inputs:
1. Selected length
2. Length `customVal`
3. `rangeMin`
4. `rangeMax`

Originally, `customVal` was cleared along with the selected validation type, and I was wondering if we would like to keep this behaviour for all the inputs.

LeonardYam:
Hi @justynoh, you were right - I was indeed using storage mode responses 😅, after changing the mode the backend validators managed to fire.

I just did a quick update on the backend tests, mainly renaming the test names to include length and updating the schemas.

During the update, I came across the test `it('should allow number with maximum length left undefined')`, and noticed that with the current implementation of our validators, we actually allow selected length validations with no customVal, even though this behaviour is not allowed in the frontend. Is this intentional?

Unfortunately, I've been very busy recently and I would need some time to finish up the PR 😅. Just to confirm, the remaining work would be to:
1. Add unit tests for range validation
2. Write the manual tests 
3. Update the existing tests (searching for `BasicField.Number` shows me some frontend and e2e tests that needs updating).
4. Confirm and implement the expected behaviour for clearing frontend inputs

Am I missing out on any additional tasks?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditNumber/EditNumber.tsx
justynoh: Hmm... Ok, thinking about it again, i think this makes sense. Let's retain the behavior of clearing all the 4 inputs you pointed out, when the selected validation is cleared. Thanks!

justynoh:
For your question regarding validator implementation - I don't think this is intentional. We should implement our validators for this feature in a way that ensures that this is not possible (i.e. if Length, it must have both `LengthSelectedValidation` and `customVal`, and if Range, it must have at least one of `customMin` and `customMax`). If there are other existing validators that are incorrect, please let us know and we can fix them separately from this PR!

As for the remaining tasks, yes, those seem like the remaining items to me. Thanks for the continuous work on this @LeonardYam! We're excited to get this merged.

LeonardYam:
Hi @justynoh, I took a look at the validators again and I believe that for our backend validators `number-validation.spec.ts` and `numberValidator.ts`, they are only responsible for validating responses and operate on an implicit assumption that the validation options are valid. 

Hence, there should not be any changes needed for the current validators. I've instead deleted the tests that are based on response validation on invalid options and re-written them as mentioned below.

For robustness, I've written automated tests for:
1. E2E tests for creation and submission, with valid options and values
2. BE tests for valid/invalid field creation and response submission (separately)
3. FE tests for valid and invalid response submissions

I believe the tests which cannot be automated currently involve testing of error messages on `EditNumber.tsx`, which I've instead included as manual tests above.

Additionally, I've also added placeholder text to selectedLengthValidation.

Let me know if there's anything more to add!

justynoh (approved):
lgtm! this is great work, thanks so much for the contribution @LeonardYam! We'll merge it in as soon as we can.

justynoh:
Closing and re-opening to get tests to run.

justynoh:
merged develop into this branch to include CI updates!

PR #6584, opened 2023-08-01
feat: upgrade sgid sdk

kschiew:
## Problem

sgID did a [major release](https://docs.id.gov.sg/important-updates/user-migrations/typescript-sdk-v2.0-major-release) for its TypeScript SDK recently and will deprecate support for v1 SDK (which is currently used by FormSG) by the end of the year.

## Solution
- Bumped major version of `sgid-client` to v2.
- Bumped major version of Mockpass to `v4.x` which is compatible to v2 of `sgid-client`.
- Code changes for Authenticated form via sgID flow.
- Code changes for Authenticated form with myInfo fields via sgID flow.
- Code changes for OGP officer-only login via sgID flow.
 
**Breaking Changes** 

    - The code changes for the sgID flow is not backwards compatible, and should only work on snapshots where `sgid-client` version is `>= 2.0.0`. 

**Improvements**:

- Introduced Proof Key for Code Exchange (PKCE) for sgID integration.

## Tests

- Created unit tests for test paths where `code_verifier` cookie is missing.
- Updated existing test cases based on the breaking changes.
- Manually tested against sgID's staging environment:
- [ ] Log in to FormSG via the `/ogp-login` route. This should work correctly.
- [ ] Create and submit a form with SGID authentication. This should work correctly.
- [ ] Create and submit a form with SGID MyInfo. This should work correctly.

## Deploy Notes

- `SGID_HOSTNAME` in `docker-compose.yml` : updated accordingly to Mockpass' changes. **No changes required for staging/prod env vars.**

**New dependencies**:

- `@opengovsg/sgid-client` : Bumped from `^1.0.4` to `v2.0.0`.
- `mockpass`: Bumped from `v3.1.3` to `v4.0.4`.


src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -274,18 +276,20 @@ export const handleGetPublicForm: ControllerHandler<
           return res.json({ form: publicForm, isIntranetUser })
         })
     case FormAuthType.SGID_MyInfo: {
-      const accessTokenCookie = req.cookies[SGID_MYINFO_COOKIE_NAME]
-      if (!accessTokenCookie) {
+      const { jwt: accessToken = '', sub = '' } = JSON.parse(
+        req.cookies[SGID_MYINFO_COOKIE_NAME] ?? '{}',
+      )
```
justynoh: Could this potentially be dangerous? If someone passes a request with a malformed cookie that is not a JSON value, `JSON.parse` would fail and terminate the process. Maybe we should extract the `JSON.parse` step and wrap it in a try-catch block.

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -456,19 +460,33 @@ export const _handleFormAuthRedirect: ControllerHandler<
...
             )
           })
         case FormAuthType.SGID_MyInfo:
           return validateSgidForm(form).andThen(() => {
+            const { codeChallenge, codeVerifier } = generatePkcePair()
+            res.cookie(
+              SGID_CODE_VERIFIER_COOKIE_NAME,
+              codeVerifier,
+              SgidService.getCookieSettings(),
+            )
```
justynoh: nit: since now we always generate the PKCE pair before redirection, I'm wondering if we should just wrap this in `SgidService.createRedirectUrl`. Concretely, it would look something like 

```typescript
validateSgidForm(form)
.andThen(() => SgidService.createRedirectUrl(...))
.andThen(({ redirectUrl, cookie: { name, val, options } }) => {
  res.cookie(name, val, options)
  return redirectUrl
})
```

It doesn't really save us much code because we still have to handle passing values via the neverthrow monad. But the main benefit here is that the specifics of the protocol between FormSG and SGID is wrapped within the SGID service layer, rather than the public form handler having to know about the internals.

src/app/modules/sgid/__tests__/sgid.routes.spec.ts
```diff
@@ -44,7 +48,7 @@ const app = setupApp('/sgid', SgidRouter)
 
 const MOCK_LOGIN_QUERY = { code: MOCK_AUTH_CODE, state: MOCK_STATE }
 
-describe('sgid.controller', () => {
+describe('sgid.routes', () => {
```
justynoh, line 51: lmfao

src/app/modules/sgid/sgid.controller.ts
```diff
@@ -66,10 +67,24 @@ export const handleLogin: ControllerHandler<
     res.cookie('isLoginError', true)
     return res.redirect(target)
   }
+
+  // const codeVerifier = req.cookies.get(SGID_CODE_VERIFIER_COOKIE_NAME)
```
justynoh: rando comment

src/app/modules/auth/sgid/auth-sgid.controller.ts
```diff
@@ -22,18 +24,28 @@ export const generateAuthUrl: ControllerHandler<
     ...createReqMeta(req),
   }
 
-  return AuthSgidService.createRedirectUrl()
-    .map((redirectUrl) => res.status(StatusCodes.OK).send({ redirectUrl }))
+  const { codeChallenge, codeVerifier } = generatePkcePair()
```
justynoh: same comment as for public forms

justynoh (commented):
Thanks so much for doing this @kschiew ! Generally LGTM with just a few small comments

justynoh (approved):

PR #6585, opened 2023-08-01
feat: show error message if no products are selected

wanlingt:
## Problem

Payment by Products: No frontend error message appears when 0 products are selected. The user should be prompted to select at least 1 product.

## Solution

Add validation to the payment product field, by checking if the total price is equal to 0.
Also, set the default value (for multi_qty products) to be the min quantity

**Breaking Changes** 

## Screenshots
image

## Tests

- [ ] On a form with multiple products, don't select any product. Click proceed to pay. The error message should appear.
- [ ] On a form with where you can choose the quantity of products, select a product but don't select a quantity. Click proceed to pay. The error message should appear.

### DIFF
--- a/frontend/src/features/public-form/components/FormPaymentPage/components/ProductPaymentItemDetailsBlock.tsx
+++ b/frontend/src/features/public-form/components/FormPaymentPage/components/ProductPaymentItemDetailsBlock.tsx
@@ -124,7 +124,7 @@ export const ProductPaymentItemDetailsBlock = ({
     validate: (value) => {
       const totalPrice = calculatePrice(value || [])
       if (totalPrice === 0) {
-        return 'Please select at least 1 product/service'
+        return 'Please select at least 1 option'
       }
     },
   })
--- a/frontend/src/features/public-form/components/FormPaymentPage/components/ProductPaymentItemDetailsBlock.tsx
+++ b/frontend/src/features/public-form/components/FormPaymentPage/components/ProductPaymentItemDetailsBlock.tsx
@@ -107,6 +107,17 @@ const PaymentItem = ({
   )
 }
 
+const isProductSelected = (productItems: Array<ProductItem>) => {
+  let isProductSelected = false
+  for (let i = 0; i < productItems.length; i++) {
+    if (productItems[i].selected) {
+      isProductSelected = true
+      break
+    }
+  }
+  return isProductSelected
+}
+
 export const ProductPaymentItemDetailsBlock = ({
   paymentDetails,
   colorTheme,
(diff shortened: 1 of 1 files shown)


frontend/src/features/public-form/components/FormPaymentPage/components/ProductPaymentItemDetailsBlock.tsx
```diff
@@ -114,8 +114,19 @@ export const ProductPaymentItemDetailsBlock = ({
...
-  register(PAYMENT_PRODUCT_FIELD_ID)
+  const {
+    register,
+    setValue,
+    formState: { errors },
+  } = useFormContext()
+  register(PAYMENT_PRODUCT_FIELD_ID, {
+    validate: (value) => {
+      const totalPrice = calculatePrice(value || [])
+      if (totalPrice === 0) {
```
KenLSM: Hm. This looks like some black magic, and might be unexpected [1]. 

Would it be possible to calculate by products selected instead?

[1] Expectedness refers to assumptions by a lay person. I felt that a lay person would assume that the state of having something selected means that "at least 1 of the buttons needs to be selected" instead of basing off pricing (even though it is technically correct to make this assumption now).

wanlingt: Makes sense, made a change here [3aad64f](https://github.com/opengovsg/FormSG/pull/6585/commits/3aad64f9e2862d2b56de3a9b6dd279bfb653ab1d)

frontend/src/features/public-form/components/FormPaymentPage/components/ProductPaymentItemDetailsBlock.tsx
```diff
@@ -107,15 +107,38 @@ const PaymentItem = ({
...
 
+const isProductSelected = (productItems: Array<ProductItem>) => {
+  let isProductSelected = false
+  for (let i = 0; i < productItems.length; i++) {
+    if (productItems[i].selected) {
+      isProductSelected = true
+      break
+    }
+  }
+  return isProductSelected
```
KenLSM: Might be able to be terser with
```
const isSelected = productItems.some(item => item.selected)
```

KenLSM (approved):
LGTM! Thanks for the quick fix. One suggestion to be terser

frontend/src/features/public-form/components/FormPaymentPage/components/ProductPaymentItemDetailsBlock.tsx
wanlingt: Thanks! [8c1861b](https://github.com/opengovsg/FormSG/pull/6585/commits/8c1861b8a943ef59e136a4211c708cac38859813)

PR #6587, opened 2023-08-01
feat: BE endpoint to allow storage submissions before encryption

LinHuiqing:
## Problem

Currently, FormSG is E2E encrypted. As such, some features are impossible as we have completely no visibility as to what the submissions look like after they leave the client.

Closes FRM-1209

## Solution

To enable some form of server-side data manipulation / analytics before the data is encrypted, we want to move away from E2E encryption, as detailed in the [RFC](https://docs.google.com/document/d/1VmNXS_xYY2Yg30AwVqzdndBp5yRJGSDsyjBnH51ktyc/edit?usp=sharing).

**Breaking Changes** 

**Features**:

- Opens up BE endpoint (`/:formId/submissions/storage`) to accept storage mode submissions that have not been encrypted with the public key yet. This heavily references @tshuli's PR https://github.com/opengovsg/FormSG/pull/6253, but with some refactoring done to share code between email and storage modes.
- BE endpoint only makes the actual submission the form has a `newEncryptionBoundary` flag set to true.

**Improvements**:

- Introduction of `receiver` module for submissions.

## Tests

To check that endpoint works:
- [ ] Create a email mode form switch off captcha.
- [ ] Go to the respondent form and open the network tab before making a submission on the form.
- [ ] Copy and save the corresponding curl request for the email submission made.
- [ ] Duplicate the form to create the exact same form, but in storage mode. Make sure captcha is off.
- [ ] Make the following edits to the curl request previously saved:
   - [ ] formId should be updated to the storage mode form's
   - [ ] update endpoint to be `/submissions/storage`
- [ ] Perform the curl request. The request should return an error saying "This endpoint has not been enabled for this form."
- [ ] Go to the form's document on the DB and add a "newEncryptionBoundary" set to true.
- [ ] Perform the curl request. Check for the following:
   - [ ] The response should say "Form submission successful."
   - [ ] The submission should show up in the form's results page.

Regression tests:
- [ ] Submit a response on an email mode form. You should receive the corresponding email for the submission.
- [ ] Submit a response on a storage mode form. You should see that the corresponding submission entry should show up on the admin results page.
- [ ] Go to an email mode form's preview page. Make a submission. The app should not crash and you should receive a corresponding email for the submission.


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -647,13 +663,219 @@ const submitEncryptModeForm: ControllerHandler<
...
+      meta: logMeta,
+      error: formResult.error,
+    })
+    const { errorMessage, statusCode } = mapRouteError(formResult.error)
+    return res.status(statusCode).json({ message: errorMessage })
+  }
+
+  // Guardrail to prevent new endpoint from being used for regular storage mode forms.
+  // TODO (FRM-1232): remove this guardrail when encryption boundary is shifted.
+  if (!formResult.value.get('newEncryptionBoundary')) {
```
ghost, line 753: I'm weighing this against using a feature flag. With this new column, I will need to update all existing records' value to `true` for the boundary shift. And after the boundary shift, I will be storing a value that will no longer be used in all form definitions.

LinHuiqing: No ah we can just remove this guardrail, then this endpoint will be available to all storage mode forms. But yes, this implementation will not provide any flexibility in terms of controlling the percentage rollout and there is no way of updating this without going into the DB. I was thinking maybe we can do a feature flag system in a separate PR maybe? Since this PR is already pretty sizeable (although it's mostly refactoring).

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -647,13 +663,219 @@ const submitEncryptModeForm: ControllerHandler<
...
+  const encryptedAttachments =
+    await getEncryptedAttachmentsMapFromAttachmentsMap(
+      attachmentsMap,
+      publicKey,
+    )
+
+  const filteredResponses = filterSendableStorageModeResponses(
+    formResult.value.form_fields as unknown as FormFieldDto[],
+    req.body.responses.map((response) => {
+      if (isAttachmentResponse(response)) {
```
ghost: Why do we need to strip out attachment content here?
Edited for more context: `filterSendableStorageModeResponses` will have taken care of that

LinHuiqing: I think we need to strip out the attachment content when getting `encryptedContent` but not here, since attachment fields will already be taken out by `filterSendableStorageModeResponses`. Will remove this in a bit!

LinHuiqing: Done in [80d1cf3](https://github.com/opengovsg/FormSG/pull/6587/commits/80d1cf33c39d24f00488a4978d0b8e6e9d05609d), thanks!

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -647,13 +663,209 @@ const submitEncryptModeForm: ControllerHandler<
...
+// TODO (FRM-1232): remove endpoint after encryption boundary is shifted
 export const handleEncryptedSubmission = [
   CaptchaMiddleware.validateCaptchaParams,
   TurnstileMiddleware.validateTurnstileParams,
   EncryptSubmissionMiddleware.validateEncryptSubmissionParams,
   submitEncryptModeForm,
 ] as ControllerHandler[]
 
+/**
+ * Reference implementation taken from frontend/src/features/public-form/utils/createSubmission.ts for purpose of demonstrating shifted encryption boundary.
```
tshuli: /nit can update the comment, this is no longer reference implementation :)

src/app/modules/submission/submission.utils.ts
```diff
@@ -1,6 +1,11 @@
 import { differenceBy, intersectionBy, keyBy, uniqBy } from 'lodash'
 import { err, ok, Result } from 'neverthrow'
 
+import {
+  ParsedClearAttachmentResponse,
+  ParsedClearFormFieldResponse,
+} from 'src/types/api'
```
tshuli: /nit: use relative import

src/app/modules/submission/receiver/__tests__/receiver.utils.spec.ts
```diff
@@ -0,0 +1,95 @@
+import { ObjectId } from 'bson'
+import { readFileSync } from 'fs'
+import { BasicField } from 'shared/types'
+
+import {
+  FieldResponse,
+  IAttachmentResponse,
+  SingleAnswerFieldResponse,
+} from 'src/types'
```
tshuli: same here with relative imports

src/app/modules/submission/receiver/__tests__/receiver.utils.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const getResponse = (_id: string, answer: string): SingleAnswerFieldResponse =>
+  ({
+    _id,
+    fieldType: BasicField.Attachment,
+    question: 'mockQuestion',
+    answer,
+  } as unknown as SingleAnswerFieldResponse)
+
+describe('email-submission.util', () => {
```
tshuli, line 34: maybe can add some test cases for `handleDuplicatesInAttachments`

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -67,3 +67,45 @@ export const validateEncryptSubmissionParams = celebrate({
     isPreview: Joi.boolean(),
   }),
 })
+
+/**
+ * Celebrate middleware for verifying shape of encrypted submission
+ */
+export const validateStorageSubmissionParams = celebrate({
```
tshuli, line 75: this is exactly the same as `validateResponseParams` in `email-submission.middleware`, can consider sharing across the two?

src/app/modules/submission/email-submission/__tests__/email-submission.util.spec.ts
```diff
@@ -13,12 +13,7 @@ import { ObjectId } from 'bson'
 import { readFileSync } from 'fs'
 import { cloneDeep, merge } from 'lodash'
 
-import {
-  FieldResponse,
-  IAttachmentResponse,
-  SingleAnswerFieldResponse,
-  SPCPFieldTitle,
-} from 'src/types'
+import { SingleAnswerFieldResponse, SPCPFieldTitle } from 'src/types'
```
tshuli: relative import

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
tshuli: Adding on to the discusison - i thought the plan was to do it only for forms with attachment first. in that case maybe the 'guardrail' can come in the form of a list of field types which are allowed (if the form has that field type, it is accepted)

which is a separate discussion from phased rollout. for phased rollout, i thought it is more of a frontend flag rather than backend flag (similar to what we did for react). because it is a phased rollout in terms of % of respondents, not in terms of % of forms.

tshuli (commented):
thanks @huiqing! lgtm except for some minor nits; also need to decide on the guardrail / phased rollout plan before proceeding

tshuli (commented):
In `receiver.service.ts`, `createMultipartReceiver`, the `fileSize` limit of 7 * MB is for email mode forms. For storage mode, you can have up to 20MB.

```
const busboy = Busboy({
      headers,
      limits: {
        fieldSize: 3 * MB,
        fileSize: 7 * MB,
      },
    })
```

- Maybe can pass in max fileSize when calling `createMultipartReceiver`?
- To add tests as well for this

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: sorry, just realised that these things from the reference implementation can be refactored to reuse what email submission already has! give me a moment

LinHuiqing: refactored here [a0ea001](https://github.com/opengovsg/FormSG/pull/6587/commits/a0ea001e4208d4a0e639d545a36fbdc88e1e262c)!

src/app/modules/submission/submission.utils.ts
LinHuiqing: fixed in [06c5edc](https://github.com/opengovsg/FormSG/pull/6587/commits/06c5edcbdf73f320c4cbd2562172fbc81871bae5)

src/app/modules/submission/receiver/__tests__/receiver.utils.spec.ts
LinHuiqing: fixed in [06c5edc](https://github.com/opengovsg/FormSG/pull/6587/commits/06c5edcbdf73f320c4cbd2562172fbc81871bae5)

src/app/modules/submission/email-submission/__tests__/email-submission.util.spec.ts
LinHuiqing: fixed in [06c5edc](https://github.com/opengovsg/FormSG/pull/6587/commits/06c5edcbdf73f320c4cbd2562172fbc81871bae5)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: I think there is a need to protect the backend endpoint itself cos this endpoint opens up an easier way to send malicious attachments. So this is slightly different from our other features which are usually just blocked on our frontend since there's no harm using those endpoints independently.

Actually to clarify, so the attachment-first approach is to showcase our new virus scanning capabilities or? Why would we open for forms with attachment fields first before the others?

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: it's not exactly the same, this one requires version too!

LinHuiqing: point taken though, let me see if I can refactor

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
tshuli: > Actually to clarify, so the attachment-first approach is to showcase our new virus scanning capabilities or? Why would we open for forms with attachment fields first before the others?

I think this was to achieve phased rollout (see the [RFC](https://docs.google.com/document/d/1VmNXS_xYY2Yg30AwVqzdndBp5yRJGSDsyjBnH51ktyc/edit) - Phase 1). Because virus scanning was the main benefit so we would start with forms with attachment fields. If there are any bugs / problems, it would be scoped to forms with attachment fields (might be easier to contact admins to remediate)

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Done in [d328c73](https://github.com/opengovsg/FormSG/pull/6587/commits/d328c73a5411168d227190fd7204c17392387b2f)

src/app/modules/submission/receiver/__tests__/receiver.utils.spec.ts
LinHuiqing: good catch, thanks! forgot to transfer the tests for this over from email submission 😓 
done in [1964a68](https://github.com/opengovsg/FormSG/pull/6587/commits/1964a68470edc0c04b6011e99fd09f57dcc7bbb8)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Ahh yes okay, sorry I missed that. I presume we could add the flag for forms with attachment fields first for the first 3 months to track numbers and limit the scope of any bugs. But this should only be out after virus scanning is in place!

LinHuiqing: but as [discussed on slack](https://opengovproducts.slack.com/archives/CKHLS3W3X/p1691113708282249?thread_ts=1691037615.319729&cid=CKHLS3W3X), I think for this particular PR, we should still have this guardrail instead of enabling it for certain forms immediately!

sebastianwzq (approved):

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
tshuli: thanks! haha yes i missed out the version bit 😅

LinHuiqing:
> In `receiver.service.ts`, `createMultipartReceiver`, the `fileSize` limit of 7 * MB is for email mode forms. For storage mode, you can have up to 20MB.

Done in [bc3d4a9](https://github.com/opengovsg/FormSG/pull/6587/commits/bc3d4a9af96dad15365bc9f640f9b48906c6216c) and [d2974f1](https://github.com/opengovsg/FormSG/pull/6587/commits/d2974f1c464846de879754f4670cdb6e548fe19c)! Went with the approach of passing in a boolean flag instead for the `createMultipartReceiver` to decide the `fileSize`.

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: for posterity, I did it wrong in [d328c73](https://github.com/opengovsg/FormSG/pull/6587/commits/d328c73a5411168d227190fd7204c17392387b2f) by replacing the params in the wrong storage-mode middleware 😓 fixed in [6c85f4f](https://github.com/opengovsg/FormSG/pull/6587/commits/6c85f4f4a7f41dad0876aed8105d52294a45ae3d)!

LinHuiqing:
Tested on `staging-alt`!

src/app/modules/submission/receiver/receiver.middleware.ts
```diff
@@ -0,0 +1,49 @@
...
+export const receiveSubmission: ControllerHandler<
+  unknown,
+  { message: string },
+  { responses: FieldResponse[] }
+> = async (req, res, next) => {
+  const logMeta = {
+    action: 'receiveSubmission',
+    ...createReqMeta(req),
+  }
+  const emailMode = req.route.path.endsWith(`/${FormResponseMode.Email}`)
```
tshuli: For this, could I suggest instead of the middleware matching the route, we can use a higher order function

```
makeReceiveSubmissionMiddleware = (isEmailMode) => 
// returns the receiveSubmissionMiddleware with the correct filecreateMultipartReceiver
```

which returns the receiver. Then the controller passes in the correct mode when calling the higher order function

tshuli (approved):
just one suggestion to not have the middleware do the route matching; else lgtm!

src/app/modules/submission/receiver/receiver.middleware.ts
LinHuiqing: Okay I've refactored this in [f167b9f](https://github.com/opengovsg/FormSG/pull/6587/commits/f167b9f0c36814ea8126c7a6c1a4ba50f5f2a090). Let me know if this is fine!

tshuli (approved):
lgtm!

PR #6588, opened 2023-08-01
feat: filter out API details from user object

wanlingt:
## Problem

The full `user` object, which includes the `apiToken` property, is returned in multiple queries.

Closes FRM-1011

## Solution

Add `select` to the `apiToken` property in the user model to filter out `apiToken` from any query that returns the full `user` object 
(from [docs](https://mongoosejs.com/docs/2.7.x/docs/schematypes.html): _select: {Boolean} - Specifies default path selection behavior. In other words, you can specify if this path should be included or excluded from query results by default._)

**Breaking Changes** 

## Tests

- [x] Update a user with the `apiToken` property (`{keyHash: .., createdAt:.., lastUsedAt:.. }`). Check that none of the following queries returns the `apiToken` property with the user object:
  - [x] `GET /api/v3/admin/forms` - visit admin form dashboard page
  - [x] `POST /api/v3/auth/otp/verify` - log out and log in again
  - [x] `GET /api/v3/user` - visit admin form dashboard page
  - [x] `POST /user/contact/otp/verify` - verify contact OTP

### DIFF
--- a/__tests__/unit/backend/helpers/jest-db.ts
+++ b/__tests__/unit/backend/helpers/jest-db.ts
@@ -24,6 +24,7 @@ import {
   IPopulatedForm,
   ISubmissionSchema,
   IUserSchema,
+  UserApiToken,
 } from 'src/types'
 
 /**
@@ -88,11 +89,13 @@ const insertUser = async ({
   userId,
   mailDomain = 'test.gov.sg',
   mailName = 'test',
+  apiToken,
 }: {
   agencyId: ObjectID
   userId?: ObjectID
   mailName?: string
   mailDomain?: string
+  apiToken?: UserApiToken
 }): Promise<IUserSchema> => {
   const User = getUserModel(mongoose)
 
@@ -100,6 +103,7 @@ const insertUser = async ({
     email: `${mailName}@${mailDomain}`,
     _id: userId,
     agency: agencyId,
+    apiToken: apiToken,
   })
 }
 
@@ -116,6 +120,7 @@ const insertFormCollectionReqs = async ({
   shortName = 'govtest',
   flags,
   betaFlags,
+  apiToken,
 }: {
   userId?: ObjectID
   mailName?: string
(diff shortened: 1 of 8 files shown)


sebastianwzq (approved):

tshuli (commented):
thanks @wanlingt! as an alternative to updating each query to de-select apiToken, would you like to consider setting `select:false` at the schema level? This should ensure that apiToken is not returned by default in all queries (in case someone forgets to deselect next time)😅 

see [https://mongoosejs.com/docs/2.7.x/docs/schematypes.html
](https://mongoosejs.com/docs/2.7.x/docs/schematypes.html
)
```
select: {Boolean} - Specifies default path selection behavior. In other words, you can specify if this path should be included or excluded from query results by default.
```

wanlingt:
> thanks @wanlingt! as an alternative to updating each query to de-select apiToken, would you like to consider setting `select:false` at the schema level? This should ensure that apiToken is not returned by default in all queries (in case someone forgets to deselect next time)😅
> 
> see [https://mongoosejs.com/docs/2.7.x/docs/schematypes.html ](https://mongoosejs.com/docs/2.7.x/docs/schematypes.html)
> 
> ```
> select: {Boolean} - Specifies default path selection behavior. In other words, you can specify if this path should be included or excluded from query results by default.
> ```

that is awesome, thanks for the suggestion @tshuli ! I've added it to the user model in [f2b8dc2](https://github.com/opengovsg/FormSG/pull/6588/commits/f2b8dc2a6f8d220a22d818cec54f455f7248c368)

tshuli (approved):
lgtm! maybe just add one unit test to check that the select:false property is working property

wanlingt:
> lgtm! maybe just add one unit test to check that the select:false property is working property

done! [c773b37](https://github.com/opengovsg/FormSG/pull/6588/commits/c773b37241fc9b7fbd4fa8783598def59cbdbd3d)

PR #6590, opened 2023-08-03
feat: toggle issue notification

foochifa:
## Problem

Closes FRM-1183

## Solution

Added `hasIssueNotification` flag for `form.server.model` with default true

Rely on `updateFormSettings` to toggle flag

Created FE toggle based on `FormCaptchaToggle` and mutate `updateFormSettings`

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:

image

## Tests

**test functionality**
- [x] Create a new form
- [x] Open form
- [x] Report issue, you should receive email notification
- [x] Create a new form
- [x] Toggle off notification on general settings page
- [x] Open form
- [x] Report issue, you shouldn't receive email notification
- [x] Check that issue is still in form->results->feedback

**test regression**
- [x] On an existing form without modifying general settings page
- [x] Open form
- [x] Report issue, you should receive email notification


frontend/src/features/admin-form/settings/components/FormIssueNotificationToggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const handleToggleIssueNotification = useCallback(() => {
+    if (!settings || isLoadingSettings || mutateFormIssueNotification.isLoading)
+      return
+    const nextHasIssueNotification = !settings.hasIssueNotification
+    return mutateFormIssueNotification.mutate(nextHasIssueNotification)
+  }, [isLoadingSettings, mutateFormIssueNotification, settings])
+
+  return (
+    <Skeleton isLoaded={!isLoadingSettings && !!settings}>
```
LinHuiqing: Do we need to check both of these? Is there a situation where the `isLoadingSettings` state is false but `settings` is undefined?

foochifa: Hmm goodpoint, I think it is unlikely, its improbable to get an empty setting after the query is successful because of our form defaults.

So if it succeeds, should have finish loaded, and setting should be defined.

Will remove the `!!settings` check!

foochifa: [13c69e3](https://github.com/opengovsg/FormSG/pull/6590/commits/13c69e3903770679ec2fb5e0f2adbdf38ae6412a)

LinHuiqing: Oh there's another part of the code with the repeated checks too! Wait ah

frontend/src/features/admin-form/settings/components/FormIssueNotificationToggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const hasIssueNotification = useMemo(
+    () => settings && settings?.hasIssueNotification,
+    [settings],
+  )
+
+  const { mutateFormIssueNotification } = useMutateFormSettings()
+
+  const handleToggleIssueNotification = useCallback(() => {
+    if (!settings || isLoadingSettings || mutateFormIssueNotification.isLoading)
```
LinHuiqing, line 21: Same comment for this part with the repeated checks! When you make a decision of which one to use, remember to update the dependencies below too.

LinHuiqing (commented):
generally lgtm! will test later in the day before approving

frontend/src/features/admin-form/settings/components/FormIssueNotificationToggle.tsx
foochifa: Ah I can't do it there @LinHuiqing, cause it will cause typing issues on the next line `const nextHasIssueNotification = !settings.hasIssueNotification`

as settings is `FormSettings | undefined`

LinHuiqing: Ah got it, because there's no "settings is not undefined" check here.

LinHuiqing (approved):
lgtm and tested!

PR #6591, opened 2023-08-03
feat: add payment summary to thank you page

wanlingt:
## Problem

Closes FRM-1199

## Solution

- add `payment_fields_snapshot` to `payments` model to capture the form's payment fields at point of payment
- return additional fields: `amount`, `payment_fields_snapshot`, `paymentDate` from `GET /payments/:paymentId/getinfo`. This information is used to populate the Payment summary in the thank you page

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

**AFTER**:

image

## Deploy Notes

**New scripts**:

- `set-payment-fields-snapshot.js` : This script adds the `payment_fields_snapshot` field to the payments collection and sets the field as the form document's `payment_field` for all existing documents.


frontend/src/features/public-form/components/FormPaymentPage/stripe/components/DownloadReceiptBlock.tsx
```diff
@@ -1,46 +1,116 @@
...
+  amount,
+  products,
+  paymentType,
+  name,
+  paymentDate,
 }: DownloadReceiptBlockProps) => {
   const toast = useToast({ status: 'success', isClosable: true })
 
+  // Extract selected product names for payment by products
+  const extractProductNames = (products: ProductItem[]) => {
```
KenLSM: /nit This function is already pure, we could also extract this function to be out of `<DownloadReceiptBlock />` as this function will be recreated whenever this component is re-rendered.

This will be ultra-nit, Might also be terser in more fp-style  😆 
```js
const getProductNames = (items: ProductItem[]) => products
  .filter(product => product.selected)
  .map(product => `${product.quantity}x ${product.data.name}`)
  .join(', ')
```

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/DownloadReceiptBlock.tsx
```diff
@@ -1,46 +1,116 @@
...
+          <Text textStyle="subhead-1">
+            Your form has been submitted and payment has been made.
+          </Text>
+        </Stack>
+        <Divider />
+        <Stack>
+          <Text textStyle="h2" mb="0.5rem">
+            Payment summary
+          </Text>
+          <PaymentSummaryItem title="Product/service" input={productName} />
```
KenLSM: Great abstraction here!

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/DownloadReceiptBlock.tsx
```diff
@@ -1,46 +1,116 @@
...
   submissionId: string
   paymentId: string
+  amount: number
+  products: ProductItem[]
+  paymentType?: PaymentType
+  name: string
+  paymentDate?: Date
+}
+
+const PaymentSummaryItem = ({
```
KenLSM: Since this might not always represent an `Item` (as in product item in the payment/product domain), it might make sense to name this as `PaymentSummaryRow`. 

Wdyt?

src/app/models/payment.server.model.ts
```diff
@@ -107,6 +108,63 @@ const PaymentSchema = new Schema<IPaymentSchema, IPaymentModel>(
         },
       ],
     },
+
+    payment_fields_snapshot: {
```
KenLSM: I'm thinking that we should set `payment_fields_snapshot` to be exactly `Form['payment_fields']` so that we only have one single source of truth (which is from `Form`)

KenLSM (approved):
LGTM! 

Only one key point on the source of truth on the snapshot that might need some changes

src/app/models/payment.server.model.ts
wanlingt: yes good point! fixed here [752654e](https://github.com/opengovsg/FormSG/pull/6591/commits/752654e4a54fdf06b478f6b8bbf8ef4c328f4a35)

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/DownloadReceiptBlock.tsx
wanlingt: [2353977](https://github.com/opengovsg/FormSG/pull/6591/commits/23539777cf1a763a9cb3d35523dc85d0063015f7)

frontend/src/features/public-form/components/FormPaymentPage/stripe/components/DownloadReceiptBlock.tsx
wanlingt: makes sense! [2353977](https://github.com/opengovsg/FormSG/pull/6591/commits/23539777cf1a763a9cb3d35523dc85d0063015f7)

PR #6592, opened 2023-08-03
fix(workspaces): remove form from workspace upon collaborator removal

### LINKED ISSUE #4073: [React/Form Workspaces] Update workspace upon removing collaborator from a form
When we remove a collaborator from a form, the collaborator should no longer be able to see the form. Hence, the form should be removed from any and all of the collaborator's workspaces

foochifa:
## Problem

Currently, upon removal of an admin as a collaborator of a form, we do not remove the form from the workspaces of the admin. Leaving the admin with a workspace that indicates that it has more forms than actuality.

This PR fixes that

Closes FRM-144
Closes #4073 

Thanks @kennethchangOPENGOV for identifying the bug

## Solution

Upon update of collaborator, will check if there are any removed emails. 

If there are, will iterate through it, and remove formId from each user email's workspaces.

**Breaking Changes** 

## Tests

- [ ] Create a form
- [ ] Add another email as collaborator

Here on out is the other email
- [ ] Log in as the other email
- [ ] Create a workspace 
- [ ] Move the form above to the workspace
- [ ] Remove yourself as collaborator
- [ ] Form should be removed from workspace and workspace tab should show the correct number of forms in it

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.service.ts
@@ -67,6 +67,7 @@ import { MissingUserError } from '../../user/user.errors'
 import * as UserService from '../../user/user.service'
 import { SmsLimitExceededError } from '../../verification/verification.errors'
 import { hasAdminExceededFreeSmsLimit } from '../../verification/verification.util'
+import { removeFormFromAllWorkspaces } from '../../workspace/workspace.service'
 import {
   FormNotFoundError,
   LogicNotFoundError,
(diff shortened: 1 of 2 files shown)


src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -790,46 +791,70 @@ export const updateFormCollaborators = (
...
+        const falseIdx = doNewCollaboratorsExist.findIndex((exists) => !exists)
+        return falseIdx < 0
+          ? okAsync(undefined)
+          : errAsync(
+              new InvalidCollaboratorError(updatedCollaboratorEmails[falseIdx]),
+            )
+      })
+      // after checking that collaborator exists
+      // remove forms from workspaces for collaborators that were removed
+      .andThen(() => {
```
tshuli, line 831: Do we need to wrap this in await Promise.all? otherwise removedCollaboratorEmails.map returns Promise<void>[], and we would return okAsync(undefined) immediately

tshuli (approved):
one comment on promise handling, otherwise lgtm!

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -790,46 +791,70 @@ export const updateFormCollaborators = (
...
+        removedCollaboratorEmails.map(async (collaborator) => {
+          await UserService.findUserByEmail(collaborator.email).map(
+            async (user) =>
+              await removeFormFromAllWorkspaces({
+                formId: form._id,
+                userId: user._id,
+              }),
+          )
+        })
+        return okAsync(undefined)
```
tshuli, line 841: new Promise((resolve, reject) => {
    setTimeout(() => {
        console.log('ok')
        resolve('Hello world')
    }, 1000)
}

tshuli: Do we need to wrap this in ResultAsync.fromPromise(Promise.all...)? otherwise removedCollaboratorEmails.map returns Promise[], and we would return okAsync(undefined) immediately

foochifa: Yep thanks for catching this Shuli! Should definitely be more defensive! [8fbeaac](https://github.com/opengovsg/FormSG/pull/6592/commits/8fbeaac340f60b28ed48940165dc546d4a84d88c)

Issue #780, opened 2020-12-02
Shard Submission service into separate Encrypted and Email services


liangyuanruo:
Storage mode responses and email responses require separate processing. Shard out mode-specific functionality from the existing Submission service into two new separate services for clarity and the avoidance of doubt. 

Blocked by #1104

mantariksh:
if the opportunity arises, we should also use this refactor to avoid modifying the shape of responses after we receive them. for example, for email submissions, we should keep attachments as its own array instead of assigning additional `filename` and `content` keys to attachment responses in the `responses` array (see `addAttachmentsToResponses`). this results in strange behaviour, e.g. we start off with an array of attachments, then distribute the information in it to the attachment responses in the `responses` array, then later on we reconstruct the attachments array that we started with.

mantariksh:
Subsumed by #1104, which will split the mode-specific service functionality into two separate classes

PR #6595, opened 2023-08-04
feat: storage submission validation with encryption shift endpoint

LinHuiqing:
## Problem

Now that the new endpoint is up for the encryption boundary shift (#6587), we want to validate the submission coming in through this endpoint so that there is some value add compared to the existing `/submissions/encrypt` endpoint.

Note: This is gated behind a form-level `encryptionBoundaryShift` flag.

Closes FRM-1211

## Solution

**Breaking Changes** 

**Features**:

- Validation of submissions coming through `/submissions/storage`.

**Improvements**:

- A more informative error message will now be shown if a submission is attempted on a form that has been deleted or closed.

## Tests

To check that existing email submission validation works:
- [ ] Create a email mode form **_with logic_** and switch off captcha.
- [ ] Go to the respondent form and open the network tab before making a submission on the form.
- [ ] Copy and save the corresponding curl request for the email submission made.
- [ ] Make an invalid edit to the email submission curl request:
   - [ ] Add an invalid value to a field before any logic. It should throw an appropriate validation error.
   - [ ] Add a value to a field that is hidden by logic. It should throw a validation error saying that the you're trying to submit with a value on a hidden field.
- [ ] To the previous request, edit it such that the field previously hidden by logic will be shown. The submission should now succeed.

To check that new endpoint is still able to make valid submissions:
- [ ] Duplicate the form to create the exact same form, but in **_storage mode_**. Make sure captcha is off.
- [ ] Make the following edits to the curl request previously saved:
   - [ ] formId should be updated to the storage mode form's
   - [ ] update endpoint to be `/submissions/storage`
   - [ ] Add `"version":1` to the response body
- [ ] Perform the curl request. The request should return an error saying "This endpoint has not been enabled for this form."
- [ ] Go to the form's document on the DB and add a "encryptionBoundaryShift" set to true.
- [ ] Perform the curl request. Check for the following:
   - [ ] The response should say "Form submission successful."
   - [ ] The submission should show up in the form's results page.

To check that submission validation for new endpoint works:
- [ ] Modify the curl request with invalid submissions.
   - [ ] Add an invalid value to a field before any logic. It should throw an appropriate validation error.
   - [ ] Add a value to a field that is hidden by logic. It should throw a validation error saying that the you're trying to submit with a value on a hidden field.
- [ ] To the previous request, edit it such that the field previously hidden by logic will be shown. The submission should now succeed.

Regression tests:
- [ ] Submit a response on an email mode form. You should receive the corresponding email for the submission.
- [ ] Submit a response on a storage mode form. You should see that the corresponding submission entry should show up on the admin results page.
- [ ] Go to an email mode form's preview page. Make a submission. The app should not crash and you should receive a corresponding email for the submission.


src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -124,6 +126,655 @@ const submitEncryptModeForm: ControllerHandler<
...
+  if (formPublicResult.isErr()) {
+    logger.warn({
+      message: 'Attempt to submit non-public form',
+      meta: logMeta,
+      error: formPublicResult.error,
+    })
+    const { statusCode, errorMessage } = mapRouteError(formPublicResult.error)
+    if (statusCode === StatusCodes.GONE) {
+      return res.sendStatus(statusCode)
+    } else {
```
tshuli: Qn:
- Does the frontend convert 410 status to appropriate error message?
- Might want to consider standardising and return error message directly from backend
(same for submitStorageModeForm)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -813,24 +1421,22 @@ const encryptSubmission: ControllerHandler<
...
 }
 
 export const handleStorageSubmission = [
   CaptchaMiddleware.validateCaptchaParams,
   TurnstileMiddleware.validateTurnstileParams,
   ReceiverMiddleware.receiveStorageSubmission,
   EncryptSubmissionMiddleware.validateStorageSubmissionParams,
-  encryptSubmission,
-  submitEncryptModeForm,
+  submitStorageModeForm,
```
tshuli: question: the logic for submitStorageModeForm and submitEncryptModeForm seems mostly repetitive; what was the consideration behind splitting them in commit 1de0e65, vs having an encryptSubmission middleware (original implementation)?

src/app/modules/submission/submission.utils.ts
```diff
@@ -23,14 +23,10 @@ type ResponseModeFilterParam = {
...
+  encrypted: boolean,
 ): (<T extends ResponseModeFilterParam>(responses: T[]) => T[]) => {
-  switch (responseMode) {
-    case FormResponseMode.Email:
-      return emailResponseModeFilter
-    case FormResponseMode.Encrypt:
-      return encryptResponseModeFilter
-  }
+  if (encrypted) return encryptResponseModeFilter
+  else return emailResponseModeFilter
```
tshuli: If you want to change the signature as such (encrypted vs non-encrypted), then would need to rename emailResponseModeFilter (since it is actually email + non-e2ee storage mode)

src/app/modules/submission/submission.utils.ts
```diff
@@ -23,14 +23,10 @@ type ResponseModeFilterParam = {
 
 // Exported for testing.
 export const getResponseModeFilter = (
-  responseMode: FormResponseMode,
+  encrypted: boolean,
```
tshuli: if boolean, the `isEncrypted` is better. Maybe `isE2EEMode?

src/app/modules/submission/submission.utils.ts
```diff
@@ -68,14 +64,10 @@ const encryptFormFieldModeFilter = <T extends FormField>(
...
 ): (<T extends FormField>(responses: T[]) => T[]) => {
-  switch (responseMode) {
-    case FormResponseMode.Email:
-      return emailResponseModeFilter
-    case FormResponseMode.Encrypt:
-      return encryptFormFieldModeFilter
-  }
+  if (encrypted) return encryptFormFieldModeFilter
+  else return emailResponseModeFilter
```
tshuli: same comment as `getResponseModeFilter`

src/app/modules/submission/submission.utils.ts
```diff
@@ -125,8 +117,14 @@ export const getFilteredResponses = (
...
 ): Result<FilteredResponse[], ConflictError> => {
-  const responseModeFilter = getResponseModeFilter(form.responseMode)
-  const formFieldModeFilter = getFormFieldModeFilter(form.responseMode)
+  const responseModeFilter = getResponseModeFilter(
+    form.responseMode === FormResponseMode.Encrypt &&
+      !form.get('newEncryptionBoundary'),
+  )
+  const formFieldModeFilter = getFormFieldModeFilter(
+    form.responseMode === FormResponseMode.Encrypt &&
+      !form.get('newEncryptionBoundary'),
```
tshuli: consider: 

const isEncryptedMode (or isE2EEMode?) = FormResponseMode.Encrypt && !form.get('newEncryptionBoundary')

const responseModeFilter = getResponseModeFilter(isEncryptedMode)
const formFieldModeFilter = getFormFieldModeFilter(isEncryptedMode)

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
```diff
@@ -37,7 +37,11 @@ export default class IncomingEncryptSubmission extends IncomingSubmission {
     ProcessingError | ConflictError | ValidateFieldError[]
   > {
     return checkIsEncryptedEncoding(encryptedContent)
-      .andThen(() => getFilteredResponses(form, responses))
+      .andThen(() => {
+        const newEncryptionBoundary: boolean = form.get('newEncryptionBoundary')
+        if (newEncryptionBoundary) return ok(responses as FilteredResponse[])
+        else return getFilteredResponses(form, responses)
+      })
```
tshuli: to add unit tests for this

src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
```diff
@@ -152,7 +152,8 @@ export default class ParsedResponsesObject {
           // encrypt mode submissions with responses on unhidden fields
           // TODO(#780): Remove this once submission service is separated into
           // Email and Encrypted services
-          form.responseMode === FormResponseMode.Encrypt
+          form.responseMode === FormResponseMode.Encrypt &&
+          !form.get('newEncryptionBoundary')
```
tshuli: to add unit test for this

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -124,6 +126,655 @@ const submitEncryptModeForm: ControllerHandler<
...
+    timestamp: (submission.created || new Date()).getTime(),
+  })
+
+  return await performEncryptPostSubmissionActions(
+    submission,
+    incomingSubmission.responses,
+  )
+}
+
+const submitStorageModeForm: ControllerHandler<
```
tshuli: if you really want to split, should duplicate the unit tests too

src/app/modules/submission/submission.utils.ts
tshuli: Suggest add a todo also to deprecate this function once we deprecate e2ee storage mode

tshuli (commented):
Thanks @LinHuiqing ! think moving along on the right track 😃  
suggested some comments for consideration

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Nope, the frontend does not. I'm assuming this has to do with the [message on formsg-clowning](https://opengovproducts.slack.com/archives/C019EQSBLTD/p1691974800148029). Okay! Will just return the error message from backend ("This form is no longer active").

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Okay, so I went with duplicating the code for commit https://github.com/opengovsg/FormSG/commit/1de0e650c52a419cc8d26e6ed76bef677492ace8 so that I just need to retrieve the form definition once, then 1) perform the relevant checks (submission limits, etc), 2) perform submission validation, 3) encrypt submission payload, then 4) submit the encrypted payload. To be very honest, I think I went with just copying the code then inserting step 2 so that I will get something working first.
I think this isn't the best way of doing this too though, with the amount of code duplication and its not very maintainable. Maybe we can refactor to do something like the [ensure middlewares suggested in Ken's doc](https://www.notion.so/opengov/API-overview-ensure-middlewares-e15206c1f7994e4ebd64d0ca06d79acc)?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Done in [3de6004](https://github.com/opengovsg/FormSG/pull/6595/commits/3de600437b87ce1105f86834bbb4eb95fab476d4)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Let me see what I can do about separating the steps into separate middleware without repeatedly retrieving the form definition from the DB.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
tshuli: Haha yes admittedly I paid attention to this because of the slack convo 😅 
but I did find it unusual that GONE was a special case where we did not return an error message explicitly
thanks for fixing!

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
tshuli: Cool, thanks @LinHuiqing !

src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
ghost: nit: `newEncryptionBoundary` is repeated several times in different files. Probably a good idea to create a constant for it so it's easier to track usages when we need to remove this flag

src/app/modules/submission/submission.utils.ts
LinHuiqing: Done in [4bf6286](https://github.com/opengovsg/FormSG/pull/6595/commits/4bf6286022cba46dd59163f21e396c98fb21fbf5)! Hope the new names are good?

src/app/modules/submission/submission.utils.ts
LinHuiqing: Done in [4bf6286](https://github.com/opengovsg/FormSG/pull/6595/commits/4bf6286022cba46dd59163f21e396c98fb21fbf5)! Hope the new names are good?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: refactored in [132021e](https://github.com/opengovsg/FormSG/pull/6595/commits/132021ee7a92d805a4482dd8a8ec8073b243ddb9) so this is no longer applicable

src/app/modules/submission/submission.utils.ts
tshuli: Sounds good!

src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
LinHuiqing: Good point, thanks! Done in [6c0c3ab](https://github.com/opengovsg/FormSG/pull/6595/commits/6c0c3ab4917d6d9501cb5a23bd3c98bf853d2183)

src/app/modules/submission/submission.utils.ts
LinHuiqing: Done in [4bf6286](https://github.com/opengovsg/FormSG/pull/6595/commits/4bf6286022cba46dd59163f21e396c98fb21fbf5) and [7072960](https://github.com/opengovsg/FormSG/pull/6595/commits/7072960a6072c29aba71f49956e20a962735809b)!

src/app/modules/submission/submission.utils.ts
LinHuiqing: done in [3b8025a](https://github.com/opengovsg/FormSG/pull/6595/commits/3b8025a70513f8472ad4e6f00d205f32ba2fe5b7), thanks!

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
LinHuiqing: added in [a61ab60](https://github.com/opengovsg/FormSG/pull/6595/commits/a61ab60ac5395dbefa077fe34cc6a424efbdca0e)

src/app/modules/submission/email-submission/ParsedResponsesObject.class.ts
LinHuiqing: I realised that this isn't used for the old storage mode submission endpoint anyway! So I removed the conditional in [0dbda94](https://github.com/opengovsg/FormSG/pull/6595/commits/0dbda9428cff48db6f3efe92efe663e6738d6334). I think this is what should be done with the TODO tagged anyway (#780)

edit: there was a bug in the above commit as I removed the `isVisible` prop completely, fixed in [c61b05a](https://github.com/opengovsg/FormSG/pull/6595/commits/c61b05a45c59462f23535e45a1d71e9be26a7a10)

shared/types/form/form.ts
```diff
@@ -150,6 +150,8 @@ export interface FormBase {
   webhook: FormWebhook
 
   responseMode: FormResponseMode
+
+  newEncryptionBoundary: boolean
```
tshuli: For this, as the encryption boundary shift only applies to storage mode, should we move this under StorageFormBase instead?

tshuli: On further thought, i think it is ok since it is an internal flag that we will remove after the rollout. We may want to test sending email mode forms to the new non-encrypted endpoint, for debugging purpose.

LinHuiqing: Hmm good point for shifting it to `StorageFormBase`. What do you mean by debugging purposes using email mode forms actually? Do you mean for virus scanning? 
Side question: for virus scanning of attachments in email mode forms, are we currently dependent on the virus scanning by email clients?

tshuli: I was thinking that in the process of debugging, we might want to send an email mode form to the non encrypted mode form endpoint. so wasn't too fixed on necessarily moving it to `StorageFormBase`

yes, there is virus scanning by email clients atm, so virus scanning for email mode is not a priority. we could implement it later for defence in depth

src/app/modules/submission/submission.service.ts
```diff
@@ -318,3 +327,48 @@ export const copyPendingSubmissionToSubmissions = (
...
+ * @returns okAsync(true) if validation passes
+ * @returns errAsync(InvalidFileExtensionError) if invalid file extensions are found
+ * @returns errAsync(AttachmentTooLargeError) if total attachment size exceeds 7MB
+ */
+export const validateAttachments = (
+  parsedResponses: ParsedClearFormFieldResponse[],
+): ResultAsync<true, InvalidFileExtensionError | AttachmentTooLargeError> => {
+  const logMeta = { action: 'validateAttachments' }
+  const attachments = mapAttachmentsFromResponses(parsedResponses)
+  if (areAttachmentsMoreThan7MB(attachments)) {
```
tshuli: since this is being shared across email and storage, need to have separate limit for storage

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
...
+    if (!formDef) return res.send(new FormDefinitionNotRetrievedError())
+
+    const logMeta = {
+      action: 'validateSubmission',
+      ...createReqMeta(req),
+      formId: formDef.id,
+    }
+
+    // Validate submission
+    return await SubmissionService.validateAttachments(req.body.responses)
```
tshuli: Here - file size limits are higher for storage mode

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -94,286 +85,258 @@ const Payment = getPaymentModel(mongoose)
...
-      return res.status(statusCode).json({
-        message: errorMessage,
-      })
-    }
-    const form = checkFormIsEncryptModeResult.value
+  const ensurePipeline = new Pipeline(
+    ensurePublicForm,
+    ensureValidCaptcha,
+    ensureFormWithinSubmissionLimits,
+  )
```
tshuli, line 109: Nice use of ensure pipeline :) (shoutout to @KenLSM  too)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -94,286 +85,258 @@ const Payment = getPaymentModel(mongoose)
...
+    res,
+  })
 
-    const hasEnsuredAll = await ensurePipeline.execute({
-      form,
-      logMeta,
-      req,
-      res,
-    })
+  if (!hasEnsuredAll) return
```
tshuli: clarification as first time reviewing the ensure middleware - what happens if one of the middleware fails, will the correct status and message be sent in the response?

KenLSM: The ensures are responsible for sending out the response and failing the pipeline. So they **must** do both as convention for this particular sets of ensures.

I have not seen other use cases of ensures yet, thus didn't require that all ensures to follow such convention. I suppose that once we have more datapoints, we can make a better call whether to have this as a convention.

src/app/models/form.server.model.ts
```diff
@@ -545,6 +545,11 @@ const compileFormModel = (db: Mongoose): IFormModel => {
         required: false,
         default: '',
       },
+
+      newEncryptionBoundary: {
```
KenLSM: nit: Would prefer not to use naming that is not futureproofed. What if there's a need to have a second boundary shift 😆 (`newerEncryptionBoundary`/`newestEncryptionBoundary`?)

Understand that this is probably temporary, thus a nit.

LinHuiqing: hmm okay will something like nonE2EEStorageMode instead work? 🤔

src/app/modules/submission/submission.service.ts
LinHuiqing: Ah, good catch, thanks!

shared/types/form/form.ts
LinHuiqing: Okay I see what you mean 🤔 Not sure if using email mode forms to debug the new storage mode endpoint will be very useful since there are things like the public key which is required by storage mode forms but not email mode forms though.

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
...
+    })
+    const { errorMessage, statusCode } = mapRouteError(formResult.error)
+    return res.status(statusCode).json({ message: errorMessage })
+  }
+
+  req.body.formsg.formDef = formResult.value
+  return next()
+}
+
+export const checkPublicKey: SharedSubmissionMiddlewareHandlerType = async (
```
tshuli, line 161: suggestion: `checkAndInjectPublicKey`, since we we storing the public key in res.locals (not just a simple validation middleware)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -670,173 +637,26 @@ export const handleEncryptedSubmission = [
   CaptchaMiddleware.validateCaptchaParams,
   TurnstileMiddleware.validateTurnstileParams,
   EncryptSubmissionMiddleware.validateEncryptSubmissionParams,
+  EncryptSubmissionMiddleware.createFormsgReqBody,
+  EncryptSubmissionMiddleware.retrieveForm,
+  EncryptSubmissionMiddleware.checkPublicKey,
+  EncryptSubmissionMiddleware.moveEncryptedPayload,
+  EncryptSubmissionMiddleware.checkEncryptMode,
```
tshuli: proposed refactor: Suggest to group this as 'processEncryptedSubmission', i.e.

const processEncryptedSubmission = [EncryptSubmissionMiddleware.createFormsgReqBody, <...>]
and then export const handleEncryptedSubmission = [<...>, ...processEncryptedSubmission]

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -670,173 +637,26 @@ export const handleEncryptedSubmission = [
...
   ReceiverMiddleware.receiveStorageSubmission,
   EncryptSubmissionMiddleware.validateStorageSubmissionParams,
-  encryptSubmission,
+  EncryptSubmissionMiddleware.createFormsgReqBody,
+  EncryptSubmissionMiddleware.retrieveForm,
+  EncryptSubmissionMiddleware.checkNewBoundaryEnabled,
+  EncryptSubmissionMiddleware.checkPublicKey,
+  EncryptSubmissionMiddleware.validateSubmission,
+  EncryptSubmissionMiddleware.encryptSubmission,
+  EncryptSubmissionMiddleware.checkEncryptMode,
```
tshuli: likewise here, can consider grouping so its neater

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
...
+    logger.warn({
+      message: 'Failed to retrieve form from database',
+      meta: logMeta,
+      error: formResult.error,
+    })
+    const { errorMessage, statusCode } = mapRouteError(formResult.error)
+    return res.status(statusCode).json({ message: errorMessage })
+  }
+
+  req.body.formsg.formDef = formResult.value
```
tshuli: should we store in res.locals? (as this was not actually part of the original request body)

src/app/modules/submission/encrypt-submission/encrypt-submission.constants.ts
```diff
@@ -0,0 +1 @@
+export const newEncryptionBoundaryFlag = 'newEncryptionBoundary'
```
tshuli: /nit: uppercase for const, i.e. NEW_ENCRYPTION_BOUNDARY_FLAG

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
...
+      return response
+    }
+  })
+
+  const encryptedContent = formsgSdk.crypto.encrypt(
+    strippedBodyResponses,
+    publicKey,
+  )
+
+  req.body.formsg.encryptedPayload = {
```
tshuli: likewise, should we consider putting this in res.locals?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
tshuli: may want to group in EncryptSubmissionMiddleware as an export, so its cleanly separated from the controller logic

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
...
+    })
+    const { statusCode, errorMessage } = mapRouteError(
+      checkFormIsEncryptModeResult.error,
+    )
+    return res.status(statusCode).json({
+      message: errorMessage,
+    })
+  }
+
+  req.body.formsg.encryptedFormDef = checkFormIsEncryptModeResult.value
```
tshuli: same here, may want to store in res.locals

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
...
+}
+
+export const checkPublicKey: SharedSubmissionMiddlewareHandlerType = async (
+  req,
+  res,
+  next,
+) => {
+  const formDef = req.body.formsg.formDef
+
+  if (!formDef) return res.send(new FormDefinitionNotRetrievedError())
```
tshuli: at this point, it should no longer be undefined, i.e. formDef it should be part of the function parameter signature so that typescript can check that it is correctly passed on

it does mean also correctly setting the type for the preceding middleware :)

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
...
+    }
+
+    return next()
+  }
+
+export const validateSubmission: StorageSubmissionMiddlewareHandlerType =
+  async (req, res, next) => {
+    const formDef = req.body.formsg.formDef
+
+    if (!formDef) return res.send(new FormDefinitionNotRetrievedError())
```
tshuli: same here - use typing to guard against missing formDef, instead of at runtime

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +119,303 @@ export const validateStorageSubmissionParams = celebrate({
     version: Joi.number().required(),
   }),
 })
+
+export const createFormsgReqBody: SharedSubmissionMiddlewareHandlerType =
+  async (req, res, next) => {
+    if (req.body.formsg) return res.send(new FormsgReqBodyExistsError())
```
tshuli: qn: is this to guard against the case where middleware is called again later and potentially overrides an existing req.body.formsg?
- if yes, may want to use typing to guard against, instead of returning error at runtime

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -94,286 +85,258 @@ const Payment = getPaymentModel(mongoose)
...
+    return res.status(statusCode).json({
+      message: errorMessage,
+    })
+  }
+  const incomingSubmission = incomingSubmissionResult.value
 
-    delete (req.body as SetOptional<EncryptSubmissionDto, 'responses'>)
-      .responses
+  // delete (req.body as SetOptional<EncryptSubmissionDto, 'responses'>)
+  //   .responses
```
tshuli: dangling comment

shared/types/form/form.ts
tshuli: Got it; fine to shift it to `StorageFormBase` in that case. Don't feel strongly eitherway

src/app/modules/submission/submission.service.ts
LinHuiqing: Fixed here: [31847c4](https://github.com/opengovsg/FormSG/pull/6595/commits/31847c41c2edd8514e97779195f46e2a7a239996)

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Same as the above comment - fixed here: [31847c4](https://github.com/opengovsg/FormSG/pull/6595/commits/31847c41c2edd8514e97779195f46e2a7a239996)

tshuli (commented):
thanks @LinHuiqing! couple of remarks for this large diff and we also discussed how to clean up the middleware a bit more :) 
will review one last time before merge in

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: That was something I did originally (can check out to [d22f450](https://github.com/opengovsg/FormSG/pull/6595/commits/d22f450baad1001336308a95bded67ecbff1849a)) but I was discussing with @timotheeg about this, and he mentioned that the purpose for `res.locals` is for template rendering purposes back before single page applications were popular. As according to [Express' docs on res.locals](https://expressjs.com/en/api.html#res.locals):

> Use this property to set variables accessible in templates rendered with [res.render](https://expressjs.com/en/api.html#res.render). The variables set on res.locals are available within a single request-response cycle, and will not be shared between requests.

And though it is not part of the original request body, the intermediate contexts passed on can be seen as context based on the requests before the final response can be returned.

LinHuiqing: It does technically work though 😂 just not very sure about the correctness / conventions

src/app/modules/submission/encrypt-submission/encrypt-submission.constants.ts
LinHuiqing: Oh, with the form schema typing I can remove this constant already. Will do so in a bit!

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Its more of guarding against future middlewares that might be added and happen to use the same namespace. We want to be able to detect and guard against that so that we don't have unintended consequences in the future.

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
tshuli: thanks! sounds good to me, lets go with `req.body<...>` then :)

src/app/models/form.server.model.ts
KenLSM: I usually name them as:
- what it actually is doing `plainResponses` or `unencryptedResponses`
- which project it belongs to `encryptionBoundaryShift` (the next shift can be called `v2`) so that it can contain some domain specific keywords (i.e., `payments`, `turnstile` vs `newRecaptcha`)

src/app/modules/submission/encrypt-submission/encrypt-submission.constants.ts
LinHuiqing: Constant removed in [28ac14c](https://github.com/opengovsg/FormSG/pull/6595/commits/28ac14c6e39d962eb83e168fe8290f9b6783a53e)!

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Discussed above in [this comment](https://github.com/opengovsg/FormSG/pull/6595#discussion_r1298074277) - conclusion is to leave it as it is 👍

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Discussed above in [this comment](https://github.com/opengovsg/FormSG/pull/6595#discussion_r1298074277) - conclusion is to leave it as it is 👍

shared/types/form/form.ts
LinHuiqing: Instead of shifting it completely, I set it as optionally `never` type for email mode instead, similar to the `publicKey` prop in the schemas. Doing this instead because the types `IFormDocument` and `IPopulatedForm` used in `getFilteredResponses` and `checkNewBoundaryEnabled` respectively need to access the flag.
Done in [175f4ea](https://github.com/opengovsg/FormSG/pull/6595/commits/175f4ea3f85645beb18f823daf58c05878d63185) and resulting broken tests fixed in [e943387](https://github.com/opengovsg/FormSG/pull/6595/commits/e94338770bca4d7b457a4adab0b1340d03e8846e).

src/app/models/form.server.model.ts
LinHuiqing: For what it is doing, I think it might not be suitable in this case as email mode forms have submission payloads that look very similar, so naming it for what its doing might be a bit confusing. I'll go with `encryptionBoundaryShift` then bah
side comment: `newRecaptcha` => `newerRecaptcha` => `newestRecaptcha`? HAHA

LinHuiqing: My thought process for `newEncryptionBoundary` was so that it somewhat reads as `newEncryptionBoundary` enabled when you set it as true haha

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: As discussed in our huddle, addressed here: [e081264](https://github.com/opengovsg/FormSG/pull/6595/commits/e081264c64aa0ebb92069bba5f1a7e8c5461c91d)
Slightly different naming for the combined middleware to highlight (imo) the more important parts of what it is doing though, which is creating the `formsg` field in the request body and retrieving the form definition.
Will close subsequent related comments so we can track all discussions here.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Addressed in [e081264](https://github.com/opengovsg/FormSG/pull/6595/commits/e081264c64aa0ebb92069bba5f1a7e8c5461c91d), resolving to streamline discussions [here](https://github.com/opengovsg/FormSG/pull/6595#discussion_r1298348792).

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Addressed in [e081264](https://github.com/opengovsg/FormSG/pull/6595/commits/e081264c64aa0ebb92069bba5f1a7e8c5461c91d), resolving to streamline discussions [here](https://github.com/opengovsg/FormSG/pull/6595#discussion_r1298348792).

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Addressed in [e081264](https://github.com/opengovsg/FormSG/pull/6595/commits/e081264c64aa0ebb92069bba5f1a7e8c5461c91d), resolving to streamline discussions [here](https://github.com/opengovsg/FormSG/pull/6595#discussion_r1298348792).

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Thanks! Removed in [b5080a5](https://github.com/opengovsg/FormSG/pull/6595/commits/b5080a5278ce7edfe139962b6b9a7ba9b9402e4f)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Actually yea related question, will the line `if (!hasEnsuredAll) return` ever enter the loop? Because I'd imagine that if any of the ensures fail, it will send the response directly. If anything fails and a response is not sent, the statement implies that there will not be a response sent, so the client will just be waiting forever.

LinHuiqing:
Okay, I think I've addressed all the comments other than the one for naming of the flag! Let me know what yall think and thanks for the comments and suggestions so far! :D

shared/types/form/form.ts
```diff
@@ -163,6 +165,7 @@ export interface StorageFormBase extends FormBase {
   payments_channel: FormPaymentsChannel
   payments_field: FormPaymentsField
   business?: FormBusinessField
+  newEncryptionBoundary?: boolean
```
tshuli: Since FormBase already has `newEncryptionBoundary?: boolean`, no longer necessary to define here

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1561,10 +1563,10 @@ export const submitEmailPreview: ControllerHandler<
   }
   const form = formResult.value
 
-  const parsedResponsesResult =
-    await EmailSubmissionService.validateAttachments(responses).andThen(() =>
-      ParsedResponsesObject.parseResponses(form, responses),
-    )
+  const parsedResponsesResult = await SubmissionService.validateAttachments(
+    responses,
+    form.responseMode === FormResponseMode.Email,
```
tshuli: /nit: hmmmm i would personally prefer this to be Enum instead of boolean. there could be more than two types of responseMode in future

src/app/modules/submission/ParsedResponsesObject.class.ts
```diff
@@ -145,18 +141,7 @@ export default class ParsedResponsesObject {
...
-          // encrypt mode submissions with responses on unhidden fields
-          // TODO(#780): Remove this once submission service is separated into
-          // Email and Encrypted services
-          form.responseMode === FormResponseMode.Encrypt
-            ? 'answer' in response &&
-              typeof response.answer === 'string' &&
-              response.answer.trim() !== ''
-            : visibleFieldIds.value.has(responseId),
+        isVisible: visibleFieldIds.value.has(responseId),
         question: formField.getQuestion(),
```
tshuli, line 145: sorry the preceeding comments (L148-L159) are a bit hard to decipher 😓 . would you mind just giving a quick 1-liner on what this change does (if any) just to. be sure we don't break anything

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
tshuli: Maybe we could add a catchall res here? note that res.status will send the latest, so we should check it has not been set first

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +116,264 @@ export const validateStorageSubmissionParams = celebrate({
...
+
+/**
+ * Moves encrypted payload present in req.body to req.body.formsg.encryptedPayload.
+ * Should only be used for the old storage mode submission endpoint (/api/v3/forms/:formId/submissions/encrypt).
+ */
+export const moveEncryptedPayload: EncryptSubmissionMiddlewareHandlerType =
+  async (req, res, next) => {
+    if (req.body.formsg.encryptedPayload) {
+      return res.send(new EncryptedPayloadExistsError())
+    }
```
tshuli: Is there any way we can enforce this in typing 🤔 Returning at runtime would be too late

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -87,3 +116,264 @@ export const validateStorageSubmissionParams = celebrate({
...
+      action: 'createFormsgAndRetrieveForm',
+      ...createReqMeta(req),
+      formId,
+    }
+
+    // Step 1: Create formsg namespace in req.body
+    if (req.body.formsg) return res.send(new FormsgReqBodyExistsError())
+    else {
+      req.body.formsg = {}
+    }
```
tshuli: likewise. we can arrange to pair if necc

tshuli (commented):
thanks @LinHuiqing for working on this, lgtm except for some last nits! 
tagging @sebastianwzq for a review too for this large diff

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Done from [132021e](https://github.com/opengovsg/FormSG/pull/6595/commits/132021ee7a92d805a4482dd8a8ec8073b243ddb9) onwards!

src/app/modules/submission/ParsedResponsesObject.class.ts
LinHuiqing: Okay so the comment was added back when email and encrypted services are not separate yet, I think. So the TODO was there to remove the conditional once it has been separated, which it already has been if we track the issue #780 tagged. So this is a dangling conditional that should have been removed a while back.

Since encrypt mode (ie old submission endpoint) does not use `ParsedResponsesObject`, we can simply remove the conditional to set the `isVisible` field to what it was being set for email mode, which is `visibleFieldIds.value.has(responseId)`.

shared/types/form/form.ts
LinHuiqing: Good point, fixed typings in [b23f39b](https://github.com/opengovsg/FormSG/pull/6595/commits/b23f39be4bf5b3c70fb21b52cee10f4397438f37)!

src/app/modules/form/admin-form/admin-form.controller.ts
LinHuiqing: That's true, refactored to use enums in [64765de](https://github.com/opengovsg/FormSG/pull/6595/commits/64765de3d5d64fd3e8622fcb27fdd85574dda230), thanks!

src/app/models/form.server.model.ts
LinHuiqing: Flag renamed to `encryptionBoundaryShift` in [18e636b](https://github.com/opengovsg/FormSG/pull/6595/commits/18e636b94996008b9da29a6d6cec18af666ded91)

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
LinHuiqing: Refactored to use a catch-all res instead in [64ca3e8](https://github.com/opengovsg/FormSG/pull/6595/commits/64ca3e8db22e4940d16c752e0a9a6c5d82430fe3)!

src/app/modules/submission/ParsedResponsesObject.class.ts
tshuli: got it, thanks @LinHuiqing !

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: I don't see a way to enforce this through typing actually. Even for the `EncryptSubmissionMiddlewareHandlerType` used for `moveEncryptedPayload`, `encryptedPayload` is `undefined` before and defined after, but using the current typing does not stop stuff at compile time.

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
timotheeg: ⚠️ I'd recommend **not** injecting data in `req.body` either! `request.body` is populated from incoming data (from whatever body parser middleware we use), and there should be a clear distinction between incoming data, and server-generated, or server-retrieved data. 

Folks may have concerns on polluting the `req` object, my approach is always to create an app-specific namespace into `req`, maybe something like this:

```javascript
if (!req.formsg) req.formsg = {}

req.formsg.something = 'foobar'
```

On using `res.locals`, this link is the answer I aligned the most with (read the whole thing because there is specifically a discussion on how locals relates to rendering and is **not** the recommended approach).
https://stackoverflow.com/a/71487552/361295

expressjs in fact specifically updated their documentation on `res.locals` to make it clear that `res.locals` relates to the `res.render()` and it not meant to be the place to dump cross-middleware variables (what @LinHuiqing shared above)

Functionally all the approaches work of course, `req` and `res` are static objects available for the lifetime of a request, so you could shove stuff anywhere, but I strongly recommend keeping a good mental model of what things are:
* keep raw incoming data separate
* keep processed incoming data separate
* keep render-concern data separate
* keep request lifecycle data (cross-middleware contextual data that gets aggregated by one or more middleware) separate

I mentally model all that like this: say you retrieve the session data and the associated user from the sessionid cookie, that data is contextual properties of **the request**, not the response! Same thing if you retrieve full product data from the db on a request because you need it to make conditional decisions. Regardless of what the response will be, the product data is also a contextual property of **the request**.

LinHuiqing: Ohh I see, thanks @timotheeg for taking the time to review this and dropping your comments! 
Hm okay let me see if I can create a `req.formsg` instead. I don't think there's a precedent in the code base yet 🤔

LinHuiqing: Refactored in [7a3608e](https://github.com/opengovsg/FormSG/pull/6595/commits/7a3608e64c761427cd85b53606e4756ffa9dfd1d) so that `req.formsg` is used to pass context instead! And `req.body` is not modified throughout.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -122,14 +122,19 @@ const submitEncryptModeForm: SubmitEncryptModeFormHandlerType = async (
...
     res,
   })
 
-  if (!hasEnsuredAll) return
+  if (ensuresError) {
+    const { errorMessage, statusCode } = mapRouteError(ensuresError)
+    return res.status(statusCode).json({
+      message: errorMessage,
+    })
+  }
```
tshuli: This seems to be quite a deviation from the usual middleware pattern - as we now we expect the pipeline to return an error, and then we use that error to set the catchall.

How about if we simply do a catch all at the end, e.g.
```
 if (!hasEnsuredAll) {
    if (!res.headersSent) {
      res.status(StatusCodes.BAD_REQUEST).json({
        message: 'Your form submission could not be processed. Please try again.',
      })
    }
    return
  }
```

LinHuiqing: Ah, got it that it's not how pipelines usually behave. Although I agree that this is more defensive and safer than just having a plain `return`, I still think it's odd that we're adding code to the `if (!hasEnsuredAll)` branch even it's impossible to fulfil that condition.

Will it be better if we follow how Express middlewares handle their errors? What they do is to pass errors into the next function, then centralise error handling with an error handling middleware, as described [here](https://gist.github.com/zcaceres/2854ef613751563a3b506fabce4501fd).

But either way, I think this can be left for another PR bah! I'll revert my refactoring commit regarding ensures and add the catch-all as you suggested. We can make further changes in a separate PR if needed.

cc @KenLSM, I think you'll be interested in this discussion as well.

LinHuiqing: Previous commit reverted in [09b367b](https://github.com/opengovsg/FormSG/pull/6595/commits/09b367b6f8a7544c82d921714b424f051a91a56b), suggested catch-all response return for errors implemented in [b9cf27a](https://github.com/opengovsg/FormSG/pull/6595/commits/b9cf27a62cd4186cc99f777b37e1d4e75fd04d4c).

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
tshuli: thanks! i investigated this a bit more, indeed it does not seem possible with express middleware

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
tshuli: thanks @LinHuiqing , agree we can take this for another PR to keep this within scope

tshuli (approved):
lgtm, thanks @LinHuiqing for seeing this through!

PR #6602, opened 2023-08-07
chore: remove angular deps

justynoh:
## Problem

There are a whole bunch of random deps that exist in our package file that were used in angular. This PR removes them, now that angular is gone.

## Solution

**Breaking Changes** 


foochifa (approved):
LGTM! Runs and builds fine locally too

PR #6604, opened 2023-08-07
feat: payment security account email check

KenLSM:
## Problem

Closes FRM-1175

## Solution

**Checking logic**
Prerequisite: feature flag of `validateStripeEmailDomain` must be enabled

1. Fetch account information from stripe just before linking account
2. Run domain test on account email that domain must be from an existing agency
3. Redirect to frontend with error indication on through query_params `error_type=100001`

This check is tested on service as other controllers/services can also attempt to connect. A generic `errAsync(InvalidDomainError)` is returned and other services should wrap over it.

**Other noteworthy changes**
**Addition of shared errors**
New `shared/constants/error.ts` is introduced to provide shared errors where backend needs to notify frontend of errors through redirects + queryparams.
Shared errors are 6 digits integers so as to future proof the error namespaces. Payment module errors will have the honour to take the inaugural `100xxx` namespace since this is the first module to implement it.

**Addition of helper function**
New `feature-flags.service` helper function `getFeatureFlag()` to handle database errors gracefully.

**Breaking Changes** 

## Before & After Screenshots
When backend service detected that the email domain is not whitelisted

Screenshot 2023-08-21 at 1 48 45 AM

## Tests

**Regression test**
- [ ] form is able to connect to stripe account with feature flag off
  - [ ] Observe oauth callback from stripe redirect admin to payment settings page
  - [ ] Observe that stripe account is now connected
  - [ ] Observe that stripe account stays connected after refresh

**Feature test**
To deploy release branch to `staging-alt2` so that we can test with live stripe credentials, credentials in [comments](https://linear.app/ogp/issue/FRM-1175#comment-5d3c762c)
- [ ] set `validateStripeEmailDomain` to `true` on `collection.featureflags`
- [ ] form is unable to connect to stripe account with feature flag off
  - [ ] Observe oauth callback from stripe redirect admin to payment settings page with query params `error_type=100001`
  - [ ] Observe that stripe account is not connected
  - [ ] Observe that an alert message is shown informing the user that connecting with a non-whitelisted domain is not allowed

## Deploy Notes
Using the staging-alt2 env (as it has live Stripe credentials), enable featureFlag `validateStripeEmailDomain`, and set to `true`.
- this flag is non-critical as it will fallback to `false` if key doesn't exist


src/app/modules/feature-flags/feature-flags.service.ts
```diff
@@ -23,3 +24,49 @@ export const getEnabledFlags = (): ResultAsync<string[], DatabaseError> => {
...
+        meta: options?.logMeta
+          ? options.logMeta
+          : {
+              action: 'getFeatureFlag',
+            },
+        error,
+      })
+      return okAsync(_fallbackValue)
+    })
+}
```
LinHuiqing, line 72: I think its better to incorporate the default `false` into the function typing definition, so something like:
```suggestion
export const getFeatureFlag = (
  flag: keyof typeof featureFlags,
  options: {
    fallbackValue: boolean
    logMeta?: CustomLoggerParams['meta']
  } = { fallbackValue: false },
): ResultAsync<boolean, DatabaseError> => {
  return getEnabledFlags()
    .andThen((featureFlagsListResult) =>
      okAsync(featureFlagsListResult.includes(flag)),
    )
    .orElse((error) => {
      logger.error({
        message: 'Error occurred whilst retrieving enabled feature flags',
        meta: options?.logMeta
          ? options.logMeta
          : {
              action: 'getFeatureFlag',
            },
        error,
      })
      return okAsync(options.fallbackValue)
    })
}
```

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
```diff
@@ -84,19 +110,19 @@ const BeforeConnectionInstructions = ({
...
+    <>
+      <InlineMessage variant="info" my="2rem">
+        <Text>
+          You are currently in test mode. You can choose to skip connecting a
+          Stripe account after clicking the button below.
+        </Text>
+      </InlineMessage>
+      <StripeConnectButton connectState={StripeConnectButtonStates.ENABLED} />
+    </>
+  )
```
LinHuiqing, line 125: Is there a reason why removing the `else` is preferred over a flow that's completely wrapped with if-else conditionals like the following?
```ts
if (isInvalidDomain) {
  return ...
} else if (isProduction) {
  return ...
} else {
  return ...
}
```

src/app/modules/feature-flags/feature-flags.service.ts
```diff
@@ -23,3 +24,49 @@ export const getEnabledFlags = (): ResultAsync<string[], DatabaseError> => {
...
+ *  logMeta,
+ * })
+ *.andThen((shouldValidateStripeEmailDomain) => ...)
+ *```
+ *
+ * @param flag
+ * @param options.fallbackValue the value to fall back to in the event that there's an error retrieving the feature flag
+ * @returns boolean that represents the status of the feature flag or the fallback value
+ */
+export const getFeatureFlag = (
```
LinHuiqing, line 48: Can be in a separate PR, but we should test this!

LinHuiqing: Or not, since we will be trying out feature flag platforms like LaunchDarkly / Growthbook 🤔

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -689,7 +692,11 @@ describe('stripe.service', () => {
     })
   })
   describe('linkStripeAccountToForm', () => {
-    it('should call func to attach payment account information', async () => {
+    beforeEach(() => {
+      jest.restoreAllMocks()
+    })
+
+    it('should attach payment account information for new accounts', async () => {
```
LinHuiqing, line 699: Add a test case for when the feature flag has not been enabled - stripe account should still be linked successfully.

LinHuiqing (commented):
Generally lgtm, just some nits and a test to add.

LinHuiqing (changes requested):
⚠️ Breaks test mode if `validateStripeEmailDomain` flag is enabled! At least it throws "StripeAccountError: Stripe account email is missing" for me when I run with test mode credentials when the flag is on.
To think about - What do we want the behaviour to be for users on UAT testing out payments? Do we want to make it impossible for them to connect if they use the "Skip this form"? Currently an error is thrown ("Error retriving Stripe account email"). Or do we want them to be able to connect since it is just a test account?

KenLSM:
> ⚠️ Breaks test mode if `validateStripeEmailDomain` flag is enabled! At least it throws "StripeAccountError: Stripe account email is missing" for me when I run with test mode credentials when the flag is on. To think about - What do we want the behaviour to be for users on UAT testing out payments? Do we want to make it impossible for them to connect if they use the "Skip this form"? Currently an error is thrown ("Error retriving Stripe account email"). Or do we want them to be able to connect since it is just a test account?

Ah yes. I've noted it on the [linear](https://linear.app/ogp/issue/FRM-1175#comment-f560108e) issue. Thanks for reminding that should be surfaced up over here for the reviewer, my bad, missed that.

This issue is partially why I've pushed to add a feature flag on this.

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
KenLSM: [I'm](https://kenlsm.com/coding/an-if-without-else/) a [strong](https://softwareengineering.stackexchange.com/a/157426) [advocate](https://stackoverflow.com/a/46875482) of using [guard](https://www.codementor.io/@clintwinter/use-guard-clauses-for-cleaner-code-1rrsczgwxp) [clauses](https://en.wikipedia.org/wiki/Guard_(computer_science)).

Using of `else` provides a path for the next developer to add nested conditionals, which from my experience an invitation to make the codebase unreadable.

LinHuiqing: Okays makes sense! Thanks thanks!

LinHuiqing:
> Ah yes. I've noted it on the [linear](https://linear.app/ogp/issue/FRM-1175#comment-f560108e) issue. Thanks for reminding that should be surfaced up over here for the reviewer, my bad, missed that.
> 
> This issue is partially why I've pushed to add a feature flag on this.

Okays understood. In this case, could you add in the deploy notes to deploy the release branch on staging-alt2 so that there are live stripe credentials to test with?

src/app/modules/feature-flags/feature-flags.service.ts
KenLSM: That changes the option argument in two ways.
1. The default `fallbackValue` will not be set if `options` was passed in from the caller. 
2. `fallbackValue` becomes a required value if I want to override the `logMeta` which may not be the case.

```ts
const options = { logMeta: { foo: "bar" }} 
getFeatureFlag('flag', options) // ts-error fallbackValue is required in "options"
```

LinHuiqing: I see, is there any other way to type `false` as a default value without these side effects then 🤔

KenLSM: Sorry for getting back late, was occupied by other urgent tasks.

I've seen different ways of providing a default but there's no clear winner, and each have their own pros and cons.

1. The way that you're seeing now
2. `Object.assign`

src/app/modules/feature-flags/feature-flags.service.ts
KenLSM: Yeah i think it's better that we look towards those tools.

src/app/modules/payments/__tests__/stripe.service.spec.ts
KenLSM: It's covered in https://github.com/opengovsg/FormSG/pull/6604/files#diff-2925cc24c9a1ab9f17fa6f5ee6f8873de02f8f9bc2900e5d01ff9f1c8210566cR823

KenLSM: (just noticed a typo on the name)

src/app/modules/feature-flags/feature-flags.service.ts
LinHuiqing: No worries, okays let's keep it as it is then, thanks Ken!

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -820,7 +820,7 @@ describe('stripe.service', () => {
       expect(result._unsafeUnwrapErr()).toBeInstanceOf(InvalidDomainError)
     })
 
-    it('should not check email whitelsiting if feature flag is disabled', async () => {
+    it('should not check email whitelistingw if feature flag is disabled', async () => {
```
LinHuiqing, line 823: @KenLSM it changed from one typo to another haha 😂

src/app/modules/payments/__tests__/stripe.service.spec.ts
```diff
@@ -738,7 +776,82 @@ describe('stripe.service', () => {
...
+      // Assert
+      expect(getFeatureFlagSpy).toHaveBeenCalledTimes(1)
+      expect(stripeAccountsRetrieveApiSpy).toHaveBeenCalledTimes(1)
+      expect(authServiceSpy).toHaveBeenCalledTimes(1)
+      expect(addPaymentAccountIdSpy).toHaveBeenCalledTimes(0)
+      expect(result.isErr()).toBeTrue()
+      expect(result._unsafeUnwrapErr()).toBeInstanceOf(InvalidDomainError)
+    })
+
+    it('should not check email whitelistingw if feature flag is disabled', async () => {
```
LinHuiqing:
```suggestion
    it('should not check email whitelisting if feature flag is disabled', async () => {
```

LinHuiqing: it changed from one typo to another haha 😂

src/app/modules/feature-flags/feature-flags.service.ts
LinHuiqing: Yea agreed! :)

LinHuiqing:
> Okays understood. In this case, could you add in the deploy notes to deploy the release branch on staging-alt2 so that there are live stripe credentials to test with?

I've added this to deploy notes btw!

LinHuiqing (approved):
lgtm! merge when ready after the last typo fix

PR #6608, opened 2023-08-08
feat: platform api v1 - get and update webhook settings

wanlingt:
## Problem

Closes FRM-1244, FRM-1242, FRM-1243

## Solution

This PR implements 
- API access for admins that are also "platforms" - i.e. they perform actions on behalf of other Form admins. Examples include Plumber and EngageSG. This is represented by the new `isPlatform` boolean property in the `apiToken` object. 

Ideally, authorisation would be via OAuth to allow for more fine-grained permissions, but for the purposes of the MVP, we are going with this simplified approach. Inspiration was taken from [GoGovSG's admin API](https://github.com/opengovsg/GoGovSG/pull/2213) which allows FormSG to create shortlinks on behalf of GoGovSG admins.
 

- In addition to authenticating the API key (shared with `/api/public/v1`), the /public/v1 route uses a middleware check `isPlatformApiUser` in `auth.middlewares.ts`. If the API user is a platform, and has provided a userEmail, it looks for the user associated with the userEmail and sets `req.session.user` as that user

New platform API endpoints:
- `POST /api/public/v1/admin/forms/:formId/webhooksettings` to retrieve webhook settings. 
- `PATCH /api/public/v1/admin/forms/:formId/webhooksettings` to update webhook settings

More details can be found in the [APIs design doc](https://www.notion.so/opengov/FormSG-External-APIs-Design-doc-8c41b23ceb184baebd97558cb10e1454?pvs=4#3927b7340a144b918a4e79fb020865f8) (WIP)

**Breaking Changes** 

**Future Improvements (TODOs)**:
- Add unit tests FRM-914
- Rate limit per API key FRM-948

## Tests

`POST /api/public/v1/admin/forms/:formId/webhooksettings` to retrieve webhook settings
```
curl --request POST \
  --url https://staging.form.gov.sg/api/public/v1/admin/forms/:formId/settings \
  --header 'Authorization: Bearer test_v1_APITOKENHERE' \
  --header 'Content-Type: application/json' \
  --data '{"userEmail": "adminOfFormId"}'
```
- [ ] Test that API call works - the webhook settings retrieved should be what's in the DB. The userEmail provided should be an admin of the formId in the URL
- [ ] Use an invalid API token. You should get a 401 Unauthorised - Invalid API Key message
- [ ] Use a user email that is a FormSG user, but isn't the admin of the form ID provided. You should get a 403 Forbidden message
- [ ] Use an invalid form ID. You should get a 404 not found error
- [ ] Use an invalid request body e.g. misspell userEmail to useEmail. You should get a 400 Bad Request validation error

`PATCH /api/public/v1/admin/forms/:formId/webhooksettings` to update webhook settings
```
curl --request PATCH \
  --url https://staging.form.gov.sg/api/public/v1/admin/forms/:formId/settings \
  --header 'Authorization: Bearer test_v1_APITOKENHERE' \
  --header 'Content-Type: application/json' \
  --data '{"userEmail": "adminOfFormId",
"webhook": {
		"url":"https://www.webhookUrl.com",
	"isRetryEnabled": true
}}'
```
- [ ] Test that API call successfully updates the webhook settings. The userEmail provided should be an admin of the formId in the URL
- [ ] Use an invalid API token. You should get a 401 Unauthorised - Invalid API Key message
- [ ] Use a user email that is a FormSG user, but isn't the admin of the form ID provided. You should get a 403 Forbidden message
- [ ] Use an invalid form ID. You should get a 404 not found error
- [ ] Use an invalid request body e.g. misspell userEmail to useEmail. You should get a 400 Bad Request validation error

## Deploy Notes
- `PLATFORM_API_RATE_LIMIT`: Per-minute, per-IP, per-instance request limit for platform APIs


src/app/config/schema.ts
```diff
@@ -319,6 +319,12 @@ export const optionalVarsSchema: Schema<IOptionalVarsSchema> = {
       default: 100,
       env: 'PUBLIC_API_RATE_LIMIT',
     },
+    platformApi: {
+      doc: 'Per-minute, per-IP, per-instance request limit for platform APIs',
```
tshuli, line 323: not within scope of this PR's code, but do consider setting overall CF rate limit for platform api endpoint too

src/app/routes/api/api.routes.ts
```diff
@@ -2,11 +2,13 @@ import { Router } from 'express'
...
 
+import { V1PlatformRouter } from './platform/v1'
 import { V1PublicRouter } from './public/v1'
 import { V3Router } from './v3'
 
 export const ApiRouter = Router()
 
 ApiRouter.use('/v3', V3Router)
 ApiRouter.use('/public/v1', V1PublicRouter)
+ApiRouter.use('/platform/v1', V1PlatformRouter)
```
tshuli: a quick clarification please - why is the new platform router not nested within the /v3 routes?
- is V1PublicRouter still in use or is this for backward compatibility?

src/app/routes/api/platform/v1/admin/forms/admin-forms.platform.routes.ts
```diff
@@ -0,0 +1,53 @@
...
+AdminFormsPlatformRouter.route('/:formId([a-fA-F0-9]{24})/webhooksettings')
+  /**
+   * Retrieve the webhook and response mode settings of the specified form
+   * @security bearer
+   * @route GET /admin/forms/:formId/settings
+   * @param body the user email
+   *
+   * @returns 200 with latest form settings
+   * @returns 401 when current user is not logged in
+   * @returns 403 when current user does not have permissions to obtain form settings
```
tshuli: Qn - when would these error codes apply for a platform user?

src/app/routes/api/platform/v1/admin/forms/admin-forms.platform.routes.ts
```diff
@@ -0,0 +1,53 @@
...
+   * Update form settings according to given subset of settings.
+   * @route PATCH /admin/forms/:formId/webhooksettings
+   * @group admin
+   * @param body the subset of settings to patch and the user email
+   * @produces application/json
+   * @consumes application/json
+   * @returns 200 with latest form settings on successful update
+   * @returns 400 when given body fails Joi validation
+   * @returns 401 when current user is not logged in
+   * @returns 403 when current user does not have permissions to update form settings
```
tshuli: likewise here

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1417,6 +1486,52 @@ export const handleGetSettings: ControllerHandler<
...
 }
 
+/**
+ * Handler for GET api/platform/v1/admin/forms/:formId/webhookSettings.
+ *
+ * @returns 200 with latest webhook and response mode settings
+ * @returns 401 when current user is not logged in
+ * @returns 403 when current user does not have permissions to obtain form settings
+ * @returns 404 when form to retrieve settings for cannot be found
+ * @returns 409 when saving form settings incurs a conflict in the database
```
tshuli: for GET request, there should not be 409 (no saving form settings) 😅

src/app/routes/api/platform/v1/admin/forms/admin-forms.platform.routes.ts
tshuli: ^ only the 401, actually. I realised 403 applies if the user listed in the userEmail doesn't have write permissions for that specific form

src/app/modules/form/admin-form/admin-form.middlewares.ts
```diff
@@ -38,3 +41,13 @@ export const updateSettingsValidator = celebrate({
     .min(1)
     .custom((value, helpers) => verifyValidUnicodeString(value, helpers)),
 })
+
+/**
+ * Joi validator for PATCH api/platform/v1/admin/forms/:formId/webhookSettings route.
+ */
+export const updateWebhookSettingsValidator = celebrate({
+  [Segments.BODY]: Joi.object<WebhookSettingsUpdateDto>({
+    userEmail: Joi.string(),
```
tshuli: suggestion: `Joi.string().email().required()` so that joi checks that it is valid email address

shared/types/user.ts
```diff
@@ -29,6 +29,7 @@ export const UserBase = z.object({
       keyHash: z.string(),
       createdAt: z.date(),
       lastUsedAt: z.date().optional(),
+      isPlatform: z.string().optional(),
```
tshuli: just a thought: if platform users are also a kind of formsg user, will this allow platforms to modify each other's user settings in future (e.g. contact number)? I was wondering if platform tokens could be issued without creating a user account for the platform

not a super major point, just food for thought

tshuli (commented):
thanks @wanlingt ! PR looks good with just some minor clarifications please 😃 will re-review again thereafter

#for future PR - the platform endpoint sounds like it could potentially give platforms too much power. I wonder if we need to build in additional safeguards against key loss or MITM attacks. e.g. require platform users to sign their requests, and host their signing keys on a well known endpoint for us to verify.

src/app/config/schema.ts
tshuli: Ok I see it is in [FRM-948](https://linear.app/ogp/issue/FRM-948). can just set CF rate limit 😆

src/app/routes/api/platform/v1/admin/forms/admin-forms.platform.routes.ts
wanlingt: thanks @tshuli ! I've refined the description in [9cdd65b](https://github.com/opengovsg/FormSG/pull/6608/commits/9cdd65b29676ff8e77cd2edbe08c7d36c5b3af03), 401 is returned when the user doesn't provide a valid API key, or is not a platform user

src/app/routes/api/platform/v1/admin/forms/admin-forms.platform.routes.ts
wanlingt: Improved the description in [9cdd65b](https://github.com/opengovsg/FormSG/pull/6608/commits/9cdd65b29676ff8e77cd2edbe08c7d36c5b3af03)!

 returns 401 when current user does not provide a valid API key, or is not a platform user
 returns 403 when user email does not have permissions to update form settings

src/app/modules/form/admin-form/admin-form.controller.ts
wanlingt: heh thanks, deleted in [9cdd65b](https://github.com/opengovsg/FormSG/pull/6608/commits/9cdd65b29676ff8e77cd2edbe08c7d36c5b3af03)

src/app/modules/form/admin-form/admin-form.middlewares.ts
wanlingt: [5898681](https://github.com/opengovsg/FormSG/pull/6608/commits/58986816d7e21e534c3bc352948c5a22daee127e)

src/app/config/schema.ts
wanlingt: great point! i've updated the description in FRM-948 to include setting CF rate limit

src/app/routes/api/api.routes.ts
wanlingt: The new platform router is separate from the internal `api/v3/` routes, mainly because of version control and for the separate authentication mechanism.

V1PublicRouter is for the public APIs (non-platforms - means that admins are only allowed to carry out actions on forms that they own) - [here's](https://github.com/opengovsg/FormSG/pull/6376) the original PR. This route is not being used currently as we've yet to publicise it to admins (nor given anyone any API tokens).

V1PlatformRouter is for "platform" users - which gives them permission to carry out actions on other form admins' forms.

Hope that clarifies!

shared/types/user.ts
wanlingt: > if platform users are also a kind of formsg user, will this allow platforms to modify each other's user settings in future (e.g. contact number)? 

Yes can, if we create an endpoint which allows them to modify other users' contact number settings. We currently only have get and update webhook settings (from this PR), but we might add more endpoints in the future depending on the need. This is potentially giving them a lot of power though, so might have to think about permissions as well. I'm intentionally ignoring that right now as this is just an MVP, but recognise that this is a possible future extension!

> I was wondering if platform tokens could be issued without creating a user account for the platform

And yes valid point too, worth considering hm

wanlingt:
> thanks @wanlingt ! PR looks good with just some minor clarifications please 😃 will re-review again thereafter
> 
> #for future PR - the platform endpoint sounds like it could potentially give platforms too much power. I wonder if we need to build in additional safeguards against key loss or MITM attacks. e.g. require platform users to sign their requests, and host their signing keys on a well known endpoint for us to verify.

thanks for the review @tshuli , made the changes. 

I think the additional safeguards is a great point. The first phase is making this available to OGP-only admins, I think we should definitely make this mechanism more secure before releasing this to WOG form admins (e.g. CPF). Might arrange a session with Eugene to better understand the safeguards we can have, will list your suggestion too

src/app/routes/api/api.routes.ts
tshuli: got it, thanks for clarifying! 😃

tshuli (approved):
thanks @wanlingt for addressing the comments, lgtm!

tshuli:
> > thanks @wanlingt ! PR looks good with just some minor clarifications please 😃 will re-review again thereafter
> > #for future PR - the platform endpoint sounds like it could potentially give platforms too much power. I wonder if we need to build in additional safeguards against key loss or MITM attacks. e.g. require platform users to sign their requests, and host their signing keys on a well known endpoint for us to verify.
> 
> thanks for the review @tshuli , made the changes.
> 
> I think the additional safeguards is a great point. The first phase is making this available to OGP-only admins, I think we should definitely make this mechanism more secure before releasing this to WOG form admins (e.g. CPF). Might arrange a session with Eugene to better understand the safeguards we can have, will list your suggestion too

Additional suggestion for future PR - perhaps you could also consider a flag to restrict platforms to managing accounts under specified domains, eg. CPFB platform account can only modify CPF forms. For OGP's platforms, they can modify any form.

shared/types/user.ts
ghost: Should `isPlatform` be a boolean?

wanlingt: yes indeed! thanks for spotting it, fixed in [e043280](https://github.com/opengovsg/FormSG/pull/6608/commits/e043280fbb02947f38f6fdc96136d6778926677d)

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1417,6 +1485,51 @@ export const handleGetSettings: ControllerHandler<
     })
 }
 
+/**
+ * Handler for GET api/platform/v1/admin/forms/:formId/webhookSettings.
```
ghost: Is there an existing endpoint for non-platform calls like `GET api/v1/admin/forms/:formId/webhookSettings.` ? Given that whether a user is an admin or not is abstracted from the caller, I wonder if it needs to be in the path. 

In future, if we want to expose an originally platform api to all users, it could be done by modifying the logic internally, instead of creating a separate endpoint for non platform users.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1315,6 +1364,26 @@ export const handleUpdateSettings = [
   _handleUpdateSettings,
 ] as ControllerHandler[]
 
+/**
+ * Handler for PATCH api/platform/v1/admin/forms/:formId/webhooksettings.
```
ghost: same comment as below

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1295,6 +1301,49 @@ export const _handleUpdateSettings: ControllerHandler<
...
 }
 
+export const _handleUpdateWebhookSettings: ControllerHandler<
+  { formId: string },
+  FormWebhookSettings | ErrorDto,
+  WebhookSettingsUpdateDto & { userEmail: string }
+> = (req, res) => {
+  const { formId } = req.params
+  const { userEmail } = req.body
+  const settingsToPatch = req.body
```
ghost: qns: still familiarising myself with typescript. settingsToPatch will include the userEmail, am I right? Can we use it to call `AdminFormService.updateFormSettings(retrievedForm, settingsToPatch) ` directly?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1417,6 +1485,51 @@ export const handleGetSettings: ControllerHandler<
...
+/**
+ * Handler for GET api/platform/v1/admin/forms/:formId/webhookSettings.
+ *
+ * @returns 200 with latest webhook and response mode settings
+ * @returns 401 when current user is not logged in
+ * @returns 403 when current user does not have permissions to obtain form settings
+ * @returns 404 when form to retrieve settings for cannot be found
+ * @returns 500 when database error occurs
+ */
+export const handleGetWebhookSettings: ControllerHandler<
```
ghost: Should we have a validator like the above API?

src/app/modules/form/admin-form/admin-form.controller.ts
wanlingt: > In future, if we want to expose an originally platform api to all users, it could be done by modifying the logic internally, instead of creating a separate endpoint for non platform users.

Yeah great point! After the discussion with the team last Friday, I think it makes sense to have just one endpoint. Will be modifying this PR accordingly and request for a review again after

wanlingt: > Is there an existing endpoint for non-platform calls like GET api/v1/admin/forms/:formId/webhookSettings. ? Given that whether a user is an admin or not is abstracted from the caller, I wonder if it needs to be in the path.

There isn't, it's just `/admin/forms/:formId/settings`. After thinking about it further + discussing with @LinHuiqing and @foochifa , I think we can just use that (not restricting the set of settings returned and updated), since we're using the same endpoint for platform and non-platform users. 

The initial consideration (about how limiting the endpoint makes this more secure) is moot because
- platform and non-platform users have the same access
- if they're API users, we can assume that they're fairly tech savvy, the likelihood that they mistakenly change the wrong setting is low.

I will filter out some of the settings fields that we don't want any admin to be able to edit, like `payments_channel` and `publicKey`.

does that make sense to you?

PR #6619, opened 2023-08-15
fix: payment date race condition

KenLSM:
## Problem

After a responder completes a payment, they would be shown a Payment summary which contains details such as payment products, and payment completion date.

However, the payment completion date has a race condition resulting in the completion date to be empty. 
image

## Solution

Move payment completion data over to `api/v3/payments/:formId/:paymentId/receipt/status`.

This works because:
1.  the whole `completedPayment` object is populated at once, this includes `receiptUrl` and `paymentDate`
2. Payment summary is not rendered until `completedPayment.receiptUrl` is populated

With the above two facts, we can be sure that `paymentDate` will be available when `<PaymentSummary >` is rendered .

https://github.com/opengovsg/FormSG/assets/12391617/f3004419-2d25-42aa-b1e7-4a8aafacfbd3

**Breaking Changes** 


wanlingt (approved):
lgtm thanks ken!

PR #6624, opened 2023-08-15
fix: skip qty validation when multiqty is disabled

KenLSM:
## Problem

1. A quantity validation check was performed preventing the field to be saved as the quantity field errors out, but this is not visible to the user.
This should be fixed such that the quantity check is not performed when the quantity limit toggle is off.
2. On the backend,`multi_qty` is always treated to be true even if the toggle is disabled, resulting in validation failure when the user is trying to add only 1 single item.

Closes FRM-1255

## Solution

1. Quantity validation will do an early return with true when quantity limit toggle is off. 
2. A `trigger` is also added so that we can update the `useForm` errors through a validation.
3. Backend checks if `multi_qty` is enabled and set the qty modifier to be 1 when disabled during calculations.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

https://github.com/opengovsg/FormSG/assets/12391617/f160b752-98ca-4086-80f1-f59d45173476

**AFTER**:

https://github.com/opengovsg/FormSG/assets/12391617/6695c5bf-6965-45e3-983b-15c54fdd3130

## Tests

Quantity below max payment amount (200k)
- [ ] Save button should be enabled when amount is 20k, and toggle disabled
- [ ] Enable toggle and set max quantity as `9`
- [ ] Save button remains enabled
- [ ] Saving items succeeds and is reflected on payment input panel with amount as `$20000.00`

Quantity at max payment amount (200k)
- [ ] Save button should be enabled when amount is 20k, and toggle disabled
- [ ] Enable toggle and set max quantity as `10`
- [ ] Save button remains enabled
- [ ] Saving items succeeds and is reflected on payment input panel with amount as `$20000.00`

Quantity above max payment amount (200k)
- [ ] Save button should be enabled when amount is 20k, and toggle disabled
- [ ] Enable toggle and set max quantity as `11`
- [ ] Save button is now **disabled**
- [ ] Saving button should not be interactable
- [ ] Disable toggle
- [ ] Saving button is now **enabled**
- [ ] Saving items succeeds and is reflected on payment input panel with amount as `$20000.00`

Regression
- [ ] Add new products
- [ ] Enabling toggle doesn't crash the modal
  - [ ] Shows error `Please enter a valid payment amount`
- [ ] Disabling toggle doesn't crash the modal 
  - [ ] Continue to show error `Please enter a valid payment amount`

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.payments.service.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.payments.service.spec.ts
@@ -387,7 +387,7 @@ describe('admin-form.payment.service', () => {
         })
         const updatedProducts = [
           {
-            multi_qty: true,
+            multi_qty: false,
             max_qty: 11,
             amount_cents: 10,
           },
@@ -406,7 +406,36 @@ describe('admin-form.payment.service', () => {
             updatedProducts,
           )
 
-        expect(putSpy).not.toHaveBeenCalled()
+        expect(putSpy).toHaveBeenCalledOnce()
+        expect(actualResult.isOk()).toBeTrue()
+      })
+      it('should allow updates when product * max qty is below max amount', async () => {
+        jest.replaceProperty(PaymentConfig, 'paymentConfig', {
+          ...PaymentConfig.paymentConfig,
+          maxPaymentAmountCents: 100,
+        })
+        const updatedProducts = [
+          {
+            multi_qty: false,
+            max_qty: 9,
+            amount_cents: 10,
+          },
+        ] as PaymentsProductUpdateDto
+
+        const putSpy = jest
+          .spyOn(EncryptFormModel, 'updatePaymentsProductById')
+          .mockResolvedValueOnce({
+            payments_field: {},
+          } as unknown as IEncryptedFormDocument)
+
+        // Act
+        const actualResult =
+          await AdminFormPaymentService.updatePaymentsProduct(
+            mockFormId,
+            updatedProducts,
+          )
+
+        expect(putSpy).toHaveBeenCalledOnce()
         expect(actualResult.isOk()).toBeTrue()
       })
     })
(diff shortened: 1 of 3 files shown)


src/app/modules/form/admin-form/__tests__/admin-form.payments.service.spec.ts
```diff
@@ -254,19 +258,157 @@ describe('admin-form.payment.service', () => {
...
+        jest.clearAllMocks()
+      })
+      it('should allow updates when product * max qty possibly exceeds max amount', async () => {
+        jest.replaceProperty(PaymentConfig, 'paymentConfig', {
+          ...PaymentConfig.paymentConfig,
+          maxPaymentAmountCents: 100,
+        })
+        const updatedProducts = [
+          {
+            multi_qty: true,
```
LinHuiqing: this should be `false` instead of `true` haha I think that's why the tests are failing too
```suggestion
            multi_qty: false,
```

KenLSM: Great catch! Added in 6a477304f72877600a12753f5417d70ba6a990f0

LinHuiqing:
2 bugs:
1. There's now an error complaining that we're trying to `.trim` on `undefined` when we just use the toggle without keying in anything:

https://github.com/opengovsg/FormSG/assets/37061143/08a4683f-8db3-4b3c-95d0-378b0d283b53

2. We need to set `multi_qty` as `false` when we don't set it, if not there'll be a joi validation error:

https://github.com/opengovsg/FormSG/assets/37061143/caa64b28-a416-4c2d-81fa-bc2efed74f58

LinHuiqing (approved):
Lgtm! Spotted 2 more payment bugs but I think they can be fixed in separate PRs (FRM-1256 and FRM-1257)

PR #6625, opened 2023-08-15
fix: product item button check for payment is not enabled

KenLSM:
## Problem

When the payment form has it's stripe account disconnected, the buttons for product items are still interactable, and enabled.

Closes FRM-1254

## Solution

Passing`isDisabled` into `IconButton`

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-08-15 at 6 55 41 PM

**AFTER**:

Screenshot 2023-08-15 at 6 55 24 PM

## Tests

Buttons are disabled and non-interactable
- [ ] As an admin, create a product payment form with at least 1 product
- [ ] Disconnect stripe account
- [ ] Observe that the buttons on `PaymentInputPanel` are now faded out (disabled variant)
- [ ] Buttons are also not interactable, it doesn't open modal or gets deleted


wanlingt (approved):
lgtm! we should probably do some refactoring in the future, the isDisabled value is being passed through many components haha

PR #6630, opened 2023-08-16
chore: reduce toast duration to 4s

foochifa:
## Problem

Discussed with @staceytan1998 and we think that the current toast duration is too long. From quick research, it seems the best duration is between 2-5s, with a preference for 3s duration. However, decided to keep it at 4s as we have some decently long toasts.

## Solution

Change `useToast` duration to `4000`

**Breaking Changes** 

## Tests

- [ ] do any action that results in a toast (Create/delete form, create/update/delete fields)
- [ ] check that the toast duration is ~4s


LinHuiqing (approved):
lgtm and tested!

PR #6631, opened 2023-08-16
feat: add logging to capture DNS resolution error

wanlingt:
## Problem

We need to capture the error message that's thrown by `dns.resolve` in `validateWebhookUrl`

## Solution

Add logging

**Breaking Changes** 

## Tests

- [ ] Update a storage mode form with a webhook url that will fail DNS resolution
- [ ] Submit a form
- [ ] Check that error records appear in CloudWatch for the following query: 
`fields @timestamp, @message, @logStream, @log`
`| filter message like 'Webhook URL failed validation' and meta.action like 'validateWebhookUrl'`

### DIFF
--- a/src/app/modules/webhook/webhook.validation.ts
+++ b/src/app/modules/webhook/webhook.validation.ts
@@ -3,8 +3,11 @@ import ip from 'ip'
 
 import { isValidHttpsUrl } from '../../../../shared/utils/url-validation'
 import config from '../../config/config'
+import { createLoggerWithLabel } from '../../config/logger'
 
 import { WebhookValidationError } from './webhook.errors'
+
+const logger = createLoggerWithLabel(module)
 
 /**
  * Checks that a URL is valid for use in webhooks.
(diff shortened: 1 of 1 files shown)


tshuli (approved):
lgtm!

LinHuiqing (commented):
Could we pass in an error code or something into the `WebhookValidationError` error message so that the admins know what went wrong? So they won't have to contact us to debug.

wanlingt:
> Could we pass in an error code or something into the `WebhookValidationError` error message so that the admins know what went wrong? So they won't have to contact us to debug.

Actually we already validate the webhook URL in the Settings page when the form settings are updated so admins should be aware - I'm not sure how this admin was able to update their webhook to an invalid one

e.g. 
image

LinHuiqing:
> Actually we already validate the webhook URL in the Settings page when the form settings are updated so admins should already be aware - I'm not sure how this admin was able to update their webhook to an invalid one

Yea we do, but for this particular case, the webhook validation sometimes passes and sometimes fails. Like prior to their report, they have been sending to the same URL, and the webhook after the 1 problematic one eventually succeed as well. Okay but fair point in the case cos at the point when the webhook is attempted, the errors will only show up on our internal logs anyway and isn't shown to users). This should be sufficient then 👍 thanks @wanlingt!

PR #6633, opened 2023-08-16
fix: default payment type to products

foochifa:
## Problem

Closes FRM-1257

## Solution

Change `paymentDeleteDefault` to `PaymentType.Products`

**Breaking Changes** 

## Tests

- [ ] Create a new payment field
- [ ] Delete the payment field
- [ ] Go to create tab payment, the default payment type should be products (and you should not see the fixed payment option)


LinHuiqing (approved):
lgtm, thanks Chi Fa!

PR #6640, opened 2023-08-18
fix: replace "whitelist" wording with "allowlist"

bentindall:
## Problem
The use of the terms "whitelist" and "blacklist" can be considered harmful, so "allowlist" and "blocklist" should be used as much as possible.

## Solution
Updated UI, code and docs to say "allowlist" instead of "whitelist"

**Breaking Changes** 


KenLSM:
Thanks @bentindall for starting this initiative, but our team is currently not ready to take this change. We'll open up a contributor ticket when we're ready.

PR #6641, opened 2023-08-20
feat: add formId for deprecatedCheck logs

LinHuiqing:
## Problem

`isPreview` was removed from the joi validation of the submission endpoints in #6587 as it has been deprecated for a while. However, this caused issues for some users who have been performing programmatic submissions, as reported in [this Zendesk ticket](https://formsg.zendesk.com/agent/tickets/46418), as they were still sending `isPreview` as one of the keys.

Based on a [query of our cloudwatch logs](https://opengovproducts.slack.com/files/U02DD6WKFEH/F05NF0CBUJF/screenshot_2023-08-18_at_7.55.01_pm.png), we can see that there have not been any storage mode submissions performed with `isPreview` for the past 4 months. As such, it is safe to remove the deprecation check for storage mode forms.

In the same cloudwatch logs query, it can be seen that there have been email mode storage mode submissions with the `isPreview` key, but since the form ids were not logged, we are unable to track which forms are performing such submissions. As such, we will want to add logging for form ids as well.

tl;dr: we do 2 things in this PR:
1. remove `deprecatedCheck` logging for storage mode submissions
2. add `formId` when logging `isPreview` `deprecatedCheck`s

After which, we will want to reach out to form admins of forms that are still making submissions with the `isPreview` key and inform them of the deprecation.

## Solution

**Breaking Changes** 

**Improvements**:

- fix: rm storage isPreview deprecatedCheck

**Bug Fixes**:

- fix: rm unused EncryptSubmissionBodyAfterProcess

### DIFF
--- a/src/app/modules/submission/email-submission/email-submission.controller.ts
+++ b/src/app/modules/submission/email-submission/email-submission.controller.ts
@@ -62,6 +62,7 @@ const submitEmailModeForm: ControllerHandler<
       meta: {
         action: 'submitEmailModeForm',
         type: 'deprecatedCheck',
+        formId,
       },
     })
   }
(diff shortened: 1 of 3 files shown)


foochifa (approved):
Thanks for taking the initiative to check for this @LinHuiqing lgtm! Also thanks for removing the unused type

PR #6642, opened 2023-08-20
feat: virus scanner for storage mode submissions

tshuli:
## Problem
- This PR implements a virus scanner which will be used to scan storage mode (and in future, email mode) attachments.
- Builds off the care360 implementation, but with a fair amount of rework as the scanning logic is different (e.g. submission blocking for FormSG vs async scanning, use of a quarantine and clean bucket) and adapted for FormSG CI/CD + infra

## Solution
- ClamAV virus scanner packaged in a docker container and deployed to lambda
- [Provision concurrency](https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/#provisioned-concurrency) so at least 1 (5 for prod) lambda function is always hot
- Scanner image is rebuilt daily via cron to maintain updated virus definitions

For integration with app, the following flow should be implemented:
- Frontend should call backend endpoint requesting for submission presign links and with file size specified
- For each attachment field, backend should return 1 pre signed S3 bucket upload link to `VIRUS_SCANNER_QUARANTINE_S3_BUCKET`, with randomised key and with filesize fixed
- Frontend should upload each attachment to the `VIRUS_SCANNER_QUARANTINE_S3_BUCKET` bucket directly  using the pre-signed link
- Thereafter, when upload is complete, frontend should send the submission to the backend, but with the upload keys instead of the actual attachments
- Backend should call the virus scanner to scan the files in `VIRUS_SCANNER_QUARANTINE_S3_BUCKET` matching these keys
- If the files are clean, the virus scanner will transfer the scanned files to `VIRUS_SCANNER_CLEAN_S3_BUCKET`, with another randomised key. The Virus scanner will then return the key to the backend
- Backend should retrieve the scanned files from `VIRUS_SCANNER_QUARANTINE_S3_BUCKET` and then encrypt the submission.

**Other technical details:**
- Some optimisation using streams to reduce virus scan time (the scanner scans the stream as it retrieves the file from s3).

## Tests

- [ ] Put clean file with valid uuid in quarantine bucket and trigger lambda with `key: <filename>`.
  - [ ]  Function should return 200 with clean file key, with <1s execution time
  - [ ] File in quarantine bucket should be deleted
  - [ ] File should be in clean bucket with clean file key
- [ ] Put infected file (e.g. from [eicar](https://www.eicar.org/download-anti-malware-testfile/)) with valid uuid in quarantine bucket and trigger lambda with `key: <filename>`.
  - [ ]  Function should return 400 
    - [ ] Response body should include 'Malicious file detected'
    - [ ] Response body should include virus metadata
  - [ ] Malicious file should be deleted from the quarantine bucket.
- [ ] Put clean file with invalid uuid key in quarantine bucket and trigger lambda with `key: <filename>`.
  - [ ]  Function should return 400
  - [ ] Quarantined file should not be deleted
  - [ ] Clean bucket should not have new file
- [ ] Put clean file with valid uuid in quarantine bucket. Put another different clean file with the same uuid as key in quarantine bucket. trigger lambda with `key: <filename>`.
  - [ ] Function should return 200 with clean file key
  - [ ] File should be in clean bucket with clean file key, and correspond to the second clean file updated.
  - [ ] File in quarantine bucket should be deleted.

## Deploy Notes
- Set up new env vars (below) before release
- Set up S3 buckets corresponding to these env vars, and configure to delete after 1 day
- Ensure that the ECR repository named `serverless-virus-scanner-[production|staging|uat]` has been created.
 
- In `/virus-scanner/[staging|production|uat]`, we have
  - VIRUS_SCANNER_QUARANTINE_S3_BUCKET (with bucket versioning)
  - VIRUS_SCANNER_CLEAN_S3_BUCKET
- in general env var for the environment, also place a copy of the two bucket env vars
- FormSg and the virus scanner lambda are different apps, hence each have their env vars


serverless/virus-scanner/serverless.yml
```diff
@@ -0,0 +1,51 @@
...
+  region: ap-southeast-1
+  stage: ${opt:stage}
+  provisionedConcurrency: 0
+  iam:
+    role:
+      statements:
+        # Allow functions to read/write objects in a bucket
+        - Effect: Allow
+          Action:
+            - 's3:GetObject'
```
spaceraccoon, line 15: You may want to separate this out further such that the role only has GetObject access to the quarantine bucket and PutObject access to the clean bucket for principle of least privilege.

serverless/virus-scanner/serverless.yml
```diff
@@ -0,0 +1,51 @@
...
+            - 's3:PutObject'
+          Resource:
+            - 'arn:aws:s3:::${env:VIRUS_SCANNER_QUARANTINE_S3_BUCKET}/*'
```
spaceraccoon, line 21: Not sure if this is possible, but would be great to have this limited further to `'arn:aws:s3:::${env:VIRUS_SCANNER_QUARANTINE_S3_BUCKET}/${env:FILE_UUID}'` so that if for some reason there is a compromise at the lambda level, it should not have access to every other file in the clean/quarantine buckets.

serverless/virus-scanner/src/clamscan.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  logger.info('Scanning file stream')
+
+  const scanner = await new NodeClam().init({
+    clamdscan: {
+      socket: '/tmp/clamd.ctl',
+    },
+  })
+
+  // create a writable stream to memory
+  const outputStream = new Stream.Writable()
```
spaceraccoon: I'm wondering if this is necessary since whenever we go from streams to fixed buffers in memory it becomes a resource exhaustion/DoS risk; can it just do an S3 API CopyObject so you don't have to do this? The IAM role should be strictly scoped to the file though.

serverless/virus-scanner/src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  } else {
+    const cleanFileKey = crypto.randomUUID()
+    logger.info({
+      message: 'clean file detected',
+      cleanFileKey,
+    })
+
+    const { cleanFile } = scanResult
+
+    await s3Client.putS3File({
```
spaceraccoon: Same comment here about whether this can be a copy operation instead. It could even be a move operation.

serverless/virus-scanner/src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (isMalicious) {
+    const { virusMetadata } = scanResult
+
+    logger.error({
+      message: 'malicious file detected!!',
+      virusMetadata,
+      quarantineFileKey,
+    })
+    return {
+      statusCode: StatusCodes.BAD_REQUEST,
```
spaceraccoon, line 135: Thoughts on deleting malicious files or prefer to keep them for records?

serverless/virus-scanner/src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+      body: JSON.stringify({
+        message: 'Missing key in body',
+      }),
+    }
+  }
+
+  const quarantineFileKey = event.key
```
spaceraccoon, line 55: Should we add a validation that `event.key` is a UUID and nothing else?

spaceraccoon (commented):
I reviewed on the basis of ensuring principle of least privilege, since we are assuming that clamscan could be exploited at some point. Generally, the best defence here would be to ensure that the IAM is tightly scoped, since the compute is already separated by Lambda. I also did a quick review of `clamscan`'s `scanStream` API and it looks fine.

serverless/virus-scanner/serverless.yml
tshuli: thanks @spaceraccoon ! done in [8f1a191](https://github.com/opengovsg/FormSG/pull/6642/commits/8f1a1917fdb59130fe280d1377b863256840d8ba)

serverless/virus-scanner/src/clamscan.service.ts
tshuli: thanks for this! as discussed separately, we keep to the current setup to guarantee that the clean file being saved is the file being scanned

resource exhaustion is less of a risk as file size limit is enforced

serverless/virus-scanner/src/index.ts
tshuli: thanks, good point! we are already logging the virus meta data; deletion implemented in [c065fec](https://github.com/opengovsg/FormSG/pull/6642/commits/c065fec3489f0fc50c3586f8a342f9aa956bd8f1)

serverless/virus-scanner/src/index.ts
tshuli: done in [152d666](https://github.com/opengovsg/FormSG/pull/6642/commits/152d66686ce75ae8b47c22076af9390903de7a72) / [a0c4a17](https://github.com/opengovsg/FormSG/pull/6642/commits/a0c4a173c5fa15df9311a5094d64a097ecb605ba)

serverless/virus-scanner/serverless.yml
tshuli: i spent some time investigating this:

- i don’t think it is possible to define the perms in advance at build/deployment time (viz serverless.yml) since the filenames are randomised

- at execution time, it should be possible for the ec2 instance to update the lambda function's role privileges to scope to the specific file being scanned, before executing the function. but i am slightly concerned that if we expect a high level of concurrency, there might be a race condition at some point, where:
  - lambda role is updated by ec2 instance A to scope to instance A's file of interest
  - lambda role is updated by ec2 instance B to scope to instance B's file of interest
  - ec2 instance A calls lambda execution, which fails because it does not have the right permissions

- the risk is somewhat reduced now that we have enforced deletion of quarantine bucket file upon scan completion, and at the application level, we will delete the clean bucket file upon successful retrieval. finally, we have lifecycle rules as a catchall to delete files in s3 bucket after preset time

- The above considered, I'm not too inclined to implement further access control on the quarantine bucket at execution time

@spaceraccoon for your further thoughts pls, thanks very much for reviewing this!

serverless/virus-scanner/src/clamscan.service.ts
tshuli: Actually we are going to try enabling bucket versioning + use versionId as unique version key, then we can use copy operation without being concerned that the file was overwritten. will circle back on this

tshuli: copy and versioning implemented in [7dea90f](https://github.com/opengovsg/FormSG/pull/6642/commits/7dea90f284efeac09a502e14f65ff1ed4dc17532), so no more use of stream -> buffer

serverless/virus-scanner/src/index.ts
tshuli: [7dea90f](https://github.com/opengovsg/FormSG/pull/6642/commits/7dea90f284efeac09a502e14f65ff1ed4dc17532)

tshuli:
> Not sure if this is possible, but would be great to have this limited further to `'arn:aws:s3:::${env:VIRUS_SCANNER_QUARANTINE_S3_BUCKET}/${env:FILE_UUID}'` so that if for some reason there is a compromise at the lambda level, it should not have access to every other file in the clean/quarantine buckets.

@spaceraccoon have addressed your comments, this one remains. 

In addition to the considerations [in the comment thread](https://github.com/opengovsg/FormSG/pull/6642#discussion_r1302554729), the file keys are randomised and unguessable. As we are only giving `GetObject` and not `ListBucket` to the lambda, it will not be able to access the files in the buckets without being explicitly given the file keys by the ec2 instance. This is effectively the same as the ec2 instance updating the IAM perms before each function call.

serverless/virus-scanner/serverless.yml
tshuli: as spoken, this is resolved through the use of uuid filenames for defence in depth

serverless/virus-scanner/src/index.ts
```diff
@@ -0,0 +1,145 @@
...
+    const { virusMetadata } = scanResult
+
+    logger.error({
+      message: 'malicious file detected!!',
+      virusMetadata,
+      key: quarantineFileKey,
+    })
+
+    // Delete from quarantine bucket
+    await s3Client.deleteS3File({
```
ghost: Will delete remove all versions of the uploaded file?

serverless/virus-scanner/src/index.ts
```diff
@@ -0,0 +1,145 @@
...
+    // If clean, move to clean bucket with randomised key and return
+  } else {
+    const cleanFileKey = crypto.randomUUID()
+    logger.info({
+      message: 'clean file detected',
+      key: quarantineFileKey,
+      versionId,
+    })
+
+    await s3Client.moveS3File({
```
ghost: Should be wrapped in a try catch

serverless/virus-scanner/src/s3.service.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    try {
+      const { Body: body, VersionId: versionId } = await this.s3Client.send(
+        new GetObjectCommand({
+          Key: objectKey,
+          Bucket: bucketName,
+        }),
+      )
+
+      if (!versionId) {
```
ghost, line 62: should we check for empty body before proceeding?

serverless/virus-scanner/src/utils.ts
```diff
@@ -0,0 +1,3 @@
+export const decodeS3URI = (uri: string) => {
```
ghost: unused

sebastianwzq:
Not critical but we should configure the expiry of the presigned URL prudently so an attacker cannot reuse the presigned URL indefinitely and has to fetch a new one frequently.

Assuming that a new presigned URL is generated from an upload attachment endpoint of sorts and the endpoint is rate-limited, it will then effectively rate-limit subsequent steps down the pipeline.

spaceraccoon:
> > Not sure if this is possible, but would be great to have this limited further to `'arn:aws:s3:::${env:VIRUS_SCANNER_QUARANTINE_S3_BUCKET}/${env:FILE_UUID}'` so that if for some reason there is a compromise at the lambda level, it should not have access to every other file in the clean/quarantine buckets.
> 
> @spaceraccoon have addressed your comments, this one remains.
> 
> In addition to the considerations [in the comment thread](https://github.com/opengovsg/FormSG/pull/6642#discussion_r1302554729), the file keys are randomised and unguessable. As we are only giving `GetObject` and not `ListBucket` to the lambda, it will not be able to access the files in the buckets without being explicitly given the file keys by the ec2 instance. This is effectively the same as the ec2 instance updating the IAM perms before each function call.

Okay, I also took a look at IAM [conditions with variables](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#policy-vars-using-variables) and didn't see one that might be suitable for this use case, but I may have missed something.

Something to keep an eye on but if the caller is restricted to the security group + UUID defence in depth + user input never influences this argument it is sufficiently mitigated.

serverless/virus-scanner/src/utils.ts
tshuli: thanks! in [45cd743](https://github.com/opengovsg/FormSG/pull/6642/commits/45cd74377c49c5b7a318a3a67226643c39eb7c38)

serverless/virus-scanner/src/s3.service.ts
tshuli: [2402958](https://github.com/opengovsg/FormSG/pull/6642/commits/240295823b2974aecf6712f48d99fdee0ad7ca3e)

serverless/virus-scanner/src/index.ts
tshuli: [87e9e45](https://github.com/opengovsg/FormSG/pull/6642/commits/87e9e4560d9391c071234e53b9328165ea941e33)

serverless/virus-scanner/src/index.ts
tshuli: [7d29dc0](https://github.com/opengovsg/FormSG/pull/6642/commits/7d29dc0c5b3600f6fe6c69bb560bb3e925fbc4e5)

tshuli:
thanks @sebastianwzq for review. I've addessed the comments, and also made the following changes:

- Added automated tests
- Fixed and tested provisioned concurrency (1 for stg/uat, 5 for prod)
- Simplified the logger from the c360 implementation and added docs
- Added try/catch + logging where necessary

serverless/virus-scanner/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+# Point clamav log and socket to /tmp folder
+RUN sed -i 's/LogFile \/var\/log\/clamav\/clamav.log/LogFile \/tmp\/clamav.log/g' /etc/clamav/clamd.conf
+RUN sed -i 's/LocalSocket \/var\/run\/clamav\/clamd.ctl/LocalSocket \/tmp\/clamd.ctl/g' /etc/clamav/clamd.conf
+
+# !! Below comment was taken from care360 repo where the virus scanner was forked from!!
+# This is a hack to give the default lambda user permissions write access to the Local UNIX socketfile and logfile
+# The appropriate GID (990) and User (sbx_user1051) to add was found out through logging the user info at lambda runtime
+# GID and User is assumed to be deterministic and hence can be hardcoded
+# An alternative would be to get the current user + group details at runtime at the app level code, then run clamd
+# TODO (Austin): If the function breaks in the future because GID / User is non-deterministic, experiment with executing command at app runtime
```
LinHuiqing: Should we be tracking this TODO as a ticket too?

serverless/virus-scanner/scripts/envLoader.mjs
```diff
@@ -0,0 +1,49 @@
...
+
+  if (process.env.ENV === 'development') {
+    console.log('In develop mode! Not fetching from SSM param store.')
+    console.log(
+      'Please reference .env.example to populate .env.development file for development environment',
+    )
+    exit(0)
+  }
+  const client = new SSMClient({ region: 'ap-southeast-1' })
+  const parameterName = `/virus-scanner/${SHORT_ENV_MAP[process.env.ENV]}`
```
LinHuiqing: nit: a little inconsistency here - "dev", "develop" and "development" are all accepted as dev env strings in `SHORT_ENV_MAP`, but only if its equal to "development" will it reference the `.env.example` file. Is this intentional?

Another nit: what do we want to do if the env is set as an arbitrary string? Do we still want it to error out?

serverless/virus-scanner/src/__tests/clamscan.service.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      // Act
+
+      const result = await scanFileStream(mockStream)
+
+      // Assert
+      expect(result.isMalicious).toBe(true)
+    })
+
+    it('should return virus metadata if virus found', async () => {
```
LinHuiqing: Should combine these 2 tests since it is testing the same case. Something like 'should return isMalicious true and virus metadata if virus found'.

serverless/virus-scanner/src/__tests/clamscan.service.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      // Act
+      const result = await scanFileStream(mockStream)
+
+      // Assert
+      expect(result.isMalicious).toBe(false)
+    })
+
+    it('should return empty virus metadata if no virus found', async () => {
```
LinHuiqing: Same for these 2 tests, we should combine them ('should return isMalicious false and empty virus metadata if no virus found') and have the 2 assertions.

serverless/virus-scanner/src/clamscan.service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  logger.info('Scanning file stream')
+
+  const scanner = await new NodeClam().init({
+    clamdscan: {
+      socket: '/tmp/clamd.ctl',
+    },
+  })
+
+  const { isInfected: isMalicious, viruses: virusMetadata } =
```
LinHuiqing, line 20: just curious, why do we want to change `isInfected` to `isMalicious`?

serverless/virus-scanner/src/clamscan.service.ts
```diff
@@ -0,0 +1,29 @@
...
+    await scanner.scanStream(s3Stream)
+
+  return isMalicious === true
```
LinHuiqing: nit: can be more concise since `isMalicious` is already typed as a `boolean`
```suggestion
  return isMalicious
```

LinHuiqing (commented):
Generally lgtm! Just some nits. Will approve after I test it 👍

LinHuiqing:
Modified manual tests, with modifications italicised:

- [ ] Put infected file (e.g. from [eicar](https://www.eicar.org/download-anti-malware-testfile/)) with valid uuid in quarantine bucket and trigger lambda with `key: <filename>`.
  - [ ]  Function should return 400 
    - [ ] Response body should include 'Malicious file detected'
    - [ ] Response body should include virus metadata
  - [ ] _Malicious file should be deleted from the quarantine bucket._
- [ ] Put clean file with invalid uuid key in quarantine bucket and trigger lambda with `key: <filename>`.
  - [ ]  Function should return _400_
  - [ ] Quarantined file should not be deleted
  - [ ] Clean bucket should not have new file
- [ ] Put clean file with valid uuid in quarantine bucket. Put another different clean file with the same uuid as key in quarantine bucket. trigger lambda with `key: <filename>`.
  - [ ] Function should return 200 with clean file key
  - [ ] File should be in clean bucket with clean file key, and correspond to the second clean file updated.
  - [ ] _File in quarantine bucket should be deleted._

Edit: added to the PR description!

LinHuiqing (approved):
Should be good to go after the above nits have been addressed! 🤞 Approving so that I won't be blocking you from merging after the nits have been fixed 👍

serverless/virus-scanner/Dockerfile
tshuli: Thanks! actually i don't think this is really a TODO, it's more of an info note in case something breaks (`If the function breaks...`). So i've removed the todo here

[d8eee93](https://github.com/opengovsg/FormSG/pull/6642/commits/d8eee931805e77272c69600afdf71eef7dc2fba4)

serverless/virus-scanner/scripts/envLoader.mjs
tshuli: thanks! good catch. [e1b20c3](https://github.com/opengovsg/FormSG/pull/6642/commits/e1b20c3bcf9aa3c9b54949a45ffd89443c218d87)

also yes it should error out if the env name is not recognised:
[3a8d77a](https://github.com/opengovsg/FormSG/pull/6642/commits/3a8d77aeaaf7ed14c685fe2e6986d44b3a196f62)

serverless/virus-scanner/src/__tests/clamscan.service.spec.ts
tshuli: [4170813](https://github.com/opengovsg/FormSG/pull/6642/commits/4170813cb0588804d15d62a51038d6177d835b24)

serverless/virus-scanner/src/__tests/clamscan.service.spec.ts
tshuli: [4170813](https://github.com/opengovsg/FormSG/pull/6642/commits/4170813cb0588804d15d62a51038d6177d835b24)

serverless/virus-scanner/src/clamscan.service.ts
tshuli: 🤔 hmm no real reason actually; for `virusMetadata` it felt more informative than `viruses` (which could have referred to the actual virus bytestream) but I must admit I didn't think too much about `isInfected` -> `isMalicious`

I'll leave it since it's just a minor naming difference that does no harm

LinHuiqing:
Closes FRM-1212

LinHuiqing:
Added to deploy notes:

- Ensure that the ECR repository named `serverless-virus-scanner-[production|staging|uat]` has been created.

PR #6646, opened 2023-08-22
fix: prevent saving empty product items

KenLSM:
## Problem

Admins are able to save when there are no products. This results in open forms with payments that has no items to select.

Closes FRM-1256

## Solution

- Saving of product items with 0 items will be blocked.
- Deleting of the final item will be disabled.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

Regression
- [ ] On admin view, deleting product items removes the deleted product
- [ ] On admin view, deleting payment component removes the payment component 
  - [ ] On responder view, the form should not have payment component

Feature Test
- [ ] On admin view, `SaveButton` is not interactive if there are no `ProductItems`
- [ ] On admin view, the final `ProductItem` cannot be deleted

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/FormFieldDrawerActions.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/common/FormFieldDrawerActions.tsx
@@ -20,7 +20,6 @@ export const FormFieldDrawerActions = ({
   isDisabled,
 }: FormFieldDrawerActionsProps): JSX.Element => {
   const isMobile = useIsMobile()
-  const isSavingDisabled = isDisabled || isLoading
 
   return (
     <Stack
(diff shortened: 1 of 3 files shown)


foochifa (approved):
Lgtm and tested!

foochifa (changes requested):
Sorry! Didn't realise the endpoint for specific products is different. Have discussed with @KenLSM and whilst this solves the saving issue without any products

However, you could save -> then delete the only product, and it will be reflected in the form

Request to disable deletion button of the product if it is the only one in the product array.

LinHuiqing:
> Sorry! Didn't realise the endpoint for specific products is different. Have discussed with @KenLSM and whilst this solves the saving issue without any products
> 
> However, you could save -> then delete the only product, and it will be reflected in the form
> 
> Request to disable deletion button of the product if it is the only one in the product array.

Ah I see, hmm. Will it make more sense to only update the products on DB when "Save field" is clicked? Is there a reason why there are separate calls made for when the individual products are added / deleted within the field? Since the products are also sent over when "Save field" clicked.

foochifa (approved):
As discussed, lets merge this in first and fix the edge case later!

PR #6648, opened 2023-08-23
fix: Myinfo Child DOB

wanlingt:
## Problem

Myinfo Child DOB is not showing up in the correct format in the frontend.

## Solution

When the Myinfo Child DOB value is set, convert the date from the formats 'yyyy-MM-dd' to 'dd/MM/yyyy'.

The way the Myinfo Child values are processed is different from normal Myinfo fields. For example, for normal Myinfo fields, the preview of the field is distinct from the value of the field that's submitted. However, for Myinfo child fields, the value shown in the field is what's submitted. Hence, in this case, the formatting takes place in 2 areas:
1. In the frontend (`ChildrenCompoundField`) before the field value is set
2. In the backend, before the value gets hashed (`myinfo.util.ts`)

(TODO for the future - Ideally, the Myinfo child values will be processed in the same way as the normal Myinfo fields)

**Details**
- Created a shared function to format Myinfo dates (`formatMyinfoDate`)

**Breaking Changes** 

## Before & After Screenshots
(Look at the Child DOB field)

**BEFORE**:
image

**AFTER**:
image

## Tests

1. Happy path test
- [ ] On a Myinfo form, create a child field with the DOB.
- [ ] Fill in the form (Staging Myinfo accounts [here](https://www.notion.so/opengov/MyInfo-Children-Fields-on-Staging-54587512c140404c91a0dde54becb277))
- [ ] The DOB should be in the format 'dd/MM/yyyy' on the form UI.
- [ ] Submit the form
- [ ] Check the email submission - the DOB for the child's field should also be in the same format.

2. Regression test
- [ ] On a Myinfo form, add the (non-child) Myinfo fields `DOB` (already has existing Myinfo data) and `Marriage Date` (has no existing Myinfo data)
- [ ] The dates should be in the format 'dd/MM/yyyy' on the form UI.
- [ ] Submit the form
- [ ] Check the email submission - the dates should show up correctly


src/app/modules/submission/email-submission/email-submission.constants.ts
```diff
@@ -3,7 +3,7 @@ export const MYINFO_PREFIX = '[MyInfo] '
 export const VERIFIED_PREFIX = '[verified] '
 export const TABLE_PREFIX = '[table] '
 export const ATTACHMENT_PREFIX = '[attachment] '
-export const CHILD_PREFIX = '[child]'
+export const CHILD_PREFIX = '[child] '
```
foochifa, line 6: Hmm is there a reason for this space here? perhaps add a comment for why it exists?

shared/constants/dates.ts
```diff
@@ -0,0 +1,3 @@
+export const DATE_DISPLAY_FORMAT = 'dd/MM/yyyy'
```
foochifa, line 1: Hmm actually our display and parse format is the same right? Is there a reason why we continue to keep them separated?

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -324,12 +326,24 @@ const ChildrenBody = ({
           const key = `${field.id}+${index}`
           const fieldPath = `${schema._id}.child.${currChildBodyIdx}.${index}`
           const myInfoValue = getChildAttr(subField)
+          let myInfoFormattedValue
```
foochifa: nit but I'm thinking it might be better to use const here instead

```
const myInfoFormattedValue = subField === MyInfoChildAttributes.ChildDateOfBirth && myInfoValue 
? formatMyinfoDate(myInfoValue)
: myInfoValue
```

shared/utils/dates.ts
```diff
@@ -0,0 +1,6 @@
+import { format, parse } from 'date-fns'
+import { MYINFO_DATE_FORMAT, DATE_PARSE_FORMAT } from '../constants'
+
+export const formatMyinfoDate = (value: string) => {
```
foochifa, line 4: Very big nit but should it be `formatMyInfoDate`? Cause we do capitalise `Info` right hahah

foochifa (approved):
Generally LGTM! Thanks for locating the problem @wanlingt

src/app/modules/submission/email-submission/email-submission.constants.ts
wanlingt: done! [6b2a7b9](https://github.com/opengovsg/FormSG/pull/6648/commits/6b2a7b92c494959fad7ba0c06a019e265ea860f7)

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
wanlingt: yup good point, done in [ee15c01](https://github.com/opengovsg/FormSG/pull/6648/commits/ee15c01c43f6f7661d89407a7602f2910a10f7c5)

shared/utils/dates.ts
wanlingt: haha so ACTUALLY I had the same thought, then I googled to check. Turns out it's actually [Myinfo](https://api.singpass.gov.sg/library/myinfo/business/implementation-myinfo-data) instead of MyInfo lolol

shared/constants/dates.ts
wanlingt: They're currently the same, but as they have distinct meanings I think it's better to keep them separate. One example is the [DateField](https://github.com/opengovsg/FormSG/blob/develop/frontend/src/templates/Field/Date/DateField.tsx#L84-L85)

foochifa: Ahh I see okay thanks @wanlingt, and thanks for abstracting them out to a shared constants file

foochifa (approved):
Approving again haha LGTM Thanks @wanlingt

PR #6652, opened 2023-08-23
feat: enable myinfo child vax field

LinHuiqing:
## Problem

Serving Child Vaccination Requirements has been approved by MyInfo since https://github.com/opengovsg/FormSG/pull/6523, so we should be enabling it.

Thanks @kenjin-work for all the prior work to make this so easy! :)

Closes FRM-1195

## Solution

**Breaking Changes** 

**Improvements**:

- Removed filter for vaccination requirements.

## Before & After Screenshots

**BEFORE**:

Vaccination Requirements not available as one of the options:

Screenshot 2023-08-23 at 2 23 40 PM

**AFTER**:

Vaccination Requirements available as one of the options and can be added:

Screenshot 2023-08-23 at 2 22 06 PM

## Tests

- [ ] Add the `children` beta flag to your admin user.
- [ ] On an email mode form, add a children field and check that "Vaccination Requirements" is one of the options and can be added to the field.


tshuli (approved):
lgtm!

PR #6653, opened 2023-08-23
fix: address should be lower case

KenLSM:
## Problem

Typo on `Email Address`, should be in lowercase.


foochifa (approved):
😅 thanks Ken

PR #6654, opened 2023-08-24
fix: failed ensures should block submission

LinHuiqing:
## Problem

Bug was introduced by #6595 where failing ensures do not return early if a response has already been sent.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Failing ensures should return early.

## Tests

Check that storage mode submissions will be gated by ensures.
   - [ ] ensureIsPublicForm:
      - [ ] Go to a public form.
      - [ ] Close the form on the admin portal.
      - [ ] Make a submission. Submission should not succeed.
         - [ ] Error should be shown to client.
         - [ ] New submission should **_not_** show up on DB
   - [ ] ensureValidCaptcha:
      - [ ] Go to a public form with captcha disabled.
      - [ ] Enable captcha on the admin portal.
      - [ ] Make a submission. Submission should not succeed.
         - [ ] Error should be shown to client.
         - [ ] New submission should **_not_** show up on DB
   - [ ] ensureFormWithinSubmissionLimits:
      - [ ] Go to a public form with a submission limit. The form should still be open at this point.
      - [ ] On the DB, edit the submission limit so that it is less than or equal to the number of existing submissions.
      - [ ] Make a submission. Submission should not succeed.
         - [ ] Error should be shown to client.
         - [ ] New submission should **_not_** show up on DB

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -121,7 +121,8 @@ const submitEncryptModeForm = async (
         new SubmissionFailedError(),
       )
       return res.status(statusCode).json({ message: errorMessage })
-    } else return // required to stop submission processing
+    }
+    return // required to stop submission processing
   }
 
   const encryptedPayload = req.formsg.encryptedPayload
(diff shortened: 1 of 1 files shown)


tshuli (approved):
lgtm

PR #6665, opened 2023-08-27
feat: % rollout of new storage submission endpoint

LinHuiqing:
## Problem

This is part of our initiative to [shift our encryption boundary away from end-to-end encryption](https://docs.google.com/document/d/1VmNXS_xYY2Yg30AwVqzdndBp5yRJGSDsyjBnH51ktyc/edit?usp=sharing) so that we have more flexibility to implement features such as submission validation and virus scanning for storage mode forms.

Now that the new storage submission endpoint with the encryption boundary shift is up (#6595), we want to serve live traffic through it. To limit the impact of any bugs, but still test a full range of payloads / workflows, this rollout is done as a percentage rollout based on the form id. 

Growthbook is used as the feature flag system as an upgrade from our various beta flags and feature flags we have been using in our DB. This PR only implements Growthbook on the frontend.

This PR also updates the Alpine LInux mirror used from "http://nl.alpinelinux.org/alpine/edge/testing" to "http://dl-cdn.alpinelinux.org/alpine/edge/testing" as the former is outdated.

Closes FRM-1277, closes FRM-1275, works towards FRM-1215

## Solution

Note: If Growthbook features fail to load for whatever reason, the features will evaluate to `null`. Docs: https://docs.growthbook.io/lib/react#built-in-fetching-and-caching. The outcome for this PR is that we will serve the old storage submission endpoint.

**Breaking Changes** 

- [x] No - this PR is backwards compatible. However, release should be more involved, with the following steps (pasted in deploy notes too):
   - [ ] 1. Check that github action secret for `APP_URL` is correctly set (e.g. `https://staging-alt2.form.gov.sg`)
   - [ ] 2. Check that cloudflare worker is set up to proxy traffic to `/api/v1/proxy/growthbook/` route
   - [ ] 3. In the `featureflags` collection of the DB, set `encryption-boundary-shift` as `true`.
   - [ ] 4. On Growthbook, turn on the `encryption-boundary-shift` feature flag for the appropriate environment and set rollout % on Growthbook to 0.
   - [ ] 5. On Growthbook, get the client key for the appropriate environment and add `GROWTHBOOK_CLIENT_KEY` to env vars.
   - [ ] 6. Deploy app.
   - [ ] 7. (Clean up) Run the `unset-encryption-boundary-shift.js` script on the DB.
   - [ ] 8. Increase rollout % on Growthbook to 1% and pray all goes well 🙏 Monitor on the main [FormSG datadog dashboard](https://app.datadoghq.com/dashboard/b3u-egb-apu/form-s-g?from_ts=1692704946967&to_ts=1693309746967&live=true), under Submissions.

**Features**:

- Directs some storage submission traffic to the new storage submission endpoint (`/submissions/storage`) based on percentage rollout specified.

**Improvements**:

- Starts using Growthbook for feature flagging on the frontend.
  - To avoid having to whitelist growthbook's domain for GSIB / HVB users, we have set up a proxy at `/api/v1/proxy/growthbook/` through cloudflare. The script that we used to do so is [here](https://github.com/opengovsg/formsg-private/pull/171).

**Bug Fixes**:

- Fixes `responseVisibilityPredicate` in `IncomingEncryptSubmission` to accept response types that have `answerArray`s instead of `answer`s. This was not an issue previously as we were only checking mobile and email fields for storage mode submissions.
- Updates the Alpine LInux mirror used from "http://nl.alpinelinux.org/alpine/edge/testing" to "http://dl-cdn.alpinelinux.org/alpine/edge/testing" so `font-wqy-zenhei@edge` and `wqy-zenhei@edge` can still be loaded. `testing` has to be used because the packages mentioned are only available in `testing`.

## Tests

Before all tests, make sure that:
- [ ] `GROWTHBOOK_CLIENT_KEY` has been set.
- [ ] `encryption-boundary-shift` in `featureflags` collection has been set to `true`.

Test that feature flagging works:
- [ ] Create a storage mode form or find a form with all the field types + logic.
- [ ] Go to the respondent view of the form.
- [ ] Open the network panel.
- [ ] Fill in the form, making sure that the submission includes answer in fields with answers that will be presented in an array (i.e. checkbox and table).
- [ ] Submit the form. Check that the submission endpoint used is either `/submissions/encrypt` or `/submissions/storage`.
- [ ] Repeat the above steps until both `/submissions/encrypt` and /submissions/storage` endpoints are tested. The probability of hitting each of these endpoints can be controlled on Growthbook.

Regression tests:
- [ ] Make sure the storage mode submissions include the following cases:
   - [ ] Fields hidden by logic.
   - [ ] Fields that might be hidden by logic but are not.
   - [ ] Attachment included.
   - [ ] Optional fields.
   - [ ] Checkboxes.
   - [ ] Table fields.
- [ ] Using cURL requests, submit invalid submissions. These should not pass. Test cases include:
   - [ ] Submission include answers on hidden fields.
   - [ ] Invalid answer format (e.g. words in number / decimal fields).

## Deploy Notes

**⚠️ Steps:**

   - [ ] 1. Check that github action secret for `APP_URL` is correctly set (e.g. `https://staging-alt2.form.gov.sg`)
   - [ ] 2. Check that cloudflare worker is set up to proxy traffic to `/api/v1/proxy/growthbook/` route
   - [ ] 3. In the `featureflags` collection of the DB, set `encryption-boundary-shift` as `true`.
   - [ ] 4. On Growthbook, turn on the `encryption-boundary-shift` feature flag for the appropriate environment and set rollout % on Growthbook to 0.
   - [ ] 5. On Growthbook, get the client key for the appropriate environment and add `GROWTHBOOK_CLIENT_KEY` to env vars.
   - [ ] 6. Deploy app.
   - [ ] 7. (Clean up) Run the `unset-encryption-boundary-shift.js` script on the DB.
   - [ ] 8. Increase rollout % on Growthbook to 1% and pray all goes well 🙏 Monitor on the main [FormSG datadog dashboard](https://app.datadoghq.com/dashboard/b3u-egb-apu/form-s-g?from_ts=1692704946967&to_ts=1693309746967&live=true), under Submissions.

- `GROWTHBOOK_CLIENT_KEY`: Growthbook client key for SDK connection configured. To get the key:
   - Login to https://app.growthbook.io/.
   - Go to SDK Configuration > SDK Connections.
   - Click on the appropriate connection entry based on the environment.
   - Copy the Client Key.

**New scripts**:

- `scripts/20230826_unset-encryption-boundary-shift/unset-encryption-boundary-shift.js`: unsets `encryptionBoundaryShift` flags for all forms.
   - A dry run can be done with `session.abortTransaction();` before we swap it to `session.commitTransaction();` to run the update operation for real.

**New dependencies**:

On frontend:
- `@growthbook/growthbook-react`: package to use growthbook on the frontend.

**⚠️ Rollback plan:**
- [ ] 1. Go to https://app.growthbook.io/features/encryption-boundary-shift. Select disable to feature for `prod`.
- [ ] 2. On the DB, go to the `featureflags` collection and set `encryption-boundary-shift` to `false`.


LinHuiqing:
Oh, I'll fix the backend tests in a bit ps.

frontend/src/features/public-form/utils/createSubmission.ts
```diff
@@ -78,20 +78,30 @@ export const createEncryptedSubmissionData = async ({
...
+ * Used for both Email mode submissions and Storage mode submissions after encryption boundary shift.
  * @returns formData containing form responses and attachments.
  * @throws Error if form inputs are invalid.
  */
-export const createEmailSubmissionFormData = (
+export const createClearSubmissionFormData = (
   formFields: FormFieldDto[],
   formInputs: FormFieldValues,
   responseMetadata?: ResponseMetadata,
+  setBody?: {
```
tshuli: suggestion: maybe something like `additionalBodyFields` might be clearer than `setBody`

frontend/src/features/public-form/PublicFormService.ts
```diff
@@ -93,6 +93,12 @@ export type SubmitStorageFormArgs = SubmitEmailFormArgs & {
   payments?: PaymentFieldsDto
 }
 
+export type SubmitStorageFormV2Args = SubmitEmailFormArgs & {
```
tshuli: suggestion: `SubmitStorageFormClearArgs`, to reduce the need to know the v1 vs v2 context

frontend/src/features/public-form/PublicFormService.ts
```diff
@@ -164,6 +170,48 @@ export const submitStorageModeForm = async ({
   ).then(({ data }) => data)
 }
 
+export const submitStorageModeFormV2 = async ({
```
tshuli: suggestion - create a fetch version as well; our longer term plan is to handle the fetch fallback more elegantly, but this can be done in another PR

frontend/src/features/public-form/mutations.ts
```diff
@@ -80,6 +81,12 @@ export const usePublicFormMutations = (
     },
   )
 
+  const submitStorageModeFormV2Mutation = useMutation(
```
tshuli: likewise, suggest to also have a fetch fallback

scripts/20230826_set-encryption-boundary-shift-true/set-encryption-boundary-shift-true.js
```diff
@@ -0,0 +1,49 @@
...
+// # of documents with encryptionBoundaryShift set to true
+const docsSetToTrueBefore = formsCollection.countDocuments({ encryptionBoundaryShift: true })
+console.log('# of documents with encryptionBoundaryShift set true:', docsSetToTrueBefore);
+
+// # of documents to modify
+console.log('# of documents to modify:', totalDocs - docsSetToTrueBefore);
+
+
+// set documents
+formsCollection.updateMany(
```
tshuli: /nit: we only need to set for storage mode forms

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
```diff
@@ -60,9 +60,12 @@ export default class IncomingEncryptSubmission extends IncomingSubmission {
...
   responseVisibilityPredicate(response: FieldResponse): boolean {
     return (
-      'answer' in response &&
-      typeof response.answer === 'string' &&
-      response.answer.trim() !== ''
+      ('answer' in response &&
+        typeof response.answer === 'string' &&
+        response.answer.trim() !== '') ||
+      ('answerArray' in response &&
+        typeof response.answerArray === 'object' &&
```
tshuli: Some gotchas here 😅 
- in nodejs, `object` is [anything that is not undefined, boolean, number, bigint, string, symbol, or fn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null).
- null is also type `object`
  ```
  const x = null
  typeof x // 'object'
  ```
  
- You might want to use `Array.isArray()` instead

frontend/src/features/public-form/utils/createSubmission.ts
```diff
@@ -78,20 +78,30 @@ export const createEncryptedSubmissionData = async ({
 }
 
 /**
+ * Used for both Email mode submissions and Storage mode submissions after encryption boundary shift.
  * @returns formData containing form responses and attachments.
  * @throws Error if form inputs are invalid.
  */
-export const createEmailSubmissionFormData = (
+export const createClearSubmissionFormData = (
```
tshuli, line 98: since version is actually non-optional for storage mode form, i feel that maybe a union type is better

`createClearSubmissionFormData = (formData: EmailSubmissionFormData | StorageSubmissionFormData)`

and then you can enforce that `StorageSubmissionFormData` contains `setBody: {version: number}`

tshuli (commented):
thanks @LinHuiqing ! Looking good 👍 
will review again after the tests are in 🙏 

quick qn - if we are just going to set encryptionBoundaryShift:true for all forms via DB, why not just disable that check in the controller? saves the effort of cleaning up the DB later

LinHuiqing:
> thanks @LinHuiqing ! Looking good 👍 will review again after the tests are in 🙏
> 
> quick qn - if we are just going to set encryptionBoundaryShift:true for all forms via DB, why not just disable that check in the controller? saves the effort of cleaning up the DB later

Thanks @tshuli for the review! To answer your question, yes, I've thought about that, but I decided to leave the flags there to be checked in the code in case something happens (e.g. security threat) and we want to quickly block the endpoint. With the  flags, we can simply change the % on growthbook and set all the flags on the DB to false without redeploying.

edit: I'll add TODOs to remove these things as part of FRM-1232

LinHuiqing:
> > thanks @LinHuiqing ! Looking good 👍 will review again after the tests are in 🙏
> > quick qn - if we are just going to set encryptionBoundaryShift:true for all forms via DB, why not just disable that check in the controller? saves the effort of cleaning up the DB later
> 
> Thanks @tshuli for the review! To answer your question, yes, I've thought about that, but I decided to leave the flags there to be checked in the code in case something happens (e.g. security threat) and we want to quickly block the endpoint. With the flags, we can simply change the % on growthbook and set all the flags on the DB to false without redeploying.
> 
> edit: I'll add TODOs to remove these things as part of [FRM-1232](https://linear.app/ogp/issue/FRM-1232)

As discussed on Slack, running a DB migration script in the case of an incident is high-risk! Instead, we have landed on using an env-wide feature flag using the `featureflags` collection on the DB to guard access to the new endpoint instead in the case of incidents.

LinHuiqing:
> Oh, I'll fix the backend tests in a bit ps.

Broken backend tests fixed in [15c90f8](https://github.com/opengovsg/FormSG/pull/6665/commits/15c90f80e4e11b2d98d238ce4dfdaf105d4aec08).

frontend/src/features/public-form/utils/createSubmission.ts
LinHuiqing: Thanks for the better nomenclature suggestion! Done here: [de584a9](https://github.com/opengovsg/FormSG/pull/6665/commits/de584a90f555b5e56344c1440a272a258cc8b562)

frontend/src/features/public-form/utils/createSubmission.ts
LinHuiqing: Thanks Shu Li! Good point, done in [9847df0](https://github.com/opengovsg/FormSG/pull/6665/commits/9847df0b2b21db1c5238a549e41718f7cbf0d03d)

src/app/modules/submission/encrypt-submission/IncomingEncryptSubmission.class.ts
LinHuiqing: Thanks for pointing this out! Done in [6b16843](https://github.com/opengovsg/FormSG/pull/6665/commits/6b16843491fc0330a551d23e5d9cdc50431f7a48).

frontend/src/features/public-form/PublicFormService.ts
LinHuiqing: Renamed in [7378d03](https://github.com/opengovsg/FormSG/pull/6665/commits/7378d0353be0e8ee013d4429d87b4126d216cd25) 👍

frontend/src/features/public-form/PublicFormService.ts
LinHuiqing: fetch version created in [d6e9db3](https://github.com/opengovsg/FormSG/pull/6665/commits/d6e9db359c1e494a792b28388408643b8bf8eb45)

frontend/src/features/public-form/mutations.ts
LinHuiqing: fetch version created in [d6e9db3](https://github.com/opengovsg/FormSG/pull/6665/commits/d6e9db359c1e494a792b28388408643b8bf8eb45)

frontend/src/features/public-form/PublicFormService.ts
LinHuiqing: Renamed more "V2"s to "Clear" in [d6e9db3](https://github.com/opengovsg/FormSG/pull/6665/commits/d6e9db359c1e494a792b28388408643b8bf8eb45)

LinHuiqing:
> > > thanks @LinHuiqing ! Looking good 👍 will review again after the tests are in 🙏
> > > quick qn - if we are just going to set encryptionBoundaryShift:true for all forms via DB, why not just disable that check in the controller? saves the effort of cleaning up the DB later
> > 
> > 
> > Thanks @tshuli for the review! To answer your question, yes, I've thought about that, but I decided to leave the flags there to be checked in the code in case something happens (e.g. security threat) and we want to quickly block the endpoint. With the flags, we can simply change the % on growthbook and set all the flags on the DB to false without redeploying.
> > edit: I'll add TODOs to remove these things as part of [FRM-1232](https://linear.app/ogp/issue/FRM-1232)
> 
> As discussed on Slack, running a DB migration script in the case of an incident is high-risk! Instead, we have landed on using an env-wide feature flag using the `featureflags` collection on the DB to guard access to the new endpoint instead in the case of incidents.

Gated new storage submission endpoint with env-level feature on the BE instead in [2593cd4](https://github.com/opengovsg/FormSG/pull/6665/commits/2593cd448dbd34004bf22369edc40b52f1d4eec2) and [8af5b52](https://github.com/opengovsg/FormSG/pull/6665/commits/8af5b52bd98529ffc2acac743393e3b5e0383ba1). 

Clean up script for form-level flags in [ab5d256](https://github.com/opengovsg/FormSG/pull/6665/commits/ab5d256863f0534fed04dffb60e52e8f5de23f2c).

frontend/src/features/public-form/PublicFormService.ts
```diff
@@ -25,13 +25,15 @@ import {
...
 import {
-  createEmailSubmissionFormData,
+  createClearSubmissionFormData,
   createEncryptedSubmissionData,
 } from './utils/createSubmission'
 import { filterHiddenInputs } from './utils/filterHiddenInputs'
 
 export const PUBLIC_FORMS_ENDPOINT = '/forms'
 
+const ENCRYPTION_BOUNDARY_SHIFT_ENCRYPTION_VERSION = 2
```
tshuli, line 37: For context - suggest just add a comment here about what the encryption boundary shift was about + link to PR number, so in future we keep context of what encryption boundary shift was all about

frontend/src/growthbook.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { GROWTHBOOK_API_HOST_PATH } from '~constants/links'
+
+export const createGrowthbookInstance = (
+  clientKey: string,
+  formsgSdkMode?: 'staging' | 'production' | 'development' | 'test',
+) => {
+  return new GrowthBook({
+    apiHost: `${
+      process.env.REACT_APP_URL ??
```
tshuli: suggestion: I think a bit odd to have the worker staging route as the fallback
- Suggest no need to nullish coaleasing here, we can assume REACT_APP_URL will be set, same as for useEnv()

scripts/20230826_unset-encryption-boundary-shift/unset-encryption-boundary-shift.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable */
```
tshuli, line 1: qn: do we still need this script?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -135,6 +136,7 @@ const submitEncryptModeForm = async (
     form,
     responses,
     encryptedContent,
+    req.formsg.featureFlags.includes(featureFlags.encryptionBoundaryShift),
```
tshuli, line 139: qn: how about if the value is false, will this still resolve to true?

tshuli (approved):
minor comments; otherwise lgtm!

scripts/20230826_unset-encryption-boundary-shift/unset-encryption-boundary-shift.js
LinHuiqing: As discussed, yes, it's a clean up script for the form-level `encryptionBoundaryShift` flags, if any.

frontend/src/features/public-form/PublicFormService.ts
LinHuiqing: comment added in [f77be1e](https://github.com/opengovsg/FormSG/pull/6665/commits/f77be1e870a75475f177caf1bfe17f7ff26453d8)

frontend/src/growthbook.ts
LinHuiqing: As discussed, we should use `isDev` to clearly convey that hardcoded link is for development only. Done in [16f255c](https://github.com/opengovsg/FormSG/pull/6665/commits/16f255c6b0d2a5dc5f1508058cebb9902245db0c) and [9d27524](https://github.com/opengovsg/FormSG/pull/6665/commits/9d27524fcf442b5171ac39ac0620588a16b75740)

scripts/20230826_set-encryption-boundary-shift-true/set-encryption-boundary-shift-true.js
LinHuiqing: Updated the script to unset the field instead as we are no longer using form-level flags for the BE when this goes out! Done in [ab5d256](https://github.com/opengovsg/FormSG/pull/6665/commits/ab5d256863f0534fed04dffb60e52e8f5de23f2c)

LinHuiqing:
Tested on staging-alt2.form.gov.sg and growthbook is working now ☺️

LinHuiqing:
Will merge after payment quantity selector PR (https://github.com/opengovsg/FormSG/pull/6669) is live to limit the scope of what can go wrong (before a long weekend) 👍

LinHuiqing:
fyi, [42e52d3](https://github.com/opengovsg/FormSG/pull/6665/commits/42e52d3cdea8e1cd60a15fe492befa32622f617e) moves feature loading to `AppRouter` level so we'll only need to load features once with auto refresh, and the children rendered just need to set attributes and evaluate features. Instead of separate components creating their own hooks to the load features, which is what was implied previously.

PR #6673, opened 2023-08-29
fix: allow submission when Myinfo child field is hidden by logic

wanlingt:
## Problem

When a Myinfo child field is hidden due to logic, the user is unable to submit the form. Instead, the user sees a `cannot read properties of undefined` error.

Closes FRM-1299

## Solution

The problem was because the user needed to populate the child field, even if the field was hidden by logic. To fix this, return an array of array of strings if `input` is undefined in `transformToChildOutput`, which transforms form inputs to their desire output shapes for sending to the server. This is similar to other field transformations, where an empty string is the default return output.

Also, add a check for a child in the `validChildAnswersNonEmpty` validator. 

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

## Tests

For staging, [these](https://www.notion.so/opengov/MyInfo-Children-Fields-on-Staging-54587512c140404c91a0dde54becb277) are the NRICs of profiles with children.
- [ ] Submit a form with a hidden Myinfo child field. The child's name should be the only Myinfo child field. The email submission should have a blank beside the Myinfo child field.
- [ ] Repeat the above, but with more child fields (BC number, DOB). The email submission should have a blank beside the Myinfo child fields.
- [ ] Submit a form where the Myinfo child field is unhidden. The child's name should be the only Myinfo child field. The email submission's Myinfo child field should be populated.
- [ ] Repeat the above, but with more child fields (BC number, DOB). The email submission's Myinfo child fields should be populated.
- [ ] Repeat the above, but do not select the child's name. Try to submit. A red border should appear around the Myinfo child field, and you should not be able to submit the form.

Regression test:
- [ ] Submit a Myinfo form without any Myinfo child fields

### DIFF
--- a/frontend/src/features/public-form/utils/inputTransformation.ts
+++ b/frontend/src/features/public-form/utils/inputTransformation.ts
@@ -189,8 +189,6 @@ const transformToChildOutput = (
   schema: ChildrenCompoundFieldSchema,
   input?: ChildrenCompoundFieldValues,
 ): ChildBirthRecordsResponse => {
-  console.log('transformToChildOutput: ', input)
-  console.log('transformToChildOutput schema: ', schema)
   const noOfChildrenSubFields = schema.childrenSubFields?.length ?? 1
   let answerArray: string[][]
   if (input !== undefined && input.child) {
@@ -198,7 +196,6 @@ const transformToChildOutput = (
   } else {
     answerArray = [Array(noOfChildrenSubFields).fill('')]
   }
-  console.log('transformToChildOutput answerArray: ', answerArray)
   return {
     ...pickBaseOutputFromSchema(schema),
     answerArray,
--- a/frontend/src/features/public-form/utils/inputTransformation.ts
+++ b/frontend/src/features/public-form/utils/inputTransformation.ts
@@ -191,7 +191,7 @@ const transformToChildOutput = (
 ): ChildBirthRecordsResponse => {
   const noOfChildrenSubFields = schema.childrenSubFields?.length ?? 1
   let answerArray: string[][]
-  if (input !== undefined && input.child) {
+  if (input?.child) {
     answerArray = input.child
   } else {
     answerArray = [Array(noOfChildrenSubFields).fill('')]
--- a/frontend/src/features/public-form/utils/inputTransformation.ts
+++ b/frontend/src/features/public-form/utils/inputTransformation.ts
@@ -187,11 +187,21 @@ const transformToSectionOutput = (
 
 const transformToChildOutput = (
   schema: ChildrenCompoundFieldSchema,
-  input: ChildrenCompoundFieldValues,
+  input?: ChildrenCompoundFieldValues,
 ): ChildBirthRecordsResponse => {
+  console.log('transformToChildOutput: ', input)
+  console.log('transformToChildOutput schema: ', schema)
+  const noOfChildrenSubFields = schema.childrenSubFields?.length ?? 1
+  let answerArray: string[][]
+  if (input !== undefined && input.child) {
+    answerArray = input.child
+  } else {
+    answerArray = [Array(noOfChildrenSubFields).fill('')]
+  }
+  console.log('transformToChildOutput answerArray: ', answerArray)
   return {
     ...pickBaseOutputFromSchema(schema),
-    answerArray: input.child,
+    answerArray,
   }
 }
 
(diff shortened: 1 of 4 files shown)


frontend/src/features/public-form/utils/inputTransformation.ts
```diff
@@ -187,11 +187,18 @@ const transformToSectionOutput = (
 
 const transformToChildOutput = (
   schema: ChildrenCompoundFieldSchema,
-  input: ChildrenCompoundFieldValues,
+  input?: ChildrenCompoundFieldValues,
 ): ChildBirthRecordsResponse => {
+  const noOfChildrenSubFields = schema.childrenSubFields?.length ?? 1
+  let answerArray: string[][]
+  if (input !== undefined && input.child) {
```
KenLSM: Nit: Looking at being terser, would `if (input?.child) {` be sufficient?

src/app/utils/field-validation/validators/childrenValidator.ts
```diff
@@ -24,7 +24,9 @@ const childrenAnswerValidator: ChildrenValidator = (response) => {
   const { answerArray } = response
 
   return answerArray.length === 0
-    ? left(`ChildrenValidator:\t Answer is empty array`)
+    ? left(
+        `ChildrenValidator (childrenAnswerValidator):\t Answer is empty array`,
```
KenLSM, line 28: Not very familiar with this part, but is the addition of ` (childrenAnswerValidator)` intended?

src/app/utils/field-validation/validators/childrenValidator.ts
```diff
@@ -47,22 +51,34 @@ const validChildAnswerFirstArray: ChildrenValidator = (response) => {
...
-  )
-    ? right(response)
-    : left(`ChildrenValidator:\t inconsistent answer array subarrays`)
+  const { childSubFieldsArray, answerArray } = response
+  const noOfChildrenSubFields = childSubFieldsArray?.length ?? 1
+  const noChildSelectedAnswerArray = Array(noOfChildrenSubFields).fill('')
+  const first = answerArray[0]
+  return Array.isArray(first) &&
+    first.length > 0 &&
+    first[0] !== noChildSelectedAnswerArray[0]
```
KenLSM: cmiiw, this is to check if `first[0]` is an empty field? If that's the case, does it make sense if we have it as `first[0] !== ''`?
 
I was initially confused with why we are comparing the first elements of both arrays.

frontend/src/features/public-form/utils/inputTransformation.ts
wanlingt: [9089e80](https://github.com/opengovsg/FormSG/pull/6673/commits/9089e8026f129e21341f060d664758b08223276d)

src/app/utils/field-validation/validators/childrenValidator.ts
wanlingt: This is to check if a child is selected. I added more comments in [d01885f](https://github.com/opengovsg/FormSG/pull/6673/commits/d01885f12aa742686227ba20c4dafa4908791a0b), hope it's slightly clearer now!

`first[0]` is the answer from the response, and `noChildSelectedAnswer` is the format that the answer should take if no child is selected (a string of empty strings). We want to check if at least 1 child is selected, before we do the validation.

src/app/utils/field-validation/validators/childrenValidator.ts
wanlingt: I added the function name in the brackets so it's clearer where the error message is thrown from, cos two of the error messages in `childrenValidator` were the same. It's not necessary for this particular one though (since the error message is unique), but left it in there anyway to keep the error messages consistent. wdyt! too messy?

src/app/utils/field-validation/validators/childrenValidator.ts
KenLSM: Got it! 👍

src/app/utils/field-validation/validators/childrenValidator.ts
KenLSM: I see. It was asking in terms of all the validator errors thrown. Thought that it was some debug text, but your reasoning makes sense too.

KenLSM (approved):
LGTM! Thanks @wanlingt

PR #6676, opened 2023-08-30
chore: remove dangling comments

tshuli:
## Problem
- Removes unused comments from #6642


tshuli:
@KenLSM could you please see if this can be added to today's release, or it can go with the next round (just removing unused comments) 🙏

KenLSM (approved):
Let's get this in

PR #6681, opened 2023-08-31
fix: payment quantity modal input inconsistencies

KenLSM:
## Problem

- input value not restoring to default on cancel press
- animation flickers when discarding input (either through backdrop, cancel button, close button)

## Solution

**Input value not restoring to default**
1. cancel press is not captured under `Modal::onClose()` which is where the reset is triggered

Fixed by resetting on modal post-animation, as it captures all onClose events, even if it's triggered manually.

**Animation flickers**
1. `onClose()` triggers the animation
3. subsequently a `resetField` was called, which updates the value **while** the close animation is still being played

Fixed by updating the fields on modal post-animation instead of together with `onClose()`

**Breaking Changes** 


foochifa (approved):
LGTM Thanks @KenLSM! tested the fix on staging-alt too. 

Just wondering if you could remove the unused import `Box`

PR #6684, opened 2023-08-31
fix: update Myinfo occupation list

wanlingt:
## Problem

The Myinfo occupation list is out of date. When a form filler has an occupation that is not in our constants list, their occupation is reflected as a blank answer in our form, and they are unable to submit.

Closes FRM-1305

## Solution

Update the occupation list according to the [latest occupations provided by MOM](https://opengovproducts.slack.com/archives/C048W1372DV/p1691648561045359).

**Breaking Changes** 

## Tests

Regression test
- [ ] On a Myinfo form, add the occupation field
- [ ] Log in as a user that has a verified occupation
- [ ] The occupation should be reflected in the form
- [ ] Submit the form
- [ ] The occupation should be reflected in the form response


KenLSM (approved):
LGTM, that's a whole bucket load of new occupations haha!

wanlingt:
> LGTM, that's a whole bucket load of new occupations haha!

yeah.... hasn't been updated in a long while. I'll open a separate PR to log when we get a new occupation/myinfo field that isn't already in our constants list

wanlingt:
> > LGTM, that's a whole bucket load of new occupations haha!
> 
> yeah.... hasn't been updated in a long while. I'll open a separate PR to log when we get a new occupation/myinfo field that isn't already in our constants list

https://github.com/opengovsg/FormSG/pull/6694

PR #6689, opened 2023-09-04
fix: payments params in new storage submission endpoint

LinHuiqing:
## Problem

There were 2 bugs with the implementation of the new storage endpoint:

1. Joi validation was not set up properly to optionally accept payment params (`paymentProducts`, `paymentReceiptEmail` and `payments`)
2. These same params were not passed on in the `encryptSubmission` middleware.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Joi validation for payments params should only be for storage submissions, and should include `paymentProducts`.
- `encryptSubmission` middleware should pass payment params on from `req.body` to `formsg.encryptedPayload`.

## Tests

- [ ] Submit a payment form with payment by products.
- [ ] Submit a payment form with variable payments.


foochifa (approved):
Tested on staging, thanks for the quick fix @LinHuiqing

PR #6694, opened 2023-09-05
feat: add logging to check for missing Myinfo field values

wanlingt:
## Problem

We want to track when we receive Myinfo field values (for dropdown fields) that are not in our Myinfo constants lists, as this prevents respondents from submitting their form.

Closes FRM-1306

## Solution

Add a check in `myinfo.service.ts`, which sends an error message when a field value for `NATIONALITIES`, `OCCUPATIONS`, `RACES`, `HOUSING_TYPES`, `HDB_TYPES` is not found.
- Other Myinfo fields which this check is not carried out for (because it's unlikely that these definitions will change): Sex, Residential status, Marital status, Workpass status, ChildGender, ChildVaxxStatus)

Added a monitor in DD [here](https://app.datadoghq.com/monitors/130812469), which pings the Slack alerts channel

**Breaking Changes** 

**Improvements**:

- Move field options into separate constants files (`HOUSING_TYPES`, `HDB_TYPES`)

## Tests

- [ ] Missing Myinfo value is captured in logs. This has already been done on staging, can search for the message "Myinfo field value not found in existing list" on Cloudwatch, under `/aws/elasticbeanstalk/formsg-staging/var/log/eb-docker/containers/eb-current-app/stdouterr.log`

Steps that were taken:
- Delete constants from the current myinfo constants files (`NATIONALITIES`, `OCCUPATIONS`, `RACES`, `HOUSING_TYPES`, `HDB_TYPES`)
- Fill in a form with a profile that has the values of the deleted constants in the previous step
- The myinfo fields should show up as blank, and you should see a "Myinfo field value not found in existing list" message for each missing value in Cloudwatch

Regression test
- [ ] Create a Myinfo form with the fields: Race, Nationality, Occupation, Dialect, Birth country, Housing type, HDB type
- [ ] Fill in the form, submit
- [ ] You should be able to submit the form successfully. The fields should be reflected in the email response

### DIFF
--- a/shared/constants/field/myinfo/myinfo-dialects.ts
+++ b/shared/constants/field/myinfo/myinfo-dialects.ts
@@ -146,5 +146,3 @@ export const myInfoDialects = [
   'WENCHOW',
   'YIDDISH',
 ]
-
-export default myInfoDialects
--- a/shared/constants/field/myinfo/myinfo-dialects.ts
+++ b/shared/constants/field/myinfo/myinfo-dialects.ts
@@ -1,4 +1,4 @@
-const myInfoDialects = [
+export const myInfoDialects = [
   'AKAN',
   'ALBANIAN',
   'AMHARIC',
--- a/src/app/modules/myinfo/myinfo.service.ts
+++ b/src/app/modules/myinfo/myinfo.service.ts
@@ -255,7 +255,6 @@ export class MyInfoServiceClass {
     currFormFields: LeanDocument<IFieldSchema[]>,
   ): ResultAsync<PossiblyPrefilledField[], MyInfoHashingError | DatabaseError> {
     const allChildAttrs: InternalAttr[] = []
-
     const prefilledFields = currFormFields.map((field) => {
       const myInfoAttr = getMyInfoAttr(field)
       // Children field prefilling is handled by the frontend.
--- a/src/app/modules/myinfo/myinfo.service.ts
+++ b/src/app/modules/myinfo/myinfo.service.ts
@@ -62,10 +62,10 @@ import {
   getMyInfoAttr,
   getMyInfoAttributeConstantsList,
   hashFieldValues,
-  isFieldValueInMyinfoList,
   isMyInfoChildrenBirthRecords,
   isMyInfoLoginCookie,
   isMyInfoRelayState,
+  logIfFieldValueNotInMyinfoList,
   validateMyInfoForm,
 } from './myinfo.util'
 import getMyInfoHashModel from './myinfo_hash.model'
@@ -277,7 +277,11 @@ export class MyInfoServiceClass {
       if (fieldValue) {
         const myInfoConstantsList = getMyInfoAttributeConstantsList(myInfoAttr)
         if (myInfoConstantsList) {
-          isFieldValueInMyinfoList(fieldValue, myInfoAttr, myInfoConstantsList)
+          logIfFieldValueNotInMyinfoList(
+            fieldValue,
+            myInfoAttr,
+            myInfoConstantsList,
+          )
         }
       }
 
--- a/src/app/modules/myinfo/myinfo.service.ts
+++ b/src/app/modules/myinfo/myinfo.service.ts
@@ -281,6 +281,7 @@ export class MyInfoServiceClass {
         myInfoAttr as InternalAttr,
       )
 
+      // Add logging to check if Myinfo field value exists in our constants lists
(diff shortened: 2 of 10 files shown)


src/app/modules/myinfo/myinfo.service.ts
```diff
@@ -270,6 +280,61 @@ export class MyInfoServiceClass {
...
+        } else if (myInfoAttr === MyInfoAttribute.HousingType) {
+          isFieldValueInMyinfoListCheck(
+            fieldValue,
+            myInfoAttr,
+            myInfoHousingTypes,
+          )
+        } else if (myInfoAttr === MyInfoAttribute.HdbType) {
+          isFieldValueInMyinfoListCheck(fieldValue, myInfoAttr, myInfoHdbTypes)
+        }
+      }
```
foochifa, line 286: Should we log an error as a 'default' case?  Or use switch-cases?

src/app/modules/myinfo/myinfo.service.ts
```diff
@@ -270,6 +280,61 @@ export class MyInfoServiceClass {
...
+              action: 'prefillAndSaveMyInfoFields',
+              myInfoFieldValue: fieldValue,
+              myInfoAttr,
+            },
+          })
+        }
+      }
+
+      // Check if field value exists in our constants lists
+      if (fieldValue) {
```
foochifa, line 277: Also, was just wondering if we could do this better... like maybe a key value map of myinfoattr -> myinfolist

hence we just query the key value mapping instead of doing this long chain of checks

shared/constants/field/myinfo/myinfo-dialects.ts
```diff
@@ -1,4 +1,4 @@
-const myInfoDialects = [
+export const myInfoDialects = [
```
foochifa, line 1: If we are exporting all these, then we have no need to export default below right?

foochifa (commented):
Just some comments and suggestions!

shared/constants/field/myinfo/myinfo-dialects.ts
wanlingt: oops yes! fixed in [ccae981](https://github.com/opengovsg/FormSG/pull/6694/commits/ccae981beabcb7d7de72daa26776b2e2bb38e9eb)

src/app/modules/myinfo/myinfo.service.ts
wanlingt: done here! [83cb27a](https://github.com/opengovsg/FormSG/pull/6694/commits/83cb27af4977997df567bec80d98593cf73c1f54)

src/app/modules/myinfo/myinfo.service.ts
wanlingt: won't log errors in default cases (as there aren't any errors to log), refactored here [83cb27a](https://github.com/opengovsg/FormSG/pull/6694/commits/83cb27af4977997df567bec80d98593cf73c1f54) :)

src/app/modules/myinfo/myinfo.service.ts
```diff
@@ -270,6 +272,15 @@ export class MyInfoServiceClass {
       const { fieldValue, isReadOnly } = myInfoData.getFieldValueForAttr(
         myInfoAttr as InternalAttr,
       )
+
+      // Check if field value exists in our constants lists. If it doesn't, log the error
+      if (fieldValue) {
+        const myInfoConstantsList = getMyInfoAttributeConstantsList(myInfoAttr)
+        if (myInfoConstantsList) {
+          isFieldValueInMyinfoList(fieldValue, myInfoAttr, myInfoConstantsList)
```
foochifa: This function actually checks and sends the logger if there is an error right. Perhaps we can make the name more specific?

`logIfFieldValueNotInMyinfoList`

Or at least identify that this method does the logging!

foochifa (approved):
LGTM Thanks for making the refactors @wanlingt, just a small nit on method naming

src/app/modules/myinfo/myinfo.service.ts
wanlingt: yes ok! good suggestion. renamed it to the suggested `logIfFieldValueNotInMyinfoList` here [f8915cd](https://github.com/opengovsg/FormSG/pull/6694/commits/f8915cdbc88e8088d27bfbdee523afaa004a75b4)

PR #6696, opened 2023-09-05
feat: soften and fix storage submission validation

LinHuiqing:
## Problem

Backend storage submission validation is currently hard but might be buggy / risky as it was done in a huge PR (https://github.com/opengovsg/FormSG/pull/6595) of +1,479 / −1,022. As we have frontend validation that is already stable, it is an option to temporarily only perform a soft validation on the backend.

We should harden our backend validation after logged validation error rates are the similar to that of email mode submissions. A ticket has been created for this: FRM-1318.

This PR fixes the [bug that prevented us from rolling out the new storage submission endpoint](https://opengovproducts.slack.com/archives/CLJPA6NDN/p1693885396583929?thread_ts=1693884652.604959&cid=CLJPA6NDN) too, which was caused by the responses being wrongly validated again in the `encryptSubmission` middleware. This should only be done for the old submission endpoint.

Closes FRM-1333

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Create `validateEncryptSubmission` middleware so that validation run in `IncomingEncryptSubmission.init` is only run for the old storage submission endpoint.

**Improvements**:

- Moves on to next middleware even if submission validation throws error.
- Use [EncryptionBoundaryShift-Production cloudwatch dashboard](https://ap-southeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-southeast-1#dashboards/dashboard/EncryptionBoundaryShift-Production) to track any errors on the backend storage submission validation middleware.
- Introduction of `req.formsg.filteredResponses` to pass filteredResponses between middlewares (instead of filtering when we need to).

## Tests

Make sure that submission still goes through if server validation doesn't pass, and make sure that the cloudwatch dashboard works:
- [ ] Deploy the app to a staging env.
- [ ] On growthbook, ensure that `encryption-boundary-shift` will be `true` the form you create (ie forced value or 100% rollout).
- [ ] In the `featureflags` DB collection, ensure that`encryption-boundary-shift-hard-validation` is unset or set to `false`.
- [ ] Make an invalid submission to the new storage submission endpoint by cURL. You should be able to submit successfully.
- [ ] The submission should show up in the form's admin results page.
- [ ] Check your logs, there should be error logs with `Error processing responses, but proceeding with submission as submission have been validated client-side`.
- [ ] Go to [EncryptionBoundaryShiftStaging cloudwatch dashboard](https://ap-southeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-southeast-1#dashboards/dashboard/EncryptionBoundaryShift-Staging), you should see the recent errored BE validation show up on the graph. The logs might take ~5min to appear.
- [ ] Check that all the instances are still healthy and didn't crash.

Make sure that the `encryption-boundary-shift-hard-validation` feature flag works:
- [ ] On growthbook, ensure that `encryption-boundary-shift` will be `true` the form you create (ie forced value or 100% rollout).
- [ ] In the `featureflags` DB collection, set `encryption-boundary-shift-hard-validation` to `true`.
- [ ] Make an invalid submission to the new storage submission endpoint by cURL. There should be a validation error shown to the respondent, preventing submission.
- [ ] The submission should **not** show up in the form's admin results page.
- [ ] Check that all the instances are still healthy and didn't crash.

Regression:
- [ ] Submit a storage mode form on the old endpoint for the following test cases:
  - [ ] Required table field hidden by logic.
  - [ ] Attachment uploaded submission.
  - [ ] Mobile and email fields:
    - [ ] Verifiable and filled
    - [ ] Required and hidden by logic
  - [ ] Fields with answer array (e.g. checkbox).
- [ ] Submit a storage mode form on the new endpoint for the above test cases.
- [ ] Submit an email mode form for the above test cases.

## Deploy Notes

1. Before deployment, in the DB's `featureflags` collection, ensure that `encryption-boundary-shift-hard-validation` is unset or set as false.
2. After deployment, monitor:
    1. "Submissions" section in the [FormSG DD monitor](https://app.datadoghq.com/dashboard/b3u-egb-apu/form-s-g?refresh_mode=sliding&from_ts=1694393221756&to_ts=1694479621756&live=true) - to ensure that email and old storage submission endpoint is still doing well.
3. Once satisfied that there are no regressions, on growthbook, increase rollout % for new storage submission endpoint to 1%.
4. Monitor:
    1. "Encryption Boundary Shift" in [FormSG Release Dashboard DD monitor](https://app.datadoghq.com/dashboard/fh2-pvk-ncb/formsg-release-dashboard?refresh_mode=sliding&from_ts=1694393504503&to_ts=1694479904503&live=true).
    2. [EncryptionBoundaryShift-Production cloudwatch dashboard](https://ap-southeast-1.console.aws.amazon.com/systems-manager/resource-groups/cloudwatch?region=ap-southeast-1&dashboard=EncryptionBoundaryShift-Production). It's okay if stuff is logged here, because these are warnings logged by the _soft_ validation and email mode still has BE validation errors as well! What we're aiming for is an error rate similar to that of email mode.

**New feature flags**:
- In the `featureflags` collection - `encryption-boundary-shift-hard-validation`: the backend uses this to decide whether hard validation is used for the new submission endpoint.

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -42,6 +42,7 @@ import {
   ValidateSubmissionMiddlewareHandlerRequest,
 } from './encrypt-submission.types'
 import { mapRouteError } from './encrypt-submission.utils'
+import IncomingEncryptSubmission from './IncomingEncryptSubmission.class'
 
 export const logger = createLoggerWithLabel(module)
 
@@ -164,7 +165,7 @@ export const checkNewBoundaryEnabled = async (
   return next()
 }
 
-export const validateSubmission = async (
+export const validateStorageSubmission = async (
   req: ValidateSubmissionMiddlewareHandlerRequest,
   res: Parameters<StorageSubmissionMiddlewareHandlerType>[1],
   next: NextFunction,
@@ -203,7 +204,7 @@ export const validateSubmission = async (
           }
         }
       }
-      req.formsg.filteredResponses = responses
+      // req.formsg.filteredResponses = responses
       return next()
     })
     .mapErr((error) => {
@@ -331,7 +332,8 @@ export const encryptSubmission = async (
 
   req.formsg.encryptedPayload = {
     attachments: encryptedAttachments,
-    responses: req.formsg.filteredResponses,
+    // filteredResponses might not have been loaded due to validation errors, default to req.body.responses
+    responses: req.formsg.filteredResponses ?? req.body.responses,
     encryptedContent,
     version: req.body.version,
     paymentProducts: req.body.paymentProducts,
@@ -356,6 +358,37 @@ export const moveEncryptedPayload = async (
   }
 
   req.formsg.encryptedPayload = req.body
+  return next()
+}
+
+/**
+ * Moves encrypted payload present in req.body to req.formsg.encryptedPayload.
+ * Should only be used for the old storage mode submission endpoint (/api/v3/forms/:formId/submissions/encrypt).
+ */
+export const validateEncryptSubmission = async (
+  req: EncryptSubmissionMiddlewareHandlerRequest,
+  res: Parameters<EncryptSubmissionMiddlewareHandlerType>[1],
+  next: NextFunction,
+) => {
+  const form = req.formsg.encryptedFormDef
+  const responses = req.body.responses
+  const encryptedContent = req.body.encryptedContent
+
+  const incomingSubmissionResult = IncomingEncryptSubmission.init(
+    form,
+    responses,
+    encryptedContent,
+    req.formsg.featureFlags.includes(featureFlags.encryptionBoundaryShift),
+  )
+  if (incomingSubmissionResult.isErr()) {
+    const { statusCode, errorMessage } = mapRouteError(
+      incomingSubmissionResult.error,
+    )
+    return res.status(statusCode).json({
+      message: errorMessage,
+    })
+  }
+
   return next()
 }
 
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -188,11 +188,7 @@ export const validateStorageSubmission = async (
     formDef.responseMode,
   )
     .andThen(() =>
-      ParsedResponsesObject.parseResponses(
-        formDef,
-        req.body.responses,
-        req.formsg.featureFlags.includes(featureFlags.encryptionBoundaryShift),
-      ),
+      ParsedResponsesObject.parseResponses(formDef, req.body.responses),
     )
     .map((parsedResponses) => {
       const responses = [] as EncryptFormFieldResponse[]
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -384,7 +384,6 @@ export const validateEncryptSubmission = async (
     form,
     responses,
     encryptedContent,
-    req.formsg.featureFlags.includes(featureFlags.encryptionBoundaryShift),
   )
   if (incomingSubmissionResult.isErr()) {
     logger.error({
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -195,7 +195,7 @@ export const validateSubmission = async (
       // TODO(FRM-1318): Set DB flag to true to harden submission validation after validation has similar error rates as email mode forms.
       if (
         req.formsg.featureFlags.includes(
-          featureFlags.encryptionBoundaryShiftValidation,
+          featureFlags.encryptionBoundaryShiftHardValidation,
         )
       ) {
         const { statusCode, errorMessage } = mapRouteError(error)
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -186,23 +186,28 @@ export const validateSubmission = async (
     )
     .map(() => next())
     .mapErr((error) => {
-      logger.error({
-        message:
-          'Error processing responses, but proceeding with submission as submission have been validated client-side',
-        meta: logMeta,
-        error,
-      })
       // TODO(FRM-1318): Set DB flag to true to harden submission validation after validation has similar error rates as email mode forms.
       if (
         req.formsg.featureFlags.includes(
           featureFlags.encryptionBoundaryShiftHardValidation,
         )
       ) {
+        logger.error({
+          message: 'Error processing responses',
+          meta: logMeta,
+          error,
+        })
         const { statusCode, errorMessage } = mapRouteError(error)
         return res.status(statusCode).json({
           message: errorMessage,
         })
       }
+      logger.warn({
+        message:
+          'Error processing responses, but proceeding with submission as submission have been validated client-side',
+        meta: logMeta,
+        error,
+      })
       return next()
     })
 }
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -173,9 +173,9 @@ export const validateStorageSubmission = async (
   const formDef = req.formsg.formDef
 
   const logMeta = {
-    action: 'validateSubmission',
+    action: 'validateStorageSubmission',
     ...createReqMeta(req),
-    formId: formDef.id,
+    formId: formDef._id.toString(),
   }
 
   // Validate submission
@@ -204,7 +204,7 @@ export const validateStorageSubmission = async (
           }
         }
       }
-      // req.formsg.filteredResponses = responses
+      req.formsg.filteredResponses = responses
       return next()
     })
     .mapErr((error) => {
@@ -230,6 +230,7 @@ export const validateStorageSubmission = async (
         meta: logMeta,
         error,
       })
+      req.formsg.filteredResponses = req.body.responses
       return next()
     })
 }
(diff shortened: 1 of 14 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -187,14 +187,17 @@ export const validateSubmission = async (
...
+        message:
+          'Error processing responses, but proceeding with submission as submission have been validated client-side',
         meta: logMeta,
         error,
       })
-      const { statusCode, errorMessage } = mapRouteError(error)
-      return res.status(statusCode).json({
-        message: errorMessage,
-      })
+      // TODO(FRM-1318): Uncomment to harden submission validation after validation has similar error rates as email mode forms.
```
KenLSM: Curious as to why we're commenting out instead of using growthbook? Is it because we haven't integrated with growthbook on the BE side of things?

KenLSM (approved):
LGTM, had a question, but non-blocking

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Ah, yes because we don't have growthbook on BE yet. And I'm hoping storage submission validation will be settled pretty soon...
But I presume I can hide this with a BE feature flag, if that makes more sense 🤔

LinHuiqing: Done in [5b85e4d](https://github.com/opengovsg/FormSG/pull/6696/commits/5b85e4dc5561da26f51a9f95ce56b53cbd822f1d) and [300d44a](https://github.com/opengovsg/FormSG/pull/6696/commits/300d44aaa08fdebbbac79ed86cfb5cfc44fdc62e)!

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -187,14 +187,23 @@ export const validateSubmission = async (
     .map(() => next())
     .mapErr((error) => {
       logger.error({
-        message: 'Error processing responses',
+        message:
+          'Error processing responses, but proceeding with submission as submission have been validated client-side',
```
KenLSM: This log might need to be moved to below the `if-block`, but before we `return next()` since technically we **are** blocking instead of proceeding.

LinHuiqing: Good point, logging improved in [4f2784d](https://github.com/opengovsg/FormSG/pull/6696/commits/4f2784d53cc3ad46cb1e54a85fe31f292c3a8fae)!

LinHuiqing:
Hold up, this doesn't actually soften the validation for some reason 😢 the stuff isn't logged, and the response is returned earlier for some reason.

LinHuiqing:
Okay so after investigation, it's because the `IncomingEncryptSubmission.init` also validates the fields again, which in the case of the old endpoint, will just be for the mobile and email fields. This validation is no longer required for the new endpoint as it can simply share what email mode is using.
Since the softening of storage submission validation will not work without the fix, the fix is included here. This solves the validator bug seen when we attempted to rollout too. I'll update the description and tests in a bit.

LinHuiqing:
Ready for review again! Up on https://staging-alt2.form.gov.sg

LinHuiqing:
Side note: I was looking at past PRs and saw this comment on how there were intentions to move away from `parseResponses` as it is tech debt more than 2 years back: https://github.com/opengovsg/FormSG/pull/1986#issuecomment-851254824
No wonder some of the code was migrated but not others 😅 And no wonder I had to spend so much time wresting `parseResponses` and related code oop

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -347,6 +361,54 @@ export const moveEncryptedPayload = async (
   return next()
 }
 
+/**
+ * Moves encrypted payload present in req.body to req.formsg.encryptedPayload.
+ * Should only be used for the old storage mode submission endpoint (/api/v3/forms/:formId/submissions/encrypt).
+ */
+export const validateEncryptSubmission = async (
+  req: EncryptSubmissionMiddlewareHandlerRequest,
+  res: Parameters<EncryptSubmissionMiddlewareHandlerType>[1],
```
tshuli: It took me a while to understand this comment 😅 

> Okay so after investigation, it's because the IncomingEncryptSubmission.init also validates the fields again, which in the case of the old endpoint, will just be for the mobile and email fields. This validation is no longer required for the new endpoint as it can simply share what email mode is using.
Since the softening of storage submission validation will not work without the fix, the fix is included here. This solves the validator bug seen when we attempted to rollout too. I'll update the description and tests in a bit.

Would be useful to add in the documentation here that:
- This middleware also validates mobile and email field via `IncomingEncryptSubmission.init`
- This applies only to old e2ee encrypt mode submission because storage mode submissions has another middleware `validateStorageSubmission` which handles validation (on that note, do add docs for `validateStorageSubmission`, and explain that it is the same validation as for email mode!)

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -159,17 +165,17 @@ export const checkNewBoundaryEnabled = async (
   return next()
 }
 
-export const validateSubmission = async (
-  req: StorageSubmissionMiddlewareHandlerRequest,
+export const validateStorageSubmission = async (
```
tshuli, line 172: to add JSDocs here

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -184,17 +190,48 @@ export const validateSubmission = async (
         req.formsg.featureFlags.includes(featureFlags.encryptionBoundaryShift),
```
tshuli: In the preceeding middleware `EncryptSubmissionMiddleware.checkNewBoundaryEnabled`, we already return an error status if `encryptionBoundaryShift` feature flag is not enabled. So `req.formsg.featureFlags.includes(featureFlags.encryptionBoundaryShift)` should always be true (same as what we do in L184 of `email-submission.controller`)

Mainly coming from the view that passing in `req.formsg.featureFlags.includes(featureFlags.encryptionBoundaryShift)` is a bit untidy; if we mistakenly disable `encryptionBoundaryShift` feature flag but the response gets to this middleware, then the storage mode submission will be processed as if it were an encrypted mode submission, and there may be unexpected outcomes because the submission shape is not what we expect, but this should not even happen.

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -184,17 +190,48 @@ export const validateSubmission = async (
...
       ),
     )
-    .map(() => next())
+    .map((parsedResponses) => {
+      const responses = [] as EncryptFormFieldResponse[]
+      for (const response of parsedResponses.getAllResponses()) {
+        if (response.isVisible) {
```
tshuli, line 196: ~qn: why is this check needed here and not in `email-submission.controller` (L181)? How does the email-submission controller deal with hidden fields~

Reviewed further against `createSubmission.ts` on the frontend and is consistent with current behavior

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -319,7 +333,7 @@ export const encryptSubmission = async (
 
   req.formsg.encryptedPayload = {
     attachments: encryptedAttachments,
-    responses: filteredResponses.value as EncryptFormFieldResponse[],
+    responses: req.formsg.filteredResponses,
```
tshuli, line 336: 🤔 was the intent here originally to check if the array is empty? (empty array is truthy so this check is indeed redundant)

tshuli (commented):
Thanks @LinHuiqing! Just minor comments / clarifications

Will do another round of test on Monday before approving 😃

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Yea I suspect this is part of why we wanted to migrate to using `IncomingEncryptSubmission`, so that encapsulate all mutations required in the class itself, as described in the description of the PR https://github.com/opengovsg/FormSG/pull/1987. Not sure if we still want to continue this journey started 2 years back though? It does seem like a worthy investment if something similar to this encryption boundary shift is possible again 🤔 If not, I think we might have bigger fish to fry haha

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Ah, so sorry about the comment! It was copied over from the `moveEncryptedPayload` middleware and I forgot to change it.

Updated in [511413d](https://github.com/opengovsg/FormSG/pull/6696/commits/511413db4d9aadda5d229122c869ed5dbf995355)!

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Added in [511413d](https://github.com/opengovsg/FormSG/pull/6696/commits/511413db4d9aadda5d229122c869ed5dbf995355), thanks!

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: True, thanks! I've removed the redundant check in [731a04b](https://github.com/opengovsg/FormSG/pull/6696/commits/731a04b015acdc300add49fdb2a6ad7ad8a799b7). 

I've done some refactoring to remove the `encryptionBoundaryShiftEnabled` param from `parseResponses` altogether as it is only used for processing submissions in the clear. Now `getFilteredResponses` simply takes in a param `isEncryptedMode` to get the right filters.

LinHuiqing:
> Will do another round of test on Monday before approving 😃

@tshuli ready for review again btw, thanks!! :)

tshuli (approved):
thanks @LinHuiqing, lgtm! can consider cutting a release for this tomorrow :)

PR #6697, opened 2023-09-07
fix: remove index from myinfo child question name

wanlingt:
## Problem

If an email response has more than 1 Myinfo Child field, the index of the child field stays at 1. For example:
image

## Solution

As a quick fix, remove the index.

A longer-term fix where the index is incremented accurately will be opened in a separate PR (FRM-1325)

**Breaking Changes** 

    - Respondents who are already collecting Myinfo responses will have the addition of an extra column (due to the new question name) when they use the data collation tool to generate a CSV. I've checked and the form admins currently using the Myinfo child field are ones that we're already in touch with, and so they are aware of the changes / have not gone live with their forms

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] On a Myinfo-enabled form, add a Myinfo child field
- [ ] Submit the form
- [ ] In the form response email, the Myinfo child fields should not contain any number in the field name. They should look like this naming convention `[child] [MyInfo] Child.childname`, where `childname` is the field type

### DIFF
--- a/src/app/modules/submission/email-submission/email-submission.util.ts
+++ b/src/app/modules/submission/email-submission/email-submission.util.ts
@@ -234,7 +234,7 @@ export const getAnswersForChild = (
         childName,
       ),
       fieldType: response.fieldType,
-      question: `Child-${childIdx + 1}.${subFields[idx]}`,
+      question: `Child.${subFields[idx]}`,
       myInfo: {
         attr: subFields[idx] as unknown as MyInfoAttribute,
       },


foochifa (approved):
Thanks @wanlingt, maybe we can make a linear task to add the index back in?

wanlingt:
> Thanks @wanlingt, maybe we can make a linear task to add the index back in?

Yes! FRM-1325. Also edited the PR description :)

PR #6701, opened 2023-09-07
fix: invalid mixed digit input

KenLSM:
## Problem

Two digit inputs are incorrect

Closes FRM-1331

## Solution

**Breaking Changes** 

## Before & After Screenshots

| Before | After |
|--------|--------|
| Screenshot 2023-09-07 at 9 44 28 PM |  Screenshot 2023-09-07 at 9 44 21 PM|
| Screenshot 2023-09-07 at 9 46 05 PM |  Screenshot 2023-09-07 at 9 45 48 PM |

## Tests

Note for all the tests:
1. Manually change the input values in order to have them stored as `string`
2. Both min and max **needs to be changed**
3. String comparison will only occur **if both operands** are `string`

Regression
On the product modal
- [ ] Ensure that `min: 1, max: 10` passes
- [ ] Ensure that `min: 10, max: 1` fails

Cases from reported example
- [ ] Ensure that `min: 59, max: 100` passes
- [ ] Ensure that `min: 10, max: 9` fails

Mixed digit strings (values should be compared as numbers instead of string)
- [ ] Ensure that `min: 2, max: 10` passes // strcmp would have failed as "2" is smaller than "1"
- [ ] Ensure that `min: 10, max: 2` fails // strcmp would have passed as "1" is smaller than "2"
- [ ] Ensure that `min: 111, max: 20` fails // strcmp would have passed as "11" is larger than "20"
- [ ] Ensure that `min: 123, max: 1200` passes // strcmp would have failed as "123" is larger than "120"

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
@@ -43,8 +43,8 @@ const DISPLAY_AMOUNT_KEY = 'display_amount'
 const MULTI_QTY_KEY = 'multi_qty'
 
 const parseIntElseNull = (val: string) => {
-  const parsed = parseInt(val, 10)
-  return Number.isNaN(parsed) ? null : parsed
+  const parsedInt = parseInt(val, 10)
+  return Number.isNaN(parsedInt) ? null : parsedInt
 }
 
 export const ProductModal = ({
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
@@ -41,6 +41,12 @@ const MIN_QTY_KEY = `min_qty`
 const MAX_QTY_KEY = `max_qty`
 const DISPLAY_AMOUNT_KEY = 'display_amount'
 const MULTI_QTY_KEY = 'multi_qty'
+
+const parseIntElseNull = (val: string) => {
+  const parsed = parseInt(val, 10)
+  return Number.isNaN(parsed) ? null : parsed
+}
+
 export const ProductModal = ({
   onClose,
   onSaveProduct,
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
```diff
@@ -41,6 +41,12 @@ const MIN_QTY_KEY = `min_qty`
 const MAX_QTY_KEY = `max_qty`
 const DISPLAY_AMOUNT_KEY = 'display_amount'
 const MULTI_QTY_KEY = 'multi_qty'
+
+const parseIntElseNull = (val: string) => {
+  const parsed = parseInt(val, 10)
```
foochifa: nit: how about `parsedInt` as the var name

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
```diff
@@ -41,6 +41,12 @@ const MIN_QTY_KEY = `min_qty`
...
+  return Number.isNaN(parsed) ? null : parsed
```
foochifa: Just wondering but is NaN truthy? if not then instead of this method we could just `parseInt` then check truthyness right?

foochifa (approved):
LGTM! Tested! Thanks for fixing it so quickly @KenLSM, a couple of small comments

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
KenLSM: `!!parseInt('0', 10)` would be falsy, which might not be what the caller expects 🤔.

As of now, we are using the falsy-ness of it, but it might be possible that in other usecase, we might want to do null checks to show some text.

`if (parseIntElseNull(val) === null) { return 'Please enter a number' }`

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
KenLSM: 8a70e1f8a

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/ProductModal.tsx
foochifa: Oh okie! Thanks for the explanation @KenLSM

PR #6712, opened 2023-09-12
fix: mobile view of unsaved modal

foochifa:
## Problem

@staceytan1998 found that our dirty modal `UnsavedChangesModal.tsx` has incorrect sizing in mobile view

## Solution

Similar to other modals, use `chakra-ui/react.useBreakpointValue`.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

On mobile mode
- [ ] build a form
- [ ] make some changes to a field
- [ ] try to navigate away without saving or canceling
- [ ] the modal should fill the entire window
- [ ] return to edit and save
- [ ] form field should be updated

On desktop mode (regression)
- [ ] build a form
- [ ] make some changes to a field
- [ ] try to navigate away without saving or canceling
- [ ] dirty modal should pop up
- [ ] do not make the changes
- [ ] form should be as before

### DIFF
--- a/frontend/src/templates/NavigationPrompt/UnsavedChangesModal.tsx
+++ b/frontend/src/templates/NavigationPrompt/UnsavedChangesModal.tsx
@@ -9,6 +9,7 @@ import {
   ModalOverlay,
   ModalProps,
   Stack,
+  useBreakpointValue,
 } from '@chakra-ui/react'
 
 import { useIsMobile } from '~hooks/useIsMobile'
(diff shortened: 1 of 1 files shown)


foochifa:
Side thoughts: we use `const modalSize = useBreakpointValue({base: 'mobile', xs: 'mobile', md: '...'})` a lot, maybe it's worth to make it a util?

wanlingt (approved):
lgtm thanks! re having this as a util function for modals, yeah makes sense. if you feel like doing some refactoring in your free time :D

PR #6715, opened 2023-09-13
fix: add check for childname when populating field value

wanlingt:
## Problem

MyInfo returns the NRIC for children above the age of 21, and does not return any other field values. We want to define children as below 21 years old.

Closes FRM-1300

## Solution

- Update the copy in the children field details to reflect that child fields are for children below 21 years old
- Check if `childname` is present in the MyInfo records. If it is not, return an empty string for all other children myinfo field values. (Note: prefilling for MyInfo child fields is done in the frontend)

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
before

**AFTER**:
after

## Tests

Regression test
- [x] On a MyInfo child form with the property `Birth Certification Number`, log in as a user with at least 1 child below the age of 21. 
- [x] Select that child, and submit the form.
- [x] The form should be submitted successfully
- [x] The form response should submit that single child

Unfortunately we're not able to test this on staging as the existing Mockpass profiles do not have children above 21yo. 
Hence, **please perform this test once it has been deployed to production**:
- [ ] On a MyInfo child form with the property `Birth Certification Number`, log in as a user with at least 1 child above the age of 21. 
- [ ] Click around the MyInfo child fields. Nothing should appear in the Birth Cert Number field.

### DIFF
--- a/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
+++ b/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
@@ -246,6 +246,14 @@ const ChildrenBody = ({
     return allChildren.filter((name) => !temp.has(name))
   }, [myInfoChildrenBirthRecords, allChildren, allSelectedNames])
 
+  const childNameValues = useMemo(() => {
+    return [childName, ...namesNotSelected()].filter((name) => {
+      if (name === '' || name === undefined) {
+        return false
+      } else return true
+    })
+  }, [childName, namesNotSelected])
+
   const indexOfChild: number = useMemo(() => {
     return (
       myInfoChildrenBirthRecords?.[MyInfoChildAttributes.ChildName]?.indexOf(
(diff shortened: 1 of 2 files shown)


frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -271,6 +271,14 @@ const ChildrenBody = ({
       ) {
         return ''
       }
+      // We use the childname to indicate if there is a child below 21.
+      // If the childname is an empty string, it means there is no child below 21.
+      // As our definition of child in FormSG means child below 21, we want to
+      // return empty strings for other child attributes even if their value is populated by myinfo
+      // if there is no childname.
+      if (myInfoChildrenBirthRecords.childname?.[0] === '') {
```
foochifa: Why `0` and not `indexOfChild`? the array index at the end is for the specific child right? Or am I not getting the shape of the myInfoChildren correct 😅

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -271,6 +271,14 @@ const ChildrenBody = ({
...
+        return ''
```
foochifa: Can we move this check above to before `result` is computed?

wanlingt: yes! done in [e35bc74](https://github.com/opengovsg/FormSG/pull/6715/commits/e35bc74e517f0eb7a00dc697ceda07bd3b5e35a1)

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
wanlingt: Yup you're right. Previously I'd only considered the case where a parent only had children above 21yo (so the first childname would definitely be `''`), but you're right that this should be according to `indexOfChild` because this assumption might not be true. Made the change in [e35bc74](https://github.com/opengovsg/FormSG/pull/6715/commits/e35bc74e517f0eb7a00dc697ceda07bd3b5e35a1).

When checking another case (1 child below 21, 1 child above 21), I realised the one above 21 still showed up in the child names selection as a blank dropdown, so I added a `useMemo` for `childNameValues` in that commit as well!

foochifa: Thanks for the additional considerations @wanlingt, man do we have a way to test this HAHAHAHA are there anyone with kids above and below 21 😂

foochifa (approved):
Thanks @wanlingt LGTM tested regression locally

PR #6729, opened 2023-09-17
fix: remove focus from myinfo child field when form is opened

wanlingt:
## Problem

When a form is opened, the focus immediately goes to the MyInfo child field. The focus should not be on any form field.

Closes FRM-1323

## Solution

To initialise the MyInfo child field, specify the focusOption `{shouldFocus: false}` in `append`. ([Docs here](https://react-hook-form.com/docs/usefieldarray)).

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

https://github.com/opengovsg/FormSG/assets/56983748/7b6b2811-6319-4638-a630-e76583f0fcfe

**AFTER**:

https://github.com/opengovsg/FormSG/assets/56983748/c57ced59-cb46-40d8-926b-7d28a5cf792c

## Tests

1. MyInfo child field - 1 child only
- [ ] Open up a MyInfo child form with 1 child field
- [ ] The focus should not be on the child field, unless the field is clicked
- [ ] Select a child
- [ ] Submit the form
- [ ] Check that the email response contains the child's info

2. MyInfo child field - 2 children only (same test as above)
- [ ] Open up a MyInfo child form with 2 children fields
- [ ] The focus should not be on the child field, unless the field is clicked
- [ ] Select a child
- [ ] Submit the form
- [ ] Check that the email response contains the children's info

3. MyInfo child field - 1 child only with option to add children
- [ ] Enable the option "Allow respondent to add multiple children" on a MyInfo child form
- [ ] Open up a MyInfo child form with 1 child field
- [ ] The focus should not be on the child field, unless the field is clicked
- [ ] Select a child
- [ ] Add another child
- [ ] Select the 2nd child
- [ ] Submit the form
- [ ] Check that the email response contains the children's info

### DIFF
--- a/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
+++ b/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
@@ -75,12 +75,11 @@ export const ChildrenCompoundField = ({
     name: schema._id,
   })
 
-  const { fields, append, remove } = useFieldArray<ChildrenCompoundFieldInputs>(
-    {
+  const { fields, append, update, remove } =
+    useFieldArray<ChildrenCompoundFieldInputs>({
       control: formContext.control,
       name: `${schema._id}.child`,
-    },
-  )
+    })
 
   useEffect(() => {
     if (schema.childrenSubFields) {
@@ -94,9 +93,9 @@ export const ChildrenCompoundField = ({
   // Initialize with a single child section
   useEffect(() => {
     if (!fields || !fields.length) {
-      append([''])
+      update(0, '')
     }
-  }, [fields, append])
+  }, [fields, update])
 
   const ariaChildrenDescription = useMemo(() => {
     let description = simplur`This is a children field. There [is|are] ${fields.length} child[|ren].`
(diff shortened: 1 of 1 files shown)


frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -94,9 +93,9 @@ export const ChildrenCompoundField = ({
   // Initialize with a single child section
   useEffect(() => {
     if (!fields || !fields.length) {
-      append([''])
+      update(0, '')
```
KenLSM: Saw that there's another parameter that we can set on append, would this work?

```suggestion
      append([''], { shouldFocus: false })
```

From the docs, update would do a re-render as it unmounts and remounts the component (`<ChildrenBody />`) which might be expensive.

wanlingt: ooh yeah, this works. thanks ken! where did you find the other parameters btw? I see `focusOptions` in the docs, but not what they are

KenLSM: Similarly, I tried to look at the docs, wasn't helpful. Then I looked into the typedocs for `append` and found it after running through a few rabbitholes 😆 .

KenLSM (approved):
LGTM! Thanks wanling!

PR #6733, opened 2023-09-19
feat(payment): invoice through s3

KenLSM:
## Problem

The receiptUrl is obtained from stripe when a payment is successful. However, this url only lasts up to 30 days.

Closes FRM-1093

Relevant eng design [doc](https://www.notion.so/opengov/Engineering-v1-Receipt-Storage-e474a79a1fbd4ae2bbd9ff25558227c7?pvs=4)

## Solution

1. Initial downloads will always be a freshly queried receipt url from stripe
2. We will generate and save the PDf over on S3, and generate a S3 Presigned URL for the user to download from S3. This will help us avoid consuming bandwidth as users will download directly from S3.

The S3 Presigned URL is only valid for a short amount of time, thus the user wouldn’t have permanent access to it.

There are two primary flow when the user tries to download the invoice.

1. Initial: we will generate and upload the receipt over to S3
2. Subsequent: return the S3 Presigned URL

**Flow for initial download**
```
1. On trigger to download receipt
2. Generate Proof of Payment
4. Store into S3
5. Save into `payment` model that pdf exist in s3
6. Generate S3 presigned link
7. Redirect user to presign link
```

**Flow for Subsequent download**
```
1. Generate S3 presigned link
2. Redirect user to presign link
```

**Breaking Changes** 

## Tests

**Regression**
Downloading proof of payments should respond with pdf file
- [ ] Complete a payment on a payment form
  - [ ] Ensure that downloading of invoice will save a PDF file
    - [ ] Ensure that Proof of payment should correctly reflect amount, quantity, and value of items purchased
  - [ ] Ensure that subsequent download of invoice will still save a PDF file
- [ ] Downloading proof of payments should respond within acceptable time limit (<10 seconds)

Disconnected Payment Forms should still allow user to download proof of payment
(record the download link from the test above)
- [ ] Disconnect payment in the Payment Settings Page 
- [ ] Ensure that downloading of invoice will save a PDF file

**New Feature**
- [ ] Ensure that subsequent download of invoice will still save a PDF file
- [ ] `hasReceiptStoredInS3` should be set in the DB for the particular payment

## Deploy Notes

1. Create s3 buckets for payment proof  on `staging`, `staging-alt`, `prod`, `uat`
2. Add config `PAYMENT_PROOF_S3_BUCKET`  to be added on `staging`, `staging-alt`, `prod`, `uat`
4. Update acl to include GET, PUT access on  `staging`, `prod`, `uat` roles (see comment on linear issue)

- `env var` : `PAYMENT_PROOF_S3_BUCKET` to configure the name of payment proof s3 bucket

~This PR is risky, as there were some issues with nginx when it was deployed to staging but unreproducible on staging-alt, staging-alt2, and uat. To be safe, we need to deploy during the low traffic period.~

~To verify that the issue doesn't occur, we will need to trigger the invoice download multiple times, and ensure that we don't see a 5xx from the ec2. Note that our express server isn't the one returning 5xx, but nginx's reverse proxy itself.~


src/app/modules/payments/stripe.service.ts
```diff
@@ -1009,15 +1011,155 @@ export const verifyPaymentStatusWithStripe = (
...
+          paymentId: payment._id,
+          objectPath,
+        },
+        error,
+      })
+      return new PaymentProofUploadS3Error()
+    },
+  )
+    .map(() => {
+      if (payment.completedPayment) {
```
tshuli: qn (probably because of lack of context) : why is the completedPayment check here, should this be at the start of the fn i.e. if the payment hasn't been completed, we should not be storing payment proof in s3?

src/app/modules/payments/stripe.service.ts
```diff
@@ -1009,15 +1011,155 @@ export const verifyPaymentStatusWithStripe = (
     })
 }
 
-export const generatePaymentInvoice = (
+const getPaymentProofObjectPath = (payment: IPaymentSchema) => {
+  return payment.formId + '/' + payment._id + '.pdf'
+}
```
tshuli: /nit: feels like something which belongs in `.utils`

src/app/modules/payments/stripe.service.ts
```diff
@@ -1009,15 +1011,155 @@ export const verifyPaymentStatusWithStripe = (
...
+            paymentId: payment._id,
+            objectPath,
+          },
+          error,
+        })
+        return new PaymentProofUploadS3Error()
+      })
+    })
+    .map(() => {
+      return undefined
```
tshuli: /opinionated: or () => true instead of undefined, as a general convention in the codebase? 😅

src/app/modules/payments/stripe.service.ts
```diff
@@ -1009,15 +1011,155 @@ export const verifyPaymentStatusWithStripe = (
...
+ *
+ * @param {IPaymentSchema} payment the payment object, used to form the object path
+ *
+ * @returns ok(string) which represents the presigned url if no errors are thrown while generating the presigned url
+ * @returns err(InvoicePresignS3Error) if an error is thrown while generating the presigned link
+ */
+const _getPaymentProofPresignedS3Url = (
+  payment: IPaymentSchema,
+): ResultAsync<string, PaymentProofPresignS3Error> => {
+  const dayInSeconds = 24 * 60 * 60
```
tshuli: /nit: this probably could go into some constants file (if it doesn't yet exist)

src/app/modules/payments/stripe.service.ts
```diff
@@ -1009,15 +1011,155 @@ export const verifyPaymentStatusWithStripe = (
...
+      action: '_getPaymentProofPresignedS3Url',
+      paymentId: payment._id,
+      objectPath,
+    },
+  })
+  return ResultAsync.fromPromise(
+    AwsConfig.s3.getSignedUrlPromise('getObject', {
+      Bucket: AwsConfig.paymentProofS3Bucket,
+      Key: objectPath,
+      Expires: 30 * dayInSeconds,
```
tshuli: /minor suggestion: add a comment here that the presigned link is used for immediate redirect, so the expiry doesn't affect user

shared/types/payment.ts
```diff
@@ -29,6 +29,7 @@ export type CompletedPaymentMeta = {
   submissionId: string
   transactionFee: number
   receiptUrl: string
+  hasS3ReceiptUrl: boolean
```
tshuli: /nit: maybe something like `hasReceiptStoredInS3` is more informative (hasS3ReceiptUrl suggests the url is stored in our DB for retrieval)

tshuli (commented):
Some minor comments for consideration :)

src/app/modules/payments/stripe.service.ts
KenLSM: Mainly due to lack of refinement of `payment.completePayment`, it was ready checked in the previous chain under `_retrieveReceiptUrlFromStripe`.

Added typing refinement in 18c84b271

src/app/modules/payments/stripe.service.ts
KenLSM: Great catch! de8c5a89e

KenLSM:
> Some minor comments for consideration :)

Thanks for the early review! Will be continuing to do some refactors around, will ping you again when it's completed.

shared/types/payment.ts
KenLSM: renamed to `hasReceiptStoredInS3` in 69002e5e1

src/app/modules/payments/stripe.service.ts
KenLSM: Left as jsdoc comment

src/app/modules/payments/stripe.service.ts
KenLSM: Yep agreed, added in 69002e5e1

KenLSM:
@tshuli could you help me to review the addition of nginx config on EBS? Specifically these files:
```
.github/workflows/deploy-eb.yml
.platform/nginx/conf.d/01_custom_ngix.conf
```
Docs for adding new [configs](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/platforms-linux-extend.html#platforms-linux-extend.example) for al2

Did a test on `staging` to make sure that the header issue is no longer present.

KenLSM:
> @tshuli could you help me to review the addition of nginx config on EBS? Specifically these files:
> 
> ```
> .github/workflows/deploy-eb.yml
> .platform/nginx/conf.d/01_custom_ngix.conf
> ```
> 
> Docs for adding new [configs](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/platforms-linux-extend.html#platforms-linux-extend.example) for al2
> 
> Did a test on `staging` to make sure that the header issue is no longer present.

Removed this. 

**Background**
@tshuli and I had done some investigation and couldn't reliably prove that adding this configuration had prevented the issue. 

We were also reminded that there was a similar issue happening previously where our staging environment had some differences with other environment. The engineers had spent significant time troubleshooting, but was unable to isolate the cause. Their recommendation was to relook after IaC migration.

src/app/constants/time.ts
```diff
@@ -0,0 +1 @@
+export const dayInSeconds = 24 * 60 * 60
```
tshuli: /nit: DAY_IN_SECONDS (upper case) for constants

src/app/modules/payments/__tests__/payment-proof.controller.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+        params: { formId: mockForm._id, paymentId: payment._id },
+      })
+      const mockRes = expressHandler.mockResponse()
+      const axiosSpy = jest
+        .spyOn(axios, 'get')
+        .mockRejectedValueOnce({ data: 'missing resource' })
+
+      const retrieveReceiptUrlFromStripeSpy = jest
+        .spyOn(PaymentProofService, '_retrieveReceiptUrlFromStripe')
+        .mockReturnValueOnce(okAsync('http://form.gov.sg'))
```
tshuli: https? XD

src/app/modules/payments/payment-proof.service.ts
```diff
@@ -0,0 +1,264 @@
...
+import { stripe } from '../../loaders/stripe'
+import { generatePdfFromHtml } from '../../utils/convert-html-to-pdf'
+
+import {
+  InvoicePdfGenerationError,
+  PaymentProofPresignS3Error,
+  PaymentProofUploadS3Error,
+} from './payment-proof.errors'
+import { getPaymentProofS3ObjectPath } from './payment-proof.utils'
+import {} from './payments.errors'
```
tshuli: is this import needed

.template-env
```diff
@@ -12,11 +12,13 @@ DB_HOST=
 
 ## AWS Config
 ATTACHMENT_S3_BUCKET=
+PAYMENT_PROOF_S3_BUCKET=
 IMAGE_S3_BUCKET=
 STATIC_ASSETS_S3_BUCKET=
 LOGO_S3_BUCKET=
 VIRUS_SCANNER_QUARANTINE_S3_BUCKET=
 VIRUS_SCANNER_CLEAN_S3_BUCKET=
+PAYMENT_PROOF_S3_BUCKET=
```
tshuli: duplicate

src/app/modules/payments/__tests__/payment-proof.controller.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+        .spyOn(axios, 'get')
+        .mockResolvedValueOnce({ data: '<html>some html</html>' })
+
+      const convertInvoiceSpy = jest
+        .spyOn(StripeUtils, 'convertToProofOfPaymentFormat')
+        .mockReturnValueOnce('<html>some converted html</html>')
+
+      const generatePdfFromHtmlSpy = jest
+        .spyOn(ConvertHtmlToPdf, 'generatePdfFromHtml')
+        .mockReturnValueOnce(Promise.resolve(Buffer.from('123')))
```
tshuli: .mockResolvedValueOnce

tshuli (approved):
thanks @KenLSM ! one failing test to resolve 😅  + some minor comments, otherwise lgtm!

src/app/modules/payments/__tests__/payment-proof.controller.spec.ts
KenLSM: eef67b99d

.template-env
KenLSM: eef67b99d

src/app/modules/payments/payment-proof.service.ts
KenLSM: Great catch! Possibly went through when I added the error code migration eef67b99d

src/app/modules/payments/__tests__/payment-proof.controller.spec.ts
KenLSM: Hah! 137bb4935

src/app/constants/time.ts
KenLSM: 07fa436af

PR #6734, opened 2023-09-20
feat(virus-scanner): invoke lambda to scan file

LinHuiqing:
## Problem

This is the second step of a 4-step effort to integrate the virus scanner into storage submission workflow after the encryption boundary shift (FRM-1292). Specifically, this PR sets up part of the middleware to scan attachments - invoking the lambda that scans a specific file in the quarantine bucket and moves it to the clean bucket if it is not malicious.

- [Design docs](https://www.notion.so/opengov/Virus-Scanner-Design-Doc-5464ffd2af88420495209baa6a55e1bf?pvs=4)
- [Overarching Linear ticket](https://linear.app/ogp/issue/FRM-1292/set-up-frontend-backend-integration-for-virus-scanner)

Closes FRM-1304

## Solution

A new middleware `scanAndRetrieveAttachments` is introduced to trigger the virus scan and download the file from the clean bucket. To do this while preventing regression for existing submissions with attachments, we use the filename to take the quarantine bucket file key.

Concretely, we need the following information from the frontend for the new flow to include virus scanning:

- Filename
- Quarantine file key
- Attachment FieldId
- Filetype

With the current `addAttachmentToResponses` implementation in `src/app/modules/submission/receiver/receiver.utils.ts`, the `response.answer`, which is what will appear as the file name to admins when they download, is forcibly set to `file.filename`. After the download file from clean bucket part is implemented, this should be changed for storage mode forms.

With the code change introduced by this PR, the lambda is triggered, but not downloaded from the clean bucket, so the submission validation will throw a invalid file extension error since the filename will not have any extensions.

**Breaking Changes** 

**Features**:

- Sets up the `scanAndRetrieveAttachments` middleware. This should be before the `validateStorageSubmission` middleware and does the following:
  1. Check if virus scanner has been enabled on BE. If virus scanner is not enabled, skip this middleware.
  2. If virus scanner is enabled, check if storage submission v2.1+. Storage submission v2.1 onwards should have virus scanning enabled. If not, skip this middleware. Note: Version number is sent by the frontend and should only be >=2.1 if virus scanning is enabled on the frontend.
  3. Trigger lambda to scan attachments.
  4. [TODO in another PR] Retrieve attachments from the clean bucket.
- Use versioning (2.1+) as an indicator that virus scanning is allowed on the FE.

**Improvements**:

- Uses aws-adk v3 for lambda to work towards upgrading aws-sdk from v2 to v3 (FRM-993).

## Tests

_Sed I can't put the requests for this set of tests on Insomnia :'(_

#### Preparations:

**Get the storage submission request:**

- [ ] Ensure that your app:
    1. Is connected to growthbook - ensure that the client key is loaded.
    2. Will direct storage submission requests to `/storage`.
- [ ] Create a storage submission form with 2 attachment fields and open it.
- [ ] Go to the respondent page and open the network panel.
- [ ] Make a submission including an attachment. To make your life easier with the tests below, attach a small file (e.g. text file with a few bytes).
- [ ] Copy the cURL request `A` for the submission. It should be to the endpoint that shows up as `storage?recaptcha...`. 
- [ ] Modify `A` to change the version to 2.1. Save this as request `B`.

**Put a valid file in the quarantine bucket:**

- [ ] Create a file with a valid uuid (e.g. `1b90195b-ce8a-4590-810b-04ebaef8e4dd`).
- [ ] Upload the file to the quarantine bucket.
  - If you are on local:
    1. Go to `formsg-backend-1`'s terminal.
    3. Create a text file with something like `echo "testing 123" > 1b90195b-ce8a-4590-810b-04ebaef8e4dd`.
    4. Copy the file over to the quarantine bucket with `aws --endpoint=http://localhost:4566 s3 cp 1b90195b-ce8a-4590-810b-04ebaef8e4dd s3://local-virus-scanner-quarantine-bucket`
    5. Make sure that the file has been successfully copied over with `aws --endpoint=http://localhost:4566 s3 ls s3://local-virus-scanner-quarantine-bucket`
  - If you are not on local, in S3, look for `[staging|prod|uat].virus.scanner.quarantine`.

**Allow virus scanner to be used on BE:**

- [ ] In the DB's `featureflags` collection, add the flag `encryption-boundary-shift-virus-scanner` and set enabled to `true`.

#### Make sure that virus scanner runs when feature flag is on and version number is 2.1+:

- [ ] Modify cURL request `B` by changing the filename to the file that you've added to the quarantine bucket. The file you've added should have a key that's a valid uuid. Save this as request `C`.
- [ ] Run request `C`. If `encryption-boundary-shift-hard-validation` is disabled, 
  - [ ] The storage submission should be submitted successfully. Note that the name of the file is now the uuid you've pasted.
  - [ ] In the logs (cloudwatch or local), there should be a warning for invalid file extension thrown by the submission validator.
  - [ ] The file you've uploaded to the quarantine bucket should have poofed! A new file will now be in the clean bucket. This file will be exactly the same as the file you've uploaded.
  - [ ] Restore the quarantine bucket file if you're still testing for effect of the `encryption-boundary-shift-hard-validation` flag.
- [ ] Run request `C`. If `encryption-boundary-shift-hard-validation` is enabled, 
  - [ ] The storage submission should fail with a validation error for invalid file extension.
  - [ ] The file you've uploaded to the quarantine bucket should have poofed! A new file will now be in the clean bucket. This file will be exactly the same as the file you've uploaded.
  - [ ] Restore the quarantine bucket file if you're still testing for effect of the `encryption-boundary-shift-hard-validation` flag.
  - [ ] Set `encryption-boundary-shift-hard-validation` to disabled.
- [ ] Rerun request `C`. This should now fail with a virus scan failure message since the quarantine bucket file has poofed.

#### Make sure that `encryption-boundary-shift-virus-scanner` feature flag works

- [ ] Unset the `encryption-boundary-shift-virus-scanner` feature flag or disable it.
- [ ] Restore the file in the quarantine bucket.
- [ ] Run cURL request `C`. The file in the quarantine bucket should remain.

#### Make sure that versioning works

- [ ] Enable the `encryption-boundary-shift-virus-scanner` feature flag.
- [ ] Ensure that the file is in the quarantine bucket.
- [ ] Run modified request `C` by changing the version number to 2. The file in the quarantine bucket should remain.

#### Make sure that only uuid file keys are accepted

- [ ] Run modified request `C`, where the filename is not a valid UUID. An error should be thrown flagging that the file key is not a valid UUID.

Regression tests:

- [ ] Submit a email mode form with multiple attachments.
- [ ]  With `encryption-boundary-shift-virus-scanner` flag disabled, submit a storage mode form (/storage endpoint) with multiple attachments.
- [ ] Try this with some of the attachment fields empty.

## Deploy Notes

Note: In production, ensure that in the DB's `featureflags` collection, the flag `encryption-boundary-shift-virus-scanner` is either **unset** or set to **false**.

- `VIRUS_SCANNER_LAMBDA_FUNCTION_NAME` : name of virus scanner lambda
  - dev: `function`
  - prod: [`virus-scanner-production-virus-scanner`](https://ap-southeast-1.console.aws.amazon.com/lambda/home?region=ap-southeast-1#/functions/virus-scanner-production-virus-scanner)
  - staging: [`virus-scanner-staging-virus-scanner`](https://ap-southeast-1.console.aws.amazon.com/lambda/home?region=ap-southeast-1#/functions/virus-scanner-staging-virus-scanner)
  - uat: [`virus-scanner-uat-virus-scanner`](https://ap-southeast-1.console.aws.amazon.com/lambda/home?region=ap-southeast-1#/functions/virus-scanner-uat-virus-scanner)

**New dependencies**:

- `@aws-sdk/client-lambda`: lambda package from aws-sdk v3, as part of FRM-993.

**New AWS policies**:

- `formsg-staging-virus-scanner-s3-lambda` policy added (for `formsg-staging-ec2-role-1`):

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "TriggerVirusScanningLambda",
      "Effect": "Allow",
      "Action": "lambda:InvokeFunction",
      "Resource": "arn:aws:lambda:ap-southeast-1:652261905145:function:virus-scanner-staging-virus-scanner"
    },
    {
      "Sid": "PutFileInQuarantine",
      "Effect": "Allow",
      "Action": "s3:PutObject",
      "Resource": "arn:aws:s3:::staging.virus.scanner.quarantine/*"
    },
    {
      "Sid": "GetFileFromClean",
      "Effect": "Allow",
      "Action": "s3:GetObject",
      "Resource": "arn:aws:s3:::staging.virus.scanner.clean/*"
    }
  ]
}
```
- Corresponding policies have been added to uat and prod too.

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -3,7 +3,7 @@ import { celebrate, Joi, Segments } from 'celebrate'
 import { NextFunction } from 'express'
 import { StatusCodes } from 'http-status-codes'
 import { chain, omit } from 'lodash'
-import { errAsync, okAsync, ResultAsync } from 'neverthrow'
+import { okAsync, ResultAsync } from 'neverthrow'
 
 import { featureFlags } from '../../../../../shared/constants'
 import {
@@ -184,10 +184,9 @@ export const checkNewBoundaryEnabled = async (
 }
 
 /**
- * Guardrail to prevent virus scanner code from being run if not enabled on frontend.
- * TODO (FRM-1232): remove this guardrail when encryption boundary is shifted.
+ * Scan attachments on quarantine bucket and retrieve attachments from the clean bucket.
  */
-export const scanAttachments = async (
+export const scanAndRetrieveAttachments = async (
   req: StorageSubmissionMiddlewareHandlerRequest,
   res: Parameters<StorageSubmissionMiddlewareHandlerType>[1],
   next: NextFunction,
@@ -197,6 +196,7 @@ export const scanAttachments = async (
     ...createReqMeta(req),
   }
 
+  // TODO (FRM-1413): remove this guardrail when virus scanning has completed rollout.
   // Step 1: If virus scanner is not enabled, skip this middleware.
 
   const virusScannerEnabled = req.formsg.featureFlags.includes(
@@ -212,6 +212,7 @@ export const scanAttachments = async (
     return next()
   }
 
+  // TODO (FRM-1413): remove this guardrail when virus scanning has completed rollout.
   // Step 2: If virus scanner is enabled, check if storage submission v2.1+. Storage submission v2.1 onwards
   // should have virus scanning enabled. If not, skip this middleware.
   // Note: Version number is sent by the frontend and should only be >=2.1 if virus scanning is enabled on the frontend.
@@ -225,43 +226,7 @@ export const scanAttachments = async (
   const triggerLambdaResult = await ResultAsync.combine(
     req.body.responses.map((response) => {
       if (isQuarantinedAttachmentResponse(response)) {
-        return triggerVirusScanning(response.filename).andThen((data) => {
-          const { statusCode, body } = JSON.parse(
-            Buffer.from(data?.Payload ?? '').toString(),
-          ) as {
-            statusCode: number
-            body: string
-          }
-          const parsedBody = JSON.parse(body)
-          const returnPayload = {
-            statusCode,
-            body: parsedBody,
-          }
-          logger.info({
-            message: 'Successfully invoked lambda function',
-            meta: {
-              ...logMeta,
-              responseMetadata: data?.$metadata,
-              returnPayload,
-            },
-          })
-
-          // If return payload is not OK, either file cannot be found or virus scan failed. This should be
-          // treated as a fatal error we should investigate.
-          if (statusCode !== StatusCodes.OK) {
-            logger.warn({
-              message: 'Some lambda invocations failed',
-              meta: {
-                ...logMeta,
-                responseMetadata: data?.$metadata,
-                returnPayload,
-              },
-            })
-            return errAsync(returnPayload)
-          }
-
-          return okAsync(returnPayload)
-        })
+        return triggerVirusScanning(response.filename)
       }
       // If response is not an attachment, return ok.
       return okAsync(true)
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -3,7 +3,7 @@ import { celebrate, Joi, Segments } from 'celebrate'
 import { NextFunction } from 'express'
 import { StatusCodes } from 'http-status-codes'
 import { chain, omit } from 'lodash'
-import { okAsync, ResultAsync } from 'neverthrow'
+import { errAsync, okAsync, ResultAsync } from 'neverthrow'
 
 import { featureFlags } from '../../../../../shared/constants'
 import {
@@ -193,7 +193,7 @@ export const scanAttachments = async (
   next: NextFunction,
 ) => {
   const logMeta = {
-    action: 'checkAttachmentQuarantineKeys',
+    action: 'scanAttachments',
     ...createReqMeta(req),
   }
 
@@ -225,29 +225,45 @@ export const scanAttachments = async (
   const triggerLambdaResult = await ResultAsync.combine(
     req.body.responses.map((response) => {
       if (isQuarantinedAttachmentResponse(response)) {
-        return triggerVirusScanning(response.answer).map((data) => {
-          const returnPayload = JSON.parse(
+        return triggerVirusScanning(response.filename).andThen((data) => {
+          const { statusCode, body } = JSON.parse(
             Buffer.from(data?.Payload ?? '').toString(),
           ) as {
             statusCode: number
-            body: {
-              message: string
-              cleanFileKey: string
-              destinationVersionId: string
-            }
+            body: string
+          }
+          const parsedBody = JSON.parse(body)
+          const returnPayload = {
+            statusCode,
+            body: parsedBody,
           }
           logger.info({
             message: 'Successfully invoked lambda function',
             meta: {
               ...logMeta,
               responseMetadata: data?.$metadata,
-              statusCode: data?.StatusCode,
               returnPayload,
             },
           })
+
+          // If return payload is not OK, either file cannot be found or virus scan failed. This should be
+          // treated as a fatal error we should investigate.
+          if (statusCode !== StatusCodes.OK) {
+            logger.warn({
+              message: 'Some lambda invocations failed',
+              meta: {
+                ...logMeta,
+                responseMetadata: data?.$metadata,
+                returnPayload,
+              },
+            })
+            return errAsync(returnPayload)
+          }
+
           return okAsync(returnPayload)
         })
       }
+      // If response is not an attachment, return ok.
       return okAsync(true)
     }),
   )
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -231,6 +231,7 @@ export const scanAttachments = async (
             meta: {
               ...logMeta,
               responseMetadata: data?.$metadata,
+              returnPayload: data?.Payload,
             },
           })
           return okAsync(true)
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -226,19 +226,26 @@ export const scanAttachments = async (
     req.body.responses.map((response) => {
       if (isQuarantinedAttachmentResponse(response)) {
         return triggerVirusScanning(response.answer).map((data) => {
+          const returnPayload = JSON.parse(
+            Buffer.from(data?.Payload ?? '').toString(),
+          ) as {
+            statusCode: number
+            body: {
+              message: string
+              cleanFileKey: string
+              destinationVersionId: string
+            }
+          }
           logger.info({
             message: 'Successfully invoked lambda function',
             meta: {
               ...logMeta,
               responseMetadata: data?.$metadata,
-              returnPayload: Buffer.from(data?.Payload ?? '').toString(),
-              logResult: Buffer.from(
-                data?.LogResult ?? '',
-                'base64',
-              ).toString(),
+              statusCode: data?.StatusCode,
+              returnPayload,
             },
           })
-          return okAsync(true)
+          return okAsync(returnPayload)
         })
       }
       return okAsync(true)
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -31,7 +31,10 @@ import {
   EncryptedPayloadExistsError,
   FormsgReqBodyExistsError,
 } from './encrypt-submission.errors'
-import { checkFormIsEncryptMode } from './encrypt-submission.service'
+import {
+  checkFormIsEncryptMode,
+  triggerVirusScanning,
+} from './encrypt-submission.service'
 import {
   CreateFormsgAndRetrieveFormMiddlewareHandlerRequest,
   CreateFormsgAndRetrieveFormMiddlewareHandlerType,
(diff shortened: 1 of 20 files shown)


src/app/modules/submission/submission.utils.ts
```diff
@@ -195,6 +195,15 @@ export const isAttachmentResponse = (
   )
 }
 
+/**
+ * Checks if a response is a quarantined attachment response to be processed by the virus scanner.
+ */
+export const isQuarantinedAttachmentResponse = (
+  response: ParsedClearFormFieldResponse,
+): response is ParsedClearAttachmentResponse => {
```
tshuli, line 203: /nit: to update jsdoc for ParsedClearAttachmentResponse (current comment only mentions for email mode form)

src/app/modules/submission/submission.utils.ts
```diff
@@ -195,6 +195,15 @@ export const isAttachmentResponse = (
   )
 }
 
+/**
+ * Checks if a response is a quarantined attachment response to be processed by the virus scanner.
+ */
+export const isQuarantinedAttachmentResponse = (
```
tshuli, line 201: how is this typeguard different from `isAttachmentResponse`?

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -578,3 +580,70 @@ export const getQuarantinePresignedPostData = (
...
+    (error) => {
+      logger.error({
+        message: 'Error encountered when invoking virus scanning lambda',
+        meta: logMeta,
+        error,
+      })
+
+      return new VirusScanFailedError()
+    },
+  ).andThen((data) => {
```
tshuli: /nit: typing for `data`

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -578,3 +580,70 @@ export const getQuarantinePresignedPostData = (
...
+    const { statusCode, body } = JSON.parse(
+      Buffer.from(data?.Payload ?? '').toString(),
+    ) as {
+      statusCode: number
+      body: string
+    }
```
tshuli: suggestion: can try a typeguard here for safety, instead of type assertion

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -578,3 +580,70 @@ export const getQuarantinePresignedPostData = (
...
+ * @returns okAsync(returnPayload) if file has been successfully scanned with status 200 OK
+ * @returns errAsync(returnPayload) if lambda invocation failed or file cannot be found
+ */
+export const triggerVirusScanning = (quarantineFileKey: string) => {
+  const logMeta = {
+    action: 'triggerVirusScanning',
+    meta: {
+      quarantineFileKey,
+    },
+  }
```
tshuli, line 738: suggestion: can add a quick check here that `quarantineFileKey` is a uuid

```
import { validate } from 'uuid'
if (!validate(...))
```

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -578,3 +580,70 @@ export const getQuarantinePresignedPostData = (
...
+        meta: {
+          ...logMeta,
+          responseMetadata: data?.$metadata,
+          returnPayload,
+        },
+      })
+      return errAsync(returnPayload)
+    }
+
+    return okAsync(returnPayload)
```
tshuli: suggestion: should type `returnPayload` (body) so that we know to expect the clean file key + versionId from it`

src/app/modules/submission/receiver/receiver.utils.ts
```diff
@@ -79,7 +79,6 @@ export const addAttachmentToResponses = (
     responses.forEach((response) => {
       if (isAttachmentResponseFromMap(attachmentMap, response)) {
         const file = attachmentMap[response._id]
-        response.answer = file.filename
         response.filename = file.filename
```
tshuli: qn: why change the key here to filename; is this a breaking change?

tshuli (commented):
thanks @LinHuiqing! looking good, some comments on the draft version first 😃

src/app/modules/submission/submission.utils.ts
LinHuiqing: Because for the new flow with the quarantine key, content will not have been sent by the FE, so there will not be content at the point of time this check is used.

src/app/modules/submission/receiver/receiver.utils.ts
LinHuiqing: Oh I'm just removing the force setting of `response.answer` to `file.filename` so I can pass the quarantine file key in the answer without it being overwritten. So a request to the virus scan enabled endpoint can still succeed while triggering the lambda in parallel.

If I were to leave the setting of `response.answer` to `file.filename` here, I can set the filename to the quarantine file key and trigger the lambda scan, but then the validator down the line will complain that the file doesn't have a valid file extension.

LinHuiqing: I don't think this is a breaking change because based on my observation of submissions, the FE should already be setting the `response.answer` as the filename.

src/app/modules/submission/submission.utils.ts
LinHuiqing: Comment added here: [7287887](https://github.com/opengovsg/FormSG/pull/6734/commits/728788750682732ee9726e588de83e83ab02c7e9)

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: Done in [de324a3](https://github.com/opengovsg/FormSG/pull/6734/commits/de324a349964247caf91bdffa20f921d20fcb86d)!

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: Thanks for the suggestion! Type guards added here: [de324a3](https://github.com/opengovsg/FormSG/pull/6734/commits/de324a349964247caf91bdffa20f921d20fcb86d)

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: Thanks! Typing added here: [de324a3](https://github.com/opengovsg/FormSG/pull/6734/commits/de324a349964247caf91bdffa20f921d20fcb86d)

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: Good point, uuid check added, thanks! [34613e7](https://github.com/opengovsg/FormSG/pull/6734/commits/34613e7a263c268061a7f935236c90b5ba267d72)

src/app/modules/submission/receiver/receiver.utils.ts
LinHuiqing: Okay on second thought, playwright tests might be failing because of this, meaning there's a regression and is a breaking change. I've restored this in [eec5388](https://github.com/opengovsg/FormSG/pull/6734/commits/eec53881d6d5e11d27a2963b73d9b23705be06c9). We can bear with the invalid file extension error since the virus scanning flag should be unset or off when this goes out.

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -176,6 +183,75 @@ export const checkNewBoundaryEnabled = async (
   return next()
 }
 
+/**
+ * Scan attachments on quarantine bucket and retrieve attachments from the clean bucket.
+ */
```
tshuli: For this PR, we are only scanning, not retrieving :) could add a TODO to indicate that we are not ready to turn on virus scanner flag yet

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
tshuli: I may have been unclear when I suggested to consider [typeguard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) :sweat_smile: basically something closer to `isQuarantinedAttachmentResponse`.
```
  const parsedPayload = parsedPayloadResult.value as {
    statusCode: number
    body: string
  }

  // Step 2a: Check if statusCode and body (unparsed) are of the correct types
  if (
    typeof parsedPayload.statusCode === 'number' &&
    typeof parsedPayload.body === 'string'
  ) {

```
for this, the typeguard would look something like
```
type ParsedPayload = {statusCode: number, body: string}
const isParsedPayload(payload: unknown): payload is ParsedPayload {
    return (
        typeof payload === 'object' &&
        !!payload &&
        hasProp(payload, 'statusCode') &&
        typeof payload.statusCode === 'number' &&
        hasProp(payload, 'body') &&
        typeof payload.body === 'string' 
    )
}
```
and then in the code, you can just call
```
if (!isParsedPayload(parsedBodyResult.value) {
   // return err(...) here
}

// continue; type of parsedBodyResult.value is now ParsedPayload
```

LinHuiqing: Ooh okay I see haha oops, will refactor to do this instead!

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Added comment in [44342cf](https://github.com/opengovsg/FormSG/pull/6734/commits/44342cff6f3614763f5e6fe34c1a735f650f9edf)! I'm keeping the comment to say retrieving because that's what the middleware should do, and instead added another line to say it's not completely implemented yet.

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: Done in [44342cf](https://github.com/opengovsg/FormSG/pull/6734/commits/44342cff6f3614763f5e6fe34c1a735f650f9edf), thanks!

tshuli (approved):
lgtm! thanks @LinHuiqing

PR #6741, opened 2023-09-21
fix: hotfix for 500 errors thrown on email bounce notification endpoint

justynoh:
## Problem

On Thurs 21 Sep, we encountered elevated errors from our `notifications/bounces/email` endpoint. This was identified to be due to AWS sending notifications that did not conform to the shape we were expecting. This PR seeks to fix this.

## Solution

- Add `parseBounceNotificationCommonHeadersToEmails` to `bounce.utils.ts` which flatmaps an array of (emails | comma-separated emails) into an array of only single emails.
- Use the new util in bounce model where SNS notifications are parsed into bounce documents. In particular, they are used in `BounceSchema.statics.fromSnsNotification` and `BounceSchema.methods.updateBounceInfo`.

**Breaking Changes** 

## Tests

Prior to release, this should be tested on staging by triggering a bounce notification to staging prior to and after deployment. If a 500 error was thrown on staging prior to deployment, we should expect it not to happen after deployment.

Update: Not able to replicate 500s, but we have tested multiple times on email submissions and the bounce notifications were successfully parsed and returned 200.


LinHuiqing (approved):
lgtm!

PR #6743, opened 2023-09-21
fix: redirect to frontend app after myinfo log in (for local dev environment)

wanlingt:
## Problem

**On local dev only**: After logging in on a myinfo form, the user gets redirected to `localhost:5001` (our backend server) instead of `localhost:3000` (our frontend server). This makes it difficult to debug errors on the frontend.

## Solution

When creating the redirect URL from myinfo back to FormSG, check if it's a `development` environment. If yes, use an absolute path with the frontend app URL instead of a relative path.

**Breaking Changes** 

Staging/Prod
- [ ] Log into a MyInfo form
- [ ] Upon logging in successfully, you should be redirected back to FormSG

Local dev
- [ ] Log into a MyInfo form
- [ ] Upon logging in successfully, you should be redirected back to FormSG on `localhost:3000`

## Deploy Notes

- `FE_APP_URL` : Frontend application URL (for local development use only)

### DIFF
--- a/src/app/modules/myinfo/myinfo.controller.ts
+++ b/src/app/modules/myinfo/myinfo.controller.ts
@@ -1,6 +1,8 @@
 import { celebrate, Joi, Segments } from 'celebrate'
 import { StatusCodes } from 'http-status-codes'
 
+import { Environment } from '../../../types'
+import config from '../../config/config'
 import { createLoggerWithLabel } from '../../config/logger'
 import { createReqMeta } from '../../utils/request'
 import { ControllerHandler } from '../core/core.types'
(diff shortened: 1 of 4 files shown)


src/app/modules/myinfo/myinfo.controller.ts
```diff
@@ -139,11 +141,17 @@ export const loginToMyInfo: ControllerHandler<
   }
   const { formId, encodedQuery } = parseStateResult.value
 
-  let redirectDestination = `/${formId}`
+  // For local dev, we need to specify the frontend app URL as this is different from the backend's app URL
+  const redirectDestinationRaw =
+    process.env.NODE_ENV === Environment.Dev
```
KenLSM, line 146: iirc `Environment.Dev` refers to `staging` and friends, which might not be intended since we only want to use this local feAppUrl for local dev 🤔 

I faced this similar problem when working with `mockTwilio`. I ended up using `.env`/`.template-env` instead.

wanlingt: `Environment` has 3 values - `development`, `production` and `test`. You might be referring to `isDev` (in `config.ts`) which refers to staging and friends and local dev.

I've tested this locally and on staging, on staging it redirects to the relative path :) which shows it's not covered by `Environment.Dev`

KenLSM: I see noted. I've checked the workflow file in GH action for staging and it is indeed the production flavour.
```
with:
  file: ./Dockerfile.production
  ...
```

KenLSM (approved):
LGTM!

PR #6745, opened 2023-09-22
feat: add index to myinfo child email response

wanlingt:
## Problem

MyInfo child responses are not numbered in the email response.

Closes FRM-1325

## Solution

There are 2 cases which would affect the numbering of the child

1. When there is >1 MyInfo child field
2. When there is >1 child per MyInfo child field (this happens when the respondent chooses to "Add another child")

The index of the child is represented by `ProcessedChildrenResponse.childIdx`, which is updated in `parseResponses`.

**Breaking Changes** 

    - This will change the response field name in the email response. This incoming change has already been communicated to the current users.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests

**Test for index** - this can't be performed on Staging as the mockpass profiles only have 1 child, but can be tested locally
- [ ] On a MyInfo form, add 2 child fields. Enable multiple children for the first child field
- [ ] Open the form
- [ ] For the first child field, add another child
- [ ] Fill in the form (there should be 3 children entries in total) and submit it
- [ ] The child fields in the email response should be numbered according to the child's index, e.g. `[MyInfo] Child 1 Name`

**Regression test**
- [ ] Add all MyInfo fields to a form
- [ ] Fill in the form and submit it
- [ ] There should be no change in the format of the MyInfo field names, other than the child field. The child field (assuming there's just 1) should be named in this format [MyInfo] Child <index> <field name> e.g. `[MyInfo] Child 1 Name`

### DIFF
--- a/src/app/modules/myinfo/myinfo.types.ts
+++ b/src/app/modules/myinfo/myinfo.types.ts
@@ -23,9 +23,9 @@ export interface IMyInfoRedirectURLArgs {
 // Field ID or a special key for a Child
 export type MyInfoKey = string | MyInfoChildKey
 
-// Field type, field ID, child attribute type, child name
+// Field type, field ID, child attribute type, child index, child name
 export type MyInfoChildKey =
-  `${MyInfoAttribute.ChildrenBirthRecords}.${string}.${MyInfoChildAttributes}.${string}`
+  `${MyInfoAttribute.ChildrenBirthRecords}.${string}.${MyInfoChildAttributes}.${number}.${string}`
 
 export type MyInfoHashPromises = Partial<
   Record<MyInfoAttribute | MyInfoChildKey, Promise<string>>
--- a/src/app/modules/myinfo/myinfo.types.ts
+++ b/src/app/modules/myinfo/myinfo.types.ts
@@ -23,9 +23,9 @@ export interface IMyInfoRedirectURLArgs {
 // Field ID or a special key for a Child
 export type MyInfoKey = string | MyInfoChildKey
 
-// Field type, field ID, child attribute type, child index, child name
+// Field type, field ID, child attribute type, child name
 export type MyInfoChildKey =
-  `${MyInfoAttribute.ChildrenBirthRecords}.${string}.${MyInfoChildAttributes}.${number}.${string}`
+  `${MyInfoAttribute.ChildrenBirthRecords}.${string}.${MyInfoChildAttributes}.${string}`
 
 export type MyInfoHashPromises = Partial<
   Record<MyInfoAttribute | MyInfoChildKey, Promise<string>>
--- a/src/app/modules/myinfo/myinfo.util.ts
+++ b/src/app/modules/myinfo/myinfo.util.ts
@@ -114,7 +114,7 @@ function hashChildrenFieldValues(
         myInfoFormattedValue = formatMyinfoDate(value)
       }
       readOnlyHashPromises[
-        getMyInfoChildHashKey(field._id, subField, childIdx, childName)
+        getMyInfoChildHashKey(field._id, subField, childName)
       ] = bcrypt.hash(myInfoFormattedValue, HASH_SALT_ROUNDS)
     })
   })
@@ -481,16 +481,14 @@ export const getMyInfoAttr = (
  * Helper function to get a MyInfo child's hash key inside an IHashes.
  *
  * @param fieldId The ID of the field the Child response belongs to.
- * @param childIdx The nth child to look for.
  * @returns An IHashes-compatible key.
  */
 export const getMyInfoChildHashKey = (
   fieldId: string,
   childAttr: MyInfoChildAttributes,
-  childIdx: number,
   childName: string,
 ): MyInfoChildKey => {
-  return `${MyInfoAttribute.ChildrenBirthRecords}.${fieldId}.${childAttr}.${childIdx}.${childName}`
+  return `${MyInfoAttribute.ChildrenBirthRecords}.${fieldId}.${childAttr}.${childName}`
 }
 
 /**
@@ -514,7 +512,7 @@ export const handleMyInfoChildHashResponse = (
   if (!subFields) {
     return
   }
-  childField.answerArray.forEach((childAnswer, childIndex) => {
+  childField.answerArray.forEach((childAnswer) => {
     // Name should be first field for child answers
     const childName = childAnswer[0]
     // Validate each answer (child)
@@ -522,7 +520,6 @@ export const handleMyInfoChildHashResponse = (
       const key = getMyInfoChildHashKey(
         field._id as string,
         subFields[subFieldIndex],
-        childIndex,
         childName,
       )
       const hash = hashes[key]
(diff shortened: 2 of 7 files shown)


src/app/modules/submission/email-submission/email-submission.util.ts
```diff
@@ -218,23 +218,19 @@ export const getAnswerForCheckbox = (
...
-      _id: getMyInfoChildHashKey(
-        response._id,
-        subFields[idx],
-        childIdx,
-        childName,
-      ),
+      _id: getMyInfoChildHashKey(response._id, subFields[idx], childName),
       fieldType: response.fieldType,
-      question: `Child.${subFields[idx]}`,
+      question: `Child-${childIdx + 1}.${subFields[idx]}`,
```
KenLSM: Trying to understand the logic on why this index now works, as it was previously removed in #6697 to address the issue of incorrect childname index.

The primary cause of the initial issue was that:
- `response` was always a 2D array of size `[1][n]`, and
- `childIdx` from the parent is the correct index to use?

KenLSM (approved):
LGTM! Have a question to aid my understanding of the root cause!

tshuli:
Sorry for jumping in late - couple of comments mainly from UI/UX perspective; also tagging @staceytan1998 

- I was wondering if it makes sense to map the fields in a more human-readable format? eg Child-1.childname could simply be Child 1: Name
- For Myinfo child, should we go for [Myinfo] [child] instead of [child] [MyInfo]?
- When would we have [child] without [Myinfo] - is this simply when the data is from MyInfo but not MyInfo verified? in that case, should we just do away with the [child] label entirely? Since all the child fields would already have `child` in the field name itself, e.g. `Child 2: Date of Birth`

staceytan1998:
Thanks for tagging @tshuli ! 

- **[Myinfo] Child 1 name** makes sense! Are there other instances where we have [Myinfo] followed by another [xxx] ? Otherwise I think we can leave out the [Child]! 
- And same question as Shu Li, are there instances where there is no [Myinfo] before the field title?

src/app/modules/submission/email-submission/email-submission.util.ts
wanlingt: Yes you're right! The function `createFormattedDataForOneField` only creates the data for each field, so this meant that previously it did not track the index of the child. This is why we now pass in `childIdx` from the parent.

wanlingt:
> Sorry for jumping in late - couple of comments mainly from UI/UX perspective; also tagging @staceytan1998
> 
> * I was wondering if it makes sense to map the fields in a more human-readable format? eg Child-1.childname could simply be Child 1: Name
> * For Myinfo child, should we go for [Myinfo] [child] instead of [child] [MyInfo]?
> * When would we have [child] without [Myinfo] - is this simply when the data is from MyInfo but not MyInfo verified? in that case, should we just do away with the [child] label entirely? Since all the child fields would already have `child` in the field name itself, e.g. `Child 2: Date of Birth`

Thanks for the comments @tshuli ! 

> When would we have [child] without [Myinfo] - is this simply when the data is from MyInfo but not MyInfo verified?
Yes that's correct.

I just synced with @staceytan1998 - We'll go with the format **[Myinfo] Child 1 Name**, and also remove the `[child]` prefix.

tshuli:
> I just synced with @staceytan1998 - We'll go with the format [Myinfo] Child 1 Name, and also remove the [child] prefix.

thanks @wanlingt and @staceytan1998 !

src/app/modules/myinfo/myinfo.types.ts
```diff
@@ -23,9 +23,9 @@ export interface IMyInfoRedirectURLArgs {
 // Field ID or a special key for a Child
 export type MyInfoKey = string | MyInfoChildKey
 
-// Field type, field ID, child attribute type, child index, child name
+// Field type, field ID, child attribute type, child name
 export type MyInfoChildKey =
-  `${MyInfoAttribute.ChildrenBirthRecords}.${string}.${MyInfoChildAttributes}.${number}.${string}`
```
tshuli: minor clarification: why remove the number here; would this not prevent the (admittedly minor) change of key collision between child records? e.g. twins with the same name 😅

src/app/modules/myinfo/myinfo.util.ts
```diff
@@ -482,16 +482,14 @@ export const getMyInfoAttr = (
...
- * @param childIdx The nth child to look for.
  * @returns An IHashes-compatible key.
  */
 export const getMyInfoChildHashKey = (
   fieldId: string,
   childAttr: MyInfoChildAttributes,
-  childIdx: number,
   childName: string,
 ): MyInfoChildKey => {
-  return `${MyInfoAttribute.ChildrenBirthRecords}.${fieldId}.${childAttr}.${childIdx}.${childName}`
```
tshuli: same here; why remove the child idx?

src/app/modules/submission/email-submission/email-submission.util.ts
```diff
@@ -484,20 +484,26 @@ const createFormattedDataForOneField = <T extends EmailDataFields | undefined>(
     response: ResponseFormattedForEmail,
     hashedFields: Set<MyInfoKey>,
   ) => T,
-): T[] => {
+  childIdx: number,
```
tshuli: I think this is not super ideal; we are passing arbitary value of `childIdx` in and ignoring it if the field is not a child field 😅

Two possible suggestions:

1) Update the type signature of `createFormattedDataForOneField` to be an union type of `createFormattedDataForOneChildField` and `createFormattedDataForOneNonChildField`, and enforce in the typing that `childIdx` exists

2) (more preferred, separation of concerns) shift the injection of child index up to `ParsedResponsesObject.class`, and include the child index in `ProcessedChildrenResponse` type. then `createFormattedDataForOneField` will be much cleaner (it becomes a pure formatter, instead of a formatter + injector)

tshuli (approved):
thanks @wanlingt, and for taking the time to polish the UI/UX 😃 

minor clarifications and comment on refactoring; can merge in once they've been addressed!

src/app/modules/myinfo/myinfo.types.ts
wanlingt: thanks for this comment @tshuli ! it made me realise the case where there's >1 child per child field (when the respondent chooses to "add another child"). Previously I assumed 1 child per child field. I've added the index back here [d173907](https://github.com/opengovsg/FormSG/pull/6745/commits/d1739075b59d65998b0b0ea73afa73f0af57b578)

src/app/modules/myinfo/myinfo.util.ts
wanlingt: re-added [d173907](https://github.com/opengovsg/FormSG/pull/6745/commits/d1739075b59d65998b0b0ea73afa73f0af57b578)

src/app/modules/submission/email-submission/email-submission.util.ts
wanlingt: went with the 2nd approach. thanks for the suggestion, it's a lot cleaner now! [c4cb6a4](https://github.com/opengovsg/FormSG/pull/6745/commits/c4cb6a4b2eaa79b8ac649a110de7abb9c4d779c0)

wanlingt:
> thanks @wanlingt, and for taking the time to polish the UI/UX 😃
> 
> minor clarifications and comment on refactoring; can merge in once they've been addressed!

thanks for the comments! I'll email the admins to inform them of the changes before merging this in

PR #6747, opened 2023-09-25
feat(virus-scanner): download clean file

LinHuiqing:
## Problem

This is the 3rd step of a 4-step effort to integrate the virus scanner into storage submission workflow after the encryption boundary shift (FRM-1292). Specifically, this PR sets up part of the middleware to scan attachments - getting the attachments from the clean bucket after they have been scanned.

- [Design docs](https://www.notion.so/opengov/Virus-Scanner-Design-Doc-5464ffd2af88420495209baa6a55e1bf?pvs=4)
- [Overarching Linear ticket](https://linear.app/ogp/issue/FRM-1292/set-up-frontend-backend-integration-for-virus-scanner)

Closes FRM-1302

## Solution

In the middleware `scanAndRetrieveAttachments`, code is implemented to download the file from the clean bucket.

**Breaking Changes** 

**Features**:

- `downloadCleanFile` is implemented at the service level to download attachments given clean file keys and version ids.

#### Preparations:

**Get the storage submission request:**

- [ ] Ensure that your app:
    1. Is connected to growthbook - ensure that the client key is loaded.
    2. Will direct storage submission requests to `/storage`.
- [ ] Create a storage submission form with 2 attachment fields (at least 1 of them optional) and open it.
- [ ] Go to the respondent page and open the network panel.
- [ ] Make a submission including an attachment. To make your life easier with the tests below, attach a small file (e.g. text file with a few bytes).
  - Hint: Content-Types:
    - `text/plain`
    - `application/pdf`
    - `image/png`
    - `image/gif`
- [ ] Copy the cURL request `A` for the submission. It should be to the endpoint that shows up as `storage?recaptcha...`. 
- [ ] Modify `A` to change the version to 2.1. Save this as request `B`.

**Put a valid file in the quarantine bucket:**

- [ ] Create a file with a valid uuid (e.g. `1b90195b-ce8a-4590-810b-04ebaef8e4dd`).
- [ ] Upload the file to the quarantine bucket.
  - If you are on local:
    1. Go to `formsg-backend-1`'s terminal.
    3. Create a text file with something like `echo "testing 123" > 1b90195b-ce8a-4590-810b-04ebaef8e4dd`.
    4. Copy the file over to the quarantine bucket with `aws --endpoint=http://localhost:4566 s3 cp 1b90195b-ce8a-4590-810b-04ebaef8e4dd s3://local-virus-scanner-quarantine-bucket`
    5. Make sure that the file has been successfully copied over with `aws --endpoint=http://localhost:4566 s3 ls s3://local-virus-scanner-quarantine-bucket`
  - If you are not on local, in S3, look for `[staging|prod|uat].virus.scanner.quarantine`.

**Allow virus scanner to be used on BE:**

- [ ] In the DB's `featureflags` collection, add the flag `encryption-boundary-shift-virus-scanner` and set enabled to `true`.

#### Make sure that virus scanner runs and clean file is downloaded when feature flag is on and version number is 2.1+:

- [ ] Modify cURL request `B` by changing the attachment field's answer to the file that you've added to the quarantine bucket. The file you've added should have a key that's a valid uuid. Save this as request `C`.
- [ ] Run request `C`.
  - [ ] The storage submission should succeed.
  - [ ] Go to the admin results page of the form. The latest response should have the attachment that you previously uploaded to the quarantine bucket, NOT the file that you used to create the cURL response.
- [ ] Rerun request `C`. This should now fail with a virus scan failure message since the quarantine bucket file has poofed.

#### Make sure that `encryption-boundary-shift-virus-scanner` feature flag works

- [ ] Unset the `encryption-boundary-shift-virus-scanner` feature flag or disable it.
- [ ] Restore the file in the quarantine bucket.
- [ ] Run cURL request `C`. The file in the quarantine bucket should remain.

#### Make sure that versioning works

- [ ] Enable the `encryption-boundary-shift-virus-scanner` feature flag.
- [ ] Ensure that the file is in the quarantine bucket.
- [ ] Run modified request `C` by changing the version number to 2. The file in the quarantine bucket should remain.

#### Make sure that only uuid file keys are accepted

- [ ] Run modified request `C`, where the filename is not a valid UUID. An error should be thrown flagging that the file key is not a valid UUID.

#### Normal E2E flows

- [ ] Submit a email mode form with multiple attachments.
- [ ]  With `encryption-boundary-shift-virus-scanner` flag disabled, submit a storage mode form (/storage endpoint) with multiple attachments.
- [ ] Try this with some of the attachment fields empty.

## Deploy Notes

**Modified AWS policies**:

- `formsg-staging-virus-scanner-s3-lambda` policy modified (for `formsg-staging-ec2-role-1`), in the Statement for `Sid` "GetFileFromClean", `GetObject` is replaced by `GetObjectVersion`. As such, the statement should look like the following:

```json
{
  "Sid": "GetFileFromClean",
  "Effect": "Allow",
  "Action": "s3:GetObjectVersion",
  "Resource": "arn:aws:s3:::staging.virus.scanner.clean/*"
}
```
- Corresponding policies have been added to uat and prod too.

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -15,6 +15,7 @@ import {
   EncryptAttachmentResponse,
   EncryptFormFieldResponse,
   FormLoadedDto,
+  ParsedClearFormFieldResponse,
 } from '../../../../types/api'
 import { paymentConfig } from '../../../config/features/payment.config'
 import formsgSdk from '../../../config/formsg-sdk'
@@ -229,9 +230,10 @@ export const scanAndRetrieveAttachments = async (
 
   // Step 3 + 4: For each attachment, trigger lambda to scan and if it succeeds, retrieve attachment from clean bucket. Do this asynchronously.
   const scanAndRetrieveFilesResult: Result<
-    boolean[], // true for attachment fields, false for non-attachment fields.
+    ParsedClearFormFieldResponse[], // true for attachment fields, false for non-attachment fields.
     VirusScanFailedError | DownloadCleanFileFailedError
   > = await ResultAsync.combine(
+    // If any scans or downloads error out, it will short circuit and return the first error.
     req.body.responses.map((response) => {
       if (isQuarantinedAttachmentResponse(response)) {
         // Step 3: Trigger lambda to scan attachments.
@@ -239,25 +241,23 @@ export const scanAndRetrieveAttachments = async (
           triggerVirusScanning(response.answer)
             .map((lambdaOutput) => lambdaOutput.body)
             // Step 4: Retrieve attachments from the clean bucket.
-            .andThen(
-              (cleanAttachment) =>
-                // Retrieve attachment from clean bucket.
-                downloadCleanFile(
-                  cleanAttachment.cleanFileKey,
-                  cleanAttachment.destinationVersionId,
-                ).map((attachmentBuffer) => {
-                  // Replace content with file retreived from clean bucket.
-                  response.content = attachmentBuffer
-                  // Replace answer with filename.
-                  response.answer = response.filename
-                  return true // Return true to indicate that the download was successful.
-                }), // If any downloads error out, it will short circuit and return the first error.
+            .andThen((cleanAttachment) =>
+              // Retrieve attachment from clean bucket.
+              downloadCleanFile(
+                cleanAttachment.cleanFileKey,
+                cleanAttachment.destinationVersionId,
+              ).map((attachmentBuffer) => ({
+                ...response,
+                // Replace content with attachmentBuffer and answer with filename.
+                content: attachmentBuffer,
+                answer: response.filename,
+              })),
             )
         )
       }
 
-      // If response is not an attachment, return false.
-      return okAsync(false)
+      // If field is not an attachment, return original response.
+      return okAsync(response)
     }),
   )
 
@@ -275,6 +275,9 @@ export const scanAndRetrieveAttachments = async (
       message: errorMessage,
     })
   }
+
+  // Step 5: Replace req.body.responses with the new responses with populated attachments.
+  req.body.responses = scanAndRetrieveFilesResult.value
 
   return next()
 }
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -268,29 +268,6 @@ export const scanAndRetrieveAttachments = async (
     })
   }
 
-  // const getBody = (response: GetObjectCommandOutput) => {
-  //   return response.Body && (response.Body as Readable);
-  // }
-
-  // const bodyToBuffer = (body: S3.Body) => {
-  //   let buffer: Buffer
-  //   if (body instanceof Uint8Array || body instanceof Blob) {
-  //     buffer = Buffer.from(body)
-  //   } else if (
-  //     typeof body === 'string' ||
-  //     body instanceof Readable ||
-  //     body instanceof ReadableStream
-  //   ) {
-  //     const chunks: Uint8Array[] = []
-  //     for await (const chunk of body) {
-  //       chunks.push(chunk instanceof Buffer ? chunk : Buffer.from(chunk))
-  //     }
-  //     buffer = Buffer.concat(chunks)
-  //   } else {
-  //     throw new Error('Invalid type for S3 object Body')
-  //   }
-  // }
-
   // Step 4: Retrieve attachments from the clean bucket.
   const downloadCleanFilesResult = await ResultAsync.combine(
     triggerLambdaResult.value.map((isAttachment) => {
@@ -300,11 +277,13 @@ export const scanAndRetrieveAttachments = async (
           isAttachment.cleanFile.cleanFileKey,
           isAttachment.cleanFile.destinationVersionId,
         ).map((result) => {
+          // Replace content with file retreived from clean bucket.
           isAttachment.response.content = result
-          return true
-        })
+          return true // Return true to indicate that the download was successful.
+        }) // If any downloads error out, it will short circuit and return the first error.
       }
 
+      // If response is not an attachment, return okAsync(false).
       return okAsync(false as const)
     }),
   )
@@ -580,8 +559,6 @@ export const createFormsgAndRetrieveForm = async (
   next: NextFunction,
 ) => {
   const { formId } = req.params
-
-  console.log('createFormsgAndRetrieveForm req.body.version', req.body.version)
 
   const logMeta = {
     action: 'createFormsgAndRetrieveForm',
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -15,7 +15,6 @@ import {
   EncryptAttachmentResponse,
   EncryptFormFieldResponse,
   FormLoadedDto,
-  ParsedClearAttachmentResponse,
 } from '../../../../types/api'
 import { paymentConfig } from '../../../config/features/payment.config'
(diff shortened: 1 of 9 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -786,3 +787,67 @@ export const triggerVirusScanning = (
...
+      meta: logMeta,
+    })
+
+    return errAsync(new InvalidFileKeyError())
+  }
+
+  let buffer = Buffer.alloc(0)
+
+  const writeStream = new Writable({
+    write(chunk, encoding, callback) {
```
tshuli: /nit can declare _encoding, since it is not used

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -786,3 +787,67 @@ export const triggerVirusScanning = (
...
+
+      readStream.on('error', (error) => {
+        reject(error)
+      })
+    }),
+    (error) => {
+      logger.error({
+        message: 'Error encountered when downloading file from clean bucket',
+        meta: logMeta,
+        error,
```
tshuli, line 847: might need to cast error to String(error) here

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -786,3 +787,67 @@ export const triggerVirusScanning = (
...
+    },
+  })
+
+  const readStream = AwsConfig.s3
+    .getObject({
+      Bucket: AwsConfig.virusScannerCleanS3Bucket,
+      Key: cleanFileKey,
+      VersionId: versionId,
+    })
+    .createReadStream()
```
tshuli: clarification: what was the consideration behind having a read stream pipe to the write stream, and then resolving the promise on stream end; might it be simpler to do something like:

```
let Body
try {
  const { Body } = await AwsConfig.s3
    .getObject({
      Bucket: AwsConfig.virusScannerCleanS3Bucket,
      Key: cleanFileKey,
      VersionId: versionId,
    })
    .promise()
} catch (e) {
    ... //Handle error here
}
```
were there any specific advantages you were trying to achieve?

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -225,13 +229,26 @@ export const scanAndRetrieveAttachments = async (
...
     req.body.responses.map((response) => {
       if (isQuarantinedAttachmentResponse(response)) {
-        return triggerVirusScanning(response.answer)
+        return triggerVirusScanning(response.answer).map((result) => ({
+          response,
+          cleanFile: result.body,
+        }))
       }
+
       // If response is not an attachment, return ok.
```
tshuli: /nit: update comment here

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -243,13 +260,50 @@ export const scanAndRetrieveAttachments = async (
...
+      if (isAttachment) {
+        // Retrieve attachment from clean bucket.
+        return downloadCleanFile(
+          isAttachment.cleanFile.cleanFileKey,
+          isAttachment.cleanFile.destinationVersionId,
+        ).map((result) => {
+          // Replace content with file retreived from clean bucket.
+          isAttachment.response.content = result
+          // Replace answer with filename.
+          isAttachment.response.answer = isAttachment.response.filename
```
tshuli: ~clarification: how does this work, are we replacing anything in `req`?~

right, so here we're passing down an object by reference across multiple .map() calls and then modifying the object property at the end. it gets pretty confusing 😅

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -243,13 +260,50 @@ export const scanAndRetrieveAttachments = async (
...
+          return true // Return true to indicate that the download was successful.
+        }) // If any downloads error out, it will short circuit and return the first error.
+      }
+
+      // If response is not an attachment, return okAsync(false).
+      return okAsync(false as const)
+    }),
+  )
```
tshuli: I think you can combine the two `Array.prototype.map()` calls in Step 3 and 4 into one, and then return a clean `response[]` at the end of `.map()` in order to replace `req.body.responses`. Then you dont need to return `false as const` along the way. Might also save on performance because you don't need all the virus scans to finish before you start downloading file.

tshuli (commented):
thanks @LinHuiqing ! looking good, mainly just a comment on how `scanAndRetrieveAttachments` can be refactored

ping me anytime to run through the tests and then i'll approve!

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: I was thinking about memory, but in this case, I think using streams won't make a difference oop 😅 will refactor to use promise!

LinHuiqing: Oh I realised why I went down this path in the first place, because `S3.GetObjectOutput.Body` can be `Buffer|Uint8Array|Blob|string|Readable`, and it'll all be `Buffer` if I just use the streams 😅 + `response.content` accepts `Buffer` type.

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Yup! This is kinda how we receive multipart form data too in `configureMultipartReceiver`. Not sure if there's a better day to do it though? 🤔

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: There seems to be no easier way to handle all the different types, sadly... https://stackoverflow.com/questions/25812384/how-do-determine-the-type-of-data-in-s3-getobject

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Sorry, which comment are you referring to? 😅

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Very good point! Refactored in [ac98173](https://github.com/opengovsg/FormSG/pull/6747/commits/ac98173b67d2a9492f52f82ece4a11b8fb9e39ff), thanks @tshuli! 🌮 🌮

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: `error` is typed as `Error` type in `readStream.on` already though, and that's the only path the promise is rejected.

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
LinHuiqing: Done in [c77b2f3](https://github.com/opengovsg/FormSG/pull/6747/commits/c77b2f3bfbeed6fbfb3043374511515dff927fb2)

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
tshuli: Ok, can go with this approach then

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
LinHuiqing: Done in [f03825e](https://github.com/opengovsg/FormSG/pull/6747/commits/f03825ed92343668572e7295863f26132d4ee342)! Thanks!

tshuli (approved):
lgtm! thanks huiqing!

PR #6751, opened 2023-09-26
fix: reinstate MyInfo verified field indicator for MyInfo child DOB

wanlingt:
## Problem

When doing tests to QA the MyInfo child fields, I realised that there was no `[MyInfo]` prefix (which represents that a field is MyInfo verified) for child Date of Birth fields, even though this data is obtained from MyInfo and should be verified.

## Solution

Remove a redundant `return` statement, which was preventing the field from being hashed. 

**Breaking Changes** 

    - This changes the field name in the email response

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-09-26 at 2 19 35 PM

**AFTER**:

Screenshot 2023-09-26 at 2 19 09 PM

## Tests

- [x] Fill in a form that contains the MyInfo child DOB field
- [x] Select a child, the child's DOB should be automatically prefilled
- [x] Submit the form
- [x] Check that in the email response, the DOB field name is prefixed with `[MyInfo]`

### DIFF
--- a/src/app/modules/myinfo/myinfo.util.ts
+++ b/src/app/modules/myinfo/myinfo.util.ts
@@ -112,7 +112,6 @@ function hashChildrenFieldValues(
       // Hence we format it here
       if (subField === MyInfoChildAttributes.ChildDateOfBirth) {
         myInfoFormattedValue = formatMyinfoDate(value)
-        return
       }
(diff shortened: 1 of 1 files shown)


foochifa (approved):
lgtm and tested! Thanks @wanlingt

PR #6754, opened 2023-09-28
feat(workspaces): remove form from workspace

foochifa:
## Problem

Closes FRM-1426

## Solution

Create a new route to remove forms from all workspaces, `POST /workspaces/remove`, with a body of `{ formIds }` to remove. This is in anticipation of the future dashboard feature which will allow selection of multiple forms and performing actions on all them.

Utilises the current `WorkspaceModel.removeFormIdsFromWorkspaces` method to remove given formIds from all workspaces under the current session user. 

Updates the FE row actions to remove form from workspace when selecting the current workspace

Additionally includes design review changes from office hour with @staceytan1998 

**Breaking Changes** 

## Tests

Functionality
- [ ] Move a form into a workspace
- [ ] Go to row actions -> workspace -> click on the current workspace that the form is in (it has a tick beside the name)
- [ ] Form should be removed from workspace
- [ ] Toast should pop up

Regression test (relating to removing forms from workspace)
- [ ] Move a form to a workspace
- [ ] Delete that form
- [ ] Ensure that it is no longer reflected in the total number of forms in the workspace
- [ ] (Using a different account) Create a form and add your other account as a collaborator
- [ ] In the other account, add the above form to a workspace
- [ ] In the previous account, remove your other account as collaborator
- [ ] The form should be removed from the workspace, and it is no longer reflected in the total number of forms


foochifa:
Heya @KenLSM! If you are free could you help me take a look at this PR? Other than the design stuffs I used the current existing `removeFormIdsFromWorkspaces` method in the schema which pulls each forms from all existing workspaces. Which is the fastest way of implementing this feature as it relies on existing code. This works as currently every form will only belong to 1 workspace at a time.

However, was wondering if it might cause complexity or unneeded time complexity down the line, thus whether we should pull only from the active workspace.

frontend/src/features/workspace/components/EmptyWorkspace/EmptyWorkspace.tsx
```diff
@@ -59,3 +60,23 @@ export const EmptyWorkspace = ({
...
+}: EmptyWorkspacePage) => (
+  <EmptyWorkspace
+    isLoading={isLoading}
+    handleOpenCreateFormModal={handleOpenCreateFormModal}
+    title={"You don't have any forms yet"}
+    subText={'Get started by creating a new form'}
+  />
+)
+
+export const EmptyNewWorkspace = ({ isLoading }: EmptyWorkspacePage) => (
```
KenLSM: /nit would prefer to have matching exported module with filename. Makes it easier to search

frontend/src/features/workspace/components/WorkspaceFormRow/RowActions/useRowAction.ts
```diff
@@ -79,14 +84,29 @@ export const useRowAction = (
...
+
   return {
     adminFormLink,
     previewFormLink,
     handleShareForm,
     handleDuplicateForm,
     handleCollaborators,
     handleDeleteForm,
-    handleMoveForm,
+    handleWorkspaceClick,
```
KenLSM: Im thinking that we should export both `handleRemoveFormFromWorkspaces` and `handleMoveFormToWorkspace` instead of combining this action. I was having some difficulty making sense of this action, as it does not tell me what it is doing[1].

I think it's ok for be not-DRY for this function and have the callers `RowActionsDrawer`, and `RowActionsDropDown` duplicate some logic for better readability.


[1] Primarily, I didn't understand what does `handleWorkspaceClick` do? Comparatively, actions like `handleLoginClick`, `handleStartButtonClick` can be inferred given the context.

src/app/modules/workspace/workspace.controller.ts
```diff
@@ -301,3 +252,40 @@ export const moveFormsToWorkspace: ControllerHandler<
...
+  return WorkspaceService.removeFormsFromAllWorkspaces({
+    formIds,
+    userId,
+  })
+    .map(() => res.sendStatus(StatusCodes.OK))
+    .mapErr((error) => {
+      logger.error({
+        message: 'Error deleting forms from all workspaces',
+        meta: {
+          action: 'deleteFormFromWorkspaces',
```
KenLSM: Should `meta.action` be `removeFormsFromWorkspaces` instead?

src/app/modules/workspace/workspace.service.ts
```diff
@@ -253,24 +240,24 @@ export const verifyWorkspaceAdmin = (
...
     WorkspaceModel.removeFormIdsFromAllWorkspaces({
       admin: userId,
-      formIds: [formId],
+      formIds,
     }),
     (error) => {
       logger.error({
         message: 'Database error encountered when archiving form',
         meta: {
           action: 'archiveForm',
```
KenLSM: Not from your changes, but in similar spirit as above, should `meta.action` be `removeFormsFromAllWorkspaces` instead?

src/app/modules/workspace/workspace.controller.ts
```diff
@@ -301,3 +252,40 @@ export const moveFormsToWorkspace: ControllerHandler<
       return res.status(statusCode).json({ message: errorMessage })
     })
 }
+
```
KenLSM, line 255: On L243 there's also the function name vs log `meta.action` mismatch. Should be `moveFormsToWorkspace` instead?

frontend/src/features/workspace/components/WorkspaceFormRow/RowActions/useRowAction.ts
foochifa: Ahh I see, I guess it can be quite confusing. Cause was thinking technically both actions are from clicking the list of workspaces dropdown. We move if its not the current workspace of the form but remove when you click on the current workspace (the one with the tick). 

I think I sacrificed some readability for the sake of too much DRY-ness for both here and the `EmptyWorkspace` component above. Will refactor them to make it clearer! Thanks Ken!

src/app/modules/workspace/workspace.controller.ts
foochifa: [b734172](https://github.com/opengovsg/FormSG/pull/6754/commits/b734172b81001f24780c9a374ae0ee1db95bdf97) 🙏

src/app/modules/workspace/workspace.service.ts
foochifa: [b734172](https://github.com/opengovsg/FormSG/pull/6754/commits/b734172b81001f24780c9a374ae0ee1db95bdf97) 🙏

src/app/modules/workspace/workspace.controller.ts
foochifa: Ahh thanks for catching all the logging inconsistencies 🙏

[b734172](https://github.com/opengovsg/FormSG/pull/6754/commits/b734172b81001f24780c9a374ae0ee1db95bdf97)

frontend/src/features/workspace/components/EmptyWorkspace/EmptyWorkspace.tsx
foochifa: [cac4a97](https://github.com/opengovsg/FormSG/pull/6754/commits/cac4a97683c5b3a7cdcaa5e96beb2d53d1991811)

frontend/src/features/workspace/components/WorkspaceFormRow/RowActions/useRowAction.ts
foochifa: [88c43f0](https://github.com/opengovsg/FormSG/pull/6754/commits/88c43f0dcaa4ce41d9d89718fc2372a98ffc7099) 

Have also added some comments and renamed the method to make more readable!

KenLSM (approved):
LGTM!

foochifa:
Thanks @KenLSM! made a slight fix to mobile margins after stacey's review! No other changes and will just merge this in!

PR #6759, opened 2023-10-02
feat: rename NRIC field to NRIC/FIN

tshuli:
## Problem
- Renames NRIC field to NRIC/FIN so it is clear to users that there is FIN validation
- Purely frontend change

## Tests
- [ ] Create a form with NRIC field before this change was pushed. Use the default field question of "NRIC".
- [ ] Deploy this change. Form submission successful on the existing NRIC field. In the submission, the field question is still "NRIC" instead of "NRIC/FIN".
- [ ] Now create a new NRIC/FIN field. Form submission successful on the new NRIC/FIN field, and in the submission, the field question is "NRIC/FIN"

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-10-02 at 12 49 28

**AFTER**:

Screenshot 2023-10-02 at 12 55 47

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,6 @@ jobs:
               - src/**
               - package.json
               - package-lock.json
-            
 
   install:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


LinHuiqing (approved):
lgtm and tested!

PR #6772, opened 2023-10-04
chore(ci): backend jest config

KenLSM:
## Problem

CI is too damn slow

## Solution

This probably requires us to tweak the numbers around, but seems like the number's I've set here is good enough for now.

[`--maxWorkers=2`](https://jestjs.io/docs/cli#--maxworkersnumstring):  this is now set to 2, which reflects the [number of CPUs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) that we are provisioned by github runner. This was previously set to 4 in `jest.config.js`. The intention is to reduce the amount of context switches that's required when the CPU switches between threads. 

~`--max-old-space-size=5376`: increased this to [75% of the amount](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes) of [RAM we're provisioned](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) by github's runner. Increasing this config defers when the JS Engine does its gc. Theoretically, this should shotern the time taken to test as it spends less time doing deep GCs.~

[`--workerIdleMemoryLimit=0.2`](https://jestjs.io/docs/configuration/#workeridlememorylimit-numberstring): setting this to 0.2 which seemed to be working. Identified that the CI crashes due to OOM, due to memory leaks after workers completes every test case, evident by the increase in memory consumption report from `--logHeapUsage`. 
Note: 0.2 was chosen as 0.4 still hits OOM.

Also, there's a new command `test:backend:ci` that's carved out specifically for `ci`-only configs. 

## Alternatives

I did some quick experiment with using [swc](https://swc.rs/docs/usage/jest) to transpile ts. It is indeed faster, and consumes less memory, but unfortunately it is not a drop-in replacement. There were tests failures (didn't do full run, but about 25% of the tests failed, issues were relating to import syntax) which requires us to make code changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,7 +156,9 @@ jobs:
       - run: npm ci --prefix serverless/virus-scanner
       - run: npm run test:backend:ci
         env:
-          NODE_OPTIONS: '--max-old-space-size=4096'
+          # https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes
+          # Recommends to be 3/4 of available memory, our runner has 7GB of memory and 0.75 * 1024 maps to 5376MB
+          NODE_OPTIONS: '--max-old-space-size=5376'
       - name: Coveralls
         uses: coverallsapp/github-action@master
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,7 +154,7 @@ jobs:
           path: '**/node_modules'
           key: ${{ runner.OS }}-node-modules-${{ hashFiles('**/package-lock.json') }}
       - run: npm ci --prefix serverless/virus-scanner
-      - run: npm run test:backend
+      - run: npm run test:backend:ci
         env:
           NODE_OPTIONS: '--max-old-space-size=4096'
       - name: Coveralls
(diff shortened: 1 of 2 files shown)


LinHuiqing:
*shorten the time taken to test

Also, how did you find how much RAM is provisioned by github's runner?

KenLSM:
> *shorten the time taken to test
> 
> Also, how did you find how much RAM is provisioned by github's runner?

7GB: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources -- added the link in the PR

KenLSM:
Hm seems like it still crashes with OOM. Will try reverting the mem increase

LinHuiqing:
still timing out :'( @KenLSM

foochifa:
Oh @KenLSM Do we want to keep the logging?

LinHuiqing (approved):
lgtm 👀 10 min is record time sia

LinHuiqing:
> Oh @KenLSM Do we want to keep the logging?

yea I saw this too, but I don't think it affects performance that much so should be fine? 🤔

KenLSM:
Yeah, I also think that this doesn't have too much of an impact on performance.

Generally, I try to have performance logging to allow for quicker debugging.

PR #6774, opened 2023-10-04
feat: add well-known domain

KenLSM:
## Problem

To support applepay, we needed to host some files in some specific urls to verify that we own the domain.

Closes FRM-1441

## Solution

Have the files be served from s3 through adding `/.well-known` as a known static asset url.

**Breaking Changes** 

## Tests

Regression
##### Existing well-known routes are not affected
- [ ] Ensure that [/singpass/.well-known/jwts.json](https://form.gov.sg/singpass/.well-known/jwks.json) does not return 4xx

##### Existing static and public routes are not affected
- [ ] Ensure that form.gov.sg is loaded with images and js running

#### Deploy Notes

**New aasa file in `.well-known`**:
- [ ] Upload `apple-developer-merchantid-domain-association` file into s3 static bucket
- [ ] Make sure that the [/.well-known/apple-developer-merchantid-domain-association](https://form.gov.sg/.well-known/apple-developer-merchantid-domain-association) does not return 404

### DIFF
--- a/src/app/loaders/express/index.ts
+++ b/src/app/loaders/express/index.ts
@@ -119,14 +119,16 @@ const loadExpressApp = async (connection: Connection) => {
   // API routes
   app.use('/api', ApiRouter)
 
-  app.get('/.well-known', catchNonExistentStaticRoutesMiddleware)
   // serve static assets. `dist/frontend` contains the root files as well as a `/static` folder
   // express.static calls next() if the file is not found
   app.use(express.static(path.resolve('dist/frontend'), { index: false }))
 
   // If requests for known static asset patterns were not served by
   // the static handlers above, middleware should try to fetch from s3 static bucket or else return 404s
-  app.get(/^\/(public|static)\//, catchNonExistentStaticRoutesMiddleware)
+  app.get(
+    /^\/(public|static|\.well-known)\//,
+    catchNonExistentStaticRoutesMiddleware,
+  )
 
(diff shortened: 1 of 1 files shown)


tshuli (commented):
thanks @KenLSM ! one minor clarifcation :)

do we expect these files to be frequently loaded? if yes, for performance optimisation it might be worthwhile loading this files onto EFS and using express.static to return the file, instead of retrieving from s3 as a proxy. (see  `src/app/routes/api/v3/corppass/corppass.routes.ts` as example)

- to also add regression test that this does not affect static serving from existing .well-known routes

KenLSM:
~I think these files should only be fetched once every week or so. Don't recall apple fetching them often!~

Need to check more! Might be fetched by the iOS device instead!

KenLSM:
> In macOS 11 and later and iOS 14 and later, apps request apple-app-site-association files from an Apple-managed content delivery network (CDN) specifically for associated domains, instead of directly from your web server. If the CDN has an old version of the file, or doesn’t already have a copy of the file, it connects to your web server to obtain the latest version. -- [developer.apple.com](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_associated-domains#:~:text=In%20macOS%2011,the%20latest%20version.)

It didn't mention whether the merchant domains AASAs will also be cached, but I think it should have similar behaviour. WDYT?

tshuli (approved):
lgtm, thanks @KenLSM

PR #6777, opened 2023-10-05
build: release v6.81.0

LinHuiqing:
## New

- feat(virus-scanner): frontend orchestration [`#6767`](https://github.com/opengovsg/FormSG/pull/6767)
- feat: add well-known domain [`#6774`](https://github.com/opengovsg/FormSG/pull/6774)
- build: merge release v6.80.0 into develop [`#6775`](https://github.com/opengovsg/FormSG/pull/6775)
- build: release v6.80.0 [`#6773`](https://github.com/opengovsg/FormSG/pull/6773)

## Dependencies

- fix(deps): bump zod from 3.22.2 to 3.22.3 in /shared [`#6768`](https://github.com/opengovsg/FormSG/pull/6768)
- fix(deps): bump zod from 3.21.4 to 3.22.3 [`#6769`](https://github.com/opengovsg/FormSG/pull/6769)

### feat(virus-scanner): frontend orchestration [`#6767`](https://github.com/opengovsg/FormSG/pull/6767)

@KenLSM 

##### Preperation

- [x] Go to https://www.eicar.org/download-anti-malware-testfile/ and download the test malicious file `eicar.com.txt`.
- [x] Find 2 non-malicious files.
- [x] In the DB `featureflags` collection, set the `encryption-boundary-shift-virus-scanner` flag as `{ enabled: true }`.
- [x] In [growthbook's feature page for "encryption-boundary-shift-virus-scanner"](https://app.growthbook.io/features/encryption-boundary-shift-virus-scanner), turn on the feature flag for the respective env and make sure the value returned will be ON.

##### Test that virus scanner works

- [x] Create a storage mode form with > 1 attachment field, with <= 1 of these attachment fields to be required.
- [x] Go to the respondent page. Open the network panel. Upload the test malicious file and submit the form. 
  - [x] The submission will pass.
  - [x] In the network panel, check that the status code of the request is 400.
  - [x] In the network panel, there should be another request that is sent immediately after to the storage submission v2 endpoint.
- [x] Go to the respondent page. Upload 1 non-malicious file.
  - [x] The submission should succeed.
  - [x] Check that the submission went through to the admin portal. Download and check that the file name and content is correct.
- [x] Go to the respondent page. Upload 2 non-malicious files.
  - [x] The submission should succeed.
  - [x] Check that the submission went through to the admin portal. Download and check that the file name and content for both files is correct.
- [x] Go to the respondent page. Open the network panel. Upload the malicious file for 1 field and non-malicious for the other.
[...]

PR #6778, opened 2023-10-05
fix: rm fallback from submit v2.1 mutation

LinHuiqing:
## Problem

Fallback for virus can mutation catches all errors. This is an issue as when malicious files are scanned, the fallback is triggered as well. This bug was identified by @KenLSM in https://github.com/opengovsg/FormSG/pull/6777.

## Solution

There are many points of failure for the mutation to fail. The original intention of the fallback mechanism is to try the submission again if there are any unexpected errors related to the virus scan itself (e.g. if lambda is down, if s3 uploads can't be done, etc). However, these block submissions anyway - the response to this should be to rollback using the growthbook flag.

As such, we want to remove the `onError` block.

**Breaking Changes** 

**Bug Fixes**:

- Malicious file scan triggering retry on submit fetch on storage mode form.

## Tests

- [ ] Go to a storage form with attachments.
- [ ] Submit a malicious file (get a test one from https://www.eicar.org/download-anti-malware-testfile/). An error message should be thrown and submission should be blocked.

### DIFF
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -579,25 +579,6 @@ export const PublicFormProvider = ({
                     timestamp,
                   })
                 },
-                onError: (error) => {
-                  // TODO(#5826): Remove when we have resolved the Network Error
-                  datadogLogs.logger.warn(
-                    `handleSubmitForm: submit with virus scan`,
-                    {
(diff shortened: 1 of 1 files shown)


LinHuiqing:
Decided to not remove the fallback for https://github.com/opengovsg/FormSG/pull/6777 as soft validation is useful (though not originally intended). Will be revisited!

PR #6787, opened 2023-10-11
chore: return InvalidFileKeyError if file not found

tshuli:
## Problem
- Returns 400 instead of 500 if file key not found in quarantine bucket

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
@@ -771,6 +771,8 @@ export const triggerVirusScanning = (
         })
 
         if (error instanceof ParseVirusScannerLambdaPayloadError) return error
+        else if (error.statusCode === StatusCodes.NOT_FOUND)
+          return new InvalidFileKeyError()
         else if (error.statusCode !== StatusCodes.BAD_REQUEST)
           return new VirusScanFailedError()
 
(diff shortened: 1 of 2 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -771,6 +771,8 @@ export const triggerVirusScanning = (
         })
 
         if (error instanceof ParseVirusScannerLambdaPayloadError) return error
+        else if (error.statusCode === StatusCodes.NOT_FOUND)
+          return new InvalidFileKeyError()
```
LinHuiqing: add custom message? the current one won't make sense in this case haha

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -771,6 +771,10 @@ export const triggerVirusScanning = (
         })
 
         if (error instanceof ParseVirusScannerLambdaPayloadError) return error
+        else if (error.statusCode === StatusCodes.NOT_FOUND)
+          return new InvalidFileKeyError(
+            'Invalid file key. The file must be uploaded first.',
```
LinHuiqing:
```suggestion
            'Invalid file key - file key is not found in the quarantine bucket. The file must be uploaded first.',
```

LinHuiqing (approved):
lgtm otherwise! thanks shuli for debugging this!

PR #6789, opened 2023-10-11
fix: add payment data to virus scanner endpoint

LinHuiqing:
## Problem

Payment forms will fail on the /submissions/storage v2.1 with 500 as the payment data is not passed to the endpoint. When there is recaptcha, the submission fails entirely as the recaptcha is retried.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Pass `formPaymentData` into the submit with virus scanning mutation.

## Tests

- [ ] Create a payment form with recaptcha. 
- [ ] Try to submit the payment form. It should succeed.


tshuli (approved):
lgtm

PR #6796, opened 2023-10-13
chore: improve nric/fin error message

tshuli:
## Problem
- Updates NRIC/FIN field error message to say 'Please enter a valid NRIC **/FIN** '

## Tests
- [ ] Create NRIC/FIN field, submit invalid FIN, should see 'Please enter a valid NRIC/FIN'


foochifa (approved):
TY shuli LGTM

PR #6808, opened 2023-10-17
fix(storybook): build crash, and missing NumberField customVal 

KenLSM:
## Problem

Picking off #6709, closes FRM-1340

1. Publish to chromatic fails due to build failure where `react-refresh` was not found. It was found that `@pwwwwh/*` package was hoisted up while react-refresh was not. Might be related to some phantom deps (possibly fixable by using `pnpm`?)
2. Webpack build failure issues
3. Storybook builds were also failing on render for some pages with `NumberFields` due to outdated configuration.

## Solution

1. Added `react-refresh` as `devDeps` on `frontend/package.json`
2. Updated node to `18` with `--openssl-legacy-provider` config 
3. Updated `customVal` config for `NumberFields` storybook. Thanks @justynoh for the fix!

**Breaking Changes** 


justynoh (approved):
Lgtm

PR #6819, opened 2023-10-19
fix(sgid-login): add graceful handling of errors when work email is invalid

KenLSM:
## Problem

BE crashes when a user tries to sign in with an invalid domain

**Breaking Changes** 


wanlingt (approved):
tested invalid work email test case

PR #6839, opened 2023-10-25
fix(workspaces): use svg for announcement and what's new

wanlingt:
## Problem

We want to use a static image for the announcement modal and what's new drawer graphics, instead of an animation.

## Solution

Replace the json with svg

- The What's New drawer already accepted a static image, from `image.url`
- For the announcement modal, I added a conditional which checks if an animation exists for the feature image (`image.animationData`). If it doesn't, it displays an Image instead - also taken from the `image.url`

## Before & After Screenshots

**AFTER**:
image

image


foochifa (approved):
Thanks @wanlingt! Also for updating the dates 🙏

PR #6841, opened 2023-10-25
fix: mobileno is always editable

wanlingt:
## Problem

The [Singpass data catalog](https://api.singpass.gov.sg/library/myinfo/business/implementation-myinfo-data) states that the mobile number field should always be editable. However, it's currently uneditable for SGID MyInfo.

## Solution

Move `MobileNoWithCountryCode` to the editable case

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- [ ] As an admin, create a form with the auth method "Singpass App-only with MyInfo" and add the "Mobile Number" MyInfo field
- [ ] Open the form and log in using a profile that has a mobile number, using the SGID Chrome extension (first profile Singaporean)
- [ ] The mobile number field should **not** be disabled
- [ ] Submit the form, the mobile number field should be reflected in the response

### DIFF
--- a/src/app/modules/sgid/sgid.adapter.ts
+++ b/src/app/modules/sgid/sgid.adapter.ts
@@ -195,7 +195,6 @@ export class SGIDMyInfoData
     if (!data || !fieldValue) return false
 
     switch (attr) {
-      case ExternalAttr.MobileNoWithCountryCode:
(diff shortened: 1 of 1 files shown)


tshuli (approved):
lgtm

PR #6843, opened 2023-10-26
fix(workspaces): remove form from collab workspace

foochifa:
## Problem

Closes FRM-1478

## Solution

Iterate thru the list of permissions, retrieve the user from the email, then remove the formId from all the workspaces under the user.

**Breaking Changes** 

    - Details ...

## Tests

- [ ] As a collaborator of the form, move the form into a workspace
- [ ] As the owner of the same form, delete the form
- [ ] The form should be removed from the collaborator's workspace, check that the total form number tally is correct.


KenLSM (approved):
LGTM

PR #6845, opened 2023-10-26
fix: use secret to sign and verify sgid jwt

wanlingt:
## Problem

SGID was not working on staging-alt as our `JWT_ALGORITHM` used to sign and verify the JWT for the profile was incompatible with the private key. We should decouple the private key used by SGID from the private key used to sign and verify the JWT.

Closes FRM-1479

## Solution

Change the algorithm used to HS256 (symmetric algorithm, only 1 secret is required).
Use a new env var, `SGID_JWT_SECRET`, to sign and verify the JWT

**Breaking Changes** 

## Tests

- [ ] Submit a form with SGID Singpass auth. You should be able to submit the form successfully.
- [ ] Submit a form with SGID MyInfo auth, that has at least 1 MyInfo field. You should be able to submit the form successfully.

## Deploy Notes

- `SGID_JWT_SECRET` : The secret key used to sign and verify JWT based on userinfo from sgID


tshuli (approved):
lgtm! thanks @wanlingt  😃

tshuli:
Added one test for SGID non-myinfo auth

PR #6850, opened 2023-10-27
fix: conditionally render POCDEX login

wanlingt:
## Problem

Closes FRM-1474

## Solution

Added a new endpoint `api/v3/intranet/is-intranet-check` so that we can check if an IP is an intranet IP on the Login page
- This involved moving the existing intranet service and middleware to a new `intranet` module (the existing intranet service is used in form.service.ts)

**Improvement**
- Added a check for `FormAuthType.SGID` and `FormAuthType.SGID_MyInfo` in `checkIsIntranetFormAccess` as forms with these authentication types cannot be accessed on GSIBs

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- [ ] On a GSIB, go to staging.form.gov.sg/login. You should **not** see the "Log in with singpass app" button
- [ ] On a non-GSIB computer, go to staging.form.gov.sg/login. You should see the "Log in with singpass app" button


KenLSM (approved):
LGTM! Thanks @wanlingt

PR #6862, opened 2023-10-31
chore(dockerfile): update mocktwilio to start asap

KenLSM:
## Problem

`mocktwilio` starts late on dev as it waits for `backend` to be first started (note: not completed). However, there's no need for `mocktwilio` to start late.

## Solution

Removed `depends_on: backend` for `mocktwilio`

**Breaking Changes** 


wanlingt (approved):
lgtm!

PR #6863, opened 2023-11-01
feat: optimise submission query

KenLSM:
## Problem

We have some DB queries that takes a long time to execute. These are found to be from querying the submission metadata for forms with a huge amount of submissions. 
Closes FRM-1495

## Solution

See [notion](https://www.notion.so/opengov/Slow-DB-Query-in-submission-metadata-f426de1f26e14d78a7480e72d55081ee?pvs=4) for more details

**Breaking Changes** 

## Tests

**Regression**
##### Storage mode submissions ordering should not change
- [ ] Create a storage mode form with multiple submissions (in excess of 30, so that we can test out the pagination skips + limit)
- [ ] Ensure that submission time on submission results page 1 are ordered in reverse chronological order

##### Storage mode submissions limit and skips should be respected
- [ ] Ensure that submission time on submission results page 2 are ordered in reverse chronological order
- [ ] Ensure that submission time on submission results page n are ordered in reverse chronological order
- [ ] Ensure that the total rows matches the total submission count.

### DIFF
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -32,35 +32,6 @@ import { PAYMENT_SCHEMA_ID } from './payment.server.model'
 
 export const SUBMISSION_SCHEMA_ID = 'Submission'
 
-function createAggregateQueryWithPaymentsProjectionParams(matchParams: any) {
-  return [
-    {
-      $match: {
-        submissionType: SubmissionType.Encrypt,
-        ...matchParams,
-      },
-    },
-    {
-      $lookup: {
-        from: 'payments',
-        localField: 'paymentId',
-        foreignField: '_id',
-        as: 'payments',
-      },
-    },
-    {
-      $project: {
-        _id: 1,
-        created: 1,
-        'payments.payout': 1,
-        'payments.completedPayment': 1,
-        'payments.amount': 1,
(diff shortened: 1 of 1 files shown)


src/app/models/submission.server.model.ts
```diff
@@ -285,14 +256,34 @@ EncryptSubmissionSchema.statics.findSingleMetadata = function (
...
-    .exec()
+  const pageResults: Promise<MetadataAggregateResult[]> = this.aggregate([
+    {
+      $match: {
+        submissionType: SubmissionType.Encrypt,
+        form: mongoose.Types.ObjectId(formId),
+        _id: mongoose.Types.ObjectId(submissionId),
+      },
+    },
+    { $limit: 1 },
```
tshuli, line 267: why do we need `{ $limit: 1 }` here; wouldn't there be at most 1 submission document returned for  a given submission id?

KenLSM: The query would continue to search for more even if we know that there should only be one -- it's still O(logn) through the index so theoretically we can remove this, but I feel that it's ok to leave it in incase we added some bad queries that prevented the use of index

tshuli (approved):
just one question - otherwise lgtm!

PR #6864, opened 2023-11-02
fix: add cloudflareinsights as allowable csp

KenLSM:
## Problem

`
Refused to load the script 'https://static.cloudflareinsights.com/beacon.min.js/<Truncated for brevity>...' because it violates the following Content Security Policy directive: <Truncated for brevity> ...
`
Closes FRM-1497

## Solution

Add `https://*.cloudflareinsights.com/` into allowed list

**Breaking Changes** 


tshuli (approved):
lgtm

PR #6867, opened 2023-11-06
chore: use non-testing branch for font-wqy-zenhei

KenLSM:
## Problem

`font-wqyt-zenhei` is [no longer on testing](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/54760) and has been promoted to the `community` stream.

Closes FRM-1504

## Solution

1. Reference `edge/community` instead of `edge/testing`
2. Switched to in-line repository flag to avoid tag for single-use installations


tshuli (approved):
lgtm! as discussed - looks like `--allow-untrusted` is not needed

PR #6875, opened 2023-11-07
fix(datepicker): webkit-related stylings

KenLSM:
## Problem

On mobile phones and safari, the text input for `DatePicker` 
- affects only `<DatePicker />`, other `<Input />`-related fields are not affected
- affects specifically when component is `disabled`

Screenshot 2023-11-07 at 5 18 56 PM
- text is set to gray but appears to be whitish-grey on chrome mobile, and safari browsers

Closes FRM-1473

## Solution

- Took reference from design-system and extracted aria fields and moved them down into the child components. 
- Fixed missing `borderRightRadius` on `<CalendarButton />`

Ideally, we should migrate to use design-system so that we don't have to patch issues like this. 

**Breaking Changes** 

## Before & After Screenshots

| Location/Modifier | Before | After |
|--------|--------|--------|
| Active | uat_active | local_active  |
| Filled | uat_filled | local_filled | 
| Error |  uat_error | local_error  | 
| Disabled | uat_disabled | local_disabled  |
| Admin | uat-admin | local-admin |
| Admin Active | uat-admin_active | local-admin_active  |
| Admin Hover | uat-admin_hover | local-admin_hover  |
| Admin Error | uat-admin_error  | local-admin_error | 

## Tests

**Regression Tests**
- [ ] Ensure that DatePicker fields cannot be interacted when field is disabled 
- [ ] Ensure that DatePicker fields cannot be selected with `Tab` when field is disabled
- [ ] Ensure that DatePicker input field is visible when viewed from Safari browser when field is disabled

### DIFF
--- a/frontend/src/components/DatePicker/DatePicker.stories.tsx
+++ b/frontend/src/components/DatePicker/DatePicker.stories.tsx
@@ -29,3 +29,14 @@ DatePickerDisallowManualInput.args = {
 
 export const Mobile = Template.bind({})
 Mobile.parameters = getMobileViewParameters()
+
+export const Prefilled = Template.bind({})
+Prefilled.args = {
+  defaultValue: new Date('2021-09-13'),
+  isDisabled: true,
+}
+
+export const Error = Template.bind({})
+Error.args = {
+  isInvalid: true,
+}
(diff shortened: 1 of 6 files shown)


wanlingt (approved):
lgtm thanks for fixing this!!

PR #6877, opened 2023-11-08
fix: encode product name and log error

sebastianwzq:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes** 

    - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/app/modules/payments/stripe.utils.ts
+++ b/src/app/modules/payments/stripe.utils.ts
@@ -410,7 +410,7 @@ const formatProductsDescriptionHtml = (products: ProductItem[]) => {
   let final =
     '<td class="Table-description Font Font--body" style="color: #525f7f;font-size: 15px;line-height: 36px;width: 100%;">'
   products.forEach((product) => {
-    const productName = encode(product?.data?.name)
+    const productName = encode(product?.data?.name || '')
     const productStr = `${productName} x ${product.quantity}<br>`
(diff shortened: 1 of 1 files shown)


src/app/modules/payments/stripe.utils.ts
```diff
@@ -410,7 +410,21 @@ const formatProductsDescriptionHtml = (products: ProductItem[]) => {
...
+        message: 'Received invalid request from Stripe webhook endpoint',
+        meta: {
+          action: 'formatProductsDescriptionHtml',
+          error: Error('Invalid product data'),
+        },
+      })
+    } else {
+      const productName = encode(product.data.name)
+      productStr = `${productName} x ${product.quantity}<br>`
+    }
```
timotheeg: The whole error logging seems overly defensive to me here 🤔. By the time we are at the display formatting stage (when this code runs), the data should be clean.If we want extra safety, we can just have something simple like this:

```typescript
encode(product?.data?.name || '')
```
but even that is probably not needed.

tshuli (approved):
lgtm

PR #6880, opened 2023-11-08
fix: add sanitisation function for products

KenLSM:
## Problem

Users can maliciously submit a product from FE to manipulate the product through scripts. This causes the product item evaluated on our BE to be incorrectly referencing the data from FE, instead of from the form payment products. 

## Solution

Don't trust FE.

Treat the product data from FE as dirty. Extract only the relevant details (quantity + selected) and replace the product data from the one in BE.

**Breaking Changes** 

## Tests

**Regression**
##### Payment by Product form submission should still work**
- [ ] Make a payment on a form with payment by products
  - [ ] Ensure that payment had succeeded
  - [ ] Ensure that payment title on payment summary displays correctly
  - [ ] Ensure that payment title on payment successfully page displays correctly
  - [ ] Ensure that payment title on receipt displays correctly

##### Variable Payment form submission should still work
- [ ] Make a payment on a form with Variable Payment
  - [ ] Ensure that payment had succeeded
  - [ ] Ensure that payment title on payment summary displays correctly
  - [ ] Ensure that payment title on payment successfully page displays correctly
  - [ ] Ensure that payment title on receipt displays correctly

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
@@ -196,7 +196,7 @@ export const getStripePaymentMethod = (
  * The payment products from incoming submission can be freely altered by the respondent
  * which could result in undesirable data seeded into our database
  * @param form
- * @param uncleanPaymentProducts
+ * @param dirtyPaymentProducts
  */
 export const sanitisePaymentProducts = (
   form: IPopulatedEncryptedForm,
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
@@ -1,5 +1,6 @@
 import _ from 'lodash'
 import moment from 'moment-timezone'
+import { Types } from 'mongoose'
 import Stripe from 'stripe'
 
 import {
@@ -201,13 +202,13 @@ export const sanitisePaymentProducts = (
   form: IPopulatedEncryptedForm,
   dirtyPaymentProducts: ProductItem[] | undefined,
 ): ProductItem[] | undefined => {
-  if (!dirtyPaymentProducts) return dirtyPaymentProducts
-  if (!form.payments_field.products) return dirtyPaymentProducts
+  if (!dirtyPaymentProducts) return
+  if (!form.payments_field.products) return
 
   const sanitisedProducts = form.payments_field.products
     .map((cleanProductData): ProductItem | null => {
-      const dirtyProduct = dirtyPaymentProducts.find(
-        ({ data }) => data._id === cleanProductData._id,
+      const dirtyProduct = dirtyPaymentProducts.find(({ data }) =>
+        (cleanProductData._id as unknown as Types.ObjectId).equals(data._id),
       )
       if (!dirtyProduct) return null
 
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import moment from 'moment-timezone'
 import Stripe from 'stripe'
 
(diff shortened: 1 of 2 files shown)


src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -387,3 +390,38 @@ export const formatMyInfoStorageResponseData = (
...
+const isNonEmpty = <T>(value: T | null | undefined): value is T => {
+  return value != null
+}
+
+/**
+ * Sanitizes the payment fields from the form and the incoming submission
+ * The payment products from incoming submission can be freely altered by the respondent
+ * which could result in undesirable data seeded into our database
+ * @param form
+ * @param uncleanPaymentProducts
```
KenLSM:
```suggestion
 * @param dirtyPaymentProducts
```

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -387,3 +390,38 @@ export const formatMyInfoStorageResponseData = (
     })
   }
 }
+
+const isNonEmpty = <T>(value: T | null | undefined): value is T => {
+  return value != null
+}
```
justynoh: does this actually satisfy the type promise? Just a sanity check, not sure how `==` and `!=` exactly work in js. 

If this is really the case then I don't know what to say lmao. whyyyyyy jsssss :(((

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -387,3 +390,38 @@ export const formatMyInfoStorageResponseData = (
...
+
+/**
+ * Sanitizes the payment fields from the form and the incoming submission
+ * The payment products from incoming submission can be freely altered by the respondent
+ * which could result in undesirable data seeded into our database
+ * @param form
+ * @param dirtyPaymentProducts
+ */
+export const sanitisePaymentProducts = (
```
justynoh, line 201: supernit: `sanitize` or `sanitise`? :)

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
```diff
@@ -186,3 +190,34 @@ export const getStripePaymentMethod = (
...
+
+  const sanitisedProducts = form.payments_field.products
+    .map((cleanProductData): ProductItem | null => {
+      const dirtyProduct = dirtyPaymentProducts.find(({ data }) =>
+        (cleanProductData._id as unknown as Types.ObjectId).equals(data._id),
+      )
+      if (!dirtyProduct) return null
+
+      return {
+        ..._.pick(dirtyProduct, ['selected', 'quantity']), // only selected and quantity are allowed to be passed through
```
justynoh, line 216: any particular reason to use lodash `pick` here as opposed to just `{ ...dirtyProduct, data: cleanProductData }` or for even more safety, `{ selected: dirtyProduct.selected, quantity: dirtyProduct.quantity, data: cleanProductData }` so that we can take advantage of the type system for typechecking? Since with `pick`, the keys are not typechecked.

justynoh (approved):
lgtm

src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
KenLSM: > any particular reason to use lodash pick here as opposed to just { ...dirtyProduct, data: cleanProductData }

My intention was to operate as an explicit whitelist of properties that we will read from `dirtyProduct`, as opposed to accepting all the fields that are passed in. I felt that these properties are critical thus, it would be better to have them explicitly inherited rather than implicitly accepted.


> Since with pick, the keys are not typechecked.

Lodash's `pick` does have typechecking, the selected property `U` must exist as a key in object `T`.

```pick<T extends object, U extends keyof T>(object: T, ...props: Array<Many<U>>): Pick<T, U>;```

PR #6900, opened 2023-11-15
feat(payments): add paynow payment method

KenLSM:
## Problem

Admins want to only support certain forms to only provide payment through paynow (and disallowing other payment methods).

Closes FRM-1515

## Solution

Added a `payment_methods: string[]` that can contain allowed payment methods
- Currently, only `Paynow: 'Paynow'` is supported

Backwards compatibility is supported through marking the new field as optional, and populated with `[]` when `form.payments_channel` is updated (during remove or add payment channel)

**Breaking Changes** 

**Features**:

- Supports `Paynow` only through backend configuration by adding `form.payments_channel.payment_methods: ["Paynow"]`

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-11-15 at 11 32 54 PM

**AFTER**:

Screenshot 2023-11-15 at 11 32 17 PM

## Tests

**Regression**
- [ ] Find an existing payment form
- [ ] Make a payment
- [ ] Ensure that both payment methods "Cards", and "Paynow" are available to be selected

**New features**
##### Newly created forms should have both "Cards" and "Paynow" payment methods
- [ ] Create a new payment form
- [ ] Make a payment
- [ ] Ensure that both payment methods "Cards", and "Paynow" are available to be selected

##### Forms with enabled toggle should only allow "Paynow" payment method
- [ ] Create a new payment form
- [ ] Upsert `form.payments_channel.payment_methods: ["Paynow"]` to the form on DB

```
  "payments_channel": {
    "channel": "Stripe",
    "target_account_id": "acct_****",
    "publishable_key": "pk_test_****",
    "payment_methods": [
      "Paynow"
    ]
  },
```
- [ ] Make a payment
- [ ] Ensure that only payment method "Paynow" is available to be selected

### DIFF
--- a/shared/types/payment.ts
+++ b/shared/types/payment.ts
@@ -20,7 +20,6 @@ export enum PaymentChannel {
 }
 
 export enum PaymentMethodType {
-  Unset = '',
   Paynow = 'Paynow',
 }
 
--- a/shared/types/payment.ts
+++ b/shared/types/payment.ts
@@ -18,6 +18,12 @@ export enum PaymentChannel {
   Stripe = 'Stripe',
   // for extensibility to future payment options
 }
+
+export enum PaymentMethodType {
+  Unset = '',
+  Paynow = 'Paynow',
+}
+
 export enum PaymentType {
   Fixed = 'Fixed',
   Variable = 'Variable',
--- a/shared/types/form/form.ts
+++ b/shared/types/form/form.ts
@@ -12,7 +12,7 @@ import {
 } from '../../constants/form'
 import { DateString } from '../generic'
 import { FormLogic, LogicDto } from './form_logic'
-import { PaymentChannel, PaymentType } from '../payment'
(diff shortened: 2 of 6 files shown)


shared/types/payment.ts
```diff
@@ -18,6 +18,12 @@ export enum PaymentChannel {
   Stripe = 'Stripe',
   // for extensibility to future payment options
 }
+
+export enum PaymentMethodType {
+  Unset = '',
```
wanlingt: can add a comment explaining what this means?

wanlingt (approved):
lgtm! just one comment about `Unset`, I don't know what it's for

shared/types/payment.ts
KenLSM: Good catch. I've missed this section when removing [Unset](e2f2281669e75496ec370b1f0d7e5ba1c92e2009). I'll update it!

KenLSM: 0dee01943

PR #6870, opened 2023-11-06
feat: myinfo for storage-mode

wanlingt:
## Problem

We want to make MyInfo fields available on storage-mode forms now that we've done the encryption boundary shift, and are able to validate form responses

Closes FRM-1430

## Solution

**Frontend**
- Allow admin to choose Singpass MyInfo or SGID MyInfo in a form's auth settings

**Backend**
- Add `MyInfo` and `SGID_MyInfo` to the form model
- Validate MyInfo hashes in `encrypt-submission.middleware`. This is done in the `validateStorageSubmission` step, before the form responses are encrypted. We also format the question name by prefixing it with [MyInfo] if it is verified in this step. Each subfield in a Child Record is listed as a new field in the form response.
- Inject the user's NRIC as verified content in `encrypt-submission.controller`

**Other fixes**
- Changed the login cookie name from `connect.sid` to `formsg.connect.sid` in local dev. 
  - Why: For MyInfo-authenticated forms, Mockpass replaces the `connect-sid` value which would log the admin out of their account. This was causing the new MyInfo playwright test to fail.
  - A separate ticket (FRM-1512) has been created to standardise the login cookie name to `formsg.connect.sid` across all environments.

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

## Before & After Screenshots
| Location/Modifier | Before | After |
|--------|--------|--------|
| Authentication settings | image | image|
| Duplicate myinfo-authenticated email-mode form as a storage-mode form | image | image |

## Tests

1. Submit normal storage mode form, without MyInfo fields (regression)
- [ ] Create a normal storage mode form with no authentication
- [ ] You should be able to submit the form successfully

**Singpass MyInfo**
2a. Submit storage mode form with Singpass MyInfo fields
- [ ] Create a normal storage mode form with Singpass MyInfo as the authentication type
- [ ] Add at least the following MyInfo fields: Name, DOB, Nationality, Vehicle Type, Child
- [ ] Add an attachment field
- [ ] Add a payment field
- [ ] Set up a webhook receiver and update the URL in the form's settings
- [ ] You should be able to submit the form successfully
- [ ] Navigate to the responses page, check that the response shows up
- [ ] Click into the response. All the results should be there
- [ ] Download the responses as a CSV. All the results should be there
- [ ] The form responses should be sent via webhook and decrypted successfully

2b. Duplicate email mode form as storage mode using the duplicate button
- [ ] Create an email mode form set to Singpass MyInfo as the authentication type
[...]

PR #6901, opened 2023-11-16
fix: remove myinfo child from storage mode

wanlingt:
## Problem

We're removing the myinfo child field from storage mode forms (which was originally implemented in #6870). This will be released when the webhook response has been properly formatted (FRM-1520)

## Solution

- Only display the child field option in the MyInfo Panel if the user has the betaflag for children, and if it's an email-mode form

**Breaking Changes** 

## Tests

- On a storage-mode form that's Myinfo authenticated, go to the form builder's MyInfo panel. You should not see Child records


tshuli (approved):
lgtm!

PR #6898, opened 2023-11-15
build: release v6.89.0

wanlingt:
## New

- fix: add learn more link [`#6897`](https://github.com/opengovsg/FormSG/pull/6897)
- chore: security upgrade axios to 1.6.2 [`#6893`](https://github.com/opengovsg/FormSG/pull/6893)
- feat: myinfo for storage-mode [`#6870`](https://github.com/opengovsg/FormSG/pull/6870)
- feat: announcement modal and what's new for myinfo storage-mode [`#6892`](https://github.com/opengovsg/FormSG/pull/6892)
- chore: remove eb shift frontend feature flags [`#6869`](https://github.com/opengovsg/FormSG/pull/6869)
- build: merge release 6.88.0 into develop  [`#6882`](https://github.com/opengovsg/FormSG/pull/6882)
- build: release v6.88.0 [`#6881`](https://github.com/opengovsg/FormSG/pull/6881)
- chore: bump version to v6.89.0 [`e511d27`](https://github.com/opengovsg/FormSG/commit/e511d27cee4cf0ad5f5687046dfb25577125942d)

## Dependencies

- chore(deps): bump axios from 1.2.1 to 1.6.0 in /frontend [`#6887`](https://github.com/opengovsg/FormSG/pull/6887)
- fix(deps): bump axios from 1.2.1 to 1.6.0 [`#6886`](https://github.com/opengovsg/FormSG/pull/6886)
- fix(deps): bump type-fest from 4.5.0 to 4.7.1 in /shared [`#6883`](https://github.com/opengovsg/FormSG/pull/6883)

## Dev-Dependencies

- chore(deps-dev): bump @types/lodash from 4.14.200 to 4.14.201 in /shared [`#6888`](https://github.com/opengovsg/FormSG/pull/6888)

### feat: myinfo for storage-mode [`#6870`](https://github.com/opengovsg/FormSG/pull/6870)
@KenLSM 

1. Submit normal storage mode form, without MyInfo fields (regression)
- [x] Create a normal storage mode form with no authentication
- [x] You should be able to submit the form successfully

**Singpass MyInfo**
 @tshuli 
2a. Submit storage mode form with Singpass MyInfo fields
- [x] Create a normal storage mode form with Singpass MyInfo as the authentication type
- [x] Add at least the following MyInfo fields: Name, DOB, Nationality, Vehicle Number, Child
- [x] Add an attachment field
- [x] Add a payment field
- [x] Set up a webhook receiver and update the URL in the form's settings
[...]

PR #6903, opened 2023-11-16
fix: hotfix v6.89.1 for proper error handling in encrypt-submission middleware

wanlingt:
## Problem

Docker containers were crashing upon because of this error `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client`

This was due to incorrect error handling - `res.status.json` was being set twice

## Solution

Set and send res.status.json only once


tshuli (approved):
lgtm! L433 to 456 can be taken out entirely:

1) In L455, we already return error. so whether or not it is hard validation, we are returning error response. So the log line that 'Error processing responses, but proceeding with submission as submission have been validated client-side', is no longer valid
2) But why do we return error - i think this was actually a mistake in the PR 😁 https://github.com/opengovsg/FormSG/pull/6898/files#diff-f6575f85355753bb9d49c6d9ff5953bd0f4cff0c5b138091f2c28b64c25149beR457

The original logic (altered by the PR) was that if there is no hard validation, then you should return next() if there is validation error.

TLDR - remove L433 to 456 and we should be good to go 😃

PR #6907, opened 2023-11-16
fix: omit isVisible property from webhook response

wanlingt:
## Problem

Remove redundant isVisible property from storage-mode response

## Solution

Do not return `isVisible` property for each form response.

**Note**: I originally wanted to use lodash's pick/omit, but this returns a Partial type which was incompatible with the return type needed for `req.body.responses` in encrypt-submission.middleware

**Another note**:
Other properties (screenshots below) returned in the webhook response are:
- `"myInfo": { "attr": <name of attribute> }`
- `"isUserVerified": boolean` (set to `true` if the field has been verified by the user, e.g. email and mobile number)
- `"signature": string` (for verified fields)
- `"isHeader": boolean` (set to `true` if it's a Heading field)

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:

Webhook responses
| Field | Screenshot |
| - | - |
| Normal field | image |
| Myinfo field | image |
| Verified field | image |
| Heading | image |

## Tests

- [ ] Create a storage mode form, with at least 1 of the following fields: short text field, Myinfo field, verified field, heading, attachment
- [ ] Set up a webhook server and update the webhook url in the form's settings
- [ ] Submit the form, check that the responses are decrypted by the webhook server and do not contain isVisible

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts
@@ -19,6 +19,10 @@ import {
   MapRouteErrors,
   SubmissionData,
 } from '../../../../types'
+import {
+  EncryptFormFieldResponse,
+  ParsedClearFormFieldResponse,
+} from '../../../../types/api'
 import { MapRouteError } from '../../../../types/routing'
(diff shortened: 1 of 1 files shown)


KenLSM (approved):
LGTM!

PR #6908, opened 2023-11-17
chore: always fallback to fetch

tshuli:
## Problem
- Currently, if submission fails due to axios network error, we trigger fetch fallback
- This is done via string matching of error message
- Fetch fallback does not seem to be triggered after last release

## Solution
- Always fallback to fetch as a retry mechanism


tshuli:
cause of the bug has been identified, closing this PR

PR #6917, opened 2023-11-21
fix(markdown): refine regex to handle newlines after indentation groups

KenLSM:
## Problem

Given
```
title 1
- item 1
- item 2

title 2
```
Title 2 will be rendered under the same indentation as `item 2` as it did not break out of the `<li>` group.
```
title 1
- item 1
- item 2

  title 2
```

Replacing `\n` with an empty space prevents markdown from detecting `\n\n` that signifies a break from the list item.

## Solution

Refining replacement of `\n` to exclude cases that are trying to break out of indentation groups.

There were two solutions considered.
1.  Provide options of having true markdown syntax for admins
2. Better qualification of regex

We've chosen to go with 2 as this qualification is not adding additional edgecase on top of what we already have. Providing true markdown will require additional learning curve for the admins.


tshuli (approved):
lgtm!

PR #6924, opened 2023-11-23
build: release 6.91.1 hotfix

tshuli:
## Problem
- This reverts commit e050bd645d06915338b2bad2b6189991da009702.
- This is because lookbehind regex is not compatible on older browsers


KenLSM (approved):
LGTM!

PR #6930, opened 2023-11-28
feat(FE): set secret key input to password type

KenLSM:
## Problem

Secret keys should be hidden

## Solution

Set `input="password"` on `<Input />`

**Breaking Changes** 

## Before & After Screenshots

Screenshot 2023-11-28 at 11 32 28 AM

## Tests

**Regression**
- [ ] Ensure secret key on storage mode form is shown as a password field (`*`)


wanlingt (approved):
lgtm!

wanlingt:
oh I'm not sure why chromatic is failing to render some tests though

PR #6937, opened 2023-11-29
fix: prevent generation of charts if number of responses exceed limit

sebastianwzq:
## Problem
Charts is lagging for forms with a large number of responses. The current limitation of 1000 responses is limited at the DB query level and although responses will not be computed for charts, they will still be decrypted, which could result in the lagginess.

Fixed another issue where chromium version upgrades are causing deployment to fail.

Closes FRM-1550

## Solution
Show empty charts and prevent decryption if total response count exceeds a configured limit.
[Update] Updating Chromium as the [Chromium version currently available on alpine3.18 is 119.0.6045.159-r0](https://pkgs.alpinelinux.org/packages?name=chromium&branch=v3.18&repo=&arch=&maintainer=)

**Breaking Changes** 
    - Details ...

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-11-30 at 1 28 26 PM

**AFTER**:
Screenshot 2023-11-30 at 1 28 12 PM

## Deployment
- [ ]  Add `charts-max-response-count` to Growthbook and set it to 1000

## Tests

- [ ] No charts will be showed if total response count exceeds limit.

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -9,7 +9,7 @@ RUN chown 1001:1001 /mongodb_data
 ENV CHROMIUM_BIN=/usr/bin/chromium-browser
 ENV NODE_ENV=development
 # --openssl-legacy-provider flag
-# A breaking change in the SSL provider was introduced in node 17. This caused 
+# A breaking change in the SSL provider was introduced in node 17. This caused
 # webpack 4 to break. This is an interim solution; we should investigate removing
 # this flag once angular has been removed and we have upgraded to CRA5 (which uses
 # webpack 5).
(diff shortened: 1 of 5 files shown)


wanlingt (approved):
lgtm! can add a deployment note to add this new feature flag to growthbook

Dockerfile.development
```diff
@@ -9,7 +9,7 @@ RUN chown 1001:1001 /mongodb_data
 ENV CHROMIUM_BIN=/usr/bin/chromium-browser
 ENV NODE_ENV=development
 # --openssl-legacy-provider flag
-# A breaking change in the SSL provider was introduced in node 17. This caused
```
ghost, line 12: Didn't make any whitespace changes not sure why they are reflected. Linting?

PR #6941, opened 2023-12-01
fix(markdown): handle newlines after indentation groups

wanlingt:
## Problem

Given
```
title 1
- item 1
- item 2

title 2
```
Title 2 will be rendered under the same indentation as `item 2` as it did not break out of the `<li>` group.
```
title 1
- item 1
- item 2

  title 2
```

Replacing `\n` with an empty space prevents markdown from detecting `\n\n` that signifies a break from the list item.

The previous PR which tackled this issue #6917 was reverted as older versions of Safari [could not handle Lookbehind in regular expressions](https://caniuse.com/js-regexp-lookbehind)

## Solution

Replace `&nbsp; \n` that comes after a list item with `\n`

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-12-03 at 9 41 11 PM

**AFTER**:
image

## Tests

- [ ] In the form builder page, create a new Paragraph question 
- [ ] For the Paragraph's content, insert the following (like the screenshot)
```
title 1
- item 1
- item 2
- item 3


title 2
```
- [ ] In the form builder, `title 2` should not be indented
- [ ] In the public form page, `title 2` should not be indented

### DIFF
--- a/frontend/src/components/MarkdownText/MarkdownText.tsx
+++ b/frontend/src/components/MarkdownText/MarkdownText.tsx
@@ -42,8 +42,8 @@ export const MarkdownText = ({
          *   \n: new line character
          */
         children
-          .replace(/\n/gi, '&nbsp; \n')
-          .replace(/(\n(-|\d+\.|\*)\s.*\n)(&nbsp; \n)/gi, '$1 \n')
+          .replace(/\n/g, '&nbsp; \n')
+          .replace(/(\n(-|\d+\.|\*)\s.*\n)(&nbsp; \n)/g, '$1 \n')
       )
     }
     return children
--- a/frontend/src/components/MarkdownText/MarkdownText.tsx
+++ b/frontend/src/components/MarkdownText/MarkdownText.tsx
@@ -23,7 +23,28 @@ export const MarkdownText = ({
   const processedRawString = useMemo(() => {
     // Create new line nodes for every new line in raw string so new lines gets rendered.
     if (multilineBreaks) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/MarkdownText/MarkdownText.tsx
```diff
@@ -23,7 +23,28 @@ export const MarkdownText = ({
...
+         *
+         *   \s: whitespace following the token, indentation groups must start with token followed by a whitespace character
+         *
+         *   .*: any character, any number of times, this is the actual text content of the line
+         *
+         *   \n: new line character
+         */
+        children
+          .replace(/\n/gi, '&nbsp; \n')
+          .replace(/(\n(-|\d+\.|\*)\s.*\n)(&nbsp; \n)/gi, '$1 \n')
```
ghost: Quite a smart approach actually! Wondering why we went with the lookbehind in the first place now. 😄 
nit: the `i` in `/gi` indicates case-insensitive matching, we can actually remove them since newline characters dont have a case.

sebastianwzq (approved):

frontend/src/components/MarkdownText/MarkdownText.tsx
wanlingt: good catch! removed here [bba36a8](https://github.com/opengovsg/FormSG/pull/6941/commits/bba36a895af473036de471b58bfaf9497ff9e1c5)

PR #6960, opened 2023-12-12
feat: multi-respondent forms

justynoh:
I'm sorry.

## Problem

We want to implement multi-respondent forms. This PR implements it.

Closes FRM-1539, Closes FRM-1599

## Solution

Crypto primitives are implemented in [the sdk library](https://github.com/opengovsg/formsg-javascript-sdk/pull/101). The new library class is `CryptoV3`.

As for this repo, here are the primary notes.

### What's new

**Shared**
- A new response DTO, `FieldResponsesV3`, is created. This new response DTO is the type representing the shape of responses sent from the frontend to the backend, and is the updated equivalent of `FieldResponse[]`. 
  - This mimics closely `FormFieldValues` from `frontend/src/templates/Field/types`, which encapsulates the types used for `react-hook-form`'s `useForm()`, which is the frontend form handler. 
  - This makes it easy for setting the defaults when future respondents need to load the form.
- A new form type, `FormResponseMode.Multirespondent` is created. The corresponding form DTOs are also added.
- A new submission type, `SubmissionType.Multirespondent` is created.  The corresponding submission DTOs are also added.

**Backend**
- `form.server.model` is updated to account for the new form type, with a new discriminator.
  - Along with this, relevant updates have been made to `form.*` and `admin-form.*`, where required.
- `submission.server.model` is updated to account for the new submission type, with a new discriminator and relevant functions. 
  - The implementation of these functions is very similar to that of `SubmissionType.Encrypt`, with the exception that the webhook-related statics are omitted (as webhooks are not enabled for MRF).
- New routes handled by functions in `multirespondent-submission.controller`. Note that both of these make use of the **new encryption scheme**.
  - `POST /:formId/submissions/multirespondent` accepts a new multirespondent submission, processes it and saves it to the DB.
  - `PUT /:formId/submisisons/:submissionId` accepts an updated existing muiltirespondent submission, processes it and saves it to the DB.

**Frontend**
- A new form type has been added in the `FormResponseOptions` screen, which is for multi-respondent forms. 
- Admin form building flow. Updates in various places of the form builder where only two modes were implicitly assumed.
  - Messages saying that certain things are disabled in certain modes (e.g. payments for email and MRF, webhooks for email and MRF, auth for MRF, etc.).
  - Attachment size limit for multirespondent forms is also 20MB.
- Admin response flow. Individual response page and decryption workers for CSV downloads are updated to make use of the **new decryption scheme**.
  - Here I want to point out that not very much has changed about the processing flow, and most of the storage mode flow has been reused.
  - A diagram below shows the new flow of how types transform the responses. 
    - In encrypt mode, `createResponsesArray()` is what transforms a `FormFieldValues` to `FieldResponse[]` by mapping the values using `transformInputsToOutputs()`. After decryption, the `FieldResponse[]` is processed by `processDecryptedContent()`, which is what's used to display the response on the individual response page. 
    - In MRF, `createResponsesV3()` is what transforms a `FormFieldValues` to a `FieldResponsesV3` via a fairly simple type transform. On decryption, the `FieldResponsesV3` then passes through `processDecryptedContentV3()`, which calls `transformInputsToOutputs()` to create a `FieldResponse[]` (or equivalent).

Multi-respondent forms - Encryption protocol (2)

- Respondent public form access flow. The original frontend path for first respondents remains the same. A new frontend path `/:formId/edit/:submissionId` is now opened, which reuses the same `PublicFormPage` component.
  - To account for this, the app router is updated. In the public form page, `:submissionId` param must now be read. Whether or not it's a completely new submission or a subsequent respondent is conditioned on whether the `:submissionId` param is provided (ref: `PublicFormProvider`).
  - If the `:submissionId` param is provided, the encrypted submission is retrieved from the backend and stored. A secret key field is displayed to request the user to input the submission secret key (ref: `FormFieldsContainer`).
- Respondent public form filling flow. Several changes are made here.
  - Firstly, the loaded form now takes `form_fields` and `form_logics` from the submission object, if the `:submissionId` param is provided. This is how consistency is achieved across respondents (ref: `PublicFormProvider`).
  - Also, all field schemas must now be passed the `responseMode` prop which is used to check for requiredness. A field response is required only if the response mode is NOT `FormResponseMode.Multirespondent`, and the field is marked required by the admin (i.e. if the response mode if `FormResponseMode.Multirespondent`, empty response is always accepted). This accounts for all the changes in `templates/Field`.

### What's moved

**Shared**
- A number of DTOs which were named `StorageModeSubmissionXYZ` have been renamed to `SubmissionXYZ` to reflect their shared nature with multirespondent submission, where necessary. 

**Backend**
- Most of the functions and values in `encrypt-submission.*` are moved to `submission.*`, since they are now generally used for both `encrypt-submission.*` and `multirespondent-submission.*`. 
  - The implementation of these functions generally looks the same, with the exception of handling here and there for now needing to handle two types of submissions. As such, most of the implementations will be similar, except for where there may be `switch (submissionType) { case SubmissionType.Encrypt: /*...*/ case SubmissionType.Multirespondent: /*...*/ }` in places. 
  - The same is true for `receiver.*`.
- Also, as a result, some routes had to be updated. For the majority of routes, this only meant repointing the routes from their `encrypt-submission.controller` to their `submisison.controller` equivalent. The route path did not change.
  - However there is only one route in particular with a path change. `/:formId/submissions/storage/get-s3-presigned-post-data`, which is used to get presigned URLs to post files to the quarantine bucket, is now renamed to `/:formId/submissions/get-s3-presigned-post-data`, in light of the fact that it is now shared between storage mode and MRF.

**Frontend**
(Nothing really _moved_, per se, only added handling for multirespondent forms.)

**Breaking Changes** 

## Merge notes

Before merging this PR, the relevant PR in the formsg-javascript-sdk repo should be merged. A new release should be cut (likely as `v0.12.0-alpha.1`). Then,`package.json` and `frontend/package.json` should be updated to import these library versions.

When the feature is generally released, we can bump the formsg-javascript-sdk version to `v0.12.0` and bump `package.json` and `frontend/package.json` accordingly.


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/field-panels/PaymentsInputPanel/PaymentsInputPanel.tsx
```diff
@@ -475,7 +475,13 @@ export const PaymentsInputPanel = (): JSX.Element | null => {
   }, [paymentsField, resetData, setData, setToEditingPayment, setToInactive])
 
   const paymentDisabledMessage = !isEncryptMode ? (
-    <Text>Payments are not available in email mode forms.</Text>
+    <Text>
```
ghost: nit: Might be easier to just say that payments is only available in storage mode so we don't always have to change this message when we add a new mode

frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -82,24 +86,55 @@ export const getDecryptedSubmissionById = async ({
...
+      break
+    }
+    case SubmissionType.Multirespondent: {
+      const decryptedContent = formsgSdk.cryptoV3.decrypt(secretKey, {
+        encryptedContent: encryptedSubmission.encryptedContent,
+        encryptedSubmissionSecretKey:
+          encryptedSubmission.encryptedSubmissionSecretKey,
+        version: encryptedSubmission.version,
+      })
+      if (!decryptedContent) throw new Error('Could not decrypt the response')
```
ghost: nit: Could be beneficial to throw a different error so we can differentiate decryption errors easily

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponseNavbar.tsx
```diff
@@ -128,7 +128,7 @@ export const IndividualResponseNavbar = (): JSX.Element => {
       <Flex gridArea="respondent" justify="center" align="center">
         <Skeleton isLoaded={!isLoading}>
           <Text textStyle="h2" as="h2">
-            Respondent
+            Response
```
ghost, line 131: Can consider renaming `currentRespondentNumber` variable as well so it's easier to read

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -150,6 +154,33 @@ export const IndividualResponsePage = (): JSX.Element => {
               </Skeleton>
             </Stack>
           )}
+          {form?.responseMode === FormResponseMode.Multirespondent && (
```
ghost, line 180: nit: Beneficial to return error early if `useIndividualSubmission` or `useAdminForm` fails to fetch data

src/app/routes/api/v3/forms/public-forms.submissions.routes.ts
```diff
@@ -66,17 +68,55 @@ PublicFormsSubmissionsRouter.route(
...
- * @route POST /forms/:formId/submissions/storage/get-s3-presigned-post-data
+ * @route POST /forms/:formId/submissions/get-s3-presigned-post-data
  * @param response.body.required - contains field ids and sizes of attachments
  * @returns 200 - presigned post data generated
  * @returns 400 - ids are invalid or attachment size exceeds limit
  * @returns 500 - failed to generate presigned post data
  */
 PublicFormsSubmissionsRouter.route(
-  '/:formId([a-fA-F0-9]{24})/submissions/storage/get-s3-presigned-post-data',
+  '/:formId([a-fA-F0-9]{24})/submissions/get-s3-presigned-post-data',
```
justynoh: retain /storage for a while, mark as deprecated and open linear ticket to clean up later

src/app/models/submission.server.model.ts
```diff
@@ -430,13 +436,196 @@ EncryptSubmissionSchema.statics.findEncryptedSubmissionById = function (
...
+    type: String,
+    trim: true,
+    required: true,
+  },
+  encryptedContent: {
+    type: String,
+    trim: true,
+    required: true,
+  },
+  verifiedContent: {
```
justynoh: delete this

src/app/models/submission.server.model.ts
```diff
@@ -430,13 +436,196 @@ EncryptSubmissionSchema.statics.findEncryptedSubmissionById = function (
     .exec()
 }
 
+export const MultirespondentSubmissionSchema = new Schema<
+  IMultirespondentSubmissionSchema,
+  IMultirespondentSubmissionModel
+>({
+  //TODO(MRF): Clean this up
```
justynoh: add more details

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
```diff
@@ -99,3 +110,35 @@ export const processDecryptedContent = (
...
+  form_fields: FormFieldDto[],
+  decrypted: DecryptedContentV3,
+): Promise<VerifiedFormField[]> => {
+  const { responses } = decrypted
+
+  // Convert decrypted content into displayable object.
+  const displayedContent = form_fields
+    .map((ff) => {
+      const response = responses[ff._id]
+      if (!response) return null
```
ghost: nit: good to log such cases since we don't have tests for this

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -89,26 +89,49 @@
...
-
-      if (!decryptedObject) {
-        throw new Error('Invalid decryption')
+      let decryptedSubmission
+      switch (submission.submissionType) {
+        case SubmissionType.Encrypt: {
+          const decryptedObject = formsgSdk.crypto.decrypt(secretKey, {
+            encryptedContent: submission.encryptedContent,
+            verifiedContent: submission.verifiedContent,
+            version: submission.version,
```
ghost, line 110: what's version for?

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -89,26 +89,49 @@
...
+          break
+        }
+        case SubmissionType.Multirespondent: {
+          const decryptedObject = formsgSdk.cryptoV3.decrypt(secretKey, {
+            encryptedSubmissionSecretKey:
+              submission.encryptedSubmissionSecretKey,
+            encryptedContent: submission.encryptedContent,
+            version: submission.version,
+          })
+          if (!decryptedObject) throw new Error('Invalid decryption')
```
ghost: nit: Different error message so we can tell them apart

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -89,26 +89,49 @@
...
+          decryptedSubmission = await processDecryptedContentV3(
+            submission.form_fields,
+            decryptedObject,
+          )
+          break
+        }
+        default: {
+          const _: never = submission
```
ghost, line 136: nice type assertion

frontend/src/features/admin-form/settings/SettingsPaymentsPage.tsx
```diff
@@ -10,7 +10,7 @@ export const SettingsPaymentsPage = (): JSX.Element => {
 
   // Payments are unsupported in email mode; show message.
```
ghost: nit: update comment too

frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx
```diff
@@ -10,7 +10,7 @@ export const SettingsWebhooksPage = (): JSX.Element => {
 
   // Webhooks are unsupported in email mode; show message.
```
ghost: nit: update comment too

frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthUnsupportedMsg.tsx
```diff
@@ -0,0 +1,23 @@
+import { Flex, Text } from '@chakra-ui/react'
+
+import { GUIDE_ENABLE_SPCP } from '~constants/links'
+import Link from '~components/Link'
+
+import { SettingsUnsupportedSvgr } from '~features/admin-form/settings/svgrs/SettingsUnsupportedSvgr'
+
+export const AuthUnsupportedMsg = (): JSX.Element => {
```
ghost, line 8: nit: `UnsupportedAuthenticationMessage`

src/app/models/submission.server.model.ts
```diff
@@ -430,13 +436,196 @@ EncryptSubmissionSchema.statics.findEncryptedSubmissionById = function (
...
+    form: formId,
+    ...createQueryWithDateParam(dateRange?.startDate, dateRange?.endDate),
+  }
+  return (
+    this.find(streamQuery)
+      .select({
+        submissionType: 1,
+        form_fields: 1,
+        form_logics: 1,
+        encryptedSubmissionSecretKey: 1,
```
wanlingt, line 574: would we need `submissionPublicKey: 1,` here too?

frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -74,6 +80,27 @@
...
+            const selected = Object.values(CountryRegion).find(
+              (option) => option.toUpperCase() === previousResponse.answer,
+            )
+            if (selected) {
+              acc[field._id] = selected
+            }
+            break
+          }
+          case BasicField.Attachment:
+          //TODO(MRF): Handling of attachments by respondent 2+
```
ghost: missing break?

frontend/package.json
```diff
@@ -11,7 +11,7 @@
     "@emotion/styled": "^11.6.0",
     "@floating-ui/react-dom-interactions": "^0.9.3",
     "@growthbook/growthbook-react": "^0.17.0",
-    "@opengovsg/formsg-sdk": "^0.11.0",
+    "@opengovsg/formsg-sdk": "github:opengovsg/formsg-javascript-sdk#feat/mrf-crypto",
```
KenLSM: Marking as TODO (Keeping here to remind me before merge)

frontend/src/components/SecretKeyVerificationInput/SecretKeyVerificationInput.tsx
```diff
@@ -0,0 +1,169 @@
...
+import { GUIDE_SECRET_KEY_LOSS } from '~constants/links'
+import { useIsMobile } from '~hooks/useIsMobile'
+import formsgSdk from '~utils/formSdk'
+
+import Button from '../Button'
+import FormLabel from '../FormControl/FormLabel'
+import Link from '../Link'
+
+const SECRET_KEY_NAME = 'secretKey'
+const SECRET_KEY_REGEX = /^[a-zA-Z0-9/+]+={0,2}$/
```
KenLSM: /nit should refer to this as base64 regex instead as we're checking if this is a base64 regex if I'm getting this right

frontend/src/features/admin-form/responses/components/SecretKeyVerification/SecretKeyVerification.tsx
```diff
@@ -1,115 +1,10 @@
...
+import { Container, Skeleton, Stack, Text } from '@chakra-ui/react'
 import simplur from 'simplur'
 
-import { GUIDE_SECRET_KEY_LOSS } from '~constants/links'
-import { useIsMobile } from '~hooks/useIsMobile'
-import formsgSdk from '~utils/formSdk'
-import Button from '~components/Button'
-import FormLabel from '~components/FormControl/FormLabel'
-import Link from '~components/Link'
+import SecretKeyVerificationInput from '~components/SecretKeyVerificationInput'
```
KenLSM, line 4: Great refactoring!

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -150,6 +154,33 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+            <>
+              <Stack
+                spacing={{ base: '0', md: '0.5rem' }}
+                direction={{ base: 'column', md: 'row' }}
+              >
+                <Text as="span" textStyle="subhead-1">
+                  Response link:
+                </Text>
+                <Skeleton isLoaded={!isLoading && !isError}>
+                  {`${window.location.protocol}//${window.location.host}/${formId}/edit/${submissionId}`}
```
KenLSM: Should refactor this out to utils

frontend/src/features/admin-form/settings/components/FormStatusToggle.tsx
```diff
@@ -49,7 +49,7 @@ export const FormStatusToggle = (): JSX.Element => {
 
     if (
       nextStatus === FormStatus.Public &&
-      responseMode === FormResponseMode.Encrypt
+      responseMode !== FormResponseMode.Email
```
KenLSM: Better to check on `Encrypt | Multirespondent` which is how L69-70 decides render the controlled modal.

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentsUnsupportedMsg.tsx
```diff
@@ -1,21 +1,33 @@
...
   const { data: paymentGuideLink } = usePaymentGuideLink()
+  const modeText =
+    responseMode === FormResponseMode.Email
+      ? 'email mode'
+      : 'multi-respondent forms'
   return (
     <Flex justify="center" flexDir="column" textAlign="center">
       <Text textStyle="h2" as="h2" color="primary.500" mb="1rem">
-        Payments are not available in Email mode
+        Payments are not available in {modeText}
```
KenLSM: Similar with above, might make sense to list `Payments is only available in storage mode` to make this more maintainable. @staceytan1998

frontend/src/features/admin-form/settings/components/WebhooksSection/WebhooksUnsupportedMsg.tsx
```diff
@@ -1,20 +1,33 @@
...
+  const modeText =
+    responseMode === FormResponseMode.Email
+      ? 'email mode'
+      : 'multi-respondent forms'
+
   return (
     <Flex justify="center" flexDir="column" textAlign="center">
       <Text textStyle="h2" as="h2" color="primary.500" mb="1rem">
-        Webhooks are not available in Email mode
+        Webhooks are not available in {modeText}
```
KenLSM: Similar to https://github.com/opengovsg/FormSG/pull/6960/files#r1426824941

frontend/src/features/public-form/mutations.ts
```diff
@@ -98,25 +98,16 @@ export const usePublicFormMutations = (
...
-    {
-      onError: (error: Error) => {
-        toast({ status: 'danger', description: error.message })
-      },
-    },
-  )
-
-  const submitStorageModeClearFormWithVirusScanningMutation = useMutation(
-    async (args: Omit<SubmitStorageFormClearArgs, 'formId'>) => {
+  const useSubmitClearFormWithVirusScanningMutation = (f: any) =>
```
KenLSM: Note: very tricky typing here

frontend/src/pages/Landing/Payments/LandingPaymentsPage.tsx
```diff
@@ -264,6 +264,7 @@ export const LandingPaymentsPage = (): JSX.Element => {
                       sendOnboardingEmailMutation.isSuccess,
                     placeholder: 'Enter your email',
                   }}
+                  responseMode={FormResponseMode.Email}
```
KenLSM: Might want to consider if we can remove `responseMode` from `<EmailField />`. Might not need the fields to know about the `responseMode`, cursory search suggest that it is used to enforce `required` validation. 

```
const requiredSingleAnswerValidationFn =
-  (schema: Pick<FieldBase, 'required'>, responseMode?: FormResponseMode) =>
+  (schema: Pick<FieldBase, 'required'>, forceRequired?: boolean) =>
```

Is there any specific reason aside from the above?

KenLSM: Similar for the rest of the fields (i.e., `frontend/src/templates/Field/Dropdown/DropdownField.tsx`)

frontend/src/utils/fieldValidation.ts
```diff
@@ -78,30 +79,39 @@ type MinimumFieldValidationPropsEmailAndMobile<T extends FieldBase> = Omit<
...
   schema: MinimumFieldValidationPropsEmailAndMobile<T>,
+  responseMode: FormResponseMode,
 ) => RegisterOptions
 
 const requiredSingleAnswerValidationFn =
-  (schema: Pick<FieldBase, 'required'>) => (value?: SingleAnswerValue) => {
-    if (!schema.required) return true
+  (schema: Pick<FieldBase, 'required'>, responseMode?: FormResponseMode) =>
+  (value?: SingleAnswerValue) => {
+    if (!schema.required || isMultirespondentForm(responseMode)) return true
```
KenLSM: I'm struggling to figure out why `isMultirespondentForm(responseMode)` should `return true` here

package.json
```diff
@@ -60,7 +60,7 @@
     "@babel/runtime": "^7.20.13",
     "@faker-js/faker": "^8.0.1",
     "@joi/date": "^2.1.0",
-    "@opengovsg/formsg-sdk": "^0.10.0",
+    "@opengovsg/formsg-sdk": "github:opengovsg/formsg-javascript-sdk#feat/mrf-crypto",
```
KenLSM: Marking as TODO (Keeping here to remind me before merge)

frontend/src/pages/Landing/Payments/LandingPaymentsPage.tsx
justynoh: Hmmmm. that's fair. I did think about whether we want to just have a generic `requiredOverride` prop or something similar. Not too sure how to name it, let me think a bit on this.

frontend/src/utils/fieldValidation.ts
justynoh: ah, `true` means that the value is accepted, while returning a string means the value is not accepted (with the string as the error message). 

So if the form is MRF, we always accept blank answers. Otherwise, the form is email or encrypt, in which case we only accept if the question is optional (i.e. `!schema.required`).

src/app/modules/submission/submission.service.ts
```diff
@@ -194,6 +615,295 @@ export const doesSubmissionIdExist = (
...
+    (modelToUse) =>
+      ResultAsync.fromPromise(
+        modelToUse.findSingleMetadata(formId, submissionId),
+        (error) => {
+          logger.error({
+            message: 'Failure retrieving metadata from database',
+            meta: {
+              action: 'getSubmissionMetadata',
+              formId,
+              submissionId,
```
wanlingt, line 642: let's log the `responseMode` too

src/app/modules/submission/submission.service.ts
```diff
@@ -194,6 +615,295 @@ export const doesSubmissionIdExist = (
...
+    (modelToUse) =>
+      ResultAsync.fromPromise(
+        modelToUse.findAllMetadataByFormId(formId, { page }),
+        (error) => {
+          logger.error({
+            message: 'Failure retrieving metadata page from database',
+            meta: {
+              action: 'getSubmissionMetadataList',
+              formId,
+              page,
```
wanlingt, line 668: log responseMode too (and all other instances where this has been added)

src/app/modules/submission/submission.types.ts
```diff
@@ -81,3 +87,75 @@ export type ProcessedFieldResponse =
   | ProcessedTableResponse
   | ProcessedAttachmentResponse
   | ProcessedChildrenResponse
+
+/**
+ * Vius scanner types
```
wanlingt: nit: Virus

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -0,0 +1,324 @@
...
+const asyncVirusScanning = (
+  responses: IdTaggedParsedClearAttachmentResponseV3[],
+): ResultAsync<
+  IdTaggedParsedClearAttachmentResponseV3,
+  | VirusScanFailedError
+  | DownloadCleanFileFailedError
+  | MaliciousFileDetectedError
+>[] =>
+  responses.map((response) =>
+    triggerVirusScanThenDownloadCleanFileChain(response.answer).map(
```
wanlingt, line 156: could you remind me why we don't need `isQuarantinedAttachmentResponse` here? (i'm comparing to `encrypt-submission.middleware`) 
Is it because we do the check (L211) in scanAndRetrieveAttachments later on?

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -0,0 +1,324 @@
...
+  req: ProcessedMultirespondentSubmissionHandlerRequest,
+  res: Parameters<ProcessedMultirespondentSubmissionHandlerType>[1],
+  next: NextFunction,
+) => {
+  const formDef = req.formsg.formDef
+  const formPublicKey = formDef.publicKey
+  const responses = req.body.responses
+
+  const attachmentsMap: Record<string, Buffer> = {}
+  const strippedAttachmentResponses: Record<string, any> = {}
```
wanlingt, line 281: what is `strippedAttachmentResponses` used for?

frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -74,6 +80,27 @@
...
+          default:
+            acc[field._id] = previousResponse.answer as FormFieldValue
+        }
+        return acc
```
ghost, line 102: Q: This means that MPF responses will now take precedence over prefills. Prefills will still work for first respondents but subsequent respondents will not see any prefills

frontend/src/features/public-form/mutations.ts
```diff
@@ -98,25 +98,16 @@
...
```
sebastianwzq, line 101:
```typescript
  type SubmitStorageModeClearFormWithVirusScanningType = (
    args: SubmitStorageFormWithVirusScanningArgs,
) => Promise<SubmissionResponseDto>
type SubmitMultirespondentFormType = (
    args: Omit<SubmitMultirespondentFormWithVirusScanningArgs, 'publicKey'>,
) => Promise<SubmissionResponseDto>
type UpdateMultirespondentSubmissionType = (
    args: Omit<SubmitMultirespondentFormWithVirusScanningArgs, 'publicKey'> & {
        submissionId?: string
    },
) => Promise<SubmissionResponseDto>
type SubmitClearFormWithVirusScanningMutationArgs =
    | SubmitStorageModeClearFormWithVirusScanningType
    | SubmitMultirespondentFormType
    | UpdateMultirespondentSubmissionType
```

frontend/src/features/public-form/mutations.ts
```diff
@@ -98,25 +98,16 @@
     },
   )
```
ghost, line 101:
```typescript
  type SubmitStorageModeClearFormWithVirusScanningType = (
    args: SubmitStorageFormWithVirusScanningArgs,
) => Promise<SubmissionResponseDto>
type SubmitMultirespondentFormType = (
    args: Omit<SubmitMultirespondentFormWithVirusScanningArgs, 'publicKey'>,
) => Promise<SubmissionResponseDto>
type UpdateMultirespondentSubmissionType = (
    args: Omit<SubmitMultirespondentFormWithVirusScanningArgs, 'publicKey'> & {
        submissionId?: string
    },
) => Promise<SubmissionResponseDto>
type SubmitClearFormWithVirusScanningMutationArgs =
    | SubmitStorageModeClearFormWithVirusScanningType
    | SubmitMultirespondentFormType
    | UpdateMultirespondentSubmissionType

const useSubmitClearFormWithVirusScanningMutation = (
    f: SubmitClearFormWithVirusScanningMutationArgs,
) => ...


```

frontend/src/features/public-form/queries.ts
```diff
@@ -30,3 +36,29 @@ export const usePublicFormView = (
...
+
+/**
+ * @precondition Must be wrapped in a Router as `useParam` is used.
+ */
+export const useEncryptedSubmission = (
+  formId: string,
+  submissionId?: string,
+  /** Extra override to determine whether query is enabled */
+  enabled = true,
+): UseQueryResult<MultirespondentSubmissionDto | undefined, ApiError> => {
```
ghost: nit: to remove undefined in return val.


```typescript
export const useEncryptedSubmission = (
  formId: string,
  submissionId?: string,
  enabled = true,
): UseQueryResult<MultirespondentSubmissionDto, ApiError> => {
  // no submissionID
  if (!submissionId) {
    return useQuery(
      publicFormKeys.submission(formId, submissionId),
      undefined,
      {
        staleTime: Infinity,
        enabled: MONGODB_ID_REGEX.test(formId) && enabled,
      },
    )
  }

  // submissionId has a value
  return useQuery(
    publicFormKeys.submission(formId, submissionId),
    () => getMultirespondentSubmissionById({ formId, submissionId }),
    {
      staleTime: Infinity,
      enabled : MONGODB_ID_REGEX.test(formId) && MONGODB_ID_REGEX.test(submissionId) && enabled
    },
  )
}
```

frontend/src/features/public-form/utils/createSubmission.ts
```diff
@@ -180,6 +189,55 @@ export const createClearSubmissionWithVirusScanningFormData = (
...
+ * Used for MRF submissions v3 (after virus scanning).
+ * @returns formData containing form responses and attachments.
+ * @throws Error if form inputs are invalid or contain malicious attachment(s).
+ */
+export const createClearSubmissionWithVirusScanningFormDataV3 = (
+  formDataArgs: CreateMultirespondentSubmissionFormDataArgs,
+  fieldIdToQuarantineKeyMap: FieldIdToQuarantineKeyType[],
+) => {
+  const { formFields, formInputs, ...formDataArgsRest } = formDataArgs
+
```
ghost, line 202: q: missed this ytd, do we perform virus scanning for mrf attachments?

frontend/src/features/public-form/utils/createSubmission.ts
```diff
@@ -284,7 +420,8 @@ const encryptAttachment = async (
     }
 
     return { id, encryptedFile: encodedEncryptedAttachment }
-  } catch (error) {
+  } catch (err) {
```
ghost, line 423: why do we need to do this?

frontend/src/features/public-form/utils/decryptSubmission.ts
```diff
@@ -0,0 +1,35 @@
...
+  secretKey,
+}: {
+  submission?: MultirespondentSubmissionDto
+  secretKey?: string
+}):
+  | (Omit<MultirespondentSubmissionDto, 'encryptedContent' | 'version'> & {
+      responses: FieldResponsesV3
+    })
+  | undefined => {
+  if (!submission || !secretKey) return
```
ghost: nit: throw error

frontend/src/features/workspace/WorkspaceService.ts
```diff
@@ -94,8 +95,8 @@ export const createEmailModeForm = async (
   )
 }
 
-export const createStorageModeForm = async (
-  body: CreateStorageFormBodyDto,
+export const createStorageModeOrMultirespondentForm = async (
```
ghost, line 98: nit: `createEncryptedForm`?

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
justynoh: the `Crypto` classes include version, to future-proof also for schema changes. Reason is that the `decrypt` function does shape validation after decryption. So for e.g., it would have been used if we only change the schema without changing the encryption scheme. Then the `Crypto` class can be reused, with the only change being that the shape is validated based on the provided version number

frontend/src/features/public-form/components/FormFields/FormFields.tsx
justynoh: correct! i presume this would be intended - we should investigate how prefills would work with MRF responses though. Will open a ticket for this, for now users should not use MRF with prefills for pilot

src/app/models/submission.server.model.ts
justynoh: Here we don't need it. The submission public key is only required when the future respondents 2+ enter their secret key, to validate that the key is correct. Since this cursor function is only used when the admin exports as CSV, so it's not required. In contrast, it is required for `findEncryptedSubmissionById` since that function retrieves the encrypted response for the respondent 2+.

frontend/src/pages/Landing/Payments/LandingPaymentsPage.tsx
justynoh: updated to `disableRequiredValidation`

src/app/modules/submission/submission.service.ts
justynoh: good point, thanks for catching that!

frontend/src/features/public-form/utils/createSubmission.ts
justynoh: yes we do! It's assumed that encryption boundary shift + virus scanning is 100%

frontend/src/features/workspace/WorkspaceService.ts
justynoh: i considered having this naming change where we have a mini hierarchy
```
                forms
              /       \
  not encrypted     encrypted
          |          /    \
        email    storage   MRF
```
But I think since a lot of naming conventions in the submissions directory mix up storage <-> encrypt, it would become pretty confusing if we do the renaming in some places but not others. I think we can refactor this another time with renamings, where even the whole `encrypt-submission` folder should become `storage-submission`, and then we can move to this naming scheme.

frontend/src/features/public-form/utils/createSubmission.ts
justynoh: this was throwing a type error for me, i don't know how it got through linting previously :'( i encountered issues without making this type change

frontend/src/components/SecretKeyVerificationInput/SecretKeyVerificationInput.tsx
justynoh: yeah technically it's a base64 regex, haha. didn't modify anything cuz it was just a refactor

in the future we may want to refactor again to combine `SecretKeyVerificationInput` and `SecretKeyActivationModal`

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
justynoh: hmmmm, won't fix this now. but will add to ticket for cleaning up individual response page.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
justynoh: ah, it's the equivalent of `strippedBodyResponses` in encrypt-submission's `encryptSubmission`. During the virus scanning flow the attachment field is augmented with a `filename` and `content` field. Here, once the virus scanning flow is complete, we remove it.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
justynoh: correct! `isQuarantinedAttachmentResponse` was lifted out of the split-mode virus scanning functions into the parent `scanAndRetrieveAttachments`.

frontend/src/features/public-form/mutations.ts
justynoh: haha. see how many cases we're handling because of all our un-cleaned-up rollouts......

frontend/src/features/public-form/mutations.ts
justynoh: me: "mom can we have proper types please??"
mom: "we have types at home"

types at home: `any`

frontend/src/templates/Field/Checkbox/CheckboxField.tsx
```diff
@@ -18,19 +18,22 @@ import FormErrorMessage from '~components/FormControl/FormErrorMessage'
...
-export const CHECKBOX_OTHERS_INPUT_VALUE =
-  '!!FORMSG_INTERNAL_CHECKBOX_OTHERS_VALUE!!'
+import {
+  CHECKBOX_OTHERS_INPUT_KEY,
+  CHECKBOX_OTHERS_INPUT_VALUE,
+} from './constants'
 
 export interface CheckboxFieldProps extends BaseFieldProps {
   schema: CheckboxFieldSchema
+  disableRequiredValidation?: boolean
```
ghost, line 28: `disableRequiredValidation` is currently used to bypass validation since MRF fields can be left empty for subsequent respondents to fill.

How do we intend to implement response validation in subsequent stages of MRF?

Do we intend to enforce it at the respondent level? e.g. field 1, 2 , 3 is required for respondent 1/submission 1.

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -51,6 +51,7 @@ import {
 
 export interface ChildrenCompoundFieldProps extends BaseFieldProps {
```
ghost, line 52: Might be better to create a `MultipleRespondentFieldProps` type in future to contain all MRF related fields. Interfaces can extend the MRF prop type instead

frontend/src/templates/Field/Email/EmailFieldInput.tsx
```diff
@@ -24,12 +25,13 @@ export interface EmailFieldInputProps {
...
 export const EmailFieldInput = ({
   schema,
+  disableRequiredValidation,
   handleInputChange,
   inputProps = {},
 }: EmailFieldInputProps): JSX.Element => {
   const validationRules = useMemo(
-    () => createEmailValidationRules(schema),
-    [schema],
+    () => createEmailValidationRules(schema, disableRequiredValidation),
```
ghost, line 33: Do we enforce the required validation upon the final submission?

frontend/src/templates/Field/Rating/RatingField.tsx
```diff
@@ -59,7 +61,7 @@ export const RatingField = ({
             colorScheme={`theme-${colorTheme}`}
             numberOfRatings={schema.ratingOptions.steps}
             variant={ratingVariant}
-            value={transform.toNumber(value)}
+            defaultValue={transform.toNumber(value)}
```
ghost, line 64: This has been wrong all along?

frontend/src/templates/Field/Table/ColumnCell.tsx
```diff
@@ -29,23 +29,29 @@ import { TableFieldInputs } from '../types'
...
 
 const ShortTextColumnCell = ({
   schema,
+  responseMode,
   inputName,
   colorTheme,
 }: FieldColumnCellProps<ShortTextColumnBase>) => {
-  const rules = useMemo(() => createBaseValidationRules(schema), [schema])
+  const rules = useMemo(
+    () => createBaseValidationRules(schema, responseMode),
```
ghost: We can't pass in `responseMode` here, `createBaseValidationRules` accepts a schema and a `disableRequiredValidation` boolean val

frontend/src/templates/Field/Table/ColumnCell.tsx
```diff
@@ -67,11 +73,15 @@ const ShortTextColumnCell = ({
...
 const DropdownColumnCell = ({
   schema,
+  responseMode,
   inputName,
   colorTheme,
 }: FieldColumnCellProps<DropdownColumnBase>) => {
   const { control } = useFormContext<TableFieldInputs>()
-  const rules = useMemo(() => createDropdownValidationRules(schema), [schema])
+  const rules = useMemo(
+    () => createDropdownValidationRules(schema, responseMode),
```
ghost: same as above
We can't pass in `responseMode` here, `createDropdownValidationRules` accepts a schema and a `disableRequiredValidation` boolean val

frontend/src/templates/Field/types.ts
```diff
@@ -82,6 +82,7 @@ export type FormFieldValue<F extends BasicField = BasicField> = F extends
   | BasicField.Nric
   | BasicField.Uen
   | BasicField.Date
+  | BasicField.CountryRegion
```
ghost, line 85: previous bug?

frontend/src/templates/Field/Email/EmailFieldInput.tsx
KenLSM: I dont think there's a concept of a final submission for MRF

shared/types/response-v3.ts
```diff
@@ -0,0 +1,117 @@
+import { BasicField, FormFieldDto, MyInfoChildAttributes } from './field'
+
+export type FieldResponsesV3 = Record<FormFieldDto['_id'], FieldResponseV3>
```
ghost, line 3: I'm missing something. Why do we need to create this instead of reusing `FieldResponse[]` directly?

shared/types/submission.ts
```diff
@@ -73,6 +74,28 @@ export type StorageModeSubmissionBase = z.infer<
   typeof StorageModeSubmissionBase
 >
 
+/**
+ * Multirespondent submission typings as stored in the database.
+ */
+
+export const MultirespondentSubmissionBase = SubmissionBase.extend({
```
ghost, line 81: Why not extend `StorageModeSubmissionBase` with the additional fields instead since we will be going for feature parity in subsequent phases anyway?

shared/types/submission.ts
```diff
@@ -101,17 +124,40 @@ export const SubmissionPaymentDto = z.object({
...
+export type StorageModeSubmissionDto = SubmissionDtoBase & {
+  submissionType: SubmissionType.Encrypt
   content: string
   verified?: string
   attachmentMetadata: Record<string, string>
   payment?: SubmissionPaymentDto
   version: number
 }
 
+export type MultirespondentSubmissionDto = SubmissionDtoBase & {
```
ghost, line 142: Similar to above - can we extend `StorageModeSubmissionDto` instead if we are aiming for feature parity in the final version?

shared/types/submission.ts
```diff
@@ -126,23 +172,49 @@ export type StorageModeSubmissionStreamDto = z.infer<
...
+  MultirespondentSubmissionBase.pick({
+    submissionType: true,
+    form_fields: true,
+    form_logics: true,
+    encryptedSubmissionSecretKey: true,
+    encryptedContent: true,
+    verifiedContent: true,
+    version: true,
+  }).extend({
+    attachmentMetadata: z.record(z.string()),
```
ghost, line 185: why not follow `attachmentMetadata: z.map(z.string(), z.string()).optional(),` in `MultirespondentSubmissionBase`?

src/app/models/__tests__/multirespondent-submission.server.model.spec.ts
```diff
@@ -0,0 +1,450 @@
...
+          await MultirespondentSubmission.findEncryptedSubmissionById(
+            formId,
+            invalidSubmissionId,
+          )
+
+        // Assert
+        expect(actual).toBeNull()
+      })
+
+      it('should return null when type of submission with given id is not SubmissionType.Encrypt', async () => {
```
ghost: is not `SubmissionType.Multirespondent`

frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
KenLSM: What will the MRF prop type contain?

frontend/src/templates/Field/Table/ColumnCell.tsx
KenLSM: Good catch, this is a change that's missed from [this comment](https://github.com/opengovsg/FormSG/pull/6960#discussion_r1426915229)

KenLSM: cafd0762

shared/types/submission.ts
KenLSM: I don't think we can have feature parity (or that I know of) for things such as myinfo, payments?

src/app/models/__tests__/multirespondent-submission.server.model.spec.ts
KenLSM: e324666d

frontend/src/templates/Field/Table/ColumnCell.tsx
KenLSM: https://github.com/opengovsg/FormSG/commit/cafd0762a90aa091ba634f1d705b70f0cafbf481

KenLSM:
Keeping formsg-sdk as alpha as we've not reached general availability.

KenLSM (approved):

src/app/modules/submission/__tests__/submission.controller.spec.ts
```diff
@@ -23,410 +32,36 @@ import {
...
+import * as SubmissionService from '../submission.service'
+import * as SubmissionsUtils from '../submission.utils'
+
+jest.mock('src/app/modules/submission/submission.service')
+jest.mock(
+  'src/app/modules/submission/encrypt-submission/encrypt-submission.service',
+)
+jest.mock('src/app/modules/user/user.service')
+jest.mock('src/app/modules/auth/auth.service')
+jest.mock('src/app/modules/feature-flags/feature-flags.service')
```
ghost, line 55: nit: feature flags service is mocked twice

src/app/modules/submission/__tests__/submission.controller.spec.ts
```diff
@@ -777,144 +401,437 @@ describe('encrypt-submission.controller', () => {
...
+    it('should return 200 with encrypted response', async () => {
+      // Arrange
+      const mockSubData: SubmissionData = {
+        _id: 'some id',
+        encryptedContent: 'some encrypted content',
+        verifiedContent: 'some verified content',
+        created: new Date('2020-10-10'),
+        paymentId: 'payment id',
+        submissionType: SubmissionType.Encrypt,
+      } as SubmissionData
```
ghost, line 508: Should be `as StorageModeSubmissionData` otherwise line 537 will throw a type error since `verifiedContent` does not exist on `MultirespondentSubmissionData`

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -0,0 +1,445 @@
...
+  SubmitMultirespondentFormHandlerType,
+  UpdateMultirespondentSubmissionHandlerRequest,
+  UpdateMultirespondentSubmissionHandlerType,
+} from './multirespondent-submission.types'
+import { createMultirespondentSubmissionDto } from './multirespondent-submission.utils'
+
+const logger = createLoggerWithLabel(module)
+const MultirespondentSubmission = getMultirespondentSubmissionModel(mongoose)
+
+const submitMultirespondentForm = async (
```
ghost, line 53: Missing type definition in function signature

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -0,0 +1,445 @@
...
+  } catch (err) {
+    logger.error({
+      message: 'Multirespondent submission save error',
+      meta: {
+        action: 'onMultirespondentSubmissionFailure',
+        ...createReqMeta(req),
+      },
+      error: err,
+    })
+    return res.status(StatusCodes.BAD_REQUEST).json({
```
ghost, line 187: Error code for DB save errors should be 500?

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -0,0 +1,445 @@
...
+      submissionId,
+      formId,
+      responseMetadata,
+    },
+  })
+
+  // TODO 6395 make responseMetadata mandatory
+  if (responseMetadata) {
+    reportSubmissionResponseTime(responseMetadata, {
+      mode: 'multirespodent',
```
ghost, line 208: nit: multi-respondent

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -0,0 +1,445 @@
...
+  //     })
+  //   } else {
+  //     attachmentMetadata = attachmentUploadResult.value
+  //   }
+  // }
+
+  const submission = await MultirespondentSubmission.findById(submissionId)
+
+  if (!submission) {
+    return res.status(StatusCodes.NOT_FOUND).json({
```
ghost, line 299: might want to log this

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -0,0 +1,445 @@
...
+  } catch (err) {
+    logger.error({
+      message: 'Encrypt submission save error',
+      meta: {
+        action: 'onEncryptSubmissionFailure',
+        ...createReqMeta(req),
+      },
+      error: err,
+    })
+    return res.status(StatusCodes.BAD_REQUEST).json({
```
ghost, line 322: Error code for DB save errors should be 500?

src/app/routes/api/v3/forms/public-forms.submissions.routes.ts
```diff
@@ -67,8 +69,63 @@ PublicFormsSubmissionsRouter.route(
...
+ * @param response.body.required - contains field ids and sizes of attachments
+ * @returns 200 - presigned post data generated
+ * @returns 400 - ids are invalid or attachment size exceeds limit
+ * @returns 500 - failed to generate presigned post data
+ */
+PublicFormsSubmissionsRouter.route(
+  '/:formId([a-fA-F0-9]{24})/submissions/get-s3-presigned-post-data',
+).post(
+  limitRate({ max: rateLimitConfig.submissions }),
+  SubmissionController.handleGetS3PresignedPostData,
```
ghost, line 121: Remember that we discussed that this change is backwards incompatible. What was our rollout strategy again?

PR #6992, opened 2024-01-03
fix(mrf): individual response page

justynoh:
## Problem
The individual response page for MRF hides some questions if they are unfilled and shows others. This PR fixes this. 

Closes FRM-1593

## Solution

The reason for this is due to schema differences between fields. Some fields have schema `answer: { value: string }` while some only have `answer: string` (this is required for special fields allowing multiple value types, such as radio, checkbox and table). Those with schema `answer: string` were filtered on save so that those with `!answer` were filtered out, which matches `undefined` and `''` values. Therefore, on decryption, we 'backfill' the value as `''` instead of ignoring the question.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
@@ -129,8 +129,7 @@ export const processDecryptedContentV3 = async (
     .map((ff) => {
       const response = responses[ff._id]
       if (!response) {
-        console.log('Unexpected empty response for field id', ff._id)
-        return null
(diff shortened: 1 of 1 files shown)


KenLSM (commented):
Thanks @justynoh for the quick turnaround!

I would like to clarify that the subsequent submissions by the next respondent will not be affected by this "prefill" as this is admin-side?

i.e., 
- 1st respondent fills in fields A, and C, and submits.
- 2nd respondent fills in field D and submits, field B will still be empty instead of `''` in the DB

justynoh:
yessu correct! Because this is purely on the rendering side on the individual response page, so nothing on the respondent side is affected

KenLSM (approved):

PR #6996, opened 2024-01-04
fix: add polyfill for Object.FromEntries

wanlingt:
## Problem

[`Object.FromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) is incompatible with older browsers. This results in an error being thrown when the public form page is loaded.

Closes FRM-1612

## Solution

Add a polyfill from the package [polyfill-object.fromentries](https://www.npmjs.com/package/polyfill-object.fromentries)

**Breaking Changes** 

## Tests

**Regression**
- [ ] Go to a public form page. You should be able to load the page and submit the form successfully.

MRF
- [ ] As a form admin, go to an MRF form that has a workflow
- [ ] Submit the form once.
- [ ] Open the form as the second respondent, using the link sent in the email.
- [ ] Submit the form. The response should be submitted successfully.

## Deploy Notes

**New dependencies**:

- `polyfill-object.fromentries` : Provides polyfill for `Object.fromEntries()`

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -50,6 +50,7 @@
         "lodash": "^4.17.21",
         "lottie-web": "^5.9.4",
         "p-queue": "^7.2.0",
+        "polyfill-object.fromentries": "^1.0.1",
         "react": "^17.0.2",
         "react-beautiful-dnd": "^13.1.0",
         "react-csv": "^2.2.2",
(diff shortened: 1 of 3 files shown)


justynoh (approved):
lgtm!

PR #6998, opened 2024-01-04
fix: move polyfills to separate file

wanlingt:
## Problem

Polyfills should be treated like global updates

## Solution

Referencing https://github.com/opengovsg/signify/pull/482, group the polyfills together in a separate file.


justynoh (approved):
lgtm

PR #7029, opened 2024-01-21
fix: add patch to feed default ValidationOptions on table shorttext

KenLSM:
## Problem

Due to this issue on [Mongoose 6.x](https://www.github.com/Automattic/mongoose/issues/12135), discriminated fields doesn't respect inherited defaults.

This currently affects our Table field where a `dropdown` column is updated to `shortText`. This would then be updated into the DB without a `ValidationOptions` field. Consequently, code that references `column.ValidationOptions.*` would throw an error as the object is `undefined`.

Note: Inserting a new `shortText` column uses the [helper function](https://github.com/opengovsg/FormSG/blob/fix/mongoose-manual-base-schema-inheritance/frontend/src/features/admin-form/create/builder-and-design/utils/columnCreation.ts#L15) that contains the `ValidationOptions` field, thus not affected by this change.

## Solution

Add a patch that checks for shortText and provide a defaulted `ValidationOptions` object to it.

## Tests
Bug fix tests, reproduction steps
- [ ] Add new table field
- [ ] Add new textField column on table
- [ ] Save field
- [ ] Ensure as a respondent submit form succeeds
- [ ] Change textField to dropdown
- [ ] Save field
- [ ] Change dropdown field to textField
- [ ] Save field
- [ ] Ensure as a respondent submit form succeeds

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.utils.ts
+++ b/src/app/modules/form/admin-form/admin-form.utils.ts
@@ -9,6 +9,7 @@ import {
   FieldUpdateDto,
   FormResponseMode,
   FormStatus,
+  TextValidationOptions,
 } from '../../../../../shared/types'
 import {
   reorder,
@@ -605,7 +606,7 @@ export const insertTableShortTextColumnDefaultValidationOptions = (
   newField: FieldUpdateDto,
 ) => {
   if (newField.fieldType === BasicField.Table) {
-    const defaultValidationOptions = {
+    const defaultValidationOptions: TextValidationOptions = {
       customVal: null,
       selectedValidation: null,
     }
--- a/src/app/modules/form/admin-form/admin-form.utils.ts
+++ b/src/app/modules/form/admin-form/admin-form.utils.ts
@@ -4,7 +4,9 @@ import { err, ok, Result } from 'neverthrow'
 import { v4 as uuidv4 } from 'uuid'
 
 import {
+  BasicField,
   DuplicateFormBodyDto,
+  FieldUpdateDto,
   FormResponseMode,
   FormStatus,
 } from '../../../../../shared/types'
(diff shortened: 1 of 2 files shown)


src/app/modules/form/admin-form/admin-form.utils.ts
```diff
@@ -586,3 +588,36 @@ export const mapGoGovErrors = (error: AxiosError): GoGovError => {
...
+ * @returns
+ */
+
+export const insertTableShortTextColumnDefaultValidationOptions = (
+  newField: FieldUpdateDto,
+) => {
+  if (newField.fieldType === BasicField.Table) {
+    const defaultValidationOptions = {
+      customVal: null,
+      selectedValidation: null,
```
justynoh, line 611: I'm wondering if we should add the prefill properties too, just for db consistency's sake

KenLSM: Hm those are marked as [optional](https://github.com/opengovsg/FormSG/blob/develop/shared/types/field/shortTextField.ts) on types. But if we look at inheritance it should be populated. 

Both are correct. I in favour of keeping it as simple as possible.

src/app/modules/form/admin-form/admin-form.utils.ts
```diff
@@ -586,3 +588,36 @@ export const mapGoGovErrors = (error: AxiosError): GoGovError => {
...
+ * Fixed in Mongoose 7.x
+ * @param newField
+ * @returns
+ */
+
+export const insertTableShortTextColumnDefaultValidationOptions = (
+  newField: FieldUpdateDto,
+) => {
+  if (newField.fieldType === BasicField.Table) {
+    const defaultValidationOptions = {
```
wanlingt:
```suggestion
    const defaultValidationOptions: TextValidationOptions = {
```
(import TextValidationOptions from '../../../../../shared/types')

wanlingt (approved):
lgtm! nit: add TextValidationOptions for type safety

src/app/modules/form/admin-form/admin-form.utils.ts
KenLSM: Makes sense. Added in b8cb4665

PR #7059, opened 2024-02-02
fix: recaptcha cn

KenLSM:
## Problem

Users from CN are unable to load FormSG as they were looking at loading `www.gstatic.cn` which was blocked by our CSP rules.
Closes FRMSD-22
## Solution

Add [`www.gstatic.cn`](https://www.whois.com/whois/gstatic.com) into our CSP.


wanlingt (approved):
thanks ken!

PR #7069, opened 2024-02-06
chore(dev): add virus scanner npm install to root install command

justynoh:
## Problem
`npm install` in the root directory only performs package installation for the frontend and backend, even though `npm run dev` attempts to run the frontend, backend and virus scanner services.

## Solution

Add virus scanner package installation into the root `postinstall` step.

**Breaking Changes** 

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     ]
   },
   "scripts": {
-    "postinstall": "npm run postinstall:frontend && npm run postinstall:shared",
+    "postinstall": "npm run postinstall:frontend && npm run postinstall:shared && npm run postinstall:virus-scanner",
     "test": "npm run test:backend && npm run test:frontend",
     "test:backend": "env-cmd -f __tests__/setup/.test-env jest",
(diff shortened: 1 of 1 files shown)


wanlingt (approved):

KenLSM:
Im thinking that this shouldn't be added to root as virus-scanner is not required to build the image for formsg

justynoh:
hmm, but by that logic, the frontend install also shouldn't be included in root install right haha 😅  or am i missing something

KenLSM:
ccmiw 
```
github: build FormFE && FormBE
eb: upload FormFE to s3 (after env configs are in)
eb: run FormBE
```

Although I think we can install them separately after we have achieved the split deployment of FE & BE

justynoh:
oh, this was more from a DX perspective haha. if you clone the repo fresh, `npm i` `npm run build:frontend` and `npm run dev`, the last command fails because it attempts to start virus scanner without having installed it's packages first. not so much from deployment perspective

that being said, the `npm i` does influence deployment speed. But i think the number of packages in the virus scanner isn't that much, the impact to deployment should be negligible. in comparison, the frontend install takes significantly longer haha

PR #7072, opened 2024-02-08
feat(mrf): workflow builder

justynoh:
## Problem
We want to build out a workflow builder for MRF with dynamic number of respondents.

Closes FRM-1624

## Solution
This implementation reuses many of the components from the logic module but with several significant changes. In particular, a lot of the inputs are removed and replaced with workflow-specific inputs (e.g. step number, respondent emails). Also, the current implementation uses a single "godlike" put endpoint to put the entire workflow array into the backend. This should be granularized so that we have specific creation, update and delete endpoints on a per-workflow-step basis.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:

https://github.com/opengovsg/FormSG/assets/25571626/c3d6947c-5264-4ef8-a2c0-64f231ad149e


frontend/src/features/admin-form/create/workflow/mutations.ts
```diff
@@ -0,0 +1,131 @@
...
+          // existing data.
+          if (!prev) throw new Error('Query should have been set')
+          if (
+            prev.responseMode !== FormResponseMode.Multirespondent ||
+            newSettings.responseMode !== FormResponseMode.Multirespondent
+          ) {
+            throw new Error('Invalid response mode')
+          }
+          prev.workflow = newSettings.workflow
+          return prev
```
KenLSM: Best to not mutate existing objects to avoid unwanted side-effects.

```suggestion

          return { ...prev, workflow: newSettings.workflow }
```

frontend/src/features/admin-form/create/workflow/mutations.ts
```diff
@@ -0,0 +1,131 @@
...
```
KenLSM: Same as above

```suggestion

          return { ...prev, workflow: newSettings.workflow }
```

frontend/src/features/admin-form/create/workflow/mutations.ts
```diff
@@ -0,0 +1,131 @@
...
```
KenLSM: Same as above

```suggestion

          return { ...prev, workflow: newSettings.workflow }
```

frontend/src/features/admin-form/create/workflow/FormWorkflowService.ts
```diff
@@ -0,0 +1,49 @@
...
+import { ApiService } from '~services/ApiService'
+
+import { ADMIN_FORM_ENDPOINT } from '~features/admin-form/common/AdminViewFormService'
+
+export const updateFormWorkflow = async (
+  formId: string,
+  newWorkflowSettings: MultirespondentFormSettings['workflow'],
+) => {
+  return ApiService.patch<FormSettings>(
+    `${ADMIN_FORM_ENDPOINT}/${formId}/settings`,
```
KenLSM, line 18: Curious, why is workflow using `:formId/settings` instead of having it's own PATCH, DEL, PUT `:formId/workflow`?

Was it to write less boilerplate code?

KenLSM: -- saw that you have already noted this on the PR, this will be resolved after we ship? Totally fine by it as we have a good amount of buffer post-release to make refinements like this.

KenLSM:
Added d2b3327 as we should be swapping these to their own PATCH / PUT / DELETE. 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments:~:text=you%20expect%20a%20fix%20to%20be%20coming%20in%20fairly%20quickly%20and%20you%20just%20need%20a%20quick%20workaround

KenLSM (approved):
LGTM! Some minor nits / questions that isn't blocking

frontend/src/features/admin-form/create/workflow/FormWorkflowService.ts
justynoh: Yup, in the spirit of making PRs as small as possible to avoid scope creep,, I decided not to do that modification here. This code was already written and I didn't want to touch it too much. Nonetheless, a linear ticket to resolve this has already been opened as [FRM-1632](https://linear.app/ogp/issue/FRM-1632/proper-object-level-rest-api-for-workflow-steps)!

frontend/src/features/admin-form/create/workflow/mutations.ts
justynoh: Thanks! all three fixed in [d5d8079](https://github.com/opengovsg/FormSG/pull/7072/commits/d5d8079eeb89c1b02e4a949733dab69df943a8c0)

frontend/src/features/admin-form/create/workflow/FormWorkflowService.ts
KenLSM: Great division of scope btw! Features are cleanly cut, and isolated.

PR #7077, opened 2024-02-12
chore: add payments by product backend validation

tshuli:
## Problem
- There is currently no backend validation for payments by product.
## Solution

**Bug Fixes**:
- Adds backend validation for payments by product

## Tests
- [ ] Create payments by product. Can submit normally.
  - [ ] Load the form. Change the payment amount in form definition. Submission should fail.
  - [ ] Load the form. Delete the product in form definition and add another product. Submission should fail.
  - [ ] Load the form. Change to multi quantity of 2 to 5. Submission with qty 1 should fail.
  - [ ] Load the form. Submission with Qty 6 should fail
  - [ ] Load the form. Now disable multi quantity. Submission with qty 4 should fail.
- [ ] Create payments by amount form. Can submit normally.

### DIFF
--- a/src/app/modules/payments/__tests__/payments.service.spec.ts
+++ b/src/app/modules/payments/__tests__/payments.service.spec.ts
@@ -2,7 +2,7 @@ import dbHandler from '__tests__/unit/backend/helpers/jest-db'
 import { ObjectId } from 'bson'
 import moment from 'moment-timezone'
 import mongoose, { Query } from 'mongoose'
-import { PaymentStatus, Product, ProductItem } from 'shared/types'
+import { PaymentStatus, Product, ProductId, ProductItem } from 'shared/types'
 
 import getAgencyModel from 'src/app/models/agency.server.model'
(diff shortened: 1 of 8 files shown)


src/app/modules/payments/__tests__/payments.service.spec.ts
```diff
@@ -155,7 +155,147 @@ describe('payments.service', () => {
...
+    })
+
+    it('should return with error if the max_qty has changed', () => {
+      // Arrange
+      const mockInvalidProductSubmission: ProductItem[] = [
+        {
+          data: {
+            ...mockValidProduct,
+            max_qty: 5,
+            _id: new ObjectId(),
```
KenLSM: Need to remove `_id` since we only want to test rejection from `max_qty` changed instead of `_id`

tshuli: 👍  thanks for catching

KenLSM (approved):

PR #7078, opened 2024-02-13
feat(mrf): dynamic routing

justynoh:
## Problem
On top of the workflow builder, we want to build dynamic routing as a feature.

Closes FRM-1621

## Solution

This PR adds instrumentation for dynamic routing in the frontend and backend. Importantly, this builds on top of #7072.

On the frontend, we add a radio toggle for admins to select whether they want to do static or dynamic routing. If static, the original text box is shown. If dynamic, we display a dropdown of all the email fields in the form for their selection.

On the backend, Joi validation for the settings PATCH endpoint is updated. Additionally, the schema and model for multirespondent form workflows is updated with a discriminator on the `workflow_type` to accommodate the two types of data now needed to be stored mutually exclusively - `emails` for static, and `field` for dynamic. Finally, on the MRF submission endpoint, the email retrieval computation is updated to include the case where the routing is dynamic.

**Breaking Changes** 

## Screenshots

https://github.com/opengovsg/FormSG/assets/25571626/5e47fc3c-a9b2-48af-8458-c23632ade345


src/app/modules/form/admin-form/admin-form.middlewares.ts
```diff
@@ -40,13 +40,23 @@ export const updateSettingsValidator = celebrate({
...
-            Joi.array().items(Joi.string().email().allow('')),
-            Joi.string().email({ multiple: true }).allow(''),
-          ),
+          emails: Joi.when('workflow_type', {
+            is: WorkflowType.Static,
+            then: Joi.alternatives()
+              .try(
+                // We might want to tighten this validation
+                Joi.array().items(Joi.string().email().allow('')),
+                Joi.string().email({ multiple: true }).allow(''),
```
KenLSM: Tightening as in making this to not allow empty strings?

KenLSM: Seems like from the schema, we should only accept `string[]`

KenLSM (commented):
LGTM except for one question on Joi validation on email field type

src/app/modules/form/admin-form/admin-form.middlewares.ts
justynoh: Yeah, by right only should accept `string[]`, but there is a `transformEmails` step in the mongoose schema which does the following:

`/src/app/modules/form/form.utils.ts`

```
// Converts 'test@hotmail.com, test@gmail.com' to ['test@hotmail.com', 'test@gmail.com']
export const transformEmailString = (v: string): string[] => {
  return v
    .split(',')
    .map((item) => item.trim().toLowerCase())
    .filter((email) => email.includes('@')) // remove ""
}

// Function that coerces the string of comma-separated emails sent by the client
// into an array of emails
export const transformEmails = (v: string | string[]): string[] => {
  // Cases
  // ['test@hotmail.com'] => ['test@hotmail.com'] ~ unchanged
  // ['test@hotmail.com', 'test@gmail.com'] => ['test@hotmail.com', 'test@gmail.com'] ~ unchanged
  // ['test@hotmail.com, test@gmail.com'] => ['test@hotmail.com', 'test@gmail.com']
  // ['test@hotmail.com, test@gmail.com', 'test@yahoo.com'] => ['test@hotmail.com', 'test@gmail.com', 'test@yahoo.com']
  // 'test@hotmail.com, test@gmail.com' => ['test@hotmail.com', 'test@gmail.com']
  if (Array.isArray(v)) {
    return v.flatMap(transformEmailString)
  } else {
    return transformEmailString(v)
  }
}
``` 

So it seems actually that it CAN take string-separated values, but we should probably just disallow that in general. Will fix this!

justynoh (commented):
[969cc00](https://github.com/opengovsg/FormSG/pull/7078/commits/969cc000fa29aff815bb047cf2bd78f931ae1c56)

oh this is quite cool! I didn't know this was a thing

src/app/modules/form/admin-form/admin-form.middlewares.ts
justynoh: done in [de93eb9](https://github.com/opengovsg/FormSG/pull/7078/commits/de93eb964d8d9259cc4f59717c55815e2936957d)

KenLSM (approved):

PR #7087, opened 2024-02-15
feat(mrf): retain workflow through a submission lifetime

justynoh:
## Problem
When a submission is made, we want to save a snapshot of the workflow at that point and use it throughout the lifetime of the submission. This involves saving the workflow in the submission object, as well as using the submission workflow for routing.

Closes FRM-1637

## Solution

Add a `workflow` field to multirespondent submissions, and populate it on creation with the form workflow. From then on, the submission object's workflow is used for routing, and passed to the frontend.

**Breaking Changes** 

## Tests

- [ ] Create an MRF form without a workflow. The usual response link passing should still work, with all fields enabled for all respondents.
- [ ] Create an MRF form.
  - [ ] Create a workflow where step 2 is a static value `a@gov.sg` and step 3 is another static value `b.gov.sg`. 
  - [ ] Submit the form once as Respondent 1. `a@gov.sg` should receive an email reminder to submit the form.
  - [ ] Now, change step 3 to be a different static value `c@gov.sg`.
  - [ ] Click the button in the email that `a@gov.sg` received to fill the form. Submit the form.
  - [ ] Check that `b@gov.sg` got the next respondent email, and not `c@gov.sg`.

## Deploy Notes

**New scripts**:

- `mrf-retain-workflow-in-submissions` : This DB migration script creates a new workflow key in all existing mrf submissions, populated with the form's existing workflow. From this point on, workflows should be captured when the submission is initially created (i.e. the workflow is started),

### DIFF
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -148,10 +148,13 @@ export const PublicFormProvider = ({
     /* enabled= */ !submissionData,
   )
 
-  // Replace form fields and logic with the previous version for MRF consistency.
+  // Replace form fields, logic, and workflow with the previous version for MRF consistency.
   if (data && encryptedPreviousSubmission) {
     data.form.form_fields = encryptedPreviousSubmission.form_fields
     data.form.form_logics = encryptedPreviousSubmission.form_logics
+    if (data.form.responseMode === FormResponseMode.Multirespondent) {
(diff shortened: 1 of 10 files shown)


justynoh:
Hold off on this. I need to do a bit of git surgery

KenLSM (approved):
LGTM!

justynoh:
no code changes, just rebase onto develop

PR #7089, opened 2024-02-18
chore(deps): fix synk vulnerabilities

KenLSM:
## Problem

1. `testcafe` is no longer referenced in the repo.
2. `bcrypt`, `@stoplight/prism` are referencing outdated `ansi-regex` which is vulnerable to certain exploits.

## Solution

1. Remove `testcafe`
2. Upgrade references


justynoh (approved):
btw, is `@stoplight/prism-cli` required as a dep or can it be downgraded to a dev-dep? iirc it's only used for our local twilio provider container right. a quick scan of the code doesn't show any import or require from `@stoplight/prism-cli`.

KenLSM:
> btw, is `@stoplight/prism-cli` required as a dep or can it be downgraded to a dev-dep? iirc it's only used for our local twilio provider container right. a quick scan of the code doesn't show any import or require from `@stoplight/prism-cli`.

Great catch, it's only required as devDeps! This change will help to reduce the number of packages installed!

PR #7095, opened 2024-02-20
fix(mrf): use undefined as default for response display

justynoh:
## Problem

User reported an issue with MRF pilot where they got an error message `a.map is not a function`. 

image

After identification, this was identified to only occur as a result of a frontend bug. The form contains a table field hidden by logic, and all hidden fields are removed from the field payload upon submission. This means that the table field did not exist in the stored data payload.

Upon retrieval, the payload was decrypted and displayed to the user on the individual response page. Therefore, when retrieved, the data object keyed by the table's field id would result in `undefined`. To deal with these undefined keys, the code used `''` as a default, in `inputTransformation`. This works in most cases, but due to the expected type of table field answers being arrays, a type with `.map` was expected, but `''` did not have this method. Therefore, the transformation process failed.

Closes FRM-1629

## Solution

We change the transformation to use `undefined` as a default. This involves changing the type signature for `transformInputsToOutputs` to accept an optional `input`, but this is okay because most sub-functions allow for an optional input. The exceptions are checkbox and radio, but even then, the functions themselves are actually already written to accommodate the case where `input` is undefined, even though the type signature doesn't specify that it's allowed. Therefore we also modify these function to allow `input` to be undefined.

**Breaking Changes** 

## Tests

- [ ] Ensure that visible fields can be displayed on the MRF individual response page.
- [ ] Ensure that hidden fields can be displayed on the MRF individual response page.

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
@@ -129,7 +129,7 @@ export const processDecryptedContentV3 = async (
     .map((ff) => {
       const response = responses[ff._id]
       if (!response) {
-        return transformInputsToOutputs(ff, '')
+        return transformInputsToOutputs(ff)
       }
       if (response.fieldType === BasicField.Attachment) {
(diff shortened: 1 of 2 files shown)


wanlingt (approved):
lgtm!

PR #7097, opened 2024-02-20
feat(config): allow R2 bucket URLs

LoneRifle:
## Problem
FormSG assumes in production environments that S3 buckets are hosted by AWS and use URLs that take after a certain pattern. This is not necessarily the case for eg, Cloudflare R2, which use a different pattern

## Solution
- Allow config to accept URLs to R2 buckets, in addition to S3 buckets
- Force path-style S3 URLs for R2 buckets, not just for dev mode

### DIFF
--- a/src/app/config/schema.ts
+++ b/src/app/config/schema.ts
@@ -58,9 +58,15 @@ const validateS3BucketUrl = (
     }
   }
   // Region should be specified correctly in production
-  const isRegionCorrect = new RegExp(`^https://s3.${region}.amazonaws.com`, 'i')
-  if (!isDev && !isRegionCorrect.test(val)) {
-    throw new Error(`region should be ${region}`)
+  const isS3RegionCorrect = new RegExp(
+    `^https://s3.${region}.amazonaws.com`,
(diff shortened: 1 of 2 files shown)


src/app/config/schema.ts
```diff
@@ -59,8 +59,11 @@ const validateS3BucketUrl = (
   }
   // Region should be specified correctly in production
   const isRegionCorrect = new RegExp(`^https://s3.${region}.amazonaws.com`, 'i')
-  if (!isDev && !isRegionCorrect.test(val)) {
-    throw new Error(`region should be ${region}`)
+  const isR2 = new RegExp(`^https://\\w+.r2.cloudflarestorage.com`, 'i')
+  if (!isDev && !isRegionCorrect.test(val) && !isR2.test(val)) {
```
KenLSM: /nit May want to rename `isRegionCorrect` to `isS3RegionCorrect` since we are allowing both S3 and R2, while the region check refers to S3's region.

KenLSM (approved):

PR #7108, opened 2024-02-22
chore: remove Anguilla from country listing

KenLSM:
## Problem

The country/region code list should be the same for both MyInfo and FormSG if the nature of the data collected is the same (i.e. if there is any issue in the list, this is not OGP's problem to manage but MyInfo team's).

Closes FRM-1665

## Solution

Remove `Anguilla` from country list

Note: `Anguilla` can still be selected as a prefix on Mobile Number Fields that accepts international numbers. The value is ultimately stored as a string, thus not removed.


kathleenkhy (approved):
lgtm!

PR #7115, opened 2024-03-01
feat: assisted form creation with Azure OpenAI

kathleenkhy:
## Problem

To migrate the AI Form Builder from Open AI to Azure OpenAI as moving to Azure OpenAI could help us to prevent logging of information for security reasons.

Closes FRM-1659, FRM-1661, FRM-1660, FRM-1682, FRM-1683, FRM-1684

**Breaking Changes** 

**Features**:

- Adding frontend and backend validation for character limit for prompts
- Adding frontend and backend validation for character limit for PDFs (to review again if this is the right approach for users)
- Updating loading image gif
- Updating upload size limit to 5mb

**Bug Fixes**:

- TODO: to fix bug that when you upload a PDF/ add a prompt and click the back button, you're still able to proceed and click the "Create Form" button
- There should be a behaviour (similar to form builder) that when you click back, there should be a prompt that asks if you're sure to discard what you have uploaded/ input

image

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


src/app/modules/form/admin-form/__tests__/admin-form.assistance.controller.spec.ts
```diff
@@ -80,7 +91,7 @@ describe('admin-form.assistance.controller', () => {
       })
 
       // Mock OpenAI API throwing an error
-      MockedOpenAIClient.prototype.chat.completions.create = jest
+      MockedOpenAIClient.prototype.getChatCompletions = jest
         .fn()
         .mockRejectedValue(new Error('Some random error message'))
```
tshuli, line 103: since this is a controller test, it suffices to mock the behavior of the service, e.g.

```
      const generateQuestionsSpy = jest
        .spyOn(AdminFormAssistanceService, 'generateQuestions')
        .mockReturnValueOnce(errAsync(new AssistanceConnectionError()) as any)

```

kathleenkhy: thank you!!

frontend/src/features/admin-form/assistance/mutations.ts
```diff
@@ -0,0 +1,148 @@
...
+  if (!formId) {
+    throw new Error('Form ID is required')
+  }
+
+  const toast = useToast({ status: 'success', isClosable: true })
+
+  const queryClient = useQueryClient()
+
+  const onCloseContext = useContext(MagicFormBuilderModalOnCloseContext)
+  const { onClose } = onCloseContext || {}
```
justynoh: Having this here feels a bit weird. This function should only contain logic specific to the mutation (i.e. the API call), and not frontend behaviors after the API call has resolved/errored. 

Should the `onClose` be factored out into the parent JSX element instead?

frontend/src/features/admin-form/assistance/mutations.ts
```diff
@@ -0,0 +1,148 @@
...
+              })
+            },
+          })
+        }),
+    {
+      onError: (error: Error) => {
+        if (error instanceof HttpError) {
+          let errorMessage
+          switch (error.code) {
+            case 429:
```
justynoh: nit: We should try to use `HTTP_STATUS_CODES` library instead of using magic numbers like this.

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/BuilderAndDesignPlaceholder/EmptyFormPlaceholder.tsx
```diff
@@ -4,70 +4,171 @@ import {
...
+          onClick={handleOpenMagicFormBuilderModal}
+        >
+          Try out Magic Form Builder
+        </Button>
+      ) : (
+        <></>
+      )
+    }, [handleOpenMagicFormBuilderModal, showMagicFormButton])
+
+    const showOrLine = useMemo(() => {
```
justynoh: the naming here feels a bit strange. usually `showX` would expect a boolean to indicate whether X should be shown or not. Maybe we can leave the boolean value `showMagicFormButton`, and then call the other two `magicFormButton` and `orLine` respectively?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FormBuilder.tsx
```diff
@@ -40,6 +48,12 @@ interface FormBuilderProps extends FlexProps {
   placeholderProps: DndPlaceholderProps
 }
 
+type OnCloseContext = {
+  onClose: () => void
+}
+export const MagicFormBuilderModalOnCloseContext =
+  createContext<OnCloseContext>({ onClose: () => undefined })
+
```
justynoh: I don't think this context should be needed at all - the child `MagicFormBuilderModal` already has `onClose` passed to it. (ref. my previous comment about calling `onClose` in the mutation hook)

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+    control,
+    setValue,
+    formState: { errors },
+  } = formMethods
+
+  const MAX_FILE_SIZE = 5 * MB
+  const VALID_EXTENSIONS = '.pdf'
+  pdfjs.GlobalWorkerOptions.workerSrc = pdfjsWorker
+
+  const [pdfFile, setPdfFile] = useState<File | undefined>(undefined)
```
justynoh: nit: can simply be `useState<File>()`

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const [pdfFileText, setPdfFileText] = useState<string>('')
+
+  function fileToArrayBuffer(file: File) {
```
justynoh: for consistency, this can be `const fileToArrayBuffer = (file: File) => {`

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+    return new Promise((resolve, reject) => {
+      const reader = new FileReader()
+      reader.onload = (event) => resolve(reader.result)
+      reader.onerror = (error) => reject(error)
+      reader.readAsArrayBuffer(file)
+    })
+  }
+
+  const handleFileUpload = async (file: any) => {
```
justynoh: We shouldn't be using `any` - is the type here `File`?

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+        setValue('pdfFileText', text)
+      } catch (e) {
+        console.error(e)
+      }
+    } else {
+      setPdfFileText('')
+      setValue('pdfFileText', '')
+    }
+  }
+  async function pdfToText(data: any) {
```
justynoh: same here. this can be `const pdfToText = async (data: T) => {` where T shouldn't be `any` - what's the correct type here?

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+    const loadingTask = pdfjs.getDocument(data)
+    const pdf = await loadingTask.promise
+
+    let combinedText = ''
+
+    for (let i = 1; i <= pdf.numPages; i++) {
+      const page = await pdf.getPage(i)
+      const content = await page.getTextContent()
+      const text = content.items.map((item: any) => item.str).join(' ')
```
justynoh: same comment here about `any` - what's the correct type?

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+                />
+              )}
+            />
+
+            <FormErrorMessage alignItems="top">
+              {!pdfFileText && <FormErrorIcon h="1.5rem" as={BxsErrorCircle} />}
+              {!pdfFileText &&
+                (pdfFile
+                  ? `${errors.pdfFileText?.message}`
+                  : 'Please upload a PDF.')}
```
justynoh: These could be combined, I think. 
```suggestion
              {!pdfFileText && <>
              <FormErrorIcon h="1.5rem" as={BxsErrorCircle} />
                (pdfFile
                  ? `${errors.pdfFileText?.message}`
                  : 'Please upload a PDF.')
                  </>
                  }
```

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderWizardProvider.tsx
```diff
@@ -0,0 +1,123 @@
...
+     * TODO: Wire up with actual mutations
+     */
+
+    const mockPdfMutation = useMutation(() => {
+      return new Promise((resolve) => {
+        setTimeout(() => {
+          resolve('success')
+        }, 2000)
+      })
+    })
```
justynoh: Never used - we can delete this

src/app/modules/form/admin-form/admin-form.assistance.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+  unknown,
+  ChatResponseMessage | ErrorDto,
+  {
+    type: string
+    content: string
+  }
+> = async (req, res) => {
+  const { error, value } = generateQuestionsSchema.validate(req.body)
+  if (error) {
+    return res.status(400).json({ message: error.details[0].message })
```
justynoh: We should have proper error logs for every error return. 

At the start of the file we should define a logger, and then `logger.error` before each of these error returns with appropriate error messages and logging metadata

src/app/modules/form/admin-form/admin-form.assistance.service.ts
```diff
@@ -0,0 +1,171 @@
...
+        function: getExpectedQuestionsListTool,
+      },
+    ],
+  } as GetChatCompletionsOptions
+
+  return ResultAsync.fromPromise(
+    azureOpenAi.getChatCompletions(deploymentId, messages, options),
+    (error) => {
+      let errorMessage = ''
+      // todo: return different error messages based on error codes
```
justynoh: is this done? if so, should we delete this comment?

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1973,6 +1973,65 @@ export const _handleCreateFormField: ControllerHandler<
...
+ * @returns 404 when form cannot be found
+ * @returns 409 when form field update conflicts with database state
+ * @returns 410 when creating form field for an archived form
+ * @returns 413 when creating form field causes form to be too large to be saved in the database
+ * @returns 422 when an invalid form field creation is attempted on the form
+ * @returns 422 when user in session cannot be retrieved from the database
+ * @returns 500 when database error occurs
+ * todo: add validation
+ */
+export const _handleCreateFormFields: ControllerHandler<
```
justynoh: This doesn't seem like it has to be preceded by an underscore. Underscore usually implies that the function is used internally by the module, and in particular doesn't really have any value when seen outside of it's surrounding scope (or is part of a larger flow that makes more sense when viewed together). 

I see maybe we intended to implement some kind of validation prior to calling this. it doesn't seem like we have this yet. So, maybe we should remove the underscore for now. Separately, we should delete the comment on validation unless it still needs to be done.

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+              }}
+              name="pdfFileText"
+              control={control}
+              render={() => (
+                <Attachment
+                  isRequired
+                  maxSize={MAX_FILE_SIZE}
+                  accept={VALID_EXTENSIONS}
+                  showFileSize
+                  title={'Upload a PDF'}
```
justynoh: This doesn't need enclosing braces (`{}`)

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PromptDetailsScreen.tsx
```diff
@@ -0,0 +1,170 @@
...
+                <Text
+                  textStyle="subhead-1"
+                  color="secondary.500"
+                  fontWeight="400"
+                  px="1rem"
+                  textAlign="justify"
+                >
+                  employee feedback on workplace environment and culture.
+                </Text>
+              </Button>
```
justynoh: the code for each of these prompt ideas are repeated. maybe we can factor them out to an array map instead? e.g. above we can define 

```
  const promptIdeas = [ 'employee feedback on workplace environment and culture.', 'event registrations and dietary preferences for dance ...]
```

 and then below when defining the component we can simply do

```
promptIdeas.map((idea: string) => <Button ....)
```

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PdfDetailsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const {
+    setError,
+    control,
+    setValue,
+    formState: { errors },
+  } = formMethods
+
+  const MAX_FILE_SIZE = 5 * MB
+  const VALID_EXTENSIONS = '.pdf'
```
justynoh: These are constants which are magic values. Should we factor them out to a `constants.ts` file?

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PromptDetailsScreen.tsx
justynoh: this can also be factored out into the constants file in my other comment :)

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/GifLoadingScreen.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const MagicFormBuilderGifLoadingScreen = (): JSX.Element => {
+  return (
+    <>
+      <ModalBody>
+        <Box display="flex" justifyContent="center">
+          <LottieAnimation
+            animationData={updatedLoadingAnimation}
+            width="47.4rem"
+            height="22rem"
+          ></LottieAnimation>
```
justynoh: nit: this can be `<LottieAnimation animationData= .... />` (i.e. we can use the self-closing syntax here)

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderModalContent/PromptDetailsScreen.tsx
```diff
@@ -0,0 +1,170 @@
...
+                maxLength: {
+                  value: 300,
+                  message: 'Please enter at most 300 characters.',
+                },
+              })}
+              placeholder="personal particulars for an event"
+            />
+            <FormErrorMessage>
+              <FormErrorIcon h="1.5rem" as={BxsErrorCircle} />
+              {errors.prompt?.message} {inputValue.length}/300
```
justynoh: this 300 also seems like a magic number - can we codify it and inject it as a variable instead?

frontend/src/features/admin-form/create/builder-and-design/MagicFormBuilderModal/MagicFormBuilderWizardContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  PromptDetails = 'prompt-details',
+  PdfDetails = 'pdf-details',
+}
+
+export type MagicFormBuilderWizardInputProps = {
+  title: string
+  magicFormBuilderMode: MagicFormBuilderMode
+  // PDF mode props
+  pdfFileText: string
+  // Prompt mode pros
```
justynoh: typo: props?

src/app/modules/form/admin-form/admin-form.assistance.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+// Validate inputs by the user for form builder:
+const generateQuestionsSchema = Joi.object({
+  type: Joi.string()
+    .valid(...Object.values(ContentTypes))
+    .required(),
+  content: Joi.string().when('type', {
+    is: ContentTypes.PROMPT,
+    then: Joi.string().max(300).required(),
```
justynoh: ah, this is where the 300 is validated. maybe we should create a constants file in `shared/constants/assistance.ts` or something like that for this number (and can be shared with other shared constants too)

src/app/modules/form/admin-form/admin-form.assistance.controller.ts
justynoh: also, we should use `HTTP_STATUS_CODES` rather than hardcoding these numbers (400, 500, 200) everywhere, if possible!

PR #7116, opened 2024-03-01
feat(fe): update copy, copy btn

KenLSM:
## Problem

There's a better test card filling UX in Airwallex, and found that we are lacking behind.

Screenshot 2024-03-01 at 2 32 15 PM

## Solution

- Update card filling copy
- Add copy button that copies test card number over to user's clipboard

**Breaking Changes** 

## Before & After Screenshots
Screenshot 2024-03-01 at 7 02 53 PM
Screenshot 2024-03-01 at 7 02 48 PM
Screenshot 2024-03-01 at 7 02 43 PM


kathleenkhy (approved):
lgtm!

PR #7120, opened 2024-03-04
chore(dd): link source code to dd

KenLSM:
## Problem

Source Code is not linked to DD. Having this link would help us to identify the likely line of code that has caused the error.

## Solution

To include DD variables into `env` that dd would load on init .
```
DD_GIT_COMMIT_SHA="<commitSha>"
DD_GIT_REPOSITORY_URL="<gitRepo>"
```

1. Rename `.ebextensions/env-file-creation.config` to `.ebextensions/01env-file-aws-ssm.config`
2. Add `.ebextensions/02env-file-datadog.config`

**Notes**
The file starts with 01 as `eb` will run the scripts in order of alphabetical order.  i.e., `01*.config` runs before `02*.config`

`.env` is created/overwritten in `01*.config` while `.ebextensions/02env-file-datadog.config` append the new envvars into `.env` directly. Thus, we have to ensure that `.ebextensions/01env-file-aws-ssm.config` gets executed first.

**Breaking Changes** 

## Tests

Regression
- [ ] Make a Form Submission
- [ ] Observe that it succeeds


KenLSM:
@wanlingt this might be useful for Signify as reference!

wanlingt:
> @wanlingt this might be useful for Signify as reference!

thanks @KenLSM !!

justynoh (approved):
LGTM

PR #7134, opened 2024-03-13
refactor(i18n): modularize and shard by feature

LoneRifle:
## Problem
Internationalization of FormSG is currently slightly held back by the hard-coding of language strings throughout the codebase. One possible cause is the monolithic nature of our i18n resource modules.

### Background
@msuhardi has invested considerable time and effort looking into making FormSG available in Bahasa Indonesia. The resulting fork that was made incorporates a refactoring of the `i18n` module, as well as replacement of a large part of FormSG's hard-coded language strings into `t()` equivalents.

This PR, and subsequent ones raised in future, aims to port some of these changes back.

## Solution
- Prepare for sharding up of locale resources by moving each locale into
its own directory. Use this as the base to manage future growth and
expansion of locale resources to cover most text in FormSG.

- Hold feature-specific locale resources in a `features/` dir for each
locale, to minimise clutter even as we add more text to hold as
resources.

**Breaking Changes** 

### DIFF
--- a/frontend/src/i18n/locales/en-sg.ts
+++ b/frontend/src/i18n/locales/en-sg.ts
@@ -1,6 +1,5 @@
-import Translation from '../types'
-
-import { login } from './features'
+import { enSG as login } from './features/login'
+import Translation from './types'
 
 export const enSG: Translation = {
   translation: {
--- a/frontend/src/i18n/locales/en-sg/features/index.ts
+++ b/frontend/src/i18n/locales/en-sg/features/index.ts
@@ -1,0 +1,1 @@
+export * from './login'
--- a/frontend/src/i18n/locales/en-sg/features/index.ts
+++ b/frontend/src/i18n/locales/en-sg/features/index.ts
@@ -1,1 +1,0 @@
-export * from './login'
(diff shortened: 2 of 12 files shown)


KenLSM (approved):
LGTM. 

Nits. Any particular reason to shard by `locale/features` instead of `feature/locale`? 

Features are usually introduced more often than languages, thus it would make more sense for us to work (and add localize) within a feature rather than across a locale.

```
// feature/locale
frontend/src/i18n/features/login/index.ts
frontend/src/i18n/features/login/en-sg.ts
frontend/src/i18n/features/login/zh-sg.ts

vs 

// locale/features
frontend/src/i18n/locales/en-sg/features/login.ts
frontend/src/i18n/locales/zh-sg/features/login.ts
```

LoneRifle:
Hmm. I don't mind, though I will point out that every locale shard will have to be consolidated into a single file for injection into i18next. This is similar to the approach taken for Chakra theming.

Having locale paths in every feature would mean that the central locale file will have imports from every feature in Form. If the team is okay with that, I'll change the PR accordingly.

Will wait for your reply @KenLSM

LoneRifle:
@KenLSM - makes sense. Let me rework this

KenLSM (approved):
LGTM. Merging

PR #7129, opened 2024-03-07
chore: add logs for virus scanning duration

kathleenkhy:
## Problem
Currently we do not keep track of how long it will take for files to be downloaded from S3 - this arose from an [alarm](https://www.notion.so/opengov/Storage-Form-Submission-taking-too-long-ab0b0e1f97d04833a25d67d617461b64?pvs=4) due to a spike in time taken for scanning and downloading clean attachments

## Solution
Added logs in `src/app/modules/submission/submission.service.ts` to track the start and end of the download from S3

**Breaking Changes** 

- No - this PR is backwards compatible  

## Manual tests

- [ ] Create a form
- [ ] Complete a submission of the form
- [ ] In datadog, search for "Successfully downloaded file from S3"
- [ ] Copy the Clean File Key
- [ ] Search for the clean file key in datadog or cloudwatch and check that there will be 2 logs with the messages of "File download from S3 has started" and "Successfully downloaded file from S3".

Datadog:
image

Cloudwatch:
image

Searching by clean file ID of form submitted:
image

PR #7135, opened 2024-03-13
chore: file download duration from S3

kathleenkhy:
## Problem
related to https://github.com/opengovsg/FormSG/pull/7129 to track duration for file to be downloaded from S3

Closes [insert issue #]

## Solution
Added logs to calculate time taken from start to end time of file download

**Breaking Changes** 

## Manual Tests
- [ ] Create a form with an attachment field
- [ ] Complete a submission of the form
- [ ] In datadog, search for "File download from S3 duration" - there should be a log result
- [ ] Check that in Datadog, there should be logs under Submissions for a graph to track the duration of submissions (under Submissions >> Duration of file download from S3 (after virus scan))


src/app/modules/submission/submission.service.ts
```diff
@@ -385,11 +385,27 @@ export const downloadCleanFile = (cleanFileKey: string, versionId: string) => {
     })
     .createReadStream()
 
+  const timeDownloadStart = Date.now()
```
KenLSM: /nit usually we have the convention of the datatype to be at the end of the variable i.e., `studentNricMap` over `mapStudentNric`.

KenLSM (approved):
lgtm, 1 nit.

PR #7136, opened 2024-03-15
refactor(logic): consolidate logic functions into shared

justynoh:
## Problem

There is currently some tech debt incurred as a result of the angular->react migration that has not yet been cleaned up. In particular, the functions used to evaluate logic have not been refactored from `src` to `shared`, because the client-side types for input data differs between angular and react. 

As a result, for a period of time, the logic for logic (ha!) had to work over three schemas if they were truly to be shared - server, angular and react. Instead of making this work, we basically duplicated the logic and made the types work for the frontend, while retaining the original backend code in src.

Now that angular is gone, however, it's time to make this code truly shared.

## Solution

This PR moves `src/shared/utils/logic` into `shared/utils/logic` and adjusts the types to account for existing behavior both on the frontend and backend.

There are no behavioral changes from the current behavior, and we should not expect any cutover period due to desynced fe/be. 

**Breaking Changes** 

## Tests

All tests are regression tests.
- [ ] Show fields logic
  - [ ] Show fields based on "greater than or equal" to a number/decimal field.
  - [ ] Show fields based on "less than or equal" to a number/decimal field.
  - [ ] Show fields based on "equal" to a decimal field.
  - [ ] Show fields based on "either" for a radio field.
  - [ ] Show fields based on "either" for a radio field with "Others" as an option.
- [ ] Prevent submit logic: repeat the same tests as above, but preventing submit instead of show fields.


justynoh:
Also, this was a side quest as a result of tackling backend validation for MRF, which requires these logic functions to work well.

frontend/src/features/logic/utils/logic-adaptor.ts
```diff
@@ -0,0 +1,107 @@
...
+      const fieldTypeAndInput = {
+        fieldType: field.fieldType,
+        input: filteredFormInputs[_id],
+      }
+      // Type narrowing to help the typechecker along with complex if-else types
+      if (isNotLogicableField(fieldTypeAndInput)) {
+        return { _id, fieldType: fieldTypeAndInput.fieldType }
+      } else if (isLogicableField(fieldTypeAndInput)) {
+        return { _id, ...fieldTypeAndInput }
+      } else {
```
KenLSM, line 57: this looks iffy... but I can't think of a better solution

frontend/src/features/logic/utils/logic-adaptor.ts
```diff
@@ -0,0 +1,107 @@
...
+      // Type narrowing to help the typechecker along with complex if-else types
+      if (isNotLogicableField(fieldTypeAndInput)) {
+        return { _id: ff._id, fieldType: fieldTypeAndInput.fieldType }
+      } else if (isLogicableField(fieldTypeAndInput)) {
+        return { _id: ff._id, ...fieldTypeAndInput }
+      } else {
+        throw new Error('Unexpected fallthrough')
+      }
+    })
+    .filter(isNotNull)
```
KenLSM: Possible to reuse `isNonEmpty` from `shared/utils/isNonEmpty.ts`?

frontend/src/features/logic/utils/logic-adaptor.ts
```diff
@@ -0,0 +1,107 @@
...
+      // Type narrowing to help the typechecker along with complex if-else types
+      if (isNotLogicableField(fieldTypeAndInput)) {
+        return { _id, fieldType: fieldTypeAndInput.fieldType }
+      } else if (isLogicableField(fieldTypeAndInput)) {
+        return { _id, ...fieldTypeAndInput }
+      } else {
+        throw new Error('Unexpected fallthrough')
+      }
+    })
+    .filter(isNotNull)
```
KenLSM: Possible to reuse `isNonEmpty` from `shared/utils/isNonEmpty.ts`?

shared/utils/logic.ts
```diff
@@ -265,102 +312,137 @@ export const getVisibleFieldIds = (
...
   let currentValue = getCurrentValue(field)
+
+  // Trim the current value
+  if (typeof currentValue === 'string') {
+    currentValue = currentValue.trim()
+  }
+
   if (
     currentValue === null ||
     currentValue === undefined ||
```
KenLSM, line 357: /nit + not part of your code change

Can be terser with
```suggestion
    currentValue == null
```

shared/utils/logic.ts
```diff
@@ -1,20 +1,19 @@
...
+  ShowFieldLogicDto,
+  PreventSubmitLogicDto,
   LogicCondition,
-} from '../../types'
+  LogicableField,
+} from '../types'
+
+// TODO: Keep this in sync with the frontend value.
+// Exported for testing
+export const RADIO_OTHERS_INPUT_VALUE = '!!FORMSG_INTERNAL_RADIO_OTHERS_VALUE!!'
```
KenLSM: Can this be moved into `shared/constants` so that both references of `frontend/src/templates/Field/Radio/constants.ts` and `shared/utils/logic.ts` be synchronized?

shared/utils/logic.ts
```diff
@@ -141,12 +193,8 @@ export const groupLogicUnitsByField = (form: IFormDocument): GroupedLogic => {
           logicUnitsGroupedByField[fieldId].push(logicUnit.conditions)
         }
       })
-    } else {
-      hasInvalidLogic = true
     }
   })
-  if (hasInvalidLogic && formId)
-    console.info(`formId="${form._id}" message="Form has invalid logic"`)
```
KenLSM, line 149: Interesting log...

shared/utils/logic.ts
```diff
@@ -265,102 +312,137 @@ export const getVisibleFieldIds = (
...
+       *     If not, continue the usual string-based check with (3).
+       * (4) If conditionValues includes "Others", check for the value being
+       *     special radio value and the othersInput subfield having a value.
+       *     If not, continue the usual string-based check with (3).
+       * (2) + (5) Use Number equality for decimals
+       * (3) + (6) Use String equality for everything else
+       */
+
+      // Special radio others check
+      if (
```
KenLSM, line 401: I had quite a hard time matching 1-6 over to each logic.

Does it make sense to indicate which case each control statement is handling?

frontend/src/features/logic/utils/logic-adaptor.ts
justynoh: the typechecker needs a lot of help in this case. like you said, I don't really have a good way of doing this other than just guaranteeing at runtime that there is no uncaught case (i.e. guaranteed only by the fact that `isNotX(args)` is defined to be `!isX(args)`). 

I suppose an alternative is just to like. do a single if-else, and then //@ts-ignore the rest but that feels even more hackish. Having a typeguard, even if self-defined and somewhat weird, is better than having no typeguard at all. I think I might just leave a comment explaining why it's a bit complicated in this case, and that we would rather have the types than not even if it results in strange unreachable code paths that aren't obvious to the typechecker.

shared/utils/logic.ts
justynoh: thanks for pointing it out! Added here [86e5e93](https://github.com/opengovsg/FormSG/pull/7136/commits/86e5e937ff3ebece7ec249bcc8ebeab48f0fd162)

frontend/src/features/logic/utils/logic-adaptor.ts
justynoh: ohh nice catch! replaced here [86e5e93](https://github.com/opengovsg/FormSG/pull/7136/commits/86e5e937ff3ebece7ec249bcc8ebeab48f0fd162)

frontend/src/features/logic/utils/logic-adaptor.ts
justynoh: [86e5e93](https://github.com/opengovsg/FormSG/pull/7136/commits/86e5e937ff3ebece7ec249bcc8ebeab48f0fd162)

shared/utils/logic.ts
justynoh: done here with shared value being ssot [f270eea](https://github.com/opengovsg/FormSG/pull/7136/commits/f270eea916e18612ac1a13c1e4aa59d62e310d4c)

KenLSM (approved):
LGTM! Yay for cleanups

PR #7180, opened 2024-03-21
feat: disabling response limit toggle for MRF

kathleenkhy:
## Problem
Imposing a response limit is only applicable for forms in email or storage mode and not for MRF (because response tracking works differently in the context of MRF). Hence the response limit toggle should be disabled for MRF forms.

Closes FRM-1693

## Solution
As a user, the response limit toggle should be disabled when using an MRF so that it is clear this feature is not intended for MRFs (and reduce potential confusion and errors if a response limit were to be set for MRFs)

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Manual Tests

- [ ] Create a multi-respondent form
- [ ] Proceed to Settings >> General
- [ ] Response limit toggle should be disabled

### DIFF
--- a/frontend/src/features/admin-form/settings/components/FormLimitToggle.tsx
+++ b/frontend/src/features/admin-form/settings/components/FormLimitToggle.tsx
@@ -109,7 +109,7 @@ export const FormLimitToggle = (): JSX.Element => {
   const { data: settings, isLoading: isLoadingSettings } =
     useAdminFormSettings()
 
-  const isMRF = settings?.responseMode === FormResponseMode.Multirespondent
+  const isMrf = settings?.responseMode === FormResponseMode.Multirespondent
 
   const { data: responseCount, isLoading: isLoadingCount } =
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/settings/components/FormLimitToggle.tsx
```diff
@@ -106,6 +109,8 @@ export const FormLimitToggle = (): JSX.Element => {
   const { data: settings, isLoading: isLoadingSettings } =
     useAdminFormSettings()
 
+  const isMRF = settings?.responseMode === FormResponseMode.Multirespondent
```
KenLSM: We usually only capitalize the first letter of an acronym so that variables that are made up of acronyms doesn't become an abomination like `XMLHTTPAsJSON` vs `xmlHttpAsJson`

KenLSM (approved):
lgtm!

PR #4275, opened 2022-07-22
feat(v2): add handling of prefill for authed public forms, show fields correctly when logged in

karrui:
## Problem

This PR implements prefill for authenticatable public forms following AngularJS implementation in #3920.
Also contains a bug fix where form fields are still not shown when user is logged in.

Closes #3630

## Solution

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Features**:

- add hooks to store and retrieve prefill query strings when user is logging into a form
  - feat: add `useStorePrefillQuery` and `useFetchPrefillQuery` hooks
- feat: reset default form values to new values if changed (allows for query to get updated when search params get set programatically)
- add `useSessionStorage` hook for session storage manipulation

**Bug Fixes**:

- fix: correctly show public auth only if spcpSession is not defined

PR #7181, opened 2024-03-21
feat(mrf): attachments

KenLSM:
## Problem

Attachments can be accepted for respondent 1, but they are not loaded onto the page for respondent 2 onwards, nor are they sent as data within the form responses object, `FormResponsesV3`

Closes FRM-1590

## Solution

Refer to Notion [doc](https://www.notion.so/opengov/MRF-v1-Attachments-c04176ff383a47439b850eecb463cca3?pvs=4)

## Before & After Screenshots

Attachments should be able to:

- See there is an existing one, download it
- Remove and re-upload
- Attachments are supported by field locking

https://github.com/opengovsg/FormSG/assets/12391617/8c6d5261-e505-4136-afce-3babc0c5ea28

## Tests

Preparation: 
- [ ] Create an MRF form
  - [ ] Add 2x attachment field
  - [ ] Assign 3 workflow steps
    - [ ] Mark AttachmentField #1 required for Step 1
    - [ ] Mark AttachmentField #2 required for Step 2
    - [ ] Mark AttachmentField #1 + #2 required for Step 3

Respondent 1
- [ ] Attach file 1 (Respondent #1) to field 1
- [ ] Ensure that field 2 is disabled
- [ ] Ensure that form can be submitted

Respondent 2
- [ ] Attach file 2 (Respondent #2) to field 2
- [ ] Ensure that field 1 is disabled
- [ ] Ensure that download of file 1 is allowed and content is correct
- [ ] Ensure that form can be submitted

Respondent 3
- [ ] Ensure that download of file 1 is allowed and content is correct
- [ ] Ensure that download of file 2 is allowed and content is correct
- [ ] Attach file 3 (Respondent #3) to field 1
- [ ] Attach file 4 (Respondent #3) to field 2
- [ ] Ensure that form can be submitted

Attachments downloaded on admin response page should be the last submitted content
- [ ] Ensure that the attachment for field 1 is file 3
- [ ] Ensure that the attachment for field 2 is file 4

Ensure that scanner works
- [ ] Using the MRF form prepared earlier
- [ ] Download the eicar [test file](https://www.eicar.org/download-anti-malware-testfile/)
- [ ] Attach eicar file to field 1
- [ ] Ensure that the form fails to submit and rejects with file has failed virus scan


frontend/src/components/Field/Attachment/AttachmentFileInfo.tsx
```diff
@@ -37,13 +45,25 @@ export const AttachmentFileInfo = ({
...
+          />
+        ) : null}
+        {showDownloadButton ? (
+          <IconButton
+            variant="clear"
+            aria-label="Click to download file"
+            icon={<BiDownload />}
+            onClick={handleDownloadFile}
+          />
+        ) : null}
```
justynoh, line 65: nit: can be `enableRemove &&` and `showDownloadButton &&`

frontend/src/templates/Field/Attachment/AttachmentField.tsx
```diff
@@ -70,6 +75,15 @@ export const AttachmentField = ({
...
           const clone = new File([buffer], file.name, { type: file.type })
+
+          /**
+           * Set a custom field to force attachment field to remain dirty.
+           * React Hook Form is unable to evaluate dirtiness file when comparing File objects https://react-hook-form.com/docs/useformstate#return
+           * React Hook Form has a custom deepEqual comparator function that checks for key values on the object https://github.com/react-hook-form/react-hook-form/blob/v7.51.1/src/utils/deepEqual.ts
+           * By introducing a new `__dirtyField` property, so we can set force the evaluation deepEqual to be false, thus remaining dirty.
+           * */
+          // @ts-expect-error __dirtyField is not a standard property of File.
+          clone.__dirtyField = 1
```
justynoh, line 86: Is this something mrf-specific or does this apply to all modes? It sounds like a generic statement

Also, I'm not sure how this helps in keeping the field dirty. What's the context for this?

shared/types/response-v3.ts
```diff
@@ -115,3 +115,48 @@ export type AttachmentFieldResponseV3 = {
...
+ * Note: JSON.parse(JSON.stringify(fooBuffer)) does not reconstruct the Buffer object
+ */
+export type HydratedClearAttachmentFieldResponseV3 =
+  AttachmentFieldResponseV3 & {
+    filename: string
+    content: {
+      data: string
+      type: 'buffer'
+    }
+  }
```
justynoh: Can clarify here what the meaning of "Hydrated" is? haha. Which part of the backend hydrates the attachment

frontend/src/templates/Field/Attachment/AttachmentField.tsx
justynoh: Update: I looked at the Notion doc again - just to be clear, so this is purely to support Singpass auth in the future? is it needed at this juncture?

shared/utils/response-v3.ts
```diff
@@ -28,10 +34,21 @@ export const isFieldResponseV3Equal = (
...
+      const lMd5 = crypto
+        .createHash('md5')
+        .update(Buffer.from(lAnswer.content))
+        .digest()
+
+      const rMd5 = crypto
+        .createHash('md5')
+        .update(Buffer.from(rAnswer.content.data))
+        .digest()
+
```
justynoh: This part feels a bit strange in principle... because equality should be a symmetric function, L and R should have the same type. But in this case, we presume asymmetry  of L and R while it not being reflected in the argument types (both of them have argument types `FieldResponseV3`). This means any user of this function has to know which value to put on the left and which one to put on the right.

Should we make it the responsibility of the backend to re-transform it's input into a format amenable to this function prior to calling it? Then we leave the key part of the code allowing for symmetry `r.answer as typeof l.answer` alone. Then we can move the parsed and hydrated and clear attachment types to only be used in the backend.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -446,18 +451,40 @@ export const validateMultirespondentSubmission = async (
...
+                      prevResField.fieldType === BasicField.Attachment
+                    ) {
+                      incomingResField.answer.answer =
+                        prevResField.answer.answer
+
+                      incomingResField.answer.filename =
+                        prevResField.answer.answer
+                    }
+
+                    return ok(undefined)
```
justynoh, line 499: what's the point of this? this replaces the response body data with the original filenames right. was this the case where someone downloads an attachment, renames it and re-uploads the same attachment?

shared/types/response-v3.ts
```diff
@@ -115,3 +115,48 @@ export type AttachmentFieldResponseV3 = {
...
+}
+
+export type ParsedClearFormFieldResponseV3 =
+  | Exclude<FieldResponseV3, AttachmentResponseV3>
+  | ParsedClearAttachmentResponseV3
+
+export type ParsedClearFormFieldResponsesV3 = Record<
+  FormFieldDto['_id'],
+  ParsedClearFormFieldResponseV3
+>
```
justynoh: I see this was moved here to be used in the shared `isFieldResponseV3Equal` function but... - see my comment in that file

frontend/src/components/Field/Attachment/AttachmentFileInfo.tsx
KenLSM: Coming from some history of RN, I usually avoid using `&&` and stick with ternary operators due to its potential pitfalls when the left-hand value is falsy but react actually renders it (i.e., `NaN` and `0`), on React Native, it crashes the whole app instead.

Usually this is fine as the implementer of the render section is the same person that created the variable. But, it may not be the same in the future where `showDownloadButton`. It had caused many headbangs and written reports in the past that we avoid it with a 10-foot pole.

shared/types/response-v3.ts
KenLSM: https://github.com/opengovsg/formsg-javascript-sdk/blob/develop/src/crypto-v3.ts#L76

I'm referring to hydration (since parsed is already used) to be `JSON.parse(foo)`

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
KenLSM: This is to handle the case where the attachment name change that is caused by [`handleDuplicatesInAttachments`](https://github.com/opengovsg/FormSG/blob/develop/src/app/modules/submission/receiver/receiver.utils.ts#L133)

frontend/src/templates/Field/Attachment/AttachmentField.tsx
KenLSM: > Is this something mrf-specific or does this apply to all modes? It sounds like a generic statement

All modes, but only applicable to MRF since it is the only mode that has attachment "prefills".

> Update: I looked at the Notion doc again - just to be clear, so this is purely to support Singpass auth in the future? is it needed at this juncture?

It is not to support Singpass auth. It is to address a specific deficit on RHF.

Specifically, I'm referring to this interaction [here](https://github.com/opengovsg/FormSG/pull/4275/files#diff-42654152f38821549821e8fd0cb30c57a7fadddc8a26b2bc714cd107c9c1f9b3R98).

1) It checks whether `formState.isDirty` and does a reset.
2) RHF is unable to evaluate dirtiness file when comparing `File` objects (read: when respondent changes the attachment field aka a `File` object, it will not be recognized as dirty by RHF)

Both 1 & 2 results in the form resetting to its defaultValues even when the respondent did change a file[a].

Then it begs the question, can we remove the offending code that does the reset? 
```
  useEffect(() => {
    if (!isDirty) {
      reset(defaultFormValues)
    }
```

Well, no. Because it is meant to populate the form fields where the query params changes when user returns from the spcp login flow, thus this means that this code is performing as intended.

The question now becomes, what is not performing as intended. For this, I've deemed the `isDirty` to be incorrect, as it is reasonable for us to presume that `isDirty` should have been `true` here since the attachmentField has been updated, but had wrongfully identified that attachmentField to be not dirty.


[a] Walkthrough of how defaultValues + `isDirty` affects the attachment field.
```
1. Respondent 1 attaches a file, and submits.
2. Respondent 2 renders form
    - defaultValue['attachmentFieldId'] = someFileObj // as per [426541](https://github.com/opengovsg/FormSG/pull/7181/files#diff-42654152f38821549821e8fd0cb30c57a7fadddc8a26b2bc714cd107c9c1f9b3R111)
    - formValue['attachmentFieldId'] = someFileObj // as populated by defaultValue
    - formState['attachmentFieldId'].isDirty // false
4. Respondent 2 changes file 
    - defaultValue['attachmentFieldId'] = someFileObj
    - formValue['attachmentFieldId'] = newFileObj // as populated by defaultValue
    - formState['attachmentFieldId'].isDirty // false!! Because https://react-hook-form.com/docs/useformstate#return and the exact comparator is also linked above
```

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
justynoh: gotcha. attachments are so annoying

so let's say R1 submits attachment Q1: `a.txt`, and R2 submits attachment Q2: `a.txt` but the two files are different.  `handleDuplicatesInAttachments` will rename Q1 to `1-a.txt` in `q1.answer.answer`, but this code snippet restores the `q1.answer.answer` to be `a.txt` - am i understanding this correctly?

shared/types/response-v3.ts
justynoh: ahhhhhh. hmmmmmm. terminology is hard. i see. I was gonna suggest "Decrypted" but that also doesn't help because it also conflates meaning with other decrypted attachment forms. I think this is fine but maybe we can add a note about what it refers to (i.e. that its specifically attachments that were decrypted from the DB/S3)

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
KenLSM: Yes exactly.

shared/types/response-v3.ts
KenLSM: Im thinking if there's a more elegant way to handle your comment [below](https://github.com/opengovsg/FormSG/pull/7181#discussion_r1546018740). This is merely an artefact from it.

frontend/src/templates/Field/Attachment/AttachmentField.tsx
justynoh: oh i see. the `reset` is necessary because of our query param handling within the spcp flow, since we restore the query params from browser storage in code. 🤦  

as a follow-up then, let's say R2 changes the file once, so that `__dirtyField = 1`. If R2 changes the file again, then `__dirtyField = 1` will continue to remain. will it work as intended in this case? it seems like it should still be dirty right? 

If so, would it make sense to do something like `__dirtyField = __dirtyField ? 0 : __dirtyField + 1` or something? (this would be capped by the browser maxint but like... no one is ever going to hit it)

or is my understanding still incorrect

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
justynoh: gotcha. thanks!

frontend/src/templates/Field/Attachment/AttachmentField.tsx
KenLSM: It will still be dirty, since rhf compares `formValue['attachmentFieldId']` with `defaultValue['attachmentFieldId']`. 

Thus no need to update `__dirtyField`.

shared/types/response-v3.ts
KenLSM: Removed in 247077854

shared/utils/response-v3.ts
KenLSM: Moved to caller 247077854

frontend/src/templates/Field/Attachment/AttachmentField.tsx
justynoh: Ah. I gotcha. Interesting... I think this is fine haha just wanted to clarify the intention

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -443,25 +450,58 @@ export const validateMultirespondentSubmission = async (
...
+                    if (prevResField.fieldType === BasicField.Attachment) {
+                      prevResField.answer.content = Buffer.from(
+                        /**
+                         * JSON.parse(JSON.stringify(fooBuffer)) does not fully reconstruct the Buffer object
+                         * it gets parsed as { type: 'Buffer', data: number[] } instead of the original Buffer object
+                         */
+                        // @ts-expect-error data does not exist on Buffer
+                        prevResField.answer.content.data,
+                      )
+                    }
```
justynoh, line 469: oof. Looks kinda complicated

justynoh (approved):
One thought I had: if the solution is going to involve us moving the shared equality function into backend, then the "generic" property of equality may not be needed anymore. The idea of writing a shared function framed as equality was because that's the most intuitive operation to compare two `FieldResponsesV3` in general. But now that it's in backend, is there value in keeping the equality function generic anymore? Maybe it should just be `isPreviousResponseAndNewResponseEqual(prevRes, newRes)` and then revert to having the two arguments have different types, rather than making the caller transform the types. Just for thought

Lgtm

PR #7005, opened 2024-01-08
chore: add advertising for Checkpoint

justynoh:
## Problem
We want to add advertising for Checkpoint on the FormSG platform. This PR does it.

## Solution

**Breaking Changes** 
- No - this PR is backwards compatible  

## Screenshots

Form creation modal

image

Form settings screen

image

PR #7191, opened 2024-03-25
chore: switch advertising for checkpoint to mrf

justynoh:
## Problem
We want to remove advertising for checkpoint and advertise mrf instead.

## Solution

Reference from #7005 

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


KenLSM (approved):
LGTM

PR #7106, opened 2024-02-21
feat(mrf): disable unsupported features

KenLSM:
## Problem

The users should not be able to use the features that do not work for the beta release
Closes FRM-1663

## Solution

**Breaking Changes** 

- No - this PR is backwards compatible  

### Attachments are disabled
disable attachments

### Mobile OTP is disabled
disable mobile otp

### Email OTP + Confirmation is disabled
email-mrf

##### New feature tests
Admin MRF Forms - Unsupported fields are disabled
- [ ] As an Admin create a non-MRF form
- [ ] Ensure that Attachment Field is disabled
  - [ ] Ensure that Attachment Field **cannot** be added as a form field
- [ ] Ensure that on Mobile Field, the OTP toggle is disabled
- [ ] Ensure that on Email Field, the OTP toggle is disabled
- [ ] Ensure that on Email Field, the OTP and email confirmation toggles are disabled

Respondent MRF Forms - Feedback section is not shown
- [ ] As a respondent, submit an MRF form
- [ ] Ensure that form feedback section is not shown

##### Regression
Non-MRF Forms - fields that are disabled on MRF should not affect non-MRF forms
- [ ] As an Admin create a non-MRF form
- [ ] Ensure that Attachment Field is **not** disabled
- [ ] Ensure that Attachment Field **can** be added as a form field 
- [ ] Ensure that on Mobile Field, the OTP toggle is not disabled
- [ ] Ensure that on Email Field, the OTP and email confirmation toggles are **not** disabled

Respondent Non-MRF Forms - Feedback section is shown
- [ ] As a respondent, submit a form
- [ ] Ensure that form feedback section is shown

PR #7192, opened 2024-03-25
fix(mrf): skip verifying OTP signatures on decryption for export

justynoh:
## Problem

On MRFs, verified fields are currently not available (email and mobile OTP verification). The primary reason is that while the verification flows all work, exporting fails due to the signatures being matched with the `createdAt` time of the submission.

Closes FRM-1615

## Solution

This PR skips the verification check for MRFs so that it is only conducted on storage mode responses. 

At the same time, the OTP verification toggles for email and mobile are re-enabled (ref: #7106).

Potential TODO: Open another ticket to investigate if re-enabling this is required.

**Breaking Changes** 

## Tests

- [ ] Create an MRF form with email and mobile otp for respondent 1 and another email and mobile otp for respondent 2. Fill the form for both respondents, then attempt to export the results as CSV. This should work.


KenLSM (approved):
lgtm

PR #7201, opened 2024-03-26
chore(mrf): prevent form activation if email confirmation is present

justynoh:
## Problem
When users duplicate their forms from email/storage mode to MRF, they may have email confirmations present. We want to prevent them from opening the form with this setting enabled. This is purely a frontend block.

Closes FRM-1677

## Solution

Perform a check on the activation toggle, pre-activation. Also, enable users to un-toggle the email confirmation in the email field themselves (i.e., only disable the toggle if email confirmations is already off).

**Breaking Changes** 

## Screenshots

image


kathleenkhy (approved):
lgtm

PR #7214, opened 2024-03-27
fix(mrf): remove any response limits for MRF

kathleenkhy:
## Problem
If a user already sets a response limit and duplicates the form as an MRF, the response limit would be set but is disabled and the user will not be able to turn off the response limit:

image

Closes FRM-1698

## Solution
If a user duplicates a form to MRF that was originally on email/storage mode, and a response limit has been set, the new MRF form should have it's response limit switched off.

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:
image

## Manual Tests

- [ ] Create a form on email or storage mode
- [ ] Switch on the response limit (using the toggle) under Settings
- [ ] Duplicate the form which has a response limit set to an MRF
- [ ] Check that within the duplicated MRF, the response limit under Settings is switched off and the toggle disabled (hence there is no way for the response limit to be on for MRFs)

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.service.ts
@@ -600,7 +600,8 @@ export const duplicateForm = (
     overrideProps.endPage = omit(originalForm.endPage, 'buttonLink')
   }
 
-  // if MRF, set submissionLimit (i.e. response limit) = null
+  // if MRF, set submissionLimit = null
+  // this is because MRF does not support submission limit (i.e. response limit)
 
   if (overrideProps.responseMode === FormResponseMode.Multirespondent) {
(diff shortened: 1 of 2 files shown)


src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -600,6 +600,12 @@ export const duplicateForm = (
     overrideProps.endPage = omit(originalForm.endPage, 'buttonLink')
   }
 
+  // if MRF, set submissionLimit (i.e. response limit) = null
```
KenLSM: /nit might want to explain why we are doing this instead

Mrf doesn't support submission limit

KenLSM (changes requested):

KenLSM (approved):

PR #7216, opened 2024-03-28
refactor: move override code closer to other overrides

KenLSM:
## Problem

There's already a function that keeps all the duplication exclusion rules. We should use that.

Thanks @justynoh for highlighting an even better way.

## Solution

Move override statement into `processDuplicateOverrideProps` function.


kathleenkhy (approved):
😮 lgtm!!

PR #7219, opened 2024-04-01
fix(mrf): handling for incorrect submission secret key in url query params

justynoh:
## Problem
We want to only allow keys to be passed via the `key` query param, and not via a secret key prompt. 
Closes FRM-1672

## Solution

Pull the decryption to the provider level to enable the screen to be swapped to the `FormNotFound` screen if the secret key is invalid.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


KenLSM (approved):
LGTM

PR #7225, opened 2024-04-02
fix: prefill myinfo dropdown field options when faking sample submission data

justynoh:
## Problem
Sample submission endpoint is causing server to crash for myinfo dropdown fields due to no field options being available to fake.

## Solution
Pre-populate myinfo dropdowns with the right fieldOptions. Additionally, add more logging upon failures.

**Breaking Changes** 

### DIFF
--- a/src/app/modules/form/public-form/public-form.controller.ts
+++ b/src/app/modules/form/public-form/public-form.controller.ts
@@ -424,6 +424,7 @@ export const handleGetPublicFormSampleSubmission: ControllerHandler<
       meta: logMeta,
       error,
     })
+    return res.sendStatus(HttpStatusCode.InternalServerError)
   }
 
   return res.json({ responses: sampleData })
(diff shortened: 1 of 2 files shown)


src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -400,16 +401,31 @@ export const handleGetPublicFormSampleSubmission: ControllerHandler<
...
   const publicForm = form.getPublicView() as PublicFormDto
 
   const formFields = publicForm.form_fields
   if (!formFields) {
-    throw new Error('unable to get form fields')
+    logger.error({
+      message: 'Form fields not found from form public view',
+      meta: logMeta,
+    })
+    return res.sendStatus(HttpStatusCode.InternalServerError)
```
KenLSM, line 415: ~I feel that this should be a 4xx since the admin didn't provide any fields to sample. The actor for this error should be the admin instead of us (i.e., the request was invalid, rather than server error)~

KenLSM: Strike that.

KenLSM (approved):
lgtm!

PR #7231, opened 2024-04-02
chore: revert "feat: disabling submit another response button (MRF)"

KenLSM:
Reverts opengovsg/FormSG#7186

PR #7232, opened 2024-04-02
fix: correcting submission button bug for MRF

kathleenkhy:
## Problem
Previous PR (https://github.com/opengovsg/FormSG/pull/7186) to prevent the Submission button from showing for MRF Respondent #2 onwards resulted in a bug (and incident) where users would not be able to access the Admin view of non-payment forms' Thank You page.

The PR was subsequently reverted to undo the problem introduced. (https://github.com/opengovsg/FormSG/pull/7231)

## Solution
To add the use of Public Form Context to the parent of EndPageBlock (i.e. FormEndPage). The EndPageBlock component is also used in the Admin View (to show as a preview while the Admin use edits the form), resulting in an error because the component that is `public-form` is also rendered in `admin-form`.

**Breaking Changes** 

## Tests

- [ ] Create a non-payment form
- [ ] Proceed to the Thank You page tab

MRF Form test
- [x] Create an MRF form
- [x] Create 2 workflows
- [x] Submit the 1st workflow - the Submit another response button should be present
- [x] Submit the 2nd workflow - the Submit another response button should not be present


KenLSM:
We should also refactor `frontend/src/features/public-form/components/FormEndPage/components/EndPageBlock.tsx` out of `public-form` and place it in a generic component folder, rather than one that implies it’s solely used for `public-form` i.e., possessing `PublicFormContext`.

It’s not very idiomatic and I can see myself getting conned by this gotcha.

KenLSM (approved):
LGTM!

PR #7239, opened 2024-04-03
chore: refactor endpageblock to generic folder

kathleenkhy:
## Problem
Related to PR (https://github.com/opengovsg/FormSG/pull/7232) to refactor by moving EndPageBlock.tsx file to a generic folder (out of the public-form folder) - this is to reduce confusion that the EndPageBlock.tsx is not entirely isolated to the public-form context and may cause downstream bugs and issues if the component is rendered incorrectly.

Closes FRM-1710

## Solution
By moving EndPageBlock.tsx to a generic folder that is not within public-form or admin-form

## Tests

- [ ] Create a non-payment form
- [ ] Proceed to the Thank You page tab

MRF Form test
- [ ] Create an MRF form
- [ ] Create 2 workflows
- [ ] Submit the 1st workflow - the Submit another response button should be present
- [ ] Submit the 2nd workflow - the Submit another response button should not be present


KenLSM (approved):
LGTM, but let's merge it after GA tomorrow.

PR #7240, opened 2024-04-03
fix: correct date validation for disabled fields

kathleenkhy:
## Problem
There is a bug in the validation for dates where if a date field had been input by Respondent 1, and the date violates the validation when it reaches Respondent 2+ (and the fields are disabled), there is still a validation in place that could prevent Respondent 2 onwards from submitting the form

image

Closes FRM-1706

## Solution
To remove the validation for disabled fields for Respondent 2 onwards

**Breaking Changes** 

## Tests

- [ ] Create an MRF form and add a date field with a validation (eg. prevent past dates)
- [ ] Go to Workflow to create multiple workflows
- [ ] Add the date fields for Workflow 1
- [ ] Add more fields for Workflow 2 onwards
- [ ] Share the form and select a future date - submit the form
- [ ] Open the form for Respondent 2
- [ ] Copy the Submission ID, proceed to the DB and search for the submission ID - change the selected date validation to "selectedDateValidation" : "Disallow future dates"
- [ ] Refresh the page for the MRF form for Respondent 2
- [ ] Fill up the remaining fields in the form and submit as Respondent 2 - you should be able to submit the form

### DIFF
--- a/frontend/src/templates/Field/Date/DateField.tsx
+++ b/frontend/src/templates/Field/Date/DateField.tsx
@@ -1,5 +1,6 @@
 import { useCallback, useMemo } from 'react'
 import { Controller, useFormContext } from 'react-hook-form'
+import { useParams } from 'react-router-dom'
 
 import { DATE_DISPLAY_FORMAT, DATE_PARSE_FORMAT } from '~shared/constants/dates'
 import { FormColorTheme } from '~shared/types'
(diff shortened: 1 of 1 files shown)


frontend/src/templates/Field/Date/DateField.tsx
```diff
@@ -32,11 +33,16 @@ export const DateField = ({
...
+  console.log('disable validaton', disableRequiredValidation)
   const validationRules = useMemo(
     () => createDateValidationRules(schema, disableRequiredValidation),
     [schema, disableRequiredValidation],
   )
 
+  console.log('validation rules', validationRules)
+  const { submissionId } = useParams()
+  console.log('submission id', submissionId)
+
```
justynoh: Is this section needed?

justynoh (changes requested):

frontend/src/templates/Field/Date/DateField.tsx
kathleenkhy: omg, sorry! 😔 thanks for pointing this out!  🙏🏼

justynoh (approved):
lgtm

We should open a ticket to investigate this since this behavior is a bit different from other fields 

In other field types, we disable the validation only for required-ness, but if an answer exists it must pass the other field validations. Now for the date field, we are disabling all validation completely. I think this is actually okay for all other fields too, since we have backend validation that the non-editable fields were not modified from the backend. we can consider doing this for all the other field types as well.

PR #7245, opened 2024-04-04
chore: move react-dev-inspector to frontend/package.json

kathleenkhy:
## Problem
`react-dev-inspector` was situated in ./package.json which resulted in local environment facing errors (screenshot below).

Why was react-dev-inspector suspected?
1. It was the code that referenced react highlighted in the stacktrace, and was installed in the wrong package.json. It should be in frontend/package.json.
2. We also added types/react@18 that might have tricked npm to assume that there was a legit react on the root-level node_modules (edited) 

image

## Solution
To move `react-dev-inspector` to frontend/package.json instead. With this, forms should be able to be rendered properly on local.

**Breaking Changes** 

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -49,7 +49,7 @@
     "react": "^17.0.2",
     "react-beautiful-dnd": "^13.1.0",
     "react-csv": "^2.2.2",
-    "react-dev-inspector": "^2.0.1",
+    "react-dev-inspector": "^1.8.6",
     "react-dom": "^17.0.2",
     "react-dropzone": "^11.4.2",
     "react-focus-lock": "^2.7.1",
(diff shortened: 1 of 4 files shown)


frontend/package.json
```diff
@@ -49,6 +49,7 @@
     "react": "^17.0.2",
     "react-beautiful-dnd": "^13.1.0",
     "react-csv": "^2.2.2",
+    "react-dev-inspector": "^2.0.1",
```
KenLSM: The version is also bumped 🤔
Is it compatible with our react? I saw that it was using [`react@18`](https://github.com/zthxxx/react-dev-inspector/blob/dev/packages/inspector/package.json#L121)

KenLSM (approved):

frontend/package.json
kathleenkhy: amended version back to 1.8.6

PR #7247, opened 2024-04-05
fix(mrf): webhook be validation

KenLSM:
## Problem

Admins cannot edit their forms, nor self-help to remove the webhooks after connecting to Plumber

1. MRF doesn’t support webhooks.
2. On FormFE, MRF will not allow admins to set webhooks
3. On FormBE, webhooks setting update can still be done programmatically (i.e., through plumber)

## Solution

Block on BE to prevent users from reaching this state.

Returns `MalformedParametersError` if populating webhook is attempted

**Breaking Changes** 

#### Regression
- [ ] create an MRF
- [ ] run a curl command to update the webhooks[1]

[1] take reference from a webhook update to a storage form and swap out the formid


kathleenkhy (approved):
lgtm

PR #7255, opened 2024-04-16
chore(dockerfile): bump chromium version to work with alpine-3.19

LoneRifle:
## Problem and Solution
Alpine bumped the version of Chromium that works with 3.19, so tweak Dockerfiles to reflect this


KenLSM (approved):
LGTM! Thanks.

KenLSM:
This should be automated, unfortunately didn't have much time to investigate on this.

PR #7259, opened 2024-04-18
chore: allow pip to install packages globally

KenLSM:
## Problem

Podman Desktop is unable to run `pip install awscli-local[ver1]` as it pip doesn't allow packages to be installed globally as per [PEP-668](https://peps.python.org/pep-0668/). However, we're installing the packages in a container, the problem of package ownership not a concern for us.

## Solution

Add `--break-system-packages` to override the error.

**Breaking Changes** 

## Tests

[Skip this test during release. This is a local-dev only, change.]
#### Regression
Ensure that local dev through `docker` is not impacted
- [ ] Run `docker compose up`
- [ ] Ensure that formsg runs

#### New feature test
Ensure that local dev through `podman` works
- [x] Run `podman compose up`
- [x] Ensure that formsg runs

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -43,7 +43,7 @@ RUN apk update && apk upgrade && \
     py-pip && \
     npm install --quiet node-gyp -g && \
     # [ver1] ensures that the underlying AWS CLI version is also installed
-    pip install awscli-local[ver1]
+    pip install awscli-local[ver1] --break-system-packages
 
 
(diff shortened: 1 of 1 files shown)


KenLSM:
Tested on `podman compose up`

kathleenkhy (approved):

KenLSM:
Some localstack services are not working. Moving back to draft

KenLSM:
Might be addressed by #7271

PR #7271, opened 2024-04-22
feat: update localstack docker image

karrui:
Update `localstack` to latest version (but pinned) so the failing SQS noise in local development console is removed.


KenLSM (approved):
LGTM, paired.

PR #7278, opened 2024-04-23
feat: upgrade prettier to v3

karrui:
This PR upgrades prettier versions used in the repo to v3, along with the new formatting changes.

PR #7277, opened 2024-04-23
feat: update eslint packages to stable latest versions

karrui:
ESLint very outdated. Time to fix it. Part of my package upgrading spree. Plan of action:

1. ESLint
2. Prettier (big changes)
3. CRA -> Vite
4. Storybook v8
5. React 17 -> 18 + Chakra UI v1 -> 2 (no choice have to do together) (mega changes)
6. Update typescript versions to v5 across app
7. Use OGPDS with FormSG theme and pray there are no regressions
8. Update react query and rhf to their latest incarnations
9. fin

PR #7279, opened 2024-04-23
fix: fix type issues on react-email templates on express server

karrui:
### TL;DR
Fix type errors on frontend since root types are of a higher version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,7 @@
     "promise-retry": "^2.0.1",
     "promise-timeout": "^1.3.0",
     "puppeteer-core": "22.6.3",
+    "react-email": "^2.1.1",
     "selectize": "0.12.6",
     "slick-carousel": "1.8.1",
     "sns-validator": "^0.3.5",
--- a/package.json
+++ b/package.json
@@ -196,7 +196,6 @@
     "@types/promise-retry": "^1.1.3",
     "@types/promise-timeout": "^1.3.0",
     "@types/puppeteer-core": "^5.4.0",
-    "@types/react": "^18.2.73",
     "@types/sns-validator": "^0.3.1",
     "@types/spark-md5": "^3.0.2",
     "@types/supertest": "^2.0.12",
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -119,7 +119,7 @@
         "@types/gtag.js": "0.0.10",
         "@types/jest": "^26.0.24",
         "@types/node": "^16.11.12",
-        "@types/react": "^17.0.37",
+        "@types/react": "^17.0.80",
         "@types/react-beautiful-dnd": "^13.1.2",
         "@types/react-csv": "^1.1.2",
         "@types/react-dom": "^17.0.11",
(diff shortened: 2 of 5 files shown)


karrui:
* **#7308** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7308?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7305** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7305?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7292** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7292?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7285** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7285?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7279** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7279?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7278** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7278?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7277** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7277?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @karrui and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

package.json
```diff
@@ -196,7 +196,6 @@
     "@types/promise-retry": "^1.1.3",
     "@types/promise-timeout": "^1.3.0",
     "@types/puppeteer-core": "^5.4.0",
-    "@types/react": "^18.2.73",
```
KenLSM, line 199: IIRC this is needed here `src/app/views/templates/MrfWorkflowEmail.tsx`

karrui: Why got tsx in serverrrrrrrr

KenLSM: Email generation. You have ideas to avoid needing the `react` types here, or even removing the need for this to be here?

karrui: new monorepo package with react-email, build via `email export`, then copy the built into the server repo (via package's script) would be my suggestion. I will see if I can do that first then.

karrui: for now i'll sync the types to `/frontend`'s folder so lint works properly

karrui: doesn't work, i'll do the monorepo package thingy first then....

karrui: i think i'm going down the route that @justynoh took lmao. need dynamic data, and a whole pipeline. I'll deprioritise it. 

(doable though, can probably just render as static, and use express's templating engine for the data instead of react-email)

justynoh: ya it's possible to just use the templating engine but every time we make a change to the email format, will need to regenerate the raw template. This was at the time i was trying to play around with different formats and send to GSIB outlook to see how they would be rendered, and I didn't really want to have to keep re-exporting every time i make one small change.

karrui: fixed it by adding the `react-email` package to root, so the types work. (and correctly, since the server is also the one rendering the email, should really have the package)

KenLSM (approved):
LGTMM

PR #7282, opened 2024-04-23
chore: reduce output, check gh login auth status only if not logged in

KenLSM:
## Problem

When releasing:
1.  output is very noisy
2. `gh auth login` executes every time even if we've just logged in

## Solution

Ideally we should use commander or something high-tech for our release scripts...

1. Hide output with `+e`
2. Test for login through `gh auth status` instead

**Breaking Changes** 

### DIFF
--- a/scripts/release_prep.sh
+++ b/scripts/release_prep.sh
@@ -1,12 +1,22 @@
 #!/bin/bash
+set +x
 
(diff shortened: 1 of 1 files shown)


kathleenkhy (approved):
lgtm

PR #7288, opened 2024-04-25
feat(i18n): stub public-facing locales

LoneRifle:
## Problem
As part of efforts to make it easier to localize FormSG, we are prioritising the public-facing elements of the frontend. 

## Solution
- Identify common error messages that a member of the public might see in a form page, and place them in a locale specifically for that.
- Replace hard-coded text in PublicFormProvider and related components
  with i18next equivalents.

#### Dropped 
The following was initially proposed but subsequently dropped:
- Default server error messages to i18next keys, and in
  PublicFormProvider, attempt to use them as such. If they are not,
  use the response messages as-is.

This will be addressed in later development work.

#### Additional work
On request, i18next locales were refactored to be optional, so that
placeholders are no longer needed. en-sg was made mandatory,
since it is a fallback language

- Allow features within Translation locales to be optional
  - Extend Translation to FallbackTranslation, which still makes them
    mandatory
- Declare en-sg as a FallbackTranslation
- Trim zh-sg to only have `login` translations, removing all other
  placeholder text

### Considerations

Even if the frontend and some parts of the backend now refers to our i18next keys, `form.server.model` defaults
the inactive message to a hard-coded string. Replacing the default with an i18next key is awkward, as the value 
will be shown as-is to the form admin, who may be unaware of what we are trying to do.


frontend/src/i18n/locales/features/public-form/zh-sg.ts
karrui, line 1: not accurate right since this should be in mandarin. Maybe we should just not include this file. i18next will fallback to default language if it does not exist anyways

src/app/modules/form/form.errors.ts
```diff
@@ -26,7 +26,7 @@ export class PrivateFormError extends ApplicationError {
    * @param formTitle Extra meta for form title
    */
   constructor(
-    message = 'If you think this is a mistake, please contact the agency that gave you the form link.',
+    message = 'features.publicForm.errors.private',
```
karrui: i think we should not put such strings into the server. We can map the server error messages to a frontend mapping. Change this to a (shared) constant instead then perform frontend mapping to the translated strings?

karrui (dismissed):

frontend/src/i18n/locales/features/public-form/zh-sg.ts
LoneRifle: That would require us to change the declared type of locales to be optional, which wasn't the norm I saw when I first started working on this. @KenLSM will have to make the call here.

src/app/modules/form/form.errors.ts
KenLSM: Agree, on not putting the translation keys into the server.

I'm rather against server providing message to the FE to display them. We should ideally be passing error codes (hoping to get to this soon) instead, and keeping additional messages to be in English for those that are using calling our APIs programmatically. Adding this will likely be introducing patterns that will go against what we want to do in the future (BE passes FE by codes, and FE does the actual translation mapping).

frontend/src/i18n/locales/features/public-form/zh-sg.ts
KenLSM: Favoring making it optional if TS is complaining about it. Reason being if we need to add new common components without translation, it should not stop us (at least for now, as we are not 100% sure if the team is ok with adopting this).

No opinion against adding EN translation in ZH locale as the default, rather it to be explicit (but favoring velocity).

src/app/modules/form/form.errors.ts
LoneRifle: ok, will drop this change now that I've drawn attention to this part of the codebase and wait for @KenLSM or someone else to refactor error mechanism.

frontend/src/i18n/locales/features/public-form/zh-sg.ts
LoneRifle: I've gone with making it optional even though TS accepts mandatory. This should satisfy both @KenLSM and @karrui - since we no longer require the creation of locales other than the fallback one, the team can move ahead without having to create placeholders or wait for translations.

KenLSM (approved):
LGTM!

KenLSM:
Merging `develop` over to this PR to fix the chromatic issue.

PR #7294, opened 2024-04-29
fix: update chromium version

KenLSM:
## Problem

Blocks release build.
Out of sync again. Perhaps we could load from the previous LTS repository instead (i.e., v3.18)

## Solution

Upgrade to latest version.

**Breaking Changes** 


KenLSM:
Bypassed branch protection to prepare for recut

PR #7295, opened 2024-04-29
fix(chromium, puppeteer): old version pin

KenLSM:
## Problem

We have to regularly update our chromium on node:alpine to unblock our builds. This makes our release very flakey.

## Solution

Pin chromium download to an older and less updated version of alpine.
Downgrade `puppeteer-core` to compatible version.

**Breaking Changes** 


KenLSM:
Chromium seems to be stable now. Closing PR

PR #7299, opened 2024-04-30
fix(btn): use different growthbook api

KenLSM:
## Problem

Used incorrect feature toggle endpoint. 

`useIsFeatureEnabled` is our BE Feature flag service, while `useFeatureValue` is our growthbook feature toggle.

## Solution

Switch to use `useFeatureValue`.

**Breaking Changes** 


KenLSM:
Bypassed release requirements -- small change + urgency

PR #7324, opened 2024-05-13
fix(mrf): attachment equality validation

justynoh:
## Problem
Attachment equality validation had to be removed previously due to backward compatibility issues in v6.119.1. This PR restores it.

## Solution

Inject a new `md5Hash` key in the database for attachment answers. 
- The key is populated with the attachment MD5 hash after virus scanning has been completed.
- The incoming attachment's MD5 is compared with the existing attachment's MD5 in order to determine equality.

At the same time, I also made a change to the frontend mutation name that was bothering me. paiseh

**Breaking Changes** 

## Tests

- [x] Create an MRF with two attachment fields. Respondent 1 can edit question 1. Respondent 2 can edit question 2.

**Backward-compatibility tests**

- [x] Submit an MRF with an attachment 1 on the old application version as respondent 1.
- [x] Now, submit the same form with an attachment 2 on the new application version as respondent 2. This should succeed.

**Tests**

- [x] Submit an MRF with an attachment 1 on the new application version as respondent 1.
- [x] Now, submit the same form with an attachment 2 on the new application version as respondent 2. This should succeed.

### DIFF
--- a/frontend/src/features/public-form/mutations.ts
+++ b/frontend/src/features/public-form/mutations.ts
@@ -87,7 +87,7 @@ export const usePublicFormMutations = (
     },
   )
 
-  const useSubmitStorageModeFormMutation = (
+  const useSubmitMutationWithAttachmentVirusScanning = (
     f: (
       args: SubmitStorageFormClearArgs & {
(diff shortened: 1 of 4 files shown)


kathleenkhy (approved):
lgtm!

PR #7325, opened 2024-05-13
fix(mrf): input transformation types for decrypting radio others

justynoh:
## Problem
A user reported an issue decrypting the radio "Others" response on MRF. This PR fixes this.

## Solution

The root cause is usage of `transformInputsToOutputs` on the wrong input type. This PR corrects the type of `transformInputsToOutputs` to unblock this decryption.

**Breaking Changes** 


kathleenkhy (approved):
lgtm!

PR #7341, opened 2024-05-23
chore(mrf): update copy questions -> fields

KenLSM:
## Problem

We use "fields" to refer to form fields, but on the placeholder, we're using "questions"

## Solution

Update to "questions"

**Breaking Changes** 


g-tejas (approved):
Looks good to me!

PR #7349, opened 2024-05-27
chore: regenerate correct package-lock

kathleenkhy:
## Problem

Synk uploaded incorrect package-lock.json, which removes mongodb by marking it peer.

## Solution

Revert back to previous good version and regenerate our own package-lock.json 

**Breaking Changes** 


KenLSM (approved):
LGTM

PR #7350, opened 2024-05-27
feat: email <> storage mode parity

### LINKED ISSUE #2264: Tighten 'emails' and 'publicKey' in form creation Joi validation
The form creation API currently allows the `emails` key to be sent even when creating a storage mode form. This key is meaningless when creating a storage mode form as it is meant to indicate the list of response recipients in email mode. Continuing to include this key for storage mode can lead to unintended bugs (e.g. the Joi validation currently rejects the request if `emails` is an empty string, even when creating a storage mode form).

Similarly, the API currently allows the `publicKey` key to be sent even when creating an email mode form.

The client should be updated to not send these irrelevant keys, and once clients have stopped sending them, the Joi validation can be updated to disallow them.

g-tejas:
## Problem

Users want email notifications for storage mode forms. This PR implements it. 

Closes #2264 

### What's new
**Frontend**
- `SettingsPage.tsx` rendering is refactored. Everything relevant to a tab is coupled under a `TabEntry` with icon, slug, label, etc
-  Storage mode forms now also have email admin inputs in the general tab. Unlike email mode forms, this can be empty.
- Storage mode forms with admin emails cannot enable payments and vice versa. This is because we don't know if the user will complete payment at form submission, which is when we send email notifications. This is okay, as admins + users get email from Stripe anyway. There is appropriate inline messages for notifying user.

**Backend**
- Storage mode forms have empty admin email array by default. This includes duplicating from non-empty email forms. The reason is for this is because of 1. Consistency with `use-template` (security issue) and 2. Consistency with default behaviour of previous storage mode forms
-  Joi validation. When creating encrypt forms, `emails` field is necessary (but it can be empty). All strings in this `emails` array must be valid emails.  Previous behaviour was this invalid email being only detected by database. This gets validated by Joi schema
-  Forbid `emails` tab when duplicating with Joi schema. Duplicated encrypt forms have empty email by default. 

**Tests**
- Mostly modifying existing tests to be consistent with new behaviour.

**Breaking Changes** 

**Features**:

The green box is the new stuff
```mermaid
sequenceDiagram
    participant Client
    participant Controller
    participant Pipeline
    participant AuthService
    participant VerifiedContentService
    participant MailService
    participant EmailSubmissionService
    participant Database

    Client->>Controller: HTTP POST /submitEncryptModeForm
    Controller->>Controller: Extract formId from req.params
    Controller->>Controller: Log metadata
    Controller->>Controller: Retrieve form definition and form
    Controller->>Controller: setFormTags(formDef)
    Controller->>Pipeline: Execute ensurePipeline
    Pipeline->>Controller: Return hasEnsuredAll

    alt hasEnsuredAll is false
        Controller->>Client: Return error response
    else
        Controller->>Controller: Extract encryptedPayload
        Controller->>AuthService: Authenticate User
        AuthService->>Controller: Return uinFin and userInfo

        alt AuthType requires verified content encryption
            Controller->>VerifiedContentService: getVerifiedContent
            VerifiedContentService->>VerifiedContentService: encryptVerifiedContent
            VerifiedContentService->>Controller: Return encryptVerifiedContentResult
        end

        rect rgba(0, 255, 0, 0.2)
        Controller->>EmailSubmissionService: extractEmailAnswers(req.body.responses)
        EmailSubmissionService->>Controller: Return replyToEmails
        Controller->>MailService: sendSubmissionToAdmin(replyToEmails, form, submission, attachments, emailData.formData)
        end

        alt EncryptedPayload has attachments
            Controller->>Database: uploadAttachments(form._id, encryptedPayload.attachments)
            Database->>Controller: Return attachmentUploadResult
        end

        Controller->>Database: Save submissionContent to Database

        alt Form has payments enabled
            Controller->>Controller: Call _createPaymentSubmission
        else
            Controller->>Controller: Call _createSubmission
        end

        Controller->>Client: Return success response
    end

```

## What's left to do?
- [ ] Add the emails field to all existing encrypt mode forms.
```ts
const result = await EncryptFormModel.updateMany(
      { emails: { $exists: false } },
      { $set: { emails: [] } } // Set the default value for emails
    );
```

## Tests
**To test if `use-template` leaks admin emails**
- [ ] 2. Ensure that the newly created form has admin emails and is not empty
- [ ] 3. Make the form public, and navigate to the live form
- [ ] 4. Append "/use-template" to the url 
- [ ] 5. Click on the top right hand corner button "Use this template"
- [ ] 6. Enter new form name, and click storage mode
- [ ] 7. Navigate to the duplicate form general settings
- [ ] 8. Observe that the emails setting is empty
- [ ] 9. Repeat from step 5 onwards for email mode forms. Observe that the admin emails is empty.


frontend/src/features/admin-form/settings/SettingsEmailNotificationsPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { CategoryHeader } from './components/CategoryHeader'
+import { EmailFormSection } from './components/EmailFormSection'
+import {
+  isEmailMode,
+  isEmailOrStorageMode,
+  useAdminFormSettings,
+} from './queries'
+
+export const EmailSection = (): JSX.Element => {
```
KenLSM: For react components, we usually want to avoid exporting multiple components in the same file. Our convention is to use the exported component name as the filename, which you have done. I suppose you've not planned for `<EmailSection />`  to be used outside of `<SettingsEmailNotificationsPage />`?

frontend/src/features/admin-form/settings/SettingsEmailNotificationsPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <Skeleton isLoaded={!isLoadingSettings && !!settings}>
+      <Stack spacing="2rem">
+        {settings && isEmailOrStorageMode(settings) ? (
+          <EmailFormSection
+            emails={settings.emails}
+            isRequired={isEmailMode(settings)}
+          />
+        ) : null}
```
KenLSM: Good practice of using ternaries (`P ? A : null`) instead of short circuiting (`P && A`), which should help in mitigating hard to find bugs coming from `A` not being a renderable data.

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -51,26 +59,50 @@ export const SettingsPage = (): JSX.Element => {
...
-    settingsTabsOrder.indexOf(settingsTab ?? ''),
-  )
+  const displayEmailNotificationPage =
+    settings && isEmailOrStorageMode(settings)
 
   // Note: Admins are not redirected to /general on invalid settings tabs as we
   // don't want to do this prematurely before displayPayments can be determined.
-  useEffect(() => {
+  const tabConfig = useMemo(() => {
+    const baseTabs = [
```
KenLSM: Good abstraction. Keeping related contents close. `label` + `icon` + `component` are v strongly coupled.

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -48,14 +40,32 @@ export const SettingsPage = (): JSX.Element => {
 
   const { ref, onMouseDown } = useDraggable<HTMLDivElement>()
 
-  const [tabIndex, setTabIndex] = useState(
-    settingsTabsOrder.indexOf(settingsTab ?? ''),
+  // Note: Admins are not redirected to /general on invalid settings tabs as we
+  // don't want to do this prematurely before displayPayments can be determined.
+  const tabConfig = useMemo(() => {
+    const baseTabs = [
```
KenLSM: Could the whole `tabConfig` be moved to a top-level constant instead of `<SettingsPage />`? 

1. nothing in `<SettingsPage />` would affect the contents of `tabConfig`
2. It regenerates a new baseTabs whenever `<SettingsPage />` unmounts (not exactly a significant perf issue)

KenLSM: i.e., at the top-level scope + no memo

```
const tabConfig = [
  { label: 'General', icon: BiCog, component: SettingsGeneralPage },
  { label: 'Singpass', icon: BiKey, component: SettingsAuthPage },
  {
    label: 'Twilio credentials',
    icon: BiMessage,
    component: SettingsTwilioPage,
  },
  { label: 'Webhooks', icon: BiCodeBlock, component: SettingsWebhooksPage },
  { label: 'Payments', icon: BiDollar, component: SettingsPaymentsPage },
]
```

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -48,14 +40,32 @@ export const SettingsPage = (): JSX.Element => {
...
+      { label: 'General', icon: BiCog, component: SettingsGeneralPage },
+      { label: 'Singpass', icon: BiKey, component: SettingsAuthPage },
+      {
+        label: 'Twilio credentials',
+        icon: BiMessage,
+        component: SettingsTwilioPage,
+      },
+      { label: 'Webhooks', icon: BiCodeBlock, component: SettingsWebhooksPage },
+      { label: 'Payments', icon: BiDollar, component: SettingsPaymentsPage },
```
KenLSM: Might want to separate the `label` from navigation `path`.

There is no guarantee that the label can be represented as a readable slug, i.e., for labels with multiple words `Email Notification` may be displayed as `email%20notification` in the url

KenLSM: Generally we don't want to add a constrain ourselves that isn't beneficial or futureproof

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
```diff
@@ -40,6 +55,22 @@ export const EmailFormSection = ({
 
   const { mutateFormEmails } = useMutateFormSettings()
 
+  const isFormPublic = useMemo(
+    () => settings.status === FormStatus.Public,
+    [settings],
```
KenLSM: Generally I would personally avoid using `useMemo` for trivial boolean comparisons.

`useMemo` internally does:
1. a comparison: to determine if the internal function should execute)
2. store record of the evaluated value for comparison

A straightforward evaluation of `const isFormPublic = settings.status === FormStatus.Public` only requires a comparison. Since we're also returning booleans, there's no difference if we return a memoized boolean, vs a new `boolean`.

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
```diff
@@ -48,22 +79,47 @@ export const EmailFormSection = ({
...
+      {isFormPublic ? (
+        <InlineMessage marginBottom="16px">
+          To change admin email recipients, close your form to new responses.
+        </InlineMessage>
+      ) : isEmailInputDisabled ? (
+        <InlineMessage variant="error">
+          To enable email notifications, please remove Payments fields from your
+          form.
+        </InlineMessage>
+      ) : settings && settings.responseMode === FormResponseMode.Encrypt ? (
```
KenLSM: /nit `settings?.responseMode === FormResponseMode.Encrypt`

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
```diff
@@ -48,22 +79,47 @@ export const EmailFormSection = ({
...
-  useEffect(() => reset({ emails: initialEmails }), [initialEmails, reset])
+  useEffect(() => reset({ emails: settings.emails }), [settings.emails, reset])
+
+  const emailModeDescription = `Add at least **2 recipients** to prevent loss of response.`
+  const storageModeDescription = `FormSG securely stores responses in an encrypted format and does not retain any associated emails.`
 
   return (
     <>
-      <MRFAdvertisingInfobox />
+      {isFormPublic ? (
```
KenLSM: Would recommend refactoring these ternary into a separate component. Usage of multiple ternaries is a sign of code smell - readability.

Something along the lines of
```
<EmailHeader
  isFormPublic={isFormPublic}
  isEmailInputDisabled={isEmailInputDisabled}
  isFormResponseModeEncrypt={...}
/>

const EmailHeader = ({isFormPublic, isEmailInputDisabled, responseMode) => {
  if(isFormPublic) return ...
  if(isEmailInputDisabled) return ...
  if(isFormResponseModeEncrypt) return ...
  return null
}
```

KenLSM: The render section of a React Component should be kept as clean as possible. Clean here refers to how close it looks like HTML (presentation).

The business logic should be minimized as much as possible in the presentation section.

frontend/src/features/admin-form/settings/SettingsEmailNotificationsPage.tsx
g-tejas: Resolved, as this file doesn't exist anymore (keeping it in general tab for now)

frontend/src/features/admin-form/settings/SettingsPage.tsx
g-tejas: Ahh that's a good point. Previously, the payments (now always visible) and the email notifications (removed) tabs were added dynamically, which I also checked for changes in `displayEmailNotif...`. Would keeping it as a top level constant also be a better choice when we have dynamically shown tabs in the future?

I'll move it to a top-level constant as well

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
g-tejas: Thanks for the catch!

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
g-tejas: Ah I see, that makes alot of sense, what I did is definitely redundant, will remove `useMemo`. Thanks!

I noticed a similar usage here by Kar rui, so I thought there might be some react stuff happening under the hood that I was unaware of. Will change this as well.
https://github.com/opengovsg/FormSG/blob/1e66a8aad7d7700da3b6b080129f87428840b3ea/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsSection.tsx#L67-L70

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
g-tejas: Thanks for the advice! This is much cleaner, got to get a hang of writing good frontend code 😅

Will take note of the point on code smell as well.

frontend/src/features/admin-form/settings/SettingsPage.tsx
g-tejas: That's a good point, will add a separate var for `path` in the `tabConfig`. Thanks!

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
```diff
@@ -45,12 +47,21 @@ const BeforeConnectionInstructions = ({
...
   const { data: paymentGuideLink } = usePaymentGuideLink()
   const [searchParams] = useSearchParams()
+  const { data: settings } = useAdminFormSettings()
 
   const queryParams = Object.fromEntries([...searchParams])
   const isInvalidDomain =
     queryParams[ERROR_QUERY_PARAM_KEY] ===
     DISALLOW_CONNECT_NON_WHITELIST_STRIPE_ACCOUNT
 
+  const isEmailsPresent = useMemo(() => {
```
KenLSM, line 73: Legit use of memo. 👍

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
```diff
@@ -48,22 +79,42 @@ export const EmailFormSection = ({
...
+          isInvalid={!isEmpty(errors)}
+          isDisabled={isFormPublic || isPaymentsEnabled}
+        >
           <FormLabel
-            isRequired
+            isRequired={settings.responseMode === FormResponseMode.Email}
             useMarkdownForDescription
-            description={`Add at least **2 recipients** to prevent loss of response. Learn more on [how to guard against email bounces](${GUIDE_PREVENT_EMAIL_BOUNCE}).`}
+            description={
+              (settings.responseMode === FormResponseMode.Email
```
KenLSM: /nit I think we can also make this to a constant that's evaluated outside of this return

frontend/src/features/workspace/components/CreateFormModal/CreateFormWizardProvider.tsx
```diff
@@ -72,21 +73,25 @@ const useCreateFormWizardContext = (): CreateFormWizardContextReturn => {
...
   const handleCreateStorageModeOrMultirespondentForm = handleSubmit(
     ({ title, responseMode }) => {
-      if (
-        !(
-          responseMode === FormResponseMode.Encrypt ||
-          responseMode === FormResponseMode.Multirespondent
-        )
-      ) {
-        return
+      switch (responseMode) {
```
KenLSM, line 76: Would recommend to have [exhaustive switch for enums](https://www.notion.so/opengov/Use-exhaustive-switch-for-enums-e273c275a684492d8aef4789b4882906?pvs=4)

src/app/models/form.server.model.ts
```diff
@@ -197,7 +197,26 @@ const EncryptedFormSchema = new Schema<IEncryptedFormSchema>({
...
+    set: transformEmails,
+    validate: [
+      (v: string[]) => {
+        if (!Array.isArray(v)) return false
+        if (v.length === 0) return true
+        return v.every((email) => validator.isEmail(email))
+      },
+      'Please provide valid email addresses',
+    ],
+    // Mongoose v5 only checks if the type is an array, not whether the array
```
KenLSM: We're no longer using mongoose v5, could this be removed? Similar question for `EmailFormSchema`

src/types/form.ts
```diff
@@ -286,7 +286,7 @@ export interface IEncryptedForm extends IForm {
   payments_channel: FormPaymentsChannel
   payments_field: FormPaymentsField
   business?: FormBusinessField
-  emails?: never
+  emails: string[] | string
```
KenLSM: I recall that we wanted to make it `string[]` only. Is that still the case?

frontend/src/features/workspace/components/CreateFormModal/CreateFormWizardProvider.tsx
g-tejas: Thanks! This is a neat ts feature, sorta like rust's exhaustive pattern match binding

shared/types/form/form.ts
```diff
@@ -171,6 +171,7 @@ export interface EmailFormBase extends FormBase {
 export interface StorageFormBase extends FormBase {
   responseMode: FormResponseMode.Encrypt
   publicKey: string
+  emails: string[]
```
g-tejas, line 174: must be optional for now, `emails?: string[]`

src/app/models/form.server.model.ts
g-tejas: Updated to say Mongoose v6!

frontend/src/features/admin-form/settings/components/EmailFormSection.tsx
g-tejas: Fixed!

src/types/form.ts
g-tejas: Thanks for the catch!

frontend/src/features/admin-form/settings/SettingsEmailsPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { data: settings } = useAdminFormSettings()
+
+  const isEmailOrStorageMode =
+    settings?.responseMode === FormResponseMode.Email ||
+    settings?.responseMode === FormResponseMode.Encrypt
+
+  if (settings && isEmailOrStorageMode) {
+    return <EmailFormSection settings={settings} />
+  } else {
+    return <EmailFormSectionSkeleton />
```
KenLSM: This looks like we'll render a skeleton if it is not `Email` or `Encrypt` mode. I understand that the `<SettingsEmailPage />` is not rendered if it is neither Email nor Encrypt, but it may not be true if there was a mistake in rendering.

Would prefer that we render `null` if it is not `isEmailOrStorageMode`, and render `<EmailFormSectionSkeleton />` if `settings` is not yet ready.

frontend/src/features/admin-form/settings/SettingsEmailsPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+export const EmailFormSectionSkeleton = (): JSX.Element => {
```
KenLSM: 1. `EmailFormSectionSkeleton` doesn't seem to be used outside of this file. We can omit the export.
2. We usually avoid exporting multiple react components from the same file.

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -46,17 +58,64 @@ export const SettingsPage = (): JSX.Element => {
...
+        path: 'payments',
+      },
+    ]
+
+    if (
+      settings?.responseMode === FormResponseMode.Encrypt ||
+      settings?.responseMode === FormResponseMode.Email
+    ) {
+      // Add the Email Notifications page into the tab list, after singpass
+      baseConfig.splice(2, 0, {
```
KenLSM: Would prefer to avoid updating `baseConfig` by post-initialization it as it may grow to become a frankenstein and prone to breakage as it depends on the index. 

My suggestion without using any explicit index👇 

```typescript
import { isNonEmpty } from '~shared/utils/isNonEmpty'

export const SettingsPage = (): JSX.Element => {
  ...
  const tabConfig = useMemo(() => {
    const emailNotificationTab =
      settings?.responseMode === FormResponseMode.Encrypt ||
      settings?.responseMode === FormResponseMode.Email
        ? {
            label: 'Email notifications',
            icon: BiMailSend,
            component: SettingsEmailsPage,
            path: 'email-notifications',
          }
        : null

    const baseConfig: (TabEntry | null)[] = [
      {
        label: 'General',
        icon: BiCog,
        component: SettingsGeneralPage,
        path: 'general',
      },
      {
        label: 'Singpass',
        icon: BiKey,
        component: SettingsAuthPage,
        path: 'singpass',
      },
      emailNotificationTab, // Add the Email Notifications page into the tab list, after singpass
      {
        label: 'Twilio credentials',
        icon: BiMessage,
        component: SettingsTwilioPage,
        path: 'twilio-credentials',
      },
      {
        label: 'Webhooks',
        icon: BiCodeBlock,
        component: SettingsWebhooksPage,
        path: 'webhooks',
      },
      {
        label: 'Payments',
        icon: BiDollar,
        component: SettingsPaymentsPage,
        path: 'payments',
      },
    ]

    return baseConfig.filter(isNonEmpty)
  }, [settings])
  ...
}
  ```

File: frontend/src/features/admin-form/settings/components/EmailFormSection.tsx [hunk 18]
```diff
@@ -83,24 +137,62 @@ const MRFAdvertisingInfobox = () => {
...
+}: EmailNotificationsHeaderProps) => {
+  if (isFormPublic) {
+    return (
+      <InlineMessage marginBottom="40px">
+        To change admin email recipients, close your form to new responses.
+      </InlineMessage>
+    )
+  }
+
+  if (isPaymentsEnabled) {
```
KenLSM, line 159: Good use of guard clauses!

src/app/modules/form/form.utils.ts
```diff
@@ -100,6 +101,18 @@ export const getCollabEmailsWithPermission = (
   }, [])
 }
 
+export const getAdminEmails = (
+  form: Partial<IFormHasEmailSchema>,
+): string[] => {
+  if (form.emails === undefined) {
```
KenLSM: Would prefer if we check for null + undefined with `form.emails == null`

src/app/modules/form/form.utils.ts
```diff
@@ -100,6 +101,18 @@ export const getCollabEmailsWithPermission = (
...
+    return []
+  } else if (typeof form.emails === 'string') {
+    return [form.emails]
+  } else {
+    return form.emails
+  }
```
KenLSM: /nit completely stylistic to omit usage of else

```suggestion
  } 
  if (typeof form.emails === 'string') {
    return [form.emails]
  }
  return form.emails
```

KenLSM:
> Add the emails field to all existing encrypt mode forms

Can we have to run this before release? I.e., this db schema change is backwards compatible

src/app/modules/form/form.utils.ts
g-tejas: Resolved!

frontend/src/features/admin-form/settings/SettingsPage.tsx
g-tejas: That makes sense! Relying on the index to insert the tab is janky indeed, thanks!

frontend/src/features/admin-form/settings/SettingsEmailsPage.tsx
g-tejas: Gotcha!

frontend/src/features/admin-form/settings/SettingsEmailsPage.tsx
g-tejas: Resolved! Changed it to return `false` if  `!isEmailOrStorageMode` since it seems that it's a better approach than `return null`

> Instead of using return null, a better approach is to use return false. When a component returns false, React treats it as if it had returned null, but it also tells React not to render any of the component's children. 
[From here](https://medium.com/@davidkelley87/stop-using-return-null-in-react-a2ebf08fc9cd)

src/app/modules/form/form.utils.ts
g-tejas: Oh wow, didn't know that the weaker equality `==` accounted for `undefined` and `null` as well. Resolved, thanks!

KenLSM (approved):

PR #7356, opened 2024-06-02
fix(husky): add execution permission to .husky files

KenLSM:
## Problem

Some pre-commit hooks were not able to be executed because they lack execution permission.
```
Running git-secrets...
hint: The '.husky/prepare-commit-msg' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
hint: The '.husky/commit-msg' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

## Solution

Update `.husky/*` files to have [execution](https://github.com/typicode/husky/issues/1177#issuecomment-1212831091) permissions.

**Breaking Changes** 


g-tejas (approved):
LGTM! Missed this, thanks for the catch!

PR #7359, opened 2024-06-03
feat(mockpass): enable admin login using mockpass locally

SeanChinJunKai:
## Problem
Previously, developers would not be able to login with sgId in local

## Solution
This PR creates a new agency called Work Allocation Singapore combined with the sgID support for pocdex.public_officer_details in the latest release of mockpass allows us to login using NRIC S9812379B.

**Breaking Changes** 


KenLSM (approved):
LGTM! Thanks for fixing this!

PR #7086, opened 2024-02-14
feat: allow wildcard domains for domain-restricted email fields

aniruddha-adhikary:
## Problem
Public officers often face the challenge of whitelisting email domains for form submissions, requiring them to manually list each .gov.sg, .edu.sg... domain. This process is not only time-consuming but also prone to errors, as it might exclude newer agency domains unintentionally.

Closes #7084

> Note: 🚨 I may have jumped the gun not waiting for a discussion on #7084.

## Solution
To address this issue, implemented a pattern-based email domain whitelisting feature. This allows form creators to use wildcard patterns such as `@*.gov.sg` and `@*.edu.sg` to automatically whitelist all subdomains under these TLDs or Suffixes.

**Breaking Changes** 
- [ ] Yes - this PR contains breaking changes
- [x] No - this PR is backwards compatible  

**Features**:
- Implemented pattern-based email domain whitelisting to allow using wildcards for domain names.

**Improvements**:
- Updated the email field validation logic to support wildcard domains, ensuring that email addresses matching the specified patterns are accepted.
- Enhanced the user interface to clearly indicate the support for wildcard domains in the email domain whitelist feature.

## Before & After Screenshots

**BEFORE**:
Screenshot 2024-02-14 at 12 20 22

**AFTER**:
(Ah need to shoot one)

PR #7363, opened 2024-06-05
revert: "feat: allow wildcard domains for domain-restricted email fields"

KenLSM:
Reverts opengovsg/FormSG#7086


KenLSM:
Reverting cause FE validation is not present. FE rejects `@nyp.edu.sg` when the a valid pattern is `@*.edu.sg`

g-tejas (approved):
lgtm

PR #7344, opened 2024-05-24
feat(i18n): replace hardcoded text in login features

SeanChinJunKai:
## Problem

This PR replaces most of the hardcoded string in the login features except some strings in SelectProfilePage.tsx. This PR can be used as a reference to translate broken up text as seen in SgidLoginButton.tsx

## Solution

**Breaking Changes** 

- No - this PR is backwards compatible

PR #7365, opened 2024-06-05
External/sean/feat/i18n/login

KenLSM:
PR created to run chromatic for external PR #7344


KenLSM:
Triggered and all chromatic tests passing

PR #7370, opened 2024-06-07
fix: check for truthiness of fieldname in receiver

KenLSM:
Closes FRM-1733

**Breaking Changes** 

###### Form with attachments can be submitted
- [ ] Create a form with attachments
- [ ] Ensure that form can be submitted with attachments


g-tejas (approved):
lgtm!

PR #7410, opened 2024-06-19
chore: expose sgid temp route

KenLSM:
## Problem

We're need to access sgid login on intranet.

## Solution

Exposes additional routes for internal testing that isn't gated. This temporary login shall be removed once the testing is completed.

**Breaking Changes** 


g-tejas (approved):
LGTM!

PR #7418, opened 2024-06-20
fix: email notifications with wrong submission id

g-tejas:
## Problem

For email notifications being sent out by encrypt mode form submissions, the submission ID and the creation date correspond with the form's id and creation date respectively. 

## Solution
Since the submission id and created time is only generated under `_createSubmission`, we have to move the `MailSubmission.sendSubmissionToAdmin` to after these are generated to ensure the created time is consistent with what's shown in the last page.

**Breaking Changes** 

## Tests
- [ ] 2. Add admin emails in the "Email notifications" tab
- [ ] 3. Open the form to public
- [ ] 4. Make a submission and note the submission ID and the created timestamp
- [ ] 5. Navigate to email
- [ ] 6. Observe that the email notification contains the ID and timestamp as what's displayed in the last page.

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -280,41 +280,6 @@ const submitEncryptModeForm = async (
     }
   }
 
-  const logMetaWithSubmission = {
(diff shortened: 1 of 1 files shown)


KenLSM (approved):
LGTM!

PR #7424, opened 2024-06-21
fix: add default values for postman env variables to docker compose

kevin9foong:
## Problem

Due to the new postman service env variables added, local dev is expecting env variables for
`POSTMAN_MOP_CAMPAIGN_ID`,  `POSTMAN_MOP_CAMPAIGN_API_KEY` and `POSTMAN_BASE_URL`. 

however, during local dev env run with `docker compose up`, these variables are not defined and hence cannot be found, leading to var not found error. seen in 'before' screenshot below. 

## Solution

Add default values for these 3 env variables in `docker.compose.yml`

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image
Error message seen when backend is serving requests due to no env variables found. 

**AFTER**:

No error should occur. 

## Tests

- [x] Go to the /dashboard page on local dev, there should not be any error seen in the server logs and page should be served without any error.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -133,9 +133,9 @@ services:
       - WOGAA_START_ENDPOINT
       - WOGAA_SUBMIT_ENDPOINT
       - WOGAA_FEEDBACK_ENDPOINT
-      - POSTMAN_MOP_CAMPAIGN_ID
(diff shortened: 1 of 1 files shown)


g-tejas (approved):
LGTM!

PR #7429, opened 2024-06-23
chore: update dockerfile casing

KenLSM:
## Problem

Received new warning when pushing to `staging`. Doesn't break our builds, but we shouldn't have too many ignored warnings.
```
 1 warning found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 3)
 ```

## Solution

Likely due to github actions deps updating, https://docs.docker.com/reference/build-checks/from-as-casing/

**Breaking Changes** 

### DIFF
--- a/Dockerfile.production
+++ b/Dockerfile.production
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-FROM node:hydrogen-alpine3.19 as build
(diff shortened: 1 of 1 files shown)


g-tejas (approved):
LGTM!

PR #7438, opened 2024-06-25
chore: replace outdated esrv id link

kevin9foong:
## Problem

Currently, for admin to get their eservice id, it says to email askNDI@tech.gov.sg but it should be http://partnersupport.singpass.gov.sg/ instead. 

Closes FRM-1754

## Solution

Update the text to the correct website link.  

**Breaking Changes** 

## Before & After Screenshots

| Topic | Before | After | 
| --- | --- | --- |
| eSrvIdBox help message | image | image | 

### helper text displays message as intended
- [ ] Create form and go to admin auth settings page 
- [ ] Select singpass authtype
- [ ] Check that the new message is as shown as screenshot. (Ie. Singpass support partner...)
- [ ] Repeat for SP with Myinfo & Corppass

### DIFF
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/EsrvcIdBox.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/EsrvcIdBox.tsx
@@ -61,7 +61,7 @@ export const EsrvcIdBox = ({
     return errors.esrvcId ? reset() : onSubmit()
   }, [errors, onSubmit, reset])
 
-  const renderedHelperText = useMemo(() => {
+  const renderHelperTextComponent = () => {
     switch (settings.authType) {
       case FormAuthType.SP:
       case FormAuthType.CP:
@@ -78,7 +78,7 @@ export const EsrvcIdBox = ({
       default:
         return null
     }
-  }, [settings.authType])
+  }
 
   const placeHolder = useMemo(
     () =>
@@ -90,7 +90,7 @@ export const EsrvcIdBox = ({
   return (
     <form onSubmit={onSubmit} onBlur={handleBlur}>
       <Stack ml="2.75rem" mb="1.25rem">
-        {renderedHelperText}
+        {renderHelperTextComponent()}
         <VisuallyHidden>
           <FormLabel htmlFor="esrvcId">e-service ID:</FormLabel>
         </VisuallyHidden>
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/EsrvcIdBox.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/EsrvcIdBox.tsx
@@ -7,7 +7,6 @@ import {
   InputGroup,
   InputRightElement,
   Stack,
-  Text,
   VisuallyHidden,
 } from '@chakra-ui/react'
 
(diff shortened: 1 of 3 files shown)


frontend/src/features/admin-form/settings/components/AuthSettingsSection/EsrvcIdBox.tsx
```diff
@@ -65,9 +66,17 @@ export const EsrvcIdBox = ({
...
       case FormAuthType.MyInfo:
-        return 'Contact askNDI@tech.gov.sg for your e-service ID'
+        return (
+          <Text textStyle="body-2" color="secondary.400">
+            Contact{' '}
+            <Link isExternal href="https://go.gov.sg/formsg-singpass-contact">
+              Singpass partner support
+            </Link>{' '}
+            for your e-Service ID
+          </Text>
```
g-tejas: Is there a reason for memoizing the entire component instead of just the string?

kevin9foong: yeah, makes sense and thanks for the catch. based on react docs, useMemo might reduce the re-rendering of this component if the settings.AuthType is the same https://react.dev/reference/react/useMemo#skipping-re-rendering-of-components

however, since this does not seem to be an expensive component to render, removing useMemo seems to be the right call here

KenLSM: This function now looks like a fully fledged component. May I suggest that we make it a component i.e., `<HelperText authType={authType} />`?

Previously, it was only responsible for determining the string value to render, however, now it had obtained additional responsibility -- to also dictate how the string renders.

frontend/src/features/admin-form/settings/components/AuthSettingsSection/EsrvcIdBox.tsx
```diff
@@ -60,16 +61,24 @@ export const EsrvcIdBox = ({
...
+  const renderHelperTextComponent = () => {
     switch (settings.authType) {
       case FormAuthType.SP:
       case FormAuthType.CP:
       case FormAuthType.MyInfo:
-        return 'Contact askNDI@tech.gov.sg for your e-service ID'
+        return (
+          <Text textStyle="body-2" color="secondary.400">
+            Contact{' '}
+            <Link isExternal href="https://go.gov.sg/formsg-singpass-contact">
```
KenLSM: Can consider moving this over to [shared/constants/links](https://github.com/opengovsg/FormSG/blob/develop/shared/constants/links) for easier lookups for external links -- I do see a couple of string constants littered around the codebase, but if possible, we want to avoid adding more to the litter 🙏

kevin9foong: resolved! thanks for pointing out the existence of ths file, didnt even realise it existed haha

frontend/src/features/admin-form/settings/components/AuthSettingsSection/EsrvcIdBox.tsx
kevin9foong: resolved! i have moved it to its own component

KenLSM (approved):
LGTM!

PR #7442, opened 2024-06-26
chore: add minor changes to sample form submission api

m0nggh:
## Problem

Data from the sample form submission API are not in the correct format.
- Checkbox field, should return the same `fieldOptions` to allow users to use every option possible
- Date field, should return month in `short` format

## Solution

- Checkbox field to return all fields options
- Date field to return months as `short`

**Breaking Changes** 


KenLSM (approved):
LGTM! Thanks!!

m0nggh (commented):
thankss, discussed to only modify date since checkbox options should inherently be randomised

KenLSM (approved):

PR #7446, opened 2024-06-26
fix: resolve eslint warnings on npm builds

g-tejas:
## Problem

This PR resolves the warnings that arise from eslint

**Breaking Changes** 

### DIFF
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -17,7 +17,7 @@
       }
     },
     {
-      "files": ["*.stories.*"],
+      "files": ["*.stories.*", "*.tsx"],
       "rules": {
         "@typescript-eslint/explicit-module-boundary-types": "off"
       }
--- a/frontend/src/app/HashRouterElement.tsx
+++ b/frontend/src/app/HashRouterElement.tsx
@@ -46,10 +46,12 @@ const pathMapper = [
   },
   {
     regex: /^\/forms\/?$/,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     getTarget: (m: FormRegExpMatchArray) => `${DASHBOARD_ROUTE}`,
   },
   {
     regex: /^\/examples\/?$/,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     getTarget: (m: FormRegExpMatchArray) => `/examples`,
   },
 ]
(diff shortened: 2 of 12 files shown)


g-tejas:
* **#7446** <a href="https://app.graphite.dev/github/pr/opengovsg/FormSG/7446?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @g-tejas and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

KenLSM:
**Regarding `no-unsed-vars`**
👍 

**Regarding `explicit-module-boundary-types`**
My personal take: for React Components that are not explicitly returning `JSX.Element`, I feel that it doesn't really add much value to add types to them. 

Usually, for these components, we are already using it as a component that will be rendered, and TS is smart enough to know whether the untyped-function is renderable function.

No qualms on adding them, but also not advocating that we should enforce a return to them.

g-tejas:
> **Regarding `no-unsed-vars`** 👍
> 
> **Regarding `explicit-module-boundary-types`** My personal take: for React Components that are not explicitly returning `JSX.Element`, I feel that it doesn't really add much value to add types to them.
> 
> Usually, for these components, we are already using it as a component that will be rendered, and TS is smart enough to know whether the untyped-function is renderable function.
> 
> No qualms on adding them, but also not advocating that we should enforce a return to them.



> **Regarding `no-unsed-vars`** 👍
> 
> **Regarding `explicit-module-boundary-types`** My personal take: for React Components that are not explicitly returning `JSX.Element`, I feel that it doesn't really add much value to add types to them.
> 
> Usually, for these components, we are already using it as a component that will be rendered, and TS is smart enough to know whether the untyped-function is renderable function.
> 
> No qualms on adding them, but also not advocating that we should enforce a return to them.

@KenLSM  In that case, should we disable `explicit-module-boundary-types` eslint rule for `.tsx` files (since these files mostly consist of frontend components? It seems like the majority of the eslint warnings emitted on builds are just this. Either that or we can disable it per file, but that feels weird. Or should we just keep the warnings?

KenLSM:
Sounds good on excluding `.tsx` files

PR #7448, opened 2024-06-27
feat: single submission per nric

kevin9foong:
## Problem

For some forms, we want to limit the form submissions to a single submission per NRIC. (Eg, each unique voter can only cast 1 vote) 

Closes FRM-1756

## Solution
Implement single limit per nric flow. 

## Storybook test: 
- toggle works 
- error messages for auth settings and payments settings reflects correctly 
- ...and more

## Unit tests: 
- test submitterId are hashed when saved 
- ensure backend validation works, also as a security practice 
- ...and more

**Breaking Changes** 

## Before & After Screenshots
| Subject | Before | After | 
| ------- | ------- | ----- |
| single submission per submitterId toggle | image | image | 
| block user from seeing form if already submitted | NA | image | 
| modal popup if user has accessed form and submitting twice from 2 diff browser, firefox/chrome or incognito/private mode | NA | image | 
| toast error if user uses browser which share same cookie | NA | image | 
 
### TC1: single submission per nric toggle is disabled/locked when payments is connected 
- [ ] create a new storage mode form 
- [ ] connect payments on settings page
- [ ] check that the singpass settings isSingleSubmission toggle is off and disabled and the reason is correctly displayed in the info box below. 

### TC2: payments cannot be connected when single submission per nric toggle is enabled, error message changes when both email notifications + single submission for storage mode 
- [ ] create a new storage mode form 
- [ ] toggle on isSingleSubmission toggle on singpass settings page
- [ ] go to payments settings, check that connect to stripe button is disabled and error message reflects correct reason. 
- [ ] go to email notif settings, add an email
- [ ] go to payments settings again, check still disabled and error message reflects correct reason  
- [ ] turn off isSingleSubmission toggle 
- [ ] go to payments settings again, check still disabled and error message reflects correct reason
- [ ] remove email notification emails 
- [ ] go to payments settings again, check now enabled and no more error message

### TC3: same nric can submit multiple times when isSingleSubmission is toggled off, does not affect other nric, same nric cannot submit once isSingleSubmission is toggled on (Remembers past submitterIds even before toggle is turned on). 
- [ ] create a new storage mode form, open the form to responses, enable singpass auth to any 
- [ ] respond to the form multiple times with the same singpass login credentials. the system should accept multiple form submissions from same singpass cred.
- [ ] go back to form and enable isSingleSubmission toggle. 
- [ ] try and respond using the same singpass account as previously. we should not be able to login to the form and will be stuck at the 'sign in with singpass' page, with an error box stating that duplicate nric submissions are not allowed. 
- [ ] use another singpass credential to login and submit the form, it should be allowed. 
- [ ] repeat the previous steps with email mode form. 

### TC4: if isSingleSubmission is toggled on from the start, same nric can only submit once. 
- [ ] create a storage mode form, choose any singpass auth setting and toggle on isSingleSubmission. 
- [ ] open the form to public and submit the form with singpass.
- [ ] complete the form and click submit another response button at the end. it should redirect back to log in with singpass with no error box anywhere. 
- [ ] log in with the same singpass credentials in previous steps. it should block from entering the form with an error box saying no duplicate submission for given singpass cred. 
- [ ] repeat with email mode form. 

### TC5: if isSingleSubmission is toggled on from the start, different nric can submit 
- [ ] create a storage mode form, choose any singpass auth setting and toggle on isSingleSubmission. 
- [ ] open the form to public and submit the form with singpass.
- [ ] complete the form and click submit another response button at the end. it should redirect back to log in with singpass with no error box anywhere. 
- [ ] log in with another singpass credentials from what is used in previous steps. it should allow submission of the form. 
- [ ] repeat with email mode form. 

### TC6: same nric cannot submit despite both logging in and having race condition, same browser 
 - [ ] create a storage mode form, choose any singpass auth setting and toggle on isSingleSubmission, open form to public. 
- [ ] open respondent form link on 2 separate tabs 
- [ ] log in with singpass on first tab
- [ ] log in with singpass on 2nd tab 
- [ ] both forms should now be logged in and enter the respondent form fields page to fill up form. 
- [ ] try to submit both forms. 
- [ ] 1st submission should be successful with no error 
- [ ] 2nd submission should have a toast message stating that something has failed (this is due to missing the cookie from forced log out) 
- [ ] repeat for email mode form 

### TC7: same nric cannot submit despite both logging in and having race condition, different browser 
 - [ ] create a storage mode form, choose any singpass auth setting and toggle on isSingleSubmission, open form to public. 
- [ ] open respondent form link on 2 separate browsers (eg, 1 chrome and 1 firefox)   
- [ ] log in with singpass on first browser
- [ ] log in with singpass on 2nd browser 
- [ ] both forms should now be logged in and enter the respondent form fields page to fill up form. 
- [ ] try to submit both forms. 
- [ ] 1st submission should be successful with no error 
- [ ] 2nd submission should have a pop up modal saying only one submission nric/fin per nric fin. 
- [ ] close the modal (click x or background) 
- [ ] try and submit the 2nd submission again. the same modal should pop up.  
- [ ] when the modal pops up, click on return to singpass login page action button on the modal. should be redirected to the login with singpass page with no error messages shown anywhere. 
- [ ] try and login with the same singpass account, it should block entrance and display an error box saying no duplicate submission per nric. 
- [ ] repeat for email mode form 

### TC8: when admin turns on isSingleSubmission while prev submitter is resubmitting form 
- [ ] create storage mode form, enable singpass auth 
- [ ] respond to form once with singpass cred
- [ ] submit another response but dont submit yet
- [ ] admin turns off form and then turns on isSingleSubmission toggle then turns form on again 
- [ ] submit the form from step 2, modal saying only one submission per nric should appear

## TC9: duplicate also retains the isSingleSubmission setting
- [ ] create email and storage form with isSingleSubmission on
- [ ] use template to duplicate the form, the duplicated form should also have the same isSingleSubmission setting on 
- [ ] duplicate action on dashboard to duplicate form, the duplicated form should also have the same isSingleSubmission setting on

### DIFF
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
@@ -8,7 +8,7 @@ interface AuthSettingsDisabledExplanationTextProps {
 }
 
 const CONTAINS_MYINFO_FIELDS_DISABLED_EXPLANATION_TEXT =
-  'To change Singpass settings, close your form to new responses. For Singpass authentication changes, remove all existing MyInfo fields.'
+  'To change Singpass settings, close your form to new responses. For Singpass authentication changes, remove all existing Myinfo fields.'
 const FORM_IS_PUBLIC_DISABLED_EXPLANATION_TEXT =
   'To change Singpass settings, close your form to new responses.'
 
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
@@ -8,7 +8,7 @@ interface AuthSettingsDisabledExplanationTextProps {
 }
 
 const CONTAINS_MYINFO_FIELDS_DISABLED_EXPLANATION_TEXT =
-  'To change Singpass settings, close your form to new responses. For Singpass authentication mode changes, additionally remove all existing MyInfo fields. '
+  'To change Singpass settings, close your form to new responses. For Singpass authentication mode changes, additionally remove all existing MyInfo fields.'
 const FORM_IS_PUBLIC_DISABLED_EXPLANATION_TEXT =
   'To change Singpass settings, close your form to new responses.'
 
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
@@ -8,9 +8,9 @@ interface AuthSettingsDisabledExplanationTextProps {
 }
 
 const CONTAINS_MYINFO_FIELDS_DISABLED_EXPLANATION_TEXT =
-  'To change Singpass authentication settings, close your form to new responses and remove all existing Myinfo fields.'
+  'To change Singpass settings, close your form to new responses. For Singpass authentication mode changes, additionally remove all existing MyInfo fields. '
 const FORM_IS_PUBLIC_DISABLED_EXPLANATION_TEXT =
-  'To change Singpass authentication settings, close your form to new responses.'
+  'To change Singpass settings, close your form to new responses.'
 
 export const AuthSettingsDisabledExplanationText = ({
   isFormPublic,
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
@@ -8,7 +8,7 @@ interface AuthSettingsDisabledExplanationTextProps {
 }
 
 const CONTAINS_MYINFO_FIELDS_DISABLED_EXPLANATION_TEXT =
-  'To change Singpass settings, close your form to new responses. For Singpass authentication mode changes, additionally remove all existing MyInfo fields.'
+  'To change Singpass settings, close your form to new responses. For Singpass authentication changes, remove all existing MyInfo fields.'
 const FORM_IS_PUBLIC_DISABLED_EXPLANATION_TEXT =
   'To change Singpass settings, close your form to new responses.'
 
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
@@ -1,0 +1,38 @@
+import { useCallback } from 'react'
+
+import { FormSettings } from '~shared/types/form'
+
+import Toggle from '~components/Toggle'
+
+import { useMutateFormSettings } from '../../mutations'
+
+interface FormSingleSubmissionPerNricToggleProps {
+  settings: FormSettings
+  isDisabled: boolean
+}
+
+export const FormSingleSubmissionToggle = ({
+  settings,
+  isDisabled,
+}: FormSingleSubmissionPerNricToggleProps): JSX.Element => {
+  const { mutateIsSingleSubmission } = useMutateFormSettings()
+  const isSingleSubmission = !!settings?.isSingleSubmission
+  const isLoading = !settings || mutateIsSingleSubmission.isLoading
+
+  const handleToggleIsSingleSubmission = useCallback(() => {
+    if (isLoading) return
+    const nextIsSingleSubmission = !settings.isSingleSubmission
+    return mutateIsSingleSubmission.mutate(nextIsSingleSubmission)
+  }, [settings, isLoading, mutateIsSingleSubmission])
+
+  return (
+    <Toggle
+      containerStyles={{ opacity: isDisabled ? 0.3 : 1 }}
+      isDisabled={isDisabled}
+      isLoading={isLoading}
+      isChecked={isSingleSubmission}
+      label="Limit each unique NRIC/FIN to one response"
+      onChange={handleToggleIsSingleSubmission}
+    />
+  )
+}
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
@@ -44,13 +44,13 @@ export const FormSingleSubmissionToggle = ({
         isDisabled={isDisabled || isPaymentsEnabled}
         isLoading={isLoading}
         isChecked={isSingleSubmission}
-        label="Limit each unique NRIC/FIN to one response"
+        label="Limit each unique NRIC/FIN/UEN to one response"
         onChange={handleToggleIsSingleSubmission}
       />
       {isPaymentsEnabled ? (
         <InlineMessage mt="1rem">
           <Text>
-            To allow only one submission per NRIC/FIN,{' '}
+            To allow only one submission per NRIC/FIN/UEN,{' '}
             <Link as={ReactLink} to={'payments'}>
               disconnect your stripe account
             </Link>
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
@@ -27,8 +27,9 @@ export const FormSingleSubmissionToggle = ({
   const isPaymentsEnabled =
     settings &&
     settings.responseMode === FormResponseMode.Encrypt &&
-    (settings.payments_channel.channel !== PaymentChannel.Unconnected ||
-      settings.payments_field.enabled)
+    ((settings?.payments_channel?.channel &&
+      settings.payments_channel.channel !== PaymentChannel.Unconnected) ||
+      settings?.payments_field?.enabled)
 
   const handleToggleIsSingleSubmission = useCallback(() => {
     if (isLoading) return
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
@@ -27,9 +27,8 @@ export const FormSingleSubmissionToggle = ({
   const isPaymentsEnabled =
     settings &&
     settings.responseMode === FormResponseMode.Encrypt &&
-    ((settings?.payments_channel?.channel &&
-      settings.payments_channel.channel !== PaymentChannel.Unconnected) ||
-      settings?.payments_field?.enabled)
+    (settings.payments_channel.channel !== PaymentChannel.Unconnected ||
+      settings.payments_field.enabled)
 
   const handleToggleIsSingleSubmission = useCallback(() => {
     if (isLoading) return
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
@@ -1,7 +1,12 @@
 import { useCallback } from 'react'
+import { Link as ReactLink } from 'react-router-dom'
+import { Text } from '@chakra-ui/react'
 
-import { FormSettings } from '~shared/types/form'
+import { PaymentChannel } from '~shared/types'
+import { FormResponseMode, FormSettings } from '~shared/types/form'
 
+import InlineMessage from '~components/InlineMessage'
+import Link from '~components/Link'
 import Toggle from '~components/Toggle'
 
 import { useMutateFormSettings } from '../../mutations'
@@ -19,6 +24,12 @@ export const FormSingleSubmissionToggle = ({
   const isSingleSubmission = !!settings?.isSingleSubmission
   const isLoading = !settings || mutateIsSingleSubmission.isLoading
 
+  const isPaymentsEnabled =
+    settings &&
+    settings.responseMode === FormResponseMode.Encrypt &&
+    (settings.payments_channel.channel !== PaymentChannel.Unconnected ||
+      settings.payments_field.enabled)
+
   const handleToggleIsSingleSubmission = useCallback(() => {
     if (isLoading) return
     const nextIsSingleSubmission = !settings.isSingleSubmission
@@ -26,13 +37,26 @@ export const FormSingleSubmissionToggle = ({
   }, [settings, isLoading, mutateIsSingleSubmission])
 
   return (
-    <Toggle
-      containerStyles={{ opacity: isDisabled ? 0.3 : 1 }}
-      isDisabled={isDisabled}
-      isLoading={isLoading}
-      isChecked={isSingleSubmission}
-      label="Limit each unique NRIC/FIN to one response"
-      onChange={handleToggleIsSingleSubmission}
-    />
+    <>
+      <Toggle
+        containerStyles={{ opacity: isDisabled ? 0.3 : 1 }}
+        isDisabled={isDisabled || isPaymentsEnabled}
+        isLoading={isLoading}
+        isChecked={isSingleSubmission}
+        label="Limit each unique NRIC/FIN to one response"
+        onChange={handleToggleIsSingleSubmission}
+      />
+      {isPaymentsEnabled ? (
+        <InlineMessage mt="1rem">
+          <Text>
+            To allow only one submission per NRIC/FIN,{' '}
+            <Link as={ReactLink} to={'payments'}>
+              disconnect your stripe account
+            </Link>
+            . Doing this will also remove all payment fields from your form.
+          </Text>
+        </InlineMessage>
+      ) : null}
+    </>
   )
 }
--- a/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
@@ -63,7 +63,7 @@ const getPaymentsNotAllowedMessageText = (
         To enable payment fields, remove all recipients from{' '}
         <Link as={ReactLink} to={'email-notifications'}>
           email notifications
-        </Link>{' '}
+        </Link>
         .
       </Text>
     )
@@ -73,7 +73,7 @@ const getPaymentsNotAllowedMessageText = (
         To enable payment fields, disable{' '}
         <Link as={ReactLink} to={'singpass'}>
           only one submission per NRIC/FIN
-        </Link>{' '}
+        </Link>
         .
       </Text>
     )
--- a/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
@@ -39,48 +39,44 @@ import {
   StripeConnectButtonStates,
 } from './StripeConnectButton'
 
-const getPaymentsNotAllowedMessageText = (
-  isEmailsPresent: boolean,
-  isSingleSubmission: boolean,
-) => {
-  if (isEmailsPresent && isSingleSubmission) {
-    return (
-      <Text>
-        To enable payment fields, remove all recipients from{' '}
-        <Link as={ReactLink} to={'email-notifications'}>
-          email notifications
-        </Link>{' '}
-        and disable{' '}
-        <Link as={ReactLink} to={'singpass'}>
-          only one submission per NRIC/FIN/UEN
-        </Link>
-        .
-      </Text>
-    )
-  }
-  if (isEmailsPresent) {
-    return (
-      <Text>
-        To enable payment fields, remove all recipients from{' '}
-        <Link as={ReactLink} to={'email-notifications'}>
-          email notifications
-        </Link>
-        .
-      </Text>
-    )
-  }
-  if (isSingleSubmission) {
-    return (
-      <Text>
-        To enable payment fields, disable{' '}
-        <Link as={ReactLink} to={'singpass'}>
-          only one submission per NRIC/FIN/UEN
-        </Link>
-        .
-      </Text>
-    )
-  }
-  return null
+const PaymentsNotAllowedMessageText = ({
+  isEmailsPresent,
+  isSingleSubmission,
+}: {
+  isEmailsPresent: boolean
+  isSingleSubmission: boolean
+}): JSX.Element => {
+  return isEmailsPresent && isSingleSubmission ? (
+    <Text>
+      To enable payment fields, remove all recipients from{' '}
+      <Link as={ReactLink} to={'email-notifications'}>
+        email notifications
+      </Link>{' '}
+      and disable{' '}
+      <Link as={ReactLink} to={'singpass'}>
+        only one submission per NRIC/FIN/UEN
+      </Link>
+      .
+    </Text>
+  ) : isEmailsPresent ? (
+    <Text>
+      To enable payment fields, remove all recipients from{' '}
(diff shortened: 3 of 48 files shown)


frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
```diff
@@ -8,9 +8,9 @@ interface AuthSettingsDisabledExplanationTextProps {
 }
 
 const CONTAINS_MYINFO_FIELDS_DISABLED_EXPLANATION_TEXT =
-  'To change Singpass authentication settings, close your form to new responses and remove all existing Myinfo fields.'
+  'To change Singpass settings, close your form to new responses. For Singpass authentication mode changes, additionally remove all existing MyInfo fields. '
```
KenLSM: Extra space after `.`?

frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { mutateIsSingleSubmission } = useMutateFormSettings()
+  const isSingleSubmission = !!settings?.isSingleSubmission
+  const isLoading = !settings || mutateIsSingleSubmission.isLoading
+
+  const isPaymentsEnabled =
+    settings &&
+    settings.responseMode === FormResponseMode.Encrypt &&
+    ((settings?.payments_channel?.channel &&
+      settings.payments_channel.channel !== PaymentChannel.Unconnected) ||
+      settings?.payments_field?.enabled)
```
KenLSM: Looks like there are quite a few unnecessary optional chaining used.

```suggestion
    (settings.payments_channel.channel !== PaymentChannel.Unconnected ||
      settings.payments_field.enabled)
```

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
```diff
@@ -39,20 +39,47 @@ import {
...
+          email notifications
+        </Link>{' '}
+        and disable{' '}
+        <Link as={ReactLink} to={'singpass'}>
+          only one submission per NRIC/FIN
+        </Link>
+        .
+      </Text>
+    )
+  } else if (isEmailsPresent) {
```
KenLSM: /nit I generally avoid using `else if`s, and this particular block of code is already written in the style of a guard statement. We can be clearer by keeping it as `if`s with `return`s after every case.


```
if (ans % 100) {
  return (
    <>
      <Food />
    </>
  )
}
if (ans % 50) {
  return (
    <>
      <Cats />
    </>
  )
}
if (ans % 5) {
  return (
    <>
      <Dogs />
    </>
  )
}
return null
```

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
```diff
@@ -78,6 +105,15 @@ const BeforeConnectionInstructions = ({
     )
   }, [settings])
 
+  const isSingleSubmission = !!settings?.isSingleSubmission
+
+  const isPaymentsDisabled = isEmailsPresent || isSingleSubmission
+
+  const PaymentsNotAllowedMessageText = getPaymentsNotAllowedMessageText(
```
KenLSM: Typo on casing for `Payments...`?
```suggestion
  const paymentsNotAllowedMessageText = getPaymentsNotAllowedMessageText(
```

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
```diff
@@ -78,6 +105,15 @@ const BeforeConnectionInstructions = ({
     )
   }, [settings])
 
+  const isSingleSubmission = !!settings?.isSingleSubmission
+
+  const isPaymentsDisabled = isEmailsPresent || isSingleSubmission
```
KenLSM, line 114: Good refactor.

frontend/src/features/admin-form/settings/mutations.ts
```diff
@@ -321,6 +322,27 @@ export const useMutateFormSettings = () => {
     },
   )
 
+  const mutateIsSingleSubmission = useMutation(
+    (nextIsSingleSubmissionPerNricEnabled: boolean) => {
+      console.log('mutating')
```
KenLSM: stray `console.log` found!

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -520,13 +544,29 @@ export const PublicFormProvider = ({
         submissionId: string
         timestamp: number
       }) => {
+        if (
+          data &&
+          form.isSingleSubmission &&
+          form.authType != FormAuthType.NIL
```
KenLSM: typo on `!==`?

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -666,6 +706,14 @@ export const PublicFormProvider = ({
                       storePaymentMemory(paymentData.paymentId)
                       return
                     }
+                    if (
+                      data &&
+                      form.isSingleSubmission &&
+                      form.authType != FormAuthType.NIL
```
KenLSM: typo on `!==`?

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -716,12 +764,19 @@ export const PublicFormProvider = ({
...
-
                   if (paymentData) {
                     navigate(getPaymentPageUrl(formId, paymentData.paymentId))
                     storePaymentMemory(paymentData.paymentId)
                     return
                   }
+                  if (
+                    data &&
+                    form.isSingleSubmission &&
+                    form.authType != FormAuthType.NIL
```
KenLSM: typo on `!==`?

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
```diff
@@ -79,9 +86,15 @@ export const PublicFormSubmitButton = ({
...
 
+  const isSingleSubmissionOnlyModalOpen = hasSingleSubmissionValidationError
+  const onSingleSubmissionModalClose = () => {
+    setHasSingleSubmissionValidationError(false)
+  }
+
   return (
     <Stack px={{ base: '1rem', md: 0 }} pt="2.5rem" pb="4rem">
-      {isOpen ? (
+      // Prevent opening of multiple modals
```
KenLSM: This will likely render the comment as text since this is in a render function.

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
```diff
@@ -79,9 +86,15 @@ export const PublicFormSubmitButton = ({
...
+      {isOpen && !isSingleSubmissionOnlyModalOpen ? (
```
KenLSM: This shouldn't be possible iirc. The payment modals should close before triggering the submission API.

frontend/src/features/public-form/components/SingleSubmissionModal/SingleSubmissionModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+}: SingleSubmissionModalProps) => {
+  const isMobile = useIsMobile()
+  const navigate = useNavigate()
+
+  const { handleLogout } = usePublicFormContext()
+
+  const singpassLogoutAndRedirectToFormLogin = () => {
+    if (!handleLogout) return
+    handleLogout()
+    navigate(getPublicFormResponseSingpassLoginUrl(formId))
```
KenLSM: Is the intention here to reload the page? -- we also shouldn't change the url if we are trying to reload, the queryparams would have been lost.

Is the intention here to dismiss the modal? -- we can use `onClose()` to dismiss it.

src/app/models/submission.server.model.ts
```diff
@@ -115,6 +118,51 @@ SubmissionSchema.statics.findFormsWithSubsAbove = function (
...
+ * This method ensures that isSingleSubmission is enforced.
+ * @param submitterId uniquely identifies the submitter
+ * @returns created submission if successful, null otherwise
+ */
+SubmissionSchema.statics.saveIfSubmitterIdIsUnique = async function (
+  formId,
+  submitterId,
+  submissionContent,
+) {
+  const session = await this.startSession()
```
KenLSM, line 132: @kevin9foong after a little more consideration I think we don't have to further optimise this to avoid the read, since it's more costly to `create` than to `read`. Thus there's some inherent benefits of keeping the `read`.

src/app/modules/form/admin-form/admin-form.payments.controller.ts
```diff
@@ -70,6 +73,9 @@ export const handleConnectAccount: ControllerHandler<{
           level: PermissionLevel.Write,
         }),
       )
+      // Payments should not be allowed to connect
+      // since single submission forms do not support it currently.
+      .andThen(assertFormIsSingleSubmissionDisabled)
```
KenLSM, line 78: We'll need to also add a `mapRouteError` for the new error (`UnsupportedSettingsError`) that's thrown, otherwise we'll get 5xx errors when user tries to connect (by manipulating the FE) when ISS is `enabled`.

src/app/modules/form/form.service.ts
```diff
@@ -274,6 +275,50 @@ export const checkFormSubmissionLimitAndDeactivateForm = (
   })
 }
 
+/**
+ * Verify that if the form is a single submission per submitterId form, the submitterId has not responded
```
KenLSM:
```suggestion
 * Verify that if the form is a single submission per submitterId form, the provided submitterId does not exist
```

src/app/modules/form/form.service.ts
```diff
@@ -274,6 +275,50 @@ export const checkFormSubmissionLimitAndDeactivateForm = (
...
+      return errAsync(
+        new ApplicationError(
+          'Cannot find submitterId which is required for single submission per submitterId form',
+        ),
+      )
+    }
+    const submissionWithSameSubmitterIdExists = SubmissionModel.exists({
+      form: form._id,
+      submitterId,
+    }).then((result) => result != null)
```
KenLSM: typo on `!==`? Since result cannot be `undefined`

src/app/modules/form/form.service.ts
```diff
@@ -274,6 +275,50 @@ export const checkFormSubmissionLimitAndDeactivateForm = (
...
+ */
+export const checkHasSingleSubmissionValidationFailure = (
+  form: PublicFormDto,
+  submitterId: string,
+): ResultAsync<boolean, ApplicationError | PossibleDatabaseError> => {
+  const logMeta = {
+    action: 'checkHasSingleSubmissionValidationFailure',
+    formId: form._id,
+  }
+  if (form.isSingleSubmission) {
```
KenLSM: /nit consider swapping it to a guard clause style
/doublenit consider the promise returned by mongoose to be as specific as possible, and allow `neverthrow` to handle the rest of the logic (comparison of non-empty-ness)

```js

  if (!form.isSingleSubmission) return okAsync(false)

  if (!submitterId) {
    return errAsync(
      new ApplicationError(
        'Cannot find submitterId which is required for single submission per submitterId form',
      ),
    )
  }

    SubmissionModel.exists({
      form: form._id,
      submitterId,
    }),
    (error) => {
      logger.error({
        message: 'Error while counting submissions for form',
        meta: logMeta,
        error,
      })
      return transformMongoError(error)
    },
  ).andThen((result) => okAsync(result !== null))
```

src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -294,65 +257,225 @@ export const handleGetPublicForm: ControllerHandler<
...
+          isIntranetUser,
+        })
+      }
+
+      myInfoFields = new SGIDMyInfoData(myInfoFieldsResult.value.data)
+      spcpSession = { userName: myInfoFields.getUinFin() }
+      break
+    }
+    default: {
+      return new UnreachableCaseError(authType)
```
KenLSM, line 321: Mental note to remove use of `ts-essentials` and / or migrate to https://github.com/sindresorhus/type-fest -- both libraries are trying to do something similar.

src/app/modules/submission/email-submission/email-submission.service.ts
```diff
@@ -89,19 +93,45 @@ export const saveSubmissionMetadata = (
...
+              'submitterId must be defined for isSingleSubmission enabled form',
+            ),
+          )
+        }
+        return EmailSubmissionModel.saveIfSubmitterIdIsUnique(
+          form._id,
+          submitterId,
+          params,
+        )
+      } else {
```
KenLSM: /nit don't need the `else`

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -730,6 +768,15 @@ const _createSubmission = async ({
   }
 
   // Send success back to client
+  // clear cookies to log out user if isSingleSubmission form
+  if (
+    form.authType != FormAuthType.NIL &&
```
KenLSM: typo on `!==`

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -730,6 +768,15 @@ const _createSubmission = async ({
   }
 
   // Send success back to client
+  // clear cookies to log out user if isSingleSubmission form
```
KenLSM, line 767: Hm. Would it make sense to clear on every case, not just when it is a `isSingleSubmission` form?

KenLSM: I recall that we want to simplify the cases, and having a login session be ended on every form submission might be simplier.

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -84,6 +85,18 @@ export const checkFormIsEncryptMode = (
     : err(new ResponseModeError(FormResponseMode.Encrypt, form.responseMode))
 }
 
+export const assertFormIsSingleSubmissionDisabled = (
```
KenLSM, line 88: we also have to handle this new error thrown in `mapRouteError()`

src/app/modules/submission/submission.utils.ts
```diff
@@ -758,3 +766,54 @@ export const getAnswersForChild = (
...
+ */
+export const checkIsIndividualSingpassAuthType = (
+  authType: FormAuthType,
+): boolean => {
+  return (
+    authType === FormAuthType.SP ||
+    authType === FormAuthType.MyInfo ||
+    authType === FormAuthType.SGID ||
+    authType === FormAuthType.SGID_MyInfo
+  )
```
KenLSM: Might be terser with this, and it's easier to extend in the future.
```suggestion
  return [
    FormAuthType.SP,
    FormAuthType.MyInfo,
    FormAuthType.SGID,
    FormAuthType.SGID_MyInfo,
  ].includes(authType)
```

src/app/models/submission.server.model.ts
kevin9foong: yeah, great point. the cost to abort a create may be higher than the cost to do a exist read check (aborting after a read should be minimal cost apart from txn overhead)

frontend/src/features/admin-form/settings/components/AuthSettingsSection/AuthSettingsDisabledExplanationText.tsx
kevin9foong: thanks for the catch. resolved!

frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormSingleSubmissionToggle.tsx
kevin9foong: removed!

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
kevin9foong: hi @KenLSM, the rationale why i used caps is because the `getPaymentsNotAllowedMessageText` method returns a jsx component, 

hence im thinking it should be capitalized for any name that references a jsx element

what are your thoughts on this?

frontend/src/features/admin-form/settings/mutations.ts
kevin9foong: thanks for the catch!

frontend/src/features/public-form/PublicFormProvider.tsx
kevin9foong: resolved!

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
kevin9foong: regarding this, i was thinking that it might be good to still have the `isPaymentsModalOpen && !isSingleSubmissionOnlyModalOpen` check, since it would be more defensive. 

any thoughts on this?

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
kevin9foong: strangely, it does not render it on my local dev even though i agree it should. 

i have removed this and refactored to `isPaymentsModalOpen && !isSingleSubmissionOnlyModalOpen` instead so the code is self-explanatory that there shouldnt be multiple modals

src/app/modules/form/form.service.ts
kevin9foong: fixed

src/app/modules/form/form.service.ts
kevin9foong: I was thinking if `!=` might be better to be used here

here i want the expression to evaluate to `false` if the result is `undefined` or `null` 

while there isnt a chance of the result being `undefined`, it might be more defensive to use `!=` as it enforces the above intent. 
image

what are your thoughts on this?

src/app/modules/form/public-form/public-form.controller.ts
kevin9foong: got it, i will not introduce any new inclusions to `ts-essentials` in the future, but will not change this for now to avoid unrelated scope for the PR.

src/app/modules/submission/submission.utils.ts
kevin9foong: yeah definitely! will change to this much cleaner code!

kevin9foong: on second look, seems ive removed this function since corppass is also now supported for single submission. will keep this `includes` function in mind in the future though

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
kevin9foong: yeah, this works! fixed!

src/app/modules/form/form.service.ts
kevin9foong: fixed and agreed this is much cleaner!

frontend/src/features/public-form/components/SingleSubmissionModal/SingleSubmissionModal.tsx
kevin9foong: no, the intention is to reload the page instead. 

this function is invoked when the user clicks "Back to Singpass Login" button after the user gets a nric is already used modal. 

so in this case, we should redirect the user to the singpass page from scratch instead of dismissing the modal.

src/app/modules/form/admin-form/admin-form.payments.controller.ts
kevin9foong: thanks for the catch!

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
kevin9foong: hi @KenLSM, i am not entirely sure if we want this behavior to force logout the user each time even for non `isSingleSubmission` forms. 

Previously, the rationale for `MyInfo` force clearing makes sense since the user needs to approve permission each time to prefill form data. 

however, for other modes of Singpass, it might be useful for the user to stay logged in to submit multiple submissions if possible. 

The second alternative way to look at this is that the same user should never need to submit the same form more than once (not always true but not always false either) 

Since the discussion on which of the above 2 is ideal is out of scope of my current nric task, i did not change the existing clearing cookie behavior. 

but if we are sure we want option 2, then i suppose i can clear login cookies for all cases. 

please let me know what your thoughts on this are

frontend/src/features/public-form/components/SingleSubmissionModal/SingleSubmissionModal.tsx
kevin9foong: im not sure if we want to keep the query params, but if we do, i can explore a way to bring along the query params (which im suspecting is used for form prefills) when navigating the route

kevin9foong: though this might not be necessary since this same user has already filled in the form according to this error message

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
KenLSM: I think we've decided to go with option 2, which is to always clear to simplify submissions for now. Makes it easier to rationalise when explaining to admins / users.

There's a counter point that can be made as well for preventing logins, where the form is displayed on a public kiosk (i.e., ipad). For that case, we should always log the respondent out, rather than to keep them logged in. Both ways seemed have sufficient usecase, but for the sake of Product Operations, I think we'll lean on the simplicity of explanation.

We do have something in mind that the team might explore, which is the citizen portal where we can properly tackle problems like this more cleanly.

KenLSM: cmiiw, also thinking that it would make your task easier if we include it in this scope, as it would mean that we get to delete code. Otherwise, agree to leave it out for a subsequent PR to sweep it out.

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
KenLSM: I think we should not need to put defensive code here, which might be swallowing mistakes rather than surfacing them up. i.e., someone made a mistake of not dismissing their modal, but had no idea that this is happening due to this piece of code "helping to hide" the mistake.

frontend/src/features/public-form/components/SingleSubmissionModal/SingleSubmissionModal.tsx
KenLSM: Understood. Sounds like we want to have something like a reload, and also agree that this specific usecase likely shouldn't retain the queryparams either.

Could I suggest naming it to be more generic such as `getPublicFormUrl(formId)`? This way, we can reuse the url builder in other situations.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
kevin9foong: deferring to future PR as discussed

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
kevin9foong: fair enough, i will remove this check

frontend/src/features/public-form/components/SingleSubmissionModal/SingleSubmissionModal.tsx
kevin9foong: renamed

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
kevin9foong: fixed by refactoring as own component and using `isPaymentsDisabled` instead of non-nullness of component as check of whether to render.

KenLSM (approved):
LGTM

PR #7457, opened 2024-06-28
feat: add model fields to represent translations for form fields

siddarth2824:
## Context
- Add fields in the model layer to represent the translations for the form fields and which languages are supported for the form.
- This PR will be the first PR to introduce the additional model fields to support for the multi translation feature.

## Changes
- Added new type called `TranslationMapping` which represents the translations for the specific language
- Added `titleTranslations` and `descriptionTranslations` for`baseField` type
- Added `paragraphTranslations` for `startPage` type
- Added `fieldOptionsTranslations` for `CheckboxFieldSchema`, `RadioFieldSchema` and `DropdownFieldSchema`

### DIFF
--- a/shared/types/form/form.ts
+++ b/shared/types/form/form.ts
@@ -74,10 +74,10 @@ export enum FormAuthType {
 }
 
 export enum Language {
-  ENGLISH = 'English',
-  CHINESE = 'Chinese',
-  MALAY = 'Malay',
-  TAMIL = 'Tamil',
+  ENGLISH = 'en-SG',
+  CHINESE = 'zh-SG',
+  MALAY = 'ms-SG',
+  TAMIL = 'ta-SG',
 }
 
 export enum FormStatus {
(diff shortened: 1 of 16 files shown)


shared/types/form/form.ts
```diff
@@ -60,12 +73,24 @@ export enum FormAuthType {
   SGID_MyInfo = 'SGID_MyInfo',
 }
 
+export enum Language {
+  ENGLISH = 'English',
```
KenLSM: Recommend to use [unicode locales](https://www.unicode.org/reports/tr35/#Unicode_locale_identifier) for the values since we might be supporting localization for other countries.
```
ENGLISH = 'en-SG',
CHINESE = 'zh-SG',
  MALAY = 'ms-SG',
  TAMIL = 'ta-SG',
```

^ opting to stick to the simpler ["ISO 639 two-letter language codes and ISO 3166 two-letter country codes"](https://en.wikipedia.org/wiki/IETF_language_tag).

KenLSM (approved):
LGTM!

PR #7472, opened 2024-07-04
chore: remove nric mask toggle if nricmask is false

kevin9foong:
## Solution
Hide Nric toggle if admin has not enabled the toggle for their form. 

**Breaking Changes** 


KenLSM (approved):
LGTM! Thanks for the quick turnaround @kevin9foong

PR #7475, opened 2024-07-04
feat: add copy for link to response page

g-tejas:
## Problem
Currrently, email mode notifications for encrypt forms do not have attachments as we plan on increasing the attachment size limit and other reasons. This PR makes it easy for admins to go from their email notifications to the settings page for downloading the attachments. 

Screenshot 2024-07-07 at 5 23 02 PM

**Breaking Changes** 

## Tests
**Regression Test:** Email mode form notifications must work exactly the same
- [ ] 1. Create a new email mode form
- [ ] 2. Make form public and create a submission
- [ ] 3. Observe that the email notification arrives in inbox, and has the same format as before. 
- [ ] 4. Observe that the email notification has attachments attached. 

**New feature**
- [ ] 1. Create a new storage mode form
- [ ] 2. Add your email as the admin
- [ ] 3. Make form public and create a submission
- [ ] 4. Observe that email notification arrives in inbox
- [ ] 4. Observe that there's a link to the admin settings page, and the response Id and time should match exactly
- [ ] 5. Observe that there are no attachments.

### DIFF
--- a/src/app/views/templates/submit-form-email.server.view.html
+++ b/src/app/views/templates/submit-form-email.server.view.html
@@ -4,9 +4,12 @@
   <body>
     <p>Hello there,</p>
     <p>A "<%= formTitle %>" form was submitted on <%= submissionTime %>.</p>
-    <br />
-    <p> View this submission and download attachments on FormSG </p>
-    <br />
+    <% if (responseMode === 'encrypt') { %>
+      <p>View this submission and download attachments on <a href="http://localhost:3000/admin/form/<%= formId %>/results/<%= refNo %>">FormSG</a></p>
+    <% } else { %>
+      <br />
+    <% } %>
+    <br/>
     <table
       id="formsg"
       width="100%"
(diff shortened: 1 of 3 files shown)


src/app/views/templates/submit-form-email.server.view.html
```diff
@@ -4,7 +4,12 @@
   <body>
     <p>Hello there,</p>
     <p>A "<%= formTitle %>" form was submitted on <%= submissionTime %>.</p>
-    <br /><br /><br />
+    <% if (responseMode === 'encrypt') { %>
+      <p>View this submission and download attachments on <a href="http://localhost:3000/admin/form/<%= formId %>/results/<%= refNo %>">FormSG</a></p>
```
KenLSM: Need to be careful of side effects when linking an authenticated url. We might need to exclude certain url groups from DD to avoid getting false positive alerts / betteruptime calls.

PR #7480, opened 2024-07-05
chore(deps): remove sentry

KenLSM:
## Problem

Sentry is no longer used by us. It was a legacy from Angular which was not removed after the react migration.

Closes FRM-411

## Solution

**Breaking Changes** 

## Tests

Regression for Admin
- [ ] Create a form
- [ ] Add a short text field
- [ ] Open the form
- [ ] Ensure that the form can be viewed by the respondent

Regression for Respondent
- [ ] Ensure that a form can be submitted

## Deploy Notes

**New dependencies**:

- `@sentry/browser`: removed
- `@sentry/integrations`: removed


KenLSM:
Was prompted by #7479 which spams us with false positives cause we have not removed sentry from our codebase.

g-tejas (approved):
lgtm!

PR #7481, opened 2024-07-05
fix(reddot): hide on mrf

KenLSM:
## Problem

Reddot shows on MRF even if there's no Email Notification configuration available.

## Solution

Check on navbar if we're on an MRF form, and hide it.

**Breaking Changes** 


g-tejas (approved):
LGTM!

PR #7499, opened 2024-07-11
fix: add the code to error tag in datadog traces

g-tejas:
## Problem
The error message for traces in datadog only has the error msg and can be difficult to deduce the error code without going to `core.errors.ts` manually to check. The `@error.type` flag also doesn't appear on the trace explorer for some reason, only the message.

## Solution
This prefixes the error message with the error code to make filtering traces easier.

**Breaking Changes** 


g-tejas:
@kevin9foong I've opened a ticket with respect to the `@error.type` not showing up on the traces. You can track the ticket [here](https://help.datadoghq.com/hc/en-us/requests/1769696)

KenLSM (approved):
LGTM

PR #7523, opened 2024-07-16
fix: enable pdf attachment for encrypt mode forms

g-tejas:
## Problem

Closes FRM-1776

## Solution
We check for `isVisible !== false` because 
1. `isVisible === true` is in email mode forms when
2. `isVisible === undefined` is in encrypt mode forms, because the field gets stripped out in the [middleware](https://github.com/opengovsg/FormSG/blob/2d68a62533f1a502c72f3cb4a9e3af12f14d3e46/src/app/modules/submission/encrypt-submission/encrypt-submission.utils.ts#L131) as mentioned in this [PR](https://github.com/opengovsg/FormSG/pull/6907). 
Hence we should allow the fields to show in the auto reply to account for these two cases

We add a new field `unencryptedAttachments` to `req.formsg` in the middleware. Here's why,
> Problem: Encrypt form attachments are stripped out in middleware, autoreply is sent at the end of the controller
**Option 1**: :no_entry: Move the autoreply sending to middleware. Issue w this is that if we move the autoreply sending to the last-est part where we do have access to attachments, the request can still fail somewhere down the pipeline and this email would have been prematurely sent to users. We need to send email out only when the submission is saved, which is after its saved to the db. Hence this option doesnt work
**Option 2**: Pass the unencrypted attachments down the middleware chain to the last controller. This allows us to keep autoreplies where they are, so they arent sent prematurely, and users also get. their attachments in the response.

Thus before we strip out the attachments in `encryptSubmission` middleware, we store it in the request, so that we can retrieve it at the end of the controller, to send with the user auto replies.

**Breaking Changes** 

## Tests
**DB Test to ensure formsg.unencryptedAttachments is not saved in DB**
- [ ] 2. Add an attachment field and an email field
- [ ] 3. Ensure that the email field has email confirmations enabled and PDF responses also enabled
- [ ] 4. and and open the form to public
- [ ] 5. Make a submission with an attachment
- [ ] 6. Obtain the form ID and make a query on the appropriate DB: `{form: ObjectId('669735b40479eaf8b6627e48')}`
- [ ] 7. Ensure that the DB doesn't contain the unencrypted attachments

**Duplicate Test form**
- [ ] 2. Add an email field and enable email confirmations
- [ ] 3. Enable PDF response as well
- [ ] 4. Duplicate the Form to an MRF form
- [ ] 5. Ensure that the email field still exists
- [ ] 6. Ensure that the email confirmation toggle is turned off and observe that the toggle is also locked.
- [ ] 7. Open the form to public
- [ ] 8. Create a submission
- [ ] 9. Observe that the submitter email doesn't receive any email at all
- [ ] 10. Repeat for email mode forms

**Regression Tests to ensure that PDF responses works for Email mode**
- [ ] 1. Create an email mode form
- [ ] 2. Add an email field and an attachment field. Add other fields as well
- [ ] 3. Enable the email confirmation for the email field and PDF responses as well.
- [ ] 4. Make form public
- [ ] 5. Make a submission with an attachment
- [ ] 6. Observe that the submitter email receives an email confirmation 
- [ ] 7. Observe that the email contains both the PDF responses + the attachments
- [ ] 8. Observe that the attachments is equivalent to the submitted attachment
- [ ] 9. Observe that the PDF response  contains all the fields in the form, and correspond to your inputs

**Ensure that payment form with NO active payment field CAN enable PDF responses**
- [ ] 1. Create encrypt mode form
- [ ] 2. Connect a stripe account
- [ ] 3. Add an email field
- [ ] 4. Enable email confirmations
- [ ] 5. Observe that the toggle for pdf responses is not locked and can be toggled on

**Unusual path: Ensure form with active payment field AND includeFormSummary enabled doesn't get response.pdf**
- [ ] 1. Create an encrypt form
- [ ] 2. Add an email field and an attachment field
- [ ] 3. For the email field, Enable email confirmations and also enable pdf response toggle
- [ ] 4. Connect a stripe account
- [ ] 5. Add a payment field
- [ ] 6. Open form to public and make a submission. Make sure the. payment goes through.
- [ ] 7. Observe that the submitter email receives an email confirmation
- [ ] 8. Observe that the email doesn't contain any attachments and also no response.pdf. 

**Ensure that payment form with active payment field CANNOT enable PDF responses**
- [ ] 1. Create encrypt mode form
- [ ] 2. Connect a stripe account
- [ ] 3. Add an email field
- [ ] 4. Enable email confirmations
- [ ] 5. Observe that the toggle for the pdf response is locked and not toggled on.

**Regression test to ensure that MRF forms cannot enable email confirmations**
- [ ] 1. Create an MRF form
- [ ] 2. Add an email field
- [ ] 3. Observe that the email confirmation toggle is locked

**Regression tests to ensure that email confirmation works for Email mode**
- [ ] 1. Create an email mode form
- [ ] 2. Add an email field and enable email confirmations. Add an attachment field as well
- [ ] 3. Observe that the toggle for pdf response is not locked
- [ ] 4. Enable pdf response
- [ ] 5. Make form visible to public
- [ ] 6. Make a submission
- [ ] 7. Observe that the submitter email received the submission confirmation
- [ ] 8. Observe that the email has PDF attachment and the other attachments as well.

**Regression tests to ensure that admin email notification still works for storage mode forms**
- [ ] 2. Create a few input fields
- [ ] 3. Add admin emails
- [ ] 4. Make the form public and make a submission
- [ ] 5. Observe that the admin email receives the email notification.
- [ ] 6. Observe that the table contains all the visible fields
- [ ] 7. Observe that the response ID, time of  submission are the same as the formsg submission page
- [ ] 8. Observe that the json payload doesn't exist at the end of the email


KenLSM (approved):
LGTM!

PR #7539, opened 2024-07-18
chore: add simple fix to date format for sample form submission api

m0nggh:
## Problem
Data from the sample form submission API are not in the correct format.

Date field should return the day in 2-digit format to standardise with formSG date format.

## Solution
Date field to return day as '2-digit'

**Breaking Changes** 


KenLSM (approved):
Lgtm

m0nggh (commented):
thanks!

PR #7542, opened 2024-07-23
chore(ci): pipe test coverage results to datadog

karrui:
Adding CI step to pipe test results to datadog


src/app/modules/datadog/__mocks__/datadog.utils.ts
karrui, line 1: don't actually knoow whether this is used, but put here ba haha

KenLSM (approved):
LGTM

PR #7285, opened 2024-04-24
feat: migrate from CRA4 to Vite, Storybook to v8, jest to Vitest

karrui:
This PR moves the frontend toolchain from CRA4 to Vite. 
Generally everything seems to be working fine, including decryption workers.

Requires a bunch of changes that has to be performed all at the same time:
- [x] Jest -> Vitest
- [x] Storybook v6 -> v8 (less painful than wrangling Storybook v6 to work with vite)

# TODO:
- [x] move back dd-chunk pipeline to vite build
- [ ] Test on GSIB

PR #7546, opened 2024-07-25
chore(deps): update date-fns, date-fns-tz

KenLSM:
## Problem

Closes #7545, #7145

## Solution

Upgrades `date-fns` to latest version, as well as updating it's corresponding downstream deps (`date-fns-tz`)

Replac `zonedTimeToUtc` to `fromZonedTime` as it is a [breaking change](https://github.com/marnusw/date-fns-tz/blob/master/CHANGELOG.md#v300-6-april-2024) listed on `date-fns-tz`

**Breaking Changes** 


KenLSM:
Chromatic is failing as the date-fns lib introduced nullish-coalescing operator into the mix. Generally most transpiler and browsers are able to handle it, but our current version of bundler (i.e., Webpack 4) doesn't support it.

There's another set of PR that migrates our bundler over. As such, we will resume this dependency upgrade after we've shipped #7285

kevin9foong (approved):
lgtm!

PR #7550, opened 2024-07-30
fix: mock statsD client during dev/test env

kevin9foong:
## Problem

Closes FRM-1800

When running tests locally (esp during verbose mode), the jest logs are often spammed with error messages due to the above. Adding to the noise and making it hard to debug test cases.  

This is a acknowledged issue and occurs all the time when backend unit tests are run. 

See example of noise added to test debug logs: 
image

## Solution

The hotshot client should be run as a mock during dev/test mode as logging to DD is not desired during test/dev.
( see hot-shot mock: https://github.com/brightcove/hot-shots )

Doing so removes the error messages above


KenLSM (approved):
LGTM! Thanks for keeping our tests clean!

PR #7561, opened 2024-08-01
feat: add settings tab and toggle for multi lang feature

siddarth2824:
## Context
- Add settings tab and toggle to allow users to turn on the multi-language translation feature for their form.
- This PR is the second PR to introduce the settings tab and toggle view to allow users to turn on and off the multi-language translation feature.
- This PR does not introduce any changes to the public form and only introduces the functionality to turn on/off support for multi-language translation.

## Changes
- Add [views](https://www.figma.com/design/wNcuzzF3ciL8f6xYNvGZgK/Form-Exploration?node-id=15273-13197&t=hvNPQdzJyBb5NYwK-0) for settings tab and toggle.
- Add mutations to update the following form's setting field:
   - Set boolean value for `hasMultiLang` to indicate if a form supports the multiple languages
   - Set the languages, in `supportedLanguages` field,  that the a form supports

## Before & After Screenshots
| ----------- | ------- | ----- |
| Addition of settings tab on the side bar and toggle for turning on/off support for multi-language translation feature| |Screenshot 2024-08-12 at 10 04 02 PM|
| Language support list when toggle is turned on and success toast| |Screenshot 2024-08-12 at 10 06 08 PM|
| Toast when support for multi-translation is turned off| |Screenshot 2024-08-12 at 10 20 23 PM|
| Tooltip when hovering over icon to able/disable language| |Screenshot 2024-08-12 at 10 24 19 PM|
| Tooltip when hovering over icon to edit translation| |Screenshot 2024-08-12 at 10 21 05 PM|
| Toast when disabling support for a specific language| |Screenshot 2024-08-12 at 10 21 29 PM|
| Toast when enabling support for a specific language| |Screenshot 2024-08-12 at 10 21 48 PM|

## Tests
TC 1: Turning on toggle adds support for default supported languages and sets the `hasMultiLang` attribute to be true
1. Turn on the toggle and check the forms table to verify that the `supportedLanguages` field is populated with the default supported languages (`[en-SG, zh-SG, ms-SG, ta-SG]`) and the `hasMultiLang` value is true

TC 2: Disable and then enable support for a specific language
1. Turn on toggle to add support for the multi-language translation feature
2. Disable "Chinese"
3. Verify in the forms table for the particular form that the `supportedLanguages` field now contains `[en-SG, ms-SG, ta-SG]`
4. Then enable "Chinese" language
5. Verify in the forms table for the particular form that the `supportedLanguages` field now contains `[en-SG, ms-SG, ta-SG]`


frontend/src/features/admin-form/settings/SettingsMultiLangPage.tsx
```diff
@@ -0,0 +1,9 @@
+import { MultiLanguageSection } from './components/MultiLanguageSection/MultiLanguageSection'
+
+export const SettingsMultiLangPage = (): JSX.Element => {
+  return (
+    <>
```
kevin9foong: we might not need a react fragment around here, since we are not returning multiple elements

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+export const FormMultiLanguageToggle = (): JSX.Element => {
+  const { data: settings, isLoading: isLoadingSettings } =
+    useAdminFormSettings()
+
+  const { mutateFormSupportedLanguages, mutateFormHasMultiLang } =
+    useMutateFormSettings()
+
+  const hasMultiLang = useMemo(
+    () => settings && settings?.hasMultiLang,
```
kevin9foong: we might not need to `useMemo` here
- `useMemo` has a cost which is to store in and lookup the memo table 

might be more efficient and simpler to do the boolean check directly

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const [isLanguageSupported, setIsLanguageSupported] = useState(
+    supportedLanguages?.indexOf(unicodeLocale) !== -1,
+  )
+
+  const languageToDisplay = convertUnicodeLocaleToLanguage(unicodeLocale)
+
+  const { mutateFormSupportedLanguages } = useMutateFormSettings()
+
+  const handleToggleSupportedLanague = useCallback(
```
kevin9foong: spelling error 
```suggestion
  const handleToggleSupportedLanguage = useCallback(
```

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+    (language: Language) => {
+      if (supportedLanguages == null) return
+
+      // Remove support for this language if it exists in supportedLanguages
```
kevin9foong: Suggestion: instead of using comments to describe `what` the code is doing, the code should probably be able to explain itself (i prefer reserving comments to describe `why` something is done)

The following code block (to line 76) could be replaced with something like this: 

```suggestion
      const isLanguageSupportedPreviously = !!supportedLanguages.find(language)
      const isAddSupport = !isLanguageSupportedPreviously
      const updatedSupportedLanguages = isAddSupport 
        ? supportedLanguage.concat(language)
        : supportedLanguages.filter(l => l !== language)
        
      setIsLanguageSupported(isAddSupport)
      
      return mutateFormSupportedLanguages.mutate({
          nextSupportedLanguages: updatedSupportedLanguages,
          selectedLanguage: language,
        })
```

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+              icon={<BiEditAlt width="44px" />}
+              colorScheme="secondary"
+              aria-label={`Add ${unicodeLocale} translations`}
+              // TODO: Will add redirection to translation section in next PR
+              onClick={() => console.log('Edit translations')}
+            />
+          </HStack>
+        )}
+      </Flex>
+      {!isLast && <Divider />}
```
kevin9foong: formsg has a convention to use 
```{!!isLast ? <Divider /> : null}``` instead of using `&&`

File: frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx [hunk 6]
```diff
@@ -0,0 +1,202 @@
...
+
+  return (
+    <Skeleton isLoaded={!isLoadingSettings && !!settings} mt="2rem">
+      <Toggle
+        onChange={handleToggleMultiLang}
+        isChecked={hasMultiLang}
+        label="Enable multi-language"
+        description="This will allow respondents to select a language they prefer to view your form in. Translations are not automated."
+      />
+      {settings && hasMultiLang && (
```
kevin9foong: same for the convention of doing `{settings && hasMultiLang ? ... : null}` instead of nullish coalesce operator

frontend/src/features/admin-form/settings/mutations.ts
```diff
@@ -138,6 +142,54 @@ export const useMutateFormSettings = () => {
...
+    {
+      onSuccess: (newData, newSupportedLanguages) => {
+        if (newSupportedLanguages && newSupportedLanguages.selectedLanguage) {
+          const supportedLanguages =
+            newSupportedLanguages.nextSupportedLanguages ?? []
+          const languageToDisplay = convertUnicodeLocaleToLanguage(
+            newSupportedLanguages.selectedLanguage,
+          )
+
+          let successMessage: string
```
kevin9foong: nit: prefer using ternary operator with a const successMessage instead of let

the rationale is that using const as much as possible reduces the likelihood of undesired mutations in the code. 

```suggestion
          const isSelectedLanguageSupported = supportedLanguages.includes(newSupportedLanguages.selectedLanguage)
          const successMessage = isSelectedLanguageSupported ? <str1> : <str2>; 
```

frontend/src/utils/multiLanguage.ts
```diff
@@ -0,0 +1,17 @@
...
+  switch (language) {
+    case Language.ENGLISH:
+      return 'English'
+    case Language.CHINESE:
+      return 'Chinese'
+    case Language.MALAY:
+      return 'Malay'
+    case Language.TAMIL:
+      return 'Tamil'
+    // This case should not occur
```
kevin9foong: comment might not be needed, since default throws an error which is expected not to occur.

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+    [settings],
+  )
+
+  const handleToggleMultiLang = useCallback(() => {
+    if (!settings || isLoadingSettings || mutateFormHasMultiLang.isLoading)
+      return
+
+    const nextHasMultiLang = !settings.hasMultiLang
+
+    const currentSupportedLanguages = settings?.supportedLanguages
```
kevin9foong: optional chaining might not be needed due to guard clause on line `161`

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+    // Add all languages as this is the first instance user turns on
+    // toggle
+    if (_.isEmpty(currentSupportedLanguages) && nextHasMultiLang) {
+      nextSupportedLanguages = Object.values(Language)
+    }
+
+    mutateFormSupportedLanguages.mutate({ nextSupportedLanguages })
+
+    return mutateFormHasMultiLang.mutate(nextHasMultiLang)
```
kevin9foong: not sure but would this cause an inconsistency error if error occurs for this but previous mutation has succeeded?

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+    const currentSupportedLanguages = settings?.supportedLanguages
+
+    // Restore back previously supported languages.
+    let nextSupportedLanguages = currentSupportedLanguages
+
+    // Add all languages as this is the first instance user turns on
+    // toggle
+    if (_.isEmpty(currentSupportedLanguages) && nextHasMultiLang) {
+      nextSupportedLanguages = Object.values(Language)
```
kevin9foong: nit: `Object.values(Language)` seems to be mentioned multiple times, can we define it once and give it a meaningful name? 

eg, `const defaultLanguages = Object.values(Language)`

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
siddarth2824: I understand. Just out of curiosity is there any reason why the convention is preferred over using `&&`?

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
siddarth2824: Thanks for catching this!

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
siddarth2824: Thanks for considering this scenario. I'll combine the two mutations into a single operation, ensuring that both must succeed; otherwise, the entire mutation will be treated as failed.

siddarth2824: Thanks for bringing up this scenario. I changed the implementation to only add the array of supported languages in the first instance that the user turns on the toggle for this feature. I also updated the model to have a default false value for the `hasMultiLang` attribute. Hence now if the mutation to update the `hasMultiLang` attribute fail, the data is not in an inconsistent state because the intended implementation was to still keep the list of supported languages even if the `hasMultiLang` value has not been updated. This is to be able to retrieve back the list of supported languages when the user turns the toggle back on.

I hope this clarifies your question.

kevin9foong:
hi @siddarth2824, 

In order for us to ensure there are no regressions and everything is working as expected, could you add in the PR description the following: 
- manual tests to run to ensure no regressions/feature works 
- UT added to cover the new functionality 
- some screenshots of before/after so others in the future know what this PR is about without executing the code locally. 

This also helps me as a reviewer who may not have much context of this feature to avoid missing anything out

You may refer to the following PR as reference: [PR 7566](https://github.com/opengovsg/FormSG/pull/7566)

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
kevin9foong: I'm not sure, I think it was a practice that has been applied before throughout the codebase (due to the belief that it is more readable) and hence just maintained for overall consistency in the codebase

siddarth2824:
I have added the screenshots and manual tests that can be run to ensure the functionality works as it should as suggested. Do let me know if anything else is required from my side as I am still getting to used to the team's practices for making PRs. Thanks!

kevin9foong:
Hi @siddarth2824, could we add some Chromatic Visual tests for the new settings page tab (& other UI components) as well? 

the rationale is to provide UI snapshots and prevent future regressions

kevin9foong:
Thanks for adding the chromatic tests!

frontend/src/features/admin-form/settings/components/MultiLanguageSection/FormMultiLanguageToggle.tsx
KenLSM: Sorry I, just saw this. 

My reason for avoiding `&&` on the repo:

This prevents us from having to invest into investigating really tricky situations where the value might be erroneously rendered. `!isLast` is definitely a boolean in this particular case, but we can never assure that another developer would use `isLast` in the next PR. To avoid such a possibility, we can refrain from using this as much as possible.

This is generally considered a footgun in my circle (even though react has evolved a lot since, and typing systems are also getting better at falsy values that will be rendered or not).

KenLSM (approved):
LGTM

PR #7563, opened 2024-08-02
chore: remove unused sgid toggle

KenLSM:
## Problem

Toggle has been already on for 10 months, but have not had a cleanup.

## Solution

Simplified and removed toggle. Constants are still kept in to avoid clashing namespace between growthbook <> source.

**Breaking Changes** 

## Tests

**Regression**
### Check that sgID MyInfo fields are available.
- [ ] Create a new form
- [ ] Enable "Singpass-App only Myinfo (Free)"
- [ ] Go to form builder
- [ ] Observe that all fields are enabled


g-tejas (approved):
LGTM!

PR #7570, opened 2024-08-06
fix: forbid stripe acc connection with pdf summary enabled

g-tejas:
## Problem
Currently, toggle stays on but locked when someone tries to connect a stripe account pdf summary enabled. This PR makes it so that he or she would have to disable pdf responses before trying to connect a stripe account.

We also change how the warnings are being shown, in bullet point instead of sentences as the combination of messages would increase factorially if we had to to create a new message for each

**Breaking Changes** 

## Before & After Screenshots
Screenshot 2024-08-06 at 1 24 48 PM

## Tests
**Ensure that payment forms cannot toggle on pdf response**
- [ ] 1. Create an encrypt mode form
- [ ] 2. Connect stripe account
- [ ] 3. Add email field
- [ ] 4. Toggle on email confirmation
- [ ] 5. Try to toggle on pdf response, should not allow, it should be locked
- [ ] 6. Observe that there's description text saying "PDF resposnse not available for payment forms"

**Ensure that forms with pdf response toggled cannot connect stripe**
- [ ] 1. Create an encrypt mode form
- [ ] 2. Add an email field
- [ ] 3. Toggle on email confirmation and also toggle pdf response.
- [ ] 4. Try to connect stripe account, it should not allow you to. 
- [ ] 5. Observe that there's a info box to remind users to disable pdf response

**Ensure that forms with pdf response disabled can connect stripe**
- [ ] 1. Create an encrypt mode form
- [ ] 2. Add an email field
- [ ] 3. Toggle on email confirmation and also DO NOT toggle pdf response.
- [ ] 4. Try to connect stripe account, it should allow you to.

**Ensure that error message is unordered list for all 6 permutations of payment exclusions**
- [ ] 1. Create an encrypt mode form
- [ ] 2. Add an admin email in the settings page
- [ ] 3. Add an email field, and enable Email confirmations + Enable pdf response
- [ ] 4. Enable singpass toggle and enable single submission toggle as well.
- [ ] 5. Navigate to payments page
- [ ] 6. Observe that the info box is in unordered list format
- [ ] 7. Repeat for 5 of the remaining permutations, ensure that if > 1 error is present, it should be an unordered list. Otherwise its just a sentence

### DIFF
--- a/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
@@ -157,7 +157,6 @@ const BeforeConnectionInstructions = ({
     const hasEmailFieldWithFormSummary = emailFields.some(
       (field) => field.autoReplyOptions.includeFormSummary,
     )
-    console.log({ hasEmailFieldWithFormSummary })
     return hasEmailFieldWithFormSummary
   }, [formDef])
 
--- a/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
@@ -57,6 +57,8 @@ const PaymentsDisabledRationaleText = ({
     isPDFResponseEnabled,
   ].filter(Boolean).length
 
+  const { data: formDef } = useAdminForm()
+
   if (disabledCount > 1) {
     return (
       <Text>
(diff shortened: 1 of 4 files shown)


frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
```diff
@@ -40,28 +43,54 @@ import {
...
 }): JSX.Element => {
-  if (isEmailsPresent && isSingleSubmission) {
+  const disabledCount = [
+    isAdminEmailsPresent,
+    isSingleSubmission,
+    isPDFResponseEnabled,
+  ].filter(Boolean).length
+
+  const { data: formDef } = useAdminForm()
+
```
KenLSM, line 62: Would prefer to do a guard check for `formDef` if it is not available.

```
if (!formDef) return <></>
```

KenLSM (approved):
LGTM, only 1 small change required.

frontend/src/features/admin-form/settings/components/PaymentSettingsSection/PaymentSettingsSection.tsx
g-tejas: 86e23b9a92d38534a58ca50fd95a45b685aaa642

Fixed!

PR #7581, opened 2024-08-08
chore: move infobox message to within tab

KenLSM:
## Problem

Infobox message regarding form closure should be associated to specific Tabs of Share modal rather than having it emplaced in its own space. This doesn't break visual continuity.

## Solution

**Breaking Changes** 

## Before & After Screenshots
|--------|--------|--------|
| Share Modal | Screenshot 2024-08-08 at 6 31 20 PM | Screenshot 2024-08-08 at 6 31 12 PM |


kevin9foong (approved):
lgtm!

PR #7597, opened 2024-08-14
feat: email notifications for mrf completed workflows 

kevin9foong:
## Problem

Admins want to be notified when an mrf is completed.   
Closes FRM-1823

## Solution

Implement MRF v2 email notifications

**Breaking Changes** 

**Features**:
Send completion email to selected emails and emails assigned to steps when mrf workflow is completion. 

**Improvements**:
**Bug Fixes and UI improvements**:
| ----------- | ------ | ----- |
| on small screens, the mrf text overflows, it should wrap to a new line. There should be a maxW on the `LogicBadge` component | image | image |
| tooltip popups should not block the component it is describing for. Hence, I have configured the popup placement with `useBreakpointValue` to prevent obfuscating the mrf fields. | image | Moved the tooltip up after consulting with Alicia |

## Before & After Screenshots

| ----------- | ------ | ----- |
| Email notifications page | | image |
| MRF builder | | image | 
| Email sent | | image |  

## Manual Tests

TC1: Verify that mrf still works (regression test)   
- [ ] Create mrf with 3 steps, assign step 1 to email field 1, step 2 to own email, step 3 to email field 2 
- [ ] Complete the workflow, ensure that the emails are still sent and all the data is filled correctly

TC2: Verify that mrf sends email workflow notifications as per steps.
- [ ] Use same mrf from TC1, add email field 1 as respondent to Step 1 of mrf
- [ ] Go to email notifications, select steps 1 and 2 for email notifications 
- [ ] Complete the workflow, ensure that can complete & email field 1 email + own email in step 2 gets the workflow completion email 
- [ ] Verify step 3 email does not get an email. 

TC3: Verify that mrf sends email workflow notifications as per steps & static emails 
- [ ] From TC2, add 2 static emails to email notifications page. Remove step 1 and 2 from email and add step 3.  
- [ ] Complete workflow. 
- [ ] Verify 2 static email + step 3 respondent email gets email completion email.


frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
```diff
@@ -84,6 +85,7 @@ export const MultiSelectProvider = ({
   downshiftComboboxProps = {},
   downshiftMultiSelectProps = {},
   inputAria,
+  overrideVirtualListItemHeight, // NOTE: for use when virtual list item height is not VIRTUAL_LIST_ITEM_HEIGHT e.g, when adding description line
```
KenLSM, line 90: Typing issue here? Don't see this declared in `MultiSelectProviderProps`

frontend/src/components/Dropdown/constants.ts
```diff
@@ -1,2 +1,4 @@
 export const VIRTUAL_LIST_MAX_HEIGHT = 12 * 16
 export const VIRTUAL_LIST_OVERSCAN_HEIGHT = 4 * 16
+export const VIRTUAL_LIST_ITEM_HEIGHT = 48
+export const VIRUTAL_LIST_ITEM_WITH_DESCRIPTION_HEIGHT = 68
```
KenLSM: Didn't see this being used / referenced

kevin9foong: Removed!

frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
kevin9foong: Added to `MultiSelectProviderProps`. Chose to keep this in so that we can override in the future since the label description has been added.

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.tsx
```diff
@@ -57,6 +57,22 @@ export const EditStepBlock = ({
...
 
   const handleSubmit = formMethods.handleSubmit((inputs: EditStepInputs) => {
+    if (isFirstStepByStepNumber(stepNumber)) {
+      if (inputs.field) {
+        return onSubmit({
+          ...inputs,
+          workflow_type: WorkflowType.Dynamic,
+          field: inputs.field,
+        })
+      } else {
```
KenLSM: /nit don't need an `else`, since we're returning in the control statement

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/QuestionsBlock.tsx
```diff
@@ -1,10 +1,18 @@
 import { Controller, UseFormReturn } from 'react-hook-form'
-import { Flex, FormControl, Icon, Stack, Text } from '@chakra-ui/react'
+import {
+  Flex,
+  FormControl,
+  Icon,
+  Stack,
+  Text,
+  useBreakpointValue,
```
KenLSM: A few unused imports here that might have been missed!

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/InactiveStepBlock/InactiveStepBlock.tsx
```diff
@@ -24,12 +27,63 @@ interface InactiveStepBlockProps {
...
+interface RespondentBadgeProps {
+  step: FormWorkflowStepDto
+  idToFieldMap: Dictionary<FormFieldWithQuestionNo<FormField>>
+}
+const FirstStepRespondentBadge = ({
+  step,
+  idToFieldMap,
+}: RespondentBadgeProps): JSX.Element | null => {
+  if (step.workflow_type === WorkflowType.Static && step.emails.length > 0) {
+    throw new Error('First step must not have static emails.')
```
KenLSM: Can throwing be avoided?

There's no way the user can self-correct and makes the workflow page completely inaccessible. In general, we don't throw on FE except on certain cases such as:
1.  `Provider/Context`: where the page completely fails to be operable
2. `mutations`: where the query has a capture on thrown errors, and a toast or feedback is shown

Edit: there are a few more cases (listing them doesn't value add to this discussion), but definitely not on data assertions

KenLSM: It feels like this is a data assertion, which does not impact whether the contents below could be shown.

frontend/src/mocks/msw/handlers/admin-form/settings.ts
```diff
@@ -33,7 +66,9 @@ export const getAdminFormSettings = ({
             }).form,
             mode === FormResponseMode.Email
               ? EMAIL_FORM_SETTINGS_FIELDS
-              : STORAGE_FORM_SETTINGS_FIELDS,
+              : mode === FormResponseMode.Encrypt
```
KenLSM: Getting slightly messier, and can only increase in messiness, how about extracting it to a map to stem it?

```ts
export const getAdminFormSettings = ({ ... }) => {
  const MODE_TO_SETTINGS_FIELD_MAP = {
    [FormResponseMode.Email]: EMAIL_FORM_SETTINGS_FIELDS,
    [FormResponseMode.Encrypt]: STORAGE_FORM_SETTINGS_FIELDS,
    [FormResponseMode.Multirespondent]: MULTIRESPONDENT_FORM_SETTINGS_FIELDS,
  }
  return rest.get<FormSettings>(
        ...
        ctx.json(
          pick(
            createMockForm({
              _id: req.params.formId as FormId,
              responseMode: mode,
              ...overrides,
            }).form,
            MODE_TO_SETTINGS_FIELD_MAP[mode],
          ),
          ...
}
```

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
```diff
@@ -42,19 +42,17 @@ export const createMultirespondentSubmissionDto = (
 }
 
 export const retrieveWorkflowStepEmailAddresses = (
-  formWorkflow: FormWorkflowDto,
-  nextWorkflowStep: number,
+  step: FormWorkflowStepDto,
   responses: FieldResponsesV3,
-): Result<string[], InvalidWorkflowTypeError> => {
```
KenLSM, line 48: I would recommend retaining `neverthrow` typings  since it can be extremely difficult to debug if the controller was not handling the correct types or the function returning the wrong types.

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+        return
+      }
+      return mutateMrfEmailNotifications.mutate({
+        emails: nextStaticEmails,
+        stepsToNotify: nextStepsToNotify,
+      })
+    },
+    [mutateMrfEmailNotifications, emails, stepsToNotify],
+  )
+  const DEBOUNCE_DELAY_IN_MS = 800
```
KenLSM: /nit can move constants to top of file

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+}: MrfEmailNotificationsFormProps) => {
+  const { isLoading, formWorkflow } = useAdminFormWorkflow()
+
+  const formWorkflowStepsWithStepNumber =
+    formWorkflow?.map((step, index) => ({
+      ...step,
+      stepNumber: index + 1,
+    })) ?? []
+
+  const checkIsEmail = useMemo(() => isEmail, [])
```
KenLSM: This doesn't seem to be doing anything special other than to reassign to a different variable name? 🤔

frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
```diff
@@ -5,43 +5,85 @@ import {
...
+  onSubmit,
+}: AdminEmailRecipientsInputProps): JSX.Element => {
+  const { getValues, setValue, control, handleSubmit } = useFormContext<{
+    emails: string[]
+    isRequired: boolean
+  }>()
+
+  const { data: settings } = useAdminFormSettings()
+
+  const tagValidation = useMemo(() => isEmail, [])
```
KenLSM: This doesn't seem to be doing anything special other than to reassign to a different variable name? 🤔

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { useAdminFormWorkflow } from '~features/admin-form/create/workflow/hooks/useAdminFormWorkflow'
+
+import { useMutateFormSettings } from '../mutations'
+
+interface MrfEmailNotificationsFormProps {
+  settings: MultirespondentFormSettings
+  isDisabled: boolean
+}
+
+const WorkflowEmailMultiSelectName = 'email-multi-select'
```
KenLSM: Great that we're using constants for reused values, but can the variable name be in `SCREAMING_SNAKE_CASE`?

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+              }) => (
+                <MultiSelect
+                  items={formWorkflowStepsWithStepNumber.map((step) => ({
+                    label: `Respondent(s) in Step ${step.stepNumber}`,
+                    value: step._id,
+                  }))}
+                  values={values}
+                  onChange={(values) => {
+                    onChange(values)
+                    handleWorkflowEmailMultiSelectChange()
```
KenLSM: This section feels a little too complex to handle. Looking at seeing how this can be simplified.

Is the reason for this additional function call + debounce due to the fact that we don't have an `onBlur` to trigger the event?

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+          >
+            Notify other parties
+          </FormLabel>
+          <Controller
+            name={OtherPartiesEmailInputName}
+            control={control}
+            rules={OPTIONAL_ADMIN_EMAIL_VALIDATION_RULES}
+            render={({ field }) => (
+              <TagInput
+                {...(getValues(OtherPartiesEmailInputName)?.length > 0
```
KenLSM: Could you help to improve the placeholder declaration with the below code? Understand that this is not related to your recent changes.

```suggestion
  // hide display of placeholder if there are values in the field
  const otherFieldPlaceholder =
    getValues(OtherPartiesEmailInputName)?.length > 0
      ? undefined
      : 'me@example.com'
    ...
      <TagInput
        placeholder={otherFieldPlaceholder}
        ...
      />
      
```

frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
```diff
@@ -5,43 +5,85 @@ import {
...
+      control={control}
+      name="emails"
+      rules={
+        settings?.responseMode === FormResponseMode.Email
+          ? REQUIRED_ADMIN_EMAIL_VALIDATION_RULES
+          : OPTIONAL_ADMIN_EMAIL_VALIDATION_RULES
+      }
+      render={({ field }) => (
+        <TagInput
+          {...(getValues('emails')?.length > 0
```
KenLSM: Similar to https://github.com/opengovsg/FormSG/pull/7597/files#r1731368617

frontend/src/components/Dropdown/SingleSelect/SingleSelectProvider.tsx
```diff
@@ -21,7 +21,7 @@ export interface SingleSelectProviderProps<
 > extends SharedSelectContextReturnProps<Item>,
     FormControlOptions {
   /** Controlled selected value */
-  value: string
+  value: string | undefined
```
KenLSM: Any specific reason why this should be switched to undefined? We usually take `''`

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
```diff
@@ -44,12 +58,54 @@ export const RespondentBlock = ({
...
+                      !emailFieldItems ||
+                      emailFieldItems.some(
+                        ({ value: fieldValue }) => fieldValue === selectedValue,
+                      ) ||
+                      'Field is not an email field'
+                    )
+                  },
+                }}
+                control={control}
+                render={({ field: { value = undefined, ...rest } }) => (
```
KenLSM: Related to https://github.com/opengovsg/FormSG/pull/7597/files#r1732039504

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
```diff
@@ -44,12 +58,54 @@ export const RespondentBlock = ({
...
+          <Flex alignItems="center" gap="0.5rem">
+            <Text textStyle="subhead-3">Respondent in this step</Text>
+            <Tooltip
+              placement="top"
+              label="Anyone you share the form link with"
+            >
+              <Icon as={BxsInfoCircleAlt} />
+            </Tooltip>
+          </Flex>
+          <FormControl isInvalid={!!errors.field} isRequired={false}>
```
KenLSM: Typo here? We can omit specifying `isRequired={false}` since it already defaults to [false](https://v2.chakra-ui.com/docs/components/form-control/props#:~:text=false-,isRequired,-Description).

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.tsx
kevin9foong: Resolved!

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
kevin9foong: hey ken, im trying to implement the onBlur logic for the form multiselect component, and i realised why the MultiSelect component may not have had an `onBlur` prop implemented previously

ive set an onBlur callback function which is invoked when the focused input box is blurred. this works fine for when selecting elements from the dropdown.

However, when clearing elements, for example when clicking the x button, the input box is not focused ever and hence the onBlur event does not trigger. This means that the onBlur callback is not invoked

image
element is focused when selecting steps from the virtual list 

however, when clearing, input component is never focused 
image
hence, admin must click on input box to focus then click out to trigger `onBlur` callback. 

Conclusion: use debounce + `onChange`

kevin9foong: Adding the following comment for documentation of what we discussed: 

Kevin: 
there are 3 approaches i can take:
1. keep current implementation in GH PR: use onChange for everything (current design) -> which is imo easier to understand. dont think it will negatively affect admin's usage of the toggle

2. i can use onBlur for when user adds steps and onChangeonly when the user clicks on xbutton. but there are some tradeoffs discovered only after I tried implementing:
- still better to have debounce for x to avoid multiple api calls when user deletes multiple steps, hence does not solve problem from 1
- different behavior for adding steps (updates only when on blur) vs deleting steps (updates immediately)
- need to add a if (selectedItems.length < prevSelectedItems.length) submitForm;
- need to call submit form twice from onBlur and onChange each

3. implement clicking x button focuses onto the input box. but this is indeed a strange behavior (and will affect the default behavior of other multiselect)  

my take is since the intention is to make the code more understandable, options 2 and 3 dont really achieve that.
I do feel that approach 1 is already reasonably understandable and we might not need to change to 2 or 3.
current implementation is probably acceptable for current use case and i should probably move on to next part instead of spending more time on this.

Ken: 
Hm i had troubles understanding 1.
i.e., why is this here
```
onChange={(values) => {
    onChange(values)
    handleWorkflowEmailMultiSelectChange()
}}
```

Then we have debounce etc to add in the complexity. When you think of using a MultiCombobox, do you have to think of debounce + onChange to track every changes, or you think that MultiCombobox should tell you (1) something changed or (2) when the user had done updating? They are have two different levels of functionality.
That’s one of the main reason i asked, why was onChange used. Was it due to convenience or restriction. It sounded like more like a convenience rather than restriction.
Natural behaviour of automatic fields updates on FormSG is generally onBlur (and as well as other projects that i’ve worked on).
You can go think about it. If approach (1) is still the best, then lets take that.

kevin9foong:
- [x] To handle the debouncing stuff 

Replaced with `onBlur` which is a cleaner implementation + less complex (no delay or debouncing intricacies needed)

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
kevin9foong: Okay, this is resolved! 

I've used option 3 which is i think the cleanest solution and upon trying it out it does look intuitive and not really odd to focus on remove and avoids the delay issue faced with debouncing.

kevin9foong:
Ready for final review!

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
```diff
@@ -31,8 +35,23 @@ export const RespondentBlock = ({
     formState: { errors },
     register,
     getValues,
+    control,
   } = formMethods
 
+  const isTest = process.env.NODE_ENV === 'test'
```
KenLSM, line 42: /nit could you also put the note here to remove the flag like you did [here](https://github.com/opengovsg/FormSG/pull/7620/files#diff-b7b4ae2945123935752455bbf96f841eb7d4bdaa1800eed17533bf5b25988401L323)? It made it easy to reference and cleanup.

frontend/src/features/admin-form/settings/SettingsEmailsPage.tsx
```diff
@@ -1,49 +1,76 @@
...
-  // should render null
-  if (!isEmailOrStorageMode) {
-    return null
+const FormEmailSectionContainer = ({
+  isDisabled,
+  settings,
+}: FormEmailSectionContainerProps): JSX.Element => {
+  if (settings.responseMode === FormResponseMode.Multirespondent) {
+    return <MrfFormEmailSection isDisabled={isDisabled} settings={settings} />
+  } else {
```
KenLSM: /nit Don't an else here. 

P.S. I generally favor guard clauses and to avoid using and else. Keeps me from creating messy deeply nested if-else.

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -60,55 +63,56 @@ export const SettingsPage = (): JSX.Element => {
...
+      component: SettingsWebhooksPage,
+      path: 'webhooks',
+    },
+    {
+      label: 'Payments',
+      icon: BiDollar,
+      component: SettingsPaymentsPage,
+      path: 'payments',
+    },
+  ].filter((tab) => tab !== null) as TabEntry[]
```
KenLSM: /nit

```suggestion
  ].filter(Boolean) as TabEntry[]
```

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -60,55 +63,56 @@ export const SettingsPage = (): JSX.Element => {
...
-      },
-    ]
-
-    return baseConfig.filter(isNonEmpty)
-  }, [settings?.responseMode])
+  const isTest = process.env.NODE_ENV === 'test'
+  // For beta flagging email notifications tab.
+  const emailNotificationsTab =
+    isUserLoading ||
+    isFormSettingLoading ||
```
KenLSM, line 70: Want to understand if we actually need to explicitly check on `isUserLoading || isFormSettingLoading`.

Wouldn't `!user?`  manage the situation where the data isn't ready yet i.e., `isUserLoading`? (similarly for `!formSettings?`)

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -279,45 +285,70 @@ const _createSubmission = async ({
...
-        currentWorkflowStep: submissionContent.workflowStep,
+        currentWorkflowStep: currentStepNumber,
         formId: form._id,
         submissionId,
       },
       error: err,
     })
   }
+
+  if (isTest || form.admin.betaFlags.mrfEmailNotifications) {
```
KenLSM, line 298: Same note here regarding the reference comment to allow quick lookup during cleanups

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -347,6 +377,115 @@ const runMultirespondentWorkflow = ({
...
+        error,
+      })
+      return errAsync(error)
+    })
+  } else {
+    return okAsync(true)
+  }
+}
+
+const sendMrfOutcomeEmails = ({
```
KenLSM: This function could likely be moved into `.service` as it doesn't look like it's dealing with `HttpRequests/Responses`.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -347,6 +377,115 @@ const runMultirespondentWorkflow = ({
   )
 }
 
+const sendMrfCompletionEmailIfWorkflowCompleted = ({
```
KenLSM: This function could likely be moved into .service as it doesn't look like it's dealing with `HttpRequests/Responses`.

KenLSM (approved):
LGTM! Some suggestions on refactors. Glad to see that this PR doesn't have much scope creep!

frontend/src/features/admin-form/settings/SettingsPage.tsx
kevin9foong: Thanks for this suggestion. Learnt something new about JS today.

frontend/src/features/admin-form/settings/SettingsPage.tsx
kevin9foong:
```
 const emailNotificationsTab =
    isUserLoading ||
    isFormSettingLoading ||
    // TODO: (MRF-email-notif) Remove isTest and betaFlag check when MRF email notifications is out of beta
    (!isTest &&
      formSettings?.responseMode === FormResponseMode.Multirespondent &&
      !user?.betaFlags?.mrfEmailNotifications)
      ? null
      : {
          label: 'Email notifications',
          icon: BiMailSend,
          component: SettingsEmailsPage,
          path: 'email-notifications',
          showRedDot: true,
        }
```

For this part of the code, when the `formSettings` is not loaded yet, `formSettings?.responseMode === FormResponseMode.Multirespondent ` evaluates to `falsy` value which causes the tab to be included. 

When the `formSettings` are finally loaded, then it will see it is not MRF and then the tab disappears. 

This causes a flickering issue. 

Alternatively, I could use `!formSettings` -> `null` as the check to replace `isFormSettingLoading` if desired. However, I felt that using the `isLoading` checks more accurately explains what this clause is doing and might be more readable and hence I chose that.

kevin9foong: For example, I could change it to: 
```
 const emailNotificationsTab =
    !user ||
    !formSettings ||
    // TODO: (MRF-email-notif) Remove isTest and betaFlag check when MRF email notifications is out of beta
    (!isTest &&
      formSettings.responseMode === FormResponseMode.Multirespondent &&
      !user.betaFlags?.mrfEmailNotifications)
      ? null
      : {
          label: 'Email notifications',
          icon: BiMailSend,
          component: SettingsEmailsPage,
          path: 'email-notifications',
          showRedDot: true,
        }
```

PR #7598, opened 2024-08-15
chore(myinfo): update myinfo race list 15-aug-2024

KenLSM:
## Problem

There were new races added since the last update which causes some desync on accepted values from our side.

## Solution

Update myinfo const list obtained from https://api.singpass.gov.sg/library/myinfo/business/implementation-myinfo-data

**Breaking Changes** 


kevin9foong (approved):
Verified that the list now is in sync with all 228 races. 

Thanks for this fix!

PR #7600, opened 2024-08-15
chore(myinfo): update myinfo nationalities list 15 aug 2024

kevin9foong:
## Problem
Myinfo has updated the nationalities list. We should keep it in sync. 
Otherwise, there will be MyInfo validation errors when respondents pull data from Myinfo that does not match the formsg local accepted values. 

## Solution

Update the constants file based on NationalityCitizenshipCode section's description downloaded from [Myinfo API Data](https://api.singpass.gov.sg/library/myinfo/business/implementation-myinfo-data) page


KenLSM (approved):
LGTM!

PR #7617, opened 2024-08-22
fix(mrf): ensure number of non-editable fields are still present

KenLSM:
## Problem

Closes FRM-1824

## Solution

**Breaking Changes** 

### DIFF
--- a/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -461,6 +461,20 @@ export const validateMultirespondentSubmission = async (
                 const previousResponses =
                   previousSubmissionDecryptedContent.responses as ParsedClearFormFieldResponsesV3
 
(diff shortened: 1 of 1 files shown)


kevin9foong (approved):
lgtm as a mitigation measure. thought i do feel that it makes the mrf code more confusing / complex as inserting the existing fields again seems odd

think we will need to refactor this and make it more understandable in future

KenLSM:
> lgtm as a mitigation measure. thought i do feel that it makes the mrf code more confusing / complex as inserting the existing fields again seems odd
> 
> think we will need to refactor this and make it more understandable in future

Yep. Mainly three approaches were tested:
1. Checking if non-editable fields were edited
2. Add non-editable fields back
3. Taking the source of truth as the previous response and if those fields exist, and are non-editable, it should similarly exist, otherwise reject those submission.

Approach 1 fails on fields that are not 

Approach 2, this current implementation, smallest change, but feels unorthodox. Takes the assumption that if it was correct, it should be correct in the future.

Approach 3, most elegant, but requires changes and more testing.

PR #7619, opened 2024-08-22
feat(nric): set default to false

KenLSM:
## Problem

We have completed the datafix, and there's no need to set as default. This PR should be released in concert with communications to our users

## Solution

Newly created forms will be defaulted to NRIC opt-in

**Breaking Changes** 


kevin9foong (approved):
lgtm! thanks for helping to update this!

PR #7620, opened 2024-08-22
feat(nric): remove override flag

KenLSM:
## Problem

We have completed our datafix for NRIC collection feature. There is no more need to retain this flag.

## Solution

Assumed true for values of the flag and remove the flag

**Breaking Changes** 


kevin9foong (approved):
Thanks for the changes! 

I have also removed the use of the flag in the FE to keep it consistent. 

Justification for why remove FE as well: 
- Without removing the FE, if this isSubmitterIdCollectionEnabled is `off`, the FE toggle disappears but collection still occurs if the field is set to `true` in the database. 
- hence, the flag no longer turns on or off submitter id collection as indicated by the flag name and hence flag should be removed from entire DB.

PR #7621, opened 2024-08-22
chore(nric): remove nric collection description

KenLSM:
## Problem

After several showing to various people, it was found that the description can be very confusing. The description itself might be making things worser than better.

## Solution

Less is more. Remove the description for the toggle, the behaviour should be rather straightforward.

**Breaking Changes** 


kevin9foong (approved):
verified the ui changes. thanks for removing the copy changes!

PR #7630, opened 2024-08-28
feat: auto-select step 1 for email notifications for mrf forms 

kevin9foong:
Note: This includes and should be merged only after PR #7957

## Problem
Based on user tests, most admins will want to send mrf outcome notifications for step 1 respondents. 
Furthermore, we no longer need the create workflow page and want a call to action to tell admin to fill in the step 1 workflow builder form.

Closes FRM-1828

## Solution
We take an opinionated approach and autoselect step 1 by default for email notifications. 
Prevent the create workflow page from showing up by setting default empty workflow step 1.
Check if only 1 step & empty workflow step 1. If so, set it to step 1 to active editing mode.

**Breaking Changes** 

## Before & After Screenshots

Step 1 respondents are auto selected by default 
image

Set step 1 edit to active if only 1 step & step 1 is empty: 
image

Message at workflow builder tab to inform user of default selection of respondent 1 for email notifications.  
image

## Manual tests 

TC 1: Create workflow page still exists for existing mrf forms 
- [ ]  Open any existing MRF form without any workflow steps 
- [ ] Go to create workflow tab and verify the `Create workflow` button still exists 

TC2: Sets email notification to step 1 by default and creates 1 step workflow upon MRF form creation 
- [ ] Create new MRF form 
- [ ] Go to create workflow tab -> verify only 1 empty workflow step exists 
- [ ] Ensure that the step 1 is in active editable mode. 
- [ ] Go to email notifications setting -> verify that step 1 is selected to receive notifications. 
- [ ] Without editing step 1, add another workflow step 2. Verify that when retuning to workflow tab, step 1 does not automatically enter active mode. 
- [ ] Delete this workflow step 2. Refresh and observe that step 1 is now automatically set to active mode again. 
- [ ] Add email field or fields to step 1 and save. Refresh and return to workflow tab, verify that step 1 does not automatically enter active mode.


kevin9foong:
- [x] Pending unit tests (single UT on `form.server.model`)
- [x] Pending manual tests

KenLSM:
@kevin9foong Just FYI, i've updated this PR to target `feat/mrf-email-notifications`

kevin9foong:
Ready for review

src/app/models/form.server.model.ts
```diff
@@ -395,6 +404,18 @@ MultirespondentFormWorkflowPath.discriminator(
   WorkflowStepDynamicSchema,
 )
 
+MultirespondentFormSchema.post('save', async function (doc, next) {
```
KenLSM, line 407: Can the admins explicitly remove the respondents from the first steps? If so, ignore this comment.

kevin9foong: Yes, they can. If an admin wants to change a field/static email from step 1, they would go to the step 1 builder and change it there.  
They can also go to the email notifications setting to unselect step 1 from receiving the email notifications.  

To elaborate, `stepsToNotify` stores the ids of the steps that have been selected (and does not contain what email fields/static emails to send to).
When sending out notification emails, we loop through the steps and extract out the static emails / fields and then forward it to the user.

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/WorkflowContent.tsx
```diff
@@ -22,6 +45,7 @@ export const WorkflowContent = (): JSX.Element | null => {
         ))}
         <NewStepBlock />
       </Stack>
+      <DefaultOutcomeNotificationBlock />
```
KenLSM: Should this be also conditionally rendered if `form.admin.betaFlags.mrfEmailNotifications`? Otherwise, the admin might be navigated to the `Settings > General` instead of `Settings > Email Notifications (MRF Flavour)`

src/app/models/form.server.model.ts
KenLSM: Hm. There will be a funky behaviour here. 

1. Delete Respondent in `Settings > Email notifications > Notify respondents in your workflow`
2. Refresh the page
3. Observe that the field remains empty
4. Add collaborator `ken+1@open.gov.sg` [a]
5. Observe that `Settings > Email notifications > Notify respondents in your workflow` has some value set.

[a] any model method that triggers a `.save()`. Examples of such actions are `updateFormCollaborators`, `updateFormFieldById`, `insertFormField`, `duplicateFormFieldByIdAndIndex`, ...


https://github.com/user-attachments/assets/71880680-9e37-4f2f-b459-473d8a5740b8

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/WorkflowContent.tsx
KenLSM: Added in 7f06a83ea

src/app/models/form.server.model.ts
kevin9foong: As discussed, will choose the simplest solution: 
- move this to controller -> update `stepsToNotify` without need for txn.

kevin9foong:
The discussion points above still hold, but will not be pushing this PR after discussion with the team. Closing PR and can reopen if we decide to implement this default step 1 auto-select functionality in the future.

PR #7636, opened 2024-09-02
feat: approvals for mrf 

kevin9foong:
## Problem
From mrf workshops and user testing sessions, admins want a way for mrf to be used for their approval processes.  

Closes FRM-1632, FRM-1829, FRM-1834, FRM-1835, FRM-1836, FRM-1856, FRM-1863

## Solution
Implement approval support for mrf according to Approvals v8. 

**Breaking Changes** 

**Features**:

**Improvements**:
- Revamped UI improve understandability of MRF builder. 
- Radio is now vertical which follows design conventions instead of horizontal. 
- Added validation for workflow email field & approval fields on per workflow step granularity. 

## Before & After Screenshots

Revamped workflow builder (inactive state):
- more obvious error message with contrast 
- addition of approval step 

| Before | After | 
| ------ | ------ | 
| image | image |

Revamped workflow builder (Active state): 
- new radio layout 
- addition of approval fields 
- more directed field prompts 

| Before | After | 
| ------ | ------ | 
| image | image |

New approvals outcome email: 
image

## Manual QA Tests

Test set up: 
- [ ] Create an MRF form called 'finance claims approval' 
- [ ] Add 2 approval yes/no fields named 'RO approve', 'Finance dept approve'. Assert that `Use for approvals` badge appears beside yes/no fields in drawer. 
- [ ] Add 2 email fields
- [ ] Assign 1 email field to step 1 to edit, Assign approval field to step 2 to edit and Assign approval field to step 3 to edit. (do not assign approval field or step 1 email notification field yet)
- [ ] Let step 2 be static email, let step 3 be dynamic email field 
  
TC1: Regression test - MRF still works
- [ ] Open the form to public 
- [ ] Respond to form, ensure that all steps of mrf can be completed and result is in admin responses tab. 
- [ ] Assert no email outcomes are sent, since not yet configured in email notification settings.  

TC2: Regression test - email notif works: 
- [ ] From TC1, config email notif setting to step 2 and static email of choice.
- [ ] Complete submission, assert that workflow completed (note that it is not an outcome status email) is sent to expected emails.  
- [ ] Remove the email notifs to prepare for TC3. 

TC3: Approve without email notif. Expect no email sent  
- [ ] Assign approval field to step 2 only. 
- [ ] Complete all 3 steps of mrf workflow and approve at step 2.  
- [ ] No email is sent. 

TC4: Reject without email notif. Expect no email sent  
- [ ] Use same form from TC3
- [ ] Reject at step 2 of mrf workflow.  
- [ ] No email is sent to step 3 respondent since workflow ends, no outcome email is sent. 

TC5: Approve and reject with email notification works: 
- [ ] Add step 1 and step 3 to email notification and a static email to email notification. 
- [ ] Repeat TC3. Assert that only email from step 3 and static email receives approve email, since step 1 does not have email field setup. 
- [ ] Add email field to step 1 in workflow. 
- [ ] Repeat TC4. Assert step 1 and step 3 and static email receive not approved email. 

TC6: Multiple approval steps - approve all
- [ ] From TC5, add approval step for step 3 of workflow. 
- [ ] Approve all approval steps, approve email should be received. 

TC7: Multiple approval steps  - Reject at step 2 (middle step) 
- [ ] From TC6, make submission
- [ ] Reject at step 2, assert R3 does not receive email and not approved email is sent to expected emails. 

TC8: Multiple approval steps - Reject at last step 
- [ ] From TC7, make submission
- [ ] Approve at step 2 but Reject at step 3, assert not approved email is sent to all expected emails. 

Edge cases: 
TC10: approval field set to optional, respondent did not fill  
- [ ] From TC9, set the approval field used for step 2 as optional. 
- [ ] Try and submit the steps of form, but at step 2, do not fill in the optional approval field. 
- [ ] assert that R3 no longer gets the next step email, no completion/approval outcome emails are sent. 
- [ ] Visit the responses page, the R2 submission is recorded. 
- [ ] Share the next step response link, it goes to R3 and can be resumed.
- [ ] Try and submit R3 and approve/reject, it should send out approved/reject email to expected emails.  

TC10: approval field deleted, admin did not reassign 
- [ ] From TC9, delete the approval field used for step 2. 
- [ ] Try and submit the steps of the form.
- [ ] assert that R3 no longer gets the next step email, no completion/approval outcome emails are sent. 
- [ ] Visit the responses page, the R2 submission is recorded. 
- [ ] Share the next step response link, it goes to R3 and can be resumed. 
- [ ] Try and submit R3 and approve/reject, it should send out approved/reject email to expected emails.


kevin9foong:
Feature Todo: 
- [x] BE send outcome email based on approve/reject 
- [x] BE validate if form field is valid yes/no form field (done on the save step form edit)
- [x] Fix FE validation error message -> focus -> need to click twice issue

kevin9foong:
Moving to testing phase: 
- [x] Pending manual QA tests
- [x] unit tests 
- [x] chromatic visual tests

kevin9foong:
Add `betaFlag` same as email notifications before merging

kevin9foong:
- Written comprehensive visual and interaction tests for FE workflow `editStepBlock` error states and inactive states. 
- Written UT for various approvals edge cases

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -95,6 +95,7 @@ export const FormLabel = ({
           >
             <Icon
               ml="0.5rem"
+              mb="0.1rem"
```
KenLSM, line 98: We usually don't use `0.1rem`. Is there an issue with the bottom margin on this?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/FieldListDrawer.stories.tsx
```diff
@@ -0,0 +1,52 @@
+import { DragDropContext } from 'react-beautiful-dnd'
```
KenLSM, line 1: Thanks for helping to add more storybook tests!

Not an issue, just taking the opportunity to think about what kind of components we're trying to do UI testing for.

Generally there are three kinds of UI that we want to test:
1. Reusable components. Components such as `<FormLabel />` which is expected to be re-used on multiple places
2. Page-level views. Generally to snapshot tests on mobile vs desktop UI
3. Features. Secret key modals, csv upload flow for conditional routing.

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
```diff
@@ -57,7 +68,7 @@ export const FieldLogicBadge = ({
   }, [defaults.variant, fieldMeta])
 
   return (
-    <LogicBadge display="inline-flex">
+    <LogicBadge display="inline-flex" bgColor={badgeBgColor}>
```
KenLSM: Was usage of `colorScheme` considered here?

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/FieldListDrawer/FieldListDrawer.stories.tsx
kevin9foong: I added this test for Alicia to verify that the `Use for approvals` badge is displayed for MRF forms only and not for other modes.

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
kevin9foong: Without it, the tooltip icon appears 1-2px too low. Perhaps there is a better way to resolve this

src/app/modules/core/core.errors.ts
```diff
@@ -63,8 +64,9 @@ export enum ErrorCodes {
   SUBMISSION_MALICIOUS_FILE_DETECTED = 100220,
   SUBMISSION_INVALID_WORKFLOW_TYPE = 100221,
   SUBMISSION_ATTACHMENT_UPLOAD = 100222,
+  SUBMISSION_EXPECTED_RESPONSE_NOT_FOUND = 100223,
   // Email Submission Errors (email mode deprecated soon)
-  EMAIL_SUBMISSION_HASH = 100223,
```
KenLSM, line 67: Generally we try to avoid changing error codes. Any specific reason why this is done? 🤔

src/app/modules/form/admin-form/admin-form.middlewares.ts
```diff
@@ -41,24 +41,6 @@ export const updateSettingsValidator = celebrate({
       gstRegNo: Joi.string().allow(''),
     }),
     payments_field: Joi.object({ gst_enabled: Joi.boolean() }),
-    workflow: Joi.array()
```
KenLSM, line 44: Likely that it might break for a very very minute set of users that were editing their workflow. Should be fixed when they refresh the page. Just need to let the team know that this is not a bug!

src/app/modules/form/admin-form/admin-form.middlewares.ts
```diff
@@ -82,3 +64,49 @@ export const getWebhookSettingsValidator = celebrate({
...
+export const createWorkflowStepValidator = celebrate({
+  [Segments.BODY]: Joi.object({
+    workflow_type: Joi.string().valid(...Object.values(WorkflowType)),
+    emails: Joi.when('workflow_type', {
+      is: WorkflowType.Static,
+      then: Joi.array().items(Joi.string().email()).required(),
+    }),
+    field: Joi.when('workflow_type', {
+      is: WorkflowType.Dynamic,
+      then: Joi.string().hex().length(24).required(),
```
KenLSM: I understand that this is to ensure that this fits the our current ObjectId restriction. However, I think these are implementation details that the application doesn't have to strictly check for. 

i.e., 
whether mongodb is using a 24-char hex string or 76-char base64 or whether our application restricts to only allow generation of 24-char hex strings.

src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1468,25 +1492,173 @@ export const _handleUpdateWebhookSettings: ControllerHandler<
...
+        AdminFormService.createWorkflowStep(form, workflowStepToCreate),
+      )
+      .map((updatedWorkflow) =>
+        res.status(StatusCodes.OK).json(updatedWorkflow),
+      )
+      .mapErr((error) => {
+        logger.error({
+          message: 'Error occurred when creating form field',
+          meta: {
+            action: '_handleCreateFormField',
```
KenLSM: /nit remove `_` for consistency (@KenLSM to fix it).

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/ApprovalsBlock.tsx
```diff
@@ -0,0 +1,140 @@
...
+}: ApprovalsBlockProps): JSX.Element => {
+  const {
+    control,
+    setValue,
+    formState: { errors },
+    clearErrors,
+    watch,
+  } = formMethods
+  const selectedApprovalField = watch(APPROVAL_FIELD_NAME)
+  const selectedEditFields = watch(FIELDS_TO_EDIT_NAME)
```
KenLSM, line 36: Didn't see where is the corresponding form field that this name is tagged to. Is this used?

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlockContainer.tsx
```diff
@@ -0,0 +1,18 @@
+import { Stack } from '@chakra-ui/react'
+
+export const EditStepBlockContainer = ({
```
KenLSM, line 3: Great abstraction!

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlockContainer.tsx
```diff
@@ -0,0 +1,18 @@
...
+  children,
+}: {
+  children?: React.ReactNode
+}): JSX.Element => {
+  return (
+    <Stack
+      direction="column"
+      spacing="0.75rem"
+      px={{ base: '1.5rem', md: '2rem' }}
+      borderTopColor="secondary.200"
```
KenLSM: Is this used? Was this supposed to be removed together along with the old `borderTopWidth="1px" borderTopColor="secondary.200"`?

Didn't see any accompanying `borderTopWidth` or `borderWidth`. Also there `<Divider />`s that were added between the blocks.

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
```diff
@@ -1,47 +1,218 @@
...
+
+  return (
+    <>
+      <Radio
+        isDisabled={isLoading}
+        isLabelFullWidth
+        allowDeselect={false}
+        value={WorkflowType.Static}
+        {...register('workflow_type', WORKFLOW_TYPE_VALIDATION)}
+        px="8px"
```
KenLSM: "0.5rem"

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
```diff
@@ -1,47 +1,218 @@
...
+
+  return (
+    <>
+      <Radio
+        isDisabled={isLoading}
+        isLabelFullWidth
+        allowDeselect={false}
+        value={WorkflowType.Dynamic}
+        {...register('workflow_type', WORKFLOW_TYPE_VALIDATION)}
+        px="8px"
```
KenLSM: "0.5rem"

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
```diff
@@ -119,158 +290,32 @@ export const RespondentBlock = ({
         </>
       ) : (
         <>
-          <Text textStyle="subhead-3">Respondent in this step</Text>
           <FormControl
             isReadOnly={isLoading}
-            id="workflowType"
```
KenLSM, line 125: Might not need `id` anymore since there's no confusion by react from `<RespondentInput />`

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
```diff
@@ -119,158 +290,32 @@ export const RespondentBlock = ({
         </>
       ) : (
         <>
```
KenLSM: `React.Fragment` is not used, can remove.

frontend/src/features/admin-form/create/workflow/hooks/useAdminFormWorkflow.ts
```diff
@@ -36,14 +37,28 @@ export const useAdminFormWorkflow = () => {
...
+          field,
+        ): field is FormFieldWithQuestionNo<FormFieldDto<YesNoFieldBase>> =>
+          field.fieldType === BasicField.YesNo,
+      ),
+    [augmentedFormFields],
+  )
+
+  const formWorkflow =
+    form?.responseMode !== FormResponseMode.Multirespondent
+      ? undefined
```
KenLSM: Any particular effects we're trying to achieve with `undefined`? Otherwise would recommend using `null` instead. 

```
const undefinedFoo = undefined;
const nullFoo = null;
console.log(JSON.stringify({ undefinedFoo, nullFoo})) // { nullFoo: null }, undefined will not be present
```

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.tsx
```diff
@@ -41,6 +43,8 @@ export const EditStepBlock = ({
   const formMethods = useForm<EditStepInputs>({
     defaultValues,
   })
+  const { user, isLoading: isUserLoading } = useUser()
+  isLoading = isLoading || isUserLoading
```
KenLSM: /nit would prefer renaming the props `isLoading: _isLoading` so that we don't lose the passed in props.

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -506,6 +519,78 @@ export const createForm = (
...
+
+    const isApprovalFieldInEditFields = workflow.every(
+      (step) =>
+        !step.approval_field ||
+        (step.edit && step.edit.includes(step.approval_field)),
+    )
+    if (!isApprovalFieldInEditFields) {
+      return errAsync(
+        new MalformedParametersError(
+          "For all steps, the selected approval field must also be included in same step's edit fields",
```
KenLSM: /nit this message sounds very technical. Suggestion: "Approval fields must be included in edit fields." since admins must save on every workflow step. The context of which step can likely be omitted without loss of specificity.

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.controller.spec.ts
```diff
@@ -43,17 +54,639 @@ describe('multirespondent-submission.controller', () => {
...
+      })
+      const mockRes = expressHandler.mockResponse({})
+
+      // Act
+      await updateMultirespondentSubmissionForTest(mockSubmitMrfReq, mockRes)
+      // Assert
+      // approval email is sent instead of completion email
+      expect(sendMrfApprovalEmailSpy).toHaveBeenCalledTimes(1)
+      expect(sendMrfWorkflowCompletionEmailSpy).not.toHaveBeenCalled()
+      expect(sendMRFWorkflowStepEmailSpy).not.toHaveBeenCalled()
```
KenLSM: Did a bit of rabbit-holing on investigating this. I noticed that we're spying through to `MailService` which is something that is usually called by `XService`[1] rather than `XController`.

What triggered this investigating was the difficulty of mocking the spies. As you may have noticed, `updateMultirespondentSubmissionForTest` doesn't actually calls what `sendMRFWorkflowStepEmail` was spying on, but through another interim function on the controller level `sendNextStepEmail`.

Arranging the control flows: 

1. `XController.updateMultirespondentSubmission -> XService.sendMrfOutcomeEmails -> MailService.XXX`
2. `XController.updateMultirespondentSubmission -> XController.sendNextStepEmail -> MailService.sendMRFWorkflowStepEmail`

**Suggestion**
1. Move `XController.sendNextStepEmail` to `XService.sendNextStepEmail`.
2. Update spies over to monitor `XService` instead of `MailService`
3. Spying of MailService can be done on `XService.spec`

**Outcome**
Generally less flakey tests since we're not testing across two different boundaries, i.e., From `Controller -> IntermediaryFoo -> Actual Spy` to `Controller -> Actual Spy`

[1] with some exception such as verification OTP, even if so it can also be argued that the `LoginService` should be calling instead of `LoginController`.

KenLSM: Also identified several additional code smell (which I'm also guilty of) [here](https://linear.app/ogp/issue/FRM-1858/encrypt-submissioncontrollerts-should-not-call-save). Something that we should address later on.

Want to just align on the code smell and the impact of such flows.

KenLSM (changes requested):
Minor issues here and there but generally no big issue with the overall design of the approval feature. Discovered some debts, but not attributable to this PR. 

Great work @kevin9foong !

src/app/modules/core/core.errors.ts
kevin9foong: i initially changed the error code for the email mode error so that email mode submission errors will use the code range 100250 - 100299 and for the generic submission errors to follow incrementing order, especially since email mode may be deprecated leaving a gap in the error code order. 

however, perhaps it is fine for the SUBMISSION_EXPECTED_RESPONSE_NOT_FOUND error to skip 1 number and be set as 100224.

kevin9foong: i initially changed the error code for the email mode error so that email mode submission errors will use the code range 100250 - 100299 and for the generic submission errors to follow incrementing order, especially since email mode may be deprecated leaving a gap in the error code order.

however, perhaps it is fine for the SUBMISSION_EXPECTED_RESPONSE_NOT_FOUND error to skip 1 number and be set as 100224.

KenLSM: Yeah ok to skip 1. There's no need for them to be gapless incremental. It just happened to be that way.

We'll generally prefer keeping the old error codes even if they are not utilised as they are already being referenced in the older logs. Updating what the error codes mean (or having multiple meanings to it) makes it hard to traceback in the future. For this scenario we'll prefer to optimise for easier debugging!

kevin9foong: for logging of rationale: 
Ken: 
Yeah ok to skip 1. There’s no need for them to be gapless incremental. It just happened to be that way.
We’ll generally prefer keeping the old error codes even if they are not utilised as they are already being referenced in the older logs. Updating what the error codes mean (or having multiple meanings to it) makes it hard to traceback in the future. For this scenario we’ll prefer to optimise for easier debugging!

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/ApprovalsBlock.tsx
kevin9foong: yes, this is being used: 
- selectedApprovalField is used to show and enable/disable the toggle upon load from db (ie, when the admin clicks on the edit step and visits the activeStepBlock)  
- selectedEditFields is used for validation of the react-hook-form state before allowing submit

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/RespondentBlock.tsx
kevin9foong: removed!

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/FieldLogicBadge.tsx
kevin9foong: thanks for highlighting this! didnt realize there was this prop!

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlockContainer.tsx
kevin9foong: thanks for catching that `borderTopColor` can be removed!  

For the others, it is still being used.
For more elaboration: 
- without `px` 
image

- for `spacing`, it is used to set the default minimum 16px spacing between the items, unless explicitly overriden by components which want more spacing.

frontend/src/features/admin-form/create/workflow/hooks/useAdminFormWorkflow.ts
kevin9foong: changed to 
```
const formWorkflow =
    form?.responseMode === FormResponseMode.Multirespondent
      ? form.workflow
      : null
```

resolved!

src/app/modules/form/admin-form/admin-form.middlewares.ts
kevin9foong: i added it as a means to harden our request validation security by reducing the number of strings that would pass our validation. 

though if preferred, i can remove this as well which may have the benefit of having less things to update if we change from using `hex().length(24)` to something else

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.controller.spec.ts
kevin9foong: let me think about this!

kevin9foong: Hi Ken, I've spent some time thinking about and understanding the concerns raised. Let me address them: 

- Regarding spying through to `MailService` being a code smell: 
I think this is a debate between the [tradeoffs of sociable vs solitary unit testing](https://medium.com/clarityai-engineering/sociable-or-solitary-unit-tests-choose-your-tradeoffs-7ced14d4baef).
image

[Source](https://martinfowler.com/articles/microservice-testing/#testing-unit-introduction)

In this case, the unit tests above are meant to be sociable unit tests, where the controller is the interface and we are testing the business functionality as a whole, treating the impl details as a black box.  
<br/>
To elaborate, here is my justification for advocating for sociable unit tests over writing per layer solitary unit tests for Approvals (these ideas are also referenced in the link above): 

1. Emulates acceptance tests and less prone to change: Implementation details may change due to refactoring (eg, what service layer function is invoked and arguments etc), however, business functionality remains the same and is the main behavior we are able to assert to prevent regressions.  
2. Save developer time: The above sociable UT are sufficient to cover all acceptance tests. It saves a lot of time (thinking about other features as well) and prevents regressions. 
3. May be more intuitive from another code reader's perspective to understand what each TC's purpose since it relates to acceptance tests: For eg, `sends approved outcome email when mrf is approved for last step of multiple step MRF` rather than `passes isApproval true and isRejected true to sendAppropriateOutcome when mrf is approved for last step of multiple step MRF` might be easier to understand. 

- Difficulty in mocking spies: 
Spies may be mocked by doing `sendMrfWorkflowCompletionEmailSpy.mockImpl(<impl here>)`. It might be similar even if we changed to solitary UT. 

- Regarding test flakiness: 
I believe you may be referring to the TC failing in the event the implementation details are changed. 
This is a strong justification for writing sociable UT which assert business outcome over per layer unit tests. 
The business functionality is unlikely to change (unless the PRD changes, which TC code changes are then inevitable either way). Writing the sociable UT above asserts no regressions to functionality while avoiding need to change due to implementation detail changes. 

Therefore, it might not be needed to split into 2 layers and test them solitarily and we should apply sociable UT for the reasons above. Please let me know your thoughts on this.

kevin9foong: Thinking about it further, the code smell might not be the test cases itself, but rather the `multirespondent.controller` itself which also partakes in invoking business logic such as `submission.save()` and checking if the step is rejected etc. 

If a change is desired, a more appropriate fix might be to refactor the controller and shift all business logic to the service layer, then apply the written sociable UT to the service layer instead (though im not sure if there's much value in doing so atm but can be done). 

The controller should only contain code which checks the http request (eg `ensurePipeline`) before invoking the appropriate service function.

frontend/src/features/admin-form/create/workflow/hooks/useAdminFormWorkflow.ts
kevin9foong: Seems like setting it to `null` breaks the workflow page, investigating 
image

kevin9foong: it seems if `null` is used instead of undefined, 

```
  const {
    yesNoFormFields = [],
    formWorkflow = [],
    isLoading,
  } = useAdminFormWorkflow()
  ```
  
  no longer sets `formWorkflow` to `[]` when null since default values only apply when the value is `undefined`. 
  hence, i will need to do an explicit check `if(!formWorkflow) formWorkflow = []` to emulate it. 
  
  thinking about it, perhaps `undefined` may be more precise since it is never defined for non-mrf forms instead of defined as `null`. hence, i will revert back to `undefined`.

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.controller.spec.ts
KenLSM: 1. Correct. The code smell that's being referred to is indeed on the structure of controller, which is discovered by looking at what the tests are doing. I've tagged tests as it's an indicator of code smell, not where the root cause of it is at. The root cause is as what you've also derived, how the controller is handling clear business logic relating to how MRF handles outcomes / notification -- which is also what I've suggestion in my comment.
2. When discussing about social vs solitary tests, it's more important to look at what we're testing. Don't want us to be applying theories blindly. The what here refers to controller, and we're specifically discussing about `.controller.specs.` which is likely that our intention is to check for correctness of response to the client. Not about what the service (business layer) do. For controller, we generally want to check on the response to the client, and that it calls to the service directly. There's no need to bring social vs solitary tests into the debate since the problem is on what you're trying to test, not how large your test scope is.
3. "business functionality is unlikely to change". This is actually the part that moves more, coding architecture change less than business functionalities. 
4. "Save developer time". I don't think that's true. I have seen too many times of bad test codes which indicate an issue that's way too broad rather than specific. i.e., "somewhere in this 3-4 step flow went wrong" vs "this went wrong", which is easier to identify?

In general, we do have a broad range of really sociable tests (e2e, services), and solitary tests (utils).

frontend/src/features/admin-form/create/workflow/hooks/useAdminFormWorkflow.ts
KenLSM: > since it is never defined for non-mrf forms

Makes sense. Just want to highlight that there's a difference here.

KenLSM (approved):
LGTM!

PR #7659, opened 2024-09-10
fix: check existence of response id in attachment map for body v3 

kevin9foong:
## Problem
Closes FRM-1844

## Solution
See ticket. 

**Breaking Changes** 

## Manual tests: 
TC1: Check that attachment forms still work for mrf (v3 body): 
- [ ] create mrf form with attachment, with >=2 steps to test it out 
- [ ] let first step respondent (R1) upload the file 
- [ ] respondent 2 (R2) download file and check it is correct R1 uploaded file 
- [ ] R2 changes file 
- [ ] submit and complete the workflow 
- [ ] assert that the uploaded file can be viewed by admin and is correct 

TC2: Check that attachment forms still work for mrf (v2.1 body): 
- [ ] Create storage form with single attachment field 
- [ ] submit form with attachment 
- [ ] verify admin can see uploaded attachment 

TC3: Check that no longer reproduce FRM-1844: 
- [ ] From TC2, submit a form and go to network tab to find the XHR req. 
- [ ] Copy the cURL (or any preferred format) req and change version to any arbitrary number >= 3. 
- [ ] send the request, verify response is: 
```
{
    "message": "Invalid file key - file key is not found in the quarantine bucket. The file must be uploaded first."
}
```
instead of `http status code 502`

### DIFF
--- a/src/app/modules/submission/receiver/receiver.utils.ts
+++ b/src/app/modules/submission/receiver/receiver.utils.ts
@@ -54,16 +54,16 @@ export const mapRouteError: MapRouteError = (error) => {
 }
 
 /**
- * Checks whether attachmentMap contains the given response
- * @param attachmentMap Map of field IDs to attachments
- * @param response The response to check
+ * Checks whether attachmentMap contains the given response id.
+ * @param attachmentMap Map of field ids to attachments
+ * @param response The response field id to check
  * @returns true if response is in map, false otherwise
  */
-const isAttachmentResponseFromMap = (
+const checkIsAttachmentResponsesIdInMap = (
   attachmentMap: Record<IAttachmentInfo['fieldId'], IAttachmentInfo>,
-  response: ParsedClearFormFieldResponse,
-): response is ParsedClearAttachmentResponse => {
-  return !!attachmentMap[response._id]
+  responseId: string,
+): boolean => {
+  return !!attachmentMap[responseId]
 }
 
 /**
@@ -82,7 +82,7 @@ export const addAttachmentToResponses = (
   // default to 0 for email mode forms where version is undefined
   // TODO (FRM-1413): change to a version existence guardrail when
   // virus scanning has completed rollout, so that virus scanning
-  // cannot be bypassed on storage mode submissions.
+  // cannot be bypassed on storage mode subscanAndRetrieveAttachmentsmissions.
   const isVirusScannerEnabled =
     (body.version ?? 0) >= VIRUS_SCANNER_SUBMISSION_VERSION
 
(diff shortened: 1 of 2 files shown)


src/app/modules/submission/receiver/receiver.utils.ts
```diff
@@ -82,7 +82,7 @@ export const addAttachmentToResponses = (
   // default to 0 for email mode forms where version is undefined
   // TODO (FRM-1413): change to a version existence guardrail when
   // virus scanning has completed rollout, so that virus scanning
-  // cannot be bypassed on storage mode submissions.
+  // cannot be bypassed on storage mode subscanAndRetrieveAttachmentsmissions.
```
KenLSM: typo on "subscan"?

kevin9foong: didnt intend to make this change, must have accidentally overwritten it.

src/app/modules/submission/receiver/receiver.utils.ts
```diff
@@ -113,7 +114,10 @@ export const addAttachmentToResponses = (
   if (isBodyVersion3AndAbove(body)) {
     Object.keys(body.responses).forEach((id) => {
       const response = body.responses[id] as ParsedClearFormFieldResponseV3
-      if (response.fieldType === BasicField.Attachment) {
+      if (
+        response.fieldType === BasicField.Attachment &&
+        checkIsAttachmentResponsesIdInMap(attachmentMap, id)
+      ) {
```
KenLSM: Could we use this instead?

```suggestion
      if (response.fieldType === BasicField.Attachment && id in attachmentMap) {
```

If this serves the same purpose, then we might be able avoid changing the function shape of `isAttachmentResponseFromMap()`.

kevin9foong: hi @KenLSM, 

i've thought about this and the reason why I changed both if clauses to use the same `checkIsAttachmentResponsesIdInMap` function is so that the logic between both functions are consistent, with exception of the json schema (`response.answer.filename` for v3 vs `response.filename`). 

When i was debugging this issue, I was looking at the 2 if clauses for versions <3 and >=3 and trying to find out the difference, which I figured was just the json schema between the diff versions. 
Hence, i think changing both to use the same `checkIsAttachmentResponsesIdInMap` has the benefit of conveying this idea to future developers. 

As for renaming the function from `isAttachmentResponseFromMap` to `checkIsAttachmentResponsesIdInMap`, the rationale is to let function's names start with verbs for better readability. Hence, I took the opportunity to enforce some code style practices at the same time. 
From Oracle's style guide: image

Ultimately, `id in attachmentMap` works as well and I'm good with using the above if desired, but i do believe the above approach has some of the above benefits mentioned.

kevin9foong: Alternatively, i can remove the function entirely and just do `id in attachmentMap` and `response._id in attachmentMap`

KenLSM: 1. Correct to note that functions should generally start with verbs, however for predicates (things that returns boolean that is meant to be used in if-else blocks), we don't use verbs.
2. This verb usage should follow conventions for specific languages. I.e., in javascript `toString()` isn't `transformToString`. It is better to have a uniform codebase (i.e., what is already present and reasonable) than to follow style guides resulting in multiple different "styles" in the codebase.

KenLSM (approved):
LGTM with approach!

PR #7663, opened 2024-09-11
test: mock date when rendering shimmer dates

KenLSM:
## Problem

Reduces flakiness of tests as the shimmer text's width is captured as a storybook snapshot. This means that if the "shimmered" dates have a different width as we get different `date()` values, storybook would highlight this a UI change. This is a low value diff, basically a noise.

## Solution

Mock the dates so that shimmer width is always the same between snapshots.

**Breaking Changes** 


kevin9foong:
This is a great fix and would remove the ui test noise we experience every single time a PR is made.

kevin9foong (approved):

PR #7665, opened 2024-09-11
chore: remove ga injection to ejs

KenLSM:
## Problem

`GA_TRACKING_ID` is no longer injected as a frontend variable in React, only in Angular. Remove all occurrences once the rollout is complete and we no longer use the Angular app

Closes FRM-231

## Solution

React uses a different method to inject `GA_TRACKING_ID`.

**Breaking Changes** 


kevin9foong (approved):
thanks for removing unused google analytics!

PR #7671, opened 2024-09-12
fix: change #black to #000000 in singpass logo

kevin9foong:
## Problem
Singpass logo #black is not recognised in svg. 

## Solution
Change #black to #000000 hex color 

**Breaking Changes** 


KenLSM (approved):
Thanks for the quick fix!

PR #7676, opened 2024-09-13
chore: remove unused deps

KenLSM:
## Problem

More unused deps found. Some of them are flooding the PR list as they are opened by dependabot.

Also raises false positive vuls.

## Solution

Identify and remove them.
Previously referenced by [`src/public/polyfills.js`](https://github.com/opengovsg/FormSG/blob/v5.74.0/src/public/polyfills.js) which we no longer use.
- fetch-readablestream
- abortcontroller-polyfill
- whatwg-fetch
- corejs
- web-streams-polyfill

Previously referenced in [`src/public/main.css`](https://github.com/opengovsg/FormSG/blob/v5.74.0/src/public/main.css) which we no longer use.
- font-awesome 

**Breaking Changes** 


kevin9foong:
Thanks for removing these!

kevin9foong (approved):

PR #7685, opened 2024-09-16
chore: remove unused deps

KenLSM:
## Problem

Removing unused deps that are false positively marked as critical. Those packages are not in used.

Generally, we should remove deps that are no longer used so as both reduce the noise by audits and as well as reducing the possible area of attack. 

## Solution

Check and remove packages that are not in used. 

**Removed**
- `selectize`, `sortablejs`: used during angular times
- `coveralls`: fully transitioned to gh actions
- `@opengovsg/credits-generator`: should be executed through npx instead of installed as devDeps

**Breaking Changes** 


kevin9foong (approved):
Thanks once again for removing all these unused deps!

PR #7694, opened 2024-09-18
chore: add comments clarifying potentially confusing terms

KenLSM:
## Problem

Confusing terminology while reviewing. 

## Solution

Add comments to clarify context that was not explicitly noted. Context was found only after reading the source PR.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/admin-form/template/UseTemplateModal/UseTemplateModal.tsx
+++ b/frontend/src/features/admin-form/template/UseTemplateModal/UseTemplateModal.tsx
@@ -16,6 +16,10 @@ export type UseTemplateModalProps = Pick<
   'onClose' | 'isOpen'
 > & { formId: string }
 
+/**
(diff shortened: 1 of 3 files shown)


kevin9foong:
Thinking about removing the use of the word 'use' in 'useTemplate' or changing it another word eg,  `utilizeTemplate`  instead to reduce need for comments as much as possible.

kevin9foong (approved):

KenLSM:
> Thinking about removing the use of the word 'use' in 'useTemplate' or changing it another word eg, `utilizeTemplate` instead to reduce need for comments as much as possible.

It's an interesting topic. "Use Template" is a business / product term that might cause confusion if we swap it out to "Utilize Template". I.e., PM / Design / Users refer to them as "Use Template" while on the code POV, it is codified as `utilizeTemplate` and the reason for that is because React chose to abuse the term `use`. 😄

PR #7718, opened 2024-09-25
feat: mrf be field validation

kevin9foong:
## Problem

BE validation is currently not performed for MRF submissions. This allows for invalid responses to be saved for MRF submissions. 

For example: 
- duplicate options selected for checkboxes
image

Closes FRM-1688

## Solution

Add validator for V3 responses and use this V3 validator in the MRF middleware which uses V3 responses. 

**Breaking Changes** 

## Tests

Regression test: 
- [ ] Create an MRF form with all types of basic fields, try a combination of different field settings.
- [ ] Create a table with dropdown field and text field col. 
- [ ] Fill in with valid values and ensure that we can still submit all fields without validation error. 

Setup for next few tests: 
- [ ] Set all fields in earlier MRF form to optional to prepare for the next few test cases.
- [ ] Go to web console (inspect -> find the XHR request -> copy as cURL), save this in a text editor to create invalidate submissions  

Checkbox field validation - no duplicates with others: 
- [ ] Create a option called `Others: A` 
- [ ] Enabled others 
- [ ] Submit the form with checkbox response selecting the option `Others: A` and a custom `A` value in others
- [ ] Should return a validation failure, which matches Storage/email mode validation. 

Table field validation: 
- [ ] From the cURL command in setup, add row to `table` field type value despite `addMoreRows` disabled, it should reject. 
- [ ] From the cURL command in setup, add row to `table` field type value, enable `addMoreRows`, it should accept.
- [ ] From the cURL command in setup, add more rows than `maxRowsAllowed` to `table` field type value with `addMoreRows` enabled, it should reject.  
- [ ] Make text field required, set the textField value in row to `''` empty string, it should reject. 
- [ ] Change dropdown to invalid value, it should reject. 
- [ ] Set text field and dropdown to not required, send empty string values for both, it should accept. 

Short text length enforced: 
- [ ] From the cURL command in setup, change the value of 'textfield' to less than the minimum text length set and submit, it should reject. 

Children field not allowed in MRF: 
- [ ] From the cURL command in setup, change the fieldType of any to `children` and submit, it should reject. 

Decimal field range is enforced: 
- [ ] From the cURL command in setup, change the value of  `decimal` fieldType to beyond the valid range and submit, it should reject. 

Short text length is enforced: 
- [ ] From the cURL command in setup, change the value of  `short` fieldType to beyond the valid range and submit, it should reject. 

Valid country/region only: 
-  [ ] From the cURL command in setup, change the value of  `country_region` fieldType to invalid value and submit, it should reject. 

NRIC must be valid: 
-  [ ] From the cURL command in setup, change the value of  `nric` fieldType to invalid nric value and submit, it should reject. 

Invalid field types not allowed: 
-  [ ] From the cURL command in setup, change the value of  fieldType to invalid field type eg, `abc` and submit, it should reject.


kevin9foong:
Pending UT. 

The core code is ready for review.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -4,6 +4,8 @@ import { NextFunction } from 'express'
 import { StatusCodes } from 'http-status-codes'
 import { err, errAsync, ok, okAsync, Result, ResultAsync } from 'neverthrow'
 
+import { FormFieldSchema } from 'src/types'
```
KenLSM: IIRC, this would likely fail as BE cannot use absolute imports

kevin9foong: resolved!

shared/constants/form.ts
```diff
@@ -91,6 +91,9 @@ export const PAYMENT_VARIABLE_INPUT_AMOUNT_FIELD_ID =
 export const CLIENT_RADIO_OTHERS_INPUT_VALUE =
   '!!FORMSG_INTERNAL_RADIO_OTHERS_VALUE!!'
 
+export const CLIENT_CHECKBOX_OTHERS_INPUT_VALUE =
```
KenLSM, line 95: Didn't see this being used by FE. Is it used somewhere? Saw this value is `CHECKED`, but not `SET`.

kevin9foong: In the FE, it is called `CHECKBOX_OTHERS_INPUT_VALUE` which is appended to checkbox value array when others option is checked. (See: `frontend/src/templates/Field/Checkbox/CheckboxField.tsx` L121) 

I used `CLIENT_CHECKBOX_OTHERS_INPUT_VALUE` to follow the convention of the above `CLIENT_RADIO_OTHERS_INPUT_VALUE`. ideally, both FE and BE will use the same shared variable, but did not do so for now as it would make sense to refactor the `CLIENT_RADIO_OTHERS_INPUT_VALUE` also together and did not want scope creep given this is a >6k LoC PR. 

Note that the `othersInput` is always sent to the BE regardless of whether the others option is checked as long as the others input text box is filled. Not ideal behavior imo but not really affecting the functionality of the app, hence i am not aiming to fix this in this PR to avoid scope creep also. 

Additional context: This behavior where `othersInput` is always sent to the BE was also appearing for radio button, which caused the bug [invalid submission responses for radio button when others input filled](https://linear.app/ogp/issue/FRM-1868/mrf-radio-bug-or-invalid-submission-responses-collected-for-radio) and has been hotfixed as it was causing responses to be submitted incorrectly warranting an immediate fix.

KenLSM: ~Thoughts on connecting these two values together by the same reference?~

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -513,8 +577,12 @@ export const validateMultirespondentSubmission = async (
                 ).map(() => undefined)
               })
               .andThen(() =>
-                // TODO: Step 4: Validate each field content with each field's validator rules individually.
-                ok(undefined),
+                validateMrfFieldResponses({
```
KenLSM, line 520: Since we're potentially blocking submissions here and we don't have full confidence that we're making the right blocks. Would recommend us to implement this through a "shadow"[1] instead.

Lmk if you have similar concerns on the confidence of the validation implementation.


[1] We log that a validation has failed, instead of block. Once we have sufficient information and confidence, we can officially swap that to a block.

src/app/utils/field-validation/index.ts
```diff
@@ -271,3 +275,223 @@ export const validateField = (
...
+          Object.values(rowObject).every((value) => value === ''),
+        ),
+      )
+    case BasicField.Attachment:
+      return ok(
+        response.answer.filename.trim() !== '' || // filename is defined only if there is a file uploaded for the response
+          response.answer.answer.trim() !== '' ||
+          !!response.answer.content,
+      )
+    case BasicField.Children:
```
KenLSM, line 340: Didn't see this in part of non-v3, this is a new check?

src/app/utils/field-validation/index.ts
```diff
@@ -271,3 +275,223 @@ export const validateField = (
...
+  response: ParsedClearFormFieldResponseV3
+  isVisible: boolean
+  formId: string
+}): Result<boolean, ValidateFieldError> => {
+  if (isGenericStringAnswerResponseV3(response)) {
+    return ok(
+      (formField.required && isVisible) ||
+        response.answer.toString().trim() !== '',
+    )
+  } else if (response.fieldType === BasicField.YesNo) {
```
KenLSM: This function is structured with lots of `else-ifs` because that's how `isValidationRequired` is done previously? Looking if we should park this as an item to refactor.

src/app/utils/field-validation/index.ts
kevin9foong: it exists for non-v3 as well, but the check performed is `isProcessedChildResponse` instead.

src/app/utils/field-validation/index.ts
kevin9foong: refactored to using switch statements

kevin9foong:
Written all UT for test coverage:

Field types:
  | HeaderResponseV3 > dont need to test since header is not submitted.
  | RadioResponseV3
  | CheckboxResponseV3
  | AttachmentResponseV3
  | TableResponseV3
  | ChildBirthRecordsResponseV3 > dont need to test since not currently used.
  | YesNoResponseV3
  | EmailResponseV3
  | MobileResponseV3
  | NumberResponseV3
  | DecimalResponseV3
  | ShortTextResponseV3
  | LongTextResponseV3
  | HomeNoResponseV3
  | DropdownResponseV3
  | RatingResponseV3
  | NricResponseV3
  | UenResponseV3
  | DateResponseV3
  | CountryRegionResponseV3
  
TC for middleware to invoke validator:
- assert children field not supported for MRF
- assert isVisible is correctly detected

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: done, when a validation error for v3 occurs, ValidateFieldErrorV3 will be thrown which will be logged on DD and we can use that for monitoring for the next 1-2 weeks.

src/app/utils/field-validation/field-validation.guards.ts
```diff
@@ -103,8 +109,20 @@ export const isPossibleEmailFieldSchema = (
   return get(field, 'fieldType') === BasicField.Email
 }
 
-export const isVerifiableMobileField = (
```
KenLSM, line 106: Good cleanup!

KenLSM (approved):
LGTM!

PR #7721, opened 2024-09-25
chore: remove container naming on docker-compose

KenLSM:
## Problem

Having a specified `container_name` causes multiple instance launches of docker to be invalid as the container_names are non-unique.

## Solution

We're not using the `container_name`, thus it is safe to remove them.

**Breaking Changes** 

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -151,7 +151,6 @@ services:
 
   database:
     image: 'bitnami/mongodb:4.4'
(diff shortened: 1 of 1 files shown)


KenLSM:
Thanks @kevin9foong great catch on the coverall change. That shouldn't be inside. Likely an artifact from my coverall debugging that bled into this PR.

kevin9foong (approved):
lgtm!

PR #7722, opened 2024-09-25
fix: coverall allow empty lcov

KenLSM:
## Problem

We've turned on selectively skipping tests which may inadvertently causes no files to be marked as testable. This results in no coverage file being generated which `coveralls` uses to upload coverage statistics.

## Solution

Upgrade `coveralls` to enable options for us to `allow-empty` lcov.

**Breaking Changes** 


KenLSM:
Bypassing merge requirements as this is needed to unblock other PRs

PR #7726, opened 2024-09-26
chore: allow empty coveralls to pass

KenLSM:
## Problem

Typo-ed. We should enable `allow-empty` instead of dis-allowing in #7722

## Solution

**Breaking Changes** 


kevin9foong (approved):
lgtm!

PR #7737, opened 2024-10-01
fix: logic active step padding

kevin9foong:


KenLSM (approved):
LGTM

KenLSM:
Failures in CI Backend are due to coveralls, which has been fixed on `develop`. Cleared to ignore the two failures!

PR #7739, opened 2024-10-02
fix: ordering of arguments for bounce service functions

kevin9foong:
## Problem
When emails bounce, admins should get an sms to their provided contact number informing them that emails have bounced. However, they are not receiving it and the following errors occur.

Closes FRM-1867

## Solution
After investigation, the cause was that the order of arguments passed to the invoked function were mis-matched.  
Use object argument with named keys to prevent mismatch of arguments in future.

**Breaking Changes** 

## Tests

`sendBouncedSubmissionSms` and `sendFormDeactivatedSms`: Admins with email submissions bounced should receive SMS notification thru Postman
- [ ] Login as admin
- [ ] Update your emergency contact number through Profile > Emergency Contact > Verify
- [ ] Create an email form, set the email notification email targets to be invalid emails eg, `abc@xyz098.com` and `abcxyz!@open.gov.sg`  
- [ ] As a respondent, make a submission to form (since all emails are invalid and bounce, a 'critical' bounce where all emails to notify have bounced occurs which triggers the next 2 emails) 
- [ ] Ensure that `bounce` SMS is received on number and `form deactivated` SMS is received. (note that 2 sms should be received)

### DIFF
--- a/src/app/modules/bounce/__tests__/bounce.service.spec.ts
+++ b/src/app/modules/bounce/__tests__/bounce.service.spec.ts
@@ -413,24 +413,24 @@ describe('BounceService', () => {
       ).toHaveBeenCalledTimes(2)
       expect(
         MockedPostmanSmsService.sendBouncedSubmissionSms,
-      ).toHaveBeenCalledWith(
-        testUser.email,
-        String(testUser._id),
-        form._id,
-        form.title,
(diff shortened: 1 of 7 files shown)


KenLSM (approved):
LGTM! Thanks for cleaning this up as well!

PR #7754, opened 2024-10-07
fix: regression on dropdown input, workflow save action step component

KenLSM:
## Problem

Regression introduced from react migration that was detected from Chromatic.

1. DateRangePicker (super huge gap, deferring now cause we are not using a wide DatePicker)
2. Workflow Approvals gap too big (need fix)
3. RightElement overlap (need fix)

## Solution

1. N/A
2. Remove `mt` that is no longer required as dividers now properly provide margins.
3. Utilise `hasInputRightElement` props to accommodate right element spacing.

**Breaking Changes** 


kevin9foong (approved):
Lgtm!

PR #7755, opened 2024-10-07
fix(react): remove strict mode, temporarily

KenLSM:
## Problem

Due to useEffect calling twice. This is as intended by the React Core team to help discover potential useEffect bugs on our system where we're not properly handling unsubscribes.

Potential workaround is to remove <React.StrictMode> from root. However, this is not recommended to be a permanent fixture.

References FRM-1871

## Solution

**Breaking Changes** 

## Tests

`react-hook-form` submit should display errors on local development
- [ ] Create a storage mdoe form
- [ ] Add short field
- [ ] Open form
- [ ] Hit the submit button
- [ ] Observe that the error appears immediately

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -30,8 +30,15 @@ window.gtag = gtag
 // Init dayjs
 dayjs.init()
 
-createRoot(document.getElementById('root')!).render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-)
+/**
+ * TODO(FRM-1855): Disable strict mode for validation to work properly
+ * This is not meant to be a permenant solution as the need to disable this
+ * reveals that there could be an existing issue with useEffect cleanup that is prevent
+ * pre-react 18.
+ */
+// createRoot(document.getElementById('root')!).render(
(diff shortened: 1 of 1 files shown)


kevin9foong:
I'm trying to debug this issue. It seems that at mainly only affects the public respondent form validation. 

When strict mode is enabled, the `FieldContainer` component only renders once upon mount and does not re-render to reflect the react-hook form error. 

will continue debugging

kevin9foong:
The code that is causing this is L321 in 
`frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx`

```
export const MemoizedFieldRow = memo(FieldRowContainer, (prevProps, newProps) =>
  isEqual(prevProps, newProps),
)
```

kevin9foong:
Trying to debug this issue as well

Some observations/clues which might be useful: 
- For nric field, does not seem to re-render when onBlur or onSubmit. For radio buttons though, it does re-render. 
- printing the `error` object from `useFormState`, it is an empty object despite validation failure.

kevin9foong (approved):

PR #7763, opened 2024-10-09
fix(DateRangePicker): react migration date range gap

KenLSM:
## Problem

Huge gaps between `Input` fields is present on `DateRangePicker` component.

Closes FRM-1870

## Solution

Chakra `Input` no longer references width on the input-props level. We'll need to supply styling tweaks on `sx` prop instead.

**Breaking Changes** 

## Before & After Screenshots
|--------|--------|--------|
| DateRangePicker | Screenshot 2024-10-09 at 4 26 55 PM | Screenshot 2024-10-09 at 4 26 41 PM | 

## Tests

Regression
- [ ] Go to Results Page
- [ ] Observe that DateRangePicker component does not have huge gaps


kevin9foong (approved):
lgtm!

PR #7766, opened 2024-10-10
fix(workflow): step 2 email subtitle

KenLSM:
## Problem

Closes FRM-1874

## Solution

Set color to `secondary.400`

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

Screenshot 2024-10-10 at 9 01 32 PM

### DIFF
--- a/frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
+++ b/frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
@@ -129,6 +129,61 @@ export const Step2Empty = {
   },
 }
 
+export const Step2FixedEmailDefault = {
+  // due to the double registration of 'workflow_type' there would be a weird interaction
+  // where the default value will be reset
+  // thus we have to manually select the field again
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement)
+    await waitFor(
+      async () =>
+        expect(await canvas.getByText('Save step')).not.toBeDisabled(),
+      {
+        timeout: 5000,
+      },
+    )
+    await waitFor(
+      async () => {
+        await userEvent.click(await canvas.getByText('Fixed email(s)'))
+      },
+      {
+        timeout: 5000,
+      },
+    )
+  },
+  args: {
+    stepNumber: 1,
+    defaultValues: {
+      workflow_type: WorkflowType.Static,
+      emails: ['me@open.gov.sg', 'invalidemail'],
+    },
+  },
+}
+export const Step2FixedEmailEmpty = {
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement)
+    await waitFor(
+      async () =>
+        expect(await canvas.getByText('Save step')).not.toBeDisabled(),
+      {
+        timeout: 5000,
+      },
+    )
+    await waitFor(
+      async () => {
+        await userEvent.click(await canvas.getByText('Fixed email(s)'))
+      },
+      {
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
```diff
@@ -129,6 +129,61 @@ export const Step2Empty = {
   },
 }
 
+export const Step2FixedEmailDefault = {
+  // due to the double registration of 'workflow_type' there would be a weird interaction
```
KenLSM, line 133: This is still weird imo. But couldn't find good and easy solution.

kevin9foong:
Having a look, trying to understand the new chromatic test case

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
kevin9foong: im not really sure what it means to `manually select the field again`. I'm currently seeing an interaction test where we are 
- expecting save step button not to be disabled (perhaps for general coverage?) 
- clicks on the fixed emails (I can understand that this is to set the state of the component into the fixed emails toggle clicked) state for visual review 

can't seem to find the code that manually selects the field.

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
```diff
@@ -129,6 +129,61 @@ export const Step2Empty = {
...
+      emails: ['me@open.gov.sg', 'invalidemail'],
+    },
+  },
+}
+export const Step2FixedEmailEmpty = {
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement)
+    await waitFor(
+      async () =>
+        expect(await canvas.getByText('Save step')).not.toBeDisabled(),
```
kevin9foong, line 167: just wondering, why are we expecting `Save step` not to be disabled in this test case? 

in `Step4ApprovalFieldNotInEditErrorMessage` test, I see the same code, which i assume is to ensure that the `Save step` button is enabled before clicking on it. 
Would awaiting and exoecting that the `Fixed emails(s)` radio button exists be more appropriate?

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
KenLSM: > await userEvent.click(await canvas.getByText('Fixed email(s)'))

is the code that manually selects the field

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
KenLSM: The interactions of radio button is dependent on `isLoading` and this value is also represented by `Save button`.

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/EditStepBlock.stories.tsx
kevin9foong: I get it now, instead of setting it using the args, we need to use interaction tests to click on the static email radio manually

kevin9foong (approved):
lgtm

PR #7767, opened 2024-10-10
chore: add paperless form banner

KenLSM:
## Problem

We want to drive more input from our admins to help us understand the existence of paper forms. 

## Solution

Add an admin only infobox on the workspace page. We've chosen not to use a banner as it should only be used for important information.

**Breaking Changes** 

## Before & After Screenshots

See chromatic

## Tests

- [ ] Ensure that the link shows up on admin dashboard
- [ ] Ensure that the link goes to our paperless survey form link


kevin9foong (approved):
lgtm!

PR #7771, opened 2024-10-14
feat: no notify for steps after reject step

kevin9foong:
## Problem
For rejected workflows, steps after the rejection should not be notified when a rejection happens earlier in the chain.  

Closes FRM-1876 

## Solution
Do not notify subsequent steps if earlier step rejects.

**Breaking Changes** 

## Tests

TC1: Regression test for approve 
- [ ] Create MRF workflow with 3 steps  
- [ ] Set step 2 and step 3 as approval step 
- [ ] Set steps 2 and 3 as notification step 
- [ ] Approve form completely. 
- [ ] See that steps 2 and 3 email get approve email. 

TC2: No notify for steps after reject step
- [ ] From MRF from TC1 
- [ ] Reject at step 2
- [ ] Ensure only step 2 gets the reject email (Step 3 does not get an email) 

TC3: Regression test for reject 
- [ ] From MRF from TC1 
- [ ] Reject at step 3
- [ ] Ensure both step 2 and 3 gets the reject email

### DIFF
--- a/src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
+++ b/src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
@@ -137,9 +137,9 @@ describe('multirespondent-submission.service', () => {
       expect(sendMrfWorkflowCompletionEmailSpy).not.toHaveBeenCalled()
       expect(sendMRFWorkflowStepEmailSpy).toHaveBeenCalledTimes(1)
       // destination emails are correct
-      expect(
-        sendMRFWorkflowStepEmailSpy.mock.calls[0][0].emails,
-      ).toContainAllValues(expectedEmails)
+      expect(sendMRFWorkflowStepEmailSpy.mock.calls[0][0].emails).toEqual(
+        expectedEmails,
+      )
       expect(sendMRFWorkflowStepEmailSpy.mock.calls[0][0].emails.length).toBe(
         expectedEmails.length,
       )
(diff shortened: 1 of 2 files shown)


src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
```diff
@@ -463,9 +578,133 @@ describe('multirespondent-submission.service', () => {
...
+            value: 'not_expected_1@example.com',
+          },
+        },
+        [yesNoFieldId1]: {
+          fieldType: BasicField.YesNo,
+          answer: 'No',
+        },
+      } as FieldResponsesV3
+
+      const fourStepApprovalWorkflow: FormWorkflowStepDto[] = [
```
KenLSM: Wouldn't this be regarded as 5 steps?

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
```diff
@@ -463,9 +578,133 @@ describe('multirespondent-submission.service', () => {
...
+        },
+        {
+          _id: stepFiveId,
+          workflow_type: WorkflowType.Static,
+          emails: ['not_expected_3@example.com'],
+          edit: [],
+        },
+      ]
+
+      const currentStepNumber = 1 // 2nd step of 3 steps workflow
```
KenLSM:
```suggestion
      const currentStepNumber = 1 // 2nd step of 5 steps workflow
```

KenLSM (approved):
Couple of nits regarding comments, otherwise LGTM.

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
kevin9foong: yes, i added another step but forgot to update the name. Thanks for this catch!

PR #7772, opened 2024-10-14
chore: update port number for vite fe in docker compose yml

kevin9foong:
## Problem
Port number in docker.compose is using the previous localhost:3000, causing maildev links to fail. 

Closes FRM-1877

## Solution
Update to port 5173.  

**Breaking Changes** 


KenLSM (approved):
Thanks! LGTM

KenLSM:
Likely isn't the last of it, will have to sweep through our `.env.example` as well

PR #7781, opened 2024-10-15
feat: email response summary

kevin9foong:
## Problem
Admins want a summary of email responses in their workflow completion emails. This is necessary for them to effectively use MRF without having to log in to the portal. (eg, step 1 respondent can see the responses etc) 

Closes FRM-1879

## Solution

Add the summary of responses in the outcome and completion email.

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:

image

## Tests

Test workflow completion email: 
- [ ] Create an MRF form with a variety of form field types (decimal, short text, email, mobile number, table and attachment). Ensure at least one of the form fields is optional. 
- [ ] Set email notification to your email inbox. 
- [ ] Set a single workflow step to fill up all the forms. 
- [ ] Submit the single step to complete the MRF workflow. Leave the optional field blank. 
- [ ] Verify workflow completion email is received with correct fields. The blank optional field should be excluded from the responses summary. 

Test workflow approved email:
- [ ] Add a yes/no field and set it to second step in workflow. 
- [ ] Submit both steps. Ensure the optional field is now filled. 
- [ ] Verify the workflow approved email is received with the correct field values. The optional field should now be included in the email since it is filled. 

Test workflow rejected email: 
- [ ] Repeat the previous TC, but reject the workflow. 
- [ ] Verify the workflow not approved email is received with the correct field values.


shared/constants/form.ts
```diff
@@ -92,6 +92,9 @@ export const PAYMENT_VARIABLE_INPUT_AMOUNT_FIELD_ID =
 export const CLIENT_RADIO_OTHERS_INPUT_VALUE =
   '!!FORMSG_INTERNAL_RADIO_OTHERS_VALUE!!'
 
+export const CLIENT_CHECKBOX_OTHERS_INPUT_VALUE =
```
KenLSM: This should be in the MRF BE validation PR? -- i.e., is there a soft deps on that PR?

kevin9foong: yes it is, will resolve the merge conflict when merging into dev.

kevin9foong: need to rebase depending on if BE validation PR or this goes into `develop` first

kevin9foong: just squashed and merged BE val, will rebase now

kevin9foong: resolved

src/app/views/templates/MrfWorkflowCompletionEmail.tsx
```diff
@@ -1,72 +1,94 @@
...
+              <Heading style={headingTextStyle}>
+                {
+                  outcome 
+                    ? `The outcome for ${formTitle}` 
+                    : `${formTitle} has been completed by all respondents.` 
+                }
+              </Heading>
+              {
+                outcome ? <>
+                  <Text style={{...outcomeTextStyle, marginTop: '2.5rem', marginBottom: '0.25rem'}}>Outcome</Text>
```
KenLSM: Might be able to utilise nested `<Text />` here as child `<Text />` should inherit parent's styles. 
```
<Text style={outcomeTextStyle}>
  <Text style={{marginTop: '2.5rem', marginBottom: '0.25rem'}}>Outcome</Text>
  <Text style={{fontWeight: 400, marginTop: '0.25rem'}}>{outcome}</Text>
</Text>
```

src/app/views/templates/styles.ts
```diff
@@ -56,3 +56,69 @@ export const linkStyle: NonNullable<LinkProps['style']> = {
   wordBreak: 'break-all',
   color: '#445fcd',
 }
+
+const textStyles = {
```
KenLSM: I'm thinking that we should separate MRF styles out from the basic plain styles.

MRF emails (sent to MOP) are fancier while our other emails (to admins to inform about various failures) are less fancy.

src/app/views/templates/MrfWorkflowCompletionEmail.tsx
```diff
@@ -1,72 +1,94 @@
...
-          </Row>
-          <Row>
-            <Column>
-              <Text style={{ ...textStyle, fontSize: '14px' }}>
+              <Text style={{...titleTextStyle, marginBottom: '0.25rem'}}>Response ID</Text>
+              <Text style={{...bodyTextStyle, marginTop: '0.25rem'}}>{responseId}</Text>
+              { 
+                formQuestionAnswers.map((qa) => {
+                return <>
+                  <Text style={{...titleTextStyle, marginBottom: '0.25rem'}}>{qa.question}</Text>
```
KenLSM: I'm seeing a couple of repeats of `{...titleTextStyle, marginBottom: '0.25rem'}` on texts that represent the "question" side of the submission, and `{...bodyTextStyle, marginTop: '0.25rem'}` on texts that presents the "answer" side of the submission.

Would it make sense to make these two styles as constants?

src/app/views/templates/MrfWorkflowCompletionEmail.tsx
kevin9foong: Doing this in react email causes the style to not apply, need to apply each style on each "component" individually.

src/app/views/templates/MrfWorkflowCompletionEmail.tsx
kevin9foong: created a `questionMargin` and `answerMargin` style, allowing us to change the padding when needed immediately. 

changed from rem to px to support gsib.

src/app/views/templates/styles.ts
kevin9foong: seems a little odd to have a style sheet just for mrf emails. I'm thinking that when we release more emails in the future, the emails sent out should conform to the same style standard as the new mrf email forms to maintain consistent product quality to all users (regardless of MOP / admins).

src/app/views/templates/MrfWorkflowCompletionEmail.tsx
kevin9foong: though i understand your intention of reducing code duplication and using common variables to make future updates simpler, i have reduced duplications as much as possible.

src/app/views/templates/styles.ts
KenLSM: I get what you mean, but having `textStyles` and `textStyle` both in the same file brings more confusion than consistency if that's the play in the current moment.

kevin9foong: split it into `commonStyle` and `mrfCompletionEmailStyle` ts files 
- common is for reusable stuff that can be used across new email templates eg, body-1, h1, design system color hex codes 
- mrfCompletionEmailStyle is for specific styles for the email itself eg, container sizing, question and answer margins

kevin9foong: I'm thinking in the future, if more templates were to be added: 
- import the common styles from `commonStyle.ts` 
- create new template specific styles in the a `..specificComponentStyle.ts` file made for the new template

KenLSM: Sounds good! Likely this can also follow figma -- don't think we have figma versions of it for our older emails, and we're just starting to add chromatic.

KenLSM (approved):
LGTM

PR #7783, opened 2024-10-16
fix: set form fields to use formfielddto type

kevin9foong:
## Problem

See linear ticket. 

Closes FRM-1881

## Solution

See linear ticket. 

**Breaking Changes** 


KenLSM (approved):
LGTM. Paired to solve this.

PR #7786, opened 2024-10-17
test(emailpreview): chromatic support

KenLSM:
## Problem

1. Our email previews are not published to chromatic, thus it is difficult for our designers to review them without developers deploying to staging / production.
2. We are also unaware of unintended changes to emails without snapshot comparison.

Closes FRM-1884

## Solution

Adds support to publish to chromatic.

**Breaking Changes** 


kevin9foong (approved):
lgtm

PR #7787, opened 2024-10-17
fix: update the form field type for v3 test cases

kevin9foong:
## Problem
Due to the update in the type that the `fieldValidatorV3` argument accepts, the test case types are mismatched. 

Closes FRM-1881

## Solution
Update the formfield to use defaultFieldGenerator which casts the type to `FormFieldDto`

**Breaking Changes** 

### DIFF
--- a/src/app/utils/field-validation/validators/__tests__/attachment-validation.spec.ts
+++ b/src/app/utils/field-validation/validators/__tests__/attachment-validation.spec.ts
@@ -1,6 +1,7 @@
 import {
   generateAttachmentResponseV3,
   generateDefaultField,
+  generateDefaultFieldV3,
   generateNewAttachmentResponse,
 } from '__tests__/unit/backend/helpers/generate-form-data'
 import { mongo as mongodb } from 'mongoose'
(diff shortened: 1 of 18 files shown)


src/app/utils/field-validation/validators/__tests__/attachment-validation.spec.ts
```diff
@@ -219,9 +224,9 @@ describe('Attachment validation V3', () => {
 
   describe('Required or optional', () => {
     it('should disallow submission with no attachment if it is required', () => {
-      const formField = generateDefaultField(BasicField.Attachment, {
+      const formField = generateDefaultFieldV3(BasicField.Attachment, {
         attachmentSize: AttachmentSize.OneMb,
-      })
+      }) as FormFieldDto
```
KenLSM: Is this typecasting still required since `generateDefaultFieldV3` returns `FormFieldDto`

kevin9foong: I missed out the removal of this cast as that was my initial solution before deciding to make a common `generateDefaultFieldV3` function. Thanks for the catch!

KenLSM (approved):
LGTM!

PR #7788, opened 2024-10-17
chore: update email bounce template

KenLSM:
## Problem

Incorrect information is present on email bounce template, there is also a problem of multiple sources of truths which makes it difficult to maintain.

Closes FRM-1883

## Solution

- Update the bounced email notifications
- Update the guide to include most of the steps on how to prevent bounce
- Update and standardise all the short links to the right format (/formsg-guide) under [form@open.gov.sg](mailto:form@open.gov.sg), [go.gov.sg](http://go.gov.sg/) account.

**Technical notes**
1. Migrated emails from `ejs` over to `tsx` so that we can get storybook previews for PM and Design to review.
2. Removed assertions that overlapping with other test (i.e., mail content is already tested in separate test case)
3. Introducing snapshot testing for email content, generally a good choice for text-based or large bulks of contents that shouldn't change.

## Tests

Critical bounces should send email to admin with affected formid and list of recipients.
- [ ] Create an email mode form
- [ ] Set email response recipient to an invalid email (`invalidemailaddress@open.gov.sg`)
- [ ] Ensure that the only recipient is the invalid email
- [ ] Open the form
- [ ] Make a submission
- [ ] Observe that an email is sent to you with the correct formid and email recipient `invalidemailaddress@open.gov.sg`


kevin9foong (approved):
lgtm

PR #7799, opened 2024-10-21
chore: fix typo on runner tag

KenLSM:
## Problem

Our chromatic-frontend action is not running due to it waiting forever for the gh runner `ubuntu-latest\`. There is a typo on the stray `\` at the end of runner tag. GH does not detect if the runner name exist or not and thus the action keeps waiting for the non-existent runner to be available (which will never be since it doesn't even exist).

## Solution

Remove the typo.

**Breaking Changes** 


kevin9foong:
Thanks for fixing this! Was wondering why the chromatic was not running

kevin9foong (approved):

PR #7810, opened 2024-10-28
chore: remove legacy routes

KenLSM:
## Problem

No longer require these routes as we've migrated them.
Closes FRM-1895

## Solution

**Breaking Changes** 


tutntut (approved):
👍

PR #7817, opened 2024-10-29
chore: small letter

KenLSM:
## Problem

We use small caps for non-special nouns.

## Solution

Update `Notify Respondent ...` to `Notify respondent ...`

**Breaking Changes** 


tutntut (approved):
lgtm

PR #7831, opened 2024-10-30
chore: update test snapshot

KenLSM:
## Problem

Test snapshot is out of sync after the recent text update.

## Solution

Checked snapshot that output is right, updated snapshot.

**Breaking Changes** 


KenLSM:
Bypassing branch protection. Small snapshot changes.

PR #7838, opened 2024-11-04
fix: set table field required false to cols instead of whole field

kevin9foong:
## Problem

The table field itself is always `required: true` and its col fields can be `required: false`. 
However, the table field was allowed to be set to `required: false` by the AI response.  

Closes FRM-1899

## Solution

Apply the `required: false` to the column fields instead of the table field itself.

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:
 image

## Tests

TC1: not required table field applies to column fields only: 
- [ ] Use MFB to create a form - include `add a table field type that is not required` in the prompt. 
- [ ] Ensure that such a table field is created and the optional is applied to the column fields instead of the table field itself. 

TC2: required table fields work as usual 
- [ ] Use MFB to create a form - include `add a table field type that is required` in the prompt. 
- [ ] Ensure that such a table field is created and that the table field and its column fields are set to required.

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.assistance.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.assistance.service.ts
@@ -51,13 +51,13 @@ const mapSuggestedFormFieldToFieldCreateDto = (
       return {
         fieldType: BasicField.Table,
         title: tableFormField.title,
-        required: tableFormField.required,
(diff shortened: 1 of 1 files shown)


KenLSM (approved):
LGTM. Short and sweet.

PR #7840, opened 2024-11-04
chore: update react-email deps

KenLSM:
## Problem

Closes #7833, Closes #7834

## Solution

1. Updates several related packages.

- "@react-email/components"
- "@react-email/render"
- "react-email"

2. Explicitly install `react-dom` as `react-email/render` has removed `react-dom` from its deps, i.e., it is no longer transitively installed, thus we have to add it ourselves.

3. Wrapped `render` with `fromPromise` as it is now async.

**Breaking Changes** 


src/app/services/mail/mail.service.ts
```diff
@@ -1079,19 +1093,30 @@ export class MailService {
       responseUrl,
     }
 
-    const html = render(MrfWorkflowEmail(htmlData))
+    console.log(render(MrfWorkflowEmail(htmlData)))
```
kevin9foong: Stray `console.log` found

KenLSM: Removed in 1541326d6

package.json
```diff
@@ -26,9 +26,9 @@
   "scripts": {
     "postinstall": "npm run postinstall:frontend && npm run postinstall:shared",
     "test": "npm run test:backend && npm run test:frontend",
-    "test:backend": "env-cmd -f __tests__/setup/.test-env jest",
-    "test:backend:ci": "env-cmd -f __tests__/setup/.test-env jest --maxWorkers=2 --logHeapUsage --workerIdleMemoryLimit=0.2",
-    "test:backend:watch": "env-cmd -f __tests__/setup/.test-env jest --watch",
+    "test:backend": "env-cmd -f __tests__/setup/.test-env node --experimental-vm-modules node_modules/jest/bin/jest.js",
```
kevin9foong, line 29: Trying to understand this change - why did we need to add this experimental-vm-modules flag? based on my research, it has something to do with ESM vs commonJS modules - did the new bump in deps change to using ESM?

package.json
```diff
@@ -224,6 +226,7 @@
     "maildev": "^2.1.0",
     "mockdate": "^3.0.5",
     "prettier": "^3.3.3",
+    "react-dom": "^18.3.1",
```
kevin9foong, line 229: Trying to understand the usage of react-dom as well - I'm seeing usage of react-dom in some stories and `index.tsx` 
However, they dont seem to be importing from `react-email` previously, so im wondering why the removal of `react-dom` from `react-email/render` causes us to have to install `react-dom`?

kevin9foong (approved):
Just a few questions for my understanding, otherwise lgtm

package.json
KenLSM: `react-email@v1` introduced the use of dynamic imports. i.e., at runtime doing `requires('somePackage')`. This is something that jest isn't very happy with.

For non-jest usage, we're running them through a different bundler / transpiler. While for tests, we're directly running it through jest.

package.json
kevin9foong: Rationale: 
- react-email used to have react-dom as a dependency in v2.x.x but they switched it to a peer dependency in v3.0.2 (react-email uses react-dom internally)

Peer dependency means it is the responsibility of the 'peer' = aka FormSG in this case to install it as a dependency -> hence we add `react-dom` to our dependencies.

package.json
kevin9foong: for more context, react email v3.0.2 does an ES modules dynamic import syntax `const { default: reactDOMServer } = await import("react-dom/server");`. Since we are using jest, it does not support ES module syntax out of the box and requires the `experimental-vm-modules` flag to support this ESM syntax

PR #7848, opened 2024-11-05
fix: add debounce to break infinite call loop

KenLSM:
## Problem

Child record fields that uses combobox (aka dropdown) freezes the page at times. This makes the form completely unusable and cannot be submitted.

Found that there's a infinite `setValue` on the combobox which starves the thread causing the page to be unresponsive.

## Solution

Adding a debounce seemed to break the loop. This is a temporary fix as we have not found the root cause of why there's an infinite loop.

**Breaking Changes** 

## Tests

**Regression on myinfo child fields combo box data**
- [ ] Create a email form with myinfo child fields with child data of Race, Secondary Race, Vaccination Status 
- [ ] As an admin through the create builder
  - [ ] Select Race field and select "Afghan"
  - [ ] Ensure that the page is still interactive by scrolling up and down, clicking on other fields
  - [ ] Select Secondary Race and select "African 
  - [ ] Ensure that the page is still interactive by scrolling up and down, clicking on other fields
- [ ] As a respondent switch the browser to render under mobile view mode
  - [ ] Select Race field and select "Afghan"
  - [ ] Ensure that the page is still interactive by scrolling up and down, clicking on other fields
  - [ ] Select Secondary Race and select "African 
  - [ ] Ensure that the page is still interactive by scrolling up and down, clicking on other fields

### DIFF
--- a/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
+++ b/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
@@ -21,7 +21,7 @@ import {
   VisuallyHidden,
   VStack,
 } from '@chakra-ui/react'
-import { get } from 'lodash'
+import { debounce, get } from 'lodash'
 import simplur from 'simplur'
 
 import { DATE_DISPLAY_FORMAT } from '~shared/constants/dates'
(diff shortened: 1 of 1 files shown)


kevin9foong (approved):
lgtm

PR #7850, opened 2024-11-05
fix: handle string response in DTO for date validation object

kevin9foong:
## Problem

Frontend crashes when the below steps are produced -> preventing MRF with date fields with validation from being edited and submitted  
image

To reproduce:
1. create a date field with custom date range validation
2. set the date field as editable in step 2 of mrf
3. step 2 fails to submit due to the screenshotted error 

See linear ticket for more info. 

Closes FRM-1900

## Solution

Handle the case where the DTO is not a date object but rather a string. 

**Breaking Changes** 

## Tests

- [ ] Create an mrf with date field with custom date range validation.
- [ ] Create 2 steps in workflow, allow date field to be modified for both steps. 
- [ ] Submit the 2 steps, should be able to submit without FE crashing.

### DIFF
--- a/frontend/src/utils/date.ts
+++ b/frontend/src/utils/date.ts
@@ -46,16 +46,15 @@ export const normalizeDateToUtc = (date: Date | null) => {
 }
 
 export const loadDateFromNormalizedDate = (date: string | Date | null) => {
-  if (typeof date === 'string') {
-    const parsedDate = parseISO(date)
-    return new Date(
-      parsedDate.getUTCFullYear(),
-      parsedDate.getUTCMonth(),
-      parsedDate.getUTCDate(),
-    )
-  }
   if (!date) return date
-  return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
+  const parsedDate = typeof date === 'string' ? parseISO(date) : date
+  return new Date(
+    Date.UTC(
+      parsedDate.getFullYear(),
+      parsedDate.getMonth(),
+      parsedDate.getDate(),
+    ),
+  )
 }
 
 /**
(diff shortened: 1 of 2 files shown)


frontend/src/utils/date.ts
```diff
@@ -45,7 +45,15 @@ export const normalizeDateToUtc = (date: Date | null) => {
...
+  if (typeof date === 'string') {
+    const parsedDate = parseISO(date)
+    return new Date(
+      parsedDate.getUTCFullYear(),
+      parsedDate.getUTCMonth(),
+      parsedDate.getUTCDate(),
+    )
+  }
   if (!date) return date
   return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
```
KenLSM: /nit Would this be terser?

```suggestion
  if (!date) return date
  const parsedDate = typeof date === 'string' ? parseISO(date) : date
  return new Date(
    Date.UTC(
      parsedDate.getFullYear(),
      parsedDate.getMonth(),
      parsedDate.getDate(),
    ),
  )
  ```

**Review by KenLSM — APPROVED** [review]
LGTM. No issues with proposed implementation. One suggestion on structure.

File: frontend/src/utils/date.ts [hunk 2]
```diff
@@ -45,9 +45,16 @@ export const normalizeDateToUtc = (date: Date | null) => {
   return new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()))
 }
 
-export const loadDateFromNormalizedDate = (date: Date | null) => {
+export const loadDateFromNormalizedDate = (date: string | Date | null) => {
   if (!date) return date
-  return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
+  const parsedDate = typeof date === 'string' ? parseISO(date) : date
+  return new Date(
+    Date.UTC(
```
KenLSM, line 52: @kevin9foong could this be the reason for the date issue on calendar?

Saw that `.UTC` was added when previously it didn't. I.e., was swapped from local time to UTC

PR #7859, opened 2024-11-07
fix(calendar): incorrect date conversion

KenLSM:
## Problem

Closes FRM-1901

## Solution

**Breaking Changes** 

## Tests

**Regression Test**
#### Selecting dates on edges should be allowed
- [ ] Create a date picker field
- [ ] Configure Date validation "custom date  range" and 10 Nov to 20 Nov
- [ ] Ensure that picker can be selected on 10 Nov
- [ ] Ensure that picker can be selected on 20 Nov


kevin9foong (approved):
lgtm! thanks for the fix!

PR #7863, opened 2024-11-07
chore: log email on error

kevin9foong:
## Problem 

We still have lots of validation failure which seemed a bit too high for us to be confident in enabling hard-blocks.

## Solution

Collect more data by logging values that we would have blocked. This will be a temporary addition to assist us in debugging.


KenLSM (approved):
LGTM!

PR #7866, opened 2024-11-07
fix: drawer date picker invalid selection bug

kevin9foong:
## Problem

image
There is a mismatch in valid dates to select and the selection for custom date validation 

Closes FRM-1901

## Solution
**Explanation of bug:** 
The reason why there is a mismatch is because the date picker component fetches the valid start and end time from either the react client or server.

- for the client: we are using local time
- for the server: it is stored in UTC -> we need to convert to local time

the problem causing this bug is:
- when the drawer is extended, the start and end validation time is from the client which provides the date in local time, but is converted once again to local time
- when the server provides the dates: it is converted from utc -> sg time as expected.

**Fix:** 
The EditDate component input which is providing the date to the DateField should be normalized to UTC. 

**Breaking Changes** 

## Tests
date reflects correctly after custom date range selection: 
- [ ] Create a storage mode form with form field
- [ ] Select custom date validation and check the range on the form field builder is correct when the drawer is extended. 
- [ ] Save the form field setting. 
- [ ] Try to preview the form and check the date field. ensure the valid date range is correct.
- [ ] Try to submit the form and check the date field. ensure the valid date range is correct.

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDate/EditDate.tsx
@@ -136,8 +136,8 @@ const transformDateEditFormToField = (
     case DateSelectedValidation.Custom: {
       nextValidationOptions = {
         selectedDateValidation: inputs.dateValidation.selectedDateValidation,
-        customMinDate: inputs.dateValidation.customMinDate,
-        customMaxDate: inputs.dateValidation.customMaxDate,
+        customMinDate: normalizeDateToUtc(inputs.dateValidation.customMinDate),
+        customMaxDate: normalizeDateToUtc(inputs.dateValidation.customMaxDate),
       }
     }
   }


kevin9foong:
the provider loads -> loads -> click -> loads -> loads 

this causes the logic state to not update for some reason

kevin9foong:
Fixed by shifting to the api call instead of the provider.

KenLSM:
I would assume that the fix to be within the flow of `<EditDateField />` <> "Preview" version of `<DateField />` rather than the Provider which may have unexpected implications?

kevin9foong:
Archiving the old PR description here:

## Problem

image
There is a mismatch in valid dates to select and the selection for custom date validation 

Closes FRM-1901

## Solution
The reason why there is a mismatch is because the date picker component fetches the valid start and end time from either the react client or server.

- for the client: we are using local time
- for the server: it is stored in UTC -> we need to convert to local time

the problem causing this bug is:
- when the drawer is extended, the start and end validation time is from the client which provides the date in local time, but is converted once again to local time
- when the server provides the dates: it is converted from utc -> sg time as expected.

we are doing the conversion of the server utc time -> local time in the date field component itself
however, the concern of providing the time in FE valid format should be at the code performing the data fetching itself (Rationale: since the data being provided in UTC is an effect related to being from the server, the code fetching the data from the server should have the responsibility as well of converting the data to a client compatible form. This is similar to what the existing `transformAllIsoStringsToDate` is doing). 

My fix: 
The code has been refactored to do the utc -> local timezone conversion at the point by which the time is fetched from the server. This moves the concern of converting to utc to local time away from the date field component (which is only responsible for rendering the date field component given compatible data) and instead to the code fetching the time from server. 

**Breaking Changes** 

## Tests: 

- [ ] Create a storage mode form with form field
- [ ] Select custom date validation and check the range on the form field builder is correct when the drawer is extended. 
- [ ] Save the form field setting. 
- [ ] Try to preview the form and check the date field. ensure the valid date range is correct.
- [ ] Try to submit the form and check the date field. ensure the valid date range is correct.

kevin9foong:
After discussion with @KenLSM , i have changed the fix to a simpler and indeed more intuitive fix

KenLSM (approved):
LGTM

PR #7867, opened 2024-11-07
feat(i18n): add en-sg locale for AddRowFooter

LoneRifle:
## Problem and Solution
As part of a wider push for i18n on FormSG, extract text from AddRowFooter into a locale, rewording as needed to suit.

### DIFF
--- a/frontend/src/templates/Field/Table/AddRowFooter.tsx
+++ b/frontend/src/templates/Field/Table/AddRowFooter.tsx
@@ -1,7 +1,7 @@
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useState } from 'react'
+import { useTranslation } from 'react-i18next'
 import { BiPlus } from 'react-icons/bi'
 import { Box, Stack, Text, VisuallyHidden } from '@chakra-ui/react'
-import simplur from 'simplur'
 
 import Button from '~components/Button'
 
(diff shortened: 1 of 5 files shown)


frontend/src/templates/Field/Table/AddRowFooter.tsx
```diff
@@ -51,19 +42,34 @@ export const AddRowFooter = ({
...
+                count: Number(maxRows),
+              })
+            : t('features.publicForm.components.table.row', {
+                count: currentRows,
+              })}
         </VisuallyHidden>
 
         <Text aria-hidden textStyle="body-2" color="secondary.400">
-          {maxRowDescription}
+          {Number.isInteger(maxRows)
```
KenLSM: Would it make sense for L53-61 and L65-72 be a constant variable since both of them should have the same content and is likely to remain that way in the future.

KenLSM (commented):
LGTM. 1 suggestion to improve maintenance.

frontend/src/templates/Field/Table/AddRowFooter.tsx
```diff
@@ -51,15 +50,16 @@ export const AddRowFooter = ({
         type="button"
         onClick={handleAddRow}
       >
-        Add another row
+        {t('features.publicForm.components.table.addAnotherRow')}
         <VisuallyHidden>
-          to the table field. {maxRowAriaDescription}
+          {t('features.publicForm.components.table.addAnotherRowAria')}
```
KenLSM, line 55: Aria for maxRow seemed to be missing here 🫣

LoneRifle: Yea, deliberate decision on my part. I can readd it, though I just think that it would be a repetition of information, by the screen reader, esp given that further along, there is a `VisuallyHidden` component articulating in a complete sentence the number of rows

wdyt?

KenLSM: Understood. Agreed on repetition. No further issues!

KenLSM (approved):

PR #7871, opened 2024-11-08
fix: make children body input components controlled from the start

kevin9foong:
## Problem
Child record fields that uses combobox (aka dropdown) freezes the page at times. This makes the form completely unusable and cannot be submitted.

Discovered that children fix is implemented as an uncontrolled component but uses `setValue` which is a property of a controlled component.

Related fix #7856

## Solution

There are some bugs likely due to usage of controlled properties on uncontrolled component on a downstream library (https://github.com/downshift-js).

Elimination of this "mixed use" seemed to mitigate this issue.

**Breaking Changes** 

#### Regression on myinfo child fields combo box data
- [ ] Create an email mode form + all child fields + some other fields (eg, short text, checkbox) 
- [ ] Make a submission to the field on mobile as a respondent, try to fill up the different child fields and change children etc. 
- [ ] Repeat submission 2-3 times and ensure that the submission does not hang.

### DIFF
--- a/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
+++ b/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useMemo, useRef } from 'react'
 import {
+  Controller,
   FieldArrayWithId,
   FieldError,
   useFieldArray,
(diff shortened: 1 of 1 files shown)


KenLSM:
Hm, the solution doesn't seem to be explain why this change addresses the issue, nor is the issue described[1].

Sorry for the nitpicking, but bug patches can be tricky and usually gets revisited, so we want to capture as much information as possible as well as assumptions taken, if any.

[1] Having the wrong explanation for the solution might give the future reader the wrong interpretation of the fix.

KenLSM (approved):
LGTM. Updated your description to improve on the articulation and intent.

kevin9foong:
Thanks for updating my PR description for future clarity. Admittedly, this is something i can work on to improve and will do so in the future!

PR #7874, opened 2024-11-11
fix: add validation rules to controller

kevin9foong:
## Problem

After making the changes for myinfo child fields introducing controller in #7871, the new controllers do not include the required property `rules` which performs FE field validation. 

## Solution

Add the required `rules={validationRules}` property to the added controllers. 

**Breaking Changes** 

## Tests 
Assert that the FE validation for my info child sub fields work 
- [ ] Create an email mode form with my info child field 
- [ ] Include all myinfo child subfields 
- [ ] Submit the form, try to submit with missing / invalid values and ensure FE validation works for all subfields.


KenLSM (approved):
LGTM!

PR #7875, opened 2024-11-11
fix: add validation for myinfo child name field 

kevin9foong:
Continuation of [PR 7874](https://github.com/opengovsg/FormSG/pull/7874)

## Tests 
myinfo child fields are properly validated as required by the FE. 
- [ ] Create a myinfo child field with all subfields 
- [ ] Submit the form and leave it empty, ensure that the required FE validation error message is shown for all fields 
- [ ] select a child which does prefill -> ensure that the prefilled fields no longer have an error 
- [ ] submit and notice that secondary race which is not prefilled still has a required error 
- [ ] fill it up and submit -> ensure that it works fine

### DIFF
--- a/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
+++ b/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef } from 'react'
+import { useCallback, useEffect, useMemo } from 'react'
 import {
   Controller,
   FieldArrayWithId,
@@ -36,6 +36,7 @@ import {
 } from '~shared/types'
 import { formatMyinfoDate } from '~shared/utils/dates'
 
+import { REQUIRED_ERROR } from '~constants/validation'
 import { createChildrenValidationRules } from '~utils/fieldValidation'
 import { Button } from '~components/Button/Button'
 import { DatePicker } from '~components/DatePicker'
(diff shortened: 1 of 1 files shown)


frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
```diff
@@ -322,33 +315,28 @@ const ChildrenBody = ({
         <FormLabel gridArea="formlabel">Child</FormLabel>
         <Flex align="stretch" alignItems="stretch" justify="space-between">
           <Box flexGrow={10}>
-            <FormControl key={field.id} isRequired isInvalid={!!childNameError}>
+            <FormControl key={field.id} isRequired isInvalid={!!error?.[0]}>
```
KenLSM: Sudden use of hardcoded index access (i.e., `[0]`) looks suspicious. Should it be the `index` of the current child instead?

i.e., Child Component has two rows, the validation error is on 2nd child instead, would this `error?.[0]` pick up the error on the correct child?

kevin9foong: hi ken, seems like the hardcoding of the index where 0th element refers to the name field is a deterministic behaviour since a while ago. (see attached ss) 

image

exploring further, the index of the error message is fixed. if an error does not exist, then it will be an empty element. for eg, this screenshot shows a case when only the 5th field has an error and the others do not - the rest of the fields still keep the same index but are empty. 
image

hence, the above `Child Component has two rows, the validation error is on 2nd child instead, would this error?.[0] pick up the error on the correct child?` is guaranteed to not occur

was considering adding a `const NAME_INDEX = 0` and using it but not sure if it will add more clarity

kevin9foong: what i can do is define a constant to document this behavior so its clearer.

KenLSM: Yeah I think `const NAME_INDEX = 0` gives it more clarity, since I was also confused why it is always the first index. Explicitly calling it out makes sense.

kevin9foong: ill add it in then

KenLSM (approved):

PR #7878, opened 2024-11-12
fix: mrf be validation isSignatureTimeValid false positive

kevin9foong:
## Problem

False positives during MRF BE validation for verifiable fields where the signature is marked as invalid when >=step 2 submissions are made past the `isSignatureTimeRangeValid` valid transaction range.

Closes FRM-1902

### Technical details

`isSignatureTimeRangeValid` is verified based on two criteria:
1. "signature time" is before "submission time"
2. "signature time" is within a reasonable bound before "submission time"

**Examples**
- if the submission is made in year 2024, it does not make sense for the signature to be generated in year 2025.
- if the submission is made on 12 nov 2024 at 10am, it does not make sense for the signature to be generated (aka when the otp is verified) on 1st jan 2023.

This works for storage mode which only has 1 submission.

For MRF, we can have >=1 submissions (and hence >= 1 submission times) per 'submission': 
- When the signature is generated in step 1, we use the step 1 submission time to validate the signature which works
- When a respondent submits at step 2 however e.g., 2 days later, the submission time of step 2 violates criteria (2) defined earlier

Hence, the verification failed and the false positive occurs.

## Solution

**Proposal 1**
Store in a field the submission time of step 1 as part of the verify field or spoof a submission timing to be submission time + 1000ms. This allows the `isSignatureTimeValid` to pass.

**Proposal 2**
Skip verification if it is an existing field and the answer & signature is unchanged.
Since must have been validated in the earlier step.

**Proposal Chosen**
Proposal 2.
1. Does not introduce a "hack" which goes against the intention of `isSignatureTimeValid` check.
2. More intuitive. If a response has been validated in earlier step and has not changed, it does not have to be validated again.
3. Does not introduce additional "time" that we need to maintain. This requires more headspace for the reader understand.

**Breaking Changes** 

## Unit tests (Automated)  
Written to enforce that: 
- if no prev response eg, not an edit field for prev step or step 1 of form submission, then the validation runs. 
- if the value or signature changes, then a validation executes. 
- if prev response value and signature are both equal, no validation runs. 
Hence, these can be excluded from manual tests. 

## Tests

### DIFF
--- a/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -515,15 +515,18 @@ export const validateMultirespondentSubmission = async (
 
                     return ok(undefined)
                   }),
-                ).map(() => undefined)
+                ).map(() => {
+                  return previousResponses
+                })
               })
-              .andThen(() => {
+              .andThen((previousResponses) => {
                 // TODO: (FRM-1688) Set to block after sure that validation logic works as expected.
                 validateMrfFieldResponses({
                   formId,
                   visibleFieldIds,
                   formFields: form_fields,
                   responses: req.body.responses,
+                  previousResponses,
                 })
 
                 return ok(req.body.responses)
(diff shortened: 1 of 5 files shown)


KenLSM (approved):
lgtm

PR #7885, opened 2024-11-13
fix: replace trash icon in MRF steps

scottheng96:
## Problem
In MRF workflow, blocks have trash icon (top right) giving users ability to delete the step in workflow. For baseline practice in UX, destructive actions should be hidden below the fold 

Closes FRM-1905

## Solution
- Replace trash icon with pencil icon
- Update logic upon clicking icon to 'edit step' instead of 'delete step'
- First step has pencil icon now also, since users can't delete the first step, but can edit the first step

- Users will not be able to click anywhere on the box to enter editing step, only on clicking of the pencil icon
- Removed highlight/background change/cursor appearance when hovering the box

## Before & After Screenshots

| Subject | Before | After | 
| ------- | ------- | ----- |
|Trash icon replaced with pencil icon (first step included) |Screenshot 2024-11-13 at 10 39 57 AM|Screenshot 2024-11-13 at 10 36 47 AM|
|On click| Screenshot 2024-11-13 at 11 10 32 AM| Screenshot 2024-11-13 at 11 11 11 AM|
|Additional changes to behavior or edit/cursor||https://github.com/user-attachments/assets/6b45ec88-0467-4bd0-bdab-97d0e7fdee55|


KenLSM:
We can also remove `handleOpenDeleteModal` property from `type InactiveStepBlockProps` as `handleOpenDeleteModal` is no longer referenced in the component.

This also means that we have to cascade the removal of `handleOpenDeleteModal` property on the parent component of `<InactiveStepBlock />`

KenLSM (commented):
LGTM but holding the approval first as I'm getting PM to review the changes in-lieu of Design.

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/InactiveStepBlock/InactiveStepBlock.tsx
```diff
@@ -190,18 +179,19 @@ export const InactiveStepBlock = ({
...
           top={{ base: '0.5rem', md: '2rem' }}
           right={{ base: '0.5rem', md: '2rem' }}
           pos="absolute"
-          aria-label="Delete step"
+          aria-label="Click to edit"
           variant="clear"
-          colorScheme="danger"
-          onClick={handleOpenDeleteModal}
-          icon={<BiTrash fontSize="1.5rem" />}
+          colorScheme="secondary"
```
KenLSM, line 189: Checked with design, we should be using `primary` instead. We can omit this line since `colorScheme` defaults to `primary` if not defined.

KenLSM: Checked with design, we should be using primary instead. We can omit this line since colorScheme defaults to primary if not defined.

KenLSM (changes requested):

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/InactiveStepBlock/InactiveStepBlock.tsx
scottheng96: Made this change. Pencil is blue-ish now (primary)

KenLSM (approved):
LGTM!

PR #7890, opened 2024-11-14
fix(mail): failure to render mail

KenLSM:
## Problem

Calls to `@react-email/render` fails as `react-dom` was not found. 

This is due to #7840 adding `react-dom` into `devDeps` instead of `deps`.

## Solution

1. Added `react-dom` to dependencies.
2. Added additional logs to ensure that we are not swallowing errors from `render`

**Breaking Changes** 

##### Bounced emails are triggered
- [ ] Create an email mode form
- [ ] Set email recipients only contain invalid emails
- [ ] Open the form
- [ ] Submit the form
- [ ] Ensure that you receive a bounced email as the form owner


kevin9foong (approved):
lgtm!

PR #7894, opened 2024-11-15
fix(mrf be val): skip validation for unchanged mrf responses

kevin9foong:
## Problem
During MRF BE Validation, false positives occur for verifiable fields (due to signature being marked as invalid) and date fields with disallow past dates validation. 

**Details**
 - for signature fields, the reproduction steps and rationale is described in https://github.com/opengovsg/FormSG/pull/7878. 
 - for date fields with disallow past dates validation: 
 1. Step 1 of MRF can be submitted on `1st jan 2024`, a respondent for step 1 can select the current date `1st jan 2024` which is valid and passes the disallow past dates validation. 
 2. Step 2 is then submitted on `3rd jan 2024` and the respondent leaves the date field value from step 1 unchanged. This fails the disallow past dates validation since the step 2 submission date of `3rd jan 2024` is after `1st jan 2024`.
 
This edge case is considered acceptable and is similar to the current (as of this PR) frontend validation behavior. For example, if the admin tries to submit the form at step 2 where the date field with disallow past dates validation is editable, they would experience a frontend validation failure.  
   
Closes FRM-1909

## Solution

During field validation, there is a check if validation is required. 
An additional check for if the field response is changed is added. In the event the field is unchanged, then the validation for the field is skipped. This is based on the idea where since a response has been validated in the previous step and is unchanged, it does not need to be re-validated in the subsequent step. This resolves the false positive issue and replicates the same frontend validation behavior described above.   

**Implementation**
A new function `checkIsResponseChangedV3` is added to a file `field-validation.utils.ts` which allows for export without having to explicitly export a `checkIsResponseChangedV3ForTest` if added to the `index.ts` file. This function is used to determine if validation should be skipped. 
This function is exported and unit tested to ensure it correctly checks for changes for the various field types.  

**Breaking Changes** 


src/app/utils/field-validation/validators/dateValidator.ts
```diff
@@ -191,7 +191,9 @@ const pastOnlyValidatorV3: ResponseValidator<DateResponseV3> = (response) => {
   const answerDate = createMomentFromDateStringV3(answer)
 
   return answerDate.isAfter(todayMax)
-    ? left(`DateValidatorV3:\t answer does not pass date logic validation`)
+    ? left(
```
kevin9foong, line 194: The changes in this file are added to make the logs more specific - not related to the skipping of validation

src/app/utils/field-validation/validators/__tests__/email-validation.spec.ts
```diff
@@ -1052,131 +1052,4 @@ describe('Email field validation V3', () => {
       new ValidateFieldError('Invalid answer submitted'),
     )
   })
-
```
kevin9foong, line 1055: These TC are removed as they are migrated to the test for checkIsResponseChangedV3 instead, avoiding the need to spy and check if the spied function is invoked

src/app/utils/field-validation/field-validation.utils.ts
```diff
@@ -0,0 +1,95 @@
...
+      // eslint-disable-next-line no-case-declarations
+      const responseValue =
+        'value' in response.answer
+          ? response.answer.value
+          : 'othersInput' in response.answer
+            ? response.answer.othersInput
+            : null
+      return prevResponseValue !== responseValue
+    case BasicField.Checkbox:
+      if (prevResponse.fieldType !== response.fieldType) {
```
kevin9foong, line 56: this is required due to TS throwing an error as it fails to infer the type of prevResponse is same as response without it, despite being checked in L19 above.

KenLSM (approved):
LGTM!

PR #7903, opened 2024-11-18
feat: unify email notification for response copy for storage and email modes  

kevin9foong:
## Problem

The copy for the same functionality (email notifications for responses) on email and storage mode differ. 

Closes FRM-1911 

## Solution

Make the copy the same for both modes. 

**Breaking Changes** 


KenLSM (approved):
LGTM!

PR #7917, opened 2024-11-19
fix: fixed cursor appearance for inactive logic block

scottheng96:
## Problem
Similar to Inactive Workflow Block, cursor should not change to 'clickable' appearance upon hover; this behaviour should only occur when hovering over the <BiPencil> icon

## Solution
Adding pointerEvents="none"

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:

https://github.com/user-attachments/assets/6d4605a2-af62-4520-a39b-dee477734d00

### DIFF
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/InactiveLogicBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/InactiveLogicBlock.tsx
@@ -98,6 +98,7 @@ export const InactiveLogicBlock = ({
         borderColor="neutral.300"
         cursor={isPreventEdit ? 'not-allowed' : 'pointer'}
         aria-disabled={isPreventEdit}
+        pointerEvents="none"
       >
         <Stack
           spacing="1.5rem"
--- a/frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/InactiveLogicBlock.tsx
+++ b/frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/InactiveLogicBlock.tsx
@@ -96,9 +96,8 @@ export const InactiveLogicBlock = ({
         bg="white"
         border="1px solid"
         borderColor="neutral.300"
-        cursor={isPreventEdit ? 'not-allowed' : 'pointer'}
+        cursor={isPreventEdit ? 'not-allowed' : 'auto'}
         aria-disabled={isPreventEdit}
-        pointerEvents="none"
       >
         <Stack
           spacing="1.5rem"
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/InactiveLogicBlock.tsx
```diff
@@ -98,6 +98,7 @@ export const InactiveLogicBlock = ({
         borderColor="neutral.300"
         cursor={isPreventEdit ? 'not-allowed' : 'pointer'}
```
KenLSM: Can check if changing 'pointer' to 'auto' works as well? This might be able to allow us to have one single control on cursors.

KenLSM (changes requested):

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/InactiveLogicBlock.tsx
kevin9foong: Might be able to change the one in the inactive step block as well - i am testing this on local

kevin9foong:
```suggestion
        cursor={isPreventEdit ? 'not-allowed' : 'default'}
```

scottheng96: cursor="auto" seems to work as well. Which one shall we go with, I can make the same changes for the workflow cursor

scottheng96: Key difference:
auto: Lets the browser decide the cursor based on the element or its context.
default: Forces the cursor to be the default (usually the arrow) regardless of context.

Maybe we use default? Then we definitely know the appearance won't change

KenLSM: We usually take the [initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#formal_definition) to revert to the original value.

scottheng96: @KenLSM @kevin9foong

KenLSM:
Good catch!

frontend/src/features/admin-form/create/logic/components/LogicContent/InactiveLogicBlock/InactiveLogicBlock.tsx
scottheng96: 'Auto' it is, let me make that change.

KenLSM (approved):
LGTM!

PR #7922, opened 2024-11-19
chore(FE): cleanup date check on sex field

KenLSM:
## Problem

Date check was implemented on Sex field to support switch overs on a particular date. Since the date has already passed, there is no longer a need for us to check the date.

Closes FRM-1741

## Solution

Remove Infoboxes and permanently change update gender -> sex.

**Breaking Changes** 


kevin9foong (approved):
lgtm

PR #7926, opened 2024-11-20
feat(i18n): extract text for form metadata

LoneRifle:
## Problem and Solution
Extract labels for form metadata, namely last modified data and form type

[i18n]
- Add Meta section to `adminForm`

[features]
- Replace constants file with i18n equivalent mapping of type labels


KenLSM (approved):
LGTM

PR #7930, opened 2024-11-20
chore(FE): use consistent vite port for dev with README

littlemight:
## Problem
Frontend dev start script uses vite's default `5173` port, while README [states 3000 as FE port](https://github.com/opengovsg/FormSG/blob/develop/README.md#prerequisites).
Not touching `vite.config.ts`, afraid it might impact github job + deployment scripts.

## Solution
Update README to reflect the vite's port.

**Breaking Changes** 

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ npm run dev
 
 After the Docker image has finished building, the following local applications can be accessed:
 
-- React application can be accessed at [localhost:3000](localhost:3000)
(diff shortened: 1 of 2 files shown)


KenLSM:
@littlemight could we update the readme instead? Thinking that it would be easier and straightforward to follow the standard vite default port `5173`. 

Appreciate the help on keeping our README up to date!

KenLSM (approved):
LGTM!

PR #7937, opened 2024-11-23
fix: trim text input before sending to backend

kevin9foong:
## Problem

While the frontend validation performs trimming of the string input for text fields before validation, no trimming is done before the inputs are sent to the backend as response. 
This leads to validation failures reported by the backend despite passing the frontend validation.  

This is also a reason for false positives reported during MRF BE validation, where the untrimmed response input eg `   5chars   ` is evaluated as 5 characters during frontend validation and passes validation for exact 5 characters, but fails in the backend since there are 6 whitespace characters leading to 11 characters being found in the backend.

Closes FRM-1913

## Solution

Trim the text field inputs before sending the response to the backend. 

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -153,7 +153,7 @@ const transformFormInputCountryRegionToUpperCase = (
   )
 }
 
-// Text inputs are trimmed duirng validation and hence should also be trimmed before sending to the backend.
+// Trim text inputs before sending to backend to match frontend validation
 const transformFormInputTrimTextInputs = (
   formInputs: Record<string, unknown>,
   form_fields: Array<{ fieldType: BasicField; _id: string }>,
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -14,6 +14,7 @@ import { useDisclosure } from '@chakra-ui/react'
 import { datadogLogs } from '@datadog/browser-logs'
 import { useGrowthBook } from '@growthbook/growthbook-react'
 import { differenceInMilliseconds, isPast } from 'date-fns'
+import { flow } from 'lodash'
 import get from 'lodash/get'
 
 import {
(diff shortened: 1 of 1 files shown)


frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -123,6 +123,65 @@ export function useCommonFormProvider(formId: string) {
...
+      } else {
+        newFormInputs[fieldId] = currentInput
+      }
+      return newFormInputs
+    },
+    {},
+  )
+}
+
+// Text inputs are trimmed duirng validation and hence should also be trimmed before sending to the backend.
```
kevin9foong: Have a typo here, will change on Monday

kevin9foong: fixed in [4db1791](https://github.com/opengovsg/FormSG/pull/7937/commits/4db17911df4844f12d24b606c50cbb557f2e11e1)

kevin9foong: made the comment more terse as well

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -515,32 +574,18 @@ export const PublicFormProvider = ({
...
-          countryRegionFieldIds.has(fieldId) &&
-          typeof currentInput === 'string'
-        ) {
-          newFormInputs[fieldId] = currentInput.toUpperCase()
-        } else {
-          newFormInputs[fieldId] = currentInput
-        }
-        return newFormInputs
-      }, {})
+      const transformedFormInputs = transformFormInputTrimTextInputs(
```
KenLSM: /nit I initially missed out that there's a nested function call. Wondering if this be made more explicit?

```tsx
let transformedFormInputs = formInput
transformedFormInputs = transformFormInputCountryRegionToUpperCase(
  transformedFormInputs,
  form.form_fields
)
transformedFormInputs = transformFormInputTrimTextInputs(
  transformedFormInputs,
  form.form_fields
)
```
vs 
```tsx
const transformedFormInputs = transformFormInputTrimTextInputs(
  transformFormInputCountryRegionToUpperCase(
    formInputs,
    form.form_fields,
  ),
  form.form_fields,
) as typeof formInputs
```

KenLSM (approved):
LGTM. One code/stylistic recommendation

frontend/src/features/public-form/PublicFormProvider.tsx
kevin9foong: Acknowledging that the nesting of functions may be messy and hard to read. Trying to keep the use of `const` as this can prevent unwanted mutations elsewhere in the code. 

hence, i am proposing the use of `flow` in lodash which is made for scenarios like this: 
  ```
    const transformedFormInputs = flow([
        transformFormInputCountryRegionToUpperCase(form.form_fields),
        transformFormInputTrimTextInputs(form.form_fields),
      ])(formInputs) as typeof formInputs
```

KenLSM: Looks v. neat!

PR #7940, opened 2024-11-25
fix(i18n): replace title,desc missed earlier

LoneRifle:
# Problem and Solution
Polish up #7134 by replacing `Field Name` and `Description` with i18n equivalent


KenLSM (approved):
LGTM!

PR #7501, opened 2024-07-11
feat: replace hardcoded text in FeedbackBlock for i18n

paradite:
## Problem

Follow up of #7427 to continue replacing hardcoded text in various components for i18n.

This PR focuses on `/frontend/src/features/public-form/components/FormEndPage/components/FeedbackBlock.tsx`

## Solution

This PR replaces all hardcoded string in `FeedbackBlock`.

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Improvements**:

- i18n locale in `/frontend/src/features/public-form/components/FormEndPage/components/FeedbackBlock.tsx` refactored

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-07-11 at 7 38 42 PM

**AFTER**:

non-payment:

Screenshot 2024-07-11 at 7 42 51 PM

payment:

Screenshot 2024-07-11 at 7 44 23 PM

PR #7941, opened 2024-11-25
feat(i18n): replace hardcoded text in FeedbackBlock for i18n

LoneRifle:
## Problem and Solution

Add i18n to FeedbackBlock, and take the opportunity to correct casing for publicFormSubmitButton

Supersedes #7501


KenLSM (approved):
LGTM!

PR #7965, opened 2024-12-02
feat(mrf-admin-view): hawkeye v1.0

kevin9foong:
## Problem

Currently, admins want a way to track the status of their multi-respondent submissions/workflows to see: 
- is the workflow completed / pending? 
- what is the current step of the workflow? 

Closes FRM-1919

## Solution

Add the above-mentioned information to the responses dashboard, csv download and individual responses page so admins can track the status. 

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image
New columns added and new name for timestamp column.

## Tests

TC1: Hawk eye view dashboard shows the new `status`, `current step` and `submission time of first step`
- [ ] Create 2 step MRF form without approval step
- [ ] Submit step 1 
- [ ] Check responses dashboard, individual submission and download CSV. It should show pending status and correct step which is 1 of 2. 
- [ ] Submit step 2. 
- [ ] Check responses dashboard, individual submission and download CSV. It should show completed status and correct step which is 2 of 2.

TC2: Reject and approve show correctly. 
- [ ] Create a 3 step form with step 2 being approval step. Also, add a compulsory attachment field at step 1. When submitting the form in subsequent steps of the test case, always upload an attachment.
- [ ] Submit step 1 and 2 of form, rejecting at step 2. (ie click `No`) 
- [ ] Check responses dashboard, individual submission and download CSV. It should show completed status and correct step which is 2 of 3. 
- [ ] Try download CSV with attachment. Ensure the csv file is correctly downloaded and the data in the csv is as expected and attachment files are downloaded. 
- [ ] Submit step 1, 2 and 3 of form, ensuring that step 2 is approve (ie click `Yes`) 
- [ ] Check responses dashboard, individual submission and download CSV. It should show completed status and correct step which is 3 of 3.
- [ ] Try download CSV with attachment. Ensure the csv file is correctly downloaded and the data in the csv is as expected and attachment files are downloaded.

TC3: Storage mode and storage mode with with payments unaffected. 
- [ ] Create a storage mode form. 
- [ ] Submit and check the dashboard. Ensure `status`, `current step` and `submission time of first step` do not exist in the dashboard and the other expected columns exist (ie, same as before the changes introduced in this PR / refer to prod env)
- [ ] Download the CSV, ensure that `status`, `current step` and `submission time of first step` do not exist in the csv. 
- [ ] Add payments to the storage mode form. 
- [ ] Submit and check the dashboard. Ensure `status`, `current step` and `submission time of first step` do not exist and the dashboard is same as previous (ie, same as before the changes introduced in this PR / refer to prod env).

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -192,21 +192,6 @@ export const IndividualResponsePage = (): JSX.Element => {
             </>
           ) : null}
           <StackRow
-            label="Status"
-            value={responseMrfStatus}
-            isLoading={isLoading}
-            isError={isError}
-          />
-          <StackRow
-            label="Current step"
-            value={getCurrentStepString(
-              data?.mrf?.workflowCurrentStepNumber,
-              data?.mrf?.workflowNumTotalSteps,
-            )}
-            isLoading={isLoading}
-            isError={isError}
-          />
-          <StackRow
             label={isMrf ? MRF_FIRST_STEP_TIMESTAMP_LABEL : 'Timestamp'}
             value={
               data?.submissionTime ?? t('features.common.loadingWithEllipsis')
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -10,6 +10,7 @@ import {
   StackDivider,
   Text,
 } from '@chakra-ui/react'
+import moment from 'moment'
 import simplur from 'simplur'
 
 import { FormResponseMode } from '~shared/types'
@@ -23,14 +24,14 @@ import { FormActivationSvg } from '~features/admin-form/settings/components/Form
 import { useUser } from '~features/user/queries'
 
 import {
-  getCurrentStepString,
+  getPendingResponseAtString,
   getStatusFromWorkflowStatus,
 } from '../common/utils/mrfSubmissionView'
 import { SecretKeyVerification } from '../components/SecretKeyVerification'
 import {
-  MRF_CURRENT_STEP_LABEL,
-  MRF_FIRST_STEP_TIMESTAMP_LABEL,
-  MRF_STATUS_LABEL,
+  MRF_PENDING_RESPONSE_AT_LABEL,
+  MRF_RESPONSE_TIMESTAMP_LABEL,
+  MRF_WORKFLOW_STATUS_LABEL,
 } from '../constants'
 import { useStorageResponsesContext } from '../ResponsesPage/storage'
 
@@ -157,6 +158,12 @@ export const IndividualResponsePage = (): JSX.Element => {
     ? getStatusFromWorkflowStatus(workflowStatus)
     : ''
 
+  const lastSubmittedAt = data?.mrf?.lastSubmittedAt
+    ? moment(data.mrf.lastSubmittedAt)
+        .tz('Asia/Singapore')
+        .format('ddd, D MMM YYYY, hh:mm:ss A')
+    : ''
+
   return (
     <Flex flexDir="column" marginTop={{ base: '-1.5rem', md: '-3rem' }}>
       <IndividualResponseNavbar />
@@ -175,26 +182,28 @@ export const IndividualResponsePage = (): JSX.Element => {
           {isMrf ? (
             <>
               <StackRow
-                label={MRF_STATUS_LABEL}
+                label={MRF_WORKFLOW_STATUS_LABEL}
                 value={responseMrfStatus}
                 isLoading={isLoading}
                 isError={isError}
               />
               <StackRow
-                label={MRF_CURRENT_STEP_LABEL}
-                value={getCurrentStepString(
-                  data?.mrf?.workflowCurrentStepNumber,
-                  data?.mrf?.workflowNumTotalSteps,
-                )}
+                label={MRF_PENDING_RESPONSE_AT_LABEL}
+                value={getPendingResponseAtString({
+                  workflowCurrentStepNumber:
+                    data?.mrf?.workflowCurrentStepNumber,
+                  workflowNumTotalSteps: data?.mrf?.workflowNumTotalSteps,
+                })}
                 isLoading={isLoading}
                 isError={isError}
               />
             </>
           ) : null}
           <StackRow
-            label={isMrf ? MRF_FIRST_STEP_TIMESTAMP_LABEL : 'Timestamp'}
+            label={isMrf ? MRF_RESPONSE_TIMESTAMP_LABEL : 'Timestamp'}
             value={
-              data?.submissionTime ?? t('features.common.loadingWithEllipsis')
+              (isMrf ? lastSubmittedAt : data?.submissionTime) ??
+              t('features.common.loadingWithEllipsis')
             }
             isLoading={isLoading}
             isError={isError}
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -10,7 +10,7 @@ import {
   StackDivider,
   Text,
 } from '@chakra-ui/react'
-import moment from 'moment'
+import { formatInTimeZone } from 'date-fns-tz'
 import simplur from 'simplur'
 
 import { FormResponseMode } from '~shared/types'
@@ -159,9 +159,11 @@ export const IndividualResponsePage = (): JSX.Element => {
     : ''
 
   const lastSubmittedAt = data?.mrf?.lastSubmittedAt
-    ? moment(data.mrf.lastSubmittedAt)
-        .tz('Asia/Singapore')
-        .format('ddd, D MMM YYYY, hh:mm:ss A')
+    ? formatInTimeZone(
+        data.mrf.lastSubmittedAt,
+        'Asia/Singapore',
+        'eee, d MMM yyyy, hh:mm:ss a',
+      )
     : ''
 
   return (
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -27,6 +27,11 @@ import {
   getStatusFromWorkflowStatus,
 } from '../common/utils/mrfSubmissionView'
 import { SecretKeyVerification } from '../components/SecretKeyVerification'
+import {
+  MRF_CURRENT_STEP_LABEL,
+  MRF_FIRST_STEP_TIMESTAMP_LABEL,
+  MRF_STATUS_LABEL,
+} from '../constants'
 import { useStorageResponsesContext } from '../ResponsesPage/storage'
 
 import { DecryptedRow } from './DecryptedRow'
@@ -87,6 +92,8 @@ export const IndividualResponsePage = (): JSX.Element => {
   if (!formId) throw new Error('Missing formId')
 
   const { data: form } = useAdminForm()
+
+  const isMrf = form?.responseMode === FormResponseMode.Multirespondent
 
   const { user } = useUser()
   const { secretKey } = useStorageResponsesContext()
@@ -160,6 +167,25 @@ export const IndividualResponsePage = (): JSX.Element => {
             isLoading={isLoading}
             isError={isError}
           />
+          {isMrf ? (
+            <>
+              <StackRow
+                label={MRF_STATUS_LABEL}
+                value={getStatusFromWorkflowStatus(data?.mrf?.workflowStatus)}
+                isLoading={isLoading}
+                isError={isError}
+              />
+              <StackRow
+                label={MRF_CURRENT_STEP_LABEL}
+                value={getCurrentStepString(
+                  data?.mrf?.workflowCurrentStepNumber,
+                  data?.mrf?.workflowNumTotalSteps,
+                )}
+                isLoading={isLoading}
+                isError={isError}
+              />
+            </>
+          ) : null}
           <StackRow
             label="Status"
             value={getStatusFromWorkflowStatus(data?.mrf?.workflowStatus)}
@@ -176,7 +202,7 @@ export const IndividualResponsePage = (): JSX.Element => {
             isError={isError}
           />
           <StackRow
-            label="Timestamp"
+            label={isMrf ? MRF_FIRST_STEP_TIMESTAMP_LABEL : 'Timestamp'}
             value={
               data?.submissionTime ?? t('features.common.loadingWithEllipsis')
             }
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
@@ -22,6 +22,7 @@ import { useAdminForm } from '~features/admin-form/common/queries'
 import { FormActivationSvg } from '~features/admin-form/settings/components/FormActivationSvg'
 import { useUser } from '~features/user/queries'
 
+import { getCurrentStepString } from '../common/utils/mrfSubmissionView'
 import { SecretKeyVerification } from '../components/SecretKeyVerification'
 import { useStorageResponsesContext } from '../ResponsesPage/storage'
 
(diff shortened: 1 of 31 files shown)


kevin9foong:
- [x] To reduce duplicate for the `Step X of Y` string into a common util function
- [x] To explore reusing types `mrfMetadata`

shared/types/submission.ts
```diff
@@ -80,6 +80,39 @@ export type StorageModeSubmissionBase = z.infer<
...
+}
+
+export const ApprovalStatus = z.enum([
+  WorkflowStatus.APPROVED,
+  WorkflowStatus.REJECTED,
+])
+
+const SubmittedNonApprovalStep = z.object({
+  isApproval: z.literal(false),
+  submittedAt: z.string(),
```
kevin9foong: does zod have a timestamp type for validation? if so, maybe we can use it

kevin9foong: yes - fixed referring to [zod's datetime documentation](https://zod.dev/?id=datetimes) with `submittedAt: z.string().datetime({ precision: 3 })`

frontend/src/features/admin-form/responses/common/utils/mrfSubmissionView.ts
```diff
@@ -0,0 +1,26 @@
+import { WorkflowStatus } from '~shared/types'
+
+/** Gets the business friendly string for current step of workflow. */
```
kevin9foong, line 31: these are helper functions to generate the view for mrf submissions responses

frontend/vite.config.ts
```diff
@@ -61,5 +61,10 @@ export default defineConfig(() => {
       plugins: () => [tsconfigPaths()],
       format: 'es' as const,
     },
+    define: {
+      // On local dev, global is undefined, causing decryption workers to fail.
+      // This is a workaround based on https://github.com/vitejs/vite/discussions/5912.
+      global: 'globalThis',
```
kevin9foong: used the bypass an error faced as described by the comment.

src/app/modules/submission/submission.utils.ts
```diff
@@ -815,3 +819,61 @@ export const getCookieNameByAuthType = (
...
+    }
+    return WorkflowStatus.COMPLETED
+  }
+  return WorkflowStatus.PENDING
+}
+
+/**
+ * Builds the metadata for a multirespondent form submission.
+ */
+export const buildMrfMetadata = ({
```
kevin9foong, line 861: mrfMetadata refers to data which is not stored but derived from stored data when `buildMrfMetadata` is executed. This metadata is used in the submission and submissionMetadata requests to display metadata info about the submission in the responses table, downloaded csv file and individual submission response detail page.

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
```diff
@@ -112,7 +112,7 @@ export class EncryptedResponseCsvGenerator extends CsvGenerator {
     if (this.hasBeenProcessed) return
 
     // Create a header row in CSV using the fieldIdToQuestion map.
-    const headers = this.isMrf ? MRF_CSV_HEADERS : NON_MRF_CSV_HEADERS
+    const headers = this.isMrf ? [...MRF_CSV_HEADERS] : [...NON_MRF_CSV_HEADERS]
```
kevin9foong: Note: The de-structuring is necessary as we are mutating the `headers` array below. De-structuring creates a new array which prevents mutations from affecting the referenced const array (ie, `MRF_CSV_HEADERS` or `NON_MRF_CSV_HEADERS`). 

Otherwise, `headers` points to the `MRF_CSV_HEADERS` or `NON_MRF_CSV_HEADERS` const array and mutates the same memory address which is shared across multiple runs of `process()`. This causes issues in cases where the memory is shared/reused,specifically test cases, which causes the headers to contain the headers from the previous runs of `process()` and for the test cases to fail.

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/ResponsesTable/ResponsesTable.tsx
```diff
@@ -193,7 +194,8 @@ const MRF_RESPONSE_TABLE_COLUMNS: Column<ResponseColumnData>[] = [
     Header: MRF_FIRST_STEP_TIMESTAMP_LABEL,
     accessor: 'submissionTime',
     width: 250,
-    disableResizing: true
+    minWidth: 250,
```
kevin9foong, line 243: Note: re-using the same fix as existing storage mode responses table columns to include minWidth which allows the bar to expand for the rightmost column when the screen is small and horizontal scroll is enabled.

src/app/modules/submission/submission.utils.ts
kevin9foong: tested to generate the correct metadata given the submission data

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/ResponsesTable/ResponsesTable.tsx
```diff
@@ -7,42 +8,89 @@ import {
...
+    display="flex"
+    textColor="success.700"
+    backgroundColor="success.100"
+  >
+    <Icon as={BiCheckDouble} mr="0.25rem" fontSize="1rem" />
+    <Text textStyle="caption-1">Completed</Text>
+  </Badge>
+)
+
+const NON_MRF_RESPONSE_TABLE_COLUMNS: Column<ResponseColumnData>[] = [
```
KenLSM: Would prefer to use "base" or "default" rather than "non-MRF". As non-MRF is not an exception. i.e., Payment is special, MRF is special, but non-MRF is not special.

```suggestion
const BASE_RESPONSE_TABLE_COLUMNS: Column<ResponseColumnData>[] = [
```

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/ResponsesTable/ResponsesTable.tsx
```diff
@@ -138,6 +238,16 @@ export const ResponsesTable = () => {
     }
   }, [filteredMetadata, metadata, submissionId])
 
+  const columns = useMemo(() => {
```
KenLSM, line 283: Good to see that this is extracted out to a function rather than choosing to add one more layer of ternary. 👍

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
KenLSM: Good explanation to highlight that we need to create new array as we are mutating it at a later stage. Would recommend to put a short note in the code.

frontend/src/features/admin-form/responses/common/utils/mrfSubmissionView.ts
```diff
@@ -0,0 +1,31 @@
...
+  workflowStatus: WorkflowStatus | undefined,
+) => {
+  switch (workflowStatus) {
+    case WorkflowStatus.COMPLETED:
+    case WorkflowStatus.APPROVED:
+    case WorkflowStatus.REJECTED:
+      return MRF_STATUS.COMPLETED
+    case WorkflowStatus.PENDING:
+      return MRF_STATUS.PENDING
+    default:
```
KenLSM: /nit since it is likely that all the `WorkflowStatus` is fully defined, we would want the [typesystem](https://www.notion.so/opengov/Use-exhaustive-switch-for-enums-e273c275a684492d8aef4789b4882906?pvs=4) to let us know that we need to update the cases here rather than it falling through to default.

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
KenLSM, line 1: General notes on this generator. It has been awhile since I've looked at this file. I feel that using in-place mutation feels iffy. For instance, it can be difficult to follow the number of `row.push(value)` especially when it gets a little more complex as it is implicit rather than explicit.

Note that this comment is not regarding the changes that were introduced on this file, but I'm simply taking this opportunity to open up a discussion on potential refactors that we might want to consider in the future. The style of coding is likely an artifact of the past almost half a decade ago where FP is not as familiar.

frontend/vite.config.ts
KenLSM: I experimented with removing the whole global access chunk in [decryption.worker.ts](https://github.com/opengovsg/FormSG/blob/feat/hawkeye-v1.0/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts#L18) and it seemed to have resolved the csv generation issue on local.

Might want to look into that solution since I would prefer deleting code than to have more config.

src/app/models/submission.server.model.ts
```diff
@@ -759,6 +791,7 @@ const buildSubmissionMetadata = (
   result: MetadataAggregateResult,
   currentNumber: number,
   paymentMeta?: PaymentAggregates,
+  mrfMeta?: MultiRespondentAggregates,
```
KenLSM: Would recommend to update the function definition to accept optional parameters as an object since I'm spotting a few uses of `foo(paramX, paramY, null, paramZ)`.
```
const buildSubmissionMetadata = (
  result: MetadataAggregateResult,
  currentNumber: number,
  options?: { 
    paymentMeta?: PaymentAggregates,
    mrfMeta?: MultiRespondentAggregates,
  }
)
```

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
```diff
@@ -469,7 +478,6 @@ export const updateMultiRespondentFormSubmission = ({
   encryptedPayload,
   logMeta,
 }: {
-  formId: string
```
KenLSM, line 472: Good spot 😆 👍

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
kevin9foong: Yes, ideally we would avoid mutation of the arrays as it can be hard to trace and debug. Will not introduce such a refactor in this PR to keep the scope defined to hawkeye.

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
kevin9foong: left a NOTE and TODO

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/ResponsesTable/ResponsesTable.tsx
kevin9foong: done

src/app/models/submission.server.model.ts
kevin9foong: refactored to utilise object argument

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/EncryptedResponseCsvGenerator/EncryptedResponseCsvGenerator.ts
kevin9foong: Left a todo to document this desired change in the future

frontend/src/features/admin-form/responses/common/utils/mrfSubmissionView.ts
kevin9foong: made it more explicit for `undefined` to become `''` and explicitly specified return literals of the function.

frontend/vite.config.ts
kevin9foong: removed the global reference + vite config, tested and it works well!

frontend/src/features/admin-form/responses/common/utils/mrfSubmissionView.ts
```diff
@@ -17,15 +17,15 @@ export const getCurrentStepString = (
...
+): MRF_STATUS | '' => {
   switch (workflowStatus) {
     case WorkflowStatus.COMPLETED:
     case WorkflowStatus.APPROVED:
     case WorkflowStatus.REJECTED:
       return MRF_STATUS.COMPLETED
     case WorkflowStatus.PENDING:
       return MRF_STATUS.PENDING
-    default:
+    case undefined:
```
KenLSM: 1. Should still have a default case for fall-through, recommend to utilise [exhaustive switch](https://www.notion.so/opengov/Use-exhaustive-switch-for-enums-e273c275a684492d8aef4789b4882906)
2. Check for undefined should be outside of switch case. Could be something before line 21.

KenLSM (changes requested):
See comments on exhaustive switch in `frontend/src/features/admin-form/responses/common/utils/mrfSubmissionView.ts`. That should be the last item before we're good to go!

frontend/src/features/admin-form/responses/common/utils/mrfSubmissionView.ts
kevin9foong: Thanks for sharing this -- it is a very cool usage of typescript

frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/ResponsesTable/ResponsesTable.tsx
```diff
@@ -300,11 +340,17 @@ export const ResponsesTable = () => {
           <Tr
             as="div"
             {...headerGroup.getHeaderGroupProps()}
+            key={headerGroup.getHeaderGroupProps().key}
             // To toggle _groupHover styles to show divider when header is hovered.
             data-group
           >
             {headerGroup.headers.map((column) => (
-              <Th as="div" pos="relative" {...column.getHeaderProps()}>
+              <Th
```
kevin9foong, line 350: fixing the warning where key should not be passed in a spread operator

src/app/modules/submission/submission.utils.ts
```diff
@@ -872,9 +873,15 @@ export const buildMrfMetadata = ({
     submittedSteps ?? [],
     workflow.length,
   )
+
+  const lastSubmittedAt =
```
kevin9foong, line 876: added a lastSubmittedAt for mrf for the responseTimestamp -- checked with Alicia what it should be

src/app/modules/submission/__tests__/submission.utils.spec.ts
```diff
@@ -103,14 +106,50 @@ describe('submission.utils', () => {
       approval_field: YES_NO_FIELD._id,
     }
 
+    it('should return undefined for lastSubmittedAt when submittedSteps is empty', () => {
```
kevin9foong, line 109: added some tc for the newly added lastSubmittedAt

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -157,6 +158,12 @@ export const IndividualResponsePage = (): JSX.Element => {
     ? getStatusFromWorkflowStatus(workflowStatus)
     : ''
 
+  const lastSubmittedAt = data?.mrf?.lastSubmittedAt
+    ? moment(data.mrf.lastSubmittedAt)
```
kevin9foong: this is necessary as individual currently includes `ddd, ...` in the front while the table and CSV mode do not. this is the cleanest way which delegates to the FE for date time formatting and the BE passing in a ISO timestamp.

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -10,7 +10,7 @@ import {
   StackDivider,
   Text,
 } from '@chakra-ui/react'
-import moment from 'moment'
```
kevin9foong: suspecting moment is causing chromatic to fail - replacing with `fns-date` FE library

kevin9foong: confirmed that it was the caused - this fixed the failing chromatic deploy

KenLSM: The reason was because `moment` didn't exist?

kevin9foong: yes - moment is in the overall package.json but not the frontend's package.json. changing to re-use `fns-date` fixes it as it is in the fe's package.json

KenLSM (approved):
LGTM

PR #7975, opened 2024-12-04
feat: add logs for uploads to s3

kevin9foong:
## Problem 
Currently, it is difficult to link object keys from the various S3 buckets to a formId and submissionId for investigation during incidents. 

## Solution 
Add logs to link the various bucket object ids to the formId.  
With the formId + estimate of the submission time (and ip), we will be able to better narrow down the submissionId.

### DIFF
--- a/src/app/modules/submission/submission.service.ts
+++ b/src/app/modules/submission/submission.service.ts
@@ -552,7 +552,6 @@ export const uploadAttachments = (
       meta: {
         action: 'uploadAttachments',
         formId,
-        attachmentMetadata: Array.from(attachmentMetadata.entries()),
       },
     })
     return attachmentMetadata
--- a/src/app/modules/submission/submission.service.ts
+++ b/src/app/modules/submission/submission.service.ts
@@ -432,33 +432,55 @@ export const triggerVirusScanThenDownloadCleanFileChain = <
     | ParsedClearAttachmentFieldResponseV3,
 >(
   response: T,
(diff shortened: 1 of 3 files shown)


src/app/modules/submission/submission.service.ts
```diff
@@ -432,33 +432,55 @@ export const triggerVirusScanThenDownloadCleanFileChain = <
...
+            error,
+          })
+          return new MaliciousFileDetectedError(response.filename)
+        }
+        return error
+      })
+      .map((lambdaOutput) => {
+        logger.info({
+          message:
+            'Successfully retrieved clean file from virus scanning lambda',
```
kevin9foong, line 464: this allows us to link the object keys with the formId. with formid, timestamp and ip, will be able to narrow down submissions.

src/app/modules/submission/submission.service.ts
```diff
@@ -524,7 +546,16 @@ export const uploadAttachments = (
       })
       return new AttachmentUploadError()
     },
-  ).map(() => attachmentMetadata)
+  ).map(() => {
+    logger.info({
+      message: 'Successfully uploaded attachments to S3',
```
kevin9foong, line 551: does not necessarily help to identify the submission. but gives info about successful s3 uploads, similar to when submission is saved to mongodb successfully.

scottheng96 (approved):

PR #7976, opened 2024-12-04
feat: add translations to public form

siddarth2824:
## Changes
- Add language control to allow public form users to toggle languages on the form
- Use admin input translations on public form
- Add translations for prevent submission from form logic
- Add fixed translations for some components on the public form which are not generated by admin. For example, labels in `FormInstructionsContainer`. This [sheet](https://docs.google.com/spreadsheets/d/1u_dH53sglH5w7jV83rU5NFIhjk80_FIJ/edit?usp=sharing&ouid=101878657257239680436&rtpof=true&sd=true) contains the fixed translations. Highlighted in green are the translations added in this PR.

## Before & After Screenshots
| ----------- | ------- | ----- |
| Language control component to allow users to select the language to view the public form in | | Screenshot 2024-12-11 at 5 24 24 PM |


LoneRifle:
Generally, we would use i18next for text. Would be happy to point out how wthat can be done.

LoneRifle (dismissed):
Yea I'm afraid I might have to ask you to refactor to incorporate the use of i18next, as we discussed the other time

shared/constants/field/myinfo/index.ts
```diff
@@ -50,6 +53,11 @@ export const types: MyInfoFieldBlock[] = [
       'The registered name of the form-filler. This field is verified by ICA for Singaporeans/PRs & foreigners on Long-Term Visit Pass, and by MOM for Employment Pass holders.',
     fieldType: BasicField.ShortText,
     previewValue: 'PHUA CHU KANG',
+    titleTranslations: [
+      { language: Language.CHINESE, translation: '姓名' },
+      { language: Language.MALAY, translation: 'Nama' },
+      { language: Language.TAMIL, translation: 'பெயர்' },
+    ],
```
LoneRifle, line 60: Er we've talked about this, can you use i18next please?

frontend/src/features/admin-form/create/builder-and-design/utils/fieldCreation.ts
```diff
@@ -220,11 +220,19 @@ export const getMyInfoFieldCreationMeta = (
...
+    | 'description'
+    | 'fieldType'
+    | 'myInfo'
+    | 'titleTranslations'
+    | 'descriptionTranslations'
   > = {
     disabled: false,
     required: true,
     title: MYINFO_ATTRIBUTE_MAP[myInfoAttribute].value,
+    titleTranslations: MYINFO_ATTRIBUTE_MAP[myInfoAttribute]?.titleTranslations,
```
KenLSM, line 235: Note to self: this is expected to remains as a temporary solution as it is complex to extract the translations to be specified in locales file.

frontend/src/features/admin-form/settings/SettingsPage.tsx
```diff
@@ -51,6 +56,18 @@ export const SettingsPage = (): JSX.Element => {
...
+    isUserLoading ||
+    isFormSettingLoading ||
+    !user?.betaFlags?.multiLangTranslation
+      ? null
+      : {
+          label: 'Multi-language',
+          icon: LanguageTranslation,
+          component: SettingsMultiLangPage,
+          path: 'language',
+        }
```
KenLSM: /nit it is easier to read the inverse.

```suggestion
    !isUserLoading &&
    !isFormSettingLoading &&
    user?.betaFlags?.multiLangTranslation
      ? {
          label: 'Multi-language',
          icon: LanguageTranslation,
          component: SettingsMultiLangPage,
          path: 'language',
        }
      : null
```

shared/constants/field/myinfo/index.ts
KenLSM, line 1: Note to self: sister file relation to translations within a constants file as interim instead of having in locales.

frontend/src/features/public-form/components/FormFields/PublicFormSubmitButton.tsx
```diff
@@ -65,6 +65,19 @@ export const PublicFormSubmitButton = ({
     })
   }, [formInputs, formFields, formLogics])
 
+  const selectedLanguage = i18n.language as Language
+  const preventSubmissionMessage = useMemo(() => {
```
KenLSM: Possible to use `getValueInSelectedLanguage()`?

frontend/src/templates/Field/Radio/RadioField.tsx
```diff
@@ -50,6 +53,8 @@ export const RadioField = ({
     [disableRequiredValidation, schema],
   )
 
+  const defaultEnglishRadioOptions = schema.fieldOptions
```
KenLSM: /supernit would suggest to not name it as `defaultEnglishRadioOptions` but rather straight up `englishRadioOptions`. 

I had initially through that this value is to be overwritten as `defaultX` suggests that the value would be replaced. However, this is not the case as we strictly only want to pass `englishRadioOptions` value to the backend.

frontend/src/templates/Field/Dropdown/DropdownField.tsx
```diff
@@ -23,12 +25,53 @@ export const DropdownField = ({
...
   const rules = useMemo(() => {
     return createDropdownValidationRules(schema, disableRequiredValidation)
   }, [schema, disableRequiredValidation])
 
   const { control } = useFormContext<SingleAnswerFieldInput>()
 
+  const selectedLanguage = i18n.language as Language
+
+  const fieldOptions: ComboboxItem[] = useMemo(() => {
+    const defaultEnglishFieldOptions = schema.fieldOptions
```
KenLSM: /supernit in similar style with `RadioField`. `defaultX` suggests that this value would be replaced, but we shouldn't since we only want to pass the english values to the backend.

frontend/src/templates/Field/Dropdown/DropdownField.tsx
```diff
@@ -23,12 +25,53 @@ export const DropdownField = ({
...
+    const translationIdx = fieldOptionsTranslations.findIndex((translation) => {
+      return translation.language === selectedLanguage
+    })
+
+    // Check if translations for field options exist and whether
+    // each field option has its own respective translation. If not
+    // render the default field options in English.
+    if (
+      translationIdx !== -1 &&
+      fieldOptionsTranslations[translationIdx].translation.length ===
```
KenLSM, line 51: Check why this change is much more than radio field

frontend/src/templates/Field/Checkbox/CheckboxField.tsx
```diff
@@ -59,11 +62,20 @@ export const CheckboxField = ({
     [disableRequiredValidation, schema],
   )
 
+  const defaultEnglishCheckboxOptions = schema.fieldOptions
```
KenLSM: /supernit in similar style with `RadioField`. `defaultX` suggests that this value would be replaced, but we shouldn't since we only want to pass the english values to the backend.

frontend/src/utils/multiLanguage.ts
```diff
@@ -16,3 +20,64 @@ export const convertUnicodeLocaleToLanguage = (language: Language) => {
...
+  // Check if translations for field options exist and whether
+  // each field option has its own respective translation. If not
+  // render the default field options in English.
+  if (
+    translationIdx !== -1 &&
+    fieldOptionsTranslations[translationIdx].translation.length ===
+      defaultValue.length
+  ) {
+    return fieldOptionsTranslations[translationIdx].translation
+  } else {
```
KenLSM: /nit this `else` can be omitted. I generally prefer an `if without else` aka guard clauses

frontend/src/utils/multiLanguage.ts
```diff
@@ -16,3 +20,64 @@ export const convertUnicodeLocaleToLanguage = (language: Language) => {
...
+  translations?: TranslationOptionMapping[]
+  selectedLanguage: Language
+}
+
+export const getValueInSelectedLanguage = ({
+  defaultValue,
+  translations,
+  selectedLanguage,
+}: SelectedLanguageProps) => {
+  let title = defaultValue
```
KenLSM: /supernit Why `title`? This function is trying to get the `value` but this suggest that the value is used for `titles`

frontend/src/templates/Field/Dropdown/DropdownField.tsx
KenLSM: Is the reason why we cannot use `getFieldOptionsInSelectedLanguage` is due to the fact that we have a custom extractor[1] here? 

Context of question: I do like the utils function as it abstracts out the selection language out of the component rendering logic, so I'm sadden to see that it cannot be utilised. When one is reading through this component, the complex logic of language selection / detection (such as detection and selection of language + its defaults) is a distraction.

[1] the assumed offending extractor
```
return translatedFieldOptions.map((translatedFieldOption, index) => {
  return {
    value: defaultEnglishFieldOptions[index],
    label: translatedFieldOption,
  }
})
```

If this is the only problem, we can likely look into a utils that supports custom extractors as an enhancement in the future.

frontend/src/templates/Field/Section/SectionField.tsx
```diff
@@ -39,6 +41,34 @@ export const BaseSectionField = forwardRef<
     },
   })
 
+  const selectedLanguage = i18n.language as Language
+
+  let title = schema.title
+  const titleTranslations = schema?.titleTranslations ?? []
```
KenLSM: Can use `getValueInSelectedLanguage`?

frontend/src/templates/Field/Section/SectionField.tsx
```diff
@@ -39,6 +41,34 @@ export const BaseSectionField = forwardRef<
...
+  })
+
+  // If there exists a translation for title based on the selected language,
+  // use that. If not default to english
+  if (titleTranslationIdx !== -1) {
+    title = titleTranslations[titleTranslationIdx].translation
+  }
+
+  let description = schema.description
+  const descriptionTranslations = schema?.descriptionTranslations ?? []
```
KenLSM: Can use `getValueInSelectedLanguage`?

frontend/src/templates/Field/Table/ColumnCell.tsx
```diff
@@ -82,12 +84,52 @@ const DropdownColumnCell = ({
...
+    // Check if translations for field options exist and whether
+    // each field option has its own respective translation. If not
+    // render the default field options in English.
+    if (
+      translationIdx !== -1 &&
+      fieldOptionsTranslations[translationIdx].translation.length ===
+        defaultEnglishFieldOptions.length
+    ) {
+      const translatedFieldOptions =
+        fieldOptionsTranslations[translationIdx].translation
```
KenLSM, line 113: Same question as in `frontend/src/templates/Field/Dropdown/DropdownField.tsx` regarding the utilisation of multi-lang utils. Was it due to a need for custom extractor?

frontend/src/templates/Field/Table/TableFieldContainer.tsx
```diff
@@ -21,8 +24,42 @@ export const TableFieldContainer = ({
...
+    },
+  )
+
+  // If there are title translations for the selected language, use the translation.
+  // If not default it to English.
+  if (titleTranslationIdx !== -1) {
+    title = titleTranslations[titleTranslationIdx].translation
+  }
+
+  let description = schema.description
```
KenLSM: Can use `getValueInSelectedLanguage`?

frontend/src/features/public-form/components/LanguageControl/LanguageControl.tsx
```diff
@@ -0,0 +1,113 @@
...
+  if (!shouldShowLanguageDropdown) return null
+
+  return (
+    <Flex
+      background={bgColour}
+      zIndex={10}
+      px={{ base: '1.5rem', md: 0 }}
+      justifyContent={{ base: 'start', md: 'center' }}
+    >
+      <HStack mt="-32px" bg="white" borderRadius="4px" shadow="md">
```
KenLSM: Able to use `rem`?

frontend/src/features/admin-form/settings/components/MultiLanguageSection/TranslationListSection.tsx
```diff
@@ -345,9 +375,26 @@ export const TranslationListSection = ({
             )
           })}
 
+          {/* Form Logic Prevent Submission Message Translation */}
+          {formLogicPreventSubmissions && (
+            <>
+              <QuestionRow
+                key="Form Logic"
```
KenLSM, line 382: Is `key` needed here since we're not rendering them in a list like manner.

frontend/src/features/admin-form/settings/components/MultiLanguageSection/TranslationListSection.tsx
```diff
@@ -345,9 +375,26 @@ export const TranslationListSection = ({
...
+                unicodeLocale={language}
+              />
+              <Divider />
+            </>
+          )}
+
           {/* End Page Translation */}
           {form.endPage && (
             <QuestionRow
+              key="End Page"
```
KenLSM, line 397: Is `key` needed here since we're not rendering them in a list like manner.

frontend/src/components/FormEndPage/EndPageBlock.tsx
```diff
@@ -43,6 +46,31 @@ export const EndPageBlock = ({
...
 
+  const selectedLanguage = i18n.language as Language
+
+  const title = getValueInSelectedLanguage({
+    defaultValue: endPage.title,
+    translations: endPage.titleTranslations,
+    selectedLanguage,
+  })
+
+  const paragraph = useMemo(() => {
```
KenLSM: Can use `getValueInSelectedLanguage`?

KenLSM (commented):
Several question. All non-blocking.

frontend/src/features/admin-form/settings/components/MultiLanguageSection/TranslationListSection.tsx
siddarth2824: Even though we are not rendering in a list manner, I think its required because there are other `<QuestionRow>` components being rendered at the same nested level as itself. Hence requiring a key to uniquely identify the `<QuestionRow>` components from each other. I keep seeing the error on my console regarding this, hence I added it. I am not very sure if its required, but when I add it, the error on console seems to disappear.

KenLSM (approved):
LGTM! Thanks for taking time to split the PRs for easier review 👍

PR #8025, opened 2025-01-02
feat: set up stg-alt3 github actions

kevin9foong:
## Problem

As part of experimentation plan for IaC, we are setting up stg-alt3 which is provisioned using IaC. 

Closes FRM-1926

## Solution

This PR includes GH action scripts to deploy code changes (eg, virus scanner and application code) to the IaC provisioned staging-alt3 infrastructure.

## IMPORTANT: Pre and post release requirements
Virus scanner name: 

Before release: 
- the virus scanner lambda function name has been edited to user-defined name. To update on SSM param store. 

After release: 
- To update the ec2 role to remove the invoke lambda permission resouces for the old virus scanner names.


ecs-task-definition.json
```diff
@@ -0,0 +1,17 @@
+
+{ 
+  "containerDefinitions": [
+    {
+      "name": "formsg-app",
+      "essential": true,
+      "command": "/bin/sh -c \"./generate-env-from-ecs-params.sh && npm start\"",
+      "portMappings": [
```
kevin9foong: to config the port mappings, ecs task and execution role etc

.github/workflows/deploy-ecs.yml
```diff
@@ -0,0 +1,120 @@
...
+        env: 
+          DD_API_KEY: ${{ secrets.DD_API_KEY }}
+          DD_ENV: ${{ secrets.DD_ENV }}
+          ECR_REPOSITORY: ${{ secrets.ECR_REPO }}-${{ env.CURRENT_ENV }}
+        with:
+          context: .  
+          file: Dockerfile.production
+          push: true
+          tags: |
+            ${{ steps.login-ecr.outputs.registry }}/${{ env.ECR_REPOSITORY }}:${{ env.IMAGE_TAG }}
```
kevin9foong: to configure the s3 bucket image upload, static file upload and check if any dd config is still needed. 

also, ensure that able to push to ECR

kevin9foong:
To push to develop: 
- need to fix the tests 
- need to clean up the code 
- look at json and how to serve it to replace static
- potentially remove unused variables

src/app/routes/api/v3/corppass/corppass.routes.ts
```diff
@@ -14,10 +15,14 @@ export const CorppassOidcRouter = Router()
  * @route GET api/v3/corppass/.well-known/jwks.json
  * @returns 200
  */
-CorppassOidcRouter.use(
-  '/.well-known/jwks.json',
-  express.static(spcpMyInfoConfig.cpOidcRpJwksPublicPath),
-)
+CorppassOidcRouter.get('/.well-known/jwks.json', (_req, res) => {
```
kevin9foong, line 18: Replacing the static asset path with the json directly

.github/workflows/deploy-virus-scanner-guardduty-staging.yml
```diff
@@ -0,0 +1,26 @@
...
+  cancel-in-progress: true
+
+permissions:
+  id-token: write
+  contents: read
+
+on:
+  push:
+    branches:
+      - staging-alt3
```
kevin9foong, line 14: omit cron job for gd virus scanner deployment, since no scanner definition upgrade required

serverless/virus-scanner/scripts/envLoader.mjs
```diff
@@ -8,14 +8,70 @@ const SHORT_ENV_MAP = {
...
+  'VIRUS_SCANNER_CLEAN_S3_BUCKET'
+]
+
+/**
+ * Fetches the parameter string for the given environment from SSM. 
+ */
+const getParamString = async (env) => {
+  const client = new SSMClient({ region: 'ap-southeast-1' })
+
+  if (isIacMigratedSsmKeys(env)) {
```
kevin9foong, line 44: load the correct SSM key format for the virus scanner based on the env (ie, is stg-alt3)

serverless/virus-scanner-guardduty/serverless.yml
```diff
@@ -68,6 +68,7 @@ provider:
 
 functions:
   virus-scanner-guardduty:
+    name: virus-scanner-guardduty-${opt:stage}
```
kevin9foong, line 71: virus scanner lambda given explicit name so that matching it to the ssm is more predictable

src/app/utils/iac.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+export const retrieveFileContent = ({
+  preIacFilePath,
+  postIacBase64EncodedString,
+}: {
+  preIacFilePath: string
+  postIacBase64EncodedString: string
+}): string => {
+  if (isIacMigrated) {
+    return Buffer.from(postIacBase64EncodedString, 'base64').toString('utf-8')
```
KenLSM, line 29: Is there a reason to encode the fileContent when we add it to SSM? Previously this wasn't encoded, and thus no need to have a decode step.

kevin9foong: This is since SSM param store types only consist `String`, `StringList`, `SecureString`, which hints that SSM param store intends for the value stored to be of type string only (Source: https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html). 

Many of the .pem, .key files etc are in binary format, which SSM param store does not support (and I did not want to try to store binary data in a param that is of type `String` for example) . Hence, I opted to convert this into string to match what the param store type and used base64 as a consistent encoding function to convert into `String` regardless of binary file type.

kevin9foong: Resolved by the rationale in the docs for future dev reference

.github/workflows/deploy-ecs-stg-alt3.yml
```diff
@@ -0,0 +1,54 @@
...
+      ecs-container-name: 'formsg-app'
+
+      # CodeDeploy configuration
+      codedeploy-application: 'formsg-stg-alt3-ecs-app'
+      codedeploy-appspec-path: deploy/appspec.yml
+      codedeploy-deployment-group: 'formsg-stg-alt3-ecs-dg'
+
+    secrets:
+      # AWS credentials and configuration
+      aws-account-id: ${{ secrets.STG_ALT3_AWS_ACCOUNT_ID }}
```
kevin9foong: While not secret or sensitive, should be shared carefully (Source: https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-identifiers.html)

.github/workflows/deploy-ecs-stg-alt3.yml
```diff
@@ -0,0 +1,54 @@
...
+      cicd-role: ${{ secrets.STG_ALT3_AWS_CI_ROLE_TO_ASSUME }}
+      ecr-repository: ${{ secrets.STG_ALT3_ECR_REPOSITORY }}
+
+      # Monitoring and analytics configuration
+      ga-tracking-id: ${{ secrets.GA_TRACKING_ID }}
+      dd-env: ${{ secrets.DD_ENV }}
```
kevin9foong: potential to move env and sample rate to inputs instead of secret

kevin9foong: but dont have to do so now

.github/workflows/deploy-virus-scanner-legacy-staging.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  deploy-scanner:
+    name: Deploy Scanner
+    uses: ./.github/workflows/aws-deploy-scanner-legacy.yml
+    with:
+      checkoutBranch: staging-alt3
+      environment: stg-alt3
+      provisionedConcurrency: 1
+    secrets: 
+      cicd-role: ${{ secrets.STG_ALT3_AWS_CI_ROLE_TO_ASSUME }}
```
KenLSM: I don't see `STG_ALT3_AWS_CI_ROLE_TO_ASSUME` in secrets 🤔 is this supposed to be ~`AWS_CI_ROLE_TO_ASSUME`~ `IAC_AWS_CI_ROLE_TO_ASSUME`?

src/app/utils/iac.ts
KenLSM: As discussed, copying for posterity, most keys are generated as an encoded Base64 for transport thus there's likely no need for us to manually encode/decode this.

Noted on checking back on this when we're operating for `production`.

.github/workflows/deploy-virus-scanner-legacy-staging.yml
KenLSM: 2674c4f

KenLSM:
@kevin9foong because there's a name change to the lambda names, `CloudFormation` may take quite some time for it to teardown the old stack + setup the new stack. Does this mean that we have downtime?

kevin9foong:
> @kevin9foong because there's a name change to the lambda names, `CloudFormation` may take quite some time for it to teardown the old stack + setup the new stack. Does this mean that we have downtime?

To avoid downtime, we will need to: 
1. Change the SSM params to the new virus scanner name. 
2. Make a release. 
3. When the EBS ssm load script executes, it will pull the new virus scanner lambda names. 

For the previous deployment, since the current EBS loads the previous lambda names in its container, changing the SSM param in step 1 should not affect it and it will execute as normal. (so we will have 4 lambdas running simultaneously)

kevin9foong:
Closing this PR in light of renaming the branch from `staging-alt3` to `stg-alt3`. will link this PR to the new one

PR #8043, opened 2025-01-13
fix: set limit to 250kb for whitelist

kevin9foong:
## Problem
The large number of NRICs needed to be whitelisted resulted in two identified issues:
1.  The processing time of whitelisting took too long
- Majority of the processing time was used to encrypt the individuals NRICs. This ensures that data remains only accessible to the form owners.
- If the processing takes too long, FormSG automatically terminates the requests as it is not responsive. 
2. The information stored in FormSG's database went over the limits
- The encryption process increases the encrypted NRIC size.
- This results in the actual file size becoming larger than the limit, crossing FormSG’s database limit per form (estimated ~250k NRICs at this point)

Closes [insert issue #]

## Solution

Reduce whitelist limit to 250kB to mitigate and avoid misleading admins on the file size limit from facing repeated issue.

**Breaking Changes** 


KenLSM (approved):
LGTM. Thanks for investigating and figuring out the practical limits

PR #8052, opened 2025-01-14
feat: speed up encryption and decryption with precomputation

kevin9foong:
## Problem
Encryption and decryption is too slow for NRIC whitelisting.

Closes FRM-1928

## Solution
1. perform pre-computation of shared key, increasing encryption and decryption speed around 10-15x
2. add loading state to file download button during nric whitelist decryption for improved UX
3. increase file limit to 3mb, which is roughly 280k whitelisted nric

**Breaking Changes** 

## Tests
Admin can upload valid nric file  
- [ ] Upload a valid whitelist csv file with 200k records without own nric 
- [ ] Open and file and try to submit 
- [ ] Ensure it blocks access when not whitelisted. 
- [ ] Download the previously uploaded whitelist file (tests decryption)  
- [ ] Update whitelist with own nric and re-upload. (tests deletion of whitelist as well) 
- [ ] Ensure it allows access when whitelisted. 
- [ ] Submit the form to ensure can submit.

### DIFF
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormWhitelistAttachmentField.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormWhitelistAttachmentField.tsx
@@ -8,7 +8,7 @@ import {
 import { useParams } from 'react-router'
 import { Box, Skeleton } from '@chakra-ui/react'
 
-import { KB } from '~shared/constants'
+import { MB } from '~shared/constants'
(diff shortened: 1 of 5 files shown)


KenLSM (changes requested):
As discussed we want to ship the encryption improvement (1+2), but want to retain the limits.

Once that change is done we should be good to go.

kevin9foong:
Reduced the file limit back to 250kB.

KenLSM (approved):
LGTM!

PR #8056, opened 2025-01-16
feat: apply bucketting and removal of hydration to speed up nric whitelist processing 

kevin9foong:
## Problem
Even with pre-computation, processing nrics is still takes too long and times out. 

Upon further investigation, saving to mongodb is the bottleneck in terms of time taken. 

To provide more detail, for a 10mb file, there are 3 steps and the time taken which are: 
1. checking validity of records which takes around 250ms
2. encrypting nric records which takes 3s (after precomputation of shared key)
3. writing to mongodb it takes around 6-8s.   

## Solution

Writing to mongodb is found to be slow due to the hydration process, where the document schema is validated each time before sending a write to the database. However, this is not necessary since: 
1. we already validate the nric earlier for invalidity, duplicates or empty values `''` 
2. the schema is adhered to in the code base, which does not change. 

To skip this process, we apply `lean: true` to the options for `insertMany`. This improves the speed of mongodb upload by a factor of 10-12x from 6-8s to 600-800ms. 

With this change, testing on staging-alt reveals no time out issues for files roughly around 10mb. 

**Breaking Changes** 

## Tests

Same tests as the precompute PR.

### DIFF
--- a/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormWhitelistAttachmentField.tsx
+++ b/frontend/src/features/admin-form/settings/components/AuthSettingsSection/FormWhitelistAttachmentField.tsx
@@ -8,7 +8,7 @@ import {
 import { useParams } from 'react-router'
 import { Box, Skeleton } from '@chakra-ui/react'
 
-import { KB } from '~shared/constants'
+import { MB } from '~shared/constants'
 import { StorageFormSettings } from '~shared/types'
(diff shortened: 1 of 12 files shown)


KenLSM:
Holding first as this isn't backwards compatible. See details on [linear comment](https://linear.app/ogp/issue/FRM-1928/nric-whitelist-or-increase-encryption-decryption-speeds#comment-99d7b070)

PR #8065, opened 2025-01-21
feat: move banner to growthbook

g-tejas:
## Problem

Closes FRM-1956

https://www.notion.so/opengov/Migration-of-Banners-to-Growthbook-18377dbba788800e8dfcdf1544d9f41f?pvs=4

As for the banner hierarchy, SSM > growthbook. Any truthy value set in SSM will override the growthbook values. 
For example,

## Solution
- Buy pro plan
- Reduce TTL for cloudflare caching by worker. 

Currently verifying that the caching is done a the CDN level and not the browser level by monitoring the API requests after reducing the TTL. 

**Breaking Changes** 
- [x] No - this PR does not contain breaking changes
    - AWS SSM is still used as fallback for this initial phase out period. If SSM has a truthy value ,then its value is used instead, and growthbook is overriden. 

## Tests
- [ ] 1. Ensure that SSM has empty banner values for staging
- [ ] 2. On GB, add override value for staging banner,
- [ ] 3. Wait for cache to expire, Observe that the staging website shows updated banner info, and is formatted correctly (info warning error)
- [ ] 4. Add banner content to SSM, with a different value from GB.
- [ ] 5. Wait for cache expire
- [ ] 6. Observe that staging shows new banner value on `staging` and is formatted correctly.

### DIFF
--- a/frontend/src/features/admin-form/common/AdminFormLayout.tsx
+++ b/frontend/src/features/admin-form/common/AdminFormLayout.tsx
@@ -1,6 +1,7 @@
 import { useMemo } from 'react'
 import { Outlet, useParams } from 'react-router-dom'
 import { Flex } from '@chakra-ui/react'
+import { useFeatureValue } from '@growthbook/growthbook-react'
 import { get } from 'lodash'
 
 import { fillHeightCss } from '~utils/fillHeightCss'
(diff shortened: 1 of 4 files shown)


KenLSM (approved):
LGTM. To verify on whether the source precedence is respected.

PR #8090, opened 2025-02-11
fix: url redirect passing in url into description bug

scottheng96:
## Problem
Upon 'Learn More' button click under 'Enable Retries' section, url does not redirect to webhooks guide.

image

## Solution
explicitly label toggle inputs instead of passing it as a an object. FYI
{ returnObjects: true }, the t function returns the entire object for retry, not just the strings for label and description.

**Breaking Changes** 

**Bug Fixes**:
(upon re-direct, see below for behaviour - tested on local)


scottheng96:
https://github.com/user-attachments/assets/c172163c-060e-4b03-ad68-21c69f8bed1f

KenLSM (approved):
LGTM!

PR #8094, opened 2025-02-12
fix: joi validation to allow emails for duplicating forms

g-tejas:
## Problem

Admins report that they get an error "[emails] is not allowed". This is due to the Joi validation not accounting for the emails being pass to the duplicate form dto. This fixes it by allowing it. 

**Breaking Changes** 


KenLSM (approved):
LGTM

PR #8100, opened 2025-02-12
chore: remove coverage restriction, inspect through datadog instead

KenLSM:
## Problem

After we've introduced [datadog test impact analysis](https://docs.datadoghq.com/tests/test_impact_analysis/) a selected set of test files would only be executed which jest's coverage isn't able to properly account for. 

## Solution

Remove coverage requirements as we should be able to monitor this through datadog and prevent devs from needing to do wasteful work by re-running the pipeline. 

**Breaking Changes** 


g-tejas (approved):
lgtm

PR #8106, opened 2025-02-13
fix: hawkeye v1.0 pending response at missing bug

kevin9foong:
## Problem 
The pending response at was not showing up for pending status results on the hawkeye dashboard. This was due to a typo in the getResponseAtString function. 

## Solution 
- Pass in workflowStatus and check if is Pending. 
- Prevent re-occurrence by adding automated TCs for the getResponseAtString to expected output for a set of inputs. 
- Prevent passing of wrong props by matching the params in the args.


KenLSM (approved):
LGTM

PR #8108, opened 2025-02-14
fix: add precomposed apple icon to avoid 404 reports

KenLSM:
## Problem

Getting lots of 403 from S3 reporting that several of these urls are 403 FORBIDDEN.

Investigation yielded that the files do not exist and these urls are requested when users tries to save FormSG as a home screen shortcut on their iOS. 

Closes FRM-1932

## Solution

Provide apple-touch-icon-precomposed.png so that we don't get distracted by the 403s.

**Breaking Changes** 


g-tejas (approved):
read every pixel for potential virus. LGTM!

PR #8142, opened 2025-02-28
feat: add admin email as a reply-to

g-tejas:
## Problem
FRM-1964

## Solution

**Breaking Changes** 

## Tests
**Form creation workflow**
- [ ] 1. Without modifying the email depr feedback form, try to create a email form
- [ ] 2. Observe that email forms can be created without errors. Check console tab as well. Observe in the network tab, that you get 200 for the feedback submission
- [ ] 3. Modify the email depr feedback form, to add the email field **AFTER** the checkbox field. IT MUST BE THE LAST ONE. Make it not required
- [ ] 4. Repeat step 2.
- [ ] 5. ??
- [ ] 6. Profit

**Repeat the above, for both duplication + use template workflow**

## Deploy Notes
- Before deploying, create an email field, turn off required for the email depr feedback form


KenLSM (approved):
LGTM

PR #8143, opened 2025-02-28
fix: e2e form testing

g-tejas:
## Problem

Closes FRM-1966

## Solution
- Add sharding
- Hardcoding test suites to speed up playwright

**Breaking Changes** 

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -28,7 +28,7 @@ export default defineConfig({
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: [['list'], ['html']],
+  reporter: [['line'], ['html']],
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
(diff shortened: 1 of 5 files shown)


KenLSM:
Did a rough review.

1. Separate build away from actual tests. Only consume the sharded runners when we’re actually running the tests. Can likely save 2mins-ish per shard?
2. Manually configure the matrix to shard by browser instead of letting playwright decide (so that it is more deterministic)

playwright.config.ts
```diff
@@ -43,52 +43,47 @@ export default defineConfig({
...
+      name: 'chromium-login',
+      testMatch: /login\.spec\.ts$/,
       use: {
         ...devices['Desktop Chrome'],
       },
     },
     {
-      name: 'firefox',
+      name: 'chromium-email-submission',
+      testMatch: /email-submission\.spec\.ts$/,
```
KenLSM, line 54: @opengovsg/formsg-engineers note that for playwright, we're adding hardcoded tests suites. i.e., new test suites permutation will need to be added here, which shouldn't happen as we want to deprecate playwright in favour of chromatic / FE unit tests.

KenLSM (approved):
LGTM

PR #8176, opened 2025-03-11
feat: deploy guardduty lambda infra

scottheng96:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes** 

    - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


.github/workflows/deploy-virus-scanner-production.yml
```diff
@@ -21,3 +21,11 @@ jobs:
       provisionedConcurrency: 10
       checkoutBranch: 'release-al2'
     secrets: inherit
+  deploy-scanner-guardduty:
+    name: Deploy Scanner Guardduty
+    uses: ./.github/workflows/aws-deploy-scanner-guardduty.yml
+    with:
+      environment: 'production'
+      provisionedConcurrency: 1
```
KenLSM: Non-blocking since we can still adjust this value after deployment. Should this be similar to `10`?

serverless/virus-scanner-guardduty/entry.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+/usr/sbin/clamd
```
KenLSM: Need to remove this clam-daemon since we're no longer using clamav

serverless/virus-scanner-guardduty/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "package": "npm run env && sls package --stage ${ENV:-'staging'} --param=\"provisionedConcurrency=${CONCURRENCY}\""
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@aws-sdk/client-s3": "^3.529.1",
+    "@aws-sdk/client-ssm": "^3.363.0",
+    "aws-lambda-ric": "^2.1.0",
+    "axios": "^1.6.4",
+    "clamscan": "^2.1.2",
```
KenLSM: No need `clamscan` anymore

serverless/virus-scanner-guardduty/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "convict": "^6.2.4",
+    "http-status-codes": "^2.2.0",
+    "pino": "8.1.0",
+    "ts-retry-promise": "^0.8.1",
+    "uuid": "^9.0.0"
+  },
+  "devDependencies": {
+    "@types/aws-lambda": "^8.10.101",
+    "@types/clamscan": "^2.0.2",
```
KenLSM: Same here, no need `clamscan`

serverless/virus-scanner-guardduty/src/s3.service.ts
```diff
@@ -0,0 +1,274 @@
...
+            this.logger.info('No Malware Scan tag found', {
+              bucketName,
+              objectKey,
+            })
+            throw Error('No Malware Scan tag found')
+          }
+          return malwareScanningTag
+        },
+        {
+          timeout: 9 * 1000, // 9 seconds. FE becomes unresponsive if set to 10s or above
```
KenLSM: To note, this means that we'll wait a maximum of 9 seconds before reporting that tag was not found and will abort. Implication here is that we *might* (to collect data) be checking too early.

scottheng96:
image
UAT virus scanner deployed successfully with this PR. Prod will follow suit once merged

serverless/virus-scanner-guardduty/src/s3.service.ts
scottheng96: I see what you are getting at. Considering the linear backoff + delay, there will be a total of 9 retries over the 9 seconds. 

heres the calculations = 0.2+0.4+0.6+0.8+1.0+1.2+1.4+1.6+1.8=9.0 seconds = 9 retries

For monitoring, maybe we can double the timeout? (to 18 seconds) to be sufficiently on the safe side. What do you think?

serverless/virus-scanner-guardduty/package.json
scottheng96: removed this!

serverless/virus-scanner-guardduty/package.json
scottheng96: removed this!

serverless/virus-scanner-guardduty/entry.sh
scottheng96: removed this!

.github/workflows/deploy-virus-scanner-production.yml
scottheng96: Changed this to 10 for prod. was a copy paste mistake

serverless/virus-scanner-guardduty/src/s3.service.ts
KenLSM: Yeah i think 18 is fine as well! Best if there's a way to us to tell what's the P99 duration (or any pattern for the matter)

KenLSM (approved):
LGTM

PR #8180, opened 2025-03-11
fix: added attachments to approval workflow end email

scottheng96:
## Problem
Part of the work of attachment support in MRF emails

## Solution
Including attachments in the email response for approval/rejection outcome emails, similar to MRF workflow completion emails (not approvals).

Attachments are only included for email responses upon outcome fulfilment:
- When MRF is approved all the way (if there are multiple approvals)
- If MRF is rejected (and workflow stops there)


kevin9foong (approved):
lgtm

PR #8181, opened 2025-03-11
feat(i18n): extract text from landing page

Jaimeloeuf:
Rebase and use develop as the base branch

1. Scaffold i18n support for pages/landing/ directory
2. Extract hard coded text out from pages/landing/ into the new i18n directory

### DIFF
--- a/frontend/src/i18n/locales/features/landing-page/en-sg.ts
+++ b/frontend/src/i18n/locales/features/landing-page/en-sg.ts
@@ -1,0 +1,189 @@
+import { LandingPage } from '.'
+
+export const enSG: LandingPage = {
+  hero: {
+    title: 'Build secure government forms in minutes.',
+    subtitle:
+      'Instant, customisable forms with zero code or cost, to safely collect classified and sensitive data.',
+    ctaButtonLabel: 'Start building your form now',
+  },
+  featureSection: {
+    title: 'Our form building and data collection features',
+    features: {
+      dragDropBuilder: {
+        title: 'Drag and drop builder',
+        description:
+          'Launch forms in minutes with the drag-and-drop builder, and over 20 types of fields to choose from',
+      },
+      singpassAndMyinfo: {
+        title: 'Singpass and Myinfo',
+        description:
+          'Authenticate individuals or businesses with Singpass, and speed up form filling with pre-filled data from Myinfo',
+      },
+      conditionalLogic: {
+        title: 'Conditional logic',
+        description:
+          'Create dynamic forms that show or hide specific fields based on previous responses',
+      },
+      emailConfirmation: {
+        title: 'Email confirmation',
+        description:
+          'Send confirmation emails to your respondents along with a copy of their responses',
+      },
+      prefill: {
+        title: 'Prefill',
+        description:
+          'Make form filling faster for respondents by pre-filling fields for them',
+      },
+      webhooks: {
+        title: 'Webhooks',
+        description:
+          'Send form responses to external applications in real time',
+      },
+    },
+  },
+  getStartedSection: {
+    title: 'No onboarding, no fees, no code.',
+    subtitle:
+      'Sign in with your government email address, and start building forms immediately. It’s free, and requires no onboarding or approvals.',
+    ctaButtonLabel: 'Get started',
+  },
+  usedByAgenciesSection: {
+    title: 'Used by most government agencies',
+    formsLaunched: 'forms launched',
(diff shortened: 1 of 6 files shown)


littlemight (approved):

frontend/src/pages/Landing/Home/LandingPage.tsx
```diff
@@ -341,64 +403,79 @@ export const LandingPage = (): JSX.Element => {
...
+                'features.landingPage.helpCenterSection.qnaAccordionItem.howDoesE2eWork.question',
+              )}
+            >
               <MarkdownText components={mdComponents}>
-                {dedent`
-                When a respondent submits a response, response data is encrypted in the respondent's browser before being sent to our servers for storage. This means that by the time Form's servers receive responses, they have already been scrambled and are stored in this unreadable form. Your response data remains in this encrypted state until you decrypt your responses with your secret key, transforming them into a readable format. 
-
-                The benefit of end-to-end encryption is that response data enters and remains in Form's servers in an encrypted state. This ensures that even if our servers are compromised by an attack, attackers will still not be able to decrypt and view your response data, as they do not possess your secret key.  
-              `}
+                {dedent(
```
LoneRifle, line 451: I would actually invoke dedent from the locale, not the component. Should be i18next's responsibility to provide properly formatted text.

frontend/src/i18n/locales/features/landing-page/en-sg.ts
```diff
@@ -0,0 +1,189 @@
+import { LandingPage } from '.'
+
+export const enSG: LandingPage = {
+  hero: {
+    title: 'Build secure government forms in minutes.',
+    subtitle:
+      'Instant, customisable forms with zero code or cost, to safely collect classified and sensitive data.',
+    ctaButtonLabel: 'Start building your form now',
+  },
+  featureSection: {
```
LoneRifle, line 10: I think generally, I wouldn't want to suffix a section with `Section`. Should be obvious enough. At your discretion

LoneRifle (commented):
lgtm otherwise, left a couple of comments

frontend/src/pages/Landing/Home/LandingPage.tsx
Jaimeloeuf: ok i can move that, this my reasoning behind this

## dedent in tsx
Regardless of the language/locale, it will all be dedent before it is displayed, since it becomes the view component's responsibility to ensure that it is dedent before rendering.

## dedent in locale
1. We have to call dedent for every single language / locale, and it becomes the translator/locale maintainer's job to ensure that it is invoked before it works nicely
2. There is no way in the component to know that it is dedent already before rendering, instead it has to trust that the locale done that already
3. It might affect tree shaking / caching for locale file, since now it cant be statically cached as it has dynamic modules/elements to it

What do you think @LoneRifle ?

frontend/src/i18n/locales/features/landing-page/en-sg.ts
Jaimeloeuf: I think i would prefer to leave it in if it makes sense?

Primarily because this would help new project onboarding / translators who only touch locale files to very quickly understand what it is for.

Like you mentioned this would probably very obvious enough for the swe looking at the site's code / easy to use from the `t` function, but im thinking this suffix would be helpful mostly for new comers / translators?

What do you think?

frontend/src/pages/Landing/Home/LandingPage.tsx
LoneRifle: mm, okay, that much is true. We can leave for now then!

LoneRifle (approved):

frontend/src/pages/Landing/Home/LandingPage.tsx
```diff
@@ -210,51 +233,83 @@ export const LandingPage = (): JSX.Element => {
...
+          <Image
+            alt={t('features.landingPage.usedByAgenciesSection.imageAlt.msfd')}
             src={restrictedMsfLogo}
           />
         </Wrap>
       </LandingSection>
       <LandingSection bg="primary.100" align="center">
         <SectionTitleText>
-          Supporting national and emergent use cases
+          {t('features.landingPage.useCaseSection.title')}{' '}
```
KenLSM: Stray `{' '}`

Jaimeloeuf: fixed, thanks for catching

PR #8190, opened 2025-03-12
feat(i18n): extract text from feedback page

Jaimeloeuf:


littlemight:
For some bloody reason, it's also showing the diff which is already there (the landing page changes). 

Pretty sure the dangling commits are showing here.

PR #8202, opened 2025-03-14
fix: update dependencies

KenLSM:
## Problem

Update `elliptic` package

## Solution

**Breaking Changes** 

- No - this PR is backwards compatible

PR #8203, opened 2025-03-14
fix: frontend deps

KenLSM:
## Problem

Missed out `frontend` in #8202 

## Solution

**Breaking Changes** 


tutntut (approved):
lgtm

PR #8240, opened 2025-03-24
fix: update clean bucket download and mrf flow of invoking guardduty lambda

scottheng96:
## Problem

- Include invocation of guardduty lambda for MRF attachment submissions. Currently MRF attachment flow is submitting to s3 quarantine bucket, but lambda is not being invoked, causing file to stay in quarantine

## Solution

- Fix the downloading of the cleanFile so that BE can download both from clean virus scanning bucket & clean guardduty bucket

Closes FRM-2001

**Breaking Changes** 

### DIFF
--- a/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -6,6 +6,7 @@ import { err, errAsync, ok, okAsync, Result, ResultAsync } from 'neverthrow'
 
 import { IAttachmentInfo } from 'src/types'
 
+import { featureFlags } from '../../../../../shared/constants'
 import {
   BasicField,
   FormDto,
@@ -42,6 +43,7 @@ import {
 } from '../submission.errors'
 import {
   getEncryptedSubmissionData,
+  triggerGuarddutyScanThenDownloadCleanFileChain,
   triggerVirusScanThenDownloadCleanFileChain,
 } from '../submission.service'
 import {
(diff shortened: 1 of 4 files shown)


src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -192,17 +194,34 @@ type IdTaggedParsedClearAttachmentResponseV3 =
...
-  )
+>[] => {
+  return responses.map((response) => {
+    if (enableGuarddutyLambdaInvoke) {
+      triggerGuarddutyScanThenDownloadCleanFileChain(
+        response.answer,
+        formId,
+      ).mapErr((err) =>
+        logger.info({
+          message: `GuardDuty Scan unsuccessful:, ${err}`,
```
KenLSM: I missed this in the previous review. Logging of errors should be done something like this. This allows the error object to be consistently mapped on datadog.

```
...
mapErr((error) =>
  logger.error({
    message: `GuardDuty Scan unsuccessful,
    meta: {
      action: 'triggerGuarddutyScanThenDownloadCleanFileChain',
    },
    error,
  })
...
```

KenLSM: Do also update this on `src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts` as well.

KenLSM (changes requested):

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: Changed this, thanks for the pointer. Will make searching for these logs easier now that we're trying to do more monitoring

KenLSM (approved):
LGTM

src/app/modules/submission/submission.service.ts
```diff
@@ -1401,6 +1406,7 @@ export const triggerGuarddutyScanThenDownloadCleanFileChain = <
         downloadCleanFile(
           cleanAttachment.cleanFileKey,
           cleanAttachment.destinationVersionId,
+          AwsConfig.guarddutyQuarantineS3Bucket,
```
scottheng96, line 1409: I'm gonna do a quick PR to fix this, supposed to be clean bucket when download the file -.- missed this, sorry

PR #8251, opened 2025-03-26
fix: mrf prefills

g-tejas:
## Problem

Closes FRM-1981

**Breaking Changes** 

## Tests

**Encrypt-form**
- [ ] 1. Create encrypt form.
- [ ] 2. Add a short text field, dropdown, rating field and table with dropdown & short text columns. 
- [ ] 3. Observe that styling in the admin dashboard is the same as before.
- [ ] 4. For the dropdown fields, add really long options that result in overflow
- [ ] 5. Make form public.
- [ ] 6. Create a submission, observe that the fields are same styling as before, with no visual differences.
- [ ] 7. Enable pre-fill, with and without locked. Observe that styling is same as before.

**MRF**
- [ ] 1. Create mrf form
- [ ] 2. Add a short text field, dropdown, rating field and table with dropdown & short text columns. 
- [ ] 3. Observe that styling in the admin dashboard is the same as before.
- [ ] 4. For the dropdown fields, add really long options that result in overflow
- [ ] 5. Add two workflow steps, the first step should have all the input fields, second step should have nothing, but point the email to yours.
- [ ] 6. Make form public
- [ ] 7. Create a submission
- [ ] 8. Click on the email for the second step
- [ ] 9. Observe that text color for all 3 components are neutral.800
- [ ] 10. Observe that dropdown field and table fields are scrollable

### DIFF
--- a/frontend/src/components/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/Checkbox/Checkbox.tsx
@@ -25,7 +25,7 @@ export interface CheckboxProps extends ChakraCheckboxProps {
    */
   colorScheme?: FieldColorScheme
   /**
-   * Whether the checkbox is in a high contrast state.
+   * Better visibility of values in the disabled component when set to true.
    */
   highContrast?: boolean
 }
--- a/frontend/src/components/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/Checkbox/Checkbox.tsx
@@ -27,7 +27,7 @@ export interface CheckboxProps extends ChakraCheckboxProps {
   /**
    * Better visibility of values in the disabled component when set to true.
    */
-  highContrast?: boolean
+  isHighContrast?: boolean
 }
 
 type CheckboxWithOthers = ComponentWithAs<'input', CheckboxProps> & {
@@ -37,7 +37,7 @@ type CheckboxWithOthers = ComponentWithAs<'input', CheckboxProps> & {
 }
 
 export const Checkbox = forwardRef<CheckboxProps, 'input'>(
-  ({ children, colorScheme = 'primary', highContrast, ...props }, ref) => {
+  ({ children, colorScheme = 'primary', isHighContrast, ...props }, ref) => {
     // Passing all props for cleanliness but the size prop is the most relevant
     const { icon: iconStyles } = useMultiStyleConfig(CHECKBOX_THEME_KEY, props)
     return (
@@ -54,7 +54,7 @@ export const Checkbox = forwardRef<CheckboxProps, 'input'>(
         colorScheme={colorScheme}
         ref={ref}
         {...props}
-        {...(highContrast && { variant: 'highContrast' })}
+        {...(isHighContrast && { variant: 'highContrast' })}
       >
         {children}
       </ChakraCheckbox>
--- a/frontend/src/components/Checkbox/Checkbox.tsx
+++ b/frontend/src/components/Checkbox/Checkbox.tsx
@@ -24,6 +24,10 @@ export interface CheckboxProps extends ChakraCheckboxProps {
    * Background and shadow colors of checkbox.
    */
   colorScheme?: FieldColorScheme
+  /**
+   * Whether the checkbox is in a high contrast state.
+   */
+  highContrast?: boolean
 }
 
 type CheckboxWithOthers = ComponentWithAs<'input', CheckboxProps> & {
@@ -33,7 +37,7 @@ type CheckboxWithOthers = ComponentWithAs<'input', CheckboxProps> & {
 }
 
 export const Checkbox = forwardRef<CheckboxProps, 'input'>(
-  ({ children, colorScheme = 'primary', ...props }, ref) => {
+  ({ children, colorScheme = 'primary', highContrast, ...props }, ref) => {
     // Passing all props for cleanliness but the size prop is the most relevant
     const { icon: iconStyles } = useMultiStyleConfig(CHECKBOX_THEME_KEY, props)
     return (
@@ -50,6 +54,7 @@ export const Checkbox = forwardRef<CheckboxProps, 'input'>(
         colorScheme={colorScheme}
         ref={ref}
         {...props}
+        {...(highContrast && { variant: 'highContrast' })}
       >
         {children}
       </ChakraCheckbox>
--- a/frontend/src/components/DatePicker/DatePicker.tsx
+++ b/frontend/src/components/DatePicker/DatePicker.tsx
@@ -18,12 +18,14 @@ export interface DatePickerProps extends DatePickerBaseProps, CalendarProps {
   onInputValueChange?: (value: string) => void
   /** Default value for uncontrolled input. */
   defaultInputValue?: string
+  /** Whether the date picker is in a high contrast state. */
+  highContrast?: boolean
 }
 
 export const DatePicker = forwardRef<DatePickerProps, 'input'>((props, ref) => {
   return (
     <DatePickerProvider {...props}>
-      <DatePickerWrapper ref={ref}>
+      <DatePickerWrapper ref={ref} highContrast>
         <DatePickerContent>
           <DatePickerCalendar />
         </DatePickerContent>
--- a/frontend/src/components/DatePicker/DatePicker.tsx
+++ b/frontend/src/components/DatePicker/DatePicker.tsx
@@ -25,7 +25,7 @@ export interface DatePickerProps extends DatePickerBaseProps, CalendarProps {
 export const DatePicker = forwardRef<DatePickerProps, 'input'>((props, ref) => {
   return (
     <DatePickerProvider {...props}>
-      <DatePickerWrapper ref={ref} isHighContrast={props.isHighContrast}>
+      <DatePickerWrapper ref={ref}>
         <DatePickerContent>
           <DatePickerCalendar />
         </DatePickerContent>
--- a/frontend/src/components/DatePicker/DatePicker.tsx
+++ b/frontend/src/components/DatePicker/DatePicker.tsx
@@ -25,7 +25,7 @@ export interface DatePickerProps extends DatePickerBaseProps, CalendarProps {
 export const DatePicker = forwardRef<DatePickerProps, 'input'>((props, ref) => {
   return (
     <DatePickerProvider {...props}>
-      <DatePickerWrapper ref={ref} highContrast>
+      <DatePickerWrapper ref={ref} highContrast={props.highContrast}>
         <DatePickerContent>
           <DatePickerCalendar />
         </DatePickerContent>
--- a/frontend/src/components/DatePicker/DatePicker.tsx
+++ b/frontend/src/components/DatePicker/DatePicker.tsx
@@ -18,7 +18,7 @@ export interface DatePickerProps extends DatePickerBaseProps, CalendarProps {
   onInputValueChange?: (value: string) => void
   /** Default value for uncontrolled input. */
   defaultInputValue?: string
-  /** Whether the date picker is in a high contrast state. */
+  /** Better visibility of values in the disabled component when set to true. */
   highContrast?: boolean
 }
 
--- a/frontend/src/components/DatePicker/DatePicker.tsx
+++ b/frontend/src/components/DatePicker/DatePicker.tsx
@@ -19,13 +19,13 @@ export interface DatePickerProps extends DatePickerBaseProps, CalendarProps {
   /** Default value for uncontrolled input. */
   defaultInputValue?: string
   /** Better visibility of values in the disabled component when set to true. */
-  highContrast?: boolean
+  isHighContrast?: boolean
 }
 
 export const DatePicker = forwardRef<DatePickerProps, 'input'>((props, ref) => {
   return (
     <DatePickerProvider {...props}>
-      <DatePickerWrapper ref={ref} highContrast={props.highContrast}>
+      <DatePickerWrapper ref={ref} isHighContrast={props.isHighContrast}>
         <DatePickerContent>
           <DatePickerCalendar />
         </DatePickerContent>
--- a/frontend/src/components/DatePicker/components/DatePickerInput.tsx
+++ b/frontend/src/components/DatePicker/components/DatePickerInput.tsx
@@ -2,7 +2,7 @@ import { useMemo } from 'react'
 import ReactInputMask from 'react-input-mask'
 import {
   forwardRef,
-  Input,
+  Input as ChakraInput,
   InputGroup,
   InputRightAddon,
   useMergeRefs,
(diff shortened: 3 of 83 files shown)


frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -81,6 +87,14 @@ export const FormLabel = ({
       flexDir="column"
       overflowWrap="break-word"
       {...labelProps}
+      sx={{
+        ...(isMRFPrefilled && {
```
KenLSM: /nit for acronyms, we usually only capitalise the first of each acronym set i.e.,  `MRFHTTPSQL` vs `mrfHttpSql`.

frontend/src/components/Input/Input.tsx
```diff
@@ -88,32 +95,37 @@ export const Input = forwardRef<InputProps, 'input'>((props, ref) => {
...
-      return props.isPrefillLocked ? (
-        <InputGroup>
+      if (props.isPrefillLocked) {
+        return (
+          <InputGroup>
+            <ChakraInput
+              ref={ref}
+              {...preventDefault}
+              {...inputProps}
+              isDisabled={true}
```
KenLSM:
```tsx
<Foo 
  somethingTrue
/>

{/* is equivalent to */}

<Foo 
  somethingTrue={true}
/>
```

frontend/src/templates/Field/ShortText/ShortTextField.tsx
```diff
@@ -42,6 +42,7 @@ export const ShortTextField = ({
         isPrefilled={isPrefilled}
         // Prevent editing of pre-filled fields if lockPrefill is true
         isPrefillLocked={isPrefillLocked}
+        isMRFPrefilled={true} /* TODO: should be conditionally applied*/
```
KenLSM: What does this TODO mean?

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -81,6 +87,14 @@ export const FormLabel = ({
...
+          _disabled: {
```
KenLSM: Could this be applied at the `theme/component/FormLabel.tsx`? This would make the change to be consistent throughout the code. 

1. Updating component to accept new modifiers
2. Updating component theme to change css based on the new modifier

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -52,6 +52,11 @@ export interface FormLabelProps extends ChakraFormLabelProps {
    * Whether markdown is enabled for description text.
    */
   useMarkdownForDescription?: boolean
+
+  /**
+   * Whether there's prefilled MRF state.
+   */
+  isMRFPrefilled?: boolean
```
KenLSM: *Discussion*
Would it make sense to name it as high-contrast or something that's not "business"-logic related?

Could we completely abstract separate the "business"-logic away from "presentation"-logic (what css should we render with). i.e., could this `isMRFPrefilled` modifier be reused in other OGP application, if so, it is a "presentation"-logic because it agnostic to whether it is on FormSG or not, otherwise it is a "business"-logic because it is specific to FormSG's context.

g-tejas: Makes sense! thanks

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
g-tejas: Good idea, will modify thanks!

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -71,6 +76,7 @@ export const FormLabel = ({
   questionNumber,
   description,
   useMarkdownForDescription = false,
+  isHighContrast = true,
```
KenLSM: Shouldn't be `isHighContrast` by default if I understood the requirements.

frontend/src/components/Checkbox/Checkbox.tsx
```diff
@@ -50,6 +54,7 @@ export const Checkbox = forwardRef<CheckboxProps, 'input'>(
         colorScheme={colorScheme}
         ref={ref}
         {...props}
+        {...(highContrast && { variant: 'highContrast' })}
```
kevin9foong, line 57:
```suggestion
       {...(highContrast ? { variant: 'highContrast' } : {})}
```

nit to be consistent with `frontend/src/components/Button/Button.tsx`

kevin9foong: using the chakra variant makes sense 👍

frontend/src/components/DatePicker/components/DatePickerInput.tsx
```diff
@@ -2,7 +2,7 @@ import { useMemo } from 'react'
 import ReactInputMask from 'react-input-mask'
 import {
   forwardRef,
-  Input,
+  Input as ChakraInput,
```
kevin9foong, line 5: curious, why is this renamed to `ChakraInput`

frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
```diff
@@ -85,7 +87,27 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
             <Text
               textStyle="body-1"
               noOfLines={1}
-              color={isDisabled ? 'neutral.500' : undefined}
+              {...(isDisabled
+                ? {
```
kevin9foong, line 111: ask during 1-1 what this code does, recommend to add the why behind why all these config exists. 

Also, might be good practice to do: 

```const <rationale for code eg, disabledHideScrollbarSettings> = ...``` to document using code instead of comments then use it to reference for better understandability by others

**Rationale:** 
This is important since when new devs make changes to the code, they may sometimes remove existing code. but its [hard to know whether something can be removed or not without knowing why](https://fs.blog/chestertons-fence/) (Chesterson's fence principle) 
The name instantly tells us what this config aims to do, the `what` without having to use code comments. 

see github/workflows/deploy-virus-scanner-staging.yml line 15 where the reason for the cron job is documented, so we know it is safe to remove the cron job when upgrading to new virus scanner

File: frontend/src/components/Field/YesNo/YesNo.tsx [hunk 10]
```diff
@@ -44,6 +44,11 @@ export interface YesNoProps {
    * Color scheme of the component to render. Defaults to `primary`.
    */
   colorScheme?: FieldColorScheme
+
+  /**
+   * Whether the component is in a high contrast state.
```
kevin9foong:
```suggestion
   * High contrast is used for better visibility of values when viewed in the disabled state.
```

frontend/src/components/Field/YesNo/YesNo.tsx
```diff
@@ -44,6 +44,11 @@ export interface YesNoProps {
...
+   */
+  highContrast?: boolean
```
kevin9foong:
```suggestion
  isHighContrast?: boolean
```

recommend for boolean values to follow naming convention eg, `isHighContrast` for better readability.

kevin9foong: eg, `isFieldEnabledByWorkflow ` seen in the code for example of this convention

frontend/src/components/Field/YesNo/YesNo.tsx
kevin9foong: recommend to include why a setting exists in the doc instead of listing what it does, which should be self explanatory based on the variable name

frontend/src/components/FormControl/FormLabel/FormLabel.tsx
```diff
@@ -52,6 +52,11 @@ export interface FormLabelProps extends ChakraFormLabelProps {
    * Whether markdown is enabled for description text.
    */
   useMarkdownForDescription?: boolean
+
```
kevin9foong, line 55: same - to include the why instead of what which is self explanatory. I understand that you're following the previous comment style, but since we are trying to improve the code quality over time, including why instead of what is more beneficial.

kevin9foong: source: https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/

frontend/src/components/Input/Input.tsx
```diff
@@ -37,6 +37,10 @@ export interface InputProps extends ChakraInputProps {
    * Whether there's an input right element. Used to provide additional padding
    */
   hasInputRightElement?: boolean
+  /**
+   * Whether the input is in a high contrast state. (e.g For MRF prefills)
```
kevin9foong: this makes sense as it includes why this was included - for MRF prefills (however, for MRF prefills may not be as comprehensible to devs who do not have the context on what MRF prefills is)

Hence, prefer to make the rationale comment more generic eg, 
- `isHighContrast is used to allow values to be read easily when state is disabled`

frontend/src/components/PhoneNumberInput/IntlPhoneNumberInput.tsx
```diff
@@ -64,6 +64,7 @@ export const IntlPhoneNumberInput = forwardRef<
         preventDefaultOnEnter
         {...props}
         placeholder={inputPlaceholder}
+        highContrast={props.highContrast}
```
kevin9foong: might not be needed since `...props` in L65 would already include destructuring of `highContrast`

frontend/src/templates/Field/Mobile/MobileFieldInput.tsx
```diff
@@ -23,13 +23,18 @@ export interface MobileFieldInputProps {
   ) => (value?: string) => void
   /** Any props to override internal input */
   phoneNumberInputProps?: Partial<PhoneNumberInputProps>
+  /**
+   * Whether the input is in a high contrast state.
```
kevin9foong: same comments should provide why instead of what, else can omit the comment

frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -13,7 +13,7 @@ import {
   Tr,
   VisuallyHidden,
 } from '@chakra-ui/react'
-import { get, head, uniq } from 'lodash'
+import { get } from 'lodash'
```
kevin9foong, line 16: thanks for cleaning this up! 👍

frontend/src/components/DatePicker/components/DatePickerInput.tsx
```diff
@@ -12,55 +12,58 @@ import {
...
-  )
-})
+    return (
+      <>
+        <VisuallyHidden aria-live="assertive">
+          {selectedDateAriaLiveText}
+        </VisuallyHidden>
+        <InputGroup>
+          <ChakraInput
+            inputMode="numeric" // Nudge Android mobile keyboard to be numeric
```
kevin9foong, line 46: what was nudging mean?

kevin9foong:
Looks good from preliminary review, with some items to change before the merge: 

Summary of review: 

- `isHighContrast` - i see you've followed this convention previously with `isMrfPrefilled` 

- Changing the comments from what to why (since the name of the param already describes what, the comments currently do not value add). We are constantly working on improving the code quality of FormSG over time, so changing comments to [mainly include why is important ](https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/) 

- Including the why for certain css config is key (recommend using code variable names instead of comments as much as possible). When working on the code, we often wonder why existing code/config exists. If this is not documented, future devs including myself [wonder if we can safely change this config](https://fs.blog/chestertons-fence/). This goes a long way to improving our codebase quality and is a standard we should enforce in the future. 

Fixing of visual tests: 
- visual test bugs can give a signal of existing bugs in the code - it will be good for us to add this for all changes and allow design function to review as well. 

Confidence of bug-free / incident-free in order to release: 
- Once the code is well tested, verified on Chromatic storybook (suggest separate visual tests for `isHighContrast=true + disabled = false` and `isHighContrast=false + disabled = true` per component) + manually via staging environments. We can have the confidence to release it.

frontend/src/components/DatePicker/components/DatePickerInput.tsx
kevin9foong: Spoke to tejas, added to be consistent with rest of code where when ChakraInput is used, it is renamed with alias. So this change makes sense 👍

frontend/src/components/DatePicker/components/DatePickerInput.tsx
```diff
@@ -35,13 +39,15 @@ export const DatePickerInput = forwardRef<object, 'input'>((_props, ref) => {
     return `Selected date: ${internalValue.toLocaleDateString()}`
   }, [internalValue])
 
+  // omit
```
kevin9foong: Stray comment

frontend/src/components/DatePicker/components/DatePickerInput.tsx
```diff
@@ -25,6 +26,9 @@ export const DatePickerInput = forwardRef<object, 'input'>((_props, ref) => {
     internalValue,
   } = useDatePicker()
 
+  const { isHighContrast: highContrast, ...filteredFcProps } =
+    fcProps as FormControlProps & { isHighContrast?: boolean }
```
kevin9foong: TS cannot resolve this - fixed in https://github.com/opengovsg/FormSG/pull/8251/commits/fef5d89dbce7f2cfc768db8517a5ca4bde276a93

frontend/src/components/DatePicker/components/DatePickerWrapper.tsx
```diff
@@ -4,40 +4,37 @@ import { useDatePicker } from '../DatePickerContext'
...
-          closeOnBlur={closeCalendarOnChange}
-          returnFocusOnClose={false}
-          {...disclosureProps}
-        >
-          <PopoverAnchor>
-            <DatePickerInput ref={ref} />
-          </PopoverAnchor>
-          {children}
-        </Popover>
+        <DatePickerInput ref={ref} isHighContrast={isHighContrast} />
```
kevin9foong: This is not required since the provider is the one that passes the `isHighContrast` prop and this is never used.

frontend/src/components/DatePicker/components/DatePickerInput.tsx
```diff
@@ -51,11 +57,12 @@ export const DatePickerInput = forwardRef<object, 'input'>((_props, ref) => {
           placeholder={placeholder}
           maskPlaceholder={placeholder}
           ref={mergedInputRef}
-          {...fcProps}
+          {...filteredFcProps}
           borderRightRadius={0}
           onBlur={handleInputBlur}
           onClick={handleInputClick}
           isReadOnly={fcProps.isReadOnly || !allowManualInput}
+          {...(highContrast && { variant: 'highContrast' })}
```
kevin9foong: Note that this `highContrast` is actually passed via the provider.

frontend/src/components/DatePicker/DatePicker.tsx
```diff
@@ -18,12 +18,14 @@ export interface DatePickerProps extends DatePickerBaseProps, CalendarProps {
   onInputValueChange?: (value: string) => void
   /** Default value for uncontrolled input. */
   defaultInputValue?: string
+  /** Better visibility of values in the disabled component when set to true. */
+  isHighContrast?: boolean
 }
 
 export const DatePicker = forwardRef<DatePickerProps, 'input'>((props, ref) => {
   return (
     <DatePickerProvider {...props}>
```
kevin9foong: The isHighContrast comes from `<DatePickerProvider {...props}>` which is then referenced in `fcProps` and not `<DatePickerWrapper ref={ref} isHighContrast={props.isHighContrast}>`, which the isHighContrast is not being used.

kevin9foong:
Noticing a couple of TS compile errors causing the CI pipeline to fail 
image

Will see if can pass the CI pipelines without fixing this - but needs to be tackled in a separate code quality PR (odd that our previous pipeline did not catch these...).

kevin9foong:
Prioritising on fixing blocking TS compile failures related to changes in mrf prefills to prevent scope creep 
image

kevin9foong:
Potential bug found:
- seems like the table field high contrast is not being applied for short text

image

kevin9foong:
Potential bug: 
image
Seems like the building name is not applying high contrast

kevin9foong:
Potential bug: 

building name does not seem to apply the high contrast
image

frontend/src/components/DatePicker/components/DatePickerWrapper.tsx
kevin9foong: fixed in https://github.com/opengovsg/FormSG/pull/8251/commits/fef5d89dbce7f2cfc768db8517a5ca4bde276a93

frontend/src/components/DatePicker/components/DatePickerInput.tsx
kevin9foong: fixed in https://github.com/opengovsg/FormSG/pull/8251/commits/fef5d89dbce7f2cfc768db8517a5ca4bde276a93

frontend/src/components/DatePicker/DatePicker.tsx
kevin9foong: fixed in https://github.com/opengovsg/FormSG/pull/8251/commits/fef5d89dbce7f2cfc768db8517a5ca4bde276a93

kevin9foong:
image

potential bug #3: others field high contrast not applied

g-tejas:
> Noticing a couple of TS compile errors causing the CI pipeline to fail image
> 
> Will see if can pass the CI pipelines without fixing this - but needs to be tackled in a separate code quality PR (odd that our previous pipeline did not catch these...).

We recently updated `react-hook-form`, which seems to be the cause of ur errors. Can try updating npm packages?

kevin9foong:
LGTM

kevin9foong (approved):

kevin9foong:
Can merge once design gives the chromatic approve

PR #8254, opened 2025-03-27
feat: guardduty switch

scottheng96:
## Problem

After monitoring using a shadowing flow, updating the code to return guardduty or local lambda flow based on feature flag toggle

## Solution
- FE still sends file to both s3 quarantine buckets
- BE will invoke both lambdas
- Toggle will choose which invoked lambda response to send downstream

**Breaking Changes** 


KenLSM (approved):
LGTM. Paired

PR #8269, opened 2025-04-01
fix: virus scanner guardduty update for unsupported files to be ok

scottheng96:
## Problem

GuardDuty virus scanner returns error when `MalwareScanTag` returns `UNSUPPORTED`. This is the case for password-protected PDFs which was discovered during the switch to use guardDuty flow on 28 March. 

Closes FRM-2004

## Solution
- Return 200 (OK) for `MalwareScanTag`: `UNSUPPORTED` files. This leaves `ERROR` to only return on scenarios of:
  - Malicious file detected
  - Access denied
  - Scan Failure

**Breaking Changes** 


KenLSM (approved):
Lgtm

PR #8291, opened 2025-04-04
fix: use forwardRef

KenLSM:
## Problem

Tooltip is shown on the top left (0,0) instead of the component that they are supposed to be near with.

Closes FRM-2008

## Solution

`ref`s needs to be passed over through `forwardRef` as they are a special type of react prop.

## Before & After Screenshots

|--------|--------|--------|
| Tooltip | Screenshot 2025-04-04 at 5 54 07 PM | Screenshot 2025-04-04 at 5 41 38 PM | 

### Regression Tests
**Tooltip appears above FormField**
- [ ] Create a form field
- [ ] Add form logic to hide field
- [ ] Observe that on mouse over, the tooltip appears just above the form field

### DIFF
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
@@ -1,4 +1,11 @@
-import { memo, useCallback, useEffect, useMemo, useRef } from 'react'
+import {
+  forwardRef,
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+} from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
```diff
@@ -234,6 +237,7 @@ const FieldRowContainer = ({
     )
   }, [isActive, isDirty, numFormFieldMutations, fieldBuilderState])
 
+  console.log({ ref2: ref })
```
scottheng96: Can remove this console log!

scottheng96 (approved):
ty for teaching me smth every day

frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignContent/FieldRow/FieldRowContainer.tsx
KenLSM: ec52870

PR #8301, opened 2025-04-10
feat(mrf): mask key

KenLSM:
## Problem

We are no longer requiring to debug keys for MRF. This should now be sanitised before the logs are sent to cloudwatch.

Closes FRM-2010

## Solution

On request buildup, we will run through an additional filter to explicitly mask the pattern.

**Breaking Changes** 


kevin9foong (approved):
LGTM!

PR #8305, opened 2025-04-11
fix: remove nested tests

KenLSM:
## Problem

> Tests cannot be nested. Test "should replace captured mrf keys with *" cannot run because it is nested within "should mask referer headers".

## Solution

Use `jest.describe` instead of `jest.it`.

**Breaking Changes** 


KenLSM:
Bypassing rules as it is a fix on tests

PR #8307, opened 2025-04-14
refactor(i18n): enforce locale enums

littlemight:
## Problem

Some dangling `en-SG` strings, it should be using the already available enums.

## Solution 
Use enums.

**Breaking Changes** 


littlemight:
defensively cleaning up stuff. fyi @LoneRifle, in preparation if eventually we need to incorporate a [lo-LA](https://simplelocalize.io/data/locale-code/lo-LA/) locale in the lao fork

LoneRifle (approved):
lgtm

PR #8311, opened 2025-04-15
feat(i18n): extract text from form validation rules

littlemight:
## Problem

_Follows on from #8310, creating a PR on @shagrirse behalf so that chromatic can be ran and reviewed.
Have vetted the code change in 8310 comments, please feel free to run through again if you're keen._

Extracting texts on the form-level validation rules to i18n locales. 

We're pulling more human readable texts to locale files. Motivation: so future forks (eg other countries wanting to adopt and translate) can just create `their-locale.ts` and reference how it's structured in `en-SG.ts`

Closes #8167.

## Solution

We're using [react-i18next](https://react.i18next.com/), which provides the `useTranslation()` hook to pull translated strings. Since it's a hook, it's meant to be used inside react components.

The form validation rule fns have been modified to facilitate it.

PS: There is a way to use pull human-readable strings for objects outside of react components (such as using the global `i18n.t()`, and other means). But this is generally not recommended [1] (the text will not update if in the future we have a language switcher) and we should strive for using the `useTranslation()` hook wherever possible.

[1] considerations
- https://www.locize.com/blog/how-to-use-i18next-t-outside-react-components (happy to attach more github issues related to this if need. I feel `Solution 3` in the blog is a bit messy)
- https://github.com/opengovsg/FormSG/pull/8310#discussion_r2043847311

**Breaking Changes** 

- [ ] Yes - this PR contains breaking changes
    - Details ...
- [x] No - this PR is backwards compatible  

**Improvements**:

- Converted constants in `src/utils/formValidation` to hooks
  - New hooks named accordingly
  - Hook invocation on respective components that use the form validation rules
- Add `useTranslation()` hook invocation
- Created locale files for human-readable text

### Regression
- [ ] Make a form
- [ ] Go to `Settings` > `General`
- [ ] Change form title below min or below max
- [ ] Go to `Settings` > `Email notifications`
- [ ] Add falsy emails
- [ ] Test that the form validation is working (for both email and title validation rules)
- [ ] Make sure it doesn't necessarily rerender (as we're using `useMemo` already)

PR #8310, opened 2025-04-14
feat(i18n): extract text from form validation rules

shagrirse:
## Problem

Closes #8167 

## Solution

**Breaking Changes** 
- [x] No - this PR is backwards compatible

**Improvements**:

- Converted constants in `src/utils/formValidation` to hooks
  - New hooks named accordingly
  - Hook invocation on respective components that use the form validation rules
- Add `useTranslation()` hook invocation
- Created locale files for human-readable text

PR #8314, opened 2025-04-16
feat(i18n): extract text from feedback page

### LINKED ISSUE #8161: feat(i18n): extract text from feedback page
Per title. Further instructions to be given

shagrirse:
## Problem

Closes #8161 

## Solution
- Extracted text from feedback form and related components. `useTranslation` is used in react components.
- The constants for the issues and reviews in `frontend/src/features/admin-form/responses/FeedbackPage/issue/IssueTable.tsx` and  `frontend/src/features/admin-form/responses/FeedbackPage/review/ReviewTable.tsx` are converted into hooks similar to the work done in https://github.com/opengovsg/FormSG/pull/8310#issue-2994292842.
- Created locale files for human-readable text
- The headers of the CSV generated files have also been extracted [1]

[1] Considerations
- the CSV generator class is not a react component, so I used i18next instead (same locales and configuration as what we've already been doing as react-i18next is just a wrapper of i18next)
- As mentioned by @littlemight in https://github.com/opengovsg/FormSG/pull/8311#issue-2995753872, it might cause some issues down the line when a language switcher is implemented. Would like some input on whether this implementation is valid, or if there are any other alternatives available (edit: perhaps checking for language and parsing it into the class when called?).

**Breaking Changes** 

### Regression
- [ ] Create form
- [ ] Leave review(s) and issue(s) for the form
- [ ] Ensure that ICU is properly implemented for plurals
- [ ] Ensure all text fields are in human-readable format for:
  - [ ] ARIA labels
  - [ ] Exported CSV file headers
  - [ ] All texts throughout the feedback page

### DIFF
--- a/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
+++ b/frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
@@ -1,4 +1,5 @@
-import { useCallback, useState } from 'react'
+import { useCallback, useMemo, useState } from 'react'
+import { useTranslation } from 'react-i18next'
 import { UseMutationResult } from 'react-query'
 import { useParams } from 'react-router-dom'
 import {
@@ -10,7 +11,6 @@ import {
   Icon,
   Text,
 } from '@chakra-ui/react'
-import simplur from 'simplur'
 
 import { ProcessedFeedbackMeta, ProcessedIssueMeta } from '~shared/types'
 
@@ -31,8 +31,8 @@ import { useAdminForm } from '~features/admin-form/common/queries'
 
 import { useFormFeedback, useFormIssues } from '../queries'
 
-import { ISSUE_TABLE_COLUMNS } from './issue/IssueTable'
-import { REVIEW_TABLE_COLUMNS } from './review/ReviewTable'
+import { useIssueTableColumns } from './issue/IssueTable'
+import { useReviewTableColumns } from './review/ReviewTable'
 import { EmptyFeedback } from './EmptyFeedback'
 import { FeedbackDownloadButton } from './FeedbackDownloadButton'
 import {
@@ -62,7 +62,54 @@ interface Review extends Feedback {
   download: UseMutationResult<void, Error, DownloadFormFeedbackMutationArgs>
 }
 
+interface ReviewInformationTranslations {
+  averageScore: string
+  reviewsToDate: (reviewCount: number) => string
+}
+
+interface IssueInformationTranslations {
+  tooltip: string
+  issuesToDate: (issueCount: number) => string
+}
+
+interface FeedbackPageTranslations {
+  issues: string
+  reviews: string
+  reviewInformation: ReviewInformationTranslations
+  issueInformation: IssueInformationTranslations
+}
+
 export const FeedbackPage = (): JSX.Element => {
+  // Translations to be parsed into helper functions
+  const { t } = useTranslation()
+
+  const translations: FeedbackPageTranslations = useMemo(
+    () => ({
+      issues: t('features.adminForm.feedback.feedbackPage.issues'),
+      reviews: t('features.adminForm.feedback.feedbackPage.reviews'),
+      reviewInformation: {
+        averageScore: t(
+          'features.adminForm.feedback.feedbackPage.reviewInformation.averageScore',
+        ),
+        reviewsToDate: (reviewCount: number) =>
+          t(
+            'features.adminForm.feedback.feedbackPage.reviewInformation.reviewsToDate',
+            { reviewCount },
+          ),
+      },
+      issueInformation: {
+        tooltip: t(
+          'features.adminForm.feedback.feedbackPage.issueInformation.tooltip',
+        ),
+        issuesToDate: (issueCount: number) =>
+          t(
+            'features.adminForm.feedback.feedbackPage.issueInformation.issuesToDate',
+            { issueCount },
+          ),
+      },
+    }),
+    [t],
+  )
   // Extract form information
   const { data: form } = useAdminForm()
   const { formId } = useParams()
@@ -94,6 +141,10 @@ export const FeedbackPage = (): JSX.Element => {
     download: issueDownload,
     isGetLoading: isIssueLoading,
   }
+
+  // Table column hooks for issue and review tables
+  const issueTableColumns = useIssueTableColumns()
+  const reviewTableColumns = useReviewTableColumns()
 
   // Download button handler
   const handleFeedbackDownloadClick = useCallback(() => {
@@ -154,6 +205,7 @@ export const FeedbackPage = (): JSX.Element => {
             currentFeedbackType,
             issueProps,
             reviewProps,
+            translations,
           )}
         </Box>
         <ButtonGroup gridArea="feedbackType" isAttached variant="outline">
@@ -165,7 +217,7 @@ export const FeedbackPage = (): JSX.Element => {
             sx={{ borderRightWidth: '0px' }}
             onClick={() => setCurrentFeedbackType(FeedbackType.Issues)}
           >
-            Issues
+            {translations.issues}
           </Button>
           <Button
             {...getFeedbackTypeButtonProps(
@@ -174,7 +226,7 @@ export const FeedbackPage = (): JSX.Element => {
             )}
             onClick={() => setCurrentFeedbackType(FeedbackType.Reviews)}
           >
-            Reviews
+            {translations.reviews}
           </Button>
         </ButtonGroup>
         <Box gridArea="export" justifySelf="flex-end">
@@ -198,8 +250,8 @@ export const FeedbackPage = (): JSX.Element => {
           }
           feedbackColumns={
             currentFeedbackType === FeedbackType.Issues
-              ? ISSUE_TABLE_COLUMNS
-              : REVIEW_TABLE_COLUMNS
+              ? issueTableColumns
+              : reviewTableColumns
           }
           currentPage={currentPage - 1}
         />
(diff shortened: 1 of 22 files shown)


littlemight:
@shagrirse hey thanks for being so on the ball with this! I'll probably can only review around monday-ish. Feeling bit under today, had minimal sleep as I had to put out and report a literal fire around my flat last night X_X.

As for the _extracting text constants_ strategy, I was sandboxing a bit for the field validation issue and came up with this early strategy that I think might work:
1. any constant variables in `some/folder/somewhere/constants.ts` should host the keyname instead of the actual hardcoded values. ie `REQUIRED_ERROR = 'This field is required'` becomes `REQUIRED_ERROR = 'mykey.path.requiredError'`
2. then, any react component or close-to-component function/variables: should only ever use `t = useTranslations()` hook and refer the constants above (like `t(REQUIRED_ERROR)`). This might involve some prop-drilling-like approach where we weave in `t` during the inception of a component to whatever downstream constants/vars having human readable text.

More or less like
```
const MyComponent = () => {
 const t = useTranslation();
 const mySuperCoolValidationFunction = validationFn(t)

 return <rules=mySuperCoolValidationFn></rules>
}

// ... then in validationFn
import REQUIRED_ERROR from constants
const validationFn = (t: TFunction) => {
 const myMessage = t(REQUIRED_ERROR)
...
}
```
I think using `useTranslation` is the idiomatic and recommended way by the react-i18n maintainers, this might change as I implement it along the way tho, but I can see this working.

# then: how do we refer those same constants in tests/assertions?
I'm still trying this out, but one approach I can see working is
1. **Use `i18n.t()` only in tests**. The point of using the `useTranslation` hook in components is because react-i18n is hook based, and react needs to know when a language is changed so that it can update the human readable texts accordingly. Thus we avoid the global `i18n.t()`, because react won't know when lang is changed and texts won't update. However in tests, it's just being used for assertion purposes, so using `i18n.t()` is ok....... I think.

```
it('some test', async () => {
   ...

    // Assert
    // Should show error message.
    const error = screen.getByText(i18n.t(REQUIRED_ERROR))
    expect(error).not.toBeNull()
  })
```

2. and you would setup a separate i18n provider for tests
```
// setupI18nTest.ts
i18n.use(initReactI18next).init({
  lng: 'en',
  ...
})
```

3. and update `jest.config.js`
```
module.exports = {
  ...
  setupFilesAfterEnv: ['<rootDir>/src/setupI18nTest.ts']
}
```

tldr; `useTranslation()` hook wherever possible in react components, but global `i18n.t()` in tests. thus constants are still reusable in both places
take with grain of salt tho: I only tried implementing with one of the constant so far, have to see how maintainable this approach is 👻 

but you don't need to worry about this for much now! I might chew on it a bit more during the review later

LoneRifle:
@littlemight - i think it's okay to use hardcoded strings for tests for the time being (you _are_ testing with the assumption that the application is using the en-SG locale after all). the transition to use `i18n.t()` can be done at leisure.

@shagrirse - CsvGenerator header setting might be tricky. Given that there is an implied specific order for the headers, and that the CsvGenerator now doesn't know what language the headers should now be in (or indeed, the name of the headers), it should be the responsibility of the _caller_ to provide the header names. This would allow us to set the headers from outside the CsvGenerator, maintaining clean separation of concerns, and avoiding both calling i18n functions from within a CsvGenerator, and the pitfalls arising from adding a lang switcher.

An alternative is to specify the currently selected locale as a ctor arg to CsvGenerator.

shagrirse:
@littlemight Hope everything is alright with the fire :face_exhaling:
With the current commit, I've drilled the computed translations from the main component down to the functions that use them. It definitely isn't the most elegant solution and it does add more complexity to the component (additional interfaces, etc.) in exchange for better (perhaps negligible) performance through memoization of all translations used for the page. It might not be exactly the same as what you're sandboxing on your end, but I'd like to hear your thoughts on it :smile:
@LoneRifle Thanks for the advice! I think it's best if the computed translations are parsed into the respective CsvGenerators (i.e. IssueCsvGenerator and FeedbackCsvGenerator) due to the reasons you specified. I feel like the alternative of specifying the locale as an arg would be a little more complex than just giving the computed translated headers.

I'll push up another commit along with the changes that come with the review. Thanks for the insights and advice :smile: !

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
```diff
@@ -62,7 +62,66 @@ interface Review extends Feedback {
...
+  issueInformation: IssueInformationTranslations
+  feedbackHeaders: string[]
+  issueHeaders: string[]
+}
+
 export const FeedbackPage = (): JSX.Element => {
+  // Translations to be parsed into helper functions
+  const { t } = useTranslation()
+
+  const translations: FeedbackPageTranslations = useMemo(
```
LoneRifle: the memoisations you have done are keyed on `t()`, but the function reference remains invariant through changes in locale. They may not work as the user intends.

What performance improvements you have in mind? If you were trying to conserve the number of `t()` invocations, would it make more sense perhaps to specify `{ returnObjects: true }`, then destructure and use the object properties as needed, ie something like....
```ts
  const { issues, reviews, averageScore, ... } = t('features.adminForm.feedback.feedbackPage', { returnObjects: true })
  // ....
```

That way, you reduce the number of invocations to once per component use or language change, and retain full language functionality.

wdyt?

LoneRifle: @shagrirse - forgot to tag you, apologies! Could you have a look at my comment?

shagrirse:
@LoneRifle sorry for the late reply!

I initially wanted to limit the invocations of t(), falsely thinking that the function would change if the locale changed (thanks for correcting me!). I think using the `returnObjects` option for the `t()` function like you mentioned and then passing the destructured object properties into the various functions would be the best approach.

#### An alternative(?)
I was also considering passing the `t` function into the other functions as an arg:
```
const FeedbackPage = () => {
  const { t } = useTranslation()
  const { issues, reviews } = t('some.key', { returnObjects: true }) // issues and reviews only needed for FeedbackPage fn
  return (
    // use issues and reviews text somewhere
    {getInformationGridComponent(..., t)} // pass t into another function
    // ...
}
```
and resolving the translations there (and only destructuring what's needed in those functions) as such:
```
const getReviewInformationComponent = (..., t: TFunction) => {
  const { averageScore } = t('another.key', { returnObjects: true })
  // use averageScore somewhere...
```

It would make the code a _little_ more structured and intuitive(?), and requires less typing for the function params, but still calls `t()` a lot more than if we were to just call it once and destructure it at the `FeedbackPage` level.

#### Issue with using the `returnObject` option
**However**, after some research and experimenting, I found no way to use the `returnObjects` option with ICU-formatted plurals, as the destructured locale key would return the unformatted text like this:
```
{
  "issuesToDate": "{issueCount, plural, =1 {issue to date} other {issues to date}}",
  "tooltip": "Feedback displayed here relates to form submission issues"
}
```

We'd thus need to call `t()` specially for the keys that have plurals, as such:
```
const { t } = useTranslation()
const { issues, ... } = t('features.adminForm.feedback.feedbackPage') // detructure non-plural keys
const translations: FeedbackPageTranslations = {
  issues,
  // ... other keys
  issuesToDate: (issueCount) => t('features.adminForm.feedback.feedbackPage.issueInformation.issuesToDate', { issueCount })
}
```

let me know what you think, or if there's something I've missed about being able to resolve plurals while still using `returnObjects`

frontend/src/features/admin-form/responses/FeedbackPage/FeedbackPage.tsx
shagrirse: @LoneRifle sorry for the late reply!

I initially wanted to limit the invocations of `t()`, falsely thinking that the function would change if the locale changed (thanks for correcting me!). I think using the `returnObjects` option for the `t()` function like you mentioned and then passing the destructured object properties into the various functions would be the best approach.

#### An alternative(?)

I was also considering passing the t function into the other functions as an arg:
```
const FeedbackPage = () => {
  const { t } = useTranslation()
  const { issues, reviews } = t('some.key', { returnObjects: true }) // issues and reviews only needed for FeedbackPage fn
  return (
    // use issues and reviews text somewhere
    {getInformationGridComponent(..., t)} // pass t into another function
    // ...
  )
}
```
and resolving the translations there (and only destructuring what's needed in those functions) as such:
```
const getReviewInformationComponent = (..., t: TFunction) => {
  const { averageScore } = t('another.key', { returnObjects: true })
  // use averageScore somewhere...
```
It would make the code a little more structured and intuitive(?), and requires less typing for the function params, but still calls `t()` a lot more than if we were to just call it once and destructure it at the `FeedbackPage` level.

#### Issue with using the `returnObject` option

**However**, after some research and experimenting, I found no way to use the `returnObjects` option with ICU-formatted plurals, as the destructured locale key would return the unformatted text like this:
```
{
  "issuesToDate": "{issueCount, plural, =1 {issue to date} other {issues to date}}",
  "tooltip": "Feedback displayed here relates to form submission issues"
}
```
We'd thus need to call `t()` specially for the keys that have plurals, as such:
```
const { t } = useTranslation()
const { issues, ... } = t('features.adminForm.feedback.feedbackPage') // detructure non-plural keys
const translations: FeedbackPageTranslations = {
  issues,
  // ... other keys
  issuesToDate: (issueCount) => t('features.adminForm.feedback.feedbackPage.issueInformation.issuesToDate', { issueCount })
}
```
let me know what you think, or if there's something I've missed about being able to resolve plurals while still using `returnObjects`

LoneRifle: Oh that's not great. How many ICU keys do we have? Could we perhaps have separate calls for those and bundle the rest under a single object-returning call?

shagrirse: Pushed up a commit that should resolve this! Let me know if all's good :-)

LoneRifle (commented):
lgtm - just need you to resolve the conflicts and we should be good

LoneRifle (approved):
lgtm

PR #8329, opened 2025-04-23
feat: set up stg-alt3 and introduce code changes enabling dual boot into both non-IaC and IaC env 

kevin9foong:
This PR is a continuation of [PR 8025](https://github.com/opengovsg/FormSG/pull/8025) which was closed as we wanted to re-name the branch to `stg-alt3` instead of `staging-alt3` to maintain consistent use of the short name across infra (eg, stack name/env name) and github branch name. 

## Problem
As part of experimentation plan for IaC, we are setting up stg-alt3 which is provisioned using IaC. 
In order to support migration, we will host both IaC and non-IaC simultaneously. Hence, the application code must be able to support both environments. 

Closes FRM-1926

## Solution
- This PR includes GH action scripts to deploy code changes (eg, virus scanner and application code) to the IaC provisioned staging-alt3 infrastructure.
- Remove unused SSM keys 
- Setup if/else check to conditionally load new IaC or non-IaC configurations as appropriate.


.github/workflows/ci.yml
```diff
@@ -2,8 +2,20 @@ name: CI
 
 on:
   push:
+    branches-ignore:
```
scottheng96, line 5: Does this mean that moving forward, when we push to branches other than `develop` we won't run frontend/backend tests & lint checks?

.github/workflows/playwright.yml
```diff
@@ -1,9 +1,22 @@
 name: Playwright Tests
+
 on:
   push:
-    branches: [develop, release-al2, master]
+    branches-ignore:
```
scottheng96, line 5: Similar question to before - we won't run playwright stories tests on branches other than `develop` or feature branches anymore?

src/app/config/features/sgid.config.ts
```diff
@@ -21,17 +25,29 @@ export const sgidVarsSchema: Schema<ISgidVarsSchema> = {
     default: null,
     env: 'SGID_CLIENT_SECRET',
   },
+  privateKey: {
```
scottheng96, line 28: For my learning, how come we had to add these keys for sgid encryption/decryption? Were these the certs originally on EFS

scottheng96 (commented):
Left a few comments on 2-3 things, some for my learning. 

I briefly checked out stg-alt3 and found that submission + attachments weren't working. Taking a look at the  [(logs):](https://ogp.datadoghq.com/logs?query=virus-scanner-stg-alt3-virus-scanner&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1745736665693&to_ts=1745823065693&live=true)

```
User: *::*:*/formsg-stg-alt3-ecs-task-role/f37b2a3dc21d454cbaf068957005f332 is not authorized to perform: lambda:InvokeFunction on resource: *:function:legacy-virus-scanner-stg-alt3:provisioned because no identity-based policy allows the lambda:InvokeFunction action
```

I believe in the infra [code](https://github.com/opengovsg/formsg-infra/blob/uat/stacks/Pulumi.stg-alt3.yaml), the name is referenced here

In `legacyVirusScanner` policy 
image

the function name is as such `...virus-scanner-stg-alt3-virus-scanner...`

I think the change would be inside `legacyVirusScanner.ts` when referencing the arn [here](https://github.com/opengovsg/formsg-infra/blob/uat/src/legacyVirusScanner.ts#L93-L97) (I think!) Hope this helps with the debugging. 

I also realised that legacy virus scanner is still being used over guardduty virus scanner - feature flag for guardduty on stg-alt3 is up, but could be a code drift. Just FYI.

kevin9foong:
> Left a few comments on 2-3 things, some for my learning.
> 
> I briefly checked out stg-alt3 and found that submission + attachments weren't working. Taking a look at the [(logs):](https://ogp.datadoghq.com/logs?query=virus-scanner-stg-alt3-virus-scanner&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1745736665693&to_ts=1745823065693&live=true)
> 
> ```
> User: *::*:*/formsg-stg-alt3-ecs-task-role/f37b2a3dc21d454cbaf068957005f332 is not authorized to perform: lambda:InvokeFunction on resource: *:function:legacy-virus-scanner-stg-alt3:provisioned because no identity-based policy allows the lambda:InvokeFunction action
> ```
> 
> I believe in the infra [code](https://github.com/opengovsg/formsg-infra/blob/uat/stacks/Pulumi.stg-alt3.yaml), the name is referenced here
> 
> In `legacyVirusScanner` policy image
> 
> the function name is as such `...virus-scanner-stg-alt3-virus-scanner...`
> 
> I think the change would be inside `legacyVirusScanner.ts` when referencing the arn [here](https://github.com/opengovsg/formsg-infra/blob/uat/src/legacyVirusScanner.ts#L93-L97) (I think!) Hope this helps with the debugging.
> 
> I also realised that legacy virus scanner is still being used over guardduty virus scanner - feature flag for guardduty on stg-alt3 is up, but could be a code drift. Just FYI.

Thanks for pointing this out. The cause is that didn't redeploy stg-alt3 after `pulumi up` hence the container envs was still using the old lambda name - re-deploying to fix

.github/workflows/ci.yml
kevin9foong: Yep, had to look at this a bit to understand this change i made a while ago - this means that we will not run the tests/lint checks on our staging/uat branches, but will still execute them on the feature/dev branches (Eg, `feat/new-feat` will still run the tests)

Rationale is the staging branches are used to deploy our code on a playground version of FormSG. Since the tests should already be run on feature branches which are where the PRs are made from, might not be a need to run the checks on staging etc to save our GH runners consumption.

.github/workflows/playwright.yml
kevin9foong: Yes - this change was also made following the removed `line 4` which im suspecting the dev who made this `playwright.yml` file probably initially desired for playwright tests to only run on dev, release and master branch (we dont use this branch anymore). However, the syntax above does not work anymore and it was executed unintendedly on all our branches. 

Rationale for ignoring uat and staging branches is same as above ^. Especially since playwright tests are very time-consuming and consume a lot of GH runner resources.

src/app/config/features/sgid.config.ts
kevin9foong: Yep! They were initially a path to a file on EFS, the files have since been converted to string representation and used directly in SSM to avoid having to use EFS. 

There are still some cleanup tasks to do but non-blocking (ie, in a separate PR) discussed with Ken previously (documented in IaC strategy notion page), specifically checking if some files can be non-base64 encoded.

kevin9foong:
ps, i missed the failing backend tests due to the newly added variables - will fix them in a new commit.

kevin9foong:
Opened https://github.com/opengovsg/FormSG/pull/8345 feature branch to run full GH CI testing

scottheng96 (approved):
Okay LGTM. Talked through with Kevin and gained better context on what this PR is in the context of the larger IaC migration.

Expected behaviour will be that nothing will break/change:
- All envs will still run on EB
- stg-alt3 will run on ECS

PR #8340, opened 2025-04-25
feat(i18n): field validation rules

### LINKED ISSUE #8168: feat(i18n): extract text from field validation rules
- Migrate constants in `frontend/src/utils/fieldValidation` to hooks, moving file to `frontend/src/hooks`
- Add `useTranslation()` hook invocation
- Create locale files for human-readable text

littlemight:
## Problem

Partially closes https://github.com/opengovsg/FormSG/issues/8168
Many more field hooks are yet to be converted since some constants are intertwined with tests, non-react code, etc. Have already drafted a change but a bit messy, will commit it after I'm back from my leave.

## Solution
- Convert rule to hooks using `t()`
- Delete single use constants for convenience (since they'll be referenced in `en-sg.ts` anyway)
- For constants used in > 2 places, the constants will be refactored to store the key path pointing to `en-sg.ts` instead, so that it's both reusable in the codebase and also in test (for assertions)
- TBA: add a i18n config for tests so that it can reuse the same constants everywhere, pending tidying up, will be committed after
- Refactors....

**Breaking Changes** 

    - Details ...

**Improvements**:

- Extraction of text as i18next interpolation values

### DIFF
--- a/frontend/src/constants/validation.ts
+++ b/frontend/src/constants/validation.ts
@@ -1,18 +1,3 @@
 export const REQUIRED_ERROR = 'This field is required'
 
 export const INVALID_EMAIL_ERROR = 'Please enter a valid email'
-
-export const INVALID_DROPDOWN_OPTION_ERROR =
-  'Entered value is not a valid dropdown option'
-
-export const INVALID_COUNTRY_REGION_OPTION_ERROR =
-  'Please select a valid country/region from the dropdown list'
-
-// local address field errors
-export const INVALID_POSTAL_CODE_ERROR = 'Please enter a valid postal code'
-export const VALID_POSTAL_CODE_NO_ADDRESS_ERROR =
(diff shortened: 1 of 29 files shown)


littlemight:
@KenLSM let me know if this PR is shaping up ok (structure and code-wise) or not!
mainly my concern would be if we should move the migrated hooks from `utils/fieldValidation` to `frontend/src/hooks` or not.

KenLSM (approved):
Noted on the work that's left to be done i.e., `createBaseValidationRules` that isn't migrated to i18n, which will be taken on as a separate piece of work.

KenLSM:
@scottheng96 FYI there might(?) be some changes on signature Validations. Need to merge with `develop` after this PR is in. Noted that this might slightly add some unavoidable delays to signatures.

PR #8352, opened 2025-04-29
chore: turn off react-refresh on non-relevant files

KenLSM:
## Problem

`react-refresh` eslint rule is applied on files which we actually want to have two exports (i.e., Context / Providers should have its hooks close).

This should help to reduce the noise when reviewing PRs.

## Solution

Detect files that ends with `**/*Context.[jt]s?(x)` or `**/*Provider.[jt]s?(x)`

**Breaking Changes** 

### DIFF
--- a/frontend/.eslintrc.cjs
+++ b/frontend/.eslintrc.cjs
@@ -45,6 +45,12 @@ module.exports = {
         'no-unsafe-optional-chaining': 'off',
       },
     },
+    {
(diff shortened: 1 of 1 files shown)


scottheng96:
From what I learnt reading around:

react-refresh is used for hot reloading React components without losing state.

There’s an ESLint rule (typically from eslint-plugin-react-refresh) that enforces React components to have a single export per file, because multiple exports can confuse Fast Refresh and break hot reloading.

scottheng96 (approved):

PR #8366, opened 2025-05-06
feat: respondent copy v2

scottheng96:
## Problem

Currently FormSG only has functionality to send respondents a copy of their results in Storage mode and through an Email Field setting that is enabled by the user. Multiple MRF admins have requested for the ability to send their respondents a copy of their results as well.

Closes FRM-2020

## Solution

As per [PRD](https://www.notion.so/opengov/Respondent-copy-v2-15a77dbba78880d7937de337088f88c3?pvs=4), feature will include:
- new respondent copy email field input on form page intended to collect respondent copy emails
- new toggle on admin email settings to enable/disable (show/hide) respondent copy field input on form

- email inputs are passed through with submission data as a new variable `respondentEmails` and email confirmation is sent out in a similar flow to existing email confirmation emails (for Storage Mode) + PDF
- for MRF, respondent copy emails will look like workflow completion emails with the refined email format with responses in the form **so far**

- Feature will be released as a pilot with betaFlag `respondentCopy` and feature-flag `respondent-copy` to limit its usage to whitelisted users as well as retain the ability to turn off the feature in the event of issues.

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:
image
image

## Tests

**Able to receive respondent copy on Encrypt Mode**
- [ ] Ensure you are whitelisted for `respondentCopy` on internal tools for the target environment
- [ ] Create an encrypt mode form with 1-2 fields
- [ ] Go to admin settings page, enable toggle `Allow respondents to receive a copy of their submission`
- [ ] Publish form
- [ ] Try to make a submission and include your email in the respondent copy email input
- [ ] See that you are able to receive an respondent copy (email confirmation) of your submission

**Able to receive respondent copy on Multi-respondent Mode**
- [ ] Ensure you are whitelisted for `respondentCopy` on internal tools for the target environment
- [ ] Create an multi-respondent mode form with at least 2 fields, create a workflow with 2 steps (1 field to one 1 step)
- [ ] Go to admin settings page, enable toggle `Allow respondents to receive a copy of their submission` on
- [ ] Publish form
- [ ] Try to make a step 1 submission and include your email in the respondent copy email input
- [ ] Observe that you are able to receive an respondent copy email of your submission
- [ ] Observe that the respondent copy email only contains responses in step 1
- [ ] Try to make a step 2 submission and include your email in the respondent copy email input
- [ ] Observe that you are able to receive an respondent copy email of your submission again
- [ ] Observe that the respondent copy email contains responses in step 1 and 2

**Test `respondentCopy` feature flag**
- [ ] Create a form with 1-2 fields
- [ ] Go to admin settings page, enable toggle `Allow respondents to receive a copy of their submission` on
- [ ] Make form public and open the form; observe that respondent copy email field is present
- [ ] Go to growthbook & turn off feature flag for target env (staging) 
- [ ] Wait ~5 mins, obersve that respondent copy email fields is absent
- [ ] Successfully make a submission for the form


frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -193,7 +205,8 @@ export const FormFields = ({
   )
 
   const hasLockedNormalPrefills = hasLockedPrefills && hasNormalPrefills
-
+  console.log(`hasrespondentcopt: ${form?.hasRespondentCopy}`)
```
kevin9foong: Issue: Potentially stray console logs - should these be removed?

frontend/src/features/admin-form/settings/SettingsService.ts
```diff
@@ -129,6 +129,14 @@ export const updateFormEmails: UpdateEmailFormFn<'emails'> = async (
   return updateFormSettings(formId, { emails: newEmails })
 }
 
+export const updateFormRespondentCopy: UpdateFormFn<
+  'hasRespondentCopy'
+> = async (formId, newHasRespondentCopy) => {
```
kevin9foong, line 134: Suggestion (non-blocking): 

Don't have to change this but a recommendation that you might find useful (and something I personally learnt about from past incidents) is to use object-style arguments eg, {formId, newRespondentCopy}: {formId: string, newRespondentCopy: type}}

Rationale is that it's easier to make sure the params are passed correctly by the function caller, can do {formId: '123', newRespondentCopy: 'abc'} instead of relying on the ordering ('123', 'abc'). 
(Fun fact: our FormSG bounces emails were not sending - causing an incident and it was discovered it was due to mis-ordered arguments, changing to object style fixed this)

frontend/src/features/public-form/components/FormFields/FormFields.tsx
```diff
@@ -164,6 +173,9 @@ export const FormFields = ({
     }
   }
 
+  //default respondent email fields
```
kevin9foong: Nit (non blocking): Shall we add a space after the //? It's an OCD thing haha

src/app/models/form.server.model.ts
```diff
@@ -746,6 +746,12 @@ const compileFormModel = (db: Mongoose): IFormModel => {
         enum: Object.values(Language),
         require: false,
       },
+
+      //boolean value to enable respondent copy emails
```
kevin9foong: Same nit with the space

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
```diff
@@ -1387,6 +1387,141 @@ describe('multirespondent-submission.service', () => {
...
+      expect(sendMrfRespondentCopyEmailSpy).toHaveBeenCalledTimes(1)
+      expect(
+        sendMrfRespondentCopyEmailSpy.mock.calls[0][0].emails,
+      ).toContainValues(respondentEmails)
+      expect(sendMrfRespondentCopyEmailSpy.mock.calls[0][0].emails.length).toBe(
+        2,
+      )
+    })
+
+    it('sends respondent copy on subsequent step submission when respondent emails are present', async () => {
```
kevin9foong, line 1437: Praise: nice work with adding and ensuring good test coverage! Goes a long way in helping us keep FormSG running well

frontend/src/features/admin-form/settings/components/EmailNotificationsSection/RespondentCopyToggle.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const nextHasRespondentCopy = !settings.hasRespondentCopy
+    return mutateFormRespondentCopy.mutate(nextHasRespondentCopy)
+  }, [isLoadingSettings, mutateFormRespondentCopy, settings])
+
+  return (
+    <Skeleton isLoaded={!isLoadingSettings && !!settings}>
+      <Toggle
+        isLoading={mutateFormRespondentCopy.isLoading}
+        isChecked={hasRespondentCopy}
+        label={t(
```
kevin9foong, line 34: Praise: nice work with using the i18n stuff, think this really helps reduce the load on the international team

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -171,16 +174,25 @@ export const performEncryptPostSubmissionActions = (
...
+      const respondentData: AutoReplyMailData[] = respondentEmails
+        ? respondentEmails?.map((val) => {
+            return {
+              email: val,
+              includeFormSummary: true,
+            }
+          })
+        : []
+
       // Send Email Confirmations
```
kevin9foong: Case study: looking at this comment, might not be necessary as proper function and variable naming helped to explain exactly what is happening (which is what is ideal - the code should be clear enough to be understood without comments, barring some low level detail but should be used as a last resort)

src/app/views/templates/MrfWorkflowCompletionEmail.tsx
```diff
@@ -39,16 +41,23 @@ export type WorkflowEmailData = {
...
   formQuestionAnswers = [], 
-  outcome
+  outcome,
+  respondentCopy,
 }: WorkflowEmailData): JSX.Element => {
-  const headingText =  
+  let headingText =  
     outcome ? `The outcome for ${formTitle}.` : `${formTitle} has been completed by all respondents.`
+  
+  // if is respondent copy, replace header
```
kevin9foong, line 57: Suggestion: comment might not be necessary as code below is self explanatory

Good example of using the code itself as documentation, leading to comments not being needed

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -171,16 +174,25 @@ export const performEncryptPostSubmissionActions = (
...
         submission,
         attachments,
         responsesData: emailData?.autoReplyData,
-        recipientData: extractEmailConfirmationData(
-          responses,
-          form.form_fields,
-        ),
+        recipientData: [
+          ...extractEmailConfirmationData(responses, form.form_fields),
+          ...respondentData, // add respondentCopyEmails
```
kevin9foong: Suggestion/nit (non-blocking):
I wonder if we could remove the use of the comment by giving it a more appropriate variable name (eg, ...respondentCopyEmails). By looking at this, we would know immediately what this destructuring is providing without the need for comments.

Rationale is that it's usually a good code quality practice for code comments to focus on explain **why**/special considerations and rely on naming etc to make the code as explicit as possible

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
```diff
@@ -801,6 +862,23 @@ export const performMultiRespondentPostSubmissionUpdateActions = ({
...
+      responses,
+      submissionId,
+      attachments,
+      respondentEmails,
+    }).mapErr((error) => {
+      logger.error({
+        message: 'Send multirespondent respondent copy email error',
+        meta: logMeta,
+        error,
+      }) // return nothing since if successful submission does not depend on this email submission
```
kevin9foong: Case study: This is a good example of comment that is used to explain why, though it'll be good at add some punctuation and potentially make it a little clearer, as it took me a couple reads to understand it.

frontend/src/features/admin-form/settings/mutations.ts
```diff
@@ -296,6 +298,20 @@ export const useMutateFormSettings = () => {
...
   )
 
+  const mutateFormRespondentCopy = useMutation(
+    (nextHasRespondentCopy: boolean) =>
+      updateFormRespondentCopy(formId, nextHasRespondentCopy),
+    {
+      onSuccess: (newData) => {
+        handleSuccess({
+          newData,
+          toastDescription: `Respondents will ${newData.hasRespondentCopy ? '' : 'not'} be able to receive a copy of their submission`,
```
kevin9foong: Suggestion: shall we also use i18n for this?

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -839,11 +842,13 @@ const _createSubmission = async ({
     timestamp: createdTime.getTime(),
   })
 
+  // send respondent copy email to respondentEmails
```
kevin9foong: Nit: I wonder if this comment should be placed here - as it might give the impression that the post actions only does 1 thing which is send respondent copy. 

Shall we place this in the post actions function's JSdocs itself? Assuming this comment is necessary

kevin9foong (commented):
Thanks for the hard work (66 commits!). Looks good and no blocking issues from preliminary review so far. 

Just a couple suggestions on the necessity of comments and if the comments can be potentially replaced by more appropriate variable/function naming instead. Goal of this is to as part of work to improve our codebase quality over time (for this feature and beyond).

In general, comments should be reserved mainly for explaining 'why' a piece of code exists instead of what/how. 
Using of comments to explain 'what' should be reserved for special cases and could be an early sign of code that is hard to understand. 

(Further reading on writing good comments/note that comments also need to be maintained: https://alexkondov.com/comments-the-good-the-bad-the-ugly/)

frontend/src/features/public-form/components/FormFields/FormFields.tsx
kevin9foong: Also, shall we replace this comment by using the variable name? Eg, DEFAULT_EMAIL_FIELD_VALUES = {...}

frontend/src/features/public-form/components/FormFields/PublicRespondentEmailField.tsx
```diff
@@ -0,0 +1,71 @@
...
+            {get(errors, `${RESPONDENT_EMAIL_FIELD_ID}.message`) ? (
+              <FormErrorMessage>
+                {get(errors, `${RESPONDENT_EMAIL_FIELD_ID}.message`)}
+              </FormErrorMessage>
+            ) : (
+              <Text
+                color="secondary.400"
+                textStyle="body-2"
+                aria-hidden
+                my="0.125rem" // same as error message margin
```
kevin9foong, line 59: Good example of a comment that explains why a piece of code exists

src/app/models/form.server.model.ts
kevin9foong: Suggestion: Is this comment necessary? Perhaps we could use clearer naming such as 'isRespondentCopyEnabled' instead

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
```diff
@@ -801,6 +862,23 @@ export const performMultiRespondentPostSubmissionUpdateActions = ({
     submissionId,
   }
 
+  //send respondent copy emails if needed (before rejection/approved etc.)
```
kevin9foong: Suggestion: consider if this comment is needed

frontend/src/features/public-form/PublicFormPage.stories.tsx
```diff
@@ -863,3 +863,18 @@ ThankYouPage.play = async ({ canvasElement }) => {
     },
   )
 }
+
+export const WithRespondentCopy = Template.bind({})
+WithRespondentCopy.parameters = {
+  msw: [
```
kevin9foong, line 869: Praise: great work with mocking the FE for chromatic stories, this will ensure respondent copy is well tested for the rest of time!

frontend/src/features/admin-form/settings/SettingsService.ts
scottheng96: Good point for this input. Though it will be tighter in terms of function handling, I think I lean towards keeping the update functions consistent syntax-wise within mutations.ts. Here are the other examples I'm following.

```
export const updateMrfEmailNotifications: UpdateMultiRespondentFormFn<
  MrfEmailNotificationSettings
> = async (formId, newSettings) => {
  return updateFormSettings(formId, newSettings)
}

export const updateFormAuthType: UpdateFormFn<'authType'> = async (
  formId,
  newAuthType,
) => {
  return updateFormSettings(formId, { authType: newAuthType })
}
```

Also UpdateFormFn specifies the input variables like so:
```
type UpdateFormFn<T extends keyof FormSettings> = (
  formId: string,
  settingsToUpdate: FormSettings[T],
) => Promise<FormSettings>
```

frontend/src/features/admin-form/settings/mutations.ts
scottheng96: Yeah, lets do it and give our future selves an easier life. Made this change

frontend/src/features/public-form/components/FormFields/FormFields.tsx
scottheng96: Might be overkill to save a DEFAULT_EMAIL_FIELD_VALUES as an empty [], but I'll remove the comments since the code is self-explanatory.

frontend/src/features/public-form/components/FormFields/FormFields.tsx
scottheng96: Removed this.

src/app/models/form.server.model.ts
scottheng96: Removed this comment since good point hasRespondentCopy is pretty self explanatory.

src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
scottheng96: Removed this comment since it is unecessary - I think this was for my personal tracing of the code during implementation.

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
scottheng96: Removed this comment! (It was Ken's 2 years ago lol)

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
scottheng96: Removed this comments and renamed variable to `respondentCopyEmailData` for tighter explicit naming

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
scottheng96: removed!

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
scottheng96: Updated this comment with punctuation

src/app/views/templates/MrfWorkflowCompletionEmail.tsx
scottheng96: This im leaving in since I'm using the WorkflowCompletionEmail for Respondent Copy emails and it might not be intuitive which content is being displayed for which email type.

kevin9foong:
Pending design approval on chromatic + fixing of the comment below before merging, otherwise code lgtm - so i approved it first

kevin9foong (approved):

frontend/src/features/public-form/components/FormFields/PublicRespondentEmailField.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <Box>
+            <FormLabel>
+              {t('features.publicForm.components.fields.respondentEmail.title')}
+            </FormLabel>
+            <TagInput
+              {...field}
+              value={field.value as string[]}
+              tagValidation={isEmail}
+            />
+            {get(errors, `${RESPONDENT_EMAIL_FIELD_ID}.message`) ? (
```
kevin9foong, line 50: Issue: we dont need the condiitional ` {get(errors, `${RESPONDENT_EMAIL_FIELD_ID}.message`) ?` here. The `FormErrorMessage` should appear based on the `FormControl`'s `isInvalid` prop which will show the `FormErrorMessage` if `isInvalid` evals to truthy. 

(Can reference: frontend/src/components/SecretKeyVerificationInput/SecretKeyVerificationInput.tsx for usage of `FormErrorMessage` and `FormControl`) 

Thus, the ternary operator should be removed

kevin9foong: Hey scott, 1 final issue before merging related to `FormControl`

frontend/src/features/public-form/components/FormFields/PublicRespondentEmailField.tsx
```diff
@@ -0,0 +1,71 @@
...
+    },
+  }
+
+  const {
+    control,
+    formState: { errors },
+  } = useFormContext<{ respondent_email_field: string[] }>()
+
+  return (
+    <FormControl isInvalid={!!errors?.[RESPONDENT_EMAIL_FIELD_ID]}>
```
kevin9foong: Related to the below,  since the `isInvalid` here will evaluate to truthy value when `get(errors, `${RESPONDENT_EMAIL_FIELD_ID}.message`)` exists, thus the FormResponseError will show up without need for the conditional check with ternary operator.

kevin9foong: Issue: should be updated to `!!errors?.[RESPONDENT_EMAIL_FIELD_ID].message` to match the conditional check

frontend/src/features/public-form/components/FormFields/PublicRespondentEmailField.tsx
scottheng96: Implementation is as such because either the error message or the helper text will be displayed underneath the input field, I could also do it similar to 
https://github.com/opengovsg/FormSG/blob/develop/frontend/src/features/admin-form/settings/components/FormEmailSection.tsx#L144-L155

but i thought this is clear as well

frontend/src/features/public-form/components/FormFields/PublicRespondentEmailField.tsx
scottheng96: Made this change. Thanks!

frontend/src/features/admin-form/settings/SettingsService.ts
KenLSM: Just as an exercise in code smell / readability, there are a few considerations before I would **very strongly** recommend object-style arguments. This is not a discussion about the above code, just taking the opportunity to explore this a bit more.

1. There are optional parameters in the function call. Nobody likes to write functions like `foo(8171, null, null, 'hello')`
2. There are more than 7 parameters in the function call. 7 is an arbitrary number that I chose. When the arguments are long, for object-style we can toss them into an object and pass it in, but for non-objects, it is a bit risker to do a spread operation.

PR #8371, opened 2025-05-06
fix: update findPaymentById to always have primary read preference

scottheng96:
## Problem
Stripe is sending us 2 `charge.succeeded` events, causing us to have to handle duplicated events. This is because the 1st `charge.succeeded` event is being failed to process due to (hypothesis) database node replication lag.

`payments.completedPayments` is written into the primary database node, but when attempting to read it again downstream in `performPaymentPostSubmissionActions` in `findPaymentById`, our application is unable to find it because there is a possibility that it spins up a session on a secondary node that does not yet have the updates from the primary node.

Closes FRM-2019

## Solution
Force read of `completedPayments` to always be on the primary node by referencing `readPreference: primary`

**Breaking Changes** 

## Tests

**Make a successful payment submission**
- [ ] Create a form with payment or use an existing payment form
- [ ] Attempt to make a payment submission
- [ ] Observe that you are successfully able to make payment and submit the form

### DIFF
--- a/src/app/modules/payments/payments.service.ts
+++ b/src/app/modules/payments/payments.service.ts
@@ -50,7 +50,9 @@ export const findPaymentById = (
       paymentId,
       null,
       // readPreference from transaction isn't respected, thus we are setting it on operation
-      session ? { session, readPreference: 'primary' } : null,
+      session
+        ? { session, readPreference: 'primary' }
+        : { readPreference: 'primary' },
     ),
     (error) => {
       logger.error({


src/app/modules/payments/payments.service.ts
```diff
@@ -50,7 +50,9 @@ export const findPaymentById = (
       paymentId,
       null,
       // readPreference from transaction isn't respected, thus we are setting it on operation
-      session ? { session, readPreference: 'primary' } : null,
+      session
```
kevin9foong, line 53: Issue: seems like there is a ternary where both paths lead to the same outcome.

Seems like it might add confusion for future developers looking at this code.

src/app/modules/payments/payments.service.ts
```diff
@@ -50,7 +50,9 @@ export const findPaymentById = (
...
+        ? { session, readPreference: 'primary' }
```
kevin9foong, line 54: Suggestion: my understanding is that when the payment id is read, the suspicion is that the changes have not yet been replicated to the secondary database nodes. 

Hence, we are forcing a read to the primary node which is where the payment document is first being written to.

Perhaps we should place a comment to explain this here. 

Also, there is a caveat where this could potentially lead to increased uneven load distribution to our primary node.

src/app/modules/payments/payments.service.ts
kevin9foong: Prefer if the ternary expression is removed entirely

kevin9foong (changes requested):

kevin9foong:
Discussed with Scott an alternative solution to pass the `completedPayment` document data directly from the txn to the `performPaymentPostSubmissionActions` instead of making a second database call, which would avoid the need to make a duplicate database query on primary node. 

Rationale that was given for going with this current approach was discussed as it is: 
- less code change involved than the alternative 
- fine to make secondary database query to the primary node since the expected number of payment volume is small, so the expected impact on primary is low 
- not 100% certain that the database secondary node replication lag is the cause

Approving in light of the above rationale

src/app/modules/payments/payments.service.ts
scottheng96: This is used sometimes when the session is passed through (writing to DB is done within a session), whereas sometimes if session isn't created yet this function creates one

kevin9foong (approved):

PR #8379, opened 2025-05-08
feat: support ssm undefined placeholder and update task def

kevin9foong:
## Problem

For some of our env vars, we do not pass in env vars (leave it as undefined) and use the default values defined in the convict schema. 
However, this does not work for ECS as the task definition expects the SSM params to exist at runtime (does not support loading conditionally based on if the SSM params exist), so SSM params must be created, which values must not be an empty string as empty strings are not supported by SSM.

## Solution

Hence, `__UNDEFINED__` is used as a placeholder to store into SSM params as values, which our application code will  interpret by `convict` as undefined and the `default` values are loaded. 

If undefined placeholder string is detected, it will reset the value to default. This approach is necessary due to the [precedence order of Convict](https://github.com/mozilla/node-convict/tree/master/packages/convict#precedence-order). 

image 

**Breaking Changes** 


scottheng96:
LGTM 👍

scottheng96 (approved):

PR #8385, opened 2025-05-14
feat: convert email to storage mode directly

kevin9foong:
## Problem

In order to drive deprecation of email mode, need a way to: 
- retain form url to minimise disruption to admins who may not be able to easily update their form touchpoints. 
- allow ops to force disable email mode forms
- allow gradual tracking of closed forms towards our OKRs. 

Closes FRM-2021

## Solution

1. Implement flag that causes email forms to have to convert to storage mode before they can set their form to public. 
2. Admins will need to download their form secret and convert the form to storage via a modal before being able to activate their previously email mode forms.  (Note that under the hood, this conversion is a delete and re-create of a form with the same id required since form `responseMode` is the mongoose discriminator) 
3. Count only the submissions matching the form type (ie, after form conversion to storage mode, the email mode submissions are no longer displayed/included in the submission response count). This decision was explicitly made to avoid over-optimizing and prevent having this legacy state (where both email + storage mode counts are displayed) from lingering in the product.    

**Breaking Changes** 

## Before & After Screenshots
**AFTER**:

Modal for converting to storage mode form with same settings and retained/same form id when toggle activate form is clicked.   
image

Toast after the conversion is complete - note that the form mode is now `storage mode`. 
image

## Tests

TC1: Email mode form can still be edited 
- [ ] Find an email mode form and set the `isForceConvertToStorageMode` and set form status to `PRIVATE`. 
- [ ] Edit form settings (singpass settings, captcha etc) 
- [ ] Add and edit form fields 
- [ ] Assert that the above operations can occur without any error. 

TC2: Refreshing before converting does not change anything and email form still works. 
- [ ] Try to set form to public. 
- [ ] Assert the convert to storage mode modal shows up. 
- [ ]  Refresh the page without clicking on confirm. Assert that the form is still email mode and can still add fields/change settings etc.

TC3: Submission limit works and resets after converting to storage mode. 
- [ ] Set the submission limit of form to 1 more than current # of email mode submissions. 
- [ ] Submit a response to email mode form. Observe that form is closed as the submission limit is hit and then responses count in the result page increases by 1. 
- [ ] Convert the form to storage mode via the modal. 
- [ ] After conversion, go to the responses page and observe the responses count is now 0. 
- [ ] Set the form as public. Assert this is now possible since form is storage mode. 
- [ ] Set the response limit to 2. 
- [ ] Make a first submission. Assert that form is not closed (despite the previous submission in email mode) 
- [ ] Make a second submission. Assert that the form is now closed (since the number of storage mode submissions = 2).


kevin9foong:
Code is ready apart from fixing the failing TC and writing new TC to test updated behavior.

frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/useSaveSecretKey.tsx
```diff
@@ -0,0 +1,120 @@
+import {
```
kevin9foong, line 1: decouple the keypair generation from the copy/download/mailto functionality so it can be reused by email mode migration modal

frontend/src/features/admin-form/settings/components/EmailModeConvertModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+      onSuccess: () => {
+        toast.closeAll()
+        toast({
+          description: t(
+            'features.adminForm.toasts.emailModeMigration.success',
+          ),
+          status: 'success',
+          isClosable: true,
+        })
+        queryClient.invalidateQueries(adminFormKeys.id(formId))
```
kevin9foong, line 75: required to flush tanstack cache to reflect the change in form response mode

src/app/models/form.server.model.ts
```diff
@@ -341,8 +341,34 @@ const EmailFormSchema = new Schema<IEmailFormSchema, IEmailFormModel>({
...
+  publicKey,
+}: {
+  publicKey: string
+}) {
+  const FormModel = mongoose.model(FORM_SCHEMA_ID)
+
+  const emailModeFormData = this.toObject()
+  const emailModeFormId = this._id
+
+  await this.deleteOne()
```
kevin9foong: note that a deletion before re-creation is necessary since the responsemode is used as the mongodb discriminator, hence, need to re-init to avoid discriminator related schema validation issues.

frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/SaveSecretKeyScreen.tsx
```diff
@@ -64,250 +49,22 @@ const useSaveSecretKeyDefault = () => {
...
-              {t('features.workspace.modals.create.secretKey.confirm')}
-            </Text>
-          </Button>
-        </Container>
-      </ModalBody>
-    </>
+    <SaveSecretKeyContent
+      contentTitle={t('features.workspace.modals.create.secretKey.title')}
+      registerStorageAck={
+        // HACK: Since CreateFormWizardInputProps must have storageAck, we can cast it to SaveSecretKeyFormInput.
```
kevin9foong: SaveSecretKeyFormInput is a subset of CreateFormWizardInputProps, where SaveSecretKeyFormInput only requires `storageAck`, but this is added as having issues resolving the type due to the UseFormRegister

kevin9foong: Open to any suggestions on a cleaner way to resolve this typescript issue apart from typecasting to `as unknown as ...`

kevin9foong:
Automated tests have been written for the following cases: 
• chromatic story for the new modal 
• test replacement of the storage mode form 
• test submission counting filters by submissionType 
• test submission counting works to filter all when no submissionType provided 
• test the controller passes the submission type correctly 
• test must not have public key before converting 
• test cannot open if form is email and isConvert

frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/SaveSecretKeyScreen.tsx
kevin9foong: https://github.com/orgs/react-hook-form/discussions/8418

Instead of doing the above (and having the `required: true` defined in the parent, outside `SaveSecretKeyContent` or using useFormContext to pass the `register` and requiring a provider, i think it is cleaner to just typecast to the register type.

frontend/src/features/admin-form/settings/components/EmailModeConvertModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+    base: 'mobile',
+    xs: 'mobile',
+    md: 'full',
+  })
+
+  const convertEmailToStorageModeMutation = useMutation(
+    convertEmailToStorageMode,
+    {
+      onSuccess: () => {
+        toast.closeAll()
```
scottheng96, line 67: Are there other toasts that get displayed onSuccess? Otherwise this could be unnecessary

frontend/src/features/analytics/AnalyticsService.ts
```diff
@@ -201,3 +201,16 @@ export const trackClickSecretKeyMailTo = (formTitle: string) => {
     form_title: formTitle,
   })
 }
+
+/**
+ * Logs clicking on mailto link to share form secret key with collaborators for email mode form that has been converted to storage mode form.
+ */
+export const trackClickSecretKeyMailToEmailToStorageConvertedForm = (
```
scottheng96, line 208: Nice! Easy way to track its usage

frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/SaveSecretKeyScreen.tsx
```diff
@@ -55,7 +37,10 @@ const useSaveSecretKeyDefault = () => {
 
   const queryClient = useQueryClient()
   const navigate = useNavigate()
-  const formId = queryClient.getQueryData(workspaceKeys.lastCreatedForm)
+
+  const formTitle = useWatch({ control, name: 'title' })
+  const formId: string =
+    queryClient.getQueryData(workspaceKeys.lastCreatedForm) ?? ''
```
scottheng96, line 43: Seems like there could be a scenario where formId is `''`, which gets passed into `SaveSecretKeyContent` and used in the hook. I might have missed some context/behaviour here - should it be the case that `formId` is always a 24-char uuid and if it's not found we deal with it?

src/app/models/__tests__/form.server.model.spec.ts
```diff
@@ -1002,10 +1003,57 @@ describe('Form Model', () => {
...
+        const pastSettings = omit(emailFormToReplace.toObject(), [
+          '__v',
+          'created',
+          'lastModified',
+          'responseMode',
+          'isForceConvertToStorageMode',
+        ])
+        expect(newStorageModeForm).toMatchObject(pastSettings)
+      })
+
```
scottheng96, line 1057: Nice well written tests sir

src/app/models/form.server.model.ts
scottheng96: I see that this is necessary to be able to convert email forms to storage mode forms under the same ID. Would it be correct to say that there is some (little) risk where we the email mode form setting is successfully deleted -> the new storage mode form setting is unsuccessfully saved -> we lose the form setting data?

src/app/modules/core/core.errors.ts
```diff
@@ -45,6 +45,7 @@ export enum ErrorCodes {
   ADMIN_FORM_MODEL_RESPONSE_INVALID_SCHEMA_FORMAT = 100113,
   ADMIN_FORM_MODEL_RESPONSE_FAILURE = 100114,
   ADMIN_FORM_MODEL_GET_CLIENT_FAILURE = 100115,
+  ADMIN_FORM_FORM_ALREADY_HAS_PUBLIC_KEY = 100116,
```
scottheng96: Good practice that I will adopt moving forward as well for more granular error handling 👍

src/app/models/form.server.model.ts
kevin9foong: yes - potentially, unless we wrap this in a txn perhaps so it can rollback

frontend/src/features/admin-form/settings/components/EmailModeConvertModal.tsx
kevin9foong: This is necessary as it follows the convention and default behavior for the other toasts eg, `FormSG/frontend/src/features/admin-form/common/mutations.ts` which do a `toast.closeAll()` before, so that it will override any current toasts (from other actions) and display this new toast at the top.

scottheng96 (commented):
Looks generally good and written cleanly. If i could suggest adding a few more test cases not related to Email -> Storage conversion, but general form creation/handling for storage mode + MRF forms. 

Since some refactoring was done for the `SaveSecretKeyScreen`, could be useful to be safe with some regression tests!

src/app/models/form.server.model.ts
kevin9foong: Added a txn to rollback in the event of partial failure

frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/SaveSecretKeyScreen.tsx
kevin9foong: In this case, i think its fine. 


Context: 
This form id is used mainly for generating the file name of the secret key text file eg, `secretKey_<formId>_<formname>.txt` 

Tracing the previous code, the `const formId = queryClient.getQueryData(workspaceKeys.lastCreatedForm)` could have returned `undefined` which was not handled, potentially leading to `secretKey_undefined_<filename>.txt` being the file name. 

In this case, if the form name is not known, it would generate ``secretKey__<formName>.txt` instead, which seems fine as well. 

Rationale: 
It is currently the same as existing behavior (out of scope for this PR) except that instead of `undefined`, it just leaves the unknown form id as empty string.

kevin9foong:
> Looks generally good and written cleanly. If i could suggest adding a few more test cases not related to Email -> Storage conversion, but general form creation/handling for storage mode + MRF forms.
> 
> Since some refactoring was done for the `SaveSecretKeyScreen`, could be useful to be safe with some regression tests!

I'm thinking perhaps the current tests are sufficient. 

Adding tests not related to Email -> Storage mode conversion would be out of scope for this PR (especially since no changes are made to the storage mode creation/handling). To cover new changes, I have written tests covering any code introduced in this PR.  

For the hooks refactor, there is already an existing Chromatic story for `SaveSecretKeyScreen`.

Added some tc for the admin-form.controller

kevin9foong:
Based on the comments, decided to a txn to the `__replaceWithStorageModeFormWithSameId` 

Agree that despite the low chance of partial failure, the consequence could be potentially disastrous (bunch of submissions with no form tied to, and not able to recover). Hence, added a transaction as a safeguard (rollback during partial fail). 

Additionally, the performance tradeoff of adding a transaction is expected to be low since the convert operation is limited to # of remaining email mode forms and will not grow.

frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/SaveSecretKeyScreen.tsx
scottheng96: If we aim to bypass this quirky typecasting, we could obtain register from `useForm<SaveSecretKeyFormInput>()` in place of `useCreateFormWizard()`? Seems to work on local haha

```
const { register } = useForm<SaveSecretKeyFormInput>()
...
  return (
    <SaveSecretKeyContent
      contentTitle={t('features.workspace.modals.create.secretKey.title')}
      registerStorageAck={register}
      isLoading={isLoading}
      formTitle={formTitle}
...
```

scottheng96:
Just an FYI: 

Found out from local testing that FormKeys extension doesn't save the secretKey during the process of email form -> storage form conversion. Doesn't affect most form admins I believe & neither does us since we won't be converting and using forms in an active manner.

scottheng96 (approved):

kevin9foong:
> Just an FYI:
> 
> Found out from local testing that FormKeys extension doesn't save the secretKey during the process of email form -> storage form conversion. Doesn't affect most form admins I believe & neither does us since we won't be converting and using forms in an active manner.

I'm not sure why it doesn't catch, might be due to Formkeys not recognizing modal name

PR #8397, opened 2025-05-20
fix(submit): reduce log pollution by only logging when email is sent

littlemight:
## Problem
`Sending admin notification mail` is sent even when `MailService.sendSubmissionToAdmin` is not called.

context: was debugging a possible email bounce issue on Lao deployment of Form. got misled by this log.

## Solution
Only log when email is actually sent.

**Breaking Changes** 

    - Details ...

**Improvements**:

- Proper logging

## Tests
- No functional logic change (thus no changed test)
- This is an observability change

## Deploy Notes
- This would reduce the number of occurrence of `Sending admin notification mail` in backend logs, since previously it logs on every submission. Now it only logs when only when email list is configured.
- Need to highlight this explicitly to not be surprised of the occurrence drop if this PR does get merged in

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -782,11 +782,6 @@ const _createSubmission = async ({
     responseMetadata,
   }
 
-  logger.info({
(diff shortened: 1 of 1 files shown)


KenLSM (approved):
LGTM!

PR #8402, opened 2025-05-21
feat: add error span tag for dd retention filter to index

kevin9foong:
## Problem

Currently, DataDog does not index all spans which include errors (this is due to error spans being marked as `status: ok` which is the default). This leads to some error spans not occurring in trace searches and inability to use APM Trace Analytic monitors as they can only filter indexed spans.

Currently, only our webhook failures are caught by the retention filter. 

Closes FRM-2022 

## Solution

Set a status: error tag for error spans so that they are indexed by the Default Error retention filter.   

Note: DD costs may increase due to the increased number of indexed spans, but this is only limited to error spans which should be indexed and the $ amount is expected to be marginal (est 2mil/month additional indexed spans) for immediate incident notifications and specific incident messages. 

**Breaking Changes** 

### DIFF
--- a/src/app/modules/datadog/datadog.utils.ts
+++ b/src/app/modules/datadog/datadog.utils.ts
@@ -22,6 +22,7 @@ export const setFormTags = (form: IPopulatedForm) => {
 export const setErrorCode = (error: ApplicationError) => {
   const span = tracer.scope().active()
   if (span && error.code) {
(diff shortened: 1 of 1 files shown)


KenLSM (approved):
LGTM. Should help to flak out more errors with improved observability -- a good problem to have.

PR #8403, opened 2025-05-21
feat: add iac tag

kevin9foong:
## Problem

Currently, on DD, there is no reliable way to identify spans from pre-IaC and IaC deployments (only can use the log source: kinesis-db, but this is not as reliable and could lead to filtering issues). 

We need a way to distinguish the spans between iac and pre-iac deployments for observability and monitoring in the event of failures. 

FRM-2023

## Solution

Add a DD tag for filtering spans / logs for iac env. 

**Breaking Changes** 


KenLSM (approved):

kevin9foong:
Verified functionality on stg-alt3, squashing

PR #8408, opened 2025-05-22
feat: add specific error messages for guardduty virus scan

kevin9foong:
## Problem

The PR solves 3 issues:

1. (main objective) Currently, our error codes do not distinguish between ClamAV and GuardDuty virus scan failures. As a result, we cannot set APM trace analytic monitors for each virus scanner separately (without using the not as maintainable log monitors).

2. There is a clamAV scanner error that has not been covered by the errorMapper in submission.utils.ts. This leads to the fallback case being triggered and unknown route error to be potentially logged. They should be included as a case.

3. The error types are not accurately typed in the virus scanner function return types. This is likely since they are hardcoded instead of using a shared variable, leading to unmaintainable code. This also likely contributed to issue 2. 

Closes FRM-2024

Solution

1. Add specific error codes for guard duty to distinguish from ClamAV virus scanner errors for DD filtering. This is implemented using exception wrapping (this is a popular pattern used in Java).

2. Add the missed error case ParseVirusScannerLambdaPayloadError to the errorMapper.

3. Use shared TS type constants to prevent need to update across multiple files.

**Breaking Changes** 

## Tests

TC1: Virus scanner works [Regression] 
- [ ] Make a submission with attachment field. 
- [ ] Ensure it works without error.  

Verified on staging-alt3 by simulating virus scanner failures to test error code piping to DataDog


src/app/modules/submission/submission.service.ts
```diff
@@ -1403,5 +1420,16 @@ export const triggerGuarddutyScanThenDownloadCleanFileChain = <
           answer: response.filename,
         })),
       )
+      .mapErr((error) => {
```
kevin9foong, line 1423: Exception wrapping - this allows us to wrap the non-gd errors and re-throw as a gd error, which is a common pattern used in Java.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -195,9 +192,8 @@ const asyncVirusScanning = (
   enableGuarddutyLambdaInvoke: boolean | undefined,
 ): ResultAsync<
   IdTaggedParsedClearAttachmentResponseV3,
-  | VirusScanFailedError
```
kevin9foong, line 198: Use TS type constants instead of hardcoding so that the upstream functions are updated when the error types change - to avoid missed error types (note that the previous error type was not updated and missed some potential error types due to the hardcode)

src/app/modules/core/core.errors.ts
```diff
@@ -70,12 +70,16 @@ export enum ErrorCodes {
...
   SUBMISSION_INVALID_WORKFLOW_TYPE = 100221,
   SUBMISSION_ATTACHMENT_UPLOAD = 100222,
+  SUBMISSION_EMAIL_HASH = 100223, // Email submission error (email mode deprecated soon)
   SUBMISSION_EXPECTED_RESPONSE_NOT_FOUND = 100224,
   SUBMISSION_SAVE_FAILURE = 100225,
   SUBMISSION_MRF_REMINDER_INVALID_PENDING_STEP = 100226,
   SUBMISSION_MRF_REMINDER_RECIPIENT_EMAILS_EMPTY = 100227,
-  // Email Submission Errors (email mode deprecated soon)
-  EMAIL_SUBMISSION_HASH = 100223,
+  SUBMISSION_GD_INVALID_FILE_KEY = 100228,
```
kevin9foong, line 78: Add new error codes and error classes to distinguish Guardduty errors from ClamAv errors

src/app/modules/submission/submission.errors.ts
```diff
@@ -167,12 +167,26 @@ export class InvalidFileKeyError extends ApplicationError {
   }
 }
 
+export class GuardDutyInvalidFileKeyError extends ApplicationError {
```
KenLSM, line 170: /nit i see a couple of casing for `GuardDuty` vs `Guardduty` in this PR. Would prefer `GuardDuty` if we're renaming the new additions.

src/app/modules/submission/submission.service.ts
KenLSM: Not too sure about this since each of the `DownloadCleanFileFailedError` (the error that's being wrapped) would have also done a `span.setTag`.

I believe the wrapping would make sense to wrap primitive errors rather than ApplicationError.

KenLSM: Looked around, couldn't find anything better. Ok with solution. But not something we want to do more often (overriding error thrown)

KenLSM (approved):

src/app/modules/submission/submission.service.ts
kevin9foong: I actually tried to directly throw the new errors from the source eg, from `downloadCleanFile` and this exception wrapping is a last resort. 

The problem was that this function is shared with `triggerVirusScannerAndScan` and `triggerGuardDutyVirusScannerAndScan` and i will either need to: 
- duplicate the function to create a guard duty version 
- pass in a flag `isGuardDuty` to conditionally throw the error based on this flag

These 2 approaches don't seem too good either - and require more invasive changes. 

Agreed that this is not something we want to do regularly, but due to the nature of the virus scanner code, think this is the least invasive way to add this functionality.

src/app/modules/submission/submission.errors.ts
kevin9foong: Okay, will change the casing to match `GuardDuty`

PR #8412, opened 2025-05-23
fix: only send mrf rc when has respondent emails

scottheng96:
## Problem

Currently, if there is an mrf submission without respondent copy emails, an attempt to send out `sendMrfRespondentCopyEmails` will occur because the `respondentEmails` is an empty `[]`, which passes the `if (respondentEmails)` check.

## Solution
Additionally check if respondentEmails has `length > 0`.

More context. Initial attempt was made to set `respondentEmails` defaultValue as `undefined`. However, if a user interacts with the input field (e.g. add an email and then remove it), `respondentEmails` with stick to an empty `[]`; this needs to be handled downstream as well. Therefore check happens right before respondent copy emails are sent out.

**Breaking Changes** 


KenLSM (approved):
LGTM!

`Boolean[]` is `true`. Just javascript things

PR #8436, opened 2025-05-29
chore: mongoose version upgrade

scottheng96:
## Problem

MongoDB v6 is reaching EOL on 31 July 2025. There is need to upgrade MongoDB to v7. Since FormSG uses mongoose, there is also a need to bump mongoose to v7.8.0.

Since Mongoose v7 is compatible with MongoDB v6, we are able to upgrade mongoose first and then mongodb subsequently.

Closes FRM-2036

## Solution

- Upgrade Mongoose to v7.8.0
- Resolved build issues and breaking changes.

Topic | Location | Solution
-- | -- | --
Removed LeanDocument and support for extends Document | form.server.model.ts,form.ts,myinfo.util.ts,myinfoField.ts | Replace LeanDocument with FlattenMaps
Removed Support for custom promise libraries | /app/config/config.ts | Remove promiseLibrary and by default use native ES6 promises  (used to be mpromise, but has since been deprecated in favor of ES6)
Unused @ts-expect-error | session.ts | @ts-expect-error was used to temporarily supress issue where ts definitions didn't match runtime behaviour in earlier versions of mongoose
TypeError: Invalid schema configuration: 'false' is not a valid type at path 'id' (test error) | form.server.model.ts | Add {id: false} as a second argument to EncryptedFormSchema.  This (1) informs mongoose not to create the virtual id getter (string version of _id (2) ensure that id: false is passed as a schema option & not mixed into the schema fieldsThe id setting was removed in Mongoose 8 due to compatibility issues
Cannot find module 'mongodb' from 'node_modules/connect-mongo/build/main/lib/MongoStore.js’ | session.ts | remove connect-mongo and mongoose from package.jsonrun npm irun npm i "connect-mongo" "mongoose@^7.4.0”mongodb was then installed as a top level package, with the same 5.9.2 version being installed ➜  FormSG git:(chore/mongoose-version-upgrade) ✗ npm why mongodbmongodb@5.9.2node_modules/mongodbmongodb@"^5.9.2" from mongodb-memory-server-core@9.5.0node_modules/mongodb-memory-server-coremongodb-memory-server-core@"^9.5.0" from the root projectmongodb@"5.9.2" from mongoose@7.8.7node_modules/mongoosemongoose@"^7.8.7" from the root project
typeError: Invalid schema configuration: false is not a valid type at path id | form.server.model.ts | replace { _id: { id: false } } by putting { _id: false } into schemaOptions (2nd argument)
TypeError: bson_1.ObjectID is not a constructor | user.service.spec.ts | ObjectId bsontype now has lowercase dreplace ObjectID with ObjectId
TypeError: Class constructor ObjectId cannot be invoked without 'new’ | analytics.service.spec.ts | In Mongoose v7, Types.ObjectId is a class and must be instantiated with newprefix variable with new
TypeError: mockTransaction.fields[1].remove is not a function | public-forms.verification.routes.spec.ts | Removed remove()Replace remove() with deleteOne()
should return a submission document with a different ObjectID if an ObjectID collision occurs while copying the pending submission to the submissions collection | submission.service.spec.ts | pendingSubmission was a Mongoose document, not a plain JS objectlodash omit did not the properties like _id etc. Mongoose always includes _id as part of the document instance, even if you try to strip it — unless you explicitly convert it to a plain objectv6 was more forgiving creating documents with unexpected or extra fields (e.g. _id) and would silently ignore internal fields. v7 is stricter how it handles fields passed into new Model(). v7 expects clean, plain objects when creating new documents — and doesn't automatically sanitize or override internal fields like _idconvert pendingSubmission.toObject()

**Breaking Changes** 

- Yes. PR contains removal of deprecated Mongoose functions + replacement with current functions.
   Examples are described above.

## Deploy Notes

Be aware that mongoose upgrade might cause some breaking changes for FormSG when interacting with the DB. Rolling back mongoose package version is a viable solution.


KenLSM (changes requested):
Ok-ed on code changes. To update PR body

KenLSM (approved):
LGTM! Discussed that the package-lock reflecting downgrade of packages is due to the usage of the default greedy diffing algorithm. Comparing it with non-greedy algo reveals that the packages were removed (due to no usage of deps) instead of downgrade.

PR #8476, opened 2025-06-16
fix: handle cf issued challenge on AJAX requests

kevin9foong:
## Problem

FormSG is proxied by CF's WAF and Super bot fight mode which sends challenges when suspicious requests are detected.
As such, when AJAX requests are made (eg, axios API calls), CF can intercept and return a challenge html as response, which do not match the current expected response of the AJAX request and causes a non-friendly `!DOCTYPE` error to be shown to users.  

Closes FRM-2032

## Solution

Intercept the error and check if it is a cloudflare challenge issued error.  
Display challenge popup to issue a cf_clearance cookie if the challenge succeeds, then replay the request. 

A challenge popup is chosen specifically so no reload is triggered and the react state (eg, for filled form fields) is retained, reducing the need for the respondent to re-fill a submission in the event of a CF issued challenge.  

Key points:
1. This popup only shows up for requests that would have been previously blocked due to CF issued challenge (ie, it should not make things worse).
2. In the event user's network blocks CF script, it will block the request. This edge case is explicitly handled.
3. The CF script is only loaded when CF issues a challenge in responses to AJAX request. It does not load otherwise.

Some benefits over the previous toast message only update:
- With the previous solution, we are relying on the idea that legit requesters can pass by retrying.
by right, requesters should not be assumed to be able to pass by retrying without completing the challenge, since this makes the challenge 'useless' given that bots can similarly retry.
in some cases, challenged users should be blocked until they pass the verification check

What's next:
- Review CF WAF/bot fight sensitivity rules and tune for optimal false positive %.

Edge cases handled: 
- In the event of turnstile script blocked by the network, the onLoadError function is invoked which will display a 'challenge load failed' error message and block the request (same as existing behavior). Note that the script is appended to the HTML body only upon error interception and will be removed in the event of script load failure - this handles users who block cloudflare scripts. 

Edge cases not handled: 
image
challenges.cloudflare.com is blocked/network fails after the Turnstile script loads. 
The `window.turnstile.render` is defined but the widget is empty

Note that this is not handled as it is an extremely unlikely occurrence similar to a network failure. 
To encounter this, a few conditions must be met: 
1. user meets challenge but does not pass this check. 
2. cloudflare script is now loaded in browser. user is re-issued challenge and blocks the cloudflare endpoint (which should not occur as it did not previously block the script unless it was intentionally done)/fails to load due to network error.  

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

doctype

**AFTER**:

image

Modal on mobile with smallest screen 

image

Modal on desktop / tablet 

## Tests

TC1: Challenge works for form submissions. 
- [x] Ensure WAF rule issues challenge to stg-alt3. 
- [x] Go to the forms submissions page (you may have to solve the turnstile) 
- [x] Clear the cookies
- [x] Try to submit, there should be a turnstile popup. Once complete, the submission should go through. 
- [x] Check that the submission details are accurate.  

TC2: Challenge works for admin login and edit fields 
- [x] Continue from TC1. 
- [x] As an admin, clear cookies and log into FormSG by triggering login email OTP. 
- [x] Assert that the pop up appears and upon solving, the request goes through and your OTP is received in your email inbox. 
- [x] Try to edit an existing form field. 
- [x] Before clicking on save form field, clear cookies once again. 
- [x] Assert that the pop up challenge appears and upon solving, the request goes through and the field is saved. 

TC3: Request is blocked when CF issues challenge but Turnstile fails to load (Eg, cloudflare script blocked) 
- [x] Continue from TC2 
- [x] Go to submit form page again (same as TC1) 
- [x] Block Cloudflare domain in browser DevTools network settings.
- [x] Clear cookies 
- [x] Attempt to submit the form. 
- [x] Assert that error is displayed that security verification failed to load and submission hence does not go through. 
- [x] Assert that the failed submission does not exist in the the admin form results/responses page.  

TC4: For GSIB 
- [ ] Retry TC1 on GSIB/comet machine

**New dependencies**:

- `dependency` : https://github.com/marsidev/react-turnstile for reliably rendering the turnstile widget in react. this is a recommended library and approved by Cloudflare. 
image


eliotlim:
Hey @kevin9foong!! this looks like a really helpful fix!

Just one note: later when we assess TC3, I'll help to run a quick check on the GSIB as well.
- staging.form.gov.sg (and form.gov.sg) are proxied directly through iasproxy.sgnet.gov.sg
- whereas cloudflare scripts might pass through Menlo SIS, and be blocked
- in the off chance that this is the case, it could cause any GSIB user facing a CloudFlare Turnstile challenge to lose form data

No biggie, I'll give it a try on stg-alt3 when it's done deploying.

kevin9foong:
> Hey @kevin9foong!! this looks like a really helpful fix!
> 
> Just one note: later when we assess TC3, I'll help to run a quick check on the GSIB as well.
> 
> * staging.form.gov.sg (and form.gov.sg) are proxied directly through iasproxy.sgnet.gov.sg
> * whereas cloudflare scripts might pass through Menlo SIS, and be blocked
> * in the off chance that this is the case, it could cause any GSIB user facing a CloudFlare Turnstile challenge to lose form data
> 
> No biggie, I'll give it a try on stg-alt3 when it's done deploying.

Sounds good - ill run it on my Comet device as well to verify this! 
In the event of CF script being blocked, im expecting the modal to close automatically and the request to be blocked with an error along the lines of 'The security verification failed to load. Please try again' with the form fields still filled for them to re-click on the submit button.  

Only thing is that since Comet has its own VPN - i will add a WAF rule for the comet VPN ip + stg-alt3 to test this (should not affect other users using non stg-alt3 domains) to trigger a managed challenge for testing

frontend/src/features/turnstile/useTurnstile.tsx
```diff
@@ -19,26 +20,6 @@ interface UseTurnstileProps extends TurnstileBaseConfig {
   enableUsage: boolean
 }
 
-type TurnstileConfig = TurnstileBaseConfig & {
```
kevin9foong, line 22: Removed since react-turnstile deps includes TS typings, which conflict with the existing self-defined type

kevin9foong:
Added stories for the new modal for small screen + desktop

frontend/src/features/turnstile/handleCloudflareChallenge.tsx
```diff
@@ -10,7 +18,79 @@ export const checkIsCloudflareChallengeError = (response: AxiosResponse) => {
...
+  const originalRequestConfigToReplay = error ? { ...error.config } : null
+
+  let overlay = document.getElementById(OVERLAY_ID)
+  if (!overlay) {
+    overlay = document.createElement('div')
+    overlay.id = OVERLAY_ID
+    document.body.insertBefore(overlay, document.body.firstChild)
+    const overlayReactRoot = createRoot(overlay)
+    return await new Promise((resolve, reject) => {
+      overlayReactRoot.render(
```
eliotlim: Hey @kevin9foong 

Wanted to say I really like the way you considered the user experience to make resubmission possible and less frustrating in [#8476](https://github.com/opengovsg/FormSG/pull/8476)

I was chatting with Ken earlier and was wondering if we might consider using a [React Context / Provider](https://react.dev/reference/react/createContext#provider) to inject the portal, so that we don't have to do direct DOM manipulation. It would probably have security and readability benefits, and reduce complexity of the code.

Wasn't sure if it was possible at first, so I gave it a poke in this [temporary commit over here](https://github.com/opengovsg/FormSG/commit/09a996219083eef112f1117cfbf333a23312f8be#diff-00493e7d46c535f8351491dad02492847e6507297378cb46baf46937c6aca436R103-R105).
- Feel free to use this snippet, or rearrange it / try something else -- it's just a proof-of-concept.
- I saw that you implemented automatic replay of the failing API request -- that's pretty cool
- (this snippet doesn't automatically cache / replay blocked requests yet)

kevin9foong: thanks for bringing this up! i'm in agreement and have shifted to the use of the provider based solution you've suggested, with some tweaks to handle CF script load failures to reload the script as discussed (also had a convo with Ken on this as well).

This new provider based approach also unlocks using chakra ui for the turnstile overlay - I've discussed with sufyan and we will be using the modal defaults and removed the hardcoded style values not possible previously due to the missing ChakaraProvider when manipulating the DOM directly.

eliotlim (approved):
Hey @kevin9foong!!

Just reviewed [the most recent changes](https://github.com/opengovsg/FormSG/pull/8476/files/0423004176f1d47cfa04dab747fcc970e4a2c506..e61287199fb0ce833bea35abddf4e824e68f6055), and verified that https://stg-alt3.form.gov.sg is working correctly :D

Screenshot 2025-06-23 at 2 54 38 PM

Again, great work with the commendable effort on improving the UX with the submission request replay 👍

eliotlim:
Oof okay just finished validating **TC4** on GSIB:

- form.gov.sg is whitelisted using the iasproxy (NUCLEUS) route
- stg-alt3.form.gov.sg seems to be whitelisted for direct connection based on the fact that Lets Encrypt certs are presented, instead of Menlo Security
- since CloudFlare is not whitelisted, the browser challenge always fails

image

However, I still support releasing this:
- it's overall still a win for user experience
  - users facing existing challenges get a better UX.
  - users with broken browsing experience remain broken
- we can whitelist GSIB access directly
- the workaround using `https://safe.menlosecurity.com/<url>` still works.

PR #8480, opened 2025-06-17
fix: onboarding and bootstrapping issues

eliotlim:
## Problem

Just a couple of onboarding rough edges:

- The onboarding and bootstrapping flow is almost ideal -- it basically bootstraps itself with just a few commands.
- When preparing a release, `scripts/release_prep.sh` isn't executable, and must be run with `bash scripts/release_prep.sh`
- Growthbook doesn't behave correctly because the client key for development isn't loaded by default

## Solution

**Breaking Changes** 

**Improvements**:

- Make **scripts/release_prep.sh** executable
- Add the (non-sensitive) Growthbook client key into `docker-compose.yml`.

## Tests

- [ ] Follow the onboarding guide: create a clean copy of the FormSG repository
- [ ] Run `npm install && npm --prefix serverless/virus-scanner install && npm run build && npm run dev`
- [ ] The entire FormSG local environment should just work (TM)


eliotlim:
whoops! the build failed.
as discussed, will remove the `workspaces` field in **package.json** to prevent dependency hoisting.

thanks @KenLSM!!

package-lock.json
KenLSM, line 1: Workspace installation causes all deps to be hoisted up, which is not what we want as the sub-repos are installed separately. This is an artifact of us structuring it as a monorepo with just `npm`.

KenLSM (approved):
LGTM! Thanks!

package-lock.json
eliotlim: 👍 agree! fixed it.

PR #8485, opened 2025-06-18
feat: add friendly message for cf intercepts

kevin9foong:
## Problem

FormSG is proxied by CF's WAF and Super bot fight mode which sends challenges when suspicious requests are detected.
As such, when AJAX requests are made (eg, axios API calls), CF can intercept and return a challenge html as response, which do not match the current expected response of the AJAX request and causes a non-friendly `!DOCTYPE` error to be shown to users.  

doctype

Closes FRM-2032

## Solution

Detect and handle the intercepted challenge responses by showing a business friendly message based on UX writing practices.    

This PR is meant as part of mitigation for incident 546. 

**Breaking Changes** 

## Tests 
TC1: Business friendly error replaces the previous !DOCTYPE error 
- [ ] Set up WAF rule to send challenges to OGP IP and connect to OGP VPN 
- [ ] On staging, go to the submit form page and clear your cookies for staging.form.gov.sg 
- [ ] Assert that the business friendly message pops up instead of the !DOCTYPE error

### DIFF
--- a/frontend/src/features/turnstile/handleCloudflareChallenge.tsx
+++ b/frontend/src/features/turnstile/handleCloudflareChallenge.tsx
@@ -1,0 +1,17 @@
+import { AxiosResponse } from 'axios'
+
+export const checkIsCloudflareChallengeError = (response: AxiosResponse) => {
+  return (
+    response &&
+    response.status === 403 &&
+    response.headers['server'] === 'cloudflare' &&
+    response.headers['cf-mitigated'] &&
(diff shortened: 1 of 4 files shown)


scottheng96 (approved):

scottheng96:
Regarding the failing test - I think what worked for me was running `npm install` and `npm run build` once again to get the module

kevin9foong:
Seem to be getting strange errors when running `npm i`

image

PR #8489, opened 2025-06-19
feat: update node / linux base

KenLSM:
## Problem

- Node 18 is reaching EOL
- Alpine 19 is reaching EOL

## Solution

- Upgrade node to 22
- Upgrade Alpine to 3.21 update matching chromium to `136.0.7103.113`
- Update `actions/setup-node` to upgrade their internal node version to >16
  - Use `node 22` for build tasks
- Update Buffer type usage to `Uint8Array` for returns by `Page.pdf` as it now returns `Uint8Array`
  - Added `Buffer.copyBytesFrom` to convert to `Buffer` since `Uint8Array` is not supported by `nodemailer`
- Upgrade `body-parser` to 2.x.x to ease future migration to `express@5`
  -  Added compatibility to restore `body` behaviour
- Update puppeteer to `24.9.0`
  - Added `--disable-gpu` to resolve issue with [failure to open new page](https://github.com/puppeteer/puppeteer/issues/11640)

**Breaking Changes** 

#### PDF Regressions
- [x] Create a form with email fields and PDF responses enabled
- [x] As a respondent make a submission
- [x] Observe that an email is received notifying of a success submission
- [x] Observe that an email is received with submitted response in PDF


src/app/loaders/express/parser.ts
```diff
@@ -14,6 +14,16 @@ const parserMiddlewares = () => {
   // processes that must be done before any other middlewares
   const preMiddlewareProcesses = [convertSnsMessageType]
 
+  const bodyParserCompatibility: RequestHandler = (req, res, next) => {
+    req.body = req.body || {}
```
eliotlim: nit: using the null coalesce operator (`??`) might be safer:
- falsy values (e.g. `""`, `0`, `false`, `[]`) should probably be preserved
- the intention seems to be to replace `null` / `undefined` with `{}`

```suggestion
    req.body = req.body ?? {}
```

eliotlim (commented):
besides the nullish coalesce operator, everything else looks good to me 👍

src/app/loaders/express/parser.ts
KenLSM: Good catch! Although to note that the lib itself [checked](https://github.com/expressjs/body-parser/commit/6cbc279dc875ba1801e9ee5849f3f64e5b42f6e1) for falsy values

eliotlim (approved):
Tested this flow and all tested flows behave correctly.

image

PR #8492, opened 2025-06-20
fix: mongoose issues

scottheng96:
## Problem

Some bugs were found during release of mongoose upgrade
1. Cannot add more than 2 collaborators
image

2. When duplicating field, duplicated field has the same _id 
image
## Solution

**Cannot add more than 2 collaborators**

- `permissionList` is returning a Dto with `_id` and `id`
- by default, **each subdocument gets its own `_id` field**. Mongoose also adds a **virtual `id` getter**, which is just `this._id.toString()`.
- setting `id: false` will disable the virtual setting of id
- To do that, creating a new schema for `permissionList`  is needed

**When duplicating field, duplicated field has the same `_id`** 

- `omit` creates a **plain JavaScript object**, not a proper Mongoose subdocument. In Mongoose v6, this works due to implicit casting. In v7, casting is stricter POJOs don’t automatically become subdocuments with new `_ids`
- Ids will then be duplicated, causing 2 fields to have the same `_id`
- Creating a new field while omit the previous `_id` will create a new subdocument with a new `_id`
- Then this needs to be casted back to a FormFieldSchema for it to be spliced into `this.form_fields`

**Breaking Changes** 

## Tests

**Adding 2+ Collaborators to a form**
- [x] Create a form/use an existing form
- [x] Try to add 2+ collaborators to the form. Observe the ability to do so.
- [x] Try changing collaborator access to viewer, and owner. Observe the ability to do so.

**Duplicating a form field**
- [x] Create a form and add any 1 field to the form
- [x] Attempt duplicating the form
- [x] Observe the ability to do so
- [x] Observe that you can independently update the original and duplicated field.
- [x] Make the form public and successfully make a submission 
- [x] Observe in admin portal responses that answers for both original and duplicated field are reflected

### DIFF
--- a/src/app/models/__tests__/form.server.model.spec.ts
+++ b/src/app/models/__tests__/form.server.model.spec.ts
@@ -2548,7 +2548,7 @@ describe('Form Model', () => {
         // Previous user should now be in permissionList with editor
         // permissions.
         expect(actual.toObject().permissionList).toEqual([
-          { email: populatedAdmin.email, write: true, _id: expect.anything() },
+          { email: populatedAdmin.email, write: true },
         ])
       })
     })
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -1122,10 +1122,12 @@ const compileFormModel = (db: Mongoose): IFormModel => {
     function (fieldId: string, insertionIndex: number) {
       const fieldToDuplicate = getFormFieldById(this.form_fields, fieldId)
       if (!fieldToDuplicate) return Promise.resolve(null)
-      const duplicatedField = omit(fieldToDuplicate, [
-        '_id',
-        'globalId',
-      ]) as FormFieldSchema
+
+      const formFieldsDocumentArray = this
+        .form_fields as Types.DocumentArray<IFieldSchema>
+      const duplicatedField = formFieldsDocumentArray.create(
+        omit(fieldToDuplicate.toObject(), ['_id', 'globalId']),
+      ) as FormFieldSchema
 
       this.form_fields.splice(insertionIndex, 0, duplicatedField)
       return this.save()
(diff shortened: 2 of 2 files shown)


eliotlim:
Tested:
- duplicating a form field ✅ 
- adding more than two collaborators ✅
- duplicating a form (whoops, not the actual test case) ✅ 


image 
image
image

eliotlim (approved):
makes sense to me! happy to have another pair of eyes on this, though! 👀 
- this fix does address the issues we observed the other day, so it's definitely required as part of the mongoose update, great stuff.

one lingering thought on my mind:
- might there be other areas in the codebase where directly writing a POJO retrieved from mongo might also contain `_id` / `globalId`?

KenLSM (approved):

KenLSM:
> makes sense to me! happy to have another pair of eyes on this, though! 👀
> 
> * this fix does address the issues we observed the other day, so it's definitely required as part of the mongoose update, great stuff.
> 
> one lingering thought on my mind:
> 
> * might there be other areas in the codebase where directly writing a POJO retrieved from mongo might also contain `_id` / `globalId`?

Had a brief chat with Scott a couple of days earlier. The error thrown was from `Joi` validation disallowing unspecified parameters (`id`). I did a rough check on validations with `_id` for patterns like this.
```
Joi.object({
     ...
})
```.

I think it also make sense to run through the errors during the window of incident if there are more issues like it.

scottheng96:
@eliotlim Good catch! I did a look over all the other areas where `_id` and `id` happen in the form definition, and I think this is the only one that was missed. There was another one that was initially handled already regarding `nricWhitelisting`. If you do find any other areas that is still in my blind spot I'll take a look at them!

scottheng96:
@KenLSM I also took a look at the 400x errors during the window of the release, seems to be that the errors that occurred were:
1. adding collaborator (~99% of them)
2. transferring ownership
   - I was unable to reproduce this after fixing the collaborator issue, I think it could be the scenario where there are more than 2 collaborators and someone tries to change ownership. I've remembered to test (without issues) this as well to be sure

PR #8500, opened 2025-06-24
fix: increase express keepalivetimeout

KenLSM:
## Problem

Getting 502s on ALB, this was also found on EC2's flow and unlikely to be solely related to IaC.

Closes FRM-2047

## Solution

Increase `keepAliveTimeout` to be more than the default alb timeout of 60s.

**Breaking Changes** 

#### Long running requests should not be terminated early
- [ ] Create a form with MFB
- [ ] Ensure that form fields are created
- [ ] As a respondent make a submission
- [ ] As an admin ensure that submission responses can be downloaded

### DIFF
--- a/src/app/loaders/express/index.ts
+++ b/src/app/loaders/express/index.ts
@@ -150,6 +150,12 @@ const loadExpressApp = async (connection: Connection) => {
 
   const server = http.createServer(app)
 
+  // To make sure that it is always more than ALB's default 60 seconds client timeout
+  // https://adamcrowder.net/posts/node-express-api-and-aws-alb-502/
+  server.keepAliveTimeout = 65 * 1000 // 65 seconds
+  // Limit the amount of time the parser will wait to receive the complete HTTP headers
+  // https://github.com/nodejs/node/issues/27363#issuecomment-2807796956
+  server.headersTimeout = 66 * 1000 // 66 seconds
   return server
 }
 


src/app/loaders/express/index.ts
```diff
@@ -150,6 +150,12 @@ const loadExpressApp = async (connection: Connection) => {
 
   const server = http.createServer(app)
 
+  // To make sure that it is always more than ALB's default 60 seconds client timeout
+  // https://adamcrowder.net/posts/node-express-api-and-aws-alb-502/
+  server.keepAliveTimeout = 65 * 1000 // 65 seconds
+  // Limit the amount of time the parser will wait to receive the complete HTTP headers
+  // https://github.com/nodejs/node/issues/27363#issuecomment-2807796956
+  server.headersTimeout = 66 * 1000 // 66 seconds
```
eliotlim, line 158: I agree -- the `headersTimeout` default value of `60` seconds [needs to be overridden](https://github.com/nodejs/node/issues/27363#issuecomment-2807796956).

eliotlim (approved):
LGTM!

We've fixed this bug before in other products (e.g. [Armoury](https://github.com/opengovsg/armoury/pull/1008)) and saw a [significant reduction in spurious 502 Gateway Errors](https://opengovproducts.slack.com/archives/C04J2QAGBEU/p1744813086298939?thread_ts=1744721345.990649&cid=C04J2QAGBEU).

On the whole, this should improve our signal to noise ratio for alerts and monitoring :)

PR #8557, opened 2025-07-21
feat: remove base64 of iac file string content

kevin9foong:
## Problem

Expectation of base64 encoding of files is unnecessary and causes potential confusion as to why it is needed.  

## Solution

Remove base64 requirement of SSM strings. 
Since previously, the file contents are extracted and read as string, it is safe to place the file string content directly into the SSM param values without first base64 encoding them. 

**Breaking Changes** 

## Tests

TC1: Singpass myinfo forms works 
- [ ] Create a singpass form with myinfo 
- [ ] Add 1-2 myinfo fields 
- [ ] Submit as a respondent and assert it works and submission is recorded etc

TC2: Singpass SGID form works 
- [ ] Create a singpass form with SGID 
- [ ] Enable nric collection opt in 
- [ ] Submit as a respondent and assert it works and submission nric is recorded etc

TC3: reCAPTCHA whitelist works for GSIB users
- [ ] Assert that GSIB users are still whitelisted after release and they are not subject to reCAPTCHA issues. 

## Deploy Notes

To apply these changes together with IaC envs (VAPT, stg-alt, uat, prod) in https://github.com/opengovsg/formsg-infra/pull/65

### DIFF
--- a/src/app/modules/auth/sgid/auth-sgid.service.ts
+++ b/src/app/modules/auth/sgid/auth-sgid.service.ts
@@ -32,7 +32,7 @@ export class AuthSgidServiceClass {
   }: ISgidVarsSchema) {
     this.privateKey = retrieveFileContent({
       preIacFilePath: privateKeyPath,
-      postIacBase64EncodedString: privateKey,
+      postIacFileContentString: privateKey,
     })
     this.client = new SgidClient({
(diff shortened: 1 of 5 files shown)


KenLSM (approved):
LGTM

PR #8565, opened 2025-07-23
chore: use npm workspaces for repo-wide npm install and better management

littlemight:
## Problem
This original install instruction in the readme...

```
To install the relevant npm packages (frontend, backend and virus-scanner), run the following in the root direcory:
npm install && npm --prefix serverless/virus-scanner install
```

...implies the command will run npm install to each folder having a `package.json`. While it is very obvious and trivial what needs to be done (e.g. just run `npm --prefix <your intended folder> install` lol), it can be easily missed.

Example: the "Running locally" section states
```
First, build the frontend for local development:

npm run build:frontend

Run the following shell commands to build the Docker image. The first time will usually take 10 or so minutes. These commands runs the backend services specified under docker-compose.yml and the React frontend on the native host.

npm run dev
```
If
1. someone forgot to run `npm --prefix frontend install`
2. `npm run build:frontend` will fail, since it needs `vite` as the build tool, but the error message is non-obvious as it's a bunch of word vomit
3. then continue to proceed to `npm run dev`, only to find out backend fails to run since one of the [controller](https://github.com/opengovsg/FormSG/blob/develop/src/app/modules/frontend/frontend.controller.ts#L16) needs to resolve the built frontend.

Found this out when piloting the FormSG self-hosting guide with someone.

## Solution
A simple solution is just write this in the readme.
```
npm install && \
  npm --prefix frontend install && \
  npm --prefix shared install && \
  npm --prefix serverless/virus-scanner install && \
  ...
```

But alternatively, I'm trying to see if we can leverage [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) for this, considering the monorepo structure of the repo.

**Breaking Changes** 

    - Details ...

**Improvements**:

- Simplified `npm i` to install `node_modules` on each listed folder.
- The usual behavior of `npm i` at root can still be achieved with `npm install --ignore-scripts --no-workspaces` if needed.

## Tests

- [ ] `npm i` at project root installs `node_modules` in all folders listed in `workspaces` section of root's `package.json`

## Deploy Notes

- Make sure backward-compatible with existing `Dockerfile.development` and `Dockerfile.production`

**New scripts**:

- `script` : Removes `postinstall` scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ nvm use
 To install the relevant npm packages (frontend, backend and virus-scanner), run the following in the root direcory:
 
 ```bash
-npm install && npm --prefix serverless/virus-scanner install
+npm install
 ```
 
 If you are on Mac OS X, you may want to allow Docker to use more RAM (minimum of 4GB) by clicking on the Docker icon on the toolbar, clicking on the "Preferences" menu item, then clicking on the "Resources" link on the left.
@@ -239,7 +239,7 @@ Please contact FormSG (support@form.gov.sg) for any details.
 ### Migrating from MongoDB to FerretDB
 [FerretDB](https://ferretdb.io) is an open source MongoDB alternative built on PostgreSQL. MongoDB can be swapped out of FormSG for FerretDB. In order for this to be done, certain changes to the code should be made as described below:
 
-- Add postgres to the list of services in the `docker.compose` file e.g. 
+- Add postgres to the list of services in the `docker.compose` file e.g.
   ```  pg:
     image: postgres:15.3-alpine3.18
     environment:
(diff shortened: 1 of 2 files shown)


littlemight:
Oh oopsie, build still failing. Changing this to draft in the meantime.

KenLSM:
Running `npm i` should be sufficient since there's a `postinstall` that also installs the other packages?

littlemight:
@KenLSM LOLz actually you're right. Hmm

littlemight:
@KenLSM Eh then I guess I was misled, maybe the error of the guy trying out this guide came from elsewhere (e.g. skipped some steps running on local, etc.).

I'll close this then

PR #8568, opened 2025-07-24
fix: radio fields logic bug

scottheng96:
## Problem

A bug was found related to radio field (others) and MRF which results respondents being unable to submit the form when a logic condition (e.g. showing/hiding a field) is dependent on the 'Others' value of the radio field.

To replicate:
- create an MRF
- add a radio field and enable "others"
- create a logic setting based on the "others" selection
- try to submit the form with "others" selected
- submission fails at the BE validation step

## Solution

Conditional field although shown in the form, is deemed as a hidden field during BE validation (expected behaviour is that conditional field is a visible field). With submission containing 2 responses & 1 visible field, BE validation fails (needs to be 1-1).

Conditional field is not deemed as visible because `isConditionFullfiled` check is failing. 

- Specifically for MRF, a radio other value is `{ othersInput: 'zxcv' }`.
- Inside the function, setting `.value` results in undefined:
  ```
    if (typeof currentValue !== 'string') {
      // Handle the case where the value is a client radio field value.
      currentValue = {
        ...currentValue,
        value: currentValue.value?.trim(),
      }
    }
  // currentValue = { othersInput: 'zxcv', value: undefined }
  ```
- It then fails the final check, specifically at `currentValue.value === CLIENT_RADIO_OTHERS_INPUT_VALUE`

**Breaking Changes** 

## Tests

**Successful submission on logic-related radio field**
- [ ] create an MRF form
- [ ] add a radio field and enable "others", and any other field
- [ ] create a logic setting based on the "others" selection (`equals to` or `either` is fine)
- [ ] successfully submit the form with "others" selected

**Regression testing for MRF**
- [ ] reuse an MRF form
- [ ] add a radio field and enable "others", and any other field
- [ ] create a logic setting based on specific radio option selection
- [ ] successfully submit the form

**Regression testing for Storage**
- [ ] create an Storage-mode form
- [ ] add a radio field and enable "others", and any other field
- [ ] create a logic setting based on the "others" selection (`equals to` or `either` is fine)
- [ ] successfully submit the form with "others" selected

### DIFF
--- a/shared/utils/logic.ts
+++ b/shared/utils/logic.ts
@@ -1,4 +1,3 @@
-import { CLIENT_RADIO_OTHERS_INPUT_VALUE } from '../constants'
 import {
   BasicField,
   FormDto,
@@ -409,10 +408,7 @@ const isConditionFulfilled = (
           }
         } else {
           // (4) Client-side handling
-          if (
-            currentValue.value === CLIENT_RADIO_OTHERS_INPUT_VALUE &&
-            !!currentValue.othersInput
-          ) {
+          if (currentValue.othersInput) {
             return true
           }
         }
--- a/shared/utils/logic.ts
+++ b/shared/utils/logic.ts
@@ -390,7 +390,7 @@ const isConditionFulfilled = (
        *     starting with "Others: ".
        *     If not, continue the usual string-based check with (3).
        * (4) If conditionValues includes "Others", check for the value being
-       *     special radio value and the othersInput subfield having a value.
+       *     undefined and the othersInput subfield having a value (XOR).
        *     If not, continue the usual string-based check with (3).
        * (2) + (5) Use Number equality for decimals
        * (3) + (6) Use String equality for everything else
@@ -408,7 +408,7 @@ const isConditionFulfilled = (
           }
         } else {
           // (4) Client-side handling
-          if (currentValue.othersInput) {
+          if (currentValue.value === undefined && currentValue.othersInput) {
             return true
           }
         }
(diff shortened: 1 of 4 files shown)


shared/utils/logic.ts
```diff
@@ -409,10 +408,7 @@ const isConditionFulfilled = (
           }
         } else {
           // (4) Client-side handling
-          if (
-            currentValue.value === CLIENT_RADIO_OTHERS_INPUT_VALUE &&
```
scottheng96, line 413: Just need to be sure that this is safe to remove - I assume it was written here for a reason in the first place. If anyone has more context that would be great :)

KenLSM:
Thanks for the quick turnaround! Actually, would the problem be `currentValue.value === CLIENT_RADIO_OTHERS_INPUT_VALUE` fails to be true when it should?

I'm thinking that the expected behaviour is actually to be `currentValue = { othersInput: 'zxcv', value: CLIENT_RADIO_OTHERS_INPUT_VALUE }`.

scottheng96:
So the current.value is populated above (in the same function):
```
// current value is `{ othersInput: 'zxcv' }`

  if (typeof currentValue !== 'string') {
    // Handle the case where the value is a client radio field value.
    currentValue = {
      ...currentValue,
      value: currentValue.value?.trim(),
    }
  }

// new current value is `{ othersInput: 'zxcv' , value: undefined}`
```

In the FE, `.value = CLIENT_RADIO_OTHERS_INPUT_VALUE` until we strip it away when creating `ResponsesV3`, right before we send it to our BE. Specifically here:

image

I think our options can be to retain it while creating the responsev3, or remove it when checking the validation.

KenLSM (approved):
LGTM, clarified that `currentValue.value` should be a XOR with `currentValue.othersInput`

scottheng96:
https://github.com/opengovsg/FormSG/pull/8574

additional bug fix linked here for posterity

PR #8569, opened 2025-07-24
feat: reduce virus scanner lambda size to 128MB

kevin9foong:
## Problem

Currently, the guard duty virus scanner consumes 2048MB of memory. However, the cost of lambda virus scanner invocations can be reduced 16x per month by reducing the memory size to 128MB with some optimizations. 

## Solution

Prevent loading of file content into lambda memory - by only loading the metadata to save memory usage. 
Change the virus scanner lambda to 128MB. 

**Breaking Changes** 

## Tests

TC1: Virus scanner works to scan files 
- [x] Create a form with 20MB file attachment fields totaling up to 20mb
- [x] Upload files to the attachment field - aiming to upload as large file size as possible
- [x] Assert virus scanner works and form can be submitted

### DIFF
--- a/serverless/virus-scanner-guardduty/serverless.yml
+++ b/serverless/virus-scanner-guardduty/serverless.yml
@@ -72,7 +72,7 @@ functions:
       name: virus-scanner-guardduty
     # max timeout per invocation
     timeout: 300
-    memorySize: 128
+    memorySize: 128 # Usage of the smallest memory size is possible since the lambda function does not load the file content into memory and simply makes API calls to S3.
     # Provision concurrency so at least one instance always hot
     # https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/#provisioned-concurrency
     # 1 for staging, 5 for prod
--- a/serverless/virus-scanner-guardduty/serverless.yml
+++ b/serverless/virus-scanner-guardduty/serverless.yml
@@ -72,8 +72,7 @@ functions:
       name: virus-scanner-guardduty
     # max timeout per invocation
     timeout: 300
-    # allocate more memory as the default 1028 MB size will cause lambda to crash
-    memorySize: 2048
+    memorySize: 128
     # Provision concurrency so at least one instance always hot
     # https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/#provisioned-concurrency
     # 1 for staging, 5 for prod
--- a/serverless/virus-scanner-guardduty/src/__tests/index.spec.ts
+++ b/serverless/virus-scanner-guardduty/src/__tests/index.spec.ts
@@ -71,15 +71,15 @@ describe('handler', () => {
       }),
     )
   })
-  it('should return 404 and log if file not found', async () => {
+
+  it('should return 404 if versionId not found', async () => {
     // Arrange
     const mockUUID = crypto.randomUUID()
     const mockEvent = { key: mockUUID }
 
-    // Mock rejection of getS3FileStreamWithVersionId
-    MockS3Service.prototype.getS3FileStreamWithVersionId = jest
+    MockS3Service.prototype.getS3ObjectVersionId = jest
       .fn()
-      .mockRejectedValueOnce(new Error('File not found'))
+      .mockRejectedValueOnce(new Error('VersionId is empty'))
 
     // Act
     const result = await handler(mockEvent as unknown as APIGatewayProxyEvent)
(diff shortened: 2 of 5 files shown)


eliotlim:
Ran into some issues with the test case 😓

I created a ZIP file with some JPG images from Unsplash:
[test-formsg-av-4.zip](https://github.com/user-attachments/files/21402597/test-formsg-av-4.zip)

But got a 504 Gateway Timeout Error

image

image

eliotlim:
I think the files are being scanned correctly by both [guardduty](https://205930608298-qycu7qst.ap-southeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-southeast-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fvirus-scanner-guardduty-stg-alt3-virus-scanner-guardduty/log-events/2025$252F07$252F24$252F$255B56$255Df43afb2513724232b0cdfed881ea742a) and the [scanner](https://205930608298-qycu7qst.ap-southeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-southeast-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fvirus-scanner-stg-alt3-virus-scanner/log-events/2025$252F07$252F24$252F$255B146$255Deeca7984fd3f43d293c8e94dd909ac4a) lambda itself, though.

eliotlim:
Tested with the updated Test Cases and everything works :)

image
image

eliotlim (dismissed):

eliotlim:
Performance compared to production

## Stg-Alt3 4 x Images: 24 seconds

image

## Stg-Alt3 17 MB ZIP: 40 seconds

image

## Production 4 x Images: 16 seconds

image

## Production 17 MB ZIP: 25 seconds
image

eliotlim:
Could we do a quick experiment to find a memory value that minimises the waiting time for fully-loaded forms?

I'm wondering if it might be worth tuning RAM to 256 / 512 MB to [increase the vCPU allocation to the lambda](https://docs.aws.amazon.com/lambda/latest/dg/configuration-memory.html)? (128MB / [1,769MB](https://docs.aws.amazon.com/lambda/latest/dg/configuration-memory.html#:~:text=At%201%2C769%20MB%2C%20a%20function%20has%20the%20equivalent%20of%20one%20vCPU%20(one%20vCPU%2Dsecond%20of%20credits%20per%20second)) = 7% vCPU)

We're seeing an increase in execution duration that could be related to vCPUs (which scales linearly to Lambda RAM)
- min: 165 ms -> 1,175 ms
- avg: 1,716 ms -> 2,800 ms

## [Production Metrics](https://652261905145-egpubaxm.ap-southeast-1.console.aws.amazon.com/lambda/home?region=ap-southeast-1#/functions/virus-scanner-guardduty-production-virus-scanner-guardduty?tab=monitoring)
image

## [Stg-Alt3 Metrics](https://205930608298-qycu7qst.ap-southeast-1.console.aws.amazon.com/lambda/home?region=ap-southeast-1#/functions/virus-scanner-guardduty-stg-alt3-virus-scanner-guardduty?tab=monitoring)
image

kevin9foong:
Thanks for bringing this up - lowering the memory reduces compute could indeed lead to slower cold starts and reduced performance. 

## Approach to compare performance/latency impact due to reducing memory to 128mb: 
Some analysis points: 
- Apart from cold start timing, I'm thinking the decreased CPU compute might not affect the actual handler execution timings too significantly, since it is making (3) API calls to S3 and not doing any compute-intensive processing itself, which 128MB should be able to handle. 
- Comparing stg-alt3 and prod timings might not be an reliable comparison for the following reasons:
1. prod has higher provisioned concurrency (similar to 10 lambdas running at all times vs stg's 1) + lambdas are kept somewhat constantly hot by regular requests. This means that the times measured for prod might not include cold start times. 
2. Another primary reason why prod might be faster is that the time taken to submit includes the time it takes for the server to invoke the controller function and its related code, including invoking the lambda, downloading & encrypt the file and responses (a compute heavy task) on prod ECS, which has 4x the compute compared to stg-alt3 (ie, 4vCPU vs stg-alt which only has 1vCPU).

Hence, a closer comparison is to compare stg-alt3 and staging, which mitigates the issue in 1 since they have similar usage patterns & provisioned concurrency. 
However, to note that staging application server still has 2x the vCPU (2vCPU) than stg-alt3. 

After doing some analysis, it seems that staging is marginally faster 2-4s faster, not accounting for the 2x vCPU which leads to faster encryption etc. This time difference seems to align with the differences in lambda invocation times - which means that the impact is accurately reflected by the differences in lambda invocation time (not overall execution time). Hence, **we can safely use the lambda invocation time differences to determine the impact of the change on file upload latency.** 

## Analyzing impact using lambda invocation times
According to this article, cold start times increase linearly with memory increases. (https://www.pluralsight.com/resources/blog/cloud/does-coding-language-memory-or-package-size-affect-cold-starts-of-aws-lambda)

So first step will to check the impact of memory size on cold start timings for our use case on the same env (stg-alt3). 

Comparing 128mb and 10gb memory size lambdas with a mix of 90kb, 5mb and 11mb file uploads. 
(using difference between billed and execution time as cold start time) 
image
image

- cold start time: 530ms(10gb) vs 750ms (128mb)
- execution time: 10gb is about 100-600ms faster per invocation based on a couple runs
image
128mb (left) 
image
10gb (right) 

Looking at the differences in lambda invocation times above, my opinion is that while reducing the memory might cause some increased latency, the **difference in timing is not significant** (ranging from 100ms - 800ms increase between 128mb and 10gb, but avg difference im expecting is ~300ms difference), since most of the time taken is for performing eg, the encryption, saving to db etc before the server returns the response. (Though if 256mb or 512mb feels safer, we can do that too - still significant annual cost savings)

## Monitoring and quick mitigation strategy
We will likely continue to monitor the latency (on DD) to determine if the above hypothesis stands (that the impact is not significant). If the impact is deemed high, the mitigation/rollback can be done by simply changing the value back to 2048mb (which is a relatively quick change). 

## Other possible approaches to optimize latency  
- perhaps we might not need a lambda to perform this checking of tags and moving of file and should move the processing to the  app server instead (esp since the lambda function has low cpu cost and its awaiting the lambda response anyway)? The previous intention seems to be to have a place to host the clamAV binary + isolate formsg app from the quarantine bucket - but this may not be necessary today if we are just pulling the versionId and never loading the file into memory.  This approach would reduce latency (potentially by 800ms for 10gb-1.5s for 128mb/invocation) as it avoids the cold start issue entirely and remove a point of failure (lambdas). 
- if the cold start is indeed a problem, maybe lambda snapstart can be used to reduce latency? Though might have to re-write the lambda code in python as it supports python (a tradeoff). 
- we can use 256/512mb, but this would still lead to increased latency assuming the plural sight article is true since the performance is linear. from my testing, the differences in function invocation time  between 128/256/512 may not be too significant.

eliotlim (approved):
Thanks again for doing this detailed analysis, @kevin9foong !!

Yup, agreed that there are confounding factors between the compute sizing for post-scan operations and the provisioned concurrency / cold-starts for the different environments.

As discussed -- let's roll this out:
1. deploy this, keep an eye out for any significant or unexpected degradation to user experience
2. roll back if necessary, since that's an easy 30 second change from the AWS console
3. if necessary, fine-tune the RAM allocation to bring the execution duration back to the baseline

kevin9foong:
Yep! Let's do this, will monitor over the coming weeks - thanks @eliotlim!

PR #8585, opened 2025-08-03
fix: handle undefined children value

kevin9foong:
## Problem

Blank screen occurs for myinfo children field form after login. 
This is caused by the `child` value from react hook form returning undefined and causing issues with `Object.values(undefined)`, causing an infinite render loop. 

image

## Solution

Handle the undefined test case. 

**Breaking Changes** 

## Tests

TC1: myinfo child loads for myinfo acc with no children 
- [x] Make a form with myinfo child field 
- [x] Login with singpass acc with no child 
- [x] Assert that no child found. 
- [x] Try to submit, assert that not submittable as field is required. 

TC2: myinfo child submittable for myinfo acc with 1 child to submit
- [x] Make a form with myinfo child field with relevant child fields. 
- [x] Login with singpass acc with child eg, `F16...64K` for mockpass.  
- [x] Assert that children are found. 
- [x] Select child and assert that child fields are found (eg, name, date of birth etc) 
- [x] Try to submit, assert that submittable. 

TC2: myinfo child submittable for myinfo acc with >=1 child to submit
- [x] Make a form with myinfo child field with relevant child fields and enable multiple children to be selected. 
- [x] Login with singpass acc with child eg, `F16...64K` for mockpass.  
- [x] Assert that children are found. 
- [x] Select 2 (or more) children in the field and assert that child fields are found (eg, name, date of birth etc) 
- [x] Try to submit, assert that submittable. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when retrieving selected child names, preventing errors if no child fields are present.

### DIFF
--- a/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
+++ b/frontend/src/templates/Field/ChildrenCompound/ChildrenCompoundField.tsx
@@ -244,9 +244,12 @@ const ChildrenBody = ({
   // useCallback to re-compute names because for some reason watch doesn't
   // work on this nested field in react-hook-form.
   const allSelectedNames = useCallback((): string[] => {
+    const child = getValues(`${schema._id}.child`)
     // Really important to note that sometimes react-hook-form stores our "array"
     // as a object with key=index and values=array entry.
-    return Object.values(getValues(`${schema._id}.child`)).map((arr) => arr[0])
+    const childValues = child ? Object.values(child) : []
+    const childNames = childValues.map((arr) => arr[0])
+    return childNames
   }, [getValues, schema._id])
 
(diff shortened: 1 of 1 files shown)


eliotlim:
Tester's note:
- the `children` beta feature flag needs to be added on the Internal Admin Portal.
- the MyInfo child fields are only available using Singpass with MyInfo (not App-only) mode.

image
image

eliotlim (approved):
since the [data schema for the **ChildrenCompoundField**](https://github.com/opengovsg/FormSG/pull/8585/files#diff-5750f9c0b3c787886715d36517dd538025a5c26209ed46a8907c0a84ae27a6c4R61-R64) is "array of arrays", this is probably the best we can do.

PR #8600, opened 2025-08-12
chore: remove email from playwright projects

KenLSM:
## Problem

We have concluded our migration of public email mode forms and there will no longer be any email mode forms available. This is part of our efforts to deprecate email mode.

However, it is noted that we have some existing test suites that is still testing email mode form which is unnecessary.

1. E2E playwright tests 👈 this PR
2. FE Chromatic stories
3. BE tests

## Solution

Remove E2E playwright for email mode.

**Breaking Changes** 

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -71,10 +71,8 @@ jobs:
         project:
           [
             'chromium-login',
-            'chromium-email-submission',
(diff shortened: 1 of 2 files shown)


kevin9foong (approved):
LGTM! The CI already feels faster from this PR!

PR #8601, opened 2025-08-12
chore: remove eb uat target

KenLSM:
## Problem

We no longer need to deploy through elasticbeanstalk!

Closes FRM-2088

## Solution

**Breaking Changes** 


kevin9foong (approved):
LGTM! Step 1 of tearing down non-IAC envs

PR #8605, opened 2025-08-13
feat: move pdf gen to lambda 

kevin9foong:
### Scaling spiky pdf gen loads independently   
Pdf generation is a compute-heavy and spiky traffic task that consumes much of the main ecs compute. Hence, it should be moved to a scalable infra solution. 
This allows the main app server and pdf generation to scale independently. 

### Potentially doubling our app server containers for the same infra spend $  
Also, with our current usage of running chromium on the main ecs compute, we require 2vCPU per container. However, node.js does not benefit much from having 2 CPUs since it is single threaded. Moving chromium away allows us to use just 1 vCPU, allowing us to double our number of running containers and double our load/traffic capacity at the same infrastructure spend.   

## Solution

Move PDF generation to its own lambda. 

This solution: 

Creating Lambda function: 
- written in TS 
- contains font zenhei to be packaged in zip for generating chinese characters (adopted from previous implementation) 
- uses esbuild to compile and package into zip. The script to build is `build_dist.sh` 
- deployment scripts in package.json 

Deploying Lambda function: 
- uses zip based deployment via SAM to AWS Lambda 
- includes S3 bucket creation to store zip files (required as zip is 87mb, exceeding 50mb limit). This s3 bucket is referenced by the SAM deployment script. 
- several GH actions deployment scripts to automate deployment 

Local testing:
- uses local SAM to emulate the lambda function (tests the actual code running on lambda exactly - ideal scenario) 
- contains utility script `npm run dev` to run locally regardless of docker software (Eg, Podman, Colima) 

Documentation: 
- new readme to explain how to manually setup (on an as-needed basis) 
- update the main readme.md to include pdf gen lambda setup on local dev. 

## Deployment Rollout Strategy 

3 key considerations: 
1. Latency comparison → DD metrics have been set up and a dashboard widget under "Pdf Generation" has been setup on the FormSG dashboard for comparison.  

2. Output comparison → internal bug bash testing to be carried out. 
To focus on: 
- Autoreply PDF generation 
( with various local languages - english, tamil, chinese, malay. Aim is to follow existing behavior )
( with varying # of file sizes and field sizes ) 
- Payment invoice 

3. Incremental rollout → Growthbook is used with target attributes to conditionally rollout 
Supports the following attributes: 
- formid
- adminEmail eg, *@hello.com 
- adminAgency eg, govtech 

**Breaking Changes** 

## Tests

TC1: Can generate auto-reply PDF 
- [x] Create a new storage mode form 
- [x] Add an email field and add PDF responses under the email field settings left drawer. 
- [ ] Add a bunch of fields (eg, signature, short text, table fields) - aim for 5 - 10 fields. 
- [x] Submit a response to the form - include chinese characters in your response. 
- [x] Assert that an email with PDF is generated and sent to the email specified in the form response.
- [x] Open and assert that the PDF content is generated without visual issues. 

TC2: Can generate payment invoice PDF
- [ ] Create a payments form 
- [ ] Submit the payments form and download PDF invoice as a respondent. 
- [ ] Assert that the payment invoice PDF is generated without visual issues. 
- [ ] Go to the admin results page for the payments form. 
- [ ] Open the response result and download PDF invoice from admin page. 
- [ ] Assert that the PDF invoice is downloaded without visual issue.


src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -4081,6 +4086,7 @@ describe('admin-form.controller', () => {
         responseMode: FormResponseMode.Encrypt,
         publicKey: 'some public key',
         title: 'mock title',
+        emails: [],
```
scottheng96, line 4089: This adding of email variable is for respondentCopy?

src/app/modules/payments/stripe.service.ts
```diff
@@ -1,5 +1,6 @@
 // Use 'stripe-event-types' for better type discrimination.
 /// <reference types="stripe-event-types" />
+import { GrowthBook } from '@growthbook/growthbook'
```
scottheng96, line 3: I'm seeing a few instances of passes of feature flags from req ---> the function that uses it. I understand this is the only way that we are able for our BE to receive ff info right now; just want to be cognisant of this trade off for code bloat: `handleStripEvent` doesn't actually use growthbook and its simply passing it through to `performPaymentPostSubmissionActions`.

We can probably go with this for now, and revisit again when an exploration into a global growthbook singleton is possible!

scottheng96 (approved):

scottheng96:
Ah one more note, if we are to deploy to stg, we don't have a `deploy-pdf-gen-stg.yaml` file yet. Should be easy to add!

Also not sure how the behaviour will work on staging (in aws production acc). Will it use the same lambda as production?

kevin9foong:
Tested on all the envs - works on IAC and Clickops. Ready for a hopeful deployment this wed!

PR #8617, opened 2025-08-19
fix: omit signature field from mfb creation

scottheng96:
## Problem

Admins can create forms with signature field without being whitelisted through Magic Form Builder

## Solution

Omit Signature field from possible field types during prompt generation

**Breaking Changes** 


eliotlim (approved):
LGTM! 👍

PR #8624, opened 2025-08-20
feat: enable mrf webhooks

eliotlim:
## Problem

Users want to configure webhooks as destinations for their Multi-Respondent Form.
At present, only Storage (Encrypt) mode forms can have Webhooks configured as a notification destination.

Closes FRM-1679

## Solution

> [!IMPORTANT]
> This pull request will allow any FormSG Admin to configure webhooks for their Multi-Respondent Form.
> The Feature Flag only affects the front-end visibility of the Webhooks section in the Settings Page.

**Breaking Changes** 

**Features**:

- Enable Webhooks for Multi-Respondent Forms
- Add a `workflowContent` field which contains the `workflowStep` counter and the `workflow` configuration

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

image

## Tests

**TC1: Verify that Multi-Respondent Forms can now be configured with a Webhook**

- [ ] Create a new Form with "Multi-Respondent" mode selected
- [ ] Create a simple form (e.g. a Short Text field called "Name")
- [ ] Add a Yes / No field for approval use
- [ ] In the Workflows tab, configure a simple Workflow with two steps:
  - [ ] Step 1: Application -- all the simple form fields should be assigned to this step.
  - [ ] Step 2: Approval -- the approval field should be assigned to this step. Use your email as the approver email.
- [ ] In the Settings page, select the Webhooks tab, then configure a Webhook URL from [https://webhook.site](https://webhook.site)
- [ ] Click outside the Webhook URL field: the URL should be saved.

**TC2: Receive a Webhook from a Multi-Respondent Form**

- [ ] Visit and fill up the form in Test Case 1
- [ ] Submit the form
- [ ] Observe that the webhook was successfully received at the Webhook URL (TC1)
- [ ] Open the approval request email, Click on "Yes" / "No", and submit the approval / rejection
- [ ] Observe that the webhook was successfully received at the Webhook URL (TC1)

**TC3: Check that MRF Form Settings can be modified**

- [ ] Visit the form in Test Case 1
- [ ] Set the Form to CLOSED.
- [ ] Set the Form to OPEN.
- [ ] Add `form@open.gov.sg` as a collaborator to the form.
- [ ] All of the operations should succeed without error.

## Deploy Notes

**New feature flags**:

- `enable-mrf-webhooks` : [enable / disable MRF webhooks for specified user segments](https://app.growthbook.io/features/enable-mrf-webhooks)

### DIFF
--- a/shared/utils/strip-workflow-emails.ts
+++ b/shared/utils/strip-workflow-emails.ts
@@ -1,0 +1,18 @@
+import {
+  FormWorkflowDto,
+  StrippedFormWorkflowDto,
+  WorkflowType,
+} from '../types'
+
+export function stripWorkflowEmails(
+  workflow: FormWorkflowDto,
+): StrippedFormWorkflowDto {
+  return workflow.map((step) => {
+    if (step.workflow_type === WorkflowType.Static) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { emails, ...rest } = step
+      return rest
+    }
+    return step
+  })
+}
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -600,6 +600,7 @@ MultirespondentSubmissionSchema.methods.getWebhookView = async function (
     workflowContent: {
       workflow: this.workflow,
       workflowStep: this.workflowStep,
+      submittedSteps: this.submittedSteps,
     },
   }
 
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -597,6 +597,10 @@ MultirespondentSubmissionSchema.methods.getWebhookView = async function (
     created: this.created,
     attachmentDownloadUrls: attachmentRecords,
     paymentContent,
+    workflowContent: {
+      workflow: this.workflow,
+      workflowStep: this.workflowStep,
+    },
   }
 
   return {
--- a/src/app/models/submission.server.model.ts
+++ b/src/app/models/submission.server.model.ts
@@ -567,6 +567,43 @@ type MultiRespondentAggregates = Pick<
 type MultiRespondentAggregateResult = MetadataAggregateResult &
   MultiRespondentAggregates
 
+/**
+ * Returns an object which represents the encrypted submission
+ * which will be posted to the webhook URL.
+ */
+MultirespondentSubmissionSchema.methods.getWebhookView = async function (
+  this: IMultirespondentSubmissionSchema,
+): Promise<WebhookView> {
+  const formId = this.populated('form')
+    ? String((this as IMultirespondentSubmissionSchema).form._id)
+    : String(this.form)
+  const attachmentRecords = Object.fromEntries(
+    this.attachmentMetadata ?? new Map(),
+  )
+
+  if (this.paymentId) {
+    await (this as IMultirespondentSubmissionSchema).populate('paymentId')
+  }
+  const paymentContent = this.populated('paymentId')
+    ? getPaymentWebhookEventObject(this.paymentId)
+    : {}
+
+  const webhookData: WebhookData = {
+    formId,
+    submissionId: String(this._id),
+    encryptedContent: this.encryptedContent,
+    verifiedContent: '',
+    version: this.version,
+    created: this.created,
+    attachmentDownloadUrls: attachmentRecords,
+    paymentContent,
+  }
+
+  return {
+    data: webhookData,
+  }
+}
+
 MultirespondentSubmissionSchema.statics.findSingleMetadata = function (
   formId: string,
   submissionId: string,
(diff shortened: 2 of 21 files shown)


src/app/models/submission.server.model.ts
```diff
@@ -567,6 +567,47 @@ type MultiRespondentAggregates = Pick<
...
+  }
+  const paymentContent = this.populated('paymentId')
+    ? getPaymentWebhookEventObject(this.paymentId)
+    : {}
+
+  const webhookData: WebhookData = {
+    formId,
+    submissionId: String(this._id),
+    encryptedContent: this.encryptedContent,
+    verifiedContent: '',
```
eliotlim, line 595: `verifiedContent` is used for MyInfo-filled fields.

This is something we might support for multi-respondent forms in the future, so it should remain reserved.

src/app/models/submission.server.model.ts
```diff
@@ -567,6 +567,47 @@ type MultiRespondentAggregates = Pick<
...
+    version: this.version,
+    created: this.created,
+    attachmentDownloadUrls: attachmentRecords,
+    paymentContent,
+    workflowContent: {
+      workflow: this.workflow,
+      workflowStep: this.workflowStep,
+    },
```
eliotlim: this is where the `workflowContent` is added to the webhook payload

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.controller.spec.ts
```diff
@@ -154,6 +154,9 @@ describe('multirespondent-submision.controller', () => {
         form: mockSubmitMrfReq.formsg.formDef,
         encryptedPayload: mockSubmitMrfReq.formsg.encryptedPayload,
         submissionId: mockSubmissionId,
+        submission: {
+          _id: mockSubmissionId,
+        },
```
eliotlim, line 159: the `submission: { _id }` field seems to be required for tests to pass

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -126,6 +126,7 @@ const getSnapshottedFormDef = (
   form_fields: mrfSubmission.form_fields,
   form_logics: mrfSubmission.form_logics,
   workflow: mrfSubmission.workflow,
+  webhook: currentFormDef.webhook,
```
eliotlim, line 129: the webhook URL is always loaded from the current form definition, along with these other fields (at time of writing):
- title
- respondent copy setting
- emails
- email notification
- step(s) to send notifications for

eliotlim:
for transparency: there's a flash of content which we should probably fix:
- it displays briefly during the loading state (like in this case)
- and worse still, may incorrectly lead the user to believe that webhooks are not available for their selected mode
image

shared/utils/strip-workflow-emails.ts
```diff
@@ -0,0 +1,18 @@
...
+): StrippedFormWorkflowDto {
+  return workflow.map((step) => {
+    if (step.workflow_type === WorkflowType.Static) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { emails, ...rest } = step
+      return rest
+    }
+    return step
+  })
+}
```
scottheng96, line 18: Nice, thanks for extracting this out

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.controller.spec.ts
scottheng96: https://github.com/opengovsg/FormSG/pull/8530

Mongoose v7 things.. where .id doesn't virtually get setted, so we have to specify _id.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
```diff
@@ -694,6 +697,28 @@ export const performMultiRespondentPostSubmissionCreateActions = ({
     })
   }
 
+  const webhookUrl = form.webhook?.url
```
scottheng96, line 700: Just a thought here, if it possibly is a good idea to feature flag the sending of MRF webhooks as additional mitigation?

If we turn feature flag off, currently we will disable the ability to update the settings, but existing forms with webhooks enabled would still have webhooks being fired.

scottheng96 (approved):

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
eliotlim: yeah, this is a great point 💯 

we discussed this offline and agreed that:
1. the scope of the rollout is small (OGPeeps only)
2. if there are unexpected issues, we can clear these webhook fields relatively easily
3. and the intention is to make these generally available sooner, rather than later

so the risk calculus would favour us releasing quickly and getting this in front of real users.

PR #8633, opened 2025-08-22
fix: ongoing mrf not using snapshotted mrf values

kevin9foong:
## Problem

MRF forms that are in progress (step >=2 onwards) are using the latest form definitions instead of the intended snapshotted form definitions. This causes corruption to the ongoing MRF submission steps if the admin changes the form in the meantime. 

## Solution

Use the snapshotted values for mrf updates. 

In the get formsg method for MRF: 
- add snapshottedFormDef to be used by the submit update mrf functions  
- update the types for functions to be minimal and only contain necessary fields so that it is compatible with various form types. 
- add TC for middleware to ensure that the formsg fields eg, snapshottedFormDef and mrfSubmission is set correctly  

**Breaking Changes** 

## Tests

### DIFF
--- a/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -4,7 +4,7 @@ import { NextFunction } from 'express'
 import { StatusCodes } from 'http-status-codes'
 import { err, errAsync, ok, okAsync, Result, ResultAsync } from 'neverthrow'
 
-import { IAttachmentInfo } from 'src/types'
+import { IAttachmentInfo, IMultirespondentSubmissionSchema } from 'src/types'
 
 import { featureFlags } from '../../../../../shared/constants'
 import {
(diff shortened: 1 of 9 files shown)


hossenmizanur:
Please fix all the problems

hossenmizanur:
Markdown is supported

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -127,41 +164,68 @@ export const createFormsgAndRetrieveForm = async (
...
+          const { statusCode, errorMessage } = mapRouteError(error)
           return res.status(statusCode).json({ message: errorMessage })
         })
-        .map((formDef) =>
-          // Step 4a: Check form is multirespondent form
-          checkFormIsMultirespondent(formDef)
+        .map((mrfSubmission) => {
+          formsg.mrfSubmission = mrfSubmission
+          return mrfSubmission
+        })
```
eliotlim, line 184: we're moving the mrf submission retrieval up so that we can include the snapshotted form definition.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -398,27 +463,27 @@ export const validateMultirespondentSubmission = async (
...
-              form_fields: submission.form_fields,
-              form_logics: submission.form_logics,
-            }))
-          : okAsync({
+              workflowStep: mrfSubmission.workflowStep + 1,
+              workflow: mrfSubmission.workflow,
+              form_fields: mrfSubmission.form_fields,
+              form_logics: mrfSubmission.form_logics,
+            })
+          : ok({
```
eliotlim, line 481: we've hoisted the snapshot retrieval to an earlier step, which is why we now reference the submission itself instead of the submissionId

eliotlim (approved):
Thanks @kevin9foong for working on this fix!

Wanna call out and specifically appreciate:
- the rapid progress and iteration in this fix
- and the way you reviewed and optimised the code changes to minimise the delta from the original solution 💯 

LGTM! 👍

PR #8641, opened 2025-08-25
fix: optional signature field

KenLSM:
## Problem

Image generation fails when an optional Signature field that's left empty is submitted

## Solution

- Ensure signature field is not empty before attempting to generate image

**Breaking Changes** 

### Regression
**Submission should work on Empty Optional Signature Fields** 
- [x] Create a Signature Field with required toggled off
- [x] Make a submission without drawing on the signature field
- [x] Ensure that submission is successful
- [x] Ensure that the signature is not attached in the response email

**Submission should work on filled in Optional Signature Fields** 
- [x] Create a Signature Field with required toggled **on**
- [x] Draw on the Signature Field
- [x] Make a submission
- [x] Ensure that submission is successful
- [x] Ensure that the signature is as attached in the response email

**Submission should work on required Signature Fields** 
- [x] Create a Signature Field with required toggled **on**
- [x] Draw on the Signature Field
- [x] Make a submission
- [x] Ensure that submission is successful
- [x] Ensure that the signature is as attached in the response email

### DIFF
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
@@ -2,6 +2,7 @@ import JoiDate from '@joi/date'
 import { celebrate, Joi as BaseJoi, Segments } from 'celebrate'
 import { AuthedSessionData } from 'express-session'
 import { StatusCodes } from 'http-status-codes'
+import { isEmpty } from 'lodash'
 import mongoose from 'mongoose'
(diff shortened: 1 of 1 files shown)


kevin9foong (approved):
lgtm - assuming that the answerArray is indeed what ts says it is :)

KenLSM:
Found a separate issue, the field still registers as `Signature_Captured_68ac4cf0426ccccdf8a1adc3.png` on the response table even though there's no signature.

A separate issue from this hotfix.

Screenshot 2025-08-25 at 7 54 37 PM

PR #8644, opened 2025-08-26
fix: ignore secret key content

KenLSM:
## Problem

There are false positives from Chromatic on Secret Key modal.

## Solution

- Add `data-chromatic="ignore"` to ignore textual diff

**Breaking Changes** 


eliotlim (approved):
LGTM! 👍 this should really improve the signal-to-noise ratio of our Chromatic diffs.

thanks for fixing this, @KenLSM

PR #8647, opened 2025-08-27
chore(deps): remove unused frontend deps

KenLSM:
## Problem

Unused dependencies can cause dependabot to update them, requiring engs to spend time reviewing unnecessary PRs.
Also slows down installs during CI.

## Solution

Removing unused or unnecessary packages from both the main and development dependencies. Focusing on `frontend/`.

Dependency removals:

* Removed `web-vitals` as we're not using this.
* Removed `react-refresh`, as `vite/plugin-react` supports this natively.

Storybook-related cleanup:

* Removed `@storybook/node-logger` and `@storybook/preset-create-react-app` from the development dependencies, which are not needed for the current Storybook setup.
* Removed `storybook-react-i18next`, as we're not using i18next in storybook.

**Breaking Changes** 

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -96,7 +96,6 @@
         "use-draggable-scroll": "^0.1.0",
         "validator": "^13.7.0",
         "vite-plugin-node-stdlib-browser": "^0.2.1",
-        "web-vitals": "^2.1.2",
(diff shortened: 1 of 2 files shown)


eliotlim (approved):
LGTM! 👍

PR #8649, opened 2025-08-27
fix: Settings Singpass page disabled changes

KenLSM:
## Problem

- Settings Page > Singpass section when disabled does not have enough contrast
- Disabled visuals looks different from other disabled sections

Closes FRM-2136

## Solution

**Breaking Changes** 

## Before & After Screenshots

| **BEFORE** | **AFTER** |
| --- | --- |
| image | image |

**(update) Singpass options have since collapsed**


frontend/src/templates/Field/Radio/RadioField.tsx
```diff
@@ -121,7 +121,6 @@ export const RadioField = ({
                 // Required should apply to radio group rather than individual radio.
                 isRequired={false}
                 isDisabled={schema.disabled}
-                isHighContrast={isHighContrast}
```
kevin9foong, line 124: curious: wondering why the radio field isHighContrast is removed

kevin9foong (commented):
lgtm if the radio field high contrast removal is justified - not sure why it was removed

frontend/src/templates/Field/Radio/RadioField.tsx
kevin9foong: Okay - i understand now. It's because the default opacity now is high contrast, so no need for setting high contrast anymore.

kevin9foong (approved):
lgtm - just have a couple merge conflicts

eliotlim (approved):
LGTM! 👍

PR #8648, opened 2025-08-27
chore(tests): upgrade msw, msw-storybook-addon, storybook

KenLSM:
## Problem

- There are a couple of migrations that's need in order for us to merge #8593

Closes #8593

## Solution

- Upgraded `msw@v1` to `msw@v2`
- Upgraded `msw-storybook-addon@v1` to `msw-storybook-addon@v2`
- Upgraded `vitest@v1` to `vitest@v3`

`msw` has a couple of breaking changes
1. `rest` -> `http`
2. `ctx.delay` -> `import { delay } from 'msw'`
3. `ctx.json` -> `import { HttpResponse } from 'msw'`
4. typing changes

Config changes on tests

**Breaking Changes** 

- No - this PR is backwards compatible

PR #8651, opened 2025-08-27
chore(test): page level delays, inherit msw handlers

KenLSM:
## Problem

We have a lot of flakey chromatic tests, which makes UI review difficult, due to the false positives, for our designers, especially when we're trying to introduce an array of UI fixes.

1. Various apis that the Page-level stories calls are not handled under `msw`, i.e. `/client/env`, `/feature-flag`. 
2. Delays not present for several Page-level stories which results in UI diff being highlighted by chromatic. This introduces a lot of noise for the reviewer.

## Solution

1. In order for us to import globally injected msw values, we have to migrate to `Object{}` syntax instead of `Array[]` during the parameters declaration. This is only supported in `>= msw@1.6` which is already addressed in #8648.
2. Add and standardise delays (`delay: 300`)to avoid chromatic capturing the snapshot as the Shimmering Loading animation is still animating.

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
+++ b/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
@@ -39,7 +39,7 @@ export default {
   component: FeedbackPage,
   parameters: {
     // Required so skeleton "animation" does not hide content.
-    chromatic: { pauseAnimationAtEnd: true, delay: 200 },
+    chromatic: { pauseAnimationAtEnd: true, delay: 300 },
     layout: 'fullscreen',
     msw: { handlers: { default: DEFAULT_MSW_ROUTES } },
   },
--- a/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
+++ b/frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
@@ -7,6 +7,7 @@ import {
   getAdminFormCollaborators,
   getAdminFormFeedback,
   getAdminFormIssue,
+  getAdminFormSubmissions,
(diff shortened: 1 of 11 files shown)


frontend/src/features/admin-form/AdminFormResultsFeedbackPage.stories.tsx
```diff
@@ -30,16 +31,17 @@ const DEFAULT_MSW_ROUTES = [
...
+  getAdminFormSubmissions(),
 ]
 
 export default {
   title: 'Pages/AdminFormPage/Results/FeedbackTab',
   component: FeedbackPage,
   parameters: {
     // Required so skeleton "animation" does not hide content.
-    chromatic: { pauseAnimationAtEnd: true },
+    chromatic: { pauseAnimationAtEnd: true, delay: 200 },
```
KenLSM:
```suggestion
    chromatic: { pauseAnimationAtEnd: true, delay: 300 },
```

frontend/src/features/admin-form/AdminFormSettingsPage.stories.tsx
```diff
@@ -51,16 +51,20 @@ export default {
   ],
   parameters: {
     // Required so skeleton "animation" does not hide content.
-    chromatic: { pauseAnimationAtEnd: true },
+    chromatic: { pauseAnimationAtEnd: true, delay: 200 },
```
KenLSM:
```suggestion
    chromatic: { pauseAnimationAtEnd: true, delay: 300 },
```

eliotlim (approved):
looks like a bunch of noise in the Chromatic diffs on the Singpass Settings page in the Form Admin view.
I'm chalking it up to better delay handling.

PR #8659, opened 2025-08-29
feat(singpass): hiding unselected collapsed options

KenLSM:
## Problem

- Admins can create Singpass forms and set to options that will be collapsed
- We want to mitigate migration risks that will happen on Sept 5 by allowing admins to switch their forms earlier

Closes FRM-2139

## Solution

- Hide collapsed options to prevent admins selected collapsed options
- Only show options that were hidden so that admins can only switch to offered options

```
Singpass App-only Login           | NA // Removed
Singpass App-only with Myinfo     | NA // Removed
Singpass                          | NA // Removed
Singpass with Myinfo              | Singpass  // renamed
Singpass (Corporate)              | Corppass
```

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

**AFTER**:

### DIFF
--- a/__tests__/e2e/helpers/createForm.ts
+++ b/__tests__/e2e/helpers/createForm.ts
@@ -329,7 +329,7 @@ const addAuthSettings = async (
 
   // Don't need to click if SGID is desired auth type
   // since SGID is the default once Singpass is enabled
-  if (formSettings.authType !== FormAuthType.SGID) {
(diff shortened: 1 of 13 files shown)


eliotlim (approved):
LGTM! 👍 

we briefly explored other ways of forcing the authType in the backend, but weren't able to come up with anything that didn't introduce major risk considerations.

will execute the data fix as planned.

PR #8665, opened 2025-09-02
fix(ui): highContrast fixes for basic fields

scottheng96:
## Problem

Email field and verifiable email field has different disabled UI, leading to user confusion.

## Solution

1. Pass `isHighContrast` through `VerifiableFieldContainer` so that FormLabels keep highContrast
2. Pass `isHighContrast through `EmailFieldInput` in `VerifiableFieldContainer` through -> `Input` to take reference of prop
3. Update <Input> to take isHighContrast prop as part of InputStyles

**Breaking Changes** 

- No - this PR is backwards compatible  

## Before & After Screenshots

|Description| **BEFORE** | **AFTER** | 
| ----| ---- | ---- |
| Prefills | image | image |
| Verifiable email field | image | image |
| Address field | image| image|
| Signature field | image |image |
| Yes/No field | image| image|
| Radio field |  new |old | 
| Dropdown field | image old1 old2 | image new1 new2 |

PR #8682, opened 2025-09-09
fix: high contrast UI fixes part 2

scottheng96:
## Problem

There are inconsistencies with disabled states across multiple fields.

## Solution

- Removed opacity differentiation when component is disabled
- Introduced `isHighContrast` boolean to make disabled color and bg color consistent
- Remove placeholder text by conditionally showing based on `isDisabled` 

Changes are related to the outstanding items in the PRD not worked on in [part 1](https://github.com/opengovsg/FormSG/pull/8665). These include:
1. MyInfo fields
2. Storage & MRF mode email settings page
3. Dropdown field cancel button disabled bg change on hover

**Breaking Changes** 

## Before & After Screenshots

| --- | --- | --- |
| MyInfo highContrast | image | image |
| Storage mode disabled settings page | image | image |
| MRF mode disabled settings page | image | image |

### DIFF
--- a/frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
@@ -153,7 +153,10 @@ export const FormEmailSection = ({
               'features.adminForm.settings.emailNotifications.section.regular.label',
             )}
           </FormLabel>
-          <AdminEmailRecipientsInput onSubmit={handleSubmitEmails} isDisabled={isDisabled} />
+          <AdminEmailRecipientsInput
+            onSubmit={handleSubmitEmails}
+            isDisabled={isDisabled}
+          />
           <FormErrorMessage>{get(errors, 'emails.message')}</FormErrorMessage>
           {isEmpty(errors) ? (
             <FormLabel.Description
--- a/frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
@@ -162,7 +162,7 @@ export const FormEmailSection = ({
             <FormLabel.Description
               color="secondary.400"
               mt="0.5rem"
-              opacity={'1'}
+              opacity='1'
             >
               {t(
                 'features.adminForm.settings.emailNotifications.section.regular.description',
--- a/frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
+++ b/frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
@@ -162,7 +162,7 @@ export const FormEmailSection = ({
             <FormLabel.Description
               color="secondary.400"
               mt="0.5rem"
-              opacity='1'
+              opacity="1"
             >
               {t(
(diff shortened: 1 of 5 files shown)


frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
```diff
@@ -143,18 +147,22 @@ export const FormEmailSection = ({
...
+            onSubmit={handleSubmitEmails}
+            isDisabled={isDisabled}
+          />
           <FormErrorMessage>{get(errors, 'emails.message')}</FormErrorMessage>
           {isEmpty(errors) ? (
             <FormLabel.Description
               color="secondary.400"
               mt="0.5rem"
-              opacity={isDisabled ? '0.3' : '1'}
+              opacity={'1'}
```
kevin9foong: nit: Since this is no longer a ternary operation, it an be simplified
```suggestion
              opacity='1'
```

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
```diff
@@ -29,6 +29,7 @@ import { RespondentCopyToggle } from './EmailNotificationsSection/RespondentCopy
 interface MrfEmailNotificationsFormProps {
   settings: MultirespondentFormSettings
   isDisabled: boolean
+  isHighContrast: boolean
```
kevin9foong, line 32: nit: some inconsistencies in `isHighContrast` type between the props, some is `boolean` and others are `boolean | undefined` - wondering if we should standardize them

kevin9foong (commented):
some nits - otherwise lgtm

frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
scottheng96: Thanks! removed this

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
scottheng96: I thought about it, but realised that the effort then would be to default `undefined` values to `false` over the same places which made it seem less worthwhile

frontend/src/features/admin-form/settings/components/FormEmailSection.tsx
```diff
@@ -162,7 +162,7 @@ export const FormEmailSection = ({
             <FormLabel.Description
               color="secondary.400"
               mt="0.5rem"
-              opacity={'1'}
+              opacity='1'
```
eliotlim:
```suggestion
              opacity="1"
```

frontend/src/features/admin-form/settings/components/MrfFormEmailSection.tsx
eliotlim: good point!
1. let's get this UI issue fixed first, and
2. re-look at the component design when we re-align with the design system in the future.

eliotlim (approved):
LGTM! 👍

PR #8687, opened 2025-09-10
build: release v6.238.0

kevin9foong:
## New

- feat: move pdf gen to lambda  [`#8605`](https://github.com/opengovsg/FormSG/pull/8605)
- build: merge release v6.237.0 to develop [`#8684`](https://github.com/opengovsg/FormSG/pull/8684)
- build: release v6.237.0 [`#8683`](https://github.com/opengovsg/FormSG/pull/8683)

## Dependencies

- fix(deps): bump helmet from 7.1.0 to 8.1.0 [`#8284`](https://github.com/opengovsg/FormSG/pull/8284)

### feat: move pdf gen to lambda [`#8605`](https://github.com/opengovsg/FormSG/pull/8605)

TC1: Can generate auto-reply PDF 
- [ ] Create a new storage mode form 
- [ ] Add an email field and add PDF responses under the email field settings left drawer. 
- [ ] Add a bunch of fields (eg, signature, short text, table fields) - aim for 5 - 10 fields. 
- [ ] Submit a response to the form - include chinese characters in your response. 
- [ ] Assert that an email with PDF is generated and sent to the email specified in the form response.
- [ ] Open and assert that the PDF content is generated without visual issues. 

TC2: Can generate payment invoice PDF
- [ ] Create a payments form 
- [ ] Submit the payments form and download PDF invoice as a respondent. 
- [ ] Assert that the payment invoice PDF is generated without visual issues. 
- [ ] Go to the admin results page for the payments form. 
- [ ] Open the response result and download PDF invoice from admin page. 
- [ ] Assert that the PDF invoice is downloaded without visual issue.


src/app/utils/convert-html-to-pdf.ts
```diff
@@ -1,39 +1,218 @@
...
-    return pdfBuffer
+    return lambdaResult.value
+  }
+
+  logger.info({
+    message:
+      'Successfully generated pdf from html - using result from local pdf generation',
+    meta: logMeta,
   })
+  return await localResult
```
eliotlim, line 214: thanks @kevin9foong!

just a few small nits on the code structure here (edited for clarity):
- this code works and is achieves the goal of getting our hypothesis tested early, but we might have to clean this up later.
- it was initially confusing for me to figure out what each return statement did
- some return statements resolve promises, and other return statements return the top-level function
- and the data flow could be tidied up a little bit (e.g. invoke all the pdf gen mechanisms that are configured, and then return the appropriate one)

it might lead to a footgun later on, since the differing coding styles (return a promise vs immediately return) could be a code readability issue.

eliotlim (approved):

src/app/utils/convert-html-to-pdf.ts
eliotlim: for clarity: 
image

kevin9foong: > thanks @kevin9foong!
> 
> just a few small nits on the code structure here (edited for clarity):
> 
> * this code works and is achieves the goal of getting our hypothesis tested early, but we might have to clean this up later.
> * it was initially confusing for me to figure out what each return statement did
> * some return statements resolve promises, and other return statements return the top-level function
> * and the data flow could be tidied up a little bit (e.g. invoke all the pdf gen mechanisms that are configured, and then return the appropriate one)
> 
> it might lead to a footgun later on, since the differing coding styles (return a promise vs immediately return) could be a code readability issue.

@eliotlim Summarizing our discussion on these items: 

We walked through 3 things brought up as concerns: 
**1. Crossing code flow due to guard clause** 
Discussed that this is mainly a stylistic issue and the comprehensibility depends on what each engineer may be used to. Based on my past experience and education, I opted for following this convention https://refactoring.guru/replace-nested-conditional-with-guard-clauses - not claiming that this is the best convention for all use cases, but this is the inspiration behind it. 

**2. About the mix between async and .then in the same function** 
Agreed that it is not possible to avoid the use of .then in this function, as we will need to avoid unnecessary awaiting the local generation if we want to use the lambda output. 
Furthermore, we discussed the rationale behind the usage of .then: 
- want to avoid changing the local gen code to keep changes minimal, this reduces the likelihood of things breaking as the existing code is proven to work and as it will be removed in the near future. 
- the usage of .map and .then are consistent, making it easier to relate the local and lambda generation logging and metrics sending. 

**3. Usage of guard clause if(resultAsync.isErr) and avoiding .map and .mapErr**
Discussed that the main consideration is following the existing code style and consistency, where .map and .mapErr has been used. While guard clause may bring some marginal benefit depending on the reader (this is subjective), the tradeoff of having code dissimilarities might not justify the change. 
It was brought up that migrating from try catch to never throw would also cause changes in code style, and I agreed that there was significant gain when done, but in this case, my thoughts are that moving to guard clause did not bring sufficient gain to justify the tradeoffs.

eliotlim: Thanks @kevin9foong for a good, solid discussion on best practices and coding conventions earlier 😄 

1. Yup, I agree that the guard clause pattern is helpful and there are advantages to an early return in the control flow.
2 and 3. 100% concur that the differing code style is an artifact of the rich history that the FormSG codebase has accumulated over the years. In fact, it's a sign of strong engineering fundamentals to be adopting the code style and focusing changes to achieve good product + engineering outcomes.

PR #8689, opened 2025-09-10
fix: add env variable for prod samconfig

kevin9foong:
## Problem

Previously missed out setting the env, causing the lambda to use the fallback `dev` env. 

## Solution

Set the env as `production`. 

**Breaking Changes** 


eliotlim (approved):
LGTM! 👍

PR #8699, opened 2025-09-15
fix: radio fields MRF bug

scottheng96:
## Problem

In MRF form, a field that is conditionally shown/hidden on logic of a radio field's 'Others' option (filled in from a previous step) is shown when the form is loaded, but then hidden when being interacted on.

Closes FRM-2149

## Solution

Fix: Rehydrate {value: 'OTHERS'} to radio data if othersInput exists on the FE (BE is untouched)

Our radio field deals with inputs using two variables: `value`  and `othersInput`
- When radio field is being filled, if 'Others' is selected, field value willl be {value: 'OTHERS*', othersInput: 'typed something here'}`
- Logic is applied conditionally on `{value: 'OTHERS'}`
- When submitting `{value: 'OTHERS'}` is stripped from the response data, leaving `{othersInput: 'typed something here'} ` to be saved
- During rehydration of data (in step 2 e.g.) the radio field only picks up `data = {othersInput: 'typed something here'}`
- When form is loaded, radio field value will be `{value=undefined, othersInput=''typed something here`
- When the form is being interacted on, RHF will default radio field value to `value=''`, resulting in the logic not being fulfilled, thus hiding the form`

*Shortened version of actual constant used in the code

**Breaking Changes** 

## Before & After Screenshots

| **BEFORE** | **AFTER** |
| --- | --- | 
| old_radio| new_radio |

## Tests
**Field shown on logic based on radio field 'others' option**

- [x] Create an MRF form, Include a radio field with 'others' enabled, and a short text field
- [x] Add logic that IF 'others' is selected, SHOW short text field
- [x] Add a 2 step workflow assigning radio field to step 1, short text field to step 2
- [x] Make the form public and make a step 1 submission with 'others' selected
- [x] Attempt to make a step 2 submission, observe that the short text field is visible consistently when being interacted on (doesn't hide away)
- [x] Successfully submit step 2


eliotlim (approved):
LGTM! 👍 

Thanks for the swift and responsive fix, @scottheng96 !! 🚀

PR #8700, opened 2025-09-15
feat: create local pdf print version of results page

kevin9foong:
## Problem

Signature fields are not useful in isolation - they need to be bundled with the items they agreeing with to be useful. 
PDFs serve as a means to bundle this information. 

For form admin on-demand pdf requirements, opt for a local generation approach to reduce maintenance / load on servers to achieve the same outcome.  

## Solution

Define printable component to be rendered by the browser and use the print functionality to build.  

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:

image

## Tests

TC1: PDF can be generated locally on-demand 
- [ ] Create a form with all possible fields, including section, email, signature, attachment, table, checkbox etc. There should be >10 fields. 
- [ ] Make a submission
- [ ] Go to admin results page and download the PDF, ensure that the format is correct. (includes section header, correct format etc. 
- [ ] Ensure the PDF page does not break in between eg, rows such as signature field/address field etc and there is at least 24px margin from top.   
note that it does not need to look the same as email pdf. 
note that for email verified fields, it does not show [verified], this is the same behavior as email pdf

**New dependencies**:

- `react-to-print` : Fairly lightweight, used to create html for the browser to render and generate PDF. High usage and still actively maintained.


frontend/src/features/admin-form/responses/IndividualResponsePage/PrintableResponse.tsx
```diff
@@ -0,0 +1,151 @@
...
+    </table>
+  )
+}
+
+const PrintableResponse = () => {
+  const { data: form } = useAdminForm()
+  const { data, isLoading, isError } = useIndividualSubmission()
+  if (isLoading || isError || !data?.responses) return null
+  return (
+    <Box py="16px" fontFamily="sans-serif">
```
kevin9foong, line 154: use precise px instead of relative as this is to be printed on A4

frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRow.tsx
```diff
@@ -137,66 +126,10 @@ const DecryptedAddressRow = ({ row }: DecryptedRowBaseProps): JSX.Element => {
...
-      <Box
-        background="white"
-        width={`${canvasSize.width}px`}
-        height={`${canvasSize.height}px`}
-        borderColor="neutral.400"
-        borderRadius="0.25rem"
-      >
-        <canvas ref={canvasRef} />
-      </Box>
+      <RenderedSignatureCanvas row={row} />
```
kevin9foong: Reuse this code as rendered signature canvas

frontend/src/utils/showOnlyWhenPrintCss.ts
```diff
@@ -0,0 +1,6 @@
+export const showOnlyWhenPrintCss = {
+  '@media print': {
```
kevin9foong, line 2: this renders the components in the dom to be ready for printing but does not make it visible to the end user.

frontend/src/features/admin-form/responses/IndividualResponsePage/RenderedSignatureCanvas.tsx
```diff
@@ -0,0 +1,82 @@
+import { useEffect, useRef, useState } from 'react'
```
kevin9foong, line 1: copy code from DecryptedRow to be reused by the pdf generator component

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponseNavbar.tsx
```diff
@@ -126,6 +130,12 @@ export const IndividualResponseNavbar = (): JSX.Element => {
 
   const isAdminPrintPdfEnabled = useFeatureIsOn(featureFlags.adminPrintPdf)
 
+  const printableResponseRef = useRef<HTMLDivElement>(null)
+  const reactToPrintFn = useReactToPrint({
+    contentRef: printableResponseRef,
```
kevin9foong, line 135: currently does not support bulk download as it relies on the [print window](https://github.com/MatthewHerbst/react-to-print?tab=readme-ov-file#can-react-to-print-be-used-to-download-a-pdf-without-using-the-print-preview-window). but this can be supported via introducing a dedicated pdf generator library eg, pdfjs / react-print renderer

kevin9foong: product has noted bulk download as not required for now

frontend/src/features/admin-form/responses/IndividualResponsePage/RenderedSignatureCanvas.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { drawStroke } from '~utils/convertSignatureOutput'
+
+import { DecryptedRowBaseProps } from './DecryptedRow'
+
+export const RenderedSignatureCanvas = ({
+  row,
+}: DecryptedRowBaseProps): JSX.Element => {
+  const canvasRef = useRef<HTMLCanvasElement | null>(null)
+  const containerRef = useRef<HTMLDivElement | null>(null)
+  const [availableWidth, setAvailableWidth] = useState<number | null>(null)
```
kevin9foong: move this rendering out to be re-used + make the scaling of the canvas sizing dynamic

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponseNavbar.tsx
```diff
@@ -115,17 +119,23 @@ export const IndividualResponseNavbar = (): JSX.Element => {
...
 
   useEffect(() => {
-    if (user) {
-      gb?.setAttributes({
-        id: user._id,
-        email: user.email,
-        agency: user.agency,
+    if (user && gb) {
+      gb.setAttributes({
+        adminEmail: user.email,
```
kevin9foong, line 124: change to match the attributes for other elements

frontend/src/features/admin-form/responses/IndividualResponsePage/RenderedSignatureCanvas.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  return (
+    <Box
+      ref={containerRef}
+      background="white"
+      width={widthPx ?? '100%'}
+      borderColor="neutral.400"
+      borderRadius="0.25rem"
+    >
+      {img && <img width={widthPx ?? containerWidth} src={img} />}
```
kevin9foong: use image which is resizable and uses less render loops

frontend/src/features/admin-form/responses/IndividualResponsePage/RenderedSignatureCanvas.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+interface RenderedSignatureCanvasSizeProps {
+  /**
+   * Size of the signature in pixels.
+   * If not provided, the signature will be scaled to fit the container.
+   */
+  widthPx?: number
+}
+
+export const RenderedSignatureCanvas = ({
```
scottheng96: This looks similar to `DecryptedSignatureRow`, would it be tricky to refactor this out into a common function, or to re-use this in DecryptedSignatureRow`?

eliotlim: Oh, the [RenderedSignatureCanvas seems to be re-used in DecryptedSignatureRow](https://github.com/opengovsg/FormSG/pull/8700/files#r2348591229) already 😅

frontend/src/features/admin-form/responses/IndividualResponsePage/SignatureCanvas.tsx
```diff
@@ -0,0 +1,118 @@
...
+      {img && (
+        <img
+          width={widthPx ?? containerWidth}
+          style={{ maxWidth: canvasSize.width, maxHeight: canvasSize.height }}
+          src={img}
+        />
+      )}
+      <canvas ref={canvasRef} style={{ display: 'none' }} />
+    </Box>
+  )
```
eliotlim, line 133: I think we need to conditionally render this canvas or omit this canvas from the printed PDF.

```
Uncaught InvalidStateError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The image argument is a canvas element with a width or height of 0.
```

image

eliotlim (commented):
I might have found an edge case 😅

- where signatures are optional
- and the response doesn't capture a signature
- the PDF renderer throws an error

frontend/src/features/admin-form/responses/IndividualResponsePage/SignatureCanvas.tsx
kevin9foong: thanks @eliotlim for pointing this out. nice catch - let me have a look and see how to resolve this

frontend/src/features/admin-form/responses/IndividualResponsePage/RenderedSignatureCanvas.tsx
kevin9foong: yep - i reused the existing logic as-is (copy pasted) and moved out the `DecryptedSignatureRow` logic into its own component, with an additional step to convert into an image so it can be scaled and resized according to screen size for mobile/a4 pdf screens. 

This approach maintains the already implemented sharpness and avoids needing to alter the `x` and `y` settings which are rather complex to do 💀.

frontend/src/features/admin-form/responses/IndividualResponsePage/SignatureCanvas.tsx
kevin9foong: fixed in [e60e7eb](https://github.com/opengovsg/FormSG/pull/8700/commits/e60e7eb62b0d1fc3f406f84d97a84dd043707f7a)!

kevin9foong: added tc to cover and prevent regressions in [b9daf80](https://github.com/opengovsg/FormSG/pull/8700/commits/b9daf8000d57411757e5f009fb7b0eb25cf211a4)

eliotlim (approved):
LGTM! 👍 

Thanks for also working together to fix the edge case handling and flaky UI diff with better tests 💯

PR #8091, opened 2025-02-12
feat: save draft v1.0

kevin9foong:
## Problem

Part of paper forms elimination initiative. 
Some paper forms remain on paper due to having too many fields and require filling up by respondents over multiple hours / days. 
On paper, the filled fields remain filled and there is an intrinsic "save draft" functionality. However, FormSG does not currently support this and hence cannot support such long paper forms effectively.  

## Solution

Implement save draft where responses are saved to the local browser - using "what you have" as a means of linking drafts to end respondents. 

1. Uses IndexedDB to store responses - which allow saving of blobs and attachments locally. 
2. Handles form definition updates by admins by comparing checksum differences - uses the direct json string as simple checksum so it is accurate and as the field definition is non-sensitive information. 
3. Excludes myinfo fields and fields that have been deleted from future saved drafts and restorations. 
4. Handles edge cases such as preview, template and form duplication.  

Chromatic stories and ut have been implemented for anti-regression. 

**Breaking Changes** 

No - this PR is backwards compatible  

## Before & After Screenshots 
Note that these screenshots are also snapshotted by chromatic + interactive tests for anti-regression. 

image

Enable save draft in form settings

image

Saving of draft by respondents

## Tests

TC1: Save draft works for mrf 
- [x] Create a new mrf form (use mrf to generate fields, define a workflow with >=2 steps)
- [x] Enable save draft in general settings 
- [x] Go to step 1, enter some fields
- [x] Click save draft
- [x] Refresh submit form page 
- [x] Assert draft is restored
- [x] Change field definitions in form builder
- [x] Refresh submit form page
- [x] Assert draft is partially restored (does not restore changed fields) 
- [x] Fill in form and submit
- [x] Go to step 2 of form 
- [x] Fill in fields and save draft for step 2
- [x] Open another step 1 of mrf form
- [x] Assert no draft exists
- [x] Go back to step 2 and refresh 
- [x] Assert step 2 draft is restored
- [x] Submit a new step 1 to go to step 2 for another mrf submission 
- [x] Assert that the other step 2 does not use the draft of the first step 2. 

TC2: Form def constant for mrf step 2 and onwards
- [x] Continue with mrf form in TC1
- [x] Submit till step 2
- [x] Partially fill in step 2 and save draft. 
- [x] Change form definitions in admin builder. 
- [x] Refresh and assert that the draft can be fully restored - since step >= 2 uses the snapshotted form definitions. 

TC3: Save draft does not save myinfo and verified fields and works for attachments
[...]

PR #8702, opened 2025-09-16
chore: temporarily revert save draft

eliotlim:
This reverts commit 17b541d631194c9fd64df721e875e9d9b1ee98d2.

## Problem

We need a bit more time on #8091 

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Revert #8091


kevin9foong:
No need to revert as we can keep the feature flag off :)

eliotlim:
ah good point, okay let's do that then.

PR #8704, opened 2025-09-16
fix: home number fields are added to the draft despite not being modified 

kevin9foong:
## Problem

Home number fields are added to the draft despite not being modified. This causes issues with displaying the correct restored draft toast message which relies on knowing which fields have been changed. 

For example, user did not edit home number field, but its value is added to the draft. Admin removes home number field. user restores draft and sees incorrect toast message telling him/her the draft is not fully restored. 

## Solution

Set the empty value from `undefined` to '""' for home number, to match the other field types. 

**Breaking Changes** 

## Tests

TC1: Home number field not added to draft if not edited 
- [x] Create form with home number field 
- [x] Save draft without editing. 
- [x] Assert draft does not include home number field. 
- [x] Edit the home number field. 
- [x] Assert draft included edited home number field and can be restored. 
- [x] Assert can submit the restored draft value.

### DIFF
--- a/frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
+++ b/frontend/src/components/PhoneNumberInput/PhoneNumberInput.tsx
@@ -89,7 +89,7 @@ export const PhoneNumberInput = forwardRef<PhoneNumberInputProps, 'input'>(
       allowInternational,
       onChange,
       onBlur,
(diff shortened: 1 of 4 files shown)


eliotlim (approved):
LGTM 👍 

thanks for the rapid and responsive fix, @kevin9foong !!

PR #8706, opened 2025-09-16
fix: improve flaky tooltip test

eliotlim:
## Problem

The test which detects the last saved draft tooltip in #8091 and #8705 may be somewhat flaky

Closes FRM-2150

## Solution

Prevent the test from failing prematurely before the tooltip has been inserted into the Document Object Model (DOM).

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Improvements**:

- Try introducing a dummy element not in the document to prevent the test from crashing

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

The UI tests should pass.

PR #8707, opened 2025-09-16
fix: save draft interaction test to never flake again

kevin9foong:
## Problem

Fix flaky tests related to save draft interaction stories. 
This is mainly caused by missing out the `await` before  each `step`. This means the tooltip check can occur before the hover occurs.  

Fixes FRM-2150.

## Solution

Add the required await statements before each step. 

This fix has been verified by running the tc locally and checking for the sequencing and that it awaits properly. 

**Breaking Changes** 


eliotlim (approved):
thanks for fixing this flaky test, @kevin9foong! 💯 

I've cherry-picked 5229220 from #8706 to make chromatic ignore always-changing data.

approved! 👍

PR #8723, opened 2025-09-17
fix: phone number fields do not work well with save draft, has +ext on blur 

kevin9foong:
## Problem

Phone number fields do not work well with save draft during restoration, has +ext on blur 

## Solution

Change to controlled component. 
Simplify the component by removing internal state

**Breaking Changes** 

## Tests
TC1: Can change, restore draft and submit 
- [ ] Create form with home number, mobile fields with intl and non intl toggles (should be 4 in total) 
- [ ] Change the field values and country 
- [ ] Ensure that the field values can change and can change country for intl numbers
- [ ] Ensure that cannot change country for non-intl 
- [ ] Save draft
- [ ] Ensure the draft is restored correctly (no +xxx in the input and the correct country is selected) 
- [ ] Submit and ensure the result is saved correctly.

### DIFF
--- a/frontend/src/components/PhoneNumberInput/IntlPhoneNumberInput.tsx
+++ b/frontend/src/components/PhoneNumberInput/IntlPhoneNumberInput.tsx
@@ -40,7 +40,6 @@ export const IntlPhoneNumberInput = forwardRef<
     innerInputRef,
     inputPlaceholder,
     handleInputChange,
-    handleInputBlur,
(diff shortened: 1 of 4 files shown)


kevin9foong:
Not currently save to merge - have issues with +1 extension and non significant national number

kevin9foong:
Actually, the current solution without this fix seems alright on further look, closing this PR

PR #8724, opened 2025-09-18
fix: signatures v1.3 bugs

scottheng96:
## Problem

1. In MRF, submission blocked (Zod validation error) at current step if signature is assigned to succeeding step
2. MRF image does not get rehydrated

## Solution

1. Update validateResponses to check for input.value instead of input. Default Signature is now changed to 'input' is ['draw', undefined], or ['draw', null]
2. Utilize useWatch to rehydrate/redraw signature if value comes async after form loads

**Breaking Changes** 

## Before & After Screenshots

| Description | **BEFORE** | **AFTER** |
| --- | --- | --- |
| MRF succeeding assigned field error | old | new |
| rehydration delay | ol1 | ne1 |

## Tests

**Rehydration for MRF signature value**
- [x] Create a 2 step workflow MRF form with signature (step 1) and short answer (step 2)
- [x] Attempt to make a step 1 submission
- [x] During attemp to make step 2 submission, observe that previous step signature is present

**Submission successful at current step with signature field in suceeding step**
- [x] Create a 2 step workflow MRF form with signature (step 2) and short answer (step 1)
- [x] Attempt to make a step 1 submission successfully
- [x] Attempt to make a step 2 submission sucecssfully

### DIFF
--- a/frontend/src/features/public-form/utils/inputTransformation.ts
+++ b/frontend/src/features/public-form/utils/inputTransformation.ts
@@ -214,7 +214,7 @@ const transformToSignatureOutput = (
   input?: SignatureFieldValues | SignatureFieldResponseV3,
 ): SignatureResponse => {
   let answerArray: string[] = []
-  if (input && input.value !== undefined && input.value !== null) {
+  if (input && input.value !== null) {
     answerArray = [input.type, convertToSignatureStringOutput(input.value)]
   }
   return {
--- a/frontend/src/features/public-form/utils/inputTransformation.ts
+++ b/frontend/src/features/public-form/utils/inputTransformation.ts
@@ -214,7 +214,7 @@ const transformToSignatureOutput = (
   input?: SignatureFieldValues | SignatureFieldResponseV3,
 ): SignatureResponse => {
   let answerArray: string[] = []
-  if (input !== undefined && input !== null) {
+  if (input && input.value !== undefined && input.value !== null) {
     answerArray = [input.type, convertToSignatureStringOutput(input.value)]
   }
   return {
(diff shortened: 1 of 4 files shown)


frontend/src/features/public-form/utils/inputTransformation.ts
```diff
@@ -214,7 +214,7 @@ const transformToSignatureOutput = (
   input?: SignatureFieldValues | SignatureFieldResponseV3,
 ): SignatureResponse => {
   let answerArray: string[] = []
-  if (input !== undefined && input !== null) {
+  if (input && input.value !== undefined && input.value !== null) {
```
kevin9foong: Hmm, seems like the TS is failing somewhere - cause based on the SingatureFieldValues type, the `value` should always be defined

frontend/src/features/public-form/utils/inputTransformation.ts
```diff
@@ -214,7 +214,7 @@ const transformToSignatureOutput = (
   input?: SignatureFieldValues | SignatureFieldResponseV3,
```
kevin9foong, line 214: would recommend updating SignatureFieldValues `value` key to be `SignatureVectorArray | undefined | null` to utilise TS safety

eliotlim: thanks for spotting this, @kevin9foong ! yup agreed.

eliotlim: @scottheng96 and I discussed it, [fixed it here](https://github.com/opengovsg/FormSG/pull/8724/commits/fd2707443472f6eb0bf114affd155d104a04642f) :)

eliotlim (approved):

PR #8726, opened 2025-09-18
fix: update signature defaultValue to standardized format

scottheng96:
## Problem

Signatures field values differ based on various interactions in the FE. This is complex due to signature field being an uncontrolled component.

If field is untouched, value = `undefined`
If field is registered, default values to = `""` (e.g. when field is hidden at first, and then shown)
If field is interacted with, value = `{type:'draw', value: [[1,1,1],...]}`
If field is cleared, value = `null`

This is creating a lot of inconsistencies when we deal with validations during submissions.

## Solution

1. By default, value will be `{type: 'draw', value; []}` whenever component is mounted. The only instance that its value remains `undefined` is if the field is hidden from form.
2. Clearing signature input defaults value back to `[]`, `{type: 'draw', value; []}` this will be the default state of a non-inputted signature field

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,30 +1,21 @@
 {
-  "eslint.enable": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll": "explicit",
     "source.fixAll.eslint": "explicit"
   },
-  "eslint.validate": [
-    "javascript",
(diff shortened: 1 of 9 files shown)


eliotlim (approved):
LGTM! 🚀

PR #8728, opened 2025-09-18
feat: remove local pdf generation

kevin9foong:
## Problem

Since PDF generation on lambda is stable, we can remove it from local to reduce our vCPUs.  

## Solution

Remove the local generation code. 

**Breaking Changes** 

## Tests 
TC1: PDF generation works 
TC2: PDF emulator works for generating PDFs locally


kevin9foong:
Closed as a noop solution is implemented - will do the cleanup separately from scratch referencing this PR.

PR #8729, opened 2025-09-18
feat: singpass with mrf first step

scottheng96:
## Problem

Singpass & MyInfo is currently not-enabled for MRF forms. This prevents MRF users from utilising these FormSG features and benefitting from these time-saving/useful features.

Closes FRM-2151

## Solution

Enable Singpass & MyInfo fields on MRF forms:
- Form-level Singpass authentication for step 1
- NRIC collection for step 1
- MyInfo fields assignable to step 1
- betaFlagged: `singpassMrf` & feature-flag: `singpass-mrf`

**Engineering Context**
| Item | Description |
| --- | ----- |
| betaFlag | controls ability for admins to see & update singpass-enabled settings on the form builder |
| feature-flag | controls ability to view form & make submission (only for singpass-enabled MRF forms) |
| verifiedContent encryption | (following encrypt mode) encryption for verifiedContent will happen independently from encrypting responses. This happens in `handleNdiResponses` middleware. the `submissionSecretKey` is used to encrypt verifiedContent, which is taken from the return of encrypting the responses. |
| verifiedContent object | is saved in this format: `"verified": { "uinFin": "S9630911B" }`. Following encrypt mode's verifiedContent of key:value pair, except adding step number to the verifiedContent (e.g.`"verified": { "uinFin (Step 1)": "S9630911B" }` . This is extensible to future implementations of Ndi data collection for subsequent steps (Step x): value |

Edge cases:
1. MyInfo children is out of scope for MRF Singpass (see [notion](https://www.notion.so/opengov/Singpass-for-MRF-v1-0-step-1-only-28077dbba78880dba1ccfc0b5f3b3374?source=copy_link))

**Breaking Changes** 

## Before & After Screenshots

| Item |**BEFORE** | **AFTER** |
| --- | --- | --- |
| whitelisted admin signatures settings page | image | image |
| feature flag off (disabled submissions) | | image |
| workflow myInfo fields in step 1 workflow | | singpassmrf |

## Tests

**Create an MRF Singpass-enabled form with no MyInfo fields**
- [x] Create an MRF form and enable Singpass on the form
- [x] Add 1-2 basic fields to the form
- [x] Add a workflow (2 steps) and assign basic fields to any step
- [x] Add your email to receive email confirmation when workflow ends
- [x] Attempt to make a step 1 submission; observe that you must sign in using Singpass on step 1 *only*
- [x] Attempt to successfully complete the workflow

**Create MRF Singpass-enabled form with MyInfo field & 2 workflow steps** 

- [x] Using the MRF form created prior, update singpass settings to toggle on 'Collect NRIC data'
- [x] Add 1-2 MyInfo fields into the form
- [x] Assert that you are unable to assign MyInfo fields onto the second step
- [x] Make the form public, and successfully make a submission of step 1
- [x] Observe in response page that MyInfo fields is displayed correctly (with [MyInfo] prefix in question)
- [x] Observe in response page that NRIC data is displayed correctly
- [x] Attempt to successfully make a step 2 submission; observe that MyInfo data is rehydrated correctly
- [x] Revisit the response page & observe that responses are displayed correctly
- [x] Observe in confirmation email that responses are displayed correctly


kevin9foong:
Flagging mechanisms: 
- when gb flag is disabled, will make mrf forms with singpass enabled not found

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -718,7 +726,17 @@ export const PublicFormProvider = ({
...
+    //TODO: FRM-2151 remove when SingpassMRF is out of beta
+    const isSingpassMrfDisabled =
+      data?.form?.responseMode === FormResponseMode.Multirespondent &&
+      data?.form?.authType !== FormAuthType.NIL &&
+      !enableSingpassMrfFeatureFlag
+
+    if (
+      isFormNotFound ||
+      isNonMultirespondentFormWithPreviousSubmissionId ||
+      isSingpassMrfDisabled
```
kevin9foong: issue: When the feature flag is disabled, it seems it shows the incorrect error message: not found -> shall we set it to form not available instead?

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
```diff
@@ -10,6 +10,7 @@ import {
   BasicField,
   FieldResponse,
   FormFieldDto,
+  FormWorkflow,
```
kevin9foong: issue (minor): seems to be unused - should this be removed?

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -1284,6 +1309,19 @@ export const PublicFormProvider = ({
...
         }
         case FormResponseMode.Multirespondent:
+          //TODO: FRM-2151 remove when SingpassMRF is out of beta
+          // if mrf form is singpass-enabled & feature flag is off, prevent submission
+          if (
+            !enableSingpassMrfFeatureFlag &&
+            form?.authType !== FormAuthType.NIL
+          ) {
+            const singpassMrfError = new Error(
+              'Submission not allowed: Singpass MRF current is disabled.',
```
kevin9foong, line 1324: issue (minor): seems to be a grammatical error here for potentially user facing message

frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormDetailsScreen.tsx
```diff
@@ -129,7 +130,7 @@ export const CreateFormDetailsScreen = (): JSX.Element => {
                   <WorkspaceRowsProvider>
                     <FormResponseOptions
                       {...field}
-                      isSingpass={isSingpass}
+                      hasMyInfoChildren={hasMyInfoChildren}
```
kevin9foong, line 133: superb use of named variable to increase understandability 👍

frontend/src/templates/Field/Signature/SignatureField.tsx
```diff
@@ -26,6 +26,8 @@ export interface SignatureFieldProps extends BaseFieldProps {
 
 const strokePressureDefault = 0.5
 
+const defaultType = 'draw'
```
kevin9foong: issue: this is not used and should be removed

src/app/modules/verified-content/verified-content.utils.ts
```diff
@@ -20,9 +20,15 @@ const isCpVerifiedContent = (
...
     VerifiedKeys.CpUid,
   ]
-  return cpKeys.every(
-    (cpKey) => cpKey in data && typeof data[cpKey] === 'string',
-  )
+  return cpKeys.every((cpKey) => {
+    if (cpKey in data && typeof data[cpKey] === 'string') return true
+
+    // Check for any step-suffixed key
+    const stepKeyPattern = new RegExp(`^${cpKey} \\(Step \\d+\\)$`)
```
kevin9foong, line 27: thoughts: not claiming that this is not ideal, but im seeing a couple usages of regexp in the code and usually when i see regex, ill wonder if we could achieve the same thing without using RegExp. Reason is that [highway](https://github.com/opengovsg/highway/pull/407) has similar code that has caused issues previously if not tested well - though that followed an [antipattern where user input was directly parsed](http://blog.regexhero.net/2010/12/when-to-use-and-when-not-to-use-regular.html) and may not be directly applicable here. For example,`new RegExp` can potentially throw exception if the arg passed into it is not accepted and cause crashing.

Hence, something that catches my attention during review to look into with more scrutiny and figure out if there's a simpler / more understandable alternative solution.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -842,3 +848,172 @@ export const encryptSubmission = async (
...
+            return uinFin
+          })
+
+        if (jwtPayloadResult.isOk()) {
+          userName = jwtPayloadResult.value
+        }
+        break
+      }
+      default:
+      // TODO: handle default
```
kevin9foong: issue: is this un-implemented TODO expected?

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -842,3 +848,172 @@ export const encryptSubmission = async (
...
+  const { formId } = req.params
+  const { authType } = formDef
+  const { submissionPublicKey } = req.formsg.encryptedPayload
+  const stepNumber: number = req.body.workflowStep
+    ? req.body.workflowStep + 1
+    : 1
+  let responses = req.formsg.encryptedPayload.responses // to add NDI data to responses (used for email payload downstream)
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let ndiResponses: Record<string, any> = {}
```
kevin9foong, line 855: thoughts: should we define a type for ndiResponses instead of defining any and skipping TS checks (see quite a few occurrences of this TS ignore line in our code)? using this might be a last resort if the above is not possible. 

usually, this ignore / disable of typescript is only used in our test files - and should ideally be avoided in our main code unless there's no other options. 

Why is this a potential issue? we use TS as a type safety mechanism to catch bugs and lack of type definitions might cause issues down the road (similar to input.`undefined` issues crashing our container which may not have been caught due to ts not being declared)  

Another benefit of defining the TS type would be to provide useful documentation for future devs on what the ndiResponses is expected to be eg, `uinFin, cp...`

hence, would be good to define it if possible

shared/types/response-v3.ts
```diff
@@ -6,7 +6,11 @@ import {
   SignatureVectorArray,
 } from './field'
 
-export type FieldResponsesV3 = Record<FormFieldDto['_id'], FieldResponseV3>
+export type FieldResponsesV3 = Record<FormFieldDto['_id'], FieldResponseV3> & {
+  [key: string]: FieldResponseV3 | NdiResponseV3
+}
+
+// export type NdiResponsesV3 = [key: string]: NdiResponseV3
```
kevin9foong: issue: stray comment?

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -842,3 +848,172 @@ export const encryptSubmission = async (
...
+      },
+    )
+
+    if (prevDecryptedSubmission?.verified) {
+      ndiResponses = { ...prevDecryptedSubmission.verified, ...ndiResponses }
+    } else {
+      logger.error({
+        message: 'Unable to get verified content',
+        meta: logMeta,
+        // error,
```
kevin9foong: issue: is this commented out `error` expected?

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: issue: is this TODO expected?

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/QuestionsBlock.tsx
```diff
@@ -37,7 +39,9 @@ export const QuestionsBlock = ({
     .filter(
       (f) =>
         // Only retain actual inputs (exclude header, statement, image)
-        !NON_RESPONSE_FIELD_SET.has(f.fieldType),
+        !NON_RESPONSE_FIELD_SET.has(f.fieldType) &&
+        // Retain MyInfo fields on the first step
+        (!('myInfo' in f) || isFirstStep),
```
kevin9foong: suggestion (understandability): while not done previously, it is a [good opportunity here](https://medium.com/engineering-managers-journal/the-boy-scout-rule-for-better-code-e6ffda090902) to start using named exports to improve code readability. 

something like... 
const isMyInfoField = (!('myInfo' in f) 
const isFillableField = !NON_RESPONSE_FIELD_SET.has(f.fieldType)
(f) => isFillableField && (!isMyInfoField || isFirstStep)

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
```diff
@@ -24,48 +25,57 @@ import {
...
  */
-const getResponseFromVerifiedField = (
-  type: VerifiedKeys,
+const getVerifiedFieldFromResponse = (
+  type: VerifiedKeys | string,
   value: string,
 ): VerifiedFormField | null => {
-  switch (type) {
+  // Extract base key and optional step number
+  const match = type.match(
```
kevin9foong: suggestion: could this be renamed to give it a [more descriptive name](http://nus-cs2103-ay1819s1.github.io/cs2103-website/se-book-adapted/chapters/codeQuality.html#intermediate-2)? for both `type` and `match` eg, for type: singpassAuthType, for match: include what is it matching? 

also, what does a 'base key' mean? could it have a more illustrative name? 

reason why i bring this up is that previously in our PublicFormProvider, the form data is simply named `data` which is then mutated in subsequent code. Explicit naming can help with avoiding such issues - especially as context is lost over the years. 


Hence, [naming to be more descriptive can be very helpful for future developers in understanding what the code does and tries to match. ](https://carlalexander.ca/importance-naming-programming/)

shared/utils/verified-content.ts
```diff
@@ -24,3 +24,29 @@ export enum SPCPFieldTitle {
...
+
+export const VerifiedKeyToSPCPTitleMap: Record<VerifiedKeys, SPCPFieldTitle> = {
+  [VerifiedKeys.SpUinFin]: SPCPFieldTitle.SpNric,
+  [VerifiedKeys.CpUen]: SPCPFieldTitle.CpUen,
+  [VerifiedKeys.CpUid]: SPCPFieldTitle.CpUid,
+  [VerifiedKeys.SgidUinFin]: SPCPFieldTitle.SpNric, // safeguarding for backwards compatibility
+}
+
+/**
+ * Maps SPCP VerifiedKeys to their titles. Used during decryption/population of outputs.
```
kevin9foong, line 36: nit: grammatical error in the docs? `handle and retain in -> it during mapping`?

suggestion: could we provide an @example annotation? (reason is that it will greatly aid with the understandability of the function and i currently have to kinda gyess what it does) 
eg, given cpUid Step-0 -> produces Corpass-step-0

kevin9foong:
hey @scottheng96, thanks for pushing out this PR and at a very quick pace to add. these changes are going to be gamechanging for mrf users and unlock the value of forms for them. 

left a few comments above (subject to discussion) and will continue to look into this PR with @eliotlim's assistance

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: Has been removed! No error gets thrown from the decryptFromSubmissionKey (it returns a null if decryption has issue FYI).

shared/utils/verified-content.ts
scottheng96: I've added an example to make it more understandable for future references

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/QuestionsBlock.tsx
scottheng96: Don't see a reason why not - I've updated the code to used named consts for these conditionals.

frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/processDecryptedContent.ts
scottheng96: A sound consideration. Let me make the naming more verbose - decided to use `singpassAuthType`, `verifiedKeyMatch`, `verifiedKey` and `stepNumber` for better clarity!

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: I agree that we should lean away from ignoring TS checks where possible; for this case however I found it tricky to define a type for `NdiResponses`.

VerifiedContent is stored as key:value pairs, e.g. (`{uinFin: S1234567A}`). To support verifiedContent collection in MRF, we have to include step information, which I have decided to add in the key (`{uinFin (Step 1): S1234567A}`). Since step number is variable, its not possible to define a static type/enum for these verified keys like the way we do for storage mode.

The tradeoff here is that we'll have to be more careful in managing the title, but it makes it easer to handle displaying verifiedContent in our email/response outputs having the step number already baked into the key.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: Missed this after implementation. I've included an error log to be fired in the event an unsupported AuthType is found. No other additional handling is needed since the jwtPayloadResult succeeding this will throw another verification error and update the res with failure status code.

scottheng96: ^

frontend/src/features/admin-form/create/workflow/components/WorkflowContent/EditStepBlock/QuestionsBlock.tsx
```diff
@@ -34,11 +36,12 @@ export const QuestionsBlock = ({
...
-    .filter(
-      (f) =>
-        // Only retain actual inputs (exclude header, statement, image)
-        !NON_RESPONSE_FIELD_SET.has(f.fieldType),
-    )
+    .filter((f) => {
+      // Only retain actual inputs (exclude header, statement, image)
+      const isFillableField = !NON_RESPONSE_FIELD_SET.has(f.fieldType)
+      const isMyInfoField = 'myInfo' in f
+      return isFillableField && (isMyInfoField || isFirstStep)
```
eliotlim:
```suggestion
      if (!isFillableField) {
        return false
      }
      if (isMyInfoField && !isFirstStep) {
        return false
      }
      return true
```

frontend/src/features/public-form/PublicFormProvider.tsx
scottheng96: Done! A bit of code bloat but its only temporary while the feature flag is being used

eliotlim (approved):
Tested on UAT, LGTM! 👍 
image

PR #8732, opened 2025-09-19
feat: revert to 240

kevin9foong:


eliotlim (approved):
agreed: let's roll back the changes made from v6.241.0 to v6.244.0

v6.245.0 will be the same as v6.240.0

image

PR #8744, opened 2025-09-22
feat: save draft v1.2

eliotlim:
## Problem

Save draft experienced an issue with MRF table fields with the add additional rows option, where >=2 step submission did not restore the added rows, causing the submission to fail since it was blocked by MRF BE validation. As such, save draft was reverted until a fix can be found and applied. 

## Solution

This issue was caused by React hook form not re-rendering to use the computed defaultValues. 

Since 
```
  // Reset default values when they change
  useEffect(() => {
    if (!isDirty) {
      reset(defaultFormValues)
    }
  ``` 
 from `FormFields.tsx` was missed and not carried over when the react hook form logic was hoisted to the provider.
 
Hence, add the logic required to reset and reload the form state to the latest default values is added to the provider. Resolving this issue. 

Also includes the restoration of save draft to prod. 

**Breaking Changes** 

**Features**:

- Save Draft v1 (#8091 and #8714)

**Improvements**:

- Logging at the MRF BE validation to identify field type of fields which do not match prev submission. This was initially added as part of save draft debugging, but simplified and retained for future observability.  

**Bug Fixes**:

- Save Draft interaction test #8707 
- MRF 2nd step value restoration not using the latest save draft values. 

## Tests

TC1: Works with table fields with ability to add more rows 
- [x] Create a mrf 2 step form
- [x] First step assign table field with ability to add more rows + some intl number fields 
- [x] Second step exclude this table field 
- [x] Submit the form, ensuring to add more rows
- [x] At the second step of submission, assert that the value restored from 1st step is correct. 
- [x] Try to submit, assert it can go through

TC2: Save draft works, status tracking no toast
- [x] On the same form as TC1, Save a draft on step 1 
- [x] Refresh and assert field vals are restored exactly 
- [x] Submit step 1
- [x] Make new step 1 submission - assert draft is cleared
- [x] Save a draft on this step 1 submission 
- [x] Go to the status tracking page of the first submission, assert no toast appears 
- [x] Go to the step 2
- [x] Assert no draft is restored 
- [x] Save draft at step 2 and refresh 
- [x] Assert draft is restored 
- [x] Submit and assert it passes


eliotlim (commented):
LGTM! 👍

kevin9foong (approved):

PR #8764, opened 2025-09-29
fix: remove optionsToRecipientsMap when duplicating form

scottheng96:
## Problem

When form is duplicated using use-template, the conditional routing dropdown email mappings also get duplicated. This is due to the mapping of the options to emails being saved in the field itself, under `optionsToRecipientsMapping` variable

## Solution

Remove `optionsToRecipientsMapping` when duplicating a form

**Breaking Changes** 

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.service.ts
@@ -670,6 +670,7 @@ export const duplicateForm = (
   newAdminId: string,
   overrideParams: DuplicateFormOverwriteDto,
   workspaceId?: string,
+  duplicateStripped?: boolean,
 ): ResultAsync<IFormDocument, FormNotFoundError | DatabaseError> => {
   const overrideProps = processDuplicateOverrideProps(
     overrideParams,
@@ -690,7 +691,7 @@ export const duplicateForm = (
   const duplicateParams = originalForm.getDuplicateParams(overrideProps)
 
   // remove conditional routing mapping from dropdown fields
-  if (duplicateParams?.form_fields) {
+  if (duplicateParams?.form_fields && duplicateStripped) {
     duplicateParams.form_fields.forEach((field) => {
       if (field.fieldType === BasicField.Dropdown) {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/src/app/modules/form/admin-form/admin-form.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.service.ts
@@ -689,9 +689,11 @@ export const duplicateForm = (
 
   const duplicateParams = originalForm.getDuplicateParams(overrideProps)
 
+  // remove conditional routing mapping from dropdown fields
   if (duplicateParams?.form_fields) {
     duplicateParams.form_fields.forEach((field) => {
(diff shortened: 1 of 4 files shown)


src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -689,6 +690,16 @@ export const duplicateForm = (
 
   const duplicateParams = originalForm.getDuplicateParams(overrideProps)
 
+  // remove conditional routing mapping from dropdown fields
+  if (duplicateParams?.form_fields && duplicateStripped) {
+    duplicateParams.form_fields.forEach((field) => {
+      if (field.fieldType === BasicField.Dropdown) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ;(field as any).optionsToRecipientsMap = undefined
```
eliotlim: I did a quick check and setting `optionsToRecipientsMap` to undefined directly should work:

```suggestion
        field.optionsToRecipientsMap = undefined
```

image

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -670,6 +670,7 @@ export const duplicateForm = (
   newAdminId: string,
   overrideParams: DuplicateFormOverwriteDto,
   workspaceId?: string,
+  duplicateStripped?: boolean,
```
eliotlim: If there are several optional params (and we find ourselves having to [include `undefined` as an argument](https://github.com/opengovsg/FormSG/pull/8764/files#diff-d25e48a691b72cce48326009d12cc208e5911e05c08ebcacb1a85651e946eb39R1059)), we could include an options structure like this:

```suggestion
  { workspaceId, duplicateStripped } ?: {
    workspaceId?: string,
    duplicateStripped?: boolean,
  }
```

(we should probably also update the JSDocs for this function with this param)

eliotlim (commented):
Thanks @scottheng96 for the quick resolution to this issue!

I've tested the template / duplication feature and both work as expected. 💯 
Screenshot 2025-09-29 at 1 04 48 PM
Screenshot 2025-09-29 at 1 05 40 PM

Just a thing or two:
1. we can set optionsToRecipientsMap to `undefined` directly (suggested)
2. when multiple optional parameters are being passed as function arguments
  a. it can be tricky to keep track of the `undefined` values and what they correspond to
  b. instead we can create an `options` struct to pass them as a named argument.

src/app/modules/form/admin-form/admin-form.service.ts
scottheng96: Committed this change! Thanks 🙏

src/app/modules/form/admin-form/admin-form.service.ts
scottheng96: I've updated the function to set this options structure and updated the tests to follow this as well.

eliotlim: thanks @scottheng96! 👍

src/app/modules/form/admin-form/admin-form.service.ts
eliotlim: works! 💯

eliotlim (approved):
LGTM 👍

PR #8770, opened 2025-09-30
feat: save draft v1.3

kevin9foong:
## Problem

This reverts commit 99abc0a9893703d4225245d855854accd730d88a to restore save draft after the Table field bug is fixed. 

## Solution

Fix the save draft bug and add regression tests in https://github.com/opengovsg/FormSG/pull/8771 

## Tests

Tests
TC1: Table field MRF works
- [ ] Create an mrf form with table field with additional rows, make min rows = 3, max = 5
- [ ] Make it 4 steps, where table field is editable only on first step and third step
- [ ] First step fill in the field and set it to 5 rows
- [ ] Submit past step 2
- [ ] For step 3, edit to 4 rows
- [ ] Submit past step 4
- [ ] Assert the full workflow can go through and the correct results are displayed in the form results page after each step.


eliotlim (approved):
🚀

PR #3922, opened 2022-05-30
feat(v2): add table field creation and preview in admin form builder

karrui:
## Problem

THIS IS THE FINAL FIELD TO GO FOR ADMIN FORM BUILDER.

This PR adds the ability to create table fields in admin form builder screen.
Closes #3398 

## Solution

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Features**:

- allow creation of table fields in admin form builder screen

## Before & After Screenshots
New stories added for EditTable component.

PR #8771, opened 2025-09-30
fix: save draft public form table field unwanted update

kevin9foong:
## Problem

- Hoisting react hook form state led to table fields not reinstating the correct number of rows, causing failures in MRF ≥2nd step submission.
- 248 fix was incorrect fix, it did not address the problem as the root cause analysis was incorrect.

To understand why this is happening, we compare the previous (pre-hoisting) with the new hoisted behavior: 

**Prev:**

In `FormFieldsContainer.tsx`, there is a guard clause

```jsx
**const** renderFields = useMemo(() **=>** {
*// Render skeleton when no data*     
*if (isLoading) {*       
	*return <FormFieldsSkeleton />*    
*}*
```

This prevents rendering of the FormFields component if the previousSubmission is not yet defined.
This is important since when `FormFields.tsx` is rendered, it will compute the `defaultFormValues`, then set it in the `useForm` hook.

```tsx
// In FormFields.tsx  
const formMethods = useForm<FormFieldValues>({
  defaultValues: defaultFormValues,
  mode: 'onTouched',
})
```

The guard clause asserts that `previousSubmission` used for computing the `defaultFormValues` exists and will always be the latest value before the `defaultFormValues` are computed. 
(Without it, `previousSubmission` may be `undefined` as it is received async via network response). 
Hence, `defaultValue` is always set to the correct values since the very first render and when the `useForm` hook is run. 

**The default value being the correct value in the very first render `useForm`  is the required condition for correctness in this case.**

### Key codes of interest that explain the rendered outcome:

This is important as in TableField there are 4 important pieces of code:

- A useEffect in `FormFields.tsx` which resets the default values cache of React hook form (A)

```tsx
  useEffect(() => {
    if (!isDirty) {
      reset(defaultFormValues)
```

- A tracker for when the schema for min rows has changed (B)

```
  const hasMinRowsChanged = useHasChanged(schema.minimumRows)
```

- A use effect which add / deletes rows when the min rows schema change is detected (C)

```
  useEffect(() => {
    // Update field array when min rows changes.
    if (hasMinRowsChanged) {
      const minRows = schema.minimumRows || 0
      const prevRowLength = fields.length
      if (minRows > prevRowLength) {
        for (let i = prevRowLength; i < minRows; i++) {
          appendTableRow()
        }
      } else {
        // Remove rows from field array
        **for (let i = prevRowLength; i > minRows; i--) {
          remove(i - 1)
        }**
      }
    }
  }, [appendTableRow, fields.length, hasMinRowsChanged, remove, schema])

```

- The `remove` function provided by React hook form  (D)

```
  const { **fields**, append, remove } = useFieldArray<TableFieldInputs>({
    control: formMethods.control,
    name: schema._id,
  })
```

The key thing about this function is that **field** is retrieved, it is equivalent to the value stored in React hook form’s cached “default values”. 

```jsx
useForm(
	{
		defaultValues: <this is retreived by 'field' and stores the cached values>
  }
)
```

### Prerequisites: Understanding React useEffect() theory

To understand what happens, we need to understand the order by which React uses `useEffect`

```jsx
// Given the following where A and B call useEffect() 
<A>
	<B /> 
</A> 

// From top to bottom ordering: 
// A mounts and begins painting
// Child component B mounts and begins painting  
// B finishes painting
// A finishes painting
// commits 
// useEffect of child B is invoked
// useEffect of parent A is invoked 
```

Let’s now study what happens when the `FormFields.tsx` and `TableField.tsx` is rendered: 

1. The `useForm` hook is defined with the correct complete `defaultFormValues` 
2. `hasMinRowsChanged` in (B) evaluates to true since the previous value is `undefined` 
3. The table field is rendered completely with the `defaultValues`.
4. After table field renders, this causes the useEffect in (C) (which is in the child `TableField.tsx` to execute and remove the fields). 
5. Then, the useEffect in (A) runs and resets the fields and reinstates the correct number of rows (A). 
Since `defaultFormValues` is correct from the first initialization, it never changes. Hence, useEffect (A) is only run once.  
6. After the reset, React hook form’s cached defaultFormValues is updated to the same complete `defaultFormValues`. 
7. Since the prevVal used in (B) was set in step 2 to the schema’s minRows, which is unchanged, `hasMinRowsChanged` is false, so no rows are removed. 
8. Hence, the final result renders the correct number of rows in the correct `defaultFormValues`.

---

**248:**

For 248, the `useForm` initialization of the `defaultValue` is hoisted to the provider.

This means that there is no guard clause to check if the form and previous submission is loaded before setting the `defaultValue` in the `useForm` hook. (Doing this check is also complicated since `useForm` is a hook and cannot be run conditionally) 

For example - the following is not possible now after it is hoisted: 

```jsx
let formMethods
if (isFormLoaded && isSubmissionLoaded) {
	formMethods = useForm({defaultValues: computeDefaultValues(prevSubmission)})
}
```

Hence, `defaultValue` first returns an invalid (does not contain the latest values) in the first render before then returning the latest values after the form and submission is loaded. 

For example, when the form response mode is not yet loaded, it returns `{}`

```jsx
const defaultFormValues = useMemo(() => {
    // return {"65c753727177d728a526be25":"","65d70123e28bc57002a961aa":{"value":"Operating Theatre(s)"},"65d7113c1a59cf739deaf270":"","660a6bb003552e0dc0531968":"","65c74f097a6e43e3d46b71f9":"kevin.foong@open.gov.sg","65c753e047e450bd2968bc97":"kevin.foong@open.gov.sg","65c74f1d97b72744d22b2f60":{"value":"kevin.foong@open.gov.sg"},"65c74fc3fbdefe5b36db7f02":"+6562223535","65c7538cd602a5e429e5ffc4":"kevin.foong@open.gov.sg","65dc1f544e0ee4bae76f35f8":"No","65c7506b7a6e43e3d46b7a3b":"30/09/2025","65c7506e8af607588945eec7":"01/10/2025","65dc21841a59cf739d60f4d4":"","65d7060d38c912c32104c65c":"","65d70624f55d126c0d36b7ab":"","65c7545cd8b3cc71bfbf00e3":"","65c750ab4f8c96bc7dae795e":"kevin.foong@open.gov.sg","65c751a1fbdefe5b36db8b92":{"value":"kevin.foong@open.gov.sg"},"65c752351b19a8e90ad2aa63":[{"65c752351b19a8e90ad2aa64":"k"},{"65c752351b19a8e90ad2aa64":"f"},{"65c752351b19a8e90ad2aa64":"w"},{"65c752351b19a8e90ad2aa64":"t"}],"662f0fc14927b8088b7f7dd8":[{"65c752351b19a8e90ad2aa65":"Halifax","65c7524216a0c0b769294179":"Halifax","662f1001ddc25f32de2cfd5e":"13"},{"65c752351b19a8e90ad2aa65":"Annex Building","65c7524216a0c0b769294179":"Annex Building","662f1001ddc25f32de2cfd5e":"23"},{"65c752351b19a8e90ad2aa65":"B2","65c7524216a0c0b769294179":"B2","662f1001ddc25f32de2cfd5e":"33"}],"6603a5c7dc6b1959b18a20f0":"","65c753107177d728a526bc69":"Yes","67a9728a43736111900a7a2d":"","respondent_email_field":[]}
    if (!form?.responseMode)
      return {}
    return getInitialFormValues({
      formResponseMode: form?.responseMode,
      previousSubmission,
      previousAttachments,
      augmentedFormFields,
      currentStepNumberWorkflowStep,
      fieldPrefillMap,
      draftResponsesToRestore,
      searchParams,
      isSaveDraftEnabled,
    })
  }, [
    form?.responseMode,
    previousSubmission,
    previousAttachments,
    augmentedFormFields,
    currentStepNumberWorkflowStep,
    fieldPrefillMap,
    draftResponsesToRestore,
    searchParams,
    isSaveDraftEnabled,
  ])
```

As a result, this in particular causes multiple (2 in particular) invocations of (A).

**Let’s study what happens when the `FormFields.tsx` and `TableField.tsx` is rendered for 248:** 

1. The computed `defaultFormValue` is set to `{}` (since form response mode is not yet loaded). 

```jsx
// In PublicFormProvider.tsx (hoisted)  
const formMethods = useForm<FormFieldValues>({
  defaultValues: computeDefaultFormValues() // = {},
  mode: 'onTouched',
})
```

Hence, `FormFields.tsx` sets React hook form’s `defaultValues` as `{}` which is not the final desired value. 

1. In `TableField.tsx`, `hasMinRowsChanged` in (B) evaluates to true since the previous value is `undefined` . This also sets the previousVal of the `hasChanged` hook to `schema.minRows`
2. The table field is rendered completely with the `defaultValues` with empty table field values, since the `defaultValues` is `{}`
3. After table field renders, this causes the useEffect in (C) (which is in the child `TableField.tsx` to execute). No fields are removed since minRows ≥ currentRows
4. Then, the useEffect in (A) runs and resets the fields and reinstates the correct number of rows (A). This time it has the correct values loaded. 
5. Resetting with a different updated `defaultValue` causes a re-render of the child component `TableField.tsx` but **where the `hasMinRowsChanged` previous value is reset to `undefined`**
    1. Hence, when `TableField.tsx` is re-rendered,  `hasMinRowsChanged` is evaluated as true. 
6. After table field renders again, this causes the useEffect in (C) (which is in the child `TableField.tsx` to execute). Fields are now removed since minRows < currentRows which use the updated `defaultValue` value. 
7. This removal of rows causes the `TableField.tsx` to re-render. 
    1. However, this re-render does not reset the previous value for `hasMinRowsChanged` and it evaluates to false. 
8. Finally, the useEffect in (A) is executed in the parent `FormField.tsx` component. But since step 8 removes the additional rows, **the `isDirty` is true and the default value `reset` is not invoked.** 
    1. This causes the final number of rows to not be updated and remain as `schema.minRows` value.   

## Solution

The main cause seems to be due to `isDirty` being detected as true, causing the reset to not run and set the form values to the latest retrieved `defaultFormValues` . However, upon further analysis, this should indeed be true, since the `remove` was invoked to remove the rows, causing the value to differ from the `defaultValues` provided. 

This means that the `defaultValues` and the `values` are actually reflecting the correct latest values. Hence, there is no data incorrectness here. 

Hence, making a “fix” such that `isDirty` does not evaluate to false seems strange. (And it is strange that for the prev version, the `isDirty` is true). Since, `isDirty` is expected to be `true` given the current values defer from the set `defaultValue`. 

Hence, instead, the error seems that the `isMinRowsChanged` is set to true, which causes a removal of the table rows when useEffect (C) is run - which should not occur in the first place for public forms (since the form schema should always be unchanged during the public form submission page, we do not need to remove/add rows based on schema min row changes). 

So, what is useEffect (C) used for then? 

This useEffect is mainly used for the admin form builder, when the schema for the Table field is updated. This is documented by this PR https://github.com/opengovsg/FormSG/issues/3922 where this `useEffect` was [added to the existing `TableField.tsx` just to support the admin form builder](https://github.com/opengovsg/FormSG/pull/3922/files#diff-387e2b9cc993f540acd069e14b6b54c58ee94b13f0cdbb46a141ae44ca1c5720) and hence it does not need to run for public form submission/preview/use-template page! 

Hence, the proposed fix is to ignore `undefined` during `isMinRowsChanged`. 

**Why is ignoring `undefined` safe?** 

Case I: For public form submission page / preview page / use template page 

Quoting…

> This useEffect is mainly used for the admin form builder, when the schema for the Table field is updated. This is documented by this PR https://github.com/opengovsg/FormSG/issues/3922 where this `useEffect` was [added to the existing `TableField.tsx` just to support the admin form builder](https://github.com/opengovsg/FormSG/pull/3922/files#diff-387e2b9cc993f540acd069e14b6b54c58ee94b13f0cdbb46a141ae44ca1c5720) and hence it does not need to run for public form submission/preview/use-template page!
> 

The useEffect (C) is mainly intended to reflect updates in schema changes in the form builder page and not for the form submission page where the schema.minimumRows should never change. Thus, it should ideally be not be invoked for these pages. 

Ignoring the initial `undefined` in `hasMinRowsChanged` causes the useEffect(C) removal to not execute on the first render, which is intended. 

Furthermore, 

Schema.minimumRows is always asserted to be defined by TS. Hence, it should never be `undefined`. Hence, comparing with `undefined` is not required and causes an unnecessary useEffect (C) execution - which is causing the bug.  

Case II: For admin form template form builder page: 

Schema.minimumRows is always asserted to be defined by TS. Hence, the useEffect (C) will still invoke when changes to the form schema are made in the builder, achieving its intended effect.

**Breaking Changes** 

## To prevent regressions: 
To prevent regressions, the following chromatic stories are added: (link: https://www.chromatic.com/build?appId=60911f69d4403b0039b266a2&number=9334)

**Admin form page:**
- When the admin updates the min rows, the rows are updated accordingly.

**Public form page**
- storage mode:
- edited fields are saved in the saved draft
- floating button appears when save draft is enabled
- restores save draft (covers rating, email, nric fields)
- prefills higher precedence than save draft
- myinfo fields are not saved in the draft
- save draft not restored for schema changed fields
- when add more rows are added, the number of rows increase in the table field. disabled when max rows are met.

1st step mrf:
- restores save draft
- when add more rows are added, the number of rows increase in the table field. disabled when max rows are met.

2nd step mrf:
- prev submission restored with correct row count
- if non editable, prev submission not overriden by saved draft
- if editable, prev submission overriden by saved draft
- if restored default value < min rows, more rows are appended in default values to meet the min rows

**Template page and preview page:**
- table field is initialised with correct number of min rows

## Tests

TC1: Table field MRF works 
- [ ] Create an mrf form with table field with additional rows, make min rows = 3, max = 5
- [ ] Make it 4 steps, where table field is editable only on first step and third step 
- [ ] First step fill in the field and set it to 5 rows 
- [ ] Submit past step 2 
- [ ] For step 3, edit to 4 rows 
- [ ] Submit past step 4
- [ ] Assert the full workflow can go through and the correct results are displayed in the form results page after each step.

### DIFF
--- a/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
+++ b/frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
@@ -18,6 +18,7 @@ import {
 import {
   augmentFormFields,
   getFieldPrefillMap,
+  getInitialFormValues,
   useCommonFormProvider,
 } from '~/features/public-form/PublicFormProvider'
 
@@ -359,13 +360,31 @@ export const PreviewFormProvider = ({
     [formFields, currentStepNumberWorkflowStep],
   )
 
-  const defaultFormValues = useMemo(() => ({}), [])
+  const isSaveDraftEnabled = Boolean(form?.isSaveDraftEnabled)
+
+  const defaultFormValues = useMemo(() => {
+    if (!form?.responseMode) return {}
+    return getInitialFormValues({
+      formResponseMode: form?.responseMode,
+      currentStepNumberWorkflowStep,
+      augmentedFormFields,
+      fieldPrefillMap,
+      draftResponsesToRestore: {},
+      searchParams,
+      isSaveDraftEnabled,
+    })
+  }, [
+    form?.responseMode,
+    currentStepNumberWorkflowStep,
+    augmentedFormFields,
+    fieldPrefillMap,
+    isSaveDraftEnabled,
+    searchParams,
+  ])
 
   const formMethods = useForm<FormFieldValues>({
     defaultValues: defaultFormValues,
   })
-
-  const isSaveDraftEnabled = Boolean(form?.isSaveDraftEnabled)
 
   if (isNotFormId) {
     return <NotFoundErrorPage />
--- a/frontend/src/features/admin-form/template/TemplateFormProvider.tsx
+++ b/frontend/src/features/admin-form/template/TemplateFormProvider.tsx
@@ -14,6 +14,7 @@ import { PublicFormContext } from '~/features/public-form/PublicFormContext'
 import {
   augmentFormFields,
   getFieldPrefillMap,
+  getInitialFormValues,
   useCommonFormProvider,
 } from '~/features/public-form/PublicFormProvider'
 
@@ -120,13 +121,31 @@ export const TemplateFormProvider = ({
     [formFields, currentStepNumberWorkflowStep],
   )
 
-  const defaultFormValues = useMemo(() => ({}), [])
+  const isSaveDraftEnabled = Boolean(form?.isSaveDraftEnabled)
+
+  const defaultFormValues = useMemo(() => {
+    if (!form?.responseMode) return {}
+    return getInitialFormValues({
+      formResponseMode: form?.responseMode,
+      currentStepNumberWorkflowStep,
+      augmentedFormFields,
+      fieldPrefillMap,
+      draftResponsesToRestore: {},
+      searchParams,
+      isSaveDraftEnabled,
+    })
+  }, [
+    form?.responseMode,
+    currentStepNumberWorkflowStep,
+    augmentedFormFields,
+    fieldPrefillMap,
+    isSaveDraftEnabled,
+    searchParams,
+  ])
 
   const formMethods = useForm<FormFieldValues>({
     defaultValues: defaultFormValues,
   })
-
-  const isSaveDraftEnabled = Boolean(form?.isSaveDraftEnabled)
 
   if (isNotFormId) {
     return <NotFoundErrorPage />
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -313,15 +313,14 @@ export const getInitialFormValues = ({
         }
       }
 
-      // Append extra empty rows to existing value until the minimum is met. 
+      // Append extra empty rows to existing value until the minimum is met.
       if (field.fieldType === BasicField.Table && acc[field._id]) {
-        const currentValue = acc[field._id] as TableFieldValues 
+        const currentValue = acc[field._id] as TableFieldValues
         if (currentValue.length < (field.minimumRows || 0)) {
           acc[field._id] = [
             ...currentValue,
-            ...times(
-              (field.minimumRows || 0) - currentValue.length,
-              () => createTableRow(field),
+            ...times((field.minimumRows || 0) - currentValue.length, () =>
+              createTableRow(field),
             ),
           ]
         }
@@ -373,7 +372,7 @@ const getSaveDraftKey = ({
   isMrf: boolean
   /**
    * @note currentMrfWorkflowStepNumber is 0-indexed.
-   */ 
+   */
   currentMrfWorkflowStepNumber: number
 }) => {
   const FORMSG_SAVE_DRAFT_PREFIX = 'formsg-save-draft'
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -887,11 +887,13 @@ export const PublicFormProvider = ({
     mode: 'onTouched',
   })
 
-  // Reset default values when they change
   const {
     formState: { isDirty },
     reset,
   } = formMethods
+
+  // Reset cached default values when they change if
+  // the user has not yet changed the form values.
   useEffect(() => {
     if (!isDirty) {
       reset(defaultFormValues)
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -242,7 +242,7 @@ export const getFieldPrefillMap = (
   }, {} as PrefillMap)
 }
 
-const getInitialFormValues = ({
+export const getInitialFormValues = ({
   formResponseMode,
   previousSubmission,
   previousAttachments,
@@ -886,19 +886,6 @@ export const PublicFormProvider = ({
     defaultValues: defaultFormValues,
     mode: 'onTouched',
   })
-
-  const {
-    formState: { isDirty },
-    reset,
-  } = formMethods
-
-  // Reset cached default values when they change if
-  // the user has not yet changed the form values.
-  useEffect(() => {
-    if (!isDirty) {
-      reset(defaultFormValues)
-    }
-  }, [defaultFormValues, isDirty, reset])
 
   const onSaveDraft = () => {
     // Used to track save draft usage
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -47,7 +47,7 @@ import {
   HttpError,
   SingleSubmissionValidationError,
 } from '~services/ApiService'
-import { FormFieldValues } from '~templates/Field'
+import { FormFieldValues, TableFieldValues } from '~templates/Field'
 import { createTableRow } from '~templates/Field/Table/utils/createRow'
 
 import NotFoundErrorPage from '~pages/NotFoundError'
@@ -313,6 +313,20 @@ export const getInitialFormValues = ({
         }
       }
 
+      // Append extra empty rows to existing value until the minimum is met. 
+      if (field.fieldType === BasicField.Table && acc[field._id]) {
+        const currentValue = acc[field._id] as TableFieldValues 
+        if (currentValue.length < (field.minimumRows || 0)) {
+          acc[field._id] = [
+            ...currentValue,
+            ...times(
+              (field.minimumRows || 0) - currentValue.length,
+              () => createTableRow(field),
+            ),
+          ]
+        }
+      }
+
       if (!acc[field._id]) {
         // Required so table column fields will render due to useFieldArray usage.
         // See https://react-hook-form.com/api/usefieldarray
@@ -357,6 +371,9 @@ const getSaveDraftKey = ({
   formId: string
   formSubmissionId?: string
   isMrf: boolean
+  /**
+   * @note currentMrfWorkflowStepNumber is 0-indexed.
+   */ 
   currentMrfWorkflowStepNumber: number
 }) => {
   const FORMSG_SAVE_DRAFT_PREFIX = 'formsg-save-draft'
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -380,6 +380,8 @@ export const PublicFormProvider = ({
   const { t, i18n } = useTranslation()
   const selectedLanguage = i18n.language as Language
 
+  const isTest = import.meta.env.STORYBOOK_NODE_ENV === 'test'
+
   // Once form has been submitted, submission data will be set here.
   const [submissionData, setSubmissionData] = useState<SubmissionData>()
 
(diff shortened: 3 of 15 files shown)


kevin9foong:
todo: figure out how to write TC to test MRF prefills for public form, preview, template and admin form page (probably use chromatic)

frontend/src/templates/Field/Table/TableField.tsx
```diff
@@ -50,7 +50,11 @@ export const TableField = ({
   isHighContrast,
 }: TableFieldProps): JSX.Element => {
   const { i18n } = useTranslation()
-  const hasMinRowsChanged = useHasChanged(schema.minimumRows)
+  // RATIONALE for setting isIgnoreUndefined to true:
+  // Since schema.minimumRows is always defined, we should ignore undefined values.
+  // This prevents unwanted update of rows for public form page while
+  // still detecting minimum row schema changes for the admin form builder page.
+  const hasMinRowsChanged = useHasChanged(schema.minimumRows, true)
```
kevin9foong, line 57: Main fix is here

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -313,6 +313,19 @@ const getInitialFormValues = ({
         }
       }
 
+      // Append extra empty rows to existing value until the minimum is met.
+      if (field.fieldType === BasicField.Table && acc[field._id]) {
```
kevin9foong, line 317: Populate the table field with at least min rows, since this is no longer done in the TableField.tsx

frontend/src/features/admin-form/template/TemplateFormProvider.tsx
```diff
@@ -120,14 +121,32 @@ export const TemplateFormProvider = ({
     [formFields, currentStepNumberWorkflowStep],
   )
 
-  const defaultFormValues = useMemo(() => ({}), [])
+  const isSaveDraftEnabled = Boolean(form?.isSaveDraftEnabled)
```
kevin9foong, line 124: Populate with correct defaults instead of empty object

frontend/src/features/admin-form/preview/PreviewFormProvider.tsx
```diff
@@ -359,14 +360,32 @@ export const PreviewFormProvider = ({
     [formFields, currentStepNumberWorkflowStep],
   )
 
-  const defaultFormValues = useMemo(() => ({}), [])
+  const isSaveDraftEnabled = Boolean(form?.isSaveDraftEnabled)
+
+  const defaultFormValues = useMemo(() => {
```
kevin9foong, line 365: Populate with correct defaults instead of empty object

frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -563,10 +581,12 @@ export const PublicFormProvider = ({
     !previousSubmission &&
     !isSubmissionSecretKeyInvalid
   ) {
-    const isValid = isKeypairValid(
-      encryptedPreviousSubmission.submissionPublicKey,
-      submissionSecretKey,
-    )
+    // During test, we want to bypass the secret key validation
+    const isValid =
```
kevin9foong, line 585: For testing, we do not want to check key correctness for developer quality of life and ease of writing test cases (Due to its difficulty, TCs were not written for multi-step MRF forms previously)

frontend/src/features/public-form/utils/decryptSubmission.ts
```diff
@@ -30,6 +30,19 @@ export const decryptSubmission = ({
   if (!submission) throw Error('Encrypted submission undefined')
   if (!secretKey) throw Error('Secret key undefined')
 
+  // For testing, do not perform decryption and return the encrypted content directly
+  // (which will be a un-encrypted FieldResponsesV3 object)
+  const isTest = import.meta.env.STORYBOOK_NODE_ENV === 'test'
```
kevin9foong, line 35: For testing, we do not want to decrypt and have to pass in encrypted values for quality of life and ease of writing test cases - hence we just pass in a valid un-encrypted payload and skip decryption for tests (Due to its inconvenience/difficulty, TCs were not written for multi-step MRF forms previously)

frontend/src/hooks/useIndexedDb.ts
```diff
@@ -1,6 +1,8 @@
 import { useCallback, useEffect, useState } from 'react'
 
-const DB_NAME = 'FormSG'
+// For testing, use a different database name to avoid conflicts with the main database.
+const isTest = import.meta.env.STORYBOOK_NODE_ENV === 'test'
+const DB_NAME = isTest ? 'TestFormSG' : 'FormSG'
```
kevin9foong: For testing, use a different DB for safety. However, note that the indexed db is domain (eg, localhost) scoped so the risk of affecting the main db is low.

eliotlim (approved):
LGTM! 👍

PR #8782, opened 2025-10-02
feat: signatures v1.4 (webhooks)

scottheng96:
## Problem

Incremental development of signatures field. For v1.4, webhooks are settled - converting the `vectorArray` to a `Signature captured` string to prevent leaking of sensitive signature information

Closes FRM-2161

## Solution

Handling webhooks in the BE has changed in order to support this update of signature field value. Originally, encryption happens once, and `encryptedContent` is saved in DB and sent in the webhook. To support a different field value in webhooks, responses need to be manipulated before encryption, and then encrypted independently from db `encryptedContent`. The diagram below depicts these decoupling changes

image

Key changes:
1. Responses are updated to desired webhook values and encrypted as `encryptedWebhookContent`
2. `encryptedWebhookContent` is saved in req (`req.formsg.encryptedWebhookContent`)
3. `encryptedWebhookContent` replaced `encryptedContent into `submission` object just before firing webhook
*For MRF, webhook responses are encrypted using `submissionPublicKey` taken from the encryption of db responses which is done first. 

4. Change to include `encryptionSubmisisonSecretKey` in MRF webhook payload.

**Breaking Changes** 

## Tests

**Regression storage mode form webhook**
- [x] Create a storage mode form with 1-2 basic fields
- [x] Enable webhooks on the form (can generate a URL on webhook.site)
- [x] Successfully make a submission
- [x] Decrypt the webhook response and observe that encryptedResponse can be successfully decrypted

**Storage mode form + signature webhook**
- [x] Add the signature field to the recently created storage mode form
- [x] Successfully make a submission
- [x] Decrypt the webhook response and observe that encryptedResponse can be successfully decrypted
- [x] Observe that under signature response, `['Signature captured']` is shown instead of the signature vector array

**Storage mode form + signature optional webhook**
- [x] Update the signature field to be optional in the recently created storage mode form
- [x] Successfully make a submission
- [x] Decrypt the webhook response and observe that encryptedResponse can be successfully decrypted
- [x] Observe that under signature response, `[]` is shown

**MRF mode form + signature webhook**
- [x] Create a MRF mode form with 1-2 basic fields and a signature field
- [x] (optional) create 2 workflow steps and assign signature field to step 1, other fields to any
- [x] Successfully complete step 1 & subsequent step submissions
- [x] At each step, decrypt the webhook response (rmb to use cryptoV3.decrypt) and observe that encryptedResponse can be successfully decrypted
- [x] Observe that under signature response, `['Signature captured']` is shown instead of the signature vector array

### DIFF
--- a/shared/types/response-v3.ts
+++ b/shared/types/response-v3.ts
@@ -144,3 +144,7 @@ export type SignatureFieldResponseV3 = {
   type: 'draw'
   value: SignatureVectorArray
 }
+
+export type SignatureFieldWebhookResponseV3 = {
+  value: string[]
+}
--- a/shared/types/response-v3.ts
+++ b/shared/types/response-v3.ts
@@ -146,6 +146,6 @@ export type SignatureFieldResponseV3 = {
 }
 
 export type SignatureFieldWebhookResponseV3 = {
-  fieldType: BasicField
+  fieldType: BasicField.Signature
   answer: string[]
 }
--- a/shared/types/response-v3.ts
+++ b/shared/types/response-v3.ts
@@ -146,5 +146,6 @@ export type SignatureFieldResponseV3 = {
 }
 
 export type SignatureFieldWebhookResponseV3 = {
-  value: string[]
+  fieldType: BasicField
+  answer: string[]
 }
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -11,7 +11,6 @@ import {
   FormResponseMode,
   isPaymentsProducts,
 } from '../../../../../shared/types'
-import { SIGNATURE_CAPTURED_STRING } from '../../../../../shared/utils/signature'
 import { IPopulatedForm } from '../../../../types'
 import {
   EncryptAttachmentResponse,
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -533,7 +533,7 @@ export const encryptSubmission = async (
   )
 
   // Modify response data for webhook responses and encrypt separately
-  const strippedBodyResponsesWebhook = strippedBodyResponses.map((response) => {
+  const strippedResponsesWebhook = strippedBodyResponses.map((response) => {
     if (
       response.fieldType === BasicField.Signature &&
       response.answerArray.length > 0
@@ -546,7 +546,7 @@ export const encryptSubmission = async (
   })
 
   const encryptedWebhookContent = formsgSdk.crypto.encrypt(
-    strippedBodyResponsesWebhook,
+    strippedResponsesWebhook,
     publicKey,
   )
   req.formsg.encryptedWebhookContent = encryptedWebhookContent
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -11,6 +11,7 @@ import {
   FormResponseMode,
   isPaymentsProducts,
 } from '../../../../../shared/types'
+import { SIGNATURE_CAPTURED_STRING } from '../../../../../shared/utils/signature'
 import { IPopulatedForm } from '../../../../types'
 import {
   EncryptAttachmentResponse,
@@ -530,6 +531,25 @@ export const encryptSubmission = async (
     strippedBodyResponses,
     publicKey,
   )
+
+  // Modify response data for webhook responses and encrypt separately
+  const strippedBodyResponsesWebhook = strippedBodyResponses.map((response) => {
+    if (
+      response.fieldType === BasicField.Signature &&
+      response.answerArray.length > 0
+    )
+      return {
+        ...response,
+        answerArray: [SIGNATURE_CAPTURED_STRING],
+      }
+    return response
+  })
+
+  const encryptedWebhookContent = formsgSdk.crypto.encrypt(
+    strippedBodyResponsesWebhook,
+    publicKey,
+  )
+  req.formsg.encryptedWebhookContent = encryptedWebhookContent
 
   req.formsg.encryptedPayload = {
     attachments: encryptedAttachments,
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
@@ -58,6 +58,7 @@ import {
 import {
   formatMyInfoStorageResponseData,
   omitResponseKeys,
+  prepareWebhookResponseContent,
 } from './encrypt-submission.utils'
 import IncomingEncryptSubmission from './IncomingEncryptSubmission.class'
 
@@ -533,17 +534,9 @@ export const encryptSubmission = async (
   )
 
   // Modify response data for webhook responses and encrypt separately
-  const strippedResponsesWebhook = strippedBodyResponses.map((response) => {
-    if (
-      response.fieldType === BasicField.Signature &&
-      response.answerArray.length > 0
-    )
-      return {
-        ...response,
-        answerArray: [SIGNATURE_CAPTURED_STRING],
-      }
-    return response
-  })
+  const strippedResponsesWebhook = prepareWebhookResponseContent(
+    strippedBodyResponses,
+  )
 
   const encryptedWebhookContent = formsgSdk.crypto.encrypt(
     strippedResponsesWebhook,
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
@@ -152,6 +152,7 @@ export const performEncryptPostSubmissionActions = ({
   emailData,
   attachments,
   respondentEmails,
+  encryptedWebhookContent,
 }: {
   submission: IEncryptedSubmissionSchema
   responses: FieldResponse[]
@@ -159,6 +160,7 @@ export const performEncryptPostSubmissionActions = ({
   emailData?: SubmissionEmailObj
   attachments?: Mail.Attachment[]
   respondentEmails?: string[]
+  encryptedWebhookContent?: string
 }): ResultAsync<
   true,
   | FormNotFoundError
@@ -183,6 +185,19 @@ export const performEncryptPostSubmissionActions = ({
         // do not await on webhook
         const webhookUrl = form.webhook?.url
         if (!webhookUrl) return okAsync(form)
+
+        // replace encryptedContent with encryptedWebhookContent before firing
+        if (encryptedWebhookContent)
+          submission.encryptedContent = encryptedWebhookContent
+        else {
+          logger.error({
+            message:
+              'Error while sending webhook, no encryptedWebhookContent found',
+            meta: {
+              action: 'useEncryptedWebhookContent',
+            },
+          })
+        }
 
         return WebhookFactory.sendInitialWebhook(
           submission,
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
@@ -187,9 +187,14 @@ export const performEncryptPostSubmissionActions = ({
         if (!webhookUrl) return okAsync(form)
 
         // replace encryptedContent with encryptedWebhookContent before firing
-        if (encryptedWebhookContent)
+        if (encryptedWebhookContent) {
           submission.encryptedContent = encryptedWebhookContent
-        else {
+          return WebhookFactory.sendInitialWebhook(
+            submission,
+            webhookUrl,
+            !!form.webhook?.isRetryEnabled,
+          ).andThen(() => okAsync(form))
+        } else {
           logger.error({
             message:
               'Error while sending webhook, no encryptedWebhookContent found',
@@ -198,13 +203,8 @@ export const performEncryptPostSubmissionActions = ({
               action: 'useEncryptedWebhookContent',
             },
           })
+          return okAsync(form)
         }
-
-        return WebhookFactory.sendInitialWebhook(
-          submission,
-          webhookUrl,
-          !!form.webhook?.isRetryEnabled,
-        ).andThen(() => okAsync(form))
       })
       // TODO [PDF-LAMBDA-GENERATION]: Remove setting of Growthbook targetting once pdf generation rollout is complete
       .map(async (form) => {
--- a/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
+++ b/src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
@@ -194,6 +194,7 @@ export const performEncryptPostSubmissionActions = ({
             message:
               'Error while sending webhook, no encryptedWebhookContent found',
             meta: {
+              ...logMeta,
               action: 'useEncryptedWebhookContent',
             },
           })
--- a/src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
+++ b/src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
@@ -9,6 +9,7 @@ import {
   WorkflowType,
 } from 'shared/types'
 
+import { WebhookFactory } from 'src/app/modules/webhook/webhook.factory'
 import MailService from 'src/app/services/mail/mail.service'
 import {
   IMultirespondentSubmissionSchema,
@@ -27,7 +28,6 @@ import {
   sendNextStepReminderEmail,
 } from '../multirespondent-submission.service'
 import * as MultirespondentSubmissionService from '../multirespondent-submission.service'
-import { WebhookFactory } from 'src/app/modules/webhook/webhook.factory'
 
 jest.mock('src/app/modules/datadog/datadog.utils')
 
@@ -1645,7 +1645,7 @@ describe('multirespondent-submission.service', () => {
       )
     })
 
-    it('should call fire webhook with the correct arguments', async () => {
+    it('should fire webhook with the correct arguments for first submission', async () => {
       const encryptedWebhookContent = 'mockEncryptedWebhookContent'
       // Act
       await performMultiRespondentPostSubmissionCreateActions({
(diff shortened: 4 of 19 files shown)


kevin9foong:
praise: just wanted to say thanks for taking the time to illustrate and highlight the key changes - it really helps with gaining context for review

shared/types/response-v3.ts
```diff
@@ -144,3 +144,8 @@ export type SignatureFieldResponseV3 = {
   type: 'draw'
   value: SignatureVectorArray
 }
+
+export type SignatureFieldWebhookResponseV3 = {
+  fieldType: BasicField
```
kevin9foong: question: should this be `BasicField.Signature`?

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -184,6 +186,19 @@ export const performEncryptPostSubmissionActions = ({
         const webhookUrl = form.webhook?.url
         if (!webhookUrl) return okAsync(form)
 
+        // replace encryptedContent with encryptedWebhookContent before firing
+        if (encryptedWebhookContent)
+          submission.encryptedContent = encryptedWebhookContent
```
kevin9foong, line 191: nit: would it be safer to use braces for this?

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -184,6 +186,19 @@ export const performEncryptPostSubmissionActions = ({
...
+        else {
+          logger.error({
+            message:
+              'Error while sending webhook, no encryptedWebhookContent found',
+            meta: {
+              action: 'useEncryptedWebhookContent',
```
kevin9foong, line 203: suggestion (non-blocking): shall we include the submission id in the meta for future debugging as well? since it is passed in the log meta object above. 

also, understand there are some inconsistencies in this function not using `logMeta` defined earlier, but might want to consider using the ...logMeta which includes some useful debugging meta values

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
```diff
@@ -674,25 +676,36 @@ export const performMultiRespondentPostSubmissionCreateActions = ({
...
-        logger.error({
-          message: 'Multirespondent submission webhook error',
-          meta: logMeta,
-          error,
-        })
+    if (!encryptedWebhookContent) {
+      logger.error({
+        message:
+          'Error while sending webhook, no encryptedWebhookContent found',
+        meta: { action: 'useEncryptedWebhookContentMrf' },
```
kevin9foong: suggestion: should we include the other debugging params in the logmeta object as well for future debugging?

kevin9foong: also, i understand there are a few inconsistencies in the code - but it seems that `meta.action` values usually corresponds to the function name eg, the main `logMeta` object used in the function. perhaps we might want to adhere to this

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
```diff
@@ -341,3 +347,41 @@ export const getQuestionTitleAnswerString = ({
...
+  > = {}
+
+  for (const [key, value] of Object.entries(input)) {
+    if (
+      value.fieldType === BasicField.Signature &&
+      value.answer.value.length > 0
+    ) {
+      result[key] = {
+        ...value,
+        answer: [SIGNATURE_CAPTURED_STRING],
```
kevin9foong, line 379: question: not sure, just double checking 

realised the result used here is `answer: [SIGNATURE_CAPTURED_STRING]` but in encrypt submissions middleware it is `answerArray: [SIGNATURE_CAPTURED_STRING]`

wanted to understand why this difference was made - does this mean that storage mode and mrf would have different formats for signatures?

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
```diff
@@ -674,25 +676,36 @@ export const performMultiRespondentPostSubmissionCreateActions = ({
   }
 
   const webhookUrl = form.webhook?.url
+
   if (webhookUrl) {
     logger.info({
-      message: 'Sending initial webhook for multirespondent submission',
+      message: 'Sending update webhook for multirespondent submission',
```
kevin9foong: question: wondering why this might have been changed to `update`? am a little confused on what this change means since the else statement below invoked is called `sendInitialWebhook`

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
```diff
@@ -674,25 +676,36 @@ export const performMultiRespondentPostSubmissionCreateActions = ({
...
+    if (!encryptedWebhookContent) {
+      logger.error({
+        message:
+          'Error while sending webhook, no encryptedWebhookContent found',
+        meta: { action: 'useEncryptedWebhookContentMrf' },
       })
+    } else {
+      submission.encryptedContent = encryptedWebhookContent
+
+      WebhookFactory.sendInitialWebhook(
```
kevin9foong, line 695: question: should this function perhaps be renamed to reflect the latest changes introduced with the encryptedWebhookContent? this is related to the comment change to `update` above if the above is intended

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
kevin9foong, line 197: nit (non blocking): also, understand there are some inconsistencies in this function itself, but might want to consider using the `...logMeta`

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
```diff
@@ -531,6 +532,25 @@ export const encryptSubmission = async (
...
 
+  // Modify response data for webhook responses and encrypt separately
+  const strippedResponsesWebhook = strippedBodyResponses.map((response) => {
+    if (
+      response.fieldType === BasicField.Signature &&
+      response.answerArray.length > 0
+    )
+      return {
+        ...response,
+        answerArray: [SIGNATURE_CAPTURED_STRING],
```
kevin9foong: from the earlier comment: for example,`answerArray` is supposed to differ from `answer` used in [src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts](https://github.com/opengovsg/FormSG/pull/8782/files/1047b929f36af3bf79ef1aeb173277f4e6657088#diff-52450cc3a3b0e1cd301e8f5847e999f40f6998f8bac9cf8e180d38cac2569bed). this might cause mrf and storage mode to have differing keys for the same field type

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
kevin9foong: what is the difference between `initial` and `update`?

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
```diff
@@ -125,6 +126,9 @@ describe('multirespondent-submission.service', () => {
 
       // Act
       await performMultiRespondentPostSubmissionUpdateActions({
+        submission: {
```
kevin9foong, line 129: praise: thanks for adding this!

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
kevin9foong: question: in the same file, there is 
```      
questionAnswerPair.push({
          question: `[signature] ${questionTitle}`,
          answer: SIGNATURE_CAPTURED_STRING,
 })```
 
 where the `SIGNATURE_CAPTURED_STRING` is not in a `[ ]`, is this inconsistency expected?

File: src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts [hunk 10]
```diff
@@ -341,3 +347,41 @@ export const getQuestionTitleAnswerString = ({
...
+  const result: Record<
+    string,
+    | ParsedClearFormFieldResponseV3
+    | StrippedAttachmentResponseV3
+    | SignatureFieldWebhookResponseV3
+  > = {}
+
+  for (const [key, value] of Object.entries(input)) {
+    if (
+      value.fieldType === BasicField.Signature &&
```
kevin9foong, line 374: suggestion: this field type check and replacement of the answer seem to be duplicated also for the encrypt submission. 

wonder if its possible to use a common function for both since they are doing the same thing

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
```diff
@@ -341,3 +347,41 @@ export const getQuestionTitleAnswerString = ({
   }
   return questionAnswerPair
 }
+
+/**
+ * Converts responses to webhook desired outputs before encryption (again)
+ */
+export const prepareWebhookResponseContent = (
```
kevin9foong: suggestion: shall we add some unit test coverage for this function, specifically to assert that signatures are replaced correctly with the SIGNATURE_CAPTURED string - since it affects external output to users to prevent unintended regressions when future code changes are made. 

Also, would it make sense to create a common `prepareWebhookResponseContent` to be used across storage mode and MRF? 
- for mrf, the processing of the signature is abstracted by this function, but not for storage mode. 
- this causes duplication of the same logic across multiple files - making it possible to miss updating one mode when another changes. 

if v1 and v3 have different formats, perhaps we can do an if/else check for the mode version 
something like: 
- prepareWebhookContent(input, formResponseMode) { 
// which would handle the creation of the webhook content 
}

src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -796,6 +799,18 @@ export const encryptSubmission = async (
       req.body.version,
     )
 
+  // Modify response data for webhook responses and encrypt separately
+  const strippedResponsesWebhook = prepareWebhookResponseContent(
```
kevin9foong: praise: nice use of abstraction to keep the middleware simple to understand

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
```diff
@@ -1601,6 +1644,110 @@ describe('multirespondent-submission.service', () => {
         2,
       )
     })
+
+    it('should fire webhook with the correct arguments for first submission', async () => {
```
kevin9foong, line 1678: thanks for adding these tests!

kevin9foong:
**PR review summary:** 
Nice work with the PR - no major blockers apart from a few suggestions: 

1. Adding of more metadata in the logs apart from just the `meta.action`. This will help identify the specific submission and make future incident (when we need it) debugging easier. 
2. understanding why storage and mrf modes use different webhook encrypted content keys for signature field (`answerArray` vs `array`)  
3. suggestion to make a common `prepareWebhookResponseContent` function to abstract and reuse the same replace of value logic with `SIGNATURE_CAPTURED` string instead of duplicating across the 2 modes. Adding of unit test to cover the output and prevent regression in webhook response. 
4. minor TS and code style check

shared/types/response-v3.ts
scottheng96: Yep, I missed that. Thanks for this catch! Made the fix

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
scottheng96: Yes encrypt & mrf will have different keys for certain field types since for encrypt we support `answerArray` and for mrf only `answer` - I believe checkbox, table, and address (afai recall) will behave this way. This is something mrf webhook users will have to manage.

until we develop `answerObject` and standardize it

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
scottheng96: Good suggestion - no harm in more verbose logging. Included this!

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
scottheng96: Ah, `initial` is meant for first step submission, and `update` is for subsequent steps (following `createActions` and `updateActions`. Although after a second look it might be cleaner to remove these differentiations from the message since its already captured in the `meta.action`. Removed this

scottheng96: Responding to `sendInitialWebhook`, i think `initial` is meant for the first attempt at firing the webhook, then subsequently will be retries.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
scottheng96: I'll include `...logMeta`.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts
scottheng96: Responding here again, I think `initial` would reflect the first attempt at firing the webhook, and then subsequent attemps would be retries.

I'll keep it this way, and have `meta.action` reference which post-submission actions are being called

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
scottheng96: Thanks for this suggestion! I did start with trying to create a common function at first, but I then realised I would not only have to if/else based on the FormResponseMode, but also expand the return types to V1 & V3, which would be more complicated to handle in the middlewares.

Ended up deciding that dealing with that logic wasn't worth it, and having only `if (
      response.fieldType === BasicField.Signature &&
      response.answerArray.length > 0
    )` duplicated a reasonable tradeoff

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
scottheng96: Responded to this above ^ let me know if the explanation is reasonable

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
scottheng96: Yes, for encrypt and mrf responses we save data differently. encrypt uses `answer` and `answerArray` (for multi-response fields, like checkbox, table, address etc.), while for MRF we save them all as `answer` - I'm thinking it was due to simplifying the rehydration of the data for subsequent steps. 

This should be solved when we tackle consolidating responses into an `answerObject` that both modes can make use of

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
scottheng96: I'll add some test coverage for this!

kevin9foong:
hey @scottheng96, read through the responses and thanks for explaining the considerations and tradeoffs. 

suggestions for test cases: was wondering if we can add the following TCs to assert the newly implemented behavior before we merge the change? 

1. For storage mode middleware, ✅ 
- given an input (with signature), formsgSdk.crypto.encrypt is invoked with the correct args ✅ 
- if the signature exists, it is removed and the constant is passed. other fields are unaffected. ✅ 
- (negative case) if the signature does not exist, the constant is not added. ✅ 

2. For `prepareWebhookResponseContent`, assert that given the a response with a bunch of fields: ✅ 
- only the signature field is modified ✅ 
- the signature field is modified to include the `SIGNATURE_CAPTURED` string (✅ in [459729f](https://github.com/opengovsg/FormSG/pull/8782/commits/459729fa0b06ca94f5ae23e7f03309b050cfbbab))
- (negative case) if signature is empty, it is not modified and does not pass the `SIGNATURE_CAPTURED`.  ✅ 

3. For [multirespondent-submission.middleware.ts] ✅ (https://github.com/opengovsg/FormSG/pull/8782/files/1047b929f36af3bf79ef1aeb173277f4e6657088#diff-50982bbcd44ee0e226a3411bd283b304ef9286f52227940620ea01f555b1fef9) 
- the `prepareWebhookResponseContent` is invoked as expected and formsgSdk.crypto.encrypt is invoked with the return value of `prepareWebhookResponseContent`. ✅ 

4. For multirespondent-submission.service.ts, ✅ 
- `sendInitialWebhook` invoked as expected ✅  [Nice, already there] 
- testing for `encryptedSubmissionSecretKey` to be included in mrf webhook payload as well which is newly added (https://github.com/opengovsg/FormSG/pull/8782#discussion_r2405483543) ✅ 

Should be able to generate them using AI rather quickly. Lmk if youll like me to generate it for you as i have access to an AI IDE.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
kevin9foong: yeah - makes sense. thanks for explaining the considerations! think the duplicate code is a fair tradeoff too

src/app/modules/submission/encrypt-submission/encrypt-submission.middleware.ts
kevin9foong: thanks for explaining this!

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
```diff
@@ -1601,6 +1644,110 @@ describe('multirespondent-submission.service', () => {
...
+          encryptedSubmissionSecretKey: 'encryptedSubmissionSecretKey',
+        } as MultirespondentSubmissionDto,
+        logMeta: {} as any,
+        encryptedWebhookContent,
+      })
+
+      expect(webhookSpy).toHaveBeenCalledWith(
+        expect.objectContaining({
+          _id: mockSubmissionId,
+          encryptedContent: encryptedWebhookContent,
```
kevin9foong, line 1705: suggestion: should we also include `encryptedSubmissionSecretKey` in the webhook call assertion as well since it is newly added in this PR?

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.utils.spec.ts
```diff
@@ -703,4 +706,26 @@ describe('multirespondent-submission.utils', () => {
...
+          type: 'draw',
+          value: [[[10, 20, 0.5]], [[40, 40, 0.5]]],
+        },
+      },
+    }
+
+    const webhookResponses = prepareWebhookResponseContent(mockResponses)
+
+    expect(webhookResponses['mockId2'].answer).toEqual(
+      expect.arrayContaining([SIGNATURE_CAPTURED_STRING]),
```
kevin9foong, line 729: suggestion: would recommend adding an expect mockId1 is unchanged as well, as this would also assert only the signature field is modified

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
scottheng96: done

src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.utils.spec.ts
scottheng96: Feel like its a bit overkill since the function is a simple one that only has logic for `BasicField.Signature` - I'll add an unchanged check to be safe anyways, always good to be safe that sorry

scottheng96:
I have added TCs for the concerns listed in (1) - (4). One change I made was to abstract out the logic to update signature values in `encrypt-submission.middleware` into its own function (similar to mrf) and write TCs for that. Considering that it's a scoped value change I don't think it garners testing the entire middleware encrypt function. 

I also realised we have no middleware tests for encrypt, and little middleware tests for mrf, we could probably invest some time in these at a later date.

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
eliotlim: actually i do agree:
1. we could consume an additional 2 bytes (one for each brace)
2. but it would make this line much more resistant to syntactic changes later
  a. for example, if we were debugging the assignment and inserted a debug / logging statement
  b. or wanted to trace some aspect of the substitution process in a subsequent signature feature release

eliotlim (approved):
LGTM! 👍 

one minor nit on the braces thing.

src/app/modules/submission/encrypt-submission/encrypt-submission.service.ts
```diff
@@ -184,6 +186,20 @@ export const performEncryptPostSubmissionActions = ({
...
+            message:
+              'Error while sending webhook, no encryptedWebhookContent found',
+            meta: {
+              ...logMeta,
+              action: 'useEncryptedWebhookContent',
+            },
+          })
+        }
+
         return WebhookFactory.sendInitialWebhook(
```
kevin9foong: seems like this is being fired even when `encryptedWebhookContent` is not defined, which defers from mrf. Concerned as this might send undesirable payload to end user despite the logs saying webhook failed

kevin9foong:
Compiling notes on pre-release testing:

**To fix before release - as it is related to this PR**
Not firing webhook when encryptedWebhookContent does not exist - to prevent sending of unintended values to end user

**(Non blocking) To be discussed in subsequent PR/iteration - Already existing behavior (no regression)**
1. Retries don't work with MRF Webhooks right now (only the [first attempt will be sent for MRF](https://webhook.site/#!/view/ce0940f5-1376-4c70-ac8d-a09facd25d68/c66ee63c-d0fd-468b-9086-e5d8b74d096d/1)) from @eliotlim 

2. Differing behavior for optional fields in mrf / storage - to be aligned in subsequent PRs since it alr exists 
The optional field behavior defers for mrf and storage mode when the value is not provided. 
- storage mode = field included but empty value
- mrf = field not included in the encrypted payload

kevin9foong (approved):

PR #8794, opened 2025-10-06
fix: hide empty tooltip for main save draft button

kevin9foong:
## Problem

Save draft button shows an empty tooltip when no draft is saved. There should ideally be no popup. 

image

## Solution

Disable (hide) the tooltip if no label exists.  

**Breaking Changes** 


eliotlim (approved):
LGTM! 👍

PR #8798, opened 2025-10-07
fix: add missing response id and time submitted in local generated pdf

kevin9foong:
## Problem

In the local pdf print button - the response ID and timestamp are not included in the generated PDF, differing from the emailed PDF. 

Expected: 
image

Actual: 
image

## Solution

Include these values 

**Breaking Changes** 

## Tests

MRF: 
- [x] Click on local print for an mrf response. 
- [x] Assert the response id and submitted time is included. 
- [x] This time and resp id should match the time in the response in the web app.  

Storage mode: 
- [ ] Click on local print for a storage mode response. 
- [ ] Assert the response id and submitted time is included. 
- [ ] This time and resp id should match the time in the response in the web app.


eliotlim (approved):
did a quick verification, works like a charm.
image

PR #8799, opened 2025-10-07
feat: duplicate form workflow and override emails

eliotlim:
## Problem

Multi-respondent Forms (MRFs) can't be duplicated along with their workflows.
Workflow creation is tedious, time-consuming, and error-prone -- which prevents users from benefiting from this feature.

Builds upon #8764
Closes FRM-2164

## Solution

**Breaking Changes** 

**Improvements**:

- Duplicate the entire MRF including its Workflow
  - If the duplication is triggered from the dashboard by the Form Owner / Editor, the workflow's emails will be preserved.
  - If the form is duplicated as a shared template, the workflow's emails will be reset to the new owner's email.

## Screenshots

image

## Tests

**Setup: Create an MRF and a workflow**
- [ ] Create a new MRF in the Form Admin Workspace
- [ ] Add a workflow that contains three steps, and a dropdown field
- [ ] The second step should request approval from 'Specific email(s)': use a distinct email (e.g. `test@example.com`)
- [ ] The third step should request approval from 'Emails assigned to options in a dropdown field': (e.g. `test+option1@example.com`)

**TC1: Duplicate an MRF as a Form Owner / Editor from the Admin Workspace**
- [ ] Return to the Admin Workspace, click the dropdown menu on the right of the newly-created Form, and click Duplicate
- [ ] Set up the duplicate as an MRF
- [ ] Visit the 'Workflow' tab in the left sidebar, and inspect the copied workflow
- [ ] The 'Specific email(s)' field should contain the original emails
- [ ] The CSV attached under 'Emails assigned to options in a dropdown field' should contain the original emails

**TC2: Create a new MRF from a template using a shared `/use-template` link.
- [ ] Click on Share > Template > Copy the shareable link to `/use-template`
- [ ] Set up the new form created from the template as an MRF
- [ ] Visit the 'Workflow' tab in the left sidebar, and inspect the copied workflow
- [ ] The 'Specific email(s)' field should contain your email
- [ ] The CSV attached under 'Emails assigned to options in a dropdown field' should only contain your email

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
@@ -3412,7 +3412,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { workspaceId: undefined },
+        { workspaceId: undefined, overrideEmails: undefined },
       )
     })
 
@@ -3464,7 +3464,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { workspaceId: mockWorkspaceId },
+        { workspaceId: mockWorkspaceId, overrideEmails: undefined },
       )
     })
 
@@ -3534,7 +3534,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { workspaceId: undefined },
+        { workspaceId: undefined, overrideEmails: undefined },
       )
     })
 
@@ -3670,7 +3670,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { workspaceId: undefined },
+        { workspaceId: undefined, overrideEmails: undefined },
       )
     })
   })
--- a/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
@@ -3900,7 +3900,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { duplicateStripped: true },
+        { overrideEmails: ['andanother@example.com'] },
       )
     })
 
@@ -3972,7 +3972,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { duplicateStripped: true },
+        { overrideEmails: ['andanother@example.com'] },
       )
     })
 
@@ -4122,7 +4122,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { duplicateStripped: true },
+        { overrideEmails: ['andanother@example.com'] },
       )
     })
   })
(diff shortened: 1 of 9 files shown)


src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -657,29 +657,32 @@ export const processCreateFormInWorkspace = async (
   return form
 }
 
+type DuplicateFormOpts = {
+  // The ID of the workspace to duplicate the form into.
+  workspaceId?: string
+  // If true, overrides the emails of MRF workflows and dropdown fields with the given overrideEmails
```
kevin9foong: nit (non blocking): should the comment be, If `defined` instead of `true`?

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -692,13 +695,45 @@ export const duplicateForm = (
     overrideProps.endPage = omit(originalForm.endPage, 'buttonLink')
   }
 
+  // If the form is an MRF, copy the workflow as well
```
kevin9foong: suggestion: before merging this in, wonder if we could include some unit tests for the service to cover this new duplication behavior in `admin-form.service` as well? 

in particular, these 3 TCs: 
1. For static workflow step, replaces with overrideEmails if provided, otherwise use existing values. 
2. For conditional, dynamic workflow step types: uses existing values regardless of if the overrideEmails exist 
3. For dropdown field, if overrideEmails is provided, will replace with the overrideEmails. Otherwise, use existing values. 

These would give us confidence that mrf workflow duplication will not regress when making future changes!

src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
```diff
@@ -3900,7 +3900,7 @@ describe('admin-form.controller', () => {
         MOCK_FORM,
         MOCK_USER_ID,
         expectedParams,
-        { duplicateStripped: true },
+        { overrideEmails: ['andanother@example.com'] },
```
kevin9foong, line 3903: praise: 
nice work covering that if use template is used - it will provide the `overrideEmails` from the user (form owner) 👍 

the test cases in duplicateAdminForm also cover that for admin duplication, no `overrideEmail` is passed in!

Together with the 3 suggested TCs below, this gives us confidence that as long as the CI passes, both TC1 and TC2 should be covered and there will be no regressions in the future

kevin9foong (commented):
hey @eliotlim, thanks for introducing this change - believe this will solve one of our admin's key issues when sharing forms and having to figure out the workflow step previously. 

Mainly looks good to merge! 

Just some suggestions on adding of a couple test cases for future regression confidence - reason is that before release, we may be able to test it manually and assert it works, but over time, we may not be able to consistently cover the same cases, hence adding some of these TCs will give us peace of mind that this change will continue to work now and in the future and make our FormSG code super robust!

src/app/modules/form/admin-form/admin-form.service.ts
eliotlim: Great point! good catch :)

```suggestion
  // If defined, overrides the emails of MRF workflows and dropdown fields with the given overrideEmails
```

src/app/modules/form/admin-form/admin-form.service.ts
kevin9foong: issue: perhaps the `workflow` should be duplicated at the `form.server.model` level instead to maintain consistency and for easier maintenance.

kevin9foong (approved):

kevin9foong (approved):

PR #8803, opened 2025-10-08
feat: add metrics error counting for webhook missing encrypted webhook content

kevin9foong:
## Problem

When webhooks are not sent due to missing encrypted webhook content, a log is fired but we are not able to trace it using the error counting mechanism. 

## Solution

Generate an error which sends the metric to datadog for observability.  

After this, we will be able to see issues with webhooks due to missing encrypted content in the dashboard. 
image

**Breaking Changes** 


eliotlim (approved):
LGTM 👍

eliotlim:
Thanks @kevin9foong for instrumenting encryptedWebhookContent usage so quickly!

We've decided to redesign the signatures payload, as opposed to the webhook sending mechanism. (#8805)
1. this will help us better the security properties of the signatures field
2. and will reduce the complexity of the webhook codepaths / dataflow.

Closing this PR because we're deprecating this codepath

PR #8814, opened 2025-10-14
feat: override workflow duplication with admin email if viewer

kevin9foong:
## Problem

For admins with collaborator access, they should not see the workflow email details when the form is duplicated in the admin panel. 

## Solution

Check if the duplicating admin has permissions to view the workflow details (ie, form write access). 
If admin has permissions, no not pass `overrideEmails`, otherwise, pass in `overrideEmails`. 

**Breaking Changes** 

## Tests

TC1: Viewer admin duplication. 
- [ ] Create a MRF form with static, dynamic and conditional routing workflow steps. 
- [ ] Share this MRF form to another admin with viewer access. 
- [ ] Duplicate the form. 
- [ ] Assert the workflow emails (conditional routing map, static emails) are replaced with the duplicating admin's email only. 

TC2: Editor admin duplication. 
- [ ] Create a MRF form with static, dynamic and conditional routing workflow steps. 
- [ ] Share this MRF form to another admin with editor access. 
- [ ] Duplicate the form. 
- [ ] Assert the workflow emails (conditional routing map, static emails) are original and NOT replaced. 

TC3: Form owner admin duplication 
- [ ] Repeat TC2 as form owner

## Deploy Notes
The following cases have been covered via BE unit tests - assuming that WRITE and above = Editor / Owner. Non-WRITE + READ = Viewer

### DIFF
--- a/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
+++ b/src/app/modules/form/admin-form/__tests__/admin-form.controller.spec.ts
@@ -55,6 +55,7 @@ import {
   IPopulatedEmailForm,
   IPopulatedEncryptedForm,
   IPopulatedForm,
+  IPopulatedMultirespondentForm,
   IPopulatedUser,
   IUserSchema,
   PublicForm,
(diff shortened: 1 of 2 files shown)


tutntut (approved):
LGTM

PR #8818, opened 2025-10-16
fix: myinfo child not reflecting correct vaccination status 

kevin9foong:
## Problem
Myinfo child not reflecting correct vaccination status as it only checks if the 1M3D is present and does not check the fulfilled boolean. 
This resolves inc-726-10_16_2025-myinfo_children_fields_possibly_returning_invalid_vaccinations. 

## Solution

Check for the fulfilled boolean and add a new NOT_FULFILLED status. 
If missing data, mark as NOT_FULFILLED.  

NOTE (Missing data is treated as not fulfilled):
All children should have the vaccination requirement 1M3D.
The official National Immunisation Registry (NIR), synced with MyInfo, must be used to mark the 1M3D requirement as fulfilled.
Hence, if the vaccination requirement is not present in MyInfo, we enforce the not fulfilled status by default.
This is instead of unknown returned previously before this PR change, which allowed the respondent to edit the value.

**Breaking Changes** 

## Tests

TC1: Myinfo child works 
- [x] Create a form with myinfo child field and vaccination status subfield. 
- [x] Respond as a parent with child. 
- [x] Assert the vaccination fulfilled subfield is prefilled correctly.


tutntut (approved):
LGTM

PR #8821, opened 2025-10-16
feat: email redact

kevin9foong:
## Solution

Split the `getPublicView` to each response mode type to simplify the typing and for simpler customization. 

Utilise TS to safeguard against misuse in the FE and enforce redaction. 

image

For example, the TS prevents `optionsToRecipientsMap` from being used in the FE public form provider. This prevents undesired bugs as the `optionsToRecipientsMap` will not be present. 

**Breaking Changes** 

### DIFF
--- a/src/app/models/form.server.model.ts
+++ b/src/app/models/form.server.model.ts
@@ -1011,6 +1011,61 @@ const compileFormModel = (db: Mongoose): IFormModel => {
       return formSettings
     }
 
+  EmailFormSchema.method<IFormDocument>(
+    'getPublicView',
+    function (): PublicForm {
+      const emailFormPublicViewFields = pick(
+        this,
+        EMAIL_PUBLIC_FORM_FIELDS,
+      ) as PublicForm
+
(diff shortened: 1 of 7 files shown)


src/app/models/form.server.model.ts
```diff
@@ -1006,35 +1011,86 @@ const compileFormModel = (db: Mongoose): IFormModel => {
...
+  )
+
+  EncryptedFormSchema.method<IEncryptedFormDocument>(
+    'getPublicView',
+    function (): PublicForm {
+      const encryptFormPublicViewFields = pick(
+        this.toObject(),
+        STORAGE_PUBLIC_FORM_FIELDS,
+      ) as PublicStorageFormDto
+      // NOTE: While encrypt mode forms do not allow adding optionsToRecipient mapping for dropdown fields,
```
eliotlim, line 1049: praise: well done for considering this non-obvious edge case.

eliotlim (approved):

PR #8825, opened 2025-10-17
feat: submission email redact

kevin9foong:
## Solution

Strip the emails when the multi respondent dto is created. 
Create a new public dto for public routes. 
Use TS for type safety. 

**Breaking Changes** 

## Tests

TC1: Automated tests have been written, but manually check 
- [x] Mrf submission network call does not expose optionsToRecipients and emails in the workflow.  

TC2: Anti regression: 
Include all possible fields in MRF form and all 3 step routing types.  
- [x] Mrf can decrypt responses
- [x] Can submit multiple step MRF with various types and routing works. 
- [x] Can send MRF reminder


tutntut (approved):
approved

PR #8828, opened 2025-10-21
feat: update save draft toast and admin toggle copy

kevin9foong:
## Problem

After respondents use save draft, they may be uncertain on how to restore and continue filling the form as no instructions are provided. 

## Solution

Provide explicit instructions in the save draft toast so respondents know how to return to their drafts. Also, update the admin settings page save draft toggle copy. 

**Breaking Changes** 


tutntut (commented):
image

Also update the admin settings copy. Thanks.

Copy change on admin settings:

"Respondents can save what they’ve filled in and continue later on the same device and browser"

tutntut (approved):
LGTM

PR #8835, opened 2025-10-22
feat: update pdf gen deployment to new acc bucket

kevin9foong:
## Problem

The pdf gen bucket to host SAM deployments has changed, leading to failing builds. 

## Solution

Update to the new bucket. 

**Breaking Changes** 


tutntut (approved):
lgtm

PR #8842, opened 2025-10-27
fix: radio others bug v2

scottheng96:
## Problem

**Context**
MRF submissions with preceding submitted data containing a radio "others" field is causing subsequent steps to fail because of BE validation error. This is because during rehydration of radio input data, value is not being rehydrated correctly
Correct value:  {value: 'OTHERS', othersInput: 'custom text'}
Current value: {value: '', othersInput: 'custom text'}  :arrow_left: This fails BE validation due to ''
This was fixed before [here](https://github.com/opengovsg/FormSG/pull/8699), but is back since we made some changes regarding handling defaultValues and this might be been missed out during that implementation.

## Solution

1. Re-implementation of https://github.com/opengovsg/FormSG/pull/8699
2. Regression unit tests have also been included

## Tests
**Field shown on logic based on radio field 'others' option**

- [ ] Create an MRF form, Include a radio field with 'others' enabled, and a short text field
- [ ] Add logic that IF 'others' is selected, SHOW short text field
- [ ] Add a 2 step workflow assigning radio field to step 1, short text field to step 2
- [ ] Make the form public and make a step 1 submission with 'others' selected
- [ ] Attempt to make a step 2 submission, observe that the short text field is visible consistently when being interacted on (doesn't hide away)
- [ ] Successfully submit step 2

### DIFF
--- a/frontend/src/features/form/utils/__tests__/extractMrfPreviousStepResponseValue.spec.ts
+++ b/frontend/src/features/form/utils/__tests__/extractMrfPreviousStepResponseValue.spec.ts
@@ -1,0 +1,95 @@
+import { CLIENT_RADIO_OTHERS_INPUT_VALUE } from '~shared/constants'
+import { CountryRegion } from '~shared/constants/countryRegion'
+import {
+  AttachmentFieldResponseV3,
+  BasicField,
+  FieldResponseV3,
+  FormFieldDto,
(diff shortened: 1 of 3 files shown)


KenLSM (approved):
LGTM! Thanks for adding additional test cases to prevent regression!

PR #8854, opened 2025-11-03
feat: respondent copy v3 (phase 1)

scottheng96:
## Problem 

- Enhancements to allow respondents to get a copy of their responses after submitting the form.
- Support signature workflows where a copy of the responses is essential
- Iterate from respondent copy v2 pilot

Note: Respondent copy v3 is split into 2 phases for incremental releases. Please refer to PRD to understand scopes of phase 1/2 delivery

Closes FRM-2171

## Solution

- [x]  Include helper text for respondent
    - [x]  Assigned state should include helper text
    - [x]  Unassigned state should not have helper text
- [x]  Default sender, subject and message displayed for user
- [x]  Move “Include a PDF response” above Subject
    - [x]  Replace “Include PDF response” with a different copy
- [x]  Move “Enable email confirmation” above OTP verification

**Breaking Changes** 

## Before & After Screenshots

| --- | --- | --- |
| Helper text & replace behaviour | image | error_replace |
| Moving toggle up in email field builder + copy changes to default placeholders | image | image |

## Tests

**Helper text shows/hides as expected**
- [x] Create a storage mode form and add an email field to the form
- [x] In the email field settings, toggle on **email confirmation** and **Include a copy of their responses**
- [x] Observe that a helper text shows up in the form builder
- [x] In the form builder, attempt to enter an invalid email, observe that the error message replaces the helper text

### DIFF
--- a/shared/constants/mail.ts
+++ b/shared/constants/mail.ts
@@ -1,0 +1,5 @@
+export const DEFAULT_RESPONDENT_COPY_EMAIL = {
+  subject: 'Thank you for submitting {formTitle}',
+  content:
+    'To whom it may concern,\n\nThank you for submitting this form.\n\nRegards,\n{agencyName}',
+}
--- a/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditEmail/EditEmail.tsx
+++ b/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditEmail/EditEmail.tsx
@@ -187,7 +187,7 @@ export const EditEmail = ({ field }: EditEmailProps): JSX.Element => {
           )}
         />
       </FormControl>
-            <Box>
+      <Box>
         <FormControl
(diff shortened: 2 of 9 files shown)


shared/constants/mail.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_RESPONDENT_COPY_EMAIL = {
+  subject: 'Thank you for submitting {formTitle}',
+  content:
+    'To whom it may concern,\n\nThank you for submitting this form.\n\nRegards,\n{agencyName}',
+}
```
scottheng96, line 5: This is a bit ugly, because the same texts are supplied in i18n en-sg (sender.placeholder etc.) Open to suggestions on a cleaner implementation! 🙏 
Issue i ran into was that I can't cleanly import en-sg text into mail.service.ts class constructor

eliotlim: 😅 this is [imported by the backend](https://github.com/opengovsg/FormSG/pull/8854/files), so it makes sense to not introduce react-i18n as a dependency in there.

not ideal, but we can clean this up another time.

eliotlim (approved):
LGTM!

PR #8858, opened 2025-11-04
feat: pinning pdf to generate from lambda

KenLSM:
## Problem

We want to stop local PDF generation.

## Solution

Fixing `generatePdfFromHtml` to always generating from lambda.

Note: Explicit ring fencing changes small and purposefully located to be in one file instead of removing the whole lambda generation.

**Breaking Changes** 


eliotlim (approved):
LGTM! 👍

PR #8865, opened 2025-11-05
fix: singpass myinfo box texts

scottheng96:
## Problem

Singpass Myinfo box recent changes had removed the display of 'hasMyInfo' text when there is myInfo fields. Admins may get confused when they find themselves unable to switch authentication modes

image

## Solution

Decouple displaying of 'hasMyInfo' text in infobox from 'isFormPublic' text. They should be independent.

**Breaking Changes** 


KenLSM (approved):
Lgtm!

PR #8868, opened 2025-11-06
fix(copy): update MyInfo fields banner text

bookwormsuf:
## Problem
MyInfo banner copy was not updated with the latest collapse of Singpass options

## Solution
Replace `MyInfo` with `Singpass` in banner


KenLSM (approved):
LGTM

PR #8900, opened 2025-11-14
fix(copy): fixed copy for toast message on form open

bookwormsuf:
## Problem
Copy for toast message when admin opens a form is not up to date - this displays incorrect toast messaging for multi respondent forms. This is since email mode has since been deprecated and thus for mrf and storage mode, the responses are stored on FormSG and there is no longer a risk of losing submissions due to mailbox being too full. 

## Solution
Changed copy to remove the warning about potential loss of submission responses when inbox is full. 

**Breaking Changes** 

### DIFF
--- a/frontend/src/features/admin-form/settings/mutations.ts
+++ b/frontend/src/features/admin-form/settings/mutations.ts
@@ -131,7 +131,7 @@ export const useMutateFormSettings = () => {
         const toastStatusPublicMessage =
           newData.responseMode === FormResponseMode.Encrypt
             ? `Your form is now open.\n\nStore your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`
-            : `Your form is now open.\n\nIf you expect a large number of responses,  [AutoArchive your mailbox](${GUIDE_PREVENT_EMAIL_BOUNCE}) to avoid losing any of them.`
+            : `Your form is now open.`
         const toastStatusClosedMessage = 'Your form is closed to new responses.'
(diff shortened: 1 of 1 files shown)


kevin9foong (approved):

frontend/src/features/admin-form/settings/mutations.ts
```diff
@@ -131,7 +131,7 @@ export const useMutateFormSettings = () => {
         const toastStatusPublicMessage =
           newData.responseMode === FormResponseMode.Encrypt
             ? `Your form is now open.\n\nStore your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`
-            : `Your form is now open.\n\nIf you expect a large number of responses,  [AutoArchive your mailbox](${GUIDE_PREVENT_EMAIL_BOUNCE}) to avoid losing any of them.`
+            : 'Your form is now open.'
```
eliotlim, line 134: Thanks for contributing this suggestion! Quick thought:

- if the FormResponseMode is Encrypt (Storage): no change to the text is expected
- if the FormResponseMode is Multirespondent (MRF): we still use a secret key, so we should probably be displaying the same message used for Storage mode!
- and we've removed FormResponseMode.Email as an option in practice, so that's sorted.

```suggestion
        const toastStatusPublicMessage =
          `Your form is now open.\n\nStore your secret key in a safe place. If you lose your secret key, all your responses will be lost permanently.`
```

PR #8901, opened 2025-11-14
feat: remove un-neeeded dd tracing and browser installation in the setup

kevin9foong:
## Problem

There is unneeded duplication of playwright browser installation and dd instrumentation in the setup step - this is not required since no test results are piped from the setup (which is the intention of this, see: https://github.com/opengovsg/FormSG/issues/7542) and no tests are run. 

## Solution

Remove these duplications. This change is verified to reduce the time for setup phase by ~1 minute (from 4mins -> 3mins) 

**Breaking Changes** 

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -33,17 +33,9 @@ jobs:
           node-version: 22
       - name: Install dependencies
         run: npm ci
-      - name: Install Playwright Browsers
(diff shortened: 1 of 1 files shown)


scottheng96 (approved):

eliotlim (approved):
LGTM! 👍 

Thanks for the improvement.

PR #8916, opened 2025-11-25
fix: myinfo children missing prefix

scottheng96:
## Problem

Children field responses are not prefixed with `[MyInfo]` tag unlike other myinfo fields.

This is because we add the prefix in the middleware by cross checking if the field._id is inside myinfo hashed fields (this gets saved to our DB when they first login to Singpass [here](https://github.com/opengovsg/FormSG/blob/76cf7bdb45c8db2c5d867ef234cdf8195fc25b7d/src/app/modules/myinfo/myinfo.service.ts#L252)).

Children field hashes have names tied to the hash id, while our response id doesn't:
hash id: `childrenbirthrecords.692411846d1da146b503aada.childname.0.Tan Wen Jie`
field id: `"childrenbirthrecords.692411846d1da146b503aada.childname.0`

response id doesn't have the child name because if there are multiple children, respondent could select children in a different order, so we don't tie the field id to the child's name.

Closes FRM-2179

## Solution

Inside function that prefixes question names, `GetMyInfoPrefix`, handle this difference.

**Breaking Changes** 

## Before & After Screenshots

| Description | BEFORE | AFTER |
| --- | --- | --- |
| [MyInfo] prefixes (IndividualResponsePage) | image | image |
| [MyInfo] prefixes (emails) | image | image |

 

## Tests

**MyInfo prefixes exist on children field outputs**
- [x] Create a storage mode form, enable Singpass on the form and add a children field to the form
- [x] (optional) Add another MyInfo field for comparison
- [x] Add an email field to the form and enable `Email confirmations`, with responses included
- [x] Under email settings, add your email to `Notifications of new responses`
- [x] Successfully make a submission, using `F1612364K` (it has children) and your email on staging singpass
- [x] Observe that on the IndividualResponsePage of the admin portal, [MyInfo] prefix exists
- [x] Observe that on received emails (email confirmation + new submission notificaiton),  [MyInfo] prefix exists


eliotlim (approved):
LGTM! 👍

PR #8919, opened 2025-11-25
fix: hidden signature field blocks submission

scottheng96:
## Problem

When a (required) signature field is hidden by logic, submission of the form is being blocked. This is due to 'required' validation error preventing successful submission

Closes FRM-2180

## Solution

Unregister signature field component when it is being unmounted:
- unmounting is tracked within the component in `useEffect`, logic can be composed in the `return`
- Since signature field is an uncontrolled component, unmounting it from the DOM does not change the fact that it is still registered in RHF, so manual unregistering is required.

This isn't needed for other fields since they are controlled components, and unregistering upon unmounting is handled in [useEditLogicBlock](https://github.com/opengovsg/FormSG/blob/c97a4f53e21b29c35faad9e53f02a66e88393bac/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditLogicBlock.tsx#L39) `shouldUnregister: true`

**Breaking Changes** 

## Tests

**Successfully submit a form with a hidden signature field**
- [x] Create a single respondent form with a signature field and a Yes/No field
- [x] Add logic to hide signature field when 'Yes' is selected
- [x] Attempt to make a successful submission

### DIFF
--- a/frontend/src/templates/Field/Signature/SignatureField.tsx
+++ b/frontend/src/templates/Field/Signature/SignatureField.tsx
@@ -32,7 +32,7 @@ export const SignatureField = ({
   isHighContrast,
 }: SignatureFieldProps): JSX.Element => {
   const formContext = useFormContext<SignatureFieldInput>()
-  const { getValues, setValue } = formContext
+  const { register, unregister, setValue } = formContext
   const { isSubmitting, isValid, errors } = useFormState<SignatureFieldInput>()
   const [showSignaturePlaceholder, setShowSignaturePlaceholder] = useState(true)
 
@@ -55,8 +55,13 @@ export const SignatureField = ({
   )
 
   useEffect(() => {
-    formContext.register(schema._id, signatureValidationRules)
-  }, [formContext, schema._id, signatureValidationRules])
+    register(schema._id, signatureValidationRules)
+
+    // when unmounting, unregister the field so that submission will not be blocked by signature validation errors
+    return () => {
+      unregister(schema._id)
+    }
+  }, [register, schema._id, signatureValidationRules, unregister])
 
   // perfect freehand variables
   const pfCanvasRef = useRef<HTMLCanvasElement>(null)


frontend/src/templates/Field/Signature/SignatureField.tsx
```diff
@@ -55,8 +55,13 @@ export const SignatureField = ({
   )
 
   useEffect(() => {
-    formContext.register(schema._id, signatureValidationRules)
-  }, [formContext, schema._id, signatureValidationRules])
+    register(schema._id, signatureValidationRules)
+
+    // when unmounting, unregister the field so that submission will not be blocked by signature validation errors
+    return () => {
+      unregister(schema._id)
```
kevin9foong, line 62: question: have not looked into detail but wondering how the other fields marked as required handle this issue - do they also perform the unregistration? 

asking this as i'm a little cautious as to whether there is an alternative way that is standard across all fields and to avoid field specific behavior as much as possible

kevin9foong: also wondering - why do we make signature field an uncontrolled element? My understanding is that signature field is utilizing an external controlled component by the library - and React hook form has an explicit mechanism for handling such components (`Controller` can be used for this purpose source: https://react-hook-form.com/docs/usecontroller/controller) which could avoid this issue and need for explicit registration/dereg

scottheng96: _1. How to other fields marked as required handle this issue?_
(reposting the reason as per the PR description) This isn't needed for other fields since they are controlled components, and unregistering upon unmounting is handled in [useEditLogicBlock](https://github.com/opengovsg/FormSG/blob/c97a4f53e21b29c35faad9e53f02a66e88393bac/frontend/src/features/admin-form/create/logic/components/LogicContent/EditLogicBlock/EditLogicBlock.tsx#L39) shouldUnregister: true

_2. Why do we make signature field an uncontrolled element?_
I tried  using `<Controller>` early on which did work as well. IIRC my considerations were mainly the frequency of re-rendering and how it would affect performance.
- Because `<Controller>` forces React to re-render the field whenever the field value changes. So for high-frequency updates of internal states (like a rapidly updating canvas drawing), I found the component to re-render **as** the stroke was drawn.
- With the uncontrolled component, I tried implementation such that re-rendering would only occur upon completion of a stroke, so essentially batching updates to avoid unnecessary re-rendering.

Hope this is a helpful explanation!

eliotlim: this was a good constructive discussion with great points raised! 👍

eliotlim (approved):

eliotlim (approved):
LGTM! 👍

PR #8921, opened 2025-11-26
feat(robots): add noindex on page router to avoid reliance on thirdparty

KenLSM:
## Problem

We have a rule setup on Cloudflare to prevent robots from indexing high-entropy routes for auth access. To reduce the reliance on Cloudflare we want to move the rule dependency to our app.

## Solution

Add a `noindex` meta tag on specific routes of concern.

Notes: the meta tag should be present even if the route is invalid so that all routes to `edit/*` are not indexed.

**Breaking Changes** 


KenLSM:
No index is respected 😄 

Screenshot 2025-11-26 at 3 25 51 PM

kevin9foong (approved):

PR #8933, opened 2025-12-02
feat: reserve cpu for formsg app

kevin9foong:
## Problem

In task definition, the cpu is split between dd-agent and formsg-app, causing overprovisioning (only 3% utilisation) for dd agent and underprovisioning (15%-60% spike) for formsg-app. 

## Solution

Hence, explicitly reserve 87.5% of cpu units for formsg-app. This should reduce latency. 

**Breaking Changes** 


scottheng96 (approved):
LGTM!

PR #8946, opened 2025-12-05
feat: partially clean up deprecated email mode submission functions and route

kevin9foong:
## Problem

Email mode is deprecated and hence these routes are no longer used. 
However, the code remains which require updates when type defs change and tcs continue to run leading to additional test run time. 

Closes FRM-2207

## Solution

Code removal: Remove deprecated code so that it does not have to be updated and test cases for deprecated functions.
Code addition: Update singpass tests written for email mode only to test against storage mode to maintain same level of test coverage.   

## Tests

TC1: (Sanity check) MRF mode still works 
- [x] Create an MRF with attachment, signature and address fields with 3 steps. 
- [x] Complete the submission and assert it can be submitted. 

Storage mode is covered by playwright e2e tests.


eliotlim (approved):
praise: great work creating a single PR for each intentional change, which has helped tremendously with reviewing and testing this deprecation.

I've also double-checked the traffic for the deprecated routes, let's go ahead with this.

LGTM! 👍

PR #8947, opened 2025-12-05
chore: remove growthbook from payment functions

kevin9foong:
## Problem

Since pdf local is no longer in use and pdf lambda is used, the growthbook flag is no longer in use. Hence, growthbook no longer needs to be passed to payments functions.  

Closes FRM-2208

## Solution

Remove passing of growthbook to payments functions. 

**Breaking Changes** 


KenLSM (approved):

eliotlim (approved):
thanks for deprecating unused feature flags and mechanisms!
good stuff 💯

PR #8948, opened 2025-12-05
feat: encapsulate data pdf formatting

kevin9foong:
## Problem

Pdf rendering requires transformation of data to a specific format, which is duplicated across mrf and storage mode. This duplication can be removed. 

Closes FRM-2217

## Solution

Encapsulate the pdf rendering to within the pdf generation function itself, reducing the chances of inconsistencies in pdf data format across form modes.  

**Breaking Changes** 

## Tests

TC1: PDF copy generated correctly for MRF with signature field
- [x] Create an MRF with signature field and a bunch of other fields (table, short text etc) 
- [x] Add an email field with respondent copy PDF. 
- [x] Submit the form. 
- [x] Assert that the respondent email is received with correct PDF details (note the date format, form name, submission id and fields and their answers are correct) 

TC1: PDF copy generated correctly for Storage mode with signature field
- [x] Create an storage mode with signature field and a bunch of other fields (table, short text etc) 
- [x] Add an email field with respondent copy PDF. 
- [x] Submit the form. 
- [x] Assert that the respondent email is received with correct PDF details (note the date format, form name, submission id and fields and their answers are correct)


src/app/services/mail/mail.utils.ts
```diff
@@ -111,8 +115,43 @@ export const generateBounceNotificationHtml = (
...
+  formTitle,
+  submissionDateTime,
+  responsesData,
+  formUrl,
+}: AutoReplyData): AutoreplySummaryRenderData => {
+  const pdfResponsesData = responsesData.map(
+    ({ question, answerTemplate, answer, fieldType }) => ({
+      question,
+      answerTemplate,
+      // In the submit-form-summary-pdf.server.view.html, for signature field, it expects the answer key, while for others, it expects the answerTemplate key.
```
scottheng96, line 137: Kudos to comments that explain 'why' instead of 'what'!

src/app/services/mail/mail.service.ts
```diff
@@ -817,14 +806,21 @@ export class MailService {
       encryptFormDef.payments_channel.channel !== PaymentChannel.Unconnected &&
       encryptFormDef.payments_field.enabled === true
 
+    const autoReplyData: AutoReplyData = {
+      refNo: submission.id,
+      formTitle: form.title,
+      submissionDateTime: submission.created || new Date(),
```
scottheng96, line 812: Can i ask about when does the submission object doesn't have a .created variable and a new Date() is used to generated submissionDateTime?

eliotlim: oh yeah, @kevin9foong and I were discussing this the other day.

It turns out that `submission.created` can be `undefined` -- and in several other places within the codebase, a new date is generated when `submission.created` is `undefined`. Here are some [examples](https://github.com/search?q=repo%3Aopengovsg%2FFormSG+%22%7C%7C+new+Date%28%29%22&type=code):
- MRF [Submission](https://github.com/opengovsg/FormSG/blob/a2412fd927d2db60ded48d1cbd0c19025fa4819c/src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts#L131) and [Update](https://github.com/opengovsg/FormSG/blob/a2412fd927d2db60ded48d1cbd0c19025fa4819c/src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts#L223)
- During Storage Mode [Submission](https://github.com/opengovsg/FormSG/blob/a2412fd927d2db60ded48d1cbd0c19025fa4819c/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts#L692) and [Notification by Email](https://github.com/opengovsg/FormSG/blob/a2412fd927d2db60ded48d1cbd0c19025fa4819c/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts#L780)

It's the scope of this [last reference](https://github.com/opengovsg/FormSG/blob/a2412fd927d2db60ded48d1cbd0c19025fa4819c/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts#L780) in which Storage mode emails are being sent that this code is being hoisted -- and therefore it would be a risky change in semantics to modify the behaviour whilst hoisting it.

(also, the behaviour of the [moment-based line being removed](https://github.com/opengovsg/FormSG/pull/8948/commits/e1d68d2dea9aa0255bfa6b05a80f6994e3822016#diff-72ca1cef9efa278dbdd3b3bf2a9a589cfda6e6cf43b16ee75b7cbd2dc22054ddL804) behaves exactly the same way)

We should really be tightening the schema to prevent `submission.created` from ever being `undefined` -- but I think it's best that we don't introduce additional complexity into this PR.

eliotlim: good on you for noticing this, though! 👏

eliotlim (approved):
LGTM! 👍 

I've run through all the test cases on UAT, all good.

PR #8949, opened 2025-12-05
feat: add admin pdf copy for mrf

kevin9foong:
## Problem

Signatures do not provide utility by themselves and need to be signed against content to be of utility. Hence, a pdf copy should be included for admins so the signature can be verified against the form fields submitted.  

A feature flag is included to protect against excessive pdf generation load. 
There are 3 flags: ['ON', 'OFF', 'SIGNATURES_ONLY'] to control when pdfs should be generated. 

Updated PR description to trigger update.

Closes FRM-2218.

## Solution

Generate and include pdf in admin mrf workflow completion. 
To reduce load, the same pdf is reused for any potential respondent pdf copy. Hence, hoisting to the post submission actions of the check for if generation is needed and generation is done. 

**Breaking Changes** 

## Tests

TC1: PDF is not included in mrf workflow completion email when flag is disabled    
- [x] Set GB flag to 'OFF' 
- [x] Create an MRF form without signature field and with email notifications for workflow completion. 
- [x] Create 3 step workflow, where 2nd step is approval step. 
- [x] Submit the form and reject at step 2. Assert the workflow completion email is received and does not contain PDF. 
- [x] Submit the form at step 3 (approved). Assert the workflow completion email is received and does not contain PDF. 

TC2: PDF is included in mrf workflow completion email when flag is enabled    
- [x] Set GB flag to 'ON' 
- [x] Create an MRF form without signature field and with email notifications for workflow completion. 
- [x] Create 3 step workflow, where 2nd step is approval step. 
- [x] Submit the form and reject at step 2. Assert the workflow completion email is received and contains PDF. 
- [x] Submit the form at step 3 (approved). Assert the workflow completion email is received and contains PDF. 

TC3: PDF is included in mrf workflow completion email when flag is signatures only    
- [x] Set GB flag to 'SIGNATURES_ONLY' 
- [x] Create an MRF form without signature field and with email notifications for workflow completion. 
- [x] Create 1 step only workflow. 
- [x] Add an email field with respondent copy + pdf to the MRF step 1. 
- [x] Submit the form. Assert the workflow completion email is received and does not contain PDF. However, assert the respondent copy email is received but with PDF. 
- [x] Add a signature field to the MRF. 
- [x] Submit the form. Assert the workflow completion email is received and contains PDF. Assert the respondent copy email is received and also contains PDF.

### DIFF
--- a/shared/constants/feature-flags.ts
+++ b/shared/constants/feature-flags.ts
@@ -28,7 +28,14 @@ export const featureFlags = {
   singpassMrf: 'singpass-mrf' as const,
   enableSaveDraftButtonFloating: 'enable-save-draft-button-floating' as const,
   enableSaveDraftButtonHeader: 'enable-save-draft-button-header' as const,
+  adminEmailPdf: 'admin-email-pdf' as const,
   ogpHeader: 'enable-ogp-header' as const,
   ogpAwareness: 'ogp-awareness' as const,
   ogpSpinner: 'ogp-spinner' as const,
+}
+
+export enum AdminEmailPdfFeatureValue {
+  OFF = 'OFF',
+  SIGNATURES_ONLY = 'SIGNATURES_ONLY',
+  ON = 'ON',
 }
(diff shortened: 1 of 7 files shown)


kevin9foong:
Some additional notes about the beta flag:
1. for SIGNATURES_ONLY flag: As long as there is a signature field in the form - even if its not filled (eg, workflow is rejected before going to the step), a pdf will still be included in workflow complete email (this should be easier to explain to admin users)

2 Does not affect whether pdf will be included respondent copy email

eliotlim (approved):
Let's do this 👍

PR #8950, opened 2025-12-05
feat: add admin pdf copy for storage mode

kevin9foong:
## Problem

Signatures do not provide utility by themselves and need to be signed against content to be of utility. Hence, a pdf copy should be included for admins so the signature can be verified against the form fields submitted.  

A feature flag is included to protect against excessive pdf generation load. 
There are 3 flags: ['ON', 'OFF', 'SIGNATURES_ONLY'] to control when pdfs should be generated. 

Closes FRM-2219

## Solution

Generate and include pdf in admin storage workflow completion. 
To reduce load, the same pdf is reused for any potential respondent pdf copy. Hence, hoisting to the post submission actions of the check for if generation is needed and generation is done. 

**Breaking Changes** 

## Tests

TC1: PDF is not included in storage admin email when flag is disabled    
- [x] Set GB flag to 'OFF' 
- [x] Create a storage form without signature field and with admin email notif.
- [x] Submit the form. Assert the admin email is received and does not contain PDF. 
- [x] Submit the form. Assert the admin email is received and does not contain PDF. 

TC2: PDF is included in storage admin email when flag is enabled    
- [x] Set GB flag to 'ON' 
- [x] Create a storage form without signature field and with admin email notif.
- [x] Submit the form. Assert the admin email is received and contains PDF. 

TC3: PDF is included in storage admin email when flag is signatures only   
- [x] Set GB flag to 'SIGNATURES_ONLY' 
- [x] Create a storage form without signature field and with admin email notif.
- [x] Add an email mode with respondent copy to the form. 
- [x] Submit the form. Assert the admin email is received and does not contain PDF. However, assert the respondent copy email is received and contains PDF. 
- [x] Add a signature field to the storage form. 
- [x] Submit the form. Assert the admin email is received and contains PDF. Assert the respondent copy email is received and contains PDF also.

### DIFF
--- a/src/app/modules/payments/payments.service.ts
+++ b/src/app/modules/payments/payments.service.ts
@@ -242,6 +242,7 @@ export const performPaymentPostSubmissionActions = (
                 performEncryptPostSubmissionActions({
                   submission,
                   responses: payment.responses,
+                  emailFields: [], // TODO [EMAIL-CONFIRMATION-BUG]: Email confirmation email to email fields does not work for payment forms, this is an existing issue to be fixed.
                 })
                   .andThen(() =>
                     // If successfully sent email confirmations, delete response data from payment document.
--- a/shared/constants/feature-flags.ts
+++ b/shared/constants/feature-flags.ts
@@ -28,7 +28,14 @@ export const featureFlags = {
   singpassMrf: 'singpass-mrf' as const,
   enableSaveDraftButtonFloating: 'enable-save-draft-button-floating' as const,
   enableSaveDraftButtonHeader: 'enable-save-draft-button-header' as const,
+  adminEmailPdf: 'admin-email-pdf' as const,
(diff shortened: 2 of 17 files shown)


src/app/modules/payments/payments.service.ts
```diff
@@ -242,6 +242,7 @@ export const performPaymentPostSubmissionActions = (
                 performEncryptPostSubmissionActions({
                   submission,
                   responses: payment.responses,
+                  emailFields: [], // TODO [EMAIL-CONFIRMATION-BUG]: Email confirmation email to email fields does not work for payment forms, this is an existing issue to be fixed.
```
eliotlim, line 245: agree that we can resolve this later.
- it's not a regression
- and payment forms send the users a confirmation through a different route

eliotlim (approved):
LGTM 👍 let's do this.

PR #8959, opened 2025-12-09
feat: replace admin results local pdf library with bulk download compatible library

kevin9foong:
## Problem

- The previous react-to-print implementation does not work efficiently for bulk pdf downloads, as it is limited by the print window. 
- The react-to-print implementation does not work on healthcare devices. 

## Solution

Change to a js blob generation and download solution by changing the library and not relying on the use of the print functionality. This solution has been verified to work on menlosecurity and is pending verification on healthcare devices.  

**Breaking Changes** 

## Tests

TC1: PDF generation works for storage mode
- [ ] Make a storage mode form with all fields (especially table, signatures, address and attachment field types) and a variety of singpass myinfo fields
- [ ] Submit the form (include malay and tamil and chinese in your responses as much as possible) 
- [ ] Download the pdf from the results page for the submission
- [ ] Assert it generates correctly. 

TC2: PDF generation works for mrf mode
- [ ] Make a mrf mode form with all fields (especially table, signatures, address and attachment field types) with 2 steps
- [ ] Submit step 1 of the form (include malay and tamil and chinese in your responses as much as possible) 
- [ ] Download the pdf from the results page for the partial submission. 
- [ ] Assert it generates correctly. 
- [ ] Submit step 2 of the form. 
- [ ] Download the pdf from the results page for the submission.
- [ ] Assert it generates correctly.


kevin9foong:
Swapping library implementation in favor for one that supports multi-lang, at the expense of being non-vectorizable (taking a screenshot) and text not being copy-able. 

But this tradeoff is made since admins usually use pdfs for data retention, audit purposes.

PR #8961, opened 2025-12-09
feat: move signature vector to data uri to shared

kevin9foong:
## Problem

Signature vector to uri is a useful function that needs to be used in the FE as well. 

## Solution

Moving signature vector to uri to shared so it can be used by the FE without duplication. 

**Breaking Changes** 

## Tests

TC1: Signature field PDF still works for respondent copy MRF 
- [ ] Create MRF with signature field and email field with respondent copy pdf. 
- [ ] Submit the MRF and ensure the respondent email gets email with PDF and has the signature field correctly rendered in the PDF.  

TC2: Signature field PDF still works for storage mode respondent copy 
- [ ] Create storage mode with signature field and email field with respondent copy pdf. 
- [ ] Submit the storage mode form and ensure the respondent email gets email with PDF and has the signature field correctly rendered in the PDF.


kevin9foong:
Closed as it is more time consuming than expected to resolve and figure out the shared dependency resolution - differing to next time as non-blocking and non-critical for now

PR #8974, opened 2025-12-11
feat: update signature clear button to clear variant

scottheng96:
## Problem

The clear button is abit too prominent and is quite out of place with other fields that looks cleaner

Closes FRM-2213

## Solution

Change button variant from 'outline' to 'clear' as per design consensus

**Breaking Changes** 

## Before & After Screenshots

| --- | --- | --- |
| Signature clear button| image| image|


tutntut:
@bookwormsuf I wonder if the clear text should be in a disabled state if there is no input in the signature field yet or only appear if there is input in the field.

kevin9foong (approved):

PR #8979, opened 2025-12-12
feat: add template_form_id to form level metadata

scottheng96:
## Problem

When forms are duplicated using `use-template`, there is no information on which parent form is was templated from. This makes forms duplicated from our template experiment difficult to track.

## Solution

Add `template_form_id` to form metadata, so as to be able to know which parent form it was duplicated form for metrics tracking.

**Breaking Changes** 

## Before & After Screenshots

**AFTER**:
Duplicated form from `use-template`
image

## Tests

**Verify `template_form_id` exists**
- [ ] With an existing form, visit the link to use its template: `Share` -> `Use template`
- [ ] On the template preview page, click `Use this template`
- [ ] Successfully duplicate the form
- [ ] Search in the DB for the newly duplicated form id, observe that there is `metadata.template_form_id` collected

### DIFF
--- a/src/app/modules/form/admin-form/admin-form.service.ts
+++ b/src/app/modules/form/admin-form/admin-form.service.ts
@@ -661,6 +661,8 @@ type DuplicateFormOpts = {
   workspaceId?: string
   // If defined, overrides the emails of MRF workflows and dropdown fields with the given overrideEmails
   overrideEmails?: string[]
+  // If duplicated from a use-template
+  isFromTemplate?: boolean
 }
 
 /**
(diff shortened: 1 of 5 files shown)


kevin9foong (approved):

src/app/modules/form/admin-form/admin-form.service.ts
```diff
@@ -720,6 +722,13 @@ export const duplicateForm = (
     })
   }
 
+  // if created from template, track original form id in metadata
+  if (isFromTemplate) {
+    duplicateParams.metadata = {
```
kevin9foong, line 727: consideration: only concerns about possible override of previous from metadata - but since metadata should be refreshed for each new form, no concerns here

kevin9foong (approved):

src/app/modules/form/admin-form/admin-form.service.ts
eliotlim: yup agreed -- if we wanted to make form metadata extensible in the future:
1. we should consider spreading the existing form metadata into the duplicated metadata
2. and then overriding the template_form_id attribute as required.

kevin9foong (approved):

PR #8982, opened 2025-12-12
feat(bulk-local-pdf): replace pdf generation with html2pdf (part 0, anytime before part 8)

kevin9foong:
## Problem

The current pdf generation, using the browser's print, is not compatible with bulk downloads. Furthermore, some devices block usage of this print functionality by web apps due to security reasons. 

## Solution

Replace the usage of browser's print with save functionality using html2pdf. 

## Tradeoffs 
Known tradeoffs are: 
- since a screenshot is made of react component to generate the pdf, text is not copy-able - this is the same behavior as the previous browser print mechanism. From talking to users, pdf are used mainly for data retention purposes and hence this feature is not critical. However, we will continue to note if this problem is a blocker to our OKRs before deciding on whether to invest more effort into this. 
- there is another solution, https://github.com/opengovsg/FormSG/pull/8959, where text can be copied, but does not support chinese/tamil glyphs without having the user download large font packages (furthermore, each ttf file has max 65k glyphs and hence some special packaging of chinese + tamil glyphs is needed since pdfmake uses pdfkit dependency which does not support fallback). hence, considering the tradeoffs, this current PR's approach may be sufficient for our OKRs. 

**Breaking Changes** 

## Tests 
Note: there is no change to the component being rendered, which should be covered by chromatic stories against regression. 

TC1: PDF generation works for storage mode
- [ ] Make a storage mode form with all fields (especially table, signatures, address and attachment field types) and a variety of singpass myinfo fields
- [ ] Submit the form (include malay and tamil and chinese in your responses as much as possible) 
- [ ] Download the pdf from the results page for the submission
- [ ] Assert it generates correctly. 

TC2: PDF generation works for mrf mode
- [ ] Make a mrf mode form with all fields (especially table, signatures, address and attachment field types) with 2 steps
- [ ] Submit step 1 of the form (include malay and tamil and chinese in your responses as much as possible) 
- [ ] Download the pdf from the results page for the partial submission. 
- [ ] Assert it generates correctly. 
- [ ] Submit step 2 of the form. 
- [ ] Download the pdf from the results page for the submission.
- [ ] Assert it generates correctly.

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponseNavbar.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponseNavbar.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef } from 'react'
+import { useCallback, useEffect, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import { BiChevronLeft, BiChevronRight, BiLeftArrowAlt } from 'react-icons/bi'
 import { FaRegFilePdf } from 'react-icons/fa6'
@@ -8,7 +8,6 @@ import {
   useNavigate,
   useParams,
 } from 'react-router-dom'
-import { useReactToPrint } from 'react-to-print'
 import {
   Box,
   ButtonGroup,
@@ -33,7 +32,7 @@ import { useUser } from '~features/user/queries'
 
 import { useUnlockedResponses } from '../ResponsesPage/storage/UnlockedResponses/UnlockedResponsesProvider'
 
-import PrintableResponse from './PrintableResponse'
+import generateResponsePdf from './utils/generateResponsePdf'
 import { useIndividualSubmission } from './queries'
 
 export const IndividualResponseNavbar = (): JSX.Element => {
@@ -56,7 +55,9 @@ export const IndividualResponseNavbar = (): JSX.Element => {
     isAnyFetching,
   } = useUnlockedResponses()
   const { data: form, isLoading: isFormLoading } = useAdminForm()
-  const { isLoading } = useIndividualSubmission()
+  const { data: submission, isLoading: isSubmissionLoading } =
+    useIndividualSubmission()
+  const isLoading = isFormLoading || isSubmissionLoading
 
   const nextSubmissionId = useMemo(
     () => getNextSubmissionId(submissionId),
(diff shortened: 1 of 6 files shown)


frontend/src/features/admin-form/responses/IndividualResponsePage/utils/generateResponsePdf.tsx
```diff
@@ -0,0 +1,55 @@
...
+  root.render(
+    <PrintableResponse
+      formTitle={form.title}
+      formId={form._id}
+      decryptedResponses={submission.responses}
+      responseId={submission.refNo}
+      submissionTime={submission.submissionTime}
+    />,
+  )
+  const pdfTitle = `${form ? `${form.title}_formId_${form._id}_` : ''}submissionId_${submission.refNo}_response.pdf`
```
scottheng96: Is this conditional inclusion of form title + id based on a scenario where a PDF will generated without this information?

I'm seeing that a `form` object with { title, _id }` must be supplied when calling the function, so we can either (1) do away with this conditional (2) make the form object as an optional argument in the function?

kevin9foong: good catch - let me have a further look!

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponseNavbar.tsx
```diff
@@ -186,11 +181,12 @@ export const IndividualResponseNavbar = (): JSX.Element => {
                         },
                       },
                     )
-                    reactToPrintFn()
+                    if (submission && form) {
+                      await generateResponsePdf({ form, submission })
+                    }
```
scottheng96, line 186: nit: any added value in adding the datadog log inside this if statement instead of at the same top level? 

I guess the only difference is if we are tracking user clicks vs. actual pdf generation calls - hard to see a scenario that submission + form is consistently missing

frontend/src/features/admin-form/responses/IndividualResponsePage/utils/generateResponsePdf.tsx
kevin9foong: fixed! thanks for highlighting this 🌮

scottheng96:
LGTM! Just 1-2 non-critical comments. 

A final note, I realised we dont yet have outcome logs of the pdf generation - I think previously it was a low-risk experiment (literally 'save as' printing on client-side) so we didn't invest time into it. Something to KIV, but I'll defer to those with more context if logs are already handled in the subsequent PRs

Good stuff @kevin9foong!

frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponseNavbar.tsx
kevin9foong: heya @scottheng96 , thanks for flagging this out for consideration. 

after thinking about why this logger was added: im inclined to leave this at the top level as this logger main intention is to track user behavior and to see if admins are clicking on the local pdf download button (to see if they are interested in the functionality). 

hence, it should be likely be logged regardless of whether the submission or form is defined and top level might be more appropriate.

scottheng96 (approved):

PR #8986, opened 2025-12-15
feat(bulk-local-pdf): abstract out decryption function (part 1)

kevin9foong:
## Problem

Currently, the `decryptIntoCsv` function handles the decryption, creation of csv rows and downloading of attachments. This makes it challenging to reuse the same decryption function for other purposes (eg, bulk pdf generation, charts, better dashboards, truly parallel local decryptions)  

Also, the decryption function uses exceptions for flow control, which makes the code hard to understand. this should be avoided, see: https://nus-cs2030s.github.io/2021-s2/19-exception.html#do-not-use-exception-as-a-control-flow-mechanism, https://medium.com/@samanwayghatak/exception-as-control-flow-anti-pattern-e3b46b079cdd

Closes FRM-2223

## Solution

Abstract the decryption to its own reusable function.  

**Breaking Changes** 

## Tests

Pre-req: 

formatted.  

formatted.  

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
@@ -70,7 +70,6 @@ async function decryptSubmissionData({
       isSubmissionDecryptionSuccessful: true
     }
   | {
-      errorMessage: string
       isSubmissionDecryptionSuccessful: false
     }
 > {
@@ -86,10 +85,8 @@ async function decryptSubmissionData({
         version,
       })
       if (!decryptedObject) {
-        const errorMessage = `Invalid decryption for storage mode response`
-        console.error(errorMessage)
+        console.error('Invalid decryption for storage mode response')
         return {
-          errorMessage,
           isSubmissionDecryptionSuccessful: false,
         }
       }
@@ -105,10 +102,8 @@ async function decryptSubmissionData({
         version,
       })
       if (!decryptedObject) {
-        const errorMessage = `Invalid decryption for multirespondent response`
-        console.error(errorMessage)
+        console.error('Invalid decryption for multirespondent response')
         return {
-          errorMessage,
           isSubmissionDecryptionSuccessful: false,
         }
       }
@@ -122,10 +117,8 @@ async function decryptSubmissionData({
     default: {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       const _: never = submissionData
-      const errorMessage = `Invalid submission type encountered.`
-      console.error(errorMessage)
+      console.error('Invalid submission type encountered.')
       return {
-        errorMessage,
         isSubmissionDecryptionSuccessful: false,
       }
     }
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
@@ -16,6 +16,10 @@ import {
 } from '../types'
 import { CsvRecord } from '../utils/CsvRecord.class'
 import { downloadAndDecryptAttachmentsAsZip } from '../utils/downloadAndDecryptAttachment'
+import {
+  processDecryptedContent,
+  processDecryptedContentV3,
+} from '../utils/processDecryptedContent'
 
 const queue = new PQueue({ concurrency: 1 })
 
@@ -59,14 +63,17 @@ async function decryptSubmissionData({
 }: {
   submissionData: SubmissionStreamDto
   secretKey: string
-}): Promise<{
-  decryptedResponses: FormField[]
-  mrfSubmissionSecretKey?: string
-}> {
-  const { processDecryptedContent, processDecryptedContentV3 } = await import(
-    '../utils/processDecryptedContent'
-  )
-
+}): Promise<
+  | {
+      decryptedResponses: FormField[]
+      mrfSubmissionSecretKey?: string
+      isSubmissionDecryptionSuccessful: true
+    }
+  | {
+      errorMessage: string
+      isSubmissionDecryptionSuccessful: false
+    }
+> {
   const { encryptedContent, verifiedContent, version, submissionType } =
     submissionData
 
@@ -79,7 +86,12 @@ async function decryptSubmissionData({
         version,
       })
       if (!decryptedObject) {
-        throw new Error('Invalid decryption for storage mode response')
+        const errorMessage = `Invalid decryption for storage mode response`
+        console.error(errorMessage)
+        return {
+          errorMessage,
+          isSubmissionDecryptionSuccessful: false,
+        }
       }
       decryptedResponses = processDecryptedContent(decryptedObject)
       break
@@ -93,7 +105,12 @@ async function decryptSubmissionData({
         version,
       })
       if (!decryptedObject) {
-        throw new Error('Invalid decryption for multirespondent response')
+        const errorMessage = `Invalid decryption for multirespondent response`
+        console.error(errorMessage)
+        return {
+          errorMessage,
+          isSubmissionDecryptionSuccessful: false,
+        }
       }
       mrfSubmissionSecretKey = decryptedObject.submissionSecretKey
       decryptedResponses = await processDecryptedContentV3(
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -102,114 +183,86 @@ async function decryptIntoCsv(
...
-            verifiedContent: submission.verifiedContent,
-            version: submission.version,
-          })
-          if (!decryptedObject) {
-            throw new Error('Invalid decryption for storage mode response')
-          }
-          decryptedSubmission = processDecryptedContent(decryptedObject)
-          break
+
+    const decryptSubmissionDataResult = await decryptSubmissionData({
```
scottheng96, line 175: For an invalid decryption, an `errorMessage` + `isSubmissionDecryptionSuccessful = false` will be returned; I don't see the `errorMessage` being logged or thrown (see line 115). Was this removed intentionally?

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -49,48 +57,121 @@ function verifySignature(
...
+  let csvRecord: CsvRecord
 
   try {
-    // Validate that the submission is of a valid shape.
     const submission = SubmissionStreamDto.parse(JSON.parse(line))
+    const {
+      _id: submissionId,
+      created: submissionCreated,
+      submissionType,
+    } = submission
```
eliotlim: nit(style): instead of adding these five lines in a destructuring operation:
- we should just reference them directly as positional arguments in the new CsvRecord constructor
- it doesn't save any lines of code or improve readability.

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -49,48 +57,121 @@ function verifySignature(
...
-  )
-
-  const { SubmissionStreamDto, SubmissionType } = await import('~shared/types')
-
   const { line, secretKey, downloadAttachments, formId, hostOrigin } = data
-
-  let csvRecord: CsvRecord
   const attachmentDownloadUrls: AttachmentsDownloadMap = new Map()
   let downloadBlob: Blob
+  let csvRecord: CsvRecord
```
eliotlim, line 149: nit(review): is this ordering change necessary?

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -102,114 +183,86 @@ async function decryptIntoCsv(
...
+      // Logic to determine which key to use to decrypt attachments.
+      const attachmentDecryptionKey =
+        // If no submission secret key present, it is a storage mode form. So, use form secret key.
+        !mrfSubmissionSecretKey
+          ? secretKey
+          : // It's an mrf, but old version
+            submissionType === SubmissionType.Multirespondent &&
+              !submission.mrfVersion
+            ? secretKey
+            : mrfSubmissionSecretKey
```
eliotlim, line 210: nit(readability): this nested ternary could be expressed more clearly

```suggestion
      let attachmentDecryptionKey = secretKey;
      if (
        // Only the new MRF submissions have a submission secret key
        submissionType === SubmissionType.Multirespondent &&
        submission.mrfVersion && mrfSubmissionSecretKey
      ){
        attachmentDecryptionKey = mrfSubmissionSecretKey
      }
```

eliotlim (commented):
praise: good to see that this pr has a single intention and is well-scoped.

just a few minor comments on readability and style.

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
kevin9foong: yes  - instead of throwing an error and using try / catch to determine what logic to execute next, opting to use `isSubmissionDecryptionSuccessful` boolean flag to determine the next logic. 

rationale: 
- decryption failures are an expected case, hence our decryption library returns `null` to prevent use of error throwing. 
- using try catch could catch even non decryption related errors, hence need to be careful what code to add within the try block, making the code less extensible. 

logging is not required since these errors are logged to the browser console, which we should ideally keep clean for users.

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
kevin9foong: suggesting to keep it as is for this pr for a couple reasons: 
- its currently using the same code over from existing implementation (just copy-pasted). this specific logic is refactored in a subsequent stacked pr to move it to its own function, readability improvements are introduced there. 
- using `let` instead of `const` has some negatives - const ensures the `attachmentDecryptionKey` is immutable and can prevent bugs.

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
kevin9foong: removed the ordering change

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
kevin9foong: resolved! removed the destructuring

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
eliotlim: fair, makes sense!

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -102,114 +178,86 @@ async function decryptIntoCsv(
...
+    })
+
+    if (!decryptSubmissionDataResult.isSubmissionDecryptionSuccessful) {
+      csvRecord.setStatus(CsvRecordStatus.Error, 'Decryption Error')
+      csvRecord.materializeSubmissionData()
+      return csvRecord as MaterializedCsvRecord
+    }
+
+    const { decryptedResponses, mrfSubmissionSecretKey } =
+      decryptSubmissionDataResult
```
eliotlim, line 187: nit(style): (i) would avoid destructuring; and (ii) for brevity, name this `decryptedContent` with an annotation `// decrypted submissionData result`

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -49,38 +57,106 @@ function verifySignature(
...
+      mrfSubmissionSecretKey?: string
+      isSubmissionDecryptionSuccessful: true
+    }
+  | {
+      errorMessage: string
+      isSubmissionDecryptionSuccessful: false
+    }
+> {
+  const { encryptedContent, verifiedContent, version, submissionType } =
+    submissionData
```
eliotlim, line 77: nit(style): would avoid destructuring to minimise code in our context window, since it takes up the same number of lines when referenced

eliotlim (approved):
tested all TCs on stg-alt2 and it works!
- it's great that this PR lays the foundation for improving admin experience through bulk decryption and PDF generation.
- just a few minor stylistic suggestions on destructuring and code readability

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
kevin9foong: for (ii), im generally against the use of comments/annotations on what something is - it usually means that the name could be more descriptive or function could have been better abstracted. (references: https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/, https://dev.to/jwwnz/be-descriptive-when-naming-33m0) 

To provide more context about the naming choice: 
- the name includes `submission` so it provides context that its the submission that is decrypted (rather than content which could be ambiguous and mean either attachment or content both decrypted in the function). 
- the use of the word `Result` is consistent with our backend which uses never throw and Result means that there is an error status that can be checked, similar to Result.isErr() in our backend

kevin9foong: (i) as for the object de-structuring, i think it might mainly a personal style thing - imo adding de-structures does not make the code worse and lines of code is not really a concern or a metric im optimising for. (reference: https://stackoverflow.com/questions/54605286/what-is-destructuring-assignment-and-its-uses)

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
scottheng96: If `errorMessage` isn't used downstream, would it be a good idea to remove it from what `decryptSubmissionData` returns, and instead only return `isSubmissionDecryptionSuccessful`. Or is it used elsewhere (maybe in a later part x PR).

kevin9foong: makes sense - since its not used, i've removed the `errorMessage`. thanks for highlighting this!

PR #8987, opened 2025-12-15
feat(bulk-local-pdf): abstract out attachment downloads (part 3)

kevin9foong:
## Problem

Currently, the `decryptIntoCsv` function handles both the creation of csv rows and downloading of attachments. This makes it challenging to reuse the same decryption function for other purposes (eg, bulk pdf generation, charts, better dashboards, truly parallel local decryptions)  

Also, the downloading of attachments uses exceptions for flow control, which makes the code hard to understand. this should be avoided, see: https://nus-cs2030s.github.io/2021-s2/19-exception.html#do-not-use-exception-as-a-control-flow-mechanism, https://medium.com/@samanwayghatak/exception-as-control-flow-anti-pattern-e3b46b079cdd

## Solution

Abstract the attachment downloads to its own reusable function.  

**Breaking Changes** 

## Tests

Pre-req: 
- [ ] Create 2 forms - one MRF and one storage mode form with attachment field + make multiple submissions for both forms. 

- [ ] Download csv + attachments from the storage form. Assert the files are downloaded and csv is correctly 
formatted.  
- [ ] Download csv only from the storage form. Assert the csv is correctly formatted.  

- [ ] Download csv + attachments from the mrf form. Assert the files are downloaded and csv is correctly 
formatted.  
- [ ] Download csv only from the mrf form. Assert the csv is correctly formatted.

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
@@ -50,6 +50,7 @@ function verifySignature(
         fieldId,
         answer,
       })
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
     } catch (error) {
       return false
     }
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -130,6 +131,85 @@ async function decryptSubmissionData({
...
+  secretKey: string
+}) {
+  const { submissionType, mrfVersion } = submission
+  if (!mrfSubmissionSecretKey) {
+    // If no mrf submission secret key present, it is a storage mode form. So, use form secret key.
+    return secretKey
+  }
+  const isOldMrfVersion =
+    submissionType === SubmissionType.Multirespondent && !mrfVersion
+  if (isOldMrfVersion) {
```
scottheng96, line 153: I did not know there was a time when we used form key instead of submission key to encrypt/decrypt mrf attachments! Nice backwards compatibility here👍

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -195,56 +273,37 @@ async function decryptIntoCsv(data: LineData): Promise<MaterializedCsvRecord> {
...
+
+      csvRecord.setStatus(
+        CsvRecordStatus.Ok,
+        'Success (with Downloaded Attachment)',
+      )
+      csvRecord.setDownloadBlob(
+        attachmentDownloadBlob.downloadedAttachmentsBlob,
+      )
     }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
scottheng96, line 306: nit: A bunch of these added to the code, which I believe the app builds fine without it. An option to consider to reduce noise in the file/codebase!

scottheng96 (approved):

scottheng96 (approved):

PR #7553, opened 2024-07-31
refactor: worker pool

g-tejas:
## Problem

When downloading responses in the order of thousands, the respones are not actually being fetched in parallel. The "staircase" bug

Screenshot 2024-07-18 at 4 51 31 PM

The network is not saturated either, preferably we want to issue downloads in parallel.
Screenshot 2024-07-18 at 5 42 29 PM

FRM-1760

## Solution
https://www.notion.so/opengov/RFC-Worker-Pool-Improvement-c32cae3dc4fc47c5b07663cd4f6361d8?pvs=4

Note: This PR will be gated behind a growthbook feature, `faster-downloads`

**Breaking Changes** 
- No - this PR is backwards compatible  

## Tests
**Before each:**
- [ ] 1. Create a download folder for these tests
- [ ] 2. Set your browsers default download location to this folder
- [ ] 3. Ensure that the option, "Ask where to save each file before downloading" is set to off
- [ ] 4. Ensure that the growthbook feature flag `faster-downloads` is turned on for staging
- [ ] 5. Navigate to the submissions page and open console tab, observe that there's a log "Faster downloads is enabled ⚡"

**After each:**
- [ ] 1. Empty the download folder created

**Regression Test: Ensure that encrypt form responses downloads work**
- [ ] 1. Create an encrypt mode form
- [ ] 2. Add some basic fields, including attachments. About > 10 mb worth of attachment fields. 
- [ ] 3. Make form public
- [ ] 4. Make about 30-40 submissions
- [ ] 5. Navigate to submissions page, and enter secret key
- [ ] 6. Download just CSV, without attachments. Open the CSV and observe that the CSV meets the following
    - "Expected total responses" is the correct number
    - "Success count" is the correct number
    - Error count is zero
    - Unverified response count is zero
    - The number of submission rows matches expected total responses
- [ ] 7. Download the CSV with attachments. Run command: `find . -name "RefNo *.zip" | wc -l` within the directory and ensure that it matches the expected submission count. 
- [ ] 8. Observe that the CSV meets the requirement (same as step 6)

**Regression Test: Ensure that MRF form responses download work**
- [ ] 1. Create an MRF mode form with multiple workflow steps
- [ ] 2. Add some basic fields, including attachments. About > 10 mb worth of attachment fields. 
- [ ] 3. Make form public
- [ ] 4. Make about 30-40 submissions
- [ ] 5. Navigate to submissions page, and enter secret key
[...]

PR #8988, opened 2025-12-15
feat(bulk-local-pdf): remove faster downloads (part 2)

kevin9foong:
## Problem

Currently, faster downloads which is not being used is blocking refactoring of decryption worker code - since subsequent changes can further invalidate its functionality. 

Faster downloads is not used due to a bug in production where attachments files for mrf do not download properly in bulk (eg, only 10/16 files download), hence, has the feature has been triggered off for the past 8 months. 

Looking into the rationale why this pr was made in the first place - this might not be the ideal/correct fix for the staircase problem (where the root cause seems to be due to the await before assigning the task to the next worker)
eg, assign chunk 1 to worker 1 -> await for it to be done -> assign chunk 2 to worker 2 -> await ... which leads to the staircase pattern described in https://github.com/opengovsg/FormSG/pull/7553. 

Note that this staircase issue is resolved in https://github.com/opengovsg/FormSG/pull/8990 (a subsequent pr).  

Hence, the code should be removed to reduce complexity in future implementations, and its code may still be used for reference if needed since it is still available in version control history.

## Solution

Remove faster downloads which is not in use. 

**Breaking Changes** 

## Tests

Pre-req: 
- [ ] Create 2 forms - one MRF and one storage mode form with attachment field + make multiple submissions for both forms. 

- [ ] Download csv + attachments from the storage form. Assert the files are downloaded and csv is correctly 
formatted.  
- [ ] Download csv only from the storage form. Assert the csv is correctly formatted.  

- [ ] Download csv + attachments from the mrf form. Assert the files are downloaded and csv is correctly 
formatted.  
- [ ] Download csv only from the mrf form. Assert the csv is correctly formatted.

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
@@ -1,7 +1,6 @@
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { useMutation, UseMutationOptions } from 'react-query'
 import { datadogLogs } from '@datadog/browser-logs'
-import { useFeatureIsOn } from '@growthbook/growthbook-react'
 
 import { waitForMs } from '~utils/waitForMs'
 
@@ -15,10 +14,7 @@ import {
 } from '~features/analytics/AnalyticsService'
 import { useUser } from '~features/user/queries'
 
-import {
-  downloadResponseAttachment,
-  downloadResponseAttachmentURL,
-} from './utils/downloadCsv'
+import { downloadResponseAttachment } from './utils/downloadCsv'
 import { EncryptedResponseCsvGenerator } from './utils/EncryptedResponseCsvGenerator'
 import {
   EncryptedResponsesStreamParams,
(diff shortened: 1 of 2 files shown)


kevin9foong:
image
As shown in the ss, the submissions are being processed sequentially - this is the likely cause of the staircase pattern. 

This is since the decrypt into csv function (which does the decryption + attachment downloading)  is being awaited and then hence processed 1 by 1. Thus, there is no actual parallelism. 
```
let i = 0
      return new Promise<DownloadResult>((resolve, reject) => {
        reader
          .read()
          .then(
            (read = async (result) => {
              const curr = i++
              console.log('starting i', curr)
              await new Promise((resolve) => setTimeout(resolve, 5000))
              console.log('finished i', curr)
              if (result.done) return
              try {
                // round-robin scheduling
                const { workerApi } =
                  workerPool[receivedRecordCount % numWorkers]
                const decryptResult = await workerApi.decryptIntoCsv({
                  line: result.value,
                  secretKey,
                  downloadAttachments,
                  formId: adminForm._id,
                  hostOrigin: window.location.origin,
                })
```

scottheng96:
🫡 @g-tejas

scottheng96 (approved):

PR #8989, opened 2025-12-15
feat: add agency to gb attributes

eliotlim:
## Problem

Some features should be rolled out on an agency-by-agency basis.

Closes FRM-2216

## Solution

**Breaking Changes** 

**Features**:

- Add agency (short name) as an targetable attribute in GrowthBook

### DIFF
--- a/frontend/src/features/public-form/PublicFormProvider.tsx
+++ b/frontend/src/features/public-form/PublicFormProvider.tsx
@@ -621,7 +621,7 @@ export const PublicFormProvider = ({
         // Only update the `formId` attribute, keep the rest the same
         ...growthbook.getAttributes(),
         formId,
-        agencyName: data?.form.admin.agency.shortName,
+        adminAgency: data?.form.admin.agency.shortName,
       })
     }
   }, [growthbook, formId, data?.form.admin.agency.shortName])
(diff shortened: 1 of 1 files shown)


frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -621,9 +621,10 @@ export const PublicFormProvider = ({
         // Only update the `formId` attribute, keep the rest the same
         ...growthbook.getAttributes(),
         formId,
+        agencyName: data?.form.admin.agency.shortName,
```
kevin9foong: suggestion: hey eliot, for consistency with other usages of gb, shall we use the term `adminAgency`? since when using gb, it will give suggestions for target attributes, it might be good to use the same term.

kevin9foong (approved):

kevin9foong:
Apart from the above suggestion, lgtm

frontend/src/features/public-form/PublicFormProvider.tsx
eliotlim: Oops that's right!

eliotlim:
```suggestion
        adminAgency: data?.form.admin.agency.shortName,
```

PR #7451, opened 2024-06-27
fix(attachment): convoy separation (#7449)

KenLSM:
## New

- fix(attachment): convoy separation [`#7449`](https://github.com/opengovsg/FormSG/pull/7449)
- build: release v6.130.0 [`#7441`](https://github.com/opengovsg/FormSG/pull/7441)

## Tests

@kevin9foong 
Regression on large submissions >10 crossing two convoys
- [x] Create 40 submissions with 20 attachments
- [x] Ensure that Download CSV (without attachment) returns 1 `.csv` file
- [x] Ensure that Download CSV with attachment returns 40 `.zip` and 1 `.csv` file

Regression on small submissions < 10
- [x] Create 7 submissions with 3 attachments only 1 convoy
- [x] Ensure that Download CSV (without attachment) returns 1 `.csv` file
- [x] Ensure that Download CSV with attachment returns 7 `.zip` and 1 `.csv` file

PR #8990, opened 2025-12-15
fix(bulk-local-pdf): parallel local response decryption with web workers (part 4) 

kevin9foong:
## Problem

Previously, despite use of multiple web worker nodes, the decryption is being done sequentially and not in parallel which is a bug. 

Explanation: The function below iterates recursively to eventually process all the reader chunks, where each chunk represents a single submission to decrypt. 
However, since the submission decryption is first awaited before carrying on to the next submission to decrypt, despite having many workers, it will never process multiple submissions in parallel.  
```
let i = 0
      return new Promise<DownloadResult>((resolve, reject) => {
        reader
          .read()
          .then(
            (read = async (result) => {
              if (result.done) return // no more chunks
              try {
                // round-robin scheduling
                const { workerApi } =
                  workerPool[receivedRecordCount % numWorkers]
                const decryptResult = await workerApi.decryptIntoCsv({ // awaiting here blocks assigning of next chunk to next available worker. 
                  line: result.value,
                  secretKey,
                  downloadAttachments,
                  formId: adminForm._id,
                  hostOrigin: window.location.origin,
                }).then(() => {read.read() // advances to the next chunk}} 
```
Assuming 2 workers and 4 chunks: 
The previous faulty flow is: 
web worker 1: assign chunk 1 
await decryption of chunk 1 by worker 1 
web worker 2: assign chunk 2
await decryption of chunk 2 by worker 2
web worker 1: assign chunk 3
await decryption of chunk 3 by worker 1
web worker 2: assign chunk 4
await decryption of chunk 4 by worker 2
done 

This leads to this "staircase" issue
image

Hence, the decryption is slow. 

Fixes FRM-2229

## Solution

Remove the await before assigning the next chunk to the next worker. 

This makes the flow become: 
web worker 1: assign chunk 1 
web worker 2: assign chunk 2
web worker 1: assign chunk 3
web worker 2: assign chunk 4
After the chunks are assigned - they can be executed in parallel by the workers. 
await(completion of all 4 chunks by the workers) 
continue logic 

once the decryption is done and the blobs are generated in memory 
next, another `then()` proceeds to download them using the convoy method to prevent browser blockage, see: https://github.com/opengovsg/FormSG/pull/7451. 
done 

image
For example, after this change - the order of decryption completion can now be interleaved - compared to previously where it must be sequential 1->2->3->4

**Breaking Changes** 

## Tests

Pre-req: 

formatted.  
- [x] Assert that the progress modal % reflects the decryption status correctly. 

formatted.  
- [x] Assert that the progress modal % reflects the decryption status correctly.

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
@@ -111,7 +111,7 @@ const useDecryptionWorkers = ({
       let errorCount = 0
       let unverifiedCount = 0
       let attachmentErrorCount = 0
-      let receivedRecordCount = 0
+      let currentSubmissionIndex = 0 // used to iterate through the workers
 
       const logMeta = {
         action: 'downloadEncryptedReponses',
@@ -158,7 +158,7 @@ const useDecryptionWorkers = ({
       await reader.read().then(
         (read = async (result) => {
           if (result.done) return
-          const { workerApi } = workerPool[receivedRecordCount % numWorkers]
+          const { workerApi } = workerPool[currentSubmissionIndex % numWorkers]
           submissionDecryptPromises.push(
             workerApi
               .decryptIntoCsv({
@@ -183,7 +183,6 @@ const useDecryptionWorkers = ({
                   case CsvRecordStatus.Ok: {
                     try {
                       csvGenerator.addRecord(decryptResult.submissionData)
-                      receivedRecordCount++
                     } catch (e) {
                       errorCount++
                       console.error('Error in getResponseInstance', e)
@@ -198,6 +197,7 @@ const useDecryptionWorkers = ({
                 return decryptResult
               }),
           )
+          currentSubmissionIndex += 1 // used to assign the next submission to the next worker
           return reader.read().then(read)
         }),
       )
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
@@ -112,6 +112,7 @@ const useDecryptionWorkers = ({
       let unverifiedCount = 0
       let attachmentErrorCount = 0
       let currentSubmissionIndex = 0 // used to iterate through the workers
+      let attachmentsToSaveCount = 0 // used for scheduling delays between attachment zip file saves
 
       const logMeta = {
         action: 'downloadEncryptedReponses',
@@ -155,10 +156,13 @@ const useDecryptionWorkers = ({
       let decryptionProgress = 0
       const submissionDecryptPromises: Promise<MaterializedCsvRecord>[] = []
 
+      let timeSinceLastXAttachmentDownload = 0
+
       await reader.read().then(
         (read = async (result) => {
           if (result.done) return
           const { workerApi } = workerPool[currentSubmissionIndex % numWorkers]
+          // Step 1: Use worker to decrypt the submission (and download and decrypt attachments if needed).
           submissionDecryptPromises.push(
             workerApi
               .decryptIntoCsv({
@@ -168,6 +172,7 @@ const useDecryptionWorkers = ({
                 formId: adminForm._id,
                 hostOrigin: window.location.origin,
               })
+              // Step 2: Update the Csv record status based on the decryption result.
               .then(async (decryptResult) => {
                 switch (decryptResult.status) {
                   case CsvRecordStatus.Error:
@@ -191,10 +196,42 @@ const useDecryptionWorkers = ({
                 }
                 return decryptResult
               })
-              .then((decryptResult) => {
+              // Step 3: Save the downloaded and decrypted attachment blobs for each submission (if required).
+              // This step is done with delays in between groups of files to space out downloads to avoid browser blocking downloads.
+              .then(async (decryptResult) => {
+                if (downloadAttachments && decryptResult.downloadBlob) {
+                  attachmentsToSaveCount += 1
+
+                  // Ensure attachments downloads are spaced out to avoid browser blocking downloads
+                  if (
+                    attachmentsToSaveCount % ATTACHMENT_DOWNLOAD_CONVOY_SIZE ===
+                    0
+                  ) {
+                    const now = new Date().getTime()
+                    const elapsedSinceXDownloads =
+                      now - timeSinceLastXAttachmentDownload
+
+                    const waitTime = Math.max(
+                      0,
+                      ATTACHMENT_DOWNLOAD_CONVOY_MINIMUM_SEPARATION_TIME -
+                        elapsedSinceXDownloads,
+                    )
+                    if (waitTime > 0) {
+                      await waitForMs(waitTime)
+                    }
+                    timeSinceLastXAttachmentDownload = now
+                  }
+                  await downloadResponseAttachment(
+                    decryptResult.downloadBlob,
+                    decryptResult.id,
+                  )
+                }
+                return decryptResult
+              })
+              // Step 4: Update the progress bar only once the attachments for the decrypted submission have been downloaded (if needed).
+              .finally(() => {
                 decryptionProgress += 1
                 onProgress(decryptionProgress)
-                return decryptResult
               }),
           )
           currentSubmissionIndex += 1 // used to assign the next submission to the next worker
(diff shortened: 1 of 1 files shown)


frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -110,7 +111,7 @@ const useDecryptionWorkers = ({
       let errorCount = 0
       let unverifiedCount = 0
       let attachmentErrorCount = 0
-      let receivedRecordCount = 0
+      let currentSubmissionIndex = 0 // used to iterate through the workers
```
kevin9foong, line 114: explanation: previously, the receivedRecordCount is used to distribute the tasks to the worker - which is updated only after the submission is decrypted by the worker. hence, there is a bottleneck where must wait for worker to complete first. 

instead the submissions should be distributed round robin even before the submission is decrypted - so that submissions can be decrypted in parallel.

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -151,30 +152,23 @@ const useDecryptionWorkers = ({
...
-        reader
-          .read()
-          .then(
-            (read = async (result) => {
-              if (result.done) return
-              try {
-                // round-robin scheduling
-                const { workerApi } =
-                  workerPool[receivedRecordCount % numWorkers]
-                const decryptResult = await workerApi.decryptIntoCsv({
```
kevin9foong, line 168: awaiting here means that the decryption must be done by the worker before the next submission is iterated and assigned to the next worker. which means that decryption is distributed across multiple workers, but done sequentially. hence, there is a bug in the use of web workers.

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -189,43 +183,57 @@ const useDecryptionWorkers = ({
...
                       errorCount++
                       console.error('Error in getResponseInstance', e)
                     }
+                  }
+                }
+                return decryptResult
+              })
+              .then((decryptResult) => {
+                decryptionProgress += 1
+                onProgress(decryptionProgress)
```
kevin9foong, line 234: Rationale: used to reflect the decryption status in the progress loader modal

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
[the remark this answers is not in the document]
eliotlim: currentSubmissionIndex is fine.

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
eliotlim: praise: yes, agreed -- this is probably why we didn't see any parallelism improvements.

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -189,43 +183,57 @@ const useDecryptionWorkers = ({
...
-                        timeSinceLastXAttachmentDownload = now
-                      }
-                      await downloadResponseAttachment(
-                        decryptResult.downloadBlob,
-                        decryptResult.id,
-                      )
-                    }
+      return new Promise<DownloadResult>((resolve, reject) => {
+        // Step 1: Decrypt all submissions and their attachments (into blobs), add the submissions into the csv.
+        Promise.all(submissionDecryptPromises)
```
eliotlim, line 244: question: should this be `Promise.allSettled` instead?
- it looks like we'd want to handle decryption of any single submission gracefully, rather than halt completely.
- (I recognise that this is part of a larger changeset of PRs that may address this elsewhere)

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -189,43 +183,57 @@ const useDecryptionWorkers = ({
...
+          // Step 2: Download the attachment blobs for each submission.
```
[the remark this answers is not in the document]
eliotlim: not necessary!

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -189,43 +183,57 @@ const useDecryptionWorkers = ({
...
+          // This step is done separately to allow for spacing out the downloads to avoid browser blocking downloads.
+          .then((decryptResults) => {
+            let timeSinceLastXAttachmentDownload = 0
+            decryptResults.forEach(async (decryptResult, index) => {
```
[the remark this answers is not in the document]
eliotlim: question: this is a valid concern -- but short of implementing a full-blown download manager (in addition to the existing decryption manager + progress UI we have)
- do we implement download progress anywhere else in the other PRs of this changeset?
- or is `Promise.allSettled` something we should consider adding to this implementation?

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -191,95 +195,99 @@ const useDecryptionWorkers = ({
                 }
                 return decryptResult
               })
+              // Step 3: Update the decryption count
               .then((decryptResult) => {
                 decryptionProgress += 1
-                onProgress(decryptionProgress)
```
kevin9foong: description: do not update modal status until the file is saved on user's device

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -191,95 +195,99 @@ const useDecryptionWorkers = ({
...
+                }
+                console.error(
+                  'Failed to download data, is there a network issue?',
+                  err,
+                )
+                throw err
+              })
+              // Step 5: Update the progress bar only once the attachments for the decrypted submission have been downloaded (if needed).
+              .finally(() => {
+                onProgress(decryptionProgress)
```
kevin9foong: instead - only update the progress once the download is saved or has failed

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -191,95 +195,99 @@ const useDecryptionWorkers = ({
...
+                    timeSinceLastXAttachmentDownload = now
+                  }
+                  await downloadResponseAttachment(
+                    decryptResult.downloadBlob,
+                    decryptResult.id,
+                  )
+                }
+                return decryptResult
+              })
+              .catch((err) => {
```
kevin9foong: note: if the download has failed - err thrown here, it is still possible for the err count to not reflect the failure, since the error count is based on the above csvRecord

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -112,6 +112,7 @@ const useDecryptionWorkers = ({
       let unverifiedCount = 0
       let attachmentErrorCount = 0
       let currentSubmissionIndex = 0 // used to iterate through the workers
+      let attachmentsToSaveCount = 0 // used for scheduling delays between attachment zip file saves
```
kevin9foong, line 115: rationale for addition: used to track the index of the attachment being downloaded independently from other variables - replicates the same behavior as previous implementation. 

this is now needed since the `decryptProgress` cannot be used anymore for the index since the decrypt and saving of attachment can be interleaved. 

Previous: decrypt 1 -> decrypt count = 1 -> save file 1 ->  decrypt 2 -> decrypt count = 2 -> save file 2 -> ... 
New: decrypt 1 -> decrypt count = 1  -> decrypt 2 -> decrypt count = 2 -> save file 1 -> save file 2 (note that the decrypt count cannot be used anymore since it will increment to 2 even before save file 1 is done). Hence, a separate variable is used to count the index for the save file (and be synchronized) to replicate the same existing behavior.

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -191,49 +196,52 @@ const useDecryptionWorkers = ({
...
+                  await downloadResponseAttachment(
+                    decryptResult.downloadBlob,
+                    decryptResult.id,
+                  )
+                }
+                return decryptResult
+              })
+              // Step 4: Update the progress bar only once the attachments for the decrypted submission have been downloaded (if needed).
+              .finally(() => {
                 decryptionProgress += 1
```
kevin9foong, line 233: rationale for change: this decryption progress should only be updated once the file has been saved to device - replicating existing behavior.

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
kevin9foong: great catch. i have updated such that the file saving should be done after the submission is decrypted to replicate previous behavior. 

this prevents the case where: attachment files are not yet saved to device but the download modal's `onProgress` is already updated to 100% and hence the download progress is out of sync with actual file downloading. 

with the new implementation, `Promise.all` is used to reject if any one of the submissions fail, so that the catch will now do the necessary logs and killing of workers

frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -203,37 +240,8 @@ const useDecryptionWorkers = ({
       )
 
       return new Promise<DownloadResult>((resolve, reject) => {
-        // Step 1: Decrypt all submissions and their attachments (into blobs), add the submissions into the csv.
+        // Step 1: Decrypt all submissions and their attachments (into blobs), add the submissions into the csv object and save the attachments into user's computer.
         Promise.all(submissionDecryptPromises)
-          // Step 2: Download the attachment blobs for each submission.
-          // This step is done separately to allow for spacing out the downloads to avoid browser blocking downloads.
-          .then((decryptResults) => {
```
kevin9foong: rationale for change: this should be moved to after decryption of each submission, this replicates existing behavior and allows the synchronization of the download progress modal and the actual file downloads

eliotlim (commented):
praise: great work identifying and addressing the issue to speedup decryption and downloads! 🚀 

LGTM 👍

eliotlim (approved):
praise: great work identifying and addressing the issue to speedup decryption and downloads! 🚀 

LGTM 👍

PR #8991, opened 2025-12-15
feat(bulk-local-pdf): minor decrypt into csv refactors (part 5) 

kevin9foong:
## Problem

Exceptions are not thrown for most of the function - however, pokemon exception catching is performed. 
Stale comment left. 

Fixes FRM-2230

## Solution

Isolate the exception catching to the line of code where exceptions can be thrown. Remove stale comment. 

**Breaking Changes** 

## Tests

Pre-req: 

formatted.  

formatted.  

### DIFF
--- a/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
+++ b/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
@@ -216,103 +216,102 @@ async function downloadAndDecryptSubmissionAttachments({
  * @param data The data to decrypt into a csvRecord.
  */
 async function decryptIntoCsv(data: LineData): Promise<MaterializedCsvRecord> {
-  // This needs to be dynamically imported due to sharing code between main app and worker code.
-  // Fixes issue raised at https://stackoverflow.com/questions/66472945/referenceerror-refreshreg-is-not-defined
-  // Something to do with babel-loader.
   const { line, secretKey, downloadAttachments, formId, hostOrigin } = data
   let csvRecord: CsvRecord
 
+  let submission: SubmissionStreamDto
   try {
-    const submission = SubmissionStreamDto.parse(JSON.parse(line))
-
-    csvRecord = new CsvRecord(
-      submission._id,
-      submission.created,
-      CsvRecordStatus.Unknown,
-      formId,
-      hostOrigin,
-      submission.submissionType === SubmissionType.Encrypt
-        ? submission.payment
-        : undefined,
-      submission.submissionType === SubmissionType.Multirespondent
-        ? {
-            workflowStatus: submission.mrfMeta.workflowStatus,
-            workflowCurrentStepNumber:
-              submission.mrfMeta.workflowCurrentStepNumber,
-            workflowNumTotalSteps: submission.mrfMeta.workflowNumTotalSteps,
-            lastSubmittedAt: submission.mrfMeta.lastSubmittedAt,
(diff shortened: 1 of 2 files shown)


frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -218,103 +218,105 @@ async function downloadAndDecryptSubmissionAttachments({
  * @param data The data to decrypt into a csvRecord.
  */
 async function decryptIntoCsv(data: LineData): Promise<MaterializedCsvRecord> {
-  // This needs to be dynamically imported due to sharing code between main app and worker code.
```
kevin9foong, line 219: this is the stale comment in the PR body

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -218,103 +218,105 @@ async function downloadAndDecryptSubmissionAttachments({
...
-  // Fixes issue raised at https://stackoverflow.com/questions/66472945/referenceerror-refreshreg-is-not-defined
-  // Something to do with babel-loader.
   const { line, secretKey, downloadAttachments, formId, hostOrigin } = data
   let csvRecord: CsvRecord
 
+  let submission: SubmissionStreamDto
   try {
-    const submission = SubmissionStreamDto.parse(JSON.parse(line))
```
kevin9foong, line 226: pokemon exception catching is done in this try block

frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -216,103 +216,102 @@ async function downloadAndDecryptSubmissionAttachments({
...
+  const decryptSubmissionDataResult = await decryptSubmissionData({
+    submissionData: submission,
+    secretKey,
+  })
 
-    const { decryptedResponses, mrfSubmissionSecretKey } =
-      decryptSubmissionDataResult
+  if (!decryptSubmissionDataResult.isSubmissionDecryptionSuccessful) {
+    csvRecord.setStatus(CsvRecordStatus.Error, 'Decryption Error')
+    csvRecord.materializeSubmissionData()
```
eliotlim, line 270: notice: `materializeSubmissionData()` [can throw because it invokes](https://github.com/opengovsg/FormSG/blob/6b7aa1dde6ca9be64afc2c3e68fc3fe82a722b6c/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/CsvRecord.class.ts#L167) `getStatusFromWorkflowStatus` ([implementation](https://github.com/opengovsg/FormSG/blob/6b7aa1dde6ca9be64afc2c3e68fc3fe82a722b6c/frontend/src/features/admin-form/responses/common/utils/mrfSubmissionView.ts#L72)).
- this is unlikely, but is something we should be careful of as we align on result type error handling

kevin9foong: thanks for pointing this out as a point of consideration and something to note. agreed that it is something to look out for 

thankfully, looking further into it, it might not be an issue for this PR as in the previous implementation, the `materializeSubmissionData` is also invoked outside the `try` block. This means the behavior if the `materializeSubmissionData` throws should be the same.  🎉

eliotlim (approved):
LGTM! 👍

PR #9001, opened 2025-12-18
fix: mrf respondent copy email body line breaks

scottheng96:
## Problem

In MRF respondent copy emails, email bodies are not accurately handling line breaks. This only affects WOG emails. Specifically, WOG email client isn't compatible with CSS properties handling whitespace.

Fixes FRM-2222

## Solution

Explicitly define line breaks using `<br>` such that WOG email clients may render the body accurately.

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

**Test MRF respondent copy email**
- [x] Create an MRF form, add an email field and enable respondent copy (Include a copy of responses.. toggle)
- [x] Inside the email field settings, add a custom body message with line breaks
- [x] Assign the email field to step 1 of the workflow
- [x] Make a submission, and key in your WOG email (@tech.gov.sg)
- [x] Using a COMET or GSIB, check that the linebreaks show as expected, as written in the email field.

### DIFF
--- a/src/app/views/templates/MrfRespondentCopyEmail.tsx
+++ b/src/app/views/templates/MrfRespondentCopyEmail.tsx
@@ -22,6 +22,7 @@ import {
   questionMargin,
   answerMargin,
 } from './mrfWorkflowCompletionEmailStyle'
+import React from 'react'
 
 export type QuestionAnswer = {
   question: string
@@ -58,6 +59,30 @@ export const MrfRespondentCopyEmail = ({
           <Hr style={{ margin: '40px 0' }} />
         </>
     )
+    
+    /**
+     * Explicitly render line breaks and paragraphs in email body, as certain email clients
+     * might not be compatible with CSS white-space property. (e.g. white-space: pre-line)
+     * @param body - email body text with line breaks
+     * @param textStyle - css properties
+     * @returns an array of JSX elements representing paragraphs and line breaks
+     */
+    const renderEmailBody = (
+      body: string,
+      textStyle: React.CSSProperties,
+    ): JSX.Element[] => {
+      return body.split(/\n{2,}/).map((paragraph, i) => (
+      <Text key={i} style={{ ...textStyle, marginBottom: '16px' }}>
+        {paragraph.split('\n').map((line, j) => (
+          <React.Fragment key={j}>
+            {line}
+            <br />
+          </React.Fragment>
+        ))}
+        </Text>
+      ))
+      }
+
 
     return (
     <Html>
@@ -70,9 +95,7 @@ export const MrfRespondentCopyEmail = ({
             <Heading style={{ ...headingTextStyle, marginBottom: '40px' }}>
               {headingText}
             </Heading>
-            <Text style={{ ...secondaryTextStyle, marginBottom: '40px', whiteSpace: 'pre-line' }}>
-              {body}
-            </Text>
+            {renderEmailBody(body, secondaryTextStyle)}
             {renderResponseId()}
               {formQuestionAnswers ? (
                 <>


src/app/views/templates/MrfRespondentCopyEmail.tsx
```diff
@@ -58,6 +59,30 @@ export const MrfRespondentCopyEmail = ({
...
+     * @param body - email body text with line breaks
+     * @param textStyle - css properties
+     * @returns an array of JSX elements representing paragraphs and line breaks
+     */
+    const renderEmailBody = (
+      body: string,
+      textStyle: React.CSSProperties,
+    ): JSX.Element[] => {
+      return body.split(/\n{2,}/).map((paragraph, i) => (
+      <Text key={i} style={{ ...textStyle, marginBottom: '40px' }}>
```
scottheng96: Finding my own bugs - this gives every text component 40px margin bottom instead of 16px, and 40px only for the last component. Fixing that now

src/app/views/templates/MrfRespondentCopyEmail.tsx
```diff
@@ -58,6 +59,39 @@ export const MrfRespondentCopyEmail = ({
...
+      textStyle: React.CSSProperties,
+    ): JSX.Element[] => {
+      const paragraphs = body.split(/\n{2,}/)
+
+      return paragraphs.map((paragraph, i) => {
+        const lines = paragraph.split('\n')
+        return (
+          <Text key={i} style={{ ...textStyle }}>
+            {lines.map((line, j) => {
+              const isLastLine = i === paragraphs.length - 1 && j === lines.length - 1
```
eliotlim: - instead of computing an `isLastLine` attribute here and dynamically adjusting the margin of the div,
- would it make sense to add a spacer div between `renderEmailBody` and `renderResponseId`?

src/app/views/templates/MrfRespondentCopyEmail.tsx
```diff
@@ -70,9 +104,7 @@ export const MrfRespondentCopyEmail = ({
             <Heading style={{ ...headingTextStyle, marginBottom: '40px' }}>
               {headingText}
             </Heading>
-            <Text style={{ ...secondaryTextStyle, marginBottom: '40px', whiteSpace: 'pre-line' }}>
-              {body}
-            </Text>
+            {renderEmailBody(body, secondaryTextStyle)}
             {renderResponseId()}
```
eliotlim, line 99: I wonder if it might make more sense to insert a spacer div here, instead of computing div margins dynamically as part of the line and paragraph styling.

src/app/views/templates/MrfRespondentCopyEmail.tsx
scottheng96: Hmm, that's a good point, would make this cleaner.

kevin9foong: I would recommend making use of the `Stack` component and its `gaps={16px}` property. 

Then you could set the overall margin bottom to 40px 

so the suggested code is 
`<Stack mb='40px' gaps='16px'>{<Text> components array}</Stack>` . This would potentially remove the need to compute `isLastLine`

src/app/views/templates/MrfRespondentCopyEmail.tsx
kevin9foong: the div margins might be needed as the 40px is after each paragraph. but it could make sense to just add a spacer after each paragraph

src/app/views/templates/MrfRespondentCopyEmail.tsx
eliotlim: true, yes! as long as Stack renders correctly on desktop Outlook, then this should work.
and we can fallback to divider components otherwise.

scottheng96: Ah okay, I tried this out. Unfortunately `Stack` doesn't exist in @react-email/components, and we're also not importing @chakra-ui in the BE where this template resides. There are components like `Row` and `Section` that might be possible, but I feel that the current implementation is clean enough.

src/app/views/templates/MrfRespondentCopyEmail.tsx
scottheng96: Thanks for the inputs - I've gone ahead to remove the dynamic margins (turns out the already [in-place divider](https://github.com/opengovsg/FormSG/blob/develop/src/app/views/templates/MrfRespondentCopyEmail.tsx#L55) takes care of that)

eliotlim (approved):
okay, cool.

let's do this first, and revisit it properly (with components / better practices) when we standardise the email templates in the future!

thanks everybody for your insights 👍

PR #9013, opened 2025-12-22
feat: abstract download attachments

kevin9foong:


kevin9foong:
Closed as use to see diff against develop

PR #9021, opened 2025-12-30
feat: Update copy for workflow approval toggle

tutntut:
## Problem
The workflow approval toggle is one of the more confusing parts for admins when setting up. It is currently presented as a role and a gate. However, it fails to mention the following:

- Which response (yes/no) counts as "approved" and "not approved"
- The approval outcome will be reflected in various touch points like the results page and the status tracking link

image

## Solution
Update the copy to make it clearer to admins setting up.

image

**Breaking Changes** 

    - Details ...


tutntut:
Not moving out of draft yet until validating with some users that this is clearer copy

kevin9foong (approved):

PR #9032, opened 2026-01-09
fix: bump react router dom version to 6.30.3

scottheng96:
## Problem

Closes FRM-2237

## Solution

**Breaking Changes** 

**Improvements**:

- Use data router variant of BrowserRouter and MemoryRouter to enable useBlocker
- Rewrite the "unsaved changes" modal using `useBlocker` APIs
- Remove usage of unstable APIs `UNSAFE_NavigationContext` which was blocking the dependency upgrade.

**Bug Fixes**:

- Update `react-router-dom` to `6.30.3`

## Tests

**TC: Check that Form Builder blocks navigation when changes are unsaved**
- [x] Visit an existing form, and make a change to any field
- [x] Without saving, navigate away from the field. An unsaved changes modal should pop up. 
- [x] Clicking Discard should drop the unsaved changes, and allow navigation.
- [x] Clicking Go back should preserve the unsaved changes, and prevent navigation.

## Deploy Notes

**New dependencies**:

- `react-router-dom` : `6.30.3`

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -73,7 +73,7 @@
         "react-markdown": "^9.0.1",
         "react-query": "^3.34.2",
         "react-remove-scroll": "^2.5.10",
-        "react-router-dom": "^6.0.2",
+        "react-router-dom": "^6.30.3",
         "react-roving-tabindex": "^3.2.0",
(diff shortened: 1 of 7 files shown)


eliotlim (approved):
LGTM! 👍

PR #9037, opened 2026-01-12
feat: add forumsg link

scottheng96:
## Solution

Adding a 'Forum' nav bar link as the main entrypoint for ForumSG. Feature is gated under a feature-flag and planned to be released scoped to *moe.gov.sg & *schools.edu.sg users

image

**Breaking Changes** 

## Before & After Screenshots

| --- | --- | --- |
| AdminNavBar | image | image |

## Tests

**Test feature flag**
- [ ] Visit FormSG and log in
- [ ] With feature-flag on (`forumsg`), observe that 'Forum' is visible on the navigation bar
- [ ] Turn feature-flag off, and observe that 'Forum' is no longer visible on the navigation bar


.vscode/settings.json
```diff
@@ -1,25 +1,17 @@
 {
+  "eslint.enable": true,
```
kevin9foong: Issue: changes to this linting file might not be a part of the change. is this perhaps an accidental commit?

docker-compose.override.yml
```diff
@@ -0,0 +1,17 @@
+version: '3'
+services:
+  mockpass:
+    image: node:22-slim
+    working_dir: /usr/src/mockpass
```
kevin9foong: Issue: is this meant to be committed?

frontend/src/app/AdminNavBar/AdminNavBar.tsx
```diff
@@ -252,6 +279,17 @@ export const AdminNavBar = ({ isMenuOpen }: AdminNavBarProps): JSX.Element => {
...
       MobileIcon: BxsHelpCircle,
     },
+    // TODO: Remove forum link after H4PG2026
+    ...(isForumSGEnabled
+      ? [
+          {
+            label: 'Forum',
+            href: FORUMSG_URL,
+            MobileIcon: MdForum,
+            shouldShowNotification: true,
```
kevin9foong, line 289: Note: This means that the red dot will always be enabled for forum. This makes sense for hackathon to draw attention to the link, but should be removed post hackathon

docker-compose.override.yml
scottheng96: mb. removing!

.vscode/settings.json
scottheng96: mb. removing!

kevin9foong (approved):

PR #9041, opened 2026-01-14
feat: expose growthbook attributes on admin nav bar

scottheng96:
## Problem

Growthbook feature flag `forumsg` is unable to utilize rules to serve different values specific to `adminEmail` or `adminAgency`. This is because these attributes are not set in the AdminNavBar component.

## Solution

Set growthbook attributes `adminEmail` and `adminAgency` in AdminNavBar.tsx.

**Breaking Changes** 


eliotlim (approved):
LGTM! :thumbe

PR #9053, opened 2026-01-21
feat: support WOG AD authentication

kevin9foong:
## Problem

Admin users frequently experience OTP delays and deliverability issues preventing them from logging in. Hence, introduce WOG AD as a login mechanism to reduce this point of failure. 

## Solution

Support WOG AD login adhering to the common OAuth2.0 code authorization flow for authentication.  
However, we use this authentication one-time to verify the email and discard the token thereafter. 
At the point the email is verified, we can generate the standard FormSG session (same as OTP) and resume from there. 

Login flow illustrated: 
image

Logout flow illustrated: 
image

## Special considerations: 
1. Growthbook feat flag is enabled to do a staged rollout to specific user groups / email domains or random % rollout.  We are able to flag according to: ALL, INTRANET_AND_RBI_ONLY, OFF

There are some environments which do not support WOG AD.  For such envs, the feature flag should be turned off for them. However, these should return a 405 Method not allowed to the client if the client somehow makes a request. (see: https://apihandyman.io/this-is-not-the-http-method-you-re-looking-for-http-status-code-404-vs-405-vs-501/ for http error code usage rationale).
image

2. Tracking of login methods is done via Datadog. We are reusing express.request.hits to reduce need to implement new metrics (and $$$ cost due to increased cardinality), avoid making DB schema changes. 

## Changes:
[After - with sufyan's design input]
image

**Breaking Changes** 

## Tests

User can still login using Singpass and OTP 
- [ ] Login via singpass 
- [ ] Login via OTP 

User can login via WOG AD 
- [ ] Login via WOG AD 
- [ ] Assert the correct email is loaded and the forms persist. 
- [ ] Logout from FormSG 
- [ ] Assert that user is redirected to Azure logout
- [ ] Assert that user is redirected back to stg.form.gov.sg 

## Deploy Notes

To test on stg, uat envs. 

- `env var` : env var details
Various WOG AD env vars - details explained in the config file. 
**New dependencies**:
msal-node: official microsoft authentication library.


deploy/ecs-task-definition.json
```diff
@@ -116,6 +116,8 @@
         { "name": "SSO_CLIENT_ID", "valueFrom": "/<ENVIRONMENT_SITE_NAME>/SSO_CLIENT_ID" }, 
         { "name": "SSO_CLIENT_SECRET", "valueFrom": "/<ENVIRONMENT_SITE_NAME>/SSO_CLIENT_SECRET" }, 
         { "name": "SSO_DISCOVERY_URL", "valueFrom": "/<ENVIRONMENT_SITE_NAME>/SSO_DISCOVERY_URL" },
+        { "name": "WOGAD_CLIENT_ID", "valueFrom": "/<ENVIRONMENT_SITE_NAME>/WOGAD_CLIENT_ID" },
+        { "name": "WOGAD_CLIENT_SECRET", "valueFrom": "/<ENVIRONMENT_SITE_NAME>/WOGAD_CLIENT_SECRET" },
```
kevin9foong, line 120: TODO: add the remaining WOG AD envs

frontend/src/features/login/LoginPage.tsx
```diff
@@ -127,6 +128,10 @@ export const LoginPage = (): JSX.Element => {
               <SgidLoginButton />
             </>
           )}
+          <>
```
kevin9foong: TODO: check with design how to represent this

frontend/src/features/login/WogadHoldingPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const WogadHoldingPage = () => {
+  const [searchParams] = useSearchParams()
+
+  const code = searchParams.get('code')
+  const csrfToken = searchParams.get('state')
+
+  const { mutate } = useVerifyWogadAuthCode()
+
+  const hasVerified = useRef(false)
```
kevin9foong, line 36: explanation: ref is used to ensure that the `mutate` is only invoked once and is only done when code and csrfToken are loaded from param

src/app/modules/auth/wogad/auth-wogad.controller.ts
```diff
@@ -0,0 +1,179 @@
...
+ */
+export const handleVerifyWithCode: ControllerHandler<
+  unknown,
+  unknown,
+  { code: string; csrfToken: string }
+> = async (req, res) => {
+  const csrfTokenFromCookie = req.cookies['csrf_token']
+  const csrfTokenFromBody = req.body['csrfToken']
+
+  if (!csrfTokenFromCookie || csrfTokenFromCookie !== csrfTokenFromBody) {
```
kevin9foong, line 106: explanation: Read more about this CSRF verification here https://www.oauth.com/oauth2-servers/signing-in-with-google/authorization-request/. 

```
It’s important to generate a “state” parameter to use to protect the client from CSRF attacks. This is a random string that the client generates and stores in the session. Our app will verify that that state parameter in the redirect from Google matches the one that was created at the start of the flow.

File: src/app/modules/auth/wogad/auth-wogad.controller.ts [hunk 5]
```diff
@@ -0,0 +1,179 @@
...
+  )
+    .map((user) => {
+      if (!req.session) {
+        return res
+          .status(StatusCodes.INTERNAL_SERVER_ERROR)
+          .json('Could not find valid session')
+      }
+
+      // Add user info to session.
+      const { _id } = user.toObject() as SessionUser
```
kevin9foong: explanation: uses the existing FormSG session mechanism to login.

src/app/modules/auth/wogad/auth-wogad.controller.ts
```diff
@@ -0,0 +1,179 @@
...
+      await ccaSingleton.acquireTokenByCode(tokenRequest)
+    account = retrievedAccount
+  } catch (error) {
+    return res.status(StatusCodes.FORBIDDEN).json({
+      message: 'Error acquiring token by code',
+      error,
+    })
+  }
+
+  await removeAccountFromTokenCache(account)
```
kevin9foong, line 131: explanation: since the token is used to verify identity at the time of retrieval, it can be disposed after the verification is done to save memory.

kevin9foong:
Added UT coverage for the controllers

eliotlim (approved):

PR #9054, opened 2026-01-22
feat: remove red dot from forumsg nav bar link

scottheng96:
## Problem

Removing the red dot from ForumSG navbar link because UTs have found that the red dot causes link to be too easily discovered.

## Solution

Removed red dot from ForumSG nav bar link

**Breaking Changes** 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image


kevin9foong (approved):

tutntut (approved):
LGTM!

PR #9067, opened 2026-02-02
fix: bug fix for singpass mrf 2+ step email OTPs 

scottheng96:
## Problem

Singpass-enabled MRF forms, that have email OTP on the 2nd step onwards is experiencing intermittent submission issues. Respondent is stuck on the email address verification step as an error message repeatedly pops up "Sorry, something went wrong, please refresh and try again."

Reason:
Before OTP generation, a validation check happens for the presence of a MyInfo JWT cookie if the form is SPCP enabled. Since it should only exist for storage mode Singpass forms + MRF forms on the 1st step (only step with Singpass login), 2+ step onwards fails this check.

Closes FRM-2241

This [check](https://github.com/opengovsg/FormSG/pull/3980) was put in place due to a VAPT finding of not validating the authorisation of the form since the endpoint included a `/:formId/fieldvericiations/*...`. This means that OTPs could still be generated programmatically even for Singpass-enabled forms

## Solution

For MRF forms only:
1. Pass a signed JWT token during MRF form load containing the `previousSubmissionId` and `workflowStep`
2. Pass this, along with the unencrypted `previousSubmissionId` with the OTP generation request
3. Decode JWT token to verify:
     1. `previousSubmissionId` from the JWT is the same as the one passed unencrypted from the FE
     2. `previousSubmissionId` exists in the DB
     3. db object's  in the DB checks corresponds to the `formId` (from the endpoint)
     4. db object's `workflowStep` matches the decoded `workflowStep`
4. If so, bypass the SPCP check

## Alternatives 
Several alternatives were discussed but were not implemented, documented here for posterity

**Passing only `previousSubmissionId` from the FE (from client public context) to BE**
- was not implemented due to security vulnerability of user being able to pass any `previousSubmissionId` and our BE trusting it which would bypass validation check

**Using `mrfSecretKey` instead of `previousSubmissionId`**
- Iffy because, we would be passing both the submissionPublicKey (in the url) and the submissionSecretKey (in the cookie) to the user when the form loads. JWT encoding can hide this but not entire full proof as well, and consensus was reached that `previousSubmissionId` would be enough to check against the DB records for validity

**Breaking Changes** 

## Tests

Email OTP verification sent for 2nd step MRF
- [ ] Create a MRF form
- [ ] Add an email field to the form, enable 'OTP verification' and assign it to the 2nd step of the workflow (1st step can have no fields assigned)
- [ ] Make the form public, and make a step 1 submission
- [ ] In step 2 submission, successfully verify your email with an OTP verfication email that is sent
- [ ] Successfully complete the step 2 submission


.vscode/settings.json
```diff
@@ -2,7 +2,7 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "explicit"
   },
-  "eslint.validate": ["javascript", "typescript"],
```
scottheng96, line 5: Just a side update on vscode lint settings. (feel like i'm the only one using vscode now haha)

src/app/modules/verification/verification.controller.ts
```diff
@@ -111,6 +111,9 @@ export const _handleGenerateOtp: ControllerHandler<
         const { authType } = form
         switch (authType) {
           case FormAuthType.CP: {
+            // Allow bypass of MyInfo verification check Singpass-enabled MRF forms
```
kevin9foong: question: the grammar for this comment is not quite right - im not understanding what it means

kevin9foong: Allow bypass of myinfo verification check for(?) singpass enabled mrf forms?

kevin9foong: might want to add an explanation on why this bypass is allowed here - might throw off future devs reading the code as to why this is being done as it is a high context change

src/app/modules/verification/verification.controller.ts
```diff
@@ -155,6 +158,9 @@ export const _handleGenerateOtp: ControllerHandler<
               })
           case FormAuthType.SGID_MyInfo:
           case FormAuthType.MyInfo:
+            // Allow bypass of MyInfo verification check Singpass-enabled MRF forms
```
kevin9foong: same as above

kevin9foong: issue: if we do this, wouldnt any client be able to bypass this verification check from the FE by just sending a fake `prevsubmissionId`? 

we need to assume that FE payload cannot be trusted and the verification needs to be done in the BE.

kevin9foong: PoC: 
client can intercept the `  return ApiService.post(
    `${FORM_API_PREFIX}/${formId}/${VERIFICATION_ENDPOINT}/${transactionId}/fields/${fieldId}/otp/generate`,
    {
      answer,
      previousSubmissionId,
    },`request using their browser and send a fake `previousSubmissionId`. 
    
using tools such as burpsuite proxy. 

Impact: This allows them to skip the verification for MyInfo even for the first step.

src/app/modules/verification/verification.errors.ts
```diff
@@ -123,3 +123,12 @@ export class OtpRequestError extends ApplicationError {
     super(message, undefined, ErrorCodes.VERIFICATION_OTP_REQUEST_ERROR)
   }
 }
+
+/**
+ * MRF JWT validation failed
+ */
+export class MrfJwtError extends ApplicationError {
```
kevin9foong: suggestion: should this be a more specific error name? eg `MrfJwtValidationError`. bringing this up as the MyinfoCookieErrors are also rather specific eg, Missing Login, Invalid Login, State error

src/app/modules/verification/verification.controller.ts
```diff
@@ -107,6 +115,67 @@ export const _handleGenerateOtp: ControllerHandler<
...
+          try {
+            const decoded = jwt.verify(
+              mrfCookie,
+              config.sessionSecret,
+            ) as MrfJwtPayload
+
+            return SubmissionService.findSubmissionById(
+              decoded.prevSubmissionId,
+            )
+              .andThen((submission) => {
```
kevin9foong: question: should there be a check to assert that the submission type is Multirespondent? so that the type narrowing is not required?

src/app/modules/verification/verification.controller.ts
```diff
@@ -107,6 +115,67 @@ export const _handleGenerateOtp: ControllerHandler<
...
+                  mrfSubmission.workflowStep !== decoded.currentWorkflowStep
+                ) {
+                  logger.error({
+                    message: 'MRF JWT submission type mismatch',
+                    meta: {
+                      ...logMeta,
+                      submissionFormId: submission.form.toString(),
+                      requestFormId: formId,
+                      submissionType: submission.submissionType,
+                      submissionWorkflowStep: mrfSubmission.workflowStep,
```
kevin9foong: issue: The type narrowing reduces the type safety that TS provides for us. 

for example, if the check for submission type !==  SubmissionType.Multirespondent fails in the above if block and short-circuits, the logger here will access undefined value, which is fine here but if any future code uses mrfSubmssion.workflowStep + 1 for example, it will cause undesirable behavior.

src/app/modules/verification/verification.controller.ts
kevin9foong: suggestion to retain type safety: 

```
            return SubmissionService.getSubmissionMetadata(
              FormResponseMode.Multirespondent,
              formId,
              decoded.prevSubmissionId,
            )
              .andThen((foundSubmissionMeta) => {
                // Validate all MRF JWT conditions
                if (
                  !foundSubmissionMeta ||  
                  foundSubmissionMeta.mrf?.workflowCurrentStepNumber !== decoded.currentWorkflowStep
                ) {
                  logger.error({
                    message: 'MRF JWT submission type mismatch',
                    meta: {
                      ...logMeta,
                      requestFormId: formId,
                      submissionWorkflowStep: foundSubmissionMeta?.mrf?.workflowCurrentStepNumber, 
                      jwtWorkflowStep: decoded.currentWorkflowStep,
                    },
                  })
                  return errAsync(new MrfJwtError('MRF JWT validation failed'))
                }

                // MRF JWT is valid, skip SPCP/MyInfo verification
                return okAsync(form)
              })
```

how about reusing the logic `getSubmissionMetadata` method? doing so we are able to only retrieve submission if the formid, submission id matches and submission type is MRF and avoid typecasting.

src/app/modules/verification/verification.controller.ts
```diff
@@ -107,6 +115,67 @@ export const _handleGenerateOtp: ControllerHandler<
...
+
+                // Validate all MRF JWT conditions
+                if (
+                  submission.form.toString() !== formId ||
+                  submission.submissionType !==
+                    SubmissionType.Multirespondent ||
+                  mrfSubmission.workflowStep !== decoded.currentWorkflowStep
+                ) {
+                  logger.error({
+                    message: 'MRF JWT submission type mismatch',
```
kevin9foong: issue(minor): the cause for this logger error could be workflow step # mismatch, formid mismatch, but the error message says its type mismatch

src/app/modules/verification/__tests__/verification.controller.spec.ts
```diff
@@ -2015,6 +2023,279 @@ describe('Verification controller', () => {
...
+      })
+
+      const MOCK_MULTIRESPONDENT_SUBMISSION = {
+        _id: new ObjectId(),
+        form: new ObjectId(MOCK_FORM_ID),
+        submissionType: SubmissionType.Multirespondent,
+        workflowStep: MOCK_WORKFLOW_STEP,
+      } as IMultirespondentSubmissionSchema
+
+      it('should return 201 when MRF JWT is valid and all checks pass', async () => {
```
kevin9foong, line 2055: praise: thanks for adding strong tc for this addition!

frontend/src/features/verifiable-fields/VerifiableFieldService.ts
```diff
@@ -56,11 +57,14 @@ export const triggerSendOtp = async ({
...
   /** The value of the verification field to verify. Usually an email or phone number */
   answer: string
+  /** The id of the previous submission */
+  previousSubmissionId?: string
 }): Promise<SendFormOtpResponseDto> => {
   return ApiService.post(
     `${FORM_API_PREFIX}/${formId}/${VERIFICATION_ENDPOINT}/${transactionId}/fields/${fieldId}/otp/generate`,
     {
       answer,
+      previousSubmissionId,
```
kevin9foong, line 67: question(answered): wondering if this previousSubmissionId param is necessary - since we can just check if there is a mrf cookie always and use the value in the cookie instead

edit: for posterity, it is needed to lookup the correct cookie, since there can be multiple submissions for the same form.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -330,6 +335,25 @@ export const handleGetMultirespondentSubmissionForRespondent: ControllerHandler<
...
+          prevSubmissionId: submissionId,
+          currentWorkflowStep: responseData.workflowStep,
+        })
+
+        res.cookie(
+          getMrfCookieName({ formId, previousSubmissionId: submissionId }),
+          mrfCookie,
+          {
+            // Important for security - access token cannot be read by client-side JS
+            maxAge: spcpMyInfoConfig.spCookieMaxAge, // 3 hours
```
kevin9foong: suggestion: shall we remove the `3 hours` comment? since it can change when the maxAge config is changed in ssm params and become a source of outdated documentation.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -330,6 +335,25 @@ export const handleGetMultirespondentSubmissionForRespondent: ControllerHandler<
...
+        const mrfCookie = createMrfCookie({
+          prevSubmissionId: submissionId,
+          currentWorkflowStep: responseData.workflowStep,
+        })
+
+        res.cookie(
+          getMrfCookieName({ formId, previousSubmissionId: submissionId }),
+          mrfCookie,
+          {
+            // Important for security - access token cannot be read by client-side JS
```
kevin9foong: nit: perhaps this comment is not required

src/app/modules/submission/multirespondent-submission/multirespondent-submission.utils.ts
```diff
@@ -509,3 +514,34 @@ export const getResponsesDataFromMrfResponses = ({
...
+  currentWorkflowStep,
+}: MrfJwtPayload): string => {
+  const payload: MrfJwtPayload = {
+    prevSubmissionId,
+    currentWorkflowStep,
+  }
+
+  return jwt.sign(payload, config.sessionSecret, {
+    // this arg must be supplied in seconds
+    expiresIn: spcpMyInfoConfig.spCookieMaxAge / 1000, // 3 hrs
```
kevin9foong: suggestion:  remove the 3hrs comment as it can change and become outdated, becoming tech debt

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -330,6 +335,25 @@ export const handleGetMultirespondentSubmissionForRespondent: ControllerHandler<
           message: 'Get encrypted response using submissionId success',
           meta: logMeta,
         })
+
+        // Set MRF cookie with submission details when loading the form
+        const mrfCookie = createMrfCookie({
```
kevin9foong, line 340: praise: thanks for taking the time to discuss and find an ideal solution which maintains security of our platform for our users. this solution is much more secure and appreciate the effort making this happen!

kevin9foong:
Before merging this in, shall we also add our discussion on the alternatives we brought up (Eg, passing the submission Id directly from FE without validating, using the mrf secret key etc). i recall that lack of alts being documented was brought up during retro and we can start from this PR in documenting our decisions for knowledge transfer with the team (us and new team members)!

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
```diff
@@ -330,6 +335,25 @@ export const handleGetMultirespondentSubmissionForRespondent: ControllerHandler<
...
+        })
+
+        res.cookie(
+          getMrfCookieName({ formId, previousSubmissionId: submissionId }),
+          mrfCookie,
+          {
+            // Important for security - access token cannot be read by client-side JS
+            maxAge: spcpMyInfoConfig.spCookieMaxAge, // 3 hours
+            httpOnly: true,
+            sameSite: 'lax', // Setting to 'strict' prevents Singpass login on Safari, Firefox
```
eliotlim: should `sameSite` be strict?
this cookie is set on form.gov.sg for use on form.gov.sg only.

src/app/modules/verification/verification.errors.ts
scottheng96: Good idea - updated as per comment

src/app/modules/verification/verification.controller.ts
```diff
@@ -107,6 +115,67 @@ export const _handleGenerateOtp: ControllerHandler<
     FormService.retrieveFullFormById(formId)
       // Step 2: Verify SPCP/MyInfo, if form requires it
       .andThen((form) => {
+        // If previousSubmissionId exists, verify MRF JWT and skip SPCP/MyInfo
```
eliotlim: This is fine, I just wanted to make this reasoning very clear:
- Step 1 MRF calls to the OTP verification endpoint will only be checked if Singpass is enabled
- Step 2+ MRF calls to this endpoint will be allowed only if:
  - this is an edit of an existing previous submission with a valid step number (can't be replayed after submission)
  - TODO: if Singpass is enabled for this step

Crucially: this is implemented as a bypass of Singpass / Myinfo cookie verification if the step doesn't require it.
- When multi-step Singpass MRF is supported in the future, this condition will still hold.
- However, we should update this check to ensure that:
  - either the step requires Singpass / Myinfo (therefore access to the verification endpoint is allowed, through Myinfo cookie validation)
  - or the step belongs to a previous submission (therefore access to the verification endpoint is allowed, through the previous submission cookie)

In other words: access should be given for the right reason.

src/app/modules/verification/verification.controller.ts
scottheng96: Works as well! I've updated it to use `getSubmissionMetadata`, and changed the corresponding variables to check against the metadata object

src/app/modules/verification/verification.controller.ts
scottheng96: Agree - removed the 'type', and logs include all the variables compared against each other. Think this is cleaner than a separate logging error for each mismatch.

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
scottheng96: Removed this!

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
scottheng96: Removed this!

src/app/modules/submission/multirespondent-submission/multirespondent-submission.controller.ts
scottheng96: Ah ok - vs. Singpass cookie which is not set by form.gov.sg. Thanks for catching this - updated as per comment

src/app/modules/verification/verification.controller.ts
```diff
@@ -1,13 +1,18 @@
...
 
 import {
   ErrorDto,
   FormAuthType,
+  FormResponseMode,
   SendFormOtpResponseDto,
+  SubmissionType,
 } from '../../../../shared/types'
 import { SALT_ROUNDS } from '../../../../shared/utils/verification'
+import { IMultirespondentSubmissionSchema } from '../../../types'
```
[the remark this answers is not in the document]
scottheng96: removed this - dismissing finding

src/app/modules/verification/verification.controller.ts
scottheng96: Added an additional TODO comment for this so we don't miss it in the future

src/app/modules/verification/verification.controller.ts
```diff
@@ -107,6 +114,83 @@ export const _handleGenerateOtp: ControllerHandler<
...
+                    message: 'MRF JWT submission mismatch',
+                    meta: {
+                      ...logMeta,
+                      jwtSubmissionIdFromMetadata: decoded.prevSubmissionId,
+                      previousSubmissionId: previousSubmissionId,
+                      submissionWorkflowStepFromMetadata:
+                        foundMrfSubmissionMetadata?.mrf
+                          ?.workflowCurrentStepNumber,
+                      jwtWorkflowStep: decoded.currentWorkflowStep,
+                      test1:
```
kevin9foong: issue: what does test1 and test2 mean? it might be an issue for a dev looking at the meta as they will need to check the code. should it be more appropriately named?

src/app/modules/verification/verification.controller.ts
```diff
@@ -107,6 +114,83 @@ export const _handleGenerateOtp: ControllerHandler<
     FormService.retrieveFullFormById(formId)
       // Step 2: Verify SPCP/MyInfo, if form requires it
       .andThen((form) => {
+        // If previousSubmissionId exists, this means it is coming from a 2+ step MRF workflow
+        // verify MRF JWT and skip SPCP/MyInfo since Singpass is currently only enabled for MRF first step
+        // TODO: revisit this logic when Singpass is enabled for all MRF steps (fix is to run validation check if step is Singpass-enabled)
+        if (previousSubmissionId) {
+          const mrfCookie =
+            req.cookies[getMrfCookieName({ formId, previousSubmissionId })]
+          try {
```
kevin9foong: issue: ideally, we should be using neverthrow similar to src/app/modules/myinfo/myinfo.service.ts's verifyLoginJwt method. especially since the rest of the function is using neverthrow.

src/app/modules/verification/__tests__/verification.controller.spec.ts
```diff
@@ -2015,6 +2022,326 @@ describe('Verification controller', () => {
...
+          MOCK_PREVIOUS_SUBMISSION_ID,
+        )
+        expect(MockOtpUtils.generateOtpWithHash).toHaveBeenCalled()
+        expect(MockVerificationService.sendNewOtp).toHaveBeenCalledWith(
+          EXPECTED_PARAMS_FOR_SENDING_FORM_OTP,
+        )
+        expect(mockRes.status).toHaveBeenCalledWith(StatusCodes.CREATED)
+      })
+
+      it('should return 400 when MRF JWT formId does not match', async () => {
```
kevin9foong: question: is this TC still necessary? since if formId does not match, it should not be able to find a submission anymore

eliotlim: i think for the sake of future proofing, we should still test if both the `formId` and `prevSubmissionId` match

src/app/modules/verification/verification.controller.ts
scottheng96: My bad - left this in during local testing. Will remove

kevin9foong: also, should it be foundMrfSubmissionMetadata?.**refNo**.toString() instead of number?

src/app/modules/verification/__tests__/verification.controller.spec.ts
kevin9foong: yeah, but in this case after the code changes, the direct formId comparison is no longer being done and the JwtToken no longer has a formid, so this test case is no longer required.

src/app/modules/verification/verification.controller.ts
eliotlim: migrated from try / catch to neverthrow!

kevin9foong:
lgtm - contingent on our bug bash being all green!

kevin9foong (approved):

src/app/modules/verification/__tests__/verification.controller.spec.ts
kevin9foong: updated the tc name!

kevin9foong:
one final suggestion before merging - shall we rename the cookie to indicate what the cookie is for? eg, similar to myInfoCookie 

suggested name: eg, mrf_formId_prevSubmissionId

PR #9084, opened 2026-02-09
fix(bulk-pdf-download): disable next button if no download options selected (part 9)

kevin9foong:
## Problem

image
 
When no download option is selected, the next button is still clickable. It should be disabled.
Closes FRM-2262

## Solution

image

Disable the 'next' button until an option is selected 

**Breaking Changes** 

## Tests

- [ ] Go to the download button in the form response page
- [ ] Assert that the next button cannot be clicked until any combination of download options is selected.


scottheng96 (approved):

scottheng96:
LGTM!

PR #9106, opened 2026-02-11
fix: modal progress bar does not reflect pdf generation time

kevin9foong:
## Problem

The download progress bar does not include the progress and time taken for generating the PDFs locally. This causes the bar to be stuck at 100% for extended periods of time causing confusion to users. 

Closes FRM-2274

## Solution

If PDFs should be generated, double the total count and include the decryption and pdf in the progress. 

**Breaking Changes** 


frontend/src/features/admin-form/responses/ResponsesPage/storage/useDecryptionWorkers.ts
```diff
@@ -309,6 +312,8 @@ const useDecryptionWorkers = ({
                   })
                   pdfZip.file(pdfTitle, pdfBlob)
                 }
+                pdfGenerationProgress += 1
```
kevin9foong: this is safe since it is incremented regardless of if the decryption is successful - so it will always be able to hit 100%.

eliotlim (approved):

PR #9110, opened 2026-02-12
feat: update pdf filenames for local pdf gen and email attachment

kevin9foong:
## Problem

The filenames in local pdf is too long and provides too much info. 
Also, the pdf filenames in the email attachments are not standardized with the local pdf download. 

Closes FRM-2275. 

## Solution

Standardize the names to `RefNo {submissionId}` following the attachment download convention. 

**Breaking Changes** 

## Tests

PDF bulk local download: 
- [ ] Bulk download the pdfs from the results page. 
- [ ] The pdf name should be similar to the screenshot. 
image

PDF single local download: 
- [ ] Select a single submission and download the PDF. 
- [ ] The pdf name should be `RefNo {submissionId}`. Verify the submission id matches the selected response. 
image

Email copy attachment:
- [ ] Make a storage mode form with email autoreply respondent + admin notification with signature field.
- [ ] Make a submission. 
- [ ] Assert that the emails received to the respondent and admin are named `RefNo {submissionId}` instead of `response.pdf` previously. 
- [ ] Do the same for MRF with an email autoreply respondent and workflow completion email. 
- [ ] Assert that the emails received to the respondent and admin are named `RefNo {submissionId}` instead of `response.pdf` previously.

### DIFF
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/utils/generateResponsePdf.tsx
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/utils/generateResponsePdf.tsx
@@ -54,11 +54,7 @@ const renderPrintableResponse = (
   }
 }
 
-const getPdfTitle = ({
-  submissionId,
-}: {
-  submissionId: string
-}) => {
+const getPdfTitle = ({ submissionId }: { submissionId: string }) => {
   return `RefNo ${submissionId}.pdf`
 }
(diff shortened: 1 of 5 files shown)


eliotlim (approved):
praise: thank you for breaking down the changes to the locally-generated / email-attached PDFs and the test cases by commits! much clearer 💯

PR #9126, opened 2026-02-24
feat: remove deprecated autoreply (part 1/3 of standardize email payload hidden fields)

kevin9foong:
Part 1 of 3 (standardise email payload hidden fields): [#9126, #9128, #9129]

## Problem

`EmailSubmissionObj.autoReplyData` is deprecated and no longer being used. (For context, it was used for storage mode autoreply previously, but no longer being used since respondent and admin email payload is standardized. mrf is not affected as it is using a separate flow (seen in diagram))
This is since the payload sent to admins and respondents are now standardized to be the same (see table below). Also, masking is no longer being done and hence the stale code can be removed. 

### Updated Email Behavior (Standardized)

| :--- | :--- | :--- | :--- |

No intended changes to MRF, which is already working as intended without issue in production. 

### Diagram of current usage (as of and after this PR): 
image
Looking at our current implementation (in red), we are only using `formData` and `autoReplyData` is no longer being used. Hence, it is safe to remove. 

In part 1, we remove stale autoReplyData types which are not being used for sending any email. 

Note that for MRF, the responses are already pre-pruned for hidden logic fields when sent to the backend, hence no changes are needed. It also follows a different flow.  

Closes FRM-2293

## Solution

Remove deprecated autoreply formatting and its associated test cases and functions. The code should not be currently used. 

**Breaking Changes** 

## Tests

- [ ] Assert the removed code paths and tests are not being used.

### DIFF
--- a/__tests__/unit/backend/helpers/generate-email-data.ts
+++ b/__tests__/unit/backend/helpers/generate-email-data.ts
@@ -1,23 +1,11 @@
 import { pick } from 'lodash'
 
 import { ProcessedSingleAnswerResponse } from 'src/app/modules/submission/submission.types'
-import {
-  EmailAdminDataField,
-  EmailDataCollationToolField,
-  EmailRespondentConfirmationField,
-} from 'src/types'
+import { EmailAdminDataField, EmailDataCollationToolField } from 'src/types'
 
 export const generateSingleAnswerJson = (
   response: ProcessedSingleAnswerResponse,
 ): EmailDataCollationToolField => pick(response, ['question', 'answer'])
-
-export const generateSingleAnswerAutoreply = (
-  response: ProcessedSingleAnswerResponse,
-): EmailRespondentConfirmationField => ({
-  question: response.question,
-  answerTemplate: response.answer.split('\n'),
-  fieldType: response.fieldType,
-})
 
 export const generateSingleAnswerFormData = (
   response: ProcessedSingleAnswerResponse,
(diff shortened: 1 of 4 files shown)


kevin9foong:
@copilot describe the changes here for a reviewer to understand

__tests__/unit/backend/helpers/generate-email-data.ts
```diff
@@ -1,24 +1,12 @@
...
 
 import { ProcessedSingleAnswerResponse } from 'src/app/modules/submission/submission.types'
-import {
-  EmailAdminDataField,
-  EmailDataCollationToolField,
-  EmailRespondentConfirmationField,
-} from 'src/types'
+import { EmailAdminDataField, EmailDataCollationToolField } from 'src/types'
 
 export const generateSingleAnswerJson = (
```
scottheng96, line 6: (A bit of scope creep) I don't believe this is being used as well - think we can remove this

scottheng96 (approved):

__tests__/unit/backend/helpers/generate-email-data.ts
kevin9foong: okay, let me look into this - thanks for bringing it up!

kevin9foong: heya @scottheng96 , had a look at what these functions do and it seems they are being used for test cases mainly for email mode. since the email mode functions are not being cleaned up in this PR (out of scope since this is just focusing on storage mode email payload), opting to leave in these helper functions as well and should be removed together with the email mode controller/service. 

This should probably be done in a separate PR.

PR #9128, opened 2026-02-24
feat: replace admin email data type with generic email data type  (part 2/3 of standardize email payload hidden fields)

kevin9foong:
Part 2 of 3 (standardise email payload hidden fields): [#9126, #9128, #9129]

## Problem

The concept of a specific "admin email data" type no longer exists, since the respondent data and admin data have been standardized to be the same. 
Hence, we should make the admin email data type generic. 

### Updated behavior for email payload 
| :--- | :--- | :--- | :--- |

As seen in this table, the email body and pdf copy have the exact same fields, which is standardized. 

Closes FRM-2294. 

### Diagram of current flow in red:  
image

We are moving towards the common `emailResponses` type in green shown in the diagram - which is used by both sending email to admin, respondent and the shared PDF. This happens to be the formData which type name is now updated to a generic `EmailDataField` instead of specific `AdminEmailDataField`. 

The following specific changes are made in this PR: 
1. Rename the `AdminEmailDataField` to `EmailDataField`. 
2. When sending email to respondent, change the type to accept the generic `EmailDataField` instead of using the old specific respondent email data type. 

**Breaking Changes** 

## Tests

TC1: Storage mode forms (Regression) 
- [ ] Create a storage mode form with admin submission email and respondent email copy with PDF and field summary. 
- [ ] Make a submission and ensure the email with PDF can be received for both admin and respondent. The output should be the same as previous.

### DIFF
--- a/src/app/modules/submission/submission.service.ts
+++ b/src/app/modules/submission/submission.service.ts
@@ -22,7 +22,7 @@ import {
   SubmissionType,
 } from '../../../../shared/types'
 import {
-  EmailRespondentConfirmationField,
+  EmailDataField,
   IEncryptSubmissionModel,
   IMultirespondentSubmissionModel,
   IPopulatedForm,
@@ -458,7 +458,7 @@ export const sendEmailConfirmations = <S extends ISubmissionSchema>({
 }: {
   form: IPopulatedForm
   submission: S
-  responsesData: EmailRespondentConfirmationField[]
+  responsesData: EmailDataField[]
   submissionAttachments?: Mail.Attachment[]
   recipientData: AutoReplyMailData[]
   pdfAttachment?: Mail.Attachment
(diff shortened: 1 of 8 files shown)


src/app/modules/submission/submission.service.ts
```diff
@@ -458,7 +458,7 @@ export const sendEmailConfirmations = <S extends ISubmissionSchema>({
 }: {
   form: IPopulatedForm
   submission: S
-  responsesData: EmailRespondentConfirmationField[]
```
kevin9foong, line 461: note: we replace the use of the autoreplydata pdf type with the generic email data field type here. 

the rest of the code changes are re-naming of EmailAdminDataField to the generic EmailDataField to be used for both respondent and admin email / pdf.

scottheng96:
We are merging this branch in develop and not `feat/remove-deprecated-autoreply` which was part 1 right?

scottheng96 (approved):

kevin9foong:
> We are merging this branch in develop and not `feat/remove-deprecated-autoreply` which was part 1 right?

yep! but i put the dest as `feat/remove-deprecated-autoreply` so that the diffs for this specific pr are reflected. it will auto update to `develop` once the prior part 1 is merged

PR #9129, opened 2026-02-24
fix: prune hidden logic fields from email data field (part 3/3 of standardize email payload hidden fields)

kevin9foong:
Part 3 of 3 (standardise email payload hidden fields): [#9126, #9128, #9129]

## Problem

The respondent and admin email and PDF contain hidden logic fields, which should not be included and should be removed from the email payload.  

### Updated behavior for email payload 
| :--- | :--- | :--- | :--- |

Closes FRM-2264.

| Before | After |
| :-- | --: |
| image | image |

### Diagram of current flow in red:  
image

In part 2, we updated `formData` towards the common `emailResponses` type in green shown in the diagram by updating the functions to use the generic `EmailFieldData` type. 

In this PR, we update `formData` to prune the hidden logic fields, achieving our intended updated behavior to remove the hidden logic fields in admin & respondent email payload (shown in the table above) 

**Breaking Changes** 

## Tests

TC1: Storage mode respondent email and admin email exclude hidden logic fields 
- [ ] Make a storage mode form with admin email notification & respondent email copy (with PDF and field summary). 
- [ ] Add logic to the form such that fields can be hidden. 
- [ ] Submit the form with hidden fields. 
- [ ] Assert the respondent copy email and pdf does not include the hidden fields. 
- [ ] Assert the admin email and pdf does not include the hidden fields. 
- [ ] Resubmit the form showing the hidden fields. 
- [ ] Assert the respondent copy email and pdf includes the hidable logic fields. 
- [ ] Assert the admin email and pdf includes the hidable logic fields. 

TC2: MRF 
- [ ] Repeat the above TC1 for MRF with 1 step.
- [ ] Repeat the above TC1 with MRF with 2 steps, where the fields can be hidden by the second step submitter.

### DIFF
--- a/src/app/modules/submission/email-submission/__tests__/email-submission.util.spec.ts
+++ b/src/app/modules/submission/email-submission/__tests__/email-submission.util.spec.ts
@@ -64,18 +64,34 @@ describe('email-submission.util', () => {
       '',
     ) as ProcessedFieldResponse
 
+    const response3 = getResponse(
+      String(new ObjectId()),
+      '',
+    ) as ProcessedFieldResponse
+
+    const response4 = getResponse(
+      String(new ObjectId()),
+      '',
+    ) as ProcessedFieldResponse
+
     response1.fieldType = BasicField.YesNo
-    response2.fieldType = BasicField.YesNo
+    response2.fieldType = BasicField.ShortText
     response1.isVisible = true
-    response2.isVisible = false
+    response2.isVisible = true
+    response3.fieldType = BasicField.LongText
+    response3.isVisible = false
+    response4.fieldType = BasicField.Mobile
+    response4.isVisible = undefined
 
     const hashedFields = new Set([
+      new ObjectId().toHexString(),
+      new ObjectId().toHexString(),
       new ObjectId().toHexString(),
       new ObjectId().toHexString(),
     ])
     const authType = FormAuthType.NIL
     const submissionEmailObj = new SubmissionEmailObj(
-      [response1, response2],
+      [response1, response2, response3, response4],
       hashedFields,
       authType,
     )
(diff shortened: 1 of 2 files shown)


src/app/modules/submission/email-submission/__tests__/email-submission.util.spec.ts
```diff
@@ -64,18 +64,35 @@ describe('email-submission.util', () => {
       '',
     ) as ProcessedFieldResponse
 
+    const response3 = getResponse(
```
kevin9foong, line 67: note: these tc cover when there are multiple responses.

src/app/modules/submission/email-submission/__tests__/email-submission.util.spec.ts
```diff
@@ -110,6 +127,61 @@ describe('email-submission.util', () => {
       ])
     })
 
+    it('should exclude non-visible fields from formData', () => {
```
kevin9foong, line 129: note: these tc cover single response and how they are specifically pruned based on the isVisible status (true, false, undefined)

scottheng96 (approved):

scottheng96:
LGTM! Thanks for splitting up the work into smaller, readable PRs.

src/app/modules/submission/email-submission/__tests__/email-submission.util.spec.ts
```diff
@@ -110,6 +126,61 @@ describe('email-submission.util', () => {
...
+      })
+
+      const emailData = new SubmissionEmailObj(
+        [response],
+        new Set(),
+        FormAuthType.NIL,
+      )
+
+      expect(emailData.dataCollationData).toEqual([
+        generateSingleAnswerJson(response),
```
scottheng96, line 177: nit: Think we can remove this - since we don't use this function anymore. Was that scope creep from part 1 - saw this in multiple places

kevin9foong: hey @scottheng96 , these helper functions are also used for test cases which test email mode form service/controller. Since which these email mode controllers/services still exists in our db, the test cases verifying it and hence the helpers should still be included - and should only be removed when the email mode service/controllers it is testing are removed from the code. 

In this case, removing of the email mode form controllers/services is out of scope of this PR since this current PR only aims to remove the concept of autoReply email notifications (which are unrelated to email mode forms despite having the word email in them) and not email mode controllers/services (eg, hashSubmission etc)

PR #185, opened 2020-08-21
Introduce minimum test coverage thresholds, coveralls.io for threshold reporting and repo badge

liangyuanruo:
## Problem

We want to get better test coverage reports and enforce a minimum threshold of coverage as we continue to refactor the codebase.

Closes #113

## Solution

Introduce coveralls.io which integrates with Jest for test reporting, and configure Jest to enforce minimum test coverage (currently 12%). This figure should be progressively raised as more tests get ported from Jasmine to Jest.

**Features**:

- Test coverage reports available on [coveralls.io](https://coveralls.io/github/opengovsg/FormSG) and locally.

**Improvements**:

- Enforces test coverage to 12%

## Screenshots

image

## Deploy Notes

_Notes regarding deployment of the contained body of work. These should note any
new dependencies, new scripts, etc._

**New dev dependencies**:

- `coveralls` : Reporting tool for uploading Jest's coverage report to coveralls.io

PR #9143, opened 2026-02-26
feat: remove unused coveralls

kevin9foong:
## Problem

Address failing BE tests due to coveralls.io is down for >12 hours. Upon further look, we are not using code coverage or coveralls.io reports to make testing decisions in the past >1.5 years. Hence, we should remove this to clear bloat and improve CI speeds. 

**Chesterson's fence:** 
For context, coveralls was added in 2020 https://github.com/opengovsg/FormSG/pull/185 mainly to track code coverage during migration from jasmine to jest. Jest was configured to a min of 12% code coverage and this 12% is not being removed in this PR. 
However, overall code coverage % has since exceeded 12% post successful migration to jest and this enforcement was since removed >2 years ago - coverage has not been referred to as a metric since then and is hence safe to remove.

Also, our test coverage from vitest is already piped to Datadog under the org's test coverage dashboard - hence no regressions are observed. 

## Solution

Remove failing coveralls from backend_tests.
Note that coveralls is being used in severless CI, but they are not failing and out of scope for this PR.  

**Breaking Changes** 


eliotlim (approved):
agreed, the failing ci step is an issue for our pipeline.

let's relook at ci coverage badges at a future date.

PR #9146, opened 2026-02-26
chore(cleanup): remove puppeteer from backend and cleanup unused local pdf generation

kevin9foong:
Notes to reviewer: 
1. please use review-by-commits
2. deployed to stg-alt3.form.gov.sg for your verification 

## Problem

This is an initiative to reduce our engineering load by reducing the packages we need to maintain. Puppeteer-core is a dependency with many transitive dependencies we need to maintain, and should be removed if unused (eg, dep alerts: https://github.com/opengovsg/FormSG/pull/9141) 

This PR removes 28 transitive dependencies we need to maintain version bumps for. 

We have previously [pinned our severless PDF generation](https://github.com/opengovsg/FormSG/issues/8858) without the ability to switch using feature flags. 
Hence, puppeteer-core is no longer used and can be safely removed.  
 
## Solution

Remove puppeteer core and isUseLambdaOutput which does not toggle anything since we [pinned our severless PDF generation](https://github.com/opengovsg/FormSG/issues/8858). 

**Breaking Changes** 

## Tests: 

TC1: PDF still works
- [x] Add email notification to a storage mode form. 
- [x] Make a submission and ensure the email is received. 

TC2: Payment proof can still be downloaded
- [x] Submit a payments form. 
- [x] Ensure that payment proof can be downloaded

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -89,7 +89,6 @@
         "perfect-freehand": "^1.2.2",
         "promise-retry": "^2.0.1",
         "promise-timeout": "^1.3.0",
-        "puppeteer-core": "24.9.0",
         "react": "^18.3.0",
         "react-dom": "^18.3.1",
         "react-email": "^4.0.11",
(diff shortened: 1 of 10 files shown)


eliotlim (approved):
LGTM! 👍

PR #9152, opened 2026-02-26
feat: bump sqs producer and consumer which uses aws sdk v3 under the hood

kevin9foong:
## Problem

Closes dep bumps for [sqs-consumer](https://github.com/opengovsg/FormSG/pull/8795) and [sqs-producer](https://github.com/opengovsg/FormSG/pull/8653)

Note that the updates use the new AWS SDK v3 from the previous AWS SDK v2 which is EOL since Sept 2025.  

## Solution

Since [keep alive is now default enabled for AWS SDK v3](https://aws.amazon.com/blogs/developer/http-keep-alive-is-on-by-default-in-modular-aws-sdk-for-javascript/), we can safely remove the sqs argument which aims to keep the SQS session alive previously recommended in [v5.7.0 of sqs-consumer](https://github.com/bbc/sqs-consumer/tree/690bcf5b17f905dcc8b9b6820e918b851e6fe257). 

**Breaking Changes** 

## Tests

Webhook can be received: 
- [x] Submit a form with webhook, to https://webhook.site/. 
- [x] Assert the webhook can be received. 

Webhook with retries can be received: 
- [x] Submit a form with webhook with retries enabled, set an invalid endpoint. 
- [x] Update the endpoint to a valid https://webhook.site/ and assert it can be received.


apps/backend/src/app/modules/webhook/webhook.consumer.ts
```diff
@@ -90,7 +75,7 @@ export const startWebhookConsumer = (
  */
 export const createWebhookQueueHandler =
   (producer: WebhookProducer) =>
-  async (sqsMessage: aws.SQS.Message): Promise<void> => {
+  async (sqsMessage: Message): Promise<Message | undefined> => {
```
kevin9foong, line 78: Note: Instead of returning resolved promise to remove the message from queue and acknowledge, the new sqs-consumer expects the original message to be returned.

apps/backend/src/app/modules/webhook/webhook.consumer.ts
```diff
@@ -37,20 +36,6 @@ export const startWebhookConsumer = (
     queueUrl,
     region: config.aws.region,
     handleMessage: createWebhookQueueHandler(producer),
-    // By default, the default Node.js HTTP/HTTPS SQS agent
```
kevin9foong, line 40: note: can be safely removed, since the new sqs-consumer uses AWS SDK v3 under the hood, which [HTTP keep-alive is on by default](https://aws.amazon.com/blogs/developer/http-keep-alive-is-on-by-default-in-modular-aws-sdk-for-javascript/).

eliotlim:
Since the monorepo migration has changed the structure, I'll be rebasing and updating this for the new monorepo structure and packages.

eliotlim (approved):
tested working on stg-alt2! 🎉 

Image

PR #9162, opened 2026-03-02
feat: migrate serverless virus scanner to pnpm

scottheng96:
## Problem

Closes FRM-2302

## Solution

1. Followed steps to migrate from npm to pnpm
2. Updated github/ci deployment scripts
3. Resolved dependencies
- because pnpm uses a stricter dependecy resolution (vs. npm which lenient phantom dependencies)

4. (unrelated) chore to remove unused arg (IS_LAMBDA) in `Dockerfile`

**Breaking Changes** 


.github/workflows/ci.yml
```diff
@@ -179,7 +179,11 @@ jobs:
           path: '**/node_modules'
           key: ${{ runner.OS }}-node-modules-${{ hashFiles('**/package-lock.json') }}
           fail-on-cache-miss: true
-      - run: npm ci --prefix serverless/virus-scanner-guardduty
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.30.3
+      - run: pnpm install --frozen-lockfile --prefix serverless/virus-scanner-guardduty
```
kevin9foong: praise: very nice using --frozen-lockfile for our CI!

serverless/virus-scanner-guardduty/src/logger.ts
```diff
@@ -3,7 +3,7 @@ import pino, { Logger, LoggerOptions as PinoLoggerOptions } from 'pino'
 
 type LoggerOptions = {
   service: string
-  prettyPrint?: ConstrainBoolean
+  prettyPrint?: boolean
```
kevin9foong, line 6: question: curious, why was this updated?

serverless/virus-scanner-guardduty/docker-compose.dev.yml
```diff
@@ -4,8 +4,6 @@ services:
     build:
       context: ./
       dockerfile: Dockerfile
-      args:
```
kevin9foong, line 7: praise: nice work spotting this can be removed and cleaning it up!

kevin9foong:
hey @scottheng96, lets also update the virus scanner script in the root package.json as well!

.github/workflows/aws-deploy-scanner-guardduty-iac.yml
```diff
@@ -48,26 +48,34 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: '.nvmrc'
-      - name: Cache Node.js modules
+      - name: Setup pnpm
```
kevin9foong: info: 

the pnpm docs suggests this - wonder if we can use this from their docs? 

```
GitHub Actions
On GitHub Actions, you can use pnpm for installing and caching your dependencies like so (belongs in .github/workflows/NAME.yml):

.github/workflows/NAME.yml
name: pnpm Example Workflow
on:
  push:

jobs:
  build:
    runs-on: ubuntu-22.04
    strategy:
      matrix:
        node-version: [20]
    steps:
      - uses: actions/checkout@v4
      - name: Install pnpm
        uses: pnpm/action-setup@v4
        with:
          version: 10
      - name: Use Node.js ${{ matrix.node-version }}
        uses: actions/setup-node@v4
        with:
          node-version: ${{ matrix.node-version }}
          cache: "pnpm"
      - name: Install dependencies
        run: pnpm install
```

kevin9foong: source: https://pnpm.io/continuous-integration

serverless/virus-scanner-guardduty/Dockerfile
```diff
@@ -13,22 +13,17 @@ RUN apt-get update && \
   automake \ 
   python3
 
+# Install pnpm
+RUN corepack enable && corepack prepare pnpm@10.30.3 --activate
+
 WORKDIR /function
 
 # install app dependencies
-COPY package.json package-lock.json ./
```
kevin9foong, line 19: praise: v nice removing debt!

serverless/virus-scanner-guardduty/src/logger.ts
scottheng96: Pino was set as v7 in `devDependency`, which supports `ConstrainBoolean`. Used to work previously because npm would flatten the dependencies and support this type. I believe pnpm instead uses a non-flattened structure, so `devDependencies` are not hoisted into my runtime dependencies.

v8 didn't have this so I had to update it! also referenced other team's [code](https://github.com/opengovsg/social360/blob/1365ff8e2122ee1462c5f88d27ba92662ddfadab/packages/shared/src/logging.ts#L6)

serverless/virus-scanner-guardduty/package.json
```diff
@@ -26,8 +27,8 @@
     "@types/aws-lambda": "^8.10.101",
     "@types/convict": "^6.1.1",
     "@types/jest": "^29.5.1",
+    "@types/lodash": "^4.17.24",
     "@types/node": "^18.0.1",
-    "@types/pino": "7.0.5",
```
kevin9foong, line 30: question: just to double check, why is types/pino removed?

kevin9foong:
mainly lgtm - we can merge this once we fix: 
1. remove caching of pnpm 
2. check why @types/pino was removed
3. ci passes and virus scanner can deploy 

for posterity, rationale for 1: 
```
ive had a look at the caching requirement for pnpm - lets not add worry about github actions caching of pnpm node_modules in phase 1 to save on unneeded work!

i think for our ci scripts, for now we dont need to add caching support first to save work - since the caching will need to be updated to the below when we swap to monorepo in phase 2

when we move to mono repo in phase 2, there will be one main node_modules in the root which we will cache, each package will have its own node_modules but they are just symlinks to the main node_modules which we dont need to cache since they install fast.

hence, the recommendation in the docs
```

kevin9foong:
hey @scottheng96 , can you help me double check if https://pnpm.io/cli/install --frozen-lockfile is needed? apparently, it is auto activated in ci environments and can detect if its GH actions! 

https://github.com/watson/ci-info#supported-ci-tools is being used under the hood of pnpm to detect ci envs

serverless/virus-scanner-guardduty/package.json
scottheng96: https://github.com/opengovsg/FormSG/pull/9162#discussion_r2870680688

Same reason as above

scottheng96:
Okay nice, I tested with `package.json` and `pnpm-lock.yaml` out of sync, without the `--frozen-lockfile` and it seems to be failing ci builds. Safe to remove!

image

serverless/virus-scanner-guardduty/pnpm-workspace.yaml
```diff
@@ -0,0 +1,8 @@
+ignoredBuiltDependencies:
```
kevin9foong: issue: we dont want to make the virus scanner the root of its own monorepo, we should be placing this in the package.json file for now

.github/workflows/aws-deploy-scanner-guardduty-iac.yml
```diff
@@ -38,36 +38,35 @@ jobs:
     name: virus scanner guardduty serverless
     runs-on: ubuntu-latest
     environment: ${{ inputs.gha-environment }}
+    strategy:
+      matrix:
```
kevin9foong: question: why do we need this?

serverless/virus-scanner-guardduty/package.json
```diff
@@ -44,5 +45,12 @@
     "ts-node-dev": "^2.0.0",
     "type-fest": "^4.17.0",
     "typescript": "^5.4.5"
+  },
+  "pnpm": {
```
kevin9foong, line 49: praise: thanks for limiting to only specifically required deps!

.github/workflows/aws-deploy-scanner-guardduty-iac.yml
kevin9foong: issue: since the matrix only has 1 entry, it is not required, we do not need to re-run our job with different node versions. lets remove this matrix 

```
name: Use Node.js ${{ matrix.node-version }}
        uses: actions/setup-node@v4
```

we can just use the same as the main `ci.yml` to keep things consistent: 
```
 - name: Use Node.js
        uses: actions/setup-node@v4
        with:
          node-version: 22
```

.github/workflows/aws-deploy-scanner-guardduty-iac.yml
```diff
@@ -38,36 +38,35 @@ jobs:
...
           fetch-depth: 0
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
         with:
-          node-version-file: '.nvmrc'
-      - name: Cache Node.js modules
-        uses: actions/cache@v4
+          version: 10.30.3
```
kevin9foong, line 50: question: wondering if we should pin to the major version instead of the specific patch

kevin9foong (approved):
thanks for the good work - approved! 👍

PR #9163, opened 2026-03-02
feat: migrate shared to pnpm

kevin9foong:
## Problem

Closes FRM-2304

Solution
- Followed steps to migrate from npm to pnpm
- Updated github/ci deployment scripts, skipped caching since they will be updated to use the root directory's 
- Resolved dependencies (added `eslint: 8.57.0` to dep dependencies)


shared/package.json
```diff
@@ -14,12 +14,16 @@
...
     "zod": "^3.11.6"
   },
+  "scripts": {
+    "preinstall": "npx only-allow pnpm"
+  },
   "devDependencies": {
     "@types/lodash": "^4.14.186",
     "@typescript-eslint/eslint-plugin": "^8.19.1",
     "@typescript-eslint/parser": "^8.19.1",
+    "eslint": "^8.57.0",
```
kevin9foong, line 21: note: required since, when the root runs ESLint on shared/, the config and plugins are resolved from shared/. pnpm installs ESLint 9 there as a transitive dependency if the eslint devDependency is not specified, but @typescript-eslint 8.x expects ESLint 8’s rule API, so the rule crashes. Adding eslint@^8.57.0 in shared pins shared to ESLint 8 and removes that mismatch.

.github/workflows/ci.yml
```diff
@@ -50,6 +50,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10
```
scottheng96: Should we specify 10.30.3? I think good to explicitly use the [latest version ](https://www.npmjs.com/package/pnpm)

.github/workflows/ci.yml
```diff
@@ -212,7 +216,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: '**/node_modules'
-          key: ${{ runner.OS }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+          key: ${{ runner.OS }}-node-modules-${{ hashFiles('**/package-lock.json', '**/pnpm-lock.yaml') }}
```
scottheng96, line 223: Are we deferring caching?

scottheng96 (approved):

.github/workflows/ci.yml
kevin9foong: yes - we will only do caching once we set up the monorepo 

rationale: 
```
ive had a look at the caching requirement for pnpm - lets not add worry about github actions caching of pnpm node_modules in phase 1 to save on unneeded work!

i think for our ci scripts, for now we dont need to add caching support first to save work - since the caching will need to be updated to the below when we swap to monorepo in phase 2

when we move to mono repo in phase 2, there will be one main node_modules in the root which we will cache, each package will have its own node_modules but they are just symlinks to the main node_modules which we dont need to cache since they install fas
```

kevin9foong: btw, caching is still being done here for the node_modules, but the pnpm install cache is not being done

.github/workflows/ci.yml
kevin9foong: yeah, good to pin to a version

PR #9164, opened 2026-03-02
feat: migrate backend to pnpm

kevin9foong:
## Problem

Closes FRM-2305

Solution
- Update dockerfiles to use pnpm and package.json 
- Added onlyBuiltDependencies to allow post install scripts, which were allowed by default for previous npm


kevin9foong:
issues with bcrypt: 
```
Error: Cannot find module '/opt/formsg/node_modules/.pnpm/bcrypt@5.1.1/node_modules/bcrypt/lib/binding/napi-v3/bcrypt_lib.node'
```
- https://github.com/vercel/next.js/discussions/76822

kevin9foong:
note: safe to remove lockfile-lint due to [pnpm's security](https://github.com/lirantal/lockfile-lint?tab=readme-ov-file#what-about-pnpm-support)

.github/workflows/ci.yml
```diff
@@ -62,14 +62,12 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ env.NODE_VERSION }}
-          cache: 'npm'
```
kevin9foong, line 65: note: removing first since we are switching to pnpm, will cache once phase II is started

src/app/services/mail/mail.types.ts
```diff
@@ -46,7 +46,7 @@ export type MailServiceParams = {
   transporter?: Mail
   senderMail?: string
   officialMail?: string
-  retryParams?: Partial<OperationOptions>
+  retryParams?: OperationOptions
```
kevin9foong, line 49: note: required since the deps are now explicitly installed and checking occurs

.github/workflows/ci.yml
```diff
@@ -213,18 +211,19 @@ jobs:
...
       # Load cached node_modules
       - name: Cache Node modules
         uses: actions/cache@v4
         with:
           path: '**/node_modules'
           key: ${{ runner.OS }}-node-modules-${{ hashFiles('**/package-lock.json', '**/pnpm-lock.yaml') }}
           fail-on-cache-miss: true
-      - run: npm run lint-ci
-      - run: npm_config_mode=yes npx lockfile-lint --type npm --path package.json --validate-https --allowed-hosts npm
+      - run: pnpm lint-ci
```
kevin9foong, line 229: note: safe to remove since [pnpm is not vulnerable to security issues this lockfile-lint script aims to protect against](https://github.com/lirantal/lockfile-lint?tab=readme-ov-file#what-about-pnpm-support).

scottheng96: ^Curious about why removing this is ok. Could you update the link? Think it links back to the PR haha

Dockerfile.production
```diff
@@ -17,24 +21,14 @@ COPY package.json package-lock.json ./
...
-# --legacy-peer-deps flag
-# A breaking change in the peer dependency resolution strategy was introduced in
-# npm 7. This resulted in npm throwing an error when installing packages:
-# npm ERR! code ERESOLVE
-# npm ERR! ERESOLVE unable to resolve dependency tree
-# See also:
-# * https://stackoverflow.com/questions/66239691/what-does-npm-install-legacy-peer-deps-do-exactly-when-is-it-recommended-wh
-# NOTE: This flag is used again later in the build process when calling npm prune.
-RUN npm ci --legacy-peer-deps
+RUN pnpm install --frozen-lockfile
```
scottheng96, line 24: I think safe to remove `--frozen-lockfile` since BE also uses ci-info

.github/workflows/ci.yml
kevin9foong: fixed the link!

package.json
```diff
@@ -52,7 +52,7 @@
     "prepare": "husky",
     "pre-commit": "lint-staged",
     "storybook": "npm run --prefix frontend storybook",
-    "postinstall:frontend": "npm --prefix frontend install",
+    "postinstall:frontend": "npm --prefix frontend install --legacy-peer-deps",
```
scottheng96, line 55: This looks like a FE script command - was there an issue with the build without it? Asking for gotchas when I do the FE migration.

package.json
```diff
@@ -112,12 +114,13 @@
     "jose": "^5.9.6",
     "jsdom": "^25.0.1",
     "json-stringify-safe": "^5.0.1",
-    "JSONStream": "^1.3.5",
```
scottheng96, line 115: nit: saw this this was moved up a couple lines - might be better for packages to be sorted alphabetically and case-insensitive

package.json
kevin9foong: yes, legacy peer deps is used the `dockerfile` and the deps that have issues with peer dep conflicts seem to only be the FE deps. 
pnpm also resolves deps differently so it doesnt have this flag anymore - so differing the legacy peer deps resolution to the FE install.

package.json
kevin9foong: hmm, this seems to be an opinionated default from pnpm - it auto updated when i ran pnpm install 

when pnpm add or installing - pnpm will update the ordering to its default to enforce an alphabetical order to keep the file clean and prevent unnecessary merge conflicts in Git

Dockerfile.production
kevin9foong: looking at [ci-info](https://github.com/watson/ci-info) from pnpm, im not seeing how it can spot that its a ci job when run within an isolated dockerfile - could i check if im missing something?

kevin9foong:
Tested on stg-alt.form.gov.sg

Dockerfile.production
scottheng96: Let's leave it in.

scottheng96 (approved):

PR #9171, opened 2026-03-03
feat: install workspaces scaffold

kevin9foong:
## Problem

Scaffold setup required to work on various parts of the monorepo migration in parallel.  

## Solution

- Remove installations from postinstall scripts for pnpm managed projected - as their topological installation is managed by pnpm workspaces.
- Install pnpm-workspaces.yaml file
- Create a placeholder backend package. 
- Update package names to follow consistent `formsg` prefix
- Update packages to be private, since they are not expected to be published
- Add existing packages to workspace, except for root src/, which will be migrated later. 
- Add pnpm caching to ci scripts.  

**Breaking Changes** 

## Tests 
TC1: Submitted attachments files can be decrypted 
- [ ] Submit a form with attachment 
- [ ] Download the attachment in the admin page and ensure it is downloaded correctly and the file matches what is uploaded.  

### Credits 
Credits to scott for finding out the cause of the virus scan deployment issue! 

Closes FRM-2312

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "formsg-backend",
+  "version": "6.311.0",
+  "private": true,
+  "homepage": ".",
+  "scripts": {},
+  "dependencies": {},
+  "devDependencies": {}
+}
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/mutations.ts
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/mutations.ts
@@ -36,7 +36,7 @@ export const useMutateDownloadAttachments = () => {
     }) => {
       const byteArray = await downloadAndDecryptAttachment(url, secretKey)
       if (!byteArray) throw new Error('Invalid file')
-      FileSaver.saveAs(new Blob([byteArray as BlobPart]), fileName)
+      FileSaver.saveAs(new Blob([byteArray]), fileName)
       return fileName
     },
     {
--- a/frontend/src/features/admin-form/responses/IndividualResponsePage/mutations.ts
+++ b/frontend/src/features/admin-form/responses/IndividualResponsePage/mutations.ts
@@ -36,7 +36,7 @@ export const useMutateDownloadAttachments = () => {
     }) => {
       const byteArray = await downloadAndDecryptAttachment(url, secretKey)
       if (!byteArray) throw new Error('Invalid file')
-      FileSaver.saveAs(new Blob([byteArray]), fileName)
+      FileSaver.saveAs(new Blob([byteArray as BlobPart]), fileName)
       return fileName
     },
     {
--- a/package.json
+++ b/package.json
@@ -250,7 +250,6 @@
       "@datadog/pprof",
       "aws-sdk",
       "aws-lambda-ric",
-      "serverless",
       "canvas",
       "dd-trace",
       "esbuild",
--- a/package.json
+++ b/package.json
@@ -1,12 +1,12 @@
 {
-  "name": "FormSG",
-  "description": "Form Manager for Government",
+  "name": "formsg-monorepo",
   "version": "6.312.0",
+  "private": true,
+  "description": "Form Manager for Government",
   "homepage": "https://form.gov.sg",
   "authors": [
-    "FormSG <formsg@data.gov.sg>"
+    "FormSG <formsg@open.gov.sg>"
   ],
-  "private": true,
   "repository": {
     "type": "git",
     "url": "https://github.com/opengovsg/formsg.git"
@@ -24,7 +24,7 @@
     ]
   },
   "scripts": {
-    "test": "pnpm test:backend && npm run test:frontend",
+    "test": "pnpm test:backend && pnpm run test:frontend",
     "test:backend": "env-cmd -f __tests__/setup/.test-env node --experimental-vm-modules node_modules/jest/bin/jest.js",
     "test:backend:ci": "env-cmd -f __tests__/setup/.test-env node --experimental-vm-modules node_modules/jest/bin/jest.js --maxWorkers=2 --logHeapUsage --workerIdleMemoryLimit=0.2",
     "test:backend:watch": "env-cmd -f __tests__/setup/.test-env node --experimental-vm-modules node_modules/jest/bin/jest.js --watch",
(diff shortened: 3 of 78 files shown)


kevin9foong:
todo: src/app/modules/submission/__tests__/submission.service.spec.ts seem wonky - to look into it

src/app/modules/submission/__tests__/submission.service.spec.ts
```diff
@@ -140,7 +140,7 @@ const MOCK_ATTACHMENTS: IAttachmentInfo[] = [
 describe('submission.service', () => {
   beforeAll(async () => await dbHandler.connect())
   afterAll(async () => await dbHandler.closeDatabase())
-  beforeEach(() => jest.resetAllMocks())
+  beforeEach(() => jest.clearAllMocks())
```
kevin9foong, line 143: note: prevent setting to undefined which is a behavior of `resetMock`, since there are test cases that share the same spy, causing failures when mocks are set to undefined before each test case run.

serverless/pdf-gen-sparticuz/package.json
```diff
@@ -1,6 +1,7 @@
 {
-  "name": "pdf-gen-sparticuz",
+  "name": "formsg-pdf-gen-sparticuz",
```
kevin9foong, line 2: note: we add a prefix to each package to follow the sdk's convention eg, `formsg-sdk`

apps/backend/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
scottheng96, line 1: Is this required or should this grouped with the BE shift for /apps/backend + the import changes

frontend/src/features/admin-form/responses/IndividualResponsePage/mutations.ts
```diff
@@ -36,7 +36,7 @@ export const useMutateDownloadAttachments = () => {
     }) => {
       const byteArray = await downloadAndDecryptAttachment(url, secretKey)
       if (!byteArray) throw new Error('Invalid file')
-      FileSaver.saveAs(new Blob([byteArray]), fileName)
+      FileSaver.saveAs(new Blob([byteArray as BlobPart]), fileName)
```
scottheng96: Can we add a test in the PR description - just so we remember to test this regression before we release?

package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "FormSG",
-  "description": "Form Manager for Government",
+  "name": "formsg-monorepo",
```
scottheng96, line 2: I have more of a stylistic opinion, if it is necessary to change the name of the root package.json and include a `-monorepo`. Reason for this is to reduce the no. of suffixes that we use (e.g. virus-scanner-guardduty**-iac**.yml)

package.json
```diff
@@ -252,13 +249,19 @@
...
       "esbuild",
       "fsevents",
       "protobufjs",
       "sharp",
-      "unix-dgram"
+      "unix-dgram",
+      "msw",
+      "microtime"
+    ],
+    "ignoredBuiltDependencies": [
```
scottheng96, line 263: Curious, why is `esbuild` being ignored?

apps/backend/package.json
kevin9foong: i think its generally safe to include this - as part of this pr, we are mainly trying to set up a scaffold for the rest of the project and define where things should be, which is the purpose of this empty package.json

package.json
kevin9foong: this naming convention was inspired by referencing other popular monorepos using pnpm workspaces e,g  vite https://github.com/vitejs/vite/blob/main/package.json, turborepo https://github.com/vercel/turborepo/blob/main/package.json, icestark https://github.com/ice-lab/icestark/blob/master/package.json, sveltekit https://github.com/sveltejs/kit/blob/main/package.json (full list here: https://pnpm.io/workspaces) 

but yes it is not strictly necessary to change it. ok with calling it `formsg` as well. let me know if this is something that you feel strongly about - we can revert it

frontend/src/features/admin-form/responses/IndividualResponsePage/mutations.ts
kevin9foong: good catch on this - there is likely a better way to solve this TS issue https://github.com/microsoft/TypeScript/issues/62546#issuecomment-3374526284

ive had a further look and this is caused mainly by the updated TS correction for typesafety against variants of Uint8Array, mainly: ArrayBuffer, SharedArrayBuffer and DataView. 
See: https://webidl.spec.whatwg.org/#js-buffer-source-types

so we should cast it to ArrayBuffer instead - before we apply the change to the sdk

package.json
kevin9foong: ideally, we want to ignore as many unneeded build scripts as possible which are not required for our build to work.

this helps protect us from eg, supply chain attacks and the default stance adopted by pnpm. see: https://pnpm.io/supply-chain-security#block-risky-postinstall-scripts

hence, we block esbuild which is not required. 

after the full migration, we should review our allowed build deps and remove deps that are not needed for our build to work.

frontend/src/features/admin-form/responses/IndividualResponsePage/mutations.ts
kevin9foong: added a test case 

can read the above link to find out more: 

but for Uint8Array -> with the new typescript update, we need to update our SDK to return Uint8Array<ArrayBuffer> see prev typescript had some errors that were patched: 

```
https://github.com/microsoft/TypeScript/issues/62546#issuecomment-3374526284
  decryptFile: (secretKey: string, { submissionPublicKey: filePublicKey, nonce, binary: encryptedBinary, }: EncryptedFileContent) => Promise<Uint8Array<ArrayBuffer> | null>;
```

not in scope for this PR so opting to bypass this by casting it - should be safe since we own the sdk and we dont use sharedArrayBuffer.

package.json
kevin9foong: can reference eliot's commit as well which blocks this - so mainly carrying it over https://github.com/opengovsg/FormSG/pull/9172/changes/7cb90d66cba23ca62d0ef1c57f1419e67aa37ec7#diff-363e93cbc68d4d676770493ba53ba98b9245d851c8a00dab730732b3dab3991e

package.json
scottheng96: Yeah I'm more inclined to leave it as is - a simplified app name without the need to explicitly describe its structure. I think it makes sense if we were to differentiate 2 different FormSG's, 1 that's not a monorepo and 1 that is. 

Not that it _really_ makes a difference, so we don't need to spend more time on this. Thanks for considering this opinion!

scottheng96 (approved):

package.json
eliotlim: whoops good catch!
- i'd left esbuild's postinstall script unapproved on my local development environment as a matter of habit
- not an intentional action for the monorepo, feel free to change this if we need to later!

PR #9173, opened 2026-03-03
feat: use formsg-shared in backend

kevin9foong:
## Problem

Update backend to use /shared package. 

Closes FRM-2315

## Solution

Enabling formsg-shared to be built and used by the BE and FE imports: 
- update the shared absolute imports in BE and tests to use `formsg-shared` workspace package. 
- add a build script for formsg-shared to be built as a 3rd party dep to be imported. This build script is run in the dockerfiles and the build is placed to be able to import by the backend and frontend. 
- add exports for formsg-shared package for imports used by the BE since it now resolves `formsg-shared` as a 3rd party dependency. note that the exports in shared/package.json are made to match the current imports in backend source code. The shared imports can be improved, but is out of scope of this PR which is to make the backend use the formsg-shared package. 
- update tsconfig to exclude shared in the build since it is now using the `formsg-shared` package. 
- update the dockerfiles to use the built `formsg-shared` workspace package.    

CI updates: 
- add formsg-shared to build in playwright CI, so that playwright can import the built files during CI. 

**Breaking Changes** 


shared/package.json
```diff
@@ -3,6 +3,197 @@
   "version": "1.0.0",
   "private": true,
   "description": "",
+  "scripts": {
+    "build": "tsc -p tsconfig.json"
+  },
+  "__comment": "To clean up the exports in shared after monorepo migration",
+  "exports": {
```
kevin9foong, line 10: rationale for this approach: 
- the backend requires the package to be built as cjs files since it cannot use .ts files. previously, before using `formsg-shared`, it was resolving `/shared` directly and building it together defined in the root `tsconfig.json`. thus, we need to build and define exports here for the backend to resolve. 

- the frontend uses vite which can resolve .ts files, hence the default export is the .ts file for vite to bundle directory, which has [benefits over bundling first into esm](https://vite.dev/guide/features) and keeps to our current behavior. 

if require is used, means it is cjs which is used by the backend which resolves to the built cjs .js files it requires. otherwise, it will default to the .ts file.

shared/package.json
```diff
@@ -3,6 +3,197 @@
...
+      "default": "./types/form/form_logo.ts"
+    },
+    "./types/form/workflow": {
+      "default": "./types/form/workflow.ts"
+    },
+    "./types/field": {
+      "default": "./types/field/index.ts"
+    },
+    "./types/submission": {
+      "default": "./types/submission.ts"
```
kevin9foong, line 41: note: for some exports, we do not include `require` since it is not imported by the backend and only by the frontend. 

the imports for shared are not currently well-organised, which is something to clean up in the future but out of scope for this PR.

frontend/src/features/public-form/components/SingleSubmissionModal/SingleSubmissionModal.tsx
```diff
@@ -12,7 +12,7 @@ import {
   Text,
 } from '@chakra-ui/react'
 
-import { FORM_SINGLE_SUBMISSION_VALIDATION_ERROR_MESSAGE } from 'formsg-shared/constants/'
+import { FORM_SINGLE_SUBMISSION_VALIDATION_ERROR_MESSAGE } from 'formsg-shared/constants'
```
kevin9foong, line 15: rationale: this is needed for the frontend to resolve and match the `exports` in `shared/package.json`

tsconfig.json
```diff
@@ -25,49 +29,51 @@
...
     // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
     // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
-
     /* Advanced Options */
-    "skipLibCheck": true,                     /* Skip type checking of declaration files. */
-    "forceConsistentCasingInFileNames": true  /* Disallow inconsistently-cased references to the same file. */
+    "skipLibCheck": true, /* Skip type checking of declaration files. */
+    "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */
   },
-  "include": ["src", "tests", "shared", "__tests__", "playwright.config.ts"],
```
kevin9foong, line 71: note: remove shared from being included in the backend's build "locally", since it imports the `formsg-shared` package instead.

tsconfig.json
```diff
@@ -25,49 +29,51 @@
...
-  "exclude": ["src/frontend"]
```
kevin9foong, line 72: note: this was outdated, updating to "frontend"

shared/tsconfig.json
```diff
@@ -23,7 +22,6 @@
     // "importHelpers": true,                       /* Import emit helpers from 'tslib'. */
     // "downlevelIteration": true,                  /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
     // "isolatedModules": true,                     /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
-
```
eliotlim, line 26: were these removed lines intentional?
- we should aim to minimise changes and maximise readability!

shared/tsconfig.json
```diff
@@ -54,19 +50,16 @@
...
     /* Experimental Options */
     // "experimentalDecorators": true,              /* Enables experimental support for ES7 decorators. */
     // "emitDecoratorMetadata": true,               /* Enables experimental support for emitting type metadata for decorators. */
-
     /* Advanced Options */
     "skipLibCheck": true /* Skip type checking of declaration files. */,
     "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */
   }
-}
+}
```
eliotlim: nit: files should end with a newline:
- so that when a new line is added at the end of a file
- the newline character doesn't cause an unnecessary diff on the old last line at the end of the file.

shared/package.json
eliotlim: yup, as discussed we'll relook at this import decision later.
- in particular, `sdk` will need to export both **CJS** and **ESM**
- and for economies of scale / readability of the codebase, we'll want to align the way we handle `sdk` and `shared` imports.

eliotlim (commented):
so happy to see relative imports going away as we clean up the codebase!! 💯 

just two minor nits on style

shared/tsconfig.json
kevin9foong: this is the prettier default lint - which was not applied previously.

PR #9180, opened 2026-03-09
feat: update frontend to use shared

scottheng96:
## Problem

As part of the migration to monorepo, adding a pnpm workspace dependency `formsg-shared` to use the local `shared` directory

Closes FRM-2316

## Solution

1. Added `formsg-shared` as a  pnpm workspace directory in `frontend/package.json` to reference local directory
2. Replace all `~shared/imports` with `formsg-shared`
3. removed `~shared` from tsconfig paths and update path in eslint.cjs

**Breaking Changes** 


frontend/tsconfig.json
```diff
@@ -22,9 +22,6 @@
     "downlevelIteration": true,
     "baseUrl": ".",
     "paths": {
-      "~shared/*": [
```
kevin9foong, line 25: praise: nice removing this, we dont need it anymore! now we can freely move our /shared directory without worrying about fe imports as pnpm workspaces will handle the symlinks for us

kevin9foong (approved):
lgtm!

PR #9183, opened 2026-03-10
feat: auxiliary pnpm workspaces

eliotlim:
## Problem

We need to migrate the auxiliary packages into pnpm workspaces to centrally manage dependencies better.

The affected packages are:
- `serverless/virus-scanner-guardduty`
- `serverless/pdf-gen-sparticuz`
- `react-email-preview`

Closes FRM-2317 and FRM-2319

## Solution

**Breaking Changes** 

**Improvements**:

- Remove the `pnpm-lock.yaml` file for affected packages
- Manage dependencies from the monorepo's top-level package
- Exclude auxiliary package dependencies from the build process of the top-level package

*Why do we need to exclude auxiliary packages from the build process?*
1. When dependencies are installed for auxiliary packages, additional build tooling is required.
2. For example, serverless packages require the aws-lambda-ric (runtime interface client)
3. Which (i) transitively requires `cmake`, `libtools`, and (ii) adds bloat to the built image
4. Shipping only the required dependencies is a good intermediate step.

Until the backend (currently the top-level package in the repository) is moved to `apps/backend`, the exclusion must be done manually and explicitly at the top level.

## Tests

**TC1: Verify that attachment scanning still works**
- [x] Create a storage mode form that contains an attachment field
- [x] Upload an attachment and submit the form
- [x] The submission should be successful, and the attachment visible in the responses page of the Admin portal

**TC2: Verify that Email and PDF generation works correctly**
- [x] Using the same form, add a Signature field and an Email notification
- [x] Submit the form, it should be successful
- [x] An email should arrive, containing the form response and a PDF attachment.

## Deploy Notes

No changes to deployment behaviour are expected.


Dockerfile.development
```diff
@@ -61,10 +61,11 @@ RUN apk add font-wqy-zenhei --repository https://dl-cdn.alpinelinux.org/alpine/e
...
 # * https://github.com/moby/moby/issues/15858
 COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
-COPY shared/package.json shared/pnpm-lock.yaml ./shared/
-COPY frontend/package.json frontend/pnpm-lock.yaml ./frontend/
+COPY shared/package.json ./shared/
+COPY frontend/ ./frontend/
+COPY frontend/package.json ./frontend/
 
-RUN pnpm install --frozen-lockfile
+RUN pnpm install --frozen-lockfile --filter "!functions/*" --filter "!serverless/*" --filter "!react-email-preview/*"
```
kevin9foong, line 68: praise: nice catch which will help us improve our docker build speeds!

services/virus-scanner-guardduty/Dockerfile
```diff
@@ -10,19 +10,26 @@ RUN apt-get update && \
...
 RUN corepack enable && corepack prepare pnpm@10.30.3 --activate
 
-WORKDIR /function
+WORKDIR /workspace
 
-# install app dependencies
-COPY package.json pnpm-lock.yaml ./
-RUN pnpm install
+# Copy workspace config and lockfile from monorepo root
+COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
```
kevin9foong, line 22: praise: thanks for using the single lock file in the monorepo root! this allows us to adhere to monorepo conventions

services/virus-scanner-guardduty/package.json
```diff
@@ -46,11 +46,5 @@
     "ts-node-dev": "^2.0.0",
     "type-fest": "^4.17.0",
     "typescript": "^5.4.5"
-  },
-  "pnpm": {
-    "__comment": "aws-lambda-ric is needed here since the dockerfile does a local installation.",
-    "onlyBuiltDependencies": [
```
kevin9foong, line 52: praise: thanks for resolving this temp workaround

services/virus-scanner-guardduty/Dockerfile
```diff
@@ -10,19 +10,26 @@ RUN apt-get update && \
...
+COPY services/virus-scanner-guardduty/package.json ./services/virus-scanner-guardduty/
```
kevin9foong, line 23: reviewer todo: check what this does

eliotlim:
P.S. did a quick check and verified that stg-alt2 is working as expected:
- emails look correct
- attachments are virus scanned
- PDF copies are rendered and attached

image

services/virus-scanner-guardduty/Dockerfile
eliotlim: righto! since we're now in a mono-repository, the lambda's structure has changed:

Before:
```
functions/ (copied directly from serverless/virus-scanner-guardduty)
  |- node_modules/ (dependencies)
  |- dist/ (build output)
  |- package.json
  etc.
```

After:
```
workspace (monorepo root)
  |- node_modules/ (filtered dependencies)
  |- services/
    |- virus-scanner-guardduty/ (copied from services/virus-scanner-guardduty)
      |- dist/ (build output)
      |- package.json
  |- pnpm-lock.yaml
  |- pnpm-workspace.yaml
  |- package.json
```

eliotlim: but that's okay because the WORKDIR has changed to `/workspaces/services/virus-scanner-guardduty`, so there's no difference from thereon.

PR #9185, opened 2026-03-11
feat: move frontend to apps folder

scottheng96:
## Problem

Closes FRM-2328

## Solution

1. Moved frontend to apps/frontend
2. Update config files to prefix `frontend` with `apps/frontend`
3. Update gh actions to reference /apps/frontend

**Breaking Changes** 


kevin9foong:
hey @scottheng96 , since the shared package builds its dist in its packages/shared folder, shall we also do the same for frontend and backend eg, apps/frontend/dist instead of dist in the root? 

edit: actually, theres some complications where the backend uses the frontend build in `frontend.controller.ts`

.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
             frontend:
               - __tests__/**
               - packages/shared/**
-              - frontend/**
+              - apps/frontend/**
```
kevin9foong, line 43: potential regression to solve: 
- previously, the frontend has a package-lock which this can detect changes in frontend dep versions for and trigger a test run. 
- now that we have a root pnpm-lock, how can we allow this to detect frontend version changes accordingly and trigger the test runs?

kevin9foong: can tackle this after the build locations

scottheng96: Does keeping the same filter also check for package.json changes? `package.json` changes = pnpm-lock.yaml changes (FE deps) -> run tests

kevin9foong: if we include pnpm-lock, we might want to search changes to the package specific lines, cause otherwise a backend dep change might cause a run of the fe tests as well

kevin9foong: ive tried it and theres no good way, lets include the `pnpm_lock.yml` in each dorny/path-filter 

eg, frontend
- pnpm-lock.yml 
backend 
- pnpm-lock.yml 

For info: 
```
1. The "Lockfile Trap"
Because you are using a single lockfile, any dependency change in the workspace (even in an unrelated package) will trigger pnpm-lock.yaml to change. With the config above, this means all your test suites will run if the lockfile is updated.

This is actually the safest behavior. In a monorepo, changing a shared dependency (like a logging utility used by both api and web) needs to be verified across all consumers. 
```

scottheng96: Basically we're saying always run FE tests whenever there are dependency changes yes

kevin9foong, line 43: issue: include pnpm-lock.yaml

PR #9187, opened 2026-03-11
feat: move backend to apps folder

kevin9foong:
## Problem

Move backend from root to its own package folder `apps/backend`. 

Closes FRM-2326.

## Solution

This PR moves the backend into the monorepo as a proper workspace package under apps/backend/, building on the structure from feat/package-monorepo (which already has packages/shared and services/).

1. Backend as a workspace package
Source: src/ is moved to apps/backend/src/ and the backend is added to the pnpm workspace as formsg-backend.
Tests: All backend tests are colocated in the package: __tests__/integration, __tests__/setup, and __tests__/unit/backend are moved to apps/backend/__tests__/.
Config: Backend gets its own package.json (with backend deps and scripts), tsconfig.json, and jest.config.js. Root package.json is trimmed to only root-level dev tooling and delegates backend commands via pnpm --filter formsg-backend (e.g. test, build, lint, dev).

2. ESLint and tooling
Root .eslintrc is removed and apps/backend/.eslintrc.js is added so ESLint runs in the backend package with tsconfigRootDir: __dirname, fixing resolution for the new paths. Lint-staged config is moved to apps/backend/.lintstagedrc.yml.

3. CI
Change detection: Path filters are updated so backend is identified by apps/backend/** (and pnpm-lock.yaml) instead of root src/ and __tests__/.
Shared as its own surface: A shared change filter is added and two jobs are introduced: shared_lint and shared_test, so changes in packages/shared trigger only shared lint/test, not the full backend pipeline.
Backend job: Backend test and lint jobs use pnpm install (with pnpm cache) and run pnpm --filter formsg-backend (and pnpm install --prefix services/... where still needed). The main lint job is renamed to lint-ci:backend.

4. Build and run
Root clean and build scripts are updated so backend build output lives under apps/backend/dist (and frontend/shared under their packages). Root no longer has a single dist/ tree; each app owns its dist/.

5. Docker
Development: Dockerfile copies apps/backend/package.json (and shared/frontend as before) and runs pnpm install with the same workspace filters.
Production: Final image copies built artifacts from apps/backend/dist, apps/frontend/dist, and packages/shared into the correct paths under /opt/formsg and no longer moves backend out of a single dist/backend/ layout.

6. E2E and Playwright
E2E helpers that load backend code (e.g. models) are updated to require from apps/backend/dist/... instead of dist/backend/.... Playwright cache/restore paths are updated to include apps/backend/dist and apps/backend/node_modules.

7. Shared test resources
Backend-only test resources under __tests__/unit/backend/resources/ are moved to __tests__/resources/ so they can be used by both backend and other packages (e.g. shared), with imports updated to the new paths.

8. Misc
.gitignore and .dockerignore are updated to ignore .pnpm-store.
Root tsconfig.json is simplified so backend compilation is fully owned by apps/backend/tsconfig.json.
Overall, the backend is now a self-contained workspace package under apps/backend/ with its own deps, scripts, and CI surface, while the root repo only orchestrates workspace commands and shared tooling.

**Breaking Changes** 


kevin9foong:
TODO: 
- shared needs to be tested in ci if changes are detected ✅ 
- need to be able to detect changes in deps for fe, be, shared independently (hint: look at pnpm-lock config) in ci.yml ✅ 
- fix FE unit test runs ✅ 
- define build locations and use the fe build path ✅ 
- verify playwright works, move the launch scripts/maildev backend pakage.json ✅

.github/workflows/ci.yml
```diff
@@ -207,4 +215,55 @@ jobs:
           cache: 'pnpm'
       - name: Install dependencies
         run: pnpm install
-      - run: pnpm lint-ci
\ No newline at end of file
+      - run: pnpm lint-ci:backend
+
+  shared_lint:
```
kevin9foong, line 220: rationale: shared is now split from backend, it runs its own tests and should have its own ci job

apps/backend/.eslintrc.js
```diff
@@ -0,0 +1,81 @@
...
+  overrides: [
+    {
+      files: ['*.ts'],
+      parser: '@typescript-eslint/parser',
+      parserOptions: {
+        sourceType: 'module',
+        ecmaFeatures: {
+          modules: true,
+        },
+        tsconfigRootDir: __dirname,
```
kevin9foong, line 26: rationale: required so that the linting check starts from this dir, and not from root.

apps/backend/package.json
```diff
@@ -3,7 +3,207 @@
...
-  "scripts": {},
-  "dependencies": {},
-  "devDependencies": {}
-}
+  "scripts": {
+    "build": "tsc -p tsconfig.build.json && pnpm copyfiles:backend",
+    "copyfiles:backend": "copyfiles -u 1 src/**/*.html ./dist/src",
+    "dev": "tsnd --poll --respawn --transpile-only --inspect=0.0.0.0 --exit-child -r dotenv/config -- src/app/server.ts",
+    "start": "node -r dotenv/config dist/src/app/server.js",
+    "test:e2e-v2:server": "env-cmd -f ./__tests__/setup/.test-env pnpm test-e2e-server",
```
kevin9foong, line 11: note: backend related e2e setup is moved here

frontend/src/features/public-form/PublicFormService.ts
```diff
@@ -52,6 +51,8 @@ export const PUBLIC_FORMS_ENDPOINT = '/forms'
 
 export const X_FORMSG_SELECTED_FORM_LANGUAGE = 'X-Formsg-Selected-Form-Language'
 
+type PresignedPost = AttachmentPresignedPostDataMapType['presignedPostData']
```
kevin9foong, line 54: rationale: after the root package removed the aws-sdk/clients/s3 dep, the frontend needs to download its own, but instead of downloading s3 sdk which is a heavy dep just for the type, it should just use the type already exported by`formsg-shared`.

packages/shared/types/field/attachmentField.ts
```diff
@@ -1,5 +1,5 @@
 import { BasicField, FieldBase } from './base'
-import { PresignedPost } from 'aws-sdk/clients/s3'
+import type { PresignedPost } from 'aws-sdk/clients/s3'
```
kevin9foong, line 2: rationale: these deps are only used for its type and only used during transpile time. to prevent to install these deps during runtime, specify the imports as type and include them as devDeps.

Dockerfile.production
```diff
@@ -26,7 +27,6 @@ COPY . ./
 # These options are only used in the build stage, not the start stage.
 ENV NODE_OPTIONS="--max-old-space-size=4096"
 
-RUN pnpm --filter formsg-shared build
 RUN pnpm build
```
kevin9foong, line 30: note: the pnpm build has been updated to also clean then build the `formsg-shared` package, since it is always required.

.github/workflows/ci.yml
```diff
@@ -41,12 +42,19 @@ jobs:
               - __tests__/**
               - packages/shared/**
               - frontend/**
+              - pnpm-lock.yaml
```
kevin9foong, line 45: rationale: since there is a common shared pnpm lock file, we want updates to deps to packages (which will trigger a change to the root pnpm-lock.yaml) to be detected and invoke the tests to verify correctness.

apps/backend/src/app/models/__tests__/submission.server.model.spec.ts
```diff
@@ -19,7 +19,7 @@ import getSubmissionModel, {
 } from 'src/app/models/submission.server.model'
 import { buildMrfMetadata } from 'src/app/modules/submission/submission.utils'
 
-import { ISubmissionSchema } from '../../../../src/types'
+import { ISubmissionSchema } from '../../../types'
```
eliotlim, line 22: praise: great catch on this relative import.

eliotlim (approved):
praise: thank you for handling this end-to-end, and what's more -- figuring out how to stage everything to minimise the number of relative import differences during the refactoring and moving step.

LGTM! 👍

PR #9188, opened 2026-03-11
fix: email passing validation

eliotlim:
## Problem

Closes FRM-2324

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Details ...

## Tests

**TC1: Submit a Payment Form**
- [ ] Submit a payment form
- [ ] Download the PDF proof of payment on the Payment Success screen

## Deploy Notes

No changes to deployment characteristics are expected.

### DIFF
--- a/src/app/modules/auth/auth.middlewares.ts
+++ b/src/app/modules/auth/auth.middlewares.ts
@@ -277,7 +277,14 @@ const isPlatformApiUser: ControllerHandler<
     })
 }
 
+const validatePlatformUserEmailParams = celebrate({
+  [Segments.BODY]: Joi.object({
+    userEmail: Joi.string().email().optional(),
+  }).unknown(true),
+})
+
 export const authenticateApiKeyAndPlatform = [
+  validatePlatformUserEmailParams,
   authenticateApiKey,
   isPlatformApiUser,
 ] as ControllerHandler[]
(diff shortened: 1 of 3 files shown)


src/app/modules/auth/auth.middlewares.ts
```diff
@@ -277,7 +277,14 @@ const isPlatformApiUser: ControllerHandler<
     })
 }
 
+const validatePlatformUserEmailParams = celebrate({
+  [Segments.BODY]: Joi.object({
+    userEmail: Joi.string().email().optional(),
+  }).unknown(true),
```
kevin9foong, line 283: question: why is this unknown(true) set here?

kevin9foong (approved):

src/app/modules/auth/auth.middlewares.ts
kevin9foong: just a curious question, but otherwise i dont see anything blocking this since this is better than no validation like previously, so with this change it can only get more secure.

eliotlim: Good question! The **AdminFormsPublicRouter** which transitively uses this middleware accepts all sorts of payloads, so we need to make sure that the only thing we're validating is the `userEmail`.

Here's the chain of usage:
- the `isPlatformApiUser` middleware
- is part of the `authenticateApiKeyAndPlatform` middleware
- that protects the [AdminFormsPublicRouter](https://github.com/opengovsg/FormSG/blob/2d68a62533f1a502c72f3cb4a9e3af12f14d3e46/src/app/routes/api/public/v1/admin/forms/admin-forms.public.routes.ts#L12).

PR #9193, opened 2026-03-13
feat: build script update

kevin9foong:
## Problem

As part of monorepo migration, we will have 2 versions:
- internal version (eg, backend, shared, frontend, virus scan which we do not export) 
- external version (eg, clients - any package we publish eg, sdk and other deps will share this same external version #) ref: https://github.com/aws/aws-sdk-js-v3

Requirements: 
- keep existing flow as much as possible (run release script, generate changelog, tag etc)
- want to keep our internal package versions in sync. 
- exclude external packages using the `path` param in the .versionrc
- extensible to support client packages (incoming packages/sdk is excluded using the `path` config)  

Closes FRM-2313

## Solution

We run version bump on all internal packages using the commit-and-tag-version utility tool. 
This tool bumps all internal package versions according to conventional commits and extracts the scopes which refer to the package.  

- Update release hotfix and release prep to support this multiple package bump. 
- Use a shared pr body creation script to reduce duplicates between the release scripts.

**Breaking Changes** 

### DIFF
--- a/scripts/generate_pr_body.sh
+++ b/scripts/generate_pr_body.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+# Usage: generate_pr_body.sh <version_num> [changelog_file]
+#   version_num    - bare tag version number, e.g. 6.314.0 (no "v" prefix)
+#   changelog_file - path to changelog (default: CHANGELOG.md)
+#
+# Outputs the PR body to stdout. Use redirection to save to a file:
+#   ./scripts/generate_pr_body.sh 6.314.0 > .pr_body_release
+#   ./scripts/generate_pr_body.sh 6.314.0 path/to/CHANGELOG.md > .pr_body_release
+
+set -euo pipefail
+
+if [[ $# -lt 1 ]]; then
+  echo "Usage: $0 <version_num> [changelog_file]" >&2
+  echo "  e.g. $0 6.314.0" >&2
+  echo "  e.g. $0 6.314.0 path/to/CHANGELOG.md" >&2
+  exit 1
+fi
+
+release_version="$1"
+changelog_file="${2:-CHANGELOG.md}"
+
+if [[ ! -f "${changelog_file}" ]]; then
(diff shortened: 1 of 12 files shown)


scripts/generate_pr_body.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+# Usage: generate_pr_body.sh <version_num> [changelog_file]
+#   version_num    - bare tag version number, e.g. 6.314.0 (no "v" prefix)
+#   changelog_file - path to changelog (default: CHANGELOG.md)
+#
+# Outputs the PR body to stdout. Use redirection to save to a file:
+#   ./scripts/generate_pr_body.sh 6.314.0 > .pr_body_release
+#   ./scripts/generate_pr_body.sh 6.314.0 path/to/CHANGELOG.md > .pr_body_release
```
kevin9foong, line 8: rationale: support for other changelog since eg, client sdk will have its own changelog.md

scripts/release_hotfix.sh
```diff
@@ -81,80 +61,66 @@ if [[ "$1" != "--nosquash" && "$2" != "--nosquash" ]]; then
...
+release_version_num=$(jq -r .version < package.json)
+release_version="v${release_version_num}"
+release_branch=release_${release_version}
+
+echo -e "\033[34mNext patch version: ${release_version}\033[0m"
 
-git commit -a -n -m "chore: bump version to ${release_version}"
-git tag ${release_version}
+
+if [[ "$1" == "--recut" || "$2" == "--recut" ]]; then
```
scottheng96: Suggestion: reduce fragility on `--recut` being limited to 1st and 2nd args, update to as long as `--recut` is included in the args

```
if [[ " $* " == *" --recut "* ]]; then
```

kevin9foong: applied to all the --recut and --nosquash flags

scottheng96 (approved):

PR #9206, opened 2026-03-17
fix(react-email-preview): update paths for template imports in email preview

scottheng96:
## Problem

Chromatic jobs are failing due to relative import paths being incorrect (in email preview)

## Solution

Update to the import paths (following `apps/backend/src`

**Breaking Changes** 


eliotlim (approved):
Thanks @scottheng96 👍

PR #9215, opened 2026-03-18
feat(shared, frontend, backend): organise shared exports dual format build

kevin9foong:
## Problem

Enable dual-format building for shared package. 

Requirements: 
- esm support for react, newer node modules 
- cjs support for older node modules 
- dual format build tool which can help automate a bunch of items (eg, creating the exports mapping to cjs, esm) 
- build tool should able to be used for sdk also (to provide standardization)

Closes [insert issue #]

## Solution

Suggesting the following structure: 
- modules: each module eg, logic has its own barrel 
- utils: omit the barrel file, since each util is independent and not really related to each other, hence should have its own export 
- types: support for barrel import and deep imports
- constants: support for barrel import and deep imports 

To support the barrel for constants, we need to fix the name conflict with `types` in `constants/fields/myinfo` and `constants/fields/basic`, hence they have been renamed to avoid conflict. 

Use tsdown to automate esm, cjs build and export resolution. 

ci.yml dependency updates should theoretically speed up our ci.yml job completion by 1-2mins per run. 

**Breaking Changes** 


.github/workflows/ci.yml
```diff
@@ -155,11 +150,11 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
           cache: 'pnpm'
       - name: Install dependencies
-        run: pnpm install
+        run: pnpm --filter formsg-frontend install
       - run: pnpm lint:frontend
 
   backend_test:
-    needs: [changes, install, build]
```
kevin9foong, line 162: rationale: backend test does not need to wait for build, saving ~2mins in ci time

.github/workflows/ci.yml
```diff
@@ -96,14 +97,6 @@ jobs:
         env:
           NODE_OPTIONS: '--max-old-space-size=4096'
           AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE: 1
-      - name: Upload build files
```
kevin9foong, line 99: note: these build files are not being used

kevin9foong:
Closing this PR since it is not actively pursued.

PR #9216, opened 2026-03-18
fix(frontend): pdf bulk download progress modal janks

kevin9foong:
## Problem

Issue 1: 
image
Due to the progress count not resetting between runs and continuing from the past undefined value, the modal count can exceed 100%. 

Issue 2: 
image
There is a progress modal timeout that spawns another modal (i believe it is for legacy reasons, where when the csv only button is clicked but is not done after 5s, a modal appears to show the progress). This behavior adds additional complexity to pdf gen / attachments download cases, and hence recommended to remove. 

Issue 3: 
Abort does not work once the function is past the submission streaming stage (eg, during decryption and pdf gen). 

Closes FRM-2333

## Solution

For issue 1: 
Reset the decryption and pdf states to 0 after each run and before. (commit 2/3 [here](519305db74b26592b36277c09dd543bb834dcc8b)) 
 
For issue 2: 
Remove this spawning behavior (commit 3/3 [here](bc6cde15fe3ade249a4952f0e781906a89a8789a))

For issue 3: 
Add checks for aborted signal from the AbortController, abort by returning when the signal is detected. (commit 1/3 [here](76c680438425530e5161cce334abe88c8aa6a1e9))   

**Breaking Changes** 

## Tests
TC1: Repeated bulk downloads work correctly 
- [x] Choose a form with many submissions (>100 preferable)
- [x] Try a combination of pdf, attachment, csv bulk downloads
- [x] Repeatedly run the bulk download (eg, 5 times) and ensure the progress modal displays the correct percentage and the download is successful. 
- [x] Assert no double modal pops up.  

TC2: Bulk download aborts work 
- [x] Start an attachment bulk download. Click on cancel. Ensure it works and the main thread is responsive, all threads are killed in chrome://inspect/#workers. 
- [x] Start a PDF bulk download. Click on cancel once it crosses 50%. Ensure that the main thread is responsive, all threads are killed in chrome://inspect/#workers

### DIFF
--- a/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/DownloadButton.tsx
+++ b/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/DownloadButton.tsx
@@ -16,7 +16,6 @@ import simplur from 'simplur'
 
 import { BxsChevronDown } from '~assets/icons/BxsChevronDown'
 import { BxsChevronUp } from '~assets/icons/BxsChevronUp'
-import { useTimeout } from '~hooks/useTimeout'
 import { useToast } from '~hooks/useToast'
 import Button from '~components/Button'
 import Checkbox from '~components/Checkbox'
@@ -154,9 +153,6 @@ export const DownloadButton = (): JSX.Element => {
     isClosable: true,
   })
 
-  const [progressModalTimeout, setProgressModalTimeout] = useState<
-    number | null
-  >(null)
   const { downloadParams, dateRangeResponsesCount } =
     useStorageResponsesContext()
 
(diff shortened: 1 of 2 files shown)


apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/UnlockedResponses/DownloadButton.tsx
```diff
@@ -224,16 +226,15 @@ export const DownloadButton = (): JSX.Element => {
...
         setDownloadMetadata(decryptResult)
+        resetDownloadProgress()
         resetDownloadOptions()
       },
     },
   })
 
   const handleBulkDownload = useCallback(() => {
     if (!downloadParams) return
-    setProgressModalTimeout(5000)
```
kevin9foong, line 236: note: this seems to be legacy behavior where the progress modal only spawns after 5s. 

thinking about chesterson's fence, probably was added if the csv only is too slow (since it is the only option that does not have a modal)
adds complexity to other modes we need to maintain though, hence recommending to remove, given we have web worker support now for large downloads.

scottheng96 (approved):

scottheng96:
LGTM!

PR #9219, opened 2026-03-19
feat(frontend): add savedraft beta badge

kevin9foong:
## Problem

To signal to admins that the save draft feature is under beta. 

Closes FRM-2336. 

## Solution

Add betaBadge support for toggles and enable it for the save draft setting toggle. 

eg, on desktop screen: 
image
eg, on iphone SE: 
image

**Breaking Changes** 

## Tests

TC1: Beta badge appears for save draft admin toggle 
- [ ] Go to the settings page. 
- [ ] Assert that the save draft toggle has the beta badge. 
- [ ] Try and use mobile screen size. 
- [ ] Assert the save draft badge does not look janky.

### DIFF
--- a/apps/frontend/src/i18n/locales/features/common/en-sg.ts
+++ b/apps/frontend/src/i18n/locales/features/common/en-sg.ts
@@ -94,7 +94,6 @@ export const enSG: Common = {
     ariaLabel: 'Click to edit the form',
   },
   moreOptions: 'More options',
-  beta: 'Beta',
   betaBadgeLabel: 'Beta',
   average: 'Average',
   notApplicable: 'N/A',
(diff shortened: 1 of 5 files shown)


apps/frontend/src/i18n/locales/features/common/en-sg.ts
```diff
@@ -95,7 +95,7 @@ export const enSG: Common = {
   },
   moreOptions: 'More options',
   beta: 'Beta',
-  betaBadgeLabel: 'beta',
+  betaBadgeLabel: 'Beta',
```
scottheng96, line 97: This is exactly the same as the above variable ^. Is it  easy enough and worth it to consolidate and remove duplicated code?

The above is only used in [here](https://github.com/opengovsg/FormSG/blob/3fa23f0ddd9f35ee90c2862c7d948641d9b630e9/apps/frontend/src/features/admin-form/responses/components/FormResultsNavbar/FormResultsNavbar.tsx#L91)

scottheng96 (approved):

PR #9220, opened 2026-03-19
fix(frontend): signatures bug

scottheng96:
## Problem

2 issues with Signatures field:
1. conditionally shown signatures field displaying empty value on second step
image
- reCaptcha is on; signature field is conditionally hidden/shown by logic
- in a subsequent step, a filled-in signature field shows a empty, disabled state

2. signatures field assigned to 2+ steps, subsequent step with previous signature data, when left untouched, experiences validation error 

image
- reCaptcha is on; signature field is assigned to 2nd step
- on the subsequent step, the form is unable to be submitted if signature field (already filled in from prev step) is left untouched

Closes FRM-2337

## Solution

Maintaining Signature field as a uncontrolled component makes it difficult to manage the formState and component state synchronously. Solution was to create a SignatureCanvas object that manages the display of strokes on `canvas`, and make that a react-controlled component under `<Controller>`. This way:
- Hidden fields are properly excluded from validation
- Values persistent through mount/unmounting
- Field lifecycle is managed by RHF

Previously:

- used `register`/ `unregister` to allow RHF to track value internally
- used `useWatch` to observe changes reactively
- used `setValue` to update formState

Now:

- passing `field.value` as a prop to `SignatureCanvas`  to receive value from RHF
- using `onChange` to update the formState and report changes upwards
- using `useEffect` syncs `pfStrokes` when `field.value` changes

**Breaking Changes** 

## Tests

**Signature value displayed as expected on subsequent step**
- [ ] Create an MRF form; leave reCaptcha on
- [ ] Add a Yes/No field and a Signature field
- [ ] Add a logic to show/hide Signature field based on Yes/No
- [ ] Assign both fields to step 1 of workflow, and create a step 2 workflow with no assigned fields
- [ ] Attempt step 1 submission with a shown signature field
- [ ] Confirm that in step 2, signature value is visible
- [ ] Successfully submit step 2 and complete workflow

**(Regression)**
- [ ] Using the same form above ^, attempt a step 1 submission with a hidden signature field
- [ ] Successfully submit step 1 and step 2 with the hidden signature field

**Form submittable when signature is assigned to both steps**
- [ ] Create an MRF form; leave reCaptcha on
- [ ] Add a Signature field
- [ ] Assign the same signature field to step 1 and step 2 of a workflow
- [ ] Attempt a step 1 submission
- [ ] Confirm that in step 2, signature value is visible, and form is able to submit without touching the signature field again


apps/frontend/src/templates/Field/Signature/SignatureField.tsx
```diff
@@ -17,67 +22,64 @@ import FormErrorMessage from '~components/FormControl/FormErrorMessage'
...
 
-const strokePressureDefault = 0.5
+export interface SignatureCanvasProps {
+  schema: SignatureFieldSchema
+  isHighContrast?: boolean
+  isSubmitting: boolean
+  isValid: boolean
+  errors: FieldErrors<SignatureFieldInput>
+  value: SignatureFieldValues | null
+  onChange: (value: SignatureFieldValues) => void
```
kevin9foong, line 43: suggestion: for `onChange`, we can consider using the `Dispatch<SetStateAction<SignatureFieldValues>>` typing instead. eg see: `apps/frontend/src/components/Calendar/CalendarBase/CalendarContext.tsx`

apps/frontend/src/templates/Field/Signature/SignatureField.tsx
```diff
@@ -17,67 +22,64 @@ import FormErrorMessage from '~components/FormControl/FormErrorMessage'
...
+  const strokePressureDefault = 0.5
 
   // perfect freehand variables
   const pfCanvasRef = useRef<HTMLCanvasElement>(null)
   const boxRef = useRef<HTMLDivElement>(null)
-
-  const watchedSignature = useWatch({ name: schema._id }) as {
-    value: SignatureVectorArray
-  }
   const [pfStrokes, setPfStrokes] = useState<SignatureVectorArray>([])
```
kevin9foong, line 72: question: instead of using derived values eg, deriving `pfStrokes` from value.value, would it make sense to just do `const pfStrokes = value.value` directly? This might make the code easier to understand and less prone to bugs.

kevin9foong: for example, in the code later, we do: 
```
 setPfStrokes([])
 onChange({ type: defaultType, value: [] })
```
if we remove the derived pfStrokes, it might simplify all future occurrences to just 
```
 onChange({ type: defaultType, value: [] })
```

kevin9foong, line 72: suggestion for [removing derived pfStrokes (example commit for reference only)](https://github.com/opengovsg/FormSG/commit/a62f62f825338811e5e695a7104bf16a45051906). 

usually, we want to avoid the use of derived state as much as possible, since these can be a source of bugs and complexity (ref: https://read.engineerscodex.com/p/4-software-design-principles-i-learned). this pr could be an opportunity to do that for sig fields since we are switching to controlled components

apps/frontend/src/templates/Field/Signature/SignatureField.tsx
```diff
@@ -17,67 +22,64 @@ import FormErrorMessage from '~components/FormControl/FormErrorMessage'
...
 
-  // Sync when form value changes
+  // Sync pfStrokes with form value when it changes externally
```
kevin9foong, line 74: suggestion: avoid deriving pfStrokes since we are using controlled components to avoid the need for this sync entirely

apps/frontend/src/templates/Field/Signature/SignatureField.tsx
```diff
@@ -17,67 +22,64 @@ import FormErrorMessage from '~components/FormControl/FormErrorMessage'
...
-    if (watchedSignature?.value) {
-      setPfStrokes(watchedSignature.value)
+    if (value?.value) {
+      setPfStrokes(value.value)
+      if (value.value.length > 0) {
+        setShowSignaturePlaceholder(false)
+      }
     }
-  }, [watchedSignature])
+  }, [value])
```
kevin9foong, line 82: suggestion: instead of `value` in the deps, can we use `value.value.length` instead? this is an int and will reduce the likelihood of re-renders caused by object reference updates, making our code more robust

apps/frontend/src/templates/Field/Signature/SignatureField.tsx
kevin9foong, line 175: suggestion: this can be removed to update the controlled `value` instead

kevin9foong, line 168: question: can we remove the need for `currentStroke` state as well? since the more pieces of state there are, the easier to [go out of sync ](https://tkdodo.eu/blog/dont-over-use-state#getting-out-of-sync)

example [commit - for reference only](https://github.com/opengovsg/FormSG/commit/3a7dcda785bc8189badac694c0857df30991b0ca)

apps/frontend/src/templates/Field/Signature/SignatureField.tsx
scottheng96: Thanks for the suggestions!

I think removing `pfStrokes` can simplify the derivation of values, but if we do so, then we'll be calling `onChange` on every pointerMove, which can be more expensive that calling onChange only on `handlePointerUp`.

For the time being I still think it makes sense to keep it for performance consistency, and spend more time verifying that removing has no drawbacks?

kevin9foong: So each time we call setPfStrokes, we are causing a re-render of the SignatureField 

React-hook-form has performance optimizations that isolate re-renders when we call `onChange` which is the main reason we are using this library instead of using our own state-based form and limits it only to the signature field (See https://www.react-hook-form.com/ for the demo). 

In terms of performance, it will be the same since the signature field component will re-render for all pointer moves  

the above changes enforce that when another dev makes changes to the sig field in the future, they do not need to remember to potentially update all pfStrokes, currentStroke and value, making it much easier to maintain and avoiding potential bugs

kevin9foong:
Seeing that its an incident and that its affecting live users - ill approve this PR first, but strongly feel the derived state should be removed.

kevin9foong (approved):

PR #8558, opened 2025-07-21
feat: Signature field beta

scottheng96:
## Problem

Paper form conversion research highlights that wet ink signature fields are a barrier for users to convert to digital forms. Admins are also hesitant to convert from a wet-ink signature to OES digital alternatives already supported (mobile/email OTP, Singpass verification)

Closes FRM-2054

## Solution

Implemented a new Basic Field (no.23) - Signature Field to provide an almost 1-1 replacement for wet-ink signatures. Instead of signing on a piece of paper users will instead sign on a digital canvas. Signature Field v1 will be released on a beta with future versions to be planned for iterative improvements.

v1:
- Signature field is on betaFlag (`signatureField`) and feature-flag(`signature-field`)
- Signature field is supported for storage & MRF mode forms + submissions
- Signature responses are:
   - in Admin response page, displayed as an image
   - in emails, attached as a .png file
   - in CSV download + attachments, as a .svg file*
   
.svg files were blocked in GSIB email inboxes and therefore there was a need to send as .png files. It is intended that CSV downloads + attachment will also download a .png file but implementation for that is trickier and thus out of scope for v1 so as not to block the release. It is planned and being worked on for v1.1.

**Breaking Changes** 

- No - this PR is backwards compatible  

**Bug Fixes**:

- A bug fix for `AddressField` find button that changes to a blue background on hover when in a disabled state. highContrast has been removed from it, along with the clear button for `SignatureField` so as to standardize the disabled behaviour with `Mobile/EmailField` OTP button.

## Before & After Screenshots

**AFTER**:

signature1
signature2
signature3
signature4
** outdated. email attachment will be PNG file

## Tests

** Make a submission on encrypt mode form**
- [ ] Ensure you have betaFlag: `signatureField` enabled on testing env
- [ ] Add an email field and enable respondent copy + PDF
- [ ] Create a storage mode form and add a signature field
- [ ] Attempt to make a submission successfully
- [ ] Observe on the IndividualResponsePage, the signature response shows an image
- [ ] Attempt to download the responses (Download CSV + attachments), observe that an .svg is downloaded and opened to the signature image
- [ ] Observe in the respondent copy, that the signature response is `Signatrue_Captured_{fieldId}.png` with the .png as an attachment
- [ ] Observe that in email notifications, the signature response is `Signatrue_Captured_{fieldId}.png` with the .png as an attachment

**Make a submission on MRF form with signatures**
- [ ] Ensure you have betaFlag: `signatureField` enabled on testing env
[...]

PR #7966, opened 2024-12-02
feat: address field

scottheng96:
## Problem
Closes FRM-1931

There is no current good way to collect address information for forms. Current users are either (1) creating 1 text field to grab address information or (2) creating multiple text fields to grab separated address information (postal code, block number, street name etc.). These can lead to multiple problems:

- address field has incomplete, incorrect information
- no type-checking for fields, e.g. 
  - unit/level/block number can have special characters/invalid formats
  - postal code is not 6 digits
- difficult to parse address information downstream (in csv & CRM integrations)

## Solution
Build a new basic field focused on collecting local address information. Explicily collect

1. postal code
2. block number
3. street name
4. building name
5. level number
6. unit number

Additionally:
- provide auto-complete of address population using OneMap API to get address info (block number and street name) based on postal code for ease of input for users
- allow editing after auto-complete for user input flexibility
- enforce type-checking for particular address fields 
image

- address output is handled 2 different ways depending on use case
   1. human-readable format ```161, BUKIT BATOK STREET 11, #1-2, SINGAPORE 650161```
     - Individual response on admin page
     - email response in storage mode email notification, PDF & email mode response
   2. machine-readable format
     - csv output (separated columns)
  image

     - email mode output (separated key-value pairs)

```
{"question":"Local address - streetName","answer":"BUKIT BATOK STREET 11"},{"question":"Local address - buildingName","answer":""},{"question":"Local address - levelNumber","answer":"1"},{"question":"Local address - unitNumber","answer":"2"},{"question":"Local address - postalCode","answer":"650161"}
```

- address webhook integration is passed as 1 field and handled accordingly (Discussed with plumber team)

```
{
  "responses": [
    {
      "_id": "6796e79ebbdba20a6e37d6cb",
      "question": "Local address",
      "fieldType": "address",
      "answerArray": [
        "blockNumber_161",
        "streetName_BUKIT BATOK STREET 11",
        "buildingName_",
        "levelNumber_1",
        "unitNumber_1",
        "postalCode_650161"
      ]
    }
  ]
}
```

**Breaking Changes** 

- [x] No - this PR is backwards compatible  

**Features**:
- New compound field Basic Address Field

**Screenshots**:
image
image

#### Can create address field in storage mode
- [ ] Create a new form
- [ ] Create a new 'local address' field on form

#### 1. Can submitted address field in storage mode
- [ ] Create new form
- [ ] Create a new 'local address' field on form
- [ ] Make form public
- [ ] Attempt a submission 
[...]

PR #9221, opened 2026-03-19
feat(sdk): migrate sdk monorepo

eliotlim:
> [!NOTE]
> This is a large PR! Please review-by-commits.

## Problem

Iterating on major features in FormSG is slower than necessary because the FormSG SDK is currently published in a [separate repository](https://github.com/opengovsg/formsg-javascript-sdk).

We estimate a Change Lead Time (CLT) of ~ two to four weeks between the decoupled SDK and the FormSG application based on new fields and features added in the past:
- Signature Field (~ 2 weeks from [sdk support](https://github.com/opengovsg/formsg-javascript-sdk/pull/118) to [feature merge](https://github.com/opengovsg/FormSG/pull/8558))
- Address Field (> 1 month from [sdk support](https://github.com/opengovsg/formsg-javascript-sdk/pull/111) to [feature merge](https://github.com/opengovsg/FormSG/pull/7966)).

Closes FRM-2297.

## Solution

Integrating the FormSG SDK development workflow into the FormSG application mono-repository should improve iteration speed and reduce change lead time.

At the same time, we built capacity and paid off tech debt by moving the FormSG SDK linking and publishing actions into this release process as well.

1. Move the FormSG SDK into the monorepo
2. Enable dual-shipping for the FormSG SDK to support both CommonJS and ESM consumers
3. (TODO: enable npm publishing from the release script)

## Non-goals

1. SDK release will take place in a separate PR
2. npm publishing will take place in a separate PR
3. we're allowing shared and sdk to use different bundling mechanisms (tsc and tsdown) at this time -- will revisit this later.

### Alternative approaches

1. we could precompile and distribute CJS, and have ESM consumers directly import the `.ts` files:
- however, ESM-using node-based backend servers would not compile node_modules/ imports by default
- which would lead to a poorer experience for consumers who are ahead of the technology curve
- we should avoid creating regressive policies.

4. we could ship ESM only, and have CommonJS consumers adopt ESM instead:
- this would be a breaking change for CommonJS SDK users, many of which might be government systems
- and lead to a bifurcation in the ecosystem: those using ESM would benefit from the latest features
- but those using legacy codebases and CommonJS imports would be left behind (a negative externality that government tries to avoid)

**Breaking Changes** 

**Improvements**:

- FormSG SDK has been incorporated into the FormSG monorepo
- FormSG SDK is now dual-shipped in both CJS and ESM by default

## Tests

**TC1: All tests pass**
- [x] Ensure that all existing CI and synthetic tests pass

**TC2: No codemods / patches to downstream consumers required**
- [x] There should be no code changes to downstream consumers of the `@opengovsg/formsg-sdk` package.

## Deploy Notes

**New dependencies**:

- `@opengovsg/formsg-sdk` : for FormSG development, this package is now linked at the workspace level.


kevin9foong:
hey @eliotlim, a convo to be had for next week. thoughts on giving [tsdown](https://tsdown.dev/) a go for both sdk and shared?

package.json
```diff
@@ -25,13 +25,15 @@
     "test:backend:ci": "pnpm --filter formsg-backend test:ci",
     "test:backend:watch": "pnpm --filter formsg-backend test:watch",
     "test:frontend": "pnpm --filter formsg-frontend test",
+    "test:sdk": "pnpm --filter @opengovsg/formsg-sdk test",
     "test:e2e-v2": "pnpm build && npx playwright test",
     "test:e2e-v2:server": "pnpm --filter formsg-backend test:e2e-v2:server",
-    "build": "pnpm clean && concurrently \"pnpm:build:*\"",
+    "build": "pnpm clean && pnpm build:sdk && pnpm build:shared && concurrently \"pnpm:build:frontend\" \"pnpm:build:backend\"",
```
eliotlim: We've explicitly specified the build order:
- pending decision on monorepo tooling
- normally this would be managed by a build system (e.g. Nx, Turbo, Rush, etc.)

packages/sdk/cjs-entry.cjs
```diff
@@ -0,0 +1,5 @@
+// CJS backward-compatibility shim.
+// TypeScript compiles `export default function` to `exports.default = fn`,
+// but CJS consumers expect `require('@opengovsg/formsg-sdk')` to return
+// the function directly (matching the old v0.15.0 behavior).
+module.exports = require('./dist/cjs/index.js').default
```
eliotlim, line 5: this shim enables downstream CommonJS projects to directly require the default function
- no codemods or updates required
- which preserves backward-compatibility

packages/sdk/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@opengovsg/formsg-sdk",
+  "version": "0.16.1",
+  "main": "./cjs-entry.cjs",
```
eliotlim, line 4: backward-compatibility support using the CommonJS shim:
- modern node versions and ESM consumers will use the "exports" directive's "import" path
- legacy node versions and CJS consumers will either use the "main" field (legacy) or the "require" path

eliotlim:
This is a large PR! Please review by commits.

package.json
kevin9foong: hey @eliotlim, good catch! 

actually, ive done the research on this previously and pnpm workspaces **does support topological builds based on dependencies**! 

The reason why its not working now is that we are using `**concurrently** \"pnpm:build:*\` (not using pnpm workspaces correctly and we should probably remove the concurrently cli tooling here), we should be doing `pnpm -r build` instead! 

source: https://pnpm.io/cli/recursive#--no-sort

packages/sdk/cjs-entry.cjs
kevin9foong: hey @eliotlim , any thoughts on considering using `tsdown` for consistency with /shared as well? it can help handle the [shims](https://tsdown.dev/options/shims#enabling-shims) for us as well so we dont need to maintain this!

kevin9foong:
i see this pr is ready for review, are we planning to include the updates to the release / publishing flow to include SDK publishing and versioning in this PR as well? ok with doing it in a separate PR if you're more comfortable with that also 

ie, the following suggested items: 
- create another versionrc extending the internal.versionrc
- update path, version tag, changelog infile config for sdk
- update release_prep and hotfix to support the sdk release tagging + changelog gen using the shared script + publish
- use the existing CHANGELOG.md from sdk as infile

.github/workflows/ci.yml
```diff
@@ -170,15 +208,31 @@ jobs:
...
+            apps/backend/node_modules
+            packages/shared/node_modules
+            packages/sdk/node_modules
+            apps/frontend/dist
+            apps/backend/dist
+            packages/shared/dist
+            packages/sdk/dist
+            services/virus-scanner/build
+            services/virus-scanner-guardduty/build
+            .next
```
kevin9foong: question: wondering why we have .next here when we are not using nextjs? could this be accidentally added?

.github/workflows/ci.yml
```diff
@@ -122,6 +140,26 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
           cache: 'pnpm'
+      - name: Restore build artifacts
+        uses: actions/cache/restore@v3
+        with:
+          path: |
+            node_modules
+            apps/frontend/node_modules
```
kevin9foong, line 145: question: wondering if the frontend tests need all these paths to be restored? eg, backend, virus scaner build and node modules may not be required 

removing the unneeded paths for frontend and backend tests could reduce the restoration time and speed up our CI

.github/workflows/ci.yml
```diff
@@ -170,15 +208,31 @@ jobs:
...
+            apps/backend/node_modules
+            packages/shared/node_modules
+            packages/sdk/node_modules
+            apps/frontend/dist
+            apps/backend/dist
+            packages/shared/dist
+            packages/sdk/dist
+            services/virus-scanner/build
```
kevin9foong, line 223: ^ same comment on the above if eg, virus scanner etc is needed

eliotlim:
Thanks @kevin9foong!

1. Let's enable publishing from the CI in a separate PR, to keep this one well-scoped on bringing the SDK into the monorepo.
2. Will implement some of the improvements to the cached folders
3. Let's consider `tsdown` usage for `sdk` specifically in a separate conversation! The CJS shim is pretty straightforward.

package.json
kevin9foong: for posterity, the command that works for building the frontend and its dependencies specifically in topological order is: 
`pnpm -r --filter formsg-frontend... build` 

explanation: https://github.com/pnpm/pnpm/issues/8335#issuecomment-4122959810

packages/sdk/cjs-entry.cjs
eliotlim: Hey @kevin9foong!

- the usage of `tsdown` in `shared/` will probably be temporary and helps us to synchronise the export paths with the code
- that's because we don't want to update the imports for `formsg-shared` right away, since it'll generate a lot of diff noise

but for external-facing packages, here's the general emergent property we want: what we publish in the repository = what the user gets when they `npm install` the library

I considered `tsdown` for sdk, and have found a few pros and cons that we should discuss (possibly in an RFC)? The gist:
- pros: automagically works, no shims and export definitions required in the codebase
- cons: what we distribute in the repository is no longer representative of what is shipped in npmjs, which is a huge gotcha from a developer experience and provenance perspective

Let's discuss this later when you're back :)
- we can use `tsdown` for `formsg-shared` as an internal package for now
- and ship `formsg-sdk` as-is, with all the backward compatibility shims and exports explicitly written out in code

package.json
eliotlim: Yup, great suggestion 💯 

Implemented the build command using the --recursive option.

.github/workflows/ci.yml
eliotlim: good catch! fixed.

.github/workflows/ci.yml
eliotlim: good catch! fixed.

.github/workflows/ci.yml
eliotlim: good catch! fixed.

.github/workflows/ci.yml
eliotlim: so it seems we need to use the exact same cache configuration
- to prevent [cache restoration failures](https://github.com/opengovsg/FormSG/actions/runs/23578531886/job/68656616539#step:5:2)
- which cause [failing tests](https://github.com/opengovsg/FormSG/actions/runs/23578531886/job/68656616539#step:10:8930)

and it's not very expensive to [restore a cache anyway](https://github.com/opengovsg/FormSG/actions/runs/23578957076/job/68657871905#step:5:1) (takes just 11 seconds), so we should just build, cache, and restore everything.

```
Cache hit for: Linux-build-32cccb9347350d0fbcb6aec1bf216537ac17c156
Received 150994944 of 400788408 (37.7%), 144.0 MBs/sec
Received 400788408 of 400788408 (100.0%), 189.7 MBs/sec
Cache Size: ~382 MB (400788408 B)
/usr/bin/tar -xf /home/runner/work/_temp/efb72d87-e554-4682-a7e7-460b44cdaf8b/cache.tzst -P -C /home/runner/work/FormSG/FormSG --use-compress-program unzstd
Cache restored successfully
Cache restored from key: Linux-build-32cccb9347350d0fbcb6aec1bf216537ac17c156
```

scottheng96 (approved):

PR #9228, opened 2026-03-20
feat(backend, shared): default admin email mrf

scottheng96:
## Problem

When MRF forms are created, the admin's email is not added to email notifications by default. Unlike storage mode, where the admin email by default is added (and configurable) during form mode selection page.

Closes FRM-2329

## Solution

Add admin email to email notifications (under 'others') by default. Done in the BE utilising authenticated user information. Same behaviour for:
- Creating MRF form
- Duplicating ^
- Copy from template ^

**Breaking Changes** 

## Before & After Screenshots

| --- | --- | --- |
| admin emails in 'others' notifications | image | image |

## Tests

**Create MRF form with default admin email added**
- [x] Create a new MRF form
- [x] Confirm that your email is added to 'Others' in email notifications

**Duplicate form with default admin email added**
- [x] (optional) add multiple emails the existing MRF form ^
- [x] Duplicate the form into a new MRF form
- [x] Confirm that only your email is added to 'Others' in email notifications

**Copy-template form with default admin email added**
- [x] With an existing MRF form, copy a template of it to create a new MRF form
- [x] Confirm that only your email is added to 'Others' in email notifications

### DIFF
--- a/apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -1052,9 +1052,14 @@ export const handleCopyTemplateForm: ControllerHandler<
   return (
     // Step 1: Retrieve currently logged in user.
     UserService.getPopulatedUserById(userId)
-      .andThen((user) =>
-        // Step 2: Check if form is currently public.
-        AuthService.getFormIfPublic(formId).andThen((originalForm) =>
+      .andThen((user) => {
+        // Step 2a: If form is MRF, add admin email to email notifications by default
+        if (overrideParams.responseMode === FormResponseMode.Multirespondent) {
+          overrideParams.emails = [user.email]
+        }
+
+        // Step 2b: Check if form is currently public.
+        return AuthService.getFormIfPublic(formId).andThen((originalForm) =>
           // Step 3: Duplicate form.
           AdminFormService.duplicateForm(originalForm, userId, overrideParams, {
             overrideEmails: [user.email],
@@ -1062,8 +1067,8 @@ export const handleCopyTemplateForm: ControllerHandler<
           })
             // Step 4: Retrieve dashboard view of duplicated form.
             .map((duplicatedForm) => duplicatedForm.getDashboardView(user)),
-        ),
-      )
+        )
+      })
       // Success; return new form's dashboard view.
       .map((dupedDashView) => res.json(dupedDashView))
       // Error; some error occurred in the chain.
--- a/apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
+++ b/apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
@@ -1221,15 +1221,23 @@ export const createForm: ControllerHandler<
     // Step 1: Retrieve currently logged in user.
     UserService.findUserById(sessionUserId)
       // Step 2: Create form with given params and set admin to logged in user.
-      .andThen((user) =>
-        AdminFormService.createForm(
+      .andThen((user) => {
+        // Step 2a: For multirespondent forms, add user's email to emails array if not provided
+        if (
+          formParams.responseMode === FormResponseMode.Multirespondent &&
+          !formParams.emails
+        ) {
+          formParams.emails = [user.email]
+        }
+
+        return AdminFormService.createForm(
           {
             ...formParams,
             admin: user._id,
           },
           workspaceId,
-        ),
-      )
+        )
+      })
       .map((createdForm) => {
         return res
           .status(StatusCodes.OK)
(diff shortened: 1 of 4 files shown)


apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1044,18 +1052,23 @@ export const handleCopyTemplateForm: ControllerHandler<
   return (
     // Step 1: Retrieve currently logged in user.
     UserService.getPopulatedUserById(userId)
-      .andThen((user) =>
-        // Step 2: Check if form is currently public.
-        AuthService.getFormIfPublic(formId).andThen((originalForm) =>
+      .andThen((user) => {
```
kevin9foong, line 1055: issue(for discussion): instead of duplicating the same check and processing of overrideProps in copy template and duplicate controller, which both call `duplicateForm` service function in the service, would it make sense to the setting of the default value of emails in `processDuplicateOverrideProps` instead? 

this ensures that if we ever call `duplicateForm` function elsewhere, the behavior where the default emails for mrf is set to `[user.email]` is not lost. 

this is since we have
```
export const processDuplicateOverrideProps = (
  params: DuplicateFormBodyDto,
  newAdminId: string,
): OverrideProps => {
  const { responseMode, title } = params

  const overrideProps: OverrideProps = {
    responseMode,
    title,
    admin: newAdminId,
  }
```
which is meant to do the processing of the overrideProps and this processing should ideally not be done in the controller which mainly handles invoking the appropriate service functions and not application logic

apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -1221,15 +1234,23 @@ export const createForm: ControllerHandler<
...
+        // Step 2a: For multirespondent forms, add user's email to emails array if not provided
+        if (
+          formParams.responseMode === FormResponseMode.Multirespondent &&
+          !formParams.emails
+        ) {
+          formParams.emails = [user.email]
+        }
+
+        return AdminFormService.createForm(
```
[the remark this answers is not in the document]
kevin9foong: eg, 
```
return AdminFormService.createForm(
          {
            ...formParams,
            admin: user._id,
          },
          workspaceId,
          adminEmail, 
        )
```

apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
scottheng96: I think it is possible to set the value at the service level - I would need to call the `UserService` to get the admin email, so maybe not in `processDuplicateOverrideProps` (util function) but in `duplicateForms` (service-level). Then that way we remove the duplicated check in duplicate/template copying which is beneficial.

It might be less efficient db wise, since this flow would need to call the UserService twice, once in controller (from existing code) and again in the service. I can try to code it up and see how clean it is.

apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
```diff
@@ -920,6 +920,14 @@ export const duplicateAdminForm: ControllerHandler<
...
+            // Step 2c: If form is MRF,  add admin email to email notifications by default
+            if (
+              overrideParams.responseMode === FormResponseMode.Multirespondent
+            ) {
+              overrideParams.emails = [user.email]
+            }
+
             return {
               originalForm,
               overrideEmails: overrideWithDuplicatingAdminEmail,
```
eliotlim, line 933: TODO: change this to overrideWorkflowEmails in another PR.

apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
eliotlim, line 942: TODO: change this to overrideWorkflowEmails in another PR.

apps/backend/src/app/modules/form/admin-form/admin-form.controller.ts
eliotlim: Had a quick look at two alternative approaches:

1. we could pass the whole user object into the **processDuplicateOverrideProps** function call instead
  - but this would mean we have one `params.email` codepath for storage mode (where the frontend provides a multi-select email field for email notifications)
  - and a second codepath that needs a user object for the MRF mode forms
2. we could have **processDuplicateOverrideProps** lookup the user's email based on the provided newAdminId iff the form is an MRF form
  - but that would cause 2x the queries to the user service and database

neither is ideal, so let's revisit this decision when we refine the form creation wizard in the future.

eliotlim (approved):
I think this is a reasonable UX improvement and will help make duplicating forms more convenient for admins.

We should revisit override emails as a feature at a later date, because there are two different kinds of emails being overridden:
- workflow email overrides
- notification email overrides <- the scope of this PR.

PR #9230, opened 2026-03-23
fix(frontend): pin mdast-util-gfm-autolink-literal version

scottheng96:
## Problem

FormSG users with older versions of Safari (iOS 15 and 16 / macOS 10.15) experience issues filling in forms.

It turns out that whilst migrating FormSG packages to pnpm workspaces, the `mdast-util-gfm-autolink-literal` package dependency was resolved to `v2.0.1` (originally `v2.0.0`)

Closes FRM-2230.

## Solution

Pin `mdast-util-gfm-autolink-literal` to `v2.0.0` in pnpm-workspace.yaml

**Breaking Changes** 

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,9 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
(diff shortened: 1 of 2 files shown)


eliotlim (approved):
LGTM! 👍

PR #9276, opened 2026-04-02
feat(ci): improve container build process

eliotlim:
## Problem

Closes FRM-2350, FRM-2351.

## Solution

**Breaking Changes** 

**Improvements**:

- use GitHub Container Registry (GHCR) to store pre-built base and release images
- build images once on the release branch to reduce deployment time
- refine hardcoded `VITE_APP_*` constants
  - only constants are built into the Vite output: APP_VERSION, DD_RUM_APP_ID, DD_RUM_CLIENT_TOKEN
  - moved to backend-loaded runtimeEnv / clientEnv: appUrl, apiBaseUrl: '/api/v3', gaTrackingId, formsgSdkMode, ddRumEnv, ddSampleRate

**Bug Fixes**:

- Replace `npx @datadog/datadog-ci` with the binary version for sourcemap upload

## Tests

**TC1: Deployment and Versioning works correctly**
- [ ] Deployments to STG and PRD should work normally.
- [ ] The application version should be correctly tagged to logs and metrics in DataDog.

### DIFF
--- a/.github/workflows/build-base-images.yml
+++ b/.github/workflows/build-base-images.yml
@@ -1,0 +1,128 @@
+name: Build Base Docker Images
+
+on:
+  workflow_call:
+    inputs:
+      force-rebuild:
+        description: 'Force rebuild even if images exist'
+        required: false
+        type: boolean
+        default: false
+    outputs:
+      build-base-image:
+        description: 'Full build-base image reference'
+        value: ${{ jobs.build-base.outputs.image }}
+      runtime-base-image:
+        description: 'Full runtime-base image reference'
+        value: ${{ jobs.build-base.outputs.runtime-image }}
+  workflow_dispatch:
+    inputs:
+      force-rebuild:
+        description: 'Force rebuild even if images exist'
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ghcr.io/opengovsg/formsg
+
+jobs:
+  build-base:
+    name: Build base images
+    runs-on: ubuntu-latest
+    outputs:
+      image: ${{ steps.set-outputs.outputs.build-image }}
+      runtime-image: ${{ steps.set-outputs.outputs.runtime-image }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Compute dependency hash
+        id: deps-hash
+        run: |
+          HASH=$(cat \
+            pnpm-lock.yaml \
+            package.json \
+            pnpm-workspace.yaml \
+            Dockerfile.base \
+            apps/backend/package.json \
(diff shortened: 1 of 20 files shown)


Dockerfile.base
```diff
@@ -0,0 +1,51 @@
...
+
+# =============================================================================
+# runtime-base: Minimal runtime image with system deps and non-root user
+# =============================================================================
+FROM node:22.22.2-alpine3.22 AS runtime-base
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable && corepack prepare pnpm@10.30.3 --activate
+
+RUN apk add --no-cache \
```
kevin9foong: question: does it make sense to also do a `apk upgrade --no-cache` here to ensure the apk package versions in runtime-base is the same as the build-base?

eliotlim: good catch! yes, it does make sense. let me add that.

Dockerfile.base
```diff
@@ -0,0 +1,52 @@
...
+# =============================================================================
+# runtime-base: Minimal runtime image with system deps and non-root user
+# =============================================================================
+FROM node:22.22.2-alpine3.22 AS runtime-base
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable && corepack prepare pnpm@10.30.3 --activate
+
+RUN apk update && apk upgrade && \
+    apk add --no-cache \
```
kevin9foong, line 38: suggestion: we can reduce the size of the image further by doing `apk upgrade --no-cache` instead of `apk update && apk upgrade` 

`apk update && apk upgrade`: Leaves the package index (cache) on your disk, making the image larger.
`apk upgrade --no-cache`	Upgrades all currently installed packages without saving the index.

kevin9foong: it can be applied to all usages since we dont need the index as we are using --no-cache

kevin9foong (approved):

.github/workflows/build-base-images.yml
```diff
@@ -0,0 +1,128 @@
...
+jobs:
+  build-base:
+    name: Build base images
+    runs-on: ubuntu-latest
+    outputs:
+      image: ${{ steps.set-outputs.outputs.build-image }}
+      runtime-image: ${{ steps.set-outputs.outputs.runtime-image }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
kevin9foong, line 43: question: do we want to pin actions as well to the specific tag eg, v4.3.1? since we are doing it for the other gh action scripts

eliotlim: good catch! done 👍

Dockerfile.base
eliotlim: makes sense!

PR #9277, opened 2026-04-03
feat: extract tract for admin form settings

### LINKED ISSUE #8452: feat(i18n): extract text from some components in admin-form setting
Components to be extracted:
- FormDetailsSection
- SecretKeyActivationModal
- SecretKeyFormModal
- SettingsTab

brendancjz:
## Problem

Closes #8452 

## Solution

**Breaking Changes** 

    - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/apps/frontend/src/features/admin-form/settings/components/FormDetailsSection.tsx
+++ b/apps/frontend/src/features/admin-form/settings/components/FormDetailsSection.tsx
@@ -32,6 +32,7 @@ export const FormTitleInput = ({
   initialTitle,
 }: FormTitleInputProps): JSX.Element => {
   const { t } = useTranslation()
+  const { formName } = t('features.common', { returnObjects: true })
   const {
     control,
     handleSubmit,
@@ -73,7 +74,7 @@ export const FormTitleInput = ({
 
   return (
     <FormControl isInvalid={!isEmpty(errors)}>
-      <FormLabel isRequired>{t('features.common.formName')}</FormLabel>
+      <FormLabel isRequired>{formName}</FormLabel>
 
       <Controller<{ title: string }>
         control={control}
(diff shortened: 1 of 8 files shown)


LoneRifle (changes requested):
Most, if not all, of the `t()` invocations within a given component rely on a single per-component prefix (eg, `whitelistCsv`). Consider invoking `t()` so that you get an object containing the text under that prefix, destructure the object, then use as desired.

ie,
```typescript
const { modalTitle, submitButton } = t('features.adminForm.settings.secretKeyModal', { returnObjects: true })
```

brendancjz:
> Most, if not all, of the `t()` invocations within a given component rely on a single per-component prefix (eg, `whitelistCsv`). Consider invoking `t()` so that you get an object containing the text under that prefix, destructure the object, then use as desired.
> 
> ie,
> 
> ```ts
> const { modalTitle, submitButton } = t('features.adminForm.settings.secretKeyModal', { returnObjects: true })
> ```

Oh sweet! Didn't know we could do that, thanks and updated~

apps/frontend/src/features/admin-form/settings/components/FormDetailsSection.tsx
```diff
@@ -30,6 +31,8 @@ interface FormTitleInputProps {
 export const FormTitleInput = ({
   initialTitle,
 }: FormTitleInputProps): JSX.Element => {
+  const { t } = useTranslation()
+  const { formName } = t('features.common', { returnObjects: true })
```
LoneRifle, line 35: Generally, if it's just one key you're retrieving, maybe just direct invocation would do.

LoneRifle (approved):
lgtm

PR #9280, opened 2026-04-06
feat: pin actions and image deps

kevin9foong:
## Problem

Pin actions and image dependencies to specific commits and versions to harden against potential supply chain attacks. 

## Solution

As above. 

**Breaking Changes** 

## Tests

All actions should still run correctly. 
- [ ] Ensure all actions are running correctly. 

Local dev and deployment works 
- [ ] Ensure the local dev image still works. 
- [ ] Ensure deployment to cloud works.

### DIFF
--- a/.github/workflows/aws-deploy-scanner-guardduty-iac.yml
+++ b/.github/workflows/aws-deploy-scanner-guardduty-iac.yml
@@ -40,21 +40,21 @@ jobs:
     environment: ${{ inputs.gha-environment }}
     steps:
       - name: Checkout branch source code into runner environment
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           ref: ${{ inputs.checkoutBranch }}
           fetch-depth: 0
       - name: Install pnpm
-        uses: pnpm/action-setup@v4
+        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v4.4.0
         with:
           version: 10.30.3
       - name: Use Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: 22.22
           cache: 'pnpm'
       - name: Grant runner AWS credentials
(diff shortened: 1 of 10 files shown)


Dockerfile.development
```diff
@@ -13,9 +13,8 @@ ENV NODE_OPTIONS="--max-old-space-size=2048"
 
 # Install pnpm
 RUN corepack enable && corepack prepare pnpm@10.30.3 --activate
-RUN apk update && apk upgrade && \
```
kevin9foong, line 16: rationale: since the final stage in the production container does not fetch the latest apk index or upgrade the world deps, dev container should follow as well to reproduce prod build as closely as possible.

Dockerfile.production
```diff
@@ -8,11 +8,10 @@ RUN corepack enable && corepack prepare pnpm@10.30.3 --activate
 FROM base AS build 
 # node-modules-builder stage installs/compiles the node_modules folder
 # Python version must be specified starting in alpine3.12
-RUN apk update && apk upgrade && \
```
kevin9foong, line 11: rationale: the build stage does not need to perform an apk update or upgrade so as the keep the apk versions in the build stage the same as the final stage of the prod container

.github/workflows/aws-deploy-scanner-guardduty-iac.yml
```diff
@@ -40,21 +40,21 @@ jobs:
     environment: ${{ inputs.gha-environment }}
     steps:
       - name: Checkout branch source code into runner environment
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
kevin9foong: note: using commit is more robust, added comment to track the current version for human-readability

Dockerfile.development
eliotlim: okay, makes sense!
this does depend on #9276 right?

eliotlim (approved):
thanks for pinning these dependency versions 👍 

one minor note on pinning actions:
- shall we adopt a policy of pinning less-trusted or third party actions by hash
- and pinning first-party actions like `actions/` and `github/` by version instead?

this would line up with [internal OGP eng guidance](https://opengovproducts.slack.com/archives/CK9GGK6EP/p1742534509860059).

PR #9282, opened 2026-04-07
feat: mrf response limit v1.0

kevin9foong:
## Problem

The response limit feature is currently unavailable on Multi-Response Forms (MRFs), creating a parity gap that is blocking the unification of Storage and MRF form types.
Closes FRM-2355

## Solution

- Enable the "Set a response limit" toggle on MRF settings pages, matching the behaviour of the existing response limit feature on standard forms.
- Remove the infobox that currently states response limits cannot be applied to multi-respondent forms.
- Response limit should be tied to the number of workflows started (not completed submissions).

Out of scope: Allowing in-progress workflows to continue after the form is closed due to a response limit being reached. This will be treated as a V1.5 enhancement.

Automated testing: 
When the submission limit is met, this submission limit is checked and used to deactivate (close) forms in 2 cases: 
- when the user tries to retrieve the public form, the form is deactivated and a 404 is returned. 
- when a submission is made directly to the submission endpoint (potentially without retrieving the public form immediately before), the form is deactivated and 404 is returned. 

Hence, tests are written for the following 2 scenarios for multi respondent forms (service method and controller at the relevant level of abstraction).
Some of these tests are expected to be updated once Response Limit (MRF) v1.5 is in progress.  

**Breaking Changes** 

## Tests

Submission limit toggle works for MRF
- [ ] Create MRF form with 2 steps 
- [ ] Enable the submission limit toggle on MRF settings and set to submission limit 3. 
- [ ] Submit MRF form entirely ie 2 steps.
- [ ] Ensure that the MRF form is still open. 
- [ ] Submit a second new MRF form and only fill 1 step. 
- [ ] The form should still be open (despite the being "3 steps worth of submissions")
- [ ] Submit a third new MRF form. 
- [ ] The form should now be closed, even for the partially completed workflow and for any new submissions. (This follows the scope of submission limit (MRF) v1.0)

### DIFF
--- a/apps/backend/__tests__/unit/backend/helpers/jest-express.ts
+++ b/apps/backend/__tests__/unit/backend/helpers/jest-express.ts
@@ -35,19 +35,25 @@ const mockRequest = <P extends Record<string, string>, B, Q = any>({
 const mockResponse = (
   extraArgs: Partial<Record<keyof Response, unknown>> = {},
 ): Response => {
+  const markHeadersSent = function (this: { headersSent: boolean }) {
+    this.headersSent = true
+    return this
+  }
+
   const mockRes = {
     locals: {},
+    headersSent: false,
     status: jest.fn().mockReturnThis(),
-    send: jest.fn().mockReturnThis(),
-    sendStatus: jest.fn().mockReturnThis(),
+    send: jest.fn().mockImplementation(markHeadersSent),
+    sendStatus: jest.fn().mockImplementation(markHeadersSent),
     type: jest.fn().mockReturnThis(),
     pipe: jest.fn().mockReturnThis(),
     emit: jest.fn().mockReturnThis(),
     on: jest.fn().mockReturnThis(),
-    end: jest.fn(),
-    json: jest.fn(),
+    end: jest.fn().mockImplementation(markHeadersSent),
+    json: jest.fn().mockImplementation(markHeadersSent),
     render: jest.fn(),
-    redirect: jest.fn(),
+    redirect: jest.fn().mockImplementation(markHeadersSent),
     cookie: jest.fn(),
     set: jest.fn(),
     clearCookie: jest.fn(),
--- a/apps/backend/src/app/modules/form/__tests__/form.service.spec.ts
+++ b/apps/backend/src/app/modules/form/__tests__/form.service.spec.ts
@@ -5,6 +5,7 @@ import {
   FormResponseMode,
   FormStatus,
   SubmissionType,
+  WorkflowType,
 } from 'formsg-shared/types'
 import { merge, times } from 'lodash'
 import mongoose from 'mongoose'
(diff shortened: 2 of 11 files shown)


apps/frontend/src/i18n/locales/features/admin-form/settings/general/en-sg.ts
```diff
@@ -20,6 +20,12 @@ export const enSG = {
       description:
         'Your form will automatically close once it reaches the set limit.',
     },
+    toast: {
+      successStorageMode:
+        'Your form will now automatically close on the {submissionLimit} submission.',
+      successMrf:
+        'Your form will now automatically close on the {submissionLimit} started workflow.',
```
kevin9foong, line 27: note: discussed with Sufyan (designer) on this copy.

kevin9foong:
TODO: modify existing and add unit tests to verify the form is closed after xth workflow is started.

apps/backend/src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.controller.spec.ts
```diff
@@ -345,6 +347,57 @@ describe('multirespondent-submision.controller', () => {
       })
     })
 
+    it('returns 404 not found when form has reached submission limit', async () => {
```
kevin9foong, line 350: note: once the submission limit is hit, aka started workflows == submission limit, all subsequent submissions should be prevented. 

see PRD: Allowing in-progress workflows to continue after the form is closed due to a response limit being reached. This will be treated as a V1.5 enhancement.

These test cases are expected to be updated for v1.5.

apps/backend/src/app/modules/form/__tests__/form.service.spec.ts
```diff
@@ -374,86 +398,162 @@ describe('FormService', () => {
...
+      })
     })
 
-    it('should not let requests through and deactivate form when form has reached submission limit', async () => {
-      // Arrange
-      const formParams = merge({}, MOCK_ENCRYPTED_FORM_PARAMS, {
-        status: FormStatus.Public,
-        submissionLimit: 5,
+    describe('for mrf forms', () => {
+      it('should let requests through when form has no submission limit', async () => {
```
kevin9foong, line 487: note: once the submission limit is hit, aka started workflows == submission limit, the form is closed and all subsequent submissions should be prevented, including in-progress workflows. 

This follows the v1.0 PRD: Allowing in-progress workflows to continue after the form is closed due to a response limit being reached. This will be treated as a V1.5 enhancement.

apps/backend/src/app/modules/form/public-form/__tests__/public-form.controller.spec.ts
```diff
@@ -130,6 +131,81 @@ describe('public-form.controller', () => {
       })
     })
 
+    describe('submission limit checks', () => {
```
kevin9foong, line 134: summary: this verifies the controller should invoke the `checkFormSubmissionLimitAndDeactivateForm` and use its response to return 200 or 404.

apps/backend/__tests__/unit/backend/helpers/jest-express.ts
```diff
@@ -35,19 +35,25 @@ const mockRequest = <P extends Record<string, string>, B, Q = any>({
 const mockResponse = (
   extraArgs: Partial<Record<keyof Response, unknown>> = {},
 ): Response => {
+  const markHeadersSent = function (this: { headersSent: boolean }) {
```
kevin9foong, line 38: ratonale: updating the mock is needed since our controller eg, submitMrfSubmission has an ensure pipeline, which sends a res if eg, the pipeline fails. 

There is then an additional check to do `res.headersSent` in the controller to not double send a response, which this mock does not support.

apps/frontend/src/features/admin-form/settings/components/FormLimitToggle.tsx
```diff
@@ -115,8 +112,6 @@ export const FormLimitToggle = (): JSX.Element => {
   const { data: settings, isLoading: isLoadingSettings } =
     useAdminFormSettings()
 
-  const isMrf = settings?.responseMode === FormResponseMode.Multirespondent
```
eliotlim, line 118: question: should we feature flag the visibility of this toggle for a controlled rollout?

eliotlim (commented):
looks like we're using existing logic for limiting new submissions:
- which means this PR primarily adds test cases and copy
- with no new business logic changes in the backend

in which case: shall we add a feature flag so that we can perform a controlled rollout?

apps/backend/src/app/modules/form/public-form/public-form.controller.ts
```diff
@@ -90,9 +92,17 @@ export const handleGetPublicForm: ControllerHandler<
     formId,
   }
 
+  const gb = req.growthbook
+  const isMrfResponseLimitEnabled =
+    isTest || (gb?.isOn(featureFlags.mrfResponseLimit) ?? true)
```
kevin9foong, line 96: explanation: if the gb flag cannot be loaded, enable the block by default

scottheng96 (approved):

scottheng96:
LGTM!

PR #9283, opened 2026-04-07
feat: email standardisation phase 1

scottheng96:
## Problem

Part 1 of email standardisation feature to consolidate out email templates and have a consistent email template for FormSG emails. Phase 1 is scoped to updating:
1. MRF workflow completion emails
2. MRF approval emails
3. MRF next step emails
4. MRF respondent copy emails

Closes FRM-2354

## Solution

1. Refactored `MrfWorkflowCompletionEmail, `MrfWorkflowEmail` and `MrfRespondentCopyEmail` to use the same (new) template
2. Refactored `mail.service.ts` duplicated code into common functions
3. (new) MRF respondent copy emails to now include Status tracker link if status-tracker is enabled in workflow
4. (new) MRF next step email will include responses of previous submission(s) for admin-requested searchability
5. (new) email buttons are now fully clickable (reduced border of container which caused some areas of 'button' to not be clickable)

Regression check:
1. Re-using the existing `questionAnswer[]` so no change to data displayed
2. PDF Admin copy logic to remain the same - to only be included when signature data is present
3. New email template was tested successfully on COMET & mobile to ensure consistency

Engineering notes:
Outlook's rendering engine doesn't support all modern CSS properties - `borderRadius` is not supported and so emails on outlook will not have the curved borders. Also, it doesn't handle table margins well, so explicit <Row> components are used to create margins between the sections.

**Breaking Changes** 

## Before & After Screenshots

| --- | ---- | ---- |
| MRF workflow completion email | image | image |
| MRF approval email | image | image |
| MRF next step email | image | image |
| MRF respondent copy email (responses + status tracker) | image | image |
| MRF respondent copy email (no response, no status tracker)| dfsdf6image | image |
| Mobile | | image |

## Tests

**TC1: MRF Workflow Completion Email, Next Step Email & Respondent Copy Email**
- [ ] Set up an MRF form
- [ ] Add an email field, toggle on `EmailConfirmation` + PDF copy of responses, assign to step 1
- [ ] Add a Yes/No field, add add it to step 2 of workflow
- [ ] Turn on status tracker
- [ ] Successfully complete the workflow submission
- [ ] Verify that respondent copy contains PDF, displays responses and has a working status tracker button
- [ ] Verify that workflow completion email is received & follows the same template as respondent copy
- [ ] Verify that next step email contains responses of the previous submission

**TC2: MRF Approval Email**
- [ ] On the MRF form created above ^, update step 2 workflow to an approval step, assigned Yes/No field as approval field
- [ ] Successfully complete another workflow submission
- [ ] Verify that approval email is received, and contains **Outcome** card in the email display


apps/backend/src/app/views/templates/EmailTemplate.tsx
```diff
@@ -0,0 +1,332 @@
...
+            )}
+
+            {/* Email end */}
+            <Text style={secondaryTextStyle}>
+              For more details, please contact the respondent(s) or form
+              administrator.
+            </Text>
+
+            {/* JSON Data Section */}
+            {responseJson && (
```
eliotlim: question: does the new format being trialled for MRF preserve the <p>-- Start of JSON --</p> parseability?

1. I noticed that it's moved from being html > body > p (JSON)
2. to html > body > Container > Section > Section > p (JSON)

apps/backend/src/app/services/mail/mail.service.ts
```diff
@@ -1010,50 +1085,34 @@ export class MailService {
...
+    const htmlData: EmailData = {
+      emailTitle: `Review and complete your part of ${formTitle}`,
       formTitle,
       responseId: responseId.toString(),
-      responseUrl,
+      reviewUrl: responseUrl,
+      formQuestionAnswers,
     }
 
-    const generatedHtml = fromPromise(
```
eliotlim, line 1026: question: would we be able to feature flag this rollout for MRF (so that we can have a corresponding feature flag for Storage Mode later)

eliotlim (commented):
just two questions:
- does this preserve parseability and structure of the JSON payload?
- could we feature flag this to reduce deployment and rollout risk for both mrf mode (now) and storage mode (later)?

apps/backend/src/app/views/templates/EmailTemplate.tsx
```diff
@@ -0,0 +1,332 @@
...
+                </Text>
+                <Text>{responseJson}</Text>
+                <Text style={{ ...secondaryTextStyle, marginTop: '8px' }}>
+                  -- End of JSON --
+                </Text>
+              </Section>
+            )}
+          </Section>
+        </Container>
+      </Body>
```
eliotlim: issue: the JSON payload is deeper in the email body structure than it previously was.

Image

this is what would make the new standard introduced with MRF backward compatible with what we currently do on storage mode:

```suggestion
          </Section>
        </Container>

        {renderMargin(40)}

        {/* JSON Data Section */}
        {responseJson && (
          <p>-- Start of JSON --</p>
          <p>{responseJson}</p>
          <p>-- End of JSON --</p>
        )}
      </Body>
```

eliotlim (changes requested):
we need to consider the structural compatibility of this template.
- it's different from what we use in storage mode
- and that will create problems if users also adopt the new MRF structure
- which will lead to a permanent difference between storage / MRF email mode templates

apps/backend/src/app/views/templates/EmailTemplate.tsx
scottheng96: Ahh, I see what you are saying. I'll commit this!

apps/backend/src/app/services/mail/mail.service.ts
scottheng96: aligned that we'll do this for phase 2 and not for phase 1 since MRF emails will not experience regression

apps/backend/src/app/views/templates/EmailTemplate.tsx
```diff
@@ -311,21 +311,19 @@ export const EmailTemplate = ({
...
-                </Text>
-                <Text>{responseJson}</Text>
-                <Text style={{ ...secondaryTextStyle, marginTop: '8px' }}>
-                  -- End of JSON --
-                </Text>
-              </Section>
-            )}
           </Section>
         </Container>
+
```
eliotlim, line 316: thanks for making this change!

eliotlim (approved):
LGTM! 👍

PR #9285, opened 2026-04-08
feat(frontend): env-independent containers via runtime env loading

eliotlim:
## Problem

Previously part of #9276 
Closes FRM-2364

The previous attempt in #9276 bundled two independent changes:
1. the container build process improvements and
2. the frontend/backend refactor to make containers environment-independent.

A bug in the decryption worker surfaced after merge and the combined PR was reverted in #9286.

This PR extracts just the environment-independent container work (plus the worker fix) so it can land on its own, ahead of a follow-up PR that re-introduces the container build process improvements.

## Solution

**Breaking Changes** 

**Improvements**:

- refine hardcoded `VITE_APP_*` constants so that the container image no longer bakes in environment-specific values
  - Vite build includes true constants: `APP_VERSION`, `DD_RUM_APP_ID`, `DD_RUM_CLIENT_TOKEN`
  - all other env vars are loaded into runtimeEnv from the backend (`window.__ENV__`): `appUrl`, `apiBaseUrl: '/api/v3'`, `gaTrackingId`, `formsgSdkMode`, `ddRumEnv`, `ddSampleRate`
- backend injects runtimeEnv `<!-- __ENV_INJECTION__ -->` in `index.html` with a CSP hash entry
- `apps/frontend/src/env.ts` reads `window.__ENV__` with fallbacks to `import.meta.env.VITE_APP_*`
- this means compatibility with CI build-time injection until the follow-up PR removes it

**Bug Fixes**:

- Decryption worker no longer transitively imports `~/env` (which touches `window`, unavailable inside a Web Worker).
- Introduce a `DecryptionCtx` carrying a per-worker `formsgSdk` instance, constructed from `formsgSdkMode` passed in via `SubmissionDataForDecryption`.
- This addresses the root cause of the parallelised-download regression reported in #9286.

## Tests

**TC1: Working Downloads**
- [x] Verify that downloads of CSVs are working correctly
- [x] Verify that downloads of attachments are working correctly
- [x] Verify that downloads of PDFs are working correctly

**TC2: Runtime env loaded correctly**
- [x] Frontend initialises Datadog RUM with the correct `env` / `appUrl` / sample rate per environment
- [x] GA tracking id, formsg-sdk mode, and API base URL are read from the backend-injected `window.__ENV__`

## Deploy Notes

- No change to the container build process in this PR
- the existing CI workflow still injects `VITE_APP_*` at build time, and `env.ts` falls back to those values


eliotlim:
Tested on stg-alt21: https://stg-alt21.form.gov.sg.

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -23,6 +24,29 @@ import {
...
+type WorkerCtx = DecryptionCtx
+
+let cachedFormsgSdk: DecryptionCtx['formsgSdk'] | null = null
+let cachedMode: string | null = null
+
+interface WorkerCtxOptions {
+  formsgSdkMode: string
+}
+
+function createWorkerCtx({ formsgSdkMode }: WorkerCtxOptions): WorkerCtx {
```
kevin9foong, line 37: question: why do we create a function that looks up the cache each time and creates a formsgPackage if it does not exist. 

would just initializing this formsgPackage at the start work? 

or is this function done for convenience to allow passing in the mode to the worker from the main thread?

kevin9foong (approved):
mainly LGTM - just wanted to understand the decision behind making a cacheable worker function

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
eliotlim: good question!

to avoid the risk of introducing a memory leak, i'm preserving the singleton behaviour here:
- in the old codebase, the formSgSdk was a singleton instantiated once during import.
- the ctx object could be recreated per-request using singleton instances of DB clients, SDKs, etc.

PR #9286, opened 2026-04-08
chore: revert 9276 container build improvements

eliotlim:
This reverts commit 9101de2a0851b961c936573caebe908960a45941.

## Problem

Parallelised Downloads are broken in #9276 

## Solution

**Breaking Changes** 


kevin9foong (approved):
Thanks for this. 

This window issue and dependency is admittedly more complex than expected - so reverting this to buy us more time to figure this out is the right call!

PR #9287, opened 2026-04-08
fix: replace npx usage

eliotlim:
## Problem

Extracted from #9276.

Fixes FRM-2350.

## Solution

**Breaking Changes** 

**Improvements**:

- Pin datadog-ci as a devDependency
- Use pnpm exec instead of npx

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -118,7 +118,7 @@ jobs:
       - name: Run Playwright tests
         env:
           NODE_OPTIONS: -r ${{ env.DD_TRACE_PACKAGE }}
-        run: npx playwright test --project=${{ matrix.project }}
+        run: pnpm exec playwright test --project=${{ matrix.project }}
       - uses: actions/upload-artifact@v4.6.2
         if: always()
         with:
--- a/Dockerfile.production
+++ b/Dockerfile.production
@@ -37,12 +37,12 @@ ARG REPO_URL
 # Mount creates a file inside /run/secrets/dd_api_key, and source to load the environment variables
 # For the content of the mounted file is only accessible in the RUN command where it’s referred in, use && command.
 RUN curl -L --fail -o /usr/local/bin/datadog-ci \
-    https://github.com/DataDog/datadog-ci/releases/download/v5.11.0/datadog-ci_linux-x64 && \
+    https://github.com/DataDog/datadog-ci/releases/download/v5.12.1/datadog-ci_linux-x64 && \
     chmod +x /usr/local/bin/datadog-ci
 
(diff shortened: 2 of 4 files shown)


.github/workflows/playwright.yml
```diff
@@ -118,7 +118,7 @@ jobs:
       - name: Run Playwright tests
         env:
           NODE_OPTIONS: -r ${{ env.DD_TRACE_PACKAGE }}
-        run: npx playwright test --project=${{ matrix.project }}
+        run: pnpm exec playwright test --project=${{ matrix.project }}
```
kevin9foong, line 121: nice! this will help ensure we are using our pinned version!

kevin9foong (approved):
approved - thanks for making this change!

Dockerfile.production
```diff
@@ -36,9 +36,13 @@ ARG REPO_URL
 # Upload datadog source maps from secret mount, id MUST be `id=dd_api_key` in the docker build command
 # Mount creates a file inside /run/secrets/dd_api_key, and source to load the environment variables
 # For the content of the mounted file is only accessible in the RUN command where it’s referred in, use && command.
+RUN curl -L --fail -o /usr/local/bin/datadog-ci \
```
kevin9foong: suggestion: curious, would it make sense to use pnpm to install the specific version instead of curl? 

for example,`pnpm add -g @datadog/datadog-ci@5.12.1`

since we are already doing this for `pnpm add -g node-gyp@12.2.0`

Dockerfile.production
```diff
@@ -36,9 +36,13 @@ ARG REPO_URL
...
+    https://github.com/DataDog/datadog-ci/releases/download/v5.11.0/datadog-ci_linux-x64 && \
```
kevin9foong: issue: shall we use the latest 5.12.1 version?

Dockerfile.production
eliotlim: the goal is ensure that only known good / bundled dependencies are installed
- if we use pnpm to install it, we'd have to install @datadog/datadog-ci as a devDependency to pin all transitive dependencies, which would pull it into every development environment
- since we only use it in CI, using the bundled binary approach is more suitable.

Dockerfile.production
eliotlim: makes sense, will bump to v5.12.1

Dockerfile.production
kevin9foong: makes sense - would this mean that we should probably apply the same for `node-gyp`?

eliotlim: yup, in a separate PR!

PR #9295, opened 2026-04-14
fix(ci): update stg-alt2 parameters

eliotlim:
## Problem

After recreating stg-alt2 as stg-alt21, these parameter values need to be updated.

Related to https://github.com/opengovsg/formsg-infra/pull/121.

## Solution

Update the deployment parameters. Tested working on #9285.

**Breaking Changes** 

**Bug Fixes**:

- Update paths and parameters to `stg-alt21`


.github/workflows/deploy-ecs-stg-alt2.yml
```diff
@@ -17,9 +17,9 @@ jobs:
       gha-environment: 'stg-alt2'
       # Environment configuration
       environment: 'stg-alt2'
```
kevin9foong, line 19: hey @eliotlim , does it make sense to update everything to just use stg21 instead? so we dont have stg-alt2 and stg-alt21 which could cause some confusion in the future and we need to remember when to use each

eliotlim: that's true. i'm hoping that this stg-alt21 naming is temporary 😅 and we can go back to stg-alt2 soon.

kevin9foong (approved):

PR #9305, opened 2026-04-15
feat(sdk): enable publishing

eliotlim:
## Problem

Building features and fixes for the FormSG SDK and keeping development in sync with the rest of FormSG is challenging.

Closes FRM-2362.

## Solution

The FormSG SDK (@opengovsg/formsg-sdk) is moving into the [FormSG monorepo] (https://github.com/opengovsg/FormSG). This allows the FormSG team to stably incorporate new features while rapidly developing fixes on the FormSG application and SDK in lockstep.

**Breaking Changes** 

**Usage**:
1. develop and merge a feature PR to `develop`
2. run `scripts/release_prep.sh` normally
3. approve the release PR to `release-al2`
4. approve the publish-sdk job on `release-al2` if SDK changes are detected
5. the new SDK version will be published automatically to npm

**Features**:

- This release will sync FormSG SDK to `v7.x` to match the application version.

**Improvements**:

- `scripts/release_prep.sh` updated to also bump SDK version if affected
- `.external.versionrc.js` is used by **commit-and-tag-version** to manage SDK versioning rules

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opengovsg/formsg-sdk",
-  "version": "0.16.1",
+  "version": "7.0.0",
   "main": "./cjs-entry.cjs",
   "module": "./dist/esm/index.js",
   "types": "./dist/cjs/index.d.ts",
@@ -35,6 +35,10 @@
   },
   "description": "Node.js SDK for integrating with FormSG",
   "license": "MIT",
+  "files": [
+    "dist/",
+    "cjs-entry.cjs"
+  ],
   "scripts": {
     "build": "tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
     "test": "jest",
--- a/scripts/release_prep.sh
+++ b/scripts/release_prep.sh
@@ -44,23 +44,45 @@ if [[ " $* " == *" --recut "* ]]; then
   may_force_push=-f
 fi
 
-# Perform the version bumps, generate changelog and create local tag. 
+# Perform the version bumps, generate changelog and create local tags.
+# 1. Monorepo release (.internal.versionrc.js) — bumps apps/*/package.json,
+#    packages/shared/package.json, etc. and creates a `v<version>` tag.
 pnpm exec commit-and-tag-version --config .internal.versionrc.js ${tag_force}
 release_version_num=$(jq -r .version < package.json)
 release_version="v${release_version_num}"
 release_branch=release_${release_version}
 
+# 2. SDK release (.external.versionrc.js) — bumps packages/sdk/package.json,
+#    updates packages/sdk/CHANGELOG.md and creates a `sdk-v<version>` tag.
+#    Only runs if there are new conventional commits touching packages/sdk
+#    since the last sdk-v* tag. If there are none, commit-and-tag-version
+#    exits non-zero and we leave sdk_release_version empty so the downstream
+#    push/recut steps skip it.
+sdk_release_version=
+if pnpm exec commit-and-tag-version --config .external.versionrc.js ${tag_force}; then
+  sdk_release_version_num=$(jq -r .version < packages/sdk/package.json)
+  sdk_release_version="sdk-v${sdk_release_version_num}"
+else
+  echo "No new SDK commits since last sdk-v* tag; skipping external (SDK) version bump."
+fi
+
 if [[ " $* " == *" --recut "* ]]; then
-  git push --delete origin ${release_version}
+  git push --delete origin ${release_version} || true
+  if [[ -n "${sdk_release_version}" ]]; then
+    git push --delete origin ${sdk_release_version} || true
+  fi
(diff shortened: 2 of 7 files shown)


packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@opengovsg/formsg-sdk",
-  "version": "0.16.1",
+  "version": "7.0.0",
```
kevin9foong, line 3: note: this change will change the published version for consumers to v7.

eliotlim: yup this is intended! the first published version will be `v7.1.0`

scripts/release_prep.sh
```diff
@@ -44,23 +44,45 @@ if [[ " $* " == *" --recut "* ]]; then
   may_force_push=-f
 fi
 
-# Perform the version bumps, generate changelog and create local tag. 
+# Perform the version bumps, generate changelog and create local tags.
```
kevin9foong, line 47: praise: thanks for leaving clear docs about what each phase in this script does

scripts/release_prep.sh
```diff
@@ -44,23 +44,45 @@ if [[ " $* " == *" --recut "* ]]; then
...
+#    push/recut steps skip it.
+sdk_release_version=
+if pnpm exec commit-and-tag-version --config .external.versionrc.js ${tag_force}; then
+  sdk_release_version_num=$(jq -r .version < packages/sdk/package.json)
+  sdk_release_version="sdk-v${sdk_release_version_num}"
+else
+  echo "No new SDK commits since last sdk-v* tag; skipping external (SDK) version bump."
+fi
+
 if [[ " $* " == *" --recut "* ]]; then
```
kevin9foong, line 69: praise: good to see recut is supported also

packages/sdk/package.json
eliotlim: as discussed, we'll sync the major versions of the app and SDK, and allow the minor and patch versions to be incremented independently.

scripts/release_prep.sh
```diff
@@ -44,23 +44,45 @@ if [[ " $* " == *" --recut "* ]]; then
...
+  fi
   git branch -D ${release_branch}
 fi
 
 git checkout -b ${release_branch}
 git branch -D ${temp_release_branch}
 
-# Push the code and tag to origin
+# Push the code and tags to origin
 git push origin ${may_force_push} HEAD:${release_branch}
```
kevin9foong, line 81: note to self: not an issue related to this PR, but our repo has an implicit agreement that new devs should name their git remote `origin`

scripts/release_prep.sh
kevin9foong, line 47: issue(for discussion): if we put this together in the main release prep script, does it mean that everytime we publish to sdk, we also need to release our app code? 

would it make sense to split this sdk release into its own release file eg `scripts/release_sdk` so we have more control in the future over what we wish to release? 

we can then invoke this `bash scripts/release_sdk` script using some flags from the main file if desired. 

this also has the effect of keeping scripts scoped to specific parts (similar to splitting app code into their own specific files) for maintainability.

eliotlim: mmm good question! here's my considerations:
- in all likelihood, we always want the FormSG app server to be running the latest version of the SDK
- therefore, releasing a new SDK is a necessary and sufficient condition for releasing a new version of the FormSG app
- the inverse isn't true -- we'll update the app-server more often than the SDK.

to avoid accidentally forgetting to release the SDK, it makes sense to bake the SDK release process into the app release process, since SDK releases are a strict subset of app releases.

kevin9foong: sure, would it make sense to have a release_sdk script storing the sdk release process and invoke it from the main release prep file? the main benefit would be the release sdk could be reused by eg, the hotfix script and possibly standalone 

and also allow for the scripts to be scoped for maintainability

packages/sdk/package.json
kevin9foong: yep aligned - just leaving an internal note for us!

scripts/release_prep.sh
eliotlim: I gave this a try, but realised that we're going to have to make some CI changes to support this:
- the SDK release process needs to commit a version bump to the repository
- this is only possible on the release-al2 PR at the moment, which we achieve through the release_prep.sh script
- we could get around this by creating a release-sdk PR as well, but that means we now have 2x the number of PRs to review and merge per release

until we overhaul the release process' version bumps to run completely in CI (as opposed to on a developer machine), i'm afraid having the SDK release process integrated into release_prep.sh is the straightforward way to move forward.

kevin9foong (approved):
LGTM

scripts/release_prep.sh
kevin9foong: thanks for giving this a shot - good call on the 2x PR issue that i didnt catch!

PR #9318, opened 2026-04-17
feat: support single submission per NRIC for MRF step 1 

kevin9foong:
## Problem

To converge storage and mrf modes while minimising user impact, parity for single submission per NRIC for step 1 MRF is required.  
Closes FRM-2366

## Solution

In order to support future expansion of NRIC features to subsequent steps, the submitterId is tied to a specific step number. The hashedSubmitterId is stored in the existing as a key of each element in the submitterSteps array.  

**Breaking Changes** 

## Tests 
Regression test: Storage mode single submission works 
- [x] Create a new storage mode form and enable singpass + single submission 
- [x] Submit the form with NRIC A
- [x] Submit the form with NRIC A again, assert it is blocked
- [x] Submit the form with NRIC B, assert it is not blocked 
- [x] Turn off single submission 
- [x] Submit the form with step A, assert it is now not blocked
- [x] Submit the form with step C
- [x] Turn on single submission again 
- [x] Submit the form with step C, assert it is blocked since we submitted with step C earlier, despite single submission being off. 

Mrf mode single submission works 
- [x] Create new multi respondent form and enable singpass + single submission with 2 steps 
- [x] Submit the form with NRIC A in step 1
- [x] Submit the form with NRIC A again in step 1, assert it is blocked
- [x] Submit the form with NRIC B in step 1, assert it is not blocked 
- [x] Try and submit step 2, assert it is not blocked 
- [x] Turn off single submission 
- [x] Submit the form with step A in step 1, assert it is now not blocked
- [x] Submit the form with step C in step 1
- [x] Turn on single submission again 
- [x] Submit the form with step C in step 1, assert it is blocked since we submitted with step C earlier, despite single submission being off.


apps/backend/src/app/modules/form/form.service.ts
```diff
@@ -387,25 +389,38 @@ export const checkHasSingleSubmissionValidationFailure = (
...
-        message: 'Error while counting submissions for form',
-        meta: logMeta,
-        error,
-      })
-      return transformMongoError(error)
-    },
-  ).andThen((result) => okAsync(result !== null))
+
+  const mrfStepOneIndex = 0
+  // NOTE: Hardcoded to step one since single submission is only supported for step one for now.
```
kevin9foong, line 399: note: this can be extended when support for eg, step 2 is needed by passing the step number argument

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -907,12 +913,25 @@ export const handleNdiResponses = async (
...
+
+      const submitterId = userName?.toUpperCase()
+      if (!submitterId) {
+        const missingSubmitterIdError = new MissingSubmitterIdError()
+        const { statusCode, errorMessage } = mapRouteError(
+          missingSubmitterIdError,
+        )
+        return res.status(statusCode).json({ message: errorMessage })
+      }
+      const hashedSubmitterId = generateHashedSubmitterId(submitterId, formId)
```
kevin9foong, line 926: note: if singpass is enabled, we will query for the submitter id and hash it for saving

apps/backend/src/app/models/submission.server.model.ts
```diff
@@ -562,6 +563,10 @@ export const MultirespondentSubmissionSchema = new Schema<
     type: Array,
     default: [],
   },
+  submitterIds: {
+    type: [String],
+    default: [],
+  },
```
eliotlim: question: would putting the submitterId into the submittedSteps object be a better architectural decision:
1. easier to introduce variable steps and conditional / routing / loops in workflows later
2. possibility of having different identity providers for multi-step respondent auth: Singpass, AdminEmail, SsoWogAd, etc.

apps/backend/src/app/modules/submission/submission.errors.ts
```diff
@@ -318,3 +318,9 @@ export class MrfReminderRecipientEmailsEmptyError extends ApplicationError {
     )
   }
 }
+
+export class MissingSubmitterIdError extends ApplicationError {
+  constructor(message = 'Failed to retrieve submitter ID from Singpass.') {
```
kevin9foong, line 323: note: Instead of throwing an application error previously, a new error code is thrown and its logged in DD if this error occurs for tracking.

apps/backend/src/app/modules/submission/submission.utils.ts
```diff
@@ -197,6 +199,16 @@ const errorMapper: MapRouteError = (
...
       }
+    case FormRespondentSingleSubmissionValidationError:
+      return {
+        statusCode: StatusCodes.BAD_REQUEST,
+        errorMessage: error.message,
+      }
+    case MissingSubmitterIdError:
+      return {
+        statusCode: StatusCodes.INTERNAL_SERVER_ERROR,
+        errorMessage: coreErrorMessage,
```
kevin9foong, line 210: note: return a generic error as end user does not need to know the specifics

apps/backend/src/app/models/submission.server.model.ts
kevin9foong: verified that there is no significant performance tradeoff for storing as a key in object array, and applied this suggestion!

apps/backend/src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.middleware.spec.ts
```diff
@@ -392,6 +393,234 @@ describe('Multirespondent Submission Middleware', () => {
       jest.resetAllMocks()
     })
 
+    describe('hashedSubmitterId is set', () => {
```
kevin9foong, line 396: automated tests written across various test files to ensure that storage mode + mrf single submitter per id works: 

for mrf at submission time: 
- ensure that the hashedSubmitterId is set if required 
- use the hashedSubmitterId to determine if duplicate exists before saving if needed

for public form at view time: 
- if mrf, query for the hashedSubmitterId in the submittedSteps array 
- if storage mode, query for the hashedSubmitterId directly

for storage mode at submisson time: 
- unchanged test cases from prev

apps/backend/src/app/models/submission.server.model.ts
```diff
@@ -614,6 +615,58 @@ MultirespondentSubmissionSchema.methods.getWebhookView = async function (
...
+    const afterCreateRes = await this.find(
+      {
+        form: formId,
+        submissionType: SubmissionType.Multirespondent,
+        [submitterIdKey]: submitterId,
+      },
+      null,
+      {
+        limit: 2,
+        readPreference: 'primary',
```
kevin9foong, line 656: note: to prevent race condition during save (write operation), the read must be done in the primary to avoid edge cases due to replication lag. 

during the view form phase (read operation), this is not enforced as the chances of race condition happening is low and the performance tradeoff can be high since all traffic must go to the primary.

apps/backend/src/app/models/submission.server.model.ts
eliotlim: awesome 👍 💯 thanks for thinking ahead with me.

apps/backend/src/app/modules/submission/multirespondent-submission/__tests__/multirespondent-submission.service.spec.ts
```diff
@@ -178,6 +175,7 @@ describe('multirespondent-submission.service', () => {
       )
       expect(saveProtoSpy).not.toHaveBeenCalled()
       expect(result.isOk()).toBe(true)
+      expect(result._unsafeUnwrap()).toEqual(mockSavedDoc)
```
kevin9foong, line 178: rationale: asserts that the submission id returned used for emailing next step is the exact same as the value returned by the respective save functions

eliotlim (approved):
LGTM! 👍

PR #9322, opened 2026-04-20
fix(ci): ignore sdk bump commit

eliotlim:
## Problem

When the SDK gets updated, a version bump commit is created.

This creates a self-perpetuating cycle: each bump commit touches packages/sdk/package.json, which becomes the only "SDK commit" triggering the next spurious bump (e.g. sdk-v7.1.1 and sdk-v7.1.3 were both empty releases).

Root cause: commit-and-tag-version exits 0 and performs a patch bump even when zero commits match the path filter. release_prep.sh assumed it would exit non-zero in this case.

Related to #9305, #9311.

## Solution

Added an explicit git log guard in release_prep.sh that checks for actual commits touching packages/sdk/ since the last sdk-v* tag before invoking commit-and-tag-version.

**Breaking Changes** 


kevin9foong (approved):
LGTM

PR #9328, opened 2026-04-22
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `apps/backend/src/app/modules/auth/constants.ts:L1`

The default bcrypt salt rounds is set to 2, which is far below recommended values for production password/OTP hashing. A low work factor significantly reduces attacker effort for offline brute-force if hashes are exposed.

## Solution

Increase the default cost factor to a secure baseline (e.g., 10-12+ depending on performance budget), make it environment-configurable, and enforce a minimum in non-test environments.

## Changes

- `apps/backend/src/app/modules/auth/constants.ts` (modified)
- `apps/backend/src/app/config/features/turnstile.config.ts` (modified)

### DIFF
--- a/apps/backend/src/app/config/features/turnstile.config.ts
+++ b/apps/backend/src/app/config/features/turnstile.config.ts
@@ -24,6 +24,22 @@ const turnstileSchema: Schema<ITurnstile> = {
   },
 }
 
-export const turnstileConfig = convict(turnstileSchema)
+const turnstileConfig = convict(turnstileSchema)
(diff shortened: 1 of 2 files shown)


kevin9foong:
Thanks for the contribution - though this code is actually not being used and should likely be removed instead of increasing the salt rounds.

kevin9foong:
closed since we have removed the referenced unused code

PR #9335, opened 2026-04-23
feat(frontend): use template nudges

eliotlim:
## Problem

Admins sometimes get confused by the Use Template page.
- their intention is to create a form using a template
- but by being brought to the form preview page, they mistakenly fill in the form instead. 

Closes FRM-2386

## Solution

Introduce three mechanisms to nudge admins towards the "Use this template" flow.

**Breaking Changes** 

**Features**:

- Add a clearer "Template Preview" frame to 
- Introduce a "wall" modal which pops up once the admin scrolls 1.5 viewports or to the end of the page.
- Introduce a tooltip to highlight the "Use this Template" button

**Improvements**:

- Add i18n translations for several controls in the Use Template flow.

## Before & After Screenshots

| Before | After |
| -------- | ------ |
| before-use-template-nudges | formsg-use-template-nudges-3 |

## Tests

**TC: Use Template Wizard works**
- [x] Open an existing form with at least two pages of fields
- [x] Click Share > Template > Visit the Link
- [x] Frame: should always be present around the Form preview
- [x] Scroll down: the modal should appear
- [x] Click Submit: the modal should appear
- [x] Tooltip: should always be present on the "Use this Template" button

## Deploy Notes

**New feature flags**:

- `use-template-frame`: Show / hide the yellow-bordered frame around the Use Template page
- `use-template-tour` : Show / hide the tour tooltip that tracks the "Use this Template" button
- `use-template-wall-scroll` : Show / hide the "wall" modal that activates at a scrolling threshold
- `use-template-wall-submit` : Show / hide the "wall" modal that when the "Submit" button is pressed


apps/frontend/src/features/admin-form/template/UseTemplateNudges/UseTemplateTour.tsx
```diff
@@ -0,0 +1,72 @@
...
+      setStickyActive(scrollTop > STICKY_ACTIVATION_OFFSET_PX)
+    }
+    onScroll()
+    window.addEventListener('scroll', onScroll, { passive: true })
+    document.addEventListener('scroll', onScroll, { passive: true })
+    return () => {
+      window.removeEventListener('scroll', onScroll)
+      document.removeEventListener('scroll', onScroll)
+    }
+  }, [])
```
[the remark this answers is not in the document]
eliotlim: fixed! we now use 75% of viewport height as the scroll trigger distance.

packages/shared/.eslintignore
```diff
@@ -0,0 +1 @@
+dist
```
scottheng96, line 1: Was this intentional to save time scanning these generated files?

scottheng96 (commented):
Looks good to me! There's a small UX nit that I don't think is blocking, basically when when I scroll and the banner transitions to sticky mode, there's a lag before the tooltip repositions to track the sticky button.

Image

Probably because the tour is using a scroll event listener while the banner is using a Waypoint, why might not synchronise at the same trigger timing. Brainstorming a potential solution, to move the stickyActive variable to the parent component and pass it down to the banner & tour.

In TemplateFormPage
```
`const { isOpen: isStickyActive, onOpen, onClose } = useDisclosure()`
then
`        <PreviewFormBannerContainer 
          isTemplate 
          isStickyActive={isStickyActive}
          onStickyChange={{ onOpen, onClose }}
        />
...
<UseTemplateTour stickyActive={isStickyActive} />
```

Haven't verified, but considering users don't scroll so slowly - I don't think it matters

scottheng96 (approved):

packages/shared/.eslintignore
eliotlim: yup we shouldn't lint generated files!
- they'll be overwritten on every build, which means that there's not much point linting them
- and dist isn't meant for human readability anyway

eliotlim:
I gave the tour improvement a try but wasn't able to get it working satisfactorily at the moment.
Given the need to solve the ops blockers sooner rather than later, we'll release this and remove any unnecessary components later.

PR #9368, opened 2026-04-28
feat: support submitter id whitelisting for MRF

kevin9foong:
## Problem

Storage and multi respondent modes cause user friction. To unify them, parity of submitter id whitelisting in the first step is needed.

## Solution

Support whitelisting of form respondents by submitter id for MRF. 

**Breaking Changes** 

## Tests

TC1: Whitelist submitter id works for Myinfo MRF >=2 steps 
- [x] Create MRF with >=2 steps. 
- [x] Add a whitelist for 1-2 NRICs 
- [x] Try to login to a form as respondent using another NRIC, assert it is blocked. 
- [x] Try to use a whitelisted NRIC, assert you can see the form fields as a respondent.
- [x] Try to submit the form step 1, assert it is successful. 
- [x] Login again with the whitelisted NRIC and go to the form fields page. 
- [x] Update the admin whitelist setting to remove the logged in NRiC. 
- [x] Try to submit the form, it should be blocked with a toast. 

TC2: Works with singpass single submission 
- [x] Whitelist the logged in NRIC again. 
- [x] Turn on single submission. 
- [x] Submit the form with the logged in NRIC, it should now be blocked showing the single submission message as error. 
- [x] Turn off singpass entirely, leaving the single submission and whitelist in place.
- [x] Go back and try to submit, the login should be gone and submission for step 1 works. 

TC3: Step 2 and beyond works and is unaffected by the whitelist setting
- [x] Check your email inbox for the step 2 submissions from TC1. 
- [x] Remove the logged in ID from step 1 from the whitelist. 
- [x] Submit the step 2. 
- [x] Assert you can submit step 2 and it does not require a login id and it does not check based on the whitelist.

TC4: Storage mode is not affected 
- [x] Try and submit a singpass enabled storage mode form without the whitelist. Assert it works. 
- [x] Enable whitelist. 
- [x] Assert login and submit works for whitelisted NRIC.
- [x] Assert login and submit is blocked with expected error infobox for non-whitelisted NRIC.


apps/backend/src/app/modules/submission/submission.errors.ts
```diff
@@ -320,7 +320,7 @@ export class MrfReminderRecipientEmailsEmptyError extends ApplicationError {
 }
 
 export class MissingSubmitterIdError extends ApplicationError {
-  constructor(message = 'Failed to retrieve submitter ID from Singpass.') {
+  constructor(message = 'Failed to retrieve submitter ID. Please try again.') {
```
kevin9foong, line 323: rationale: keep this message generic, since submitter id can be from CP/SP or other providers in the future.

apps/backend/src/app/modules/form/__tests__/form.service.spec.ts
```diff
@@ -82,6 +90,319 @@ describe('FormService', () => {
...
+      const result = await FormService.checkHasRespondentNotWhitelistedFailure(
+        form,
+        'mockSubmitterId',
+      )
+
+      expect(result.isOk()).toBe(true)
+      expect(result._unsafeUnwrap()).toBe(false)
+    })
+
+    it('should return false if form auth type is nil', async () => {
```
kevin9foong, line 195: note: covers the earlier incident with single sub to prevent re-ocurrence

eliotlim: noice 💯

kevin9foong:
these copilot reviews are good! 

ive resolved all of them - with the exception of the submitterId in middleware propagation, since we likely do not want to do the check in the middleware and want to do it in the ensures pipeline instead.

apps/backend/src/types/form.ts
```diff
@@ -379,6 +379,7 @@ export interface IMultirespondentForm extends IForm {
   stepsToNotify: string[]
   stepOneEmailNotificationFieldId: string
   hasStatusTracker: boolean
+  whitelistedSubmitterIds?: WhitelistedSubmitterIds
```
eliotlim, line 382: question: does this create the implicit assumption that the entire form (including all steps) will share the `whitelistedSubmitterIds`?

1. this might work when we want to enforce `whitelistedSubmitterIds` on the first step
2. but on the second step onwards, we'd have to extend this construct.
3. this might be true for now since access control is governed by possession of a one-time link sent by email
4. but would complicate us adding per-step access control / whitelisting in the future

what if we considered an extension like this:

```suggestion
  submissionAccessControl?: {
    type: 'form',
    whitelistedSubmitterIds: [],
  } | {
    type: 'steps',
    steps: [],
  },
```

apps/backend/src/app/models/form.server.model.ts
```diff
@@ -463,6 +473,12 @@ const MultirespondentFormSchema = new Schema<IMultirespondentFormSchema>({
   },
 })
 
+MultirespondentFormSchema.methods.getWhitelistedSubmitterIds = function () {
+  return this.get('whitelistedSubmitterIds', null, {
+    getters: false,
+  })
+}
+
```
eliotlim, line 481: note: to build two-way doors, we could consider how this can be extended:

(e.g. to help us extend multi-step submission access control in the future):
- we could consider how we might pass a bit more context
- at least the step number

```suggestion
MultirespondentFormSchema.methods.getWhitelistedSubmitterIds = function (ctx: {stepNumber: number}) {
  return this.get('whitelistedSubmitterIds', null, {
    getters: false,
  })
}

```

eliotlim (commented):
quick check in on the extensibility of this design: would this be able to support multi-step submission access control in the future?

apps/backend/src/types/form.ts
eliotlim: right, we'd discussed this previously and agreed that this would be a good extension when we fix response limits as well.

eliotlim (approved):
Tested all TCs and everything works on stg-alt3! LGTM 👍 

- We'd [discussed the extensibility considerations previously](https://opengovproducts.slack.com/archives/C097RMGC1LK/p1776826042911989?thread_ts=1776663789.931239&cid=C097RMGC1LK)
- and agreed that this would be a good extension when we fix response limits as well.

PR #9371, opened 2026-04-29
feat: answer object decryption

scottheng96:
## Problem

Closes FRM-2388 
Closes FRM-2389

## Solution

1. Introduced answer object schema (labelled as v4) - naming is up for discussion
2. Implemented usage of answer object schema into decryption flows for MRF:
- decrypting responses when displaying in admin portal (IndividualSubmissionPage)
- decrypting responses when downloading responses (in csv etc.)
3. answerObject decryption flow is gated under feature-flag `answer-object

There should be no change to the outputs and admin user experience. This is a validation that the new schema is backwards compatible and decryption flows in our app can be supported.

**Breaking Changes** 

## Tests

**TC1: Decrypting responses in admin portal**
- [ ] Prepare/use an MRF form with responses including at least 1 basic field, 1 checkbox with others, 1 dropdown with others, 1 table field, 1 address field and at least 1 myinfo field
- [ ] Turn on feature flag `answer-object-decryption` in the testing env
- [ ] Visit the response page for an individual submission, observe that all responses are accurately displayed
- [ ] Turn off the feature flag `answer-object-decryption`
- [ ] On the same individual submisson page, observe that there is no regression to the responses displayed

**TC2: Decrypting responses for download**
- [ ] Prepare/use an MRF form with responses including at least 1 basic field, 1 checkbox with others, 1 dropdown with others, 1 table field, 1 address field and at least 1 myinfo field
- [ ] Turn on feature flag `answer-object-decryption` in the testing env
- [ ] Download responses (select all csv, PDF, attachments)
- [ ] Turn off the feature flag `answer-object-decryption`
- [ ] Download responses (select all csv, PDF, attachments)
- [ ] Verify that both downloaded content is the same: (1) No regression in csv (2) Attachments are successfully downloaded and zipped (3) No regression in PDF


apps/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -121,11 +124,14 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+  const useV4Display = useFeatureIsOn(featureFlags.answerObject)
 
   const { user } = useUser()
   const { secretKey } = useStorageResponsesContext()
-  const { data, isLoading, isError } = useIndividualSubmission()
-
+  const { data, isLoading, isError } = useIndividualSubmission({
+    useV4: useV4Display,
+  })
+  console.log(`data:`, data)
```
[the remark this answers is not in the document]
eliotlim: TODO: we should remove this before merging the PR.

packages/shared/constants/feature-flags.ts
```diff
@@ -33,7 +33,7 @@ export const featureFlags = {
   ogpSpinner: 'ogp-spinner' as const,
   wogadLogin: 'wogad-login' as const,
   mrfResponseLimit: 'mrf-response-limit' as const,
-  answerObject: 'answerObject' as const,
+  answerObject: 'answer-object' as const,
```
eliotlim: question: perhaps we should split this into separate feature flags:
- answer-object-decryption
- answer-object-submission
- answer-object-encryption

```suggestion
  answerObjectDecryption: 'answer-object-decryption' as const,
```

packages/sdk/spec/resources/crypto-v3-data-20260428.ts
```diff
@@ -0,0 +1,322 @@
...
+      fieldType: 'date',
+      answer: '02/09/2028',
+    },
+    '69eedf89fd2757b0584e3a9b': {
+      fieldType: 'textfield',
+      answer: 'SDK6800X',
+    },
+    '69eedf91ed3e04f8606c401c': {
+      fieldType: 'textfield',
+      answer: '752 BEDOK RESERVOIR ROAD, #05-08, SINGAPORE 479257',
```
eliotlim: please tell me this isn't a real address 😅

packages/sdk/src/adapt-v3-to-v4.ts
```diff
@@ -0,0 +1,205 @@
...
+    value: answer.answer,
+    hasBeenScanned: answer.hasBeenScanned,
+    ...(answer.md5Hash !== undefined && { md5Hash: answer.md5Hash }),
+  }
+}
+
+const convertTableAnswer = (answer: Record<string, string>[]): TableAnswerV4 => {
+  const result: TableAnswerV4 = {}
+  for (let i = 0; i < answer.length; i++) {
+    const rowId = `row${i}`
```
eliotlim: note:
- perhaps we should consider using a uuid or random id instead of running numbers
- let's say we append a row in the future:
  - how do we determine what row id to insert?
  - do we increment all the other row ids (very expensive)
  - do we look up the last row id and + 1? (not feasible, there may have been a row + 1 moved into the middle of the table)

using a random row id makes it immediately less likely for people to depend on `row${index}` behaviour by accident later on:
- they might be tempted to iterate through the table like this: row = table[`row${index++}`]
- which would error if a row was deleted
- and behave incorrectly if table rows were re-ordered

eliotlim (commented):
good progress! I like that this starts with the two-way door: a risk-managed way to test decryption before irreversibly encrypting responses in the new format.

three comments to consider before we move forward.

packages/sdk/src/adapt-v3-to-v4.ts
```diff
@@ -0,0 +1,205 @@
...
+  'buildingName',
+  'levelNumber',
+  'unitNumber',
+] as const
+
+const convertStringAnswer = (answer: string): StringAnswerV4 => {
+  return { value: answer }
+}
+
+const convertYesNoAnswer = (answer: string): YesNoAnswerV4 => {
```
kevin9foong, line 26: question: is this typesafe to cast as 'Yes' | 'No'? would it make sense to have a check here to ensure the value is equals to 'Yes' or 'No' before casting? 

eg, if answer ==='Yes' || 'answer' ==='No', else throw error

this might be crucial since we are relying on this converter for all submissions and this typecast assumes the value is yes no, checking here could assert this assumption and save us from potential value inconsistency issues down the road and notify us early if there is such inconsistency.

packages/sdk/src/adapt-v3-to-v4.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+  return { value: answer.value, isOthersInput: false }
+}
+
+const convertCheckboxAnswer = (answer: {
+  value: string[]
+  othersInput?: string
+}): CheckboxAnswerV4 => {
+  return {
+    value: answer.value,
```
kevin9foong, line 57: question: checkbox has a potential value called `CHECKBOX_OTHERS_INPUT_VALUE`, do we still need to store this value in schema v4?

packages/sdk/src/adapt-v3-to-v4.ts
```diff
@@ -110,15 +109,16 @@ const convertChildrenAnswer = (answer: {
...
     for (let j = 0; j < answer.childFields.length; j++) {
       const attr = answer.childFields[j]
-      entry[attr] = {
+      value[attr] = {
         value: answer.child[i][j] ?? '',
         myInfo: { attr },
       }
     }
-    result[childKey] = entry
+    result[childKey] = { value }
```
kevin9foong, line 101: question: in the docs, 
the schema is 
```
Name",
"myInfo": {
"attr": "childname"
},
``` 
however, in this implementation, 

```
"Name": {
  "value": { 
"myInfo": {
"attr": "childname"
},
  }
} 
```
seems to be the schema - is this expected? otherwise, should we update the docs to match?

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
```diff
@@ -0,0 +1,290 @@
...
+        return (
+          <DecryptedSignatureRowV4
+            field={field}
+            questionNumber={questionNumber}
+          />
+        )
+      case BasicField.Checkbox: {
+        const checkboxAnswer = field.answer as CheckboxAnswerV4
+        const selectedValues = checkboxAnswer.value.filter(
+          (v) => v !== '!!FORMSG_INTERNAL_CHECKBOX_OTHERS_VALUE!!',
```
kevin9foong: issue: seems like TS is throwing a warning here - Parameter 'v' implicitly has an 'any' type.

kevin9foong: issue: should we import the constant instead of using the string literal here? 
eg, exported `CHECKBOX_OTHERS_INPUT_VALUE` in the constants file

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
```diff
@@ -0,0 +1,168 @@
...
+  StringAnswerV4,
+  TableAnswerV4,
+  VerifiableAnswerV4,
+} from '@opengovsg/formsg-sdk'
+import { FormField } from '@opengovsg/formsg-sdk/dist/types'
+
+import { BasicField } from 'formsg-shared/types'
+
+const OTHERS_PREFIX = 'Others: '
+const CHECKBOX_OTHERS_VALUE = '!!FORMSG_INTERNAL_CHECKBOX_OTHERS_VALUE!!'
```
kevin9foong: issue: should we use the exported `CHECKBOX_OTHERS_INPUT_VALUE` instead of redeclaring a constant here?

packages/sdk/spec/adaptor/adapt-v3-to-v4.spec.ts
```diff
@@ -0,0 +1,439 @@
...
+      const result = adaptV3ToV4(v3)
+
+      expect(result.field1.answer).toEqual({
+        value: '+6591234567',
+        signature: 'sig',
+      })
+    })
+  })
+
+  describe('radiobutton field', () => {
```
kevin9foong, line 114: suggestion: shall we add a TC to test for the edge case regarding `'!!FORMSG_INTERNAL_RADIO_OTHERS_VALUE!!'` as well?

packages/sdk/spec/adaptor/adapt-v3-to-v4.spec.ts
```diff
@@ -0,0 +1,439 @@
...
+      const result = adaptV3ToV4(v3)
+
+      expect(result.field1.answer).toEqual({
+        value: 'Custom answer',
+        isOthersInput: true,
+      })
+    })
+  })
+
+  describe('checkbox field', () => {
```
kevin9foong, line 148: suggestion: shall we add a TC to cover the `'!!FORMSG_INTERNAL_CHECKBOX_OTHERS_VALUE!!'` edge case as well?

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -49,3 +50,36 @@ export const augmentDecryptedResponses = (
...
+
+    return {
+      fieldId,
+      field,
+      questionNumber: isNonResponse
+        ? undefined
+        : index + 1 - nonResponseFieldsCount,
+      downloadUrl: attachmentMetadata[fieldId],
+    }
+  })
```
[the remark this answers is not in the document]
kevin9foong: TODO: evaluate if we are rendering CSV, admin page using the snapshotted form def which has the ordering. If we are using the ordering of the snapshotted form def, this might not be an issue

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -125,24 +130,53 @@ async function decryptSubmissionData(
...
         }
+        mrfSubmissionSecretKey = decryptedV4.submissionSecretKey
+        const v4Responses = decryptedV4.verified
+          ? {
+              ...decryptedV4.responses,
+              ...convertVerifiedToV4(decryptedV4.verified),
+            }
+          : decryptedV4.responses
+        decryptedResponses = flattenV4ToFormFields(v4Responses)
+      } else {
```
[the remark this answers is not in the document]
kevin9foong: TODO: evaluate if we are rendering CSV, admin page using the snapshotted form def which has the ordering. If we are using the ordering of the snapshotted form def, this might not be an issue

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
```diff
@@ -0,0 +1,290 @@
...
+      }
+      case BasicField.Radio: {
+        const radioAnswer = field.answer as RadioAnswerV4
+        return (
+          <Stack>
+            <DecryptedQuestionLabelV4
+              field={field}
+              questionNumber={questionNumber}
+            />
+            <Text textStyle="body-1">{radioAnswer.value}</Text>
```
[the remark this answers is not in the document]
kevin9foong: TODO: verify if this comment is true, if so, could be something we want to address

kevin9foong:
nice work with the PR and appreciate the effort to use review by commit which made this PR much easier to understand and look through! 

this is a big step towards unifying the modes and the changes mostly look good - just a few clarifying questions and suggestions to help cover some edge cases and reuse common exports 

ive looked through the comments by copilot, seems like we just need to verify the one about the `Others:` and address it if its true and check if we are using the snapshotted form def to decide if the ordering in the answer object matters

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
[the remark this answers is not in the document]
scottheng96: This is because `Others: ` gets appended to the value during decryption. No change needed here

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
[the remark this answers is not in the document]
scottheng96: Fixed this as per comment.

packages/sdk/src/types-v4.ts
```diff
@@ -0,0 +1,180 @@
...
+    | 'decimal'
+    | 'textfield'
+    | 'textarea'
+    | 'homeno'
+    | 'dropdown'
+    | 'rating'
+    | 'nric'
+    | 'uen'
+    | 'date'
+    | 'country_region'
```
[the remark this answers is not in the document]
scottheng96: Fixed as per comment

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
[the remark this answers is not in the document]
scottheng96: Fixed as per comment

packages/sdk/src/adapt-v3-to-v4.ts
scottheng96: cmiiw, I believe the `value` and `myInfo` variables will be on the same level through this. Did a second pass to check this
```
    value[attr] = {
      value: answer.child[i][j] ?? '',
      myInfo: { attr: attr },
    }

e.g.
{
  "child0": {
    "value": {
      "childgender": { "value": "MALE", "myInfo": { "attr": "childgender" } },
      "childname":   { "value": "Alice", "myInfo": { "attr": "childname" } }
    }
  }
}
```

packages/sdk/spec/adaptor/adapt-v3-to-v4.spec.ts
scottheng96: From my read of its usage, `'!!FORMSG_INTERNAL_RADIO_OTHERS_VALUE!!'` doesn't get saved in the responses - it's used during response [rehydration](https://github.com/opengovsg/FormSG/blob/ae32227fa15f757f4794a8dcc90e05f5895c0382/apps/frontend/src/features/form/utils/extractMrfPreviousStepResponseValue.ts#L47), maybe to allow the FE to handle a radio other's input.

This is how the data gets saved in db:
```
    '69eeddcf8844a134ddbadc56': {
      fieldType: 'radiobutton',
      answer: {
        value: 'Option 2',
      },
    },
    '69eeddd53c9ffa7a2b464687': {
      fieldType: 'radiobutton',
      answer: {
        othersInput: 'adg',
      },
    },
```

packages/sdk/spec/adaptor/adapt-v3-to-v4.spec.ts
scottheng96: Good idea!

packages/sdk/src/adapt-v3-to-v4.ts
scottheng96: Updated as per comment!

packages/sdk/src/adapt-v3-to-v4.ts
scottheng96: Yeah, you make a good point. I've updated the code to reflect a randomly generated uuid for each row. One note, is that there different row uuids across responses, which I believe is intended for scenarios like you mentioned above ^. It doesn't break downstream since we don't use rowUUID for anything at the moment.

packages/shared/constants/feature-flags.ts
kevin9foong: this is resolved - marking it as so

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
```diff
@@ -121,11 +124,14 @@ export const IndividualResponsePage = (): JSX.Element => {
...
+  const useV4Display = useFeatureIsOn(featureFlags.answerObjectDecryption)
 
   const { user } = useUser()
   const { secretKey } = useStorageResponsesContext()
-  const { data, isLoading, isError } = useIndividualSubmission()
-
+  const { data, isLoading, isError } = useIndividualSubmission({
+    useV4: useV4Display,
+  })
+  console.log(`data:`, data)
```
kevin9foong: hey @scottheng96 , shall we remove this console log before merging it in?

kevin9foong (approved):
LGTM - please remove the last stray console log before merging in!

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/IndividualResponsePage.tsx
scottheng96: removed!

PR #9375, opened 2026-04-30
chore(auth): remove unused auth functions

kevin9foong:
## Problem

Remove unused auth functions, which were previously used as part of an experiment to expose some APIs for external users (https://github.com/opengovsg/FormSG/pull/6376/changes). 

## Solution

Since this experiment is not being done, these functions should be removed to clean up the codebase. 
Checked with engineering team that this is safe to remove, since this experiment is no longer being pursued, to adhere to chesterson's fence principle. 

**Breaking Changes** 


eliotlim (approved):
thanks for helping to clean this up!

PR #9383, opened 2026-05-04
feat: support charts for MRF

kevin9foong:
## Problem

To converge MRF and storage modes, parity for charts is a pre-requisite. 
This PR aims to support charts for MRF forms as well. 

Closes FRM-2397

## Solution

Previously, charts on storage mode used a charts specific endpoint and decryption flow. This includes querying 1000 submissions sorted by latest created submission document key (note that key is indexed).

However, we can extend and reuse the streaming endpoint used by the csv bulk download flow. This allows us to use streaming + web workers (allowing us to potentially increase the # of submissions supported by charts as a secondary objective) + avoids needing to create and maintain additional routes.

### Feature flagging: 
- A growthbook flag charts-max-response-count is used to control the max submissions supported - we can set it to an arbitrarily high value to make it "unlimited".

### Secondary improvements: 
Scale improvements: 
- The # of decryption workers follows the same values as the csv/pdf bulk download decrypt.

FE quality improvements: 
- Previously, during decryption / form loading, the 'no charts found' message is shown. Now, a skeleton is shown with a loading animation instead of the `no charts found` to feedback to the user that the page is loading.  

**Breaking Changes** 

## Tests

**TC1: Charts works on storage mode forms** 
- [x] Choose a storage mode form with small number of submissions eg, 5
- [x] Go to charts, enter secret key and assert the charts displays the correct values according to the responses and the decrypt count is correct.  
- [x] Choose a storage mode form with large number of submissions eg >= 100 
- [x] Go to charts, enter secret key and assert the charts displays the correct decrypted responses count. 

**TC2: Charts works on MR mode forms** 
- [x] Choose a MR mode form with small number of submissions eg, 5
- [x] Go to charts, enter secret key and assert the charts displays the correct values according to the responses and the decrypt count is correct.  
- [x] Choose a MR mode form with large number of submissions eg >= 100 
- [x] Go to charts, enter secret key and assert the charts displays the correct decrypted responses count. 

**TC3: Charts show empty when no supported charts fields** 
- [x] Create a storage mode form with non-charts supported field eg, only attachment 
- [x] Submit the form 
- [x] Assert that charts shows the empty chart screen, since attachment is not supported. 
- [x] Create a MR mode form with non-charts supported field eg, only attachment 
- [x] Submit the form 
- [x] Assert that charts shows the empty chart screen, since attachment is not supported.


apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -179,29 +187,61 @@ export const getAllDecryptedSubmission = async ({
...
+  const workerPool: CleanableDecryptionWorkerApi[] = []
+  const currentSubmissionIndex = 0
 
-  const allEncryptedData = await getAllEncryptedSubmission({
-    formId,
+  for (let i = workerPool.length; i < numWorkers; i++) {
+    workerPool.push(makeWorkerApiAndCleanup())
+  }
+
+  const submissionsStream = await getEncryptedResponsesStream(formId, {
```
kevin9foong, line 208: consideration: i believe charts should use a generic submissions fetch instead of a specialized charts fetch endpoint. 

this would help support long term scale of charts which needs to use streaming + avoid maintenance of unnecessary endpoints

apps/frontend/src/features/admin-form/responses/ChartsPage/queries.ts
```diff
@@ -24,10 +26,22 @@ export const useAllSubmissionData = (dateRange?: DateString[]) => {
...
-    () => getAllDecryptedSubmission({ formId, secretKey, startDate, endDate }),
+    () =>
+      getAllDecryptedSubmission({
+        formId,
+        secretKey,
+        startDate,
+        endDate,
+        downloadAttachments: false,
+        isSortByLatest: true,
+        limit: CHART_MAX_SUBMISSION_RESULTS,
```
kevin9foong: consideration: my main concern is whether this can be used as a form of DOS on this endpoint if the limit is specified in the FE. but the csv/attachment download would expose us to the same issue.  

after thought: 
this is a protected route so the exposure is limited to admin users, this is not a concern with proper audit logging that we have

apps/backend/src/app/modules/submission/submission.controller.ts
```diff
@@ -278,6 +278,8 @@ const validateDateRange = celebrate({
       startDate: Joi.date().format('YYYY-MM-DD').raw(),
       endDate: Joi.date().format('YYYY-MM-DD').min(Joi.ref('startDate')).raw(),
       downloadAttachments: Joi.boolean().default(false),
+      isSortByLatest: Joi.boolean().default(false),
+      limit: Joi.number().integer().min(1).optional(),
```
kevin9foong, line 282: note: joi validation + guard clause is enforced to validate the limit. we can also specify an upper bound to limit against DOS, but this existing endpoint does not currently have a limit and for future consideration

apps/backend/src/app/models/submission.server.model.ts
```diff
@@ -472,18 +474,24 @@ EncryptSubmissionSchema.statics.getSubmissionCursorByFormId = function (
...
-        encryptedContent: 1,
-        verifiedContent: 1,
-        attachmentMetadata: 1,
-        paymentId: 1,
-        created: 1,
-        version: 1,
-        id: 1,
-      })
+    query
       .batchSize(2000)
```
kevin9foong, line 495: note: the batch size and read on secondary can help to relief load on the DB

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -179,29 +188,68 @@ export const getAllDecryptedSubmission = async ({
...
+      )
+      currentSubmissionIndex++
+      return reader.read().then(read)
+    }),
+  )
 
-    return { ...decryptedContent, submissionTime: encryptedData.created }
+  const decryptionResults = await Promise.all(
+    decryptSubmissionPromises,
+  ).finally(() => {
```
kevin9foong, line 248: note: finally is used to ensure that the workers are killed regardless of success or failure/error. 
this is the same behavior as the bulk results (eg, csv, attachments, pdf) decryption workers.

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -179,29 +188,68 @@ export const getAllDecryptedSubmission = async ({
...
+        workerApi
+          .parseAndDecryptSubmissionData({
+            submissionStreamDtoString: result.value,
+            secretKey,
+            workerCtxOptions: {
+              formsgSdkMode: env.formsgSdkMode,
+            },
+          })
+          .then((result) => {
+            if (!result.isParseSuccessful || !result.isDecryptionSuccessful) {
```
kevin9foong, line 235: note: this fails fast and halts the function during partial decryption - similar to existing charts behavior.

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -1,23 +1,30 @@
...
+  makeWorkerApiAndCleanup,
+} from './common/utils/decryptionWorker'
+import { getEncryptedResponsesStream } from './ResponsesPage/storage/StorageResponsesService'
+import { CleanableDecryptionWorkerApi } from './ResponsesPage/storage/types'
 import { augmentDecryptedResponses } from './ResponsesPage/storage/utils/augmentDecryptedResponses'
 import {
   processDecryptedContent,
   processDecryptedContentV3,
 } from './ResponsesPage/storage/utils/processDecryptedContent'
+import { DecryptionResult } from './ResponsesPage/storage/worker/decryption.worker'
```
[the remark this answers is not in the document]
kevin9foong: fixed!

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -148,60 +154,76 @@ export const getDecryptedSubmissionById = async ({
...
-  endDate?: DateString
+  secretKey: string
+  startDate: string
+  endDate: string
+  downloadAttachments: boolean
+  isSortByLatest: boolean
+  limit: number
 }): Promise<DecryptedSubmission[]> => {
-  if (!secretKey) return []
+  const numWorkers = window.navigator.hardwareConcurrency ?? 1
```
kevin9foong, line 200: note: this should be safe as it is the same behavior as csv bulk download. note that more workers might not mean faster decryption due to overhead, but for simplicity of code, we use the default following the bulk downloads.

apps/backend/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
```diff
@@ -820,79 +804,3 @@ export const handleStorageSubmission = [
   EncryptSubmissionMiddleware.encryptSubmission,
   submitEncryptModeForm,
 ] as ControllerHandler[]
-
```
kevin9foong, line 823: note: the previous charts specific endpoints are removed.

apps/frontend/src/features/admin-form/responses/ChartsPage/UnlockedCharts/UnlockedChartsContainer.tsx
```diff
@@ -161,45 +177,47 @@ export const UnlockedChartsContainer = () => {
...
-        <EmptyChartsContainer
-          title={t(
-            'features.adminForm.responses.charts.emptyChartContainer.noSupportedFields.title',
-          )}
-          subtitle={t(
-            'features.adminForm.responses.charts.emptyChartContainer.noSupportedFields.subtitle',
-          )}
-        />
-      )}
+      <Skeleton isLoaded={!isLoading}>
```
kevin9foong, line 180: fe-quality: instead of showing the `features.adminForm.responses.charts.unlockedChartsContainer.noChartsForDateRange.title` block, we should show a skeleton loading animation if loading is still in progress.

apps/backend/src/app/modules/submission/encrypt-submission/encrypt-submission.controller.ts
eliotlim: note: we should leave previous charts endpoints to preserve backward compatibility
- there isn't a mechanism to bump all older client versions at this time
- and dropping the API route will be a breaking change for about 5% of users over 48 hours

eliotlim (commented):
praise: good work considering how to capitalise on our previous work on parallelised decryption with web workers to eliminate the previous Charts limit.

tested working on stg-alt3

one small note on API compatibility and deprecation of the older Charts endpoint.

kevin9foong:
> praise: good work considering how to capitalise on our previous work on parallelised decryption with web workers to eliminate the previous Charts limit.
> 
> tested working on stg-alt3
> 
> one small note on API compatibility and deprecation of the older Charts endpoint.

for posterity on API compatibility: 
1. we discussed about how removing the deprecated BE endpoints can break previous FE clients - ie, is a breaking change
2. discussed for the flow to be: 
- release the new FE version which uses the new endpoint whilst still supporting the old BE endpoints
- Monitor until the FE clients use the new version
- once all (or reasonable majority) of FE clients no longer use the old BE ie have been updated to new version, we can more safely deprecate the old BE functions

hence, the following steps have been taken: 
- dropped the removal of charts endpoints commit from the [charts PR ](https://github.com/opengovsg/FormSG/pull/9383)
- re-deployed to stg-alt3, will run 1 more round of testing 
- cherry-picked the removal of charts endpoints commit to its own PR [feat: remove unused charts specific submission retrieval endpoint](https://github.com/opengovsg/FormSG/pull/9395) to isolate the removal of deprecated charts PR, to merge and release only once FE clients are verified to no longer use these deprecated endpoints on Datadog. This PR is a major version bump as it introduces the breaking change.

eliotlim (approved):
LGTM! 👍

PR #9248, opened 2026-03-29
fix(duplications): secret key file has old form name

ketanwani:
## Problem

After a storage mode form is duplicated, the dashboard query is invalidated and refetches. When the refetch completes, `dashboardForms` changes (now includes the newly created form), causing the `useEffect` in `DupeFormWizardProvider` to fire again and call `reset()` with a newly-generated title based on the **original** form name. This overwrites the form state title, so the `SaveSecretKeyScreen` reads the wrong title via `useWatch` and names the downloaded secret key file incorrectly.

Fixes #9050

## Solution

Guard the `reset()` call so it only runs while on the `Details` step (i.e. before the form has been created and the wizard has advanced to the save secret key screen).

## Changes

- `DupeFormWizardProvider.tsx` — add `currentStep !== CreateFormFlowStates.Details` guard to the `useEffect`
- `DupeFormWizardProvider.test.tsx` — add tests covering the bug scenario and the fix

## Test plan

- [ ] Duplicate a storage mode form and proceed to the Save Secret Key screen — verify the downloaded file is named after the **duplicate** title, not the original
- [ ] Unit tests pass: `npx vitest run src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.test.tsx`

https://github.com/user-attachments/assets/51a13f29-e719-4d58-859a-ae493cbd119b

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9391, opened 2026-05-06
fix(duplications): secret key file has old form name

### LINKED ISSUE #9050: fix(duplications): secret key file has old form name
When duplicating a form, we expect the duplicated form's secret key to bear the name of the new form, not the old one.

**To Reproduce**
1. Create a form
2. Once created, duplicate the form, specifying a different form name
3. Download the secret key for the duplicated form
4. Inspect the file name for the key

**Expected behavior**
The file name for the key for the duplicated form should contain the name for it

**Screenshots**

Image

LoneRifle:
## Problem
After a storage mode form is duplicated, the dashboard query is invalidated and refetches. 

When the refetch completes, dashboardForms changes  (now includes the newly created form), causing the useEffect in DupeFormWizardProvider to fire again and call `reset()` with a newly-generated title based on the original form name. 

This overwrites the form state title, so the SaveSecretKeyScreen reads the wrong title via useWatch and names the downloaded file incorrectly.

Fixes #9050
Merges #9248

Closes FRM-2030.

## Solution
Fix by guarding the reset() call so it only runs while on the Details step (i.e., before the form has been created and we've moved to the save secret key screen).

**Breaking Changes**

### DIFF
--- a/apps/frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.test.tsx
+++ b/apps/frontend/src/features/workspace/components/DuplicateFormModal/DupeFormWizardProvider.test.tsx
@@ -1,0 +1,155 @@
+import { act, renderHook } from '@testing-library/react'
+
+import { usePreviewForm } from '~features/admin-form/common/queries'
+import { useUser } from '~features/user/queries'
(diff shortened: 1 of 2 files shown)


eliotlim (approved):
LGTM! 👍

PR #9394, opened 2026-05-07
fix(security): 2026-05-07 bump dependencies

eliotlim:
## Problem

Routine dependency bump

Closes FRM-2393.

## Solution

**Breaking Changes** 

## Tests

**TC: Check that FormSG is behaving correctly**
- [ ] Visit form.gov.sg and log in as an admin
- [ ] Submit a Storage Mode Form with Attachments and PDF respondent copy
- [ ] Attachment Upload and Submission should Succeed
- [ ] PDF and Email should arrive normally


package.json
```diff
@@ -78,7 +78,101 @@
       "handlebars": "^4.7.9",
       "jspdf": "^4.2.1",
       "fast-xml-parser@<5": "^4.5.5",
-      "fast-xml-parser@>=5": "^5.5.7"
+      "fast-xml-parser@>=5": "^5.5.7",
```
kevin9foong, line 81: thoughts: wondering if theres a way to reduce the number of overrides we need to do here

apps/frontend/package.json
```diff
@@ -133,18 +133,18 @@
...
-    "@storybook/addon-interactions": "8.3.1",
-    "@storybook/addon-links": "8.3.1",
-    "@storybook/addon-themes": "8.3.1",
-    "@storybook/blocks": "8.3.1",
-    "@storybook/preview-api": "8.3.1",
-    "@storybook/react": "8.3.1",
-    "@storybook/react-vite": "8.3.1",
-    "@storybook/test": "8.3.1",
-    "@storybook/theming": "8.3.1",
+    "@storybook/addon-a11y": "8.6.18",
```
kevin9foong, line 136: joke: hopefully theres no supply chain vuln exposure here? 😆

eliotlim: haha i hope not 🤣

package.json
eliotlim: yes, we should reduce the number of overrides over time!
- these overrides are necessary because the vulnerable / affected versions are not a direct dependency of FormSG.
- [pnpm overrides are a common way to address vulnerable transitive dependencies](https://blog.logto.io/pnpm-upgrade-transitive-dependencies#the-solution) imported by third-party libraries

kevin9foong (approved):

PR #9395, opened 2026-05-07
feat!: remove unused charts specific submission retrieval endpoint

kevin9foong:
Note: do not merge until: 
1. charts MRF is merged into dev (rebase on top of dev then)
2. FE clients are no longer using these routes, monitor on DD. 

## Problem

As discussed in mrf charts PR, these endpoints should be removed after monitoring FE clients have moved to the new version in feat/mrf-charts to prevent breaking old FE clients using (now) legacy charts endpoint. 

The flow: 
1. release the new FE version (ie, charts MRF PR) which uses the new endpoint whilst still supporting the old BE endpoints
2. Monitor until the FE clients use the new version
3. Once all (or reasonable majority) of FE clients no longer use the old BE ie have been updated to new version, we can more safely deprecate the old BE functions (ie release this PR) 

## Solution

Remove deprecated charts functions, ensuring to merge and release after reasonable majority/all FE clients no longer use these removed charts functions in favor of the endpoints introduced in charts MRF PR. 

**Breaking Changes** 

Yes - this PR is not backwards compatible. Specifically, the older FE's charts endpoint will no longer be supported in the BE. 

## Tests 
TC1: Charts can be loaded and works 
- [ ] Open results page with charts. Verify it loads the correct values.


eliotlim (approved):
based on our latest api endpoint usage, we can consider this fully deprecated and merge it!

PR #9398, opened 2026-05-08
fix: signature field draggability interaction

scottheng96:
## Problem

After implementing a fully draggable form field, Signature field experiences a bug where the user is unable to draw on the field and experiences the drag action instead (only in the form builder).

`stopPropagation()` is not working as intended since it is a bubble-phase action, which doesn't block a capture-phase listener (`provided.dragHandleProps` spread onto `HighlightableFlex`) which is higher in the DOM tree and gets fired first and takes precedence.

## Solution

- Remove existing `stopPropagation()` code
- Set canvas element as an editable interactive elements, which then doesn't inherit draggable function (similar to other interactive elements e.g. TextFieldArea etc.)

**Breaking Changes** 

## Before & After Screenshots

| --- | ---- | ----- |
| Signature field | s_wrong | s_correct |

### DIFF
--- a/apps/frontend/src/templates/Field/Signature/SignatureField.tsx
+++ b/apps/frontend/src/templates/Field/Signature/SignatureField.tsx
@@ -272,18 +272,12 @@ const SignatureCanvas = ({
                   </Flex>
                 )}
                 <Box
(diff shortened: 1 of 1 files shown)


eliotlim (approved):
LGTM! 👍

PR #9414, opened 2026-05-13
fix: mrf v3 webhook retries, payload inconsistencies and incorrect submittedSteps values

kevin9foong:
## Problem

There are a several bugs related to MRF webhooks: 

image

1. Webhook retries for Multirespondent (MRF) form submissions are broken. This has 2 causes: 
- after sending the initial webhook, the mrf post submission code tries to update the sent webhook status in the MR submission document in the DB via `EncryptSubmission.addWebhookResponse`. However, it fails to find the submission id and throws an error before adding the webhook retry message into the queue. 

This is fixed in commits: 444f296f8ac19895e404760717d35ffa5160add6^..746479084d477097ef48e16e5a5d2a7d4fdd1072

- Secondly, after the MR webhook retry message is added to the queue. The retry consumer fetches submission info via `EncryptSubmission.retrieveWebhookInfoById(...)` which is defined only on the Encrypt discriminator model. Hence, the webhook info fails to retrieve and the webhook retry fails. 

This is fixed in commits: 26f93c121b6a54628e3be88ff6078991139c1f48^..31a5d6d31e89b5b77ef44d2310e0d340ae7ee1ea

| Initial webhook | Retry webhook | 
| --- | --- | 
| image | image | 
2. Webhook initial payload and retry payload are inconsistent. Specifically, in second step onwards, the `submittedStep` object has different `nextStepEmails` values. The initial webhook payload is `undefined`, while subsequent payloads are `null`.  

This is due to differences in in-memory object and rehydrated object, the saving and rehydration causes the retry webhook payloads to be populated with `null`. This is addressed by specifying the type of `submittedSteps` key, which ensures both are set to a consistent `[]` if empty. 

This is fixed in commit: 038910a247aa2523d4e1df9dd0a0edd1aedd6c62

3. For >=2 steps, despite the step not being an approval step, as long as the form has an approval step, all >=2 steps in `submittedSteps` are marked as `isApproval = true`. This is a bug. 
This is caused by the check for `isApproval = true` relying on the wrong boolean `isFormApproval` instead of `isStepApproval`. 
This is fixed in commit: 235ab19c3cdff88c713e73dead21e0009d5c6e70

Closes FRM-2402

## Solution

This change is expected to apply to v4 webhooks for unified mode as well. 

Define `retrieveWebhookInfoById` onto the base `SubmissionSchema` so the lookup resolves through Mongoose's submission type discriminator, enabling finding the relevant webhook info based on the submission type. 

**Breaking Changes**
No — backwards compatible. Encrypt-mode webhook retries continue to work identically; MRF retries now work too.

## Before & After Screenshots

N/A — backend-only refactor; no user-facing UI change.

## Tests
Note: to accelerate testing and reduce the retry duration, i have pushed a commit to reduce the retry delay to stg-alt3 (see: https://github.com/opengovsg/FormSG/commit/f10ed7e5cdd1400566a608b28bf44177f2d2ec74) This commit is not included as part of this change. You should receive 3 retries at 10s intervals. 

TC1: Webhook retries work for storage mode 
- [ ] Create a webhook endpoint on eg, webhook.site. edit the response value to 500 to emulate failure. 
- [ ] Create storage mode form with webhooks, use the created webhook endpoint. 
- [ ] Submit the form. 
- [ ] Assert that a webhook is sent and received immediately (ie, initial webhook is sent) 
- [ ] Wait 5 mins. Assert that the retry is sent and identical as the first initial webhook. (check the submission id to verify this) 

TC2: Webhook retries work for MRF step 1 
- [x] Create a webhook endpoint on eg, webhook.site. edit the response value to 500 to emulate failure. 
- [x] Create multi respondent mode form with webhooks and 3 steps (2nd should be non-approval, 3rd should be approval), use the created webhook endpoint. 
- [x] Submit the form. 
- [x] Assert that a webhook is sent and received immediately (ie, initial webhook is sent) 
- [ ] Wait 5 mins. Assert that the retry is sent and identical as the first initial webhook. Assert the MRF related keys are included in the retry webhook payload. (check the submission id to verify this) 

TC3: Webhook retries work for MRF step >=2 
- [ ] Continue from TC2. 
- [ ] Submit the form's step 2. 
- [ ] Assert that a webhook is sent and received immediately (ie, initial webhook is sent) 
- [ ] Wait 5 mins. Assert that the retry is sent and identical as the first initial webhook. Assert the MRF related keys are included in the retry webhook payload. (check the submission id to verify this) 
- [x] Assert that the isApproval = false in the submittedSteps key of the webhook payload. 

TC4: Webhook retries work for MRF step >=2 approval step. 
- [ ] Continue from TC3. 
- [ ] Submit the form's step 3. 
- [x] Assert that a webhook is sent and received immediately (ie, initial webhook is sent) 
- [x] Wait 5 mins. Assert that the retry is sent and identical as the first initial webhook. Assert the MRF related keys are included in the retry webhook payload. (check the submission id to verify this) 
- [ ] Assert that the isApproval = false in the submittedSteps key of the webhook payload and the status is reflected correctly.

### DIFF
--- a/apps/backend/src/app/models/__tests__/submission.server.model.spec.ts
+++ b/apps/backend/src/app/models/__tests__/submission.server.model.spec.ts
@@ -66,6 +66,7 @@ describe('Submission Model', () => {
       responseHash: 'This is a hash',
       responseSalt: 'This is a salt',
       hasBounced: false,
+      webhookResponses: [],
     },
     MOCK_SUBMISSION_PARAMS,
   )
@@ -102,6 +103,7 @@ describe('Submission Model', () => {
       encryptedContent: MOCK_ENCRYPTED_CONTENT,
       version: 3,
       workflowStep: 0,
+      webhookResponses: [],
     },
     MOCK_SUBMISSION_PARAMS,
   )
@@ -1045,6 +1047,50 @@ describe('Submission Model', () => {
 
         // Assert
         expect(actualSubmission).toBeNull()
+      })
+
+      it('should push a webhook response onto a multirespondent submission', async () => {
+        // Arrange
+        const formId = new ObjectId()
+        const submission = await MultirespondentSubmission.create({
+          form: formId,
+          submissionType: SubmissionType.Multirespondent,
+          form_fields: [],
+          form_logics: [],
+          workflow: [],
+          submissionPublicKey: 'test public key',
+          encryptedSubmissionSecretKey: 'test secret key',
+          encryptedContent: MOCK_ENCRYPTED_CONTENT,
+          version: 1,
+          workflowStep: 0,
+          submittedSteps: [],
+        })
+
+        const webhookResponse: WebhookResponse = {
+          signature: 'mrf signature',
+          webhookUrl: 'https://form.gov.sg/mrf-endpoint',
+          response: {
+            data: '{"result":"mrf-ok"}',
+            status: 200,
+            headers: '{}',
+          },
+        }
+
+        // Act — call via the base Submission model to confirm the static
+        // resolves through the discriminator for MRF submissions.
+        const actualSubmission = await Submission.addWebhookResponse(
+          String(submission._id),
+          webhookResponse,
+        )
+        const webhookResponses = actualSubmission!.toObject().webhookResponses!
+
+        // Assert
+        expect(webhookResponses).toHaveLength(1)
+        expect(webhookResponses[0].signature).toEqual(webhookResponse.signature)
+        expect(webhookResponses[0].webhookUrl).toEqual(
+          webhookResponse.webhookUrl,
+        )
+        expect(webhookResponses[0].response).toEqual(webhookResponse.response)
       })
     })
   })
--- a/apps/backend/src/app/models/__tests__/submission.server.model.spec.ts
+++ b/apps/backend/src/app/models/__tests__/submission.server.model.spec.ts
@@ -4,6 +4,7 @@ import { promises as dns } from 'dns'
 import {
   BasicField,
   FormAuthType,
+  FormResponseMode,
   PaymentType,
   SubmissionType,
   WebhookResponse,
@@ -12,6 +13,7 @@ import {
 import { merge, omit, times } from 'lodash'
 import mongoose from 'mongoose'
 
+import { getMultirespondentFormModel } from 'src/app/models/form.server.model'
 import getSubmissionModel, {
   getEmailSubmissionModel,
   getEncryptSubmissionModel,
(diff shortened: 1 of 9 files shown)


apps/backend/src/types/submission.ts
```diff
@@ -109,6 +112,29 @@ export interface ISubmissionModel extends Model<ISubmissionSchema> {
...
   saveIfSubmitterIdIsUnique: SaveIfSubmitterIdIsUniqueType
+
+  /**
+   * Retrieves webhook-related info for a given submission. Resolves via
+   * the discriminator, so Encrypt and Multirespondent submissions are
+   * both supported.
+   * @param submissionId
+   * @returns Object containing webhook destination and data
+   */
+  retrieveWebhookInfoById(
```
kevin9foong, line 123: note: i have hoisted this to the base submission and rely on the mongoose submission type discriminator, adheres to DRY, instead of checking submission type and then loading the specific model. 

this is since: 
- makes use of in build discriminator mechanism
- is safe, since email mode is deprecated and no longer in use + email mode never calls the webhook sendWebhook() function.

apps/backend/src/types/submission.ts
```diff
@@ -109,6 +112,29 @@ export interface ISubmissionModel extends Model<ISubmissionSchema> {
...
+  ): Promise<SubmissionWebhookInfo | null>
+
+  /**
+   * Adds a record of a webhook response to a submission. Resolves via
+   * the discriminator, so Encrypt and Multirespondent submissions are
+   * both supported.
+   * @param submissionId ID of submission to update
+   * @param webhookResponse Response data to push
+   */
+  addWebhookResponse(
```
kevin9foong, line 134: note: similar to retrieveWebhookInfoById, i have hoisted this to the base submission and rely on the mongoose submission type discriminator, adheres to DRY, instead of checking submission type and then loading the specific model.

this is since:
- makes use of in build discriminator mechanism
- is safe, since email mode is deprecated and no longer in use + email mode never calls the webhook addWebhookResponse() function.

apps/backend/src/types/submission.ts
kevin9foong: hoisting to the base is a cleaner approach considering we only have storage and mr modes and 1 unified mode in the future

apps/backend/src/app/models/__tests__/submission.server.model.spec.ts
```diff
@@ -502,6 +506,61 @@ describe('Submission Model', () => {
           },
         })
       })
+
+      it('should return the populated submission for a multirespondent submission', async () => {
```
kevin9foong, line 510: note: 
storage mode is already covered by existing test cases, we need to extend to ensure mr mode is also supported.

apps/backend/src/app/models/submission.server.model.ts
```diff
@@ -564,7 +589,7 @@ export const MultirespondentSubmissionSchema = new Schema<
     type: Number,
   },
   submittedSteps: {
-    type: Array,
+    type: [submittedStepSchema],
```
kevin9foong, line 592: why is this change needed? 

the payload for initial webhook & webhook retries differ 
- specifically, in the submittedSteps array, the `nextStepRecipientEmails` is undefined for the initial webhook but null for retries. this is a discrepancy between the 2 payloads, which should be identical.

why is this happening? 

Initial webhook for step N (N ≥ 1) — fired from [multirespondent-submission.service.ts:1362] is using the in-memory submission object that was just mutated and saved in apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.service.ts:1260

this in-memory representation has the value of eg, `nextStepRecipientEmails` set as undefined. 

Retry webhook — Submission.retrieveWebhookInfoById(id) does a fresh findById(...).populate(...) and then calls getWebhookView() on the DB-hydrated document. The DB hydrated document sets `nextStepRecipientEmails` as `null`.

This causes the discrepancy. Specifying the type of `submittedSteps` helps enforce consistency between the webhook initial payload & retry payload.

apps/backend/src/app/models/submission.server.model.ts
```diff
@@ -520,6 +521,30 @@ EncryptSubmissionSchema.statics.findEncryptedSubmissionById = function (
     .exec()
 }
 
+const submittedStepSchema = new Schema(
```
kevin9foong, line 524: for reviewer's reference: this is ther comparison with previous defined type - ive checked that the schema is aligned and typesafe. 

const SubmittedNonApprovalStep = z.object({
  isApproval: z.literal(false),
  submittedAt: z.string().datetime({ precision: 3 }),
  nextStepRecipientEmails: z.array(z.string()).optional(),
  submitterId: z.string().optional(),
})

const SubmittedApprovalStep = SubmittedNonApprovalStep.extend({
  isApproval: z.literal(true),
  status: ApprovalStatus,
})

apps/backend/src/app/models/submission.server.model.ts
```diff
@@ -520,6 +521,30 @@ EncryptSubmissionSchema.statics.findEncryptedSubmissionById = function (
...
+  {
+    isApproval: {
+      type: Boolean,
+      required: true,
+    },
+    submittedAt: {
+      type: String,
+      required: true,
+    },
+    nextStepRecipientEmails: {
```
kevin9foong, line 534: note that instead of returning `undefined` in the initial webhook payload previously (mrf retry payload not a concern since it didnt work), it will return an empty array `[]` now. this is consistent with mrf retries and retry & initial have identical payload. 

this payload will persist for v4 webhooks as well.

scottheng96:
Nice - thanks for the helpful explanations. I tested it successfully on stg-alt3, can ignore the checkboxes; updates are janky to them

scottheng96 (approved):

apps/backend/src/app/models/submission.server.model.ts
eliotlim: praise: thanks for defining this clearly

apps/backend/src/types/submission.ts
eliotlim: comment: this note was helpful, I appreciated the context.

PR #9432, opened 2026-05-17
fix: use POSIX compliant syntax in husky hook scripts for shell interoperability

kevin9foong:
## Problem

Our current pre-commit hooks only work on bash which is the current login shell on MacOS. 
However, husky uses sh which requires POSIX compliant syntax on linux machines. 
This causes issues running the hooks on other machines such as isolated sandboxes eg, codespaces.  

## Solution

Specify `sh` usage with shebang. Update to use POSIX compliant syntax in husky hook scripts whilst maintaining the same script behavior. 

**Breaking Changes** 

## Tests

TC1: 
- [x] Can still commit on local macOS with bash 

TC2: 
- [x] Can also commit on linux machines eg, Github codespaces

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,5 +1,7 @@
+#!/usr/bin/env sh
+
 # Section for git-secrets
-if ! command -v git-secrets &> /dev/null
+if ! command -v git-secrets > /dev/null 2>&1
 then
     echo "git-secrets is not installed. Please run 'brew install git-secrets' or visit https://github.com/awslabs/git-secrets#installing-git-secrets"
     exit 1
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,7 @@
+#!/usr/bin/env sh
+
 # Section for git-secrets
-if ! command -v git-secrets &> /dev/null
+if ! command -v git-secrets > /dev/null 2>&1
 then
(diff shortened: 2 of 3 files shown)


scottheng96:
While running TC2, I realise husky doesn't run on codespaces. I believe it because 
- husky's prepare script checks for `CI=true`, and skips hooks installation if so
- Github Codespace sets `CI=true` in the env (during the postCreate) phase

image


Think to updating the "prepare" to manually install the hooks should work? I guess the larger question is if we are making any changes on Codespaces as well which makes this necessary haha

kevin9foong:
> While running TC2, I realise husky doesn't run on codespaces. I believe it because
> 
> * husky's prepare script checks for `CI=true`, and skips hooks installation if so
> * Github Codespace sets `CI=true` in the env (during the postCreate) phase
> 
> image
> Think to updating the "prepare" to manually install the hooks should work? I guess the larger question is if we are making any changes on Codespaces as well which makes this necessary haha

let me check on this in a bit - the precommit hooks were running and causes issues for me hence this PR.

kevin9foong:
hey @scottheng96 , for husky to work, since its a `devDependency`, we need to do a `pnpm install` first - i think thats the reason why it was not running. 

I've run `pnpm install` to install husky and verified that it works on codespaces and avoids the syntax incompatibility issues.

scottheng96 (approved):

PR #9441, opened 2026-05-18
fix: signature hover state to match form theme color

### LINKED ISSUE #9347: Signature field's hover state does not match form theme colour
https://github.com/user-attachments/assets/fbaa7d8c-faec-4dfa-a9a1-e774d161134f

Proposed changes:

- Hover state should not be hardcoded and follow the form's theme color just like the other fields.

rickysurya:
## Problem
The signature field's hover uses a hardcoded blue (#445fcd) instead of following the form's theme color like other fields

Closes #9347

## Solution
I added colorTheme to SignatureCanvas so it uses the form's theme color. Swapped the hardcoded color for semantic tokens matching what other fields use

## Tests
Verified in Storybook that the field renders with theme colors

### DIFF
--- a/apps/frontend/src/templates/Field/Signature/SignatureField.tsx
+++ b/apps/frontend/src/templates/Field/Signature/SignatureField.tsx
@@ -5,10 +5,10 @@ import {
   useFormContext,
   useFormState,
 } from 'react-hook-form'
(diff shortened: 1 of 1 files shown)


LoneRifle (approved):
lgtm

PR #9445, opened 2026-05-19
fix: increase myinfo circuit breaker timeout to 30000ms

scottheng96:
## Problem

Myinfo service is timing out with 5000ms

## Solution

Increase timeout to 30000ms (30s).

**Breaking Changes** 


eliotlim (approved):
LGTM 👍 

We're seeing p99 latency spikes of up to 30s on Myinfo API endpoints, so this should address the ongoing timeout issues whilst upstream sorts it out.

PR #9449, opened 2026-05-20
fix: update augmented responses v4 to include unaswered questions & headers

scottheng96:
## Problem

Found bugs with v4 decryption onto response pages:
1. Numberings don't align with response-only questions; non-response fields increment the question number, but aren't shown on the page, leading to mismatch between form and response question numbers
2. unanswered fields don't get displayed on the response page
3. Header's aren't shown in individual response page (regression from v3)

## Solution

When augmenting v4 responses,
1.  stop question number incrementing based on formField, not response (since responses will never contain non-response fields)
2. Determine if formField exists but not yet responded to - if so create a synthetic responseV4 with empty value
3. Include headers in augmented response v4

**Breaking Changes** 

## Before & After Screenshots
| --- | ----- | ----- |
| Headers, unaswered questions 
image | image | image |

## Tests

TC1: Verify unanswered and headers behaviour on IndividualResponsePage
- [ ] Create an MRF form with 2 basic text fields (make 1 optional), 1 header field, 1 section/image field
- [ ] Make a submissions of the form, leave optional field blank
- [ ] Visit the IndividualResponsePage and observe that
   - [ ] Responses are numbered matching the questions on the form
   - [ ] Unanswered text field shows a question without an answer
   - [ ] Headers are included, but not numbered

TC2: Verify download csv responses (regression)
-  [ ] Download responses into CSV with PDF attachment
-  [ ] Verify that CSV columns correspond with all response-related fields, and optional field is not answered
-  [ ] Verify that PDF shows optional field unanswered & presence of headers

### DIFF
--- a/apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
+++ b/apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
@@ -209,10 +209,23 @@ export const DecryptedRowV4 = memo(
     questionNumber,
     downloadUrl,
     attachmentDecryptionKey,
+    unanswered,
   }: DecryptedRowV4Props): JSX.Element => {
-    switch (field.fieldType) {
-      case BasicField.Section:
+    if (unanswered) {
+      if (field.fieldType === BasicField.Section) {
         return <DecryptedHeaderRowV4 field={field} />
+      }
+      return (
+        <Stack>
+          <DecryptedQuestionLabelV4
+            field={field}
+            questionNumber={questionNumber}
+          />
+        </Stack>
+      )
+    }
+
+    switch (field.fieldType) {
       case BasicField.Attachment:
         return (
           <DecryptedAttachmentRowV4
--- a/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
+++ b/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
@@ -105,17 +105,18 @@ export const augmentDecryptedResponsesV4 = (
   const results: AugmentedDecryptedResponseV4[] = []
 
   orderedFieldIds.forEach((fieldId, index) => {
-    const field = responsesV4[fieldId]
+    const fieldFromResponse = responsesV4[fieldId]
     const formField = formFieldMap.get(fieldId)
 
-    const fieldType = field?.fieldType ?? (formField?.fieldType as BasicField)
-    if (!fieldType || !formField || !isBasicField(fieldType)) return
+    const fieldType =
+      fieldFromResponse?.fieldType ?? (formField?.fieldType as BasicField)
+    if (!isBasicField(fieldType)) return
 
     const isNonResponse = NON_RESPONSE_FIELD_SET.has(fieldType)
     if (isNonResponse) {
       nonResponseFieldsCount++
       // Only Section headers are rendered; other non-response fields (e.g. Statement) are skipped
-      if (fieldType === BasicField.Section) {
+      if (fieldType === BasicField.Section && formField) {
         results.push({
           fieldId,
           field: buildSyntheticField(formField),
@@ -127,10 +128,12 @@ export const augmentDecryptedResponsesV4 = (
 
     results.push({
       fieldId,
-      field: field ?? buildSyntheticField(formField),
+      field:
+        fieldFromResponse ??
+        (formField ? buildSyntheticField(formField) : undefined),
       questionNumber: index + 1 - nonResponseFieldsCount,
       downloadUrl: attachmentMetadata[fieldId],
-      unanswered: !field,
+      unanswered: !fieldFromResponse,
     })
   })
 
--- a/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
+++ b/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
@@ -1,4 +1,8 @@
-import { FieldResponsesV4, FieldResponseV4 } from '@opengovsg/formsg-sdk'
+import {
+  AnswerV4,
+  FieldResponsesV4,
+  FieldResponseV4,
+} from '@opengovsg/formsg-sdk'
 import type {
   EncryptedAttachmentRecords,
   FormField,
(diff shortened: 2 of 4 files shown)


apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -80,24 +106,31 @@ export const augmentDecryptedResponsesV4 = (
 
   orderedFieldIds.forEach((fieldId, index) => {
     const field = responsesV4[fieldId]
-    if (!field) return
+    const formField = formFieldMap.get(fieldId)
 
-    if (!isBasicField(field.fieldType)) return
+    const fieldType = field?.fieldType ?? (formField?.fieldType as BasicField)
+    if (!fieldType || !formField || !isBasicField(fieldType)) return
```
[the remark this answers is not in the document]
scottheng96: Yep. Good catch Copilot. Removed !formFIeld as per comment

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -80,24 +106,31 @@ export const augmentDecryptedResponsesV4 = (
...
+      return
     }
 
     results.push({
       fieldId,
-      field,
-      questionNumber: isNonResponse
-        ? undefined
-        : index + 1 - nonResponseFieldsCount,
+      field: field ?? buildSyntheticField(formField),
```
kevin9foong: reviewer verification: this fixes unanswered fields don't get displayed on the response page ✅ 

we build a synthetic field when the `fieldFromResponse` does not exist and push this

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -58,17 +62,39 @@ export type AugmentedDecryptedResponseV4 = {
...
+ * @param formFields Ordered snapshot of form field definitions for the submission
+ * @param responsesV4 Decrypted responses in V4 format, keyed by field ID
+ * @param attachmentMetadata Map of field ID to attachment download URL
+ * @returns Ordered list of augmented responses ready for display
+ */
 export const augmentDecryptedResponsesV4 = (
   formFields: FormFieldDto[],
   responsesV4: FieldResponsesV4,
   attachmentMetadata: EncryptedAttachmentRecords,
 ): AugmentedDecryptedResponseV4[] => {
```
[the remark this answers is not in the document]
scottheng96: Added tests

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -80,24 +106,31 @@ export const augmentDecryptedResponsesV4 = (
 
   orderedFieldIds.forEach((fieldId, index) => {
     const field = responsesV4[fieldId]
```
kevin9foong: suggestion (understandability): instead of calling this `field`, should we call this eg, `fieldFromResponse` instead? 

rationale: this would help with understandability below
```
results.push({
      fieldId,
      field: fieldFromResponse ?? buildSyntheticFieldResponse(formField),
      unanswered: !fieldFromResponse
```

my main thought as a code reader is that i have to reason that the formFieldMap `formField` and `field` is different. which can be confusing

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -58,17 +62,39 @@ export type AugmentedDecryptedResponseV4 = {
...
+/**
+ * Builds a synthetic FieldResponseV4 from a form field definition.
+ * Used for non-response fields (e.g. Section headers) and unanswered optional fields,
+ * so they can be rendered with question text but no answer body.
+ * Safe to use only when paired with unanswered: true, which bypasses answer rendering.
+ */
+const buildSyntheticField = (formField: FormFieldDto): FieldResponseV4 => ({
+  fieldType: formField.fieldType as FieldResponseV4['fieldType'],
+  question: formField.title,
+  answer: {} as AnswerV4,
```
kevin9foong, line 77: suggestion: instead of providing an empty object - could we use `null` instead? it would be more idiomatic since `null` means no answer which `{}` might be used to signal.

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
```diff
@@ -209,10 +209,23 @@ export const DecryptedRowV4 = memo(
     questionNumber,
     downloadUrl,
     attachmentDecryptionKey,
+    unanswered,
   }: DecryptedRowV4Props): JSX.Element => {
-    switch (field.fieldType) {
-      case BasicField.Section:
+    if (unanswered) {
```
kevin9foong, line 214: thought: why do we need to add a derived unanswered param?

could we directly check for fieldType === BasicField.Section instead?
instead of using unanswered, suppose we are using null = no answer, could we check field.response === null instead?

rationale: each param we have is a form of derived state which is potential for more bugs
for example, the addition of `unanswered` introduces this new constraint in the comments below `* Safe to use only when paired with unanswered: true, which bypasses answer rendering.`

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -80,24 +106,31 @@ export const augmentDecryptedResponsesV4 = (
 
   orderedFieldIds.forEach((fieldId, index) => {
```
kevin9foong, line 107: verifying by reviewer: this ensures the ordering below matches v3 ✅

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
```diff
@@ -58,17 +62,39 @@ export type AugmentedDecryptedResponseV4 = {
...
+  unanswered?: boolean
 }
 
+/**
+ * Builds a synthetic FieldResponseV4 from a form field definition.
+ * Used for non-response fields (e.g. Section headers) and unanswered optional fields,
+ * so they can be rendered with question text but no answer body.
+ * Safe to use only when paired with unanswered: true, which bypasses answer rendering.
+ */
+const buildSyntheticField = (formField: FormFieldDto): FieldResponseV4 => ({
```
kevin9foong, line 74: suggestion: instead of the term synthetic, could we name it eg, `buildNullResponseField` or something along these lines which might represent the intention of this. 

this might help with code understandability and avoid new terms

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
scottheng96: Agree - renamed to `fieldFromResponse`

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
scottheng96: Updated as per comment above!

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
scottheng96: Similar comment as above, it might not make sense that a FieldResponseV4.answer can be `null`, since its encrypted and decrypted as-is, its probably better that it should be strictly non-null. This then leads the implementation to
- build unanswered fields as synthetic fields with a non-null `.answer`
- include a variable if it is `unanswered`
- check against this derived variable, so we don't have to worry about handling those synthetic fields non-null `.answers` at the specific fieldRow level

Why v3 doesn't have this problem is because of the loosely typed FormField in the sdk (`any`), which despite allowing this flexibility, sacrifices type-security. I also thought about setting up a v4 decrypted type to handle `null`, but honestly I'm leaning towards not handling so many types (it's one of the pitfalls of our V1/V3 types)

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/augmentDecryptedResponses.ts
scottheng96: Hmm, I'm hesitant to utilise `null`, since that would need FieldResponseV4 type to also handle it `| null/undefined`, and I think it might make sense to keep it as strictly non-null answers. I'm open to other ideas though, I think this isn't as clean as we might like it to be.

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/DecryptedRowV4.tsx
scottheng96: Yknow on second thought, I see a lot of usages of AugmentedDecryptedResponse everywhere. Maybe it would make our lives easier to figure out a way to type v4responses into AugmentedDecryptedResponse instead of handling a whole new separate flow

PR #9458, opened 2026-05-21
fix: refactors for answer-object-decryption

scottheng96:
## Problem

Found bugs with v4 decryption onto response pages:

1. Question numbering mismatch in Individual response page (non-response fields are incrementing the question number)
2. Unanswered fields not displayed on the response page & csv
3. Headers not shown in individual response page

## Solution

Refactored the decryption handling flow to remove v4 pipeline and instead utilise existing v3 pipeline since the increased expressibility of v4 responses are not valuable to decryption handling (yet), so prioritising code cleanliness.

1. Introduced `processedDecryptedContentV4`, which outputs compatible type to be handled in the same augmentation flow as V3 (and V1).
2. Included unanswered fields & headers at `processDecryptedContentV4` layer to leverage existing pipeline's handling of these fields
3. Removed usage of `augmentedDecryptedResponseV4`, 

*Additionally, I've refactored v4 decryption to run as a shadow flow to v3, so if the feature-flag is turned on but v4 decryption fails, v3 decryption fallback is still used. Minimises risk to users while maintaining observability

**Breaking Changes** 

## Before & After Screenshots
| --- | ----- | ----- |
| Headers, unaswered questions 
image | image | image |

## Tests

TC1: Verify unanswered and headers behaviour on IndividualResponsePage
- [ ] Create an MRF form with 2 basic text fields (make 1 optional), 1 header field, 1 section/image field
- [ ] Make a submissions of the form, leave optional field blank
- [ ] Visit the IndividualResponsePage and observe that
   - [ ] Responses are numbered matching the questions on the form
   - [ ] Unanswered text field shows a question without an answer
   - [ ] Headers are included, but not numbered

TC2: Verify download csv responses (regression)
-  [ ] Download responses into CSV with PDF attachment
-  [ ] Verify that CSV columns correspond with all response-related fields, and optional field is not answered
-  [ ] Verify that PDF shows optional field unanswered & presence of headers


apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -167,29 +160,23 @@ export const getDecryptedSubmissionById = async ({
...
-              ...convertVerifiedToV4(decryptedV4.verified),
-            }
-          : decryptedV4.responses
-        responsesV4 = augmentDecryptedResponsesV4(
-          encryptedSubmission.form_fields,
-          processedContentV4,
-          encryptedSubmission.attachmentMetadata,
-        )
       }
       submissionSecretKey = decryptedContent.submissionSecretKey
```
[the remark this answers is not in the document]
scottheng96: moved the v3 assignment before the v4 attempt!

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -146,7 +145,29 @@ async function decryptSubmissionData(
...
+      decryptedResponses = await processDecryptedContentV3(
+        submissionData.form_fields,
+        decryptedObject,
+      )
+
       if (useV4) {
+        console.log('Using v4 decryption for MRF response', {
+          submissionId: submissionData._id,
+        })
         const formFieldsMeta = buildFormFieldMetaMap(submissionData.form_fields)
```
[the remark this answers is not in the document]
scottheng96: Removed

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -146,7 +145,29 @@ async function decryptSubmissionData(
...
+        return {
+          isSubmissionDecryptionSuccessful: false,
+        }
+      }
+      mrfSubmissionSecretKey = decryptedObject.submissionSecretKey
+      decryptedResponses = await processDecryptedContentV3(
+        submissionData.form_fields,
+        decryptedObject,
+      )
+
```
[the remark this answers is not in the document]
scottheng96: Good point, updated as per comment

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
```diff
@@ -47,31 +47,29 @@ export const flattenV4ToFormFields = ({
...
+    const fieldResponse = v4Responses[ff._id]
+    if (!fieldResponse) {
+      v1Fields.push({
+        _id: ff._id,
+        question: ff.title,
+        fieldType: ff.fieldType,
+        answer: '',
+        ...(ff.fieldType === BasicField.Section && { isHeader: true }),
+      })
+      continue
```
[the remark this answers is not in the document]
scottheng96: Resolved by using common transformInputToOutput function

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -159,46 +158,40 @@ async function decryptSubmissionData(
...
+      // Perform v3 decryption if v4 deryption fails or is not attempted (useV4 = false)
+      const decryptedObject = ctx.formsgSdk.cryptoV3.decrypt(secretKey, {
+        encryptedSubmissionSecretKey:
+          submissionData.encryptedSubmissionSecretKey,
+        encryptedContent,
+        verifiedContent,
+        version,
+      })
+      if (!decryptedObject) {
+        console.error('Invalid decryption for multirespondent response')
```
eliotlim: question: should we use datadogLogs for this to make sure we can capture the error and improve visibility.
```suggestion
        datadogLogs.logger.error('Invalid decryption for multirespondent response')
```

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
```diff
@@ -47,31 +49,29 @@ export const flattenV4ToFormFields = ({
...
+      // Reuse v3 transform to get the correct empty output shape per field type.
+      // This returns null for Statement/Image (excluded, matching v3 behavior)
+      // and the correct answerArray shape for compound fields like Address, Table, etc.
+      const emptyOutput = transformInputsToOutputs(ff)
+      if (emptyOutput) {
+        v1Fields.push(emptyOutput as unknown as FormField)
+      }
+      continue
+    }
+    const { fieldType, question } = fieldResponse
```
eliotlim: nit: if the semantics of this didn't change, perhaps we should have kept this variable name as `field` to eliminate the unnecessary diffs below

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
```diff
@@ -47,31 +49,29 @@ export const flattenV4ToFormFields = ({
...
+    const fieldId = ff._id
 
     // Generic string fields (including yes_no)
     if (GENERIC_STRING_FIELD_TYPES.has(fieldType)) {
-      const answer = field.answer as StringAnswerV4
+      const answer = fieldResponse.answer as StringAnswerV4
```
eliotlim: nit: ditto ^

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -159,46 +158,40 @@ async function decryptSubmissionData(
...
+          mrfSubmissionSecretKey = decryptedV4.submissionSecretKey
+          decryptedResponses = processDecryptedContentV4(
+            submissionData.form_fields,
+            decryptedV4.responses,
+            decryptedV4.verified,
           )
-          return {
-            isSubmissionDecryptionSuccessful: false,
-          }
+          break
```
eliotlim, line 169: comment: the success case returns early (not a guard clause) -- which creates a pitfall.
- the break jumps us out of the MRF decryption flow on success if useV4 and decryptedV4.
- we could rewrite this in a different way that is simpler and easier to read
- and that implements a guard clause structure: fail early, such that the break handles the error condition branch of the successful flow

```suggestion
        // equivalent to !(useV4 && decryptedV4) by deMorgan's law
        if (!useV4 || !decryptedV4) {
          // Perform v3 decryption if v4 deryption fails or is not attempted (useV4 = false)
          const decryptedObject = ctx.formsgSdk.cryptoV3.decrypt(secretKey, {
            ...
          })
          break
```

eliotlim (commented):
just a few comments on how we might improve code readability!

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
scottheng96: Thanks for the suggestion!

Did some exploring into this, but obstacle would be to handle `responses` that may be left undefined if we set it in conditionally in both if statements. Current implementation is still the cleanest way!

eliotlim (approved):

PR #9459, opened 2026-05-21
fix(backend): update sso integration behavior

eliotlim:
## Problem

When the SSO discovery URL is not available at container start, the FormSG application may error out.

Previously related to #9029.
Closes FRM-2259.

## Solution

**Breaking Changes** 

**Improvements**:

- Retry discovery so that SSO can recover without a container restart

**Bug Fixes**:

- Handle the error when SSO_DISCOVERY_URL cannot be loaded

### DIFF
--- a/apps/backend/src/app/config/features/sso.config.ts
+++ b/apps/backend/src/app/config/features/sso.config.ts
@@ -11,8 +11,7 @@ convict.addFormat(url)
 
 export const optionalValuesFromSsm: Path<ISsoVarsSchema>[] = [] //['hostname']
 
-// Placeholder defaults used when SSO env vars are not provided. Referenced by
-// both the schema below and isSsoConfigured() so the two cannot drift apart.
+// RATIONALE: shared between the schema and isSsoConfigured() so they can't drift.
 const SSO_PLACEHOLDER_CLIENT_ID = 'client-id'
 const SSO_PLACEHOLDER_CLIENT_SECRET = 'test'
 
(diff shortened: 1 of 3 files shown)


apps/backend/src/app/config/features/sso.config.ts
```diff
@@ -19,20 +27,71 @@ export const ssoVarsSchema: Schema<ISsoVarsSchema> = {
...
+        isDefaultClientId: clientId === SSO_PLACEHOLDER_CLIENT_ID,
+        isDefaultClientSecret: clientSecret === SSO_PLACEHOLDER_CLIENT_SECRET,
+      },
+    })
+    return false
+  }
+
+  // Validate that discoveryUrl is a valid URL
+  try {
+    new URL(discoveryUrl)
```
kevin9foong, line 73: TODO: double check if `new URL(undefined)` is safe

apps/backend/src/app/modules/auth/sso/auth-sso.service.ts
```diff
@@ -34,26 +64,54 @@ export class AuthSsoServiceClass {
...
+        meta: {
+          action: 'AuthSsoServiceClass.initializeClientConfig',
+          error,
+        },
+        message:
+          'Error while parsing SSO discovery URL. SSO login is unavailable.',
+        error,
       })
+      this.discoveryFailedAt = Date.now()
+      // Create a rejected promise
```
kevin9foong: thought: i wonder if some of these LLM comments are too verbose. should we prompt it to remove these redundant comments?

apps/backend/src/app/modules/auth/sso/auth-sso.service.ts
```diff
@@ -63,13 +121,50 @@ export class AuthSsoServiceClass {
...
+      )
+    }
+
+    // Initialize the client config if not already done
+    this.initializeClientConfig()
+
+    // TypeScript doesn't know that initializeClientConfig always sets clientConfigPromise,
+    // so we need to check for it explicitly
+    const configPromise = this.clientConfigPromise
+    if (!configPromise) {
```
kevin9foong, line 136: note: this comment might be redundant. im thinking we need to add to our context some rules for LLM comments. it tends to be very verbose and over time our codebase will be filled with comments which we need to maintain.

apps/backend/src/app/modules/auth/sso/auth-sso.service.ts
kevin9foong: suggested prompt: 

```
limit your changes to this PR. 
reduce verbosity of the comments introduced. In general, comments should highlight the rationale behind changes eg, RATIONALE: ...` 
for the what and how, focus on naming of variables and constants to make it as clear as possible 
```

apps/backend/src/app/config/features/sso.config.ts
```diff
@@ -2,12 +2,20 @@ import convict, { Path, Schema } from 'convict'
...
 
+const logger = createLoggerWithLabel(module)
+
 convict.addFormat(url)
 
 export const optionalValuesFromSsm: Path<ISsoVarsSchema>[] = [] //['hostname']
 
+// Placeholder defaults used when SSO env vars are not provided. Referenced by
+// both the schema below and isSsoConfigured() so the two cannot drift apart.
+const SSO_PLACEHOLDER_CLIENT_ID = 'client-id'
```
kevin9foong, line 15: praise: good practice to use constants

kevin9foong (approved):
mainly LGTM, just a comment about making the comments in this PR less verbose for your consideration

Issue #9454, opened 2026-05-20
DuplicateFormModal: default to MRF + escape hatch + storage-mode page


kevin9foong:
## What to build

Change the **Duplicate Form** modal so that the new (duplicated) form defaults to **Multirespondent Form (MRF)** regardless of the source form's response mode. The mode-selection screen no longer offers a storage-mode option directly; an "old version of FormSG" escape-hatch link is always visible with copy composed from the user's beta flags (via #9451).

Clicking the link navigates the modal to a storage-mode page (title input + create button). Submitting calls the existing `dupeStorageModeForm` mutation. A back affordance returns to the MRF default screen.

Visual layout and copy come from the Figma file: <https://www.figma.com/design/LCHsRuUJ0HkTCMjIcxFtQ4/Form-Design-Master-Part-2?node-id=3424-764>.

Backend is unchanged — gating is frontend-only per `docs/adr/0001-frontend-only-storage-form-gating.md`.

Verify end-to-end through the **TDD loop**: Storybook stories (none currently exist for `DuplicateFormModal` — add them), agent-browser interaction, Chromatic visual gate, local dev verification.

See repo-root `CONTEXT.md` for terminology.

## Acceptance criteria

- [ ] `DuplicateFormModal` defaults `responseMode` to `Multirespondent` regardless of the source form's mode
- [ ] The mode-selection screen no longer offers a storage-mode option directly
- [ ] An "old version of FormSG" escape-hatch link is always visible
- [ ] The link copy is produced by the composer from #9451
- [ ] Clicking the link navigates to a storage-mode page (title input + create button)
- [ ] Submitting the storage-mode page duplicates the source as a storage-mode form via the existing mutation
- [ ] A back affordance returns the user to the MRF default screen
- [ ] New Storybook stories cover: MRF default, escape-hatch open, beta-flag copy variants
- [ ] Chromatic visual baseline updated
- [ ] No backend changes

## Blocked by

- #9456 (mrf-cutover feature flag)
- #9451 (escape-hatch copy composer)

Issue #9451, opened 2026-05-20
Build escape-hatch copy composer (pure fn + unit tests)


kevin9foong:
## What to build

A pure function that takes a user's `betaFlags` and returns the composed escape-hatch link copy for the form creation modals. The copy is composed from the applicable reasons an admin might still legitimately need a storage-mode form:

- Payments → always applicable (implicit for all admins)
- Children fields → applicable if `betaFlags.children` is true
- Webhooks v1 → applicable if `betaFlags.createStorageModeForV1Webhook` is true

When multiple flags are set, the reasons compose into a single coherent line (one link, not multiple). When the user has no extra flags, the link still appears for the payments reason.

This is a deep, isolated, framework-free module. It is the only piece of conditional logic in the broader feature and the highest-value unit-test target.

Final copy strings come from the Figma file: <https://www.figma.com/design/LCHsRuUJ0HkTCMjIcxFtQ4/Form-Design-Master-Part-2?node-id=3424-764>. Pull the exact text from the relevant frames.

See repo-root `CONTEXT.md` for terminology.

## Acceptance criteria

- [ ] Pure function exported from a shared frontend location, importable by all three modal flows
- [ ] Returns the composed copy line for every combination of `children` and `createStorageModeForV1Webhook` (with payments implicit)
- [ ] No React, no hooks, no side effects
- [ ] Unit-test coverage for all flag combinations: none, children only, v1-webhook only, both
- [ ] Copy strings match Figma source of truth

## Blocked by

None - can start immediately

PR #9465, opened 2026-05-21
feat(mrf-cutover): default CreateFormModal to MRF with storage-mode escape hatch (4/6)

### LINKED ISSUE #9453: CreateFormModal: default to MRF + escape hatch + storage-mode page
## What to build

Change the **Create Form** modal so that newly created forms default to **Multirespondent Form (MRF)** instead of storage mode. The mode-selection screen no longer offers a storage-mode option directly; instead, an "old version of FormSG" escape-hatch link is always visible, with copy composed from the user's beta flags (via the composer from #9451).

Clicking the link navigates the modal to a new **storage-mode page** containing a title input and a create button. Submitting the title calls the existing storage-mode form creation mutation. A back affordance returns to the MRF default screen.

This applies to all entry points into `CreateFormModal`, including the empty-workspace "Create your first form" CTA.

Visual layout and copy come from the Figma file (look for the "storage mode page" frame): <https://www.figma.com/design/LCHsRuUJ0HkTCMjIcxFtQ4/Form-Design-Master-Part-2?node-id=3424-764>.

Backend is unchanged — gating is frontend-only per `docs/adr/0001-frontend-only-storage-form-gating.md`.

Verify end-to-end through the **TDD loop**: Storybook stories for each state (MRF default, escape-hatch open, each beta-flag copy variant), agent-browser interaction, Chromatic visual gate, and local dev verification.

See repo-root `CONTEXT.md` for terminology.

## Acceptance criteria

- [ ] `CreateFormModal` defaults `responseMode` to `Multirespondent`
- [ ] The mode-selection screen no longer offers a storage-mode option directly
- [ ] An "old version of FormSG" escape-hatch link is always visible
- [ ] The link copy is produced by the composer from #9451 and reflects the user's beta flags
- [ ] Clicking the link navigates to a new storage-mode page (title input + create button)
- [ ] Submitting the storage-mode page creates a storage-mode form via the existing mutation
- [ ] A back affordance returns the user to the MRF default screen without closing the modal
- [ ] Empty-workspace CTA opens the same modal with the same behaviour
- [ ] Storybook stories cover: MRF default, escape-hatch open / storage-mode page, each beta-flag copy variant
- [ ] Chromatic visual baseline updated
- [ ] No backend changes

## Blocked by

- #9456 (mrf-cutover feature flag)
- #9450 (beta flag plumbing)
- #9451 (escape-hatch copy composer)

kevin9foong:
> Stacked PR 4/6 for MRF cutover. Base: `refactor/modal-cutover-prep`.

## Problem

Per the PRD (issue #9454): the Create Form modal currently defaults to storage mode and offers it side-by-side with MRF, which slows org-wide MRF migration and leaves admins on the legacy v1 webhook schema by default. Closes #9453.

## Solution

Under the `mrf-cutover` flag:

- Default `CreateFormModal` to **Multirespondent** — no mode toggle on the first screen.
- New `CreateFormStorageModeScreen` reachable via the escape-hatch link ("old version of FormSG"), composed from the `escapeHatchCopy` foundation in PR 1. Title input + create button, with a "back" affordance to return to the MRF default screen.
- Match Figma fidelity: hide the data-classification infobox in the cutover flow, use 2.5rem spacing.
- Fix sliding-animation bug + `undefined` email crash that surfaced during escape-hatch navigation.
- Storybook stories for the three escape-hatch copy variants (payments-only, +children, +webhook v1).

When the flag is off, the original modal renders unchanged.

**Breaking Changes**

No - gated behind `mrf-cutover` feature flag. Backend `createForm` still accepts `Encrypt` without flag checks (see `docs/adr/0001-frontend-only-storage-form-gating.md`).

## Tests

**TC1: flag on, default MRF flow**

- [ ] Enable `mrf-cutover` flag; open Create Form
- [ ] Confirm MRF default screen renders (no mode toggle, no data-classification infobox)
- [ ] Submit a title → confirm an MRF is created

**TC2: escape hatch → storage mode**

- [ ] On the MRF default screen, click "old version of FormSG"
- [ ] Confirm `CreateFormStorageModeScreen` renders with title input + create button
- [ ] Submit → confirm a storage-mode form is created
- [ ] From the storage screen, click "back" → confirm return to MRF screen without state loss

**TC3: escape-hatch copy by beta flag**

- [ ] User with no extra flags → copy mentions payments only
- [ ] User with `betaFlags.children` → copy mentions payments + children fields
- [ ] User with `betaFlags.createStorageModeForV1Webhook` → copy mentions payments + webhooks v1

**TC4: flag off**

- [ ] Disable `mrf-cutover` flag; open Create Form
- [ ] Confirm original modal renders unchanged

### DIFF
--- a/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModal.stories.tsx
+++ b/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModal.stories.tsx
@@ -133,6 +133,29 @@ MrfCutoverOnWebhookV1Beta.parameters = {
   ],
 }
 
+export const MrfCutoverOnAllExceptions = Template.bind({})
+MrfCutoverOnAllExceptions.decorators = [
+  (Story) => (
+    <GrowthBookProvider growthbook={mrfCutoverOn}>
+      <Story />
+    </GrowthBookProvider>
+  ),
+]
+MrfCutoverOnAllExceptions.parameters = {
+  msw: [
+    getUser({
+      delay: 0,
+      mockUser: {
+        ...MOCK_USER,
+        betaFlags: {
+          children: true,
+          createStorageModeForV1Webhook: true,
+        },
+      },
+    }),
+  ],
+}
+
 export const StorageModeAckScreen = () => {
   const secretKey = 'mock-secret-key'
 
--- a/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModal.stories.tsx
+++ b/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModal.stories.tsx
@@ -9,13 +9,15 @@ import {
   useClipboard,
   useDisclosure,
 } from '@chakra-ui/react'
+import { GrowthBook, GrowthBookProvider } from '@growthbook/growthbook-react'
 import { Meta, StoryFn } from '@storybook/react'
 
+import { featureFlags } from 'formsg-shared/constants'
 import { UserId } from 'formsg-shared/types'
 import { PublicFormViewDto } from 'formsg-shared/types/form'
 import { Workspace, WorkspaceId } from 'formsg-shared/types/workspace'
 
-import { userHandlers } from '~/mocks/msw/handlers/user'
+import { getUser, MOCK_USER, userHandlers } from '~/mocks/msw/handlers/user'
 
 import { ApiError } from '~typings/core'
 
@@ -80,6 +82,56 @@ const Template: StoryFn<CreateFormModalProps> = (args) => {
   )
 }
 export const Default = Template.bind({})
+
+const mrfCutoverOn = new GrowthBook({
+  features: { [featureFlags.mrfCutover]: { defaultValue: true } },
+})
+
+export const MrfCutoverOn = Template.bind({})
+MrfCutoverOn.decorators = [
+  (Story) => (
+    <GrowthBookProvider growthbook={mrfCutoverOn}>
+      <Story />
+    </GrowthBookProvider>
+  ),
+]
+
+export const MrfCutoverOnChildrenBeta = Template.bind({})
+MrfCutoverOnChildrenBeta.decorators = [
+  (Story) => (
+    <GrowthBookProvider growthbook={mrfCutoverOn}>
+      <Story />
+    </GrowthBookProvider>
+  ),
+]
+MrfCutoverOnChildrenBeta.parameters = {
+  msw: [
+    getUser({
+      delay: 0,
+      mockUser: { ...MOCK_USER, betaFlags: { children: true } },
+    }),
+  ],
+}
+
+export const MrfCutoverOnWebhookV1Beta = Template.bind({})
+MrfCutoverOnWebhookV1Beta.decorators = [
+  (Story) => (
+    <GrowthBookProvider growthbook={mrfCutoverOn}>
+      <Story />
+    </GrowthBookProvider>
+  ),
+]
+MrfCutoverOnWebhookV1Beta.parameters = {
+  msw: [
+    getUser({
+      delay: 0,
+      mockUser: {
+        ...MOCK_USER,
+        betaFlags: { createStorageModeForV1Webhook: true },
+      },
+    }),
+  ],
+}
 
 export const StorageModeAckScreen = () => {
   const secretKey = 'mock-secret-key'
--- a/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
+++ b/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
@@ -1,48 +1,22 @@
-import { RegisterOptions } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
 import { BiArrowBack, BiRightArrowAlt } from 'react-icons/bi'
-import {
-  Container,
-  Flex,
-  FormControl,
-  ModalBody,
-  ModalHeader,
-  Skeleton,
-  Text,
-} from '@chakra-ui/react'
+import { Container, Flex, ModalBody, ModalHeader, Text } from '@chakra-ui/react'
 
-import { useFormTitleValidationRules } from '~utils/formValidation'
 import Button from '~components/Button'
-import FormErrorMessage from '~components/FormControl/FormErrorMessage'
-import FormFieldMessage from '~components/FormControl/FormFieldMessage'
-import FormLabel from '~components/FormControl/FormLabel'
 import IconButton from '~components/IconButton'
-import Input from '~components/Input'
 
-import {
-  CreateFormWizardInputProps,
-  useCreateFormWizard,
-} from '../CreateFormWizardContext'
+import { useCreateFormWizard } from '../CreateFormWizardContext'
 
-const FORM_TITLE_LENGTH_WARNING = 65
+import { FormTitleInput } from './FormTitleInput'
 
 export const CreateFormStorageModeScreen = (): JSX.Element => {
   const { t } = useTranslation()
   const {
-    formMethods,
     handleCreateStorageModeOrMultirespondentForm,
     isLoading,
     isFetching,
     goToMrfDetails,
   } = useCreateFormWizard()
-  const {
-    register,
-    formState: { errors },
-    watch,
-  } = formMethods
-
-  const titleInputValue = watch('title')
-  const formTitleValidationRules = useFormTitleValidationRules()
 
   return (
     <>
@@ -67,31 +41,7 @@ export const CreateFormStorageModeScreen = (): JSX.Element => {
             use this if you need a feature not yet available in the current
             version.
           </Text>
-          <FormControl isRequired isInvalid={!!errors.title} mb="2.5rem">
-            <FormLabel useMarkdownForDescription>
-              {t('features.workspace.modals.forms.create.details.name.label')}
-            </FormLabel>
-            <Skeleton isLoaded={!isFetching}>
-              <Input
-                autoFocus
-                {...register(
-                  'title',
-                  formTitleValidationRules as RegisterOptions<
-                    CreateFormWizardInputProps,
-                    'title'
-                  >,
-                )}
-              />
-            </Skeleton>
-            <FormErrorMessage>{errors.title?.message}</FormErrorMessage>
-            {titleInputValue?.length > FORM_TITLE_LENGTH_WARNING ? (
-              <FormFieldMessage>
-                {t(
-                  'features.workspace.modals.forms.create.details.name.message',
-                )}
-              </FormFieldMessage>
-            ) : null}
-          </FormControl>
+          <FormTitleInput mb="2.5rem" />
           <Button
             rightIcon={<BiRightArrowAlt fontSize="1.5rem" />}
             type="submit"
--- a/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
+++ b/apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
@@ -56,12 +56,17 @@ export const CreateFormStorageModeScreen = (): JSX.Element => {
               onClick={goToMrfDetails}
               mr="0.5rem"
             />
-            {t('features.workspace.modals.forms.create.title.setup')}
+            Set up a Storage mode form
           </Flex>
         </Container>
       </ModalHeader>
       <ModalBody whiteSpace="pre-wrap">
         <Container maxW="45rem" p={0}>
+          <Text textStyle="body-2" color="secondary.500" mb="2rem">
+            Storage mode is outdated and no longer receives new features. Only
+            use this if you need a feature not yet available in the current
+            version.
+          </Text>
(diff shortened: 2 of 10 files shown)


eliotlim (approved):

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
```diff
@@ -0,0 +1,112 @@
...
+              mr="0.5rem"
+            />
+            Set up a Storage mode form
+          </Flex>
+        </Container>
+      </ModalHeader>
+      <ModalBody whiteSpace="pre-wrap">
+        <Container maxW="45rem" p={0}>
+          <Text textStyle="body-2" color="secondary.500" mb="2rem">
+            Storage mode is outdated and no longer receives new features. Only
```
kevin9foong, line 40: **issue (non-blocking):** new user-facing copy here is hardcoded English instead of going through `t()`.

**What's happening:** the back `aria-label="Back"` (line 54), the header `Set up a Storage mode form` (line 59), and the body text `Storage mode is outdated…` (line 66) are inline strings — but the same screen already uses `t('features.workspace.modals.forms.create…')` for the title label and the create button.

**Why it bites later:** these strings won't appear in the ms / ta / zh locales, and copy now lives in two different conventions within one file.

**Suggestion:** route all three through `t()` with new keys, like the sibling screens. `SaveSecretKeyContent.tsx` even wraps its `aria-label`s in `t()`.

**Confidence: 95** — verified; flagged independently by the Standards, Spec, Architecture and Divergent axes.

🤖 This comment was generated by an AI code review. Please verify before acting on it.

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
```diff
@@ -0,0 +1,112 @@
...
+import FormLabel from '~components/FormControl/FormLabel'
+import IconButton from '~components/IconButton'
+import Input from '~components/Input'
+
+import {
+  CreateFormWizardInputProps,
+  useCreateFormWizard,
+} from '../CreateFormWizardContext'
+
+const FORM_TITLE_LENGTH_WARNING = 65
```
kevin9foong: **suggestion (non-blocking):** extract the duplicated title input rather than copy-pasting it.

**What's happening:** `FORM_TITLE_LENGTH_WARNING = 65` (line 27) and the whole title `FormControl` block (label, Skeleton, Input, validation, length-warning, ~lines 71–90) are copied near-verbatim from `CreateFormDetailsScreen.tsx` (constant at line 35, block at lines 91–115).

**Why it bites later:** the two copies drift — a change to the warning threshold or the validation has to be made in both, and this cutover screen is temporary so divergence is likely.

**Suggestion:** pull the shared part into a small `<FormTitleInput/>` (both screens already share `useFormTitleValidationRules` and one `formMethods`), or at minimum export the constant.

**Confidence: 82** — verified; raised by the Standards and Divergent axes.

🤖 This comment was generated by an AI code review. Please verify before acting on it.

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormWizardProvider.tsx
```diff
@@ -118,21 +119,23 @@ const useCreateFormWizardContext = (
...
+        case FormResponseMode.Encrypt: {
+          const cutoverDefaultEmails = user?.email ? [user.email] : []
           return createStorageModeFormMutation.mutate(
             {
               title,
               responseMode,
               publicKey: keypair.publicKey,
               workspaceId,
-              emails: emails.filter(Boolean),
+              emails: emails ? emails.filter(Boolean) : cutoverDefaultEmails,
```
kevin9foong: **suggestion (non-blocking):** set the cutover email default at the form's `defaultValues`, not inside the submit handler.

**What's happening:** `emails: emails ? emails.filter(Boolean) : cutoverDefaultEmails` guards a field typed non-optional `string[]` (`CreateFormWizardContext.tsx:26`), but the new storage screen never registers `emails`, so at runtime it is `undefined` — the type and the runtime disagree. The `[user.email]` default itself is correct (it matches the legacy `EmailFormRecipientsInput` default).

**Suggestion — pick one:**
- seed `defaultValues.emails = user?.email ? [user.email] : []` when `isMrfCutoverEnabled`, next to the existing `responseMode` default, so the create handler stays honest with its type; or
- narrow the type to `emails?: string[]`.

**Confidence: 75** — verified type/runtime mismatch; below the correctness gate's 80 bar, kept as Architecture / Divergent design feedback.

🤖 This comment was generated by an AI code review. Please verify before acting on it.

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModal.stories.tsx
```diff
@@ -81,6 +83,79 @@ const Template: StoryFn<CreateFormModalProps> = (args) => {
 }
 export const Default = Template.bind({})
 
+const mrfCutoverOn = new GrowthBook({
+  features: { [featureFlags.mrfCutover]: { defaultValue: true } },
+})
+
+export const MrfCutoverOn = Template.bind({})
```
kevin9foong, line 90: **issue (non-blocking):** the storage-mode page itself has no Storybook story / Chromatic baseline.

**What's happening:** the added stories (`MrfCutoverOn` + the three beta-flag variants) all render only the MRF default screen. None has a `play` / `userEvent` step that clicks the escape-hatch link through to `CreateFormStorageModeScreen`, so that screen is never captured.

**Why it matters:** issue #9453 asks for stories covering "escape-hatch open / storage-mode page", and without one Chromatic won't baseline the new screen.

**Suggestion:** add one story with a `play` function that clicks the escape-hatch link and asserts the storage-mode page renders.

**Confidence: 75** — verified unmet acceptance criterion; scored just under the Spec gate's 80 bar, included here as a follow-up.

🤖 This comment was generated by an AI code review. Please verify before acting on it.

kevin9foong (commented):
## 🤖 AI code review

Reviewed `git diff origin/develop...HEAD` (target `develop`) across four axes — Standards, Spec, Architecture, Divergent — each verified against false positives. 4 inline comments posted; all non-blocking.

**Dropped during verification (for transparency):**
- *MRF default leaking into the Duplicate Form modal* — the `responseMode: Multirespondent` injection lives in the shared `useCommonFormWizardProvider`, but that line is **already on `develop`** and isn't part of this PR's diff, so it's out of scope here (the dupe modal is #9454).
- *`-1` animation-direction hack* in `goToStorageModeDetails` — acceptable: correctly breadcrumbed `TODO [MRF-CUTOVER]`, and the real slide-animation fix is out of scope for code the cutover deletes.
- *Escape-hatch copy hardcoded in `escapeHatchCopy.ts`* — out of this diff (introduced by the merged #9451).

🤖 This review was generated by an AI code review. Please verify before acting on it.

kevin9foong:
# Code Review: PR #9468 — refactor(mrf-cutover): lift wizard navigation + pass source as prop (2/6)

**Author:** kevin9foong · **Base:** `develop` ← `refactor/modal-cutover-prep` · **+235 / −56 · 15 files** · stacked PR 2/6

## Overview

Scaffolding for the upcoming MRF cutover. Three concrete changes:

1. **Lifts MRF-cutover navigation** (`goToStorageModeDetails` / `goToMrfDetails`, new `StorageModeDetails` flow state, `isMrfCutoverEnabled` flag) into the shared `useCommonFormWizardProvider`, threaded through all three modals (Create / Duplicate / Use-Template).
2. **Decouples `DuplicateFormModal` from React context** — `source` (form id + workspace id) now passed as explicit props instead of read from `WorkspaceRowsContext` / `WorkspaceContext`, so it can be mounted outside the workspace tree (e.g. from settings in PR 3/6).
3. **Adds two flags** — `mrf-cutover` growthbook flag (defined, not activated) and `createStorageModeForV1Webhook` user beta flag (schema + zod type + test), plus a new `composeEscapeHatchCopy` util.

The refactor (#1, #2) is clean and the behavioral claim holds: with `mrf-cutover` off, defaults and navigation are unchanged.

## Strengths

- **Prop-drilling over context is the right call** — `DupeFormWizardSource` makes the modal's dependencies explicit and testable, and the test file is correctly simplified (drops `WorkspaceContext` / `WorkspaceRowsContext` mocks, passes `formIdToDuplicate` directly).
- The default-value override is correctly gated and spread-ordered so the flag wins only when on.
- Good test coverage on the genuinely-logic-bearing piece (`composeEscapeHatchCopy`) including the 1/2/3-reason Oxford-comma branches and the falsy-flag case. Backend persistence test for the new beta flag is appropriate.
- The whitespace-only cleanup of the Chakra `RemoveScroll` HACK comment is harmless.

## Issues & Suggestions

**1. Dead code in this PR (confirmed by grep).** None of `composeEscapeHatchCopy`, `StorageModeDetails`, `goToStorageModeDetails`, or `goToMrfDetails` is consumed by any UI in this diff — they're inert until later PRs. That's defensible for stacked scaffolding, but two consequences worth deciding on:
   - If `mrf-cutover` were ever flipped on *before* PR 3+ lands, `goToStorageModeDetails` sets `currentStep` to `StorageModeDetails`, for which **no screen renders** → blank wizard step. Since the flag is off this is latent, but it's a footgun in a stacked series. Consider a code comment noting the screen arrives in a later PR.
   - Verify your lint/CI doesn't fail on unused exports.

**2. `composeEscapeHatchCopy` bypasses i18n.** Every other string in these modals goes through `t(...)`, but this util hardcodes English (`'Need payments? Use the '`, `'old version of FormSG'`). If this copy is user-facing it should likely be translatable, or at least follow the project's i18n convention. Flag for intent — acceptable if MRF cutover is deliberately SG-English-only and temporary.

**3. `formIdToDuplicate: FormId | undefined` is required-but-nullable.** Making the prop required (not `?:`) while allowing `undefined` is a slightly awkward contract — every caller must pass it explicitly, but it can still be empty. That's intentional (forces callers to think about source), so fine, but consider documenting *why* it's `| undefined` rather than optional.

**4. `formId as FormId` cast in `PreviewFormBanner`.** `formId` comes from `usePublicFormContext()` (a `string`). The cast is reasonable for a branded id type, but if the context can yield an empty/undefined formId, the cast masks it. Low risk given `usePreviewForm` is gated on `!!sourceFormId` downstream.

## Risk Assessment

**Low.** The behavior-changing path is feature-flagged off, the context→prop migration preserves the existing `isDefaultWorkspace ? undefined : activeWorkspace._id` logic (just relocated to `WorkspaceRowsProvider`), and tests were updated to match. Main residual risk is the latent blank-screen if the flag is enabled out of order — mitigated by the stacked-PR sequencing.

## Verdict

**Approve with minor comments.** Solid, well-scoped refactor that does what the description says. I'd want answers on (2) i18n intent and a comment addressing (1) the order-of-activation footgun before merge, but neither blocks. The TC1 manual checklist (flag-off modal flows) is still unchecked — confirm that's been run.

made with claude's /review skill

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
kevin9foong: Ignored for now since we do not wish to support i18n since deprecation message is going to be temporary.

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormWizardProvider.tsx
kevin9foong: emails is indeed set as must be defined - it is a type violation for me to do a check for email definition here. will apply a fix to this and subsequent PRs.

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
kevin9foong: accepted. clarified with kenneth (pm) that we probably will keep this setup modal in the near/mid term, so ideally want to ensure this component is well implemented without dups. this will also accelerate AI coding in the future.

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModal.stories.tsx
kevin9foong: lets skip this for now, it is a good to have, but not strictly necessary.

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormWizardProvider.tsx
kevin9foong: lets do option 2, following the mental model “We might not have emails in form state; fix it when we send.”

this keeps the same decision made in this PR but resolves the type violation

kevin9foong: fixed!

apps/frontend/src/features/workspace/components/CreateFormModal/CreateFormModalContent/CreateFormStorageModeScreen.tsx
kevin9foong: fixed, moved it into its own reusable component

Issue #9455, opened 2026-05-20
UseTemplateModal: default to MRF + escape hatch + storage-mode page


kevin9foong:
## What to build

Change the **Use Template** modal so that the new (instantiated) form defaults to **Multirespondent Form (MRF)** regardless of the source template's response mode. The mode-selection screen no longer offers a storage-mode option directly; an "old version of FormSG" escape-hatch link is always visible with copy composed from the user's beta flags (via opengovsg/FormSG#9451).

Clicking the link navigates the modal to a storage-mode page (title input + create button). Submitting calls the existing storage-mode template mutation. A back affordance returns to the MRF default screen.

Visual layout and copy come from the Figma file: [https://www.figma.com/design/LCHsRuUJ0HkTCMjIcxFtQ4/Form-Design-Master-Part-2?node-id=3424-764](<https://www.figma.com/design/LCHsRuUJ0HkTCMjIcxFtQ4/Form-Design-Master-Part-2?node-id=3424-764>).

Backend is unchanged — gating is frontend-only per `docs/adr/0001-frontend-only-storage-form-gating.md`.

Verify end-to-end through the **TDD loop**: Storybook stories (none currently exist for `UseTemplateModal` — add them), agent-browser interaction, Chromatic visual gate, local dev verification.

See repo-root `CONTEXT.md` for terminology.

## Acceptance criteria

- [ ] `UseTemplateModal` defaults `responseMode` to `Multirespondent` regardless of the source template's mode
- [ ] The mode-selection screen no longer offers a storage-mode option directly
- [ ] An "old version of FormSG" escape-hatch link is always visible
- [ ] The link copy is produced by the composer from opengovsg/FormSG#9451
- [ ] Clicking the link navigates to a storage-mode page (title input + create button)
- [ ] Submitting the storage-mode page instantiates a storage-mode form from the template via the existing mutation
- [ ] A back affordance returns the user to the MRF default screen
- [ ] New Storybook stories cover: MRF default, escape-hatch open, beta-flag copy variants
- [ ] Chromatic visual baseline updated
- [ ] No backend changes

## Blocked by

* opengovsg/FormSG#9456 (mrf-cutover feature flag)
* opengovsg/FormSG#9451 (escape-hatch copy composer)

Issue #9452, opened 2026-05-20
Webhook v1 schema infobox on storage-mode webhook settings


kevin9foong:
## What to build

On the form webhook settings page for storage-mode forms only, render a new infobox describing the v1 webhook schema (which is what their webhooks will emit). MRF webhook settings remain unchanged — they continue to show the existing "unavailable" message; no v4 schema infobox is rendered yet (v4 is not implemented).

The "latest version of FormSG" phrase in the infobox copy is a clickable link that opens the **Duplicate Form modal** for the current form, so the admin can migrate this storage-mode form into an MRF in one step. The modal lives in the workspace; from a non-workspace route the cleanest path is to navigate to the workspace with a query param (e.g. `?duplicate=<formId>`) that auto-opens the modal — the exact mechanism is up to the implementer, but the end behaviour is: click → duplicate modal open, prefilled with the current form as source.

The whole infobox is gated on the `mrf-cutover` GrowthBook flag — when OFF, do not render it.

Copy and visual layout come from the Figma file: <https://www.figma.com/design/LCHsRuUJ0HkTCMjIcxFtQ4/Form-Design-Master-Part-2?node-id=3424-764>.

See repo-root `CONTEXT.md` for terminology.

## Acceptance criteria

- [ ] A new presentational component renders on the webhook settings page when the form's response mode is `Encrypt` AND the `mrf-cutover` flag is ON
- [ ] The infobox is not rendered for MRF or any other response mode
- [ ] MRF webhook settings continue to show the existing "unavailable" message unchanged
- [ ] "latest version of FormSG" is a clickable link that opens the Duplicate Form modal with the current form as source
- [ ] Storybook story covers the storage-mode webhook settings page with the new infobox (Chromatic visual coverage)
- [ ] Copy matches Figma

## Blocked by

- #9456 (mrf-cutover feature flag)

PR #9468, opened 2026-05-21
refactor(mrf-cutover): lift wizard navigation + pass source as prop (2/6)

kevin9foong:
> Stacked PR 2/6 for MRF cutover. Base: `feat/mrf-cutover-foundations`.

## Problem

Preparing the create / duplicate / use-template modals for the cutover requires shared wizard scaffolding. Each modal currently owns its own navigation handling, and the duplicate modal pulls its "source" form from React context, making the upcoming MRF-default + escape-hatch wiring duplicative and hard to test. The duplicate modal also needs to be mountable outside the workspace tree so the webhook-v1 infobox (PR 3/6) can open it from settings.

Refs #9452, #9453, #9454, #9455.

## Solution

- Introduces `mrf-cutover` growthbook flag, which will be used subsequently. Not activated in this PR. 
- Introduces `createStorageModeForV1Webhook` user beta flag, it is only used to conditionally construct the escape hatch message string. 

Pure refactor — no user-visible behavior change.

- Lift `mrf-cutover` navigation (switching between MRF default screen and storage-mode escape-hatch screen) into the common wizard provider used by all three modals.
- Pass `source` (form id + workspace id) as explicit props to `DuplicateFormModal` instead of reading from `WorkspaceRowsContext` / `WorkspaceContext`. Lets callers outside the workspace tree mount the modal.
- Remove the now-unused `WorkspaceRowsProvider` wrapping around `CreateFormDetailsScreen`.

**Breaking Changes**

No - backwards compatible.

## Tests

**TC1: existing modal flows unchanged**

- [ ] Open Create Form, Duplicate Form, Use Template modals (flag off)
- [ ] Confirm wizard steps + navigation work exactly as before

### DIFF
--- a/apps/backend/src/app/models/__tests__/user.server.model.spec.ts
+++ b/apps/backend/src/app/models/__tests__/user.server.model.spec.ts
@@ -146,6 +146,18 @@ describe('User Model', () => {
         'This email is not a valid agency email',
       )
     })
+
+    it('should persist createStorageModeForV1Webhook beta flag', async () => {
+      const saved = await User.create({
+        email: VALID_USER_EMAIL,
+        agency: agency._id,
+        contact: VALID_CONTACT,
+        betaFlags: { createStorageModeForV1Webhook: true },
+      })
+
+      const user = await User.findById(saved._id).lean()
+      expect(user?.betaFlags?.createStorageModeForV1Webhook).toBe(true)
+    })
   })
 
   describe('Methods', () => {
--- a/apps/backend/src/app/models/user.server.model.ts
+++ b/apps/backend/src/app/models/user.server.model.ts
@@ -84,6 +84,7 @@ const compileUserModel = (db: Mongoose) => {
         respondentCopy: Boolean,
         statusTracker: Boolean,
         signatureField: Boolean,
+        createStorageModeForV1Webhook: Boolean,
       },
       flags: {
         type: Schema.Types.Map, // of SeenFlags
--- a/apps/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
+++ b/apps/frontend/src/features/admin-form/common/components/PreviewFormBanner/PreviewFormBanner.tsx
@@ -20,6 +20,8 @@ import {
   useDisclosure,
 } from '@chakra-ui/react'
 
+import { FormId } from 'formsg-shared/types/form/form'
+
 import { FORMSG_UAT } from '~constants/links'
 import { ADMINFORM_ROUTE, DASHBOARD_ROUTE } from '~constants/routes'
 import Button, { ButtonProps } from '~components/Button'
@@ -166,7 +168,11 @@ export const PreviewFormBanner = ({
             onClose={onModalClose}
           />
         ) : (
-          <DuplicateFormModal isOpen={isModalOpen} onClose={onModalClose} />
+          <DuplicateFormModal
+            isOpen={isModalOpen}
+            onClose={onModalClose}
+            formIdToDuplicate={formId as FormId}
+          />
         )}
         <Drawer
           placement="bottom"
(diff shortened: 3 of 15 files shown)


eliotlim (approved):
LGTM! 👍 

praise: thank you for structuring the commits in a reviewable way! 💯

kevin9foong:
# Code Review: PR #9468 — refactor(mrf-cutover): lift wizard navigation + pass source as prop (2/6)

**Author:** kevin9foong · **Base:** `develop` ← `refactor/modal-cutover-prep` · **+235 / −56 · 15 files** · stacked PR 2/6

## Overview

Scaffolding for the upcoming MRF cutover. Three concrete changes:

1. **Lifts MRF-cutover navigation** (`goToStorageModeDetails` / `goToMrfDetails`, new `StorageModeDetails` flow state, `isMrfCutoverEnabled` flag) into the shared `useCommonFormWizardProvider`, threaded through all three modals (Create / Duplicate / Use-Template).
2. **Decouples `DuplicateFormModal` from React context** — `source` (form id + workspace id) now passed as explicit props instead of read from `WorkspaceRowsContext` / `WorkspaceContext`, so it can be mounted outside the workspace tree (e.g. from settings in PR 3/6).
3. **Adds two flags** — `mrf-cutover` growthbook flag (defined, not activated) and `createStorageModeForV1Webhook` user beta flag (schema + zod type + test), plus a new `composeEscapeHatchCopy` util.

The refactor (#1, #2) is clean and the behavioral claim holds: with `mrf-cutover` off, defaults and navigation are unchanged.

## Strengths

- **Prop-drilling over context is the right call** — `DupeFormWizardSource` makes the modal's dependencies explicit and testable, and the test file is correctly simplified (drops `WorkspaceContext` / `WorkspaceRowsContext` mocks, passes `formIdToDuplicate` directly).
- The default-value override is correctly gated and spread-ordered so the flag wins only when on.
- Good test coverage on the genuinely-logic-bearing piece (`composeEscapeHatchCopy`) including the 1/2/3-reason Oxford-comma branches and the falsy-flag case. Backend persistence test for the new beta flag is appropriate.
- The whitespace-only cleanup of the Chakra `RemoveScroll` HACK comment is harmless.

## Issues & Suggestions

**1. Dead code in this PR (confirmed by grep).** None of `composeEscapeHatchCopy`, `StorageModeDetails`, `goToStorageModeDetails`, or `goToMrfDetails` is consumed by any UI in this diff — they're inert until later PRs. That's defensible for stacked scaffolding, but two consequences worth deciding on:
   - If `mrf-cutover` were ever flipped on *before* PR 3+ lands, `goToStorageModeDetails` sets `currentStep` to `StorageModeDetails`, for which **no screen renders** → blank wizard step. Since the flag is off this is latent, but it's a footgun in a stacked series. Consider a code comment noting the screen arrives in a later PR.
   - Verify your lint/CI doesn't fail on unused exports.

**2. `composeEscapeHatchCopy` bypasses i18n.** Every other string in these modals goes through `t(...)`, but this util hardcodes English (`'Need payments? Use the '`, `'old version of FormSG'`). If this copy is user-facing it should likely be translatable, or at least follow the project's i18n convention. Flag for intent — acceptable if MRF cutover is deliberately SG-English-only and temporary.

**3. `formIdToDuplicate: FormId | undefined` is required-but-nullable.** Making the prop required (not `?:`) while allowing `undefined` is a slightly awkward contract — every caller must pass it explicitly, but it can still be empty. That's intentional (forces callers to think about source), so fine, but consider documenting *why* it's `| undefined` rather than optional.

**4. `formId as FormId` cast in `PreviewFormBanner`.** `formId` comes from `usePublicFormContext()` (a `string`). The cast is reasonable for a branded id type, but if the context can yield an empty/undefined formId, the cast masks it. Low risk given `usePreviewForm` is gated on `!!sourceFormId` downstream.

## Risk Assessment

**Low.** The behavior-changing path is feature-flagged off, the context→prop migration preserves the existing `isDefaultWorkspace ? undefined : activeWorkspace._id` logic (just relocated to `WorkspaceRowsProvider`), and tests were updated to match. Main residual risk is the latent blank-screen if the flag is enabled out of order — mitigated by the stacked-PR sequencing.

## Verdict

**Approve with minor comments.** Solid, well-scoped refactor that does what the description says. I'd want answers on (2) i18n intent and a comment addressing (1) the order-of-activation footgun before merge, but neither blocks. The TC1 manual checklist (flag-off modal flows) is still unchecked — confirm that's been run.

PR #9469, opened 2026-05-21
fix(frontend): react translate compatibility

eliotlim:
## Problem

Browser translation engines substitute text nodes with font nodes, which can cause application crashes.

Closes FRM-2417.

## Solution

**Breaking Changes** 

**Bug Fixes**:

- Wrap other text nodes in spans for browser translation compatibility.

### DIFF
--- a/apps/frontend/src/features/admin-form/create/workflow/components/WorkflowContent/NewStepBlock/NewStepBlock.tsx
+++ b/apps/frontend/src/features/admin-form/create/workflow/components/WorkflowContent/NewStepBlock/NewStepBlock.tsx
@@ -1,9 +1,10 @@
 import { useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 import { BiPlus } from 'react-icons/bi'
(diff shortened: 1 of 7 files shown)


apps/frontend/src/features/admin-form/create/workflow/components/WorkflowContent/NewStepBlock/NewStepBlock.tsx
```diff
@@ -1,10 +1,11 @@
 import { useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 import { BiPlus } from 'react-icons/bi'
-import { Button } from '@chakra-ui/react'
 
 import { FormWorkflowStep } from 'formsg-shared/types'
 
+import Button from '~components/Button'
```
kevin9foong, line 7: comment: nice to use the common button instead of base chakra button

kevin9foong (approved):
verified dropdown highlight works - lgtm

PR #9482, opened 2026-05-22
fix(frontend): close 0-byte upload bypass paths in attachment field

### LINKED ISSUE #9479: Frontend: close 0-byte upload bypass paths in Attachment field
## What to build

Three frontend paths currently allow a 0-byte file to be handed to `onChange` and uploaded to S3, landing in the GuardDuty quarantine bucket and breaking the virus-scanner lambda. Close all three in the Attachment field components.

**Path 1 — non-image bypass in validator**
In `Attachment.tsx`, the `file.size === 0` check sits inside `if (!IMAGE_UPLOAD_TYPES_TO_COMPRESS.includes(file.type))`. Move it out so it runs for all file types (images included) in the dropzone `fileValidator`.

**Path 2 — image compression produces empty blob**
In `Attachment.tsx`, after `imageCompression(acceptedFile, {...})` resolves, check the resulting `File.size`. If 0:
- Do NOT call `onChange`.
- Surface a user-facing error using a new i18n key `error.imageProcessingError` → "There was an issue processing your image. Please try again or upload a different file."
- Emit `datadogLogs.logger.warn` with payload `{ formId, fileName, originalSize, originalType, reason: 'imageCompressionEmpty' }`.

**Path 3 — file clone resolves to 0 bytes**
In `AttachmentField.tsx`, after the `new File([await fileArrayBuffer(file)], ...)` clone, check `clone.size === 0` (and ideally compare against `file.size`). If empty:
- Do NOT call `onChange(clone)`.
- Surface the existing "error reading your file" message.
- Emit `datadogLogs.logger.warn` with payload `{ formId, fileName, originalSize: file.size, cloneSize: clone.size, reason: 'fileCloneEmpty' }`.

`formId` should be pulled from the public form context (confirm the exact hook during implementation). Datadog payloads must contain metadata only — never file contents.

## Acceptance criteria

- [ ] A 0-byte file of any type (image or non-image) is rejected at the dropzone validator with the existing empty-file message
- [ ] `imageCompression()` returning an empty blob does not call `onChange`; user sees the new `error.imageProcessingError` message
- [ ] One Datadog warn log emitted on compression-empty case with `formId`, file metadata, and `reason: 'imageCompressionEmpty'`
- [ ] `fileArrayBuffer` clone resolving to 0 bytes does not call `onChange(clone)`; user sees existing "error reading your file" message
- [ ] One Datadog warn log emitted on clone-empty case with `formId`, both sizes, and `reason: 'fileCloneEmpty'`
- [ ] No file contents (only filename, sizes, type, formId) in Datadog payloads

## Blocked by

None - can start immediately.

kevin9foong:
## Problem

A 0-byte attachment can reach S3 through three independent frontend paths,
landing in the GuardDuty quarantine bucket and breaking the virus-scanner
lambda (it retries on `HeadObject.ContentLength === 0` and logs a
misleading "Body is empty"). This PR closes all three paths in the
Attachment field components.

Closes #9479.

## Solution

Three short-circuits in the attachment flow, all returning the user to the
empty dropzone without invoking `onChange`:

### Path 1 — dropzone validator

The `file.size === 0` check sat inside the non-image branch, so empty PNG/JPEG
files slipped through to the compression branch. Moved the check above the
image-type guard so it runs for every file type.

### Path 2 — empty compression blob

When `browser-image-compression` resolves to a 0-byte blob, surface the new
`attachment.error.imageProcessingError` message and emit a Datadog warn with
`{ formId, fileName, originalSize, originalType, reason: 'imageCompressionEmpty' }`.

### Path 3 — empty file clone

The `new File([await fileArrayBuffer(file)], …)` clone can resolve to 0 bytes
on Android + Google Drive even when the source has content. After cloning we
now compare sizes; if the clone is empty we surface the existing reading-error
message and warn `{ formId, fileName, originalSize, cloneSize, reason: 'fileCloneEmpty' }`.

**Alternatives considered**

- Path 2's log lives in `Attachment.tsx`, which has no form context (the component
  is reused by admin-side form builder / whitelist / workflow code, so
  `usePublicFormContext` would throw there). We considered routing the log up
  to `AttachmentField` via a new `onCompressionEmpty` callback, but that adds
  surface area for the same outcome. Instead `Attachment` accepts an optional
  `formId?: string` prop and `AttachmentField` reads `PublicFormContext` with
  raw `useContext` (non-throwing) and threads it down. Admin usages pass
  nothing and the Datadog payload just carries `formId: undefined`.

**Breaking Changes**

No - backwards compatible. `formId` on `Attachment` is optional.

## Tests

**TC1: 0-byte image of any type is rejected at the dropzone**

- [ ] Open a public form with an attachment field
- [ ] Create a 0-byte file with `.png` extension (e.g. `: > empty.png`) and drag it in
- [ ] Confirm the empty-file error appears and no file is attached
- [ ] Repeat with a 0-byte `.pdf` (non-image path still works)

**TC2: Image-compression producing an empty blob is handled**

- [ ] On a field with a small `attachmentSize` (e.g. 1MB), upload an image that
      forces compression
- [ ] If compression produces an empty result, confirm the new
      "There was an issue processing your image…" message appears and no file
      is attached
- [ ] Confirm one Datadog warn fires with `reason: 'imageCompressionEmpty'`,
      `formId`, and file metadata only (no contents)

**TC3: Empty file clone is rejected**

- [ ] From an Android device, pick an attachment via the Google Drive picker
- [ ] If the clone resolves to 0 bytes, confirm the existing reading-error
      message appears
- [ ] Confirm one Datadog warn fires with `reason: 'fileCloneEmpty'`,
      `formId`, both sizes, and no file contents

> Screenshots omitted — change is functional (validation paths) and the only
> visible UI delta is the new `imageProcessingError` toast string.

### DIFF
--- a/apps/frontend/src/components/Field/Attachment/Attachment.tsx
+++ b/apps/frontend/src/components/Field/Attachment/Attachment.tsx
@@ -236,6 +236,16 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
 
     const fileValidator = useCallback<NonNullable<DropzoneProps['validator']>>(
       (file) => {
+        // 0-byte check runs for all file types — images included — so that
+        // empty files cannot bypass the validator via the compression branch.
+        if (file.size === 0) {
+          return {
+            code: 'file-empty',
+            message: t(
+              'features.publicForm.components.fields.attachment.error.zipParsing',
+            ),
+          }
+        }
         if (!IMAGE_UPLOAD_TYPES_TO_COMPRESS.includes(file.type)) {
           if (maxSize && file.size > maxSize) {
             return {
@@ -243,14 +253,6 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
               message: t(
                 'features.publicForm.components.fields.attachment.error.fileTooLarge',
                 { readableMaxSize },
-              ),
-            }
-          }
-          if (file.size === 0) {
-            return {
-              code: 'file-empty',
-              message: t(
-                'features.publicForm.components.fields.attachment.error.zipParsing',
               ),
             }
           }
--- a/apps/frontend/src/components/Field/Attachment/Attachment.tsx
+++ b/apps/frontend/src/components/Field/Attachment/Attachment.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from 'react'
+import { useCallback, useContext, useMemo } from 'react'
 import { DropzoneProps, useDropzone } from 'react-dropzone'
 import { useTranslation } from 'react-i18next'
 import {
@@ -18,6 +18,8 @@ import { MB } from 'formsg-shared/constants/file'
 
 import { ATTACHMENT_THEME_KEY } from '~theme/components/Field/Attachment'
 import { ThemeColorScheme } from '~theme/foundations/colours'
+
+import { PublicFormContext } from '~features/public-form/PublicFormContext'
 
 import { downloadFile } from './utils/downloadFile'
 import { AttachmentStylesProvider } from './AttachmentContext'
@@ -114,13 +116,6 @@ export interface AttachmentProps extends UseFormControlProps<HTMLElement> {
    * Callback function that is invoked when replace button is clicked. 'Show Replace' must be present for function to work.
    */
   handleReplaceFileOverride?: () => void
-
-  /**
-   * Public form id. When present, attached to Datadog warn logs emitted by
-   * defensive empty-file short-circuits so the originating form can be
-   * identified during quarantine-bucket triage.
-   */
-  formId?: string
 }
 
 export const Attachment = forwardRef<AttachmentProps, 'div'>(
@@ -144,12 +139,16 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
       handleDownloadFileOverride,
       handleRemoveFileOverride,
       handleReplaceFileOverride,
-      formId,
       ...props
     },
     ref,
   ) => {
     const { t } = useTranslation()
+    // Read PublicFormContext non-throwingly so admin-side usages (form builder,
+    // storybook) that mount Attachment outside the public form provider
+    // continue to render. formId is used only for telemetry on the defensive
+    // empty-file short-circuit and may legitimately be undefined.
+    const formId = useContext(PublicFormContext)?.formId
     // Merge given props with any form control props, if they exist.
     const inputProps = useFormControl(props)
     // id to set on the rendered max size FormFieldMessage component.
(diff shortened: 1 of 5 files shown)


apps/frontend/src/components/Field/Attachment/Attachment.tsx
```diff
@@ -246,14 +281,6 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
               ),
             }
           }
-          if (file.size === 0) {
-            return {
-              code: 'file-empty',
-              message: t(
-                'features.publicForm.components.fields.attachment.error.zipParsing',
```
kevin9foong, line 253: note: the error message here should have been `fileEmpty`

apps/frontend/src/components/Field/Attachment/Attachment.tsx
```diff
@@ -221,21 +225,43 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
             initialQuality: 0.8,
             useWebWorker: false,
             preserveExif: true,
-          }).then((blob) =>
-            onChange(
+          }).then((blob) => {
+            if (blob.size === 0) {
+              datadogLogs.logger.warn('attachment empty after compression', {
```
kevin9foong, line 230: note: logs are added for tracking when such incidents occur.

apps/frontend/src/components/Field/Attachment/Attachment.tsx
```diff
@@ -221,21 +225,43 @@ export const Attachment = forwardRef<AttachmentProps, 'div'>(
...
         }
         onChange(acceptedFile)
       },
-      [accept, maxSize, onChange, onError, t],
+      [accept, publicFormId, maxSize, onChange, onError, t],
     )
 
     const fileValidator = useCallback<NonNullable<DropzoneProps['validator']>>(
       (file) => {
+        if (file.size === 0) {
```
kevin9foong, line 257: note: the empty file check should be above the image check since we do not accept empty files including images. we do not log this since this is not an edge occurrence.

eliotlim:
hmm there's a failing CI test: https://github.com/opengovsg/FormSG/actions/runs/26388087581/job/77671431968?pr=9482#step:8:1928

eliotlim (approved):
LGTM 👍 

thank you for fixing this longstanding issue @kevin9foong 🎉

kevin9foong:
> hmm there's a failing CI test: https://github.com/opengovsg/FormSG/actions/runs/26388087581/job/77671431968?pr=9482#step:8:1928

yep, it was due to the change in error code from zip parse error -> empty file error. i have updated the TC to align with this change.

PR #9483, opened 2026-05-22
fix(backend): reject 0-byte uploads at presigned POST policy

### LINKED ISSUE #9480: Backend: reject 0-byte uploads at presigned POST policy
## What to build

Defence in depth against 0-byte attachment uploads landing in the GuardDuty quarantine bucket. Even if a frontend regression slips through, S3 should reject the upload at the edge.

In `apps/backend/src/app/utils/aws-s3.ts`, the presigned POST conditions currently set `['content-length-range', 0, params.size]`. Change the minimum from `0` to `1` so S3 rejects 0-byte uploads.

Verify no legitimate flow relies on a 0-byte presigned upload (none expected — attachment uploads always have content).

## Acceptance criteria

- [ ] Presigned POST policy `content-length-range` minimum is `1` (not `0`)
- [ ] Existing upload flows for non-empty attachments continue to work
- [ ] A direct attempt to POST a 0-byte file using a valid presigned URL is rejected by S3 with `EntityTooSmall`

## Blocked by

None - can start immediately.

kevin9foong:
## Problem

The presigned POST policy allowed `content-length-range` down to 0, so 0-byte uploads succeeded at the S3 edge and landed in the GuardDuty quarantine bucket. The virus-scan lambda then failed with `Body is empty`. No legitimate workflow relies on empty attachment uploads.

We previously introduced FE gating to prevent 0B file uploads, noting that there are no detected compression or cloning failures. This means that 0B files are not likely originating from the FE. 

Instead of allowing 0B file uploads to quarantine bucket and rejecting it at the virus scanner step, we should not accept 0B file uploads to the quarantine bucket in the first place. 

Closes #9480.

## Solution

Bump the policy minimum from `0` to `1` in `createPresignedPostDataPromise`. S3 now returns `EntityTooSmall` for 0-byte PUTs against a valid presigned URL. This is a defense-in-depth backstop for the frontend short-circuits shipped in #9479 — the frontend should never produce a 0-byte upload, but if it does (new entry path, regression), S3 refuses it instead of quarantine.

**Breaking Changes**

No - backwards compatible. No legitimate flow uploads 0-byte files.

## Tests

**TC1: 0-byte upload is rejected at S3**

- [ ] Request a presigned POST URL from staging.
- [ ] `curl` a 0-byte body to that URL with the returned fields.
- [ ] Confirm response is HTTP 400 with `EntityTooSmall`.

**TC2: Normal uploads still work**

- [ ] Submit a storage-mode form on staging with a small (>0B) attachment.
- [ ] Confirm the attachment lands in the clean bucket and decrypts on the responses page.

### DIFF
--- a/apps/backend/src/app/modules/submission/__tests__/submission.service.spec.ts
+++ b/apps/backend/src/app/modules/submission/__tests__/submission.service.spec.ts
@@ -2545,7 +2545,7 @@ describe('submission.service', () => {
             ...expectedCalledWithSubset,
             Bucket: aws.guarddutyQuarantineS3Bucket,
             Fields: { key: uuid1 },
-            Conditions: [['content-length-range', 0, 1]],
+            Conditions: [['content-length-range', 1, 1]],
           },
           expect.any(Function), // anonymous error handling function
         ],
@@ -2554,7 +2554,7 @@ describe('submission.service', () => {
             ...expectedCalledWithSubset,
             Bucket: aws.guarddutyQuarantineS3Bucket,
             Fields: { key: uuid2 },
-            Conditions: [['content-length-range', 0, 2]],
+            Conditions: [['content-length-range', 1, 2]],
           },
           expect.any(Function), // anonymous error handling function
         ],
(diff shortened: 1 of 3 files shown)


scottheng96 (approved):

scottheng96:
LGTM!

PR #9485, opened 2026-05-25
fix: signature canvas for v4 display

scottheng96:
## Problem

When flattening v4 to v1 responses, signature canvas doesn't render due to a wrong transformation of signature `answerArray` value not including signature type.
- Signature canvas is not rendered on individual response page & PDF

## Solution

Add `answer.type` into answerArray when flattening v4 to v1 responses.

**Breaking Changes** 

## Before & After Screenshots

| --- | ---- | ------ |
| Signature canvas | image | image |
| PDF | image | image |

## Tests

TC1: Signature canvas display
- [ ] Create an MRF form, add a signature field
- [ ] Make a submission 
- [ ] Verify that on response page & PDF, signature canvas is shown
- [ ] (regression) When download in CSV, verify that `Signature captured` is displayed as a column value

### DIFF
--- a/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
+++ b/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
@@ -171,7 +171,7 @@ export const flattenV4ToFormFields = ({
           _id: fieldId,
           question,
           fieldType,
-          answerArray: [JSON.stringify(answer.value)],
+          answerArray: [answer.type, JSON.stringify(answer.value)],
         })
         break
       }
--- a/apps/frontend/src/features/public-form/utils/inputTransformation.ts
+++ b/apps/frontend/src/features/public-form/utils/inputTransformation.ts
@@ -213,7 +213,7 @@ const transformToSignatureOutput = (
   schema: SignatureFieldSchema,
   input?: SignatureFieldValues | SignatureFieldResponseV3,
 ): SignatureResponse => {
-  let answerArray: string[] = []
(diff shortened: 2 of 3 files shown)


apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
```diff
@@ -171,7 +171,7 @@ export const flattenV4ToFormFields = ({
           _id: fieldId,
           question,
           fieldType,
-          answerArray: [JSON.stringify(answer.value)],
```
kevin9foong, line 174: question: 

it seems that the signature answerArray needs exactly 2 elements, should we add type safety by using zod's `.length(2)` method to define the type? 

```
export const SignatureResponse = MultiAnswerResponse.extend({
  fieldType: z.literal(BasicField.Signature),
  // Enforces exactly 2 items in the array at runtime and type level
  answerArray: z.array(z.string()).length(2) as unknown as z.Schema<[string, string]>,
})
```

this helps ensure something like this will not occur again.

scottheng96: Makes sense to add type safety, which I believe `z.tuple([z.string(), z.string()])` would work better - z.Schema<[string, string]> cast only satisfies typescript type checker which is not enforced at runtime.

Anyways, `SignatureResponse` doesn't get used during `flattenV4ToFormFIelds.ts` since it pushed directly to `FormField[]`.

let me quickly add it anyways!

kevin9foong (approved):

PR #9486, opened 2026-05-25
chore: email subject title update

scottheng96:
## Problem

Some concerns have risen from users after comms went out informing of FormSG's change in email subject title, and the confusion of the new email template.
- Email body title is confusing: `Completed by all respondents` does not make sense for forms with only one submission
- Email subject title `Completed - ` is not desired
     - Use case: Admins use FormSG form as an incident reporting form, where admins reply to the FormSG submission email that gets sent to the respondent - in many cases, responding to the submission does not necessarily mean that the incident has resolved, and this leads to respondents being confused with `Completed -`, as there may be subsequent back and forths

## Solution

Aligned with Prod and Design, implementation of:
1. Removing email body title for admin emails
2. Update `Completed - ` with `Responses received`

**Breaking Changes** 

### DIFF
--- a/apps/backend/src/app/services/mail/mail.service.ts
+++ b/apps/backend/src/app/services/mail/mail.service.ts
@@ -860,7 +860,7 @@ export class MailService {
       return this.#sendEmailWithTemplate({
         emails: adminEmails,
         formId: String(form._id),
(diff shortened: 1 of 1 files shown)


kevin9foong (approved):
LGTM

Issue #9470, opened 2026-05-21
Fix Browser Translations Part 2


eliotlim:
**Description**

**Problem**

A clear and concise description of what the problem is. Ex. I'm always frustrated when \[...\]

**Solution**

*A clear and concise description of what you want to happen.*

**Alternatives**

*A clear and concise description of any alternative solutions or features you've considered.*

**Additional context**

*Add any other context or screenshots about the feature request here.*

If this is a front end work, please provide 

* **image / screenshot**
* **link to the Figma**

PR #9501, opened 2026-05-28
fix(table): show full dropdown label in disabled MRF cells

### LINKED ISSUE #9498: Table dropdown cells: reveal full selected label when disabled (MRF subsequent step)
## What to build

In a multi-respondent form (MRF), when a Table field has a Dropdown column and the field is rendered disabled for a subsequent step's respondent (or in any other disabled context — admin preview, response view), the selected dropdown option is currently truncated with `...` and there is no way for the respondent to read the full label. This slice fixes that end-to-end.

The disabled Dropdown column cell continues to look like a dropdown (the chevron caret and clear button remain visible to match editable rows), but the selected label is no longer ellipsis-truncated:

- the label allows horizontal scrolling within the cell (`overflow: auto; whiteSpace: nowrap`, no `noOfLines`, no `pointerEvents: none` on the label node),
- the label sets a native `title` attribute equal to the full selected label, giving a hover tooltip on desktop.

Row height and column layout are unchanged. The dropdown itself remains non-interactive — `isDisabled` semantics (cannot open, cannot change selection) are preserved.

The behaviour is gated behind a new opt-in prop on the shared select component (`isDisabledScrollable`), forwarded by `SingleSelect`, and set by the Table's Dropdown column cell renderer. Other call sites of `SingleSelect` elsewhere in the app keep their current disabled-truncation styling.

Affected areas (subject to drift):
- `apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx`
- `apps/frontend/src/components/Dropdown/SingleSelect.tsx` (prop forwarding + types)
- `apps/frontend/src/templates/Field/Table/ColumnCell.tsx` (`DropdownColumnCell`)

## Acceptance criteria

- [ ] New opt-in boolean prop `isDisabledScrollable` exists on `SelectCombobox` and is forwarded by `SingleSelect`.
- [ ] When `isDisabled && isDisabledScrollable`, the selected label is rendered without ellipsis truncation, allows horizontal scroll, and exposes the full label via a native `title` attribute.
- [ ] The chevron caret and clear button remain visually present in this state.
- [ ] The dropdown cannot be opened or have its value changed while disabled (existing semantics preserved).
- [ ] `DropdownColumnCell` in the Table field sets `isDisabledScrollable` on its `SingleSelect`, so every disabled Table dropdown cell (MRF subsequent step, admin preview, response view) gets the new behaviour.
- [ ] Without `isDisabledScrollable`, existing `SingleSelect` consumers render disabled state identically to before (no visual regression).
- [ ] Unit test on `SelectCombobox`: with `isDisabled` + `isDisabledScrollable`, the full selected label is queryable in the DOM with a `title` attribute matching the label; caret and clear button are present; without the new prop, the existing truncated rendering still applies.
- [ ] Integration test on `DropdownColumnCell` within a Table field: when the table is disabled and a row has a long selected dropdown value, the full label is reachable (no ellipsis in accessible name), and the dropdown cannot be opened.

## Blocked by

None - can start immediately.

kevin9foong:
## Problem

In a multi-respondent form, when a Table dropdown column cell is disabled for the current step's respondent (e.g. it was filled by an earlier step and not re-assigned), the selected option was rendered with an ellipsis (`...`) and no way to reveal the rest. Subsequent respondents could not see what was previously chosen. Closes #9498.

## Solution

Disabled dropdown column cells now drop the ellipsis: the selected label allows horizontal scrolling within the cell and exposes the full text via a native `title` tooltip on hover. The caret and clear button stay visible so the cell still reads as a dropdown, and row height is unchanged. The behaviour is gated behind a new opt-in `isDisabledScrollable` prop on `SingleSelect` so other call sites keep their current rendering; `DropdownColumnCell` is the only consumer.

**Alternatives considered**

We considered rendering the disabled cell as plain wrapped text instead of a select trigger, but that changes row height and makes the column look inconsistent with editable rows. We also considered mimicking the disabled text input exactly (no caret, no clear button), but we want the affordance that it is still a dropdown so a respondent recognises the kind of value at a glance.

**Screenshots**

Storybook: `TableField/DisabledWithLongDropdownOption`, selected value = `"This is a very long dropdown option label that should not be truncated when disabled"`.

| State | Screenshot |
| --- | --- |
| Before — ellipsis-truncated, no scroll, no tooltip | before |
| After — same render at rest (row height + controls unchanged) | after |
| After — horizontally scrolled to reveal the rest of the label | after-scrolled |
| After - with hover tooltip | image | 

The native `title` tooltip on hover is not visible in headless capture.

**Breaking Changes**

No - backwards compatible. `isDisabledScrollable` defaults to `false` and is only opted into by `DropdownColumnCell`.

## Tests

**TC1: MRF subsequent step sees full previous-step dropdown selection**

- [ ] Open an MRF where step 1 fills a Table field with a Dropdown column containing a long option label.
- [ ] Submit step 1 and proceed to step 2 (where the Table field is not assigned).
- [ ] Hover the disabled dropdown cell — the native browser tooltip shows the full label.
- [ ] Drag horizontally inside the cell — the label scrolls to reveal the rest of the text.
- [ ] Confirm caret and clear button are visible but the dropdown cannot be opened or cleared.

**TC2: No regression on standalone disabled dropdown**

- [ ] Open Storybook `Templates/Field/DropdownField` (or any non-table disabled `SingleSelect`).
- [ ] Confirm the disabled rendering is unchanged from `develop`: ellipsis truncation, no `title` attribute, no horizontal scroll.

**TC3: No regression on editable Table dropdown cell**

- [ ] Open Storybook `TableField/Default` and select a long dropdown option in a row.
- [ ] Confirm the cell behaves exactly as before — clickable, truncated with ellipsis while editable.

### DIFF
--- a/apps/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
+++ b/apps/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
@@ -300,6 +300,7 @@ export const MultiSelectProvider = ({
     <SelectContext.Provider
       value={{
         fullWidth: false,
+        isDisabledScrollable: false,
         inputRef,
         isClearable: false,
         selectedItem: null,
--- a/apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
+++ b/apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
@@ -36,6 +36,17 @@ const hideScrollbarStyles = {
   },
 } as const
 
+const disabledLabelColor = 'neutral.800'
+
+const truncatedLabelStyles = { noOfLines: 1 } as const
+
+const disabledTruncatedLabelStyles = {
+  ...truncatedLabelStyles,
+  ...hideScrollbarStyles,
+  color: disabledLabelColor,
+  pointerEvents: 'none',
+} as const
+
 export const SelectCombobox = forwardRef<HTMLInputElement>(
   (_props, ref): JSX.Element => {
     const {
@@ -103,27 +114,26 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
                 aria-disabled={isDisabled}
               />
             ) : null}
-            <Text
-              textStyle="body-1"
-              {...(isDisabled && isDisabledScrollable
-                ? {
-                    ...hideScrollbarStyles,
-                    color: 'neutral.800',
-                    title: selectedItemMeta.label,
-                  }
-                : {
-                    noOfLines: 1,
-                    ...(isDisabled
-                      ? {
-                          ...hideScrollbarStyles,
-                          color: 'neutral.800',
-                          pointerEvents: 'none',
-                        }
-                      : {}),
-                  })}
-            >
-              {selectedItemMeta.label}
-            </Text>
+            {(() => {
+              const disabledScrollableLabelStyles = {
+                ...hideScrollbarStyles,
+                color: disabledLabelColor,
+                title: selectedItemMeta.label,
+              }
+
+              const labelStyles =
+                isDisabled && isDisabledScrollable
+                  ? disabledScrollableLabelStyles
+                  : isDisabled
+                    ? disabledTruncatedLabelStyles
+                    : truncatedLabelStyles
+
+              return (
+                <Text textStyle="body-1" {...labelStyles}>
+                  {selectedItemMeta.label}
+                </Text>
+              )
+            })()}
           </Stack>
           <Input
             isReadOnly={!isSearchable || isReadOnly}
--- a/apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
+++ b/apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
@@ -44,6 +44,7 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
       getInputProps,
       styles,
       isDisabled,
+      isDisabledScrollable,
       isSearchable,
       isReadOnly,
       isInvalid,
@@ -104,14 +105,22 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
             ) : null}
             <Text
               textStyle="body-1"
-              noOfLines={1}
-              {...(isDisabled
+              {...(isDisabled && isDisabledScrollable
                 ? {
                     ...hideScrollbarStyles,
                     color: 'neutral.800',
-                    pointerEvents: 'none',
+                    title: selectedItemMeta.label,
                   }
-                : {})}
+                : {
+                    noOfLines: 1,
+                    ...(isDisabled
+                      ? {
+                          ...hideScrollbarStyles,
+                          color: 'neutral.800',
+                          pointerEvents: 'none',
+                        }
+                      : {}),
+                  })}
             >
               {selectedItemMeta.label}
             </Text>
--- a/apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
+++ b/apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
@@ -130,7 +130,7 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
 
               return (
                 <Text textStyle="body-1" {...labelStyles}>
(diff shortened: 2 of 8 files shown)


apps/frontend/src/components/Dropdown/MultiSelect/MultiSelectProvider.tsx
```diff
@@ -300,6 +300,7 @@ export const MultiSelectProvider = ({
     <SelectContext.Provider
       value={{
         fullWidth: false,
+        isDisabledScrollable: false,
```
kevin9foong, line 303: note: instead of applying to all disabled single select, we only apply to within table fields to prevent unwanted regressions, hence this `isDisabledScrollable` prop

apps/frontend/src/templates/Field/Table/ColumnCell.tsx
```diff
@@ -144,6 +144,7 @@ const DropdownColumnCell = ({
       render={({ field }) => (
         <SingleSelect
           isDisabled={isDisabled}
+          isDisabledScrollable
```
kevin9foong, line 147: note: this change is limited only to table field single selects (aka dropdown fields)

apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
```diff
@@ -104,14 +105,22 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
...
-              {...(isDisabled
+              {...(isDisabled && isDisabledScrollable
                 ? {
                     ...hideScrollbarStyles,
                     color: 'neutral.800',
-                    pointerEvents: 'none',
+                    title: selectedItemMeta.label,
                   }
-                : {})}
+                : {
```
kevin9foong: note: this is the same style as previous. kept so that the existing single select without the prop have the same style

apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
```diff
@@ -104,14 +105,22 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
...
             <Text
               textStyle="body-1"
-              noOfLines={1}
-              {...(isDisabled
+              {...(isDisabled && isDisabledScrollable
                 ? {
                     ...hideScrollbarStyles,
                     color: 'neutral.800',
-                    pointerEvents: 'none',
+                    title: selectedItemMeta.label,
```
kevin9foong: note: this is required for native tooltip to appear after hovering for 1sec.

apps/frontend/src/components/Dropdown/components/SelectCombobox/SelectCombobox.tsx
```diff
@@ -102,19 +114,26 @@ export const SelectCombobox = forwardRef<HTMLInputElement>(
...
+              const labelStyles =
+                isDisabled && isDisabledScrollable
+                  ? disabledScrollableLabelStyles
+                  : isDisabled
+                    ? disabledTruncatedLabelStyles
+                    : truncatedLabelStyles
+
+              return (
+                <Text textStyle="body-1" {...labelStyles}>
+                  {selectedItemMeta.label}
```
eliotlim: comment: don't mind me, just gonna slip a little browser translation fix (see #9470) in here.

```suggestion
                  <span>{selectedItemMeta.label}</span>
```

apps/frontend/src/templates/Field/Table/TableField.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { render, screen } from '@testing-library/react'
+
+import * as stories from './TableField.stories'
+
+const { DisabledWithLongDropdownOption } = composeStories(stories)
+
+const LONG_DROPDOWN_OPTION =
+  'This is a very long dropdown option label that should not be truncated when disabled'
+
+describe('TableField — disabled dropdown column cell', () => {
```
eliotlim, line 11: praise: great work adding an additional chromatic test to prevent regressions! 💯

eliotlim (approved):
LGTM! 👍 

thank you for thinking ahead to prevent regressions 💯

eliotlim:
All UI tests still passing since adding the span fix for browser translation compatibility!

PR #9502, opened 2026-05-28
feat: answer object encryption

scottheng96:
## Problem

As part of answerObject project, this PR focuses on milestone 2, which is introducing v4 encryption into MRF flow.
1. encrypt submissions into v4, feature-flagged
2. include v4-> v3 adaptor at
    - next step submission decryption (FE rehydration)
    - submission validation (when previous response is compared with current response)

Closes FRM-2430

## Solution

1. Transform response from v3 to v4 before encryption @ `encryptSubmission`
2. Transform response from v4 to v3 at 
     - FE rehydration
     - submission validation (`validateSubmission`)
3. (pre-requisite) remove feature-flag of v4 decryption to always serve decryption in v4, so as to reduce extra work of including v4->v3 support for decryption (milestone #1).
4. answerObject v4 encryption is gated behind a gb featureflag (answer-object-encryption)
5. Update cjs-entry.cjs to let BE import some new named exports

To easily differentiate between v3 and v4 response since they are using the same submission flow:
1. At the field response level, we can check for presence of `provenance`
2. At the submission level, we can check for `mrfVersion: 2`. Previous MRF submissions are 1, or `null`.

**Breaking Changes** 

- Yes - this PR contains breaking changes
**v4encryption** is a one-way door as it will start encrypting responses in v4. Thus, a phased-rollout will be implemented through whitelisting formIds. growthbook feature-flag has been updated to support exception rules relating to `formId` for this.

## Tests

**TC1: Test answerbject v4 encryption**
- [ ] Create an MRF form with all fields, including Myinfo
- [ ] In Singpass settings, toggle on `collect NRIC`
- [ ] Add a 2-step workflow, assigning all the fields to step 1 and none to a step 2
- [ ] Attempt a full workflow completion
   - [ ] Verify step 1 can be submitted successfullly
   - [ ] Verify that at step 2, responses are rehydrated correctly in the FE
   - [ ] Verify step 2 can be submitted successfully
   - [ ] Verify that responses are correctly displayed on the response page
   - [ ] Verify that responses are correct displayed on csv
      - [ ] special attention to address field, signature field, table field, singpass validated NRIC

## Deploy Notes

Before deployment, ensure that `answer-object-encryption` is toggled to `OFF`, serving `FALSE` for the production environment

### DIFF
--- a/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -1,5 +1,5 @@
 import type { FieldResponsesV4, FormFieldMeta } from '@opengovsg/formsg-sdk'
-import { adaptV3ToV4 } from '@opengovsg/formsg-sdk'
+import { adaptV3ToV4, adaptV4ToV3 } from '@opengovsg/formsg-sdk'
 import { celebrate, Joi, Segments } from 'celebrate'
 import crypto from 'crypto'
 import { NextFunction } from 'express'
@@ -460,6 +460,7 @@ export const validateMultirespondentSubmission = async (
               previousSubmission: {
                 encryptedContent: mrfSubmission.encryptedContent,
                 version: mrfSubmission.version,
+                mrfVersion: mrfSubmission.mrfVersion,
               },
               workflowStep: mrfSubmission.workflowStep + 1,
               workflow: mrfSubmission.workflow,
@@ -557,6 +558,9 @@ export const validateMultirespondentSubmission = async (
                     req.body.submissionSecretKey,
                     previousSubmission,
                   )
+                console.log(
+                  `previousSubmissionDecryptedContent: ${JSON.stringify(previousSubmissionDecryptedContent)}`,
+                )
 
                 if (!previousSubmissionDecryptedContent) {
                   return err(
@@ -564,8 +568,26 @@ export const validateMultirespondentSubmission = async (
                   )
                 }
 
-                const previousResponses =
-                  previousSubmissionDecryptedContent.responses as ParsedClearFormFieldResponsesV3
+                /**
+                 * Since the incoming client responses are in V3,
+                 * if previous submission was encrypted in V4 format, convert to V3
+                 * to facilitate comparison.
+                 */
+                const previousResponses = (() => {
+                  const responses = previousSubmissionDecryptedContent.responses
+                  const entries = Object.values(responses)
+                  if (
+                    entries.length > 0 &&
+                    'provenance' in (entries[0] as object) && //TODO: verify preferred option
+                    previousSubmission.mrfVersion === 2
+                  ) {
+                    return adaptV4ToV3(
+                      responses as FieldResponsesV4,
+                    ) as ParsedClearFormFieldResponsesV3
+                  }
+                  // Response is in v3 format, return as is
+                  return responses as ParsedClearFormFieldResponsesV3
+                })()
 
                 const previousNonEditableFieldIdsWithResponses = Object.keys(
                   previousResponses,
--- a/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -3,7 +3,7 @@ import {
   adaptV3ToV4,
   adaptV4ToV3,
   isFieldResponsesV4,
-} from '@opengovsg/formsg-sdk'
+} from '@opengovsg/formsg-sdk/adapters'
 import { celebrate, Joi, Segments } from 'celebrate'
 import crypto from 'crypto'
 import { NextFunction } from 'express'
--- a/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -735,6 +735,12 @@ export const encryptSubmission = async (
   res: Parameters<ProcessedMultirespondentSubmissionHandlerType>[1],
   next: NextFunction,
 ) => {
+  // Set growthbook formId attribute for targetted feature flagging
+  void req.growthbook?.setAttributes({
+    ...req.growthbook.getAttributes(),
+    formId: req.params.formId,
+  })
+
   const formDef = req.formsg.formDef
   const formPublicKey = formDef.publicKey
   const responses = req.body.responses
--- a/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -1,6 +1,9 @@
+import type { FieldResponsesV4, FormFieldMeta } from '@opengovsg/formsg-sdk'
+import { adaptV3ToV4 } from '@opengovsg/formsg-sdk'
 import { celebrate, Joi, Segments } from 'celebrate'
 import crypto from 'crypto'
 import { NextFunction } from 'express'
+import { featureFlags } from 'formsg-shared/constants'
 import {
   BasicField,
   FormAuthType,
@@ -781,12 +784,38 @@ export const encryptSubmission = async (
     req.formsg.unencryptedAttachments = unencryptedAttachments
   }
 
+  const useV4Encryption =
+    req.growthbook?.isOn(featureFlags.answerObjectEncryption) ?? false
+
+  let responsesToEncrypt:
+    | Record<
+        string,
+        ParsedClearFormFieldResponseV3 | StrippedAttachmentResponseV3
+      >
+    | FieldResponsesV4 = strippedAttachmentResponses
+
+  if (useV4Encryption) {
+    // Build FormFieldMeta map from form definition for question text
+    const formFieldMeta: Record<string, FormFieldMeta> = {}
+    for (const field of formDef.form_fields) {
+      formFieldMeta[field._id] = {
+        question: field.title,
+        ...(field.myInfo?.attr && { myInfo: { attr: field.myInfo.attr } }),
+      }
+    }
+
+    responsesToEncrypt = adaptV3ToV4(strippedAttachmentResponses, {
+      formFields: formFieldMeta,
+      provenance: {},
+    })
+  }
+
   const {
     encryptedContent,
     encryptedSubmissionSecretKey,
     submissionSecretKey,
     submissionPublicKey,
-  } = formsgSdk.cryptoV3.encrypt(strippedAttachmentResponses, formPublicKey)
+  } = formsgSdk.cryptoV3.encrypt(responsesToEncrypt, formPublicKey)
 
   const encryptedAttachments =
     await getEncryptedAttachmentsMapFromAttachmentsMap(
@@ -806,12 +835,10 @@ export const encryptSubmission = async (
     workflowStep: req.body.workflowStep,
     responses,
     /**
-     * MRF Version: 1
-     * ====================
-     * - Encrypted payload does not contain attachment contents
-     * - Encrypted Attachment now encrypted by mrf / submission Public Key instead of Form Public Key
+     * MRF Version: 1 — V3 encrypted responses
+     * MRF Version: 2 — V4 encrypted responses (with provenance)
      */
-    mrfVersion: 1,
+    mrfVersion: useV4Encryption ? 2 : 1,
   }
 
   return next()
--- a/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -558,9 +558,6 @@ export const validateMultirespondentSubmission = async (
                     req.body.submissionSecretKey,
                     previousSubmission,
                   )
-                console.log(
-                  `previousSubmissionDecryptedContent: ${JSON.stringify(previousSubmissionDecryptedContent)}`,
-                )
 
                 if (!previousSubmissionDecryptedContent) {
                   return err(
--- a/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
+++ b/apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
@@ -1,5 +1,9 @@
 import type { FieldResponsesV4, FormFieldMeta } from '@opengovsg/formsg-sdk'
-import { adaptV3ToV4, adaptV4ToV3 } from '@opengovsg/formsg-sdk'
+import {
+  adaptV3ToV4,
+  adaptV4ToV3,
+  isFieldResponsesV4,
+} from '@opengovsg/formsg-sdk'
 import { celebrate, Joi, Segments } from 'celebrate'
 import crypto from 'crypto'
 import { NextFunction } from 'express'
@@ -572,12 +576,7 @@ export const validateMultirespondentSubmission = async (
                  */
                 const previousResponses = (() => {
                   const responses = previousSubmissionDecryptedContent.responses
-                  const entries = Object.values(responses)
-                  if (
-                    entries.length > 0 &&
-                    'provenance' in (entries[0] as object) && //TODO: verify preferred option
-                    previousSubmission.mrfVersion === 2
-                  ) {
+                  if (isFieldResponsesV4(responses)) {
                     return adaptV4ToV3(
                       responses as FieldResponsesV4,
                     ) as ParsedClearFormFieldResponsesV3
@@ -757,7 +756,6 @@ export const encryptSubmission = async (
   res: Parameters<ProcessedMultirespondentSubmissionHandlerType>[1],
   next: NextFunction,
 ) => {
-  // Set growthbook formId attribute for targetted feature flagging
   void req.growthbook?.setAttributes({
     ...req.growthbook.getAttributes(),
     formId: req.params.formId,
--- a/apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
+++ b/apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
@@ -27,7 +27,6 @@ import { augmentDecryptedResponses } from './ResponsesPage/storage/utils/augment
 import {
   buildFormFieldMetaMap,
   processDecryptedContent,
-  processDecryptedContentV3,
   processDecryptedContentV4,
 } from './ResponsesPage/storage/utils/processDecryptedContent'
 
--- a/apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
+++ b/apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
@@ -91,12 +91,10 @@ export const getDecryptedSubmissionById = async ({
   formId,
   submissionId,
   secretKey,
-  useV4,
 }: {
   formId: string
   submissionId: string
   secretKey?: string
-  useV4?: boolean
 }) => {
   if (!secretKey) return
 
@@ -120,57 +118,38 @@ export const getDecryptedSubmissionById = async ({
       break
     }
     case SubmissionType.Multirespondent: {
-      const decryptedContent = formsgSdk.cryptoV3.decrypt(secretKey, {
-        encryptedContent: encryptedSubmission.encryptedContent,
-        encryptedSubmissionSecretKey:
-          encryptedSubmission.encryptedSubmissionSecretKey,
-        verifiedContent: encryptedSubmission.verifiedContent,
-        version: encryptedSubmission.version,
-      })
-      if (!decryptedContent)
-        throw new Error('Could not decrypt the multirespondent form response')
-      processedContent = await processDecryptedContentV3(
+      const formFieldsMeta = buildFormFieldMetaMap(
         encryptedSubmission.form_fields,
-        decryptedContent,
       )
-      submissionSecretKey = decryptedContent.submissionSecretKey
+      const decryptedV4 = formsgSdk.cryptoV3.decryptToV4(
+        secretKey,
+        {
+          encryptedContent: encryptedSubmission.encryptedContent,
+          encryptedSubmissionSecretKey:
+            encryptedSubmission.encryptedSubmissionSecretKey,
+          verifiedContent: encryptedSubmission.verifiedContent,
+          version: encryptedSubmission.version,
+        },
+        formFieldsMeta,
+      )
 
-      if (useV4) {
-        const formFieldsMeta = buildFormFieldMetaMap(
-          encryptedSubmission.form_fields,
-        )
-        const decryptedV4 = formsgSdk.cryptoV3.decryptToV4(
-          secretKey,
-          {
-            encryptedContent: encryptedSubmission.encryptedContent,
-            encryptedSubmissionSecretKey:
-              encryptedSubmission.encryptedSubmissionSecretKey,
-            verifiedContent: encryptedSubmission.verifiedContent,
-            version: encryptedSubmission.version,
+      if (!decryptedV4) {
+        datadogLogs.logger.error('Could not decrypt MRF response in v4', {
+          meta: {
+            action: 'getDecryptedSubmissionById',
+            formId,
+            submissionId,
           },
-          formFieldsMeta,
-        )
-
-        if (!decryptedV4) {
-          datadogLogs.logger.warn(
-            'Could not decrypt MRF response in v4, falling back to v3',
-            {
-              meta: {
-                action: 'getDecryptedSubmissionById',
-                formId,
-                submissionId,
-              },
-            },
-          )
-        } else {
-          processedContent = processDecryptedContentV4(
-            encryptedSubmission.form_fields,
-            decryptedV4.responses,
-            decryptedV4.verified,
-          )
-          submissionSecretKey = decryptedV4.submissionSecretKey
-        }
+        })
+        throw new Error('Could not process submission content')
       }
+
+      processedContent = processDecryptedContentV4(
+        encryptedSubmission.form_fields,
+        decryptedV4.responses,
+        decryptedV4.verified,
+      )
+      submissionSecretKey = decryptedV4.submissionSecretKey
       mrfVersion = encryptedSubmission.mrfVersion
       break
     }
--- a/apps/frontend/src/features/admin-form/responses/IndividualResponsePage/queries.ts
+++ b/apps/frontend/src/features/admin-form/responses/IndividualResponsePage/queries.ts
@@ -1,8 +1,6 @@
 import { useQuery } from 'react-query'
 import { useParams } from 'react-router-dom'
-import { useFeatureIsOn } from '@growthbook/growthbook-react'
 
-import { featureFlags } from 'formsg-shared/constants'
 import { FormResponseMode } from 'formsg-shared/types'
 
 import { useToast } from '~hooks/useToast'
@@ -38,12 +36,10 @@ export const useGetIndividualDecryptedSubmission = ({
   })
   const { secretKey } = useStorageResponsesContext()
   const { data: { responseMode } = {} } = useAdminForm()
-  const useV4 = useFeatureIsOn(featureFlags.answerObjectDecryption)
 
   return useQuery(
-    adminFormResponsesKeys.individual(formId, submissionId, useV4),
-    () =>
-      getDecryptedSubmissionById({ formId, submissionId, secretKey, useV4 }),
+    adminFormResponsesKeys.individual(formId, submissionId),
+    () => getDecryptedSubmissionById({ formId, submissionId, secretKey }),
     {
       staleTime:
         responseMode === FormResponseMode.Multirespondent
--- a/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
+++ b/apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
@@ -71,7 +71,7 @@ export const flattenV4ToFormFields = ({
 
     // Generic string fields (including yes_no)
     if (GENERIC_STRING_FIELD_TYPES.has(fieldType)) {
-      const answer = field.answer as StringAnswerV4
+      const answer = fieldResponse.answer as StringAnswerV4
       v1Fields.push({
         _id: fieldId,
         question,
@@ -84,7 +84,7 @@ export const flattenV4ToFormFields = ({
     switch (fieldType) {
       case BasicField.Email:
       case BasicField.Mobile: {
-        const answer = field.answer as VerifiableAnswerV4
+        const answer = fieldResponse.answer as VerifiableAnswerV4
         v1Fields.push({
           _id: fieldId,
           question,
@@ -98,7 +98,7 @@ export const flattenV4ToFormFields = ({
       }
 
       case BasicField.Radio: {
-        const answer = field.answer as RadioAnswerV4
+        const answer = fieldResponse.answer as RadioAnswerV4
         v1Fields.push({
           _id: fieldId,
           question,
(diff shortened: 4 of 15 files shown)


apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -554,15 +558,36 @@ export const validateMultirespondentSubmission = async (
                     req.body.submissionSecretKey,
                     previousSubmission,
                   )
+                console.log(
+                  `previousSubmissionDecryptedContent: ${JSON.stringify(previousSubmissionDecryptedContent)}`,
+                )
```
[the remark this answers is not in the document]
scottheng96: Fixed.

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/worker/decryption.worker.ts
```diff
@@ -146,57 +146,36 @@ async function decryptSubmissionData(
...
-        encryptedContent,
-        verifiedContent,
-        version,
-      })
-      if (!decryptedObject) {
-        datadogLogs.logger.error(
-          'Invalid decryption for multirespondent response',
-        )
+      const formFieldsMeta = buildFormFieldMetaMap(submissionData.form_fields)
+      const decryptedV4 = ctx.formsgSdk.cryptoV3.decryptToV4(
```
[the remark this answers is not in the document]
scottheng96: Removed!

kevin9foong (commented):
Automated review pass on the V4 (answer-object) encryption rollout for MRF. Overall this looks solid and well-tested, with the right read-before-write sequencing. Inline comments are **suggestions and questions**, not blockers — the main things I'd want resolved before merge are the unresolved `//TODO` and the hand-rolled vs. `isFieldResponsesV4` detection in the middleware.

_🤖 This review was generated by an AI code reviewer. Please verify each point._

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -561,8 +565,26 @@ export const validateMultirespondentSubmission = async (
...
                 }
 
-                const previousResponses =
-                  previousSubmissionDecryptedContent.responses as ParsedClearFormFieldResponsesV3
+                /**
+                 * Since the incoming client responses are in V3,
+                 * if previous submission was encrypted in V4 format, convert to V3
+                 * to facilitate comparison.
+                 */
+                const previousResponses = (() => {
```
kevin9foong, line 577: **Suggestion:** could this reuse the `isFieldResponsesV4` helper that this PR now exports from the SDK? `decryptSubmission.ts` (public form) uses it for exactly this V4 detection, whereas here it's hand-rolled as `'provenance' in entries[0]`. Reusing the helper would keep the two paths in sync.

**Question:** the condition combines the structural probe **and** `previousSubmission.mrfVersion === 2` with `&&`. If they ever disagree (e.g. `mrfVersion === 2` but the first entry lacks `provenance`), it silently falls through to "treat as V3" and the non-editable comparison would run against the wrong shape. Is the dual check intentional, or would a single authoritative source (`mrfVersion === 2`, or `isFieldResponsesV4(...)`) be safer here?

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -561,8 +565,26 @@ export const validateMultirespondentSubmission = async (
...
+                  const responses = previousSubmissionDecryptedContent.responses
+                  const entries = Object.values(responses)
+                  if (
+                    entries.length > 0 &&
+                    'provenance' in (entries[0] as object) && //TODO: verify preferred option
```
kevin9foong: This `//TODO: verify preferred option` looks unresolved — worth settling (or removing) before merge so the detection heuristic is final. Related to the suggestion above about reusing `isFieldResponsesV4`.

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -735,6 +757,12 @@ export const encryptSubmission = async (
   res: Parameters<ProcessedMultirespondentSubmissionHandlerType>[1],
   next: NextFunction,
 ) => {
+  // Set growthbook formId attribute for targetted feature flagging
+  void req.growthbook?.setAttributes({
```
kevin9foong, line 759: Just flagging for awareness (no change needed, I believe): `setAttributes` isn't awaited before the `isOn(answerObjectEncryption)` check below. That's fine because GrowthBook assigns `_ctx.attributes` synchronously and `isOn` reads it synchronously — so the `formId` targeting does take effect. It matches the existing pattern in `encrypt-submission.service.ts`. Mentioning only so a future reader doesn't "fix" it into an `await`.

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -800,12 +854,10 @@ export const encryptSubmission = async (
...
     /**
-     * MRF Version: 1
-     * ====================
-     * - Encrypted payload does not contain attachment contents
-     * - Encrypted Attachment now encrypted by mrf / submission Public Key instead of Form Public Key
+     * MRF Version: 1 — V3 encrypted responses
+     * MRF Version: 2 — V4 encrypted responses (with provenance)
      */
-    mrfVersion: 1,
+    mrfVersion: useV4Encryption ? 2 : 1,
```
kevin9foong, line 858: **Question:** the encrypt stamp here (`? 2 : 1`) and the V4->V3 conversion above both key off an exact `mrfVersion === 2`. A future `mrfVersion: 3` would silently be treated as V3 in the comparison path. Would a `>= 2` check (or a named constant) be more forward-proof, or is exact-match deliberate for now?

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/queries.ts
```diff
@@ -38,12 +36,10 @@ export const useGetIndividualDecryptedSubmission = ({
   })
   const { secretKey } = useStorageResponsesContext()
   const { data: { responseMode } = {} } = useAdminForm()
-  const useV4 = useFeatureIsOn(featureFlags.answerObjectDecryption)
 
   return useQuery(
-    adminFormResponsesKeys.individual(formId, submissionId, useV4),
-    () =>
-      getDecryptedSubmissionById({ formId, submissionId, secretKey, useV4 }),
+    adminFormResponsesKeys.individual(formId, submissionId),
```
kevin9foong, line 41: Dropping `useV4` from the cache key means the admin read path is now always V4 (`decryptToV4` + `processDecryptedContentV4`), so the `answerObjectDecryption` flag is effectively retired on this path. That seems intentional (reads ungated before writes), but two things worth confirming:
- Should the `answerObjectDecryption` constant in `feature-flags.ts` be removed/deprecated now that nothing reads it?
- There's no longer a display-side kill-switch: every admin immediately renders legacy V3 submissions through `adaptV3ToV4` -> `processDecryptedContentV4`. Have we confirmed that produces output identical to the old `processDecryptedContentV3` for existing data?

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -27,7 +27,6 @@ import { augmentDecryptedResponses } from './ResponsesPage/storage/utils/augment
 import {
   buildFormFieldMetaMap,
   processDecryptedContent,
-  processDecryptedContentV3,
   processDecryptedContentV4,
```
kevin9foong, line 30: With this and the worker both dropping `processDecryptedContentV3`, that function appears to be dead code now (only referenced in a doc comment). Suggestion: remove it, or add a note on why it's retained — otherwise it's unclear which decryption pipeline is canonical.

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
```diff
@@ -171,7 +171,7 @@ export const flattenV4ToFormFields = ({
           _id: fieldId,
           question,
           fieldType,
-          answerArray: [JSON.stringify(answer.value)],
+          answerArray: [answer.type, JSON.stringify(answer.value)],
```
kevin9foong, line 174: Adding `answer.type` as the leading element changes the `answerArray` shape for Signature fields. Looks intentional (aligns with the signature-canvas v4 display fix), but could you confirm the downstream consumers — CSV export and the decrypted-row renderer — expect the leading `type` element and won't shift columns?

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -578,7 +578,7 @@ export const PublicFormProvider = ({
           if (response.fieldType === BasicField.Attachment) {
             previousAttachments[id] = Uint8Array.from(
               //@ts-expect-error 'content' required for backward compatibility, but
-              // does not exist on AttachmentFieldResponseV3 in mrfVersion === 1 versions
+              // does not exist on AttachmentFieldResponseV3 in mrfVersion  >= 1 versions
```
kevin9foong: Tiny nit: double space in `mrfVersion  >= 1`. Also, since the guard above flipped from `=== 1` to `!= null`, `mrfVersion: 2` now flows through `decryptAttachments()` too — looks correct, just confirming the `else` (legacy `undefined`) branch is the only intended fallback.

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

packages/sdk/cjs-entry.cjs
```diff
@@ -2,4 +2,11 @@
 // TypeScript compiles `export default function` to `exports.default = fn`,
 // but CJS consumers expect `require('@opengovsg/formsg-sdk')` to return
 // the function directly (matching the old v0.15.0 behavior).
-module.exports = require('./dist/cjs/index.js').default
+const _mod = require('./dist/cjs/index.js')
+module.exports = _mod.default
+// Re-export named exports so CJS consumers can access them
+Object.keys(_mod).forEach((key) => {
```
kevin9foong: Nit / edge case: copying named exports onto the default *function* object works, but a named export colliding with an intrinsic function property (`name`, `length`) would be silently skipped or clobbered. Fine for the current export names — just noting it in case the export surface grows.

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

kevin9foong:
Disclaimer: using this PR as a testbed for AI review experimentation, please ignore the comments marked generated by AI. I am WIP to verify and optimise them to improve the signal-noise ratio

packages/sdk/cjs-entry.cjs
kevin9foong: **Architectural suggestion (altitude): expose the adapters via a subpath export instead of decorating the factory function.**

This loop copies named exports onto the default factory function so CJS consumers (the backend) can reach `adaptV3ToV4` / `adaptV4ToV3`. It works, but it's a workaround for overloading one entry point with two unrelated responsibilities: a *callable* default (the configured SDK factory) and *stateless* utility functions. Hanging utilities off the function object is what creates the intrinsic-property-collision footgun (`name`, `length`, …) and makes the package's runtime export surface implicit.

**There's already an in-package precedent for the cleaner approach.** `packages/sdk/package.json` ships a subpath export `"./dist/types"` whose `require` condition resolves to a plain compiled CJS module (`./dist/cjs/types.js`) — **not** through this `cjs-entry.cjs` flatten shim. Its named exports therefore resolve natively, with no decoration and no collision risk. (`formsg-shared` uses this same subpath pattern at scale.) The only entry that *has* to flatten is `"."`, precisely because it carries the legacy "the module *is* the factory function" contract — the adapters don't need that contract.

Suggested shape — add a subpath alongside the existing ones:
```jsonc
"./adapters": {
  "import":  { "types": "./dist/esm/adapters.d.ts", "default": "./dist/esm/adapters.js" },
  "require": { "types": "./dist/cjs/adapters.d.ts", "default": "./dist/cjs/adapters.js" }
}
```
Backend then imports `import { adaptV3ToV4 } from '@opengovsg/formsg-sdk/adapters'`, the `require` condition hits a normal CJS namespace, named values resolve natively — and `cjs-entry.cjs` can go back to its original one-liner (no loop, no collision caveat).

Not a blocker for this PR — the current approach unblocks the backend. But it follows the repo's own established pattern rather than introducing a decorate-the-function approach that has no precedent here. If keeping the single entry for now, prefer explicit named assignments over the dynamic loop so the export surface is legible and intrinsic-name collisions surface instead of silently no-op'ing.

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -735,6 +757,12 @@ export const encryptSubmission = async (
   res: Parameters<ProcessedMultirespondentSubmissionHandlerType>[1],
   next: NextFunction,
 ) => {
+  // Set growthbook formId attribute for targetted feature flagging
```
kevin9foong: nit: this comment is redundant, it might be good to remove unneeded comments to reduce comment bloat

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -561,8 +565,26 @@ export const validateMultirespondentSubmission = async (
...
 
-                const previousResponses =
-                  previousSubmissionDecryptedContent.responses as ParsedClearFormFieldResponsesV3
+                /**
+                 * Since the incoming client responses are in V3,
+                 * if previous submission was encrypted in V4 format, convert to V3
+                 * to facilitate comparison.
+                 */
+                const previousResponses = (() => {
+                  const responses = previousSubmissionDecryptedContent.responses
```
kevin9foong, line 578: question: should we be using the `isFieldResponsesV4` helper that this PR now exports here instead? Reusing the helper would keep the two paths in sync.

also edge case/potential defensive change for consideration: the condition combines the provenance check and previousSubmission.mrfVersion === 2 with &&. If they ever disagree (e.g. mrfVersion === 2 but the first entry lacks provenance), it silently falls through to "treat as V3" and the non-editable comparison would run against the wrong shape. would using isFieldResponsesV4(...)) be safer here?

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: aligned that it is not a required change to make it `await` since it works

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -27,7 +27,6 @@ import { augmentDecryptedResponses } from './ResponsesPage/storage/utils/augment
 import {
   buildFormFieldMetaMap,
   processDecryptedContent,
-  processDecryptedContentV3,
```
kevin9foong, line 30: question: since this is a one-way door, shall we clean up/remove this function as well? we can always recover the code from Github commit history. 

keeping it might make it unclear which decryption pipeline is canonical.

packages/sdk/cjs-entry.cjs
```diff
@@ -2,4 +2,11 @@
 // TypeScript compiles `export default function` to `exports.default = fn`,
 // but CJS consumers expect `require('@opengovsg/formsg-sdk')` to return
 // the function directly (matching the old v0.15.0 behavior).
-module.exports = require('./dist/cjs/index.js').default
+const _mod = require('./dist/cjs/index.js')
```
kevin9foong: suggestion: instead of adding a loop which copies named exports onto the default factory function so CJS consumers (the backend) can reach adaptV3ToV4 / adaptV4ToV3. Shall we use the current `packages/sdk/package.json` subpath export approach instead to keep things consistent? 

For context, "./dist/types" also faced the same issue which was resolved with subpath exports, similar to our `formsg-shared` package. 

Suggested addition to `packages/sdk/package.json`: 
```
"./adapters": {
  "import":  { "types": "./dist/esm/adapters.d.ts", "default": "./dist/esm/adapters.js" },
  "require": { "types": "./dist/cjs/adapters.d.ts", "default": "./dist/cjs/adapters.js" }
}
```

This achieves 2 benefits: 
- prevents potential named export colliding with an intrinsic function property (name, length) which would be silently skipped or clobbered. Fine for the current export names — but noting it in case the export surface grows.
- Prevents having 2 different methods in the codebase of resolving these named exports

packages/sdk/cjs-entry.cjs
kevin9foong: collated into the final comment

kevin9foong: prompted the model to understand if this change is required and if it differs from how we were doing it previously

kevin9foong: collated into hand written comment

apps/frontend/src/features/public-form/PublicFormProvider.tsx
kevin9foong: nit(typo): double space in comment

apps/frontend/src/features/admin-form/responses/ResponsesPage/storage/utils/flattenV4ToFormFields.ts
kevin9foong: this should be fine and expected change

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: question: is this TODO expected? 

what does this mean? could it be related to reusing `isFieldResponsesV4`?

apps/frontend/src/features/admin-form/responses/IndividualResponsePage/queries.ts
kevin9foong: Should the answerObjectDecryption constant in feature-flags.ts be removed/deprecated now that nothing reads it?
no, our convention is to retain prior flags in code. 

There's no longer a display-side kill-switch: every admin immediately renders legacy V3 submissions through adaptV3ToV4 -> processDecryptedContentV4. Have we confirmed that produces output identical to the old processDecryptedContentV3 for existing data?
yes, we have verified this for the past week.

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: added in the comment on R574

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: lets make it deliberate for now, we might not wish to treat a future v3 the same way as v2 anyways.

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -120,57 +117,38 @@ export const getDecryptedSubmissionById = async ({
...
-        const decryptedV4 = formsgSdk.cryptoV3.decryptToV4(
-          secretKey,
-          {
-            encryptedContent: encryptedSubmission.encryptedContent,
-            encryptedSubmissionSecretKey:
-              encryptedSubmission.encryptedSubmissionSecretKey,
-            verifiedContent: encryptedSubmission.verifiedContent,
-            version: encryptedSubmission.version,
+      if (!decryptedV4) {
+        datadogLogs.logger.error('Could not decrypt MRF response in v4', {
```
kevin9foong, line 136: praise: nice to add logs for tracking

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
```diff
@@ -120,57 +117,38 @@ export const getDecryptedSubmissionById = async ({
...
-        } else {
-          processedContent = processDecryptedContentV4(
-            encryptedSubmission.form_fields,
-            decryptedV4.responses,
-            decryptedV4.verified,
-          )
-          submissionSecretKey = decryptedV4.submissionSecretKey
-        }
+        })
+        throw new Error('Could not process submission content')
```
kevin9foong, line 143: recommendation: i wonder if it also makes sense to either set up a DD alarm for the logs or errorCode to alert us if decryption issues do occur.

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong, line 577: **Suggestion (not blocking): keep the “which format is this?” logic in one place.**

The screen that *shows* responses got a tidy helper this PR — `decryptToV4` — which decrypts a submission **and** quietly sorts out whether it was saved in the old (V3) or new (V4) format, so whoever calls it never has to think about formats. This validation code doesn’t have the matching helper for the *opposite* direction, so it has to work the format out by hand right here: peek for a `provenance` key, check `mrfVersion === 2`, then convert V4→V3 itself.

**Why this bites later:** the rule for “is this the new format?” now lives in two different places — inside the SDK *and* in this block. The next time the format changes, someone has to remember to update both. If they update only one, this comparison quietly runs against the wrong data shape — and this is the security-sensitive check that stops a respondent editing a field they’re not allowed to edit.

**Suggestion:** give the SDK a sibling to `decryptToV4` — say, one that always hands back V3 no matter how the data was stored — and call that here. The hand-written detect-and-convert block then disappears, and the “what format is this” knowledge lives in exactly one easy-to-test spot. Not required for this PR.

_🤖 This is an AI-generated architecture suggestion — non-blocking, please sanity-check before acting on it._

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: **Suggestion (not blocking): decide “is this the new format?” one consistent way.**

Today the code answers that question two different ways: (a) by peeking inside the data for a `provenance` key (`isFieldResponsesV4`), and (b) by reading the saved `mrfVersion` number. This line uses **both**, joined with `&&`. Elsewhere in the app, `mrfVersion` gets read with four *different* comparisons: `=== 1`, `=== 2`, `!= null`, and `!mrfVersion`.

**Why this bites later:** when the two checks disagree — e.g. an empty response, or some future version number — the `&&` makes this silently fall back to “treat it as the old format,” which is the wrong answer on a security check. And because the same `mrfVersion` number is interpreted four different ways, every reader is re-deciding what each number means; the day someone adds `mrfVersion: 3`, they have to find and re-check all of them.

**Suggestion:** pick the saved `mrfVersion` as the single source of truth and wrap it in one tiny helper (e.g. `isV4(submission)`), used everywhere. Keep the `provenance` peek only as the SDK’s internal fallback for old data that has no `mrfVersion`. (This also resolves the unresolved `//TODO` on this line.)

_🤖 This is an AI-generated architecture suggestion — non-blocking, please sanity-check before acting on it._

apps/frontend/src/features/public-form/PublicFormProvider.tsx
```diff
@@ -565,7 +565,7 @@ export const PublicFormProvider = ({
       setPreviousAttachments(decryptedAttachments)
     }
 
-    if (encryptedPreviousSubmission?.mrfVersion === 1) {
+    if (encryptedPreviousSubmission?.mrfVersion != null) {
```
kevin9foong, line 568: **Suggestion (bigger-picture, not blocking): `mrfVersion` is now doing two jobs at once.**

`mrfVersion` originally described **how attachments are stored**. This PR also makes it describe **the response format** (old V3 vs new V4). So one number now stands for two unrelated things. The giveaway is right here: this check used to be `=== 1` (an *attachment* question), and had to be loosened to `!= null` — because the version number moved for a reason (a new response format) that has nothing to do with attachments.

**Why this bites later:** every time *either* concern changes, someone has to re-examine every `mrfVersion === n` check in the app to make sure it didn’t accidentally change meaning for the *other* concern. That’s exactly how an innocent-looking `=== 2` somewhere becomes a silent bug.

**Suggestion:** split the two ideas apart — track “attachment scheme” and “response format” separately, or derive named flags like `hasInlineAttachments` / `isV4` from `mrfVersion` once — so each piece of code asks about the one thing it actually cares about. This is the most involved of the three suggestions; weigh it against the cost of a small schema/semantics change.

_🤖 This is an AI-generated architecture suggestion — non-blocking, please sanity-check before acting on it._

kevin9foong, line 568: note: this assumes that if the mrfVersion is null or undefined, attachments are stored separately. this should be a safe assumption

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: thoughts: good thing to consider, but might not be required if the end goal is to actually decrypt to v4 for v4 responses. 
after reading through the implementation phases, it seems that the end goal after implementation might be to still convert both v4 and v3 into v3 since FE will still take to take responses in v3 form. 

based on this, is this something we should consider? @scottheng96

apps/frontend/src/features/public-form/PublicFormProvider.tsx
kevin9foong: thoughts: might be a rather large change + i think the mrfVersion being present makes sense to indicate how attachments should be processed for now. 

my thoughts are that it is fine as is and thus do not recommend this change

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
kevin9foong: thoughts: this is related to review comment regarding the reuse of the exported `isFieldResponseV4` above. resolving the above might help with this

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -775,12 +806,38 @@ export const encryptSubmission = async (
     req.formsg.unencryptedAttachments = unencryptedAttachments
   }
 
+  const useV4Encryption =
+    req.growthbook?.isOn(featureFlags.answerObjectEncryption) ?? false
```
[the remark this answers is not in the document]
scottheng96: Added tests to verify v3 and v4 encryptions based on feature flag

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
```diff
@@ -554,15 +558,36 @@ export const validateMultirespondentSubmission = async (
...
+                 */
+                const previousResponses = (() => {
+                  const responses = previousSubmissionDecryptedContent.responses
+                  const entries = Object.values(responses)
+                  if (
+                    entries.length > 0 &&
+                    'provenance' in (entries[0] as object) && //TODO: verify preferred option
+                    previousSubmission.mrfVersion === 2
+                  ) {
+                    return adaptV4ToV3(
```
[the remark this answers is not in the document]
scottheng96: Added tests for this

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: Yep thanks for catching that! I implemented this before creating the isFieldResponsesV4 - hence the TODO, and forgot to comeback to it. Fixed!

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
scottheng96: Good idea - I have removed the function as well.

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: Responded above ^ implemented before `isFieldResponsesV4 was created, didn't comeback to it, hence the TODO. fixed!

packages/sdk/cjs-entry.cjs
scottheng96: Thanks for surfacing this. I agree that the loop solution wasn't optimal, but didn't find another approach after time boxing this. Learnt something and happy to keep things consistent with how we're already doing it.

apps/frontend/src/features/admin-form/responses/AdminSubmissionsService.ts
scottheng96: Agree. I'd add it to the list of items as part of milestone #2.

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: Hmm, point taken. There can be some value to have a function in the sdk that always decrypts to V3, but afaik it will be used only for us - since webhooks aren't widely supported in MRF yet, we shouldn't allow and encourage the use of v3 schema to downstream users, and instead push to use for v4.

W.r.t to the value for us, Phase 3 of answerObject is about handling responses in v4 format in both the FE and BE. We'll eventually get to a point when we're encrypting & decrypting in v4, and performing the v3->v4 conversion if necessary. This v4->3 transformation is temporary.

My thoughts are to leave as is, and we'll tackle removing this code in milestone #3

apps/backend/src/app/modules/submission/multirespondent-submission/multirespondent-submission.middleware.ts
scottheng96: Yep! Fixed this

apps/frontend/src/features/public-form/PublicFormProvider.tsx
scottheng96: I'll document this mrfVersion somewhere so we don't forget the lineage of it's changes.

test-decrypt.js
```diff
@@ -0,0 +1,1499 @@
+/**
```
kevin9foong: **This looks like an accidentally-committed scratch/debug script — should it be in the PR?**

`test-decrypt.js` (+1499 lines, repo root) is a manual harness: header says `Usage: node test-decrypt.js ... Toggle between decryptFromSubmissionKey (MRF) and decryptToV4`, it `require()`s the SDK via a hardcoded `./apps/backend/node_modules/...` path, runs `formsg({ mode: 'production' })`, and embeds a hardcoded `secretKey` plus a full `encryptedContent` / `encryptedSubmissionSecretKey` / `verifiedContent` payload.

Reads as a dev aid used while building this PR rather than something meant to ship. Suggest removing it from the branch (recoverable from git history). Beyond repo noise, committing a production-mode decrypt script with an inline key sets a poor precedent even if the data is throwaway.

_🤖 This comment was generated by an AI code review. Please verify before acting on it._

eliotlim: yup agree we should remove this from the PR.

eliotlim (approved):
Had a good long look at this, looks reasonable to me.
Good that it's feature flagged and we've considered the edge cases:
- myInfo attributes
- attachment field handling for early MRF responses

test-decrypt.js
scottheng96: Mb. Removed!

kevin9foong (approved):

PR #9524, opened 2026-06-02
fix(admin-form): show error state when webhook settings fetch fails

### LINKED ISSUE #9512: [Bug] Webhooks settings page shows "only available in storage mode" when settings fetch fails
## Summary

On a storage-mode form, if the admin-form settings request fails or hangs, `SettingsWebhooksPage` renders the "Webhooks are only available in storage mode" message instead of an error state. The UI misdiagnoses the form's response mode — admins are told their form is the wrong type when the fetch simply never succeeded.

## Impact

- **Who:** All form admins on storage-mode forms whose settings fetch fails (transient 5xx, slow backend, CSRF/session drift on the settings endpoint).
- **Severity:** Degraded UX — misleading, not blocking. The admin's actual data is fine; they are just told something incorrect about it.
- **Workaround:** Hard refresh. Works for pure network blips (react-query's `refetchOnReconnect` auto-recovers those). Does **not** recover for failures that don't fire a reconnect event (5xx, auth-adjacent 4xx on this specific endpoint).

## Steps to Reproduce

1. Create a storage-mode form ("One respondent" option).
2. Navigate to that form → Settings → Webhooks. Confirm the Endpoint URL input renders normally.
3. Open DevTools → Network. Right-click the `GET /api/v3/admin/forms/<formId>/settings` request → **Block request URL** (block *only* this one request — not Offline, and not "Block request domain"; both also block the app's own HTML/JS, so you just get the browser's offline page and the app never boots).
4. **Reload the whole page (F5).** The network stays online; only that request is blocked, so the app loads normally and only the settings fetch fails. A full page reload is required — react-query keeps the cached settings on in-app tab navigation, so the error only surfaces on a fresh load with an empty cache. ("Block request URL" matches the exact URL incl. `<formId>`, so re-test on the same form, or use a `*/forms/*/settings` pattern in the Network request blocking panel.)
5. Observed result: page shows **"Webhooks are only available in storage mode"** with the storage-mode upsell illustration — even though this form *is* in storage mode.

## Expected

A distinct error state — e.g. "Couldn't load settings, please try again" with a retry affordance — clearly separated from the legitimate "this form's response mode does not support webhooks" message.

## Actual

The unsupported-mode message renders on a form that does in fact support webhooks. There is no indication that anything failed. The admin reads it as a definitive statement about their form.

## Evidence

`apps/frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx` (around lines 15–49): `useAdminFormSettings()` is destructured for `data` and `isLoading` only; `isError` is ignored. When the query errors, `isLoading` becomes false, `settings` stays `undefined`, `enableWebhooks` collapses to `false`, and the unsupported branch renders.

## Scope

- **Affected surfaces:** Web admin app only (`SettingsWebhooksPage`).
- **Affected accounts:** All form admins on storage-mode and (with `enableMrfWebhooks`) MRF forms.
- **Possibly elsewhere:** Other `Settings*Page` components that read from `useAdminFormSettings` may share this pattern — not audited yet.

## Suspected Cause

`SettingsWebhooksPage` treats the React Query result as a binary "loading vs ready" state. The page asks two questions ("is it still loading?" and "is the response mode supported?") but not the third ("did the fetch actually succeed?"). When the fetch fails, "no data" gets reclassified as "wrong mode" because the supported-mode check on `undefined` is falsy.

## Proposed Fix

- Read `isError` (and `error`) from `useAdminFormSettings()` in `SettingsWebhooksPage`.
- Render an explicit error/retry state when `isError` is true, *before* evaluating `enableWebhooks`.
- Only render `WebhooksUnsupportedMsg` once settings have loaded successfully and `responseMode` is confirmed unsupported.
- Add a component test that mocks the hook with `{ isError: true, isLoading: false, data: undefined }` and asserts the unsupported message is **not** shown.
- Sweep other `Settings*Page` files for the same destructuring pattern as a follow-up.

## Questions

- Worth a hotfix, or fine to ship in the next normal release? The bug is degraded UX rather than data loss.
- Is there a shared error-state component already in `admin-form/settings/` that should be reused, or should one be introduced?
- Should the fix extend to the other `Settings*Page` components in the same audit, or stay narrowly scoped to webhooks?

### LINKED ISSUE #9525: [Task] Localise the Webhooks settings section copy (i18n)
## Background
`SettingsWebhooksPage` already pulls its heading from i18n (`features.adminForm.settings.webhooks.title`), but `WebhooksErrorMsg` still hardcodes English. Raised on #9524 by @kevin9foong, split out as a follow-up by @LoneRifle.

Scope is `WebhooksErrorMsg` only: per @LoneRifle the sibling `WebhooksUnsupportedMsg` is now a historical artifact (webhooks are fully supported across storage mode and MRF), so it's intentionally left out.

## What to do
Move the hardcoded copy in `WebhooksErrorMsg.tsx` to the existing webhooks i18n files, matching how the title already works:

- "Couldn't load webhook settings", the body sentence, "Try again" / "Trying again…"

Keys live in `apps/frontend/src/i18n/locales/features/admin-form/settings/webhooks/` (`index.ts` for the type, `en-sg.ts` for the English values). English values only — no ms/ta/zh translation needed (these feature folders ship English only).

## Done when
- [ ] `WebhooksErrorMsg` uses `useTranslation`; no hardcoded user-facing strings left
- [ ] Existing tests/stories still pass

Follow-up to #9512 / #9524.

LoneRifle:
## Problem and Solution

Closes #9512
Closes #9525
Merges #9523

### DIFF
--- a/apps/frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx
+++ b/apps/frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx
@@ -23,8 +23,6 @@ export const SettingsWebhooksPage = (): JSX.Element => {
 
   const enableMrfWebhooks = useFeatureIsOn(featureFlags.enableMrfWebhooks)
 
-  // The settings fetch failed. Show an explicit error/retry state instead of
-  // misreporting the form's response mode as unsupported.
   if (isError) {
     return <WebhooksErrorMsg onRetry={refetch} isRetrying={isRefetching} />
   }
--- a/apps/frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx
+++ b/apps/frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx
@@ -7,18 +7,31 @@ import { FormResponseMode } from 'formsg-shared/types/form'
 
 import { CategoryHeader } from './components/CategoryHeader'
 import { WebhooksSection } from './components/WebhooksSection'
+import { WebhooksErrorMsg } from './components/WebhooksSection/WebhooksErrorMsg'
 import { WebhooksUnsupportedMsg } from './components/WebhooksSection/WebhooksUnsupportedMsg'
 import { useAdminFormSettings } from './queries'
 
 export const SettingsWebhooksPage = (): JSX.Element => {
   const { t } = useTranslation()
-  const { data: settings, isLoading } = useAdminFormSettings()
+  const {
+    data: settings,
+    isLoading,
+    isError,
+    isRefetching,
+    refetch,
+  } = useAdminFormSettings()
 
   const enableMrfWebhooks = useFeatureIsOn(featureFlags.enableMrfWebhooks)
 
+  // The settings fetch failed. Show an explicit error/retry state instead of
+  // misreporting the form's response mode as unsupported.
+  if (isError) {
+    return <WebhooksErrorMsg onRetry={refetch} isRetrying={isRefetching} />
+  }
+
   const enableWebhooks =
     !isLoading &&
-    (settings?.responseMode == FormResponseMode.Encrypt ||
+    (settings?.responseMode === FormResponseMode.Encrypt ||
       (settings?.responseMode === FormResponseMode.Multirespondent &&
         enableMrfWebhooks))
 
--- a/apps/frontend/src/features/admin-form/settings/components/WebhooksSection/WebhooksErrorMsg.tsx
+++ b/apps/frontend/src/features/admin-form/settings/components/WebhooksSection/WebhooksErrorMsg.tsx
@@ -24,7 +24,7 @@ export const WebhooksErrorMsg = ({
         <Button
           isLoading={isRetrying}
           loadingText="Trying again…"
-          onClick={onRetry}
+          onClick={() => void onRetry()}
         >
           Try again
         </Button>
--- a/apps/frontend/src/features/admin-form/settings/components/WebhooksSection/WebhooksErrorMsg.tsx
+++ b/apps/frontend/src/features/admin-form/settings/components/WebhooksSection/WebhooksErrorMsg.tsx
@@ -1,3 +1,4 @@
+import { useTranslation } from 'react-i18next'
 import { Flex, Text } from '@chakra-ui/react'
 
 import Button from '~components/Button'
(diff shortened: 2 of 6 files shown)


apps/frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx
```diff
@@ -27,9 +34,15 @@ export const SettingsWebhooksPage = (): JSX.Element => {
 
   const enableMrfWebhooks = useFeatureIsOn(featureFlags.enableMrfWebhooks)
 
+  // The settings fetch failed. Show an explicit error/retry state instead of
```
kevin9foong: nit: should we remove these comments, since the code already explains what is happening

apps/frontend/src/features/admin-form/settings/components/WebhooksSection/WebhooksErrorMsg.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export const WebhooksErrorMsg = ({
+  onRetry,
+  isRetrying = false,
+}: WebhooksErrorMsgProps): JSX.Element => {
+  return (
+    <Flex justify="center" flexDir="column" textAlign="center" role="alert">
+      <Text textStyle="h2" as="h2" color="primary.500" mb="1rem">
+        Couldn't load webhook settings
```
kevin9foong: question: shall we also support i18n for this?

LoneRifle: Done as 3881d1f

apps/frontend/src/features/admin-form/settings/SettingsWebhooksPage.tsx
```diff
@@ -27,9 +34,15 @@ export const SettingsWebhooksPage = (): JSX.Element => {
...
+  // misreporting the form's response mode as unsupported.
+  if (isError) {
+    return <WebhooksErrorMsg onRetry={refetch} isRetrying={isRefetching} />
+  }
+
   const enableWebhooks =
     !isLoading &&
-    (settings?.responseMode == FormResponseMode.Encrypt ||
+    (settings?.responseMode === FormResponseMode.Encrypt ||
```
kevin9foong, line 32: praise: nice catch!

LoneRifle: Copilot caught that..

apps/frontend/src/features/admin-form/settings/components/WebhooksSection/WebhooksErrorMsg.tsx
kimyungju: Sure, happy to take this on. I'll file a separate issue to track i18n support for the webhooks settings section — covering both `WebhooksErrorMsg` and the sibling `WebhooksUnsupportedMsg`, which is currently hardcoded too — and link it back here. Thanks!

kimyungju: Filed as #9525. I scoped it to cover the sibling `WebhooksUnsupportedMsg` too, since it hardcodes its copy the same way and is part of the same inconsistency. Happy to narrow it to just `WebhooksErrorMsg` if you'd prefer. @LoneRifle

eliotlim (approved):

PR #9527, opened 2026-06-02
fix(sig): match hover state to form theme color

### LINKED ISSUE #9347: Signature field's hover state does not match form theme colour

LoneRifle:
## Problem and Solution
Closes #9347
Merges #9441

### DIFF
--- a/apps/frontend/src/templates/Field/Signature/SignatureField.tsx
+++ b/apps/frontend/src/templates/Field/Signature/SignatureField.tsx
@@ -8,7 +8,7 @@ import {
 import { Box, Flex, FormControl, Stack, Text } from '@chakra-ui/react'
 import getStroke from 'perfect-freehand'
 
-import { SignatureVectorArray } from 'formsg-shared/types'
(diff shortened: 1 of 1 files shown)


scottheng96:
Thanks for working on this @LoneRifle. There are a couple of other areas where the button color is also inconsistently not following the form theme color:
1. Save draft button
2. Address field 'Find Address' button

Since we're tracking this as part of 1 ticket, would it be possible to ask for scope creep and implement these changes to those components as well?

LoneRifle:
@scottheng96 I defer to @rickysurya . Could you create a separate GitHub issue for this so that I can have the changes merged in via external contributions? Alternatively, you can launch a GitHub Copilot agent to work on it

scottheng96:
@LoneRifle @rickysurya Can - we can go ahead with this first then 👍

scottheng96 (approved):

PR #9554, opened 2026-06-05
feat: track form origin (paper-form question) in post-cutover Create flow

kevin9foong:
## Problem

As FormSG moves through the MRF cutover, the team has no visibility into whether a new form is digitising an existing paper process. The creation modals collect nothing about a form's origin today, so the data simply doesn't exist. This is the foundational slice of the Form Origin & Storage Mode Reason PRD (issue #1): it establishes the reusable plumbing — the `formOrigin` field, the create-payload extension, the answer→metadata mapper, and the paper-form screen — that the Duplicate, Use-Template, and Storage-reasons slices build on.

## Solution

After the MRF cutover, the Create modal asks **"Is this currently a paper form?"** (Yes/No) as a step after the title screen and before the secret key. The answer is mapped to a `formOrigin` value (`paper` / `unspecified`) by a pure mapper and rides along in the create request, so a form is born with its origin in a single atomic write. `formOrigin` is added to the form record as an extensible enum and persisted by the backend create path. Forms created before this feature have no value and read back as `unspecified` (no backfill).

Scope is the **Create** modal, **MRF** flow, **post-cutover** only. `goToPaperFormQuestion` lives on the shared wizard provider so the Duplicate/Use-Template wizards still satisfy the context, but their flows are wired in later slices.

**Alternatives considered**

- We kept `formOrigin` **required** on the `FormBase`/`FormDto` type (every persisted form has exactly one origin, guaranteed by the schema default) and made it optional only on the create-input types — rather than making it optional everywhere, which would have pushed `?? unspecified` fallbacks onto every future reader.
- For the Yes/No options we built a small bespoke selector instead of reusing the `Tile` component (it requires an icon) or the design-system `Radio` (it renders a radio circle) — neither matches the iconless two-up cards in the design.
- The paper answer is gated with a `validate` rule read through `useFormState`, not a plain `required` rule. `required` didn't reliably block the custom control, and because `formMethods` is shared via React context the error message wasn't reactive without `useFormState`.

**Screenshots**

| Page | After |
| --- | --- |
| Create modal → paper-form question (post-cutover) | after |

_Net-new screen, so no "before". Layout verified against the design source in the visual gate._

**Breaking Changes**

No - backwards compatible. `formOrigin` is optional input and defaults to `unspecified`; existing forms read back as `unspecified`.

## Tests

**TC1: Paper form → paper origin (post-cutover MRF)**

- [ ] With `mrf-cutover` ON, create a form: enter a title, click Next.
- [ ] On "Is this currently a paper form?", choose **Yes**, click Next step.
- [ ] Form is created; its stored `formOrigin` is `paper`.

**TC2: Not a paper form → unspecified**

- [ ] Repeat TC1 but choose **No**; stored `formOrigin` is `unspecified`.

**TC3: Answer is required**

- [ ] On the paper-form screen, click Next step without choosing an option.
- [ ] The flow does not advance and an inline error is shown.

**TC4: Legacy form reads as unspecified**

- [ ] Open a form created before this change; its `formOrigin` reads as `unspecified` (no migration run).

### DIFF
--- a/apps/backend/src/app/models/__tests__/form.server.model.spec.ts
+++ b/apps/backend/src/app/models/__tests__/form.server.model.spec.ts
@@ -15,6 +15,7 @@ import {
   FormField,
   FormFieldDto,
   FormLogoState,
+  FormOrigin,
   FormPermission,
   FormResponseMode,
   FormStartPage,
@@ -113,6 +114,7 @@ const FORM_DEFAULTS = {
     isRetryEnabled: false,
   },
   status: 'PRIVATE',
+  formOrigin: FormOrigin.Unspecified,
   submissionLimit: null,
   goLinkSuffix: '',
   noSmsLimit: false,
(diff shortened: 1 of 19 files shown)


kevin9foong:
closing as this was for AI demo, likely not the desired end product for tracking forms

PR #9557, opened 2026-06-07
feat(frontend): link "What's new" to Gitbook guide

tutntut:
## Problem

The admin nav bar "What's new" tab opens an in-app drawer listing feature
updates. However, the information is outdated and was last updated in 2024. We want it to instead take admins to our guide where updates are more frequent.

## Solution

"What's new" is now a plain external nav link to
https://go.gov.sg/formsg-guides-whats-new. 

**Screenshots**

image

image

**Breaking Changes**

No - backwards compatible.

## Tests

**TC1: "What's new" links out to the guide (desktop)**

- [ ] Log in as an admin and open the dashboard
- [ ] Click "What's new" in the top nav bar
- [ ] Confirm it opens https://go.gov.sg/formsg-guides-whats-new in a new tab (no in-app drawer appears)

**TC2: "What's new" links out to the guide (mobile)**

- [ ] Open the dashboard at a mobile viewport
- [ ] Tap the rocket icon in the nav bar
- [ ] Confirm it opens the What's New guide in a new tab, with no notification dot shown


apps/frontend/src/app/AdminNavBar/AdminNavBar.tsx
```diff
@@ -89,69 +96,76 @@ const AdminNavBarLink = ({ MobileIcon, href, label }: AdminNavBarLinkProps) => {
...
-      >
-        {WHATS_NEW_LABEL}
-      </Button>
-      {shouldShowNotiifcation && (
-        <Icon as={GoDotFill} color="danger.500" position="absolute" ml="-5px" />
-      )}
-    </Box>
-  )
-}
+// [whats-new drawer] Preserved for future reuse. The "What's new" tab now links
```
tutntut: **Why commented out, not deleted:** product may reuse the in-app What's New drawer in future. The component, disclosure/notification wiring, `<WhatsNewDrawer />` render and now-unused imports are preserved in place (each tagged `[whats-new drawer]`) so the trigger can be re-wired from one spot, and the feature files under `~features/whats-new/` are untouched. We considered deleting and relying on git history, but inline restore notes are faster to act on.

eliotlim: comment: we should use git history for these changes, because it's important to manage context and readability in our code!

if it's removed cleanly in this commit, then we can do `git revert <commit-hash-for9557>` and it all comes back.

Here's the situation that we want to avoid:
1. we comment out the code, so that we can restore it easily later
2. another contribution messes up the comments, so that it no longer works
3. it continues to occupy space in our context window as we review it
4. we uncomment the code later, and it doesn't work because things have moved

tutntut:
done @eliotlim thanks for the feedback!

eliotlim (approved):

apps/frontend/src/app/AdminNavBar/AdminNavBar.tsx
kevin9foong: for understanding: this comment is an example of the /tdd breadcrumb improvement over the original skill

```
# Breadcrumb Convention

During implementation, the agent drops **breadcrumbs** — small structured decision notes — into `.scratch/<feature>/decisions.md` in this repo. `prepare-for-review` reads them at PR time and surfaces each one either in the PR body (as a bullet under "Alternatives considered") or as an inline review comment anchored to a specific line.

Breadcrumbs exist so that rationale is captured at the moment of choice, not reconstructed after the fact. ADRs cover hard-to-reverse architectural decisions; breadcrumbs cover the smaller in-impl judgment calls where an ADR would be overkill but a reviewer would still benefit from knowing *why this and not that*.
```

apps/frontend/src/app/AdminNavBar/AdminNavBar.tsx
```diff
@@ -332,10 +246,6 @@ export const AdminNavBar = ({ isMenuOpen }: AdminNavBarProps): JSX.Element => {
           </AvatarMenu>
         </HStack>
       </AdminNavBar.Container>
-      <WhatsNewDrawer
```
kevin9foong, line 335: suggestion: since this is no longer in use, lets remove the  `WharsNewDrawer` component file and its associated story entirely as well. 

These files can be retrieived from git commit history if needed.

apps/frontend/src/app/AdminNavBar/AdminNavBar.tsx
```diff
@@ -306,10 +224,6 @@ export const AdminNavBar = ({ isMenuOpen }: AdminNavBarProps): JSX.Element => {
           {navLinks.map((link, index) => (
             <AdminNavBarLink key={index} {...link} />
           ))}
-          <WhatsNewNavBarTab
```
kevin9foong, line 309: suggestion: same as`WhatsNewDrawer`

kevin9foong:
**AI code review — `/review` (Standards · Spec · Architecture · Divergent)**

The changed code itself is clean: the new What's New link correctly reuses the `AdminNavBarLink` pattern (external, `target="_blank"`, mobile icon), `FORM_GUIDE_WHATS_NEW` matches the `links.ts` naming convention, and `AdminNavBar.tsx` has no leftover unused imports.

All findings are one theme: **the feature's consumer was removed cleanly, but its now-orphaned remnants were left behind.** None of these live in the two changed files, so they can't be anchored inline — grouping them here. Confidence scores are the reviewer's calibrated 0–100 estimate that each finding is real.

---

### issue (non-blocking): orphaned `whats-new` drawer subtree — _confidence 88_

**What's happening:** Removing the `<WhatsNewDrawer>` mount from `AdminNavBar.tsx` left `WhatsNewDrawer` referenced **only by its own `WhatsNewDrawer.stories.tsx`** (verified: `git grep WhatsNewDrawer` finds nothing else on this branch). `WhatsNewContent.tsx` is reachable only via the drawer.

**Suggestion:** Delete the dead subtree in this PR — `features/whats-new/WhatsNewDrawer.tsx`, `WhatsNewContent.tsx`, `WhatsNewDrawer.stories.tsx`, and the `i18n/locales/features/whats-new/` locale.

**Keep:** `FeatureUpdateList` / `FEATURE_UPDATE_LIST` — still used by `rollout-announcement` and `user/constants.ts`.

---

### issue (non-blocking): stale Storybook stories for the deleted notification dot — _confidence 80_

**What's happening:** `app/AdminNavBar/AdminNavBar.stories.tsx` still defines `WhatsNewFeatureNotificationShown`, `...NotShown`, and `...MobileNotificationShown`. These exist only to toggle the notification dot this PR deleted, so they now render identically to `Default` / `Mobile` and assert nothing.

**Why it bites later:** They mislead — a reader thinks the dot behaviour is still covered. The `FEATURE_UPDATE_LIST` import becomes dead with them.

**Suggestion:** Remove the three stories (and the now-unused import) alongside the drawer cleanup.


> 🤖 This comment was generated by an AI code review. Please verify before acting on it.

kevin9foong:
There's another finding, but lets skip this in this PR since removing it from the BE is out of scope and we should remove it together with the backend. 

### issue (non-blocking): dead `SeenFlags.LastSeenFeatureUpdateVersion` map entry — _confidence 85_

**What's happening:** The PR removed both the only frontend reader (`getShowFeatureFlagLastSeen(...)`) and writer (`updateLastSeenFlagMutation`) of this flag. Its only remaining frontend reference is the `SeenFlagsMapVersion` entry in `features/user/constants.ts` — nothing reads or writes it now.

**Suggestion:** Drop the dead entry in the frontend. (As for the backend, since the backend still sets it inside an already-`@deprecated` route — it is out of scope, leave it.)

kevin9foong (approved):
great stuff - thanks for cleaning what's new from being coded and using gitbook instead so we have a single source of truth! will help maintenance in the long run! @tutntut

tutntut:
## Review outcome

A `/review` run (Standards · Spec · Architecture · Divergent, each a parallel
sub-agent, findings verified against false positives) returned **0 surviving
findings**. Three findings were raised and dropped during filtering:

- **Spec — implementation contradicts the `decisions.md` breadcrumb.** The
  breadcrumb originally recorded "comment out the drawer, don't delete"; the
  shipped code deletes it. Dropped: the deletion was a deliberate change made on
  reviewer request, and the breadcrumb lives under gitignored `.scratch/`
  (untracked, won't ship). The breadcrumb has since been updated to record the
  delete-and-rely-on-git-history decision.
- **Architecture — `FeatureUpdateList` / `LastSeenFeatureUpdateVersion` left as
  orphaned dead code.** Dropped: premise refuted on verification —
  `FeatureUpdateList` is still live-consumed by `rollout-announcement`, and the
  seen-flag is a shared enum read by the backend; deleting either would break
  the build.
- **Divergent — move the `whatsNew` i18n key into `linkLabel.*` for
  shape-consistency.** Dropped: factually correct but touches a pre-existing key
  the PR never modified (single consumer, no drift risk) — an out-of-scope
  cosmetic nit.

PR #9562, opened 2026-06-08
fix(frontend): show "(optional)" beside Description for input fields

### LINKED ISSUE #9345: Missing (optional) beside several fields' Description
Description for:

- Short answer
- Long answer
- Number
- Decimal
- Email
- Mobile number
- Local number 

Issue:

Image

Correct behavior: 

Image

tutntut:
Closes #9345

## Problem

The "(optional)" indicator was missing beside the **Description** label in the admin field editor (EditFieldDrawer) for 7 field types: 

- Short answer
- Long answer
- Number
- Decimal
- Email
- Mobile number
- Local number 

## Cause

In these 7 fields, the description `FormControl` was marked `isRequired`. The shared `FormLabel.OptionalIndicator` only renders "(optional)" when the control is **not** required (`FormLabel.tsx`), so the flag suppressed it. It also set a stray `aria-required` on the textarea.

The flag was purely a mislabel: `description` is registered with `register('description')` and carries **no validation rule**, so nothing actually required it. Every other field type already omits `isRequired` on this control.

## Fix

Remove `isRequired` from the description `FormControl` in the 7 affected fields, aligning them with the ~13 other field types. The two editors where the description *is* genuinely required content — `EditParagraph` and `EditImage` (both use `requiredValidationRule`) — were intentionally left untouched.

No validation logic changes; the field's separate `required` toggle (for respondents) is unaffected.

## Before and after

Issue:

Image

Correct behavior: 

Image

### DIFF
--- a/apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDecimal/EditDecimal.tsx
+++ b/apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditDecimal/EditDecimal.tsx
@@ -107,11 +107,7 @@ export const EditDecimal = ({ field }: EditDecimalProps): JSX.Element => {
         <Input autoFocus {...register('title', requiredValidationRule)} />
         <FormErrorMessage>{errors?.title?.message}</FormErrorMessage>
       </FormControl>
-      <FormControl
-        isRequired
-        isReadOnly={isLoading}
-        isInvalid={!!errors.description}
-      >
+      <FormControl isReadOnly={isLoading} isInvalid={!!errors.description}>
         <FormLabel>
           {t(
             'features.adminForm.sidebar.fields.commonFieldComponents.description',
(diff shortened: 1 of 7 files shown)


tutntut:
# Code Review — fix/optional-label-field-description

Multi-axis review of the diff vs `develop`. Each axis ran as an independent sub-agent.

**Summary:** Standards 0 findings · Spec n/a · Architecture 1 (non-blocking, deferred) · Divergent 0. Worst issue: pre-existing boilerplate duplication, explicitly out of scope. Nothing blocking; nothing dropped below threshold. The diff is the correct minimal fix.

---

## Standards

No findings. After the change, all 7 description `FormControl`s are byte-identical to the 13 other input-type drawers. The removed `isRequired` was inconsistent with behavior — `register('description')` carries no validation rule, so the implied constraint was never enforced. The change moves code toward the established convention.

*(0 findings)*

---

## Spec



Skipped — no originating issue/PRD/spec was linked to the branch at review time (the fix maps to issue #9345, "Missing (optional) beside several fields' Description").

*(n/a)*

---

## Architecture

One **non-blocking, out-of-scope** finding (confidence 90): the title+description `FormControl` boilerplate is copy-pasted across ~21 editors with no shared component — which is exactly why this one-boolean bug required editing 7 files by hand, and why `EditImage` expresses "required" a third way (on `FormLabel`, not `FormControl`). A deeper fix would extract a `FieldDescriptionControl` taking `{ register, errors, isLoading, isRequired? }`. Correctly flagged as a follow-up, **not** something to fold into this targeted bugfix.

*(1 finding, non-blocking)*

---

## Divergent

No findings. The chosen approach (deleting the prop) is optimal vs. the alternatives: `isRequired={false}` on `FormLabel` adds a prop where deletion suffices and diverges from the pattern; the shared-component refactor is scope creep. The diff makes the outliers conform to the existing 13-editor convention — the textbook signal of the right fix.

*(0 findings)*

kevin9foong (approved):
lgtm

PR #9565, opened 2026-06-08
feat: short answer prefill progressive disclosure

### LINKED ISSUE #9356: Progressive disclosure for Pre-fill feature
Problem:
Pre-fill editing is a more advanced feature. However, short answer fields are likely one of the fields admin work with. A side effect of having these additional options is that the "Create field" is hidden from view. 

Proposed changes:
- "Prevent pre-fill editing" should appear only after "Enable pre-fill" is toggled on. 
- If "Prevent pre-fill editing" is enabled, but "Enable pre-fill" is disabled, pre-fills should not work.

Image

Image

tutntut:
## Problem

Closes #9356.

On the **Short Answer** field editor, both pre-fill controls — "Enable pre-fill" and "Prevent pre-fill editing" — are always shown ("Prevent pre-fill editing" merely greyed out when pre-fill is off). Pre-fill is an advanced feature used by a minority of admins on a very common field, so the always-visible controls clutter the editor and push the "Create field" button out of view.

## Solution

Progressive disclosure: **"Prevent pre-fill editing" is hidden until "Enable pre-fill" is on**. The lock value is **remembered** while pre-fill is off (not reset), so re-enabling pre-fill restores the prior choice.

image

image

## Test cases

**TC1: Lock toggle is hidden until pre-fill is enabled**
- [ ] Add a new Short Answer field and open its editor
- [ ] Confirm "Prevent pre-fill editing" is **not** shown while "Enable pre-fill" is OFF
- [ ] Turn "Enable pre-fill" ON, and confirm the lock toggle appears, OFF and fully usable (not greyed out)

**TC2: Lock choice is remembered when pre-fill is toggled off and on**
- [ ] Enable pre-fill, turn the lock ON, then turn "Enable pre-fill" OFF
- [ ] Confirm the lock toggle (and Field ID box) disappear
- [ ] Turn "Enable pre-fill" ON again, and confirm the lock reappears still ON

**TC3: Saved field reopens correctly, and disabling pre-fill hides the lock** *(regression)*
- [ ] Enable pre-fill + lock ON, Save, then reopen the field
- [ ] Confirm pre-fill is ON and the lock toggle is visible and ON
- [ ] Turn "Enable pre-fill" OFF, and confirm both the lock toggle and Field ID box hide

**TC4: Field saves with pre-fill OFF but lock ON (no validation error)**
- [ ] Enable pre-fill, turn the lock ON, turn pre-fill OFF, then Save
- [ ] Confirm it saves with no "Cannot lock prefill if prefill is not enabled" error
- [ ] Reopen, enable pre-fill, and confirm the lock comes back ON

**TC5: Respondent — lock only takes effect when a pre-fill value is present**
- [ ] On a field with pre-fill ON + lock ON, open the live form with a pre-fill value in the URL → confirm the field is pre-filled and not editable
- [ ] Open it without a pre-fill value (and also test a field saved as pre-fill OFF + lock ON) → confirm the field is a normal editable field

### DIFF
--- a/apps/backend/src/app/models/field/shortTextField.ts
+++ b/apps/backend/src/app/models/field/shortTextField.ts
@@ -22,20 +22,9 @@ const createShortTextFieldSchema = () => {
       default: false,
     },
     lockPrefill: {
-      // locks prefill if it is enabled
       type: Boolean,
       default: false,
       required: false,
-      // Only allow lock prefill if prefill is enabled
-      validate: {
-        validator: function (this: IShortTextFieldSchema) {
-          if (!this.allowPrefill && this.lockPrefill) {
-            return false
-          }
-          return true
-        },
-        message: 'Cannot lock prefill if prefill is not enabled',
-      },
     },
   })
 
--- a/apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditShortText/EditShortText.stories.tsx
+++ b/apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditShortText/EditShortText.stories.tsx
@@ -75,3 +75,12 @@ PrefillWithFieldId.args = {
     _id: 'mock-field-id-allow-copy',
   },
 }
+
+export const PrefillLocked = Template.bind({})
+PrefillLocked.args = {
+  field: {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    ...PrefillWithFieldId.args.field!,
+    lockPrefill: true,
+  },
+}
--- a/apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditShortText/EditShortText.tsx
+++ b/apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditShortText/EditShortText.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useRef } from 'react'
+import { useEffect, useMemo } from 'react'
 import { Controller, RegisterOptions } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
 import {
@@ -6,7 +6,6 @@ import {
   InputGroup,
   InputRightElement,
   SimpleGrid,
-  useMergeRefs,
 } from '@chakra-ui/react'
 import { extend, isEmpty, pick } from 'lodash'
 
@@ -120,26 +119,7 @@ export const EditShortText = ({ field }: EditShortTextProps): JSX.Element => {
     'ValidationOptions.selectedValidation',
   )
 
-  const hasLockPrefillRef = useRef<HTMLInputElement>(null)
-
-  const lockPrefillRegister = useMemo(() => register('lockPrefill'), [register])
-
-  const mergedLockPrefillRef = useMergeRefs(
-    hasLockPrefillRef,
-    lockPrefillRegister.ref,
-  )
-
   const watchAllowPrefill = watch('allowPrefill')
-  const watchLockPrefill = watch('lockPrefill')
-
-  useEffect(() => {
-    // Prefill must be enabled for lockPrefill
-    // We cannot simply use setValue as it does not update
-    // the UI
-    if (!watchAllowPrefill && watchLockPrefill) {
-      hasLockPrefillRef.current?.click()
-    }
-  }, [watchAllowPrefill, watchLockPrefill])
 
   const customValValidationOptions: RegisterOptions<
     EditShortTextInputs,
@@ -272,15 +252,23 @@ export const EditShortText = ({ field }: EditShortTextProps): JSX.Element => {
           </>
         ) : null}
       </FormControl>
-      <FormControl isReadOnly={isLoading}>
-        <Toggle
-          {...lockPrefillRegister}
-          ref={mergedLockPrefillRef}
-          label="Prevent pre-fill editing"
-          description="This prevents respondents from clicking the field to edit it. However, field content can still be modified via the URL."
-          isDisabled={!watchAllowPrefill}
-        />
-      </FormControl>
+      {watchAllowPrefill ? (
+        <FormControl isReadOnly={isLoading}>
+          <Controller
+            name="lockPrefill"
+            control={control}
+            render={({ field: { value, onChange, ...rest } }) => (
+              <Toggle
+                {...rest}
+                isChecked={!!value}
+                onChange={onChange}
+                label="Prevent pre-fill editing"
+                description="This prevents respondents from clicking the field to edit it. However, field content can still be modified via the URL."
+              />
+            )}
+          />
+        </FormControl>
+      ) : null}
       <FormFieldDrawerActions
         isLoading={isLoading}
         buttonText={buttonText}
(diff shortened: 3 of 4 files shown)


tutntut:
## Multi-axis review — Short Answer pre-fill lock progressive disclosure

**Scope:** commit `0d2e16523` (4 files) vs `a0fc653e8`
**Spec:** `docs/prd/progressive-disclosure-prefills.md` (Closes #9356)
**Result:** ✅ Clean — no findings survived verification.

Reviewed on four independent axes (Standards, Spec, Architecture, Divergent). Six raised concerns were all dropped under false-positive verification — each turned out to be the *deliberate, spec-backed* design rather than a defect.

### What changed
- **Backend** (`shortTextField.ts`): removed the mongoose validator that rejected `lockPrefill: true` while `allowPrefill: false`; its test was inverted to assert the lock now **persists** independently.
- **Frontend** (`EditShortText.tsx`): the "Prevent pre-fill editing" toggle is now rendered only when pre-fill is enabled (progressive disclosure), driven via a `Controller` so it paints in its remembered state on mount. Removed the old `useRef` + `useMergeRefs` + `useEffect`/`.click()` reset hack.
- **Storybook**: added a `PrefillLocked` story (pre-fill on + lock on).

### Axis results
| Axis | Outcome |
|------|---------|
| **Spec** | Faithful to the PRD — every "Done when" item verified (hide-until-needed, ready-to-use, lock remembered, save/reopen round-trip, backend constraint removed + test inverted). No findings. |
| **Standards** | 2 raised, both verified **false positive (score 0)**. |
| **Architecture** | 2 raised, both dropped. |
| **Divergent** | 2 raised, both dropped; 1 self-confirmed (no action). |

### Design decisions validated (pre-empting reviewer questions)
These are the choices the review probed hardest and confirmed correct:

1. **Why remove the backend validator instead of normalising on save?**
   The PRD requires the lock choice to be *remembered* when pre-fill is toggled off and restored when it's turned back on (mirroring the Singpass "Collect NRIC/FIN/UENs" persistence). A normalise-/reset-on-save approach would destroy that remembered value and defeat the feature. The persisted `{allowPrefill: false, lockPrefill: true}` state is intentional and inert — the lock only takes effect respondent-side when a pre-fill value is present (`getFieldPrefillMap` gates on `allowPrefill`).

2. **Why a `Controller` for the lock toggle when siblings use `{...register()}`?**
   The toggle is conditionally mounted, so it remounts mid-session; an uncontrolled `register` checkbox can paint stale state on remount (the bug the old `.click()` hack worked around). `Controller` with `isChecked={!!value}` keeps the painted state in sync with form state. The sibling "Enable pre-fill" toggle stays on `register` because it's always mounted — the asymmetry is justified, not accidental.

kevin9foong:
> Why remove the backend validator instead of normalising on save?
The PRD requires the lock choice to be remembered when pre-fill is toggled off and restored when it's turned back on (mirroring the Singpass "Collect NRIC/FIN/UENs" persistence). A normalise-/reset-on-save approach would destroy that remembered value and defeat the feature. The persisted {allowPrefill: false, lockPrefill: true} state is intentional and inert — the lock only takes effect respondent-side when a pre-fill value is present (getFieldPrefillMap gates on allowPrefill).

This mirrors how the nric settings applies the settings, so the past settings are resumed once the prefill is enabled again. Since the prefill lock only applies when there is a prefill, no issues here

apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditShortText/EditShortText.tsx
```diff
@@ -272,15 +252,23 @@ export const EditShortText = ({ field }: EditShortTextProps): JSX.Element => {
...
-          {...lockPrefillRegister}
-          ref={mergedLockPrefillRef}
-          label="Prevent pre-fill editing"
-          description="This prevents respondents from clicking the field to edit it. However, field content can still be modified via the URL."
-          isDisabled={!watchAllowPrefill}
-        />
-      </FormControl>
+      {watchAllowPrefill ? (
+        <FormControl isReadOnly={isLoading}>
+          <Controller
```
kevin9foong, line 257: praise: good practice to use a controller here.

from the react hook form docs, this is an ideal place to use a controller 
> it's hard to avoid working with external controlled component such as [React-Select](https://github.com/JedWatson/react-select), [AntD](https://github.com/ant-design/ant-design) and [MUI](https://mui.com/). This wrapper component will make it easier for you to work with them.

apps/frontend/src/features/admin-form/create/builder-and-design/BuilderAndDesignDrawer/EditFieldDrawer/edit-fieldtype/EditShortText/EditShortText.stories.tsx
```diff
@@ -75,3 +75,12 @@ PrefillWithFieldId.args = {
     _id: 'mock-field-id-allow-copy',
   },
 }
+
+export const PrefillLocked = Template.bind({})
+PrefillLocked.args = {
+  field: {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
kevin9foong, line 82: note: ok to do this since it is an existing pattern in our tests

apps/backend/src/app/models/field/shortTextField.ts
```diff
@@ -22,20 +22,9 @@ const createShortTextFieldSchema = () => {
       default: false,
     },
     lockPrefill: {
-      // locks prefill if it is enabled
       type: Boolean,
       default: false,
       required: false,
-      // Only allow lock prefill if prefill is enabled
-      validate: {
```
kevin9foong, line 30: verification: verified it is safe to remove this check, since we are only prefilling from the FE in the public form provider. 

```
export const getFieldPrefillMap = (
  formFields: FormFieldDto[],
  searchParams: URLSearchParams,
) => {
  // Return object containing field id and query param value only if id exists in form fields.
  return formFields.reduce((acc, field) => {
    if (
      field.fieldType === BasicField.ShortText &&
      field.allowPrefill &&
      searchParams.has(field._id)
    ) {
      acc[field._id] = {
        prefillValue: searchParams.get(field._id) ?? '',
        lockPrefill: field.lockPrefill ?? false,
      }
    }
    return acc
  }, {} as PrefillMap)
}
```

this uses allowPrefill as a gate to set the prefillValue, which is then used together with lockPrefill subsequently.

kevin9foong: There is a tradeoff here: 

either we prioritize: 
- smaller likelihood of bugs => enforce lockPrefill is false when allowPrefill is false, or mitigate it using an explicit `allowPrefill` check where lockPrefill is used.    
- product consistency with NRIC single sub behavior

i think the tradeoff is in favor of `product consistency with NRIC single sub behavior`, since it makes sense to check prefillValue before locking the field, which is already done.

kevin9foong (approved):

kevin9foong:
note: the lock toggle disappears quite abruptly when the allow prefill toggle is disabled. might not be the smoothest animation 

wonder if we can fix this

tutntut:
@kevin9foong Another weird behavior that could be picked up later. For Singpass settings, there isn't an animation for the hidden toggles that have already been saved as true.

Singpass behavior:

https://github.com/user-attachments/assets/9a98d424-42a8-478c-9561-2ddd56b56408


Email confirmation + new progressive disclosure:

https://github.com/user-attachments/assets/d3927743-f3fb-4fd6-9109-f63333c01cf6

Decided not to fix for this.

tutntut:
> note: the lock toggle disappears quite abruptly when the allow prefill toggle is disabled. might not be the smoothest animation
> 
> wonder if we can fix this

The same behavior appears in email confirmation + include a pdf copy too. So i think i should fix both in a separate pr

Left it as out of scope. Will update the PRD and created an issue (#9578)

